PR #3942, opened 2021-04-27
Use the new utility functions to create temporary directories / files in tests

sschuberth:

### DIFF
--- a/test-utils/src/main/kotlin/Extensions.kt
+++ b/test-utils/src/main/kotlin/Extensions.kt
@@ -47,7 +47,7 @@ fun FileArchiver.Companion.createDefault(): FileArchiver =
         storage = LocalFileStorage(DEFAULT_ARCHIVE_DIR)
     )
 
-fun TestConfiguration.createTestTempDir(vararg infixes: String): File {
+fun TestConfiguration.createSpecTempDir(vararg infixes: String): File {
     val dir = createOrtTempDir(*infixes)
 
     afterSpec {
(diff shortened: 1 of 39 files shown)


oheger-bosch (approved):

oheger-bosch (approved):

test-utils/src/main/kotlin/Extensions.kt
```diff
@@ -66,3 +66,23 @@ fun TestConfiguration.createSpecTempFile(prefix: String? = null, suffix: String?
...
+    val dir = createOrtTempDir(*infixes)
+
+    afterTest {
+        dir.safeDeleteRecursively(force = true)
+    }
+
+    return dir
+}
+
+fun TestConfiguration.createTestTempFile(prefix: String? = null, suffix: String? = null): File {
```
fviernau, line 80: Why did you decide to have separate functions for 'spec' / 'test'? 
I'm asking because it would be possible to do the cleanup always consistently at the later point in time of the two, ´afterSpec` or `afterTest`. (I don't recall right now which that is)

If that's an optimization, do you think it's worth the added complexity?

fviernau (changes requested):

test-utils/src/main/kotlin/Extensions.kt
sschuberth: > Why did you decide to have separate functions for 'spec' / 'test'?

I didn't want to change the functionality compared to before at the same time when doing this refactoring to ease the review.

> it would be possible to do the cleanup always consistently at the later point in time

Not necessarily. Individual test might write to the same file in the temporary directory. and the second test might not expect the file to already be present from the first test.

> If that's an optimization, do you think it's worth the added complexity?

It's not a conscious optimization that I did, I just kept the existing behavior. In that sense I also did not *add* complexity.

fviernau: > Individual test might write to the same file in the temporary directory.

Convinced, missed that point.

fviernau (approved):

sschuberth:

Issue #3503, opened 2021-01-13
Enhance the evaluator to be able to create rules for the Advisor result


MarcelBochtler:
Add an `AdvisorRule` to be able to evaluate results from the Advisor using the evaluator.

neubs-bsi:
Suggested implementation by Test Driven Development:
Write the dsl rule the way you would want the syntax to compile and develop the `AdvisorRule` based on that example

mnonnenmacher:
@MarcelBochtler  FYI, I'm working on an update to the rules API that will make the rules more readable and will also make it easier to add new rule types, so I'll assign this issue to me.

sschuberth:
@mnonnenmacher we need this feature urgently. Does it make sense for us to implement this *now*, an you migrate the syntax later to your new DSL?

mnonnenmacher:
Yes.

PR #3953, opened 2021-04-28
TemplateHelper: Generate human-readable string for vulnerability scores

bs-ondem:
Add a helper method to generate human-readable string based on the
given score and scoring system of a vulnerability. This method can
be used in the template files in order to provide further information
about the vulnerabilities in the report.

### DIFF
--- a/model/src/main/kotlin/Vulnerability.kt
+++ b/model/src/main/kotlin/Vulnerability.kt
@@ -51,58 +51,6 @@ data class Vulnerability(
             "A Vulnerability must have at least one reference."
         }
     }
-
-    /**
-     * The rating attaches human-readable semantics to the score number according to CVSS version 2, see
-     * https://www.balbix.com/insights/cvss-v2-vs-cvss-v3/#CVSSv3-Scoring-Scale-vs-CVSSv2-6.
-     */
-    enum class Cvss2Rating(private val upperBound: Float) {
-        LOW(4.0f),
-        MEDIUM(7.0f),
-        HIGH(10.0f);
-
-        companion object {
-            /**
-             * Get the [Cvss2Rating] from a [score], or null if the [score] does not map to any [Cvss2Rating].
-             */
-            fun fromScore(score: Float): Cvss2Rating? =
-                when {
-                    score < 0.0f || score > HIGH.upperBound -> null
-                    score < LOW.upperBound -> LOW
-                    score < MEDIUM.upperBound -> MEDIUM
-                    score <= HIGH.upperBound -> HIGH
-                    else -> null
-                }
-        }
-    }
-
-    /**
-     * The rating attaches human-readable semantics to the score number according to CVSS version 3, see
-     * https://www.first.org/cvss/v3.0/specification-document#Qualitative-Severity-Rating-Scale.
-     */
-    enum class Cvss3Rating(private val upperBound: Float) {
-        NONE(0.0f),
-        LOW(4.0f),
-        MEDIUM(7.0f),
-        HIGH(9.0f),
-        CRITICAL(10.0f);
-
-        companion object {
-            /**
-             * Get the [Cvss3Rating] from a [score], or null if the [score] does not map to any [Cvss3Rating].
-             */
-            fun fromScore(score: Float): Cvss3Rating? =
-                when {
-                    score < 0.0f || score > CRITICAL.upperBound -> null
-                    score == NONE.upperBound -> NONE
-                    score < LOW.upperBound -> LOW
-                    score < MEDIUM.upperBound -> MEDIUM
-                    score < HIGH.upperBound -> HIGH
-                    score <= CRITICAL.upperBound -> CRITICAL
-                    else -> null
-                }
-        }
-    }
 }
 
 /**
--- a/model/src/main/kotlin/VulnerabilityReference.kt
+++ b/model/src/main/kotlin/VulnerabilityReference.kt
@@ -49,4 +49,56 @@ data class VulnerabilityReference(
      * severity.
      */
     val severity: String?
-)
+) {
+    /**
+     * The rating attaches human-readable semantics to the score number according to CVSS version 2, see
+     * https://www.balbix.com/insights/cvss-v2-vs-cvss-v3/#CVSSv3-Scoring-Scale-vs-CVSSv2-6.
+     */
+    enum class Cvss2Rating(private val upperBound: Float) {
+        LOW(4.0f),
+        MEDIUM(7.0f),
+        HIGH(10.0f);
+
+        companion object {
+            /**
+             * Get the [Cvss2Rating] from a [score], or null if the [score] does not map to any [Cvss2Rating].
+             */
+            fun fromScore(score: Float): Cvss2Rating? =
+                when {
+                    score < 0.0f || score > HIGH.upperBound -> null
+                    score < LOW.upperBound -> LOW
+                    score < MEDIUM.upperBound -> MEDIUM
+                    score <= HIGH.upperBound -> HIGH
+                    else -> null
+                }
+        }
+    }
+
+    /**
+     * The rating attaches human-readable semantics to the score number according to CVSS version 3, see
+     * https://www.first.org/cvss/v3.0/specification-document#Qualitative-Severity-Rating-Scale.
+     */
+    enum class Cvss3Rating(private val upperBound: Float) {
+        NONE(0.0f),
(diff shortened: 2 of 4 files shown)


reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
```diff
@@ -330,6 +330,17 @@ class FreemarkerTemplateProcessor(
         @Suppress("UNUSED") // This function is used in the templates.
         fun filterForUnresolvedVulnerabilities(vulnerabilities: List<Vulnerability>): List<Vulnerability> =
                 vulnerabilities.filter { resolutionProvider.getVulnerabilityResolutionsFor(it).isEmpty() }
+
```
sschuberth, line 333: Commit message:

> to generate human-readable string

"to generate a human-readable string" or "to generate human-readable strings"

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
```diff
@@ -330,6 +330,17 @@ class FreemarkerTemplateProcessor(
...
+        /**
+         * Return a human-readable string that is determined based on the given score and scoring system.
+         */
+        @Suppress("UNUSED") // This function is used in the templates.
+        fun getSeverityString(scoreSystem: String?, score: String) =
+            when (scoreSystem) {
+                "CVSS2" -> Vulnerability.Cvss2Rating.fromScore(score.toFloat())?.name?.toLowerCase()
+                "CVSS3" -> Vulnerability.Cvss3Rating.fromScore(score.toFloat())?.name?.toLowerCase()
+                else -> "unknown"
+            }
```
sschuberth, line 343: How about moving this function to a `companion object` inside the `Vulnerability` class instead?

sschuberth: Actually, I believe we should do some more small refactorings first. now that we have the `VulnerabilityReference` class:

1. Commit: Move the `Cvss2Rating` and `Cvss3Rating` *enums* from `Vulnerability` to `VulnerabilityReference`.
2. Commit: Add this function to a `companion object` in `VulnerabilityReference`.

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
```diff
@@ -330,6 +330,17 @@ class FreemarkerTemplateProcessor(
         @Suppress("UNUSED") // This function is used in the templates.
         fun filterForUnresolvedVulnerabilities(vulnerabilities: List<Vulnerability>): List<Vulnerability> =
                 vulnerabilities.filter { resolutionProvider.getVulnerabilityResolutionsFor(it).isEmpty() }
+
+        /**
+         * Return a human-readable string that is determined based on the given score and scoring system.
+         */
+        @Suppress("UNUSED") // This function is used in the templates.
+        fun getSeverityString(scoreSystem: String?, score: String) =
```
sschuberth, line 338: We should use `scoringSystem` instead of `scoreSystem` to align with the variable name in `VulnerabilityReference`.

sschuberth (changes requested):

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
oheger-bosch: It might make sense to align these refactorings with requirements that come out of the work on #3503. Rule authors will need mechanisms to handle different scoring systems as well.

model/src/main/kotlin/Vulnerability.kt
```diff
@@ -51,58 +51,6 @@ data class Vulnerability(
             "A Vulnerability must have at least one reference."
         }
     }
-
```
sschuberth, line 54: The commit message lacks a rationale. The rationale is:

As the severity is now stored in VulnerabilityReference, it also makes more sense to move the enums that map severity scores to ratings to that class.

model/src/main/kotlin/VulnerabilityReference.kt
```diff
@@ -50,6 +50,18 @@ data class VulnerabilityReference(
      */
     val severity: String?
 ) {
+    companion object {
+        /**
+         * Return a human-readable string that is determined based on the given score and scoring system.
+         */
+        @Suppress("UNUSED") // This function is used in the templates.
```
sschuberth, line 57: Is this still needed?

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
```diff
@@ -120,7 +121,8 @@ class FreemarkerTemplateProcessor(
             "licenseTextProvider" to input.licenseTextProvider,
             "LicenseView" to LicenseView,
             "helper" to TemplateHelper(input.ortResult, input.licenseClassifications, input.resolutionProvider),
-            "projectsAsPackages" to projectsAsPackages
+            "projectsAsPackages" to projectsAsPackages,
+            "vulnerabilityReference" to VulnerabilityReference
```
sschuberth, line 125: I'd prefer the changes to this file to be done only when the templates actually start using it.

sschuberth (changes requested):

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
bs-ondem: So custom templates depends on the default template?

sschuberth: Ah, you mean that we should *provide* the feature even if it's not (yet) *used* in the default template. Yes, I'd agree to that. But then again please amend you existing commit message with something like "Also make the method available to all templates by default."

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
MarcelBochtler: Pushed my first draft of #3503 to this PR: https://github.com/oss-review-toolkit/ort/pull/3961

I designed the rules API in a way that the actual value from the `VulnerabilityReference` is being used and needs to be provided by the rules author.
So I don't think that this will be affected by this PR.

model/src/test/kotlin/VulnerabilityTest.kt
```diff
@@ -27,43 +27,40 @@ import io.kotest.matchers.shouldBe
 
 import java.lang.IllegalArgumentException
 
-import org.ossreviewtoolkit.model.Vulnerability.Cvss2Rating
-import org.ossreviewtoolkit.model.Vulnerability.Cvss3Rating
-
 class VulnerabilityTest : StringSpec({
     "The CVSS 2 rating should be correct for a given score" {
-        Cvss2Rating.fromScore(-0.1f) should beNull()
+        VulnerabilityReference.Cvss2Rating.fromScore(-0.1f) should beNull()
```
sschuberth, line 32: Please just change the imports so all code can stay the same, for a smaller diff.

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
sschuberth: > The method is available to all templates by default, because it's already added to the tempalte data model.

I'm perfectly fine with not using my suggested wording, but this sentence read odd to me (besides the typo in "tempalte"). How about: "The method is made available to all templates by default by adding to the template data model."

sschuberth (changes requested):

model/src/main/kotlin/VulnerabilityReference.kt
```diff
@@ -50,6 +50,17 @@ data class VulnerabilityReference(
...
+        /**
+         * Return a human-readable string that is determined based on the given score and scoring system.
+         */
+        fun getSeverityString(scoringSystem: String?, severity: String) =
+            when (scoringSystem) {
+                "CVSS2" -> Cvss2Rating.fromScore(severity.toFloat())?.name?.toLowerCase()
+                "CVSS3" -> Cvss3Rating.fromScore(severity.toFloat())?.name?.toLowerCase()
+                else -> "unknown"
+            }
+    }
```
sschuberth, line 63: Sorry, missed that before: Missing blank line after this one.

sschuberth (approved):

sschuberth:

PR #3954, opened 2021-04-28
Gradle: Upgrade the SW360 client to version 13.2.0-28

bs-ondem:


sschuberth:
Note: We usually link to the release notes for dependency upgrades in the commit message. Here, there are no release notes, but we happen to know exactly what fix we're after by doing this upgrade. And that should be described in the commit message, please.

oheger-bosch (approved):

sschuberth:

PR #3955, opened 2021-04-28
scanner: Align the return code of the ScannerCommand

nnobelis:
Following the discussion on Slack, this commit aligns the return codes of `ScannerCommand` with the recent changes made to the Evaluator (see https://github.com/oss-review-toolkit/ort/pull/3767).

Only return code '2' if there are some issues with severity higher than 'HINT'.

PR #3956, opened 2021-04-28
Severity: Swap ordering of enum members

sschuberth:
This way a more natural comparison like "severity > Severity.HINT" can
be done.

### DIFF
--- a/model/src/main/kotlin/Severity.kt
+++ b/model/src/main/kotlin/Severity.kt
@@ -22,13 +22,13 @@ package org.ossreviewtoolkit.model
 import org.apache.logging.log4j.Level
 
 /**
- * A generic class describing a severity, e.g. of issues.
+ * A generic class describing a severity, e.g. of issues, sorted from least severe to most severe.
  */
 enum class Severity {
     /**
-     * An error is something that has to be addressed.
+     * A hint is something that is provided for information only.
      */
-    ERROR,
+    HINT,
 
     /**
      * A warning is something that should be addressed.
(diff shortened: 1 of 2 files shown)


nnobelis (dismissed):

nnobelis:
when this is merged, I will adapt https://github.com/oss-review-toolkit/ort/pull/3955

oheger-bosch (commented):
While I do not object this change, I find the approach to rely on the order of enum constants a bit brittle - especially as everything else in the codebase is alphabetically ordered. I had a look into Bloch's "Effective Java", 3rd edition: There are the items 35 "Use instance fields instead of ordinals" and 37 "Use EnumMap instead of ordinal indexing" related to the topic; but both talk about the `ordinal()` function and not directly about the order of enum constants.

A more explicit approach could be to have the `Severity` class provide methods to compare constants or a `Comparator`.

sschuberth:
> I find the approach to rely on the order of enum constants a bit brittle

I agree it's a bit brittle, but that's why I decided to at least add a code comment here.

> especially as everything else in the codebase is alphabetically ordered.

Note, however, that *this* enum wasn't alphabetically sorted before. Also other enums like `HashAlgorithm` are not. In general, we tend to not sort enum members alphabetically, but by meaning.

mnonnenmacher:
I'm a bit undecided about this, I don't really like relying on the order of the enum, but on the other hand you cannot override `compareTo` for enums, so any other solution would require a separate function or comparator which would be less elegant to use. We have the dev meeting in an hour, we could discuss this there?

sschuberth:
> We have the dev meeting in an hour, we could discuss this there?

I've put it onto the agenda.

mnonnenmacher (approved):

fviernau (approved):

PR #3957, opened 2021-04-28
Scanner fixes

sschuberth:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -231,7 +231,6 @@ subprojects {
             jvmTarget = "11"
             apiVersion = "1.4"
             freeCompilerArgs = freeCompilerArgs + customCompilerArgs
-            useIR = true
         }
     }
 
--- a/scanner/src/main/kotlin/LocalScanner.kt
+++ b/scanner/src/main/kotlin/LocalScanner.kt
@@ -227,9 +227,11 @@ abstract class LocalScanner(
 
         val downloadDirectory = createOrtTempDir()
 
-        val resultsFromScanner = remainingPackages.scan(outputDirectory, downloadDirectory)
-
-        downloadDirectory.safeDeleteRecursively(force = true)
+        val resultsFromScanner = try {
+            remainingPackages.scan(outputDirectory, downloadDirectory)
+        } finally {
+            downloadDirectory.safeDeleteRecursively(force = true)
+        }
 
         return resultsFromStorage + resultsFromScanner
     }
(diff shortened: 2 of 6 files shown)


oheger-bosch (approved):

build.gradle.kts
```diff
@@ -231,7 +231,6 @@ subprojects {
             jvmTarget = "11"
             apiVersion = "1.4"
             freeCompilerArgs = freeCompilerArgs + customCompilerArgs
-            useIR = true
```
mnonnenmacher, line 234: Is it worth disabling the compiler for this? I don't know what the performance improvements of the new compiler are, but if it is relevant, isn't there another workaround?
Btw, which commit do you mean here, the next commit is only removing calls to `getSpdxLicenseRef()` so it should not cause this issue?

mnonnenmacher (changes requested):

build.gradle.kts
sschuberth: > Btw, which commit do you mean here, the next commit is only removing calls to `getSpdxLicenseRef()` so it should not cause this issue?

It does. And that's the shocking part. If such a trivial change triggers a compiler bug, my confidence in that new backend becomes shattered.

So yes, that's why I believe it's worth disabling the new backend. Also because I have a hunch that other issues that we started to see recently, like bogus error highlighting issues in the IDE, might also be related to this.

The new compiler backend will become the default with Kotlin 1.5, for which there's already an RC with fixes compared to the new compiler backend version in Kotlin 1.4.32, so I'm hoping that we can "undo" this change implicitly again when we upgrade to Kotlin 1.5.

mnonnenmacher: Ok, I was wondering about it because the linked issue talks about typealiases and it did not seem to be related.

mnonnenmacher (approved):

build.gradle.kts
sschuberth: That's why I had added a [comment](https://youtrack.jetbrains.com/issue/KT-45308#focus=Comments-27-4864409.0-0) to that ticket.

sschuberth:
Merging despite the unrelated test timeout.

build.gradle.kts
sschuberth: For reference, the PR to upgrade to Kotlin 1.5 (which does not have this particular bug anymore) is at https://github.com/oss-review-toolkit/ort/pull/3987.

PR #3959, opened 2021-04-29
WIP postgres: Overhaul configuration

heliocastro:
By default, Postgres define the public schema as available for every
created database. Newly created databases can benefit of this as not
usual create custom schema on test databases.

### DIFF
--- a/model/src/main/kotlin/config/ScanStorageConfiguration.kt
+++ b/model/src/main/kotlin/config/ScanStorageConfiguration.kt
@@ -74,7 +74,7 @@ data class PostgresStorageConfiguration(
     /**
      * The name of the database to use.
      */
-    val schema: String,
+    val schema: String = "public",
 
     /**
      * The username to use for authentication.
--- a/README.md
+++ b/README.md
@@ -492,7 +492,7 @@ ort {
     storages {
       postgresStorage {
         url = "jdbc:postgresql://example.com:5444/database"
-        schema = "schema"
+        schema = "public"
         username = "username"
         password = "password"
         sslmode = "verify-full"
@@ -510,7 +510,9 @@ ort {
 }
 ```
 
-While the specified schema already needs to exist, the _scanner_ will itself create a table called `scan_results` and
+The database needs to exist. If the schema is defined and is not the default postgress _public_, it need to be previously created and grant usage for the specific database user. Left schema entry empty will always default to _public_.
+
+The _scanner_ will itself create a table called `scan_results` and
 store the data in a [jsonb](https://www.postgresql.org/docs/current/datatype-json.html) column.
 
 If you do not want to use SSL set the `sslmode` to `disable`, other possible values are explained in the
--- a/model/src/main/kotlin/utils/DatabaseUtils.kt
+++ b/model/src/main/kotlin/utils/DatabaseUtils.kt
@@ -42,10 +42,6 @@ object DatabaseUtils {
             "URL for PostgreSQL storage is missing."
         }
 
-        require(config.schema.isNotBlank()) {
-            "Schema for PostgreSQL storage is missing."
-        }
-
         require(config.username.isNotBlank()) {
             "Username for PostgreSQL storage is missing."
         }
--- a/model/src/main/resources/reference.conf
+++ b/model/src/main/resources/reference.conf
@@ -54,7 +54,7 @@ ort {
 
       postgresStorage {
         url = "url"
-        schema = "schema"
+        schema = "public"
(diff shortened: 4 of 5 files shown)


model/src/main/kotlin/config/ScanStorageConfiguration.kt
```diff
@@ -72,9 +72,9 @@ data class PostgresStorageConfiguration(
     val url: String,
 
     /**
-     * The name of the database to use.
+     * The schema used by the database. Default is public.
```
sschuberth, line 75: I'd say `Defaults to the "public" schema which exists by default.`

model/src/main/kotlin/config/ScanStorageConfiguration.kt
```diff
@@ -72,9 +72,9 @@ data class PostgresStorageConfiguration(
...
      */
-    val schema: String,
+    val schema: String = "public",
```
sschuberth, line 77: Does this imply the risk to overwrite existing data in that default scheme? Wouldn't it be better to create any non-existing scheme, and set the default to something like "ort" here?

model/src/main/kotlin/config/ScanStorageConfiguration.kt
sschuberth: In the commit message:

> Postgres define the

"defines"

> as not usual create custom schema on test databases.

I do not get this sentence. Is there an "it is" missing after "as"? Also "*a* custom schema". And why "test" database?

model/src/main/kotlin/config/ScanStorageConfiguration.kt
heliocastro: Will not overwrite data in default scheme, but then the idea of create own ort scheme is very good.
Let me think about it

mnonnenmacher: I don't think there's a risk of data loss, ORT is never deleting tables or columns, it only adds missing tables or columns. Scan results are also never overwritten, but IIRC for file archives this might be the case, but I don't see a problem there.
On the other hand it is likely that the role used for adding/reading scan results and file archives does not have permission to create schemas or change schema permissions. And I also think that it's a good idea to use a role without any permissions outside of the schema that should be used by ORT in a production environment.
Therefore I think we should not bother with schema creation, in a testing environment it's fine to use "public" by default, and in a production environment you will want to have more fine grained control over the permission anyway.

mnonnenmacher:
@heliocastro You mentioned on Slack that the schema property is missing in the documentation somewhere, can you add another commit to this PR and fix this as well? This would be useful even if we have a default value for the schema.

scanner/src/main/kotlin/storages/PostgresStorage.kt
```diff
@@ -90,11 +96,20 @@ class PostgresStorage(
             defaultFetchSize(1000)
         }
 
+        var hik = dataSource as HikariDataSource
+        var config = hik.getHikariConfigMXBean() as HikariConfig
+        val schema = Schema(config.schema)
+
         transaction {
-            withDataBaseLock {
+            withDataBaseLock {
```
sschuberth, line 104: Does this line contain trailing whitespace? It looks like so. These should be removed.

model/src/main/kotlin/config/ScanStorageConfiguration.kt
```diff
@@ -67,14 +67,24 @@ data class FileBasedStorageConfiguration(
  */
 data class PostgresStorageConfiguration(
     /**
-     * The database URL in JDBC format.
+     * Hostname or IP of postgres server
```
sschuberth, line 70: In general, please write all docs like you'd write sentences in a printed book. I.e. use a dot at the end of a sentence.

sschuberth (changes requested):

model/src/main/kotlin/config/ScanStorageConfiguration.kt
heliocastro: Changed approach

scanner/src/main/kotlin/storages/PostgresStorage.kt
heliocastro: Fixed lints

sschuberth:
@heliocastro, has this PR been superseded by https://github.com/oss-review-toolkit/ort/pull/3982? If so, can we close this one? Also, why did you not simply update this PR?

sschuberth:
Closing this as #3982 has been merged.

PR #3912, opened 2021-04-20
declared-license-mapping: Remove the entry for "BSD License"

fviernau:
The key cannot be mapped because the value is ambigious. For example,
the string "BSD License" in [1] most likely refers to "BSD-2-Clause"
[2].

[1] https://github.com/testing-cabal/mock/blob/3.0.5/setup.cfg#L8
[2] https://github.com/testing-cabal/mock/blob/3.0.5/LICENSE.txt

Signed-off-by: Frank Viernau <frank.viernau@here.com>

Please ensure that your pull request adheres to our [contribution guidelines](https://github.com/oss-review-toolkit/ort/blob/master/CONTRIBUTING.md). Thank you!

PR #3960, opened 2021-04-29
declared-license-mapping: Add LGPL-2.1 mapping

tsteenbe:
Found "GNU Lesser General Public License v2 (LGPLv2)" in [1].

[1]: https://github.com/PyCQA/astroid/blob/v2.5.6/setup.cfg#L15

Signed-off-by: Thomas Steenbergen <thomas.steenbergen@here.com>

### DIFF
--- a/spdx-utils/src/main/resources/declared-license-mapping.yml
+++ b/spdx-utils/src/main/resources/declared-license-mapping.yml
@@ -252,6 +252,7 @@
 "GNU Lesser General Public License 2.1": LGPL-2.1-only
 "GNU Lesser General Public License Version 2.1": LGPL-2.1-only
 "GNU Lesser General Public License Version 2.1, February 1999": LGPL-2.1-only
+"GNU Lesser General Public License v2 (LGPLv2)": LGPL-2.1-only
 "GNU Lesser General Public License v2 or later (LGPLv2+)": LGPL-2.0-or-later
 "GNU Lesser General Public License v3 (LGPLv3)": LGPL-3.0-only
 "GNU Lesser General Public License v3 or later (LGPLv3+)": LGPL-3.0-or-later


spdx-utils/src/main/resources/declared-license-mapping.yml
```diff
@@ -252,6 +252,7 @@
 "GNU Lesser General Public License 2.1": LGPL-2.1-only
 "GNU Lesser General Public License Version 2.1": LGPL-2.1-only
 "GNU Lesser General Public License Version 2.1, February 1999": LGPL-2.1-only
+"GNU Lesser General Public License v2 (LGPLv2)": LGPL-2.1-only
```
sschuberth, line 255: Shouldn't this be `LGPL-2.0-only` instead?

fviernau: To me it seems this again is an ambiguous case, so cannot be added from correctness point of view?!

sschuberth: Indeed. Only because the [license file](https://github.com/PyCQA/astroid/blob/b964e1f5e04194221b795594d3924b7eb3e8e312/LICENSE#L3) states "Version 2.1" we cannot assume that everybody who declares "License v2" actually means version 2.*1*.

tsteenbe: LGPL-2.0-only is a rare license for OSS projects,  FSF released it in 1991 but quickly superseded with LGPL-2.1-only when it was renamed the GNU Lesser General Public License to reflect the FSF's position that not all libraries should use it. When people say LGPLv2 they mean 99.99% of the time LGPL-2.1-only and not LGPL-2.0-only.

sschuberth: I'd say 99.99% is not enough for our newly stated goal to make the contents of *this* file 100% correct in all cases. And we *need* 100% correctness as the goal was to outsource / share this file with https://github.com/spdx/package-licenses-mapping.

tsteenbe: I don't recall when we agreed upon the 100% correctness target as I see 100% right for each mapping as an impossible target - no compliance tool is 100% right. In my view this is also not a desired balance between the level of automation and an acceptable compliance level desired by most ORT users especially those who are doing OSS compliance at scale. Plus consider the facts that the declared license can be wrong compared to the project's main license.

I would actually like to propose a different approach to mapping ambiguous declared licenses - introduce an option in ORT to let it map automatically all ambiguous cases and then throw an issue if during scanning the main license files or if no main license files are present all source code excluding package metadata files reveals a different license.  

Example 

[PyPI::pytest-cov](https://github.com/pytest-dev/pytest-cov/tree/v2.11.1) declares itself as [MIT](https://github.com/pytest-dev/pytest-cov/blob/v2.11.1/setup.py#L89) and [BSD License](https://github.com/pytest-dev/pytest-cov/blob/v2.11.1/setup.py#L105).  Under my proposal, ORT would map BSD License to BSD-3-Clause but after the scanner is run ORT detects only MIT in the [LICENSE](https://github.com/pytest-dev/pytest-cov/blob/v2.11.1/LICENSE) so then would throw an error/warning issue with message 'Mapped BSD License to BSD-3-Clause for package  PyPI::pytest-cov:2.11.1 but scan results do not support this mapping. Recommend to add a curation to clarify this package licensing.'

By having ORT "guess" the SPDX id for an ambiguous declared license and then "correct itself" when "wrong" would save ORT users a lot of effort of having to manually add declared license mapping for a lot of packages (which is something we are now forced to do thanks to https://github.com/oss-review-toolkit/ort/pull/3912).
I realize that not all ORT users are using `ort scan` but I bet even those users would like to have this guessing option either to save having to create curations or simply not caring whether BSD means BSD-2-Clause, BSD-3-Clause or any of the other 20+ BSDs.  In ORT's export to https://github.com/spdx/package-licenses-mapping we would simply not share all ambiguous mappings or on the SPDX we could add an special label for ambiguous mappings.

sschuberth: > I don't recall when we agreed upon the 100% correctness target 

Meanwhile, see https://github.com/oss-review-toolkit/ort/pull/5127 in that context.

> I see 100% right for each mapping as an impossible target

Well, I believe it *is* possible if we restrict ourselves to the indisputable cases, like mapping "Apache v2.0" to "Apache-2.0" etc.

sschuberth:
Closing this in favor of the approach discussed in the [community meeting on 2022-11-24](https://github.com/oss-review-toolkit/ort/wiki/Community-Meeting#2022-11-24).

PR #3961, opened 2021-04-29
Add RuleMatcher for security vulnerabilities found by the Advisor

MarcelBochtler:


evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -70,6 +70,31 @@ open class PackageRule(
         licenseRules.forEach { it.evaluate() }
     }
 
+    /**
+     * A [RuleMatcher] that checks whether any severity of the [package][pkg] using the specified [scoringSystem] is
+     * higher than [maxValue]. Tested using the [severityComparator]. Returns false if the severity is equal to
+     * [maxValue].
+     */
+    fun containsSeverityOver(maxValue: String, scoringSystem: String, severityComparator: Comparator<String>) =
```
tsteenbe, line 78: @MarcelBochtler  I would rename "containsSeverityOver" to "hasVulnerabilityWithSeverityOver" to align the function name with existing package rulematchers

model/src/main/kotlin/AdvisorRun.kt
```diff
@@ -31,12 +31,12 @@ data class AdvisorRun(
     /**
      * The [Instant] the advisor was started.
      */
-    val startTime: Instant,
+    val startTime: Instant = Instant.EPOCH,
```
sschuberth, line 34: > To align with the other *Run classes.

Are you aware the in the other *Run classes the comment says "The default value exists only for backward compatibility"? So IMO we do not need / want this here.

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -70,6 +70,35 @@ open class PackageRule(
...
 
+    /**
+     * A [RuleMatcher] that checks whether a severity of any vulnerability of the [package][pkg] using the specified
+     * [scoringSystem] is higher than [maxValue]. Tested using the [severityComparator]. Returns false if the severity
+     * is equal to [maxValue].
+     */
+    fun hasVulnerabilityWithSeverityOver(
+        maxValue: String,
+        scoringSystem: String,
+        severityComparator: Comparator<String>
```
sschuberth, line 81: Would it make sense to make these parameters optional and name the function just `hasVulnerability`, so user can also easy check whether there's *any* known vulnerability, without taking the score into account?

sschuberth (changes requested):

cli/src/funTest/assets/semver4j-analyzer-result.yml
```diff
@@ -325,4 +325,38 @@ scanner:
...
+    processors: 4
+    max_memory: 12884901888
+    variables:
+      JAVA_HOME: "/opt/java/openjdk"
+      ANDROID_HOME: "/opt/android-sdk"
+      GOPATH: "/go"
+    tool_versions: {}
+  config:
+    nexus_iq:
+      server_url: "https://ossreviewtoolkit.org"
```
tsteenbe, line 344: The official ORT website URL is https://oss-review-toolkit.org

examples/rules.kts
```diff
@@ -179,6 +179,25 @@ val ruleSet = ruleSet(ortResult, licenseInfoResolver) {
...
+        require {
+            -isExcluded()
+            +hasVulnerabilityWithSeverityOver(maxAcceptedSeverity, scoringSystem) { o1, o2 ->
+                o1.toFloat().compareTo(o2.toFloat())
+            }
+        }
+
+        issue(
+            Severity.ERROR,
+            "Vulnerability with $scoringSystem severity > $maxAcceptedSeverity in package " +
```
tsteenbe, line 195: Broken English - I would make it "The package ${pkg.id.toCoordinates()} has a vulnerability with $scoringSystem severity > $maxAcceptedSeverity" so it aligns with the other rule messages.

evaluator/src/test/kotlin/TestData.kt
```diff
@@ -195,6 +208,44 @@ val ortResult = OrtResult(
...
+        results = AdvisorRecord(
+            advisorResults = sortedMapOf(
+                packageWithVulnerabilities.id to listOf(
+                    AdvisorResult(
+                        vulnerabilities = listOf(
+                            Vulnerability(
+                                id = "CVE-2021-critical",
+                                references = listOf(
+                                    VulnerabilityReference(
+                                        url = URI("https://ossreviewtoolkit.org"),
```
tsteenbe, line 225: Fix ORT URL to https://oss-review-toolkit.org/

tsteenbe (dismissed):

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -70,6 +70,35 @@ open class PackageRule(
         licenseRules.forEach { it.evaluate() }
     }
 
+    /**
+     * A [RuleMatcher] that checks whether a severity of any vulnerability of the [package][pkg] using the specified
+     * [scoringSystem] is higher than [maxValue]. Tested using the [severityComparator]. Returns false if the severity
+     * is equal to [maxValue].
+     */
+    fun hasVulnerabilityWithSeverityOver(
```
oheger-bosch, line 78: This API is rather basic. Is it planned to provide some abstractions over scoring systems? AFAIK, we already have a few classes to represent concrete scoring systems. So the mechanism to query vulnerabilities for a specific scoring system could be aligned to these classes (or vice versa). For instance, the scoring system classes could already provide a `Comparator`.

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -70,6 +70,35 @@ open class PackageRule(
...
+    ) =
+        object : RuleMatcher {
+            override val description =
+                "Higher severity than $maxValue defined by $scoringSystem found for ${pkg.id.toCoordinates()}."
+
+            override fun matches(): Boolean {
+                val advisorResults = ruleSet.ortResult.getAdvisorResults(true)
+                val resultsForPackage = advisorResults[pkg.id]
+
+                return resultsForPackage
```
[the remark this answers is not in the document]
MarcelBochtler: Yes it can be used.
While the merging of the vulnerabilities is not required for this functionality, I agree that `getVulnerabilites()` should be reused here.

evaluator/src/main/kotlin/PackageRule.kt
MarcelBochtler: As spoken on the phone:
I think we require this kind of low level API to support all scoring systems.
But for the known scoring systems (as of now `CVSS2` and `CVSS3`) I think it would be a nice addition to the API and I will create a follow up PR for this.

evaluator/src/main/kotlin/PackageRule.kt
MarcelBochtler: Renamed it again to `hasVulnerability` so it can also be a little bit more generalised.
See also https://github.com/oss-review-toolkit/ort/pull/3961#discussion_r623162217

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -70,6 +70,27 @@ open class PackageRule(
         licenseRules.forEach { it.evaluate() }
     }
 
+    /**
+     * A [RuleMatcher] that checks whether a severity of any vulnerability of the [package][pkg] using the specified
+     * [scoringSystem] is higher than [maxValue]. Checked using the [severityComparator]. Returns false if the severity
+     * is equal to [maxValue].
```
sschuberth, line 91: How about this instead:

A [RuleMatcher] that checks whether any vulnerability for the [package][pkg] exceeds [maxValue] according to the [scoringSystem] and the belonging [severityComparator].

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -70,6 +70,27 @@ open class PackageRule(
...
+     */
+    fun hasVulnerability(maxValue: String, scoringSystem: String, severityComparator: Comparator<String>) =
+        object : RuleMatcher {
+            override val description =
+                "Higher severity than $maxValue defined by $scoringSystem found for ${pkg.id.toCoordinates()}."
```
sschuberth, line 96: Note that the `description` is not actually user-facing, but only used in logs currently. So it might make more sense to use something shorter here; see the descriptions of other rules.

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -70,6 +70,27 @@ open class PackageRule(
         licenseRules.forEach { it.evaluate() }
     }
 
+    /**
+     * A [RuleMatcher] that checks whether a severity of any vulnerability of the [package][pkg] using the specified
+     * [scoringSystem] is higher than [maxValue]. Checked using the [severityComparator]. Returns false if the severity
+     * is equal to [maxValue].
+     */
+    fun hasVulnerability(maxValue: String, scoringSystem: String, severityComparator: Comparator<String>) =
```
sschuberth, line 93: Maybe rename `maxValue` to `threshold`?

evaluator/src/test/kotlin/PackageRuleTest.kt
```diff
@@ -172,5 +172,52 @@ class PackageRuleTest : WordSpec() {
                 }
             }
         }
+
+        "hasVulnerability()" should {
+            "return true if a severity of a vulnerability is higher than the threshold" {
+                val rule = createPackageRule(packageWithVulnerabilities)
+                val matcher = rule.hasVulnerability("8.9", "CVSS3") { o1, o2 ->
```
sschuberth, line 193: Instead of `o1` and `o2` I'd prefer names that make more clear which variable is the value, and which one the threshold. Also in `examples/rules.kts` below.

sschuberth (changes requested):

evaluator/src/main/kotlin/PackageRule.kt
MarcelBochtler: I opted for `maxValue` as this implied that the set value is included.
But as I like `threshold` better and the meaning is described in the documentation, I changed it.

sschuberth: Thinking about this again, as a user I probably want to report a vulnerability as soon as it's e.g. of "high" severity, which according to CVSS2/3 would be starting with a score of 7.0. Which means a user should specify the score value when it should start to break, not the score value that is still acceptable.

So we should reword the docs yet again to

A [RuleMatcher] that checks whether any vulnerability for the [package][pkg] has a score that equals or is greater than [threshold] according to the [scoringSystem] and the belonging [severityComparator].

and `?.any { it > 0 }` needs to be changed to `?.any { it >= 0 }`.

What do you think @MarcelBochtler?

sschuberth (approved):

PR #3964, opened 2021-04-29
Invalidate Azure build caches

sschuberth:
Invalidate the Azure cache for Gradle, as the Gradle build cache seems
to be corrupt due to Kotlin compiler backend changes.


tsteenbe (approved):

sschuberth:
Note, while this doesn't fix all of the problem we see in CI, it's at least a step in the right direction; only Windows-specific issues seem to remain.

fviernau (approved):

PR #3968, opened 2021-04-30
HowToFixTextProvider: Add more packages to the default imports

mnonnenmacher:
Add more default imports to the `HowToFixTextProvider` scripts so that
they can use the same model classes and utility functions as evaluator
scripts.


sschuberth (approved):

mnonnenmacher:
Merged, ignoring the unrelated test errors.

PR #3972, opened 2021-05-03
Netrc parsing fix

sschuberth:

### DIFF
--- a/utils/src/test/kotlin/OrtAuthenticatorTest.kt
+++ b/utils/src/test/kotlin/OrtAuthenticatorTest.kt
@@ -52,7 +52,7 @@ class OrtAuthenticatorTest : WordSpec({
             }
         }
 
-        "recognize the default machine" {
+        "recognize a default entry that comes after all machine entries" {
             val authentication = getNetrcAuthentication("""
                 machine github.com
                 login git
--- a/utils/src/test/kotlin/OrtAuthenticatorTest.kt
+++ b/utils/src/test/kotlin/OrtAuthenticatorTest.kt
@@ -66,18 +66,26 @@ class OrtAuthenticatorTest : WordSpec({
             }
         }
 
-        "prefer machine-specific entries over the default" {
+        "recognize a default entry if it is the only entry" {
+            val authentication = getNetrcAuthentication("""
+                default login foo password bar
+            """.trimIndent(), "gitlab.com")
+
+            authentication shouldNotBeNull {
+                userName shouldBe "foo"
+                password shouldBe "bar".toCharArray()
+            }
+        }
+
+        "disregard a default entry that comes before machine entries" {
             val authentication = getNetrcAuthentication("""
                 default login foo password bar
                 machine github.com
                 login git
                 password hub
-            """.trimIndent(), "github.com")
+            """.trimIndent(), "gitlab.com")
 
-            authentication shouldNotBeNull {
-                userName shouldBe "git"
-                password shouldBe "hub".toCharArray()
-            }
+            authentication should beNull()
         }
 
         "ignore superfluous statements" {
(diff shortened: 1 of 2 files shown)


mnonnenmacher (approved):

utils/src/test/kotlin/OrtAuthenticatorTest.kt
```diff
@@ -66,18 +66,26 @@ class OrtAuthenticatorTest : WordSpec({
...
+        }
+
+        "disregard a default entry that comes before machine entries" {
             val authentication = getNetrcAuthentication("""
                 default login foo password bar
                 machine github.com
                 login git
                 password hub
-            """.trimIndent(), "github.com")
+            """.trimIndent(), "gitlab.com")
```
MarcelBochtler, line 86: Can you add an additional test case that checks if a host can be found if the `.netrc` format is ok?
As this was the original purpose of this test.

MarcelBochtler (changes requested):

utils/src/test/kotlin/OrtAuthenticatorTest.kt
sschuberth: Have you also checked the other existing tests that I left untouched? Can you state exactly what additional case you'd like to see covered?

MarcelBochtler: I checked the other tests now. The case I'd like to see covered is this `netrc`:
```
machine host.com login user password pass
default login foo password bar
```

MarcelBochtler: Your test name is `"disregard a default entry that comes before machine entries"`
But your test checks if `gitlab.com` can be found, while it should actually check for `github.com` to check the if the default entry is disregarded.

MarcelBochtler: My previous comment is invalid, as you're right, there is already a test that covers that scenario.

However:
I'd suggest to add a test case that actually tests the behaviour that is mentioned in the `netrc` documentation:
Something like that:
```kt
"prioritise a default entry if it comes before the actual host" {
    val authentication = getNetrcAuthentication("""
        default login foo password bar
        machine github.com
        login git
        password hub
    """.trimIndent(), "github.com")

    authentication shouldNotBeNull {
        userName shouldBe "foo"
        password shouldBe "bar"
    }
}
```

utils/src/test/kotlin/OrtAuthenticatorTest.kt
```diff
@@ -66,18 +66,26 @@ class OrtAuthenticatorTest : WordSpec({
...
 
-            authentication shouldNotBeNull {
-                userName shouldBe "git"
-                password shouldBe "hub".toCharArray()
-            }
+            authentication should beNull()
```
MarcelBochtler, line 88: This assumption is wrong.
The `.netrc` contains a default entry. It should be used for `gitlab.com`.

MarcelBochtler (changes requested):

utils/src/test/kotlin/OrtAuthenticatorTest.kt
sschuberth: But the default entry is the *first* in the file, so it should not be used, or?

MarcelBochtler: The specification does not mention what should happen if the default entry is **not** placed last.

But Git's parsing of the `.netrc` file uses the `default` entry if one is available before a machine entry.
https://github.com/git/git/blob/7e391989789db82983665667013a46eabc6fc570/contrib/credential/netrc/git-credential-netrc.perl

This can be checked by using such a .netrc file:
```
default login <valid-user> password <valid-password>
machine github.com login <invalid-user> password <invalid-password>
```
and then try to clone or pull from github.

sschuberth:
After an internal discussion it turns out the implementation is still wrong. Please see https://github.com/oss-review-toolkit/ort/pull/3973 instead.

PR #3975, opened 2021-05-03
Cargo: Map the license-file property to an unknown license

mnonnenmacher:
Cargo allows to declare non-SPDX licenses only by referencing a license
file. If the `license-file` property is used, map the value to
`LicenseRef-ort-unknown-license-reference` to indiciate that some
license is declared, instead of ignoring the property. This
implementation is similar to how the "SEE LICENSE IN ..." values are
handled in NPM.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Cargo.kt
+++ b/analyzer/src/main/kotlin/managers/Cargo.kt
@@ -236,10 +236,21 @@ private fun checksumKeyOf(metadata: JsonNode): String {
 
 private fun extractCargoId(node: JsonNode) = node["id"].textValueOrEmpty()
 
-private fun extractDeclaredLicenses(node: JsonNode): SortedSet<String> =
-    node["license"].textValueOrEmpty().split('/')
+private fun extractDeclaredLicenses(node: JsonNode): SortedSet<String> {
+    val declaredLicenses = node["license"].textValueOrEmpty().split('/')
         .map { it.trim() }
         .filterTo(sortedSetOf()) { it.isNotEmpty() }
+
+    // Cargo allows to declare non-SPDX licenses only by referencing a license file. If a license file is specified, add
+    // an unknown declared license to indicate that there is a declared license but we cannot know which it is at this
+    // point.
+    // See: https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields
+    if (node["license_file"].textValueOrEmpty().isNotBlank()) {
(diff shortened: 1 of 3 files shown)


analyzer/src/main/kotlin/managers/Cargo.kt
```diff
@@ -236,11 +236,21 @@ private fun checksumKeyOf(metadata: JsonNode): String {
...
 private fun extractCargoId(node: JsonNode) = node["id"].textValueOrEmpty()
 
-private fun extractDeclaredLicenses(node: JsonNode): SortedSet<String> =
-    node["license"].textValueOrEmpty().split('/')
+private fun extractDeclaredLicenses(node: JsonNode): SortedSet<String> {
+    val declaredLicenses = node["license"].textValueOrEmpty().split('/')
         .map { it.trim() }
         .filterTo(sortedSetOf()) { it.isNotEmpty() }
 
+    // Cargo allows to declare non-SPDX licenses only by referencing a license file. Avoid reporting an [OrtIssue] by
```
sschuberth, line 244: Isn't the comment about avoiding an OrtIssue a bit misleading, as before this change no OrtIssue would have been created, or? It would only be created if whatever `node["license_file"]` was assigned as-is to `declaredLicenses`.

sschuberth: Ah, I see the comment was taken from `Npm.kt`. However, there the case is slightly different, as "SEE LICENSE IN " might be part of the regular field for a declared license. So for Npm, the mapping is necessary to avoid an OrtIssue. But as Cargo has a dedicated field for that, that argumentation does not hold. I'd simply drop that sentence.

mnonnenmacher: I wanted to use the same comment as in `Npm` because it's the same concept.

sschuberth: As I tried to explain, the "Avoid reporting an OrtIssue" part is *not* the same concept here. You're talking about suppressing an issue that would not have been there without your change.

mnonnenmacher: I only saw your first comment when I added my response. I have replaced the sentence with a better explanation.

sschuberth (approved):

mnonnenmacher:
Merged, because the `GitRepoFunTest` errors are unrelated to this change.

PR #3976, opened 2021-05-04
VcsInfo usage improvements and fixes

fviernau:
See individual commits.


analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -172,7 +172,7 @@ class NuGetSupport(serviceIndexUrls: List<String> = listOf(DEFAULT_SERVICE_INDEX
             VcsInfo(
                 type = VcsType(it.type.orEmpty()),
                 url = it.url.orEmpty(),
-                revision = (it.branch ?: it.commit).orEmpty(),
+                revision = (it.commit ?: it.branch).orEmpty(),
```
sschuberth, line 175: IIRC we intentionally (preferably) used `branch` for `revision` and `commit` for `resolvedRevision`, which matches ORT semantics to eventually have a symbolic name in `revision`, but always have a fixed commit in `resolvedRevision`. So I believe we should drop this change.

sschuberth: Hmpf. Having this commit first would have saved me from typing the comment before, as I now get what you're after.

But I actually think the `VcsInfo` docs are overly specific here when saying "The VCS-specific revision resolved during downloading from the VCS". Back then we simply couldn't imaginge any other way to resolve the revision than downloading. I'd rather change the docs here instead.

model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
```diff
@@ -70,7 +70,7 @@ class LicenseInfoResolverTest : WordSpec() {
             type = VcsType.GIT,
             url = "https://github.com/oss-review-toolkit/ort.git",
             revision = "master",
-            resolvedRevision = "master"
+            resolvedRevision = "0000000011111111222222223333333344444444"
```
sschuberth, line 73: Nit: I believe we usually used all zeros, see `git grep resolvedRevision.*0000`.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: This `NugetSupport` is the only place which uses the `resolvedRevision` in a way which is incompatible with the current `KDoc`. The work I did so far for the removal of `Provenance.origninalVcsInfo` relies on the current semantics of the `resolvedRevision` according to the KDoc. Allowing the analyzers to set the resolved revision will ((IIUC) make it impossible to eliminate the mentioned `origninalVcsInfo`. What do you think?

Regarding ,

> IIRC we intentionally (preferably) used branch for revision and commit for resolvedRevision, which matches ORT semantics to eventually have a symbolic name in revision, but always have a fixed commit in resolvedRevision. So I believe we should drop this change.

Why do you think it is better to have a symbolic name in `revision`? The documentation says that it can be (1) branch name (2) tag name (3) sha1, so any of these options is allowed so having (2) or (3) would match ORT's semantics also, not just (1) would.

model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
fviernau: The output of that grep command yields 3 matches for the zero case and 2 matches for the non-zero case.
Can we keep it as is?

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
mnonnenmacher: @sschuberth First a question regarding the current implementation for NuGet, because I don't know how it works. Are the dependencies cloned from a VCS, and is the `commit` property a result of this or is it coming from metadata? Because if it's coming from the metadata we cannot rely on it to be correct. It could not exist, for example, and the downloader might actually clone a different revision by guessing a tag.
Apart from this, we also do not set the resolved revision for projects, where we do indeed know that exact revision, and we also don't set it in `GoMod` where dependencies are cloned using Git. Doing it different in NuGet is inconsistent with that and makes it indeed impossible to remove the `originalVcsInfo` property.
My suggestion would be that we later do another refactoring where we:
* If required or wanted, add another property like `reference` to VcsInfo which can be used for symbolic names like branches or tags.
* Remove the `resolvedRevision` property from `VcsInfo` and instead have it as a separate property in `RepositoryProvenance`, to align the model with the so far predominant interpretation that `VcsInfo` is for metadata, except for the `resolvedRevision` field which is supposed to be only set by the downloader.

sschuberth: > This `NugetSupport` is the only place which uses the `resolvedRevision` in a way which is incompatible with the current `KDoc`.

Yes, but that's only because none of the other package managers has a dedicated metadata field for the resolved revision / commit SHA1.

> The work I did so far for the removal of `Provenance.origninalVcsInfo` relies on the current semantics of the `resolvedRevision` according to the KDoc. Allowing the analyzers to set the resolved revision will ((IIUC) make it impossible to eliminate the mentioned `origninalVcsInfo`.

Can you maybe explain in detail why this is the case? Sorry, I have so many context switches, I don't have all the constraints of the scan-by-repo topic and its preparations in mind.

> Why do you think it is better to have a symbolic name in `revision`?

Assuming that it's fine for the analyzer to populate both `revision` and `resolvedRevision` (which so far nothing was speaking against in my view), then it does not make sense to set `revision` and `resolvedRevision` to the same value, but I do see some value in having both the (unresolved) `revision` and the `resolvedRevision`.

> The documentation says that it can be (1) branch name (2) tag name (3) sha1, so any of these options is allowed so having (2) or (3) would match ORT's semantics also, not just (1) would.

Which docs are you referring to? At https://docs.microsoft.com/en-us/nuget/reference/nuspec#repository I'm reading "the individual branch name and / or commit SHA-1 hash", so `commit` should always be a hash.

sschuberth: > Are the dependencies cloned from a VCS, and is the `commit` property a result of this or is it coming from metadata? 

It's coming from metadata.

> Because if it's coming from the metadata we cannot rely on it to be correct. It could not exist, for example, and the downloader might actually clone a different revision by guessing a tag.

That's correct.

> Apart from this, we also do not set the resolved revision for projects, where we do indeed know that exact revision, and we also don't set it in `GoMod` where dependencies are cloned using Git. Doing it different in NuGet is inconsistent with that and makes it indeed impossible to remove the `originalVcsInfo` property.

Good points about projects and `GoMod`.

>     * If required or wanted, add another property like `reference` to VcsInfo which can be used for symbolic names like branches or tags.

That probably does not work in general, as a resolved revision / commit SHA1 might be part of multiple branches.

>     * Remove the `resolvedRevision` property from `VcsInfo` and instead have it as a separate property in `RepositoryProvenance`, to align the model with the so far predominant interpretation that `VcsInfo` is for metadata, except for the `resolvedRevision` field which is supposed to be only set by the downloader.

I'd agree that if we know decide now that a `VcsInfo`'s `resolvedRevision` may not be set by *any* user of `VcsInfo`, but *only* by the downloader, then `resolvedRevision` should not be part of `VcsInfo` to avoid misuse.

model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
sschuberth: I won't block it if you keep it, but I'll probably do an alignment-commit afterwards.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
mnonnenmacher: > > The work I did so far for the removal of `Provenance.origninalVcsInfo` relies on the current semantics of the `resolvedRevision` according to the KDoc. Allowing the analyzers to set the resolved revision will ((IIUC) make it impossible to eliminate the mentioned `origninalVcsInfo`.
> 
> Can you maybe explain in detail why this is the case? Sorry, I have so many context switches, I don't have all the constraints of the scan-by-repo topic and its preparations in mind.

I'll try to summarize it shortly. The `originalVcsInfo` is required to match a scan result to the VCS metadata of a package, because the provenance of the scan result can be different if the revision was guessed by the downloader (e.g. because there is no revision in the metadata, or the revision is wrong). With the removal of the `originalVcsInfo` this matching logic needs to be changed, and one assumption of the new logic is that the `resolvedRevision` can only be set by the downloader.

After checking the new logic again, I think it might still work without changing the code here, but I think for consistency it's better to still change it. However, I just found another issue with using the branch for `revision` and the SHA1 for `resolvedRevision` here: The downloader never uses `resolvedRevision` as input, that means even though we know the exact revision here it will be ignored and only the branch name will be tried.

I also think the term "resolved revision" can be misleading here, because it was (at least in my understanding, and according to the KDoc) meant to mean the resolution happening in the downloader which includes e.g. checking if the revision exists or guessing a revision based on the version, but it can also be interpreted as resolving a branch or tag name to a SHA1. Therefore I would suggest to apply this commit and, as proposed before, later introduce separate properties in `VcsInfo` for symbolic names and SHA1s. And also make it more clear in documentation and how we use those classes, that `VcsInfo` is used for capturing metadata, and `Provenance` is used for source code origin that are confirmed by a "real" download.

> That probably does not work in general, as a resolved revision / commit SHA1 might be part of multiple branches.

I believe that is not a problem here, because we are just capturing metadata. That means, if the metadata provides a branch name and a revision, it doesn't really matter if other branches contain that commit as well, the only purpose of the symbolic name would be to act as a fallback in case the revision is missing or cannot be downloaded.

@sschuberth @fviernau My answer became a bit longer than intended, if you have any questions I would suggest we have a quick call.

mnonnenmacher: Just one more thought, the fact that changes in this area are so difficult to understand seems to be a good indicator that some refactoring is required.

mnonnenmacher (approved):

fviernau:
@sschuberth  is this good to go from your side?

sschuberth:
> @sschuberth is this good to go from your side?

Please give me some more time to review. I hope to finish the review today. To me, the most important thing is that we get the rationale(s) properly documented in the commit messages. With the changes per se I agree basically.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
sschuberth: > The downloader never uses `resolvedRevision` as input

That's actually a good point, and probably a good rationale why it can be removed / should not be set, as it's of no use.

> but it can also be interpreted as resolving a branch or tag name to a SHA1.

That exactly was my interpretation so far: To resolve a (potentially) symbolic revision to a commit SHA1. So "resolve" in the Git-sense. And in the spirit of capturing as much facts / metadata as we can, I wanted to capture what NuGet believes to be the resolved revision, too.

> later introduce separate properties in `VcsInfo` for symbolic names and SHA1s and remove the `resolvedRevision` property.

Maybe we don't even need the former (as so far really only NuGet provides both in the metadata), and we can only do the removal of `resolvedRevision` to keep things as simple as possible.

sschuberth:
Ok, here's my "wishlist" of nit-fixes:
- [ ] In the first two commits, please write camel-cased "NuGet..." instead of "Nuget...".
- [ ] Please swap the order of the frist two commits. I find it much easier to argue in the commit message why `revision` should be set to a fixed commit when `resolvedRevision` is gone as an alternative.
- [ ] For commit "NuGetSupport: Stop assigning the resolved revision of packages" I'd propose the following body:
```
The resolved revision is not supposed to be assigned by the analyzer at
all, but only by the downloader. That is because what counts it not to what
a `revision` resolved to at the time of analysis, or what package metadata
claims a `revision` resolves to, but what commit of the source code was
really downloaded and scanned. This is especially important for moving
revisions, like branch names, which could have pointed to something
different at the time of analysis than at the time of downloading /
scanning.
```
- [ ] Then for commit "NuGetSupport: Prefer `commit` over `branch` as revision of a package" I propose the following body:
```
Try to be as concrete as possible when populating the `revision` by
preferring a commit SHA1 over a branch name. This also helps the downloader
to eventually avoid finding commit candidates when resolving the revision.
```
- [ ] I'd prefer to drop "VcsHost: Inline a couple of variables" as I find the resulting clustering of parentheses and curly braces hard to read.
- [ ] Commit "LicenseInfoResolverTest: Use a better value for `resolvedRevision`" renames an `archive.zip`, but I guess that's not intended.

fviernau:
Thanks a lot @sschuberth for taking the time and for crafting that wishlist!
 
I've implemented it as you proposed, except for / in addition to  
- I slightly adjusted the commit message of commit: "VcsHost: Remove unnecessary curly braces from several `when` cases"
- For commit "LicenseInfoResolverTest: Use a better value for `resolvedRevision`" renaming the directory was actually intended,    so I adjusted the commit message to say so

sschuberth (approved):

Issue #3727, opened 2021-03-05
Add creation date to scan-report-web-app.html


hanna-modica:
In order to find out easily, while studying a report, when it was created, I would like to have the creation date present in the report.

neubs-bsi:
branch for this ticket: https://github.com/oss-review-toolkit/ort/tree/web-app-reporter-start-time
(PR not created yet)

MarcelBochtler:
Hi @hanna-modica, in https://github.com/oss-review-toolkit/ort/pull/3978 we added the time the analyzer result was created, as the report may be created days later we thought that this might be the more relevant date.
Is this sufficient for your use case?

hanna-modica:
Hi @MarcelBochtler, yes, this is totally enough to keep me from going insane when having opened a lot of scan reports for the same project simultaneously.

sschuberth:
@hanna-modica, if you "only goal" is traceability, isn't the back-linking from reports that we implemented recently anyway the better solution? Is this then even still needed?

But leaving the traceability aside, I do believe that having a timestamp makes sense. However, I also still believe that arbitrarily picking only the analyzer time is not a good idea. The best would be an overview over *all* times and durations, for all steps. And it only *one* time must be used, I'd still vote for the report generation time, as that's then at least not arbitrarily chosen.

Issue #3901, opened 2021-04-16
Miscellaneous web app reporter improvement ideas


sschuberth:
The "About" screen in the web app reporter could use some improvements, like:

- [x] More padding below the logo before the first text paragraph starts. (Done in https://github.com/oss-review-toolkit/ort/pull/4092/commits/e887d57e47bc0ab76e9159e86e8c80820e91eeb6)
- [ ] #8701
- [x] Also, on the "Violations" tab, it would be nice if the "Rule" column (and also others) would be filterable and sortable. (Done in https://github.com/oss-review-toolkit/ort/pull/4092/commits/ebfc2e4f6b75068526122def26689c3eca4c604e)

sschuberth:
Also, on the "Violations" tab, it would be nice if the "Rule" column (and also others) would be filterable and sortable.

sschuberth:
The last item from the original list is now tracked separately as https://github.com/oss-review-toolkit/ort/issues/8701 for visibility.

PR #3978, opened 2021-05-04
WebAppReporter: Add Analyzer start time to about dialog

MarcelBochtler:
Resolves: #3727.

### DIFF
--- a/reporter-web-app/src/components/AboutModal.js
+++ b/reporter-web-app/src/components/AboutModal.js
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2019-2021 HERE Europe B.V.
+ * Copyright (C) 2021 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -46,7 +47,21 @@ SyntaxHighlighter.registerLanguage('yaml', yaml);
 const AboutModal = (props) => {
     const { webAppOrtResult } = props;
     const { repositoryConfiguration } = webAppOrtResult;
-    const { labels } = webAppOrtResult;
+    const {
+        labels,
+        metaData
+    } = webAppOrtResult;
+
+    const { analyzerStartTime } = metaData;
+    const analyzerStartDate = new Date(analyzerStartTime).toLocaleDateString(
+        undefined,
+        {
+            weekday: 'long',
+            year: 'numeric',
+            month: 'long',
+            day: 'numeric'
+        }
+    );
 
     return (
         <Modal
@@ -167,6 +182,17 @@ const AboutModal = (props) => {
                         third-party software components under other open source licenses.
                         See OSS Review Toolkit code repository for further details.
                     </p>
+                    {
+                        !!analyzerStartDate
+                        && (
+                            <p>
+                                This ORT report is based on an analysis started on
+                                {' '}
+                                {analyzerStartDate}
+                                .
+                            </p>
+                        )
+                    }
                 </TabPane>
             </Tabs>
         </Modal>
--- a/reporter-web-app/src/models/MetaData.js
+++ b/reporter-web-app/src/models/MetaData.js
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ * Copyright (C) 2021 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+class MetaData {
+    #analyzerStartTime;
+
+    constructor(obj) {
+        if (obj instanceof Object) {
+            if (obj.analyzer_start_time || obj.analyzerStartTime) {
+                this.#analyzerStartTime = obj.analyzer_start_time || obj.analyzerStartTime;
+            }
+        }
+    }
+
+    get analyzerStartTime() {
+        return this.#analyzerStartTime;
+    }
+}
+
+export default MetaData;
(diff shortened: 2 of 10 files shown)


reporter-web-app/src/components/AboutModal.js
```diff
@@ -46,7 +46,21 @@ SyntaxHighlighter.registerLanguage('yaml', yaml);
 const AboutModal = (props) => {
     const { webAppOrtResult } = props;
     const { repositoryConfiguration } = webAppOrtResult;
-    const { labels } = webAppOrtResult;
+    const {
```
fviernau, line 49: commit message: The sign off for "Marcel Bochtler" is duplicated. Is a sign off for "Stephanie Neubauer" missing as well?

reporter-web-app/src/models/MetaData.js
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+class MetaData {
+    #startTime;
+
+    constructor(obj) {
+        if (obj instanceof Object) {
+            if (obj.start_time || obj.startTime) {
```
fviernau, line 25: Have you considered naming this `analyzer_start_time`, because just `start_time` is a bit ambiguous?

..same for several other places.

reporter-web-app/src/models/MetaData.js
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+class MetaData {
```
fviernau, line 20: Instead of just `MetaData` how about `OrtJob` or `OrtRun` to make the name less generic?

reporter/src/main/kotlin/model/MetaData.kt
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
fviernau, line 4: Is it intended to not add HERE as copyright holder?

reporter/src/main/kotlin/model/EvaluatedModelMapper.kt
```diff
@@ -120,7 +121,8 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
             ),
             repository = input.ortResult.repository,
             repositoryConfiguration = yamlMapper.writeValueAsString(input.ortResult.repository.config),
-            labels = input.ortResult.labels
+            labels = input.ortResult.labels,
+            metaData = MetaDataCalculator().getMetaData(input.ortResult)
```
fviernau, line 125: The `MetaDataCalculator` class seems so small - how about inlining it below line 101?

fviernau (changes requested):

reporter-web-app/src/components/AboutModal.js
```diff
@@ -167,6 +181,17 @@ const AboutModal = (props) => {
                         third-party software components under other open source licenses.
                         See OSS Review Toolkit code repository for further details.
                     </p>
+                    {
+                        !!creationDate
+                        && (
+                            <p>
+                                This ORT run started on
```
tsteenbe, line 188: https://github.com/oss-review-toolkit/ort/issues/3727 "In order to find out easily, while studying a report, when it was created, I would like to have the creation date present in the report." 

I presumed this was the WebApp report creation date itself not the Analyzer start time.`
It may occur that the analyzer, scanner and reporter were run on different dates so I get why you implemented the analyzer time but then should change the text here to "This ORT report is based on analysis started on" 
Grammar-wise using "This ORT run" is weird as this sentence is part of a static HTML file so nothing is "running"

reporter-web-app/src/models/MetaData.js
tsteenbe: @fviernau  We want to add more data into `MetaData` in follow up PRs such as the ORT version, as requested in https://github.com/oss-review-toolkit/ort/issues/3901

tsteenbe (changes requested):

reporter-web-app/src/models/MetaData.js
fviernau: hm, I see - my proposal doesn't really improve the situation then. I'm good with just MetaData then.

reporter/src/main/kotlin/model/EvaluatedModelMapper.kt
MarcelBochtler: We had this discussion ourselves, but as we expect that more data will be added to this model, we thought that it is beneficial to add a separate `Calculator` right now instead of refactoring at a later point.

reporter-web-app/src/components/AboutModal.js
MarcelBochtler: Yes you're right, but I changed your suggestion to `This ORT report is based on an analysis started on`.

sschuberth: But to me the question remains: Why pick the analyzer time instead of, say, the scanner time? I cannot currently think of a good justification for using another tool's time than the time the report was generated. That is, if we want to stick to single time. Another option of course would be to simply include the time (and duration, maybe) of *all* tools that were run.

neubs-bsi: If we simply add the time the report was created, it would not reflect the time the data in the report was collected, which is ultimately what matters, in my opinion.
We could add the start time of the scanner as well, since data is collected there too.

sschuberth: > We could add the start time of the scanner as well, since data is collected there too.

And the advisor, and the evaluator, and ... you see where this is going. Hence my proposal to stick to the report generation time.

neubs-bsi: I do see where this is going. We had this discussion during implementation. 
But as I say, I feel like the time the report was created, is the most uninteresting of all the times.

mnonnenmacher: I'm not saying it needs to happen in this change, but why not go one step further and add the start and end times for all tools? Then the report could also show the durations which would be useful information.

sschuberth: > why not go one step further and add the start and end times for all tools?

Quoting myself from [above](https://github.com/oss-review-toolkit/ort/pull/3978#discussion_r626432039):

> Another option of course would be to simply include the time (and duration, maybe) of all tools that were run.

😉

mnonnenmacher: See, I support that 🙂

fviernau: @mnonnenmacher @sschuberth, do you want these multiple values to be added in scope of this PR? (I'd approve otherwise as it LGTM)

sschuberth: It's ok from my side, we can improve this later.

fviernau (dismissed):
LGTM

sschuberth (approved):

sschuberth:

PR #3981, opened 2021-05-05
SpdxExternalReference: Make the type safer to use

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -186,7 +186,7 @@ internal fun SpdxPackage.getVcsInfo(): VcsInfo? {
  * [SpdxPackage], or null if there is no such reference.
  */
 private fun SpdxPackage.locateExternalReference(type: SpdxExternalReference.Type): String? =
-    externalRefs.find { it.referenceType == type.typeName && it.referenceCategory == type.category }?.referenceLocator
+    externalRefs.find { it.referenceType == type }?.referenceLocator
 
 /**
  * Return the organization from an "originator", "supplier", or "annotator" string, or null if no organization is
@@ -327,7 +327,7 @@ class SpdxDocumentFile(
 
         return Package(
             id = id,
-            purl = locateExternalReference(SpdxExternalReference.Type.PURL) ?: id.toPurl(),
+            purl = locateExternalReference(SpdxExternalReference.Type.Purl) ?: id.toPurl(),
             // TODO: Find a way to track authors.
             authors = sortedSetOf(),
             declaredLicenses = sortedSetOf(licenseDeclared),
--- a/spdx-utils/src/main/kotlin/model/SpdxExternalReference.kt
+++ b/spdx-utils/src/main/kotlin/model/SpdxExternalReference.kt
@@ -20,6 +20,12 @@
 package org.ossreviewtoolkit.spdx.model
 
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonValue
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer
 
 /**
  *  References an external source of additional information, metadata, enumerations, asset identifiers, or downloadable
@@ -47,7 +53,8 @@ data class SpdxExternalReference(
      * The references type as specified by
      * https://github.com/spdx/spdx-spec/blob/master/chapters/appendix-VI-external-repository-identifiers.md.
      */
-    val referenceType: String
+    @JsonDeserialize(using = ReferenceTypeDeserializer::class)
+    val referenceType: Type
 ) {
     enum class Category {
         SECURITY,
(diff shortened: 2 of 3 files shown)


spdx-utils/src/main/kotlin/model/SpdxExternalReference.kt
```diff
@@ -56,25 +63,38 @@ data class SpdxExternalReference(
...
+    }
+}
 
-        require(referenceType.isNotBlank()) { "The referenceType must not be blank." }
+private class ReferenceTypeDeserializer : StdDeserializer<SpdxExternalReference.Type>(
+    SpdxExternalReference.Type::class.java
+) {
+    override fun deserialize(p: JsonParser, ctxt: DeserializationContext): SpdxExternalReference.Type {
+        val node = p.codec.readTree<JsonNode>(p)
+        val typeName = node.textValue()
```
MarcelBochtler, line 102: Shouldn't you access the `typeName` property of `Type` to get this value?

sschuberth: No, that's part of the trick: This deserializer is only for the `referenceType` field. Any while in the YAML / JSON only associates a string value to that field, we construct a more complex / nested type here from that string.

MarcelBochtler (approved):

spdx-utils/src/main/kotlin/model/SpdxExternalReference.kt
```diff
@@ -56,25 +63,45 @@ data class SpdxExternalReference(
...
+
+        class Other(typeName: String) : Type(typeName, Category.OTHER)
     }
 
     init {
         require(referenceLocator.isNotBlank()) { "The referenceLocator must not be blank." }
 
-        require(referenceType.isNotBlank()) { "The referenceType must not be blank." }
+        if (referenceType.category != Category.OTHER) {
+            require(referenceType.category == referenceCategory) {
```
mnonnenmacher, line 89: Could remove one level of nesting by: `require(referenceType.category == Category.OTHER || referenceType.category == referenceCategory)`

mnonnenmacher: If `referenceType.category` is `Category.OTHER`, why must then `referenceCategory` not be `Category.OTHER` as well?

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -186,7 +186,7 @@ internal fun SpdxPackage.getVcsInfo(): VcsInfo? {
  * [SpdxPackage], or null if there is no such reference.
  */
 private fun SpdxPackage.locateExternalReference(type: SpdxExternalReference.Type): String? =
-    externalRefs.find { it.referenceType == type.typeName && it.referenceCategory == type.category }?.referenceLocator
+    externalRefs.find { it.referenceType == type }?.referenceLocator
```
mnonnenmacher, line 189: Commit message: Any explanation why deserialization was not "safe" before but now is?

spdx-utils/src/main/kotlin/model/SpdxExternalReference.kt
```diff
@@ -56,25 +63,45 @@ data class SpdxExternalReference(
...
+}
+
+private class ReferenceTypeDeserializer : StdDeserializer<SpdxExternalReference.Type>(
+    SpdxExternalReference.Type::class.java
+) {
+    override fun deserialize(p: JsonParser, ctxt: DeserializationContext): SpdxExternalReference.Type {
+        val node = p.codec.readTree<JsonNode>(p)
+        val typeName = node.textValue()
+        val type = SpdxExternalReference.Type::class.sealedSubclasses.mapNotNull { it.objectInstance }
+            .find { it.name == typeName }
```
mnonnenmacher, line 111: Optional (shorter): `.find { it.objectInstance?.name == typeName }`

mnonnenmacher (changes requested):

spdx-utils/src/main/kotlin/model/SpdxExternalReference.kt
sschuberth: I wasn't sure how the read the [specification for this](https://spdx.github.io/spdx-spec/appendix-VI-external-repository-identifiers/), but my thinking was that if a user wants to refer to package manager unknown to SPDX, would make sense to allow the user to still use `PACKAGE_MANAGER` as the category, even if our "auto-detecting" of the category would say `OTHER`.

spdx-utils/src/main/kotlin/model/SpdxExternalReference.kt
sschuberth: That does not work as the return type would be `KClass` instead of `SpdxExternalReference.Type` then.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: I agree "type safety" was a bad choice of words. I've improved the commit message.

spdx-utils/src/main/kotlin/model/SpdxExternalReference.kt
mnonnenmacher: Thanks, that makes sense.

mnonnenmacher (approved):

sschuberth:
Merging despite the unrelated test failures, which will be fixed by https://github.com/oss-review-toolkit/ort/pull/3986.

PR #3982, opened 2021-05-05
postgresql: Use default schema if not defined

heliocastro:
schema now is not required and always fallback on postgresql default
public schema.
Custom schemas need to be created and granted previously on database

### DIFF
--- a/README.md
+++ b/README.md
@@ -492,7 +492,7 @@ ort {
     storages {
       postgresStorage {
         url = "jdbc:postgresql://example.com:5444/database"
-        schema = "schema"
+        schema = "public"
         username = "username"
         password = "password"
         sslmode = "verify-full"
@@ -510,7 +510,9 @@ ort {
 }
 ```
 
-While the specified schema already needs to exist, the _scanner_ will itself create a table called `scan_results` and
+The database needs to exist. If the schema is set to something else than the default of `public`, it needs to exist and be accessible by the configured username.
+
+The _scanner_ will itself create a table called `scan_results` and
 store the data in a [jsonb](https://www.postgresql.org/docs/current/datatype-json.html) column.
 
 If you do not want to use SSL set the `sslmode` to `disable`, other possible values are explained in the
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -510,7 +510,9 @@ ort {
 }
 ```
 
-While the specified schema already needs to exist, the _scanner_ will itself create a table called `scan_results` and
+The database needs to exist. If the schema is defined and is not the default postgress _public_, it need to be previously created and grant usage for the specific database user. Left schema entry empty will always default to _public_.
```
sschuberth, line 513: Second sentence:

    If the schema is set to something else than the default of "public", it needs to exist and be accessible by the configured username.

model/src/main/resources/reference.conf
```diff
@@ -100,8 +100,8 @@ ort {
 
       postgres {
         url = "jdbc:postgresql://your-postgresql-server:5444/your-database"
-        schema = schema
-        username = username
+        schema = "public"
+        username = username
```
sschuberth, line 104: Please remove the trailing space.

README.md
```diff
@@ -492,7 +492,7 @@ ort {
     storages {
       postgresStorage {
         url = "jdbc:postgresql://example.com:5444/database"
-        schema = "schema"
+        schema = "public"
```
sschuberth, line 495: Commit message:

```
postgres: Use default schema if none defined

The schema now is an optional parameter as the postgres default of "public"
is used.
```

There is no need to repeat too much of the README docs here.

sschuberth (dismissed):

model/src/main/kotlin/utils/DatabaseUtils.kt
```diff
@@ -42,10 +42,6 @@ object DatabaseUtils {
             "URL for PostgreSQL storage is missing."
         }
 
-        require(config.schema.isNotBlank()) {
-            "Schema for PostgreSQL storage is missing."
-        }
```
mnonnenmacher, line 47: The schema could be set to a blank string in the configuration file so this check is still valid.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #3987, opened 2021-05-06
Upgrade to Kotlin 1.5

sschuberth:


fviernau:
> @mnonnenmacher @fviernau while the tests pass now, I'm not so happy that I had to disable parallel execution of tasks across projects. There is a _chance_ that is gets fixed with Kotlin 1.5.10, but I don't know when it will be released, or whether it would really fix the issue. How do we want to go forward? Merge this now and enable parallel task execution again on the upgrade to Kotlin 1.5.10?

The only thing I am considering now is the impact on feedback loops. In particular how does this affect the execution time of (1) build + tests on CI (2) local incremental build. Do you have any idea / guess? 

If it's hard to guess, I'd throw in the idea of splitting out and merging the commit "Gradle: Disable parallel execution of tasks in different projects " now, to give it a try. If it does not harm then merge the rest in a few days.

mnonnenmacher:
> @mnonnenmacher @fviernau while the tests pass now, I'm not so happy that I had to disable parallel execution of tasks across projects. There is a _chance_ that is gets fixed with Kotlin 1.5.10, but I don't know when it will be released, or whether it would really fix the issue. How do we want to go forward? Merge this now and enable parallel task execution again on the upgrade to Kotlin 1.5.10?

I tend to wait for 1.5.10 or a later version that fixes this if there is nothing that pushes us to upgrade Kotlin ASAP. But it depends of course on how big the impact on build times is if parallel execution is disabled, especially during development. If it's negligible I'm fine with upgrading.

sschuberth:
> I tend to wait for 1.5.10 or a later version that fixes this if there is nothing that pushes us to upgrade Kotlin ASAP.

Agreed, let's wait. Maybe by then also our other dependencies have upgraded to Kotlin 1.5, which might make things a bit smoother.

sschuberth:
Note to myself, Kotest "will release 4.6 within 2 weeks with 1.5 compatbility".

sschuberth:
This is now ready for review.

spdx-utils/src/main/kotlin/Extensions.kt
```diff
@@ -124,6 +124,12 @@ fun String.isSpdxExpression(): Boolean =
 fun String.isSpdxExpressionOrNotPresent(): Boolean =
     SpdxConstants.isNotPresent(this) || isSpdxExpression()
 
+/**
+ * Return this string lower-cased except for the first character which is upper-cased.
+ */
+fun String.titlecase() =
+    lowercase().replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() }
```
oheger-bosch, line 131: Refer to other remark with regards to usage of the default locale.

utils/src/main/kotlin/Extensions.kt
```diff
@@ -368,6 +368,11 @@ fun String.percentEncode(): String =
 fun String.stripCredentialsFromUrl() =
     toUri { URI(it.scheme, null, it.host, it.port, it.path, it.query, it.fragment).toString() }.getOrDefault(this)
 
+/**
+ * Return this string lower-cased except for the first character which is upper-cased.
+ */
+fun String.titlecase() = lowercase().uppercaseFirstChar()
```
oheger-bosch, line 374: It might be (slightly) confusing that there are different implementations of `titlecase()`. I assume, a central one is not possible because of restrictions regarding the dependencies between modules. But shouldn't then all implementations look the same?

sschuberth: > I assume, a central one is not possible because of restrictions regarding the dependencies between modules.

Exactly.

> But shouldn't then all implementations look the same?

I introduced the `uppercaseFirstChar()` utility function only later as it was also needed separately, but only in the main code. I'm undecided et whether it's worth it to duplicate this function also in the independent modules if it's only ever used once there (and thus could be inlined).

clients/clearly-defined/src/main/kotlin/Extensions.kt
```diff
@@ -27,3 +29,9 @@ import okio.Buffer
  * Return the request body as a string, see https://github.com/square/okhttp/issues/1891.
  */
 fun RequestBody.string() = Buffer().also { writeTo(it) }.readUtf8()
+
+/**
+ * Return this string lower-cased except for the first character which is upper-cased.
+ */
+fun String.titlecase() =
+    lowercase().replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() }
```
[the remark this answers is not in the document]
sschuberth: I was simply following the advice from the deprecation message, see

https://github.com/JetBrains/kotlin/blob/e927764aafc4571373bd2053497f8179e3cf5911/libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt#L721

I'm not sure whether using `Char.titlecaseChar()` is the right approach here, as it's documented as "This function performs one-to-one character mapping. To support one-to-many character mapping use the titlecase function", where as the `String.lowercase()` function is documented as "This function supports one-to-many and many-to-one character mapping, thus the length of the returned string can be different from the length of the original string".

So with your proposal one implementation would support many-to-one mappings while the other would not. However, I wonder whether instead of `Locale.getDefault()` rather `Locale.ROOT` should be used, as in the implementation of `String.lowercase()`. Any opinion on that?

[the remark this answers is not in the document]
oheger-bosch: This Unicode stuff is indeed confusing. I would then indeed prefer `Locale.ROOT` here; with the default locale, the results of these functions can vary depending on the local setup.

spdx-utils/src/main/kotlin/Extensions.kt
sschuberth: See the discussion above.

oheger-bosch (approved):

PR #3989, opened 2021-05-06
reporter: Use PURLs in PDF templates

sschuberth:
PURLs are of wider use than ORT's own coordinates, so use them instead
in the reports.


MarcelBochtler (approved):

MarcelBochtler:
Would this be a `breaking change` as the IDs in the report are changed?

PR #3990, opened 2021-05-06
fossid-webapp: Change some extension functions return type

nnobelis:
As explained in the toList function, the FossID server is not
consistent. It usually returns a List or a Map, but if there is no
entry, it returns data:false.
The toList function abstracts this behaviour, but we nevertheless need
a Kotlin representation that could be all these types at once, to be
able to deserialize them from JSON.
That's why we are forced to use EntityPostResponseBody<Any> in
several functions. However such generic placeholder is not nice for
API users. Therefore this commit changes the return type of the
extension functions and inline the toList call.

### DIFF
--- a/clients/fossid-webapp/src/main/kotlin/Extensions.kt
+++ b/clients/fossid-webapp/src/main/kotlin/Extensions.kt
@@ -23,6 +23,8 @@ package org.ossreviewtoolkit.clients.fossid
 
 import com.fasterxml.jackson.module.kotlin.convertValue
 
+import org.ossreviewtoolkit.clients.fossid.model.Scan
+
 private const val SCAN_GROUP = "scans"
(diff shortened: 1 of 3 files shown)


nnobelis:
I found another solution which is maybe more elegant:
https://github.com/bosch-io/oss-review-toolkit/tree/nnobelis/rework_functions_with_any2

sschuberth:
I like the idea of https://github.com/bosch-io/oss-review-toolkit/commit/39d2e162924adfd62876194510a58cd432488e3d to use custom deserializers to completely get rid of the `toList()` function. (That's the goal, right?) Please try to implement that approach. I already found some nits to address in your POC commit, but we can address there later as part of a real PR review.

nnobelis:
Replaced by https://github.com/oss-review-toolkit/ort/pull/4045.

PR #3992, opened 2021-05-06
Improve debug logging in CLI commands

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -37,6 +37,7 @@ import java.io.File
 
 import kotlin.system.exitProcess
 import kotlin.time.measureTime
+import kotlin.time.measureTimedValue
 
 import org.apache.logging.log4j.Level
 import org.apache.logging.log4j.core.config.Configurator
@@ -45,6 +46,7 @@ import org.ossreviewtoolkit.commands.*
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.config.LicenseFilenamePatterns
 import org.ossreviewtoolkit.model.config.OrtConfiguration
+import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.model.writeValue
 import org.ossreviewtoolkit.utils.Environment
 import org.ossreviewtoolkit.utils.ORT_CONFIG_DIR_ENV_NAME
@@ -213,6 +215,19 @@ class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
 
         return header.joinToString("\n", postfix = "\n")
     }
+}
+
+/**
+ * Read [ortFile] into an [OrtResult] and return it.
+ */
+fun CliktCommand.readOrtResult(ortFile: File): OrtResult {
+    val (ortResult, duration) = measureTimedValue { ortFile.readValue<OrtResult>() }
+
+    log.perf {
+        "Read ORT result from '${ortFile.name}' (${ortFile.formatSizeInMib}) in ${duration.inMilliseconds}ms."
+    }
+
+    return ortResult
 }
 
 /**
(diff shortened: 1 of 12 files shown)


cli/src/main/kotlin/OrtMain.kt
```diff
@@ -71,7 +71,7 @@ data class GlobalOptions(
     val forceOverwrite: Boolean
 )
 
-class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
+object OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
```
fviernau, line 83: Does this change work well together with the succeeding commit wrt. to our log function (see below)?
I wonder what `this::class.java` evaluates when it's an object. I guess it's an anonymous class but haven't checked.

```
val Any.log: KotlinLogger
    inline get() = loggerOfClass.getOrPut(this::class.java) {
        this::class.qualifiedName?.let { name ->
            require(name.startsWith("org.ossreviewtoolkit.")) {
                "Logging is only allowed on ORT classes, but '$name' is used."
            }
        }

        loggerOf(this::class.java)
    }
```

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -210,6 +215,20 @@ object OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
     }
 }
 
+/**
+ * Write the [ortResult] to all [outputFiles] for the given [resultName].
+ */
+fun CliktCommand.writeOrtResult(ortResult: OrtResult, outputFiles: Collection<File>, resultName: String) {
+    outputFiles.forEach { file ->
+        println("Writing $resultName result to '$file'.")
```
fviernau, line 241: Have you considered moving the `println()` above the for loop, like:
`println("Writing $resultName result to '$outputFiles'.")

The output IMO gives kind of a heads up that this can take a little bit.
So that heads up can be given right away for all files which are planned to be written.

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -45,6 +46,7 @@ import org.ossreviewtoolkit.commands.*
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.config.LicenseFilenamePatterns
 import org.ossreviewtoolkit.model.config.OrtConfiguration
+import org.ossreviewtoolkit.model.readValue
```
fviernau, line 50: Consider swapping the order of this commit with the previous one.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -229,39 +228,33 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
             }
         }
 
-        var (ortResultInput, readDuration) = measureTimedValue { ortFile?.readValue<OrtResult>() }
-
-        ortFile?.let { file ->
-            log.perf {
-                "Read ORT result from '${file.name}' (${file.formatSizeInMib}) in ${readDuration.inMilliseconds}ms."
-            }
+        val existingOrtFile = requireNotNull(ortFile) {
```
fviernau, line 231: Moving this here could be a separate commit preceding this one.
IMO it would be nice to at least mention that: while at it this check has been moved further up to fail early.

fviernau (approved):

cli/src/main/kotlin/OrtMain.kt
sschuberth: No, I haven't considered this (as it was like that before). But what's the advantage of giving this "heads-up" for all files at once? Isn't it even better the way it is now, where the heads-up is given before each file is written, and then by the time of the next heads-up you know that the previous file was fully written? Like a crude progress indicator?

cli/src/main/kotlin/OrtMain.kt
sschuberth: I liked it better to first refactor the advisor / scanner to have the extracted code at the same location as the other subcommands, and then extract the code.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
sschuberth: > Moving this here could be a separate commit preceding this one.

You're right, done.

fviernau (approved):

PR #3993, opened 2021-05-07
flutter: update version to v2.0.6

JeroenKnoops:
Signed-off-by: Jeroen Knoops <jeroen.knoops@gmail.com>

- Update version of default flutter to 2.0.6
- Change call to `pub` to `dart pub` resolving the error below.

## Error running `pub`
```
06:31:07.727 [DefaultDispatcher-worker-2] ERROR org.ossreviewtoolkit.analyzer.managers.Pub - Resolving dependencies for 'pubspec.yaml' failed with: IOException: Cannot run program "pub"
```

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -75,7 +75,7 @@ private const val GRADLE_VERSION = "5.6.4"
 private const val PUB_LOCK_FILE = "pubspec.lock"
 
 private val flutterCommand = if (Os.isWindows) "flutter.bat" else "flutter"
-private val pubCommand = if (Os.isWindows) "pub.bat" else "pub"
+private val pubCommand = if (Os.isWindows) "pub.bat" else "dart pub"
 
 private val flutterVersion = Os.env["FLUTTER_VERSION"] ?: "2.0.6-stable"
 private val flutterInstallDir = "$ortToolsDirectory/flutter-$flutterVersion"
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -77,7 +77,7 @@ private const val PUB_LOCK_FILE = "pubspec.lock"
 private val flutterCommand = if (Os.isWindows) "flutter.bat" else "flutter"
 private val pubCommand = if (Os.isWindows) "pub.bat" else "pub"
 
-private val flutterVersion = Os.env["FLUTTER_VERSION"] ?: "v1.12.13+hotfix.9-stable"
+private val flutterVersion = Os.env["FLUTTER_VERSION"] ?: "2.0.6-stable"
 private val flutterInstallDir = "$ortToolsDirectory/flutter-$flutterVersion"
 
 private val flutterHome by lazy {


JeroenKnoops:
I will dive in the errors in the buildpipeline.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -75,9 +75,9 @@ private const val GRADLE_VERSION = "5.6.4"
 private const val PUB_LOCK_FILE = "pubspec.lock"
 
 private val flutterCommand = if (Os.isWindows) "flutter.bat" else "flutter"
-private val pubCommand = if (Os.isWindows) "pub.bat" else "pub"
+private val pubCommand = if (Os.isWindows) "pub.bat" else "dart pub"
```
sschuberth, line 78: Please explain in the commit message why this change is necessary as part of the update.

Also, for single-file changes our convention is to use the name of the file as the topic, and start with a capitalized letter after the colon: "Pub: Upgrade Flutter to version 2.0.6"

Finally, a link to the change log of Flutter 2.0.6 would be nice for completeness.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -77,7 +77,7 @@ private const val PUB_LOCK_FILE = "pubspec.lock"
 private val flutterCommand = if (Os.isWindows) "flutter.bat" else "flutter"
 private val pubCommand = if (Os.isWindows) "pub.bat" else "pub"
 
-private val flutterVersion = Os.env["FLUTTER_VERSION"] ?: "v1.12.13+hotfix.9-stable"
+private val flutterVersion = Os.env["FLUTTER_VERSION"] ?: "2.0.6-stable"
```
sschuberth, line 80: Does this work without the next commit? If not, commits should be squashed as each individual commit ideally should pass all tests.

analyzer/src/main/kotlin/managers/Pub.kt
MarcelBochtler, line 78: This is not strictly necessary and would maybe break backwards compatibility.
When downloading the flutter SDK from [flutter.dev](https://flutter.dev/docs/development/tools/sdk/releases) the `pub` binary is included ([it is deprecated](https://dart.dev/tools/pub/cmd)).

The issue is that (some?) package managers do not include it anymore (incl. Homebrew and Macports on MacOS).
Which is how most developers would install flutter on their environment.

Would it be possible to add an additional check for the below configured `flutterVersion` and adjust the command accordingly?
`dart pub` is available from dart version 2.10 ([source](https://dart.dev/tools/pub/cmd)) which is included in flutter version >= 1.22.0.

MarcelBochtler (changes requested):

sschuberth:
@JeroenKnoops is there any interest from your side to move this forward? Otherwise we'll probably take over, as also @fviernau was doing something similar in https://github.com/oss-review-toolkit/ort/pull/4215.

analyzer/src/main/kotlin/managers/Pub.kt
MarcelBochtler: > When downloading the flutter SDK from flutter.dev the pub binary is included (it is deprecated).

I just rechecked. This is incorrect. The SDK doesn't package the `pub` binary anymore.

Unfortunately this change doesn't work as at some point ORT tries to run `dart pub --version` which is not a valid command.

JeroenKnoops:
I can take take another look, but I don't know how to run the Azure Pipelines locally... So it's hard to get the test passing...

sschuberth:
> I can take take another look, but I don't know how to run the Azure Pipelines locally... So it's hard to get the test passing...

You shouldn't aim to run *Azure Pipelines* locally, but just the same (functional) tests that we run *on* Azure Pipelines. For *Pub* stuff that would be `./gradlew analyzer:funTest --tests org.ossreviewtoolkit.analyzer.managers.PubFunTest`.

sschuberth:
Closing this in favor of https://github.com/oss-review-toolkit/ort/pull/4429.

PR #3994, opened 2021-05-07
Add CocoaPods dependency manager

pietbrauer:
analyzer: Add initial support for the CocoaPods dependency manager

Constructing the dependency tree for any given `Podfile` solely from the
output of the `pod` sub-commands requires running `pod` sub-commands
which only run on macOS. This implementation avoids all macOS-only
commands by constructing the dependency tree from the lockfile and
therefore the analysis of any `Podfile` requires the precence of the
corresponding `Podfile.lock`.

In order to obtain the meta-data for a dependency, the corresponding
`.podspec` file is determined via the command `pod spec which`. These
paths always point under `~/cocoapods/repos` which is where CocoaPods
sets up its so called Specs repositories. Which repositories are to be
used is defined by the respective `Podfile` via the `source` property,
which defaults to [1]. This implementation disregards the specified
repositories and always assumes that default. That default specs
repository is however not used via GitHub, but via CDN as this is much
faster [2]. Cloning [1] takes 10-30 minutes and takes a bit more than
5GB while the CDN approach initializes in no time and fetches only
required data.

Even with that CDN based approach the implementation can
and should be extended to adhere to the specified repositories in the
future, see #4188.

[1] https://github.com/CocoaPods/Specs
[2] CocoaPods/CocoaPods#7046

------------------------------------- Original description from Piet below -----------------------------------

This PR adds CocoaPods as a dependency manager.

I tested it on macOS and inside the docker container. The Gem runs fine on Linux, the only thing that doesn't work is `pod install` because Xcode is missing. As far as I am concerned `pod install` is not needed, if the `Podfile.lock` is present.

Things I am unsatisfied with/didn't know better:
1. How to mock the `Command` invocations to test the actual flow
2. JSON parsing
3. Parsing the strings like `Alamofire/Something (~> 1.0)` into a `PackageReference`
4. I also don't like cloning the spec repo inside the docker container which takes about 5 Minutes and adds a lot of storage. Setting it up on demand is even worse, if you do it every time you launch the docker container.

Any pointers are welcome.


sschuberth:
Thanks for the contribution! Unfortunately, at least I'd need a few more days before I could start looking at this.

fviernau:
Hi @pietbrauer ,

thanks for your contribution! I'm picking up this review now - mind helping me with the below initial questions?

1. You mentioned "As far as I am concerned pod install is not needed, if the Podfile.lock is present.". Can you give me a hint
    _why_ install is necessary if there is no lock file?
2. Regarding "I also don't like cloning the spec repo inside the docker container which takes about 5 Minutes and adds a lot of storage. Setting it up on demand is even worse, if you do it every time you launch the docker container."  Have you considered the approach to not clone it initially but on-the-fly, and at the same time recommend the user to either (a) bind mount a host directory into the Docker or (b) use a docker volume for that directory?
3. Have there been any higher level implementation / design decisions which you made, which aren't obvious from the PR description or commit (e.g. like choosing whether to parse lock files)? If so, mind sharing them e.g. in the PR description?

fviernau:
@pietbrauer In order move a bit faster, I've setup a branch [1] in this repository and pushed your changes (rebased onto master). I'll make some modifications to the code there (force push) and as soon as done we may create a new PR from that branch replacing this PR. Fine with you?

[1] https://github.com/oss-review-toolkit/ort/tree/cocoa-pods-support

pietbrauer:
@fviernau I'm currently on parental leave and won't work on it until August. We have it in use at Porsche so we don't have time pressure.

I don't get the *working on a different branch* but if that's how you want to review it so be it.
I enabled that maintainers can push onto this branch so rebasing and committing should be possible without splitting all the work onto multiple PRs or branches I can't push to.

fviernau:
@pietbrauer @sschuberth @mnonnenmacher 

I refactored or re-wrote a couple of things mainly for simplification / clarity. 
I've addressed the points 2, 3 mentioned by Piet in the description. Point 1 I'd see not as an issue anymore.
Regarding point 4, I'd prefer to address this in a follow up PR.
Also I'd like to add documentation in a follow-up PR in order to move this forward now.

It's up for your review now.

mnonnenmacher (changes requested):
Please also update the list of supported package managers in the README.

reporter-web-app/yarn.lock
```diff
@@ -92,16 +92,16 @@
     source-map "^0.5.0"
 
 "@babel/core@^7.1.0", "@babel/core@^7.12.3", "@babel/core@^7.7.5", "@babel/core@^7.8.4":
-  version "7.14.5"
-  resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.14.5.tgz#d281f46a9905f07d1b3bf71ead54d9c7d89cb1e3"
-  integrity sha512-RN/AwP2DJmQTZSfiDaD+JQQ/J99KsIpOCfBE5pL+5jJSt7nI3nYGoAXZu+ffYSQ029NLs2DstZb+eR81uuARgg==
+  version "7.14.6"
```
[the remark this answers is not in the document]
fviernau: > Did you verify that these version updates do not break anything in the web app report?

not yet.

> And is this rebased on the current master?

yes.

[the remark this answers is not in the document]
fviernau: I've dropped the commit.

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -46,7 +86,206 @@ class CocoaPods(
...
+    override fun beforeResolution(definitionFiles: List<File>) = checkVersion(analyzerConfig.ignoreToolVersions)
+
     override fun resolveDependencies(definitionFile: File): List<ProjectAnalyzerResult> {
-        TODO("not implemented") // To change body of created functions use File | Settings | File Templates.
+        // CocoaPods originally used and may still use the Specs repository on GitHub [1] as package meta-data database.
+        // Using [1] requires an initial clone which is slow to do and consumes already more than 5 GB on disk, see
+        // also [2]. (Final) CDN support has been added in version 1.7.2 [3] to speed things up.
+        //
+        // Temporarily re-configure the repos such that 'https://cdn.cocoapods.org' is the only repository in order to
+        // avoid having to clone and fetch the large Git repository. This ensure that the repositories used are
```
mnonnenmacher, line 105: init: "ensure**s**"

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -46,7 +86,206 @@ class CocoaPods(
...
+        // [3] https://blog.cocoapods.org/CocoaPods-1.7.2/
+
+        stashDirectories(File("~/.cocoapods/repos")).use {
+            run("repo", "add-cdn", "trunk", "https://cdn.cocoapods.org")
+
+            try {
+                return resolveDependenciesInternal(definitionFile)
+            } finally {
+                // The cache entries are not re-usable across definition files because the keys do not contain the
+                // dependency version. If non-default Specs repositories were supported, then these would also need to
```
mnonnenmacher, line 119: nit: "to **be** part"

mnonnenmacher (changes requested):

mnonnenmacher (approved):

mnonnenmacher (dismissed):

.azure-pipelines/LinuxAnalyzerTest.yml
```diff
@@ -43,6 +43,13 @@ jobs:
       mkdir -p $HOME/go/bin
       curl https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | sh
 
+      # Install 'CocoaPods'. As https://github.com/CocoaPods/CocoaPods/pull/10609 is needed but not yet released, make a
+      # build from the latest revision of the master branch.
+      git clone https://github.com/CocoaPods/CocoaPods.git
```
sschuberth, line 48: How about downloading exactly the revision we need (so the download does not change all the time), and without the Git history: https://github.com/CocoaPods/CocoaPods/archive/9461b346aeb8cba6df71fd4e71661688138ec21b.zip (This is the merge commit for the wanted PR.)

Dockerfile
```diff
@@ -149,6 +150,15 @@ RUN /opt/ort/bin/import_proxy_certs.sh && \
         SDK_MANAGER_PROXY_OPTIONS="--proxy=http --proxy_host=${PROXY_HOST_AND_PORT%:*} --proxy_port=${PROXY_HOST_AND_PORT##*:}"; \
     fi && \
     yes | $ANDROID_HOME/cmdline-tools/bin/sdkmanager $SDK_MANAGER_PROXY_OPTIONS --sdk_root=$ANDROID_HOME "platform-tools" && \
+    # Install 'CocoaPods'. As https://github.com/CocoaPods/CocoaPods/pull/10609 is needed but not yet released, make a
+    # build from the latest revision of the master branch.
+    git clone https://github.com/CocoaPods/CocoaPods.git && \
+        cd CocoaPods && \
+        git checkout 9461b346aeb8cba6df71fd4e71661688138ec21b && \
```
sschuberth, line 157: The same idea from above applies here to (hopefully) improve download performance.

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -79,7 +79,8 @@ abstract class PackageManager(
             "src/test/resources",
             // Ignore virtual environments in Python.
             "lib/python2.*/dist-packages",
-            "lib/python3.*/site-packages"
+            "lib/python3.*/site-packages",
+            "Pods"
```
sschuberth, line 83: This line should be headed by a new separating comment, explaining where this comes from (not from Python).

analyzer/src/funTest/kotlin/managers/CocoaPodsFunTest.kt
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright (C) 2017-2021 HERE Europe B.V.
+ *
```
sschuberth, line 3: Should this also have `Copyright (C) 2021 Dr. Ing. h.c. F. Porsche AG`, or was this file created anew?

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -19,25 +20,64 @@
...
+ * As pre-condition for the analysis each respective definition file must have a sibling lockfile named 'Podfile.lock'.
+ * The dependency tree is constructed solely based on parsing that lockfile. So, the dependency tree can be constructed
+ * on any platform. Note that obtaining the dependency tree from the 'pod' command without a lock file has Xcode
+ * dependencies and is not supported by this class.
+ *
+ * The only interactions with the 'pod' command happen in order to obtain meta-data for dependencies. Therefore
+ * 'pod repo update' and 'pod spec which' get executed, which works also under Linux.
+ *
+ * Note: This class depends on https://github.com/CocoaPods/CocoaPods/pull/10609 which is not yet released, so a build
+ * of the master branch is needed.
```
sschuberth, line 71: I would not be as specific here, esp when changing to the ZIP download above. Simply omitting the part after the comma would be fine with me.

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -46,7 +86,206 @@ class CocoaPods(
         ) = CocoaPods(managerName, analysisRoot, analyzerConfig, repoConfig)
     }
 
+    private val podspecCache: MutableMap<String, Podspec> = mutableMapOf()
```
sschuberth, line 89: Nit: We usually do `private val podspecCache = mutableMapOf<String, Podspec>()`

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -46,7 +86,206 @@ class CocoaPods(
...
+
+    override fun command(workingDir: File?) = "pod"
+
+    override fun getVersionRequirement(): Requirement = Requirement.buildIvy("[1.10.1,)")
+
+    override fun getVersionArguments(): String = "--version --allow-root"
```
sschuberth, line 95: Nit: I believe it'd be fine to also omit the explicit return types here, like in the functions above and below, as it's very clear from the function expression.

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -46,7 +86,206 @@ class CocoaPods(
...
+        //
+        // [1] https://github.com/CocoaPods/Specs
+        // [2] https://github.com/CocoaPods/CocoaPods/issues/7046.
+        // [3] https://blog.cocoapods.org/CocoaPods-1.7.2/
+
+        stashDirectories(File("~/.cocoapods/repos")).use {
+            run("repo", "add-cdn", "trunk", "https://cdn.cocoapods.org")
+
+            try {
+                return resolveDependenciesInternal(definitionFile)
```
sschuberth, line 116: Nit: Move out `return` to before `stashDirectories`?

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -46,7 +86,206 @@ class CocoaPods(
...
+    }
+
+    private fun getPodspec(id: Identifier, workingDir: File): Podspec {
+        podspecCache[id.name]?.let { return it }
+
+        val podspecName = id.name.substringBefore("/")
+
+        val podspecFile = run("spec", "which", podspecName, "--version=${id.version}", "--allow-root",
+            "--regex", workingDir = workingDir
+        ).requireSuccess().stdout.trim().let { File(it) }
```
sschuberth, line 203: Nit: Formatting is a bit odd, IMO. Maybe a line break after `run(` to put all rags into a separate line?

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -46,7 +86,206 @@ class CocoaPods(
...
+    val dependenciesForName = mutableMapOf<String, MutableSet<String>>()
+    val root = yamlMapper.readTree(podfileLock)
+
+    root.get("PODS").asIterable().forEach { node ->
+        val entry = when (node) {
+            is ObjectNode -> node.fieldNames().asSequence().first()
+            else -> node.textValue()
+        }
+
+        val (name, version) = NAME_AND_VERSION_REGEX.find(entry)!!.groups.let {
```
sschuberth, line 228: How can we be sure the `!!` in this line never causes an NPE?

sschuberth (changes requested):

analyzer/src/main/kotlin/PackageManager.kt
fviernau: I believe it should not be added after reading https://guides.cocoapods.org/using/using-cocoapods.html.

analyzer/src/funTest/kotlin/managers/CocoaPodsFunTest.kt
fviernau: The copyright is correct as-is because the file was written only by me.

analyzer/src/main/kotlin/managers/CocoaPods.kt
fviernau: To my understanding the file format of the lockfile implies that `!!` does not cause an NPE.

.azure-pipelines/LinuxAnalyzerTest.yml
fviernau: I'd like to leave this improvement up to a follow-up PR as this does cause me quite a bit: context switch and testing time.

Dockerfile
fviernau: I'd like to leave this improvement up to a follow-up PR as this does cause me quite a bit: context switch and testing time. Ok?

analyzer/src/main/kotlin/managers/CocoaPods.kt
sschuberth: So you're saying we *should* throw an exception on malformed files / not be lenient by skipping invalid entries or so. Ok.

fviernau: I'd prefer adding that fault tolerance only in case need arises.

.azure-pipelines/LinuxAnalyzerTest.yml
sschuberth: Ok.

analyzer/src/main/kotlin/managers/CocoaPods.kt
sschuberth: You moved this to before `try`, not to before `stashDirectories`. Was this intended?

analyzer/src/main/kotlin/managers/CocoaPods.kt
sschuberth: This comment also referred to line 93.

analyzer/src/main/kotlin/managers/CocoaPods.kt
fviernau: I intended to do what you suggested, so I did wrong. Moved it to before `stashDirectories`.

sschuberth (approved):

.azure-pipelines/LinuxAnalyzerTest.yml
fviernau: See #4204.

PR #3995, opened 2021-05-07
Download improvements

sschuberth:


sschuberth:
FYI, I plan to also make more use of the new helper function, but wanted to split this into new PRs.

utils/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -120,23 +120,23 @@ object OkHttpClientHelper {
      * [Result] wrapping an [IOException] (which might be a [HttpDownloadError]) on failure.
      */
     fun downloadText(url: String): Result<String> {
-        val request = Request.Builder().get().url(url).build()
-        val response = runCatching { execute(request) }.getOrElse { return Result.failure(it) }
+        val getRequest = Request.Builder().get().url(url).build()
+        val responseForGetRequest = runCatching { execute(getRequest) }.getOrElse { return Result.failure(it) }
 
-        return response.use {
+        return responseForGetRequest.use { response ->
```
fviernau, line 130: Is the reason for adding `get` into request / response variable the avoiding of a naming conflict when finding a better name for `it` below? 

If so (and as I like just `request` `response` a bit more, how about extracting `runCatching { execute(getRequest) }.getOrElse { return Result.failure(it) }` into a function instead, e.g. `fun request.execute(): Response`.

The you can just write: 

```
val request = Request.Builder().get().url(url).build()

return request.execute().use { response ->
...
```

fviernau: Reading up the documentation of  [1] it seems that the closing (Closable) interface only affects the response body.
So, using `use` is only necessary for the body. If you just modify the line 136 to `val text = response.use { it.body?.string().orEmpty()` }, then you can keep the shorter variables `request` and `response` for the rest of the function scope.


[1] https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response/

fviernau (changes requested):

utils/src/main/kotlin/OkHttpClientHelper.kt
sschuberth: > Is the reason for adding `get` into request / response variable the avoiding of a naming conflict when finding a better name for `it` below?

Yes.

> Reading up the documentation of [1] it seems that the closing (Closable) interface only affects the response body.

That was a very good hint actually, thanks for that! This makes the code more nicer.

fviernau (changes requested):

utils/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -138,6 +142,73 @@ object OkHttpClientHelper {
...
+            return Result.failure(HttpDownloadError(response.code, response.message))
+        }
+
+        // Depending on the server, we may only get a useful target file name when looking at the response
+        // header or at a redirected URL. In case of the Crates registry, for example, we want to resolve
+        //     https://crates.io/api/v1/crates/cfg-if/0.1.9/download
+        // to
+        //     https://static.crates.io/crates/cfg-if/cfg-if-0.1.9.crate
+        //
+        // On the other hand, e.g. for GitHub exactly the opposite is the case, as there a get request for URL
```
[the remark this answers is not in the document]
sschuberth: I've improved the noun to say "get-request".

utils/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -138,6 +142,73 @@ object OkHttpClientHelper {
...
+            filenames.firstOrNull()?.removeSurrounding("\"")
+        }
+
+        listOf(response.request.url, getRequest.url).mapTo(candidateNames) {
+            it.pathSegments.last()
+        }
+
+        val fileName = candidateNames.find {
+            ArchiveType.getType(it) != ArchiveType.NONE
+        } ?: candidateNames.first()
```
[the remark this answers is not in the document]
sschuberth: IMO a `require()` wouldn't really make it clearer that we can *guarantee* that `candidateNames` is not empty at this point. After reading through https://www.bignerdranch.com/blog/write-better-code-using-kotlins-require-check-and-assert/ I went for `assert()` instead.

fviernau (approved):

PR #3999, opened 2021-05-10
Fix sorting of Identifiers

sschuberth:

### DIFF
--- a/analyzer/src/funTest/kotlin/managers/PubFunTest.kt
+++ b/analyzer/src/funTest/kotlin/managers/PubFunTest.kt
@@ -70,7 +70,7 @@ class PubFunTest : WordSpec() {
                 }
             }
 
-            "Resolve dependencies for a project with flutter correctly" {
+            "resolve dependencies for a project with flutter correctly" {
                 val workingDir = projectsDir.resolve("project-with-flutter")
                 val packageFile = workingDir.resolve("pubspec.yaml")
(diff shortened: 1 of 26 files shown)


fviernau:
@sschuberth why is this a "breaking change" ?

sschuberth:
> @sschuberth why is this a "breaking change" ?

I just wanted to be conservative in case someone has workflow steps that rely on the sorting by `id`, e.g. in the `curations.yml`. We actually have pre-submit checks in CI to enforce alphabetical sorting in that file. And this change would break it.

But I agree it's debatable. Should I remove the "breaking change" label?

fviernau:
> But I agree it's debatable. Should I remove the "breaking change" label?

I'm fine with the label. I just wanted to double check whether I'm not missing anything breaking here.

analyzer/src/funTest/kotlin/managers/PubFunTest.kt
```diff
@@ -108,7 +108,7 @@ class PubFunTest : WordSpec() {
                 result.toYaml() shouldBe expectedResult
             }
 
-            "Error is shown when no lockfile is present" {
+            "shown an error if no lockfile is present" {
```
oheger-bosch, line 111: Nit: "shown" -> "show"

oheger-bosch (approved):

MarcelBochtler (dismissed):

mnonnenmacher (approved):

PR #4002, opened 2021-05-11
LocalScanner: Name temporary directories also after the provenance

sschuberth:
Esp. when having reconfigured the sourceCodeOrigins, it can happen that
the stored scan results contain multiple results for the same package
with different provenance. If also createMissingArchives is enabled,
this could have led to name clashes in the created temporary
directories. Solve that by adding another directory level that is named
after the provenance.

### DIFF
--- a/scanner/src/main/kotlin/LocalScanner.kt
+++ b/scanner/src/main/kotlin/LocalScanner.kt
@@ -320,7 +320,7 @@ abstract class LocalScanner(
 
         try {
             missingArchives.forEach { (pkg, provenance) ->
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/LocalScanner.kt
```diff
@@ -320,7 +320,7 @@ abstract class LocalScanner(
```
oheger-bosch, line 320: Commit message nit: I think this should be "...this could have led to name *clashes* in the...".

PR #4003, opened 2021-05-11
getPathFromEnvironment: Support basic expansion of nested environment variables

sschuberth:
For simplicity, only support word-like characters in environment
variables names to be expanded. While on specific OSes also other
characters are valid as part of environment variable names (including
e.g. "\0"), these are edge-cases that should barely occur in reality,
and if so, it is probably not good practice anyway.

### DIFF
--- a/utils/src/main/kotlin/Utils.kt
+++ b/utils/src/main/kotlin/Utils.kt
@@ -172,14 +172,28 @@ fun getCommonFileParent(files: Collection<File>): File? =
  * Return the full path to the given executable file if it is in the system's PATH environment, or null otherwise.
  */
 fun getPathFromEnvironment(executable: String): File? {
+    fun String.expandVariable(referencePattern: Regex, groupName: String): String =
+        replace(referencePattern) {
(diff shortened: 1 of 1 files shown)


utils/src/main/kotlin/Utils.kt
```diff
@@ -175,11 +175,27 @@ fun getPathFromEnvironment(executable: String): File? {
...
+            }
+
+            resolveWindowsExecutable(File(expandedPath, executable))
         }.firstOrNull()
     } else {
-        paths.map { path -> File(path, executable) }.find { it.isFile }
+        val nestedVariableRegex = Regex("\\$\\{?(?<reference>\\w+)}?")
+
+        paths.map { path ->
+            val expandedPath = path.replace(nestedVariableRegex) {
```
oheger-bosch, line 192: There is some redundancy here. The replacement of variables in the path could be done in a helper function that takes the regular expression as a parameter.

oheger-bosch (approved):

PR #4004, opened 2021-05-11
GoMod: Extract the revision from versions in "pseudo version" format

fviernau:
See https://golang.org/ref/mod#pseudo-versions.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -189,7 +189,7 @@ class GoMod(
             VcsInfo(
                 type = VcsType.GIT,
                 url = "https://${id.name.removeSuffix("/")}.git",
-                revision = id.version
+                revision = getRevision(id.version)
             )
         } else {
             VcsInfo.EMPTY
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -286,3 +286,15 @@ private fun Collection<Edge>.toPackageReferenceForest(
 
     return nodes.values.filter { it.incomingEdges.isEmpty() }.map { getPackageReference(it.id) }.toSortedSet()
 }
+
+// See https://golang.org/ref/mod#pseudo-versions.
+private val PSEUDO_VERSION_REGEX = "^v0.0.0-([\\d]{14}-([0-9a-f]*)$)".toRegex()
+
+internal fun getRevision(version: String): String {
+    PSEUDO_VERSION_REGEX.find(version)?.let { matchResult ->
+        println("matches!")
```
sschuberth, line 295: Debug left-over?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -286,3 +286,15 @@ private fun Collection<Edge>.toPackageReferenceForest(
 
     return nodes.values.filter { it.incomingEdges.isEmpty() }.map { getPackageReference(it.id) }.toSortedSet()
 }
+
+// See https://golang.org/ref/mod#pseudo-versions.
+private val PSEUDO_VERSION_REGEX = "^v0.0.0-([\\d]{14}-([0-9a-f]*)$)".toRegex()
```
sschuberth, line 291: How about using both a non-capturing and a named group, like

    "^v0.0.0-(?:[\\d]{14}-(?<sha1>[0-9a-f]+)$)"

Note that I also replaced `*` with `+`.

sschuberth (changes requested):

PR #4010, opened 2021-05-12
Use correct database connection when starting transactions

### LINKED ISSUE #4009: Multiple database connections get mixed up
When using Postgres for both storing scan results and file archives, a connection pool is created for each of these use cases. In our case, these connection pools refer to the same database, but different schemas.

The way [Exposed](https://github.com/JetBrains/Exposed) is used in `PostgresStorage` and `PostgresFileArchiverStorage` to execute transactions, however, causes wrong connections to be used. The reason is the usage of the `transaction()` function without specifying the database connection to be used. The function then takes a connection from a thread-local storage.

For us, the effect when running a `scan` command is as follows:
* The first connection pool for the `PostgresStorage` is created.
* The scan results available in the storage are queried successfully.
* The second connection pool for the `PostgresFileArchiverStorage` is created.
* When new scan results are available `PostgresStorage` is invoked to store them. This fails, because the storage now uses the second connection pool. As this pool refers to a different schema, the table with scan results cannot be found.

A solution is to store the Exposed `Database` object representing the connection in the storage classes and explicitly refer to it when starting a new transaction.

oheger-bosch:
This PR fixes #4009 by explicitly referring to the correct database connection when starting a new transaction.

### DIFF
--- a/model/src/main/kotlin/utils/DatabaseUtils.kt
+++ b/model/src/main/kotlin/utils/DatabaseUtils.kt
@@ -22,8 +22,13 @@ package org.ossreviewtoolkit.model.utils
 import com.zaxxer.hikari.HikariConfig
 import com.zaxxer.hikari.HikariDataSource
 
+import kotlinx.coroutines.Deferred
+
+import org.jetbrains.exposed.sql.Database
 import org.jetbrains.exposed.sql.Transaction
 import org.jetbrains.exposed.sql.transactions.TransactionManager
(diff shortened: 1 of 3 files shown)


mnonnenmacher (approved):

sschuberth:
Merging this despite the unrelated test failures which get addressed by https://github.com/oss-review-toolkit/ort/pull/4011.

PR #4011, opened 2021-05-12
PipFunTest: Update the expected test results

fviernau:

### DIFF
--- a/analyzer/src/funTest/assets/projects/external/example-python-flask-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/external/example-python-flask-expected-output.yml
@@ -20,16 +20,20 @@ project:
     dependencies:
     - id: "PyPI::flask:0.12"
       dependencies:
-      - id: "PyPI::click:7.1.2"
-      - id: "PyPI::itsdangerous:1.1.0"
+      - id: "PyPI::click:8.0.0"
+      - id: "PyPI::itsdangerous:2.0.0"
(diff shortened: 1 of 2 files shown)


mnonnenmacher (dismissed):

sschuberth:
Hmm, for some reason (only) on Windows "PyPI::click:8.0.0" seems to have a dependency on "PyPI::colorama:0.4.4", which is missing in the expected output. So until we have another way to handle this, we probably need to start using different expected result files per OS.

fviernau:
> Hmm, for some reason (only) on Windows "PyPI::click:8.0.0" seems to have a dependency on "PyPI::colorama:0.4.4", which is missing in the expected output. So until we have another way to handle this, we probably need to start using different expected result files per OS.
> 
> _Edit:_ Indeed `colorama` is only requited on Windows, see https://github.com/pallets/click/blob/9da166957f5848b641231d485467f6140bca2bc0/setup.py#L3

@sschuberth  would you be fine if I disable the windows test in scope of this PR.
Then we can re-enable it in a follow-up PR?
I'm asking because I don't have a windows development setup right now.

sschuberth:
I would also be fine to disable that specific test on Windows for now.

sschuberth:
> So until we have another way to handle this, we probably need to start using different expected result files per OS.

PS: As I didn't find an issue yet where we've documented the need for this, I created one at https://github.com/oss-review-toolkit/ort/issues/4013.

sschuberth (approved):

PR #4018, opened 2021-05-12
Make use of a withoutSuffix() utility function

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -41,6 +41,7 @@ import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.utils.CommandLineTool
 import org.ossreviewtoolkit.utils.Os
 import org.ossreviewtoolkit.utils.stashDirectories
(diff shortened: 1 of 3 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -280,7 +281,7 @@ private fun Collection<Edge>.toPackageReferenceForest(
 private val PSEUDO_VERSION_REGEX = "^v0.0.0-(?:[\\d]{14}-(?<sha1>[0-9a-f]+)$)".toRegex()
 
 private fun getRevision(version: String): String {
-    if (version.endsWith("+incompatible")) return getRevision(version.removeSuffix("+incompatible"))
+    version.withoutSuffix("+incompatible")?.let { return getRevision(it) }
```
oheger-bosch, line 284: I found the original version easier to read and understand; the new variant is certainly more elegant. So I won't object here.

oheger-bosch (approved):

analyzer/src/main/kotlin/managers/GoMod.kt
sschuberth: It's probably just again one of those things one needs to get used to 😀 Thanks for approving anyway.

PR #4019, opened 2021-05-12
Embedded Postgres improvements

sschuberth:

### DIFF
--- a/model/src/test/kotlin/utils/PostgresFileArchiverStorageTest.kt
+++ b/model/src/test/kotlin/utils/PostgresFileArchiverStorageTest.kt
@@ -21,9 +21,9 @@ package org.ossreviewtoolkit.model.utils
 
 import com.opentable.db.postgres.embedded.EmbeddedPostgres
 
-import io.kotest.core.spec.IsolationMode
 import io.kotest.core.spec.Spec
 import io.kotest.core.spec.style.WordSpec
+import io.kotest.core.test.TestCase
 import io.kotest.matchers.shouldBe
 
 import java.io.File
@@ -68,14 +68,21 @@ class PostgresFileArchiverStorageTest : WordSpec() {
 
     override fun beforeSpec(spec: Spec) {
         postgres = EmbeddedPostgres.builder().setPGStartupWait(PG_STARTUP_WAIT).start()
+    }
+
+    override fun beforeTest(testCase: TestCase) {
+        postgres.postgresDatabase.connection.use { c ->
+            val s = c.createStatement()
+            s.execute("DROP SCHEMA public CASCADE")
+            s.execute("CREATE SCHEMA public")
+        }
+
         storage = PostgresFileArchiverStorage(postgres.postgresDatabase)
     }
 
     override fun afterSpec(spec: Spec) {
         postgres.close()
     }
-
-    override fun isolationMode() = IsolationMode.InstancePerTest
 
     init {
         "hasArchive()" should {
--- a/scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
+++ b/scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
@@ -128,7 +128,6 @@ abstract class AbstractStorageFunTest : WordSpec() {
     private lateinit var storage: ScanResultsStorage
 
     override fun beforeTest(testCase: TestCase) {
-        super.beforeTest(testCase)
         storage = createStorage()
     }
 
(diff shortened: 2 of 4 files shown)


model/src/test/kotlin/utils/PostgresFileArchiverStorageTest.kt
```diff
@@ -67,7 +72,7 @@ class PostgresFileArchiverStorageTest : WordSpec() {
         }
 
     override fun beforeSpec(spec: Spec) {
-        postgres = EmbeddedPostgres.builder().setPGStartupWait(PG_STARTUP_WAIT).start()
+        postgres = EmbeddedPostgres.builder().setPGStartupWait(PG_STARTUP_WAIT).setPort(port).start()
```
oheger-bosch, line 75: I had a look in the code of the `EmbeddedPostgres.Builder` class. IIUC, per default, the builder uses a free random port if no port is explicitly set. (Refer to the `detectPort()` function.) This is actually what we want, so it should not be necessary to set a random port explicitly. You change has the disadvantage that there is a risk that the selected port is already in use.

utils/src/funTest/kotlin/storage/HttpFileStorageFunTest.kt
```diff
@@ -43,7 +43,7 @@ import org.ossreviewtoolkit.utils.log
 class HttpFileStorageFunTest : WordSpec() {
     private val loopback = InetAddress.getLoopbackAddress()
     private val port = Random.nextInt(1024, 49152) // See https://en.wikipedia.org/wiki/Registered_port.
-        .also { log.debug { "Using port $it for test server." } }
+        .also { log.debug { "Using port $it for HTTP server." } }
```
oheger-bosch, line 46: Maybe it would make sense to rewrite this test using WireMock. WireMock directly supports starting a test server at a free random port.

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
```diff
@@ -128,7 +128,6 @@ abstract class AbstractStorageFunTest : WordSpec() {
     private lateinit var storage: ScanResultsStorage
```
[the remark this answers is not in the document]
sschuberth: In the general case I believe so, yes.

I'm emphasizing this here as for the change at https://github.com/oss-review-toolkit/ort/pull/4019/files#diff-19568fa757cf629a7054a4fbee0975cd0189a2bd594face19c39f81788356db5R44-R51, `super.beforeTest(testCase)` *must* be called last. And in order to understand why, you have to *know* that the super-implementation actually creates the storage, so recreating the schema has to come before it.

What I'm trying to say is, if you call *any* function (a super-method or anything else) you always have to have some understanding about what it does in order to call it at the right place. With that in mind, there can be no rule-of-thumb saying something like "always call a super-method first (or last) in your overridden method" (like sometimes being suggested on the Internet).

model/src/test/kotlin/utils/PostgresFileArchiverStorageTest.kt
sschuberth: You're right. I admit I didn't dig that deeply and simply added this on spec as I happened to be reminded of the similar code in *HttpFileStorageFunTest*. I'll simply drop the commit.

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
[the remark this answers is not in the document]
oheger-bosch: Probably a misunderstanding: I just wanted to point out a typo in the commit message. "known" should be replaced by "know".

utils/src/funTest/kotlin/storage/HttpFileStorageFunTest.kt
sschuberth: I'll leave that as a future exercise then 😁

Speaking about WireMock, since we're using OkHttp I was recently wondering whether we should also prefer to use [MockWebServer](https://github.com/square/okhttp/tree/master/mockwebserver) instead. Here is [some](https://www.baeldung.com/spring-mocking-webclient) [tutorials](https://rieckpil.de/test-spring-webclient-with-mockwebserver-from-okhttp/). Any opinion about that?

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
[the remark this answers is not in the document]
sschuberth: My bad 😆 Fixed!

utils/src/funTest/kotlin/storage/HttpFileStorageFunTest.kt
oheger-bosch: I had a short look at MockWebServer, and it indeed looks promising. My first impression is that the framework is pretty light-weight and can be easily used. WireMock currently offers more functionality, e.g. a DSL to stub responses for specific requests, matchers on request properties, JSON support, ...

oheger-bosch (approved):

utils/src/funTest/kotlin/storage/HttpFileStorageFunTest.kt
sschuberth: ... and pre-recording of real responses, right? While *MockWebServer* also has a *RecordedRequest* class, it seems to be something else.

PR #4022, opened 2021-05-14
NodeSupport: Avoid invalid logging on non-ORT classes

sschuberth:


oheger-bosch (approved):
An ugly work-around, but I do not have a better idea.

PR #4026, opened 2021-05-17
PackageManager: Improve log and issue messages

fviernau:
Make some log and issue messages unambiguous, to enable better resolutions and for debugging.

Note: the change may break the matching of existing issue resolutions.

### DIFF
--- a/analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
@@ -4321,10 +4321,10 @@ analyzer:
       'NPM::submodules/test-data-npm/long.js/package.json:':
       - timestamp: "1970-01-01T00:00:00Z"
         source: "NPM"
-        message: "Resolving dependencies for 'package.json' failed with: IllegalArgumentException:\
-          \ No lockfile found in 'submodules/test-data-npm/long.js'. This potentially\
-          \ results in unstable versions of dependencies. To allow this, enable support\
-          \ for dynamic versions."
+        message: "Resolving dependencies for 'submodules/test-data-npm/long.js/package.json'\
+          \ failed with: IllegalArgumentException: No lockfile found in 'submodules/test-data-npm/long.js'.\
+          \ This potentially results in unstable versions of dependencies. To allow\
+          \ this, enable support for dynamic versions."
         severity: "ERROR"
     has_issues: true
 scanner: null
--- a/analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
@@ -4321,7 +4321,7 @@ analyzer:
       'NPM::submodules/test-data-npm/long.js/package.json:':
       - timestamp: "1970-01-01T00:00:00Z"
         source: "NPM"
-        message: "Resolving dependencies for 'submodules/test-data-npm/long.js/package.json'\
+        message: "Resolving NPM dependencies for 'submodules/test-data-npm/long.js/package.json'\
           \ failed with: IllegalArgumentException: No lockfile found in 'submodules/test-data-npm/long.js'.\
           \ This potentially results in unstable versions of dependencies. To allow\
           \ this, enable support for dynamic versions."
--- a/analyzer/src/main/kotlin/PackageManager.kt
+++ b/analyzer/src/main/kotlin/PackageManager.kt
@@ -242,7 +242,7 @@ abstract class PackageManager(
                     val id = if (e is ProjectBuildingException && e.projectId?.isEmpty() == false) {
                         Identifier("Maven:${e.projectId}")
                     } else {
-                        val relativePath = definitionFile.absoluteFile.relativeTo(analysisRoot).invariantSeparatorsPath
+                        val relativePath = definitionFile.relativeTo(analysisRoot).invariantSeparatorsPath
                         Identifier.EMPTY.copy(type = managerName, name = relativePath)
                     }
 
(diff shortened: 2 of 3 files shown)


analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -255,7 +255,7 @@ abstract class PackageManager(
                     val errors = listOf(
                         createAndLogIssue(
                             source = managerName,
-                            message = "Resolving dependencies for '$definitionFile' failed with: " +
+                            message = "Resolving $managerName dependencies for '$definitionFile' failed with: " +
```
sschuberth, line 258: Commit message:

A word seems to be missing in "matching  with resolutions unambiguously". Or at least there's a double-space in there.

Also, isn't the package manager clear from the definition file? I.e. can you given an example where the same definition file is use by different package managers? Are you e.g. referring to `package.json` eventually being used by both NPM and Yarn?

sschuberth (changes requested):

analyzer/src/main/kotlin/PackageManager.kt
fviernau: > A word seems to be missing in "matching with resolutions unambiguously". Or at least there's a double-space in there.

Done.

> Also, isn't the package manager clear from the definition file? I.e. can you given an example where the same definition file is use by different package managers? Are you e.g. referring to package.json eventually being used by both NPM and Yarn?

`package.json` would be an example, yes. As package managers are plugin-able, it could be any definition file.

sschuberth: > As package managers are plugin-able, it could be any definition file.

That argument does not hold as it's ORT looking for the definition files here, not the package manager.

fviernau: > That argument does not hold as it's ORT looking for the definition files here, not the package manager.

Do you prefer to ditch the commit, or do you want a different commit message?

sschuberth: I'm undecided. I currently do not really see a value in this change, esp. as at least the issue source already contains the package manager name. If you'd like to keep the commit, please at least improve the commit messaging by naming the NPM vs Yarn example. Then it'd be fine with me.

sschuberth: One more thought: If dropping the last commit would get rid of the "breaking change" label (as issue resolutions do not need to be adjusted), I'd have a *slight* tendency towards dropping it.

fviernau: Ok great,I still have a preference towards keeping it. I've updated the commit message accordingly.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -163,6 +165,8 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
             inputDir, distinctPackageManagers, curationProvider, repositoryConfigurationFile
         ).mergeLabels(labels)
 
+        File("/home/frank/out.yaml").writeText(yamlMapper.writeValueAsString(ortResult))
```
sschuberth, line 168: Debug left-over?

sschuberth (changes requested):

analyzer/src/main/kotlin/PackageManager.kt
fviernau: > One more thought: If dropping the last commit would get rid of the "breaking change" label (as issue resolutions do not need to be adjusted), I'd have a slight tendency towards dropping it.

Already the first commit modifies the issue message, so the breaking change would remain even if dropping that commit IIUC.

sschuberth (dismissed):

mnonnenmacher (dismissed):

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -255,7 +257,8 @@ abstract class PackageManager(
                     val errors = listOf(
                         createAndLogIssue(
                             source = managerName,
-                            message = "Resolving dependencies for '${definitionFile.name}' failed with: " +
+                            message = "Resolving $managerName dependencies for " +
+                                    "'${definitionFile.relativeTo(analysisRoot)}' failed with: " +
```
sschuberth, line 261: I believe you also need `.invariantSeparatorsPath` here for the test on Windows to succeed.

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -257,8 +257,8 @@ abstract class PackageManager(
                     val errors = listOf(
                         createAndLogIssue(
                             source = managerName,
-                            message = "Resolving dependencies for '${definitionFile.name}' failed with: " +
-                                    e.collectMessagesAsString()
+                            message = "Resolving dependencies for '${definitionFile.relativeTo(analysisRoot)}' " +
```
sschuberth: I believe you also need `.invariantSeparatorsPath` here for the test on Windows to succeed.

sschuberth (approved):

analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
```diff
@@ -4321,10 +4321,10 @@ analyzer:
       'NPM::submodules/test-data-npm/long.js/package.json:':
       - timestamp: "1970-01-01T00:00:00Z"
         source: "NPM"
-        message: "Resolving dependencies for 'package.json' failed with: IllegalArgumentException:\
-          \ No lockfile found in 'submodules/test-data-npm/long.js'. This potentially\
-          \ results in unstable versions of dependencies. To allow this, enable support\
-          \ for dynamic versions."
+        message: "Resolving dependencies for 'submodules/test-data-npm/long.js/package.json'\
```
sschuberth: Commit message nit:

> I've not used

As usual, we try to avoid "I" / "we" etc. in commit messages and use passive voice like "The absolute path was not used ...". But as this is only a nit I've approved anyway.

PR #4029, opened 2021-05-17
GradleDependencyHandler: Make use of `takeIf` to simplify code a bit

fviernau:


analyzer/src/main/kotlin/managers/GradleDependencyHandler.kt
```diff
@@ -119,12 +119,8 @@ class GradleDependencyHandler(
...
  * Determine the [PackageLinkage] for this [Dependency].
  */
-private fun Dependency.linkage() =
-    if (isProjectDependency()) {
-        PackageLinkage.PROJECT_DYNAMIC
-    } else {
-        PackageLinkage.DYNAMIC
-    }
+private fun Dependency.linkage(): PackageLinkage =
+    PackageLinkage.PROJECT_DYNAMIC.takeIf { isProjectDependency() } ?: PackageLinkage.DYNAMIC
```
mnonnenmacher, line 123: What about this, to me classic `if/else` is easier to grasp:
```
if (isProjectDependency()) PackageLinkage.PROJECT_DYNAMIC else PackageLinkage.DYNAMIC
```

sschuberth: I think I'd also prefer the `if` / `else` one-liner. Personally, I tend to use `takeIf` /  `takeUnless` in cases where there is a clear default case, and an "exceptional" fallback. But here, the likeliness of coming across project or non-project linkage is probably the same. Or if at all, I'd rather swap the order:

    PackageLinkage.DYNAMIC.takeUnless { isProjectDependency() } ?: PackageLinkage.PROJECT_DYNAMIC

Also, it's a bit unusual IMO to use `takeIf` /  `takeUnless` in cases where you don't need to refer to `it`.

sschuberth:
Here's somewhat of a replacement for this PR: https://github.com/oss-review-toolkit/ort/pull/4031

PR #4030, opened 2021-05-17
analyzer: Fail gracefully when dependency graph contains invalid IDs

fviernau:
The dependency graph representation uses `String` as type for the
package identifiers. The package managers Maven and Gradle which use this
dependency graph representation do not parse the identifiers of the
packages into `Identifier` instances until right after the analyzer
result creation. Actually, that parsing happens when
`ortResult.getProjects()` is called by the analyzer command, see [1].

As consequence, the pre-condition checks implemented in the constructor
of `Identifier` are executed after the analyzer result is created. When
that check fails the analyzer command exits with an exception, e.g.
without any result.

Change the return type of `GradleDependencyHandler.identifierFor()` to
`Identifier` to execute these checks as early as possible. This in fact
turns that uncaught runtime exception into a gracefully handled
OrtIssue. It is a fix-up for [2].

Note that the problem was observed for a Gradle project with two levels
of nested submodules.

[1] ./cli/src/main/kotlin/commands/AnalyzerCommand.kt:166
[2] 5086c686019f3048802f6e66fa9d57b1658fd306

### DIFF
--- a/analyzer/src/main/kotlin/managers/GradleDependencyHandler.kt
+++ b/analyzer/src/main/kotlin/managers/GradleDependencyHandler.kt
@@ -28,6 +28,7 @@ import org.eclipse.aether.repository.RemoteRepository
 
 import org.ossreviewtoolkit.analyzer.managers.utils.DependencyHandler
 import org.ossreviewtoolkit.analyzer.managers.utils.MavenSupport
+import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtIssue
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageLinkage
@@ -53,8 +54,13 @@ class GradleDependencyHandler(
      */
     var repositories: List<RemoteRepository> = emptyList()
 
-    override fun identifierFor(dependency: Dependency): String =
-        "${dependency.dependencyType()}:${dependency.groupId}:${dependency.artifactId}:${dependency.version}"
+    override fun identifierFor(dependency: Dependency): Identifier =
+        Identifier(
+            type = dependency.dependencyType(),
+            namespace = dependency.groupId,
+            name = dependency.artifactId,
+            version = dependency.version
+        )
 
     override fun dependenciesFor(dependency: Dependency): Collection<Dependency> = dependency.dependencies
 
--- a/analyzer/src/main/kotlin/managers/MavenDependencyHandler.kt
+++ b/analyzer/src/main/kotlin/managers/MavenDependencyHandler.kt
@@ -56,16 +56,13 @@ class MavenDependencyHandler(
      */
     val sbtMode: Boolean
 ) : DependencyHandler<DependencyNode> {
-    override fun identifierFor(dependency: DependencyNode): Identifier {
-        val id = dependency.identifier()
-
-        return Identifier(
-            type = if (isLocalProject(id)) managerName else "Maven",
+    override fun identifierFor(dependency: DependencyNode): Identifier =
+        Identifier(
+            type = if (isLocalProject(dependency.identifier())) managerName else "Maven",
             namespace = dependency.artifact.groupId,
             name = dependency.artifact.artifactId,
             version = dependency.artifact.version
         )
-    }
 
     override fun dependenciesFor(dependency: DependencyNode): Collection<DependencyNode> {
         val childrenWithoutToolDependencies = dependency.children.filterNot { node ->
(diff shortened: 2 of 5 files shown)


analyzer/src/main/kotlin/managers/GradleDependencyHandler.kt
```diff
@@ -28,6 +28,7 @@ import org.eclipse.aether.repository.RemoteRepository
 
 import org.ossreviewtoolkit.analyzer.managers.utils.DependencyHandler
 import org.ossreviewtoolkit.analyzer.managers.utils.MavenSupport
+import org.ossreviewtoolkit.model.Identifier
```
sschuberth, line 31: Commit message:

> [1] ./cli/src/main/kotlin/commands/AnalyzerCommand.kt:166

Would be nice if this was a permalink, i.e. https://github.com/oss-review-toolkit/ort/blob/c596bc2b0f77e138ed9e6ee38dc55108e7c48fbb/cli/src/main/kotlin/commands/AnalyzerCommand.kt#L166

analyzer/src/main/kotlin/managers/MavenDependencyHandler.kt
```diff
@@ -55,10 +56,16 @@ class MavenDependencyHandler(
...
-    override fun identifierFor(dependency: DependencyNode): String {
+    override fun identifierFor(dependency: DependencyNode): Identifier {
         val id = dependency.identifier()
-        val type = if (isLocalProject(id)) managerName else "Maven"
-        return "$type:$id"
+        val parts = id.split(':')
+
+        return Identifier(
+            type = if (isLocalProject(id)) managerName else "Maven",
+            namespace = parts.getOrNull(0).orEmpty(),
```
sschuberth, line 65: Could be `getOrElse(0) { "" }` here and similar below.

analyzer/src/main/kotlin/managers/utils/DependencyGraphBuilder.kt
```diff
@@ -161,7 +161,7 @@ class DependencyGraphBuilder<D>(
     private fun addDependencyToGraph(scopeName: String, dependency: D, transitive: Boolean): DependencyReference {
         val identifier = dependencyHandler.identifierFor(dependency)
         val issues = dependencyHandler.issuesForDependency(dependency).toMutableList()
-        val index = updateDependencyMappingAndPackages(identifier, dependency, issues)
+        val index = updateDependencyMappingAndPackages(identifier.toString(), dependency, issues)
```
sschuberth, line 164: Should this really be `toString()`, or rather `toCoordinates()`? Similar below.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/MavenDependencyHandler.kt
```diff
@@ -55,10 +56,16 @@ class MavenDependencyHandler(
...
-        val type = if (isLocalProject(id)) managerName else "Maven"
-        return "$type:$id"
+        val parts = id.split(':')
+
+        return Identifier(
+            type = if (isLocalProject(id)) managerName else "Maven",
+            namespace = parts.getOrElse(0) { "" },
+            name = parts.getOrElse(1) { "" },
+            version = parts.getOrElse(2) { "" },
+        )
```
sschuberth, line 68: Sorry, looking at this again, there's no need to concat and split the `id` again, but we can just use

    override fun identifierFor(dependency: DependencyNode): Identifier {
        val id = dependency.identifier()

        return Identifier(
            type = if (isLocalProject(id)) managerName else "Maven",
            namespace = dependency.artifact.groupId,
            name = dependency.artifact.artifactId,
            version = dependency.artifact.version
        )
    }

sschuberth: As a follow-up, I'd personally also inline the two uses of the `DependencyNode.identifier()` convenience function.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/MavenDependencyHandler.kt
fviernau: I'll add your above suggestion to my follow-up PR.

oheger-bosch:
I like these changes. Just out of curiosity: What was the exact issue with the failing Gradle project you mentioned?

IIUC, this PR does not change the output of the analyzer, but makes the expected format of identifiers explicit. So were the coordinates in this project invalid?

oheger-bosch (approved):
The only minor problem I see is that there is now an asymmetry between the `identifierFor()` function and `createPackage()`, which is still passed a String as identifier. But this can be addressed later. After having implemented dependency handlers for multiple package managers, I think that this identifier parameter for `createPackage()` is not needed and thus could be removed at all. I may create a PR, which does this.

sschuberth (approved):

fviernau:
> What was the exact issue with the failing Gradle project you mentioned?

The issue was an identifier where the name part contained `:`. So more than three `:` in the identifier. This happens if nested submodules are used and there is an error in resolving that submodule. Then you get something like "Unknow:<project>:parent-module:child-module:"

> The only minor problem I see is that there is now an asymmetry between the `identifierFor()` function and `createPackage()`

I'm working on this already. I just wanted to get the bugfix out asap.

fviernau:
@oheger-bosch here is the follow-up for using the Identifier type in more places: #4035

PR #4034, opened 2021-05-18
NPM: Switch to dependency graph format

oheger-bosch:
This PR changes the NPM implementation to use the dependency graph format.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -32,6 +32,8 @@ import java.util.SortedSet
 
 import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
 import org.ossreviewtoolkit.analyzer.PackageManager
+import org.ossreviewtoolkit.analyzer.PackageManagerResult
+import org.ossreviewtoolkit.analyzer.managers.utils.DependencyGraphBuilder
 import org.ossreviewtoolkit.analyzer.managers.utils.expandNpmShortcutUrl
 import org.ossreviewtoolkit.analyzer.managers.utils.hasNpmLockFile
 import org.ossreviewtoolkit.analyzer.managers.utils.mapDefinitionFilesForNpm
@@ -40,6 +42,7 @@ import org.ossreviewtoolkit.analyzer.managers.utils.readRegistryFromNpmRc
 import org.ossreviewtoolkit.analyzer.parseAuthorString
 import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.downloader.VersionControlSystem
+import org.ossreviewtoolkit.model.DependencyGraph
 import org.ossreviewtoolkit.model.Hash
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
@@ -47,7 +50,6 @@ import org.ossreviewtoolkit.model.PackageReference
 import org.ossreviewtoolkit.model.Project
 import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.RemoteArtifact
-import org.ossreviewtoolkit.model.Scope
 import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
@@ -91,6 +93,15 @@ open class Npm(
     }
 
     companion object {
+        /** Name of the scope with the regular dependencies. */
+        private const val DEPENDENCIES_SCOPE = "dependencies"
+
+        /** Name of the scope with optional dependencies. */
+        private const val OPTIONAL_DEPENDENCIES_SCOPE = "optionalDependencies"
+
+        /** Name of the scope with development dependencies. */
+        private const val DEV_DEPENDENCIES_SCOPE = "devDependencies"
+
         /**
          * Parse information about licenses from the [package.json][json] file of a module.
          */
@@ -311,6 +322,8 @@ open class Npm(
         }
     }
 
+    private val graphBuilder = DependencyGraphBuilder(NpmDependencyHandler(npmRegistry))
+
     /**
      * Array of parameters passed to the install command when installing dependencies.
      */
@@ -334,6 +347,10 @@ open class Npm(
         ortProxySelector.removeProxyOrigin(managerName)
     }
 
+    override fun createPackageManagerResult(projectResults: Map<File, List<ProjectAnalyzerResult>>):
+            PackageManagerResult =
+        PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages().toSortedSet())
+
     override fun resolveDependencies(definitionFile: File): List<ProjectAnalyzerResult> {
         val workingDir = definitionFile.parentFile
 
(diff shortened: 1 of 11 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -90,6 +90,96 @@ open class Npm(
         ) = Npm(managerName, analysisRoot, analyzerConfig, repoConfig)
     }
 
+    companion object {
+        /**
+         * Parse information about licenses from the [package.json][json] file of a module.
+         */
+        internal fun parseLicenses(json: JsonNode): SortedSet<String> {
```
sschuberth, line 108: IIRC the outcome of a discussion in one of the last developer meetings was that we should only put public static functions into companion objects for discoverability, and move static function with narrower visibility to the top-level. Do I recall that correctly @fviernau?

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -171,99 +171,15 @@ open class Npm(
         }
 
         /**
-         * Split the given [rawName] of a module to a pair with namespace and name.
+         * Construct a [Package] by parsing its _package.json_ file and - if applicable - querying additional
```
sschuberth, line 185: Commit message nits: I noticed different "Npm" vs. "NPM" spellings in this and the previous commit message. If you refer to the class, it should be "Npm", and if you refer to the tool, it should be "NPM".

analyzer/src/main/kotlin/managers/NpmDependencyHandler.kt
```diff
@@ -0,0 +1,63 @@
...
+    val dependencies: Set<NpmModuleInfo>
+)
+
+/**
+ * A specialized [DependencyHandler] implementation for NPM.
+ */
+class NpmDependencyHandler(
+    /** The URI for queries to the NPM registry. */
+    val npmRegistryUri: String
+    ) : DependencyHandler<NpmModuleInfo> {
```
sschuberth, line 54: Indentation is off.

analyzer/src/main/kotlin/managers/utils/DependencyGraphBuilder.kt
```diff
@@ -118,8 +118,8 @@ class DependencyGraphBuilder<D>(
     /** The mapping from scopes to dependencies constructed by this builder. */
     private val scopeMapping = mutableMapOf<String, List<RootDependencyIndex>>()
 
-    /** Stores all packages encountered in the dependency tree. */
-    private val resolvedPackages = mutableSetOf<Package>()
+    /** Stores all packages encountered in the dependency tree keyed by their ID. */
```
sschuberth, line 121: Maybe "associated by" similar to the name of the Kotlin function?

analyzer/src/main/kotlin/managers/utils/DependencyGraphBuilder.kt
```diff
@@ -143,6 +143,16 @@ class DependencyGraphBuilder<D>(
...
 
+    /**
+     * Add the given [packages] to this builder. They are stored internally and also returned when querying the set of
+     * known packages. This function can be used by package managers that have to deal with packages not part of
+     * normal scope dependencies. One example would be Yarn; here packages can be defined in the workspace and are not
+     * necessarily referenced by manifest files.
+     */
+    fun addPackages(packages: Collection<Package>): DependencyGraphBuilder<D> = apply {
+        resolvedPackages += packages.associateBy { it.id }
+    }
```
sschuberth, line 154: Formatting should be like above.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Npm.kt
oheger-bosch: I remember the discussion, but am not exactly sure about the final conclusion. What I want to achieve here is a safe reuse of these functions between the `Npm` and the `NpmDependencyHandler` classes, but without polluting the package namespace. The functions have quite generic names like `parseLicenses()`; otherwise, they would have to be renamed to something more specific, e.g. `parseNpmLicenses()` - I prefer the approach with the companion object in this case.

analyzer/src/main/kotlin/managers/Npm.kt
oheger-bosch: I hope I have fixed all spelling problems. It is sometimes a bit ambiguous whether the class or the tool should be referenced.

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: I agree. So from my point of view this can stay as-is, even if it might differ a bit from the rule of thumb we agreed on.

analyzer/src/main/kotlin/managers/NpmDependencyHandler.kt
```diff
@@ -0,0 +1,63 @@
...
+    /** A set with information about the modules this module depends on. */
+    val dependencies: Set<NpmModuleInfo>
+)
+
+/**
+ * A specialized [DependencyHandler] implementation for NPM.
+ */
+class NpmDependencyHandler(
+    /** The URI for queries to the NPM registry. */
+    val npmRegistryUri: String
```
sschuberth, line 53: This could be private.

analyzer/src/main/kotlin/managers/NpmDependencyHandler.kt
```diff
@@ -0,0 +1,63 @@
...
+data class NpmModuleInfo(
+    /** The identifier for the represented module. */
+    val id: Identifier,
+
+    /** The file pointing to the package.json for this module. */
+    val packageFile: File,
+
+    /** A set with information about the modules this module depends on. */
+    val dependencies: Set<NpmModuleInfo>
+)
```
sschuberth, line 46: Just wondering, we now have this class in addition to `Npm`'s private `ModuleInfo` class. Do you see a change to consolidate these? Or make their names more distinct by renaming either one?

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/NpmDependencyHandler.kt
oheger-bosch: Consolidating these classes would be difficult, as there is a major difference in their content. I changed the following things:
* Renamed the private `ModuleInfo` class to `RawModuleInfo` and added a comment about its purpose.
* Renamed the `getModuleInfo()` function to `parsePackageJson()`.
* Renamed `getPackageReferenceForModule()` to `getModuleInfo()`; the old name was a remainder from the original implementation based on package references.

sschuberth: Thanks!

analyzer/src/main/kotlin/managers/NpmDependencyHandler.kt
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * A specialized [DependencyHandler] implementation for NPM.
+ */
+class NpmDependencyHandler(
+    /** The URI for queries to the NPM registry. */
+    private val npmRegistryUri: String
```
sschuberth, line 53: Sorry, one more thing at I saw just now: Shouldn't this be URL / ...Url instead of URI / ...Uri?

analyzer/src/test/kotlin/managers/NpmDependencyHandlerTest.kt
```diff
@@ -0,0 +1,103 @@
...
+            declaredLicenses should containExactly("MIT")
+            authors should containExactly("Thomas Watson Steen")
+            homepageUrl shouldBe "https://github.com/watson/bonjour/local"
+            description shouldBe "A Bonjour/Zeroconf implementation in pure JavaScript (local)"
+        }
+    }
+})
+
+/** The URI to be used for queries to the NPM registry. */
+const val NPM_REGISTRY_URI = "http://npm-test.example.org"
```
sschuberth, line 82: Same here, URL / ...Url instead of URI / ...Uri?

Also, even if just a dummy, I'd prefer to have "https" here instead of "http". This avoid false matches when searching the code base for places where still "http" is used.

sschuberth (changes requested):

sschuberth (approved):

sschuberth:
Merging despite the unrelated failure in `PipFunTest` which will get fixed by https://github.com/oss-review-toolkit/ort/pull/4049.

PR #4035, opened 2021-05-18
DependencyGraph: Consistently use Identifier as type for IDs

fviernau:
Change the Identifier type from String to Identifier in several classes related to the dependency graph and rename a few variables. See the individual commit messages.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GradleDependencyHandler.kt
+++ b/analyzer/src/main/kotlin/managers/GradleDependencyHandler.kt
@@ -86,7 +86,11 @@ class GradleDependencyHandler(
     override fun linkageFor(dependency: Dependency): PackageLinkage =
         if (dependency.isProjectDependency()) PackageLinkage.PROJECT_DYNAMIC else PackageLinkage.DYNAMIC
 
-    override fun createPackage(identifier: String, dependency: Dependency, issues: MutableList<OrtIssue>): Package? {
+    override fun createPackage(
+        identifier: Identifier,
+        dependency: Dependency,
+        issues: MutableList<OrtIssue>
+    ): Package? {
         // Only look for a package if there was no error resolving the dependency and it is no project dependency.
         if (dependency.error != null || dependency.isProjectDependency()) return null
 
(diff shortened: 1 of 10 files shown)


analyzer/src/main/kotlin/managers/utils/DependencyGraphBuilder.kt
```diff
@@ -149,7 +149,7 @@ class DependencyGraphBuilder<D>(
      * Construct the [DependencyGraph] from the dependencies passed to this builder so far.
      */
     fun build(): DependencyGraph = DependencyGraph(
-        dependencyIds.map { it.toCoordinates() },
+        dependencyIds,
```
sschuberth, line 152: Commit message: Double-space in "`String`  eventually".

sschuberth (dismissed):

analyzer/src/main/kotlin/managers/GradleDependencyHandler.kt
```diff
@@ -86,7 +86,11 @@ class GradleDependencyHandler(
     override fun linkageFor(dependency: Dependency): PackageLinkage =
         if (dependency.isProjectDependency()) PackageLinkage.PROJECT_DYNAMIC else PackageLinkage.DYNAMIC
 
-    override fun createPackage(identifier: String, dependency: Dependency, issues: MutableList<OrtIssue>): Package? {
+    override fun createPackage(
```
oheger-bosch, line 89: While you are at it, I think the `identifier` parameter can be completely dropped from `createPackage()`. It is not necessary, as the whole information is available via the `dependency` object. `GradleDependencyHandler` is currently the only implementation that accesses this parameter, but only for logging purposes; this could be solved differently.

oheger-bosch (approved):
The changes are fine with me. I have added one comment regarding the `id` parameter for `createPackage()`. I leave it to you whether you want to change this or not.

analyzer/src/main/kotlin/managers/GradleDependencyHandler.kt
fviernau: I saw this but wasn't sure...thanks for the hint! I'll file a follow up PR this afternoon.

fviernau: @oheger-bosch I've addressed it with #4037.

PR #4036, opened 2021-05-18
Remove the "all managers" test directory

sschuberth:


analyzer/src/test/kotlin/PackageManagerTest.kt
```diff
@@ -19,9 +19,11 @@
 
 package org.ossreviewtoolkit.analyzer
 
+import io.kotest.assertions.assertSoftly
```
mnonnenmacher, line 22: Commit message type: "dete**c**ting"
The commit message should explain the motivation for this change, because it makes the test case a lot longer. It slightly improves because before the test did not verify that we look into subdirectories.
As an alternative, to be able to skip empty definition files, couldn't we just make the file in all-managers non-empty to not couple the test with the package manager specific tests?

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -130,7 +130,16 @@ abstract class PackageManager(
                             // That is, at the example of a PIP project, if a directory contains all three files
                             // "requirements-py2.txt", "requirements-py3.txt" and "setup.py", only consider the
                             // former two as they match the glob with the highest priority, but ignore "setup.py".
-                            result.getOrPut(manager) { mutableListOf() } += matchesPerGlob.first()
+                            val priorityMatches = matchesPerGlob.first()
```
mnonnenmacher, line 133: We currently have the problem that we do no support Gradle projects which define multiple projects in a single `build.gradle`. The only workaround I know for such projects is to put empty `build.gradle` files into the directories of the subprojects. I didn't test it but I expect that this change would make the workaround impossible.

mnonnenmacher (changes requested):

analyzer/src/funTest/assets/projects/synthetic/stack-yesodweb-simple/.dir-locals.el
```diff
@@ -0,0 +1,4 @@
+((haskell-mode . ((haskell-indent-spaces . 4)
```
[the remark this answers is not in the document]
sschuberth: My original plan was to replace (bit by bit) "real" projects that we incorporate via Git submodules with synthetic projects that are fully under our control. With that in mind, synthetic projects should actually not necessarily be as small as possible, but sufficiently large to cover the wanted use-cases. Also, to me it was attractive to use a project that can be auto-generated, which might simplify any future updates when the package manager changes its implementation.

analyzer/src/test/kotlin/PackageManagerTest.kt
sschuberth: > make the file in all-managers non-empty

I'm really not a fan of those fake-files, which we wouldn't expect the analyzer to parse correctly.

> to not couple the test with the package manager specific tests?

Actually, I found exactly that reuse elegant...

analyzer/src/main/kotlin/PackageManager.kt
sschuberth: Wouldn't it be good practice to put a comment into those Gradle files then anyway, might makes them non-empty, and it would be working again?

analyzer/src/test/kotlin/PackageManagerTest.kt
mnonnenmacher: > > make the file in all-managers non-empty
> 
> I'm really not a fan of those fake-files, which we wouldn't expect the analyzer to parse correctly.

If the files exist only to test the detection of definition files I see absolutely no problem with that.

> > to not couple the test with the package manager specific tests?
> 
> Actually, I found exactly that reuse elegant...

What I don't like about the coupling is that it connects test cases which should not be connected because they test different parts of the code. For example, if we add a new test project for Gradle we have to update this test as well, even if the new file does not improve the test for detection of the definition files at all. And the other way round, the test projects are not necessarily structured in way that tests all aspects of the detection algorithm we might want to test.

Maybe a third opinion would help here, @oheger-bosch @fviernau ?

analyzer/src/funTest/assets/projects/synthetic/stack-yesodweb-simple/.dir-locals.el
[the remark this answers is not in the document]
mnonnenmacher: > With that in mind, synthetic projects should actually not necessarily be as small as possible, but sufficiently large to cover the wanted use-cases.

That's actually what I meant. However, this test project seems to be a lot larger than required.

> Also, to me it was attractive to use a project that can be auto-generated, which might simplify any future updates when the package manager changes its implementation.

That's a good argument for me to have additional files which are not required for the test. I would like if for such cases we add a `README.md` to the project directory that explains how the project was generated and how it can be updated.

analyzer/src/test/kotlin/PackageManagerTest.kt
oheger-bosch: I tend to agree with Martin's argument about the coupling; unrelated changes in the tests of specific package managers could now break this test.

Personally, I am more a fan of unit tests. So I would probably have tested the algorithm to detect definition files on a temporary folder structure, in which some definition files have been generated. The test could use a couple of stub PackageManagerFactories to test some corner cases such as multiple globs for definition files are handled, sub directories are processed, etc.

For this it is not necessary to test with all supported package managers; in the tests for specific managers, it should then be verified that the expected definition file globs are correct.

analyzer/src/funTest/assets/projects/synthetic/stack-yesodweb-simple/.dir-locals.el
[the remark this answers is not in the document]
sschuberth: > That's a good argument for me to have additional files which are not required for the test. I would like if for such cases we add a `README.md` to the project directory that explains how the project was generated and how it can be updated.

There already is a `README.md` being added as part of the project generation, so there would be a name clash. Isn't the explanation in the Git commit message about how the projected was generated good enough?

analyzer/src/test/kotlin/PackageManagerTest.kt
sschuberth: > As an alternative, to be able to skip empty definition files, couldn't we just make the file in all-managers non-empty to not couple the test with the package manager specific tests?

Simply making them non-empty would not work, unfortunately; they'd also need to contain valid syntax in most cases. Otherwise parsing e.g. with Jackson would fail.

sschuberth: > If the files exist only to test the detection of definition files I see absolutely no problem with that.

The problem with these files actually occurs when scanning ORT with ORT, which makes the analyzer fail are empty files cannot be parsed properly. Thus, I'd like to simply skip empty files, but that would make the original test fail.

> So I would probably have tested the algorithm to detect definition files on a temporary folder structure, in which some definition files have been generated.

I like the idea of generating the definition files in a temporary folder structure. However, also those files would need to be non-empty and syntactically valid *if* we simply run the analyzer on the folder structure.

But as the original test only intended to test the discovery of the definition files, not their analysis, I agree with @oheger-bosch that we probably should come up with unit tests for the globs.

Anyway, I'd appreciate if we could merge this as an intermediate step into the right direction of getting rid of the commit empty definition files, and to be able skip empty definition files.

mnonnenmacher: > I like the idea of generating the definition files in a temporary folder structure. However, also those files would need to be non-empty and syntactically valid if we simply run the analyzer on the folder structure.

This would be a replacement for the test of `PackageManager.findManagedFiles()` so the generated files could be empty. This approach could make `PackageManagerTest` a lot smaller because it would not have to contain the large list of files from our test project directory.

> Anyway, I'd appreciate if we could merge this as an intermediate step into the right direction of getting rid of the commit empty definition files, and to be able skip empty definition files.

What is the motivation for skipping empty definition files? Do you have issues with empty definition files except for the ones removed in this PR? Are we sure that across all supported package managers there are no cases where empty definition files are valid?

analyzer/src/funTest/assets/projects/synthetic/stack-yesodweb-simple/.dir-locals.el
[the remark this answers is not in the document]
mnonnenmacher: Yes, I think we won't be doing many manual changes to the test project so it should be easy enough to find.

analyzer/src/test/kotlin/PackageManagerTest.kt
sschuberth: > What is the motivation for skipping empty definition files?

See the [commit message](https://github.com/oss-review-toolkit/ort/pull/4036/commits/338415e57b2e82356f5a6aed74bc8d4dac41f7cd): "This change is a preparation for skipping pathologic cases of empty definition files, which otherwise would cause parsing errors with several third-party libraries that we use, like Jackson."

> Do you have issues with empty definition files except for the ones removed in this PR?

Yes. Empty definition files in a real project made ORT crash in production. But I could not simply ignore empty definition files until now as the mentioned test depends on finding empty definition files.

> Are we sure that across all supported package managers there are no cases where empty definition files are valid?

Even if an upstream package manager would accept (i.e. not crash when reading them) empty definition files, they would not contain any information, so it should be fine to skip them in any case.

mnonnenmacher: > Yes. Empty definition files in a real project made ORT crash in production.

This sound like there is a try/catch missing somewhere in `PackageManager` or `AnalyzerCommand` that would prevent ORT from crashing. This would be useful also for other cases where files are unparsable.

> Even if an upstream package manager would accept (i.e. not crash when reading them) empty definition files, they would not contain any information, so it should be fine to skip them in any case.

Gradle is the one exception I know. I think I would have preferred a solution where at least a warning is added to the ORT result instead of silently ignoring those files, but I'm not going to block this now.

mnonnenmacher (approved):

analyzer/src/test/kotlin/PackageManagerTest.kt
sschuberth: > This sounds like there is a try/catch missing somewhere in `PackageManager` or `AnalyzerCommand` that would prevent ORT from crashing.

The term "crash" might have been wrong... sorry this happened too long ago. Maybe the real problem was that the exception *was* catched, but then showed up as an issue in the analyzer result, that we needed / wanted to get rid of. And while investigating it occurred to me that the issue should not be filed in the first place for empty files.

> Gradle is the one exception I know.

So what does Gradle do with empty definition files? Does it apply all defaults that indeed result in a valid project definition?

> I think I would have preferred a solution where at least a warning is added to the ORT result instead of silently ignoring those files

I should be able to add that later.

sschuberth: > I like the idea of generating the definition files in a temporary folder structure. However, also those files would need to be non-empty and syntactically valid _if_ we simply run the analyzer on the folder structure.

See the follow-up in https://github.com/oss-review-toolkit/ort/pull/4330. A correction to my statement above, though: While the temporarily generated definition files need to be non-empty (due to [this](https://github.com/oss-review-toolkit/ort/pull/4036/commits/a87da6e872871b2ec99bdb9bc03f72082d61f58f)), they do not need to be syntactically valid.

PR #4024, opened 2021-05-15
Support long paths when running Git on Windows

sschuberth:
Please have a look at the individual commit messages for the details.

PR #3894, opened 2021-04-15
Maven: Switch to the dependency graph format

oheger-bosch:
This PR is related to #3825. It converts the package manager implementation for Maven to the dependency graph format.

PR #4039, opened 2021-05-19
downloader: Return a Result object from updateWorkingTree()

### LINKED ISSUE #3952: Provide more information when VCS download fails
The [`updateWorkingTree`](https://github.com/oss-review-toolkit/ort/blob/ce30534806361486097700fb0c0fbab611a2d239/downloader/src/main/kotlin/VersionControlSystem.kt#L308) function only returns a `Boolean` to describe if the update was successful or not. If it fails, the actual errors get only written to the logs. For example, if Git fails to checkout a revision because it does not exist, this error is written to the logs, but will not be visible in the result files and reports, these will only contain a generic error message.

This makes the investigation of VCS download issues unnecessarily hard. It could be improved by either throwing an exception in `updateWorking` that provides more details in the message, or by returning a result object with more information instead of a `Boolean`.

sschuberth:
Instead of just a Boolean, return a Result object encapsulating the
originally requested revsion for convenience, or any occurred exception.
This eases the debugging of download failures.

Resolves #3952.

### DIFF
--- a/downloader/src/main/kotlin/VersionControlSystem.kt
+++ b/downloader/src/main/kotlin/VersionControlSystem.kt
@@ -220,7 +220,9 @@ abstract class VersionControlSystem {
         // revision candidates instead of a single revision.
         val revisionCandidates = mutableSetOf<String>()
 
-        try {
+        val emptyRevisionCandidatesException = DownloadException("Unable to determine a revision to checkout.")
+
+        runCatching {
             pkg.vcsProcessed.revision.also {
                 if (it.isNotBlank() && (allowMovingRevisions || isFixedRevision(workingTree, it))) {
                     if (revisionCandidates.add(it)) {
@@ -230,16 +232,18 @@ abstract class VersionControlSystem {
                     }
                 }
             }
-        } catch (e: IOException) {
-            e.showStackTrace()
+        }.onFailure {
+            it.showStackTrace()
 
             log.info {
-                "Meta-data has invalid $type revision '${pkg.vcsProcessed.revision}': ${e.collectMessagesAsString()}"
+                "Meta-data has invalid $type revision '${pkg.vcsProcessed.revision}': ${it.collectMessagesAsString()}"
             }
+
+            emptyRevisionCandidatesException.addSuppressed(it)
         }
 
         fun addGuessedRevision(project: String, version: String): Boolean =
-            try {
+            runCatching {
                 workingTree.guessRevisionName(project, version).also {
                     if (revisionCandidates.add(it)) {
                         log.info {
@@ -248,18 +252,16 @@ abstract class VersionControlSystem {
                         }
                     }
                 }
-
-                true
-            } catch (e: IOException) {
-                e.showStackTrace()
+            }.onFailure {
+                it.showStackTrace()
 
                 log.info {
                     "No $type revision for package '$project' and version '$version' found: " +
-                            e.collectMessagesAsString()
+                            it.collectMessagesAsString()
                 }
 
-                false
-            }
+                emptyRevisionCandidatesException.addSuppressed(it)
+            }.isSuccess
 
         if (!addGuessedRevision(pkg.id.name, pkg.id.version)) {
             when {
@@ -279,9 +281,7 @@ abstract class VersionControlSystem {
             }
         }
 
-        if (revisionCandidates.isEmpty()) {
-            throw DownloadException("Unable to determine a revision to checkout.")
-        }
+        if (revisionCandidates.isEmpty()) throw emptyRevisionCandidatesException
 
         val results = mutableListOf<Result<String>>()
 
--- a/downloader/src/main/kotlin/VersionControlSystem.kt
+++ b/downloader/src/main/kotlin/VersionControlSystem.kt
@@ -283,11 +283,17 @@ abstract class VersionControlSystem {
             throw DownloadException("Unable to determine a revision to checkout.")
         }
 
-        var i = 0
-        val workingTreeRevision = revisionCandidates.find { revision ->
-            log.info { "Trying revision candidate '$revision' (${++i} of ${revisionCandidates.size})..." }
-            updateWorkingTree(workingTree, revision, pkg.vcsProcessed.path, recursive)
-        } ?: throw DownloadException("$type failed to download from URL '${pkg.vcsProcessed.url}'.")
+        val results = mutableListOf<Result<String>>()
+
+        revisionCandidates.forEachIndexed { index, revision ->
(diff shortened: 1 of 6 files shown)


downloader/src/main/kotlin/VersionControlSystem.kt
```diff
@@ -283,11 +283,17 @@ abstract class VersionControlSystem {
...
-        } ?: throw DownloadException("$type failed to download from URL '${pkg.vcsProcessed.url}'.")
+        val results = mutableListOf<Result<String>>()
+
+        revisionCandidates.forEachIndexed { index, revision ->
+            log.info { "Trying revision candidate '$revision' (${index + 1} of ${revisionCandidates.size})..." }
+            results += updateWorkingTree(workingTree, revision, pkg.vcsProcessed.path, recursive)
+            if (results.last().isSuccess) return@forEachIndexed
+        }
+
+        val workingTreeRevision = results.last().getOrElse {
```
fviernau, line 294: I believe that in case `results.last() is Failure == true` the handling is now missing and needs to be added.

downloader/src/main/kotlin/vcs/GitRepo.kt
```diff
@@ -189,17 +189,15 @@ class GitRepo : VersionControlSystem(), CommandLineTool {
             runRepo(workingTree.workingDir, *syncArgs.toTypedArray())
 
             log.debug { runRepo(workingTree.workingDir, "info").stdout }
-
-            true
-        } catch (e: IOException) {
-            e.showStackTrace()
+        }.onFailure {
+            it.showStackTrace()
```
fviernau, line 193: Consider renaming `it` to `e`, same above.

downloader/src/main/kotlin/vcs/Subversion.kt
```diff
@@ -226,15 +226,15 @@ class Subversion : VersionControlSystem() {
             }
 
             true
-        } catch (e: SVNException) {
-            e.showStackTrace()
+        }.onFailure {
+            it.showStackTrace()
```
fviernau, line 230: commit-msg: Would it make sense to mention that the DownloadExceptions are not changed?

fviernau (changes requested):

downloader/src/main/kotlin/vcs/GitRepo.kt
sschuberth: I actually thought about this when @oheger-bosch did a [similar change](https://github.com/oss-review-toolkit/ort/pull/3894/files#diff-439602bba9d79021802c46707b56c056fdb9f49373dac17ba938cdc4552605cdR88), but in exactly the opposite way: I tend to prefer `it` for rather simple code (esp. one-liners) where there's no clash with another (nested) `it`. For the special case of `runCatching` I tend to only use a named lambda parameter for `onFailure` if also done so for an `onSuccess`.

I *do* see a small point here to use `e` to keep the diff a bit smaller, but then again `e` is not much more telling than `it` anyway (`exception` would have been more telling). So I chose to stick to `it`.

But I'm open to get overruled here. Any opinion from @oheger-bosch?

downloader/src/main/kotlin/vcs/Subversion.kt
sschuberth: I'm not exactly sure what you're after, but doesn't "or any occurred exception" imply that?

downloader/src/main/kotlin/VersionControlSystem.kt
sschuberth: The "OrElse" part is what gets executed in that case.

fviernau: If result contains more than zero elements and all elements are of the type `Failure`, then, you say, `orElse` gets executed?

downloader/src/main/kotlin/vcs/GitRepo.kt
fviernau: I'm ok with `it`.

downloader/src/main/kotlin/vcs/Subversion.kt
fviernau: Right.

downloader/src/main/kotlin/VersionControlSystem.kt
sschuberth: Yes. But only the message of the last failure is included into the `DownloadException`.

What probably confuses you is that the `getOrElse` here is not the function on a collection, but the function on a `Result` object.

sschuberth: And `results` cannot be empty as `revisionCandidates` is ensured to not be empty.

fviernau: > What probably confuses you is that the getOrElse here is not the function on a collection, but the function on a Result object.

Argh, yes that was exactly what was confusing me. Thanks for pointing it out.

fviernau (approved):

downloader/src/main/kotlin/vcs/GitRepo.kt
oheger-bosch: IMHO we should not be too dogmatic, but give the developers a certain freedom to follow their preferred style. I am fine with both.

downloader/src/main/kotlin/VersionControlSystem.kt
```diff
@@ -283,11 +283,17 @@ abstract class VersionControlSystem {
...
+            throw DownloadException("$type failed to download from URL '${pkg.vcsProcessed.url}'.", it)
```
mnonnenmacher, line 295: In the `if (revisionCandidates.isEmpty())` case we still don't propagate the exceptions that could have happened before (lines 233 and 253). I think it would be good to add them to the exception thrown in line 283 to completely address #3952 . Also I wonder if it is sufficient to figure out the problem to only propagate the last exception. But it probably is, at least I have no idea currently in which situation it wouldn't be.

mnonnenmacher (changes requested):

downloader/src/main/kotlin/VersionControlSystem.kt
sschuberth: Issue #3952 specifically talks about `updateWorkingTree`, so that's what I was looking at, and the current changes already properly provide the "too long path" root cause I'm looking at in #4024. But let's see what I can do about the other exceptions...

mnonnenmacher: Thanks.

mnonnenmacher (approved):

PR #4045, opened 2021-05-19
fossid-webapp: use custom Jackson deserializer

nnobelis:
The FossID server is not consistent. It usually returns a List or a Map,
but if there is no entry, it returns data:false.
Hence this commit wires a custom Jackson deserializer that abstracts
this behavior.
It also removes the previous toList() function who attempted at doing
this abstraction.

### DIFF
--- a/clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
+++ b/clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
@@ -21,15 +21,31 @@
 
 package org.ossreviewtoolkit.clients.fossid
 
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.core.JsonToken
+import com.fasterxml.jackson.databind.BeanDescription
+import com.fasterxml.jackson.databind.DeserializationConfig
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonDeserializer
 import com.fasterxml.jackson.databind.MapperFeature
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.PropertyNamingStrategies
+import com.fasterxml.jackson.databind.deser.BeanDeserializerModifier
+import com.fasterxml.jackson.databind.deser.std.DelegatingDeserializer
 import com.fasterxml.jackson.databind.json.JsonMapper
-import com.fasterxml.jackson.module.kotlin.registerKotlinModule
+import com.fasterxml.jackson.databind.type.CollectionType
+import com.fasterxml.jackson.databind.type.MapType
+import com.fasterxml.jackson.module.kotlin.kotlinModule
 
 import okhttp3.OkHttpClient
 import okhttp3.ResponseBody
 
 import org.ossreviewtoolkit.clients.fossid.model.Project
+import org.ossreviewtoolkit.clients.fossid.model.Scan
+import org.ossreviewtoolkit.clients.fossid.model.identification.identifiedFiles.IdentifiedFile
+import org.ossreviewtoolkit.clients.fossid.model.identification.ignored.IgnoredFile
+import org.ossreviewtoolkit.clients.fossid.model.identification.markedAsIdentified.MarkedAsIdentifiedFile
+import org.ossreviewtoolkit.clients.fossid.model.result.FossIdScanResult
 import org.ossreviewtoolkit.clients.fossid.model.status.DownloadStatus
 import org.ossreviewtoolkit.clients.fossid.model.status.ScanStatus
 
@@ -44,12 +60,50 @@ interface FossIdRestService {
         /**
          * The mapper for JSON (de-)serialization used by this service.
          */
-        val JSON_MAPPER = JsonMapper()
+        val JSON_MAPPER: ObjectMapper = JsonMapper()
             .setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE)
             // FossID has a bug in get_results/scan.
             // Sometimes the match_type is "ignored", sometimes it is "Ignored".
             .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
-            .registerKotlinModule()
+            .registerModule(kotlinModule().setDeserializerModifier(FossIdDeserializerModifier()))
+
+        /**
+         * A class to modify the standard Jackson deserialization to deal with inconsistencies in responses
+         * sent by the FossID server.
+         * FossID usually returns data as a List or Map, but in case of no entries it returns a Boolean (which is set to
+         * false). To handle this special case, this class makes sure that a custom deserializer is used for lists or
+         * maps that detects such a result and converts it to an empty list or map, respective. That way, the FossID
+         * service interface can use meaningful result types.
+         */
+        private class FossIdDeserializerModifier : BeanDeserializerModifier() {
+            override fun modifyMapDeserializer(
+                config: DeserializationConfig,
+                type: MapType,
+                beanDesc: BeanDescription,
+                deserializer: JsonDeserializer<*>
+            ): JsonDeserializer<*> = FalseValueDeserializer(deserializer) { mutableMapOf<Any, Any>() }
+
+            override fun modifyCollectionDeserializer(
+                config: DeserializationConfig?,
+                type: CollectionType?,
+                beanDesc: BeanDescription?,
+                deserializer: JsonDeserializer<*>
+            ): JsonDeserializer<*> = FalseValueDeserializer(deserializer) { mutableListOf<Any>() }
+        }
+
+        /**
+         * Custom Jackson deserializer that abstracts the behaviour explained above. [creator] is a function that
+         * creates the 'replacement' object when a Boolean value set to false is received. When not, the standard
+         * Kotlin/Jackson deserializer is called by delegation.
+         */
+        private class FalseValueDeserializer(delegate: JsonDeserializer<*>, private val creator: () -> Any) :
+            DelegatingDeserializer(delegate) {
+            override fun deserialize(p: JsonParser, ctxt: DeserializationContext): Any =
+                if (p.currentToken == JsonToken.VALUE_FALSE) creator() else super.deserialize(p, ctxt)
+
+            override fun newDelegatingInstance(newDelegatee: JsonDeserializer<*>): JsonDeserializer<*> =
+                FalseValueDeserializer(newDelegatee, creator)
+        }
 
         /**
          * Create a FossID service instance for communicating with a server running at the given [url],
@@ -70,7 +124,7 @@ interface FossIdRestService {
     suspend fun getProject(@Body body: PostRequestBody): EntityResponseBody<Project>
 
     @POST("api.php")
-    suspend fun listScansForProject(@Body body: PostRequestBody): EntityResponseBody<Any>
+    suspend fun listScansForProject(@Body body: PostRequestBody): MapResponseBody<Scan>
 
     @POST("api.php")
     suspend fun createProject(@Body body: PostRequestBody): MapResponseBody<String>
@@ -94,19 +148,19 @@ interface FossIdRestService {
     suspend fun checkScanStatus(@Body body: PostRequestBody): EntityResponseBody<ScanStatus>
 
     @POST("api.php")
-    suspend fun listScanResults(@Body body: PostRequestBody): EntityResponseBody<Any>
+    suspend fun listScanResults(@Body body: PostRequestBody): MapResponseBody<FossIdScanResult>
 
     @POST("api.php")
-    suspend fun listIdentifiedFiles(@Body body: PostRequestBody): EntityResponseBody<Any>
+    suspend fun listIdentifiedFiles(@Body body: PostRequestBody): MapResponseBody<IdentifiedFile>
 
     @POST("api.php")
-    suspend fun listMarkedAsIdentifiedFiles(@Body body: PostRequestBody): EntityResponseBody<Any>
+    suspend fun listMarkedAsIdentifiedFiles(@Body body: PostRequestBody): MapResponseBody<MarkedAsIdentifiedFile>
 
     @POST("api.php")
-    suspend fun listIgnoredFiles(@Body body: PostRequestBody): EntityResponseBody<Any>
+    suspend fun listIgnoredFiles(@Body body: PostRequestBody): EntityResponseBody<List<IgnoredFile>>
 
     @POST("api.php")
-    suspend fun listPendingFiles(@Body body: PostRequestBody): EntityResponseBody<Any>
+    suspend fun listPendingFiles(@Body body: PostRequestBody): MapResponseBody<String>
 
     @GET("index.php?form=login")
     suspend fun getLoginPage(): ResponseBody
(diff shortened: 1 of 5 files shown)


clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -44,12 +60,58 @@ interface FossIdRestService {
...
+            .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS).copy()
+            .registerModule(
+                kotlinModule()
+                    .setDeserializerModifier(FossIdDeserializerModifier())
+            )
+
+        /**
+         * The FossID server is not consistent. It usually returns a List or a Map, but if there is no
+         * entry, it returns data:false.
+         * We register two custom deserializers to address this issue.
```
oheger-bosch, line 76: I think the comment of this class could be improved. I would use something like this:

"A class to modify the standard Jackson deserialization to deal with inconsistencies in responses sent by the FossID server.

FossID usually returns a List or a Map, but if there is no entry, it returns data:false. To handle this special case, this class makes sure that a custom deserializer is used for lists or maps that detects such a result and converts it to an empty list or map, respective. That way, the FossID service interface can use meaningful result types."

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
```diff
@@ -107,8 +101,8 @@ class FossIdClientNewProjectTest : StringSpec({
 
     "Scans for project can be listed when there is no scan" {
         service.listScansForProject("", "", PROJECT_CODE) shouldNotBeNull {
-            checkResponse("list scans")
-            toList<Scan>() should beEmpty()
+            checkResponse("list scans", true)
+            data!!.values should beEmpty()
```
oheger-bosch, line 105: It is a bit ugly to have these "!!" everywhere. Could this be circumvented somehow, maybe by having another helper function `checkNonNullResponse()`, which returns a non-nullable object?

oheger-bosch (commented):
This new deserialization is indeed a huge improvement.

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
mnonnenmacher: Or maybe just be a bit more verbose and use kotest functions, e.g.
```
data shouldNotBeNull { values should beEmpty() }
```
`with(data!!)` and `data!!.should` could be replaced with this, and `with(data!!.values)` could be replaced with `data shouldNotBeNull { with(data.vlues) { ... } }`.

nnobelis: Fixed thanks for the hint.
Just for the record, these `!!` were not unsafe because `data` was never `null`: the previous call in `checkResponse` made sure of that.
However, using `shouldNotBeNull` as suggested makes it more straightforward to the reader.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -44,12 +60,61 @@ interface FossIdRestService {
...
+            ): JsonDeserializer<*> {
+                return LenientDeserializer(deserializer) { mutableListOf<Any>() }
+            }
+        }
+
+        /** Custom Jackson deserializer that abstracts the behaviour explained above.
+         * [creator] is a function that creates an empty object when data:false is received.
+         * Otherwise, the standard Kotlin/Jackson deserializer is called by delegation.
+         */
+        private class LenientDeserializer(delegate: JsonDeserializer<*>, val creator: () -> Any) :
```
sschuberth, line 105: The name `LenientDeserializer` doesn't really explain in what way the deserializer is "lenient". I propose something more telling like e.g. "FalseValueDeserializer".

sschuberth: `val creator` could be private.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -44,12 +60,61 @@ interface FossIdRestService {
...
+            DelegatingDeserializer(delegate) {
+            override fun deserialize(p: JsonParser, ctxt: DeserializationContext): Any {
+                if (p.currentToken == JsonToken.VALUE_FALSE) {
+                    return creator()
+                }
```
sschuberth, line 110: Could be a one-liner.

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
```diff
@@ -107,8 +101,10 @@ class FossIdClientNewProjectTest : StringSpec({
 
     "Scans for project can be listed when there is no scan" {
         service.listScansForProject("", "", PROJECT_CODE) shouldNotBeNull {
-            checkResponse("list scans")
-            toList<Scan>() should beEmpty()
+            checkResponse("list scans", true)
```
sschuberth, line 104: `true` is the default for the `withDataCheck` parameter and should be omitted here and in several other places.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -70,7 +135,7 @@ interface FossIdRestService {
     suspend fun getProject(@Body body: PostRequestBody): EntityResponseBody<Project>
 
     @POST("api.php")
-    suspend fun listScansForProject(@Body body: PostRequestBody): EntityResponseBody<Any>
+    suspend fun listScansForProject(@Body body: PostRequestBody): EntityResponseBody<Map<String, Scan>>
```
sschuberth, line 138: Why is this not `MapResponseBody<Scan>`? Similar in several places below.

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
```diff
@@ -107,8 +101,10 @@ class FossIdClientNewProjectTest : StringSpec({
...
+            data shouldNotBeNull {
+                values should beEmpty()
```
sschuberth, line 106: IMO it's a bit odd to only check the values here (even if it implies that the keys, and therefore the whole map, is empty, too). I'd prefer `this should beEmpty()`.

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
```diff
@@ -196,30 +194,34 @@ class FossIdClientNewProjectTest : StringSpec({
     "Marked files can be listed when there are none" {
         service.listMarkedAsIdentifiedFiles("", "", SCAN_CODE) shouldNotBeNull {
             checkResponse("list marked as identified files")
-            toList<MarkedAsIdentifiedFile>() should beEmpty()
+            data shouldNotBeNull {
+                values should beEmpty()
```
sschuberth, line 198: Again `this should beEmpty()`.

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
```diff
@@ -196,30 +194,34 @@ class FossIdClientNewProjectTest : StringSpec({
...
             checkResponse("list pending files")
-
-            val files = toList<String>()
-            files.size shouldBe 2
-            files.first() shouldBe "src/extra_file.txt"
+            data shouldNotBeNull {
+                values.should {
+                it.size shouldBe 2
+                it.first() shouldBe "src/extra_file.txt"
+            }
```
sschuberth, line 223: Indentation is off.

clients/fossid-webapp/src/test/kotlin/FossIdClientReturnTypeTest.kt
```diff
@@ -80,85 +71,109 @@ class FossIdClientReturnTypeTest : StringSpec({
...
+            }
         }
     }
 
     "Marked as identified  files can be listed when there is some" {
         service.listMarkedAsIdentifiedFiles("", "", SCAN_CODE_2) shouldNotBeNull {
             checkResponse("list marked as identified files")
-            toList<MarkedAsIdentifiedFile>() shouldNot beEmpty()
+            data shouldNotBeNull {
+                values shouldNot beEmpty()
```
sschuberth, line 139: Hint: To avoid a name clash of `beEmpty()` for collections (see below) and maps (here), you could use

    import io.kotest.matchers.maps.beEmpty as beEmptyMap

and then

    this should beEmptyMap()

here.

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -44,12 +60,61 @@ interface FossIdRestService {
...
          * The mapper for JSON (de-)serialization used by this service.
          */
-        val JSON_MAPPER = JsonMapper()
+        val JSON_MAPPER: ObjectMapper = JsonMapper()
             .setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE)
             // FossID has a bug in get_results/scan.
             // Sometimes the match_type is "ignored", sometimes it is "Ignored".
-            .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
-            .registerKotlinModule()
+            .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS).copy()
```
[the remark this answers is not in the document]
nnobelis: This was a remnant from previous implementation, forgotten.

clients/fossid-webapp/src/test/kotlin/FossIdClientReturnTypeTest.kt
nnobelis: Yes ! This was the bit I was missing and the reason I checked only the values.

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
```diff
@@ -156,7 +152,7 @@ class FossIdClientNewProjectTest : StringSpec({
 
     "Scan status can be queried" {
         service.checkScanStatus("", "", SCAN_CODE) shouldNotBeNull {
-            checkResponse("get scan status", false)
+            checkResponse("get scan status")
```
sschuberth, line 155: This now changes behavior as `false` (which is not the default) is omitted. Is this intended?

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -44,12 +60,58 @@ interface FossIdRestService {
...
-        val JSON_MAPPER = JsonMapper()
+        val JSON_MAPPER: ObjectMapper = JsonMapper()
             .setPropertyNamingStrategy(PropertyNamingStrategies.SNAKE_CASE)
             // FossID has a bug in get_results/scan.
             // Sometimes the match_type is "ignored", sometimes it is "Ignored".
             .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
-            .registerKotlinModule()
+            .registerModule(
+                kotlinModule().setDeserializerModifier(FossIdDeserializerModifier())
+            )
```
sschuberth, line 70: This still fits one line 😉

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -44,12 +60,58 @@ interface FossIdRestService {
...
+         */
+        private class FossIdDeserializerModifier : BeanDeserializerModifier() {
+            override fun modifyMapDeserializer(
+                config: DeserializationConfig,
+                type: MapType,
+                beanDesc: BeanDescription,
+                deserializer: JsonDeserializer<*>
+            ): JsonDeserializer<*> {
+                return FalseValueDeserializer(deserializer) { mutableMapOf<Any, Any>() }
+            }
```
sschuberth, line 88: Nit: Could be

            ): JsonDeserializer<*> =
                FalseValueDeserializer(deserializer) { mutableMapOf<Any, Any>() }

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -44,12 +60,58 @@ interface FossIdRestService {
...
+
+            override fun modifyCollectionDeserializer(
+                config: DeserializationConfig?,
+                type: CollectionType?,
+                beanDesc: BeanDescription?,
+                deserializer: JsonDeserializer<*>
+            ): JsonDeserializer<*> {
+                return FalseValueDeserializer(deserializer) { mutableListOf<Any>() }
+            }
```
sschuberth, line 97: Nit: Could be

            ): JsonDeserializer<*> =
                FalseValueDeserializer(deserializer) { mutableListOf<Any>() }

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -44,12 +60,58 @@ interface FossIdRestService {
...
+        private class FalseValueDeserializer(delegate: JsonDeserializer<*>, private val creator: () -> Any) :
+            DelegatingDeserializer(delegate) {
+            override fun deserialize(p: JsonParser, ctxt: DeserializationContext): Any {
+                return if (p.currentToken == JsonToken.VALUE_FALSE) creator() else super.deserialize(p, ctxt)
+            }
+
+            override fun newDelegatingInstance(newDelegatee: JsonDeserializer<*>): JsonDeserializer<*> {
+                return FalseValueDeserializer(newDelegatee, creator)
+            }
+        }
```
sschuberth, line 114: Proposed formatting:

        private class FalseValueDeserializer(
            delegate: JsonDeserializer<*>, private val creator: () -> Any
        ) : DelegatingDeserializer(delegate) {
            override fun deserialize(p: JsonParser, ctxt: DeserializationContext) : Any =
                if (p.currentToken == JsonToken.VALUE_FALSE) creator() else super.deserialize(p, ctxt)

            override fun newDelegatingInstance(newDelegatee: JsonDeserializer<*>): JsonDeserializer<*> =
                FalseValueDeserializer(newDelegatee, creator)
        }

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
```diff
@@ -147,7 +143,7 @@ class FossIdClientNewProjectTest : StringSpec({
 
     "A scan can be deleted" {
         service.deleteScan("", "", SCAN_CODE) shouldNotBeNull {
-            checkResponse("delete scan", true)
+            checkResponse("delete scan")
```
sschuberth, line 146: Not: Ideally, the (somewhat unrelated) omission of the default argument would ideally go to a separate / preceding commit.

clients/fossid-webapp/src/test/kotlin/FossIdClientReturnTypeTest.kt
```diff
@@ -81,84 +73,108 @@ class FossIdClientReturnTypeTest : StringSpec({
...
             checkResponse("list ignored files")
-            toList<IgnoredFile>() shouldNot beEmpty()
+            data shouldNotBeNull {
+                this shouldNot beEmpty()
+            }
         }
     }
 
-    "Pending files can be listed when there is some" {
+    "Pending files can be listed when there is none" {
```
sschuberth, line 163: Note: Ideally, the (somewhat unrelated) test case name fixes (here and below) would go to a separate / preceding commit.

sschuberth (changes requested):

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
nnobelis: Please see new commit for the explanation.

sschuberth (approved):

PR #4046, opened 2021-05-19
DependencyGraphBuilder: Return the managed packages as a sorted set

oheger-bosch:
When constructing analyzer results the packages typically need to be
provided as a sorted set. So far, all PackageManager implementations
making use of DependencyGraphBuilder do this conversion manually.
Simplify this by changing the return type of the packages() function
from Set<Package> to SortedSet<Package>.

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/DependencyGraphBuilder.kt
+++ b/analyzer/src/main/kotlin/managers/utils/DependencyGraphBuilder.kt
@@ -19,6 +19,8 @@
 
 package org.ossreviewtoolkit.analyzer.managers.utils
 
+import java.util.SortedSet
+
 import org.ossreviewtoolkit.model.DependencyGraph
 import org.ossreviewtoolkit.model.DependencyReference
 import org.ossreviewtoolkit.model.Identifier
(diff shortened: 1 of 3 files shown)


analyzer/src/main/kotlin/managers/utils/DependencyGraphBuilder.kt
```diff
@@ -153,9 +155,9 @@ class DependencyGraphBuilder<D>(
     )
 
     /**
-     * Return a set with all the packages that have been encountered for the current project.
+     * Return a sorted set with all the packages that have been encountered for the current project.
      */
-    fun packages(): Set<Package> = resolvedPackages
+    fun packages(): SortedSet<Package> = resolvedPackages.toSortedSet()
```
sschuberth, line 160: I'm wondering whether we should do somewhat the opposite, and convert to a sorted set as late as possible. The reason is that I recall we had performance issues when inserting into sorted sets. And as we only need the sorting in user-facing output / the serialized ORT result files, I wonder whether we could e.g. instead make `PackageManagerResult` not expect a sorted set, and do the sorting later. Or is that class already part of the serialization?

oheger-bosch: No, this class is not serialized. I like this idea. The creation of the sorted set can then be deferred until the analyzer result is constructed by the corresponding builder.
I would then close this PR and open a new one following this other approach, okay?

sschuberth: Yes, fine with me!

oheger-bosch:
Closing to follow a different approach as discussed.

PR #4047, opened 2021-05-20
helper-cli: Add a command to set the labels in an ORT result

mnonnenmacher:
Add a command to set labels in an ORT result file and optionally remove
existing labels.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2021 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.associate
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.flag
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.file
+
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.model.utils.mergeLabels
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+internal class SetLabelsCommand : CliktCommand(
+    help = "Set the labels in an ORT result file."
+) {
+    private val inputOrtFile by option(
+        "--input-ort-file", "-i",
+        help = "The input ORT result file."
+    ).convert { it.expandTilde() }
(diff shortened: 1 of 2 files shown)


helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.file
+
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.model.utils.mergeLabels
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+internal class SetLabelsCommand : CliktCommand(
```
sschuberth, line 37: Just wondering, as the command can also remove labels, whether it should simply be called "LabelCommand" (*to* label as a verb), like `orth label`. In exchange I'd probably rename `--label` / `-l` to `--set` / `-s`.

helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
```diff
@@ -0,0 +1,76 @@
...
+        .defaultLazy { inputOrtFile }
+
+    private val labels by option(
+        "--label", "-l",
+        help = "Set a label in the ORT result, overwriting any existing label of the same name. Can be used multiple " +
+                "times. For example: --label distribution=external"
+    ).associate()
+
+    private val removeExistingLabels by option(
+        "--remove-existing-labels",
```
sschuberth, line 63: `--remove-all`?

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
mnonnenmacher: I'd like to keep "Set" because all other commands start with a verb as well. And I'd also like to keep `--label` to be aligned with the `AnalyzerCommand`.

helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
mnonnenmacher: I think both names are fine but I like the current name a bit more.

helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
sschuberth: Just as a remark (and as explained above) *to* "label" is also a verb 😉

helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.ossreviewtoolkit.model.utils.mergeLabels
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+internal class SetLabelsCommand : CliktCommand(
+    help = "Set the labels in an ORT result file."
+) {
+    private val inputOrtFile by option(
+        "--input-ort-file", "-i",
+        help = "The input ORT result file."
```
fviernau, line 42: The variable name and help do not indicate that the file is actually written to if no output ort file is provided as option. How about eliminating the overwriting of the input for simplicity?

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/SetLabelsCommand.kt
```diff
@@ -0,0 +1,76 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands
```
fviernau, line 20: Instead of this approach, have you considered doing a command which 

(1) extracts the labels to a yaml file
(2) set the labels from a yaml file

This way you can just open up a yaml file and edit the fields. I believe this is quicker than doing it via the command line.

fviernau: After discussing this it seems that as-is fits your use case better.

fviernau (approved):

PR #4048, opened 2021-05-20
SubtractScanResultsCommand: Fix parameter descriptions

mnonnenmacher:


helper-cli/src/main/kotlin/commands/SubtractScanResultsCommand.kt
```diff
@@ -41,23 +41,23 @@ internal class SubtractScanResultsCommand : CliktCommand(
...
+        help = "The ORT result containing the right-hand side scan result."
     ).convert { it.expandTilde() }
         .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
         .convert { it.absoluteFile.normalize() }
         .required()
 
     private val outputOrtFile by option(
         "--output-ort-file",
-        help = "The ORT result containing the left-hand-side scan result."
+        help = "The output ORT result."
```
sschuberth, line 60: This sounds a bit as if the output is already there. How about "The file to write the output ORT result to."?

sschuberth (approved):

PR #4054, opened 2021-05-21
helper-cli: Make sure that `withResolvedScopes()` is called

mnonnenmacher:

### DIFF
--- a/helper-cli/src/main/kotlin/commands/ExtractRepositoryConfigurationCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/ExtractRepositoryConfigurationCommand.kt
@@ -25,9 +25,8 @@ import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.file
 
+import org.ossreviewtoolkit.helper.common.readOrtResult
 import org.ossreviewtoolkit.helper.common.write
-import org.ossreviewtoolkit.model.OrtResult
-import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.utils.expandTilde
 
 class ExtractRepositoryConfigurationCommand : CliktCommand(
@@ -50,8 +49,7 @@ class ExtractRepositoryConfigurationCommand : CliktCommand(
         .required()
 
     override fun run() {
-        ortFile
-            .readValue<OrtResult>()
+        readOrtResult(ortFile)
             .repository
             .config.write(repositoryConfigurationFile)
     }
--- a/helper-cli/src/main/kotlin/commands/GenerateTimeoutErrorResolutionsCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/GenerateTimeoutErrorResolutionsCommand.kt
@@ -27,8 +27,8 @@ import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.file
 
 import org.ossreviewtoolkit.helper.common.getScanIssues
+import org.ossreviewtoolkit.helper.common.readOrtResult
 import org.ossreviewtoolkit.helper.common.replaceConfig
-import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.config.IssueResolution
 import org.ossreviewtoolkit.model.config.IssueResolutionReason
 import org.ossreviewtoolkit.model.config.Resolutions
@@ -68,7 +68,7 @@ internal class GenerateTimeoutErrorResolutionsCommand : CliktCommand(
     ).flag()
 
     override fun run() {
-        val ortResult = ortFile.readValue<OrtResult>().replaceConfig(repositoryConfigurationFile)
+        val ortResult = readOrtResult(ortFile).replaceConfig(repositoryConfigurationFile)
 
         val resolutionProvider = DefaultResolutionProvider().apply {
             var resolutions = Resolutions()
--- a/helper-cli/src/main/kotlin/commands/ImportScanResultsCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/ImportScanResultsCommand.kt
@@ -25,8 +25,7 @@ import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.file
 
-import org.ossreviewtoolkit.model.OrtResult
-import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.helper.common.readOrtResult
 import org.ossreviewtoolkit.scanner.storages.FileBasedStorage
 import org.ossreviewtoolkit.utils.expandTilde
 import org.ossreviewtoolkit.utils.storage.LocalFileStorage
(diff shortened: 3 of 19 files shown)


helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -100,3 +112,33 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
         printStackTrace = stacktrace
     }
 }
+
+/**
+ * Read [ortFile] into an [OrtResult] and return it. Make sure that information about project scopes is available
+ * (by calling [OrtResult.withResolvedScopes]), so that it can be processed.
```
sschuberth, line 118: Hmm, I actually do not see `withResolvedScopes` being called here...

helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -100,3 +112,33 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
...
+ */
+fun CliktCommand.readOrtResult(ortFile: File): OrtResult {
+    log.debug { "Input ORT result file has SHA-1 hash ${HashAlgorithm.SHA1.calculate(ortFile)}." }
```
sschuberth, line 121: Just as a not, I'm not sure how useful the SHA1 calculation is for the `helper-cli` module. It was mostly added for debugging "hand-over" issues in pipelines running ORT. So you might want to remove it here and below.

helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -120,7 +120,7 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
 fun CliktCommand.readOrtResult(ortFile: File): OrtResult {
     log.debug { "Input ORT result file has SHA-1 hash ${HashAlgorithm.SHA1.calculate(ortFile)}." }
 
-    val (ortResult, duration) = measureTimedValue { ortFile.readValue<OrtResult>() }
+    val (ortResult, duration) = measureTimedValue { ortFile.readValue<OrtResult>().withResolvedScopes() }
```
sschuberth, line 123: Ah, here it comes. Then maybe the function docs should also only start mentioning it in this commit?

sschuberth (changes requested):

helper-cli/src/main/kotlin/HelperMain.kt
mnonnenmacher: Yes, that sentence should have gone to the next commit.

helper-cli/src/main/kotlin/HelperMain.kt
mnonnenmacher: I wasn't sure what the use case was so I have kept it, but I have removed it now.

helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -100,3 +111,28 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
         printStackTrace = stacktrace
     }
 }
+
+/**
+ * Read [ortFile] into an [OrtResult] and return it.
+ */
+fun CliktCommand.readOrtResult(ortFile: File): OrtResult {
```
fviernau, line 119: Why are these functions of `CliktCommand`?
Isn't the `CliktCommand` not used?

helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -49,6 +54,12 @@ import org.ossreviewtoolkit.helper.commands.VerifySourceArtifactCurationsCommand
 import org.ossreviewtoolkit.helper.commands.packageconfig.PackageConfigurationCommand
 import org.ossreviewtoolkit.helper.commands.repoconfig.RepositoryConfigurationCommand
 import org.ossreviewtoolkit.helper.common.ORTH_NAME
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.formatSizeInMib
+import org.ossreviewtoolkit.utils.log
+import org.ossreviewtoolkit.utils.perf
```
fviernau, line 62: I believe it would make sense to move them into the Utils class.

helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -100,3 +111,28 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
...
     }
 }
+
+/**
+ * Read [ortFile] into an [OrtResult] and return it.
+ */
+fun CliktCommand.readOrtResult(ortFile: File): OrtResult {
+    val (ortResult, duration) = measureTimedValue { ortFile.readValue<OrtResult>() }
+
+    log.perf {
```
fviernau, line 122: I think performance logging is not needed in the helper cli and prefer to not add it.

fviernau (changes requested):

helper-cli/src/main/kotlin/HelperMain.kt
sschuberth: Because we need an object to log an / where the log extension is defined.

helper-cli/src/main/kotlin/HelperMain.kt
sschuberth: If the last renaming `log` would go away, then this would indeed not need to be an extension function anymore.

sschuberth (dismissed):

helper-cli/src/main/kotlin/HelperMain.kt
mnonnenmacher: I wanted to have them in a similar place as in the `cli` to make it easier if both need updates in the future.

helper-cli/src/main/kotlin/HelperMain.kt
mnonnenmacher: But does it hurt to have optional performance logging? It's only visible if the debug flag is set.

sschuberth: ...or the `--perf` flag 😉 Speaking of performance (for *serializing* result files, though), did you see an drop in your number since https://github.com/oss-review-toolkit/ort/pull/3999/commits/738834057e41e99e6a44255fb322dd109f725c1d was merged?

fviernau: > But does it hurt to have optional performance logging? 

I believe that's the wrong question to ask, because unneeded code IMO should be avoided.

helper-cli/src/main/kotlin/HelperMain.kt
fviernau: Is it really harder to updated if they are in the Utils file?

fviernau: It feels more natural to me that the subcommands use function from `common/Utils` rather than from `Main`, because they don't need to know about main. I won't insist on it though.

mnonnenmacher: Not easier to implement, but easier to find. However, I've implemented the change.

helper-cli/src/main/kotlin/HelperMain.kt
mnonnenmacher: I think "unneeded" is a matter of opinion in this case. However, I have removed it to not prolong the discussion.

fviernau (approved):

helper-cli/src/main/kotlin/HelperMain.kt
mnonnenmacher: > ...or the `--perf` flag wink Speaking of performance (for _serializing_ result files, though), did you see an drop in your number since [7388340](https://github.com/oss-review-toolkit/ort/commit/738834057e41e99e6a44255fb322dd109f725c1d) was merged?

The duration went down a bit, but due to the different result sizes it's difficult to draw conclusions. The deserialization:serialization ratio went roughly from 4:1 to 5:1, but that could also be coincidence or caused by other factors. I think the best way to check the impact would be a before/after comparison for a large file.

PR #4056, opened 2021-05-21
Fix formatting of some test assets

mnonnenmacher:
Align the indentation with how Jackson would serialize those files.

### DIFF
--- a/cli/src/funTest/assets/analyzer-result-with-dependency-graph.yml
+++ b/cli/src/funTest/assets/analyzer-result-with-dependency-graph.yml
@@ -30,182 +30,182 @@ analyzer:
     allow_dynamic_versions: false
   result:
     projects:
-      - id: "Maven:com.vdurmont:semver4j:3.1.0"
-        definition_file_path: "pom.xml"
+    - id: "Maven:com.vdurmont:semver4j:3.1.0"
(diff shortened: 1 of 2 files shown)


sschuberth:
This is not only about indentation, but also about ordering. Is this due to the recent merge of https://github.com/oss-review-toolkit/ort/pull/4050?

mnonnenmacher:
> This is not only about indentation, but also about ordering. Is this due to the recent merge of #4050?

I used IntelliJ to reformat the files, I don't think any order has changed.

mnonnenmacher:
> > This is not only about indentation, but also about ordering. Is this due to the recent merge of #4050?
> 
> I used IntelliJ to reformat the files, I don't think any order has changed.

Looking at the diff, it seems that Git is really bad at doing diffs if only indentation is changed, maybe that gave the impression that the order has changed.

sschuberth:
> Looking at the diff, it seems that Git is really bad at doing diffs if only indentation is changed, maybe that gave the impression that the order has changed.

It's also GitHub (not Git) being bad at this, but you're right, I verified locally with a more sophisticated diff tool that only indentation changed.

sschuberth (approved):

PR #4057, opened 2021-05-21
Minor improvements for reference.conf

mnonnenmacher:

### DIFF
--- a/model/src/main/resources/reference.conf
+++ b/model/src/main/resources/reference.conf
@@ -31,19 +31,19 @@ ort {
 
   licenseFilePatterns {
     licenseFilenames = ["license*"]
-    patentFilenames = ["patents"]
+    patentFilenames = [patents]
     rootLicenseFilenames = ["readme*"]
   }
 
   downloader {
     // Only used by the CLI tool when the '--license-classifications-file' option is specified.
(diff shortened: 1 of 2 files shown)


model/src/main/resources/reference.conf
```diff
@@ -31,19 +31,19 @@ ort {
 
   licenseFilePatterns {
     licenseFilenames = ["license*"]
-    patentFilenames = ["patents"]
+    patentFilenames = [patents]
```
sschuberth, line 34: Hmm, I'm not sure esp. about this one. Isn't it confusing to omit quotes here, but not above and below?

sschuberth (approved):
Approving despite my comment, as the syntax seems to be valid after all.

model/src/main/resources/reference.conf
mnonnenmacher: I first thought the same, but then decided it's more consistent to use quotes only if they are required in all places in the file. But I'd be fine with changing this one back as well if you prefer that.

sschuberth: No, I think it's fine to keep it as a show-case that wildcard patterns need to be quoted.

PR #4063, opened 2021-05-21
Some code inspection fixes

fviernau:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pip.kt
+++ b/analyzer/src/main/kotlin/managers/Pip.kt
@@ -302,12 +302,15 @@ class Pip(
         }
 
         // Amend information from "setup.py" with that from "requirements.txt".
-        val projectName = when (Pair(setupName.isNotEmpty(), requirementsName.isNotEmpty())) {
-            Pair(true, false) -> setupName
+        val hasSetupName = setupName.isNotEmpty()
+        val hasRequirementsName = requirementsName.isNotEmpty()
+
+        val projectName = when {
+            hasSetupName && !hasRequirementsName -> setupName
             // In case of only a requirements file without further meta-data, use the relative path to the analyzer
             // root as a unique project name.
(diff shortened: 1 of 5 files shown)


model/src/main/kotlin/utils/DatabaseUtils.kt
```diff
@@ -110,7 +110,7 @@ object DatabaseUtils {
     /**
      * Start a new asynchronous transaction to execute the given [statement] on this [Database].
      */
-    suspend fun <T> Database.asyncTx(statement: suspend Transaction.() -> T): Deferred<T> =
+    suspend fun <T> Database.startTransactionAsync(statement: suspend Transaction.() -> T): Deferred<T> =
```
mnonnenmacher, line 113: I would remove the `start` prefix because it doesn't add much value and only makes it longer. Renaming should also be aligned with the other `tx` function for consistency. So e.g. name them `transaction()` and `transactionAsync()`. Or maybe call them `transaction()` and `suspendedTransaction()` to align with exposed: https://github.com/JetBrains/Exposed/wiki/Transactions#working-with-coroutines

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -302,12 +302,15 @@ class Pip(
         }
 
         // Amend information from "setup.py" with that from "requirements.txt".
-        val projectName = when (Pair(setupName.isNotEmpty(), requirementsName.isNotEmpty())) {
-            Pair(true, false) -> setupName
+        val hasSetupName = setupName.isNotEmpty()
```
sschuberth, line 305: Commit message: s/boolen/boolean/

sschuberth (changes requested):

sschuberth (approved):

PR #4068, opened 2021-05-25
Don't use a sorted set for shared packages in PackageManagerResult

oheger-bosch:
This is a minor simplification/optimization for package managers using the dependency graph format. Refer to the commit messages for further details.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -347,9 +347,8 @@ open class Npm(
         ortProxySelector.removeProxyOrigin(managerName)
     }
 
-    override fun createPackageManagerResult(projectResults: Map<File, List<ProjectAnalyzerResult>>):
-            PackageManagerResult =
-        PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages().toSortedSet())
+    override fun createPackageManagerResult(projectResults: Map<File, List<ProjectAnalyzerResult>>) =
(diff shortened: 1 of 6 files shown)


fviernau (dismissed):

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -349,7 +349,7 @@ open class Npm(
 
     override fun createPackageManagerResult(projectResults: Map<File, List<ProjectAnalyzerResult>>):
             PackageManagerResult =
-        PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages().toSortedSet())
+        PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages())
```
sschuberth, line 352: Nit: This formatting is inconsistent with the formatting in the changes above, and I wonder whether we should take the opportunity to reformat all of these to omit the explicit return type, as the `PackageManagerResult` is directly constructed afterwards:

    override fun createPackageManagerResult(projectResults: Map<File, List<ProjectAnalyzerResult>>) =

sschuberth (approved):
Approving as my comment was only a nit.

analyzer/src/main/kotlin/managers/Npm.kt
oheger-bosch: Changed the formatting for all implementations of `createPackageManagerResult()` and dropped the explicit return type.

sschuberth (approved):

PR #4072, opened 2021-05-25
Docker: Conditionally start ssh agent for SSH downloads

Martin-Idel-SI:
We want to run ORT within a Kubernetes pod, but we also need to be able to resolve repositories by SSH since some packages have a download link as SSH-link. 

Previously, we would start an agent on the host and then just share the socket with the container, but this is not as simple in a Kubernetes setup. Therefore we need a way to open an ssh agent within the container, which is what this PR is about.

The PR will only change anything if an identity file is mounted into the Docker container.

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# Copyright (C) 2021 Bosch.IO GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+set -euo pipefail
+
+if [[ -f /root/.ssh/id_rsa ]]; then
+  eval $(ssh-agent -s)
+  ssh-add /root/.ssh/id_rsa
+fi
+
+opt/ort/bin/ort "$@"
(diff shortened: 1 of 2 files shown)


mnonnenmacher:
I'm not sure if this is useful default behavior for other users for the Docker image, did you consider using a custom entrypoint script in your setup instead?

sschuberth:
Commit message:

> This allows setting up SSH downloads in containers where SSH socket can
> be shared from the host system.

Shouldn't this say "*cannot* be shared"? Also, naming Kubernetes pod as a concrete example would be nice, IMO. Maybe even with a link to docs explaining *why* this is not possible / supported in a Kubernetes setup.

scripts/entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+set -euo pipefail
+
+if [[ -f /root/.ssh/id_rsa ]]; then
+  eval $(ssh-agent -s)
+  ssh-add /root/.ssh/id_rsa
```
sschuberth, line 24: I'm wondering why this should be necessary. `id_rsa` is one of the default keys OpenSSH tries, so if we *just* want to authenticate with *that* key, there should be no need to run an agent at all. Can you please elaborate?

scripts/entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+set -euo pipefail
+
+if [[ -f /root/.ssh/id_rsa ]]; then
```
sschuberth, line 22: I'm not sure this is the right logic to determine whether run the agent *inside* the container, or whether to rely on it running on the *host*. I probably would have expected a condition that checks for the `SSH_AUTH_SOCK` environment variable to be set. If yes, the agent is running on the host and it gets forwarded to the container. If no, we need to run the agent ourselves.

sschuberth (changes requested):

scripts/entrypoint.sh
Martin-Idel-SI: Actually, we would override that key - and therefore have to explicitly add it.
But it would probably be better to use a different directory and not overwrite the existing one, I agree.

Martin-Idel-SI:
@mnonnenmacher That's what we were wondering. I'm also not sure, so I'll close this for now.

PR #4077, opened 2021-05-27
SpdxExpression: Revert sorting expressions when computing the DNF

mnonnenmacher:

### DIFF
--- a/reporter/src/funTest/assets/static-html-reporter-test-expected-output.html
+++ b/reporter/src/funTest/assets/static-html-reporter-test-expected-output.html
@@ -631,7 +631,7 @@ Resolved by: CANT_FIX_EXCEPTION - Apache-2 is not an issue.</p>
               </dl>
               <em>Effective License:</em>
               <dl>
-                <dd>Apache-2.0 AND CC-BY-NC-3.0 AND GPL-3.0-only WITH GCC-exception-3.1 AND MIT</dd>
+                <dd>GPL-3.0-only WITH GCC-exception-3.1 AND CC-BY-NC-3.0 AND Apache-2.0 AND MIT</dd>
               </dl>
             </td><td>
               <ul></ul>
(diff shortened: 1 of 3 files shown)


spdx-utils/src/test/kotlin/SpdxExpressionTest.kt
```diff
@@ -291,25 +295,25 @@ class SpdxExpressionTest : WordSpec() {
             "not change an expression already in DNF" {
                 val spdxExpression = "(a AND b) OR (c AND d)".toSpdx()
 
-                spdxExpression.disjunctiveNormalForm() shouldBe spdxExpression
+                spdxExpression.disjunctiveNormalForm() shouldBeString spdxExpression.toString()
```
sschuberth, line 298: Wouldn't it be easier to just call "toString()" on the left-hand side (i.e. on `disjunctiveNormalForm()`) and then keep the string (but without the `.toSpdx()`) on the right-hand side?

reporter/src/funTest/assets/static-html-reporter-test-expected-output.html
```diff
@@ -631,7 +631,7 @@ <h3 class="ort-excluded">Packages</h3>
               </dl>
               <em>Effective License:</em>
               <dl>
-                <dd>Apache-2.0 AND CC-BY-NC-3.0 AND GPL-3.0-only WITH GCC-exception-3.1 AND MIT</dd>
+                <dd>GPL-3.0-only WITH GCC-exception-3.1 AND CC-BY-NC-3.0 AND Apache-2.0 AND MIT</dd>
```
sschuberth, line 634: > SPDX expression

expression*s*

sschuberth: The commit lacks a signoff.

sschuberth (changes requested):

spdx-utils/src/test/kotlin/SpdxExpressionTest.kt
mnonnenmacher: I went one step further and duplicated the string to align with the style of the other tests.

spdx-utils/src/test/kotlin/SpdxExpressionTest.kt
```diff
@@ -289,27 +293,25 @@ class SpdxExpressionTest : WordSpec() {
 
         "disjunctiveNormalForm()" should {
             "not change an expression already in DNF" {
-                val spdxExpression = "(a AND b) OR (c AND d)".toSpdx()
-
-                spdxExpression.disjunctiveNormalForm() shouldBe spdxExpression
+                "(a AND b) OR (c AND d)".toSpdx().disjunctiveNormalForm() shouldBeString "(a AND b) OR (c AND d)"
```
sschuberth, line 296: Shouldn't the expected / normalized string also drop the parentheses here, like in the other examples, as `AND` binds stronger than `OR` by default?

spdx-utils/src/test/kotlin/SpdxExpressionTest.kt
```diff
@@ -594,3 +596,8 @@ class SpdxExpressionTest : WordSpec() {
         }
     }
 }
+
+private infix fun <T> T.shouldBeString(expected: String) {
+    assertionCounter.inc()
+    eq(this.toString(), expected)?.let(errorCollector::collectOrThrow)
```
sschuberth, line 602: I'm having a hard time understanding what this does. I guess you copied this from somewhere? Can you link to the source or add some code comments why some counter (?!) needs to be incremented here?

mnonnenmacher: I decided to implement this with a custom matcher instead. It's a bit more code but I find it cleaner because it does not have to mess with kotest internals.

sschuberth (approved):

PR #4082, opened 2021-05-28
Make issues for unmapped licenses consider concluded licenses

sschuberth:


model/src/main/kotlin/CuratedPackage.kt
```diff
@@ -49,15 +49,17 @@ data class CuratedPackage(
     /**
      * Check if this package contains any erroneous data.
      */
-    fun collectIssues(): List<OrtIssue> =
-        pkg.declaredLicensesProcessed.unmapped.map { unmappedLicense ->
+    fun collectIssues(): List<OrtIssue> {
```
fviernau, line 52: The issue is just a warning. I believe it's better to not do this change to keep the default behavior simple and thus easier to understand for the user. I suspect that the behavior 'as-is' before and after this change any does not match the needs of all users. 

Instead of adding more complexity here I propose to (1) add a policy rule to our defaults which implements your idea with a rule violation and to (2) optionally eliminate creating this ORT issue completely.

(In general I think we should move towards reporting issues via rule violations where applicable. It's already valuable for just customizing the severity, let alone the added flexibility)

What do you think?

fviernau (changes requested):

model/src/main/kotlin/AnalyzerResult.kt
```diff
@@ -84,7 +84,7 @@ data class AnalyzerResult(
         }
```
oheger-bosch, line 84: Commit message in the first line of the body: "preparaed" -> "preparation"

model/src/main/kotlin/CuratedPackage.kt
sschuberth: Leaving this as-is is not an option for us. We have an "everything must be green" mandate, so also warnings need to be addressed.

While we could have gotten rid of the warning by creating a mapping as part of a curation, there is little reason to create such a mapping if its result is anyway going to be overridden by a concluded license.

But I do see some more options:
* lower the severity to hint (which for use results in a "green" run) if a concluded license if set
* only *log* a warning for this, but not create an issue if a concluded license if set

How about these?

fviernau: The fact that Bosch fixes all warning and HERE doesn't makes it hard to make a good proposal.
I believe it indicates that severity should be under user control, so that everybody can decide what exactly is a must fix.
Anyhow, the only additional idea I'd throw into the basket of options is the following (which tries to avoid adding complexity into ORT)

1. Turn this issue into a hint
2. Implement a rule violation with your desired (more complex) logic

sschuberth: > On the other hand warning seems just right, so I can't see any short term options which is ideal imo. I'd actually prefer @sschuberth 's proposal to "lower the severity to hint (which for us results in a "green" run) if a concluded license is set"

Ok, I'll make the change, as I also like to have a warning for this in the general case.

fviernau (approved):

oheger-bosch (approved):

sschuberth:
I'll work on the `PipFunTest` failure separately...

PR #4086, opened 2021-05-28
Support NPM dependencies that point to VCS hosts (like GitHub)

sschuberth:

### DIFF
--- a/downloader/src/main/kotlin/VcsHost.kt
+++ b/downloader/src/main/kotlin/VcsHost.kt
@@ -238,6 +238,11 @@ enum class VcsHost(
         private val GIT_REVISION_FRAGMENT = Regex("git.+#[a-fA-F0-9]{7,}")
 
         /**
+         * Return the [VcsHost] for a [vcsUrl].
+         */
+        fun toVcsHost(vcsUrl: URI): VcsHost? = values().find { host -> host.isApplicable(vcsUrl) }
+
+        /**
          * Return all [VcsInfo] that can be parsed from [projectUrl] without actually making a network request.
          */
         fun toVcsInfo(projectUrl: String): VcsInfo {
@@ -310,11 +315,6 @@ enum class VcsHost(
             return values().find { host -> host.isApplicable(vcsInfo) }
                 ?.toPermalinkInternal(vcsInfo.normalize(), startLine, endLine)
         }
-
-        /**
-         * Return the [VcsHost] for a [vcsUrl].
-         */
-        fun toVcsHost(vcsUrl: URI): VcsHost? = values().find { host -> host.isApplicable(vcsUrl) }
     }
 
     private val supportedTypes = supportedTypes.asList()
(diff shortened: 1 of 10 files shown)


sschuberth:
FYI @neubs-bsi.

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -237,6 +237,11 @@ enum class VcsHost(
         private val SVN_TRUNK_PATTERN = Regex("(.*svn.*)/(trunk)/?(.*)")
```
oheger-bosch, line 250: Commit message typo: ...it makes sense to list _it_ first...

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -385,7 +385,7 @@ private fun projectUrlToUserOrOrgAndProject(projectUrl: URI): Pair<String, Strin
```
oheger-bosch, line 430: Commit message: It would be nice to give a reason why the suffix is removed.

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -358,6 +386,23 @@ enum class VcsHost(
 
     protected abstract fun toVcsInfoInternal(projectUrl: URI): VcsInfo
 
+    /**
+     * Return the download URL to an archive generated for the referenced [vcsInfo], or null if no download URL can be
+     * determined.
+     */
+    fun toArchiveDownloadUrl(vcsInfo: VcsInfo): String? {
```
oheger-bosch, line 393: I do not understand this duplication of the function from the companion object. Is this intended?

downloader/src/main/kotlin/VcsHost.kt
sschuberth: Oh, I though 'Strip any ".git" suffix from the project name' already makes clear that ".git" should not be regarded as a part of the project name, but I'll see what I can do.

downloader/src/main/kotlin/VcsHost.kt
sschuberth: Yes, it's intended. That's the pattern we currently follow in this class, see also e.g. the `toPermalink()` function, so that you can call these functions in case you already know the VcsHost, and also in case you don't already know the VcsHost (and want to get it auto-detected).

downloader/src/main/kotlin/VcsHost.kt
oheger-bosch: Thanks for bearing with my limited knowledge ;-)

downloader/src/main/kotlin/VcsHost.kt
oheger-bosch: Thank you for explaining.

oheger-bosch (approved):

neubs-bsi (approved):

PR #4092, opened 2021-05-31
Reporter WebApp Improvements

tsteenbe:


reporter-web-app/src/utils/index.js
```diff
@@ -40,4 +40,54 @@ const randomStringGenerator = (length = (Math.floor(Math.random() * 501) + 20))
     return alpha(l, symbol(1, alpha(r, []))).join('');
 };
 
-export { randomStringGenerator };
+// Utility function to generate color for given license string 
+const licenseToHslColor = (license) => {
+    let hash = 0;
+    let min;
+    let max;
+    // Basic license classification based on https://scancode-licensedb.aboutcode.org
```
sschuberth, line 48: IMO we shouldn't do this. I understand that we cannot easily use our own `license-classifications.yml` as its freely defined categories have no machine-readable semantics attached. But we should stick to our principle to not give any advice and not do any interpretation, even if it's just coloring according to something "dangerours" or "safe".

*Edit:* But maybe we could add a `color` property to the categories in `license-classifications.yml`, and make use of them here.

reporter-web-app/package.json
```diff
@@ -21,7 +21,7 @@
     },
     "dependencies": {
         "@ant-design/icons": "^4.6.2",
-        "antd": "^4.16.1",
+        "antd": "^4.16.2",
```
sschuberth, line 24: Can we squash this to the other dependency update commit?

sschuberth (changes requested):

reporter-web-app/src/utils/index.js
tsteenbe: Agreed, removed the code in question and created https://github.com/oss-review-toolkit/ort/issues/4150

reporter-web-app/src/reducers/index.js
```diff
@@ -446,3 +446,5 @@ export default (state = initState, action) => {
         return state;
     }
 };
+
+export default states;
```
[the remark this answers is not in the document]
sschuberth: According to GitHub rendering, this issue is still there.

reporter-web-app/src/components/Shared.js
```diff
@@ -0,0 +1,79 @@
...
+        ? record[dataIndex].toString().toLowerCase().includes(value.toLowerCase())
+        : false),
+    onFilterDropdownVisibleChange: (visible) => {
+        if (visible) {
+            setTimeout(() => that.searchInput.select());
+        }
+    }
+});
+
+export { getColumnSearchProps };
```
sschuberth, line 79: Nit: Also here a final newline is missing.

reporter-web-app/src/utils/index.js
```diff
@@ -40,4 +40,11 @@ const randomStringGenerator = (length = (Math.floor(Math.random() * 501) + 20))
     return alpha(l, symbol(1, alpha(r, []))).join('');
 };
 
-export { randomStringGenerator };
+// Utility function to generate color for given license string 
+const licenseToHslColor = (license) => {
+    return 'hsl(' + (Math.random() * (360 - 22) + 22) + ',' +
+        (25 + 70 * Math.random()) + '%,' + 
+        (55 + 10 * Math.random()) + '%)';
```
sschuberth, line 47: I don't see how "license" is being taken into account here, is it? Because the same license should generate the same color, or?

reporter-web-app/src/reducers/index.js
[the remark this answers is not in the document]
sschuberth: This is still there, but I guess I'll approve anyway to move things forward.

reporter-web-app/src/components/Shared.js
sschuberth: Same here.

sschuberth (approved):

PR #4093, opened 2021-06-01
Npm: Explicitly declare the type of `graphBuilder`

fviernau:
Make the type more obvious.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -326,7 +326,8 @@ open class Npm(
         }
     }
 
-    private val graphBuilder = DependencyGraphBuilder(NpmDependencyHandler(npmRegistry))
+    private val graphBuilder: DependencyGraphBuilder<NpmModuleInfo> =
+        DependencyGraphBuilder(NpmDependencyHandler(npmRegistry))
 
     /**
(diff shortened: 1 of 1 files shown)


sschuberth:
> Explicitly declared

"Explicitly declare" (omit "d")

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -326,7 +326,8 @@ open class Npm(
         }
     }
 
-    private val graphBuilder = DependencyGraphBuilder(NpmDependencyHandler(npmRegistry))
+    private val graphBuilder: DependencyGraphBuilder<NpmModuleInfo> =
+        DependencyGraphBuilder(NpmDependencyHandler(npmRegistry))
```
sschuberth, line 330: Just for my understanding: Are you after making the `DependencyGraphBuilder` type explicit, or the template argument `NpmModuleInfo`? I would understand the latter somewhat, but for the first "main" type having something like

    val ... : Type = Type()

seem overly verbose to make `Type` explicit, as there is a immediate duplication of the `Type`.

fviernau: I'm after making the `<NpmModuleInfo` obvious.

sschuberth (approved):

PR #4094, opened 2021-06-01
UploadResultToSw360Command: Add a new command line option to attach sources

### LINKED ISSUE #3998: Improve the UploadResultToSw360Command to optionally also upload source code bundles
There are cases when one would like to bundle and attach downloaded source code to SW360 releases, e.g. as a backup if the upstream source vanishes. We should add a CLI option like "--attach-sources", "-a" to support attaching source code bundles to SW360 components. If that option is given, source code should get downloaded / archived using the ORT downloader beforehand.

bs-ondem:
Add a new command line option --attach-sources in order to attach sources
of packages if they exist. The Downloader is used to download the sources of
the package and it's packed as a ZIP file in a temporary directory.

Resolves #3998.

Signed-off-by: Onur Demirci <onur.demirci@bosch.io>

### DIFF
--- a/cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
+++ b/cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
@@ -24,13 +24,16 @@ import com.fasterxml.jackson.databind.DeserializationFeature
 import com.github.ajalt.clikt.core.CliktCommand
 import com.github.ajalt.clikt.core.requireObject
 import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.file
 
+import org.eclipse.sw360.clients.adapter.AttachmentUploadRequest
 import org.eclipse.sw360.clients.adapter.SW360ProjectClientAdapter
 import org.eclipse.sw360.clients.adapter.SW360ReleaseClientAdapter
 import org.eclipse.sw360.clients.rest.resource.SW360Visibility
+import org.eclipse.sw360.clients.rest.resource.attachments.SW360AttachmentType
 import org.eclipse.sw360.clients.rest.resource.projects.SW360Project
 import org.eclipse.sw360.clients.rest.resource.releases.SW360Release
 import org.eclipse.sw360.clients.utils.SW360ClientException
@@ -38,6 +41,7 @@ import org.eclipse.sw360.clients.utils.SW360ClientException
 import org.ossreviewtoolkit.cli.GlobalOptions
 import org.ossreviewtoolkit.cli.utils.inputGroup
 import org.ossreviewtoolkit.cli.utils.readOrtResult
+import org.ossreviewtoolkit.downloader.Downloader
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Package
@@ -46,9 +50,12 @@ import org.ossreviewtoolkit.model.config.Sw360StorageConfiguration
 import org.ossreviewtoolkit.model.jsonMapper
 import org.ossreviewtoolkit.model.utils.toPurl
 import org.ossreviewtoolkit.scanner.storages.Sw360Storage
+import org.ossreviewtoolkit.utils.archive
 import org.ossreviewtoolkit.utils.collectMessagesAsString
+import org.ossreviewtoolkit.utils.createOrtTempDir
 import org.ossreviewtoolkit.utils.expandTilde
 import org.ossreviewtoolkit.utils.log
+import org.ossreviewtoolkit.utils.safeDeleteRecursively
 
 class UploadResultToSw360Command : CliktCommand(
     name = "upload-result-to-sw360",
@@ -64,6 +71,11 @@ class UploadResultToSw360Command : CliktCommand(
         .convert { it.absoluteFile.normalize() }
         .required()
         .inputGroup()
+
+    private val attachSources by option(
+        "--attach-sources", "-a",
+        help = "Download sources of packages and upload them as attachments to SW360 releases."
+    ).flag()
 
     private val globalOptionsForSubcommands by requireObject<GlobalOptions>()
 
@@ -81,13 +93,50 @@ class UploadResultToSw360Command : CliktCommand(
         val sw360Connection = Sw360Storage.createConnection(sw360Config, sw360JsonMapper)
         val sw360ReleaseClient = sw360Connection.releaseAdapter
         val sw360ProjectClient = sw360Connection.projectAdapter
+        val downloader = Downloader(globalOptionsForSubcommands.config.downloader)
 
         getProjectWithPackages(ortResult).forEach { (project, pkgList) ->
             val linkedReleases = pkgList.mapNotNull { pkg ->
                 val name = createReleaseName(pkg.id)
-                sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
+                val release = sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
                     .flatMap { sw360ReleaseClient.enrichSparseRelease(it) }
                     .orElseGet { createSw360Release(pkg, sw360ReleaseClient) }
+
+                if (attachSources) {
+                    val tempDirectory = createOrtTempDir(pkg.id.toPath())
+                    try {
+                        // First, download the sources of the package into a sources directory, whose parent directory
+                        // is temporary.
+                        val sourcesDirectory = tempDirectory.resolve("sources")
+                        downloader.download(pkg, sourcesDirectory)
+
+                        // After downloading the source files successfully in a sources directory, create a
+                        // ZIP file of the sources directory and save it in the root directory of it.
+                        // Finally the created ZIP file of the sources can be uploaded to SW360 as an attachment
+                        // of the release.
+                        val zipFile = tempDirectory.resolve("${pkg.id.toPath("-")}.zip")
+                        val archiveResult = archive(sourcesDirectory, zipFile)
+
+                        val uploadResult = sw360ReleaseClient.uploadAttachments(
+                            AttachmentUploadRequest.builder(release)
+                                .addAttachment(archiveResult.getOrThrow().toPath(), SW360AttachmentType.SOURCE)
+                                .build()
+                        )
+
+                        if (uploadResult.isSuccess) {
+                            log.info {
+                                "Successfully uploaded source attachment '${zipFile.name}' to release " +
+                                        "${release.id}:${release.name}"
+                            }
+                        } else {
+                            log.error { "Could not upload source attachment: " + uploadResult.failedUploads() }
+                        }
+                    } finally {
+                        tempDirectory.safeDeleteRecursively(force = true)
+                    }
+                }
+
+                release
             }
 
             val sw360Project = sw360ProjectClient.getProjectByNameAndVersion(project.id.name, project.id.version)
--- a/utils/src/main/kotlin/Utils.kt
+++ b/utils/src/main/kotlin/Utils.kt
@@ -26,6 +26,8 @@ import java.security.Permission
 
 import kotlin.reflect.full.memberProperties
 
+import org.ossreviewtoolkit.spdx.VCS_DIRECTORIES
+
 /**
  * The directory to store ORT (read-only) configuration in.
  */
@@ -63,6 +65,25 @@ val ortDataDirectory by lazy {
  * Global variable that gets toggled by a command line parameter parsed in the main entry points of the modules.
  */
 var printStackTrace = false
+
+/**
+ * Archive the contents of [inputDir], omitting common [VCS_DIRECTORIES], to [zipFile] where an optional [prefix] is
+ * added to each file. Return a [Result] wrapping the [zipFile] on success, or an exception of failure.
+ */
+fun archive(inputDir: File, zipFile: File, prefix: String = ""): Result<File> {
+    return runCatching {
+        inputDir.packZip(
+            zipFile,
+            prefix,
+            directoryFilter = { it.name !in VCS_DIRECTORIES },
+            fileFilter = { it != zipFile }
+        )
+
+        zipFile
+    }.onFailure {
+        it.showStackTrace()
+    }
+}
 
 /**
  * Return whether [T] (usually an instance of a data class) has any non-null property.
--- a/cli/src/main/kotlin/commands/DownloaderCommand.kt
+++ b/cli/src/main/kotlin/commands/DownloaderCommand.kt
@@ -37,8 +37,6 @@ import com.github.ajalt.clikt.parameters.options.switch
 import com.github.ajalt.clikt.parameters.types.enum
 import com.github.ajalt.clikt.parameters.types.file
 
-import java.io.File
-
 import org.ossreviewtoolkit.cli.GlobalOptions
 import org.ossreviewtoolkit.cli.GroupTypes.FileType
 import org.ossreviewtoolkit.cli.GroupTypes.StringType
@@ -61,19 +59,17 @@ import org.ossreviewtoolkit.model.licenses.LicenseClassifications
 import org.ossreviewtoolkit.model.licenses.LicenseInfoResolver
 import org.ossreviewtoolkit.model.licenses.LicenseView
 import org.ossreviewtoolkit.model.licenses.ResolvedLicenseInfo
-import org.ossreviewtoolkit.model.licenses.orEmpty
 import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.model.utils.createLicenseInfoResolver
-import org.ossreviewtoolkit.spdx.VCS_DIRECTORIES
 import org.ossreviewtoolkit.spdx.model.LicenseChoice
 import org.ossreviewtoolkit.utils.ArchiveType
 import org.ossreviewtoolkit.utils.ORT_LICENSE_CLASSIFICATIONS_FILENAME
+import org.ossreviewtoolkit.utils.archive
 import org.ossreviewtoolkit.utils.collectMessagesAsString
 import org.ossreviewtoolkit.utils.encodeOrUnknown
 import org.ossreviewtoolkit.utils.expandTilde
 import org.ossreviewtoolkit.utils.log
 import org.ossreviewtoolkit.utils.ortConfigDirectory
-import org.ossreviewtoolkit.utils.packZip
 import org.ossreviewtoolkit.utils.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.showStackTrace
 
@@ -370,20 +366,5 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
             .filter { it.id in effectiveLicenses }
             .flatMap { it.categories }
             .toSet()
-    }
-}
-
-private fun archive(inputDir: File, zipFile: File, prefix: String = ""): Result<File> {
-    return runCatching {
-        inputDir.packZip(
-            zipFile,
-            prefix,
-            directoryFilter = { it.name !in VCS_DIRECTORIES },
-            fileFilter = { it != zipFile }
-        )
-
-        zipFile
-    }.onFailure {
-        it.showStackTrace()
     }
 }


cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -89,9 +107,27 @@ class UploadResultToSw360Command : CliktCommand(
         getProjectWithPackages(ortResult).forEach { (project, pkgList) ->
             val linkedReleases = pkgList.mapNotNull { pkg ->
                 val name = createReleaseName(pkg.id)
-                sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
+                var release = sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
                     .flatMap { sw360ReleaseClient.enrichSparseRelease(it) }
                     .orElseGet { createSw360Release(pkg, sw360ReleaseClient) }
+
+                if (attachSources) {
+                    val tempDirectory = createTempDirectory(pkg.id.toCoordinates()).toFile()
```
sschuberth, line 115: Please use `createOrtTempDir()` instead which uses an ORT-specific prefix automatically.

sschuberth: `toCoordinates()` is not safe to use as part of a directory name as it contains colons. Please use `pkg.id.toPath()` instead, similar to like in the `DownloaderCommand`:

https://github.com/oss-review-toolkit/ort/blob/768a585b1ccf72cbd952fe7f95216a15d11c947c/cli/src/main/kotlin/commands/DownloaderCommand.kt#L244

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -89,9 +107,27 @@ class UploadResultToSw360Command : CliktCommand(
...
+                    val saveSourcesTemporarily = saveSourcesTemporarily(pkg, tempDirectory)
```
sschuberth, line 116: Maybe give `saveSourcesTemporarily` a more telling name, like `sourceCodeArchive`?

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -181,4 +217,43 @@ class UploadResultToSw360Command : CliktCommand(
...
+                log.debug("Uploaded attachment ${path.fileName} with type $type to SW360.")
+            }
+        } else {
+            release.also {
+                log.error("Could not upload attachment: " + uploadAttachmentResponse.failedUploads())
+            }
+        }
+    }
+
+    private fun saveSourcesTemporarily(pkg: Package, tempDirectory: File): File? {
```
sschuberth, line 244: This does more than just "saving sources temporarily". Also, we should probably not assume here that the caller is passing in a *temporary* directory; it could be just any directory.

In general I believe it makes sense to generalize the `DownloaderCommand`'s `archive(All)` functions (by returning a `File?`), moving them to a common place in the `cli` module, and making use of them here. Maybe you could pair with @MarcelBochtler for this?

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -89,9 +107,27 @@ class UploadResultToSw360Command : CliktCommand(
...
+                var release = sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
                     .flatMap { sw360ReleaseClient.enrichSparseRelease(it) }
                     .orElseGet { createSw360Release(pkg, sw360ReleaseClient) }
+
+                if (attachSources) {
+                    val tempDirectory = createTempDirectory(pkg.id.toCoordinates()).toFile()
+                    val saveSourcesTemporarily = saveSourcesTemporarily(pkg, tempDirectory)
+
+                    saveSourcesTemporarily?.let {
+                        release = uploadAttachmentToRelease(
```
sschuberth, line 119: Just curious: Is the `release` returned here really any different from the passed-in `release`? I would have assumed that attaching sources does not change the release per se. Because if it doesn't, we could simplify the code here a bit.

sschuberth (changes requested):

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
sschuberth: I'll do some prep-work for this.

sschuberth: Here's a proposal: https://github.com/oss-review-toolkit/ort/pull/4098

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
oheger-bosch: It actually is, as it has been replaced by the instance returned from the server, which now contains the attachment file. That said, given the way the release is later used (just to establish the link to the project), this difference should not really matter, as only the release ID is relevant. So I think it should be safe to stick to the original release instance.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -89,9 +107,27 @@ class UploadResultToSw360Command : CliktCommand(
...
+                            sw360ReleaseClient,
+                            release,
+                            saveSourcesTemporarily.toPath(),
+                            SW360AttachmentType.SOURCE
+                        )
+                    }
+
+                    tempDirectory.safeDeleteRecursively(force = true)
```
oheger-bosch, line 127: To make sure that the directory is always deleted - even when exceptions occur -, the whole block should probably be wrapped in a try-finally statement.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -181,4 +217,43 @@ class UploadResultToSw360Command : CliktCommand(
...
     private fun createReleaseName(pkgId: Identifier) =
         listOf(pkgId.namespace, pkgId.name).filter { it.isNotEmpty() }.joinToString("/")
+
+    private fun uploadAttachmentToRelease(
+        sw360ReleaseClient: SW360ReleaseClientAdapter,
+        release: SW360Release,
+        path: Path,
+        type: SW360AttachmentType
+    ): SW360Release? {
+        val build = AttachmentUploadRequest.builder(release)
```
oheger-bosch, line 227: This variable could have a better name, e.g. `request` or `uploadRequest`.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
bs-ondem: Yes i think it's better to generalize the archive funtion in a common place and both commands can use it.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -181,4 +217,43 @@ class UploadResultToSw360Command : CliktCommand(
 
     private fun createReleaseName(pkgId: Identifier) =
         listOf(pkgId.namespace, pkgId.name).filter { it.isNotEmpty() }.joinToString("/")
+
+    private fun uploadAttachmentToRelease(
+        sw360ReleaseClient: SW360ReleaseClientAdapter,
+        release: SW360Release,
+        path: Path,
+        type: SW360AttachmentType
+    ): SW360Release? {
```
oheger-bosch, line 226: Can this function actually return *null*?

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -181,4 +217,43 @@ class UploadResultToSw360Command : CliktCommand(
...
+        if (download.matches(pkg)) {
+            val zipFile = File(tempDirectory, "${pkg.id.toPath("-")}.zip")
+            tempDownloadDirectory.packZip(
+                zipFile,
+                directoryFilter = { it.name !in VCS_DIRECTORIES }
+            )
+
+            return zipFile
+        }
+        return null
```
oheger-bosch, line 257: In case of an error, there should be a log statement that the sources for the package affected could not be uploaded.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
bs-ondem: Yes, it's different, but I just noticed that it doesn't matter because only the `releaseId` is relevant to for the further run.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
bs-ondem: No 🤣

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
bs-ondem: You mean not downloaded?

oheger-bosch: Well, as a consequence of the failing download, the upload for this package - which is the purpose of this command - will fail, too.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
bs-ondem: @sschuberth Can you check the new commit https://github.com/oss-review-toolkit/ort/pull/4094/commits/04082f32a50ed6f474f58a2f6517f48831e2592c and the reuse of it in the second one?

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -85,9 +101,32 @@ class UploadResultToSw360Command : CliktCommand(
...
                 val name = createReleaseName(pkg.id)
-                sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
+                val release = sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
                     .flatMap { sw360ReleaseClient.enrichSparseRelease(it) }
                     .orElseGet { createSw360Release(pkg, sw360ReleaseClient) }
+
+                if (attachSources) {
+                    val tempDirectory = createOrtTempDir(pkg.id.toPath())
+                    try {
+                        val sourceCodeArchive = saveSourcesTemporarily(pkg, tempDirectory)
```
sschuberth, line 111: As discussion, this could be inlined and the Downloader be created only once with the global configuration:

    val downloader = Downloader(globalOptionsForSubcommands.config.downloader)

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -156,4 +195,45 @@ class UploadResultToSw360Command : CliktCommand(
...
+            }
+        } else {
+            release.also {
+                log.error("Could not upload attachment: " + uploadAttachmentResponse.failedUploads())
+            }
+        }
+    }
+
+    private fun saveSourcesTemporarily(pkg: Package, tempDirectory: File): File? {
+        val tempDownloadDirectory = tempDirectory.resolve("sources")
```
sschuberth, line 223: Maybe better: "sourcesDirectory"

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -85,9 +101,32 @@ class UploadResultToSw360Command : CliktCommand(
...
+
+                        if (sourceCodeArchive != null) {
+                            uploadAttachmentToRelease(
+                                sw360ReleaseClient,
+                                release,
+                                sourceCodeArchive.toPath(),
+                                SW360AttachmentType.SOURCE
+                            )
+                        } else {
+                            log.error("The sources for the package '${pkg.id.toCoordinates()}' could not be " +
```
sschuberth, line 121: `log.error {}`

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -85,9 +101,32 @@ class UploadResultToSw360Command : CliktCommand(
...
                     .flatMap { sw360ReleaseClient.enrichSparseRelease(it) }
                     .orElseGet { createSw360Release(pkg, sw360ReleaseClient) }
+
+                if (attachSources) {
+                    val tempDirectory = createOrtTempDir(pkg.id.toPath())
+                    try {
+                        val sourceCodeArchive = saveSourcesTemporarily(pkg, tempDirectory)
+
+                        if (sourceCodeArchive != null) {
+                            uploadAttachmentToRelease(
```
sschuberth, line 114: Please make this simply return a `Boolean` instead of a `SW360Release` and handle the logging in the caller.

sschuberth (changes requested):
See comments.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
bs-ondem: I completely removed the function and put it to the caller. Now I handle the logging based on the `uploadResult` in the caller directly.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -81,13 +93,55 @@ class UploadResultToSw360Command : CliktCommand(
...
                 val name = createReleaseName(pkg.id)
-                sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
+                val release = sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
                     .flatMap { sw360ReleaseClient.enrichSparseRelease(it) }
                     .orElseGet { createSw360Release(pkg, sw360ReleaseClient) }
+
+                if (attachSources) {
+                    val tempDirectory = createOrtTempDir(pkg.id.toPath())
+                    try {
+                        // First, download the sources of the package into a sources directory, whose root directory is
```
sschuberth, line 108: Nit: Shouldn't this say "whose *parent* directory"?

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -81,13 +93,55 @@ class UploadResultToSw360Command : CliktCommand(
...
+                        // temporary.
+                        val sourcesDirectory = tempDirectory.resolve("sources")
+                        val download = downloader.download(pkg, sourcesDirectory)
+
+                        if (!download.matches(pkg)) {
+                            log.debug { "The sources for the package '${pkg.id.toCoordinates()}' could not be " +
+                                    "uploaded to SW360 due to missing source artifact." }
+                        } else {
```
sschuberth, line 116: I believe we can safely skip this check: If the download fails, a `DownloadException` is thrown. And if it succeeds, the provenance is guaranteed to match the `pkg`. Looking at the provenance is mostly relevant for calls that need to understand from *where* the source code has been downloaded. But in this case we do not care, as long as we *have* the source.

utils/src/main/kotlin/Utils.kt
```diff
@@ -64,6 +66,24 @@ val ortDataDirectory by lazy {
  */
 var printStackTrace = false
 
+/**
+ * Archives a given [input directory][inputDir] into the given []ZIP file][zipFile].
```
sschuberth, line 70: s/Archives/Archive, s/[]/[/

Maybe we should make more clear that not `inputDir` itself is archived, but only its contents (i.e. the ZIP does not contain a single top-level directory). And we should also document `prefix` and mention the return value. Like:

> Archive the contents of [inputDir], omitting common [VCS_DIRECTORIES], to [zipFile] where an optional [prefix] is added to each file. Return a [Result] wrapping the [zipFile] on success, or an exception of failure.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -81,13 +93,55 @@ class UploadResultToSw360Command : CliktCommand(
...
+                val release = sw360ReleaseClient.getSparseReleaseByNameAndVersion(name, pkg.id.version)
                     .flatMap { sw360ReleaseClient.enrichSparseRelease(it) }
                     .orElseGet { createSw360Release(pkg, sw360ReleaseClient) }
+
+                if (attachSources) {
+                    val tempDirectory = createOrtTempDir(pkg.id.toPath())
+                    try {
+                        // First, download the sources of the package into a sources directory, whose root directory is
+                        // temporary.
+                        val sourcesDirectory = tempDirectory.resolve("sources")
```
sschuberth, line 110: Note that theoretically it would not even be required to put sources and the ZIP at different directory levels, as the `archive()` function is smart enough to not ZIP the ZIP itself if sources and ZIP are in the same directory (see the `fileFilter = { it != zipFile }`). But I see how relying on this feature can be a bit confusing, so I'm also fine with keeping this as-is; your call.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -81,13 +93,55 @@ class UploadResultToSw360Command : CliktCommand(
...
+                            // ZIP file of the sources directory and save it in the root directory of it.
+                            // Finally the created ZIP file of the sources can be uploaded to SW360 as an attachment
+                            // of the release.
+                            val zipFile = tempDirectory.resolve("${pkg.id.toPath("-")}.zip")
+                            val archiveResult = archive(sourcesDirectory, zipFile)
+
+                            if (archiveResult.isSuccess) {
+                                val uploadResult = sw360ReleaseClient.uploadAttachments(
+                                    AttachmentUploadRequest.builder(release)
+                                        .addAttachment(zipFile.toPath(), SW360AttachmentType.SOURCE)
```
sschuberth, line 127: If archiving fails, we currently get no notification at all, as any occurred exception is wrapped into the `Result` object returned by `archive()`. A simple way to solve this is to not check `archiveResult.isSuccess` at all, but to replace `zipFile.toPath()` in this line with `archiveResult.getOrThrow().toPath()`.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -81,13 +93,55 @@ class UploadResultToSw360Command : CliktCommand(
...
+                                        .build()
+                                )
+
+                                if (uploadResult.isSuccess) {
+                                    log.debug { "Uploaded attachment ${zipFile.toPath().fileName} with type " +
```
sschuberth, line 132: Maybe lower the log level to `info`?

sschuberth: There is not need to convert `zipFile` to a `Path` first just to get its name.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -81,13 +93,55 @@ class UploadResultToSw360Command : CliktCommand(
...
+                                            "${SW360AttachmentType.SOURCE} to SW360." }
```
sschuberth, line 133: Why is `SW360AttachmentType.SOURCE` expanded as a variable here? Do you really want the result of `SW360AttachmentType.SOURCE.toString()` instead of literal "SW360AttachmentType.SOURCE"?

Anyway, I believe we should reword the log message to also include the release, like

    log.info {
        "Successfully uploaded source attachment '${zipFile.name}' to release $release."
    }

Note that if just `$release` (i.e. `release.toString()`) does not output something meaningful, we should probably explicitly use the *name* and *version* of the release instead.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -81,13 +93,55 @@ class UploadResultToSw360Command : CliktCommand(
...
+                                } else {
+                                    log.error { "Could not upload attachment: " + uploadResult.failedUploads() }
```
sschuberth, line 135: "source attachment" to align with my proposed wording from above.

sschuberth (changes requested):

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
bs-ondem: I prefer to split the sources from the zip file although the `archive()` function is smart enough.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -81,13 +93,52 @@ class UploadResultToSw360Command : CliktCommand(
...
+                        downloader.download(pkg, sourcesDirectory)
+
+                        // After downloading the source files successfully in a sources directory, create a
+                        // ZIP file of the sources directory and save it in the root directory of it.
+                        // Finally the created ZIP file of the sources can be uploaded to SW360 as an attachment
+                        // of the release.
+                        val zipFile = tempDirectory.resolve("${pkg.id.toPath("-")}.zip")
+                        val archiveResult = archive(sourcesDirectory, zipFile)
+
+                        if (archiveResult.isSuccess) {
```
sschuberth, line 120: This success check is still here, which means `archiveResult.getOrThrow().toPath()` in line 123 will never fail, so we have no error handling at all. Please simply remove the check as discussed.

sschuberth (changes requested):

sschuberth (approved):

PR #4097, opened 2021-06-01
cli: Move Constants and Extensions to the utils package

sschuberth:
These are not commands.

### DIFF
--- a/.detekt.yml
+++ b/.detekt.yml
@@ -61,7 +61,7 @@ style:
     excludes: ''
     excludeImports: '
       org.ossreviewtoolkit.analyzer.managers.*,
-      org.ossreviewtoolkit.commands.*,
+      org.ossreviewtoolkit.cli.commands.*,
       org.ossreviewtoolkit.downloader.vcs.*,
       org.ossreviewtoolkit.helper.commands.*,
       org.ossreviewtoolkit.reporter.reporters.*,
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -43,7 +43,7 @@ plugins {
 
 application {
     applicationName = "ort"
-    mainClassName = "org.ossreviewtoolkit.OrtMainKt"
+    mainClassName = "org.ossreviewtoolkit.cli.OrtMainKt"
 }
 
 tasks.withType<ShadowJar> {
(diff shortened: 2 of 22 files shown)


oheger-bosch (dismissed):

MarcelBochtler (changes requested):
You need to add imports for extensions used in `AdvisorCommand`, `AnalyzerCommand`, `DownloaderCommand` and some more.

MarcelBochtler (commented):
`./gradlew :cli:compileFunTestKotlin` is still failing.
But I'm not sure why, as I see that you added the imports.

sschuberth:
> `./gradlew :cli:compileFunTestKotlin` is still failing.
> But I'm not sure why, as I see that you added the imports.

Thanks, I believe it's due to some subtle package name clash. I'll try to resolve it in a preparing commit.

MarcelBochtler (approved):
LGTM.

However, this requires every developer to change their Run configurations (`cli.OrtMainKt`). You might want to announce this in the Slack channel.
Also `./gradlew detekt` fails for me locally, but CI seems to be fine with it.

sschuberth:
> However, this requires every developer to change their Run configurations (Main Class: `org.ossreviewtoolkit.cli.OrtMainKt`). You might want to announce this in the Slack channel.

Good point. Do you thin this deserves a "breaking change" label?

> Also `./gradlew detekt` fails for me locally, but CI seems to be fine with it.

There is a glitch with our detekt-rules not being rebuilt correct, it seems, but I had no time to look into it.

MarcelBochtler:
> Do you think this deserves a "breaking change" label?

From my understanding this label is only required if user facing configurations / functionalities change.
IMO developers are not users therefore the label is not required.

MarcelBochtler:
> There is a glitch with our detekt-rules not being rebuilt correct, it seems, but I had no time to look into it.

Do you know a work around to fix it?
Cleaning and rerunning tasks seems not to work.

sschuberth:
> Do you know a work around?

Yes, `rm ./detekt-rules/build/libs/detekt-rules.jar` should do the trick.

sschuberth:
> There is a glitch with our detekt-rules not being rebuilt correct, it seems, but I had no time to look into it.

I'll try to look into it tonight.

sschuberth:
> > Also `./gradlew detekt` fails for me locally, but CI seems to be fine with it.
> 
> There is a glitch with our detekt-rules not being rebuilt correct, it seems, but I had no time to look into it.

Looks like this is a [known issue](https://github.com/detekt/detekt/issues/2645#issuecomment-622217315) with custom *detekt* rules and the Gradle daemon / Java's `ServiceLoader` 😢

PR #4100, opened 2021-06-02
Remove (some) references to JCenter

sschuberth:
Note that the test projects partly still refer to JCenter, but that probably is even a good idea, to verify that we can deal with projects that still refer to JCenter.

### DIFF
--- a/scripts/import_proxy_certs.sh
+++ b/scripts/import_proxy_certs.sh
@@ -22,7 +22,7 @@ if [ -z "$https_proxy" ]; then
     exit
 fi
 
-CONNECT_SERVER="jcenter.bintray.com:443"
+CONNECT_SERVER="repo.maven.apache.org:443"
 
 TEMP_DIR="$(mktemp -d)"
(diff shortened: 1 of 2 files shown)


scripts/import_proxy_certs.sh
```diff
@@ -22,7 +22,7 @@ if [ -z "$https_proxy" ]; then
     exit
 fi
 
-CONNECT_SERVER="jcenter.bintray.com:443"
+CONNECT_SERVER="repo.maven.apache.org:443"
```
fviernau, line 25: commit message: typo in "do esnot"

fviernau: such much with server: with -> which

fviernau (changes requested):

scripts/import_proxy_certs.sh
fviernau, line 25: Commit-msg: why should the server **only** be reachable through the proxy?

PR #4101, opened 2021-06-02
GoMod: Use the source artifact URL only as fallback

fviernau:
Using the source artifact URL pointing to a GO proxy is more fragile
compared to using the VCS info. So, avoid using / scanning that source
artifact URL by setting it only in case the VCS info is not available.

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
@@ -106,7 +106,7 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: "https://proxy.golang.org/github.com/fatih/color/@v/v1.7.0.zip"
+    url: ""
     hash:
       value: ""
       algorithm: ""
@@ -132,7 +132,7 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: "https://proxy.golang.org/github.com/golang/protobuf/@v/v1.2.0.zip"
+    url: ""
     hash:
       value: ""
       algorithm: ""
@@ -158,7 +158,7 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: "https://proxy.golang.org/github.com/hashicorp/errwrap/@v/v1.0.0.zip"
+    url: ""
     hash:
       value: ""
       algorithm: ""
@@ -184,7 +184,7 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: "https://proxy.golang.org/github.com/hashicorp/go-multierror/@v/v1.0.0.zip"
+    url: ""
     hash:
       value: ""
       algorithm: ""
@@ -210,7 +210,7 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: "https://proxy.golang.org/github.com/mattn/go-colorable/@v/v0.1.4.zip"
+    url: ""
     hash:
       value: ""
       algorithm: ""
@@ -236,7 +236,7 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: "https://proxy.golang.org/github.com/mattn/go-isatty/@v/v0.0.10.zip"
+    url: ""
     hash:
       value: ""
       algorithm: ""
@@ -262,7 +262,7 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: "https://proxy.golang.org/github.com/mattn/go-isatty/@v/v0.0.8.zip"
+    url: ""
     hash:
       value: ""
       algorithm: ""
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -185,17 +185,24 @@ class GoMod(
         return result
     }
 
-    private fun createPackage(id: Identifier): Package =
-        Package(
+    private fun createPackage(id: Identifier): Package {
+        val vcsInfo = id.toVcsInfo()
+
+        return Package(
             id = Identifier(managerName, "", id.name, id.version),
             authors = sortedSetOf(), // Go mod doesn't support author information.
             declaredLicenses = sortedSetOf(), // Go mod doesn't support declared licenses.
             description = "",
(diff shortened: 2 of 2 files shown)


tsteenbe:
@fviernau  I noticed in OSS reviews as well that source artifact URL are sometimes invalid so this look like a good idea to me.

sschuberth:
> pointing to a GO proxy is more fragile

Please explain *why* it is fragile. In our pipeline, we've configured ORT to prefer source artifacts over VCS locations. So this change could have a negative impact on us.

sschuberth:
> Please explain _why_ it is fragile.

Ah, that's probably because of this comment in `getSourceArtifactForPackage`, right:

https://github.com/oss-review-toolkit/ort/blob/9af404c2a684507525ead90c74e106096e09157e/analyzer/src/main/kotlin/managers/GoMod.kt#L202-L203

sschuberth:
As I believe all Go modules are hosted as source in Git (and mostly on GitHub), I wonder whether we could do something else and simply leverage the new `VcsHost.toArchiveDownloadUrl` functionality to always create artifact download URLs from VCS locations in the Go case.

fviernau:
> As I believe all Go modules are hosted as source in Git (and mostly on GitHub), I wonder whether we could do something else and simply leverage the new `VcsHost.toArchiveDownloadUrl` functionality to always create artifact download URLs from VCS locations in the Go case.

Why would `VcsHost.toArchiveDownloadUrl()` be the option of choice?

sschuberth:
> Why would `VcsHost.toArchiveDownloadUrl()` be the option of choice?

Why not? From skipping through the comments in `getSourceArtifactForPackage`, the use of `toArchiveDownloadUrl` sounds way more reliable.

fviernau:
> Why not? 

Because nowhere in ORT we use and archive URL for obtaining the source tree from VCS INFO?

sschuberth:
> Because nowhere in ORT we use an archive URL for obtaining the source tree from VCS INFO?

We actually do as of https://github.com/oss-review-toolkit/ort/pull/4086.

fviernau:
>  I wonder whether we could do something else and simply leverage the new `VcsHost.toArchiveDownloadUrl` functionality to always create artifact download URLs from VCS locations in the Go case.

I wonder if there was a misunderstanding of this sentence on my side. Did you actually mean to do what you described above (1) in addition to setting the VCS INFO or (2) instead of?

I understood it as (2), but if you meant (1) then I would certainly agree.

sschuberth:
> I somehow understood / assumed you mean (2), but if you meant (1) then I would certainly agree.

Yes, I meant additionally to VcsInfo, not instead of VcsInfo. (But instead of the GOPROXY stuff.)

fviernau:
While doing the implementation I noticed that `toArchiveDownloadUrl()` does not adhere to the VCS path. 
Because of that I believe that's not an option. 

 @sschuberth any thoughts on how to go forward now?

sschuberth:
> While doing the implementation I noticed that `toArchiveDownloadUrl()` does not adhere to the VCS path.

I wasn't aware that you / Go needs to go down to the path level. How does Go itself do this then, i.e. what happens internally in Go when a Go module is downloaded whose source code is in the sub-directory of a bigger (mono-)repo? Does it use Git sparse checkout then?

Anyway, going one step back again: Quoting the [mod docs](https://golang.org/ref/mod#modules-overview) "Modules may be downloaded directly from version control repositories or from module proxy servers". So I wonder why we try to construct source artifact URLs based on *every* package, instead of only for those packages where a proxy URL was used in the original `go.mod` file. In other words, we should simply take any proxy URL to a source artifact from `go.mod`, but only if it's present, and leave the source artifact URL empty in cases where there is no proxy URL in `go.mod`.

sschuberth:
> So I wonder why we try to construct source artifact URLs based on _every_ package, instead of only for those packages where a proxy URL was used in the original `go.mod` file.

Looks like my assumption was wrong that using a proxy URL would have an explicit syntax in `go.mod`. Apparently, the syntax to refer to modules from `go.mod` is always the same, and only the `GOPROXY` environment variable determines whether the module should be retrieved from a proxy, for from VCS / upstream directly.

fviernau:
@sschuberth reading the above I'm not certain if you changed your mind about the PR as-is. If you're fine with it as-is in general I can make minor fixes and merge it. Otherwise I would just close it, up to you.

sschuberth:
> I'm not certain if you changed your mind about the PR as-is.

I'm not sure either 😆 

> I assume this PROXY is supposed to be some transparent concept for performance optimization and is thus not the best way to specify a source code location.

I though so, too, but what's still unclear to me if it's even possible that a module is *only* available through a proxy, and *not* via VCS. Because if that's not possible, we'd not even need the fallback to the source artifact that you've implemented, and we'd simply *never* have source artifacts for Go modules.

fviernau:
> I though so, too, but what's still unclear to me if it's even possible that a module is only available through a proxy, and not via VCS. Because if that's not possible, we'd not even need the fallback to the source artifact that you've implemented, and we'd simply never have source artifacts for Go modules.

I'd agree with aspects of it, in particular I'd say "if we had a complete correct implementation of determining the VCS / that means we ideally use the original tool / library for that" then a proxy URL is not needed. As-is I'm just not 100% certain if the VCS detection is good enough to strip the fallback. I believe we should keep it for now. 

Also removing the fallback it would nicely make up a next incremental improvement step.

Note that the VCS detection currently relies on VCS host class. So it is VCS host specific. The Go mod internals are VCS host specific as well, but in a different way. So there might be VCS hosts where our current approach does not work. Is that last sentence alone convincing enough to keep the fallback ?

sschuberth:
> Note that the VCS detection currently relies on VCS host class. So it is VCS host specific.

Actually, what makes the code *really* host specific is that we call `VcsHost` only if the `Identifier`'s `name` starts with "github.com", which is quite dumb, as we then don't benefit from the auto-detection / all hosts that `VcsHost` supports. I'll make a PR based on yours to change that.

So let's merge this; I believe the logic becomes clearer with my follow-up PR.

sschuberth (approved):

sschuberth:
> So let's merge this; I believe the logic becomes clearer with my follow-up PR.

Here it is: https://github.com/oss-review-toolkit/ort/pull/4115

PR #4102, opened 2021-06-02
fossid-webapp: Fix regression in response deserialization

nnobelis:
This bug addresses a regression introduced by https://github.com/oss-review-toolkit/ort/pull/4045.

The FossID `get_all_scans` operation has a bug and returns an array if only one scan is returned, otherwise a map.
We have to handle this case, plus the one the other PR was dealing with i.e. handling `data:false` when there is no result.

The result is a new Jackson serializer that handles all these cases.

This commit also contains a new test exhibiting the bug.

### DIFF
--- a/clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
+++ b/clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
@@ -22,7 +22,7 @@ import com.github.tomakehurst.wiremock.client.WireMock
 import com.github.tomakehurst.wiremock.core.WireMockConfiguration
 
 import io.kotest.core.spec.style.StringSpec
-import io.kotest.matchers.maps.beEmpty
+import io.kotest.matchers.collections.beEmpty
 import io.kotest.matchers.maps.shouldContain
 import io.kotest.matchers.nulls.beNull
 import io.kotest.matchers.should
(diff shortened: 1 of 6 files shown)


sschuberth:
There seems to be a compile error @nnobelis.

nnobelis:
> There seems to be a compile error @nnobelis.

Fixed thanks.

oheger-bosch (approved):
This is really a cool solution for these problems with the FossId-API. You are now the Jackson-Serialization-Overlord!

clients/fossid-webapp/src/test/kotlin/FossIdClientNewProjectTest.kt
```diff
@@ -103,7 +103,7 @@ class FossIdClientNewProjectTest : StringSpec({
         service.listScansForProject("", "", PROJECT_CODE) shouldNotBeNull {
             checkResponse("list scans")
             data shouldNotBeNull {
-                this should beEmpty()
+                this.isEmpty() shouldBe true
```
sschuberth, line 106: You can drop `this.` here now.

clients/fossid-webapp/src/test/kotlin/FossIdClientReturnTypeTest.kt
```diff
@@ -74,7 +81,19 @@ class FossIdClientReturnTypeTest : StringSpec({
...
+                this should beEmpty()
+            }
+        }
+    }
+
+    "Scans for project can be listed when there is exactly one" {
+        service.listScansForProject("", "", PROJECT_CODE_3) shouldNotBeNull {
+            checkResponse("list scans")
+            data shouldNotBeNull {
+                this.size shouldBe 1
```
sschuberth, line 93: You can drop this. here now.

sschuberth (approved):

Issue #3764, opened 2021-03-16
Dockerfile shouldn't be run with ROOT permissions


Martin-Idel-SI:
When using the default Dockerfile, we get errors when using bower package manager:

> ERROR org.ossreviewtoolkit.analyzer.PackageManager - Resolving dependencies for 'bower.json' failed with: IOException: Running 'bower install' in '/input/project/...' failed with exit code 1:
bower ESUDO         Cannot be run with sudo
Additional error details:
Since bower is a user command, there is no need to execute it with superuser permissions.
If you're having permission errors when using bower without sudo, please spend a few minutes learning more about how your system should work and make any necessary repairs.
http://www.joyent.com/blog/installing-node-and-npm
https://gist.github.com/isaacs/579814
You can however run a command with sudo using "--allow-root" option

This issue will also appear in the reports later on, so it's a problem.

We could either add an option to run bower commands with `--allow-root` as suggested in the error message or add a user to the official Dockerfile and run without root permissions. This would be the cleaner solution in my opinion.

So the real question is whether there is something in the Dockerfile which inherently requires sudo permissions while running.

sschuberth:
> So the real question is whether there is something in the Dockerfile which inherently requires sudo permissions while running.

I'm not aware of anything. We just run as root as that's Docker's default.

tsteenbe:
Maybe we should run https://github.com/hadolint/hadolint - which is a Docker best practices linter -  over our DockerFile using the https://github.com/marketplace/actions/hadolint-action.

We could also look at implementing the recommendations in https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html - include steps how to use an unprivileged user

mnonnenmacher:
Also see my comment here: https://github.com/oss-review-toolkit/ort/pull/4103#issuecomment-854137392

sschuberth:
> Maybe we should run https://github.com/hadolint/hadolint - which is a Docker best practices linter - over our DockerFile using the https://github.com/marketplace/actions/hadolint-action.

I gave `hadolint` a try, and I'm not convinced of the tool. I was immediately running into multiple issues [[1]] [[2]] [[3]].

[1]: https://github.com/hadolint/hadolint/issues/285
[2]: https://github.com/hadolint/hadolint/issues/422
[3]: https://github.com/hadolint/hadolint/issues/679

sschuberth:
[...]

PR #4103, opened 2021-06-02
Dockerfile: Fix an issue with the GoMod support

fviernau:
When running ORT in the docker image as non-root user, GoMod fails with
[1] when running [2]. Grant the required file system permissions to fix
that issue.

[1] `go: writing go.mod cache: mkdir /go/pkg: permission denied`
[2] `go mod vendor -v`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -137,6 +137,7 @@ RUN /opt/ort/bin/import_proxy_certs.sh && \
     tar -C /opt -xzf go$GO_VERSION.linux-amd64.tar.gz && \
     rm go$GO_VERSION.linux-amd64.tar.gz && \
     mkdir -p $GOPATH/bin && \
+    chmod a+rw $GOPATH -R \
     curl -ksS https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | sh && \
     curl -ksS https://raw.githubusercontent.com/commercialhaskell/stack/v$HASKELL_STACK_VERSION/etc/scripts/get-stack.sh | sh && \
     # Install SDKs required for analysis.


sschuberth:
We don't currently support running our Docker image as non-root, also see https://github.com/oss-review-toolkit/ort/issues/3764. 

Instead of this band-aid, I feel we should address https://github.com/oss-review-toolkit/ort/issues/3764 and add proper / generic support for running as non-root. That would also allow to revert https://github.com/oss-review-toolkit/ort/pull/3879.

mnonnenmacher:
> We don't currently support running our Docker image as non-root, also see #3764.
> 
> Instead of this band-aid, I feel we should address #3764 and add proper / generic support for running as non-root. That would also allow to revert #3879.

I think we should merge this as a quick fix just like #3879 because it seems that fixing #3764 is a lot more effort. Also, I couldn't find it anywhere in the docs that were require users to run the image as root?

sschuberth:
> I think we should merge this as a quick fix just like #3879 because it seems that fixing #3764 is a lot more effort.

It's basically just a matter of restoring the `runAsUser` logic from 43130c72a0ef9332afde73e3ef36394612e6f1b0 IMO; not really much effort.

> Also, I couldn't find it anywhere in the docs that were require users to run the image as root?

See this [comment](https://github.com/oss-review-toolkit/ort/issues/3764#issuecomment-800388243), that's simply `docker run``s default, so there probably no need to document that. On the contrary, if we'd start to *require* a non-root user (in order to revert #3879), then *that* would need to get documented.

mnonnenmacher:
> > I think we should merge this as a quick fix just like #3879 because it seems that fixing #3764 is a lot more effort.
> 
> It's basically just a matter of restoring the `runAsUser` logic from [43130c7](https://github.com/oss-review-toolkit/ort/commit/43130c72a0ef9332afde73e3ef36394612e6f1b0) IMO; not really much effort.

I didn't know that you have an idea how to fix that, it wasn't mentioned before. I'm no Docker expert after all so I would have to do research first to understand what the best solution is for this issue. The OWASP document @tsteenbe linked below your comment also recommends to not use root, and the second suggestion to create a user inside the container seems to be what @Martin-Idel-SI proposed in #3764.

Anyway, I haven't talked to @fviernau yet so I don't know how urgent this fix is, or if it would be an option to look deeper into the root cause. I think I just found it a bit contradictory that you rejected this quick fix when you actually went for a quick fix for Bower as well.

sschuberth:
> I think I just found it a bit contradictory that you rejected this quick fix when you actually went for a quick fix for Bower as well.

That's because the Bower quick fix goes exactly in the opposite direction: It allows to run Bower (in certain cases) with the current default of running Docker as root. @fviernau's change on the other hand is a work-around to run Docker in the currently unsupported non-root mode.

sschuberth:
So is this issue specially about GitLab CI runners? I'm asking because I'm thinking about solving #3764 properly by running as a user inside the Docker container. But the question is: As which user? There are scenarios where the uid / name of a user on the host, and the user running in Docker should be aligned, e.g. if you mount a volume from the host; then the Docker user needs to have permission to access the host's directory. So in case of GitLab CI runners, is there a volume mounted? And which users does the GitLab host give access to that volume?

sschuberth:
> And which users does the GitLab host give access to that volume?

Or phrasing that differently: When GitLab does `docker run` for you, what does it pass as `-u`? Or is it not GitLab itself that does `docker run`, but you explicitly yourself, in your CI script?

mnonnenmacher:
I don't know the details because I haven't worked with it myself. But I believe on GitLab it runs with the root users, I don't know if this is configurable, but it seems that basically can just provide the name of the image and an optional entry point script:
https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#available-settings-for-image

This is also not specific to GitLab, for example in Jenkins pipelines you can also provide the name of an image plus additional options using the pipelines DSL:
https://www.jenkins.io/doc/book/pipeline/docker/

And on Azure Pipelines it's also similar:
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops#single-job

The point is, I think our image should work out of the box without requiring a separate launch script to make it easy to use. How we should configure the user, I don't know. It's probably best to look how other images solve this issue (e.g. the Jenkins image has a jenkins user with UID 1000: https://github.com/jenkinsci/docker/blob/master/README.md) or ask a Docker expert for help.

The non-root user issue we have on Jenkins, but there we use a custom script. However, for now we have solved the issue by creating our own image that inherits from the ORT image and put our custom configuration there.

mnonnenmacher:
@fviernau Based on what I wrote above I think we can close this PR because we have a solution that works for us and it seems we need more time to find a proper solution for the public image.

fviernau:
> @fviernau Based on what I wrote above I think we can close this PR because we have a solution that works for us and it seems we need more time to find a proper solution for the public image.

Agree.

PR #4104, opened 2021-06-02
- Set proxy for git in case of reverse proxy

heliocastro:


downloader/src/main/kotlin/vcs/Git.kt
```diff
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.downloader.vcs
 
 import com.jcraft.jsch.JSch
+import com.jcraft.jsch.ProxyHTTP
```
sschuberth, line 23: Commit message:

> - Set proxy for git in case of reverse proxy 

Please omit the leading "- " and s/git/Git/

downloader/src/main/kotlin/vcs/Git.kt
```diff
@@ -69,8 +71,14 @@ class Git : VersionControlSystem(), CommandLineTool {
...
             val sessionFactory = object : JschConfigSessionFactory() {
-                @Suppress("EmptyFunctionBlock")
-                override fun configure(hc: OpenSshConfig.Host, session: Session) {}
+                override fun configure(hc: OpenSshConfig.Host, session: Session) {
+                    // If SSH protocol is using in environments where reverse proxy
+                    // is needed ( corkscrew style ), set env proxy if exists
+                    if (! Os.env["http_proxy"].isNullOrEmpty()) {
+                        val uri = URI(Os.env["http_proxy"])
+                        session.setProxy(ProxyHTTP(uri.host, uri.port))
+                    }
```
sschuberth, line 80: I'm afraid that this would break things for setups where you want / need to use an HTTP proxy, but want to connect via SSH directly. I currently have no good proposal how to distinguish these use-cases. Is there anything in the syntax of the string returned by `Os.env["http_proxy"]` that we could look at?

sschuberth (changes requested):

downloader/src/main/kotlin/vcs/Git.kt
heliocastro: The syntax is simple: http_proxy=http://localhost:3128
We use in our machines a ntml proxy on dev machines to avoid setup individually every single app, but is not transparent

Maybe add an option on ort.conf global downloader config like use_gitssh_proxy = true would solve the issue above ?

downloader/src/main/kotlin/vcs/Git.kt
heliocastro: Done

downloader/src/main/kotlin/vcs/Git.kt
sschuberth: > The syntax is simple: http_proxy=http://localhost:3128

So there is nothing special about the syntax that would help us to decide whether the proxy should be used to tunnel SHH through HTTP, or as a "regular" HTTP proxy only.

It looks like tools like `corkscrew` don't consider the `http(s)_proxy` environment variables by themselves; and that's exactly for the above mentioned reason. 

I see two possible ways forward: Either do like you propose and introduce an ORT downloader configuration variable to choose whether `http(s)_proxy` should be considered for SSH over HTTP(S), too. Or we simply come up with different environment variables to use, like `ssh_over_http_proxy.` I have a tendency towards the former as it avoid coming up with very ORT-specific conventions, and there is a change that SSH over HTTP user already have set `http(s)_proxy` correctly.

sschuberth: So I propose to add `val useHttpProxyForSsh: Boolean` to `DownloaderConfiguration`, but that will require some refactoring as the `Git` class does not know anything about a `DownloaderConfiguration` yet...

heliocastro:
No needed anymore

PR #4106, opened 2021-06-02
SpdxExpression: Add a function to sort expressions lexicographically

mnonnenmacher:
Add a function to sort SPDX expressions lexicographically and use it for the effective license in the Static HTML reporter for a start.

### DIFF
--- a/spdx-utils/src/main/kotlin/SpdxExpression.kt
+++ b/spdx-utils/src/main/kotlin/SpdxExpression.kt
@@ -125,6 +125,11 @@ sealed class SpdxExpression {
     abstract fun normalize(mapDeprecated: Boolean = true): SpdxExpression
 
     /**
+     * Sort this expression lexicographically.
+     */
+    open fun sort(): SpdxExpression = this
+
+    /**
      * Validate this expression. [strictness] defines whether only the syntax is checked
      * ([ALLOW_ANY][Strictness.ALLOW_ANY]), or semantics are also checked but deprecated license identifiers are
      * allowed ([ALLOW_DEPRECATED][Strictness.ALLOW_DEPRECATED]) or only current license identifiers are allowed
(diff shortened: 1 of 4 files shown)


spdx-utils/src/main/kotlin/SpdxExpression.kt
```diff
@@ -255,6 +260,28 @@ class SpdxCompoundExpression(
...
+            }
+
+            addChildren(expression.left)
+            addChildren(expression.right)
+
+            return children
+        }
+
+        val children = getSortedChildrenWithSameOperator(this)
+        return parse(children.map { "($it)" }.sorted().joinToString(" ${operator.name} "))
```
sschuberth, line 282: Can you maybe add some code comments about the idea of the algorithm?

sschuberth: I was asking specifically as I thought `WITH` would still be an operator, but we changed that a while back. But what IÄm after: Would this break as soon as a non-commutative operator is added to `SpdxOperator`? Should we therefore explicitly check for `AND` and `OR`?

spdx-utils/src/main/kotlin/SpdxExpression.kt
```diff
@@ -124,6 +124,11 @@ sealed class SpdxExpression {
      */
     abstract fun normalize(mapDeprecated: Boolean = true): SpdxExpression
 
+    /**
+     * Sort this expression lexicographically.
```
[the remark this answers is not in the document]
mnonnenmacher: Isn't this (a) obvious and (b) an implementation detail that shouldn't be in the docs of the public API? We don't have a similar comment for e.g. `disjunctiveNormalForm()`.

spdx-utils/src/main/kotlin/SpdxExpression.kt
mnonnenmacher: I have rewritten the algorithm to use `reduce` instead of building and parsing a string for better efficiency. I think this should be self-explanatory now. Also it contains a `when` on the operator now which enforces revisiting the code if any new operators are added.

spdx-utils/src/main/kotlin/SpdxExpression.kt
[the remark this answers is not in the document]
sschuberth: You're right that my "Only needs to be overridden" is an implementation detail an should not be mentioned. The "support multiple sub-expressions" part was not super-obvious to me, and it took me a few seconds to realize, but I'm fine with leaving the comment as-is.

spdx-utils/src/main/kotlin/SpdxExpression.kt
mnonnenmacher: I have also added a comment to the helper function now to explain what it does.

sschuberth (approved):

PR #4108, opened 2021-06-03
Align severity handling for CLI exit codes

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -36,19 +36,13 @@ import com.github.ajalt.clikt.parameters.types.file
 import java.io.File
 
 import kotlin.system.exitProcess
-import kotlin.time.measureTime
-import kotlin.time.measureTimedValue
 
 import org.apache.logging.log4j.Level
 import org.apache.logging.log4j.core.config.Configurator
 
 import org.ossreviewtoolkit.cli.commands.*
-import org.ossreviewtoolkit.model.HashAlgorithm
-import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.config.LicenseFilenamePatterns
 import org.ossreviewtoolkit.model.config.OrtConfiguration
-import org.ossreviewtoolkit.model.readValue
-import org.ossreviewtoolkit.model.writeValue
 import org.ossreviewtoolkit.utils.Environment
 import org.ossreviewtoolkit.utils.ORT_CONFIG_DIR_ENV_NAME
 import org.ossreviewtoolkit.utils.ORT_CONFIG_FILENAME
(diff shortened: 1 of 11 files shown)


cli/src/main/kotlin/commands/AdvisorCommand.kt
```diff
@@ -36,10 +36,12 @@ import com.github.ajalt.clikt.parameters.types.file
 
 import org.ossreviewtoolkit.advisor.Advisor
 import org.ossreviewtoolkit.cli.GlobalOptions
+import org.ossreviewtoolkit.cli.printSeverityStats
```
mnonnenmacher, line 39: Commit message typo: "iss**u**es"

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -280,16 +281,6 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
...
-        val warningCount = counts[Severity.WARNING] ?: 0
-        val hintCount = counts[Severity.HINT] ?: 0
-
-        if (errorCount > 0 || warningCount > 0) {
-            println("Found $errorCount errors, $warningCount warnings, $hintCount hints.")
-            throw ProgramResult(2)
-        }
-
-        println("Found $hintCount hints only.")
+        if (printSeverityStats(counts, Severity.HINT)) throw ProgramResult(2)
```
mnonnenmacher, line 284: This reads like `ProgramResult(2)` is thrown if the stats could be printed, I'd prefer to find a better name for the function but don't have a good suggestion. Another option would be to split `printSeverityStats` and the `hasSevereIssues` check into separate functions.

mnonnenmacher (dismissed):

cli/src/main/kotlin/commands/EvaluatorCommand.kt
sschuberth: Or how about simply moving the `throw ProgramResult(2)` into the function (as that's what all callers currently do) and not return anything?

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -71,6 +72,35 @@ data class GlobalOptions(
...
+
+    fun getSeverityCount(severity: Severity) =
+        counts.getOrDefault(severity, 0).also { hasSevereIssues = it > 0 && severity > threshold }
+
+    val hintCount = getSeverityCount(Severity.HINT)
+    val warningCount = getSeverityCount(Severity.WARNING)
+    val errorCount = getSeverityCount(Severity.ERROR)
+
+    println("Found $errorCount errors, $warningCount warnings, $hintCount hints.")
+    if (hasSevereIssues) println("There are severe issues.")
```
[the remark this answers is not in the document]
sschuberth: I went for something even more verbose.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
sschuberth: I did that now in the new iteration.

oheger-bosch (approved):

cli/src/main/kotlin/commands/EvaluatorCommand.kt
mnonnenmacher: I still find it very unexpected that a function with the name "print..." has such a side effect, but it's merged now, so be it.

sschuberth: I'm more than happy to make a follow-up PR to rename, it's not carved in stone. But the alignment of the exit code was more important to me right now. Do you have a proposal for a name by now?

PR #4115, opened 2021-06-04
GoMod: Support all hosts known to VcsHost

sschuberth:
Previously, VcsHost was only used for "github.com". Make use of the host
detection in VcsHost instead to support more hosts.

This way the fallback to proxy artifacts, for which we have only limited
support according to the code comments, might be triggered less often.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -36,6 +36,7 @@ import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.RemoteArtifact
 import org.ossreviewtoolkit.model.Scope
 import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.utils.CommandLineTool
@@ -186,7 +187,7 @@ class GoMod(
     }
 
     private fun createPackage(id: Identifier): Package {
-        val vcsInfo = id.toVcsInfo()
+        val vcsInfo = id.toVcsInfo().takeUnless { it.type == VcsType.UNKNOWN } ?: VcsInfo.EMPTY
 
         return Package(
             id = Identifier(managerName, "", id.name, id.version),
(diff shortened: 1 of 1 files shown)


sschuberth:
The `PipFunTest` failure is getting fixed separately by https://github.com/oss-review-toolkit/ort/pull/4118.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -297,12 +298,8 @@ private fun getRevision(version: String): String {
...
-    if (name.startsWith("github.com")) {
-        val projectUrl = "https://${name.removeSuffix("/")}"
-        val vcsInfo = VcsHost.toVcsInfo(projectUrl)
-
-        vcsInfo.copy(revision = getRevision(version))
-    } else {
-        VcsInfo.EMPTY
-    }
+internal fun Identifier.toVcsInfo(): VcsInfo {
+    val projectUrl = name.takeIf { it.startsWith("http") } ?: "https://$name"
```
fviernau, line 302: This code can now potentially be called for VCS types other than `git`, e.g. subversion.
IIUC the project URL construction for subversion does not work as the URL should start with 'svn:'?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
sschuberth: I've generalized the check accordingly.

fviernau: Maybe one step back. Your change seems to assume that the `name` part of the `Identifier` can have a `protocol` prefix.
Is that assumption actually reasonable to make?

Apart from that, do `https` URLs work for  SVN?

sschuberth: > Your change seems to assume that the `name` part of the `Identifier` can have a `protocol` prefix. Is that assumption actually reasonable to make?

I don't know enough about Go to tell. I just didn't want prepend a protocol without checking whether it's already there.

*Edit:* But according to https://golang.org/ref/mod#go-mod-file-ident ("limited ASCII punctuation") it can never contain a protocol as ":" is not allowed. So I'll simply drop the check.

> Apart from that, do `https` URLs work for SVN?

[Yes](https://subversion.apache.org/quick-start#what-is-a-repository).

fviernau: Thanks!

fviernau (approved):

PR #4116, opened 2021-06-04
fossid-webapp: Search and reuse existing scans

nnobelis:
The logic of the scanner has been changed. Now we search for existing
scans of a given package before creating a new scan. If such an existing
scan is found, the scanner will check if repository data has been
downloaded and if this scan has completed. If some of these conditions is
not fulfilled, the scanner will wait until it has.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -42,6 +42,7 @@ import org.ossreviewtoolkit.clients.fossid.getProject
 import org.ossreviewtoolkit.clients.fossid.listIdentifiedFiles
 import org.ossreviewtoolkit.clients.fossid.listIgnoredFiles
 import org.ossreviewtoolkit.clients.fossid.listMarkedAsIdentifiedFiles
+import org.ossreviewtoolkit.clients.fossid.listScansForProject
 import org.ossreviewtoolkit.clients.fossid.model.Project
 import org.ossreviewtoolkit.clients.fossid.model.status.DownloadStatus
 import org.ossreviewtoolkit.clients.fossid.model.status.ScanState
@@ -221,17 +222,31 @@ class FossId(
                         .checkResponse("create project")
                 }
 
-                log.info { "Creating scan for URL $url and revision $revision ..." }
+                val scans = service.listScansForProject(user, apiKey, projectCode)
+                    .checkResponse("list scans for project").data
+                checkNotNull(scans)
 
-                val scanCode = createScan(projectCode, url, revision)
+                val existingScan = scans.sortedByDescending { it.id }.find { scan ->
+                    scan.gitBranch == revision && scan.gitRepoUrl == url
+                }
 
-                log.info { "Initiating download from Git repository ..." }
+                val scanCode = if (existingScan == null) {
+                    log.info { "No scan found for $url and revision $revision. Creating scan ..." }
+                    val scanCode = createScan(projectCode, url, revision)
 
-                service.downloadFromGit(user, apiKey, scanCode)
-                    .checkResponse("download data from Git", false)
+                    log.info { "Initiating data download ..." }
+                    service.downloadFromGit(user, apiKey, scanCode)
+                        .checkResponse("download data from Git", false)
+                    scanCode
+                } else {
+                    log.info { "Scan found for $url and revision $revision." }
+
+                    requireNotNull(existingScan.code) {
+                        "FossId returned a null scancode for an existing scan"
+                    }
+                }
 
                 checkScan(scanCode)
-
                 val rawResults = getRawResults(scanCode)
                 val resultsSummary = createResultSummary(startTime, rawResults)
 


sschuberth:
> The logic of the scanner has been changed.

What does this refer to? To me it sounds that (a new release of) the upstream FOSSID server as somehow changed its behavior, and we need to adapt to it. Is it that what you want to say? Or do you rather want to say that this commit of yours changes the logic our scanner implementation? If so, please stick to the imperative / present tense for clarity.

> Now we search for existing scans of a given package before creating a new scan.

Please avoid "we", "I" etc, and use passive voice in commit messages, like you correctly do in the next sentence ("scan is found").

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,35 @@ class FossId(
                         .checkResponse("create project")
                 }
 
-                log.info { "Creating scan for URL $url and revision $revision ..." }
+                val scans = service.listScansForProject(user, apiKey, projectCode)
+                    .checkResponse("list scans for project").data
+                requireNotNull(scans)
```
sschuberth, line 227: AFAIU this can be dropped as `withDataCheck = true` is the default for `checkResponse`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,35 @@ class FossId(
...
 
-                val scanCode = createScan(projectCode, url, revision)
-
-                log.info { "Initiating download from Git repository ..." }
+                val existingScan = scans.sortedByDescending { scan -> scan.id }.find { scan ->
```
sschuberth, line 229: Nit: For one-liners / when there is no variable shadowing, we usually stick to `it`, i.e. change `scan -> scan.id` to just `it.id`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,35 @@ class FossId(
...
+                    scan.gitBranch == revision && scan.gitRepoUrl == url
+                }
 
-                service.downloadFromGit(user, apiKey, scanCode)
-                    .checkResponse("download data from Git", false)
+                val scanCode = kotlin.run {
```
sschuberth, line 233: Why wrapping into `run`? IMO you can just drop that outer level.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,35 @@ class FossId(
...
+                    if (existingScan == null) {
+                        log.info { "No scan found for url=$url and revision=$revision. Creating scan ..." }
```
sschuberth, line 235: As from the syntax of a URL it's usually clear that it's a URL, I believe you can drop "url=", and in "revision=" please replace the "=" with " " to make a fluently readable sentence.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,35 @@ class FossId(
...
+                        val scanCode = createScan(projectCode, url, revision)
+
+                        log.info { "Launching data download ..." }
```
sschuberth, line 238: Nit: Maybe "Initiating" or simply "Starting" instead of "Launching"?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,35 @@ class FossId(
...
+                        responseDownload.checkResponse("download data from Git", false)
+                        scanCode
+                    } else {
+                        log.info { "Scan found for url=$url and revision=$revision." }
+
+                        val scanCode = existingScan.code
+                        requireNotNull(scanCode) {
+                            "FossId returned a null scancode for an existing scan"
+                        }
+                        scanCode
```
sschuberth, line 249: As `requireNotNull` returns a non-null version of its argument, you can omit this line (and then inline `existingScan.code` to `requireNotNull`).

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch, line 233: If I understand correctly, the `run` call is only necessary for the case that `existingScan` is *null*. Maybe this could be better expressed by changing the logic to something like:

```
existingScan?.let {
  log.info { "Scan found for url=$url and revision=$revision." }

  val scanCode = existingScan.code
  ...
} ?: kotlin.run {
  log.info { "No scan found for url=$url and revision=$revision. Creating scan ..." }
  val scanCode = createScan(projectCode, url, revision)
  ...
}

```

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: No it unfortunately does not work. Because the check is done in the `checkResponse` function, Kotlin cannot infer non-nullability from outside.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Is this really more readable as a plain if..else ? :)

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth: In other code areas you expressed this via `data!!`, so I would have assumed that to be used here, too.

But a more speaking / semantically better way to express this is via `checkNotNull`: This function documents that the argument is guaranteed to be non-null. That's a different thing from `requireNotNull` which acknowledges that the value might be non-null, but a non-null value is required.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Granted, this may be a matter of taste. My point was more that `run` should only be called if necessary. I would also be fine with moving this call inside the `if`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Ok makes sense.
I try to prevent using `!!` as this is a code smell.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Yes, I did it regarding the other review.

oheger-bosch (dismissed):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,32 @@ class FossId(
...
+                    val scanCode = createScan(projectCode, url, revision)
 
-                service.downloadFromGit(user, apiKey, scanCode)
-                    .checkResponse("download data from Git", false)
+                    log.info { "Initiating data download ..." }
+                    val responseDownload = service.downloadFromGit(user, apiKey, scanCode)
+                    responseDownload.checkResponse("download data from Git", false)
+                    scanCode
+                } else {
+                    log.info { "Scan found for url=$url and revision=$revision." }
```
sschuberth, line 242: Nit: My comment about above's log statement also applies here actually.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,32 @@ class FossId(
...
 
-                checkScan(scanCode)
+                    val scanCode = existingScan.code
+                    requireNotNull(scanCode) {
```
sschuberth, line 245: You forgot the inlining here, I guess.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -221,17 +222,32 @@ class FossId(
...
-                log.info { "Initiating download from Git repository ..." }
+                val scanCode = if (existingScan == null) {
+                    log.info { "No scan found for $url and revision $revision. Creating scan ..." }
+                    val scanCode = createScan(projectCode, url, revision)
 
-                service.downloadFromGit(user, apiKey, scanCode)
-                    .checkResponse("download data from Git", false)
+                    log.info { "Initiating data download ..." }
+                    val responseDownload = service.downloadFromGit(user, apiKey, scanCode)
+                    responseDownload.checkResponse("download data from Git", false)
```
sschuberth, line 239: As `responseDownload` is used nowhere else, maybe just call `checkResponse()` directly on `downloadFromGit()`?

sschuberth (approved):

Issue #3934, opened 2021-04-26
GoMod considers the entire graph of dependencies without checking if the code uses them


vbasem:
currently the logic relies on go mod graph to extract the dependency tree.
For example if you are using [viper](https://github.com/spf13/viper) as a dependency ORT will report the [crypt](https://github.com/bketelsen/crypt) package as a dependency and an input for scan and download. 

This is only true, if the code uses the remote provider in viper and not the local implementation. To avoid this issue, we could use `go mod why <package>` to verify whether a dependency from the graph is actually used within the application or not, since by default the graph doesnt consider this.

Suggestion:

After building the dependency graph , verify each element through ```go mod why``` as explained [here](https://golang.org/ref/mod#go-mod-why)

<sub> Basem Vaseghi <basem.vaseghi@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

sschuberth:
> since by default the graph doesnt consider this.

Doh, that's quite a bummer. Thanks for pointing this out.

So taking `go mod why` into account should also help to address #3860.

vbasem:
yes the pseudo code would be:
1. get the dependency tree using ```go mod graph``` which should print the module requirements
2. run each entry through ```go mod why``` to verify whether that dependency is part of the compilation or an unused dependency

sschuberth:
> 2. run each entry through `go mod why` to verify whether that dependency is part of the compilation or an unused dependency

That bit unfortunately sounds very involving when you have hundreds or thousands of dependencies in the graph. But at least `go mod why` seems to accept multiple packages, so we can probably do one huge bulk-request, or at least go through the graph's nodes in chunks.

vbasem:
thats true, luckily even in the kubernetes project we dont come anywhere near that many dependencies. Most larger projects converge anywhere between 100-500 deps. If the we know of an optimized way, the better!

tsteenbe:
[...]

Issue #3860, opened 2021-04-09
Analyzer creates extremely large result for GoMod project and creates issues during scanning phase


nate-droid:
I am currently running an Analyze on this [repo](https://github.com/valocode/bubbly/). In addition to taking over 10 minutes, the result is approx 10GB in size. The entire project is about 300mb, so this seems a bit too large. 

When running Scan on the results, it will take between 30-40 min before it will eventually result in `java.lang.OutOfMemoryError: Java heap space`. This process has been confirmed on several machines.

`ort analyze -i ../path-my-project -o ../path-to-output -m GoMod`
`ort scan -i ../output/analyzer-result.yml -o ../output --skip-excluded`

sschuberth:
Probably one cause is that we haven't ported the Go-analyzers to the new dependency graph format yet, also see https://github.com/oss-review-toolkit/ort/issues/3825.

sschuberth:
As a side note, please be aware that Go*Mod* support is still marked as *experimental* "as only proxy-based source artifacts but no VCS locations are supported".

sschuberth:
While #4119 is already merged, #4249 still needs to be implemented to fully resolve this issue.

fviernau:
I've recently did a larger re-work of ORT's GoMod support which affected the reported dependency tree. So, I re-tested this bug with:
```
./cli/build/install/ort/bin/ort --info analyze -i ~/sources/github/valocode/bubbly -o . -m gomod
``` 
This took less than two minutes to execute and produced a result of size 0.16MB without any analyzer issues.
The analyzer examined revision aca7f8deb2e02c62583a5a0e3599f5c1cb096a3b.

So, the issue is fixed and this ticket can be closed as fixed.

PR #4119, opened 2021-06-04
GoMod: Remove unused packages from the dependency graph

oheger-bosch:
The current implementation makes use of the "go mod graph" command to
obtain all the dependencies of a project. This command, however,
yields a full dependency graph that might contain packages, which are
not directly used by the main module.

To fix this, run the "go mod why" command on the found packages and
filter out those packages that are reported to be unused.

This is a fix for #3934.


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -235,54 +227,57 @@ class GoMod(
...
-        val id: Identifier,
-        val outgoingEdges: MutableSet<Identifier> = mutableSetOf(),
-        val incomingEdges: MutableSet<Identifier> = mutableSetOf()
-    )
+/**
+ * A type alias for a representation of the dependency graph used by this class. This is basically an adjacency list
+ * implemented by a map whose keys are package identifiers and whose values are the identifiers of packages these
+ * packages depend on.
+ */
+private typealias Graph = MutableMap<Identifier, Set<Identifier>>
```
fviernau, line 259: Instead of defining a `typealias`, I believe it is worth to define a `class`.
This allows to define all the below extension functions simply as member functions,
provides better encapsulation and is simpler to use.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -92,35 +92,27 @@ class GoMod(
         val projectDir = definitionFile.parentFile
 
         stashDirectories(projectDir.resolve("vendor")).use {
-            val edges = getDependencyGraph(projectDir)
+            val graph = getDependencyGraph(projectDir)
```
fviernau: commit-message: Was there an issue with the performance of the code? 
If there wasn't I believe the code should rather be optimized in terms of complexity / readability.
Please mentioned that in the commit message.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -235,54 +227,57 @@ class GoMod(
...
+    filter { it.key in nodes }
+        .mapValuesTo(mutableMapOf()) { e -> e.value.filterTo(mutableSetOf()) { it in nodes } }
 
-        if (source != null && target != null) {
-            source.outgoingEdges += target.id
-            target.incomingEdges += source.id
-        }
-    }
+/**
+ * Convenience function to query the identifiers of the packages the package with the passed in [id] depends on.
```
fviernau, line 279: Maybe shorter: "Return the identifiers of the direct dependencies of the package denoted by [id]."

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -235,54 +227,57 @@ class GoMod(
...
+}
 
-    forEach { edge ->
-        val source = addNode(edge.source)
-        val target = addNode(edge.target)
+/**
+ * Return a new [Graph] that is restricted to the given [nodes]. This can be used to construct graphs for specific
+ * scopes.
+ */
+private fun Graph.retainNodes(nodes: Set<Identifier>): Graph =
```
fviernau, line 274: I believe this is commonly called `subgraph`. So how about naming the function `subgraph()` or `getSubgraph()` ?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -41,6 +41,7 @@ import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.utils.CommandLineTool
 import org.ossreviewtoolkit.utils.Os
+import org.ossreviewtoolkit.utils.log
```
fviernau, line 44: commit message: It says "which are
not directly used by the main module". I believe using the term "directly" is unclear. What do you mean by that?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -283,6 +307,15 @@ private fun Graph.toPackageReferenceForest(root: Identifier): SortedSet<PackageR
 // See https://golang.org/ref/mod#pseudo-versions.
 private val PSEUDO_VERSION_REGEX = "^v0.0.0-(?:[\\d]{14}-(?<sha1>[0-9a-f]+)$)".toRegex()
 
+/** Separator string indicating that data of a new package follows in the output of the go mod why command. */
+private const val PACKAGE_SEPARATOR = "# "
+
+/**
```
fviernau, line 313: Why is chunking necesssary?  e.g. what problem does it avoid?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -178,6 +179,29 @@ class GoMod(
         return result
     }
 
+    /**
```
fviernau, line 182: Instead of the functions `filterUsedPackages()` and `parseWhyOutput()`, I'd prefer
`getUsedPackages(): Set<Identifier>`. This IMO better aligns with Single responsibiltiy principle and minimized the dependencies of the functions, e.g. the function which determines the used packages does not involve a `Graph` and thus is easier to test / can be tested independently of a Graph.

The implementation then could simply do `graph.subgraph(getUsedPackages())`.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -297,3 +330,25 @@ internal fun Identifier.toVcsInfo(): VcsInfo {
...
+/**
+ * Parse the given [lines] generated by the _go mod why_ command and the names of packages that are reported to be
+ * used by the main module.
+ * See https://golang.org/ref/mod#go-mod-why.
+ */
+internal fun parseWhyOutput(lines: Sequence<String>): Set<String> {
+    val usedPackages = mutableSetOf<String>()
+    var currentPackage: String? = null
+
+    lines.forEach { line ->
```
fviernau, line 343: The output for a single package is structured as follow

1. First line must be of the form: "# $ID"
2. Followed by 0-N lines which don't start with '(' of the form "$ID"
3. Followed by  a blank line

I believe it would be good to assume that format and make strict checks for it. What do you think?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -297,3 +330,25 @@ internal fun Identifier.toVcsInfo(): VcsInfo {
     val vcsInfo = VcsHost.toVcsInfo("https://$name")
     return vcsInfo.copy(revision = getRevision(version))
 }
+
+/**
+ * Parse the given [lines] generated by the _go mod why_ command and the names of packages that are reported to be
+ * used by the main module.
+ * See https://golang.org/ref/mod#go-mod-why.
+ */
+internal fun parseWhyOutput(lines: Sequence<String>): Set<String> {
```
fviernau, line 339: How about making the parameter just `String` for simplicity?

analyzer/src/test/kotlin/managers/GoModTest.kt
```diff
@@ -70,4 +73,59 @@ class GoModTest : WordSpec({
             id.toVcsInfo().revision shouldBe "v43.3.0"
         }
     }
+
+    "parseWhyOutput" should {
+        "detect packages that are used" {
+            val output = listOf(
```
fviernau, line 79: If we assume [1] as output format then blank lines should be added.
Same below.

[1] https://golang.org/ref/mod#go-mod-why

fviernau (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -180,12 +173,35 @@ class GoMod(
...
+     * Remove the packages from the given [graph] that are not used by the module by running the _go mod why_ command
+     * repeatedly in [projectDir].
+     */
+    private fun filterUsedPackages(graph: Graph, projectDir: File): Graph {
+        val usedPackageNames = mutableSetOf<String>()
+
+        graph.keys.chunked(WHY_CHUNK_SIZE).forEach { ids ->
+            val pkgNames = ids.map { it.name }.toTypedArray()
+            usedPackageNames +=
+                parseWhyOutput(run(projectDir, "mod", "why", *pkgNames).requireSuccess().stdout.lineSequence())
```
fviernau, line 192: https://golang.org/ref/mod#go-mod-why says that "The -vendor flag causes go mod why to ignore imports in tests of packages outside the main module (as go mod vendor does).". So, would it make sense to add the `-vendor´ flag?

analyzer/src/test/kotlin/managers/GoModTest.kt
```diff
@@ -70,4 +73,59 @@ class GoModTest : WordSpec({
...
+
+        "handle a missing current package" {
+            val output = listOf("foo").asSequence()
+
+            parseWhyOutput(output) should beEmpty()
+        }
+
+        "skip unused packages" {
+            val output = listOf(
+                "# cloud.google.com/go",
```
fviernau, line 103: My proposal to use `String` instead of `Sequence` would lead to multi line strings in this test which is IMO a bit nicer to read and write.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: Makes sense. However, the `parseWhyOutput()` function has been exposed, so that it can be tested separately, also including corner cases of the output of the `why` command. With the two functions being merged, this would not be possible any more in an easy way. Any suggestions how to achieve this?

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: You mean the call to `lineSequence()` should be done in the function?

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Sorry, I wrote it wrongly. I didn't mean to ask to ditch `parseWhyOutput()`. Having that for the tests makes sense to me. Replacing `filterUsedPackages()` with a combination of `getUsedPackages` and `graph.subgraph(getUsedPackages())` was more what I was after.

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Yes.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: The current approach is indeed pretty tolerant against quirks in the output (although I do not necessarily see this as a disadvantage). In order to apply stricter checks, we would need to carry more state during the iteration over the output lines. There are multiple things that could be wrong, for instance
* multiple blank lines in a series
* blank lines before a line that does not introduce a new package ID
* Multiple lines starting with '#' following each other
* A line starting with '(' not directly after a line with '#'

Should we check all of those?

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: I decided against this because in the following it is stated that from go 1.16 or higher that flag does not make any difference. OTOH, it would probably do no harm to add it.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: I hoped to make this clear in the comment: to prevent an unrestricted, large output of a single command when there is a big number of dependencies. But I can try to make the comment more explicit.

fviernau: Would a very conservative upper bound be let's say 300 dependencies, with 300 lines per dependency, with 70 bytes per line, I'd arrive at a bit less than 7MB ? May this is really quite opinionated, so I'll leave it up to you. I don't have a strong opinion on that one.

pro:
- constant memory consumption
contra:
- a bit more complex
- a bit worse execution time

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: I tried to write in code what I had in mind, but it turned out to be more complex than I expected.
I'm good with leaving it as is.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -302,3 +330,25 @@ internal fun Identifier.toVcsInfo(): VcsInfo {
...
+ * used by the main module.
+ * See https://golang.org/ref/mod#go-mod-why.
+ */
+internal fun parseWhyOutput(lines: Sequence<String>): Set<String> {
+    val usedPackages = mutableSetOf<String>()
+    var currentPackage: String? = null
+
+    lines.forEach { line ->
+        if (line.startsWith(PACKAGE_SEPARATOR)) {
+            currentPackage = line.substring(PACKAGE_SEPARATOR.length)
```
fviernau, line 345: consider using `.removePrefix`

sschuberth: ... or `.withoutPrefix`.

analyzer/src/main/kotlin/managers/GoMod.kt
sschuberth: We *do* have some reports about our Go analysis being slow, see e.g. https://github.com/oss-review-toolkit/ort/issues/3860. That also matches my personal experience. I'd say, any performance improvement is welcome in that regard.

sschuberth: On top of that it's even slightly less code than before, so 👍🏻 from my side.

analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
```diff
@@ -20,42 +20,13 @@ project:
     dependencies:
     - id: "GoMod::github.com/fatih/color:v1.7.0"
       linkage: "PROJECT_STATIC"
-    - id: "GoMod::github.com/hashicorp/go-multierror:v1.0.0"
```
sschuberth, line 23: Commit message nit: I believe there should be no comma in "contain packages, which are".

sschuberth: Out of curiosity: Can you explain to me what "not directly used by the main module" means in more detail? I mean, probably no transitive dependency is directory used (also) by the main module, but still some transitive dependencies remain after your change (which is good). Is it some sort of "dead code elimination" that's going on here, so if there is no code path that really makes use of a transitive dependency, it gets removed?

analyzer/src/main/kotlin/managers/GoMod.kt
sschuberth: Ah, looks like I had a similar question [above](https://github.com/oss-review-toolkit/ort/pull/4119/commits/75e835073286deeed6e3d17c843833428c493aed#r645843909).

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: `Graph` is now a class.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: While I do not have any benchmarks, I noticed that the original code was iterating over the list of edges multiple times. As the number of edges is typically greater than the number of nodes, the new representation should be more efficient, especially as packages and their dependencies can now be looked up directly in the map.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: I would argue that the `chunked()` function from the standard library makes it quite easy to handle these chunks, so the additional complexity is low. I always feel a bit uneasy if no size restrictions are applied. And there is #3860, which mentions analyzer results for go projects with a size of up to 10 GB.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: Yes, I also think changing this would be rather complex and would not have that much benefit.

analyzer/src/test/kotlin/managers/GoModTest.kt
oheger-bosch: The current implementation is actually a bit more robust regarding empty lines. So the tests cover both cases, existing and missing empty lines.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: I went for `removePrefix()`, although I think the implementation with `substring()` is probably slightly more efficient, as the `startsWith()` check has to be done anyway.

sschuberth: That was exactly the movitation to introduce `withoutPrefix()` which allows you to do:

    lines.forEach { line ->
        line.withoutPrefix(PACKAGE_SEPARATOR)?.let {
            currentPackage = it
        } ?: run {
            if (!line.startsWith('(') && line.isNotBlank()) {
                currentPackage?.let { usedPackages += it }
            }
        }
    }

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: This is a good question. Unfortunately, my knowledge of go modules is quite limited, and I was mainly following the advice given in #3934. As I understand it from the description in https://golang.org/ref/mod#modules-overview, a module can consist of multiple packages. The dependency graph per default seems to include all the packages of a module. `go mod why` obviously inspects the `import` statements starting from the main module to determine whether specific packages are actually referenced.

I slightly reworded the commit message to align to the terminology used by `go mod why`. The command prints "main module does not need package XXX" for the packages that are now filtered out.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: Indeed stylish. But for the problem at hand, I find the current code more readable; it better expresses that the line is checked for different patterns which trigger specific actions.

analyzer/src/test/kotlin/managers/GoModTest.kt
fviernau: Maybe it's a nit: I still believe the input format according to above link does contain blank lines, so this test is now testing implementation details it shouldn't be testing.s Even worse, that implementation detail is tested multiple times redundantly.
As consequence, a valid refactoring may break test which should not be the case.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -235,54 +227,68 @@ class GoMod(
...
-        if (source != null && target != null) {
-            source.outgoingEdges += target.id
-            target.incomingEdges += source.id
-        }
+/**
+ * A class to represent a graph with dependencies. This representation is basically an adjacency list implemented by a
+ * map whose keys are package identifiers and whose values are the identifiers of packages these packages depend on.
+ */
+private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>) {
+    constructor() : this(mutableMapOf())
```
sschuberth, line 258: How about moving this to a default parameter instead?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -308,3 +340,25 @@ internal fun Identifier.toVcsInfo(): VcsInfo {
...
+ * used by the main module.
+ * See https://golang.org/ref/mod#go-mod-why.
+ */
+internal fun parseWhyOutput(output: String): Set<String> {
+    val usedPackages = mutableSetOf<String>()
+    var currentPackage: String? = null
+
+    output.lineSequence().forEach { line ->
+        if (line.startsWith(PACKAGE_SEPARATOR)) {
+            currentPackage = line.removePrefix(PACKAGE_SEPARATOR)
```
sschuberth, line 355: I have to admit, personally I liked the original code better then, as it was not less readable IMO, but avoided the duplicate `startsWith()`. But I'll leave it to you guys, I'm just throwing in my comments 😉

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -235,54 +227,68 @@ class GoMod(
...
-
-        if (source != null && target != null) {
-            source.outgoingEdges += target.id
-            target.incomingEdges += source.id
-        }
+/**
+ * A class to represent a graph with dependencies. This representation is basically an adjacency list implemented by a
+ * map whose keys are package identifiers and whose values are the identifiers of packages these packages depend on.
+ */
+private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>) {
```
fviernau, line 257: Why is `nodeMap` a constructore parameter? Can that parameter be removed?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: It is used by `subgraph()`.

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: Switched back to the implementation using `substring()`; I liked it better, too.

sschuberth:
LGTM now, but I'll leave the approval to @fviernau who did a more thorough review than I.

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Missed that, thanks. I believe the following would be still good to do - I'll leave it up to you as I also think it's already good enough.

1. Use a public noarg constructor and make the constructor which accepts the map private
2. Add an pre-condition check to that private constructor which verifies that all identifiers of all sets are contained as key in the given map.

fviernau (approved):

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -231,9 +231,7 @@ class GoMod(
  * A class to represent a graph with dependencies. This representation is basically an adjacency list implemented by a
  * map whose keys are package identifiers and whose values are the identifiers of packages these packages depend on.
  */
-private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>) {
-    constructor() : this(mutableMapOf())
-
+private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>> = mutableMapOf()) {
```
fviernau, line 257: Maybe forgot to squash this into the preceeding change?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
oheger-bosch: Oh yes. Thanks for spotting this.

oheger-bosch:
@fviernau regarding the changes proposed for the constructors of the `Graph` class: If it was a public, exposed class, I would directly go for it; but as it is used only as an internal data structure, I do not think that additional checks are necessary. The logic to construct and update the graph should ensure that the map passed to the constructor is consistent.

fviernau (approved):

PR #4129, opened 2021-06-08
model: Introduce and use a generic findFilesWithKnownExtensions() function

sschuberth:

### DIFF
--- a/model/src/main/kotlin/FileFormat.kt
+++ b/model/src/main/kotlin/FileFormat.kt
@@ -64,6 +64,14 @@ enum class FileFormat(val mapper: ObjectMapper, val fileExtension: String, varar
          * Return the [FileFormat] for the given [file], or `null` if there is none.
          */
         fun forFile(file: File): FileFormat = forExtension(file.extension)
+
+        /**
+         * Return a list of all files inside [directory] with known extensions that can be deserialized.
+         */
+        fun findFilesWithKnownExtensions(directory: File): List<File> {
+            val allFileExtensions = enumValues<FileFormat>().flatMap { it.fileExtensions }
+            return directory.walkBottomUp().filter { it.isFile && it.extension in allFileExtensions }.toList()
+        }
     }
 
     /**
--- a/helper-cli/src/main/kotlin/commands/packageconfig/FindCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/packageconfig/FindCommand.kt
@@ -25,10 +25,10 @@ import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.file
 
+import org.ossreviewtoolkit.model.FileFormat
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.config.PackageConfiguration
 import org.ossreviewtoolkit.model.readValue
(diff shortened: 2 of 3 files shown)


model/src/main/kotlin/FileFormat.kt
```diff
@@ -64,6 +64,16 @@ enum class FileFormat(val mapper: ObjectMapper, val fileExtension: String, varar
...
          */
         fun forFile(file: File): FileFormat = forExtension(file.extension)
+
+        /**
+         * Return a list of all files inside [directory] with known extensions that can be deserialized.
+         */
+        fun findDeserializableFiles(directory: File): List<File> {
+            val allFileExtensions = enumValues<FileFormat>().flatMap { it.fileExtensions }
+            return directory.walkBottomUp().filter {
+                it.isFile && it.extension in allFileExtensions && !it.isHidden
```
sschuberth, line 74: @fviernau do you guys actually make use of the hidden flag to "disable" files, or could we drop `!it.isHidden` for simplicity?

fviernau: We are not using the "hidden" flag to "disable" package configurations.

model/src/main/kotlin/FileFormat.kt
```diff
@@ -64,6 +64,16 @@ enum class FileFormat(val mapper: ObjectMapper, val fileExtension: String, varar
          * Return the [FileFormat] for the given [file], or `null` if there is none.
          */
         fun forFile(file: File): FileFormat = forExtension(file.extension)
+
```
fviernau, line 67: commit: What's the reason for extracting this function?

fviernau (changes requested):

model/src/main/kotlin/FileFormat.kt
```diff
@@ -64,6 +64,16 @@ enum class FileFormat(val mapper: ObjectMapper, val fileExtension: String, varar
...
+        /**
+         * Return a list of all files inside [directory] with known extensions that can be deserialized.
+         */
+        fun findDeserializableFiles(directory: File): List<File> {
```
fviernau, line 71: Would it make sense to name this e.g. `findFilesWithKnownExtension`
to not communicate that the returned files can actually be deserialized?

model/src/main/kotlin/FileFormat.kt
fviernau: I'm not certain how we do it for `ort.yml` and other config files.
However, would it make sense to limit the support to just `yml`?

sschuberth: We're planning to make reuse of it in an upcoming PR for other YAML / JSON files in a directory tree.

model/src/main/kotlin/FileFormat.kt
sschuberth: Makes sense, will do.

model/src/main/kotlin/FileFormat.kt
sschuberth: > However, would it make sense to limit the support to just `yml`?

I don't think so. While unusual and not really documented, we actually would support `json` files anywhere where we usually use `yml` files, so I'd like to keep the function generic.

MarcelBochtler (approved):

fviernau (dismissed):

fviernau (approved):

MarcelBochtler (approved):

Issue #4069, opened 2021-05-25
Improve handling of dependency graphs


mnonnenmacher:
Currently some package managers store their dependencies as dependency tree, and some use the dependency graph. The goal is that all package managers should use the dependency graph representation.

As converting a package manager to natively create the new format takes some effort, as a first step convert the dependency trees after running the package managers and later convert them one by one to support the new graph representation.

Also, currently it is required to call the `AnalyzerResult.withResolvedScopes()` function to make sure that the `scope` properties of the `Project`s are properly initialized. This can lead to wrong results if the user forgets to call that function. To eliminate the need for calling that function, identify which code is currently using the `Project.scope` property and provide an alternative API with the same functionality but which operates on the dependency graphs. The benefit is that converting the graphs to trees to initialize the scopes uses more memory because the same data exist in two separate representations, providing an API on top of the graphs eliminates that issue.

The implementation should be backward compatible so that old ORT result files that use the dependency tree representation can still be deserialized.

oheger-bosch:
I did a search for code using the `Project.scopes` property. Find the full list of references with short remarks below.

While some of these references are pretty easy to convert, there is the general problem that the API of dealing with the dependencies in an `OrtResult` is often tight to a concrete implementation - namely the data classes `Project`, `Scope`, and `PackageReference`. At least the latter two are not used by the dependency graph representation. So I think in order to provide an alternative API for these clients, there needs to be some kind of abstraction or a different approach (e.g. operate on scope names rather than `Scope` objects; provide functions that return the identifiers of the dependencies of a specific package rather than dealing with a `PackageReference` etc.)

To make this new API more explicit, do you think it is worth to introduce a dedicated interface for this purpose? Maybe something like `DependencyHandler`?

Here are the current usages of the `Project.scopes` property:

- Project
  - collectDependencies(): Set<Identifier>
  - collectIssues(): Map<Identifier, Set<OrtIssue>>
  - collectSubProjects(): Set<Identifier>
  - findReferences(): List<PackageReference>
- OrtResult
  - packages: Map<Identifier, PackageEntry>
- RuleSet
  - execution of dependency rules, which get initialized with a package, a package reference, the list of ancestors
- GenerateScopesExcludesCommand
  - filters over scope names
- RemoveEntriesCommand
  - collects all scope names
- ProjectAnalyzerResult
  - collectIssues(): Map<Identifier, OrtIssue>
  - collectPackagesByScope(): List<Package>
[...]

PR #4130, opened 2021-06-08
Write the dependency graph format for all package managers

oheger-bosch:
This PR is part of #4069. It adds a generic converter for analyzer results that generates dependency graphs if necessary. It also adds a custom serializer that triggers this conversion automatically before writing the analyzer result.

Adding the custom serializer was a bit problematic because the conversion logic is part of the `analyzer` module; so the serializer cannot be placed in `model` where the standard object mappers are created. I therefore introduced a mechanism via Java's `ServiceLoader` that allows customization of these mappers from other modules. If you have better suggestions, please let me know.

Note that there are still some quirks until #4069 is completely solved. Currently, there are format conversions when loading and saving ORT results in single steps. (These conversions may change the exact dependency graph representation, if projects are processed in a different order.) This will be solved when the dependency graph becomes the "native" format.

### DIFF
--- a/analyzer/src/funTest/kotlin/GitRepoFunTest.kt
+++ b/analyzer/src/funTest/kotlin/GitRepoFunTest.kt
@@ -37,6 +37,7 @@ import org.ossreviewtoolkit.utils.Ci
 import org.ossreviewtoolkit.utils.ORT_NAME
 import org.ossreviewtoolkit.utils.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.test.DEFAULT_ANALYZER_CONFIGURATION
+import org.ossreviewtoolkit.utils.test.convertToDependencyGraph
 import org.ossreviewtoolkit.utils.test.patchActualResult
 import org.ossreviewtoolkit.utils.test.patchExpectedResult
 
@@ -67,10 +68,12 @@ class GitRepoFunTest : StringSpec() {
         "Analyzer correctly reports VcsInfo for git-repo projects".config(enabled = !Ci.isAzureWindows) {
             val ortResult = Analyzer(DEFAULT_ANALYZER_CONFIGURATION).analyze(outputDir)
             val actualResult = ortResult.withResolvedScopes().toYaml()
-            val expectedResult = patchExpectedResult(
-                File("src/funTest/assets/projects/external/git-repo-expected-output.yml"),
-                revision = REPO_REV,
-                path = outputDir.invariantSeparatorsPath
+            val expectedResult = convertToDependencyGraph(
+                patchExpectedResult(
+                    File("src/funTest/assets/projects/external/git-repo-expected-output.yml"),
+                    revision = REPO_REV,
+                    path = outputDir.invariantSeparatorsPath
+                )
             )
 
             patchActualResult(actualResult, patchStartAndEndTime = true) shouldBe expectedResult
--- a/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
+++ b/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
@@ -119,7 +119,7 @@ class AnalyzerResultBuilderTest : WordSpec() {
                 val serializedMergedResults = yamlMapper.writeValueAsString(mergedResults)
                 val deserializedMergedResults = yamlMapper.readValue<AnalyzerResult>(serializedMergedResults)
 
-                deserializedMergedResults shouldBe mergedResults
+                deserializedMergedResults.withScopesResolved() shouldBe mergedResults
             }
 
             "be serialized and deserialized correctly with a dependency graph" {
@@ -157,6 +157,20 @@ class AnalyzerResultBuilderTest : WordSpec() {
                 val serializedResult2 = yamlMapper.writeValueAsString(deserializedResult)
 
                 serializedResult2 shouldBe serializedResult
+            }
+
+            "use the dependency graph representation on serialization" {
+                val mergedResults = AnalyzerResultBuilder()
+                    .addResult(analyzerResult1)
+                    .addResult(analyzerResult2)
+                    .build()
+
+                val serializedMergedResults = yamlMapper.writeValueAsString(mergedResults)
+                val resultTree = yamlMapper.readTree(serializedMergedResults)
+
+                resultTree["dependency_graphs"] shouldNotBeNull {
+                    count() shouldBe 2
+                }
             }
         }
 
--- a/test-utils/src/main/kotlin/Utils.kt
+++ b/test-utils/src/main/kotlin/Utils.kt
@@ -93,4 +93,20 @@ fun patchActualResult(
 ): String =
     patchActualResult(yamlMapper.writeValueAsString(result), patchStartAndEndTime)
 
-fun readOrtResult(file: String) = File(file).let { it.mapper().readValue<OrtResult>(patchExpectedResult(it)) }
+fun patchActualResultObject(result: OrtResult, patchStartAndEndTime: Boolean = false): OrtResult =
+    yamlMapper.readValue(patchActualResult(result, patchStartAndEndTime))
+
+fun readOrtResult(file: String) = readOrtResult(File(file))
+
+fun readOrtResult(file: File) = file.mapper().readValue<OrtResult>(patchExpectedResult(file))
+
+/**
+ * Deserialize an [OrtResult] from the given [string representation][result] and then serialize it again. As for some
+ * of the components in the result custom serializers or deserializers are registered, this can cause some
+ * modifications and conversions on the original result. This function is also used to convert expected test results
+ * stored in the assets to the most recent serialization format.
+ */
+fun convertToDependencyGraph(result: String): String {
+    val ortResult = yamlMapper.readValue<OrtResult>(result)
+    return yamlMapper.writeValueAsString(ortResult)
+}
--- a/analyzer/src/funTest/kotlin/managers/SbtFunTest.kt
+++ b/analyzer/src/funTest/kotlin/managers/SbtFunTest.kt
@@ -19,6 +19,8 @@
 
 package org.ossreviewtoolkit.analyzer.managers
 
+import com.fasterxml.jackson.module.kotlin.readValue
+
 import io.kotest.core.spec.style.StringSpec
 import io.kotest.matchers.shouldBe
 
@@ -27,11 +29,14 @@ import java.io.File
 import org.ossreviewtoolkit.analyzer.Analyzer
 import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.downloader.vcs.Git
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.yamlMapper
 import org.ossreviewtoolkit.utils.Ci
 import org.ossreviewtoolkit.utils.normalizeVcsUrl
 import org.ossreviewtoolkit.utils.test.DEFAULT_ANALYZER_CONFIGURATION
-import org.ossreviewtoolkit.utils.test.patchActualResult
+import org.ossreviewtoolkit.utils.test.patchActualResultObject
 import org.ossreviewtoolkit.utils.test.patchExpectedResult
+import org.ossreviewtoolkit.utils.test.readOrtResult
 
 class SbtFunTest : StringSpec({
     "Dependencies of the external 'directories' single project should be detected correctly".config(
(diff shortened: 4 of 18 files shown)


analyzer/src/test/kotlin/managers/utils/DependencyGraphConverterTest.kt
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            convertedResult.withScopesResolved() shouldBe result
+            convertedResult.packages should beTheSameInstanceAs(result.packages)
+        }
+
+        "convert a result with a partial dependency graph" {
+            val gradleProject = createProject("Gradle", 1)
+            val goProject1 = createProject("GoMod", 2)
+            val goProject2 = createProject("GoMod", 3)
```
mnonnenmacher, line 86: Please consider using named arguments to let the numbers appear less magic.

analyzer/src/test/kotlin/managers/utils/DependencyGraphConverterTest.kt
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+})
+
+/**
+ * Create a test project with a (small) dependency tree. Use [managerName] and [index] to generate identifiers.
+ */
+private fun createProject(managerName: String, index: Int): Project {
+    val mainScope = Scope("main", createDependencies(managerName, index * 100, 4))
+    val testScope = Scope("test", createDependencies(managerName, index * 100 + 10, 8))
```
mnonnenmacher, line 112: Please consider using named arguments to let the numbers appear less magic.

analyzer/src/main/kotlin/managers/utils/DependencyGraphConverter.kt
```diff
@@ -0,0 +1,115 @@
...
+ * The main conversion function takes an [AnalyzerResult] and iterates over all projects contained in it. Each
+ * project that still uses the dependency tree representation (i.e. the set of scopes) is converted, and a dependency
+ * graph for the associated package manager is created.
+ *
+ * Via this converter, optimized results can be generated for package managers that do not yet support this format.
+ * This approach is, however, not ideal, because during the conversion both the dependency tree and the dependency
+ * graph need to be hold in memory. So this is rather a temporary solution until all package managers support the new
+ * format natively.
+ */
+object DependencyGraphConverter {
```
mnonnenmacher, line 43: Is there any reason that this class has to be in the analyzer module? I think it could be in `model.utils` as well which would make the reverse dependency you introduce in the next commits unnecessary. This would require moving `DependencyGraphBuilder` and maybe some other classes to the model as well, but I think I would prefer that over the service loader approach. As this is an architectural question I'd like to have another opinion from @sschuberth or @fviernau as well.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/utils/DependencyGraphConverter.kt
```diff
@@ -0,0 +1,115 @@
...
+     */
+    fun convert(result: AnalyzerResult): AnalyzerResult {
+        val projectsToConvert = result.projectsWithScopes()
+        if (projectsToConvert.isEmpty()) return result
+
+        val graphs = buildDependencyGraphs(projectsToConvert)
+
+        return result.copy(
+            dependencyGraphs = result.dependencyGraphs + graphs,
+            projects = result.projects.mapTo(sortedSetOf()) { it.convertToScopeNames() })
```
sschuberth, line 56: Nit: Closing parenthesis should be on new line.

analyzer/src/main/kotlin/managers/utils/DependencyGraphConverter.kt
sschuberth: I'm not a fan of Java's `ServiceLoader` mechanism in general, so whenever we can (reasonably) avoid it, I'm in for it. And simply moving (hopefully only) a few classes to the `model` module to do so sounds reasonable to me.

oheger-bosch: I was not sure about this, so I was hoping for your feedback.

So I take it, the agreement is to move the `DependencyGraphBuilder` and related classes to `model.utils`. Then the custom serializer for `AnalyzerResult` can be configured in the usual way via annotations, and no manipulations on standard mappers are needed.

oheger-bosch: The following classes are now under `model.utils`:
* `DependencyGraphBuilder`
* `DependencyHandler`
* `DependencyGraphConverter`

analyzer/src/main/resources/META-INF/services/org.ossreviewtoolkit.model.MapperCustomizer
```diff
@@ -0,0 +1 @@
+org.ossreviewtoolkit.analyzer.managers.utils.CustomAnalyzerResultSerializerModule
```
sschuberth, line 1: Is this now still needed? I thought we got rid of the `ServiceLoader`-based approach in the new iteration.

analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
```diff
@@ -158,6 +159,18 @@ class AnalyzerResultBuilderTest : WordSpec() {
...
+            "use the dependency graph representation on serialization" {
+                val mergedResults = AnalyzerResultBuilder()
+                    .addResult(analyzerResult1)
+                    .addResult(analyzerResult2)
+                    .build()
+
+                val serializedMergedResults = yamlMapper.writeValueAsString(mergedResults)
+                val deserializedMergedResults = yamlMapper.readValue<AnalyzerResult>(serializedMergedResults)
+
+                deserializedMergedResults.dependencyGraphs.keys should containExactlyInAnyOrder("type-1", "type-2")
```
sschuberth, line 172: According to the test name, shouldn't you directly check on `serializedMergedResults`? Otherwise you're rather testing the *de*serialization, which might apply its own magic.

test-utils/src/main/kotlin/Utils.kt
```diff
@@ -94,4 +94,18 @@ fun patchActualResult(
...
+
+fun readOrtResult(file: File) = file.mapper().readValue<OrtResult>(patchExpectedResult(file))
+
+/**
+ * Convert a result in serialized form by deserializing it and then serializing it again. This causes a conversion to
+ * the most recent serialization format.
+ */
+fun convertResult(orgResult: String): String {
+    val ortResult = yamlMapper.readValue<OrtResult>(orgResult)
+    return yamlMapper.writeValueAsString(ortResult)
```
sschuberth, line 110: Without knowing about the custom serializer, it would be totally unclear to the read why this would convert anything. So I wonder whether we should rename this function so something that doe snot include "convert" at all, like "reserializeOrtResult" or so.

sschuberth: Also "orgResult" and "ortResult" look awfully similar. "org" could also be misinterpreted as "organization". So why not just "result"? Or "oldResult" and "newResult"?

sschuberth (changes requested):

analyzer/src/main/resources/META-INF/services/org.ossreviewtoolkit.model.MapperCustomizer
oheger-bosch: Well spotted, you are right. The file was removed.

analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
oheger-bosch: Well, these are white-box tests; we know that the standard Jackson deserialization is used.

Testing on the serialized form is problematic. It would probably require some string matching. If you prefer, I can try this out.

test-utils/src/main/kotlin/Utils.kt
oheger-bosch: I renamed the function to `applyCustomSerialization()` and gave some more background in the comment.

oheger-bosch:
Rebased again on master, as some conflicts were reported. @mnonnenmacher, @sschuberth, is there anything left here you want me to change?

analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
mnonnenmacher: Just as an idea, another option would be to use `readTree` instead and do the assertion on the node, this would decouple the test from any deserialization logic.

test-utils/src/main/kotlin/Utils.kt
mnonnenmacher: I'm not sure if I understand why this function is required. Couldn't the same be achieved by just fixing the expected result files once to the new format?

oheger-bosch: Yes, this would be possible, but this would have some drawbacks:
* Many files would have to be adapted, as there are a lot of expected result files.
* The new format is less readable.
* When you change both the implementation and the expected results, you cannot really be sure whether the new implementation yields correct results.

Therefore, I would prefer to keep the expectations in the old format, so to say as a reference. The tests then have to do some kind of conversion to check the content of the produced results.

sschuberth: In my view the only argument that really counts somewhat is "The new format is less readable". But how readable do expected test result files actually need to be? I still need to make up my mind about this.

oheger-bosch: With the old format, you can understand the dependency relations between projects and packages quite easily. This is not possible with the new format where you just have a tree of indices. So with reasonable effort, it is hardly possible to derive test expectations from the result file.

analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
oheger-bosch: Good idea! I reworked the test case to read the JSON tree and operate on the nodes.

analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
```diff
@@ -158,6 +158,20 @@ class AnalyzerResultBuilderTest : WordSpec() {
...
+                val mergedResults = AnalyzerResultBuilder()
+                    .addResult(analyzerResult1)
+                    .addResult(analyzerResult2)
+                    .build()
+
+                val serializedMergedResults = yamlMapper.writeValueAsString(mergedResults)
+                val resultTree = yamlMapper.readTree(serializedMergedResults)
+
+                resultTree["dependency_graphs"] shouldNotBeNull {
+                    this.count() shouldBe 2
```
mnonnenmacher, line 172: Is `this` required?

test-utils/src/main/kotlin/Utils.kt
mnonnenmacher: I just realized that my last comment was wrong, in case `applyCustomSerialization` is used the new format would be shown in the diff, right? Anyway, I would be fine with merging this as is and discuss later if we want to change this.

analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
oheger-bosch: No, removed.

mnonnenmacher (approved):

analyzer/src/funTest/kotlin/GitRepoFunTest.kt
```diff
@@ -67,10 +68,12 @@ class GitRepoFunTest : StringSpec() {
         "Analyzer correctly reports VcsInfo for git-repo projects".config(enabled = !Ci.isAzureWindows) {
             val ortResult = Analyzer(DEFAULT_ANALYZER_CONFIGURATION).analyze(outputDir)
             val actualResult = ortResult.withResolvedScopes().toYaml()
-            val expectedResult = patchExpectedResult(
-                File("src/funTest/assets/projects/external/git-repo-expected-output.yml"),
-                revision = REPO_REV,
-                path = outputDir.invariantSeparatorsPath
+            val expectedResult = applyCustomSerialization(
```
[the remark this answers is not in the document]
mnonnenmacher: I have to admit, I'm a bit confused now about which format is used where. Maybe we could have a quick call with @oheger-bosch to clarify that?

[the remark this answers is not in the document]
oheger-bosch: A call is fine with me; I should be available after 10.30 CET today.

Just to point out that the serialization problem is probably temporary. The custom serializer is currently needed because in order to access the dependency information, it has to be converted to the dependency tree format. In #4166 I am working on introducing an API to access this information independently on its representation; then no conversion will be required any more.

test-utils/src/main/kotlin/Utils.kt
```diff
@@ -93,4 +93,20 @@ fun patchActualResult(
...
+
+fun readOrtResult(file: File) = file.mapper().readValue<OrtResult>(patchExpectedResult(file))
+
+/**
+ * Deserialize an [OrtResult] from the given [string representation][result] and then serialize it again. As for some
+ * of the components in the result custom serializers or deserializers are registered, this can cause some
+ * modifications and conversions on the original result. This function is also used to convert expected test results
+ * stored in the assets to the most recent serialization format.
+ */
+fun applyCustomSerialization(result: String): String {
```
sschuberth, line 109: As discussed, `convertToDependencyGraph`.

analyzer/src/funTest/kotlin/GitRepoFunTest.kt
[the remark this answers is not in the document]
oheger-bosch: As discussed in a call:
* In this case, we keep the comparison of strings in the new format, but the `applyCustomSerialization()` function was given a better name.
* In future, all expected result files are going to be converted to the new graph format; conversions are then no longer necessary.
* For the conversion, a helper-cli command will be added.

sschuberth (approved):

PR #3926, opened 2021-04-23
Handle user-facing empty configuration files leniently

sschuberth:
Please have a look at the individual commit messages for the details.

PR #4131, opened 2021-06-08
AnalyzerCommand: Add option to to provide multiple curation files

### LINKED ISSUE #4128: Load curations from a directory structure
To avoid huge `curation.yml` files, ORT should be able to use curations from a directory structure similar to the `--package-configuration-dir` option.

The separate yml-files should be searched for recursively so the user can maintain their wanted directory structure.
As I don't see a technical reason to support both the `--package-curations-file` option and the new `--package-curations-dir` should not be mutually exclusive but merge the curations from both sources.

MarcelBochtler:
Resolves #4128.


analyzer/src/test/assets/package-curations-dIr/curations-1.yml
```diff
@@ -0,0 +1,4 @@
+---
```
sschuberth, line 1: Is the "I" in "package-curation-dIr" intentionally upper-cased?

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -156,9 +166,13 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
         val config = globalOptionsForSubcommands.config
         val analyzer = Analyzer(config.analyzer)
 
+        val curationFiles = mutableListOf<File>()
+        packageCurationsFile.takeIf { it.isFile }?.let { curationFiles.add(it) }
```
sschuberth, line 170: Use `+=` instead of `.add()`.

sschuberth: Same below.

sschuberth (changes requested):

oheger-bosch (commented):
There seems to be a failing test.
Note: The documentation about curation files should be updated (could be done later).

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -28,11 +28,13 @@ import org.ossreviewtoolkit.model.PackageCuration
...
- * A [PackageCurationProvider] that loads [PackageCuration]s from a single file. Supports all file formats specified
- * in [FileFormat].
+ * A [PackageCurationProvider] that loads [PackageCuration]s from all [curationFiles]. Supports all file formats
+ * specified in [FileFormat].
  */
-class FilePackageCurationProvider(curationFile: File) : PackageCurationProvider {
-    internal val packageCurations by lazy { curationFile.readValueOrDefault(emptyList<PackageCuration>()) }
+class FilePackageCurationProvider(curationFiles: List<File>) : PackageCurationProvider {
+    internal val packageCurations by lazy {
+        curationFiles.map { it.readValueOrDefault(emptyList<PackageCuration>()) }.flatten()
```
fviernau, line 36: Please put reading a single file into `try / catch` and log a message when reading fails. 
As-is it's hard to figure out for which of the files the deserialization failed.
Also see the `PackageConfigurationProvider` class.

fviernau (changes requested):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
sschuberth: Speaking about the `(Simple)PackageConfigurationProvider)`, I plan to refactor the class a bit to *also* allow multiple package configurations in a single file when multiple files are spread across a directory tree. And we should probably then do the same here to align the behavior. Any objections @fviernau?

fviernau: > "I plan to refactor the class a bit to also allow multiple package configurations in a single file"

Hm, not sure what exactly you have in mind. I guess that would change the file format of these package configurations, is that correct? What's the reason to do that?

Regarding this commit vs. `(Simple)PackageConfigurationProvider`, I find it quite nice that is has `configurations: Collection<PackageConfiguration>` as opposed to `List<File>`. I believe this PR can consider aligning with that approach.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -28,11 +28,13 @@ import org.ossreviewtoolkit.model.PackageCuration
...
 
 /**
- * A [PackageCurationProvider] that loads [PackageCuration]s from a single file. Supports all file formats specified
- * in [FileFormat].
+ * A [PackageCurationProvider] that loads [PackageCuration]s from all [curationFiles]. Supports all file formats
+ * specified in [FileFormat].
  */
-class FilePackageCurationProvider(curationFile: File) : PackageCurationProvider {
-    internal val packageCurations by lazy { curationFile.readValueOrDefault(emptyList<PackageCuration>()) }
+class FilePackageCurationProvider(curationFiles: List<File>) : PackageCurationProvider {
```
fviernau, line 34: Please use `Collection` instead of `List` as parameter type.

fviernau (changes requested):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -28,11 +28,13 @@ import org.ossreviewtoolkit.model.PackageCuration
 import org.ossreviewtoolkit.model.readValueOrDefault
 
 /**
- * A [PackageCurationProvider] that loads [PackageCuration]s from a single file. Supports all file formats specified
- * in [FileFormat].
+ * A [PackageCurationProvider] that loads [PackageCuration]s from all [curationFiles]. Supports all file formats
```
fviernau, line 31: Take a look at `(Simple)PackageConfigurationProvider`, I find it quite nice that is has `configurations: Collection<PackageConfiguration>` as opposed to `List<File>`. I believe at least considering aligning with that approach would make sense.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
sschuberth: > I guess that would change the file format of these package configurations, is that correct?

No, it would simply be tried to either read `PackageConfiguration` or `List<PackageConfiguration>` from files found by `findFilesWithKnownExtensions`.

> What's the reason to do that?

That way you could have a directory hierarchy that nests package type / namespace / name, but put different `PackageConfiguration`s for different versions into the same file. But the current mode of one `PackageConfiguration` per file in the tree would also still work.

Also, @MarcelBochtler's change here goes into the direction of always reading `List<PackageCuration>` from curation files. I'd find it confusing for the user to remember that, in case of a directory hierarchy, for `PackageConfiguration`s you must put only one per file, but for `PackageCuration` you must put multiple per file. IMO we should align that by allow both formats for both file types.

fviernau: I find one package configuration quite aligned with the `ort.yml` so I'm not with you in the confusing user bit. However, I agree with your argumentation. So I'm fine implementing it as you just outlined.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
MarcelBochtler: While I agree that in general it's better to handle defined objects, for the `FilePackageCurationProvider` I think it's better to pass the actual file.

sschuberth: Actually, I'd agree with @fviernau here. If we go into the direction of an ORT "server" backend and a frontend, then our APIs should abstract away from `File`s where possible. I argumented similarly in e.g. [this change](https://github.com/oss-review-toolkit/ort/pull/3926/commits/6d048c02236d4129584bc893b3fc6b4bf71f2f36).

However, such a refactoring should be done in a separate commit, or maybe even in a preparing separate PR, IMO. Esp. as probably more changes make sense to do then, like renaming "*File*PackageCurationProvider" to "*Simple*PackageCurationProvider".

That said, such a refactoring could also be done *after* merging this PR.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
MarcelBochtler: > Please put reading a single file into try / catch and log a message when reading fails.
As-is it's hard to figure out for which of the files the deserialization failed.
Also see the PackageConfigurationProvider class.

Good point.
I opted to catch `Exception` instead of `IOException` which is done in `PackageConfigurationProvider` to also catch e.g. `IllegalArgumentExceptions` if the file is not valid.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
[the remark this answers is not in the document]
MarcelBochtler: I explicitly opted against `varargs`. But you're right a secondary constructor would increase the convenience.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
fviernau: Agree, out of scope of this one makes sense. Just a small hint: testing is also easier without files.

fviernau (dismissed):

MarcelBochtler:
Added documentation.

MarcelBochtler:
I'll also add this option to the _evaluator_ in an upcoming PR.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -28,11 +29,27 @@ import org.ossreviewtoolkit.model.PackageCuration
...
+ * specified in [FileFormat].
  */
-class FilePackageCurationProvider(curationFile: File) : PackageCurationProvider {
-    internal val packageCurations by lazy { curationFile.readValueOrDefault(emptyList<PackageCuration>()) }
+class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCurationProvider {
+    constructor(curationFile: File) : this(listOf(curationFile))
```
sschuberth, line 36: BTW, an IMO nice pattern to combine both constructors is to use a signature like

    FilePackageCurationProvider(curationFile: File, vararg moreFiles: Collection<File>)

and then let the implementation merge `curationFile` and `moreFiles`. That way you ensure at compile time that at least one file is passed, but more can be passed as well. This is also the approach I'm following in my branch for upcoming `SimplePackageConfigurationProvider` changes.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -28,11 +29,27 @@ import org.ossreviewtoolkit.model.PackageCuration
...
+
+    @Suppress("TooGenericExceptionCaught")
+    internal val packageCurations by lazy {
+        curationFiles.map { curationFile ->
+            try {
+                curationFile.readValueOrDefault(emptyList<PackageCuration>())
+            } catch (e: Exception) {
```
sschuberth, line 43: Why is this not just `IOException` like in `SimplePackageConfigurationProvider`? If it was, we could also just rethrow it (again see `SimplePackageConfigurationProvider`) instead of coming up with our own exception.

MarcelBochtler: See: https://github.com/oss-review-toolkit/ort/pull/4131#discussion_r648073792

sschuberth: Sorry, that conversion was resolved (and thus hidden) too early for me 😀

sschuberth: But sorry, one more question: Do we really stop parsing for all remaining files once an exception occurred? In my upcoming [multi-location-package-config](https://github.com/oss-review-toolkit/ort/tree/multi-location-package-config) branch I'm using [this code](https://github.com/oss-review-toolkit/ort/blob/75f93cdc65c533d63449231b0139a785734dfd17/model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt#L58-L66) to only log warnings about those files, but continue parsing the other files.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
MarcelBochtler: > That way you ensure at compile time that at least one file is passed

But this will take away the possibility to pass empty lists, which is being done in this case when no curation file is given.

My prefered option was to use `vararg`. But I get slapped on the fingers by detekt in the calling code:
`FilePackageCurationProvider(*curationFiles.toTypedArray()),`: https://detekt.github.io/detekt/performance.html#spreadoperator

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
MarcelBochtler: I agree. Changed it to logging the warning instead of throwing an error.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
sschuberth: Then your *detekt*-pre-commit-check is somehow wrong, as we've [disabled that rule](https://github.com/oss-review-toolkit/ort/blob/bd2494938a82843932738780ce27707900976334/.detekt.yml#L41-L43) 😁

sschuberth (approved):

PR #4132, opened 2021-06-08
Move resolvedRevision from VcsInfo to Provenance

mnonnenmacher:
This PR mainly moves the `resolvedRevision` property from `VcsInfo` to `RepositoryProvenance` to better model that this property is only supposed to be set by the downloader. Please see the commit messages for details.

### DIFF
--- a/analyzer/src/funTest/kotlin/integration/DrupalIntegrationFunTest.kt
+++ b/analyzer/src/funTest/kotlin/integration/DrupalIntegrationFunTest.kt
@@ -43,10 +43,9 @@ class DrupalIntegrationFunTest : AbstractIntegrationSpec() {
         binaryArtifact = RemoteArtifact.EMPTY,
         sourceArtifact = RemoteArtifact.EMPTY,
         vcs = VcsInfo(
-            VcsType.GIT,
-            "https://github.com/drupal/drupal.git",
-            "4a765491d80d1bcb11e542ffafccf10aef05b853",
-            ""
+            type = VcsType.GIT,
+            url = "https://github.com/drupal/drupal.git",
+            revision = "4a765491d80d1bcb11e542ffafccf10aef05b853"
         )
     )
 
--- a/analyzer/src/test/kotlin/managers/GoModTest.kt
+++ b/analyzer/src/test/kotlin/managers/GoModTest.kt
@@ -36,10 +36,10 @@ class GoModTest : WordSpec({
             id.toVcsInfo().type shouldBe VcsType.GIT
         }
 
-        "return null as resolved revision" {
+        "return the revision" {
             val id = Identifier("GoMod::github.com/chai2010/gettext-go:v1.0.0")
 
-            id.toVcsInfo().resolvedRevision shouldBe null
+            id.toVcsInfo().revision shouldBe "v1.0.0"
         }
 
         "return the VCS URL and path for a package from a single module repository" {
--- a/model/src/main/kotlin/Provenance.kt
+++ b/model/src/main/kotlin/Provenance.kt
@@ -62,24 +62,22 @@ data class RepositoryProvenance(
     /**
      * The VCS repository that was downloaded.
      */
-    val vcsInfo: VcsInfo
+    val vcsInfo: VcsInfo,
+
+    /**
+     * The VCS revision of the source code that was downloaded, resolved from [vcsInfo] during download. Must not be
+     * blank, and must also be fixed revision, e.g. the SHA1 of a Git commit instead of a branch or tag name.
+     */
+    val resolvedRevision: String
 ) : KnownProvenance() {
-    override fun matches(pkg: Package): Boolean {
-        // If no VCS information is present either, or it does not have a resolved revision, there is no way of
-        // verifying matching provenance.
-        if (vcsInfo.resolvedRevision == null) return false
-
-        // This provenance was definitely created when downloading this package.
-        if (pkg.vcsProcessed.equalsIgnoreResolvedRevision(vcsInfo)) return true
-
-        return listOf(pkg.vcs, pkg.vcsProcessed).any {
-            if (it.resolvedRevision != null) {
-                it.resolvedRevision == vcsInfo.resolvedRevision
-            } else {
-                it.revision == vcsInfo.revision || it.revision == vcsInfo.resolvedRevision
-            } && it.type == vcsInfo.type && it.url == vcsInfo.url && it.path == vcsInfo.path
-        }
+    init {
+        require(resolvedRevision.isNotBlank()) { "The resolved revision must not be blank." }
     }
+
+    /**
+     * Return true if this provenance matches the VCS information of the [package][pkg].
+     */
+    override fun matches(pkg: Package): Boolean = vcsInfo == pkg.vcs || vcsInfo == pkg.vcsProcessed
 }
 
 /**
@@ -95,12 +93,11 @@ private class ProvenanceDeserializer : StdDeserializer<Provenance>(Provenance::c
             }
             node.has("vcs_info") -> {
                 val vcsInfo = jsonMapper.treeToValue<VcsInfo>(node["vcs_info"])!!
-                RepositoryProvenance(vcsInfo)
+                // For backward compatibility, if there is no resolved_revision use the revision from vcsInfo.
+                val resolvedRevision = node["resolved_revision"]?.textValue() ?: vcsInfo.revision
+                RepositoryProvenance(vcsInfo, resolvedRevision)
             }
             else -> UnknownProvenance
         }
     }
 }
-
-private fun VcsInfo.equalsIgnoreResolvedRevision(other: VcsInfo): Boolean =
-    copy(resolvedRevision = "") == other.copy(resolvedRevision = "")
--- a/model/src/main/kotlin/VcsInfo.kt
+++ b/model/src/main/kotlin/VcsInfo.kt
@@ -19,7 +19,6 @@
 
 package org.ossreviewtoolkit.model
 
-import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.core.JsonParser
 import com.fasterxml.jackson.databind.DeserializationContext
 import com.fasterxml.jackson.databind.JsonNode
@@ -54,13 +53,6 @@ data class VcsInfo(
     val revision: String,
 
     /**
-     * The VCS-specific revision resolved during downloading from the VCS. In contrast to [revision] this must not
-     * contain symbolic names like branches or tags.
-     */
-    @JsonInclude(JsonInclude.Include.NON_NULL)
-    val resolvedRevision: String? = null,
-
-    /**
      * The path inside the VCS to take into account, if any. The actual meaning depends on the VCS type. For
      * example, for Git only this subdirectory of the repository should be cloned, or for Git Repo it is
      * interpreted as the path to the manifest file.
@@ -76,7 +68,6 @@ data class VcsInfo(
             type = VcsType.UNKNOWN,
             url = "",
             revision = "",
-            resolvedRevision = null,
             path = ""
         )
     }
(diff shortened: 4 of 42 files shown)


mnonnenmacher:
@sschuberth @fviernau This can already be reviewed, even though I expect that some tests might fail due to the last two commits I added. The tests already worked for the main commit which moves the property. This change is basically the outcome of the discussion we had a while ago, I hope it matches everyone's requirements. The only thing I didn't do is to add a property for symbolic names like branches or tags, but I think this can be done later if required.

sschuberth:
> This change is basically the outcome of the discussion we had a while ago, I hope it matches everyone's requirements. The only thing I didn't do is to add a property for symbolic names like branches or tags, but I think this can be done later if required.

I have to admit I was slightly confused by the PR's title, as I though the outcome of our discussion was rather to change the docs than the implementation of `VcsInfo`. But I'll first have a look before I complain 😉

mnonnenmacher:
Yes, I mainly captured the "spirit" of our discussion, I have tried some approaches and found that this produced the cleanest code.

sschuberth (commented):
Will continue the review tomorrow.

analyzer/src/funTest/kotlin/integration/DrupalIntegrationFunTest.kt
```diff
@@ -43,10 +43,10 @@ class DrupalIntegrationFunTest : AbstractIntegrationSpec() {
         binaryArtifact = RemoteArtifact.EMPTY,
         sourceArtifact = RemoteArtifact.EMPTY,
         vcs = VcsInfo(
-            VcsType.GIT,
-            "https://github.com/drupal/drupal.git",
-            "4a765491d80d1bcb11e542ffafccf10aef05b853",
-            ""
+            type = VcsType.GIT,
```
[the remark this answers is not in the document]
mnonnenmacher: > Would you mind adding that commit's SHA1 as a reference so I remember?

Done.

> Which "requirement"?

This meant the requirement described in the paragraph above, I have rephrased the sentence to make it clearer.

> s/previouly/previously/

Done.

> This means we cannot currently store the (symbolic) revision and its corresponding resolved revision at the same time , e.g. if provided by a package manager.

Yes, just like before. But now there is a way for package managers to show that a revision is resolved which wasn't possible before. I have only done it for NuGet in this PR because it was doing this before and I wasn't sure which other package managers support that, probably GoMod and NPM/Gradle with VCS dependencies.

As mentioned above I think if we want this to be possible it can be done separately. I didn't want to do this now because I mainly wanted to improve the current situation, not add new functionality. Also I don't see that much value in capturing a branch/tag name if we already have a specific revision. I could be used as a fallback if the revision doesn't work or be shown in the reports (with explaining that it could probably have nothing to do with the actually scanned revision, depending on where the data is coming from), but it doesn't seem that any of the package managers I work most with supports two properties for the revision anyway.

> It also means we need to know if a revision is resolved or not. 

No, we don't. If we set the flag to false even though the revision was resolved by the package manager nothing bad happens, except that the downloader might try to guess a revision which is the same behavior as before. On the contrary we now have a way to flag resolved revisions outside the downloader. I already started using this in a separate PR for projects, because if the analyzer runs on a VCS working tree we always know that the revisions of detected projects are resolved.

> The tests already worked for the main commit which moves the property.

Turns out I still have to fix some tests. I wrongly thought that we always execute all test tasks, but it seems that either Gradle stops executing tests tasks if one of them had failing tests, or that we have issues with test reporting on Azure, so everytime I fixed tests a new set of failing tests from another module popped up.

analyzer/src/funTest/kotlin/integration/DrupalIntegrationFunTest.kt
```diff
@@ -43,10 +43,10 @@ class DrupalIntegrationFunTest : AbstractIntegrationSpec() {
...
+            url = "https://github.com/drupal/drupal.git",
+            revision = "4a765491d80d1bcb11e542ffafccf10aef05b853",
+            isResolvedRevision = true
```
sschuberth, line 49: As discussed in the developer meeting, `isVerifiedRevision` or `isConfirmedRevision` would probably fit the semantics better.

mnonnenmacher: I have renamed it to `isVerifiedRevision` and also removed the last commit because you mentioned in the meeting today that NuGet takes the commit SHA1 from metadata so it would be wrong to set the flag there.

mnonnenmacher: As discussed today I have now completely removed the `isVerifiedRevision` property so that we can address this topic later in a more sophisticated way.

analyzer/src/test/kotlin/managers/GoModTest.kt
```diff
@@ -36,10 +36,10 @@ class GoModTest : WordSpec({
             id.toVcsInfo().type shouldBe VcsType.GIT
         }
 
-        "return null as resolved revision" {
+        "return an unresolved revision" {
```
sschuberth, line 39: Just a note: Personally I'm fine with the wording. But given the past discussions about the meaning of "resolved", we might not want to use "unresolved" here. Maybe just "return the revision"?

model/src/main/kotlin/Provenance.kt
```diff
@@ -62,23 +62,26 @@ data class RepositoryProvenance(
...
+     * blank, and must also be fixed revision, e.g. the SHA1 of a Git commit instead of a branch or tag name.
+     */
+    val resolvedRevision: String
 ) : KnownProvenance() {
-    override fun matches(pkg: Package): Boolean {
-        // If no VCS information is present either, or it does not have a resolved revision, there is no way of
-        // verifying matching provenance.
-        if (vcsInfo.resolvedRevision == null) return false
+    init {
+        require(resolvedRevision.isNotBlank()) { "The resolvedRevision must not be blank." }
```
sschuberth, line 74: Nit: IN the free test we should probably say "resolved revision".

model/src/main/kotlin/Provenance.kt
```diff
@@ -62,23 +62,26 @@ data class RepositoryProvenance(
...
+    }
 
-        // This provenance was definitely created when downloading this package.
-        if (pkg.vcsProcessed.equalsIgnoreResolvedRevision(vcsInfo)) return true
+    /**
+     * Return true if this provenance matches the VCS information of the [package][pkg].
+     */
+    override fun matches(pkg: Package): Boolean {
+        // If there is no resolved revision, there is no way of verifying matching provenance.
+        if (resolvedRevision.isBlank()) return false
```
sschuberth, line 82: This can't happen due to the check in `init`.

sschuberth (changes requested):

analyzer/src/funTest/kotlin/integration/DrupalIntegrationFunTest.kt
```diff
@@ -43,10 +43,9 @@ class DrupalIntegrationFunTest : AbstractIntegrationSpec() {
         binaryArtifact = RemoteArtifact.EMPTY,
         sourceArtifact = RemoteArtifact.EMPTY,
         vcs = VcsInfo(
-            VcsType.GIT,
-            "https://github.com/drupal/drupal.git",
-            "4a765491d80d1bcb11e542ffafccf10aef05b853",
-            ""
+            type = VcsType.GIT,
+            url = "https://github.com/drupal/drupal.git",
```
sschuberth, line 47: Commit message:

> The intentation that `resolvedRevision` is only set by the downloader

"The intention to set `resolvedRevision` only by the downloader"

> also the the

"also the"

> , later commits can then address the issue of
> providing more detailed information about the revision in `VcsInfo`.

I propose to make this a separate sentence. Or maybe drop it completely, as we're not sure yet AFAIK which way we should go.

sschuberth (changes requested):

mnonnenmacher:
I have marked this as a breaking change because it partly invalidates stored scan results. In one test with ~500 stored scan results ~400 could still be matched, the other ~100 had to be scanned again. After re-scan of those packages the results can be retrieved from the storage again.

sschuberth:
> I have marked this as a breaking change because it partly invalidates stored scan results.

I had a [similar discussion](https://github.com/oss-review-toolkit/ort/pull/4097#issuecomment-852254603) with @MarcelBochtler in one of my PRs, and as this is not a change that requires our users to react, nor does it change the results. It would only be perceived as a (temporary) performance "regression". So from my point of view, we could drop the "breaking change" label.

sschuberth (approved):

model/src/main/kotlin/VcsInfo.kt
```diff
@@ -128,8 +121,9 @@ private class VcsInfoDeserializer : StdDeserializer<VcsInfo>(VcsInfo::class.java
         return VcsInfo(
             VcsType(node["type"].textValueOrEmpty()),
             node["url"].textValueOrEmpty(),
-            node["revision"].textValueOrEmpty(),
-            (node["resolved_revision"] ?: node["resolvedRevision"])?.textValue(),
+            // For backward compatibility, if resolved_revision is set, prefer it over revision because it is more
+            // specific.
+            node["resolved_revision"]?.textValue() ?: node["revision"].textValueOrEmpty(),
```
fviernau, line 126: This favors an empty `resolvedRevision` over a non-empty `revision`, intended? ...Same for RepositoryProvenance deserializer.

model/src/main/kotlin/Provenance.kt
```diff
@@ -95,12 +93,11 @@ private class ProvenanceDeserializer : StdDeserializer<Provenance>(Provenance::c
             }
             node.has("vcs_info") -> {
                 val vcsInfo = jsonMapper.treeToValue<VcsInfo>(node["vcs_info"])!!
-                RepositoryProvenance(vcsInfo)
+                // For backward compatibility, if there is no resolved_revision use the revision from vcsInfo.
+                val resolvedRevision = node["resolved_revision"]?.textValue() ?: vcsInfo.revision
```
fviernau, line 97: Why is it necessary to fallback to the `revision`? The `match()` function before never matched if there is not `resolvedRevision` and due to this fallback it would match.

```
        // If no VCS information is present either, or it does not have a resolved revision, there is no way of
        // verifying matching provenance.
        if (vcsInfo.resolvedRevision == null) return false
```

fviernau (changes requested):

model/src/main/kotlin/VcsInfo.kt
sschuberth: Note that `resolvedRevision` is / was nullable (unlike `revision`). So if there was no `resolvedRevision` is should have been *null* instead of empty. That's why I believe it's highly unlikely to get an empty `resolvedRevision` here.

model/src/main/kotlin/Provenance.kt
mnonnenmacher: That's right, but in the backward compatibility case the deserialization of `VcsInfo` also prefers `resolvedRevision` over `revision`, that's why we have to fallback to `revision` to get the resolved revision of the scan result. Like I mentioned above, this does not always work which is why some packages will have to be rescanned, but at least it works in most cases.

mnonnenmacher:
> > I have marked this as a breaking change because it partly invalidates stored scan results.
> 
> I had a [similar discussion](https://github.com/oss-review-toolkit/ort/pull/4097#issuecomment-852254603) with @MarcelBochtler in one of my PRs, and as this is not a change that requires our users to react, nor does it change the results. It would only be perceived as a (temporary) performance "regression". So from my point of view, we could drop the "breaking change" label.

Makes sense, I have added the release notes label instead.

fviernau (approved):

mnonnenmacher:
> > > I have marked this as a breaking change because it partly invalidates stored scan results.
> > 
> > 
> > I had a [similar discussion](https://github.com/oss-review-toolkit/ort/pull/4097#issuecomment-852254603) with @MarcelBochtler in one of my PRs, and as this is not a change that requires our users to react, nor does it change the results. It would only be perceived as a (temporary) performance "regression". So from my point of view, we could drop the "breaking change" label.
> 
> Makes sense, I have added the release notes label instead.

I have re-added the breaking change label, because if ORT is used programmatically the API change does require adaption.

PR #4138, opened 2021-06-09
fossid-webapp: add credentials from netrc file to repo URL

nnobelis:
The usual way of dealyíng with repositories requiring authentication
in FossID is by adding the public key of the FossID instance in the
repository and using the SSH protocol.
Unfortunately, in several environments, SSH is not available or
forbidden and FossID has no mechanisms of managing HTTP credentials.
The usual workaround is to use a netrc file but this requires admin
access to the system running the FossID instance.
This commit is an alternative: it extracts credentials from the system
running ORT and prepends them to the repository URL, before giving it
to FossID.
Please note that doing so is highly insecure because the credentials
will be visible in FossID logs. Therefore this behavior is an opt-in
option that must be enabled with the scanner option
"credentialsFromNetrc".

Please note that I tried to write a functional test for this feature, but since `OrtAuthenticator` does not expose the `netrc` file configuration, this seems impossible to do it without creating a file `~/.netrc` with test contents, which taints the filesystem where the tests run (mocking seems also out of the question since the file names are not accessed through an interface).
Please advice if you have a suggestion.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.scanner.scanners.fossid
 
 import java.io.File
 import java.io.IOException
+import java.net.Authenticator
 import java.time.Instant
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
@@ -59,6 +60,8 @@ import org.ossreviewtoolkit.scanner.AbstractScannerFactory
 import org.ossreviewtoolkit.scanner.RemoteScanner
 import org.ossreviewtoolkit.utils.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.log
+import org.ossreviewtoolkit.utils.replaceCredentialsInUri
+import org.ossreviewtoolkit.utils.toUri
 
 import retrofit2.Retrofit
 import retrofit2.converter.jackson.JacksonConverterFactory
@@ -111,6 +114,33 @@ class FossId(
 
             return projectName
         }
+
+        /**
+         * This function fetches credentials for [repoUrl] and insert them between the URL scheme and the host. If no
+         * matching host is found by [Authenticator], the [repoUrl] is returned untouched.
+         */
+        private fun queryAuthenticator(repoUrl: String): String {
+            val repoUri = repoUrl.toUri().getOrElse {
+                log.warn { "Host cannot be extracted for $repoUrl." }
+                return repoUrl
+            }
+
+            log.info { "Requesting authenticator for host ${repoUri.host} ..." }
+
+            val creds = Authenticator.getDefault().requestPasswordAuthenticationInstance(
+                repoUri.host,
+                null,
+                0,
+                null,
+                null,
+                null,
+                null,
+                Authenticator.RequestorType.SERVER
+            )
+            return creds?.let {
+                repoUrl.replaceCredentialsInUri("${creds.userName}:${String(creds.password)}")
+            } ?: repoUrl
+        }
     }
 
     private val serverUrl: String
@@ -121,6 +151,8 @@ class FossId(
 
     private val service: FossIdRestService
     private val packageNamespaceFilter: String
+
+    private val addAuthenticationToUrl: Boolean
 
     override val version: String
 
@@ -138,6 +170,7 @@ class FossId(
         user = fossIdScannerOptions["user"]
             ?: throw IllegalArgumentException("No FossId User configuration found.")
         packageNamespaceFilter = fossIdScannerOptions["packageNamespaceFilter"].orEmpty()
+        addAuthenticationToUrl = fossIdScannerOptions["addAuthenticationToUrl"].toBoolean()
 
         val client = OkHttpClientHelper.buildClient()
 
@@ -249,8 +282,10 @@ class FossId(
 
                 val scanCode = if (existingScan == null) {
                     log.info { "No scan found for $url and revision $revision. Creating scan ..." }
-                    val scanCode = createScan(projectCode, url, revision)
 
+                    val newUrl = if (addAuthenticationToUrl) queryAuthenticator(url) else url
+
+                    val scanCode = createScan(projectCode, newUrl, revision)
                     log.info { "Initiating data download ..." }
                     service.downloadFromGit(user, apiKey, scanCode)
                         .checkResponse("download data from Git", false)
--- a/utils/src/main/kotlin/Extensions.kt
+++ b/utils/src/main/kotlin/Extensions.kt
@@ -362,11 +362,13 @@ fun String.percentEncode(): String =
         .replace("%7E", "~")
 
 /**
- * Strip any user name / password off the URL represented by this [String]. Return the unmodified [String] if it does
- * not represent a URL or if it does not include a user name.
+ * Replace any user name / password in the URI represented by this [String] with [userInfo]. If [userInfo] is null, the
+ * user name / password are stripped. Return the unmodified [String] if it does not represent a URI.
  */
-fun String.stripCredentialsFromUrl() =
-    toUri { URI(it.scheme, null, it.host, it.port, it.path, it.query, it.fragment).toString() }.getOrDefault(this)
+fun String.replaceCredentialsInUri(userInfo: String? = null) =
+    toUri {
+        URI(it.scheme, userInfo, it.host, it.port, it.path, it.query, it.fragment).toString()
+    }.getOrDefault(this)
 
 /**
  * Return this string lower-cased except for the first character which is upper-cased.
--- a/downloader/src/main/kotlin/Downloader.kt
+++ b/downloader/src/main/kotlin/Downloader.kt
@@ -46,9 +46,9 @@ import org.ossreviewtoolkit.utils.collectMessagesAsString
 import org.ossreviewtoolkit.utils.createOrtTempDir
 import org.ossreviewtoolkit.utils.log
 import org.ossreviewtoolkit.utils.perf
+import org.ossreviewtoolkit.utils.replaceCredentialsInUri
 import org.ossreviewtoolkit.utils.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.safeMkdirs
-import org.ossreviewtoolkit.utils.stripCredentialsFromUrl
 import org.ossreviewtoolkit.utils.unpack
 
 /**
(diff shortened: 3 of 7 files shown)


sschuberth:
> The usual way of dealyíng with repositories requiring authentication

s/dealyíng/dealing/

> in FossID is by adding the public key of the FossID instance in the

s/in the/to the/

> This commit is an alternative: it extracts credentials from the system

s/is an/implements an/
s/it extracts/It extracts/

> running ORT and prepends them to the repository URL, before giving it
> to FossID.

But maybe a better wording would be anyway: "It uses the credentials configured for ORT and injects them to the repository URL sent to FossID for cloning." (Or so.)

nnobelis:
> 
> But maybe a better wording would be anyway: "It uses the credentials configured for ORT and injects them to the repository URL sent to FossID for cloning." (Or so.)

I changed it to
```
This commit implements an alternative: It uses the credentials
configured on ORT's host system and injects them to the repository
URL sent to FossID for cloning the repository.
```
I think this is important to say the credentials are coming from the host configuration, not from ORT configuration.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -111,6 +116,34 @@ class FossId(
 
             return projectName
         }
+
+        /**
+         * This function fetches credentials for [repoUrl] and insert them between the url scheme and the host. If no
+         * matching host is found the in netrc file, the [repoUrl] is returned untouched.
+         */
+        fun getCredentialsFromNetrc(repoUrl: String): String {
```
oheger-bosch, line 124: Is this function only used internally? Then it could be *private*.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -111,6 +116,34 @@ class FossId(
...
+            val matcher = HOST_EXTRACTOR_REGEX.matchEntire(repoUrl)
+            return if (matcher?.groupValues?.size == 4) {
+                val (_, scheme, host, remainder) = matcher.groupValues
+
+                log.info { "Checking .netrc file for host $host ..." }
+
+                val creds = OrtAuthenticator(Authenticator.getDefault()).requestPasswordAuthenticationInstance(
```
oheger-bosch, line 131: Rather than manually creating an `OrtAuthenticator` every time, I think this function should just rely on the default authenticator (and assume that an `OrtAuthenticator` has been installed properly).

This would also help with the testing problems you mentioned: In the test, you could temporarily replace the default `Authenticator` with a mock one and thus get configurable credentials. (Just remember to restore the original default `Authenticator` afterwards, maybe inside a *try*-*finally* block or by means of the KoTest framework.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -111,6 +116,34 @@ class FossId(
...
         }
+
+        /**
+         * This function fetches credentials for [repoUrl] and insert them between the url scheme and the host. If no
+         * matching host is found the in netrc file, the [repoUrl] is returned untouched.
+         */
+        fun getCredentialsFromNetrc(repoUrl: String): String {
+            val matcher = HOST_EXTRACTOR_REGEX.matchEntire(repoUrl)
+            return if (matcher?.groupValues?.size == 4) {
+                val (_, scheme, host, remainder) = matcher.groupValues
```
sschuberth, line 127: Instead of this, I think it would be nicer to convert to an URI (also see our `String.toUri()` extension function) and then use the URI elements.

sschuberth (changes requested):

sschuberth:
> I think this is important to say the credentials are coming from the host configuration, not from ORT configuration.

But actually the opposite is the case: You're asking the *Ort*Authenticator for credentials, and the fact that it (also) uses the `.netrc` internally is an implementation detail. The `OrtAuthenticator` also uses other sources, like environment variables. As such I'd also rename the `credentialsFromNetrc` property to something else, like `useOrtAuthenticator`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: I thought about it but then considering the URL reconstruction, the result seems cumbersome:
` "${it.scheme}${it.schemeSpecificPart}${it.authority}$scheme${creds.userName}:${String(creds.password)}${it.host}${it.port}${it.path}${it.query}${it.fragment}"`
or
` repoUrl.replaceFirst(it.host, "$scheme${creds.userName}:${String(creds.password)}${it.host}")`

URI are immutable objects so we cannot do `setUrerInfo`. 
I find the regex here particularly readable. Or am I missing something ?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Well spotted.

In the end I can use `getDefault()` because the `OkHttpClientHelper` installs it.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: I wanted to access it in testing but since I do not come further with testing I will make it `private`.

nnobelis:
> > I think this is important to say the credentials are coming from the host configuration, not from ORT configuration.
> 
> But actually the opposite is the case: You're asking the _Ort_Authenticator for credentials, and the fact that it (also) uses the `.netrc` internally is an implementation detail. The `OrtAuthenticator` also uses other sources, like environment variables. As such I'd also rename the `credentialsFromNetrc` property to something else, like `useOrtAuthenticator`.

I see your point. I've changed the ocmmit message, the function and property names. Please review.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Sounds reasonable, I will do it.

utils/src/main/kotlin/Extensions.kt
```diff
@@ -362,11 +362,13 @@ fun String.percentEncode(): String =
...
 
 /**
- * Strip any user name / password off the URL represented by this [String]. Return the unmodified [String] if it does
- * not represent a URL or if it does not include a user name.
+ * Strip or replace any user name / password off the URL represented by this [String]. Return the unmodified [String] if
+ * it does not represent a URL or if it does not include a user name.
  */
-fun String.stripCredentialsFromUrl() =
-    toUri { URI(it.scheme, null, it.host, it.port, it.path, it.query, it.fragment).toString() }.getOrDefault(this)
+fun String.replaceCredentialsFromUrl(userInfo: String? = null) =
```
sschuberth, line 368: As [mentioned](https://github.com/oss-review-toolkit/ort/pull/4138#discussion_r648232567), the new name should be `replaceCredentialsInUri`. So "In" instead of "From" and "Uri" instead of "Url".

utils/src/main/kotlin/Extensions.kt
```diff
@@ -362,11 +362,13 @@ fun String.percentEncode(): String =
         .replace("%7E", "~")
 
 /**
- * Strip any user name / password off the URL represented by this [String]. Return the unmodified [String] if it does
- * not represent a URL or if it does not include a user name.
+ * Strip or replace any user name / password off the URL represented by this [String]. Return the unmodified [String] if
+ * it does not represent a URL or if it does not include a user name.
```
sschuberth, line 366: This should say:

> Replace any user name / password in the URI represented by this [String] with [userInfo]. If [userInfo] is null, the user name / password are stripped. Return the unmodified [String] if it does not represent a URI.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -111,6 +115,33 @@ class FossId(
 
             return projectName
         }
+
+        /**
+         * This function fetches credentials for [repoUrl] and insert them between the url scheme and the host. If no
```
sschuberth, line 120: s/url/URL/

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -111,6 +115,33 @@ class FossId(
...
+         * matching host is found by [OrtAuthenticator], the [repoUrl] is returned untouched.
+         */
+        private fun queryOrtAuthenticator(repoUrl: String): String {
+            val repoUri = repoUrl.toUri().getOrElse {
+                log.warn { "Host cannot be extracted for $repoUrl." }
+                return repoUrl
+            }
+
+            log.info { "Checking .netrc file for host ${repoUri.host} ..." }
```
sschuberth, line 129: This still talk specifically about `.netrc`...

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.scanner.scanners.fossid
 
 import java.io.File
 import java.io.IOException
+import java.net.Authenticator
```
sschuberth, line 24: The commit message needs to be adapted to the changes.

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth: The commit mess still talks about "credentialsFromNetrc". And while at it, please start capitalized after the ":" in commit message titles.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -111,6 +115,33 @@ class FossId(
 
             return projectName
         }
+
+        /**
+         * This function fetches credentials for [repoUrl] and insert them between the URL scheme and the host. If no
+         * matching host is found by [OrtAuthenticator], the [repoUrl] is returned untouched.
+         */
+        private fun queryOrtAuthenticator(repoUrl: String): String {
```
sschuberth, line 123: The function name (and docs) still talk about "OrtAuthenticator" whereas you do not directly query the `OrtAuthenticator` anymore, so we should just say "Authenticator".

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -138,6 +171,7 @@ class FossId(
         user = fossIdScannerOptions["user"]
             ?: throw IllegalArgumentException("No FossId User configuration found.")
         packageNamespaceFilter = fossIdScannerOptions["packageNamespaceFilter"].orEmpty()
+        queryAuthenticatorForHttpURLs = fossIdScannerOptions["queryAuthenticatorForHttpURLs"].toBoolean()
```
sschuberth, line 174: Why are these properties called "...HttpURLs"? If I'm not mistaken, you query unconditionally for any protocol, so we should not create the impression it's restricted to HTTP. I'd also use a name that emphasizes the security implications that you mention in the commit message. Something like "addAuthenticationToUrl" or so.

sschuberth (changes requested):

sschuberth (approved):

PR #4141, opened 2021-06-09
helper-cli: Add a command to split the curations.yml

MarcelBochtler:
The command creates a directory structure in the format
 type/namespace/package.yml.
If no namespace is available the structure is type/package.yml.
This directory structure is intended to be used with the
`--package-curations-dir` command.

Related to #4128.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/SplitCurations.kt
+++ b/helper-cli/src/main/kotlin/commands/SplitCurations.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.UsageError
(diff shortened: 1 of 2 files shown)


sschuberth:
> The command creates a directory structure in the format
> type/namespace/package.yml.
> If no namespace is available the structure is type/package.yml.
> This directory structure is intended to be used with the
> `--package-curations-dir` command.

Shouldn't you use `name.yml` instead of `package.yml` to reflect the `Identifier` field it comes from, like for the directory levels? Also, I'd indent the second line with 4 spaces, surround it with blank lines, and use no dot at the end to make it stand out, and add `<>` to emphasize it's a template:

```
The command creates a directory structure in the format

    <type>/<namespace>/<name>.yml

```

helper-cli/src/main/kotlin/commands/SplitCurations.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun run() {
+        if (!outputCurationsDir.list().isNullOrEmpty()) {
+            throw UsageError("Output directory $outputCurationsDir must be empty.")
+        }
+
+        val packageCurations = inputCurationsFile.readValue<List<PackageCuration>>()
+
+        packageCurations.forEach { curation ->
+            val type = curation.id.type.encodeOrUnknown()
+            val namespace = curation.id.namespace.percentEncode()
```
sschuberth, line 64: Note that `percentEncode()` might not be enough, which is why we have `fileSystemEncode()`. With `toPath()` you'd get it for free:

    curation.id.toPath().substringBeforeLast('/')

(If the version is empty it gets replaced by "unknown", so `substringBeforeLast('/')` should still be correct in that case.)

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/SplitCurations.kt
MarcelBochtler: > With toPath() you'd get it for free

I explicitly decided to not use it, as this will create an `unknown` directory when a namespace is not set.
And `curation.id.toPath(emptyValue = "")` is not possible due to:
https://github.com/oss-review-toolkit/ort/blob/8307eb70536dfa2a800de3d7390a01d31343c4da/utils/src/main/kotlin/Extensions.kt#L281-L289

So I'd go with using `fileSystemEncode()` on the namespace as well. For type and name it's already used implicitly.

sschuberth (dismissed):

sschuberth (approved):

PR #4142, opened 2021-06-09
WorkingTree: Make clear that returned revisions are resolved

mnonnenmacher:
WIP: This depends on #4132 to be merged first.

Revisions returned by the working tree are already resolved, so change
the implementation of `getInfo()` to set `isResolvedRevision` to true.
This will affect the VCS properties of projects, because
`PackageManager.processProjectVcs()` merges the VCS infromation captured
from metadata with the VCS information of the working tree.

As a result, the ORT downloader will not try to guess revisions for
projects anymore and instead only use the (resolved) revisions provided
by the analyzer.


mnonnenmacher:
Closed for now, will be continued later once it is clear how we want to model resolved revisions in the `VcsInfo`.

PR #4148, opened 2021-06-10
fossid-webapp: Various improvements

nnobelis:
Please refer to individual commit messages for detail.

### DIFF
--- a/clients/fossid-webapp/src/main/kotlin/model/result/MatchType.kt
+++ b/clients/fossid-webapp/src/main/kotlin/model/result/MatchType.kt
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.clients.fossid.model.result
 
 enum class MatchType {
+    ERROR,
     FAILED,
     FULL,
     IGNORED,
--- a/clients/fossid-webapp/src/main/kotlin/model/status/ScanState.kt
+++ b/clients/fossid-webapp/src/main/kotlin/model/status/ScanState.kt
@@ -25,6 +25,7 @@ enum class ScanState {
     @JsonProperty("AUTO-ID")
     AUTO_ID,
     FINISHED,
+    INTERRUPTED,
     @JsonProperty("NOT STARTED")
     NOT_STARTED,
     QUEUED,
--- a/clients/fossid-webapp/src/main/kotlin/model/status/ScanState.kt
+++ b/clients/fossid-webapp/src/main/kotlin/model/status/ScanState.kt
@@ -21,14 +21,44 @@ package org.ossreviewtoolkit.clients.fossid.model.status
 
 import com.fasterxml.jackson.annotation.JsonProperty
 
+/**
+ * An enumeration for the state of a FossID scan, as returned by the "check_status" operation.
+ */
 enum class ScanState {
+    /**
+     * FossID is automatically applying top matched component.
+     */
     @JsonProperty("AUTO-ID")
     AUTO_ID,
+
+    /**
+     * The scan has been completed.
+     */
     FINISHED,
+
+    /**
+     * The scan has been stopped in the UI. The issue needs to be resolved manually and the scan restarted.
+     */
     INTERRUPTED,
+
+    /**
+     * The scan has not started yet.
+     */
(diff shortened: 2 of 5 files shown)


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -170,6 +170,7 @@ class FossId(
         user = fossIdScannerOptions["user"]
             ?: throw IllegalArgumentException("No FossId User configuration found.")
         packageNamespaceFilter = fossIdScannerOptions["packageNamespaceFilter"].orEmpty()
+        packageAuthorsFilter = fossIdScannerOptions["packageAuthorsFilter"].orEmpty()
```
oheger-bosch, line 176: Is this just a plain string referencing a single author? Then this would be rather limited. Should it be at least a comma-separated list? (Although this would introduce the problem if an author name can contain a comma.)

clients/fossid-webapp/src/main/kotlin/model/result/MatchType.kt
```diff
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.clients.fossid.model.result
 
 enum class MatchType {
+    ERROR,
```
sschuberth, line 23: BTW, I believe you've clarified with FOSSID by now what these enums exactly mean. So would you mind adding according docs for all of these (also for `ScanState`) in another commit?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth, line 176: BTW, I believe all of these FOSSID-specific options also deserve some extra docs, explaining what exactly they're good for (what's the use-case). Can I also expect another commit for that?

sschuberth: And the commit message should make more clear that this is a generic, scanner-specific option, not a top-level `ort.conf` entry.

nnobelis: I wanted to keep things simple.
The standard use case if you have a project with it's dependencies, and you want to only scan the project, i.e. there is only one author.
We used this option for several projects and we never faced this as a limitation.

clients/fossid-webapp/src/main/kotlin/model/result/MatchType.kt
nnobelis: We clarified only the `ScanState` therefore I'll add documentation only for it.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Ok, fine with me. As long as the feature is not needed, we do not have to complicate things.

clients/fossid-webapp/src/main/kotlin/model/status/ScanState.kt
```diff
@@ -22,13 +22,40 @@ package org.ossreviewtoolkit.clients.fossid.model.status
 import com.fasterxml.jackson.annotation.JsonProperty
 
 enum class ScanState {
```
oheger-bosch, line 24: For me as a non-FossID expert, it would be helpful to give some context in the class comment as well, e.g. something like "An enumeration with the status codes reported by FossID when the state of a scan is queried(?)".

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -153,8 +153,21 @@ class FossId(
     private val secretKeys = listOf("serverUrl", "apiKey", "user")
 
     private val service: FossIdRestService
+
+    /**
+     * Scanner option: If this optional filter is set, only packages having an identifier in given namespace will be
+     * scanned.
+     */
     private val packageNamespaceFilter: String
```
oheger-bosch, line 161: Other scanner implementations (at least ScanCode) list the configuration properties they support in the class comment. This is quite nice because you have an overview over the options available at a central place.

sschuberth (approved):

PR #4153, opened 2021-06-10
SplitCurations: Ensure curations for one package are persisted into a single file

MarcelBochtler:
This is a fixup for 503842005511772b695ce23bdbd520cad0b9e658.
Curations also need to be persisted as a list, which is the required
format for the `--package-curations-dir` option.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/SplitCurations.kt
+++ b/helper-cli/src/main/kotlin/commands/SplitCurations.kt
@@ -58,17 +58,14 @@ internal class SplitCurations : CliktCommand(
         }
 
         val packageCurations = inputCurationsFile.readValue<List<PackageCuration>>()
+        val groupedCurations = packageCurations.groupBy {
+            outputCurationsDir.resolve(it.id.type.encodeOrUnknown())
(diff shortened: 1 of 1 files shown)


helper-cli/src/main/kotlin/commands/SplitCurations.kt
```diff
@@ -58,17 +58,18 @@ internal class SplitCurations : CliktCommand(
         }
 
         val packageCurations = inputCurationsFile.readValue<List<PackageCuration>>()
+        val groupedCurations = packageCurations.groupBy { it.copy(id = it.id.copy(version = "<dummy>")).id }
```
fviernau, line 61: Consider grouping it by output file instead of by (dummy) ID.

fviernau (approved):

helper-cli/src/main/kotlin/commands/SplitCurations.kt
MarcelBochtler: Thanks for the hint. That makes way more sense. Changed the implementation.

sschuberth (approved):

fviernau (approved):

PR #4154, opened 2021-06-10
Environment class improvements

sschuberth:


sschuberth:
@mnonnenmacher, in a follow-up PR I plan to add the ORT version to the evaluated model in order to display it in the web app report's about dialog. WOuld you simply add the whole `Environment` to the `EvaluatedModel` class, or only add the ORT version as a `metaData` field to `EvaluatedModel`?

mnonnenmacher (approved):

mnonnenmacher:
> @mnonnenmacher, in a follow-up PR I plan to add the ORT version to the evaluated model in order to display it in the web app report's about dialog. Would you simply add the whole `Environment` to the `EvaluatedModel` class, or only add the ORT version as a `metaData` field to `EvaluatedModel`?

Yes, I think the other data from `Environment` would be useful to show in the web app report as well.

PR #4157, opened 2021-06-11
Add ORT helper command to create a new curation in the correct directory structure

MarcelBochtler:
The directory structure from split-curations is reused.

### DIFF
--- a/helper-cli/src/main/kotlin/HelperMain.kt
+++ b/helper-cli/src/main/kotlin/HelperMain.kt
@@ -42,11 +42,11 @@ import org.ossreviewtoolkit.helper.commands.ListStoredScanResultsCommand
 import org.ossreviewtoolkit.helper.commands.MapCopyrightsCommand
 import org.ossreviewtoolkit.helper.commands.MergeRepositoryConfigurationsCommand
 import org.ossreviewtoolkit.helper.commands.SetLabelsCommand
-import org.ossreviewtoolkit.helper.commands.SplitCurationsCommand
 import org.ossreviewtoolkit.helper.commands.SubtractScanResultsCommand
 import org.ossreviewtoolkit.helper.commands.TransformResultCommand
 import org.ossreviewtoolkit.helper.commands.VerifySourceArtifactCurationsCommand
 import org.ossreviewtoolkit.helper.commands.packageconfig.PackageConfigurationCommand
+import org.ossreviewtoolkit.helper.commands.packagecuration.PackageCurationCommand
 import org.ossreviewtoolkit.helper.commands.repoconfig.RepositoryConfigurationCommand
 import org.ossreviewtoolkit.helper.common.ORTH_NAME
 import org.ossreviewtoolkit.utils.printStackTrace
@@ -84,9 +84,9 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
             MapCopyrightsCommand(),
             MergeRepositoryConfigurationsCommand(),
             PackageConfigurationCommand(),
+            PackageCurationCommand(),
             RepositoryConfigurationCommand(),
             SetLabelsCommand(),
-            SplitCurationsCommand(),
             SubtractScanResultsCommand(),
             TransformResultCommand(),
             VerifySourceArtifactCurationsCommand()
--- a/helper-cli/src/main/kotlin/HelperMain.kt
+++ b/helper-cli/src/main/kotlin/HelperMain.kt
@@ -42,7 +42,7 @@ import org.ossreviewtoolkit.helper.commands.ListStoredScanResultsCommand
 import org.ossreviewtoolkit.helper.commands.MapCopyrightsCommand
 import org.ossreviewtoolkit.helper.commands.MergeRepositoryConfigurationsCommand
 import org.ossreviewtoolkit.helper.commands.SetLabelsCommand
-import org.ossreviewtoolkit.helper.commands.SplitCurations
+import org.ossreviewtoolkit.helper.commands.SplitCurationsCommand
 import org.ossreviewtoolkit.helper.commands.SubtractScanResultsCommand
 import org.ossreviewtoolkit.helper.commands.TransformResultCommand
 import org.ossreviewtoolkit.helper.commands.VerifySourceArtifactCurationsCommand
@@ -86,7 +86,7 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
             PackageConfigurationCommand(),
             RepositoryConfigurationCommand(),
             SetLabelsCommand(),
-            SplitCurations(),
+            SplitCurationsCommand(),
             SubtractScanResultsCommand(),
             TransformResultCommand(),
             VerifySourceArtifactCurationsCommand()
--- a/helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationCommand.kt
@@ -25,6 +25,7 @@ import com.github.ajalt.clikt.core.subcommands
 internal class PackageCurationCommand : CliktCommand() {
     init {
         subcommands(
+            CreateCommand(),
             SplitCommand()
         )
     }
--- a/helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationCommand.kt
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands.packagecuration
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.subcommands
+
+internal class PackageCurationCommand : CliktCommand() {
+    init {
+        subcommands(
+            SplitCommand()
+        )
+    }
+
+    override fun run() {
+        // Intentionally empty, because all logic is implemented inside the sub commands.
+    }
+}
--- a/helper-cli/src/main/kotlin/common/Utils.kt
+++ b/helper-cli/src/main/kotlin/common/Utils.kt
@@ -66,6 +66,8 @@ import org.ossreviewtoolkit.model.writeValue
 import org.ossreviewtoolkit.spdx.SpdxExpression
 import org.ossreviewtoolkit.spdx.SpdxSingleLicenseExpression
 import org.ossreviewtoolkit.utils.CopyrightStatementsProcessor
+import org.ossreviewtoolkit.utils.encodeOrUnknown
+import org.ossreviewtoolkit.utils.fileSystemEncode
 import org.ossreviewtoolkit.utils.isSymbolicLink
 import org.ossreviewtoolkit.utils.replaceCredentialsInUri
 import org.ossreviewtoolkit.utils.withoutPrefix
@@ -118,6 +120,14 @@ internal fun findRepositories(directory: File): Map<String, VcsInfo> {
 
     return ortResult.repository.nestedRepositories
 }
+
+/**
+ * Build the file for the split curations.
+ */
+internal fun getSplitCurationFile(parent: File, packageId: Identifier, fileExtension: String) =
+    parent.resolve(packageId.type.encodeOrUnknown())
+        .resolve(packageId.namespace.fileSystemEncode())
+        .resolve("${packageId.name.encodeOrUnknown()}.$fileExtension")
 
 /**
  * Return an approximation for the Set-Cover Problem, see https://en.wikipedia.org/wiki/Set_cover_problem.
--- a/helper-cli/src/main/kotlin/commands/SplitCurationsCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/SplitCurationsCommand.kt
@@ -33,7 +33,7 @@ import org.ossreviewtoolkit.utils.encodeOrUnknown
 import org.ossreviewtoolkit.utils.expandTilde
 import org.ossreviewtoolkit.utils.fileSystemEncode
 
-internal class SplitCurations : CliktCommand(
+internal class SplitCurationsCommand : CliktCommand(
     help = "Split a single curations file into a directory structure using the format '<type>/<namespace>/<name>.yml'."
 ) {
     private val inputCurationsFile by option(
--- a/helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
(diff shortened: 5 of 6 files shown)


helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -72,6 +72,7 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
         }
 
         subcommands(
+            CurationHelperCommand(),
```
fviernau, line 75: commit-msg: commant -> command

helper-cli/src/main/kotlin/commands/curationshelper/CurationHelperCommand.kt
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands.curationshelper
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.subcommands
+
+internal class CurationHelperCommand : CliktCommand() {
```
fviernau, line 25: I'd prefer doing the "organization" thing later when there is more knowledge. Now there is just one command so it does not make too much sense as-is, because the context of commands which will be added in the future is unknown. Maybe other options for organizing that pop up by then.

helper-cli/src/main/kotlin/commands/curationshelper/CurationHelperCommand.kt
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
```
fviernau, line 19: Please remove "helper" from command and package name and for "curations" please decide for plural or singular consistently. Using "PackageCuration" and "packagecuration" (package) would be my proposal.

helper-cli/src/main/kotlin/commands/curationshelper/CreatePackageCurationCommand.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+internal class CreatePackageCurationCommand : CliktCommand(
+    help = "Create a single curation file for a package id complying to the directory structure created by the" +
+            "`split-curations` sub-command."
+) {
+    private val packageId by option(
+        "--package-id",
+        help = "The target package id for which a curation template should be generated."
```
fviernau, line 47: Shall we ditch "template" here to better align with the command name? (and maybe because it's not relevant)

helper-cli/src/main/kotlin/commands/curationshelper/CreatePackageCurationCommand.kt
```diff
@@ -0,0 +1,84 @@
...
+    ).convert { Identifier(it) }
+        .required()
+
+    private val outputDir by option(
+        "--package-curations-dir",
+        help = "The parent directory where the curation template should be created."
```
fviernau, line 53: How about shorter: "The output package curations directory."?

helper-cli/src/main/kotlin/commands/curationshelper/CreatePackageCurationCommand.kt
```diff
@@ -0,0 +1,84 @@
...
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = false, canBeDir = true, mustBeWritable = true, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+
+    private val fileFormat by option(
```
fviernau, line 59: Is this really needed? "helper-cli" so far IIRC only outputs YAML and I prefer to keep that unless there is a specific requirement.

helper-cli/src/main/kotlin/common/Utils.kt
```diff
@@ -119,6 +121,14 @@ internal fun findRepositories(directory: File): Map<String, VcsInfo> {
     return ortResult.repository.nestedRepositories
 }
 
+/**
+ * Build the file for the split curations.
+ */
+internal fun getSplitCurationFile(parent: File, packageId: Identifier, fileExtension: String) =
```
fviernau, line 127: Other bits of the "helper-cli" use `Identifier.toPath()`. See "generate package config" command.
Would it make sense to re-use it here as well? (A small modification is needed to exclude the version from the path)

helper-cli/src/main/kotlin/common/Utils.kt
```diff
@@ -119,6 +121,14 @@ internal fun findRepositories(directory: File): Map<String, VcsInfo> {
...
+    parent.resolve(packageId.type.encodeOrUnknown())
+        .resolve(packageId.namespace.fileSystemEncode())
+        .resolve("${packageId.name.encodeOrUnknown()}.$fileExtension")
```
fviernau, line 130: Note that file.resolve("a/b/c") also works.

MarcelBochtler:
I'm thinking about dropping the `curations` from `split-curations` and `create-package-curation` as these are already given by the `curations-helper` sub-command.
What do you think?

fviernau (changes requested):

helper-cli/src/main/kotlin/common/Utils.kt
MarcelBochtler: This is intentionally not used.
See the discussion here: https://github.com/oss-review-toolkit/ort/pull/4141#discussion_r648406981

helper-cli/src/main/kotlin/commands/curationshelper/CurationHelperCommand.kt
MarcelBochtler: As I added a second command in a later commit of this PR. Is your change request still valid?

helper-cli/src/main/kotlin/commands/curationshelper/CreatePackageCurationCommand.kt
MarcelBochtler: It's not required for our use case. I'll remove it if the intention of the `helper-cli` is not intended to support all of ORT's data file formats.

helper-cli/src/main/kotlin/commands/curationshelper/CurationHelperCommand.kt
MarcelBochtler: I renamed the the command and the package name.
The singular/plural choice for `create-package-curation` and `split-curations` was intentional as the first one creates a single file, the second one splits multiple curations into multiple files.

But I wouldn't care to adjust the command naming to singular if you think this is better for the usability.

helper-cli/src/main/kotlin/commands/curationshelper/CurationHelperCommand.kt
fviernau: no, I wasn't aware when writing the comment.

helper-cli/src/main/kotlin/common/Utils.kt
MarcelBochtler: While it works, does it offer any benefits?

MarcelBochtler: IMO chaining the `resolve()` calls is more clear than adding everything into `"${...}/${...}"`

helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationCommand.kt
```diff
@@ -0,0 +1,35 @@
...
+
+package org.ossreviewtoolkit.helper.commands.packagecuration
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.subcommands
+
+internal class PackageCurationCommand : CliktCommand() {
+    init {
+        subcommands(
+            SplitCurationsCommand()
```
fviernau, line 29: The command line call would be `orth package-curation split-curations`.
What do you think about just `orth package-curation split`?

(maybe this slightly gives a hint why introducing the grouping that early is difficult. E.g. does the subcommand name `package-curation` fit further future subcommands?)

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
```diff
@@ -0,0 +1,76 @@
...
+
+import org.ossreviewtoolkit.helper.common.getSplitCurationFile
+import org.ossreviewtoolkit.model.FileFormat
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.PackageCuration
+import org.ossreviewtoolkit.model.PackageCurationData
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+internal class CreatePackageCurationCommand : CliktCommand(
```
fviernau, line 39: The CLI call would be `orth package-configuration create-package-configuration`.
How about `orth package-configuration create`?

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
```diff
@@ -0,0 +1,76 @@
...
+        val outputFile = getSplitCurationFile(outputDir, packageId, FileFormat.YAML.fileExtension)
+
+        if (outputFile.exists()) {
+            throw UsageError("File ${outputFile.absolutePath} already exists.")
+        }
+
+        try {
+            outputFile.parentFile.mkdirs()
+            outputFile.writeValue(listOf(
+                // Adding an empty comment to avoid `{}` for an empty PackageCurationData.
```
fviernau, line 67: Is there a reason not to exclude null comment values from serialization?

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
```diff
@@ -0,0 +1,76 @@
...
+                PackageCuration(packageId, PackageCurationData(comment = ""))
+            ))
+        } catch (e: IOException) {
+            throw IOException("Failed to create ${outputFile.absoluteFile}.", e)
```
fviernau, line 71: Not sure if it makes sense. Have you checked if throwing ProgramResult fits ?

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.ossreviewtoolkit.model.FileFormat
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.PackageCuration
+import org.ossreviewtoolkit.model.PackageCurationData
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+internal class CreatePackageCurationCommand : CliktCommand(
+    help = "Create a single curation file for a package id complying to the directory structure created by the" +
+            "`split-curations` sub-command."
```
fviernau, line 41: How about not mentioning the other command, because that hierarchical structure will be
common anyway I presume.

Create a curation file for a package id using a hierarchical directory structure.

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/curationshelper/CreatePackageCurationCommand.kt
```diff
@@ -0,0 +1,84 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands.curationshelper
```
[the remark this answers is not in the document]
MarcelBochtler: This was already done. I think you reviewed an old version here.

helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationCommand.kt
MarcelBochtler: > What do you think about just `orth package-curation split`?

Yes. I had this thought as well: https://github.com/oss-review-toolkit/ort/pull/4157#issuecomment-859522680

> E.g. does the subcommand name `package-curation` fit further future subcommands?

I think it does, but I also tried to name it a little more generic by using `curation-helper`, which is somehow duplicated wording for ORT's `helper-cli`.

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
MarcelBochtler: Yes see: https://github.com/oss-review-toolkit/ort/pull/4157#issuecomment-859522680.

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
MarcelBochtler: Changed it to your suggestion. Do you think it makes sense to mention the structure from `split-curations` again?
https://github.com/oss-review-toolkit/ort/blob/b093d2ec08b69abfe73f4eacdb8564f58479edd1/helper-cli/src/main/kotlin/commands/SplitCurations.kt#L37

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
MarcelBochtler: Clikt's `ProgramResult` is not an Exception. It only allows me to set the `statusCode` the exception would therefore be lost.
As throwing an `IOException` already results in the statusCode `1`, I think this is the better way to handle this.

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
MarcelBochtler: Null comments are already excluded from serialization. See: https://github.com/oss-review-toolkit/ort/blob/eed301187ed11d8527ee89b558ba29ac0151e8ad/model/src/main/kotlin/PackageCurationData.kt#L36-L42

The issue here is with the alternative: `PackageCuration(packageId, PackageCurationData())` this is how the result looks like:
```yml
- id: "Maven:aaa.bbb.ccc:package-name:5.6.0"
  curations: {}
```

`curations` could only be removed from the resulting file by making `PackageCurationData` nullable in `PackageCuration`. I do not want to do this for a `helper-cli` script.

This is why I opted to add the comment with an empty String, as the comment is the only value I can expect to be added into most curation.
Which results in
```yml
- id: "Maven:aaa.bbb.ccc:package-name:5.6.0"
  curations:
    comment: ""
```

helper-cli/src/main/kotlin/common/Utils.kt
fviernau: For package configuration, I believe '_' is used instead of 'unknown'. Would that make sense here as well (for consistency) ?

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
fviernau: Thanks, makes sense now - you're right.

helper-cli/src/main/kotlin/commands/packagecuration/CreatePackageCurationCommand.kt
fviernau: Thanks.

helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -72,6 +72,7 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
         }
 
         subcommands(
+            PackageCurationCommand(),
```
fviernau, line 75: Please move below line 87 to adhere to ordering.

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/curationshelper/CreatePackageCurationCommand.kt
fviernau: What the intention of the `helper-cli` is is not written in stone, it's also not supposed to be dictated by a single person but ideally a result of collective work changing iteratively.

Anyhow, I'd propose to drop it for consistency (at least) still and as I believe it falls under YAGNI.

fviernau (approved):

helper-cli/src/main/kotlin/common/Utils.kt
MarcelBochtler: Reference to the package configuration:
https://github.com/oss-review-toolkit/ort/blob/f8d90dee54e1e6cdca5d0a6bed99e92bdec5e650/helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt#L109

For consistency this makes sense, yes.
But I don't really like the resulting directory structure if using `_` as parent folder if the namespace is unknown.

For multiple package manager that will mean that everything is moved to the "useless" folder.
e.g. every `Crate`, `GoMod` and most `NPM` dependencies will be moved into a `_` folder.

MarcelBochtler:
After discussion with @fviernau, I'll change the directory structure for packages with unknown namespaces to use the `_` folder to be consistent with `package configuration` and to maintain one level per id-coordinate.

PR #3827, opened 2021-03-29
Detekt: Enable AbsentOrWrongFileLicense

mnonnenmacher:
Enable the `AbsentOrWrongFileLicense` [1] rule to verify that the
correct license header is used in all Kotlin files.

[1] https://detekt.github.io/detekt/comments.html#absentorwrongfilelicense

PR #4163, opened 2021-06-13
Various fixes to license headers

sschuberth:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+#     https://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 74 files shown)


oheger-bosch (dismissed):

nnobelis (approved):

oheger-bosch (approved):
Just wondering: Many projects I was working on in the past used some plugins to check the license headers of project files. Would this be an option here as well, so that it can be ensured that no files are missed by such updates?

sschuberth:
> Just wondering: Many projects I was working on in the past used some plugins to check the license headers of project files. Would this be an option here as well, so that it can be ensured that no files are missed by such updates?

We actually *do* have a check in place. However, as the check is performed by *detekt* (also see [this](https://github.com/oss-review-toolkit/ort/pull/3827)), we only cover files known by *detekt*. Maybe we should consider doing it like we did before 1cdb9b04214c4044a4911e8a4a69497105088816, i.e. in a custom Gradle task.

mnonnenmacher (approved):

sschuberth:
@mnonnenmacher I had to rebase to resolve a "conflict" that GitHub showed, but my local Git didn't even complain about...

mnonnenmacher (approved):

PR #4166, opened 2021-06-14
Abstract the access to dependencies in an OrtResult via an interface

oheger-bosch:
This PR is related to #4069. In order to change the representation of dependency information in `OrtResult`, code that accesses this information has to be adapted. To have better control over the changes required, introduce an interface to query information about dependencies. This interface can then be implemented in terms of the classic dependency tree and the new dependency graph.

Unfortunately, there are many consumers of dependency information, which have to be modified. To prevent this PR from getting too big, not all of them have been converted by now. Follow-up PRs will deal with the remaining ones and then also remove possibilities to circumvent the new API as far as possible.


oheger-bosch:
This is still WIP, but I would be happy to receive some feedback early, as this PR has to touch a number of different areas.

The basic idea is that there is now a generic API for accessing dependency information from an `OrtResult` that is independent on concrete data structures. I am still in the progress of identifying use cases for this API, add corresponding functions to the interface, and rewrite client code to use this API rather than dealing with `Scope`, `Project`, or `PackageReference` directly.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -0,0 +1,94 @@
...
+ * dependency is matched or not.
+ */
+typealias DependencyMatcher = (DependencyNode) -> Boolean
+
+/**
+ * An interface allowing the navigation through the dependency information contained in an [OrtResult], independent on
+ * the concrete storage representation.
+ *
+ * The dependencies detected by the analyzer are represented in an [OrtResult] in an optimized form which is not
+ * accessible easily. To simplify dealing with dependencies, this interface is introduced. It offers functionality
```
mnonnenmacher, line 37: Nit: Remove "this interface is introduced", just say "...this interface offers functionality...".

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -0,0 +1,94 @@
...
+         * A pre-defined [DependencyMatcher] that matches all dependencies. It can be used to traverse the whole
+         * dependency graph.
+         */
+        val MATCH_ALL: DependencyMatcher = { true }
+    }
+
+    /**
+     * Return a set with the names of all the scopes defined for the given [project].
+     */
+    fun scopeNames(project: Project): Set<String>
```
mnonnenmacher, line 67: The function could have a default implementation `= scopeNames(project.id)`. Same for the functions below. Is the duplication of the function for `project` and `projectId` really required?

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Return a set with the names of all the scopes defined for the project with the given [projectId]. If the project
+     * cannot be resolved, return an empty set.
```
mnonnenmacher, line 71: It depends on where this function is used, but in general I would prefer if it would throw an exception if the project is not part of the ORT result, otherwise it could hide bugs. Same for the functions below.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -0,0 +1,94 @@
...
+     */
+    fun scopeNames(projectId: Identifier): Set<String>
+
+    /**
+     * Return a sequence with [DependencyNode]s for the given [project] and [scopeName]. From this sequence, the whole
```
mnonnenmacher, line 76: "...[DependencyNode]s **for the direct depedendencies** for the given..."

model/src/main/kotlin/DependencyTreeNavigator.kt
```diff
@@ -0,0 +1,56 @@
...
+
+package org.ossreviewtoolkit.model
+
+/**
+ * A [DependencyNavigator] implementation based on the classic dependency tree format, which is represented by the
+ * [Project.scopes] property.
+ */
+class DependencyTreeNavigator(
+    /** The underlying [OrtResult]. */
+    val result: OrtResult
```
mnonnenmacher, line 28: Consider making this private.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -464,4 +464,9 @@ data class OrtResult(
                 result = analyzer.result.withScopesResolved()
             )
         )
+
+    /**
+     * Return an object that can be used to navigate the dependency information contained in this result.
+     */
+    fun dependencyNavigator(): DependencyNavigator = DependencyTreeNavigator(this)
```
mnonnenmacher: This could be a lazy property instead to make the call not require parenthesis and since you change it to reuse the same object in a later commit anyway.

model/src/test/kotlin/DependencyTreeNavigatorTest.kt
```diff
@@ -0,0 +1,179 @@
...
+            "return a map with scopes and their dependencies for an unknown project ID" {
+                val navigator = OrtResult.EMPTY.dependencyNavigator()
+
+                navigator.scopeDependencies(Identifier.EMPTY).keys should beEmpty()
+            }
+
+            "return a map with scopes and their dependencies up to a given maxDepth" {
+                val navigator = testResult.dependencyNavigator()
+
+                val scopeDependencies = navigator.scopeDependencies(PROJECT_ID, maxDepth = 1)
```
mnonnenmacher, line 141: I think it would be good to also test with another depth, especially as 1 is the simplest case.

mnonnenmacher (changes requested):
I only managed to review the first three commits so far, I only looked at the other commits very shortly. I agree with the general approach, but I would have to look at the other commits in detail to see how the new navigator is used to before I can provide more high-level feedback about the approach.

model/src/main/kotlin/DependencyNode.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    /** The [PackageLinkage] of this dependency. */
+    val linkage: PackageLinkage
+
+    /** A list with issues that occurred while resolving this dependency. */
+    val issues: List<OrtIssue>
+
+    /**
+     * Visit the dependencies of this [DependencyNode] by calling the specified [block] with a sequence of all child
+     * nodes. The code block can produce a result, which is returned by this function.
```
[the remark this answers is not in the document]
oheger-bosch: The class comment has been improved in this regard.

model/src/main/kotlin/DependencyNode.kt
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * Notes:
+     * - Not all storage formats used for dependency information use structures that can be mapped directly to this
+     *   interface; therefore, some kind of wrapping may be required. To allow for optimizations, this function
+     *   operates of a sequence rather than a collection; a collection would require wrapping a whole level of
+     *   dependencies, while this is not necessarily required with a sequence.
+     * - The dependency graph to navigate should be free of cycles; so the caller does not have to bother about this.
```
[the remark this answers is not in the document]
oheger-bosch: What I want to express is, that the dependency graphs, as produced by package managers, are free of cycles - the package manager implementations contain extra checks for this purpose. Therefore, client code does not need to protect itself against infinite loops. I changed the wording to hopefully make this clearer.

model/src/main/kotlin/DependencyNavigator.kt
oheger-bosch: This duplication bothers me, too. I took another look at the call sites, and it seems that always a full `Project` is used. So I will remove the variants that work on a project ID; this will get rid off a bunch of functions, and also the test class gets much leaner.

model/src/main/kotlin/DependencyNavigator.kt
oheger-bosch: The functions accepting a project ID have been dropped.

model/src/main/kotlin/OrtResult.kt
oheger-bosch: Done

model/src/main/kotlin/DependencyTreeNavigator.kt
oheger-bosch: The class should be private, I agree. It is, however, currently still used by a test helper class from another package. This should no longer be necessary when the expected analyzer test results are reworked.

model/src/main/kotlin/DependencyNavigator.kt
oheger-bosch: There is now no need any more to keep a reference to an `OrtResult`. As the class has no more state, it has been converted to an object.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -463,4 +463,9 @@ data class OrtResult(
                 result = analyzer.result.withScopesResolved()
             )
         )
+
+    /**
+     * Return an object that can be used to navigate the dependency information contained in this result.
+     */
+    fun dependencyNavigator(): DependencyNavigator = DependencyTreeNavigator
```
sschuberth: Why is this hard-coded to a "Dependency*Tree*Navigator"? Is it because the *tree*, as opposed to the *graph*, is available for all package manager implementations?

helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
```diff
@@ -71,7 +71,7 @@ internal class GenerateScopeExcludesCommand : CliktCommand(
 
 private fun OrtResult.generateScopeExcludes(): List<ScopeExclude> {
     val projectScopes = getProjects().flatMap { project ->
-        dependencyNavigator().scopeNames(project)
+        dependencyNavigator.scopeNames(project)
```
sschuberth, line 74: > While at it, make the dependency navigator in OrtResult a property; this simplifies access from clients.

Is there a reason to not do it like this from the beginning, when the dependency navigator in OrtResult was introduced?

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -82,6 +82,19 @@ interface DependencyNavigator {
...
+     * Return a set with the [Identifier]s of packages that are dependencies of package with the given [packageId] in
+     * the given [project]. Starting from [project], the dependency graph is searched for the package in question;
+     * then its dependencies are collected. As usual, it is possible to restrict the dependencies to be fetched with
+     * [maxDepth] and [matcher].
+     */
+    fun packageDependencies(
+        project: Project,
+        packageId: Identifier,
+        maxDepth: Int = -1,
+        matcher: DependencyMatcher = MATCH_ALL
```
sschuberth, line 119: Also here we should maybe think about the parameter order.

model/src/test/kotlin/DependencyTreeNavigatorTest.kt
```diff
@@ -184,7 +185,7 @@ class DependencyTreeNavigatorTest : WordSpec() {
                     Identifier("Maven:com.fasterxml.jackson.core:jackson-annotations:2.8.0") to listOf(
                         Identifier("Maven:net.logstash.logback:logstash-logback-encoder:4.11"),
                         Identifier("Maven:com.fasterxml.jackson.core:jackson-databind:2.8.9")
-                        ),
+                    ),
```
sschuberth, line 191: Nit: This change should go to a previous commit.

reporter/src/main/kotlin/utils/StatisticsCalculator.kt
```diff
@@ -92,9 +92,11 @@ internal class StatisticsCalculator {
         ortResult
             .getProjects()
             .filterNot { project -> ignoreExcluded && ortResult.isExcluded(project.id) }
-            .flatMap { project -> project.scopes }
-            .filterNot { scope -> ignoreExcluded && ortResult.repository.config.excludes.isScopeExcluded(scope) }
-            .map { scope -> scope.getDependencyTreeDepth() }
+            .flatMap { project ->
+                val scopes = ortResult.dependencyNavigator.scopeNames(project)
+                (MutableList(scopes.size) { project }).zip(scopes)
```
sschuberth, line 97: Nit: The parentheses should not be required here.

reporter/src/main/kotlin/utils/StatisticsCalculator.kt
```diff
@@ -92,9 +92,11 @@ internal class StatisticsCalculator {
...
+            }.filterNot { pair -> ignoreExcluded && ortResult.repository.config.excludes.isScopeExcluded(pair.second) }
+            .map { pair -> ortResult.dependencyNavigator.dependencyTreeDepth(pair.first, pair.second) }
```
sschuberth, line 99: Nit: Maybe use a destructing declaration to give the pair member a more speaking name.

cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
```diff
@@ -200,7 +200,8 @@ class UploadResultToSw360Command : CliktCommand(
     private fun getProjectWithPackages(ortResult: OrtResult): Map<Project, List<Package>> =
         ortResult.getProjects(omitExcluded = true).associateWith { project ->
             // Upload the uncurated packages because SW360 also is a package curation provider.
-            project.collectDependencies().mapNotNull { ortResult.getUncuratedPackageById(it) }
+            ortResult.dependencyNavigator.projectDependencies(project)
```
sschuberth, line 203: Commit message:

> When using the dependency graph format

Comma after this part for readability.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -112,6 +112,17 @@ interface DependencyNavigator {
...
         scopeDependencies(project, matcher = MATCH_SUB_PROJECTS).collectDependencies()
 
+    /**
+     * Return a set with the [Identifier]s of all dependencies of the given [project] across all scopes. As usual,
+     * it is possible to restrict the dependencies to be fetched with [maxDepth] and [matcher].
+     */
+    fun projectDependencies(
+        project: Project,
+        maxDepth: Int = -1,
+        matcher: DependencyMatcher = MATCH_ALL
```
sschuberth, line 144: Reminder: Order of parameters.

analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
```diff
@@ -46,6 +46,7 @@ class AnalyzerResultBuilderTest : WordSpec() {
     private val issue2 = OrtIssue(source = "source-2", message = "message-2")
     private val issue3 = OrtIssue(source = "source-3", message = "message-3")
     private val issue4 = OrtIssue(source = "source-4", message = "message-4")
+    private val issue5 = OrtIssue(source = "source-5", message = "message-5")
```
sschuberth, line 49: Commit message: Superfluous line with only "." in it.

model/src/main/kotlin/AnalyzerResult.kt
```diff
@@ -84,8 +84,12 @@ data class AnalyzerResult(
     fun collectIssues(): Map<Identifier, Set<OrtIssue>> {
         val collectedIssues = issues.mapValuesTo(mutableMapOf()) { it.value.toMutableSet() }
 
-        projects.forEach { project ->
-            project.collectIssues().forEach { (id, issues) ->
+        // Collecting issues from projects is necessary only if they use the dependency tree format; otherwise, the
+        // issues can be retrieved from the graph. So, once analyzer results are created with dependency graphs
+        // exclusively, this step can be removed.
+        projects.filter { it.scopeDependencies != null }.forEach { project ->
```
sschuberth, line 90: There's a dedicated `filterNotNull`.

sschuberth (changes requested):

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Return a map with information of the dependencies of a [project] grouped by scopes. The resulting map has the
+     * names of scopes as keys and a set with the identifiers of the dependencies found in this scope as values. With
+     * [maxDepth] the depth of the dependency tree to be traversed can be restricted; negative values mean that there
+     * is no restriction. Use the specified [matcher] to filter for specific dependencies.
+     */
+    fun scopeDependencies(
+        project: Project,
+        maxDepth: Int = -1,
+        matcher: DependencyMatcher = MATCH_ALL
```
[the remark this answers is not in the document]
oheger-bosch: Just two remarks/questions:
* The order of the `maxDepth` and `mather` parameters is analogous to corresponding functions in the old API, especially `PackageReference.collectDependencies()` and `Project.collectDependencies()`.
* As `matcher` is a lambda, isn't it beneficial to have it as the last parameter, or is this irrelevant when using named arguments?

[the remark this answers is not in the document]
sschuberth: Apparently, I don't remember the old / current API well enough 😆 Also, I forgot about `DependencyMatcher` being a lambda, as from this signature it looks more like an `enum`. So long story short, let's keep the order as-is then.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -41,6 +43,20 @@ interface DependencyNavigator {
          * dependency graph.
          */
         val MATCH_ALL: DependencyMatcher = { true }
+
+        /**
+         * A pre-defined [DependencyMatcher] that matches only dependencies with a linkage indicating sub projects.
+         */
+        private val MATCH_SUB_PROJECTS: DependencyMatcher = { node ->
```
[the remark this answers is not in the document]
oheger-bosch: OTOH, the interface is named `DependencyNode`, which seems to be fitting, as dependencies do form a graph. In this sense, the term "dependency tree" is rather a misnomer; actually, there is a set of dependency trees, a forest so to say.

model/src/main/kotlin/AnalyzerResult.kt
sschuberth: Sorry, that seem to be applicable only if the code was `filter { it != null }`.

model/src/main/kotlin/DependencyNavigator.kt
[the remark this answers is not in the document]
sschuberth: Agreed.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -95,6 +95,11 @@ data class OrtResult(
         )
     }
 
+    /** An object that can be used to navigate the dependency information contained in this result. */
+    @get:JsonIgnore
+    val dependencyNavigator: DependencyNavigator
+    get() = DependencyTreeNavigator
```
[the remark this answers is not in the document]
oheger-bosch: `DependencyTreeNavigator` is an object, so no new instances are created. In a later state of the refactoring, it becomes a lazy property when there are different implementations of `DependencyNavigator`.

helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
oheger-bosch: Well, I have to admit, that this course is not really straight-forward. Initially, I thought there might be more dynamic required to select the navigator implementation. Do you want me to change the commits?

sschuberth: It would certainly be nice to get this part of this change squashed into the commit that introduces the dependency navigator. Hoping that you don't run into too many conflicts.

sschuberth: Or at least the change to a property should be extracted to a commit proceeding this commit directly.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -95,6 +95,10 @@ data class OrtResult(
         )
     }
 
+    /** An object that can be used to navigate the dependency information contained in this result. */
+    @get:JsonIgnore
```
sschuberth, line 99: Now this annotation should not be required anymore, or?

oheger-bosch: Yes, it was dropped.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -82,6 +82,19 @@ interface DependencyNavigator {
         matcher: DependencyMatcher = MATCH_ALL
     ): Map<String, Set<Identifier>>
 
+    /**
+     * Return a set with the [Identifier]s of packages that are dependencies of package with the given [packageId] in
+     * the given [project]. Starting from [project], the dependency graph is searched for the package in question;
+     * then its dependencies are collected. As usual, it is possible to restrict the dependencies to be fetched with
```
mnonnenmacher, line 112: nit: I'd remove "as usual".

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -82,6 +82,19 @@ interface DependencyNavigator {
         matcher: DependencyMatcher = MATCH_ALL
     ): Map<String, Set<Identifier>>
 
+    /**
+     * Return a set with the [Identifier]s of packages that are dependencies of package with the given [packageId] in
+     * the given [project]. Starting from [project], the dependency graph is searched for the package in question;
```
mnonnenmacher, line 111: I found the first sentence difficult to read, maybe: "...that are dependencies of **the** package with the given [packageId] in the **dependency tree of** the given [project]."

model/src/test/kotlin/DependencyTreeNavigatorTest.kt
```diff
@@ -170,6 +170,57 @@ class DependencyTreeNavigatorTest : WordSpec() {
...
+                    Identifier("Maven:org.reactivestreams:reactive-streams:1.0.1") to listOf(
+                        Identifier("Maven:com.typesafe.akka:akka-stream_2.12:2.5.6")
+                    ),
+                    Identifier("Maven:org.slf4j:jcl-over-slf4j:1.7.25") to emptyList(),
+                    Identifier("Maven:org.slf4j:slf4j-api:1.7.25") to listOf(
+                        Identifier("Maven:ch.qos.logback:logback-classic:1.2.3")
+                    ),
+                )
+            }
+        }
```
mnonnenmacher, line 226: I like the previous implementation of this test more because:
* It shows the dependency tree next to the expected results which makes it easier to understand if the test is correct.
* The dependency tree was designed to test certain cases, like the same dependency on multiple levels of the tree. I don't know if these cases are also covered by the current input.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -95,9 +97,69 @@ interface DependencyNavigator {
...
     fun collectSubProjects(project: Project): SortedSet<Identifier> =
         scopeDependencies(project, matcher = MATCH_SUB_PROJECTS).collectDependencies()
+
+    /**
+     * Determine the map of shortest paths for all the dependencies of a project, given its map of [scopeDependencies]
+     * and a [scopeResolver] function that retrieves the direct dependency nodes of a scope.
+     */
+    private fun getShortestPathForProject(
+        scopeDependencies: Map<String, Set<Identifier>>,
+        scopeResolver: (String) -> Sequence<DependencyNode>
```
mnonnenmacher, line 167: Is there a case where this is not the `directDependencies()` function? If not I would inline the function to call to simplify the signature.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -95,9 +97,69 @@ interface DependencyNavigator {
...
+    ): Map<String, Map<Identifier, List<Identifier>>> =
```
mnonnenmacher, line 168: Could this be a memory issue for projects with many scopes and large dependency trees (e.g. Android projects), compared to having only the map for one scope at a time? I'm asking because I didn't check if there was a place where they were all in memory at the same time also with the old code.

sschuberth:
Hmm, looks like the ignore statement for `dependencyNavigator` was needed after all?!

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -162,4 +169,12 @@ interface DependencyNavigator {
 
         return result
     }
+
+    /**
+     * Traverse the given sequence of [dependencies] recursively to determine the depth of the dependency tree.
+     */
+    private fun getTreeDepthRec(dependencies: Sequence<DependencyNode>): Int =
```
mnonnenmacher, line 215: Better write out "Recursive".

model/src/test/kotlin/DependencyTreeNavigatorTest.kt
```diff
@@ -231,6 +232,17 @@ class DependencyTreeNavigatorTest : WordSpec() {
                 subProjectIds should containExactly(PROJECT_ID)
             }
         }
+
+        "dependencyTreeDepth" should {
+            "calculate the dependency tree depth for a project" {
+                navigator.dependencyTreeDepth(testProject, "compile") shouldBe 3
+                navigator.dependencyTreeDepth(testProject, "test") shouldBe 2
+            }
```
mnonnenmacher, line 278: I liked about the previous implementation of the test that it was self contained and also tested the `0` and `1` cases.

reporter/src/main/kotlin/utils/StatisticsCalculator.kt
```diff
@@ -92,9 +92,11 @@ internal class StatisticsCalculator {
         ortResult
             .getProjects()
             .filterNot { project -> ignoreExcluded && ortResult.isExcluded(project.id) }
-            .flatMap { project -> project.scopes }
-            .filterNot { scope -> ignoreExcluded && ortResult.repository.config.excludes.isScopeExcluded(scope) }
-            .map { scope -> scope.getDependencyTreeDepth() }
+            .flatMap { project ->
+                val scopes = ortResult.dependencyNavigator.scopeNames(project)
+                MutableList(scopes.size) { project }.zip(scopes)
+            }.filterNot { (_, scope) -> ignoreExcluded && ortResult.repository.config.excludes.isScopeExcluded(scope) }
```
mnonnenmacher, line 98: I'm not sure, but didn't you introduce a function for the excluded check to the dependency navigator in a previous commit?

model/src/test/kotlin/DependencyTreeNavigatorTest.kt
```diff
@@ -222,6 +222,41 @@ class DependencyTreeNavigatorTest : WordSpec() {
             }
         }
 
+        "projectDependencies" should {
+            "return the dependencies of a project" {
+                val scopeDependencies = navigator.scopeDependencies(testProject)
+                val projectDependencies = navigator.projectDependencies(testProject)
+
+                projectDependencies should containExactlyInAnyOrder(
+                    scopeDependencies["compile"].orEmpty() + scopeDependencies["test"].orEmpty()
```
mnonnenmacher, line 234: Maybe use `getValue("compile/test")` to be on the safe side in case of unexpected changes to the test data.

analyzer/src/funTest/kotlin/managers/Extensions.kt
```diff
@@ -55,18 +56,20 @@ fun PackageManager.resolveSingleProject(definitionFile: File, resolveScopes: Boo
...
 /**
  * Return only those packages from the given set of [allPackages] that are referenced by this [Project].
+ * NOTE: The project is known to use the scopes structure for storing its dependencies; therefore, a
+ * [DependencyTreeNavigator] can be used to access this information.
  */
-private fun Project.filterReferencedPackages(allPackages: Set<Package>): SortedSet<Package> =
-    collectDependencies().run { allPackages.filter { it.id in this }.toSortedSet() }
+private fun Project.filterReferencedPackages(allPackages: Set<Package>): SortedSet<Package> {
+    val projectDependencies = DependencyTreeNavigator.projectDependencies(this)
+    return projectDependencies.run { allPackages.filter { it.id in this }.toSortedSet() }
```
mnonnenmacher, line 74: `run` is not needed anymore: `return allPackages.filter { it.id in projectDependencies }.toSortedSet()`

model/src/main/kotlin/ProjectAnalyzerResult.kt
```diff
@@ -46,21 +46,6 @@ data class ProjectAnalyzerResult(
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     val issues: List<OrtIssue> = emptyList()
 ) {
```
mnonnenmacher, line 48: This check helped finding bugs in the past, is there any similar check in the dependency graph builder?

mnonnenmacher (changes requested):

oheger-bosch:
> 
> 
> Hmm, looks like the ignore statement for `dependencyNavigator` was needed after all?!

Indeed. Fortunately, the tests cover this case.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -41,6 +43,20 @@ interface DependencyNavigator {
...
+         */
+        private val MATCH_SUB_PROJECTS: DependencyMatcher = { node ->
+            node.linkage in PackageLinkage.PROJECT_LINKAGE
+        }
+
+        /**
+         * Return a sorted set with all the [Identifier]s contained in the given map of scope dependencies. This is
+         * useful if all dependencies are needed independent on the scope they belong to.
+         */
+        fun Map<String, Set<Identifier>>.collectDependencies(): SortedSet<Identifier> =
```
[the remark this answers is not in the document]
oheger-bosch: Yes, made private.

model/src/test/kotlin/DependencyTreeNavigatorTest.kt
oheger-bosch: The original test was re-added (in a slightly modified form to call the `DependencyNavigator`. So there are now two test cases, one with a real project and one with the synthetic dependency tree.

model/src/main/kotlin/OrtResult.kt
sschuberth: [Turn out](https://github.com/oss-review-toolkit/ort/pull/4166#issuecomment-874807742) it was still needed.

model/src/main/kotlin/DependencyNavigator.kt
oheger-bosch: Done.

model/src/main/kotlin/DependencyNavigator.kt
oheger-bosch: This is definitive something we need to have an eye on. I think the old code mostly used nested loops to iterate over all the scopes of a project and then over the dependencies of a scope; so the access here was sequential.

What I have done for now, is to introduce a new `dependenciesForScope()` function in `DependencyNavigator`, which only fetches the dependencies of a single scope. (There is a default implementation of `scopeDependencies()`, which constructs a map of all scopes.) `getShortestPathsForProject()` now uses this new function and therefore should behave analogously to the previous implementation. We can think about changing the other callers of `scopeDependencies()` in a similar way if necessary.

model/src/test/kotlin/DependencyTreeNavigatorTest.kt
oheger-bosch: Restored the original tests.

reporter/src/main/kotlin/utils/StatisticsCalculator.kt
oheger-bosch: I only added an overloaded function to `Excludes`, which accepts a scope name rather than a full `Scope`.

model/src/main/kotlin/ProjectAnalyzerResult.kt
oheger-bosch: There is no such check yet. I think, it would be possible to add one that checks the packages added to the builder against the list of package identifiers referenced by the graph. But I would prefer to do this in a separate PR.

I am a bit on the fence regarding such runtime checks. Are they really necessary if you have good test coverage?

mnonnenmacher: The problem here is that the amount of different project setups that can be passed to the analyzer is virtually unlimited, so we cannot test for everything. Therefore it's good to have some safeguards in place that fail with a meaningful error message instead of failing in random other places where it might be difficult to figure out the root cause. But I'm fine with doing that in a separate PR.

mnonnenmacher (approved):

PR #4174, opened 2021-06-15
Git: Make the logic to create a missing local tag more robust

sschuberth:
As parsing human-readable output is generally fragile, instead check
whether the revision is a remote tag in order to conditionally create a
matching local tag.

Also, as right now (before the scan-by-repository topic is implemented)
there could be race conditions with multiple paths from a single Git
repository being downloaded in parallel, force to create the local tag
in case it already exists.

### DIFF
--- a/downloader/src/main/kotlin/vcs/Git.kt
+++ b/downloader/src/main/kotlin/vcs/Git.kt
@@ -53,7 +53,6 @@ import org.ossreviewtoolkit.utils.installAuthenticatorAndProxySelector
 import org.ossreviewtoolkit.utils.log
 import org.ossreviewtoolkit.utils.safeMkdirs
 import org.ossreviewtoolkit.utils.showStackTrace
-import org.ossreviewtoolkit.utils.withoutPrefix
 
 // TODO: Make this configurable.
 const val GIT_HISTORY_DEPTH = 50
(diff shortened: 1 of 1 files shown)


sschuberth:
@heliocastro can you please check whether this makes your bogus `fatal: tag '1.1.1' already exists` output go away?

MarcelBochtler (approved):

sschuberth:
The rationale was wrong, and I had to tweak the implementation slightly.

downloader/src/main/kotlin/vcs/Git.kt
```diff
@@ -183,17 +182,13 @@ class Git : VersionControlSystem(), CommandLineTool {
...
-            // not fetched. So create it manually afterwards.
-            val tag = fetch.stderr.lineSequence().mapNotNull {
-                it.withoutPrefix(" * tag ")?.substringBefore("->")?.trim()
-            }.firstOrNull()
-
-            if (revision == tag) {
-                workingTree.runGit("tag", revision, "FETCH_HEAD")
+            // not fetched. So create the local tag manually afterwards.
+            if (revision in workingTree.listRemoteTags()) {
+                workingTree.runGit("tag", "-f", revision, "FETCH_HEAD")
```
MarcelBochtler, line 191: Do you want to mention your rational why the `-f` flat is required in the comment above?
IMO it's important enough to not _only_ be in the commit message.

MarcelBochtler (approved):

heliocastro (approved):
Works for me

downloader/src/main/kotlin/vcs/Git.kt
sschuberth: > IMO it's important enough to not _only_ be in the commit message.

I think it's fine 😛 More seriously, it's temporary eventually anyway, as the `-f` could be removed once scan-by-repo is implemented.

PR #4175, opened 2021-06-15
Evaluator: Add option to provide multiple curation files

MarcelBochtler:
Analogously to #4128.

### DIFF
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -47,7 +47,7 @@ class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCura
         }
     }
 
-    internal val packageCurations: Set<PackageCuration> = run {
+    val packageCurations: Set<PackageCuration> = run {
         val allCurations = curationFiles.mapNotNull { curationFile ->
             runCatching {
                 curationFile.readValueOrDefault(emptyList<PackageCuration>())
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -37,6 +37,16 @@ import org.ossreviewtoolkit.utils.log
 class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCurationProvider {
     constructor(curationFile: File) : this(listOf(curationFile))
 
+    companion object {
+        fun from(file: File? = null, dir: File? = null): FilePackageCurationProvider {
+            val curationFiles = mutableListOf<File>()
+            file?.takeIf { it.isFile }?.let { curationFiles += it }
+            dir?.let { curationFiles += FileFormat.findFilesWithKnownExtensions(it) }
+
+            return FilePackageCurationProvider(curationFiles)
+        }
+    }
+
     internal val packageCurations: Set<PackageCuration> = run {
         val allCurations = curationFiles.mapNotNull { curationFile ->
             runCatching {
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -27,6 +27,7 @@ import org.ossreviewtoolkit.model.FileFormat
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.PackageCuration
 import org.ossreviewtoolkit.model.readValueOrDefault
+import org.ossreviewtoolkit.spdx.getDuplicates
 import org.ossreviewtoolkit.utils.log
 
 /**
@@ -36,15 +37,25 @@ import org.ossreviewtoolkit.utils.log
 class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCurationProvider {
     constructor(curationFile: File) : this(listOf(curationFile))
 
-    internal val packageCurations by lazy {
-        curationFiles.mapNotNull { curationFile ->
+    internal val packageCurations: Set<PackageCuration> = run {
+        val allCurations = curationFiles.mapNotNull { curationFile ->
             runCatching {
                 curationFile.readValueOrDefault(emptyList<PackageCuration>())
             }.onFailure {
                 log.warn { "Failed parsing package curation from '${curationFile.absoluteFile}'." }
             }.getOrNull()
         }.flatten()
+
+        val duplicates = allCurations.getDuplicates()
+
+        if (duplicates.isNotEmpty()) {
+            throw DuplicatedCurationException("Duplicated curation for $duplicates found.")
+        }
+
+        allCurations.toSet()
     }
 
     override fun getCurationsFor(pkgId: Identifier) = packageCurations.filter { it.isApplicable(pkgId) }
 }
+
+private class DuplicatedCurationException(message: String?) : Exception(message)
--- a/analyzer/src/test/kotlin/curation/FilePackageCurationProviderTest.kt
+++ b/analyzer/src/test/kotlin/curation/FilePackageCurationProviderTest.kt
@@ -55,6 +55,24 @@ class FilePackageCurationProviderTest : StringSpec() {
             }
         }
 
+        "Provider can read from a single file and directories" {
+            val provider = FilePackageCurationProvider.from(curationsFile, curationsDir)
+
+            val idsWithExistingCurations = listOf(
+                // File
+                Identifier("npm", "", "ramda", "0.21.0"),
+                // Directory
+                Identifier("maven", "org.ossreviewtoolkit", "example", "1.0"),
+                Identifier("maven", "org.foo", "bar", "0.42")
+            )
+
+            idsWithExistingCurations.forEach {
+                val curations = provider.getCurationsFor(it)
+
+                curations should haveSize(1)
+            }
+        }
+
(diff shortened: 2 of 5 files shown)


cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -247,8 +257,21 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
...
         }
 
-        packageCurationsFile?.let {
-            val curations = it.readValueOrDefault(emptyList<PackageCuration>())
+        val curationFiles = mutableListOf<File>()
+        packageCurationsFile?.takeIf { it.isFile }?.let { curationFiles += it }
+        packageCurationsDir?.let { curationFiles += FileFormat.findFilesWithKnownExtensions(it) }
+
+        val curations = curationFiles.mapNotNull { curationFile ->
+            runCatching {
```
fviernau, line 265: The lines from 265 - 271 are redundant with the ones existing in `FilePackageCurationProvider`, except for one inconsistency. Please remove the duplication.

e.g. would it make sense to refactor it to end up with something like below?
```
val curationsFiles = FilePackageCurationProvider.readFile(packageCurationsFile) 
   + FilePackageCurationProvider.readFile(packageCurationsFile).readDir(packageCurationsDir)
```

model/src/main/kotlin/OrtResult.kt
```diff
@@ -340,7 +340,7 @@ data class OrtResult(
     /**
      * Return a copy of this [OrtResult] with the [PackageCuration]s replaced by the given [curations].
      */
-    fun replacePackageCurations(curations: List<PackageCuration>): OrtResult =
+    fun replacePackageCurations(curations: Collection<PackageCuration>): OrtResult =
```
fviernau, line 343: Just a hint: the result of this function depends on the ordering of the given curations.

fviernau (changes requested):

cli/src/main/kotlin/commands/EvaluatorCommand.kt
fviernau: Maybe even such that there is no more redundancy regarding the ordering, like e.g.

`FilePackageCurationProvider.for(file=packageCurationsDir, dir=packageCurationsFile)`

MarcelBochtler: I refactored it.
The calling code now looks like this:
`val curations = FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations`

Not completely happy with making the parameters `from` nullable, but all solutions I could think of would make the calling code pretty ugly.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -43,7 +43,7 @@ class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCura
             }.onFailure {
                 log.warn { "Failed parsing package curation from '${curationFile.absoluteFile}'." }
             }.getOrNull()
-        }.flatten()
+        }.flatten().toSet()
```
fviernau, line 58: I believe this change may have an effect on the result of applying the curations. It wouldn't have an effect if the ordering wasn't effected and always the foremost of the duplicated curations were removed. I doubt this is the case.

Apart from that, is it deterministic still?

I'm not 100% convinced that this is the right way to address the problem. 
What do you think about the following two approaches
- throw an exception instead
- accept duplicates, but address it elsewhere. Maybe have a duplicates check in your PSV.

I'd be interested in what other devs think about this - maybe @oss-review-toolkit/core-devs ?

analyzer/src/test/kotlin/curation/FilePackageCurationProviderTest.kt
```diff
@@ -55,6 +55,24 @@ class FilePackageCurationProviderTest : StringSpec() {
...
+                Identifier("npm", "", "ramda", "0.21.0"),
+                // Directory
+                Identifier("maven", "org.ossreviewtoolkit", "example", "1.0"),
+                Identifier("maven", "org.foo", "bar", "0.42")
+            )
+
+            idsWithExistingCurations.forEach {
+                val curations = provider.getCurationsFor(it)
+
+                curations should haveSize(1)
```
fviernau, line 72: hint: Maybe it suffices to just do `provider.packageCurations.size shouldBe x`

fviernau (changes requested):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
MarcelBochtler: As I can't see a valid use case to use contradicting duplicated curations, I think explicitly throwing an exception would be the best way to solve this.

fviernau: Should we then also throw an exception when the parsing fails (instead of just logging a warning) ? 

..and probably also ditch the `by lazy` in order to fail early?

MarcelBochtler: > Should we then also throw an exception when the parsing fails

This was suggested by @sschuberth here: https://github.com/oss-review-toolkit/ort/pull/4131#discussion_r648273567
For our internal workflow, our CI checks if the curations can be deserialised anyways.

Does it make sense to support use cases where the user uses a `curations/` directory with other `.yml` and `.json` files (and just print a warning)?
Otherwise I'd suggest to throw an exception and fail the ORT run.

analyzer/src/test/kotlin/curation/FilePackageCurationProviderTest.kt
MarcelBochtler: I'd like to check for the IDs specifically so this test doesn't fail whenever a new test curation is added to the `curations.yml`.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
MarcelBochtler: For now I just added a check and the `Exception` if curations are duplicated.

fviernau: > Does it make sense to support use cases where the user uses a curations/ directory with other .yml and .json files (and just print a warning)?
Otherwise I'd suggest to throw an exception and fail the ORT run.

I believe it makes sense to assume that underneath the given package curations directory all files are package curation files, because it's simpler and its a requirement one can IMO easily ask for. It does not harm any user if it can't put other files there.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -36,15 +37,35 @@ import org.ossreviewtoolkit.utils.log
...
+        fun from(file: File? = null, dir: File? = null): FilePackageCurationProvider {
+            val curationFiles = mutableListOf<File>()
+            file?.takeIf { it.isFile }?.let { curationFiles += it }
+            dir?.let { curationFiles += FileFormat.findFilesWithKnownExtensions(it) }
+
+            return FilePackageCurationProvider(curationFiles)
+        }
+    }
+
+    val packageCurations = run {
```
fviernau, line 50: I'd prefer explicit type `packageCurations: Set<PackageCurations>`, to make it obvious.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -36,15 +37,35 @@ import org.ossreviewtoolkit.utils.log
...
             runCatching {
                 curationFile.readValueOrDefault(emptyList<PackageCuration>())
             }.onFailure {
                 log.warn { "Failed parsing package curation from '${curationFile.absoluteFile}'." }
             }.getOrNull()
         }.flatten()
+
+        val duplicates = allCurations.getDuplicates()
+
+        if (duplicates.isEmpty()) {
```
fviernau, line 61: hint: If the `if-`condition is inverted, the `else` can be dropped.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -36,15 +37,35 @@ import org.ossreviewtoolkit.utils.log
 class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCurationProvider {
     constructor(curationFile: File) : this(listOf(curationFile))
 
-    internal val packageCurations by lazy {
-        curationFiles.mapNotNull { curationFile ->
+    companion object {
```
fviernau, line 40: The `from` function should have gone into the following commit?!

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -47,7 +47,7 @@ class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCura
         }
     }
 
-    val packageCurations = run {
+    internal val packageCurations = run {
```
fviernau: Should this be taken into use directly in the analyzer command?

fviernau (changes requested):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
MarcelBochtler: Not the `packageCurations`, but the `from` function.
Added a commit to reuse the functionality.

fviernau (approved):

PR #4178, opened 2021-06-16
Dockerfile: Change base image to JDK

mmurto:
Analyzing some Gradle projects with JRE fails with a
NullPointerException. Using JDK works, as described in
https://github.com/keeganwitt/docker-gradle/issues/155#issuecomment-706779404.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ RUN --mount=type=cache,target=/root/.gradle/ \
     sed -i -r 's,(^distributionUrl=)(.+)-all\.zip$,\1\2-bin.zip,' gradle/wrapper/gradle-wrapper.properties && \
     ./gradlew --no-daemon --stacktrace -Pversion=$ORT_VERSION :cli:distTar :helper-cli:startScripts
 
-FROM adoptopenjdk:11-jre-hotspot-bionic
+FROM adoptopenjdk:11-jdk-hotspot-bionic
 
 ENV \
     # Package manager versions.


sschuberth:
It's still a bit unclear to me why we'd (sometimes) require a JDK, as we're not actually compiling any Java code when the ORT analyzer runs on a project. At least we shouldn't. Is this only required for ill-configured builds that execute the compiler during Gradle's configuration phase instead of during Gradle's execution phase?

Also, can we please get some numbers about how much the Docker image size increases due to this change?

mnonnenmacher:
> Also, can we please get some numbers about how much the Docker image size increases due to this change?

I just checked on https://hub.docker.com/_/adoptopenjdk?tab=tags&page=1&name=hotspot-bionic, the JRE image is 80.3MB, the JDK image 223.97MB.

sschuberth:
> the JRE image is 80.3MB, the JDK image 223.97MB.

So almost a threefold size increase, quite a bit. Again, what's the use-case *exactly*?

mmurto:
> > Also, can we please get some numbers about how much the Docker image size increases due to this change?
> 
> I just checked on https://hub.docker.com/_/adoptopenjdk?tab=tags&page=1&name=hotspot-bionic, the JRE image is 80.3MB, the JDK image 223.97MB.

This seems to translate for 3.08 GB for the final image with JRE and 3.3 GB with JDK.

> Again, what's the use-case exactly?

I unfortunately am not very familiar with Gradle, so I can't assess whether the project is ill-configured or not. What I can say is that with JDK as base ORT supports _more_ projects, be they properly configured or with erroneous configuration. At least for us, that's a good thing, as we're often not in control of the projects we're analyzing, but just need to get the project analyzed with all its errors. Whether that's worth the increase in image size is of course up to you.

sschuberth:
> Whether that's worth the increase in image size is of course up to you.

I'm really undecided about that one. @mnonnenmacher, what's your opinion? Weren't you mentioning something similar for Scala projects in the chat?

mnonnenmacher:
> > Whether that's worth the increase in image size is of course up to you.
> 
> I'm really undecided about that one. @mnonnenmacher, what's your opinion? Weren't you mentioning something similar for Scala projects in the chat?

That was just the assumption that the same issue could appear with SBT projects, but not based on any issue I have seen. But for Gradle we also analyzed projects where Gradle required the JDK, that's one of the reasons we already use a derived Docker image which has the JDK installed.
IMO the increase in image size is worth it if it means that ORT supports more projects, especially for cases where the ORT user has no control over the project source code.

sschuberth:
> we also analyzed projects where Gradle required the JDK

But what I'd still like to under stand: *Why* / *what for* does just determining the dependencies make Gradle require a JDK?

sschuberth:
> But for Gradle we also analyzed projects where Gradle required the JDK, that's one of the reasons we already use a derived Docker image which has the JDK installed.

So, the concrete use-case seems to be projects with a [`buildSrc` directory](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) that compile build logic at build time.

fviernau:
> So, the concrete use-case seems to be projects with a [`buildSrc` directory](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) that compile build logic at build time.

Interresting! Thanks for sharing!

sschuberth:
Superseded by https://github.com/oss-review-toolkit/ort/pull/4934 which upgrades to Java 17 LTS at the same time.

PR #4179, opened 2021-06-16
Helper curations commands: Use block comments

MarcelBochtler:
Ensure that curations created via `helper-cli` use a block comment.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
@@ -28,12 +28,12 @@ import com.github.ajalt.clikt.parameters.types.file
 
 import java.io.IOException
 
+import org.ossreviewtoolkit.helper.common.createBlockYamlMapper
 import org.ossreviewtoolkit.helper.common.getSplitCurationFile
 import org.ossreviewtoolkit.model.FileFormat
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.PackageCuration
 import org.ossreviewtoolkit.model.PackageCurationData
-import org.ossreviewtoolkit.model.writeValue
 import org.ossreviewtoolkit.utils.expandTilde
 
 internal class CreateCommand : CliktCommand(
@@ -60,12 +60,19 @@ internal class CreateCommand : CliktCommand(
             throw UsageError("File ${outputFile.absolutePath} already exists.")
         }
 
+        val mapper = createBlockYamlMapper()
+
         try {
+            val text = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(
+                listOf(
+                    // A block comment without any text is not valid in YAML. Therefore add a dummy comment with a line
+                    // break to force the YAML mapper to create a block comment.
+                    PackageCuration(packageId, PackageCurationData(comment = "Curation comment.\n"))
+                )
+            )
+
             outputFile.parentFile.mkdirs()
-            outputFile.writeValue(listOf(
-                // Adding an empty comment to avoid `{}` for an empty PackageCurationData.
-                PackageCuration(packageId, PackageCurationData(comment = ""))
-            ))
+            outputFile.writeText(text)
         } catch (e: IOException) {
             throw IOException("Failed to create ${outputFile.absoluteFile}.", e)
         }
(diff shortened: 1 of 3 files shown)


helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
```diff
@@ -26,11 +26,16 @@ import com.github.ajalt.clikt.parameters.options.option
...
 import org.ossreviewtoolkit.helper.common.getSplitCurationFile
+import org.ossreviewtoolkit.helper.common.wrapAt
 import org.ossreviewtoolkit.model.PackageCuration
 import org.ossreviewtoolkit.model.readValue
-import org.ossreviewtoolkit.model.writeValue
 import org.ossreviewtoolkit.utils.expandTilde
+import org.ossreviewtoolkit.utils.safeMkdirs
+
+// Wrap at column 120 minus 6 spaces of indentation.
+private const val COMMENT_WRAP_COLUMN = 120 - 6
```
MarcelBochtler, line 38: @fviernau I added this as it is required for our use case.
Do you think it makes sense to add a flag to enable / disable the block comments?

helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
```diff
@@ -61,8 +66,24 @@ internal class SplitCommand : CliktCommand(
             getSplitCurationFile(outputCurationsDir, it.id, inputCurationsFile.extension)
         }
 
+        val mapper = createBlockYamlMapper()
         groupedCurations.forEach { (outputFile, curations) ->
-            outputFile.writeValue(curations)
+            curations.forEach { originalCuration ->
+                val comment = originalCuration.data.comment?.wrapAt(COMMENT_WRAP_COLUMN)
```
oheger-bosch, line 72: Suggestion: It may be slightly more concise to move the conversion of the comment to an extension function on the curation.

oheger-bosch (approved):

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
```diff
@@ -60,12 +60,19 @@ internal class CreateCommand : CliktCommand(
             throw UsageError("File ${outputFile.absolutePath} already exists.")
         }
 
+        val mapper = createBlockYamlMapper()
+
         try {
+            val text = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(
+                listOf(
+                    // A block comment without any text is not valid in yaml. Therefore add a dummy command to force a
```
sschuberth, line 68: Nit: "YAML".

sschuberth: s/command/comment/?

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
```diff
@@ -60,12 +60,19 @@ internal class CreateCommand : CliktCommand(
...
+                    // block comment.
+                    PackageCuration(packageId, PackageCurationData(comment = "Curation comment.\n"))
```
sschuberth, line 70: Why the `\n` at the end?

sschuberth: Ah, for the same reason as [below](https://github.com/oss-review-toolkit/ort/pull/4179/files#diff-b78e8fca71542309a4a90ff9c1f29927ec8d341894ff60f102efc9eb96bf0f58R75-R76)? Then we should probably also copy the comment.

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
MarcelBochtler: Yes, it's the same reason.
I did add the reason for the `\n` to the comment.

sschuberth (approved):

helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
fviernau: In the long term I couldn't see a reason why this wrapping shouldn't be done by all helper cli commands which output YAML consistently. I wouldn't add a flag for now. Later on may considering introducing a helper-cli config file for such things.

PR #4182, opened 2021-06-17
Dockerfile: Be explicit about the required frontend syntax version

sschuberth:
This is a fixup for 8b07e14 which dropped this header completely, but
Azure DevOps seems to run a version of Docker that defaults to a
different frontend syntax version, so explicitly specify it again.

Also see [1].

[1] https://github.com/moby/buildkit/blob/fa632c5/frontend/dockerfile/docs/syntax.md#build-mounts-run---mount

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1.2
+
 # Copyright (C) 2020 Bosch Software Innovations GmbH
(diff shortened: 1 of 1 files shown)


oheger-bosch (approved):
Looks good. Thanks for the explicit reference to buildkit documentation.

MarcelBochtler (approved):

PR #4183, opened 2021-06-17
ON HOLD: AnalyzerCommand: Print the amount of PackageReference entries by project

fviernau:
For large analyzer results the call to `writeOrtResult()` may run out of
memory. In such cases the size of the Ort result is usually dominated by
the size of the package reference entries. Print the amount of entries
by project to ease debugging.

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -48,6 +48,7 @@ import org.ossreviewtoolkit.cli.utils.inputGroup
 import org.ossreviewtoolkit.cli.utils.outputGroup
 import org.ossreviewtoolkit.cli.utils.writeOrtResult
 import org.ossreviewtoolkit.model.FileFormat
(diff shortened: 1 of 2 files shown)


sschuberth:
My proposal would be to put this on hold until https://github.com/oss-review-toolkit/ort/pull/4130, https://github.com/oss-review-toolkit/ort/pull/4166 and probably some more upcoming PRs related to the dependency graph topic are merged. One we use the dependency graph format by default not only for serialization (#4130) but also in memory, I expect this PR to become obsolete.

Would you agree @fviernau and @oheger-bosch?

oheger-bosch:
Yes, I was just about to comment that this PR is conflicting with the changes in #4166.

Also, the approach to count the package references does not cover the dependencies stored in the dependency graph format. So the results will be incomplete for some package managers.

fviernau:
Ok, then I'll just do my debugging with this unmerged PR and when done close it.

fviernau:
not desired, see above.

PR #4189, opened 2021-06-18
analyzer: Remove the obsolete "experimental" state from GoMod

fviernau:
Meanwhile VCS detection capababilities have been added which are not
identical to / using the originial (GoMod) tool, but are good enough to
remove the experimental state.


mnonnenmacher (approved):

fviernau:
Merge since the failed markdown check is unrelated.

PR #4191, opened 2021-06-18
Helper-cli: Add a command to convert results to the graph format

oheger-bosch:
In future, all ORT results will use the dependency graph format. It
can therefore be useful to have a command that does a corresponding
conversion on existing result files.

The command supports different types of result files. It is also
intended to be used to convert the expected test results to the new
format.

### DIFF
--- a/helper-cli/src/main/kotlin/HelperMain.kt
+++ b/helper-cli/src/main/kotlin/HelperMain.kt
@@ -41,6 +41,7 @@ import org.ossreviewtoolkit.helper.commands.ListPackagesCommand
 import org.ossreviewtoolkit.helper.commands.ListStoredScanResultsCommand
 import org.ossreviewtoolkit.helper.commands.MapCopyrightsCommand
 import org.ossreviewtoolkit.helper.commands.MergeRepositoryConfigurationsCommand
+import org.ossreviewtoolkit.helper.commands.SetDependencyRepresentationCommand
 import org.ossreviewtoolkit.helper.commands.SetLabelsCommand
 import org.ossreviewtoolkit.helper.commands.SubtractScanResultsCommand
 import org.ossreviewtoolkit.helper.commands.TransformResultCommand
@@ -86,6 +87,7 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
             PackageConfigurationCommand(),
             PackageCurationsCommand(),
             RepositoryConfigurationCommand(),
+            SetDependencyRepresentationCommand(),
             SetLabelsCommand(),
             SubtractScanResultsCommand(),
             TransformResultCommand(),
--- a/helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
@@ -1,0 +1,157 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.UsageError
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.enum
+import com.github.ajalt.clikt.parameters.types.file
+
+import org.ossreviewtoolkit.analyzer.AnalyzerResultBuilder
+import org.ossreviewtoolkit.model.AnalyzerResult
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.ProjectAnalyzerResult
+import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.model.utils.DependencyGraphConverter
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+/**
+ * A command to convert the dependency representation in an ORT result file between the different supported formats.
+ *
+ * The input file can contain the serialized form of either a full [OrtResult], an [AnalyzerResult], or a single
+ * [ProjectAnalyzerResult]. If one of the supported results is found, it is converted to the specified target format.
+ * If this target format is already used, then this command does nothing.
+ */
+class SetDependencyRepresentationCommand : CliktCommand(
+    help = "Set the dependency representation of an ORT result to a specific target format."
+) {
+    /**
+     * An enumeration class representing the different supported formats to represent dependencies in an ORT result.
+     * It is used to define the target format that should be set by this command.
+     */
+    enum class TargetFormat {
+        /** The dependency graph format. */
+        GRAPH {
+            override fun convert(result: AnalyzerResult): AnalyzerResult =
+                DependencyGraphConverter.convert(result)
+        };
+
+        /**
+         * Convert the given [result] to the target format defined by this literal.
+         */
+        abstract fun convert(result: AnalyzerResult): AnalyzerResult
+    }
+
+    private val ortFile by option(
+        "--ort-file", "-i",
+        help = "The ORT result file to read as input. This can be the serialized form of an OrtResult, an " +
+                "AnalyzerResult, or a ProjectAnalyzerResult. NOTE: If no output file is specified, this file is " +
+                "overwritten."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+
+    private val outputFile by option(
+        "--output-file", "-o",
+        help = "The file in which to write the result of the conversion. If missing, the input file is overwritten."
+    ).convert { it.expandTilde() }
+        .file(mustExist = false, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+
+    private val targetFormat by option(
+        "--format", "-f",
+        help = "The target format for the conversion."
+    ).enum<TargetFormat>().default(TargetFormat.GRAPH)
+
+    override fun run() {
+        val converters = sequenceOf(::convertOrtResult, ::convertAnalyzerResult, ::convertProjectAnalyzerResult)
+        converters.mapNotNull { it() }.firstOrNull()?.let { writeResult(it) }
+            ?: throw UsageError("$ortFile does not contain a supported result.")
+    }
+
+    /**
+     * Convert a full ORT result file.
+     */
+    private fun convertOrtResult(): Any? =
+        convertResult<OrtResult> { result ->
+            println("Converting a full ORT result file.")
+
+            convertToTarget(result.analyzer?.result)?.let { analyzerResult ->
+                result.copy(
+                    analyzer = result.analyzer?.copy(result = analyzerResult)
+                )
+            }
+        }
+
+    /**
+     * Convert an [AnalyzerResult] file.
+     */
+    private fun convertAnalyzerResult(): Any? =
+        convertResult<AnalyzerResult> {
+            println("Converting an analyzer result file.")
+
+            convertToTarget(it)
+        }
+
+    /**
+     * Convert a [ProjectAnalyzerResult]. In this case, the result file contains an [AnalyzerResult], as a
+     * [ProjectAnalyzerResult] does not contain a dependency graph.
+     */
+    private fun convertProjectAnalyzerResult(): Any? =
+        convertResult<ProjectAnalyzerResult> { result ->
+            println("Converting a project analyzer result file.")
+
+            val analyzerResult = AnalyzerResultBuilder().addResult(result).build()
+            convertToTarget(analyzerResult)
+        }
+
+    /**
+     * Convert a result of type [T]. Read this result from the input file and invoke [block] with it. Return *false*
+     * if the result file could not be loaded (which probably means that it contains a result of a different type).
+     */
+    private inline fun <reified T : Any> convertResult(block: (T) -> Any?): Any? =
+        runCatching {
+            block(ortFile.readValue())
+        }.getOrNull()
+
+    /**
+     * Convert the given [analyzerResult] to the selected target format if it is present.
+     */
+    private fun convertToTarget(analyzerResult: AnalyzerResult?): AnalyzerResult? =
+        analyzerResult?.let(targetFormat::convert)
+
+    /**
+     * Write the result of the conversion to the configured output file. If no output file was specified, overwrite the
+     * input file.
+     */
+    private inline fun <reified T : Any> writeResult(result: T) {
+        val targetFile = outputFile ?: ortFile
+
+        println("Writing converted result to $targetFile.")
+
+        targetFile.writeValue(result)
+    }
+}


helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -72,6 +73,7 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
         }
 
         subcommands(
+            ConvertToDependencyGraphCommand(),
```
sschuberth, line 76: Is there still the plan to eventually also convert in the other direction, from graph to tree, e.g. to make result better comparable / diffable? I believe the outcome from our discussion was "no, we want to give diffing the graph format a try first", but I'm still wondering whether we should generalize the name of the command, like "ConvertDependencies" or "ConvertDependencyRepresentation", and then from the option(s) tell from which direction into which. Or simply auto-detect it, and always convert into the other format than what is in the input.

helper-cli/src/main/kotlin/commands/ConvertToDependencyGraphCommand.kt
```diff
@@ -0,0 +1,153 @@
...
+        val analyzerResult = AnalyzerResultBuilder().addResult(result).build()
+        writeResult(convertToGraph(analyzerResult))
+    }
+
+    /**
+     * Convert a result of type [T]. Read this result from the input file and invoke [block] with it. Return *false*
+     * if the result file could not be loaded (which probably means that it contains a result of a different type).
+     */
+    private inline fun <reified T : Any> convertResult(block: (T) -> Unit): Boolean =
+        try {
```
sschuberth, line 110: Maybe a bit nicer: `runCatching` / `onFailure` / `isSuccess`

helper-cli/src/main/kotlin/commands/ConvertToDependencyGraphCommand.kt
```diff
@@ -0,0 +1,153 @@
...
+            block(ortFile.readValue())
+            true
+        } catch (e: JsonMappingException) {
+            println("Could not convert result type ${T::class.simpleName}: ${e.message}")
```
sschuberth, line 114: Isn't it a bit confusing to see this message while "auto-detecting" the input file format? IMO we could drop this, and only keep the error message if no supported input file format could be detected.

helper-cli/src/main/kotlin/commands/ConvertToDependencyGraphCommand.kt
```diff
@@ -0,0 +1,153 @@
...
+    override fun run() {
+        if (!convertOrtResult() && !convertAnalyzerResult() && !convertProjectAnalyzerResult()) {
+            failure("No supported data type could be found in the input file.", 1)
+        }
+    }
+
+    /**
+     * Convert a full ORT result file.
+     */
+    private fun convertOrtResult(): Boolean = convertResult<OrtResult> { result ->
```
sschuberth, line 74: Formatting nit: Line break after `=`. Some below.

helper-cli/src/main/kotlin/commands/ConvertToDependencyGraphCommand.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    private val outputFile by option(
+        "--output-file", "-o",
+        help = "The file in which to write the result of the conversion. If missing, the input file is overwritten."
+    ).convert { it.expandTilde() }
+        .file(mustExist = false, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+
+    override fun run() {
+        if (!convertOrtResult() && !convertAnalyzerResult() && !convertProjectAnalyzerResult()) {
```
sschuberth, line 66: How about instead:

        val converters = listOf(::convertOrtResult, ::convertAnalyzerResult, ::convertProjectAnalyzerResult)
        if (converters.none { it() }) {

helper-cli/src/main/kotlin/commands/ConvertToDependencyGraphCommand.kt
```diff
@@ -0,0 +1,153 @@
...
+            false
+        }
+
+    /**
+     * Convert the given [analyzerResult] to the dependency graph format. Fail if no result is available or the
+     * result is already in the dependency graph format.
+     */
+    private fun convertToGraph(analyzerResult: AnalyzerResult?): AnalyzerResult {
+        val convertedResult =
+            analyzerResult?.let(DependencyGraphConverter::convert) ?: failure("No analyzer result available.", 1)
```
sschuberth, line 124: How about:

        val convertedResult = DependencyGraphConverter.convert(
            analyzerResult ?: failure("No analyzer result available.", 1)
        )

sschuberth (changes requested):

helper-cli/src/main/kotlin/HelperMain.kt
fviernau: Maybe consider also "SetDependencyRepresentationCommand".

helper-cli/src/main/kotlin/commands/ConvertToDependencyGraphCommand.kt
```diff
@@ -0,0 +1,153 @@
...
+        help = "The ORT result file to read as input. This can be the serialized form of an OrtResult, an " +
+                "AnalyzerResult, or a ProjectAnalyzerResult."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+
+    private val outputFile by option(
+        "--output-file", "-o",
+        help = "The file in which to write the result of the conversion. If missing, the input file is overwritten."
```
fviernau, line 60: Is "If missing, the input file is overwritten." really needed? Why did you decide to add this?

oheger-bosch: The tool is going to be used to convert the files with the test expectations to the new format. I thought, for this use case it is more convenient to do the modification in-place, rather than write the result to a temporary file, delete the original file, and eventually rename the output file.

helper-cli/src/main/kotlin/HelperMain.kt
oheger-bosch: I liked Frank's suggestion to renamed the command to `SetDependencyRepresentationCommand`.

Regarding the output format: I prefer explicitly specifying the target format of the conversion. If the command always did a conversion, you would lose the use case to ensure that a file has a specific format by just running the command on it. Therefore, I added an option to set the target format based on an enum. This is currently a bit moot, as there is only a single supported output format, but can be extended easily.

helper-cli/src/main/kotlin/commands/ConvertToDependencyGraphCommand.kt
oheger-bosch: Now using `runCatching`.

helper-cli/src/main/kotlin/commands/ConvertToDependencyGraphCommand.kt
oheger-bosch: The main reason for this message was `detekt` complaining about a swallowed exception. This no longer applies due to the `runCatching` construct. So the message has been dropped.

helper-cli/src/main/kotlin/HelperMain.kt
sschuberth: > If the command always did a conversion, you would lose the use case to ensure that a file has a specific format by just running the command on it.

Good point.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    private val outputFile by option(
+        "--output-file", "-o",
+        help = "The file in which to write the result of the conversion. If missing, the input file is overwritten."
+    ).convert { it.expandTilde() }
+        .file(mustExist = false, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+
+    private val targetFormat by option(
+        "--target", "-t",
```
sschuberth, line 83: Why not `"--format", "-f"`, as "target" could be misunderstood as the output file?

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,176 @@
...
+        help = "The target format for the conversion. Allowed values: GRAPH."
```
sschuberth, line 84: We don't usually hard-code allowed values here, but leave it to *clikt* to print the supported values if a wrong one is given.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,176 @@
...
+    ).enum<TargetFormat>().default(TargetFormat.GRAPH)
+
+    override fun run() {
+        val converters = listOf(this::convertOrtResult, this::convertAnalyzerResult, this::convertProjectAnalyzerResult)
```
sschuberth, line 88: Please drop `this`.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+    /**
+     * Convert a result of type [T]. Read this result from the input file and invoke [block] with it. Return *false*
+     * if the result file could not be loaded (which probably means that it contains a result of a different type).
+     */
+    private inline fun <reified T : Any> convertResult(block: (T) -> Unit): Boolean =
+        runCatching {
+            block(ortFile.readValue())
+        }.fold(onSuccess = { true }, onFailure = { false })
```
sschuberth, line 138: I believe `.isSuccess` is still more readable while being shorter.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,176 @@
...
+     * Write the result of the conversion to the configured output file. If no output file was specified, overwrite the
+     * input file.
+     */
+    private inline fun <reified T : Any> writeResult(result: T): Boolean {
+        val targetFile = outputFile ?: ortFile
+
+        println("Writing converted result to $targetFile.")
+
+        targetFile.writeValue(result)
+        return true
```
sschuberth, line 166: This bogus return value can be dropped.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /**
+     * Exit the program with the given failure [message] and status [code].
+     */
+    private fun failure(message: String, code: Int): Nothing {
+        println(message)
+        throw ProgramResult(code)
+    }
```
sschuberth, line 175: I believe instead of this function we could simply throw a `UsageError` as all current causes are actually user errors.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,176 @@
...
+     * result is already in the target format.
+     */
+    private fun convertToTarget(analyzerResult: AnalyzerResult?): AnalyzerResult {
+        val convertedResult = targetFormat.convert(
+            analyzerResult ?: failure("No analyzer result available.", 1)
+        )
+
+        if (convertedResult == analyzerResult) {
+            failure("The result already uses the dependency graph format. No conversion is done.", 2)
+        }
```
sschuberth, line 151: In [this comment](https://github.com/oss-review-toolkit/ort/pull/4191#discussion_r655095130) you argued that a user should be able to run the comment to ensure a specific output format, independent of the input format. With that in mind, it IMO should be no failure if the input already is in the desired format. Also, the user might have specified a different output file, effective asking for a copy of the input in that case.

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
oheger-bosch: Ok, saw this in another command, but this was probably a bad example.

sschuberth: Which one? I also did a quick check, and did not find any such code.

oheger-bosch: It was `ListLicensesCommand`.

sschuberth: I'd indeed consider that to be a bad example. I'll file a separate PR to fix this.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
oheger-bosch: I wanted the command to return a different exit code when no conversion was done. Currently, exit code 1 is used for real errors and 2 if no conversion was done. This is a bit similar to the handling of issues and hints in ORT's main commands.

Is it desired to have such different exit codes, and - if yes -, what would be the best way to achieve it?

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
sschuberth: See https://github.com/oss-review-toolkit/ort/pull/4194.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
sschuberth: IMO it would not be *required* to have a different exit code in this case, but I can see how it makes sense. As I was mostly disturbed by the semantics of the word "failure", I guess my issue would go away if we'd just throw a `UsageError` directly instead.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
oheger-bosch: Done

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
oheger-bosch: The function has been replaced by throwing `UsageError` directly.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,170 @@
...
+        help = "The ORT result file to read as input. This can be the serialized form of an OrtResult, an " +
+                "AnalyzerResult, or a ProjectAnalyzerResult."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+
+    private val outputFile by option(
+        "--output-file", "-o",
+        help = "The file in which to write the result of the conversion. If missing, the input file is overwritten."
```
fviernau, line 77: Would it make sense to mention the following for the input parameter to avoid surprises? "If missing, the input file is overwritten."

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,170 @@
...
+class SetDependencyRepresentationCommand : CliktCommand(
+    help = "Convert an ORT result from the dependency tree representation to the dependency graph format."
+) {
+    /**
+     * An enumeration class representing the different supported formats to represent dependencies in an ORT result.
+     * It is used to define the target format that should be set by this command.
+     */
+    enum class TargetFormat {
+        /** The dependency graph format. */
+        GRAPH {
```
fviernau, line 55: just out of interrest: Are there plans to implement conversion from `Graph` to `Scope`?

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,170 @@
...
+            block(ortFile.readValue())
+        }.isSuccess
+
+    /**
+     * Convert the given [analyzerResult] to the selected target format. Fail if no result is available or the
+     * result is already in the target format.
+     */
+    private fun convertToTarget(analyzerResult: AnalyzerResult?): AnalyzerResult {
+        val convertedResult = targetFormat.convert(
+            analyzerResult ?: throw UsageError("No analyzer result available.")
```
fviernau, line 146: I believe the command is easier to use if you change semantics to let's say "the command converts all analyzer results contained in the ORT result to the specified format". This way the caller does not need to check first whether the ORT result contains an analyzer result. This makes it easier to use from a script.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,170 @@
...
+        )
+
+        if (convertedResult == analyzerResult) {
+            throw UsageError(
+                "The result already uses the dependency graph format. No conversion is done.",
+                statusCode = 2
+            )
+        }
```
fviernau, line 154: I'd also prefer to drop this error for same reason as above.

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
fviernau: e.g. when the command exists successfully, it is ensured that the target file has specified format, no matter what format the input file used. I cannot see any use cases which satisfy the added complexity for these errors.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
oheger-bosch: Not concrete yet. What we have discussed was mainly to see the diffs for failing tests in the more readable tree format; but this could be achieved via a custom matcher. However, the conversion to the classic format is already implemented, so if there is a concrete use case, why not.

fviernau: I believe debugging could be a use case, e.g. making it more readable to easier grasp it.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
oheger-bosch: I agree to the approach to make the command easier to use in scripts. So I changed (or basically removed) the error handling. It is now only visible from the command's output whether a conversion was done or not.

The only thing I am not sure is whether an error code should be returned if the input file could not be processed at all. (Currently it is not, but this would be rather easy to change.)

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
oheger-bosch: No error is thrown any more.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
fviernau: If execution is not successful because the input file does not exist or cannot be deserialized IMO a non zero exit code would be fine.

fviernau (approved):

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * A command to convert the dependency representation in an ORT result file between the different supported formats.
+ *
+ * The input file can contain the serialized form of either a full [OrtResult], an [AnalyzerResult], or a single
+ * [ProjectAnalyzerResult]. If one of the supported results is found, it is converted to the specified target format.
+ * If this target format is already used, then this command does nothing.
+ */
+class SetDependencyRepresentationCommand : CliktCommand(
+    help = "Convert the dependency representation of an ORT result to a specific target format."
```
sschuberth, line 48: The help text should now probably also say "Set" instead of "Convert", or?

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -0,0 +1,157 @@
...
+        .convert { it.absoluteFile.normalize() }
+
+    private val targetFormat by option(
+        "--format", "-f",
+        help = "The target format for the conversion."
+    ).enum<TargetFormat>().default(TargetFormat.GRAPH)
+
+    override fun run() {
+        val converters = listOf(::convertOrtResult, ::convertAnalyzerResult, ::convertProjectAnalyzerResult)
+        converters.asSequence().mapNotNull { it() }.firstOrNull()?.let { writeResult(it) }
```
sschuberth, line 91: If we view `converters` as a sequence, we could as well right away use `sequenceOf()` instead of `listOf()`.

sschuberth (changes requested):

sschuberth (approved):

sschuberth (approved):

PR #4198, opened 2021-06-22
cli: Introduce a default value for the `--package-curations-dir` option

fviernau:


utils/src/main/kotlin/Constants.kt
```diff
@@ -61,9 +61,14 @@ const val ORT_COPYRIGHT_GARBAGE_FILENAME = "copyright-garbage.yml"
 const val ORT_CUSTOM_LICENSE_TEXTS_DIRNAME = "custom-license-texts"
 
 /**
- * The name of the ORT curations configuration file.
+ * The name of the ORT package curations directory.
  */
-const val ORT_CURATIONS_FILENAME = "curations.yml"
+const val ORT_PACKAGE_CURATIONS_DIRNAME = "curations.yml"
```
sschuberth, line 66: I guess the *directory* name should not end in `.yml`, or?

fviernau: Now I'm considering naming it `package-curations` and renaming `curations.yml` to `package-curations.yml`. What do you think?

MarcelBochtler (approved):

utils/src/main/kotlin/Constants.kt
MarcelBochtler: This would be a breaking change and would cause us to do some work first.

sschuberth: > Now I'm considering naming it `package-curations` and renaming `curations.yml` to `package-curations.yml`. What do you think? (I'd do that in a separate PR though)

On the one hand I like the idea, as there were continuous confusions on our side whether curations also apply to *projects* or not. Such a renaming would make that a bit more clear, and it would align with "package configurations".

On the other hand, like Marcel mentioned it'd be a breaking change, and I believe we all got very used to just calling them "curations" instead of "package curations". But in total, I'm slightly in favor of the change.

fviernau: Thanks @sschuberth @MarcelBochtler . I'll just make a separate PR for that, then we can have a vote there.

PR #4199, opened 2021-06-22
helper-cli: Add a command to set the package curations for a ORT file

fviernau:
One use case for this is removing organizatinon internal (curation)
data from a given ORT file prior to sharing it externally.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationsCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationsCommand.kt
@@ -28,6 +28,7 @@ internal class PackageCurationsCommand : CliktCommand(
     init {
         subcommands(
             CreateCommand(),
+            SetCommand(),
             SplitCommand()
         )
     }
--- a/helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2021 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands.packagecuration
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.file
+
+import org.ossreviewtoolkit.analyzer.curation.FilePackageCurationProvider
+import org.ossreviewtoolkit.helper.common.writeOrtResult
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+class SetCommand : CliktCommand(
+    help = "(Re-)set all package curations for a given ORT file to the curations specified via package curations " +
+            "file and directory. If no curations are given then all curations get removed."
+) {
+    private val ortFile by option(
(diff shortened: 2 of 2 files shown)


helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationsCommand.kt
```diff
@@ -28,6 +28,7 @@ internal class PackageCurationsCommand : CliktCommand(
     init {
         subcommands(
             CreateCommand(),
+            SetCommand(),
```
MarcelBochtler, line 31: Commit Message: `organizatinon` --> `organisation`

helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.ossreviewtoolkit.helper.common.writeOrtResult
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+class SetCommand : CliktCommand(
+    help = "(Re-)set all package curations for a given ORT file."
+) {
+    private val ortFile by option(
+        "--ort-file",
```
MarcelBochtler, line 38: Please also add the short `-i` option as it is done in other places for the `--ort-file` option.

helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt
```diff
@@ -0,0 +1,66 @@
...
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+
+    override fun run() {
+        val curations = FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations
+
+        val ortResult = ortFile.readValue<OrtResult>().replacePackageCurations(curations)
+
+        writeOrtResult(ortResult, ortFile)
```
MarcelBochtler, line 64: I prefer that every command should have at least some output. Otherwise the user wouldn't know if something happened.
What do you think about something like `println("Successfully set curations for ${ortResult.absolutePath}."`?

MarcelBochtler (changes requested):

helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt
fviernau: I wouldn't like to add that requriement to all the commands as I believe it's not needed.
E.g. the commands `cp` or `mv` also don't print that they successfully copied something by default.
I prefer not to add a comment to keep things minimal and safe time discussing it.

MarcelBochtler: > E.g. the commands cp or mv also don't print that they successfully copied something by default.

That's why I aliased `cp` to `cp -v` :)
I'd still prefer feedback from the commands, but I get your argumentation and it's okay for me to not add it in the `helper-cli`.

helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt
```diff
@@ -0,0 +1,66 @@
...
+        "--package-curations-file",
+        help = "A file containing package curation data."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+
+    override fun run() {
+        val curations = FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations
+
+        val ortResult = ortFile.readValue<OrtResult>().replacePackageCurations(curations)
```
MarcelBochtler, line 62: If neither `--package-curations-dir` nor `--package-curations-file` is set, will all curations be removed? Or will nothing be done?
Can this be made clear in the documentation and maybe provide some CLI feedback? I want to avoid printing nothing in the success case but also in the case when nothing was being changed.

fviernau: I've updated the `help` text.

helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.utils.expandTilde
+
+class SetCommand : CliktCommand(
+    help = "(Re-)set all package curations for a given ORT file to the curations specified via package curations " +
+            "file and directory. If no curations are given then all curations get removed."
+) {
+    private val ortFile by option(
+        "--ort-file, -i",
```
MarcelBochtler, line 39: This should be `"--input-file", "-i",`.

MarcelBochtler (changes requested):

MarcelBochtler (approved):

PR #4201, opened 2021-06-22
SplitCommand: Append text instead overwriting existing files

MarcelBochtler:


sschuberth (dismissed):

fviernau (approved):

helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
```diff
@@ -82,7 +82,12 @@ internal class SplitCommand : CliktCommand(
                 val text = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(listOf(curation))
 
                 outputFile.parentFile.safeMkdirs()
-                outputFile.writeText(text)
+
+                if (outputFile.exists()) {
```
sschuberth, line 86: Even if not strictly required here, please use `isFile` instead of `exists()`  which is good style to also verify the "thing" is of the expected type.

helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
```diff
@@ -82,7 +82,12 @@ internal class SplitCommand : CliktCommand(
...
+                    outputFile.appendText("\n$text")
+                } else {
+                    outputFile.appendText(text)
```
sschuberth, line 89: This can and should be `writeText()` now again.

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
```diff
@@ -82,7 +82,12 @@ internal class SplitCommand : CliktCommand(
                 val text = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(listOf(curation))
 
                 outputFile.parentFile.safeMkdirs()
-                outputFile.writeText(text)
+
+                if (outputFile.exists()) {
+                    outputFile.appendText("\n$text")
+                } else {
```
fviernau, line 88: Thinking about this change again, I propose to refactor it to the following approach
- in the outer `for-loop` deserialize the existing file
- in the outer `for-loop` merge the de-serialized curations with the `curations` variable content
- in the outer `for-loop` write all curations at once with a single `writeText()` call

So the inner loop gets eliminated. This approach
- allows easily adding sorting of curations
- allows easily adding duplicate removal
- It properly handles the case where target files exist already prior to running the command. E.g. if they have unexpected format the command fails. 
- remove unnecessary case distinctions
- puts the mkDirs call in the outer loop

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
sschuberth: Sounds good to me, but I'd be fine to do this in a follow-up PR.

sschuberth (approved):

fviernau (approved):

PR #4202, opened 2021-06-22
ScanSummary: Remove fileCount

### LINKED ISSUE #4197: Remove fileCount from ScanSummary
In an ORT run with an SPDX dependency with a VCS info that referred to only one file we encountered the problem that when using ScanCode and that file contains no license or copyright information, the fileCount is set to 0, even though a file gets scanned. See here:

https://github.com/oss-review-toolkit/ort/blob/4615824c8aaa29fa276eaf01fd6c34d486e08976/model/src/main/kotlin/ScanSummary.kt#L100-L106

When the file count is 0, the scan result is considered to have failed andthis log output is printed:

https://github.com/oss-review-toolkit/ort/blob/4615824c8aaa29fa276eaf01fd6c34d486e08976/scanner/src/main/kotlin/ScanResultsStorage.kt#L276-L280

When checking the code, I could not see another place than the above check in `ScanSummary` where the `fileCount` was used other than reaching it through to be printed in e.g. the WebAppReporter. 

I propose to delete the variable all together, since it doesn't seem to add valuable functionality. This also removes the necessity to re-scan empty repositories or dependencies with just one file but no license or copyright findings.

neubs-bsi:
FileCount does not seem to add any important functionality and is
simply reached through until output.
The only check that is done on fileCount additionally leads to a wrong failure
when only one file was scanned without licenseFindings and copyrightFindings.

Thus, removing the variable is a reasonable choice.

Resolves #4197.


scanner/src/test/kotlin/scanners/scancode/ScanCodeResultParserTest.kt
```diff
@@ -41,28 +41,6 @@ import org.ossreviewtoolkit.spdx.SpdxConstants
 
 @Suppress("LargeClass")
 class ScanCodeResultParserTest : WordSpec({
-    "ScanCode 2 results" should {
```
sschuberth, line 44: I believe we shouldn't remove these completely, but simply test other properties of a ScanSummary.

sschuberth: Done for you in https://github.com/oss-review-toolkit/ort/pull/4233, so you just need to rebase after that got merged.

scanner/src/test/kotlin/storages/CompositeStorageTest.kt
```diff
@@ -82,8 +82,9 @@ private fun licenseFinding(index: Int): LicenseFinding =
 private fun createScanResult(resultCount: Int): ScanResult {
     val licenseFindings = List(resultCount) { licenseFinding(it) }
     val summary = ScanSummary(
-        startTime = Instant.now(), endTime = Instant.now(), fileCount = resultCount,
-        packageVerificationCode = "test$resultCount", licenseFindings = licenseFindings.toSortedSet(),
+        startTime = Instant.now(), endTime = Instant.now(),
+        packageVerificationCode = "test$resultCount",
+        licenseFindings = licenseFindings.toSortedSet(),
```
sschuberth, line 87: I'm fine with reformatting here, but then we should consistently use one argument per line.

.azure-pipelines/LinuxAnalyzerTest.yml
```diff
@@ -63,9 +63,9 @@ jobs:
   # Gradle build cache, see: https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops
   - task: Cache@2
     inputs:
-      key: '"$(Agent.OS)" | analyzer-test | gradle-caches | v3 | gradle.properties, settings.gradle.kts, **/build.gradle.kts'
+      key: '"$(Agent.OS)" | analyzer-test | gradle-caches | v2 | gradle.properties, settings.gradle.kts, **/build.gradle.kts'
```
sschuberth, line 66: I'd hope that we can drop this commit after another rebase.

sschuberth (changes requested):

.azure-pipelines/LinuxAnalyzerTest.yml
neubs-bsi: I checked the merged PRs from the last days and none seemed to specifically make any changes here. Is there a reason why you think this can be removed? (I'd be more than happy to)

sschuberth: I simply think this is just a band aim which *might* help *sometimes*, but it does not fix the root cause. We already did bump the cache version in the past to work around the issue, and while it seemed to have fixed the issue, it apparently did not (as you're still seeing the issue). So instead of bumping the cache version over and over again, I believe we can as well merge despite the test failure and try to address the root cause in parallel.

neubs-bsi: I dropped the commit for now. I can always add it again, if it doesn't work :)

neubs-bsi:
Currently pipeline breaks, due to the known error of the cache problem, for which I removed the additional commit, as was suggested [here](https://github.com/oss-review-toolkit/ort/pull/4202#discussion_r661172507)

sschuberth:
> Currently pipeline breaks, due to the known error of the cache problem, for which I removed the additional commit, as was suggested [here](https://github.com/oss-review-toolkit/ort/pull/4202#discussion_r661172507)

Instead of that dropped commit, I tried https://github.com/oss-review-toolkit/ort/pull/4240, and will retrigger CI checks for this PR eventually.

sschuberth:
@neubs-bsi, as retriggering CI did not work, but other PRs passed recently, can you please try to rebase?

sschuberth:
@mnonnenmacher could you please also review despite the failing tests, to check if the substitute for the *FileCounterScannerFunTest* is like you imagined?

scanner/src/main/kotlin/LocalScanner.kt
```diff
@@ -508,7 +508,7 @@ abstract class LocalScanner(
     /**
      * Return the scanner's raw result in a JSON representation.
      */
-    internal abstract fun getRawResult(resultsFile: File): JsonNode
+    abstract fun getRawResult(resultsFile: File): JsonNode
```
mnonnenmacher, line 511: Do you have to remove the `internal` here?

mnonnenmacher (changes requested):

mnonnenmacher:
> @mnonnenmacher could you please also review despite the failing tests, to check if the substitute for the _FileCounterScannerFunTest_ is like you imagined?

Yes, looks good to me.

scanner/src/main/kotlin/LocalScanner.kt
sschuberth: Yes, unfortunately `internal` doe snot work with our custom `funTest` source set, only with the built-in `test` source set. Also see https://github.com/oss-review-toolkit/ort/blob/1fa1d749d65ee9a17dc3214d2abcb26aaf6978f8/build.gradle.kts#L174-L182

mnonnenmacher:
@sschuberth @neubs-bsi I was wondering why the reporter-web-app build is still failing in this PR. I noticed that it should never be built for the analyzer tests, so I have created: https://github.com/oss-review-toolkit/ort/pull/4253
But this will not help with the "LinuxTest" task. I noticed that this inotify errors appears before the task fails, do you think this could be related?
```
> Task :reporter-web-app:yarnInstall FROM-CACHE
Unable to watch the file system for changes. The inotify watches limit is too low.
Stopping file watching and invalidating VFS after an error happened
```

Maybe it's worth trying to [increase the inotify limit](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:inotify_watches_limit) or to run the CI tasks with [`--no-watch-fs`](https://docs.gradle.org/current/userguide/gradle_daemon.html#enabling_and_disabling_file_system_watching) because I think for CI builds it does not make sense to watch the filesystem anyway, right?

mnonnenmacher (approved):

sschuberth:
> I noticed that this inotify error appears before the task fails, do you think this could be related?

Good finding, could indeed be related (although if so, I'd regard that as a bug in Gradle).

> run the CI tasks with [`--no-watch-fs`](https://docs.gradle.org/current/userguide/gradle_daemon.html#enabling_and_disabling_file_system_watching) because I think for CI builds it does not make sense to watch the filesystem anyway, right?

I agree, we should pass `--no-watch-fs` on CI.

sschuberth:
Merging despite the unrelated build failure in `yarnBuild`.

sschuberth:
> I agree, we should pass `--no-watch-fs` on CI. Will you make that change?

Done as part of https://github.com/oss-review-toolkit/ort/pull/4256.

PR #4204, opened 2021-06-22
Docker/Azure: Avoid cloning the CocoaPods Git repository

fviernau:
Download the archive instead of cloning the repository to reduce
execution time.

### DIFF
--- a/.azure-pipelines/LinuxAnalyzerTest.yml
+++ b/.azure-pipelines/LinuxAnalyzerTest.yml
@@ -43,10 +43,9 @@ jobs:
       mkdir -p $HOME/go/bin
       curl https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | sh
 
-      # Install 'CocoaPods'. As https://github.com/CocoaPods/CocoaPods/pull/10609 is needed but not yet released, make a
-      # build from the latest revision of the master branch.
-      git clone https://github.com/CocoaPods/CocoaPods.git
-      cd CocoaPods
+      # Install 'CocoaPods'. As https://github.com/CocoaPods/CocoaPods/pull/10609 is needed but not yet released.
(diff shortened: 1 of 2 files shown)


.azure-pipelines/LinuxAnalyzerTest.yml
```diff
@@ -43,10 +43,10 @@ jobs:
       mkdir -p $HOME/go/bin
       curl https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | sh
 
-      # Install 'CocoaPods'. As https://github.com/CocoaPods/CocoaPods/pull/10609 is needed but not yet released, make a
-      # build from the latest revision of the master branch.
-      git clone https://github.com/CocoaPods/CocoaPods.git
-      cd CocoaPods
+      # Install 'CocoaPods'. As https://github.com/CocoaPods/CocoaPods/pull/10609 is needed but not yet released.
+      curl -ksSL -o cocoapods.zip https://github.com/CocoaPods/CocoaPods/archive/9461b346aeb8cba6df71fd4e71661688138ec21b.zip
```
sschuberth, line 47: You could use `-O` (uppercase) instead of `-o cocoapods.zip`. Or yet simpler, pipe `curl`'s output directly to `unzip`. That way you also don't have to manually remove the ZIP file, see the similar code below.

sschuberth (changes requested):

.azure-pipelines/LinuxAnalyzerTest.yml
fviernau: Why should `-O` instead of `-o` be used?

The code below pipes to `tar`, not to `unzip` and I was not able to figure out how it works with `unzip`.
Can you give a hint?

sschuberth: These lines still use the previous approach.

sschuberth (approved):

sschuberth (approved):

PR #4206, opened 2021-06-23
Npm: Omit blank declared licenses 

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -106,7 +106,7 @@ open class Npm(
          * Parse information about licenses from the [package.json][json] file of a module.
          */
         internal fun parseLicenses(json: JsonNode): SortedSet<String> {
-            val declaredLicenses = sortedSetOf<String>()
+            val declaredLicenses = mutableListOf<String>()
 
             // See https://docs.npmjs.com/files/package.json#license. Some old packages use a "license" (singular) node
             // which ...
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -127,15 +127,17 @@ open class Npm(
                 licenseNode["type"]?.textValue()
             }
 
-            return declaredLicenses.mapTo(sortedSetOf()) { declaredLicense ->
+            return declaredLicenses.mapNotNullTo(sortedSetOf()) { declaredLicense ->
                 when {
                     // NPM does not mean https://unlicense.org/ here, but the wish to not "grant others the right to use
                     // a private or unpublished package under any terms", which corresponds to SPDX's "NONE".
                     declaredLicense == "UNLICENSED" -> SpdxConstants.NONE
+
                     // NPM allows to declare non-SPDX licenses only by referencing a license file. Avoid reporting an
                     // [OrtIssue] by mapping this to a valid license identifier.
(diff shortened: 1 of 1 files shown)


neubs-bsi:
Could you add a test case?

sschuberth:
> Could you add a test case?

I wonder whether instead of a (package manager specific) test it would make more sense to add a `require` in an `init` block for `Project` and `Package`, so we become aware of the issue for *all* package managers. Opinions @fviernau, @mnonnenmacher?

mnonnenmacher:
> > Could you add a test case?
> 
> I wonder whether instead of a (package manager specific) test it would make more sense to add a `require` in an `init` block for `Project` and `Package`, so we become aware of the issue for _all_ package managers. Opinions @fviernau, @mnonnenmacher?

I'd rather have a test for each package manager and proactively fix the issue instead of waiting for it to fail at runtime.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -106,7 +106,7 @@ open class Npm(
          * Parse information about licenses from the [package.json][json] file of a module.
          */
         internal fun parseLicenses(json: JsonNode): SortedSet<String> {
-            val declaredLicenses = sortedSetOf<String>()
+            val declaredLicenses = mutableListOf<String>()
```
oheger-bosch, line 109: Instinctively, I would have used `mutableSetOf` to get rid off duplicates directly. But this won't make a bit difference.

oheger-bosch (commented):
Could the filtering of blank licenses be done more centrally, so that not all package manager implementations would have to deal with this? Theoretically, there could be a similar problem with author names, too. So some functionality to sanitize these properties would be nice.

Placing this logic in the constructors of `Project` and `Package` is probably not ideal, as these are plain data classes. Maybe some checks or filtering could be done by the classes used to build analyzer results, e.g. `ProjectAnalyzerResult` or `AnalyzerResultBuilder`?

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: I went for a list as we can spare the implicit deduplication here as we do it in the final mapping again. And maybe for debugging it might even be valuable to see where duplicate license declarations occur.

sschuberth:
> Could the filtering of blank licenses be done more centrally, so that not all package manager implementations would have to deal with this?

Unfortunately not (yet), IMO, because we have no defined API that package managers need to implement in order to emit projects / packages. This actually does a bit into the direction of your https://github.com/oss-review-toolkit/ort/pull/4166, except that it would be an API to *create* the dependency graph / packages / projects, instead of consuming it.

> Maybe some checks or filtering could be done by the classes used to build analyzer results, e.g. `ProjectAnalyzerResult` or `AnalyzerResultBuilder`?

Any such late checks are problematic IMO as they likely involve copying / modifying of originally immutable collections. But `ProjectAnalyzerResult` actually already has some sanity checks in `init`, so we could still place some more checks there.

sschuberth:
> I'd rather have a test for each package manager and proactively fix the issue instead of waiting for it to fail at runtime.

At least such a runtime failure would not be as bad as it sounds, as `PackageManager` catches any `Exception` and wraps it into issues.

oheger-bosch (approved):
These changes are certainly an improvement over the current state. We should, however, be aware that there might still be problems with other package manager implementations and other properties which might be blank as well.

sschuberth:
I'll add a test later.

Issue #4096, opened 2021-06-01
Align on what counts as an issue that needs to be resolved


sschuberth:
The CLI tools currently return inconsistent exit codes regarding issues.

The analyzer and advisor return non-0 if there are issues of *any* severity:

https://github.com/oss-review-toolkit/ort/blob/768a585b1ccf72cbd952fe7f95216a15d11c947c/cli/src/main/kotlin/commands/AnalyzerCommand.kt#L178-L181

https://github.com/oss-review-toolkit/ort/blob/768a585b1ccf72cbd952fe7f95216a15d11c947c/cli/src/main/kotlin/commands/AdvisorCommand.kt#L126-L129

The scanner and evaluator return non-0 if there are issues of a severity *higher than hint*:

https://github.com/oss-review-toolkit/ort/blob/768a585b1ccf72cbd952fe7f95216a15d11c947c/cli/src/main/kotlin/commands/ScannerCommand.kt#L183-L190

https://github.com/oss-review-toolkit/ort/blob/768a585b1ccf72cbd952fe7f95216a15d11c947c/cli/src/main/kotlin/commands/EvaluatorCommand.kt#L279-L286

Finally, the web app report displays unresolved / resolved issue counts in its "Violations" and "Issues" tabs, but what counts as "unresolved" is only issues of a severity *higher than warning*:

![image](https://user-images.githubusercontent.com/349154/120305020-95f85c80-c2d0-11eb-95c4-6b6569381a80.png)

As probably no consensus can be reached about whether hints / warnings *must* be addressed in general (as it's a workflow specific question), the only way forward might be to make this configurable and make all tools adhere to the configuration.

sschuberth:
@tsteenbe would you be able to help with resolving the final bit, and make the web app report adhere to the new `severeIssueThreshold` configuration in its issue counts in the tabs?

sschuberth:
Closing this in favor of tracking the remaining work at #4453.

PR #4207, opened 2021-06-23
reporter: Make the TemplateHelper adhere to severeIssueThreshold by default

sschuberth:
This is part of resolving #4096. Note that this implicitly changes the
default value used for hasUnresolvedIssues() and
hasUnresolvedRuleViolations() from Severity.HINT to Severity.WARNING.

### DIFF
--- a/reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
+++ b/reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
@@ -41,14 +41,12 @@ import org.ossreviewtoolkit.model.Vulnerability
 import org.ossreviewtoolkit.model.VulnerabilityReference
 import org.ossreviewtoolkit.model.config.VulnerabilityResolution
 import org.ossreviewtoolkit.model.licenses.DefaultLicenseInfoProvider
-import org.ossreviewtoolkit.model.licenses.LicenseClassifications
 import org.ossreviewtoolkit.model.licenses.LicenseInfoResolver
 import org.ossreviewtoolkit.model.licenses.LicenseView
 import org.ossreviewtoolkit.model.licenses.ResolvedLicense
 import org.ossreviewtoolkit.model.licenses.ResolvedLicenseFileInfo
 import org.ossreviewtoolkit.model.licenses.ResolvedLicenseInfo
 import org.ossreviewtoolkit.model.licenses.filterExcluded
-import org.ossreviewtoolkit.model.utils.ResolutionProvider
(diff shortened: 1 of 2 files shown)


reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
```diff
@@ -122,7 +120,7 @@ class FreemarkerTemplateProcessor(
             "ortResult" to input.ortResult,
             "licenseTextProvider" to input.licenseTextProvider,
             "LicenseView" to LicenseView,
-            "helper" to TemplateHelper(input.ortResult, input.licenseClassifications, input.resolutionProvider),
+            "helper" to TemplateHelper(input),
```
fviernau, line 123: This change changes the default value from `hint` to `warning`. Is it intended?

fviernau (changes requested):

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
sschuberth: Yes. That's why I marked this as a breaking change. But AFAIK that's what HERE wants (i.e. you don't require hints to be addressed), and it's also what we decided to settle on now, too.

fviernau: How about mentioning that default value change in the commit message?

sschuberth: Makes sense, done.

fviernau: > But AFAIK that's what HERE wants (i.e. you don't require hints to be addressed), and it's also what we decided to settle on now, too.

Almost, we also don't require warnings to be addressed, but that's ok as it is configurable.

sschuberth: > Almost, we also don't require warnings to be addressed, but that's ok as it is configurable.

Do you need to make a configuration change then? If so, please tell me (or only approve by then) when you're good to merge this.

fviernau (approved):

reporter/src/main/kotlin/utils/FreemarkerTemplateProcessor.kt
fviernau: You can merge it now. Thanks for asking!

PR #4209, opened 2021-06-23
Adapt more clients to the dependency navigator API

oheger-bosch:
This is a follow-up PR to #4166. It addresses further consumers of dependency information in an `OrtResult`.

### DIFF
--- a/evaluator/src/main/kotlin/RuleSet.kt
+++ b/evaluator/src/main/kotlin/RuleSet.kt
@@ -19,12 +19,11 @@
 
 package org.ossreviewtoolkit.evaluator
 
+import org.ossreviewtoolkit.model.DependencyNode
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Package
-import org.ossreviewtoolkit.model.PackageReference
 import org.ossreviewtoolkit.model.Project
 import org.ossreviewtoolkit.model.RuleViolation
-import org.ossreviewtoolkit.model.Scope
 import org.ossreviewtoolkit.model.licenses.LicenseInfoResolver
 import org.ossreviewtoolkit.model.utils.createLicenseInfoResolver
 import org.ossreviewtoolkit.utils.log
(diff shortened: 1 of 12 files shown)


evaluator/src/main/kotlin/RuleSet.kt
```diff
@@ -60,31 +60,31 @@ class RuleSet(
...
             level: Int,
-            scope: Scope,
+            scopeName: String,
             project: Project,
-            visitedPackages: MutableSet<PackageReference>
+            visitedPackages: MutableSet<Identifier>
         ) {
-            if (visitedPackages.contains(pkgRef)) {
-                log.debug { "Skipping rule $name for already visited dependency ${pkgRef.id.toCoordinates()}." }
+            if (visitedPackages.contains(node.id)) {
```
oheger-bosch, line 76: @mnonnenmacher could you please double-check this change?

In the original implementation, a single package could have been visited multiple times when it was part of different sub trees. This is no longer the case for the new implementation that only checks for package IDs.

Was it actually the original intension to distinguish between different sub trees and are there rules that take this information into account? Otherwise, the comparison of sub trees is not that trivial to implement for the dependency graph. (And the existing implementation is probably quite costly for large dependency structures, as always tree comparisons are made.)

evaluator/src/test/kotlin/RuleSetTest.kt
```diff
@@ -155,6 +158,38 @@ class RuleSetTest : WordSpec() {
...
+                            -isStaticallyLinked()
+                        }
+
+                        if (ancestors.any { !it.id.name.endsWith("-ref") }) {
+                            error("Node is not a reference.", howToFix)
+                        }
+                    }
+                }
+
+                ruleSet.violations should haveSize(0)
```
MarcelBochtler, line 191: Nit: I'd prefer `beEmpty()` to `haveSize(0)`.

MarcelBochtler (approved):
The changes look good to me.
Approving it but I'd wait for a response by @mnonnenmacher to @oheger-bosch's note above.

model/src/test/kotlin/config/ExcludesTest.kt
```diff
@@ -482,21 +482,21 @@ class ExcludesTest : WordSpec() {
             "return true if the scope is excluded" {
                 val excludes = Excludes(scopes = listOf(scopeExclude1))
 
-                excludes.isScopeExcluded(scope1) shouldBe true
+                excludes.isScopeExcluded(scope1.name) shouldBe true
```
nnobelis, line 485: Is it intended that this line is exactly the same as the next ?

Same comment for the changes underneath.

oheger-bosch: This must somehow be crept in during merging. Thanks for spotting, fixed.

evaluator/src/main/kotlin/RuleSet.kt
mnonnenmacher: The purpose of the dependency rule is indeed to visit each occurrence of a package in the dependency tree. This is for rules which require access to linkage, tree level or ancestors because these are not available in the `PackageRule`. Therefore calling the rule only for one occurrence could break existing rule sets and contradict the idea of the `DependencyRule`.

> Otherwise, the comparison of sub trees is not that trivial to implement for the dependency graph.

Why is that, is there no way to simply iterate over the whole dependency tree?

oheger-bosch: Thanks for clarifying. I think, I have an idea now how this can be achieved with the dependency graph.

oheger-bosch: I changed the implementation of `RuleSet` back, so that the node object is compared again, and not only its identifier. The comment of `DependencyNode.getStableReference()` now mentions that a meaningful implementation of `equals()` and `hashCode()` must be present. For `PackageReference` objects, this is already the case; for the data structures of the dependency graph, this can be achieved as well.

Unrelated to this change, I am a bit worried if the API of `DependencyRule` is clear in this regard. Are rule authors aware that the rule can be invoked multiple times for the same package? A typical implementation will probably do checks on the current package and might produce issues. If invoked multiple times, these issues will get duplicated. There is not really an easy way to prevent this, or do I miss something?

sschuberth: > Unrelated to this change, I am a bit worried if the API of `DependencyRule` is clear in this regard. Are rule authors aware that the rule can be invoked multiple times for the same package?

On a related note, that's one of the reasons why I had a long-standing desire to rename `DependencyRule` to `PackageReferenceRule`. That makes more clear that *references* to the (same) packages are handled here, and until the recent `DependencyGraph` changes, "dependency" was not a term that resembled a data structure in ORT.

mnonnenmacher: @oheger-bosch I read the code again and realized what you meant by many tree comparisons. Actually, what I said was partly wrong, identical subtrees are not visited multiple times since 672eef7, but the documentation was not updated accordingly.

> Unrelated to this change, I am a bit worried if the API of DependencyRule is clear in this regard. Are rule authors aware that the rule can be invoked multiple times for the same package?

It's at least documented in `RuleSet.dependencyRule()`, but the rule API documentation could be improved.

oheger-bosch: @mnonnenmacher Ok, I see. Then this PR does (after my latest changes) not change the traversal done by the rules: In case of the dependency tree, a `PackageReference` is used to check whether the current sub tree has already been visited; so this is exactly as before.
When later the implementation for the dependency graph is added, `DependencyReference` objects will be used for this purpose. This is then more efficient, because these objects have been de-duplicated (i.e. if a single package occurs multiple times in the dependency graph with the same set of transitive dependencies, the `DependencyReference` for this package is reused); hence a `==` comparison is sufficient. But the semantic is the same.

mnonnenmacher (approved):
@oheger-bosch  I have added the breaking-change and release-notes labels because this PR changes the dependency rule API.

PR #4212, opened 2021-06-23
OrtMain: Also print the actual count of severe issues

sschuberth:
While at it, generalize the wording to match for both single and
multiple issues.

### DIFF
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -78,19 +78,24 @@ data class GlobalOptions(
  * than [threshold], print an according note and throw a ProgramResult exception with [severeStatusCode].
  */
 fun concludeSeverityStats(counts: Map<Severity, Int>, threshold: Severity, severeStatusCode: Int) {
-    var hasSevereIssues = false
+    var severeIssueCount = 0
 
     fun getSeverityCount(severity: Severity) =
-        counts.getOrDefault(severity, 0).also { hasSevereIssues = it > 0 && severity >= threshold }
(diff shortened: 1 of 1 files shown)


cli/src/main/kotlin/OrtMain.kt
```diff
@@ -78,19 +78,24 @@ data class GlobalOptions(
  * than [threshold], print an according note and throw a ProgramResult exception with [severeStatusCode].
  */
 fun concludeSeverityStats(counts: Map<Severity, Int>, threshold: Severity, severeStatusCode: Int) {
-    var hasSevereIssues = false
+    var severeIssueCount = 0
 
     fun getSeverityCount(severity: Severity) =
-        counts.getOrDefault(severity, 0).also { hasSevereIssues = it > 0 && severity >= threshold }
+        counts.getOrDefault(severity, 0).also {
+            severeIssueCount += (it > 0 && severity >= threshold).compareTo(false)
```
oheger-bosch, line 85: This one is actually tricky. Can we be sure that the counter is always incremented by one? The JavaDoc of `Boolean.compareTo()` says that it returns

> zero if this object represents the same boolean value as the argument; a positive value if this object represents true and the argument represents false; and a negative value if this object represents false and the argument represents true

So "a positive value", but not necessarily 1.

sschuberth: True, I've changed the implementation.

oheger-bosch (approved):

sschuberth:
Ignoring the failed Docker build which is due to Eclipse infrastructure being down, see https://www.eclipsestatus.io/.

PR #4216, opened 2021-06-24
AbstractIntegrationSpec: Support results with dependency graphs

oheger-bosch:
Change the test assertions to accept results that either use a
dependency graph or the classic tree format. Otherwise, tests for
package managers that produce a dependency graph fail.

Note: Some test classes derived from this base class must have been
failing for a while now. This has not yet been noticed because these
tests are not executed by the default CI pipeline.

### DIFF
--- a/analyzer/src/funTest/kotlin/integration/AbstractIntegrationSpec.kt
+++ b/analyzer/src/funTest/kotlin/integration/AbstractIntegrationSpec.kt
@@ -22,11 +22,11 @@ package org.ossreviewtoolkit.analyzer.integration
 import io.kotest.core.spec.Spec
 import io.kotest.core.spec.style.StringSpec
 import io.kotest.inspectors.forAll
-import io.kotest.matchers.collections.beEmpty
+import io.kotest.inspectors.forAtLeastOne
+import io.kotest.inspectors.forOne
 import io.kotest.matchers.collections.containExactly
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
-import io.kotest.matchers.shouldNot
 import io.kotest.matchers.types.shouldBeTypeOf
 
 import java.io.File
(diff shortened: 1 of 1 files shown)


analyzer/src/funTest/kotlin/integration/AbstractIntegrationSpec.kt
```diff
@@ -133,11 +132,16 @@ abstract class AbstractIntegrationSpec : StringSpec() {
                     VersionControlSystem.forType(result.project.vcsProcessed.type) shouldBe
                             VersionControlSystem.forType(pkg.vcs.type)
                     result.project.vcsProcessed.url shouldBe pkg.vcs.url
-                    result.project.scopes shouldNot beEmpty()
-                    result.packages shouldNot beEmpty()
+                    result.project.hasScopes() shouldBe true
+                    (results.dependencyGraph != null || result.packages.isNotEmpty()) shouldBe true
```
sschuberth, line 136: Can we split this into two assertions wrapped in `forOne` instead? And could we use a similar appraoch to get rid of the `Project.hasScopes()` helper function?

sschuberth: For details see https://kotest.io/docs/assertions/inspectors.html.

oheger-bosch: From the description of inspectors, it is not clear to me how `forOne` could be used in this scenario. Do you mean something like `listOf(results.dependencyGraph != null, result.packages.isNotEmpty()).forOne { 
  it shouldBe true 
}` ?
I seem to miss something obvious.

sschuberth: > Do you mean something like `listOf(results.dependencyGraph != null, result.packages.isNotEmpty()).forOne { it shouldBe true }` ?

Yes.

oheger-bosch: Changed for both cases.

sschuberth (approved):

PR #4217, opened 2021-06-24
CreateCommand: Create new curations in existing files

MarcelBochtler:


helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
```diff
@@ -56,27 +57,32 @@ internal class CreateCommand : CliktCommand(
     override fun run() {
         val outputFile = getSplitCurationFile(outputDir, packageId, FileFormat.YAML.fileExtension)
 
-        if (outputFile.exists()) {
-            throw UsageError("File ${outputFile.absolutePath} already exists.")
+        val curations = if (outputFile.exists()) {
```
sschuberth, line 60: `.isFile`...

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
```diff
@@ -56,27 +57,32 @@ internal class CreateCommand : CliktCommand(
...
+            outputFile.readValue<MutableSet<PackageCuration>>()
+        } else {
+            mutableSetOf()
         }
 
+        if (packageId in curations.map { it.id }) {
+            throw IllegalArgumentException(
+                "Curation for ${packageId.toCoordinates()} already exists in ${outputFile.absolutePath}."
```
sschuberth, line 68: Nit: We usually single-quote paths (and also package coordinates) in case they contain spaces. Similar below.

sschuberth (dismissed):

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
```diff
@@ -56,27 +57,32 @@ internal class CreateCommand : CliktCommand(
...
 
-        if (outputFile.exists()) {
-            throw UsageError("File ${outputFile.absolutePath} already exists.")
+        val curations = if (outputFile.isFile) {
+            outputFile.readValue<MutableSet<PackageCuration>>()
+        } else {
+            mutableSetOf()
         }
 
+        if (packageId in curations.map { it.id }) {
```
fviernau, line 66: Why is it invalid to have multiple curations for a single `id`?

For example, the following could not be put anymore into separate curations for a single package.
* concluded license
* declared_license_mapping
* vcs info
* source artifact url

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
sschuberth: Still missing here...

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
```diff
@@ -20,20 +20,21 @@
...
 
 import com.github.ajalt.clikt.core.CliktCommand
-import com.github.ajalt.clikt.core.UsageError
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.file
 
 import java.io.IOException
+import java.lang.IllegalArgumentException
```
fviernau, line 29: Can you give a hint on the use case for the added logic?

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
MarcelBochtler: That would be done in a single entry for one identifier, right?
```yml
- id: "Maven:com.foo:bar:1.0"
  curations:
    comment: "An explanation why the curation is needed or the reasoning for a license conclusion"
    concluded_license: "valid SPDX license expression to override the license findings."
    declared_licenses:
    - "license a"
    - "license b"
    description: "Curated description"
    homepage_url: "http://example.com"
    binary_artifact:
      url: "http://example.com/binary.zip"
      hash: "ddce269a1e3d054cae349621c198dd52"
      hash_algorithm: "MD5"
    source_artifact:
      url: "http://example.com/sources.zip"
      hash: "ddce269a1e3d054cae349621c198dd52"
      hash_algorithm: "MD5"
    vcs:
      type: "git"
      url: "http://example.com/repo.git"
      revision: "1234abc"
      path: "subdirectory"
    is_meta_data_only: true
    is_modified: true
    declared_license_mapping:
      "license a": "Apache-2.0"
```

And not by adding a separate entry for `- id: "Maven:com.foo:bar:1.0"` (This would cause issues in ORT, because the user wouldn't know which curation is applied).


The check is being done to avoid overwriting or additionally adding `- id: "Maven:com.foo:bar:1.0"` again to the same file.

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
MarcelBochtler: The curation files support multiple entries by file.
This enhances the `create` command to add new curations to this file.

the file is being created in `<type>/<namespace>/<name>.yml` and multiple versions of packages live in this file.

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
fviernau: > That would be done in a single entry for one identifier, right?

I would prefer not to introduce the assumption the such curations are always done in a single entry.
Actually doing one thing at a time to me seems to be cleaner and has several advantages.
Just consider the `comment`. It has to provide a rationale for the curation. describing a single thing is much more straight forward compared to describing all things at once. It's a bit similar to "atomic commits" best practice IMO.

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
fviernau: Let's assume the file already exist and thus probably has >= 1 curations in it.
The use case is that a person wants to add a curation to that file. That user has to 

1. find the file location
2. open the file
3. insert a new curation

The rationale would be: the command helps with 1 / 2 since it prints the file location.
It helps with 3. by saving a copy and paste step. Instead of the user copying the curation
from directly above a template is already inserted. So was the intention of this commit to speed up 1,2,3?
Or maybe just 1 / 2 ?

MarcelBochtler: I wanted to cover all 3 mentioned use cases.

Additionally it covers use case 4 for me:
As a user that writes a curation, I don't need to know if a curation file already exists. The command behaves the same if a new file is created or if one already exists.

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
MarcelBochtler: Does ORT actually merge multiple curations for a single `package-id`?
What does that mean for this change? Should the command add a new curation with exactly the same package id?

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
fviernau: Thanks! makes sense now.

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
sschuberth: > Does ORT actually merge multiple curations for a single `package-id`?

Yes, somewhat: They are not merged, but applied in order, one after the other.

fviernau: > What does that mean for this change? Should the command add a new curation with exactly the same package id?

IMO it should at least not throw an exception if multiple curations with same identifier are in the file.

Whether it should always add a new curation: initially I was wondering whether it should add that template at all or just (create and) open that file. No strong opinion on that, I'd tend towards either always or never creating the template.

MarcelBochtler: > IMO it should at least not throw an exception if multiple curations with same identifier are in the file.

Yes I agree. I wasn't aware that ORT applies all curations for one `identifier`.

> I'd tend towards either always or never creating the template. What do you prefer?

In that case I would prefer to net creating the curation but printing the message that is as of now set as Exception message: `"Curation for ${packageId.toCoordinates()} already exists in '${outputFile.absolutePath}'."`

> open that file

That is not an option in the `helper-cli` but a wrapper script would do this, correct?

fviernau: > That is not an option in the helper-cli but a wrapper script would do this, correct?

Would agree.

MarcelBochtler: I removed the `Exception`. Please have another look.

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
```diff
@@ -56,27 +56,30 @@ internal class CreateCommand : CliktCommand(
...
+        val curations = if (outputFile.isFile) {
+            outputFile.readValue<MutableSet<PackageCuration>>()
+        } else {
+            mutableSetOf()
         }
 
-        val mapper = createBlockYamlMapper()
+        if (packageId in curations.map { it.id }) {
+            println("Curation for ${packageId.toCoordinates()} already exists in '${outputFile.absolutePath}'.")
+        } else {
```
fviernau, line 67: Consider adding a return into the `if-`branch to eliminate one level of nesting.

fviernau (approved):

helper-cli/src/main/kotlin/commands/packagecuration/CreateCommand.kt
MarcelBochtler: @fviernau Done.

fviernau (approved):

Issue #2848, opened 2020-07-09
Allow to prefer source artifacts over source code from VCS


sschuberth:
BTW, can we have an option to try artifact source before vcs?

_Originally posted by @kneep in https://github.com/oss-review-toolkit/ort/issues/2846#issuecomment-655879046_

sschuberth:
Currently, it is hard-coded to try downloading sources from VCS first, and then try a source artifact:

https://github.com/oss-review-toolkit/ort/blob/46a9ae9f730d132bd363bfa72c3d61fce630b2ea/downloader/src/main/kotlin/Downloader.kt#L124-L155

There could be an option to change the order.

kneep:
Here in China, as you may all know, we have some network limitation. The public vcs repositories are hosted on various sites. Some of these sites are not accessible (e.g. sourceforge) while others have poor speed (e.g. Github, especially after evening). But generally artifacts repositories have reasonable speed here and are more predictable. So maybe it's better to have some flexibility on the priority of the source of source code.

sschuberth:
Another problem that would get solved by this:

Suppose a Maven (binary) artifact has both a belonging sources artifact and a SCM / VCS location. Then the current implementation would scan the source code from the VCS, but the VCS typically contains test code for that dependency which is not part of the sources artifact. As test code often has very different detected licenses than the real product code, scanning the VCS results in potentially many irrelevant (but not false) findings that would need to be excluded by a [package configuration](https://github.com/oss-review-toolkit/ort/blob/master/docs/config-file-package-configuration-yml.md), which is cumbersome. With that in mind it makes sense to focus as much as possible on the source code that makes up the binary, which means to prefer the sources artifact.

fviernau:
I believe it makes sense to make this configurable. I could see the following use cases

1. only scan source artifact
2. only scan VCS
3. prefer source artifact, fallback to VCS
4. prefer VCS, fallback to source artifact
5. prefer cached, then fallback to VCS, then fallback to source artifact
6. prefer cached, then fallback to source artifact, then fallback to source artifact

I believe what we currently have implemented is #5.
I'd propose to make the behavior configurable.
I'd have a strong interrest in having one configuration option which mimics the exact current behavior.

mnonnenmacher:
[...]

PR #4221, opened 2021-06-24
helper-cli: Add command to delete entries from postgres scan storage

MarcelBochtler:
While this is quite a specific command, it is required for https://github.com/oss-review-toolkit/ort/issues/2848.

Inconsistencies in the scan storage caused by switching the prefered source can be cleaned up using this `helper-cli` command.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands.scanstorage
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.associate
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.split
+import com.github.ajalt.clikt.parameters.types.file
+
+import org.jetbrains.exposed.sql.Database
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.isNotNull
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.like
+import org.jetbrains.exposed.sql.compoundAnd
+import org.jetbrains.exposed.sql.compoundOr
+import org.jetbrains.exposed.sql.deleteWhere
+
+import org.ossreviewtoolkit.helper.common.ORTH_NAME
+import org.ossreviewtoolkit.model.SourceCodeOrigin
+import org.ossreviewtoolkit.model.config.OrtConfiguration
+import org.ossreviewtoolkit.model.config.PostgresStorageConfiguration
+import org.ossreviewtoolkit.model.utils.DatabaseUtils
+import org.ossreviewtoolkit.model.utils.DatabaseUtils.transaction
+import org.ossreviewtoolkit.model.utils.rawParam
+import org.ossreviewtoolkit.scanner.storages.utils.ScanResults
+import org.ossreviewtoolkit.utils.ORT_CONFIG_FILENAME
+import org.ossreviewtoolkit.utils.expandTilde
+import org.ossreviewtoolkit.utils.log
+import org.ossreviewtoolkit.utils.ortConfigDirectory
+
+internal class DeleteCommand : CliktCommand(
+    help = "Removes stored scan results matching the options or all results if no options are given."
+) {
+    private val configFile by option(
+        "--config",
+        help = "The path to the ORT configuration file that configures the scan results storage."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+        .default(ortConfigDirectory.resolve(ORT_CONFIG_FILENAME))
+
+    private val configArguments by option(
+        "-P",
+        help = "Override a key-value pair in the configuration file. For example: " +
+                "-P scanner.postgresStorage.schema=testSchema"
+    ).associate()
+
+    private val sourceCodeOrigins by option(
+        "--source-code-origins",
+        help = "The origin of the scan results that should be deleted."
+    ).convert { SourceCodeOrigin.valueOf(it) }
+        .split(",")
+        .default(emptyList())
+
+    private val packageType by option(
+        "--package-type",
+        help = "The package manager type to delete from the scan storage. Like 'Maven' or 'NPM'."
+    )
+
+    override fun run() {
+        val database = initDatabase(OrtConfiguration.load(configArguments, configFile))
+
+        val provenanceKeys = sourceCodeOrigins.map { origin ->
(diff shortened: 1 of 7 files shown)


helper-cli/src/main/kotlin/common/DatabaseHelper.kt
```diff
@@ -0,0 +1,47 @@
...
+
+        log.info {
+            "Using Postgres storage with URL '${storageConfig.url}' and schema '${storageConfig.schema}'."
+        }
+
+        val dataSource = DatabaseUtils.createHikariDataSource(
+            config = storageConfig,
+            applicationNameSuffix = ORTH_NAME,
+            // Use a value slightly higher than the number of threads accessing the storage.
+            maxPoolSize = LocalScanner.NUM_STORAGE_THREADS + 3
```
oheger-bosch, line 42: This is probably a copy & paste error. There is only a single thread accessing the database. So you could set `maxPoolSize` to a rather low value, maybe even 1.

helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
```diff
@@ -0,0 +1,102 @@
...
+    ).associate()
+
+    private val provenance by option(
+        "--provenance",
+        help = "The origin of the scan results that should be deleted."
+    ).enum<SourceCodeOrigin>()
+
+    private val type by option(
+        "--type",
+        help = "The type of the package to delete from the scan storage."
```
oheger-bosch, line 70: From the help text, it is not really clear what _type_ actually means. It would be better to explicitly state that the type of package manager like "Maven" or "Gradle" is referred to.

helper-cli/src/main/kotlin/common/DatabaseHelper.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import org.jetbrains.exposed.sql.Database
+
+import org.ossreviewtoolkit.model.config.OrtConfiguration
+import org.ossreviewtoolkit.model.config.PostgresStorageConfiguration
+import org.ossreviewtoolkit.model.utils.DatabaseUtils
+import org.ossreviewtoolkit.scanner.LocalScanner
+import org.ossreviewtoolkit.utils.log
+
+internal object DatabaseHelper {
```
oheger-bosch, line 30: Is this class actually needed? It does not much more than calling `DatabaseUtils.createHikariDataSource()`. So should this logic be inlined into the command? Or is this function generic enough to be moved to `DatabaseUtils`?

helper-cli/src/main/kotlin/common/DatabaseHelper.kt
MarcelBochtler: As this is implemented as a Utitlity class I think that it will be used for different use cases as well, which might access the database in multiple threads.
Do you think we'll remember to raise the pool size then?

However, you're right, this is a copy paste error, I'd suggest to just use the default pool size (5).
What do you think?

helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
MarcelBochtler: How about `The package manager type to delete from the scan storage. Like 'Maven' or 'NPM'.`?

helper-cli/src/main/kotlin/common/DatabaseHelper.kt
MarcelBochtler: I'll move it to a private method in the `DeleteCommand`.

helper-cli/src/main/kotlin/common/DatabaseHelper.kt
oheger-bosch: Hikari will remind us to increase the pool size by spitting out strange errors related to timeouts ;-)

But kidding aside, if it is a generic utility function, then the pool size should be a parameter - as you said, it depends on the use case.

helper-cli/src/main/kotlin/commands/ScanStorageCommand.kt
```diff
@@ -0,0 +1,39 @@
...
+) {
+    init {
+        subcommands(
+            DeleteCommand()
+        )
+    }
+
+    override fun run() {
+        // Intentionally empty, because all logic is implemented inside the sub commands.
+    }
```
sschuberth, line 38: You might want to derive from `NoOpCliktCommand` then instead.

helper-cli/src/main/kotlin/common/DatabaseHelper.kt
MarcelBochtler: I moved this functionality to a private method.
So it is no longer a utility function and I reduced the `maxPoolSize` to `1`.

helper-cli/src/main/kotlin/commands/ScanStorageCommand.kt
sschuberth: I see other helper-cli commands do the same. So maybe add another commit to address this for all of them?

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/ScanStorageCommand.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.ossreviewtoolkit.helper.commands
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.subcommands
+
+import org.ossreviewtoolkit.helper.commands.scanstorage.DeleteCommand
+
+class ScanStorageCommand : CliktCommand(
```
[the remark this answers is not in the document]
MarcelBochtler: > What's the reason for adding a subcommand?

Because I can see multiple commands regarding the scan-results. Like the `ListStoredScanResultsCommand` that would also fit this subcommand. E.g. as `read`.

> Listing and deleting are very similar because in terms of SQL the only difference is "DELETE FROM" vs "SELECT FROM", thus all other logic would be redundant.

While that is correct, the `ListStoredScanResultsCommand` also works for other storage types. The `DeleteCommand` only works for the postgres storage, because a delete functionality is not yet implemented in the production code, and the way I implemented it here is IMO too specific to reuse it in any meaningful way.

What is your suggested change?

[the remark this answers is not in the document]
fviernau: > What is your suggested change?

No change suggested. I just wasn't able to judge whether a subcommand makes sense based on the information provided. Which is why I asked.

helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
```diff
@@ -0,0 +1,122 @@
...
+        .convert { it.absoluteFile.normalize() }
+        .default(ortConfigDirectory.resolve(ORT_CONFIG_FILENAME))
+
+    private val configArguments by option(
+        "-P",
+        help = "Override a key-value pair in the configuration file. For example: " +
+                "-P scanner.postgresStorage.schema=testSchema"
+    ).associate()
+
+    private val provenance by option(
```
fviernau, line 66: How about naming this `sourceCodeOrigin` as the term `provenance` is used already with a different semantics.

helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
```diff
@@ -0,0 +1,122 @@
...
+        "--provenance",
+        help = "The origin of the scan results that should be deleted."
+    ).enum<SourceCodeOrigin>()
+
+    private val type by option(
+        "--type",
```
fviernau, line 72: Would it make sense to rename this to `--package-type` ?

helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
```diff
@@ -0,0 +1,122 @@
...
+        help = "The package manager type to delete from the scan storage. Like 'Maven' or 'NPM'."
+    )
+
+    override fun run() {
+        val database = initDatabase(OrtConfiguration.load(configArguments, configFile))
+
+        val provenanceKey = when (provenance) {
+            SourceCodeOrigin.ARTIFACT -> "source_artifact"
+            SourceCodeOrigin.VCS -> "vcs_info"
+            null -> null
```
fviernau, line 82: I believe making the `provenance` required, and turning it into a collection (created from CSV) would simplify the logic here. What do you think?

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
MarcelBochtler: I like the suggestion. Please have a look at the changes.

fviernau (approved):

PR #4224, opened 2021-06-25
Improve String.wrapAt()

fviernau:


helper-cli/src/main/kotlin/common/Utils.kt
```diff
@@ -418,16 +418,9 @@ internal fun String.wrapAt(column: Int): String {
...
-        var index = spaceIndex
-        while (index > column) {
-            // Find the last space before the column.
-            while (index > 0 && text[--index] != ' ');
-        }
-
-        // If there is no space before the column, take the first space after the column.
-        if (index == 0) index = spaceIndex
+        val firstSpaceAfterColumnIndex = text.indexOf(' ', column)
+        val lastSpaceBeforeColumnIndex = text.substring(0, column).lastIndexOf(' ')
```
sschuberth: I believe this could even be simplified further by only using `val lastSpaceBeforeColumnIndex = text.lastIndexOf(' ', column - 1)`.

helper-cli/src/main/kotlin/common/Utils.kt
```diff
@@ -418,16 +418,9 @@ internal fun String.wrapAt(column: Int): String {
...
+        val index = lastSpaceBeforeColumnIndex.takeIf { it != -1 } ?: firstSpaceAfterColumnIndex
```
sschuberth: Nit: Maybe `takeUnless { it == -1 }` to better denote what the expected common case is.

sschuberth (changes requested):

helper-cli/src/main/kotlin/common/Utils.kt
fviernau: Hm, `column - 1` would be the `startIndex` parameter. So it couldn't match a space before that index?!

fviernau: I see, thanks! Nice simplification btw!

helper-cli/src/main/kotlin/common/Utils.kt
```diff
@@ -418,16 +418,9 @@ internal fun String.wrapAt(column: Int): String {
...
-        while (index > column) {
-            // Find the last space before the column.
-            while (index > 0 && text[--index] != ' ');
-        }
-
-        // If there is no space before the column, take the first space after the column.
-        if (index == 0) index = spaceIndex
+        val firstSpaceAfterColumnIndex = text.indexOf(' ', column)
+        val lastSpaceBeforeColumnIndex = text.lastIndexOf(' ', column - 1)
+        val index = lastSpaceBeforeColumnIndex.takeUnless { it == -1 } ?: firstSpaceAfterColumnIndex
```
sschuberth: Nit: Do you think it makes sense to inline `firstSpaceAfterColumnIndex` now so it's only calculated in the of the fallback?

fviernau: I tried out the inlineing when writing this code, decided against it as I found it less readable.

sschuberth (approved):

PR #4225, opened 2021-06-25
helper-cli: Miscellaneous package curation command improvements

fviernau:


helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
```diff
@@ -56,18 +58,15 @@ internal class SplitCommand : CliktCommand(
...
         }
 
         val mapper = createBlockYamlMapper()
         groupedCurations.forEach { (outputFile, curations) ->
-            val curationsToPersist = if (outputFile.isFile) {
-                outputFile.readValue<MutableSet<PackageCuration>>()
-            } else {
-                mutableSetOf()
-            }
+            val curationsToPersist = readPackageCurations(outputFile).toMutableSet()
```
sschuberth: Please add a rationale to the commit message, esp. as there seems no reuse being involved, and it's only a small amount of code anyway. The "reuse" in line 61 is not required as the input file is forced to exist.

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
```diff
@@ -64,14 +64,13 @@ internal class SplitCommand : CliktCommand(
...
         val mapper = createBlockYamlMapper()
         curationsByOutputFile.forEach { (outputFile, curations) ->
-            val curationsToPersist = readPackageCurations(outputFile).toMutableSet()
-
-            curationsToPersist += curations
-
-            val curationsWithBlockComment = curationsToPersist.mapTo(mutableSetOf()) { it.formatComment() }
+            val curationsToPersist = (readPackageCurations(outputFile) + curations)
+                .map { it.formatComment() }
+                .distinct()
```
MarcelBochtler, line 61: Why is `.distinct()` better than just using `.toSet()`?
Especially because `.distinct()` uses `toSet()` internally and then converts it to a `List` again.

helper-cli/src/main/kotlin/common/Utils.kt
```diff
@@ -839,7 +841,18 @@ internal fun PackageConfiguration.write(targetFile: File) {
...
-internal fun createBlockYamlMapper(): ObjectMapper =
+/**
+ * Serialize [PackageCuration] to the given [targetFile] as YAML.
+ */
+internal fun Collection<PackageCuration>.writeAsYaml(targetFile: File) {
+    targetFile.parentFile.safeMkdirs()
+
+    createBlockYamlMapper()
+        .writerWithDefaultPrettyPrinter()
+        .writeValueAsString(this)
```
MarcelBochtler, line 877: This is not actually writing the value to `targetFile`.

MarcelBochtler (changes requested):

helper-cli/src/main/kotlin/commands/packagecuration/SplitCommand.kt
fviernau: Due to the sorting right below it becomes a list anyway.
I found distinct slightly more speaking and a bit more emphasizing that this deals with a list, not a set.

Do you want me to change it?

MarcelBochtler: > I found distinct slightly more speaking

I agree.

> Do you want me to change it?

No.

MarcelBochtler (approved):

fviernau:
Merge regardless of the unrelated failing analyzer test

PR #4228, opened 2021-06-25
CocoaPods: Gracefully handle absent Podspecs

fviernau:
Don't throw a NPE in case a Podspec cannot be found in the default Specs
repository. Use an empty package instead. This is particularly important
in case the Podfile uses non-default Specs repositories, which are not
yet supported in ORT.

### DIFF
--- a/analyzer/src/main/kotlin/managers/CocoaPods.kt
+++ b/analyzer/src/main/kotlin/managers/CocoaPods.kt
@@ -204,7 +204,7 @@ class CocoaPods(
 
         val podspecFile = podspecCommand.stdout.trim().let { File(it) }
 
-        podspecFile.readValue<Podspec>().withSubspecs().associateByTo(podspecCache, { it.name })
+        podspecFile.readValue<Podspec>().withSubspecs().associateByTo(podspecCache) { it.name }
 
         return podspecCache.getValue(id.name)
     }
--- a/analyzer/src/main/kotlin/managers/CocoaPods.kt
+++ b/analyzer/src/main/kotlin/managers/CocoaPods.kt
@@ -169,7 +169,8 @@ class CocoaPods(
     }
 
     private fun getPackage(id: Identifier, workingDir: File): Package {
-        val podspec = getPodspec(id, workingDir)
+        // TODO: Emit a warning / hint when no Podspec is found.
+        val podspec = getPodspec(id, workingDir) ?: return Package.EMPTY.copy(id = id)
 
         val vcs = podspec.source["git"]?.let { url ->
             VcsInfo(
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -169,7 +169,7 @@ class CocoaPods(
     }
 
     private fun getPackage(id: Identifier, workingDir: File): Package {
-        val podspec = getPodspec(id, workingDir)
+        val podspec = getPodspec(id, workingDir) ?: return Package.EMPTY.copy(id = id)
```
sschuberth, line 172: Shouldn't this at least also emit a warning / hint?

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -192,14 +192,14 @@ class CocoaPods(
...
-    private fun getPodspec(id: Identifier, workingDir: File): Podspec {
+    private fun getPodspec(id: Identifier, workingDir: File): Podspec? {
         podspecCache[id.name]?.let { return it }
 
         val podspecName = id.name.substringBefore("/")
 
         val podspecFile = run(
             "spec", "which", podspecName, "--version=${id.version}", "--allow-root", "--regex", workingDir = workingDir
-        ).requireSuccess().stdout.trim().let { File(it) }
+        ).takeIf { it.isSuccess }?.stdout?.trim()?.let { File(it) } ?: return null
```
sschuberth, line 202: This is becoming a bit unreadable, IMO. How about splitting this like

        val podspecCommand = run(
            "spec", "which", podspecName, "--version=${id.version}", "--allow-root", "--regex", workingDir = workingDir
        ).takeIf { it.isSuccess } ?: return null

        val podspecFile = File(podspecCommand.stdout.trim())

analyzer/src/main/kotlin/managers/CocoaPods.kt
```diff
@@ -192,14 +192,14 @@ class CocoaPods(
...
 
         podspecFile.readValue<Podspec>().withSubspecs().associateByTo(podspecCache, { it.name })
```
sschuberth, line 204: While at it, in adding a commit to move out the lambda of the parentheses here?

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/CocoaPods.kt
fviernau: I made this a TODO for now, as I wanted to eliminate the NPE quickly.

sschuberth (approved):

PR #4235, opened 2021-06-29
fossid-webapp: Add an asynchronous mode to the scanner

nnobelis:
When the 'noWait' parameter is set, ORT doesn't wait for repositories
to be downloaded nor scans to be completed.
Scan results are then only available in the FossID instance.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -51,6 +51,7 @@ import org.ossreviewtoolkit.clients.fossid.model.status.ScanState
 import org.ossreviewtoolkit.clients.fossid.runScan
 import org.ossreviewtoolkit.model.OrtIssue
 import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
 import org.ossreviewtoolkit.model.ScanSummary
 import org.ossreviewtoolkit.model.Severity
@@ -82,6 +83,8 @@ import retrofit2.converter.jackson.JacksonConverterFactory
  * will be scanned.
  * * **"packageAuthorsFilter":** If this optional filter is set, only packages from a given author will be scanned.
  * * **"addAuthenticationToUrl":** When set, ORT will add credentials from its Authenticator to the URLs sent to FossID.
+ * * **"waitForResult":** When set to false, ORT doesn't wait for repositories to be downloaded nor scans to be
+ * completed. As a consequence, scan results won't be available in ORT result.
  */
 class FossId(
     name: String,
@@ -153,7 +156,7 @@ class FossId(
     private val serverUrl: String
     private val apiKey: String
     private val user: String
-
+    private val waitForResult: Boolean
     private val secretKeys = listOf("serverUrl", "apiKey", "user")
 
     private val service: FossIdRestService
@@ -179,6 +182,9 @@ class FossId(
         packageNamespaceFilter = fossIdScannerOptions["packageNamespaceFilter"].orEmpty()
         packageAuthorsFilter = fossIdScannerOptions["packageAuthorsFilter"].orEmpty()
         addAuthenticationToUrl = fossIdScannerOptions["addAuthenticationToUrl"].toBoolean()
+        waitForResult = fossIdScannerOptions["waitForResult"]?.toBooleanStrictOrNull() ?: true
+
+        log.info { "waitForResult parameter is set to '$waitForResult'" }
 
         val client = OkHttpClientHelper.buildClient()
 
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -311,11 +315,13 @@ class FossId(
                     }
                 }
 
-                checkScan(scanCode)
-                val rawResults = getRawResults(scanCode)
-                val resultsSummary = createResultSummary(startTime, rawResults)
+                if (!noWait) {
```
sschuberth, line 318: I stumbled over this double negation. Do you think it makes sense to use positive logic and rename `noWait` to something like `waitForResult`? Or is `noWait` a term coined by FOSSID itself?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -311,11 +315,13 @@ class FossId(
...
+                    checkScan(scanCode)
+                    val rawResults = getRawResults(scanCode)
+                    val resultsSummary = createResultSummary(startTime, rawResults)
 
-                results.getOrPut(pkg) { mutableListOf() } += resultsSummary
+                    results.getOrPut(pkg) { mutableListOf() } += resultsSummary
```
oheger-bosch, line 323: IIUC, there is currently no indication that a package has been processed if `noWait` is *true*. Would it make sense to have one?

It is clear that no actual result is available yet, but could there be a kind of dummy result - ideally with information from where the final result can be obtained? Does the ORT data model offer anything that could be used for this purpose? (The support for asynchronous scanners is rather limited, I assume.)

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: No `noWait` comes from me :)
But you are right I will address it.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: I thought about creating an ORT issue with a level of `HINT` with something like:
`Scanner has run in asynchronous mode. Scan results are available on the FossID instance.`

but since issue is a sub property from `ScanSummary`, it would have meant that this issue repeats identically for each package.

Now if I cross it with your first comment, we could change the issue content to:

`Package xyz has been scanned in asynchronous mode. Scan results are available on the FossID instance.`

What to you think about it ?

oheger-bosch: This would be fine with me as a work-around - until asynchronous scanners are treated as first-class citizens in the ORT model.

oheger-bosch (dismissed):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -179,6 +181,9 @@ class FossId(
         packageNamespaceFilter = fossIdScannerOptions["packageNamespaceFilter"].orEmpty()
         packageAuthorsFilter = fossIdScannerOptions["packageAuthorsFilter"].orEmpty()
         addAuthenticationToUrl = fossIdScannerOptions["addAuthenticationToUrl"].toBoolean()
+        waitForResult = (fossIdScannerOptions["waitForResult"] ?: "true").toBoolean()
```
sschuberth, line 184: I believe it's not so nice to hard-code the `true` default as a string here. I propose instead:

    fossIdScannerOptions["waitForResult"]?.toBooleanStrictOrNull() ?: true

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -311,11 +316,33 @@ class FossId(
...
+                        fileCount = 0,
+                        packageVerificationCode = "",
+                        licenseFindings = sortedSetOf(),
+                        copyrightFindings = sortedSetOf(),
+                        issues = listOf(
+                            issue
+                        )
+                    )
+
+                    results.getOrPut(pkg) { mutableListOf() } += ScanResult(UnknownProvenance, details, summary)
```
sschuberth, line 344: The `UnknownProvenance` here strikes me. I know it has been like this before in `createResultSummary()`, but *why* is the provenance actually unknown currently in any case? We *do* know what's been scanned, as we told FOSSID what to scan, or?

nnobelis: Yes this is right. We can replace it by `RepositoryProvenance` right ?
Does it needs to be a separate commit ?

sschuberth (approved):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -311,11 +317,36 @@ class FossId(
                     }
                 }
 
-                checkScan(scanCode)
-                val rawResults = getRawResults(scanCode)
-                val resultsSummary = createResultSummary(startTime, rawResults)
+                val provenance = RepositoryProvenance(pkg.vcsProcessed, pkg.vcsProcessed.revision)
```
sschuberth, line 320: Thanks, approved. Yet one more question, though: Do we currently support scanning source artifacts with FOSSID? Because if so, we can't hard-code the provenance like this here. And if not, we should support scanning source artifacts with FOSSID.

nnobelis: Looking at the API documentation, the way to scan source artefacts is through the file upload mechanism we don't support (yet).
My suggestion would be to leave the provenance implementation as it is in current PR and tweak it when we implement file upload mode.
I created a ticket for it:
https://github.com/oss-review-toolkit/ort/issues/4242

PR #4237, opened 2021-06-29
VulnerabilityReference: Rewrite getSeverityString() to match more systems

sschuberth:
A VulnerabilityReference has a nullable severity, so also handle that in
in getSeverityString(), and generalize the scoring system matching to be
able to handle more cases.

### DIFF
--- a/model/src/main/kotlin/VulnerabilityReference.kt
+++ b/model/src/main/kotlin/VulnerabilityReference.kt
@@ -52,14 +52,15 @@ data class VulnerabilityReference(
 ) {
     companion object {
         /**
-         * Return a human-readable string that is determined based on the given score and scoring system.
+         * Return a human-readable string that is determined based on the given [scoringSystem] and [severity].
          */
-        fun getSeverityString(scoringSystem: String?, severity: String) =
-            when (scoringSystem) {
-                "CVSS2" -> Cvss2Rating.fromScore(severity.toFloat())?.name?.lowercase()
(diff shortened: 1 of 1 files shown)


model/src/main/kotlin/VulnerabilityReference.kt
```diff
@@ -52,14 +52,15 @@ data class VulnerabilityReference(
...
-        fun getSeverityString(scoringSystem: String?, severity: String) =
-            when (scoringSystem) {
-                "CVSS2" -> Cvss2Rating.fromScore(severity.toFloat())?.name?.lowercase()
-                "CVSS3" -> Cvss3Rating.fromScore(severity.toFloat())?.name?.lowercase()
-                else -> "unknown"
-            }
+        @Suppress("UNUSED") // This function is used in the templates.
+        fun getSeverityString(scoringSystem: String?, severity: String?) =
+            when (scoringSystem?.uppercase()) {
+                "CVSS2", "CVSSV2" -> severity?.toFloat()?.let { Cvss2Rating.fromScore(it)?.toString() }
```
oheger-bosch, line 60: As we do not have control over the quality of data returned by a vulnerability provider, I would prefer a more robust implementation for the conversion to a numeric value. `toFloat()` may just blow up for unexpected input. Maybe have an extension function `toScore()` on `String` that catches `NumberFormatException`?

sschuberth: I'm simply using `toFloatOrNull()` now.

oheger-bosch (approved):

fviernau (approved):

PR #4239, opened 2021-06-30
Introduce and use VcsInfo.orEmpty()

fviernau:


model/src/main/kotlin/VcsInfo.kt
```diff
@@ -100,6 +100,11 @@ data class VcsInfo(
     fun toCuration() = VcsInfoCurationData(type, url, revision, path)
 }
 
+/**
+ * Return this [VcsInfo] if not null or else [VcsInfo.EMPTY].
+ */
+fun VcsInfo?.orEmpty(): VcsInfo = this ?: VcsInfo.EMPTY
```
sschuberth, line 106: I'm myself a bit undecided, but shouldn't this strictly speaking be put into `model/src/main/kotlin/utils/Extensions.kt`?

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -32,9 +32,9 @@ import org.ossreviewtoolkit.model.AnalyzerResult
 import org.ossreviewtoolkit.model.AnalyzerRun
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Repository
-import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.model.orEmpty
```
sschuberth, line 37: Looks like you missed to replace this:

https://github.com/oss-review-toolkit/ort/blob/a2d82888eecc3049e1872140b20359e486207d61/analyzer/src/main/kotlin/managers/Npm.kt#L234

sschuberth (changes requested):

model/src/main/kotlin/VcsInfo.kt
fviernau: I prefer to keep it here but wouldn't mind moving it if you want me to. Note that we have similar extension e.g. in `LicenseClassifications.kt`, `Resolutions.kt`.

PR #4243, opened 2021-06-30
DependencyGraphConverter: Take issues of dependencies into account

oheger-bosch:
DependencyGraphConverter missed the issues stored at PackageReferences
when constructing a DependencyGraph. By overriding the
issuesForDependency() function in the private ScopesDependencyHandler
class, they are now preserved.

### DIFF
--- a/model/src/test/kotlin/utils/DependencyGraphConverterTest.kt
+++ b/model/src/test/kotlin/utils/DependencyGraphConverterTest.kt
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.model.utils
 
 import io.kotest.assertions.fail
 import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.collections.beEmpty
 import io.kotest.matchers.collections.containExactlyInAnyOrder
 import io.kotest.matchers.nulls.beNull
 import io.kotest.matchers.should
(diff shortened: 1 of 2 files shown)


model/src/test/kotlin/utils/DependencyGraphConverterTest.kt
```diff
@@ -135,9 +157,17 @@ private fun createIdentifier(managerName: String, index: Int, forProject: Boolea
...
+        PackageReference(createIdentifier(managerName, index, forProject = false), issues = createIssues(index))
     }
 
+/**
+ * Create a list with issues for a test dependency based on its [index].
+ */
+private fun createIssues(index: Int): List<OrtIssue> =
+    emptyList<OrtIssue>().takeIf { index % 2 == 0 } ?: listOf(
+        OrtIssue(source = "test", message = "Test issue $index.")
+    )
```
sschuberth, line 169: Alternatively:

    listOfNotNull(
        OrtIssue(source = "test", message = "Test issue $index.").takeUnless { index % 2 == 0 }
    )

oheger-bosch: IMHO the original approach is more efficient as an `OrtIssue` object is only created if necessary. Regarding the readability, I see no major difference.

sschuberth (approved):

PR #4244, opened 2021-07-01
Add a DependencyNavigator implementation for dependency graphs

oheger-bosch:
This is going to become the default implementation of the `DependencyNavigator` interface. It allows access to dependency information in ORT results that completely use the dependency graph format.

This PR also reworks the representation of dependency graphs in ORT result files again. The new format has the following advantages:
* It is more compact, especially for complex graphs referencing dependency sub trees multiple times. (To give you an impression: for a bigger JavaScript project, the size of the analyzer result could be reduced from 25 MB to ~900 KB.)
* No duplication of objects happens during serialization or deserialization.
* The serialized form may be easier to understand, as it is based on a typical graph representation with nodes and edges.

The code can handle the original dependency graph format as well; so existing result files can still be processed.

Note that on loading an ORT result, the single commands still do a conversion to the classic dependency tree format. There is one more PR to come that does the final integration.

### DIFF
--- a/model/src/main/kotlin/DependencyGraph.kt
+++ b/model/src/main/kotlin/DependencyGraph.kt
@@ -19,10 +19,16 @@
 
 package org.ossreviewtoolkit.model
 
+import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonInclude
 
 import java.lang.IllegalStateException
 import java.util.SortedSet
+
+/**
+ * Type alias for a [Map] that associates a [DependencyGraphNode] with the nodes representing its dependencies.
+ */
+typealias NodeDependencies = Map<DependencyGraphNode, List<DependencyGraphNode>>
 
 /**
  * A data class that represents the graph of dependencies of a project.
@@ -41,25 +47,40 @@ import java.util.SortedSet
  * with [Scope] and [PackageReference] objects (with shared references, so keeping memory consumption low, too), so
  * that it can be processed in the usual ways.
  */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
 data class DependencyGraph(
     /**
      * A list with the identifiers of the packages that appear in the dependency graph. This list is used to resolve
-     * the numeric indices contained in the [DependencyReference] objects.
+     * the numeric indices contained in the [DependencyGraphNode] objects.
      */
     val packages: List<Identifier>,
 
     /**
      * Stores the dependency graph as a list of root nodes for the direct dependencies referenced by scopes. Starting
      * with these nodes, the whole graph can be traversed. The nodes are constructed from the direct dependencies
-     * declared by scopes that cannot be reached via other paths in the dependency graph.
+     * declared by scopes that cannot be reached via other paths in the dependency graph. Note that this property
+     * exists for backwards compatibility only; it is replaced by the lists of nodes and edges.
      */
-    val scopeRoots: SortedSet<DependencyReference>,
+    val scopeRoots: SortedSet<DependencyReference> = sortedSetOf(),
 
     /**
      * A mapping from scope names to the direct dependencies of the scopes. Based on this information, the set of
      * [Scope]s of a project can be constructed from the serialized form.
      */
-    val scopes: Map<String, List<RootDependencyIndex>>
+    val scopes: Map<String, List<RootDependencyIndex>>,
+
+    /**
+     * A list with the nodes of this dependency graph. Nodes correspond to packages, but in contrast to the [packages]
+     * list, there can be multiple nodes for a single package. The order of nodes in this list is relevant; the
+     * edges of the graph reference their nodes by numeric indices.
+     */
+    val nodes: List<DependencyGraphNode>? = null,
+
+    /**
+     * A list with the edges of this dependency graph. By traversing the edges, the dependencies of packages can be
+     * determined.
+     */
+    val edges: List<DependencyGraphEdge>? = null
 ) {
     companion object {
         /**
@@ -93,6 +114,12 @@ data class DependencyGraph(
             // namespace, the name, and the version.
             scopeName.split(':', limit = 4).getOrElse(3) { scopeName }
     }
+
+    /**
+     * A mapping that allows fast access to the dependencies of a node in this graph.
+     */
+    @get:JsonIgnore
+    val dependencies: NodeDependencies by lazy { constructNodeDependencies() }
 
     /**
      * Stores a mapping from dependency indices to [PackageReference] objects. This is needed when converting the
@@ -136,36 +163,44 @@ data class DependencyGraph(
      */
     private fun constructReferenceMapping(): Map<String, PackageReference> {
         val refMapping = mutableMapOf<String, PackageReference>()
+        val allNodes = nodes ?: scopeRoots.map(DependencyReference::toGraphNode)
 
-        scopeRoots.forEach { tree ->
-            constructReferenceTree(tree, refMapping)
-        }
+        allNodes.forEach { constructReferenceTree(it, refMapping) }
 
         return refMapping
     }
 
     /**
-     * Construct the tree with [PackageReference]s by navigating the dependency tree starting with [ref] and
+     * Construct the tree with [PackageReference]s by navigating the dependency graph starting with [node] and
      * populate the given [refMapping].
      */
     private fun constructReferenceTree(
-        ref: DependencyReference,
+        node: DependencyGraphNode,
         refMapping: MutableMap<String, PackageReference>
     ): PackageReference {
-        val indexKey = RootDependencyIndex.generateKey(ref.pkg, ref.fragment)
+        val indexKey = RootDependencyIndex.generateKey(node.pkg, node.fragment)
         return refMapping.getOrPut(indexKey) {
-            val dependencies = ref.dependencies.mapTo(sortedSetOf()) {
+            val refDependencies = dependencies[node].orEmpty().mapTo(sortedSetOf()) {
                 constructReferenceTree(it, refMapping)
             }
 
             PackageReference(
-                id = packages[ref.pkg],
-                dependencies = dependencies,
-                linkage = ref.linkage,
-                issues = ref.issues
+                id = packages[node.pkg],
+                dependencies = refDependencies,
+                linkage = node.linkage,
+                issues = node.issues
             )
         }
     }
+
+    /**
+     * Construct a mapping that allows fast navigation from a graph node to its dependencies.
+     */
+    private fun constructNodeDependencies(): NodeDependencies =
+        when {
+            nodes != null && edges != null -> constructNodeDependenciesFromGraph(nodes, edges)
+            else -> constructNodeDependenciesFromScopeRoots(scopeRoots)
+        }
 
     /**
      * Return a map of all de-duplicated [OrtIssue]s associated by [Identifier].
@@ -276,4 +311,100 @@ class DependencyReference(
         } else {
             fragment - other.fragment
         }
+}
+
+/**
+ * A data class representing a node in the dependency graph.
+ *
+ * A node corresponds to a package, which is referenced by a numeric index. A package may, however, occur multiple
+ * times in the dependency graph with different transitive dependencies. In this case, different fragment indices are
+ * used to distinguish between these occurrences.
+ */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+data class DependencyGraphNode(
+    /**
+     * Stores the numeric index of the package dependency referenced by this object. The package behind this index can
+     * be resolved by evaluating the list of identifiers stored in [DependencyGraph] at this index.
+     */
+    val pkg: Int,
+
+    /**
+     * Stores the index of the fragment in the dependency graph where the referenced dependency is contained. This is
+     * needed to uniquely identify the target if the dependency occurs multiple times in the graph.
+     */
+    val fragment: Int = 0,
+
+    /**
+     * The type of linkage used for the referred package from its dependent package. As most of our supported
+     * package managers / languages only support dynamic linking or at least default to it, also use that as the
+     * default value here to not blow up our result files.
+     */
+    @JsonInclude(value = JsonInclude.Include.CUSTOM, valueFilter = PackageLinkageValueFilter::class)
+    val linkage: PackageLinkage = PackageLinkage.DYNAMIC,
+
+    /**
+     * A list of [OrtIssue]s that occurred handling this dependency.
+     */
+    val issues: List<OrtIssue> = emptyList()
+)
+
+/**
+ * A data class representing an edge in the dependency graph.
+ *
+ * An edge corresponds to a directed depends-on relationship between two packages. The packages are identified by the
+ * numeric indices into the list of nodes.
+ */
+data class DependencyGraphEdge(
+    /** The index of the source node of this edge. */
+    val from: Int,
+
+    /** The index of the destination node of this edge. */
+    val to: Int
+)
+
+/**
+ * Convert this [DependencyReference] to a [DependencyGraphNode].
+ */
+private fun DependencyReference.toGraphNode() = DependencyGraphNode(pkg, fragment, linkage, issues)
+
+/**
+ * Construct a mapping of dependencies based on the given [roots].
+ */
+private fun constructNodeDependenciesFromScopeRoots(roots: SortedSet<DependencyReference>): NodeDependencies {
+    val mapping = mutableMapOf<DependencyGraphNode, List<DependencyGraphNode>>()
+
+    fun construct(refs: SortedSet<DependencyReference>) {
+        refs.forEach { ref ->
+            val node = ref.toGraphNode()
+            if (node !in mapping) {
+                mapping[node] = ref.dependencies.map(DependencyReference::toGraphNode)
+                construct(ref.dependencies)
+            }
+        }
+    }
+
+    construct(roots)
+    return mapping
+}
+
+/**
+ * Construct a mapping of dependencies based on the lists of graph [nodes] and [edges].
+ */
+private fun constructNodeDependenciesFromGraph(
+    nodes: List<DependencyGraphNode>,
+    edges: List<DependencyGraphEdge>
+): NodeDependencies {
+    val mapping = mutableMapOf<DependencyGraphNode, MutableList<DependencyGraphNode>>()
+
+    edges.forEach { edge ->
+        val srcNode = nodes[edge.from]
+        val dstNode = nodes[edge.to]
+        val dependencies = mapping.getOrPut(srcNode) { mutableListOf() }
+        dependencies += dstNode
+    }
+
+    // Add entries for nodes without dependencies.
+    nodes.filter { it !in mapping }.forEach { mapping[it] = mutableListOf() }
+
+    return mapping
 }
--- a/model/src/main/kotlin/DependencyGraph.kt
+++ b/model/src/main/kotlin/DependencyGraph.kt
@@ -41,11 +41,32 @@ typealias NodeDependencies = Map<DependencyGraphNode, List<DependencyGraphNode>>
  * using different dependencies in different scopes. Then the dependency graph is split into multiple fragments, and
  * each fragment has a consistent view on the dependencies it contains.
  *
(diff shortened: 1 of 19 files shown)


model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
```diff
@@ -0,0 +1,360 @@
...
+        "scopeNames" should {
+            "return the scope names of a project" {
+                navigator.scopeNames(testProject) should containExactlyInAnyOrder("compile", "test")
+            }
+        }
+
+        "directDependencies" should {
+            "return the direct dependencies of a project" {
+                navigator.directDependencies(testProject, "test")
+                    .map { it.id }.toList() should containExactly(
```
sschuberth, line 71: Nit: There should be no break between these lines to fix indentation.

sschuberth: You can also avoid the `.toList()` conversion by importing the specialized `io.kotest.matchers.sequences.containExactly` (eventually under a different name, if you also need `containExactly` for collections).

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
```diff
@@ -0,0 +1,360 @@
...
+                    Identifier("Maven:org.scalacheck:scalacheck_2.12:1.13.5"),
+                    Identifier("Maven:org.scalatest:scalatest_2.12:3.0.4")
+                )
+            }
+
+            "return an empty sequence for an unknown scope" {
+                navigator.directDependencies(testProject, "unknownScope").toList() should beEmpty()
```
sschuberth, line 78: Similar to my above comment, this could be

    navigator.directDependencies(testProject, "unknownScope") should beEmptySequence()

with

    import io.kotest.matchers.sequences.beEmpty as beEmptySequence

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
```diff
@@ -0,0 +1,360 @@
...
+
+                scopeDependencies.keys should containExactlyInAnyOrder("compile", "test")
+
+                val compileDependencies = scopeDependencies["compile"].orEmpty()
+                compileDependencies should haveSize(17)
+                compileDependencies should containAll(
+                    Identifier("Maven:com.typesafe.akka:akka-actor_2.12:2.5.6"),
+                    Identifier("Maven:org.scala-lang:scala-reflect:2.12.2"),
+                    Identifier("Maven:org.scala-lang:scala-library:2.12.3")
+                )
```
sschuberth, line 94: How about making the non-null check explicit via

                scopeDependencies["compile"].shouldNotBeNull {
                    this should haveSize(17)
                    this should containAll(
                        Identifier("Maven:com.typesafe.akka:akka-actor_2.12:2.5.6"),
                        Identifier("Maven:org.scala-lang:scala-reflect:2.12.2"),
                        Identifier("Maven:org.scala-lang:scala-library:2.12.3")
                    )
                }

Similar in more places below.

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
```diff
@@ -0,0 +1,360 @@
...
+            }
+        }
+
+        "getShortestPaths" should {
+            "return the shortest paths for a project" {
+                val paths = navigator.getShortestPaths(testProject)
+
+                paths.keys should haveSize(2)
+
+                paths["compile"]!! should org.ossreviewtoolkit.utils.test.containExactly(
```
sschuberth, line 228: Can we make the non-null check explicit here?

sschuberth: And maybe globally import `org.ossreviewtoolkit.utils.test.containExactly` under a different name rather than using the FQN here?

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
```diff
@@ -0,0 +1,360 @@
...
+            }
+        }
+
+        "projectDependencies" should {
+            "return the dependencies of a project" {
+                val scopeDependencies = navigator.scopeDependencies(testProject)
+                val projectDependencies = navigator.projectDependencies(testProject)
+
+                projectDependencies should containExactlyInAnyOrder(
+                    scopeDependencies["compile"].orEmpty() + scopeDependencies["test"].orEmpty()
```
sschuberth, line 279: Reminder: Please consider avoiding the implicit non-null checks also here, and in other places using `orEmpty()`.

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
```diff
@@ -0,0 +1,360 @@
...
+                val dependencies = navigator.projectDependencies(testProject, 0)
+
+                dependencies should beEmpty()
+            }
+        }
+
+        "collectSubProjects" should {
+            "find all the sub projects of a project" {
+                val projectId = Identifier("SBT:com.pbassiner:multi1_2.12:0.1-SNAPSHOT")
+                val project = testResult.getProject(projectId)!!
```
sschuberth, line 311: Reminder: Please consider avoiding the implicit non-null checks also here, and in other places using `!!`.

model/src/main/kotlin/DependencyNavigator.kt
```diff
@@ -215,7 +215,7 @@ interface DependencyNavigator {
 
             val newParents = item.parents + item.pkgRef.id
             item.pkgRef.visitDependencies { dependencyNodes ->
-                dependencyNodes.forEach { node -> queue.offer(QueueItem(node, newParents)) }
+                dependencyNodes.forEach { node -> queue.offer(QueueItem(node.getStableReference(), newParents)) }
```
sschuberth, line 218: Just for my information, how would you generally identify such code spots where a call to `getStableReference()` is missing? Apparently, before this commit no test was broken, so we do not seem to have test coverage for this, or?

model/src/main/kotlin/DependencyGraph.kt
```diff
@@ -277,3 +312,100 @@ class DependencyReference(
...
+
+    /** The index of the destination node of this edge. */
+    val to: Int
+)
+
+/**
+ * Convert this [DependencyReference] to a [DependencyGraphNode].
+ */
+private fun DependencyReference.toGraphNode(): DependencyGraphNode =
+    DependencyGraphNode(pkg, fragment, linkage, issues)
```
sschuberth, line 390: Nit: Could be a one-liner with the explicit return type omitted.

model/src/main/kotlin/DependencyGraphNavigator.kt
```diff
@@ -0,0 +1,245 @@
...
+     * Check for equality with [other]. The check is implemented as a reference comparison to the current node. As the
+     * nodes in the dependency graph have been de-duplicated, a reference comparison is sufficient. Note that this
+     * implementation is not necessarily consistent when comparing nodes during an iteration; but it works correctly
+     * for stable references.
+     */
+    override fun equals(other: Any?): Boolean =
+        when (other) {
+            is DependencyRefCursor -> current == other.current
+            else -> false
+        }
```
sschuberth, line 179: Maybe just `(other as? DependencyRefCursor)?.current == current`?

model/src/main/kotlin/DependencyGraphNavigator.kt
```diff
@@ -0,0 +1,245 @@
...
+ * Generate a sequence that allows accessing the given [dependencies] from the provided [graph] via the
+ * [DependencyNode] interface.
+ */
+private fun dependenciesSequence(
+    graph: DependencyGraph,
+    dependencies: Collection<DependencyGraphNode>
+): Sequence<DependencyNode> =
+    dependencies.takeIf { it.isNotEmpty() }?.let {
+        DependencyRefCursor(graph, it).asSequence()
+    } ?: emptySequence()
```
sschuberth, line 224: Just use `.orEmpty()`.

model/src/main/kotlin/DependencyGraphNavigator.kt
```diff
@@ -0,0 +1,245 @@
...
+    nodes: Sequence<DependencyNode>,
+    maxDepth: Int,
+    matcher: DependencyMatcher,
+    ids: MutableSet<Identifier>
+) {
+    if (maxDepth != 0) {
+        nodes.forEach { node ->
+            if (matcher(node)) {
+                ids += node.id
+            }
```
sschuberth, line 240: Nit: Could be one-liner.

model/src/test/kotlin/DependencyGraphNavigatorTest.kt
```diff
@@ -0,0 +1,95 @@
...
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
+
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.utils.Environment
+
+class DependencyGraphNavigatorTest : AbstractDependencyNavigatorTest() {
+    override val resultFileName: String = RESULT_FILE
+
+    override val resultWithIssuesFileName: String = RESULT_WITH_ISSUES_FILE
```
sschuberth, line 32: Nit: Could omit the explicit types.

sschuberth (dismissed):

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
oheger-bosch: I tried this, but then, surprisingly, the test failed. After a longer investigation, I think the problem is in the way the Kotest matcher compares sequences. The sequence with dependencies is constructed using `generateSequence()` and backed by a function that calculates the next element. It can therefore be traversed only once (which I would assume is true for most of the sequences generated this way).

The sequence matcher, however, first determines the count of the sequence (which requires a traversal) and afterwards checks the elements (another traversal). This second traversal no longer returns the full sequence, causing the comparison to fail.

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
oheger-bosch: Done, but for this specific place, is there a more elegant solution than nesting `shouldNotBeNull` blocks?

sschuberth: I don't think so... maybe @mnonnenmacher has an idea?

model/src/main/kotlin/DependencyNavigator.kt
oheger-bosch: Before this commit, the only `getStableReference()` implementation existing was for `PackageReference`, which returned itself. With `DependencyGraphNavigator`, there is now a new implementation, and this one would break tests.

That said, I am not very happy myself with this `getStableReference()` function and the necessity to call it explicitly; this is quite easy to forget. The main reason to introduce it was because of the `RuleSet` class, which keeps track of the `DependencyNode`s that are ancestors of the current package passed to a dependency rule. Maybe, in a later refinement, it is possible to provide an implementation of the _Visitor_ pattern or something like that, that allows a controlled iteration over all dependencies and hides these details.

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
sschuberth: @sksamuel acknowledged it's an [issue](https://github.com/kotest/kotest/issues/2412).

sschuberth:
Can you please rebase due to recent (unrelated) test fixes?

oheger-bosch:
> 
> 
> Can you please rebase due to recent (unrelated) test fixes?

Done

model/src/main/kotlin/OrtResult.kt
```diff
@@ -372,7 +372,7 @@ data class OrtResult(
      */
     @JsonIgnore
     fun getProjects(omitExcluded: Boolean = false): Set<Project> =
-        analyzer?.result?.withScopesResolved()?.projects.orEmpty().filterTo(mutableSetOf()) { project ->
+        analyzer?.result?.projects.orEmpty().filterTo(mutableSetOf()) { project ->
```
mnonnenmacher, line 375: Commit message: It sounds like it was impossible to access projects if the ORT result is in dependency graph format, but I believe what you mean is that it is impossible to access projects without converting them to the scope format.

mnonnenmacher (changes requested):
I only had time for the first three commits today, I will review the remaining commits tomorrow.

model/src/test/assets/result-with-issues.yml
```diff
@@ -0,0 +1,1371 @@
+---
```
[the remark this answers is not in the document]
oheger-bosch: Renamed to `result-with-issues-scopes.yml`.

model/src/main/kotlin/OrtResult.kt
oheger-bosch: Reworded the commit message to make the intension (hopefully) clearer.

model/src/test/kotlin/AbstractDependencyNavigatorTest.kt
mnonnenmacher: You could write it like this to avoid nesting:
```
scopeDependencies should containKeys("compile", "test")
val expectedDependencies = scopeDependencies.getValue("compile") + scopeDependencies.getValue("test")
projectDependencies should containExactlyInAnyOrder(compileDependencies + testDependencies)
```

model/src/test/kotlin/DependencyGraphNavigatorLegacyTest.kt
```diff
@@ -0,0 +1,36 @@
...
+    override val resultFileName = RESULT_FILE
+
+    override val resultWithIssuesFileName = RESULT_WITH_ISSUES_FILE
+}
+
+/** Name of a file with a more complex ORT result that is used by multiple test cases. */
+private const val RESULT_FILE = "src/test/assets/sbt-multi-project-example-graph-old.yml"
+
+/** Name of the file with a result that contains some issues. */
+private const val RESULT_WITH_ISSUES_FILE = "src/test/assets/result-with-issues-graph-old.yml"
```
mnonnenmacher, line 36: Why not assign the values directly to the overridden variables above?

model/src/test/kotlin/DependencyGraphNavigatorTest.kt
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+    }
+}
+
+/** Name of a file with a more complex ORT result that is used by multiple test cases. */
+private const val RESULT_FILE = "src/test/assets/sbt-multi-project-example-graph.yml"
+
+/** Name of the file with a result that contains some issues. */
+private const val RESULT_WITH_ISSUES_FILE = "src/test/assets/result-with-issues-graph.yml"
```
mnonnenmacher, line 95: Why not assign the values directly to the overridden variables above?

model/src/test/kotlin/DependencyGraphNavigatorTest.kt
```diff
@@ -0,0 +1,95 @@
...
+                    )
+                )
+
+                shouldThrow<IllegalArgumentException> {
+                    DependencyGraphNavigator(result)
+                }
+            }
+        }
+
+        "DependencyNode.equals" should {
```
mnonnenmacher, line 57: What is the reason not to add these test cases to the abstract class?

mnonnenmacher (changes requested):

model/src/main/kotlin/DependencyGraph.kt
```diff
@@ -41,25 +47,40 @@ import java.util.SortedSet
...
 
     /**
      * Stores the dependency graph as a list of root nodes for the direct dependencies referenced by scopes. Starting
      * with these nodes, the whole graph can be traversed. The nodes are constructed from the direct dependencies
-     * declared by scopes that cannot be reached via other paths in the dependency graph.
+     * declared by scopes that cannot be reached via other paths in the dependency graph. Note that this property
+     * exists for backwards compatibility only; it is replaced by the lists of nodes and edges.
      */
-    val scopeRoots: SortedSet<DependencyReference>,
+    val scopeRoots: SortedSet<DependencyReference> = sortedSetOf(),
```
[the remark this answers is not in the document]
oheger-bosch: Using `@Deprecated` would probably be the best solution. Unfortunately, this breaks the build, because the field is still referenced in some places that now cause compiler warnings, and *detekt* is configured to treat warnings as errors.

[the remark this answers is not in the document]
sschuberth: > _detekt_ is configured to treat warnings as errors.

Nit: It's not about _detekt_, but our [Kotlin compiler options](https://github.com/oss-review-toolkit/ort/blob/cb118d8e306310cec16877804a02a95ac3862613/build.gradle.kts#L225).

model/src/main/kotlin/DependencyGraphNavigator.kt
```diff
@@ -0,0 +1,239 @@
...
+package org.ossreviewtoolkit.model
+
+/**
+ * A [DependencyNavigator] implementation based on the dependency graph format. It obtains the information about a
+ * project's dependencies from the shared [DependencyGraph] stored in the [OrtResult].
+ */
+class DependencyGraphNavigator(ortResult: OrtResult) : DependencyNavigator {
+    /** The map with shared dependency graphs from the associated result. */
+    private val graphs: Map<String, DependencyGraph> =
+        requireNotNull(ortResult.analyzer?.result?.dependencyGraphs?.takeIf { it.isNotEmpty() }) {
```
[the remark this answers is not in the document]
oheger-bosch: It is not clear to me how the proposed construct can be integrated with the assignment to `graphs`. `require()` returns `Unit`, and what is *this* in this context?

model/src/main/kotlin/DependencyGraphNavigator.kt
```diff
@@ -0,0 +1,239 @@
...
+     * A data structure allowing fast access to a specific [DependencyReference] based on its index and fragment.
+     */
+    private val graphDependencyRefMappings: Map<String, Array<MutableList<DependencyGraphNode>>> by lazy {
+        graphs.mapValues { it.value.dependencyRefMapping() }
+    }
+
+    override fun scopeNames(project: Project): Set<String> = project.scopeNames.orEmpty()
+
+    override fun directDependencies(project: Project, scopeName: String): Sequence<DependencyNode> {
+        val manager = project.managerName
```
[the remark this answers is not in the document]
oheger-bosch: Inlined.

model/src/main/kotlin/DependencyGraphNavigator.kt
```diff
@@ -0,0 +1,239 @@
...
+ * Traverse the given [nodes] recursively up to the given [maxDepth], filter using [matcher] and adds all identifiers
+ * found to [ids].
+ */
+private fun collectDependencies(
+    nodes: Sequence<DependencyNode>,
+    maxDepth: Int,
+    matcher: DependencyMatcher,
+    ids: MutableSet<Identifier>
+) {
+    if (maxDepth != 0) {
```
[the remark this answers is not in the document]
oheger-bosch: The test for `!= 0` is actually required to handle `maxDepth < 0` in the meaning of "no depth limit" correctly.

model/src/test/kotlin/DependencyGraphNavigatorTest.kt
oheger-bosch: They have originally been written specifically for the `equals()` implementation of the internal `DependencyRefCursor` class; but you are right, the tests can as well be moved to the base class to check `equals()` for all `DependencyNode` implementations. Done.

model/src/main/kotlin/DependencyGraph.kt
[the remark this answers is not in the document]
mnonnenmacher: Ok, then let's not add `@Deprecated`.

model/src/main/kotlin/DependencyGraphNavigator.kt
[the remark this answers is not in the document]
mnonnenmacher: Sorry, I didn't realize that the return value of `requireNotNull` was actually used.

mnonnenmacher (approved):

PR #4248, opened 2021-07-01
Package provenance resolver

mnonnenmacher:
The first component for a new experimental scanner, see the commit messages for details.

### DIFF
--- a/scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
+++ b/scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
@@ -1,0 +1,144 @@
+/*
+ * Copyright (C) 2021 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.scanner.experimental
+
+import java.io.IOException
+import java.net.HttpURLConnection
+
+import okhttp3.Request
+
+import org.ossreviewtoolkit.downloader.VersionControlSystem
+import org.ossreviewtoolkit.model.ArtifactProvenance
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.RemoteArtifact
+import org.ossreviewtoolkit.model.RepositoryProvenance
+import org.ossreviewtoolkit.model.SourceCodeOrigin
+import org.ossreviewtoolkit.model.UnknownProvenance
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.utils.OkHttpClientHelper
+import org.ossreviewtoolkit.utils.collectMessagesAsString
+import org.ossreviewtoolkit.utils.createOrtTempDir
+import org.ossreviewtoolkit.utils.log
+import org.ossreviewtoolkit.utils.showStackTrace
+
+/**
+ * The [PackageProvenanceResolver] provides a function to resolve the [Provenance] of a [Package].
+ */
+interface PackageProvenanceResolver {
+    /**
+     * Resolve the [Provenance] of [pkg] based on the provided [sourceCodeOriginPriority]. If no provenance can be
+     * resolved [UnknownProvenance] is returned.
+     */
+    fun resolveProvenance(pkg: Package, sourceCodeOriginPriority: List<SourceCodeOrigin>): Provenance
+}
+
+/**
+ * The default implementation of [PackageProvenanceResolver].
+ */
+class DefaultPackageProvenanceResolver : PackageProvenanceResolver {
+    /**
+     * Resolve the [Provenance] of [pkg] based on the provided [sourceCodeOriginPriority]. For source artifacts it is
+     * verified that the [RemoteArtifact] does exist. For a VCS it is verified that the revision exists. If the revision
+     * provided by the [package][pkg] metadata does not exist or is missing, the function tries to guess the tag based
+     * on the name and version of the [package][pkg].
+     */
+    override fun resolveProvenance(pkg: Package, sourceCodeOriginPriority: List<SourceCodeOrigin>): Provenance {
+        sourceCodeOriginPriority.forEach { sourceCodeOrigin ->
+            runCatching {
+                when (sourceCodeOrigin) {
+                    SourceCodeOrigin.ARTIFACT -> {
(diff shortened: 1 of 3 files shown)


scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
```diff
@@ -0,0 +1,135 @@
...
+ */
+class DefaultPackageProvenanceResolver : PackageProvenanceResolver {
+    /**
+     * Resolve the [Provenance] of [pkg] based on the provided [sourceCodeOriginPriority]. For source artifacts it is
+     * verified that the [RemoteArtifact] does exist. For a VCS it is verified that the revision exists. If the revision
+     * provided by the [package][pkg] metadata does not exist or is missing, the function tries to guess the tag based
+     * on the name and version of the [package][pkg].
+     */
+    override fun resolveProvenance(pkg: Package, sourceCodeOriginPriority: List<SourceCodeOrigin>): Provenance {
+        sourceCodeOriginPriority.distinct().forEach { sourceCodeOrigin ->
```
sschuberth, line 64: Should we make `sourceCodeOriginPriority` a `Set` to be able to spare the `distinct()` here?

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
```diff
@@ -0,0 +1,135 @@
...
+            kotlin.runCatching {
```
sschuberth, line 65: Please drop the `kotlin.`.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
```diff
@@ -0,0 +1,135 @@
...
+import org.ossreviewtoolkit.utils.OkHttpClientHelper
+import org.ossreviewtoolkit.utils.collectMessagesAsString
+import org.ossreviewtoolkit.utils.createOrtTempDir
+import org.ossreviewtoolkit.utils.log
+import org.ossreviewtoolkit.utils.showStackTrace
+
+/**
+ * The [PackageProvenanceResolver] provides a function to resolve the [Provenance] of a [Package].
+ */
+interface PackageProvenanceResolver {
```
sschuberth, line 45: I'm having a bit of a problem with the term "resolve" here. To me, "resolve" implies working on something that *is* already there, but gets "transformed" in some way (like resolving a link to the file pointed to). However, what I believe this does, is to *determine* the provenance. Or even "create" / "calculate" the provenance.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
```diff
@@ -0,0 +1,135 @@
...
+                return ArtifactProvenance(sourceArtifact)
+            }
+
+            throw IOException("Could not verify existence of source artifact at ${sourceArtifact.url}.")
+        }
+    }
+
+    private fun resolveVcs(pkg: Package): RepositoryProvenance {
+        // TODO: Is it right to always determine type from URL?
+        val vcs = VersionControlSystem.forUrl(pkg.vcsProcessed.url)
```
sschuberth, line 104: Maybe do `?: throw IOException("Could not determine VCS type for URL ${pkg.vcsProcessed}.")` here to avoid nesting below?

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
mnonnenmacher: IMO the term "resolve" is okay, but others would work as well. Based on your interpretation, it does take something which is already there, namely the package metadata, and transforms it into a provenance. For example the VCS info gets amended with a resolved revision.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
mnonnenmacher: No, because the order defines the priority, and the iteration order of a `Set` depends on the implementation.

sschuberth: Then maybe generalize to use a `Collection` so that *also* sets can be passed in (as practically sets in Kotlin are created via `mutableSetOf()`, which creates a set that maintains the iteration order)? And then I'd also omit the `.distinct()` (which internally also just converts to a mutable set and then back to a list) and leave it to the caller to not pass in any duplicates.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
sschuberth: What I mean is: In "resolve" there is a 1:1 relationship between input and output, you always resolve *to* the same thing. But here, you determine the provenance from metadata, and what the provenance *is* changes depending on the `sourceCodeOriginPriority`. I'd strongly prefer another term, also to avoid the confusion with "resolving revisions" similar to like we had before in the `VcsInfo` context; otherwise one could think this somehow resolves the (symbolic) revision of a `RepositoryProvenance`.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
mnonnenmacher: I don't understand why this `distinct()` call is bothering you so much. I find it important that the function requires the user to provide the sources sorted by priority. With a set or collection it is possible to accidentally pass the wrong order, e.g. by using the wrong set implementation, with a list not. And why leave it to the caller to make the mistake to pass duplicates if it can be prevented with a simple `distinct()` call?

Btw, the config setting which is what will be passed to this function later on is a list already:
https://github.com/oss-review-toolkit/ort/blob/fd32da954a63ae7ceaf0e1b8d81af7e8718682e5/model/src/main/kotlin/config/DownloaderConfiguration.kt#L33-L36

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
mnonnenmacher: Ok, I'd suggest one of those solutions:
* Please suggest a specific name and I'll change it.
* We'll leave the naming discussion for later.
Just because I don't want to get stuck on a naming discussion right now. My current plan is to try to get a working first version of the new scanner ready before my holiday and then maybe after the holiday we could make a extensive review session for the concept and implementation.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
sschuberth: > I don't understand why this `distinct()` call is bothering you so much

Most because its implementation is to inelegant, with its crude conversion to a set and back to a list, creating temporary collections.



> And why leave it to the caller to make the mistake to pass duplicates if it can be prevented with a simple `distinct()` call?

Because passing duplicates actually is no mistake, or? What matters here is the relative order of *different* items in the list; specifying the same item multiple times should not change the outcome, but it's just confusing.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
sschuberth: I was giving the naming some thought again, but couldn't come up with a better name. And as the existence *is* actually checked (e.g. via a `HEAD` request in case of a source artifact), the artifact *is* actually implicitly resolved. So let's keep the naming as-is then for now.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
mnonnenmacher: I have removed the `distinct()` call but kept the signature.

sschuberth (approved):

PR #4250, opened 2021-07-02
helper-cli: Support placeholders when changing the dependency format

oheger-bosch:
One use case for SetDependencyRepresentationCommand is to convert the
YAML files with expected results of ORT's functional tests. This is
currently not possible, however, because most of these files contain
special placeholders that are dynamically replaced during test
execution and that cause the deserialization to result objects to
fail.

Therefore, add logic to handle such placeholders. If enabled via a new
command line flag, critical placeholders are replaced by allowed
values before deserialization. When writing the result of the
conversion they are restored again.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
@@ -23,10 +23,13 @@ import com.github.ajalt.clikt.core.CliktCommand
 import com.github.ajalt.clikt.core.UsageError
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.enum
 import com.github.ajalt.clikt.parameters.types.file
+
+import java.util.regex.Pattern
 
 import org.ossreviewtoolkit.analyzer.AnalyzerResultBuilder
 import org.ossreviewtoolkit.model.AnalyzerResult
@@ -35,6 +38,7 @@ import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.model.utils.DependencyGraphConverter
 import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.model.yamlMapper
 import org.ossreviewtoolkit.utils.expandTilde
 
 /**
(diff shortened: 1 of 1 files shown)


helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -86,6 +114,11 @@ class SetDependencyRepresentationCommand : CliktCommand(
         help = "The target format for the conversion."
     ).enum<TargetFormat>().default(TargetFormat.GRAPH)
 
+    private val handlePlaceholders by option(
+        "--placeholders", "-p",
+        help = "Handles special placeholders in the input file that are used by ORT functional tests."
+    ).flag("--no-placeholders", "-P", default = false)
```
sschuberth, line 120: Nit: The default default (😁) value is `false`, you could drop this.

helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -64,6 +72,26 @@ class SetDependencyRepresentationCommand : CliktCommand(
...
+        val placeholder: String,
+
+        /** The value to insert for the placeholder to make the input file valid YAML. */
+        val replacement: String
+    )
+
+    /** A list of [PlaceholderInfo] objects that are handled during processing. */
+    private val placeholders = listOf(
+        PlaceholderInfo("processors", "REPLACE_PROCESSORS", "11"),
+        PlaceholderInfo("max_memory", "REPLACE_MAX_MEMORY", "1234")
```
sschuberth, line 92: Why are these not in here?

REPLACE_REVISION, REPLACE_URL, REPLACE_URL_PROCESSED, REPLACE_RAW_URL, REPLACE_JAVA, REPLACE_OS, REPLACE_PATH, REPLACE_DEFINITION_FILE_PATH, REPLACE_GITHUB_PROJECT

oheger-bosch: Those are assigned to string values and do not cause any problems during conversion. The only critical placeholders are the ones assigned to numeric properties; they cause the serialization to their Kotlin representation to fail.

sschuberth: Ah, good hint. Can we get that information as a code comment, and / or as part of the variable name, like `numericPlaceholders`?

oheger-bosch: Yes, makes sense. I would, however, not use the name `numericPlaceholders`. That all of these placeholders are numeric may not always hold. There are for instance placeholders for URL values, and if sometimes some validation logic is implemented on such properties, they would have to be treated in a special way, too.

So maybe something like `problematicPlaceholders` or `criticalPlaceholders`? Or simply `placeholdersToReplace`?

sschuberth: How about "typedPlaceholders" as the replacement is done based on whether there is some kind of type checking on the values? And in the code comment explain further what this refers to?

I'm just trying to find a name that further hints at in *what way* these placeholders are "problematic" or "critical".

oheger-bosch: Renamed the variable and added more details to the comment.

sschuberth (approved):

PR #4252, opened 2021-07-02
DependencyRule: Improve level-related documentation

sschuberth:
Avoid the term "root", as the root of a dependency tree usually is the
project itself. However, here "level 0" refers to the direct
dependencies of a project.

### DIFF
--- a/evaluator/src/main/kotlin/DependencyRule.kt
+++ b/evaluator/src/main/kotlin/DependencyRule.kt
@@ -45,13 +45,14 @@ class DependencyRule(
     val dependency: PackageReference,
 
     /**
-     * The ancestors of the [dependency] in the dependency tree. The first entry is the root of the tree, the last entry
-     * (at [level] - 1) is the direct parent.
+     * The ancestors of the [dependency] in the dependency tree, sorted from farthest to closest: The first entry is the
+     * direct dependency of a project, the last entry (at index `level - 1`) is a direct parent of this [dependency].
+     * If the list is empty it means that this dependency is a direct dependency.
      */
     val ancestors: List<PackageReference>,
 
     /**
-     * The level of the [dependency] inside the dependency tree. Starts with 0 for root level entries.
+     * The level of the [dependency] inside the dependency tree. Starts with 0 for a direct dependency of a project.
      */
     val level: Int,
 
--- a/examples/rules.kts
+++ b/examples/rules.kts
@@ -229,7 +229,7 @@ val ruleSet = ruleSet(ortResult, licenseInfoResolver) {
 
     dependencyRule("COPYLEFT_LIMITED_STATIC_LINK_IN_DIRECT_DEPENDENCY") {
         require {
-            +isAtTreeLevel(1)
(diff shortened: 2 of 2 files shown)


sschuberth:
@mmurto, does this make it clear enough?

evaluator/src/main/kotlin/DependencyRule.kt
```diff
@@ -45,13 +45,13 @@ class DependencyRule(
     val dependency: PackageReference,
 
     /**
-     * The ancestors of the [dependency] in the dependency tree. The first entry is the root of the tree, the last entry
-     * (at [level] - 1) is the direct parent.
+     * The ancestors of the [dependency] in the dependency tree, sorted from farthest to closest: The first entry is the
+     * direct dependency of a project, the last entry (at index [level] - 1) is the direct parent of this [dependency].
```
fviernau, line 49: How about removing "(at index [level] - 1)" as "the last entry" already describes it sufficiently?

fviernau (approved):

evaluator/src/main/kotlin/DependencyRule.kt
sschuberth: I liked it, as it emphasizes that the number of entries in `ancestors` equals the `level`.

mmurto:
I think this should also be corrected to `isAtTreeLevel(0)`.

https://github.com/oss-review-toolkit/ort/blob/fd32da954a63ae7ceaf0e1b8d81af7e8718682e5/examples/rules.kts#L230-L234

mnonnenmacher:
> I think this should also be corrected to `isAtTreeLevel(0)`.
> 
> https://github.com/oss-review-toolkit/ort/blob/fd32da954a63ae7ceaf0e1b8d81af7e8718682e5/examples/rules.kts#L230-L234

@sschuberth Could you add this fix to your PR?

evaluator/src/main/kotlin/DependencyRule.kt
mnonnenmacher: Maybe write it as "(at index `level - 1`)".

mnonnenmacher: You could make it even clearer by adding: "If the list is empty it means that this dependency is a direct dependency."

evaluator/src/main/kotlin/DependencyRule.kt
```diff
@@ -45,13 +45,13 @@ class DependencyRule(
...
      */
     val ancestors: List<PackageReference>,
 
     /**
-     * The level of the [dependency] inside the dependency tree. Starts with 0 for root level entries.
+     * The level of the [dependency] inside the dependency tree. Starts with 0 for the direct dependency of a project.
```
mnonnenmacher, line 54: Replace "the" with "a", otherwise it sounds like there can only be one direct dependency.

mnonnenmacher (changes requested):
See above...

mnonnenmacher (approved):

sschuberth:

PR #4256, opened 2021-07-02
Make use of GRADLE_OPTS

sschuberth:

### DIFF
--- a/.azure-pipelines/LinuxAnalyzerTest.yml
+++ b/.azure-pipelines/LinuxAnalyzerTest.yml
@@ -140,10 +140,6 @@ jobs:
       javaHomeOption: JDKVersion
       jdkVersionOption: 1.11
 
-  # Ensure that any running Gradle daemon is stopped before caching starts.
-  - bash: ./gradlew --stop
-    displayName: Stop Gradle Daemon
-
   - script: bash <(curl -s https://codecov.io/bash) -X gcov -f '!*/src/*/assets/*';
     displayName: Upload to codecov.io
     condition: eq(variables['Build.Reason'], 'Schedule')
--- a/.azure-pipelines/LinuxAnalyzerTest.yml
+++ b/.azure-pipelines/LinuxAnalyzerTest.yml
@@ -3,7 +3,7 @@ jobs:
   pool:
     vmImage: ubuntu-18.04
   variables:
-    GRADLE_OPTS: -Dorg.gradle.daemon=false
+    GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.vfs.watch=false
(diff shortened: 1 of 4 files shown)


sschuberth:
Hmm, we [still getting](https://dev.azure.com/oss-review-toolkit/ort/_build/results?buildId=4927&view=logs&j=0b98c9ad-b4dd-509f-97af-ff0d5e511683&t=d4a109d9-63b6-59b7-7349-067146a0a444&l=81) "Unable to watch the file system for changes. The inotify watches limit is too low." despite ``.

mnonnenmacher:
> Hmm, we [still getting](https://dev.azure.com/oss-review-toolkit/ort/_build/results?buildId=4927&view=logs&j=0b98c9ad-b4dd-509f-97af-ff0d5e511683&t=d4a109d9-63b6-59b7-7349-067146a0a444&l=81) "Unable to watch the file system for changes. The inotify watches limit is too low." despite `-Dorg.gradle.vfs.watch=false`.

It looks like `GRADLE_OPTS` is not working as expected because it also prints this line at the beginning:
```
Starting a Gradle Daemon (subsequent builds will be faster)
```

mnonnenmacher:
Maybe you have to set the `gradleOpts` option of the Gradle tasks like this:
```
gradleOptions: ${{ variables.GRADLE_OPTS }}
```
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/gradle?view=azure-devops#arguments

sschuberth:
> Maybe you have to set the `gradleOpts` option of the Gradle tasks like this:

Hmm, thanks for the hint, but I'm but sure if it's even worth a try, as that would defeat the purpose to not touch every Gradle task call to pass common options.

sschuberth:
> It looks like `GRADLE_OPTS` is not working as expected because it also prints this line at the beginning:
> 
> ```
> Starting a Gradle Daemon (subsequent builds will be faster)
> ```

I also find the Gradle docs in that regard confusing. Some sources claim that even with `--no-daemon` the daemon *is* actually started, but shut down again immediately after the build. Other sources indicate that really no daemon is started...

mnonnenmacher:
> > Maybe you have to set the `gradleOpts` option of the Gradle tasks like this:
> 
> Hmm, thanks for the hint, but I'm but sure if it's even worth a try, as that would defeat the purpose to not touch every Gradle task call to pass common options.

Maybe it could be set in `.azure-pipelines.yml`, then there would at least only one central place to define those options.

sschuberth:
This is obsolete now that CI tests are not run in Azure Pipelines anymore as of https://github.com/oss-review-toolkit/ort/pull/5031.

PR #4257, opened 2021-07-04
scanner: Add a NestedProvenanceResolver

mnonnenmacher:
The `ProvenanceTreeResolver` is used to resolve nested repositories in a
`RepositoryProvenance`. It will be used by a new experimental scanner
implementation in a later commit, therefore it is also located in the
"experimental" package for now.

### DIFF
--- a/scanner/src/funTest/kotlin/experimental/DefaultNestedProvenanceResolverFunTest.kt
+++ b/scanner/src/funTest/kotlin/experimental/DefaultNestedProvenanceResolverFunTest.kt
@@ -1,0 +1,202 @@
+/*
+ * Copyright (C) 2021 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.scanner.experimental
+
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldBe
+
+import java.io.IOException
+
+import org.ossreviewtoolkit.model.ArtifactProvenance
+import org.ossreviewtoolkit.model.Hash
+import org.ossreviewtoolkit.model.RemoteArtifact
+import org.ossreviewtoolkit.model.RepositoryProvenance
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.utils.Os
+import org.ossreviewtoolkit.utils.test.containExactly
+
+class DefaultNestedProvenanceResolverFunTest : WordSpec() {
+    private val resolver = DefaultNestedProvenanceResolver()
+
+    init {
+        "Resolving an artifact provenance" should {
+            "Return only the artifact provenance" {
+                val provenance = ArtifactProvenance(
+                    sourceArtifact = RemoteArtifact(
+                        url = "",
+                        hash = Hash.NONE
+                    )
+                )
+
+                resolver.resolveNestedProvenance(provenance) shouldBe
+                        NestedProvenance(root = provenance, subRepositories = emptyMap())
+            }
+        }
+
+        "Resolving a repository provenance" should {
+            "Return the correct root provenance" {
+                val provenance = RepositoryProvenance(
+                    vcsInfo = VcsInfo(
+                        type = VcsType.GIT,
+                        url = "https://github.com/oss-review-toolkit/ort-governance.git",
+                        revision = "1440358978888bfc4b6fbf356e4e6be005c9a25e"
+                    ),
+                    resolvedRevision = "1440358978888bfc4b6fbf356e4e6be005c9a25e"
+                )
+
+                resolver.resolveNestedProvenance(provenance) shouldBe
+                        NestedProvenance(root = provenance, subRepositories = emptyMap())
+            }
+
+            "Find recursive Git submodules" {
+                val provenance = RepositoryProvenance(
+                    vcsInfo = VcsInfo(
+                        type = VcsType.GIT,
+                        url = "https://github.com/oss-review-toolkit/ort-test-data-git-submodules.git",
+                        revision = "fcea94bab5835172e826afddb9f6427274c983b9"
+                    ),
+                    resolvedRevision = "fcea94bab5835172e826afddb9f6427274c983b9"
+                )
+
+                val nestedProvenance = resolver.resolveNestedProvenance(provenance)
+
+                nestedProvenance.root shouldBe provenance
+                nestedProvenance.subRepositories should containExactly(
+                    "commons-text" to RepositoryProvenance(
+                        vcsInfo = VcsInfo(
+                            type = VcsType.GIT,
+                            url = "https://github.com/apache/commons-text.git",
+                            revision = "7643b12421100d29fd2b78053e77bcb04a251b2e"
+                        ),
+                        resolvedRevision = "7643b12421100d29fd2b78053e77bcb04a251b2e"
+                    ),
+                    "test-data-npm" to RepositoryProvenance(
+                        vcsInfo = VcsInfo(
+                            type = VcsType.GIT,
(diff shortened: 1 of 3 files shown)


scanner/src/main/kotlin/experimental/ProvenanceTree.kt
```diff
@@ -0,0 +1,40 @@
...
+
+package org.ossreviewtoolkit.scanner.experimental
+
+import org.ossreviewtoolkit.model.KnownProvenance
+import org.ossreviewtoolkit.model.RepositoryProvenance
+
+/**
+ * This class contains information about a [root] provenance and all nested [subRepositories].
+ */
+data class ProvenanceTree(
```
sschuberth, line 28: Sorry to go into a naming discussion again, but as this only applies to `RepositoryProvenance`, not to `ArtifactProvenance`, and as in the context of a repository the term "tree" is commonly used for the file tree, I'd propose a name like `NestedRepositoryProvenance`. That would mach the `nestedRepositories` from `Repository`.

scanner/src/main/kotlin/experimental/ProvenanceTree.kt
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * The root provenance.
```
sschuberth, line 30: What *is* a "root provenance" (I assume it's the super-project), and why is it just `KnownProvenance` instead of `RepositoryProvenance` (as there probably cannot be a thing like a `ArtifactProvenance` that refers to multiple child `RepositoryProvenance`)?

scanner/src/main/kotlin/experimental/ProvenanceTreeResolver.kt
```diff
@@ -0,0 +1,74 @@
...
+            ?: VersionControlSystem.forUrl(provenance.vcsInfo.url)
+            ?: throw IOException("Could not determine VCS type for ${provenance.vcsInfo}.")
+
+        val targetDir = createOrtTempDir()
+        val vcsInfo = provenance.vcsInfo.copy(revision = provenance.resolvedRevision)
+
+        val workingTree = vcs.initWorkingTree(targetDir, vcsInfo)
+
+        val result = vcs.updateWorkingTree(workingTree, vcsInfo.revision, path = vcsInfo.path, recursive = true)
+        result.onFailure { throw it }
```
sschuberth, line 65: Nit: If you don't use `result` otherwise, you can just chain the calls and get rid of the variable.

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/ProvenanceTree.kt
mnonnenmacher: I went with `NestedProvenace` (see below).

scanner/src/main/kotlin/experimental/ProvenanceTree.kt
mnonnenmacher: It's the provenance that contains the nested provenances, I have added this to the comment. The reason that root is a `KnownProvenance` is because this makes the code using a lot simpler because it only has to deal with a single class. It's documented in the `subRepositories` property that it is always empty if `root` is an `ArtifactProvenance`.

scanner/src/main/kotlin/experimental/NestedProvenanceResolver.kt
```diff
@@ -0,0 +1,74 @@
...
+ */
+class DefaultNestedProvenanceResolver : NestedProvenanceResolver {
+    override fun resolveProvenanceTree(provenance: KnownProvenance): NestedProvenance {
+        return when (provenance) {
+            is ArtifactProvenance -> NestedProvenance(root = provenance, subRepositories = emptyMap())
+            is RepositoryProvenance -> resolveVcsSourceTree(provenance)
+        }
+    }
+
+    private fun resolveVcsSourceTree(provenance: RepositoryProvenance): NestedProvenance {
```
sschuberth, line 54: I'd prefer to also use the term "nested" here, to avoid any confusion with the working*Tree* below.

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/NestedProvenanceResolver.kt
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * The [NestedProvenanceResolver] provides a function to resolve nested provenances.
+ */
+interface NestedProvenanceResolver {
+    /**
+     * Resolve nested [Provenance]s of the provided [provenance]. For an [ArtifactProvenance] the returned
+     * [NestedProvenance] always contains only the provided [ArtifactProvenance]. For a [RepositoryProvenance] the
+     * resolver looks for nested repositories, for example Git submodules or Mercurial subrepositories.
+     */
+    fun resolveProvenanceTree(provenance: KnownProvenance): NestedProvenance
```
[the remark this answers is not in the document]
mnonnenmacher: I just forgot to rename these functions, I have called it `resolvedNestedProvenance` now.

scanner/src/main/kotlin/experimental/NestedProvenanceResolver.kt
```diff
@@ -0,0 +1,74 @@
...
+            is RepositoryProvenance -> resolveNestedRepository(provenance)
+        }
+    }
+
+    private fun resolveNestedRepository(provenance: RepositoryProvenance): NestedProvenance {
+        val vcs = VersionControlSystem.forType(provenance.vcsInfo.type)
+            ?: VersionControlSystem.forUrl(provenance.vcsInfo.url)
+            ?: throw IOException("Could not determine VCS type for ${provenance.vcsInfo}.")
+
+        val targetDir = createOrtTempDir()
```
sschuberth, line 59: Looks like this call again causes

```
error: Cannot fetch oss-review-toolkit/ort-test-data-git-submodules\test-data-npm\isarray (GitError: oss-review-toolkit/ort-test-data-git-submodules\test-data-npm\isarray init: C:/Users/VssAdministrator/AppData/Local/Temp/ort-DefaultNestedProvenanceResolver4660852545205957772/.repo/projects/submodules.git/subprojects/test-data-npm.git/subproject-objects/oss-review-toolkit/ort-test-data-git-submodules/test-data-npm/isarray.git: Filename too long
```

on Windows. If you're lucky, you can work around it like in

https://github.com/oss-review-toolkit/ort/blob/a754d7bfc2e4687a884d62d0f516a23c83b3e38a/analyzer/src/funTest/kotlin/GitRepoFunTest.kt#L52-L54

mnonnenmacher: Thanks, for `Git` we have:
https://github.com/oss-review-toolkit/ort/blob/a754d7bfc2e4687a884d62d0f516a23c83b3e38a/downloader/src/main/kotlin/vcs/Git.kt#L144-L146

Wouldn't it be an option to do the same for `GitRepo`?

sschuberth: I tried doing this in https://github.com/oss-review-toolkit/ort/pull/4024, but as `repo` is written in Python, somewhere in the JVM -> ORT - Python -> repo process chain not all environment variables seem to get passes on as expected, and also another approach via setting / restoring `git config` did not succeed initially, and then I ran out of time to try further.

mnonnenmacher: Ok, I decided to disable the one failing test on Windows and leave the problem for later, because the workaround didn't work.

scanner/src/funTest/kotlin/experimental/DefaultNestedProvenanceResolverFunTest.kt
```diff
@@ -0,0 +1,202 @@
...
+                val provenance = RepositoryProvenance(
+                    vcsInfo = VcsInfo(
+                        type = VcsType.GIT,
+                        url = "https://github.com/oss-review-toolkit/ort-test-data-git-submodules.git",
+                        revision = "fcea94bab5835172e826afddb9f6427274c983b9"
+                    ),
+                    resolvedRevision = "fcea94bab5835172e826afddb9f6427274c983b9"
+                )
+
+                val tree = resolver.resolveNestedProvenance(provenance)
```
sschuberth, line 81: This is still called "tree", not sure if it's intentional. Could be e.g. just "repo".

scanner/src/funTest/kotlin/experimental/DefaultNestedProvenanceResolverFunTest.kt
```diff
@@ -0,0 +1,202 @@
...
+                    vcsInfo = VcsInfo(
+                        type = VcsType.GIT_REPO,
+                        url = "https://github.com/oss-review-toolkit/ort-test-data-git-repo.git",
+                        revision = "31588aa8f8555474e1c3c66a359ec99e4cd4b1fa",
+                        path = "manifest.xml"
+                    ),
+                    resolvedRevision = "31588aa8f8555474e1c3c66a359ec99e4cd4b1fa"
+                )
+
+                val tree = resolver.resolveNestedProvenance(provenance)
```
sschuberth, line 130: This is still called "tree", not sure if it's intentional. Could be e.g. just "repo".

sschuberth (dismissed):
Approving despite my comments, which are just hints.

sschuberth (approved):

PR #4258, opened 2021-07-05
Fully integrate the dependency graph format

oheger-bosch:
This PR makes sure that ORT results using the dependency graph format can be read and written transparently, without any special manual conversion effort.


sschuberth:
LGTM, but I'd like @mnonnenmacher to also take a look, as he was also involved in the other reviews.

PR #4259, opened 2021-07-05
fossid-webapp: Introduce pattern options for naming projects and scans

nnobelis:
These patterns allow to control the name when a scan or project is created.
When a pattern is omitted, a default naming convention is used.

### DIFF
--- a/model/src/main/resources/reference.conf
+++ b/model/src/main/resources/reference.conf
@@ -81,6 +81,22 @@ ort {
         commandLine = --copyright --license --info --strip-root --timeout 300
         parseLicenseExpressions = true
       }
+
+      FossId {
+        serverUrl = "https://fossid.example.com/instance/"
+        user = "user"
+        apiKey = "XYZ"
+
+        namingProjectPattern = "$Var1_$Var2"
+        namingScanPattern = "$Var1_#projectBaseCode_$Var3"
+        namingVariableVar1 = "myOrg"
+        namingVariableVar2 = "myTeam"
+        namingVariableVar3 = "myUnit"
+
+        waitForResult = "false"
+        packageNamespaceFilter = "myorg.myunit"
+        addAuthenticationToUrl = "false"
+      }
     }
 
     storages {
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -320,7 +320,10 @@ class FossId(
                 checkNotNull(scans)
 
                 val existingScan = scans.sortedByDescending { it.id }.find { scan ->
-                    scan.gitBranch == revision && scan.gitRepoUrl == url
+                    // The scans in the server contain the url with the credentials so we have to remove it for the
+                    // comparison. If we don't, the scans won't be matched if the password changes!
+                    val urlWithoutCredentials = scan.gitRepoUrl?.replaceCredentialsInUri()
+                    scan.gitBranch == revision && urlWithoutCredentials == url
                 }
 
                 val scanCode = if (existingScan == null) {
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -23,8 +23,6 @@ import java.io.File
 import java.io.IOException
 import java.net.Authenticator
 import java.time.Instant
-import java.time.LocalDateTime
-import java.time.format.DateTimeFormatter
 
 import kotlin.time.measureTimedValue
 
@@ -85,6 +83,12 @@ import retrofit2.converter.jackson.JacksonConverterFactory
  * * **"addAuthenticationToUrl":** When set, ORT will add credentials from its Authenticator to the URLs sent to FossID.
  * * **"waitForResult":** When set to false, ORT doesn't wait for repositories to be downloaded nor scans to be
  * completed. As a consequence, scan results won't be available in ORT result.
+ *
+ * Naming conventions options. If they are not set, default naming convention are used.
+ * * **"namingProjectPattern":** A pattern for project names when projects are created on the FossID instance. Contains
+ * variables prefixed by "$" e.g. "$Var1_$Var2". Variables are also passed as options and are prefixed by
+ * [NAMING_CONVENTION_VARIABLE_PREFIX] e.g. namingVariableVar1 = "foo".
+ * * **"namingScanPattern":** A pattern for scan names when scans are created on the FossID instance.
  */
 class FossId(
     name: String,
@@ -108,6 +112,11 @@ class FossId(
 
         @JvmStatic
         private val WAIT_REPETITION = 50
+
+        /**
+         * The scanner options beginning with this prefix will be used to parametrize project and scan names.
+         */
+        private const val NAMING_CONVENTION_VARIABLE_PREFIX = "namingVariable"
 
         /**
          * Convert a Git repository URL to a valid project name, e.g.
@@ -158,6 +167,7 @@ class FossId(
     private val user: String
     private val waitForResult: Boolean
     private val secretKeys = listOf("serverUrl", "apiKey", "user")
+    private val namingProvider: FossIdNamingProvider
 
     private val service: FossIdRestService
     private val packageNamespaceFilter: String
@@ -185,6 +195,19 @@ class FossId(
         waitForResult = fossIdScannerOptions["waitForResult"]?.toBooleanStrictOrNull() ?: true
 
         log.info { "waitForResult parameter is set to '$waitForResult'" }
+
+        val namingProjectPattern = fossIdScannerOptions["namingProjectPattern"]?.apply {
+            FossId.log.info { "Naming pattern for projects is $this." }
+        }
+        val namingScanPattern = fossIdScannerOptions["namingScanPattern"]?.apply {
+            FossId.log.info { "Naming pattern for scans is $this." }
+        }
+
+        val namingConventionVariables = fossIdScannerOptions
+            .filterKeys { it.startsWith(NAMING_CONVENTION_VARIABLE_PREFIX) }
+            .mapKeys { it.key.substringAfter(NAMING_CONVENTION_VARIABLE_PREFIX) }
+
+        namingProvider = FossIdNamingProvider(namingProjectPattern, namingScanPattern, namingConventionVariables)
 
         val client = OkHttpClientHelper.buildClient()
 
@@ -282,7 +305,8 @@ class FossId(
 
                 val url = pkg.vcsProcessed.url
                 val revision = pkg.vcsProcessed.revision.ifEmpty { "HEAD" }
-                val projectCode = convertGitUrlToProjectName(url)
+                val projectName = convertGitUrlToProjectName(url)
+                val projectCode = namingProvider.createProjectCode(projectName)
 
                 if (getProject(projectCode) == null) {
                     log.info { "Creating project '$projectCode' ..." }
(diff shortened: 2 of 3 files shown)


model/src/main/resources/reference.conf
```diff
@@ -81,6 +81,22 @@ ort {
...
       }
+
+      FossId {
+        serverUrl = "https://fossid.example.com/instance/"
+        user = "user"
+        apiKey = "XYZ"
+
+        namingProjectPattern = "$Var1_$Var2"
+        namingScanPattern = "$Var1_#projectBaseCode_$Var3"
+        namingVar1 = "myOrg"
```
oheger-bosch, line 92: Is this example correct? I would have expected that the variable must be named `namingVariable1`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -282,7 +309,8 @@ class FossId(
 
                 val url = pkg.vcsProcessed.url
                 val revision = pkg.vcsProcessed.revision.ifEmpty { "HEAD" }
-                val projectCode = convertGitUrlToProjectName(url)
+                val baseProjectCode = convertGitUrlToProjectName(url)
```
oheger-bosch, line 312: I find these variable names confusing. What is now called `projectCode`, isn't this actually the `projectName`? And `projectBaseCode` could then be `projectCode` (or `projectUrl`, `projectId` ?).

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -296,21 +324,24 @@ class FossId(
                 checkNotNull(scans)
 
                 val existingScan = scans.sortedByDescending { it.id }.find { scan ->
-                    scan.gitBranch == revision && scan.gitRepoUrl == url
+                    // The scans in the server contain the url with the credentials so we have to remove it for the
```
oheger-bosch, line 327: This change does not seem to be related to the naming conventions, or is it?

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * This class provides names for projects and scans when the FossID scanner creates them, following a given pattern.
+ * If one or both patterns is null, a default naming convention is used.
+ *
+ * [namingScanPattern] and [namingProjectPattern] are patterns describing the name using variables, e.g. "$var1_$var2".
+ * Variable values are given in the map [namingConventionVariables].
+ *
+ * There also are builtin variables. Builtin variables are are prefixed in the pattern with "#" e.g. "$var1_#builtin".
+ * Available builtin variables:
+ * * **projectBaseCode**: The name of the project (i.e. the part of the URL before .git).
```
oheger-bosch, line 36: Would it make sense to have the current timestamp as a built-in variable?

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Replace the naming convention variables with their values. Used for projects and scans.
+     */
+    private fun replaceNamingConventionVariables(
+        namingConventionPattern: String, baseProjectCode: String,
+        namingConventionVariables: Map<String, String>
+    ): String {
+        log.info { "Parametrizing the name with naming=$namingConventionPattern, baseProjectCode=$baseProjectCode" }
+        var replacedName = namingConventionPattern
```
oheger-bosch, line 62: I guess, this can be simplified. You can create a map that contains both the built-in and the user-defined variables. Then the replacement can be done in a single step. You can also use `fold`, e.g.:
```
val replacedName = namingConventionVariables.fold(namingConventionPattern) { str, variable ->
  str.replace("\$${variable.key}", variable.value)
}
```

model/src/main/resources/reference.conf
nnobelis: No, all variables are prefixed with `naming` but the prefix is not in the variable name.
I did so to be able to group everything naming convention-related in the `ort.conf`, since it's no possible to have structured data such as:
```
        namingConvention {
          projectPattern = "$Var1_$Var2"
          scanPattern = "$Var1_#projectBaseCode_$Var3"
          variables {
            Var = "myOrg"
          }
        }
```

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Yes you are right, thinks got messy when I tried to have FossID semantic (scanCode, projectCode) and the concept of "naming convention".
I revamped it and would appreciate your opinion.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -186,6 +196,23 @@ class FossId(
...
+            .filterKeys { it.startsWith(NAMING_CONVENTION_VARIABLE_PREFIX) }
+            .mapKeys { it.key.substringAfter(NAMING_CONVENTION_VARIABLE_PREFIX) }
+
+        namingProjectPattern?.let {
+            log.info { "Naming pattern for projects is $namingProjectPattern." }
+        }
+
+        namingScanPattern?.let {
+            log.info { "Naming pattern for scans is $namingScanPattern." }
+        }
```
sschuberth, line 212: Nit: Maybe move these straight to the variable assignment above for more compact code?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Yes, the new names are better. Only the `createScan()` function still uses the old ones as parameters names.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -360,9 +391,11 @@ class FossId(
     /**
      * Create a new scan in the FossID server and return the scan code.
      */
-    private suspend fun createScan(projectCode: String, url: String, revision: String): String {
-        val formatter = DateTimeFormatter.ofPattern("'$projectCode'_yyyyMMdd_HHmmss")
-        val scanCode = formatter.format(LocalDateTime.now())
+    private suspend fun createScan(projectCode: String, baseProjectCode: String, url: String, revision: String):
+            String {
```
sschuberth, line 395: Nit: Please rather format like

    private suspend fun createScan(
        projectCode: String, baseProjectCode: String, url: String, revision: String
    ): String {

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -186,6 +196,23 @@ class FossId(
 
         log.info { "waitForResult parameter is set to '$waitForResult'" }
 
+        val namingProjectPattern = fossIdScannerOptions["namingProjectPattern"]
+        val namingScanPattern = fossIdScannerOptions["namingScanPattern"]
+
+        val namingConventionVariables = fossIdScannerOptions
+            .filterKeys { it.startsWith(NAMING_CONVENTION_VARIABLE_PREFIX) }
+            .mapKeys { it.key.substringAfter(NAMING_CONVENTION_VARIABLE_PREFIX) }
```
sschuberth, line 204: From just reading the code, this looks strange: You're basically just stripping prefixed from key names. Which raises the question: Why bother adding these prefixes in the first place if they get stripped anyway?

nnobelis: Please see my answer above to Oliver:

> I did so to be able to group everything naming convention-related in the ort.conf, since it's no possible to have structured data

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Moved to another commit.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -320,7 +320,10 @@ class FossId(
                 checkNotNull(scans)
 
                 val existingScan = scans.sortedByDescending { it.id }.find { scan ->
-                    scan.gitBranch == revision && scan.gitRepoUrl == url
+                    // The scans in the server contain the url with the credentials so we have to remove it for the
+                    // comparison. If we don't, the scans won't be matched if the password changes!
+                    val urlWithoutCredentials = scan.gitRepoUrl!!.substringAfter("@")
```
sschuberth, line 325: This also strips the protocol part, so does the comparison really work? Can't / shouldn't we make use of `replaceCredentialsInUri()` instead?

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
nnobelis: 1. the built-ins are prefix with "#" so you cannot replace variables and build-ins in one step...
2. I don't find `.fold` for `Map`. Do you mean `foldTo` for `MutableMap` ?

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
nnobelis: YAGNI but ok :)

oheger-bosch: I think it is already used for the default pattern; so if you just slightly wanted to adapt the default pattern, you would need such a variable.

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
oheger-bosch: at 1: This depends how you organize the keys in the map and the replacement logic, but I won't insist.
at 2: You can use `Map.entries().fold()`.

model/src/main/resources/reference.conf
oheger-bosch: Ok, got this; but from my understanding, the constant `NAMING_CONVENTION_VARIABLE_PREFIX` has the value "namingVariable", and you filter for keys starting with this value. So the key `namingVar1` used in the example would not be matched - or am I wrong?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -186,6 +196,19 @@ class FossId(
 
         log.info { "waitForResult parameter is set to '$waitForResult'" }
 
+        val namingProjectPattern = fossIdScannerOptions["namingProjectPattern"]?.apply {
```
sschuberth, line 199: Nit: Using `let` (and `it` in the next line) would be a bit more common for this use-case, IMO. Same below.

nnobelis: Understood but then you have to return something from `let` i.e.

```
val namingProjectPattern = fossIdScannerOptions["namingProjectPattern"]?.let {
            FossId.log.info { "Naming pattern for projects is $it." }
            it
}
```
which is more code or am I missing something ?

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
nnobelis: It's ok...I reworked the code to do it in one operation.

Thanks for the suggestions !

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth: My bad, you're right. I missed that.

model/src/main/resources/reference.conf
nnobelis: You are of course right. Fixed !

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -320,7 +320,10 @@ class FossId(
                 checkNotNull(scans)
 
                 val existingScan = scans.sortedByDescending { it.id }.find { scan ->
-                    scan.gitBranch == revision && scan.gitRepoUrl == url
+                    // The scans in the server contain the url with the credentials so we have to remove it for the
+                    // comparison. If we don't, the scans won't be matched if the password changes!
+                    val urlWithoutCredentials = scan.gitRepoUrl?.replaceCredentialsInUri()
+                    scan.gitBranch == revision && urlWithoutCredentials == url
```
sschuberth, line 326: One last commit message nit: "GIT" should be "Git" (as it's not an acronym).

sschuberth (approved):

sschuberth:
Merging despite the unrelated `PipFunTest` failure which got fixed by https://github.com/oss-review-toolkit/ort/pull/4262.

PR #4260, opened 2021-07-05
DependencyGraphConverter: filter out empty projects

oheger-bosch:
Such projects are created as fall-back by PackageManager if resolving
of dependencies fails. They can cause problems if there is already a
dependency graph for their package manager type; then this graph is
overridden by an empty one.

### DIFF
--- a/model/src/main/kotlin/utils/DependencyGraphConverter.kt
+++ b/model/src/main/kotlin/utils/DependencyGraphConverter.kt
@@ -86,7 +86,8 @@ object DependencyGraphConverter {
      * Determine the projects in this [AnalyzerResult] that require a conversion. These are the projects that manage
      * their dependencies in a scope structure.
      */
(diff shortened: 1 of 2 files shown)


sschuberth:
> Such projects are created as fall-back by PackageManager if resolving of dependencies fails.

So what does this mean as a consequence for this change? If empty projects are created in the error case to have something to attach issues to (that's how I understood your comment), then what are issues attached to in the dependency graph format if there is no more empty project?

oheger-bosch:
The empty project is not removed from the result; it stays there with all of its issues. It is only no longer taken into account when the dependency graph is constructed. (It would not contribute to it anyway as it does not define any dependencies.)

sschuberth:
> The empty project is not removed from the result; it stays there with all of its issues. It is only no longer taken into account when the dependency graph is constructed. (It would not contribute to it anyway as it does not define any dependencies.)

Thanks for clarifying. Could you maybe amend the commit message with this info? And while at it, capitalize "filter" after the colon in the title?

sschuberth (approved):

PR #4265, opened 2021-07-06
Add a notifier module that informs the user based on an ORT result

### LINKED ISSUE #4236: ORT notifications module
As discussed in the [ORT developer meeting from 2021-03-04](https://github.com/oss-review-toolkit/ort/wiki/ORT-Developer-Meetings#2021-03-04), an ORT notification module should be created.

This module should have the following functions:
- A modular system offers multiple helpers to send notifications (e.g. E-Mail, Slack messages, ...).
- A _scripting engine_ is provided similar to the `Evaluator` which allows sending notifications based on _rules_ defined in a `notifications.kts` script.
- The user should be able to customize the notifications based on data from the `ortResult`.

MarcelBochtler:
The user can write custom notification rules in a `notifications.kts` which is evaluated.
For this script multiple modules / notification helpers are provided (for now email only).
More ideas for possible modules:
* Create Jira issue
* Send Slack notification
* ...

Resolves #4236.

CC: @neubs-bsi, @Martin-Idel-SI


cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -162,6 +170,35 @@ class ExamplesFunTest : StringSpec() {
             report shouldHaveSize 1
         }
 
+        "emails can be send using notifications.kts" {
```
neubs-bsi, line 173: I would change this to match the other test names:
`Notifications.kts can be complied and executed`

cli/src/main/kotlin/commands/NotifierCommand.kt
```diff
@@ -0,0 +1,80 @@
...
+        "--ort-file", "-i",
+        help = "The ORT result file to read as input."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+        .inputGroup()
+
+    private val notificationsFile by option(
+        "--notification-file", "-n",
```
neubs-bsi, line 48: Variable is called `notificationsFile` but the one taken is called `notification-file`. Would propbably be nice to align

cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -162,6 +170,35 @@ class ExamplesFunTest : StringSpec() {
...
+                        port = 3025,
+                        username = "no-reply@ossreviewtoolkit.org",
+                        password = "pwd",
+                        useSsl = false,
+                        fromAddress = "no-reply@ossreviewtoolkit.org"
+                    )
+                )
+            )
+
+            notifier.run(takeExampleFile("notifications.kts").readText())
```
neubs-bsi, line 192: Extract `takeExampleFile("notifications.kts").readText()` into separate variable would increase readability and align with other test for rules

model/src/main/kotlin/config/EmailConfiguration.kt
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model.config
+
+data class EmailConfiguration(
```
oheger-bosch, line 22: Not sure about the name: With "email" I associate a single message. So should this better be _EmailServerConfiguration_ or _EmailSenderConfiguration_?

notifier/src/main/kotlin/modules/Email.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.commons.mail.DefaultAuthenticator
+import org.apache.commons.mail.Email
+import org.apache.commons.mail.SimpleEmail
+
+import org.ossreviewtoolkit.model.config.EmailConfiguration
+
+/**
+ * Notification module that provides a configured email client.
+ */
+class Email(config: EmailConfiguration) {
```
oheger-bosch, line 31: Again, I find the class name `Email` confusing. Maybe something like `EmailNotifier`?

cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -162,6 +170,35 @@ class ExamplesFunTest : StringSpec() {
...
+            val greenMail = GreenMail()
+            greenMail.setUser("no-reply@ossreviewtoolkit.org", "no-reply@ossreviewtoolkit.org", "pwd")
+            greenMail.start()
+
+            val ortResult = File("src/funTest/assets/semver4j-analyzer-result.yml").readValue<OrtResult>()
+            val notifier = Notifier(
+                ortResult, NotifierConfiguration(
+                    EmailConfiguration(
+                        hostName = "localhost",
+                        port = 3025,
```
oheger-bosch, line 183: It is preferable to use a dynamic port here if possible, rather than a hard-coded one, which might already be used in a specific environment. I think, the documentation of GreenMail mentions that this is supported.

oheger-bosch (commented):
It is not yet clear to me, what the actual scope of the notifier module is. The existing API and functionality provided are very basic; for instance, you can send only plain text mails, not HTML-formatted ones. All logic has to be implemented in a script, which is on the one hand of course very flexible, but on the other hand does not help you that much when implementing more complex use cases.

Therefore, at this early stage, it may be better to not focus on concrete transport mechanisms (what is not the core domain of ORT), but think about use cases we want to support and how an API for those could look like.

Just as an example, an alternative approach could be that ORT itself produces a number of notifications or events (such as issues, rule violations, but maybe also life-cycle events like a run starts or finishes), and scripting logic can be used to take up some of these notifications and forward them via specific extensible channels (email, Jira, a chat system, whatever).

model/src/main/kotlin/config/EmailConfiguration.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Configuration if SSL/TLS encryption should be enabled with the SMTP server.
+     */
+    val useSsl: Boolean = true,
+
+    /**
+     * The 'from' field of the outgoing email.
+     */
+    val fromAddress: String = ""
```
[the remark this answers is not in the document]
sschuberth: Or maybe better "no-reply@oss-review-toolkit.org" like you do in the reference config.

model/src/main/resources/reference.conf
```diff
@@ -144,4 +144,15 @@ ort {
...
   }
+
+  notifier {
+    email {
+      hostName = "localhost"
+      username = "user"
+      password = "secret"
+      port = 25
+      useSsl = true
+      fromAddress = "no-reply@ossreviewtoolkit.org"
```
sschuberth, line 155: Please use the TLD with dashes instead: no-reply@oss-review-toolkit.org

cli/src/main/kotlin/commands/NotifierCommand.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        notifier.run(script)
+    }
+
+    private fun readDefaultNotificationsFile(): String {
+        val notificationsFile = ortConfigDirectory.resolve("notifications.kts")
+
+        if (!notificationsFile.isFile) {
+            throw UsageError(
+                "No notification-file option specified and no default notification-file found at " +
```
sschuberth, line 73: Please either use "--notification-file" or "notification file", but not something in between. I'd prefer "notification file" in both places to decouple from the concrete name of the option.

notifier/src/main/kotlin/Notifier.kt
```diff
@@ -0,0 +1,58 @@
...
+
+import java.time.Instant
+
+import org.ossreviewtoolkit.model.NotifierRun
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.config.NotifierConfiguration
+import org.ossreviewtoolkit.notifier.modules.Email
+import org.ossreviewtoolkit.utils.ScriptRunner
+
+class Notifier(ortResult: OrtResult = OrtResult.EMPTY, config: NotifierConfiguration) : ScriptRunner() {
```
sschuberth, line 30: Why a default for `ortResult`, but none for `config`?

notifier/src/main/kotlin/modules/Email.kt
```diff
@@ -0,0 +1,52 @@
...
+        client.setAuthenticator(DefaultAuthenticator(config.username, config.password))
+        client.setSSLOnConnect(config.useSsl)
+        client.setFrom(config.fromAddress)
+    }
+
+    @Suppress("UNUSED") // This is intended to be used by notification script implementations.
+    fun sendEmail(subject: String, message: String, vararg receivers: String) {
+        client.subject = subject
+        client.setMsg(message)
+        receivers.forEach { client.addTo(it) }
```
sschuberth, line 48: Could be just `client.addTo(*receivers)`.

cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -162,6 +170,37 @@ class ExamplesFunTest : StringSpec() {
...
+                        username = "no-reply@ossreviewtoolkit.org",
```
sschuberth, line 184: Dashed domain version.

cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -162,6 +170,37 @@ class ExamplesFunTest : StringSpec() {
...
+                        password = "pwd",
+                        useSsl = false,
+                        fromAddress = "no-reply@ossreviewtoolkit.org"
```
sschuberth, line 187: Dashed domain version.

cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -162,6 +170,37 @@ class ExamplesFunTest : StringSpec() {
             report shouldHaveSize 1
         }
 
+        "Notifications.kts can be complied and executed" {
```
sschuberth, line 173: Nit: Lowercase "notifications.kts".

README.md
```diff
@@ -53,6 +53,8 @@ use):
   classifications.
 * [_Reporter_](#reporter) - presents results in various formats such as visual reports, Open Source notices or
   Bill-Of-Materials (BOMs) to easily identify dependencies, licenses, copyrights or policy rule violations.
+* [_Notifier_](#notifier) - notifies the user for found issues or license / copyright findings based on a customizable
```
sschuberth, line 56: IMO we should not explicitly say what *about* the notifier notifies; it's the whole point that this is completely customizable, but just give examples.

Like: "sends status notifications (like about issues in ORT result files) to arbitrary receivers."

sschuberth (dismissed):

MarcelBochtler:
> Therefore, at this early stage, it may be better to not focus on concrete transport mechanisms (what is not the core domain of ORT), but think about use cases we want to support and how an API for those could look like.

Had a talk with @oheger-bosch about this.

For now I'll move these changes into `experimental` packages and remove the changes in the `README.md` (as these do not fit to the rest of the README anyways. E.g. the missing logo).

CC: @sschuberth

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -63,6 +63,11 @@ data class OrtConfiguration(
      */
     val downloader: DownloaderConfiguration = DownloaderConfiguration(),
 
+    /**
+     * The configuration of the notifier.
+     */
+    val notifier: NotifierConfiguration = NotifierConfiguration(),
```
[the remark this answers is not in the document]
MarcelBochtler: After the scanner, as the `notifier` will most likely run at a later point to have as much possible information in the `ortResult`.
Moved it.

cli/src/main/kotlin/commands/NotifierCommand.kt
MarcelBochtler: I took this from another command. But I agree.

neubs-bsi (approved):

notifier/src/main/kotlin/modules/EmailNotifier.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.ossreviewtoolkit.model.config.SendMailConfiguration
+
+/**
+ * Notification module that provides a configured email client.
+ */
+class EmailNotifier(config: SendMailConfiguration) {
+    private val client: Email
+
+    init {
+        client = SimpleEmail()
```
oheger-bosch, line 35: Just out of curiosity: Is Commons Mail thread-safe? When using a single instance like here probably not. This is not an issue for now, but something to keep in mind when optimizations are done later. (E.g. as in the reporter where reports are generated in parallel.)

model/src/main/kotlin/config/NotifierConfiguration.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model.config
+
+data class NotifierConfiguration(
+    val email: SendMailConfiguration? = null
```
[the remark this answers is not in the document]
MarcelBochtler: IMO `email` is still more on point for this user facing naming:
```
  notifier {
    email {
      hostName = "localhost"
      username = "user"
      password = "secret"
      port = 465
      useSsl = true
      fromAddress = "no-reply@oss-review-toolkit.org"
    }
  }
```

vs:
```
  notifier {
    mailSender {
      hostName = "localhost"
      username = "user"
      password = "secret"
      port = 465
      useSsl = true
      fromAddress = "no-reply@oss-review-toolkit.org"
    }
  }
```

Maybe a naming with `server` in the name, like you suggested, would be the most intuitive.
What do you think. Rename `SendMailConfiguration` once again? Or would you still prefer `mailSender` for the configuration?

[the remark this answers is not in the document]
oheger-bosch: It is indeed difficult to find sound names :-(
Using _server_ in the name is probably also not fully correct, as the `fromAddress` is not strictly part of a server configuration.
From a user PoV, I can understand that the parts of the configuration should directly identify the notifier modules they belong to. So _email_ is probably not that bad. Or, as we already dropped the 'e' in the configuration class, should it then just be _mail_?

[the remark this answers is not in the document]
MarcelBochtler: > Or, as we already dropped the 'e' in the configuration class, should it then just be mail?

I'm ok with that.

oheger-bosch (approved):

PR #4266, opened 2021-07-06
Add else branch to when to make it exhaustive

MarcelBochtler:


spdx-utils/src/main/kotlin/SpdxExpression.kt
```diff
@@ -470,6 +470,10 @@ class SpdxLicenseWithExceptionExpression(
                     )
                 }
             }
+            else -> throw SpdxException(
```
sschuberth, line 471: Nit: Blank line before this one to align with spacing above.

sschuberth: And can we even reach this? Because shouldn't in this case the `as` cast in line 455 fail before?

spdx-utils/src/main/kotlin/SpdxExpression.kt
```diff
@@ -449,12 +449,10 @@ class SpdxLicenseWithExceptionExpression(
...
     override fun exception() = exception
 
-    override fun normalize(mapDeprecated: Boolean): SpdxExpression {
+    override fun normalize(mapDeprecated: Boolean) =
         // Manually cast to SpdxLicenseException, because the type resolver does not recognize that in all subclasses of
         // SpdxSimpleExpression normalize() returns an SpdxSingleLicenseExpression.
-        val normalizedLicense = license.normalize(mapDeprecated) as SpdxSingleLicenseExpression
-
-        return when (normalizedLicense) {
+        when (val normalizedLicense = license.normalize(mapDeprecated) as SpdxSingleLicenseExpression) {
```
sschuberth, line 455: Please also mention in the commit message that this allows turning the function into an expression (which is sort of an incentive to do this inlining).

sschuberth (changes requested):

spdx-utils/src/main/kotlin/SpdxExpression.kt
MarcelBochtler: > And can we even reach this? Because shouldn't in this case the as cast in line 455 fail before?

`SpdxSingleLicenseExpression` can be:
- `SpdxLicenseIdExpression`
- `SpdxLicenseReferenceExpression`
- `SpdxLicenseWithExceptionExpression`
- `SpdxSimpleExpression`

From what I see this would cover everything, so the else branch would not be required.
However: IntelliJ complains that it is required and marks it as an error.

How should I change this? Replace the `when` with an if / else? :/

sschuberth: I would have said to suppress the inspection hint, but there is [contradicting information whether this is possible](https://stackoverflow.com/questions/61756519/how-can-i-disable-the-non-exhaustive-when-expression-inspection). So, can't we just ignore this like we did until now?

MarcelBochtler: I did not find a way to disable it. I'll remove the commit.

sschuberth (approved):

spdx-utils/src/main/kotlin/SpdxExpression.kt
mnonnenmacher: I noticed that after a recent update IntelliJ started complaining about this but it was fine when I wrote this code. I was hoping that it would get fixed in another update of the Kotlin plugin but didn't check if there is a bug report for this. We had another case recently where IntelliJ wrongly reported a compiler error, but I don't remember where this was right now. This is getting a bit annoying.

PR #4267, opened 2021-07-06
OkHttpClientHelper: Improve error logging

oheger-bosch:
In case of a request that failed because of an exception, log some
information about the current request. This is especially useful when
requests are executed asynchronously, as it is then sometimes hard to
say, which request caused the failure.

### DIFF
--- a/utils/src/main/kotlin/OkHttpClientHelper.kt
+++ b/utils/src/main/kotlin/OkHttpClientHelper.kt
@@ -81,13 +81,23 @@ object OkHttpClientHelper {
         return OkHttpClient.Builder()
             .addNetworkInterceptor { chain ->
                 val request = chain.request()
-                chain.proceed(
-                    if (request.header("User-Agent") == null) {
-                        request.newBuilder().header("User-Agent", Environment.ORT_USER_AGENT).build()
-                    } else {
-                        request
+                @Suppress("TooGenericExceptionCaught")
+                try {
+                    chain.proceed(
+                        if (request.header("User-Agent") == null) {
+                            request.newBuilder().header("User-Agent", Environment.ORT_USER_AGENT).build()
+                        } else {
+                            request
+                        }
+                    )
+                } catch (e: Exception) {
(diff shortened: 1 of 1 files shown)


neubs-bsi (approved):

utils/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -81,13 +81,20 @@ object OkHttpClientHelper {
...
+                try {
+                    chain.proceed(
+                        if (request.header("User-Agent") == null) {
+                            request.newBuilder().header("User-Agent", Environment.ORT_USER_AGENT).build()
+                        } else {
+                            request
+                        }
+                    )
+                } catch (e: Exception) {
+                    log.error { "HTTP request to '${request.url}' failed with exception." }
```
nnobelis, line 94: Better:
`logger.error(e) {"HTTP request to '${request.url}' failed with exception." }`

Then you don't need the `showStackTrace`.

nnobelis (changes requested):

utils/src/main/kotlin/OkHttpClientHelper.kt
sschuberth, line 94: "*an* exception"

Also, should we add `${e.message}` to the string?

sschuberth (changes requested):

utils/src/main/kotlin/OkHttpClientHelper.kt
sschuberth: Should be `log` instead of `logger`, but nice otherwise, I didn't know about this overload.

oheger-bosch: `showStackTrace` takes the command line option into account whether stack traces should be printed. I think, it is therefore common to use it.

nnobelis (approved):

sschuberth (approved):

PR #4271, opened 2021-07-07
analyzer: Replace jcenter with mavenCentral in test projects

mnonnenmacher:
Replace `jcenter()` with `mavenCentral()` in all synthetic test
projects, because JCenter is goin to be shutdown [1]. The external test
project will be addressed separately.

[1] https://blog.gradle.org/jcenter-shutdown

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/gradle-android/build.gradle
+++ b/analyzer/src/funTest/assets/projects/synthetic/gradle-android/build.gradle
@@ -1,7 +1,7 @@
 buildscript {
     repositories {
         google()
-        jcenter()
+        mavenCentral()
     }
 
     dependencies {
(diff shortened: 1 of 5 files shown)


sschuberth:
Note that I deliberately did not replace JCenter with Maven Central in examples when I did https://github.com/oss-review-toolkit/ort/pull/4100 as the blog page says

> UPDATE: JFrog has decided to keep JCenter as a read-only repository indefinitely.

So I though keeping JCenter is a good test case for legacy projects that we might need to also analyze in real-life.

sschuberth:
> The external test project will be addressed separately.

... or we merge this PR still, but do not address the external test projects.

mnonnenmacher:
Then I vote for keeping the external projects but updating the synthetic projects, also to get rid of the deprecation warnings.

sschuberth (approved):

PR #4272, opened 2021-07-07
ClearlyDefinedPackageCurationProvider: Handle all exceptions

oheger-bosch:
In a concrete scenario, ClearlyDefined did not send a response within
the configured timeout, causing a SocketTimeoutException to be thrown.
This exception was not handled and eventually caused the analyzer to
crash. To avoid this, add a catch block for all exceptions.

Add a test case using a mock server to verify that timeout exceptions
are now handled properly.

### DIFF
--- a/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
@@ -96,29 +96,41 @@ class ClearlyDefinedPackageCurationProvider(
         val (type, provider) = pkgId.toClearlyDefinedTypeAndProvider() ?: return emptyList()
         val namespace = pkgId.namespace.takeUnless { it.isEmpty() } ?: "-"
 
-        val curation = try {
+        val curation = runCatching {
             // TODO: Maybe make PackageCurationProvider.getCurationsFor() a suspend function; then all derived
             //       classes could deal with coroutines more easily.
             runBlocking(Dispatchers.IO) { service.getCuration(type, provider, namespace, pkgId.name, pkgId.version) }
-        } catch (e: HttpException) {
-            // A "HTTP_NOT_FOUND" is expected for non-existing curations, so only handle other codes as a failure.
-            if (e.code() != HttpURLConnection.HTTP_NOT_FOUND) {
-                e.showStackTrace()
+        }.onFailure { e ->
+            when (e) {
+                is HttpException -> {
+                    // A "HTTP_NOT_FOUND" is expected for non-existing curations, so only handle other codes as a
+                    // failure.
+                    if (e.code() != HttpURLConnection.HTTP_NOT_FOUND) {
+                        e.showStackTrace()
 
-                log.warn {
-                    val message = e.response()?.errorBody()?.string() ?: e.collectMessagesAsString()
-                    "Getting curations for '${pkgId.toCoordinates()}' failed with code ${e.code()}: $message"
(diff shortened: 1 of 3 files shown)


analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
```diff
@@ -110,6 +118,14 @@ class ClearlyDefinedPackageCurationProvider(server: Server = Server.PRODUCTION)
 
             log.warn { "Deserializing the ClearlyDefined curation for '${pkgId.toCoordinates()}' failed." }
 
+            return emptyList()
+        } catch (e: Exception) {
```
neubs-bsi, line 122: Currently we only experience SocketTimeoutExceptions. 

Might it not be better to start with a specific exception that actually occurs and catch it, instead of all?
It might happen that other, important exceptions are overlooked, when a user becomes accustomed to the warning.

analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
```diff
@@ -96,6 +96,7 @@ class ClearlyDefinedPackageCurationProvider(
         val (type, provider) = pkgId.toClearlyDefinedTypeAndProvider() ?: return emptyList()
         val namespace = pkgId.namespace.takeUnless { it.isEmpty() } ?: "-"
 
+        @Suppress("TooGenericExceptionCaught")
```
sschuberth, line 99: Hint: You could avoid this suppression by using `runCatching` / `onFailure` / `when` like

        val curation = runCatching {
            // TODO: Maybe make PackageCurationProvider.getCurationsFor() a suspend function; then all derived
            //       classes could deal with coroutines more easily.
            runBlocking(Dispatchers.IO) { service.getCuration(type, provider, namespace, pkgId.name, pkgId.version) }
        }.onFailure { e ->
            when (e) {
                is HttpException -> {
                    // A "HTTP_NOT_FOUND" is expected for non-existing curations, so only handle other codes as a failure.
                    if (e.code() != HttpURLConnection.HTTP_NOT_FOUND) {
                        e.showStackTrace()

                        log.warn {
                            val message = e.response()?.errorBody()?.string() ?: e.collectMessagesAsString()
                            "Getting curations for '${pkgId.toCoordinates()}' failed with code ${e.code()}: $message"
                        }
                    }
                }

                is JsonMappingException -> {
                    e.showStackTrace()

                    log.warn { "Deserializing the ClearlyDefined curation for '${pkgId.toCoordinates()}' failed." }
                }

                else -> {
                    e.showStackTrace()

                    log.warn {
                        "Querying ClearlyDefined curation for '${pkgId.toCoordinates()}' failed: ${e.collectMessagesAsString()}"
                    }
                }
            }
        }.getOrNull() ?: return emptyList()

oheger-bosch: Yes, this looks good. Changed to this construct.

analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
oheger-bosch: In this special case, I see the problem that any unexpected exception will cause the analyzer to crash. For a relatively unimportant task like loading optional curations from ClearlyDefined, this should not be the case IMHO.

sschuberth (approved):

PR #4274, opened 2021-07-08
fossid-webapp: Delta scans

nnobelis:
When set, ORT will create delta scans. When only changes in a repository
need to be scanned, delta scans reuse the identifications of latest scan
on the same repository to reduce the amount of findings.
If deltaScans is set and no scan exist yet, an initial scan called
"origin" scan will be created.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -44,6 +44,7 @@ import org.ossreviewtoolkit.clients.fossid.listMarkedAsIdentifiedFiles
 import org.ossreviewtoolkit.clients.fossid.listPendingFiles
 import org.ossreviewtoolkit.clients.fossid.listScansForProject
 import org.ossreviewtoolkit.clients.fossid.model.Project
+import org.ossreviewtoolkit.clients.fossid.model.Scan
 import org.ossreviewtoolkit.clients.fossid.model.status.DownloadStatus
 import org.ossreviewtoolkit.clients.fossid.model.status.ScanState
 import org.ossreviewtoolkit.clients.fossid.runScan
@@ -319,30 +320,7 @@ class FossId(
                     .checkResponse("list scans for project").data
                 checkNotNull(scans)
 
-                val existingScan = scans.sortedByDescending { it.id }.find { scan ->
-                    // The scans in the server contain the url with the credentials so we have to remove it for the
-                    // comparison. If we don't, the scans won't be matched if the password changes!
-                    val urlWithoutCredentials = scan.gitRepoUrl?.replaceCredentialsInUri()
-                    scan.gitBranch == revision && urlWithoutCredentials == url
-                }
-
-                val scanCode = if (existingScan == null) {
-                    log.info { "No scan found for $url and revision $revision. Creating scan ..." }
-
-                    val scanCode = namingProvider.createScanCode(projectName)
-                    val newUrl = if (addAuthenticationToUrl) queryAuthenticator(url) else url
-                    createScan(projectCode, scanCode, newUrl, revision)
-                    log.info { "Initiating data download ..." }
-                    service.downloadFromGit(user, apiKey, scanCode)
-                        .checkResponse("download data from Git", false)
-                    scanCode
-                } else {
-                    log.info { "Scan ${existingScan.code} found for $url and revision $revision." }
-
-                    requireNotNull(existingScan.code) {
-                        "FossId returned a null scancode for an existing scan"
-                    }
-                }
+                val scanCode = checkAndCreateScan(scans, revision, url, projectCode, projectName)
 
                 val provenance = RepositoryProvenance(pkg.vcsProcessed, pkg.vcsProcessed.revision)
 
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -162,6 +166,13 @@ class FossId(
         }
     }
 
+    /**
+     * The qualifier of a scan when delta scans are enabled.
+     */
+    internal enum class DeltaTag {
+        ORIGIN, DELTA
```
sschuberth, line 173: Can you please document each enum member?

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
```diff
@@ -47,26 +49,33 @@ class FossIdNamingProvider(
...
+            "#projectName" to projectName
+        )
+        replaceNamingConventionVariables(namingProjectPattern, builtins, namingConventionVariables)
     } ?: projectName
 
-    fun createScanCode(projectName: String): String {
-        val pattern = namingScanPattern ?: "${projectName}_#currentTimestamp"
-        return replaceNamingConventionVariables(pattern, projectName, namingConventionVariables)
+    fun createScanCode(projectName: String, deltaTag: FossId.DeltaTag? = null): String {
+        val defaultPattern = "${projectName}_#currentTimestamp${if (deltaTag == null) "" else "_#deltaTag"}"
```
sschuberth, line 59: BTW, it just occurred to me: Why is `${projectName}` already inserted here? Shouldn't it be `#projectName` to let `replaceNamingConventionVariables()` do the replacement?

nnobelis: Oh yes, well spotted.

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
```diff
@@ -47,26 +49,33 @@ class FossIdNamingProvider(
...
+            "#projectName" to projectName
+        )
+        replaceNamingConventionVariables(namingProjectPattern, builtins, namingConventionVariables)
     } ?: projectName
 
-    fun createScanCode(projectName: String): String {
-        val pattern = namingScanPattern ?: "${projectName}_#currentTimestamp"
-        return replaceNamingConventionVariables(pattern, projectName, namingConventionVariables)
+    fun createScanCode(projectName: String, deltaTag: FossId.DeltaTag? = null): String {
+        val defaultPattern = "#projectName_#currentTimestamp${if (deltaTag == null) "" else "_#deltaTag"}"
```
sschuberth, line 59: Thanks for the pattern fix, but ideally it would go to a separate commit (the first) in this PR, as it has been wrong before, and it's an independent change.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -319,35 +339,15 @@ class FossId(
...
-                    scanCode
+                val scanCode = if (deltaScans) {
+                    checkAndCreateDeltaScan(scans, revision, url, projectCode, projectName)
                 } else {
-                    log.info { "Scan ${existingScan.code} found for $url and revision $revision." }
-
-                    requireNotNull(existingScan.code) {
-                        "FossId returned a null scancode for an existing scan"
-                    }
+                    checkAndCreateScan(scans, revision, url, projectCode, projectName)
```
sschuberth, line 345: For a more pleasant review experience, could also the factoring-out of this function go to a separate (non-functional-change) commit? That allows a reviewer to only skim over commit that do not change / add logic, and to concentrate on the real changes.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -394,6 +416,64 @@ class FossId(
                 "FossId returned a null scancode for an existing scan"
             }
         }
+        if (waitForResult) checkScan(scanCode)
```
sschuberth, line 419: Why does this check now happen *after* getting the raw result?

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
```diff
@@ -35,8 +35,10 @@ import org.ossreviewtoolkit.utils.log
  * Available builtin variables:
  * * **projectName**: The name of the project (i.e. the part of the URL before .git).
  * * **currentTimestamp**: The current time.
+ * * **deltaTag** (scan code only): If delta scans is enabled, this qualify the scan as an *origin* scan or a *delta*
+ * scan.
  */
-class FossIdNamingProvider(
+internal class FossIdNamingProvider(
```
sschuberth, line 41: This lowering of the visibility is also unrelated to this change, strictly spoken, or?

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
```diff
@@ -47,26 +49,33 @@ class FossIdNamingProvider(
...
-        val pattern = namingScanPattern ?: "#projectName_#currentTimestamp"
-        return replaceNamingConventionVariables(pattern, projectName, namingConventionVariables)
+    fun createScanCode(projectName: String, deltaTag: FossId.DeltaTag? = null): String {
+        val defaultPattern = "#projectName_#currentTimestamp${if (deltaTag == null) "" else "_#deltaTag"}"
+        val pattern = namingScanPattern ?: defaultPattern
+        val builtins = mutableMapOf("#projectName" to projectName)
+        deltaTag?.let {
+            builtins += "#deltaTag" to deltaTag.name.lowercase()
+        }
+        return replaceNamingConventionVariables(pattern, builtins, namingConventionVariables)
```
sschuberth, line 65: How about checking for `deltaTag != null` only once (and inlining `pattern`), like

        var defaultPattern = "#projectName_#currentTimestamp"
        val builtins = mutableMapOf("#projectName" to projectName)

        if (deltaTag != null) {
            defaultPattern += "_#deltaTag"
            builtins += "#deltaTag" to deltaTag.name.lowercase()
        }

        return replaceNamingConventionVariables(namingScanPattern ?: defaultPattern, builtins, namingConventionVariables)

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Heu, no it still happen before: the call is in `checkAndCreateScan` and the latter is called before `getRawResults`.

scanner/src/main/kotlin/scanners/fossid/FossIdNamingProvider.kt
nnobelis: No it is required because `createScanCode` takes a `DeltaTag` in parameters which is `internal` and leads to:
`'public' function exposes its 'internal' parameter type DeltaTag`

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth: What I meant is: Is it intended that the original `checkScan` call from line 328 is removed?

PS: What does "Heu" mean? 😀

nnobelis: Yes it is intended because the call moved to `checkAndCreateScan` and `checkAndCreateDeltaScan`.
Heu = Hrm in french :)

sschuberth (approved):

PR #4277, opened 2021-07-12
SpdxDocumentFile: Use packageFilename before downloadLocation for VcsInfo

### LINKED ISSUE #4276: SpdxDocumentFile: Use packageFilename if present before using downloadLocation
In https://github.com/oss-review-toolkit/ort/blob/master/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt#L320-L325 I propose we change the order of the check e.g. use packageFilename if present else use downloadLocation.
This make sense as if packageFilename is defined the code is already present locally so why download it again. Also this always the use case where a user uses downloadLocation to specify where he/she orginally got the code but wants to scan their local fork (with perhaps minor changes).

neubs-bsi:
If the `packageFilename` is defined, the code is already present, thus, need
not be downloaded again. Additionally, this ensures that the code is scanned
that is actually used by the project, which could include minor changes.

Resolves #4276.

This was discussed in [this thread](https://ort-talk.slack.com/archives/C9NNJ54B1/p1625841198374400) in Slack.

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -317,12 +317,10 @@ class SpdxDocumentFile(
     private fun SpdxPackage.toPackage(workingDir: File): Package {
         val packageDescription = description.takeUnless { it.isEmpty() } ?: summary
 
-        // If the VCS information cannot be determined from the download location, fall back to try getting it from the
-        // VCS working tree itself.
-        val vcs = getVcsInfo() ?: run {
-            val packageDir = workingDir.resolve(packageFilename)
-            VersionControlSystem.forDirectory(packageDir)?.getInfo()
-        }.orEmpty()
+        // If the VCS information cannot be determined from the VCS working tree itself, fall back to try getting it
+        // from the download location.
+        val packageDir = workingDir.resolve(packageFilename)
+        val vcs = VersionControlSystem.forDirectory(packageDir)?.getInfo() ?: getVcsInfo().orEmpty()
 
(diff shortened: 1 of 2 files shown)


neubs-bsi:
@tsteenbe this might be interesting for you.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -319,10 +319,10 @@ class SpdxDocumentFile(
 
         // If the VCS information cannot be determined from the download location, fall back to try getting it from the
         // VCS working tree itself.
-        val vcs = getVcsInfo() ?: run {
```
oheger-bosch, line 322: The comment before this construct now has to be updated as well.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -319,10 +319,10 @@ class SpdxDocumentFile(
...
+        val vcs = run {
```
oheger-bosch, line 322: The `run` is no longer necessary. This can be simplified to
`val vcs = VersionControlSystem.forDirectory(packageDir)?.getInfo() ?: getVcsInfo().orEmpty()`

oheger-bosch (changes requested):

tsteenbe:
> @tsteenbe this might be interesting for you

Thanks @neubs-bsi  for implementing this!

tsteenbe:
> @tsteenbe this might be interesting for you

Thanks @neubs-bsi  for implementing this!

oheger-bosch (approved):

PR #4278, opened 2021-07-12
fossid-webapp: Refine the "reuse identification" behavior

nnobelis:
Tweak the logic when reusing existing scans: If the scan is not finished, wait for it to complete.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -383,6 +383,31 @@ class FossId(
         return results
     }
 
+    private suspend fun List<Scan>.findLatestPendingOrFinishedScan(url: String, revision: String? = null): Scan? =
+        filter {
+            // The scans in the server contain the url with the credentials so we have to remove it for the
+            // comparison. If we don't, the scans won't be matched if the password changes!
+            val urlWithoutCredentials = it.gitRepoUrl?.replaceCredentialsInUri()
+            urlWithoutCredentials == url && (revision == null || (it.gitBranch == revision))
+        }.sortedByDescending { scan -> scan.id }.find { scan ->
+            val scanCode = requireNotNull(scan.code) {
+                "FossId returned a null scancode for an existing scan."
+            }
+
+            val response = service.checkScanStatus(user, apiKey, scanCode)
+                .checkResponse("check scan status", false)
+            when (response.data?.state) {
+                ScanState.FINISHED -> true
+                null, ScanState.NOT_STARTED, ScanState.INTERRUPTED -> false
+                ScanState.STARTED, ScanState.SCANNING, ScanState.AUTO_ID, ScanState.QUEUED -> {
+                    log.warn { "Found previous scan but it is still running." }
+                    log.warn { "Ignoring the 'waitForResult' option and waiting ..." }
+                    waitScanComplete(scanCode)
+                    true
+                }
+            }
+        }
+
     private suspend fun checkAndCreateScan(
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -383,19 +383,41 @@ class FossId(
...
     }
 
+    private suspend fun List<Scan>.findLatestPendingOrFinishedScan(url: String, revision: String? = null): Scan? {
+        return sortedByDescending { scan -> scan.id }.find { scan ->
+            // The scans in the server contain the url with the credentials so we have to remove it for the
+            // comparison. If we don't, the scans won't be matched if the password changes!
+            val urlWithoutCredentials = scan.gitRepoUrl?.replaceCredentialsInUri()
+            if (urlWithoutCredentials == url && (revision == null || (scan.gitBranch == revision))) {
+                val scanCode = requireNotNull(scan.code) {
+                    "FossId returned a null scancode for an existing scan"
```
oheger-bosch, line 393: Missing '.' at the end of the sentence.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -383,19 +383,41 @@ class FossId(
...
+                }
```
oheger-bosch, line 394: Blank line after closing '}'.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -383,19 +383,41 @@ class FossId(
         return results
     }
 
+    private suspend fun List<Scan>.findLatestPendingOrFinishedScan(url: String, revision: String? = null): Scan? {
+        return sortedByDescending { scan -> scan.id }.find { scan ->
+            // The scans in the server contain the url with the credentials so we have to remove it for the
+            // comparison. If we don't, the scans won't be matched if the password changes!
+            val urlWithoutCredentials = scan.gitRepoUrl?.replaceCredentialsInUri()
+            if (urlWithoutCredentials == url && (revision == null || (scan.gitBranch == revision))) {
```
oheger-bosch, line 391: Maybe you could extract the `if` condition to an additional `filter()` call executed before `sortedByDescending`. This would simplify the lambda passed to `find` and would be slightly more efficient because the list is reduced before it is sorted.

oheger-bosch (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Are you sure about this ? This is not a log message rather an exception message....

oheger-bosch: I think, this is a convention in the complete ORT codebase.

nnobelis: After a quick search this is in fact inconsistent in ORT. However I will fix it for you :)

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -383,19 +383,41 @@ class FossId(
         return results
     }
 
+    private suspend fun List<Scan>.findLatestPendingOrFinishedScan(url: String, revision: String? = null): Scan? {
```
oheger-bosch, line 386: One more minor nit: As the function now consists of a single expression, you can drop the curly braces and the return statement.

oheger-bosch (approved):

PR #4281, opened 2021-07-13
fossid-webapp: Various improvements

nnobelis:

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -36,6 +36,7 @@ import org.ossreviewtoolkit.clients.fossid.checkResponse
 import org.ossreviewtoolkit.clients.fossid.checkScanStatus
 import org.ossreviewtoolkit.clients.fossid.createProject
 import org.ossreviewtoolkit.clients.fossid.createScan
+import org.ossreviewtoolkit.clients.fossid.deleteScan
 import org.ossreviewtoolkit.clients.fossid.downloadFromGit
 import org.ossreviewtoolkit.clients.fossid.getProject
 import org.ossreviewtoolkit.clients.fossid.listIdentifiedFiles
@@ -188,6 +189,11 @@ class FossId(
     private val waitForResult: Boolean
     private val secretKeys = listOf("serverUrl", "apiKey", "user")
     private val namingProvider: FossIdNamingProvider
+    // A list of all scans created in an ORT run, to be able to delete them in case of error.
+    // The reasoning is that either all these scans are successful, either none is created at all (clean slate).
+    // A use case is that an ORT run is created regularly e.g. nightly, and we want to have exactly the same amount
+    // of scans for each package.
+    private val createdScans = mutableSetOf<String>()
 
     private val service: FossIdRestService
     private val packageNamespaceFilter: String
@@ -386,6 +392,11 @@ class FossId(
 
                     val scanResult = ScanResult(provenance, details, summary)
                     results.getOrPut(pkg) { mutableListOf() } += scanResult
+
+                    createdScans.forEach {
+                        log.warn("Deleting previous scan $it.")
+                        deleteScan(it)
+                    }
                 }
             }
 
@@ -544,7 +555,7 @@ class FossId(
         requireNotNull(scanId) { "Scan could not be created. The response was: ${response.message}." }
 
         log.info { "Scan has been created with ID $scanId." }
-
+        createdScans.add(scanCode)
         return scanCode
     }
 
@@ -635,6 +646,16 @@ class FossId(
         requireNotNull(result) { "Timeout while waiting for the scan to complete" }
 
         log.info { "Scan has been completed." }
+    }
+
+    /**
+     * Delete a scan with [scanCode].
+     */
+    private suspend fun deleteScan(scanCode: String) {
+        val response = service.deleteScan(user, apiKey, scanCode)
+        response.error?.let {
+            log.error { "Cannot delete scan $scanCode: $it." }
+        }
     }
 
     /**
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -357,9 +357,9 @@ class FossId(
                     checkNotNull(scans)
 
                     val scanCode = if (deltaScans) {
-                        checkAndCreateDeltaScan(scans, revision, url, path, projectCode, projectName)
+                        checkAndCreateDeltaScan(scans, url, revision, path, projectCode, projectName)
                     } else {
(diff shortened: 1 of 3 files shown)


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -385,6 +388,11 @@ class FossId(
                 } catch (e: IllegalStateException) {
                     e.showStackTrace()
                     log.error("Package at url=$url and path='$path' cannot be scanned")
+
+                    createdScans.forEach {
+                        log.warn("Deleting previous scan $it.")
+                        deleteScan(it)
```
neubs-bsi, line 394: does this delete all created scans, or just failed scans, or is this the same?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -383,12 +389,17 @@ class FossId(
         return results
     }
 
-    private suspend fun List<Scan>.findLatestPendingOrFinishedScan(url: String, revision: String? = null): Scan? =
+    private suspend fun List<Scan>.findLatestPendingOrFinishedScan(
```
oheger-bosch, line 407: Minor nit: The order of parameters here is different than to the `createScan()` function. It would probably make sense to align it everywhere.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -332,52 +334,57 @@ class FossId(
...
+                            )
                         )
-                    )
 
-                    val scanResult = ScanResult(provenance, details, summary)
-                    results.getOrPut(pkg) { mutableListOf() } += scanResult
+                        val scanResult = ScanResult(provenance, details, summary)
+                        results.getOrPut(pkg) { mutableListOf() } += scanResult
+                    }
+                } catch (e: IllegalStateException) {
```
oheger-bosch, line 388: Is it possible (and does it make sense) to not only log the exception, but create a result with an issue for the current package?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -188,6 +189,8 @@ class FossId(
     private val waitForResult: Boolean
     private val secretKeys = listOf("serverUrl", "apiKey", "user")
     private val namingProvider: FossIdNamingProvider
+    // a list of all created scans, to be able to delete them in case of error
+    private val createdScans = mutableSetOf<String>()
```
oheger-bosch, line 193: Could this be made a local variable of the `scanPackages()` function? I think, the scope of this field is larger than it needs to be.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: It deletes all scans created **in this run** i.e. all scans that passed `createScan`, whether they were successful or not.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: It need to be accessed by `createScan` ... A global variable is better than modifying a parameter no ?

oheger-bosch: But `createScan()` returns the scan code, so the caller could store it in a local variable.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
neubs-bsi: I don't know that much about FossId. Is there a reason to delete all scans, and not keep at least the successful ones?

nnobelis: Some further explanation:
In one ORT run, several scans are created, one for each package to scan.

The idea with this change is that either all these scans are successful, either none is created at all (clean slate).
The reasoning being that a ORT run is created regularly e.g. nightly, and we want to have exactly the same amount of scan for each package.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Yes but `createScan` is called by `checkAndCreateScan` and `checkAndCreateDeltaScan`.
You could argue that each one of them is returning the `scanCode` to `scanPackages` where the deletion should occur.

However the problem is that in the second branch of `checkAndCreateScan`, no scan is created: previous scan is found and reused. The latter is a scan created in a previous separate run therefore it should not be deleted.

(We delete scans only from this run).

scanner/src/main/kotlin/scanners/fossid/FossId.kt
neubs-bsi: That makes sense. Thanks for explaining.

oheger-bosch: Yes, thanks for the clarification. Could this be added as a comment in the code? I think, this atomicity is an important property, which is not really obvious otherwise.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Ok, I won't insist.

nnobelis:
Please review everything as I pushed a new commit.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Done.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -188,6 +189,11 @@ class FossId(
     private val waitForResult: Boolean
     private val secretKeys = listOf("serverUrl", "apiKey", "user")
     private val namingProvider: FossIdNamingProvider
+    // A list of all scans created in an ORT run, to be able to delete them in case of error.
+    // The reasoning is that either all these scans are successful, either none is created at all (clean slate).
+    // A use case is that a ORT run is created regularly e.g. nightly, and we want to have exactly the same amount
+    // of scan for each package.
+    private val createdScans = mutableSetOf<String>()
```
oheger-bosch, line 196: Thanks for the comment with background information.
Some typos:
* (clean slate) -> (clean state)
* a ORT run -> an ORT run
* amout of scan -> amount of scans

oheger-bosch (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Points 2 and 3 are addressed.
Point 1: I meant indeed "clean slate" :)

oheger-bosch (approved):

PR #4288, opened 2021-07-14
Improve docs examples

tsteenbe:

### DIFF
--- a/docs/config-file-curations-yml.md
+++ b/docs/config-file-curations-yml.md
@@ -73,6 +73,8 @@ The structure of the curations file consist of one or more `id` entries:
     is_meta_data_only: true
     is_modified: true
 ````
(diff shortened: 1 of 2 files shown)


docs/config-file-curations-yml.md
```diff
@@ -73,6 +73,8 @@ The structure of the curations file consist of one or more `id` entries:
     is_meta_data_only: true
```
oheger-bosch, line 73: Commit message nit: The last part of the sentence seems to be incomplete. Should it be
"...curation, so *they* can learn *about* available options." ?

tsteenbe: Good catch, fixed

oheger-bosch (approved):

PR #4297, opened 2021-07-20
fossid-webapp: Improvements to target_path handling

nnobelis:
See individual commit messages for detail about the changes.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -319,6 +319,26 @@ class FossId(
             val filteredPackages = packages
                 .filter { packageNamespaceFilter.isEmpty() || it.id.namespace == packageNamespaceFilter }
                 .filter { packageAuthorsFilter.isEmpty() || packageAuthorsFilter in it.authors }
+                .onEach {
+                    if (it.vcsProcessed.path.isNotEmpty()) {
+                        log.warn {
+                            "Ignoring package with url ${it.vcsProcessed.url} " +
+                                    "with non-null path ${it.vcsProcessed.path}"
+                        }
+
+                        val provenance = RepositoryProvenance(it.vcsProcessed, it.vcsProcessed.revision)
+                        val summary = createSingleIssueSummary(
+                            it.id.toCoordinates(),
+                            "This package has been ignored because it contains a non-empty VCS path." +
+                                    "FossID does not support partial checkouts of a Git repository.",
+                            Severity.HINT,
+                            Instant.now()
+                        )
+
+                        val scanResult = ScanResult(provenance, details, summary)
+                        results.getOrPut(it) { mutableListOf() } += scanResult
+                    }
+                }.filter { it.vcsProcessed.path.isEmpty() }
 
             if (filteredPackages.isEmpty()) {
                 log.warn { "There is no package to scan !" }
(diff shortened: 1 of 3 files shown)


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -319,6 +319,11 @@ class FossId(
             val filteredPackages = packages
                 .filter { packageNamespaceFilter.isEmpty() || it.id.namespace == packageNamespaceFilter }
                 .filter { packageAuthorsFilter.isEmpty() || packageAuthorsFilter in it.authors }
+                .onEach {
+                    if (it.vcs.path.isNotEmpty()) {
+                        log.warn { "Ignoring package with url ${it.vcs.url} with non-null path ${it.vcs.path}" }
+                    }
+                }.filter { it.vcsProcessed.path.isEmpty() }
```
neubs-bsi, line 326: Is this tested anywhere?
To me it looks like you first check if a path is set in the vcs.path, stating that it will be ignored, and then filter out all the path of vcs processed that are NOT set. 
This seems counter-intuitive to your commit message.

neubs-bsi (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
neubs-bsi, line 326: why do you use vcsProcessed here and vcs above?

nnobelis: I was wrong, thanks.

neubs-bsi (dismissed):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -319,6 +319,11 @@ class FossId(
             val filteredPackages = packages
                 .filter { packageNamespaceFilter.isEmpty() || it.id.namespace == packageNamespaceFilter }
                 .filter { packageAuthorsFilter.isEmpty() || packageAuthorsFilter in it.authors }
+                .onEach {
+                    if (it.vcsProcessed.path.isNotEmpty()) {
+                        log.warn { "Ignoring package with url ${it.vcs.url} with non-null path ${it.vcs.path}" }
+                    }
+                }.filter { it.vcsProcessed.path.isEmpty() }
```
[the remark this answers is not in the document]
nnobelis: According to my experience, generally the whole repository is scanned with another project-package, therefore we can safely ignore these packages based on a sub-path of the repository.
However putting an `OrtIssue` is a good idea to track such filtering. But I don't want some pipeline to fail because such skips were done. Therefore I suggest we put an `OrtIssue` but with severity set to `HINT`.
Would it be acceptable ?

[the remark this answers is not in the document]
oheger-bosch: Yes, this would be fine with me.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -319,6 +319,26 @@ class FossId(
...
+                    if (it.vcsProcessed.path.isNotEmpty()) {
+                        log.warn {
+                            "Ignoring package with url ${it.vcsProcessed.url} " +
+                                    "with non-null path ${it.vcsProcessed.path}"
+                        }
+
+                        val provenance = RepositoryProvenance(it.vcsProcessed, it.vcsProcessed.revision)
+                        val summary = createSingleIssueSummary(
+                            it.id.toCoordinates(),
+                            "This package has been ignored because it contains a non-empty VCS path." +
```
oheger-bosch, line 332: Minor nit: A whitespace is missing between the two sentences.

oheger-bosch (approved):

oheger-bosch (approved):

PR #4300, opened 2021-07-22
Add a new JiraNotifier

bs-ondem:
This introduces the Jira notifier, which uses the Java REST client of
Jira in order to make the CRUD operations. Currently, the JiraNotifier
only provides a wrapper function to create new issues in Jira.

### DIFF
--- a/notifier/src/main/kotlin/modules/JiraNotifier.kt
+++ b/notifier/src/main/kotlin/modules/JiraNotifier.kt
@@ -1,0 +1,130 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.notifier.modules
+
+import com.atlassian.jira.rest.client.api.JiraRestClient
+import com.atlassian.jira.rest.client.api.RestClientException
+import com.atlassian.jira.rest.client.api.domain.BasicIssue
+import com.atlassian.jira.rest.client.api.domain.Comment
+import com.atlassian.jira.rest.client.api.domain.input.IssueInputBuilder
+import com.atlassian.jira.rest.client.internal.async.AsynchronousJiraRestClientFactory
+
+import java.net.URI
+
+import org.ossreviewtoolkit.model.Failure
+import org.ossreviewtoolkit.model.Result
+import org.ossreviewtoolkit.model.Success
+import org.ossreviewtoolkit.model.config.JiraConfiguration
+import org.ossreviewtoolkit.utils.collectMessagesAsString
+import org.ossreviewtoolkit.utils.log
+
+class JiraNotifier(
+    val config: JiraConfiguration,
+    private val restClient: JiraRestClient = AsynchronousJiraRestClientFactory()
+        .createWithBasicHttpAuthentication(URI(config.host), config.username, config.password)
+) {
+    /**
+     * Returns a [ProjectIssueBuilder] object, which can be used to do operations for the given [projectKey].
+     */
+    fun projectIssueBuilder(projectKey: String): ProjectIssueBuilder =
+        ProjectIssueBuilder(projectKey, restClient)
+
+    class ProjectIssueBuilder(
+        private val projectKey: String,
+        private val restClient: JiraRestClient
+        ) {
+        private val issueTypes = restClient.projectClient.getProject(projectKey).claim()
+            .issueTypes.associateBy { it.name }
+
+        /**
+         * Create an issue for the project with the usage of [summary], [description] and [issueType]. If an [assignee]
+         * is given, the assignee for the issue is also set. The [avoidDuplicates] flag decides whether duplicate issues
+         * will be created or not. If the flag is set to true (default), it will not create a new issue if there is
+         * already exactly one issue with the same [summary], but it will add a comment to the issue with the given
+         * [description]. If false, it will always create a new issue even if there are issues with the same [summary].
+         * Return a [Result] that encapsulates the created issue on success, or the caught exception on failure.
+         */
+        fun createIssue(
+            summary: String,
+            description: String,
+            issueType: String,
+            assignee: String? = null,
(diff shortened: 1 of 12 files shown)


notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -0,0 +1,140 @@
...
+import org.ossreviewtoolkit.utils.collectMessagesAsString
+import org.ossreviewtoolkit.utils.log
+
+class JiraNotifier(
+    val config: JiraConfiguration,
+    private val restClient: JiraRestClient = AsynchronousJiraRestClientFactory()
+        .createWithBasicHttpAuthentication(URI(config.host), config.username, config.password)
+) {
+    /** Concurrent map of all available issue types for each project */
+    private val projectIssueTypes: ConcurrentMap<String, Map<String, IssueType>> = ConcurrentHashMap()
```
oheger-bosch, line 47: Could you elaborate why this needs to be a concurrent map? I think, the check whether a project's issue types have already been loaded is not actually thread-safe, since it is not done in an atomic way.

notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -0,0 +1,140 @@
...
+                .map { it.name to it }
+                .toMap()
+        }
+
+        if (!isIssueTypeValid(projectKey, issueType)) {
+            return Failure("The issue type '$issueType' is not valid. Use a valid issue type, specified in " +
+                    "your project '$projectKey'")
+        }
+
+        val issueInputBuilder = IssueInputBuilder()
```
oheger-bosch, line 84: I am not familiar with the JIRA API, so just a question: As far as I know, JIRA can be configured to support many more properties for issues. Is this somehow reflected in the API, e.g. via generic properties? If so, should this be supported by the notifier as well?

notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -0,0 +1,140 @@
...
+     */
+    fun createIssue(
+        projectKey: String,
+        summary: String,
+        description: String,
+        issueType: String,
+        assignee: String? = null,
+        avoidDuplicates: Boolean = true
+    ): Result<BasicIssue> {
+        if (!projectIssueTypes.containsKey(projectKey)) {
```
oheger-bosch, line 72: Thinking again about this map with issue types per project, I wonder whether in a typical scenario actually multiple projects are involved. I would rather assume that all issues are created in a single project only. Based on this assumption, we could get rid off the map (and probably make the API of the notifier more convenient to use) with a slightly different approach:

Rather than providing a `createIssue()` function, `JiraNotifier` could offer a function `issuesBuilder(projectKey: String)`, which is passed a project key and returns a builder object. The builder now has the `createIssue()` function and creates all issues in this project. The issue types available for this project can be loaded once when the builder is created; so there should be no longer threading issues.

This is just a suggestion, not saying it has to be implemented that way - but do you think this would make sense?

notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -0,0 +1,140 @@
...
+     * @param issueType The issue type of the issue. The given type has to be specified in the given project in Jira.
+     * @param assignee The assignee name of the issue. The given name has to be present in the Jira instance.
+     * @param avoidDuplicates If true (default), the function will not create a new issue if there is already exactly
+     *                        one issue with the same summary, but it will add a comment to the issue with the given
+     *                        description. If false, it will always create a new issue even if there are issues with the
+     *                        same summary.
+     *
+     * @return Return a [Result] that encapsulates the created issue on success, or the caught exception on failure.
+     */
+    fun createIssue(
```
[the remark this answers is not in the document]
bs-ondem: I used the `@param` tags on purpose because the parameter list is long and the use of the tags can still be found in other places in the code. Should I rewrite it or keep it with the the tags for readability?

notifier/src/main/kotlin/modules/JiraNotifier.kt
bs-ondem: Honestly, there is no reason to use the concurrent map here. I just changed it to a `MutableMap`.

notifier/src/main/kotlin/modules/JiraNotifier.kt
bs-ondem: I think there are certainly use cases for this approach, because depending on the violation/failure, the issues could possibly be created in different projects?

notifier/src/main/kotlin/modules/JiraNotifier.kt
sschuberth: While at it, please omit the explicit type by rewriting this as `private val projectIssueTypes = mutableMapOf<String, Map<String, IssueType>>()`.

notifier/src/main/kotlin/modules/JiraNotifier.kt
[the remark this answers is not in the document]
sschuberth: We do not have strict rule for this, but as the `@param` style is only found in three other files in the source code, and so the vast majority of files use the leaner KDoc format, I personally prefer to align on using the more commonly used lean KDoc format.

notifier/src/main/kotlin/modules/JiraNotifier.kt
bs-ondem: IMO we should keep the notifier as simple as possible, if there is a need for this, we could add it later.

notifier/src/main/kotlin/modules/JiraNotifier.kt
oheger-bosch: Actually, my intension was to simplify the implementation. With the proposed approach, you would not have to maintain a map with projects and issue types, but load the information for a specific project once the builder for this project is created. But, as I said, just a proposal.

sschuberth: BTW, in Kotlin this line should be written as `if (projectKey !in projectIssueTypes) {`.

notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -0,0 +1,130 @@
...
+
+class JiraNotifier(
+    val config: JiraConfiguration,
+    private val restClient: JiraRestClient = AsynchronousJiraRestClientFactory()
+        .createWithBasicHttpAuthentication(URI(config.host), config.username, config.password)
+) {
+    /**
+     * Returns a [ProjectIssueBuilder] object, which can be used to do operations for the given [projectKey].
+     */
+    fun issueBuilder(projectKey: String): ProjectIssueBuilder =
```
oheger-bosch, line 46: Minor nit: Should this function now better be called `projectIssueBuilder()`? Sorry for not noticing earlier.

oheger-bosch (approved):
Looks good to me except for this minor naming issue.

notifier/src/main/kotlin/modules/JiraNotifier.kt
bs-ondem: Yes, it makes sense. I changed the function name to `projectIssueBuilder`.

oheger-bosch (approved):

PR #4306, opened 2021-07-27
vulnerabilities added to EvaluatedModel and make dependent tests green

ashykhov:
Signed-off-by: Artem Shykhov <ext-artem.shykhov@here.com>


reporter/src/main/kotlin/model/EvaluatedModelMapper.kt
```diff
@@ -188,6 +179,26 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
         }
     }
 
+    private fun createVulnerabilities() {
+        input.ortResult.advisor
+            ?.results
+            ?.advisorResults
+            ?.flatMap { (identifier, resultsList) ->
```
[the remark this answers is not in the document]
fviernau: Suggest to rename `resultsList` to just `results`.

reporter/src/main/kotlin/model/EvaluatedModelMapper.kt
```diff
@@ -188,6 +179,26 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
...
+                val pkg = packages[identifier] ?: createEmptyPackage(identifier)
+                resultsList.flatMap { result ->
```
fviernau, line 188: nit: new line before this one.

reporter/src/main/kotlin/model/EvaluatedVulnerability.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude
+import org.ossreviewtoolkit.model.VulnerabilityReference
+import org.ossreviewtoolkit.model.config.VulnerabilityResolution
+
+data class EvaluatedVulnerability(
+    val id: String,
+    val pkg: EvaluatedPackage,
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    val references: List<VulnerabilityReference>,
+    val resolutions: List<VulnerabilityResolution>
```
fviernau, line 12: Why did you decide to use "include non-empty" only for one of the two properties?

fviernau (changes requested):

reporter/src/funTest/kotlin/reporters/EvaluatedModelReporterFunTest.kt
```diff
@@ -23,39 +23,68 @@ import io.kotest.core.TestConfiguration
...
 class EvaluatedModelReporterFunTest : WordSpec({
     "EvaluatedModelReporter" should {
-        "create the expected JSON output" {
-            val expectedResult = File(
-                "src/funTest/assets/evaluated-model-reporter-test-expected-output.json"
-            ).readText()
-            val ortResult = readOrtResult("src/funTest/assets/static-html-reporter-test-input.yml")
 
-            generateReport(ortResult).normalizeLineBreaks() shouldBe expectedResult
+        "create the general expected JSON output" {
```
fviernau, line 48: Why did you decide to create a separate test asset?
e.g. instead of adding vulnerabilities to the existing one.

fviernau (changes requested):

reporter/src/funTest/assets/evaluated-model-reporter-test-expected-output.json
```diff
@@ -831,6 +831,8 @@
     "message" : "BSD-3-Clause warning",
     "how_to_fix" : "* *Step 1*\n* __Step 2__\n* ***Step 3***\n```\nSome long text verify that overflow:scroll is working as expected.\n```"
```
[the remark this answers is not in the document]
ashykhov: got it, i will fix it

reporter/src/main/kotlin/model/EvaluatedModel.kt
```diff
@@ -29,24 +29,13 @@ import com.fasterxml.jackson.databind.json.JsonMapper
...
-import org.ossreviewtoolkit.model.RuleViolation
-import org.ossreviewtoolkit.model.config.IssueResolution
-import org.ossreviewtoolkit.model.config.LicenseFindingCuration
-import org.ossreviewtoolkit.model.config.PathExclude
-import org.ossreviewtoolkit.model.config.RepositoryConfiguration
-import org.ossreviewtoolkit.model.config.RuleViolationResolution
-import org.ossreviewtoolkit.model.config.ScopeExclude
-import org.ossreviewtoolkit.model.mapperConfig
+import org.ossreviewtoolkit.model.*
+import org.ossreviewtoolkit.model.config.*
```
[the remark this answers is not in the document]
ashykhov: sorry for that, it will be fixed

reporter/src/main/kotlin/model/EvaluatedModel.kt
```diff
@@ -29,12 +29,24 @@ import com.fasterxml.jackson.databind.json.JsonMapper
 import com.fasterxml.jackson.databind.node.ArrayNode
 import com.fasterxml.jackson.databind.node.ObjectNode
 import com.fasterxml.jackson.dataformat.yaml.YAMLMapper
-import org.ossreviewtoolkit.model.*
-import org.ossreviewtoolkit.model.config.*
+
```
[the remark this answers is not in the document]
ashykhov: wil be fixed as well

reporter/src/funTest/kotlin/reporters/EvaluatedModelReporterFunTest.kt
ashykhov: It's really hard to understand what is wrong when something fails, as well as create new test when you are working with assertion about 1000 rows. I  thikn it would be better to have few separated tests for different parts and cases instead of one test for everything. Am i wrong with my thoughts?

fviernau: The approach of this test already before was testing many different things. E.g. the test asset IMO intends to cover as many use cases as possible. I prefer to not refactor this approach now, but just add the vulnerabilities to it, so that it's cover by the static html and evaluated model test. 

Moving more towards testing less things at a time might make sense in general. I believe applying that approach is a bigger topic and needs way more refactoring than done HERE. Separating out only the vulnerabilities while testing all other things together seems inconsistent to me and introduces redundancy. Also the need for testing both JSON and YAML does not exist anymore for both cases.

PR #4310, opened 2021-07-28
Unmanged: Improve the docs for this special override of resolveDependencies()

sschuberth:


analyzer/src/main/kotlin/managers/Unmanaged.kt
```diff
@@ -55,10 +55,8 @@ class Unmanaged(
     }
 
     /**
-     * Return a [ProjectAnalyzerResult] for the [Project] contained in the [definitionFile] directory, but does not
-     * perform any dependency resolution.
-     *
-     * @param definitionFile The directory containing the unmanaged project.
+     * Return a list with a single [ProjectAnalyzerResult] for the "unmanaged" [Project] defined by the
```
fviernau, line 58: Following Liskov substitution principle, it is sufficient for the user to look only at the documentation of the base class. So, every documentation in any overridden function can be missed. Therefore I wonder whether we should invent the general rule to not allow documentation in overridden functions and instead add this documentation to 
the base class.

Another option could be to move this documentation from this function to the Unmanage class documentation.

What do you think?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Unmanaged.kt
sschuberth: I think that in this special case it doe snot apply, as *this* override is fundamentally different compared to the base functionality. That's why the docs existed in the first place, I believe. But generally, I agree that generic function docs that apply to all overrides should go to the base class only.

fviernau (approved):

PR #4318, opened 2021-07-30
Extract a configuration for FossId

oheger-bosch:
This PR is a minor refactoring of FossId with the aim to make this class easier to test. The idea is to extract the various parameters into a dedicated configuration class, which also supports the creation of helper objects. By passing a mock configuration, it is then possible to inject mocks for these helpers.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -61,7 +61,6 @@ import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.config.ScannerOptions
 import org.ossreviewtoolkit.scanner.AbstractScannerFactory
 import org.ossreviewtoolkit.scanner.RemoteScanner
-import org.ossreviewtoolkit.scanner.scanners.fossid.FossId.Companion.NAMING_CONVENTION_VARIABLE_PREFIX
 import org.ossreviewtoolkit.utils.log
 import org.ossreviewtoolkit.utils.replaceCredentialsInUri
 import org.ossreviewtoolkit.utils.showStackTrace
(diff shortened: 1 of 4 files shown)


sschuberth:
Tests fail due to `Logging is only allowed on ORT classes, but 'kotlin.String' is used.`

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -111,4 +122,23 @@ internal data class FossIdConfig(
             )
         }
     }
+
+    /**
+     * Create a [FossIdNamingProvider] helper object based on the configuration stored in this object.
+     */
+    fun createNamingProvider(): FossIdNamingProvider {
+        val namingProjectPattern = options[NAMING_PROJECT_PATTERN_PROPERTY]?.apply {
```
sschuberth, line 135: Using `also` here (and below) instead of `apply` should get rid of the log-related test failures.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -111,4 +122,23 @@ internal data class FossIdConfig(
...
+            log.info { "Naming pattern for projects is $this." }
```
sschuberth, line 136: With `also`, `$this` needs to be changed to `$it`.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
sschuberth: I still have a slight preference for using `also` instead of `apply` to avoid the need for the explicit `FossIdConfig.` prefix, but I'll approve anyway.

sschuberth (approved):

nnobelis:
Can you prefix the commit message with `fossid-webapp` for easy searching ?

scanner/src/test/kotlin/scanners/fossid/FossIdConfigTest.kt
```diff
@@ -113,6 +113,44 @@ class FossIdConfigTest : WordSpec({
...
+                "user" to USER,
+                "namingProjectPattern" to "#projectName_\$Org_\$Unit",
+                "namingVariableOrg" to "TestOrganization",
+                "namingVariableUnit" to "TestUnit"
+            ).toScannerConfig()
+
+            val fossIdConfig = FossIdConfig.create(scannerConfig)
+            val namingProvider = fossIdConfig.createNamingProvider()
+
+            val projectName = namingProvider.createProjectCode("TestProject")
```
nnobelis, line 138: IMH you are mixing the tests here. A config test should only be about retrieving the configuration.
The call to `createProjectCode` and the verification of its return value should be in a `namingConventionTest.kt` or something.
Same comment underneath.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -141,4 +146,20 @@ internal data class FossIdConfig(
 
         return FossIdNamingProvider(namingProjectPattern, namingScanPattern, namingConventionVariables)
     }
+
+    /**
+     * Create the [FossIdRestService] to interact with the FossID instance described by this configuration.
+     */
+    fun createService(): FossIdRestService {
+        val client = OkHttpClientHelper.buildClient()
+        log.info { "FossID server URL is $serverUrl." }
```
nnobelis, line 155: Could you please also modify  `log4j2.xml` to add this class ?
Since you moved the log message to this class it won't get logged by default.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -97,14 +92,15 @@ import retrofit2.converter.jackson.JacksonConverterFactory
  * [NAMING_CONVENTION_VARIABLE_PREFIX] e.g. namingVariableVar1 = "foo".
```
nnobelis, line 92: Should we not remove all the duplicated documentation to allow having to maintain documentation in two places ?

For instance we could put here `Options of the scanner including documentation is available in [FossIdConfig].`

nnobelis (changes requested):

scanner/src/test/kotlin/scanners/fossid/FossIdConfigTest.kt
oheger-bosch: The purpose of this test is to check whether a correct naming provider is created (which is functionality of `FossIdConfig`). The naming provider does not allow checking its internal properties, so the only way to check its initialization is to invoke it to generate names. It is true that these tests could be done in a separate test class, but testing this here was the easiest way to achive this.

nnobelis: I think the solution here is to mark the properties of the `NamingProvider`  as `internal` so you can test them.
But the decision is up to you I won't fight about it.

oheger-bosch: For `FossIdNamingProvider` this would be possible, but for `FossIdRestService`, I see no easy way how this could be done.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Moved the description of configuration options to the `FossIdConfig` class.

nnobelis (dismissed):
You can merge, my comments have been addressed.

nnobelis (approved):

sschuberth (approved):

PR #4320, opened 2021-07-30
helper-cli: Add a command which shows the licenses of a package

fviernau:
The command allows querying the detected license as well as the root
license from the configured scan result storage, given any package
identifier.

This allows to determine the detected license given a package
identifier relatively quickly compared to determining the detected
license by locating, downloading and parsing an ORT result file
containing a scan result for a given package identifier. In particular
for large ORT result file the latter is much slower.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
@@ -1,0 +1,134 @@
+/*
+ * Copyright (C) 2021 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


sschuberth:
Instead of only summing up in prose what the code does, could you please also explain the usage scenario in the commit message, when such a comment becomes useful?

fviernau:
> Instead of only summing up in prose what the code does, could you please also explain the usage scenario in the commit message, when such a comment becomes useful?

I updated the commit message accordingly.

helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+private fun ScanResultsStorage.getScanResults(id: Identifier): List<ScanResult> =
+    when (val status = read(id)) {
+        is Success -> status.result
+        is Failure -> emptyList()
+    }
+
+data class Result(
```
mnonnenmacher, line 129: Can this class be private as well?

mnonnenmacher (dismissed):

helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
fviernau: Yes, done.

mnonnenmacher (approved):

PR #4327, opened 2021-08-03
Change for FossID 21.2

nnobelis:
This changes are required to make ORT work with FossID 21.2.
ORT is still backward compatible with older FossID versions.

Refer to individual commit messages for description.


clients/fossid-webapp/src/main/kotlin/model/status/Description.kt
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.clients.fossid.model.status
+
+interface Description {
```
oheger-bosch, line 22: The name `Description` is quite generic. What does it describe? Should the interface be named `ScanDescription`, and another name should be searched for the class? Also some comments would be helpful.

clients/fossid-webapp/src/main/kotlin/VersionedFossIdService21dot2.kt
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.clients.fossid
+
+import org.ossreviewtoolkit.clients.fossid.model.status.Description
+
+class VersionedFossIdService21dot2(
```
oheger-bosch, line 24: Regarding naming: The new classes are now named with a specific version, but the meaning is actually "this version or newer". I am just thinking whether it would be clearer to name the most recent class without a version and somehow mark the legacy classes. Not sure what is better.

clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
```diff
@@ -23,6 +23,7 @@ class EntityResponseBody<T>(
     val operation: String? = null,
     val status: Int? = null,
     val message: String? = null,
+    val messageParameters: Array<String>? = null,
```
oheger-bosch, line 26: Just a question: Do we evaluate this property somewhere or is it just mandatory to declare it because it is sent by FossID? If so, would it make sense to configure the JSON mapper to ignore unknown properties, so that we could keep our models minimalistic?

nnobelis: Currently I don't know the content of this property, it is non documented.
I would like to have it inside to see what is coming: this could be useful.

clients/fossid-webapp/src/main/kotlin/VersionedFossIdService21dot2.kt
nnobelis: Yes I think about this too. The problem is we have not so much visibility about this...
What we could do, we could mark the legacy classes with the highest supported version. And the current one without ?

The problem I see is the Git history: when a new version is coming, you need to move the current class to a legacy class with version in one commit. Then in another commit you can create a new class for the actual version.
Otherwise you risk of losing Git history for the legacy class.

clients/fossid-webapp/src/main/kotlin/model/status/Description.kt
nnobelis: So you understand the problem I had: `ScanStatus` is the status of a scan. `Scan` is the scan itself. `ScanState` is the state of a scan. `ScanDescriptionXXX` is the description of a scan in version XXX.
Since here we are talking about a non-version specific scan description, what about `GenericScanDescription` or `UnversionedScanDescription` ?

oheger-bosch: Yes, the terminology used by FossID seems to be quite confusing. Especially the distinction between `ScanStatus` and `ScanState` I find difficult to grasp, but this is out of our control. Regarding `Description`, I would be fine with either name, only the plain `Description` I find rather meaningless.

oheger-bosch (approved):
Regarding the naming of the old and new classes I am still a bit undecided. Other than that, the changes look good to me. Classes can be renamed in a later PR if there is an agreement.

PR #4335, opened 2021-08-04
PackageManager: Align on logging the relative path of the definiton file

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/PackageManager.kt
+++ b/analyzer/src/main/kotlin/PackageManager.kt
@@ -229,7 +229,9 @@ abstract class PackageManager(
         beforeResolution(definitionFiles)
 
         definitionFiles.forEach { definitionFile ->
-            log.info { "Resolving $managerName dependencies for '$definitionFile'..." }
+            val relativePath = definitionFile.relativeTo(analysisRoot).invariantSeparatorsPath
+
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -229,7 +229,7 @@ abstract class PackageManager(
         beforeResolution(definitionFiles)
 
         definitionFiles.forEach { definitionFile ->
-            log.info { "Resolving $managerName dependencies for '$definitionFile'..." }
+            val relativePath = definitionFile.relativeTo(analysisRoot).invariantSeparatorsPath
```
mnonnenmacher, line 232: Sometimes the analyzer gets stuck with a project, in such situations the removed log line was very useful. It's also useful to know which definition file the following log statements belong to. Therefore I would prefer to keep it.

sschuberth: > It's also useful to know which definition file the following log statements belong to.

Each of the following log statements already show the definition file by themselves.

> Sometimes the analyzer gets stuck with a project, in such situations the removed log line was very useful.

Ok, agreed. I'll then change the intention of this PR to only consistently show the relative path of the definition file.

mnonnenmacher (approved):

PR #4343, opened 2021-08-06
helper-cli: GetPackageLicensesCommand improvements

fviernau:


helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
```diff
@@ -101,13 +102,12 @@ class GetPackageLicensesCommand : CliktCommand(
                 .applyAll(nonExcludedLicenseFindings, licenseFindingCurations)
                 .mapNotNull { it.curatedFinding }
 
-            val detectedLicense = curatedFindings.map { it.license.toString() }
-                .distinct().sorted().joinToString(" AND ")
+            val detectedLicense = curatedFindings.toSpdxExpression()
```
mnonnenmacher, line 107: Commit message nit: `asSequence()`

helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
```diff
@@ -121,7 +122,7 @@ class GetPackageLicensesCommand : CliktCommand(
 }
 
 private fun Collection<LicenseFinding>.toSpdxExpression(): String =
-    asSequence().map { it.license.toString() }.distinct().sorted().joinToString(" AND ")
+    asSequence().map { it.license.toString() }.distinct().joinToString(" AND ").toSpdx().sort().toString()
```
mnonnenmacher: There is no need to convert to string before concatenating and parsing back into an SPDX expression (could also be applied in the previous commit already):
```
asSequence().map { it.license }.reduce(SpdxExpression::and).sort().toString()
```
(not sure if `asSequence()` still makes sense with this code)

mnonnenmacher (changes requested):

helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
fviernau: > (not sure if `asSequence()` still makes sense with this code)

The IDE still suggested it, so I kept it.

helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
```diff
@@ -121,7 +122,7 @@ class GetPackageLicensesCommand : CliktCommand(
 }
 
 private fun Collection<LicenseFinding>.toSpdxExpression(): String =
-    asSequence().distinct().sorted().joinToString(" AND ")
+    asSequence().distinct().joinToString(" AND ").toSpdx().sort().toString()
```
mnonnenmacher: You are now converting the `LicenseFinding`s to strings and still not use `reduce()`:
```
asSequence().map { it.license }.distinct().reduce(SpdxEpxression::and).sort().toString()
```

mnonnenmacher (changes requested):

helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
fviernau: I've added the reduce in a separate commit now.

mnonnenmacher (approved):

PR #4344, opened 2021-08-09
spdx-utils: Regenerate SPDX enums to update to license list version 3.14

tsteenbe:
Done by running "./gradlew :spdx-utils:generateSpdxEnums". For details
see https://github.com/spdx/license-list-XML/releases/tag/v3.14.

Signed-off-by: Thomas Steenbergen <thomas.steenbergen@here.com>


fviernau (approved):

fviernau:
@tsteenbe have you checked whether `generateSpdxLicenseExceptionEnum` also has some updates?

spdx-utils/src/main/resources/licenses/GPL-2.0-or-later
```diff
@@ -10,343 +10,3 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with
 this program; if not, write to the Free Software Foundation, Inc., 51 Franklin
 Street, Fifth Floor, Boston, MA  02110-1301, USA.
-
```
fviernau, line 13: The (CI) test fails due to the changes in this file. @tsteenbe can you confirm the changes in this file are good?

tsteenbe: @fviernau Then it must be some logic on our side as https://github.com/spdx/license-list-XML/blob/v3.14/src/GPL-2.0-or-later.xml has not changed since July 4th 2019.

fviernau: I wasn't aware that ORT actually fetches the license texts for the SPDX licenses from ScanCode.
In particular this license is fetched from the file `licenses/scancode-toolkit/gpl-2.0-plus.LICENSE`.
A recent change in ORT aligned the version of the imported scancode resources with the version of ScanCode used in ORT [2]. So, as of [2] the license texts are not fetched anymore from ScanCode's `develop` branch, but from tag `3.2.1rc2`.
This explains the downgrade of the license text observed here.

That's the situation - the question now is whether it is good as is...?

[1] https://github.com/nexB/scancode-toolkit/commit/cdda9c0fda513af86e864998ab7f6e4facfab70e#diff-db221c21b7d27dbf84985089c9702a5092ba5fcb9db519dbe7e86f51e0daea49
[2] https://github.com/oss-review-toolkit/ort/commit/ded4f202862678f9d3be13e902539dd72bbed3e0

sschuberth: > Does anyone call the reason why SPDX texts are imported from ScanCode instead of from the SPDX license list repository?

IIRC the license text formatting in the SPDX list was severely broken, and so we preferred to fetch from ScanCode instead. However, that [issue](https://github.com/spdx/LicenseListPublisher/issues/78) seems to have been fixed by now.

So I guess what we should do is: Yes, align ScanCode version and ScanCode-specific license texts, but get SPDX license texts from the SPDX list only.

fviernau: Thanks @sschuberth . Your proposal makes sense to me.

sschuberth: I'd propose then to do the "fetch SPDX license texts from SPDX list only" change as a separate commit (or even separate PR) to better see the impact of that change only on the formatting of license texts, to verify the formatting *really* is not broken anymore.

mnonnenmacher:
This update will be done as part of #4523.

spdx-utils/src/main/resources/licenses/GPL-2.0-or-later
mnonnenmacher: I have tried this when working on #4523 but the license texts provided by SPDX are still badly formatted so we decided to stick with the ScanCode texts for now.

PR #4349, opened 2021-08-10
evaluator: Add the new rule type `OrtResultRule`

fviernau:
Having a generic rule which is executed once per ORT result helps with use
cases for which the more specific rules aren't a good fit. As a pre-requisite the field
`RuleViolation.pkgId` is made nullable, which eliminates the assumption that a rule violation always corresponds to a package.

NOTE: The PR has to be amended to properly handle the now nullable attribute in the `web-app`.

### DIFF
--- a/evaluator/src/main/kotlin/OrtResultRule.kt
+++ b/evaluator/src/main/kotlin/OrtResultRule.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2021 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.evaluator
+
+import org.ossreviewtoolkit.model.OrtResult
+
+/**
+ * A [Rule] to check an [OrtResult].
+ */
+open class OrtResultRule(
+    ruleSet: RuleSet,
+    name: String,
+
+    /**
+     * The [OrtResult] to check.
+     */
+    val ortResult: OrtResult,
+) : Rule(ruleSet, name) {
+    override val description = "Evaluating ORT result rule '$name'."
+
+    override fun issueSource() = "$name - ORT result"
+
+    fun error(message: String, howToFix: String): Unit =
+        error(
+            pkgId = null,
+            license = null,
+            licenseSource = null,
+            message = message,
+            howToFix = howToFix
+        )
+}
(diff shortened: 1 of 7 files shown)


reporter/src/main/kotlin/reporters/StaticHtmlReporter.kt
```diff
@@ -295,7 +295,7 @@ class StaticHtmlReporter : Reporter {
                 }
             }
             td { +ruleViolation.violation.rule }
-            td { +ruleViolation.violation.pkg.toCoordinates() }
+            td { +ruleViolation.violation.pkg?.toCoordinates().orEmpty() }
```
mnonnenmacher, line 298: Consider `?: "-"` instead of `orEmpty()` to be consistent with the next column.

reporter/src/main/kotlin/model/EvaluatedRuleViolation.kt
```diff
@@ -32,7 +32,7 @@ import org.ossreviewtoolkit.model.config.RuleViolationResolution
 data class EvaluatedRuleViolation(
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     val rule: String,
-    val pkg: EvaluatedPackage,
+    val pkg: EvaluatedPackage?,
```
mnonnenmacher, line 35: This might require changes in the reporter-web-app.

evaluator/src/main/kotlin/OrtResultRule.kt
```diff
@@ -0,0 +1,53 @@
...
+     * The [OrtResult] to check.
+     */
+    val ortResult: OrtResult,
+) : Rule(ruleSet, name) {
+    override val description = "Evaluating ORT result rule '$name'."
+
+    override fun issueSource() = "$name - ORT result"
+
+    init {
+        matchers += RuleMatcher.MATCH_ALWAYS
```
mnonnenmacher, line 42: This should not be required, if no matchers are configured the default behavior is already that the rule always applies.

mnonnenmacher (changes requested):

reporter/src/main/kotlin/model/EvaluatedRuleViolation.kt
fviernau: Yes, it does. See also description of the PR.

evaluator/src/main/kotlin/OrtResultRule.kt
fviernau: Oh, thanks. I must have misunderstood the code then.

sschuberth:
> helps with use cases for which the more specific rules aren't a good fit.

Could you amend the commit message with an example for such a use-case?

reporter/src/main/kotlin/reporters/StaticHtmlReporter.kt
fviernau: Good catch!

fviernau:
> Could you amend the commit message with a concrete example for such a use-case?
Done.

reporter/src/main/kotlin/model/EvaluatedRuleViolation.kt
fviernau: Addressed by #4450 .

mnonnenmacher (approved):
Can you please add an example to `examples/rules.kts`, maybe with a simple check for a label? And also check if any documentation needs to be updated? I'm fine with doing that in a separate PR.

fviernau:
> Can you please add an example to `examples/rules.kts`, maybe with a simple check for a label? And also check if any documentation needs to be updated? I'm fine with doing that in a separate PR.

Yes, will do a follow-up PR.

fviernau:
> Can you please add an example to `examples/rules.kts`, maybe with a simple check for a label? And also check if any documentation needs to be updated? I'm fine with doing that in a separate PR.

@mnonnenmacher I propose to merge this example: #4499 .

PR #4352, opened 2021-08-10
Conan: Take local configuration (like remotes) into account

### LINKED ISSUE #4312: Support custom Conan repositories
By default Conan expects to find dependencies on [ConanCenter](https://conan.io/center/). However, similar to Maven and NPM, also Conan allows to declare custom / internal artifact repositories via [remotes](https://docs.conan.io/en/latest/uploading_packages/remotes.html). ORT currently does not support these, but it obviously should.

sschuberth:
This applies configuration stored to a conventional "conan_config"
directory before the dependency resolution. This is needed in order to
support custom remotes besides the default ConanCenter.

Resolves #4312.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Conan.kt
+++ b/analyzer/src/main/kotlin/managers/Conan.kt
@@ -108,8 +108,15 @@ class Conan(
         // [2]: https://docs.conan.io/en/latest/configuration/download_cache.html#download-cache
         val conanStoragePath = conanHome.resolve("data")
 
-        stashDirectories(conanStoragePath).use {
-            val workingDir = definitionFile.parentFile
+        val workingDir = definitionFile.parentFile
+        val conanConfig = listOf(workingDir, analysisRoot).map { it.resolve("conan_config") }
+            .find { it.isDirectory }
+        val directoryToStash = conanConfig?.let { conanHome } ?: conanStoragePath
+
+        stashDirectories(directoryToStash).use {
+            conanConfig?.also {
+                run("config", "install", it.absolutePath)
+            }
 
             installDependencies(workingDir)
 


oheger-bosch (approved):
Changes look good to me.

As usual, I feel uneasy if non-trivial changes are introduced without tests. Unfortunately, there is no unit test class for Conan, and testing additional remote repositories via a functional test, is probably tricky as well.

sschuberth:
> Unfortunately, there is no unit test class for Conan, and testing additional remote repositories via a functional test, is probably tricky as well.

Indeed. At least I can assure you that I have tested these changes locally with a concrete internal project. Thanks for the review!

nnobelis:
@sschuberth a customer of us has his `remotes.txt` file in a non standard location `tools/conan` and does not want to create a `conan_config` directory.

Looking at the specs, Conan does not enforce where the `remotes.txt` should be located in the repository
https://docs.conan.io/en/latest/reference/commands/consumer/config.html.

I also saw the TODO here :
https://github.com/oss-review-toolkit/ort/blob/be5b5f1ef3f0d8126e74eaf20ad69d003137399b/analyzer/src/main/kotlin/managers/Conan.kt#L136

but I think an alternative could be that ORT automagically calls `config install` on any `remotes.txt` file it founds in the repository. What do you think ?
This would remove the need of a specific ORT configuration.

sschuberth:
> an alternative could be that ORT automagically calls `config install` on any `remotes.txt` file it founds in the repository.

Sounds a little bit too risky to me given that `remotes.txt` is a rather generic name. Also, maybe people store different `remotes.txt` files in their repo that are supposed to be used mutually exclusively depending on the product flavor to build. So I'd still prefer the configurable `remotes.txt` location.

PR #4360, opened 2021-08-13
Dockerfile: Upgrade Node to version 16

fviernau:
Upgrading Node from version 12 to version 16 also leads to an upgrade of
Npm. So, pin the Npm version accordingly to the latest release. See also
[1] and [2].

This change fixes [3] for unknown reason.

[1] https://github.com/nodejs/node/blob/7ca38f05a023666274569343f128c5aed81599f3/doc/changelogs/CHANGELOG_V16.md
[2] https://github.com/npm/cli/blob/59b9851d1649b8f78203cae187b1337f2441219d/CHANGELOG.md#v7206-2021-08-12
[3] https://stackoverflow.com/questions/68021121/fsevents-not-accessible-from-jest-haste-map.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ ENV \
     GO_DEP_VERSION=0.5.4 \
     GO_VERSION=1.16.5 \
     HASKELL_STACK_VERSION=2.1.3 \
-    NPM_VERSION=6.14.2 \
+    NPM_VERSION=7.20.6 \
     PYTHON_PIPENV_VERSION=2018.11.26 \
     PYTHON_VIRTUALENV_VERSION=15.1.0 \
     SBT_VERSION=1.3.8 \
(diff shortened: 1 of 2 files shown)


sschuberth:
Doesn't upgrading Node also upgrade the bundled `npm`? And if so, to what version? Does that version still match our `Requirement.buildNPM("5.7.* - 6.14.*")`?

fviernau:
> Doesn't upgrading Node also upgrade the bundled `npm`? And if so, to what version? Does that version still match our `Requirement.buildNPM("5.7.* - 6.14.*")`?

Thanks for the hint! Let me check it...

Dockerfile
```diff
@@ -54,7 +54,7 @@ ENV \
     GO_DEP_VERSION=0.5.4 \
     GO_VERSION=1.16.5 \
     HASKELL_STACK_VERSION=2.1.3 \
-    NPM_VERSION=6.14.2 \
+    NPM_VERSION=7.20.3 \
```
mnonnenmacher, line 57: Why not update to the latest version 7.20.6?

mnonnenmacher: Commit message:
* It's a bit confusing that you mention node in the title but NPM in the body.
* Could you shortly explain why it fixes the linked issue? Because it's not apparent from the Stackoverflow thread how that's related to the node version.
* Could you link the changelog / release notes as we usually do?

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #4361, opened 2021-08-13
Spdx document file handling improvements

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -340,7 +340,7 @@ class SpdxDocumentFile(
     }
 
     /**
-     * Get the [SpdxPackage] for the given [identifier] by resolving against packages an external document references
+     * Get the [SpdxPackage] for the given [identifier] by resolving against packages or external document references
      * contained in [doc].
      */
     private fun getSpdxPackageForId(doc: SpdxDocument, identifier: String, workingDir: File): SpdxPackage {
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -350,7 +350,7 @@ class SpdxDocumentFile(
         val externalDocumentReference = doc.externalDocumentRefs.find {
             it.externalDocumentId == documentRef
         } ?: throw IllegalArgumentException(
-            "No single package or externalDocumentRef with ID '$identifier' found."
+            "ID '$identifier' could neither be resolved as a package nor as an externalDocumentRef."
```
mnonnenmacher, line 350: "resolved to" instead of "resolved as".

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: I was using "as" intentionally, but I'm also fine with "to". Will change in both places.

sschuberth: Or did you really only want to get "as" changed to "to" in the first occurrence?

mnonnenmacher: I meant both occurrences of "as", I'm not a native speaker but "resolve as" just sounds wrong to me.

mnonnenmacher: Anyway, if you want to keep it as is that's fine with me.

sschuberth: Changed in both places.

fviernau (approved):

PR #4366, opened 2021-08-16
FossID: Limit the number of delta scans

oheger-bosch:
Please refer to the single commits for further details.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -320,9 +320,9 @@ class FossId internal constructor(
                     val scanResult = ScanResult(provenance, details, summary)
                     results.getOrPut(pkg) { mutableListOf() } += scanResult
 
-                    createdScans.forEach {
-                        log.warn("Deleting previous scan $it.")
-                        deleteScan(it)
+                    createdScans.forEach { code ->
+                        log.warn { "Deleting scan $code during exception cleanup." }
+                        deleteScan(code)
                     }
                 }
             }
@@ -345,16 +345,13 @@ class FossId internal constructor(
         return ScanSummary(startTime, Instant.now(), "", sortedSetOf(), sortedSetOf(), listOf(issue))
     }
 
-    private suspend fun List<Scan>.findLatestPendingOrFinishedScan(
-        url: String,
-        revision: String? = null
-    ): Scan? =
-        filter {
-            // The scans in the server contain the url with the credentials so we have to remove it for the
-            // comparison. If we don't, the scans won't be matched if the password changes!
-            val urlWithoutCredentials = it.gitRepoUrl?.replaceCredentialsInUri()
-            urlWithoutCredentials == url && (revision == null || (it.gitBranch == revision))
-        }.sortedByDescending { scan -> scan.id }.find { scan ->
+    /**
+     * Find the latest [Scan] in this list with a finished state. If necessary, wait for a scan to finish. Note that
+     * this function expects that [recentScansForRepository] has been applied to this list for the current
+     * repository.
+     */
+    private suspend fun List<Scan>.findLatestPendingOrFinishedScan(): Scan? =
+        find { scan ->
             val scanCode = requireNotNull(scan.code) {
                 "FossId returned a null scancode for an existing scan."
             }
@@ -374,6 +371,20 @@ class FossId internal constructor(
             }
         }
 
+    /**
+     * Filter this list of [Scan]s for the repository defined by [url] and an optional [revision] and sort it by
+     * scan ID, so that the most recent scan comes first and the oldest scan comes last.
+     */
+    private fun List<Scan>.recentScansForRepository(
+        url: String,
+        revision: String? = null
+    ): List<Scan> = filter {
+        // The scans in the server contain the url with the credentials so we have to remove it for the
+        // comparison. If we don't, the scans won't be matched if the password changes!
+        val urlWithoutCredentials = it.gitRepoUrl?.replaceCredentialsInUri()
+        urlWithoutCredentials == url && (revision == null || (it.gitBranch == revision))
+    }.sortedByDescending { scan -> scan.id }
+
     private suspend fun checkAndCreateScan(
         scans: List<Scan>,
         url: String,
@@ -381,7 +392,7 @@ class FossId internal constructor(
         projectCode: String,
         projectName: String
     ): String {
-        val existingScan = scans.findLatestPendingOrFinishedScan(url, revision)
+        val existingScan = scans.recentScansForRepository(url, revision).findLatestPendingOrFinishedScan()
 
         val scanCode = if (existingScan == null) {
             log.info { "No scan found for $url and revision $revision. Creating scan ..." }
@@ -414,7 +425,8 @@ class FossId internal constructor(
         projectName: String
     ): String {
         // we ignore the revision because we want to do a delta scan
-        val existingScan = scans.findLatestPendingOrFinishedScan(url)
+        val recentScans = scans.recentScansForRepository(url)
+        val existingScan = recentScans.findLatestPendingOrFinishedScan()
 
         val scanCode = if (existingScan == null) {
             log.info { "No scan found for $url and revision $revision. Creating origin scan ..." }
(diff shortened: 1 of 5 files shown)


scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -45,6 +45,9 @@ import retrofit2.converter.jackson.JacksonConverterFactory
  * * **"deltaScans":** When set, ORT will create delta scans. When only changes in a repository need to be scanned,
  * delta scans reuse the identifications of latest scan on this repository to reduce the amount of findings. If
  * deltaScans is set and no scan exist yet, an initial scan called "origin" scan will be created.
+ * * **"deltaScanLimit":** This setting can be used to limit the number of delta scans for a repository. So if another
```
nnobelis, line 48: ...the number of delta scan **to keep** for a given...

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -45,6 +45,9 @@ import retrofit2.converter.jackson.JacksonConverterFactory
...
+ * delta scan is created, older delta scans are deleted until this number is reached. If unspecified, no limit is
+ * enforced on the number of delta scans. This property is evaluated only if *deltaScans* is enabled.
```
nnobelis, line 50: ...of delta scans **to keep**.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -345,16 +345,12 @@ class FossId internal constructor(
...
-        revision: String? = null
-    ): Scan? =
-        filter {
-            // The scans in the server contain the url with the credentials so we have to remove it for the
-            // comparison. If we don't, the scans won't be matched if the password changes!
-            val urlWithoutCredentials = it.gitRepoUrl?.replaceCredentialsInUri()
-            urlWithoutCredentials == url && (revision == null || (it.gitBranch == revision))
-        }.sortedByDescending { scan -> scan.id }.find { scan ->
+    /**
+     * Find the latest [Scan] in this (already filtered and sorted) list with a finished state. If necessary, wait for
```
nnobelis, line 349: ...(already filtered and sorted)... according which criteria ?
IMO you can remove this comment.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -132,6 +141,7 @@ internal data class FossIdConfig(
             val addAuthenticationToUrl = fossIdScannerOptions[CREDENTIALS_IN_URL_PROPERTY].toBoolean()
             val waitForResult = fossIdScannerOptions[WAIT_FOR_RESULT_PROPERTY]?.toBooleanStrictOrNull() ?: true
             val deltaScans = fossIdScannerOptions[DELTA_SCAN_PROPERTY].toBoolean()
+            val deltaScanLimit = fossIdScannerOptions[DELTA_SCAN_LIMIT_PROPERTY]?.toInt() ?: Int.MAX_VALUE
```
nnobelis, line 144: You should enforce `deltaScanLimit` > 0 (it's noticeable where you use it and do `-1`).

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -567,6 +594,7 @@ class FossId internal constructor(
      * Delete a scan with [scanCode].
      */
     private suspend fun deleteScan(scanCode: String) {
+        log.info { "Deleting scan $scanCode." }
```
nnobelis, line 597: In the other place where the function is called we already do `log.warn("Deleting previous scan $it.")`.
Therefore you should also delete latter logging code.

nnobelis (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Also pass the `Level` and use `log.log` to still be able to do it at `WARN` level in the other case and `INFO` in this one.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Reworded the comment.

nnobelis (dismissed):

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
sschuberth: So, what semantics does `Int.MAX_VALUE` have here? Does it mean "unlimited", or does it really mean the numeric value of `Int.MAX_VALUE`?

What I'm after is: I'm not so happy with "misusing" numeric values that are within the valid range for something special, like the meaning of "unlimited"; I would have rather said a value of `0` means "unlimited", if `0` does not make sense to be use otherwise.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -447,11 +460,27 @@ class FossId internal constructor(
...
         return scanCode
     }
 
+    /**
+     * Make sure that only the configured number of delta scans exists for the current package. Based on the list of
+     * [existingScans], delete older scans until the maximum number of delta scans is reached.
+     */
+    private suspend fun enforceDeltaScanLimit(existingScans: List<Scan>) {
+        log.info { "Number of delta scans to keep: ${config.deltaScanLimit}." }
+        if (existingScans.size >= config.deltaScanLimit) {
```
sschuberth, line 476: Why is the comparison `>=` here? If we have 10 existing scans, and we're fine to keep 10 scans, then no scan should be deleted. Or is this because the current scan is not yet included in the number of existing scans?

oheger-bosch: Exactly, the current scan is not included in the list; it needs to be counted separately.

sschuberth: Thanks for confirming. Can we probably get a code comment for this fact? And maybe change the comparison slightly to `existingScans.size + 1 > config.deltaScanLimit` to emphasize?

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
oheger-bosch: It means "unlimited". This may be a difference from a logic point of view, but in practice this can be treated as equivalent (as we will never reach this number of delta scans). Using this high value has the advantage that no conditional logic is needed when enforcing this limit.

But if you insist in changing this, I would rather use a *null* value as undefined, hoping that some of Kotlin's constructs would make the necessary distinctions more straight-forward.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Added a comment and changed the condition.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
sschuberth: While I have to admit that I don't like this mix of semantics, I do see the value of handling it this way for code simplicity. So let's keep it.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -447,11 +460,28 @@ class FossId internal constructor(
...
+     * [existingScans], delete older scans until the maximum number of delta scans is reached.
+     */
+    private suspend fun enforceDeltaScanLimit(existingScans: List<Scan>) {
+        log.info { "Number of delta scans to keep: ${config.deltaScanLimit}." }
+        // The current scan needs to be counted as well, in addition to the already existing scans.
+        if (existingScans.size + 1 > config.deltaScanLimit) {
+            log.info { "Deleting ${existingScans.size - config.deltaScanLimit + 1} older scans." }
+        }
+
+        existingScans.drop(config.deltaScanLimit - 1)
```
sschuberth, line 481: `drop()` "Returns a list containing all elements except first *n* elements", so it drops the *n* most *recent* scans. But don't we want to drop / delete the *oldest* scans that exceed the limit?

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: This is a bit tricky, maybe a comment is needed here as well. From the list the _n_ most recent scans are removed, but then the *remaining* ones are deleted. This basically means that the most recent scans are kept.

sschuberth: Ah, sorry, mindfart on my side. So yes, maybe saying some like `// Drop the scans to keep in order to iterate over the remaining ones to delete them.` would help.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -564,9 +594,10 @@ class FossId internal constructor(
     }
 
     /**
-     * Delete a scan with [scanCode].
+     * Delete a scan with [scanCode]. Log a message about this operation with the given [logLevel].
      */
-    private suspend fun deleteScan(scanCode: String) {
+    private suspend fun deleteScan(scanCode: String, logLevel: Level = Level.INFO) {
+        log.log(logLevel) { "Deleting scan $scanCode." }
```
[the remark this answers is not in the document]
oheger-bosch: Done. Now different log statements are used.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Added a comment.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -374,14 +371,28 @@ class FossId internal constructor(
             }
         }
 
+    /**
+     * Filter this list of [Scan]s for the repository defined by [url] and an optional [revision] and sorts it by
```
sschuberth, line 375: This still says "sorts" instead of "sort".

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
oheger-bosch: Sorry, missed this one. Should be fixed now.

sschuberth (approved):

PR #4372, opened 2021-08-17
reporter: Align the default output names of CycloneDX and SPDX files

sschuberth:
These days it's all about (S)BOMs, so align the names of these commonly
used BOM formats to appear next to each other when sorted
alphabetically, for a better overview.

### DIFF
--- a/reporter/src/main/kotlin/reporters/CycloneDxReporter.kt
+++ b/reporter/src/main/kotlin/reporters/CycloneDxReporter.kt
@@ -61,7 +61,7 @@ import org.ossreviewtoolkit.utils.isFalse
  */
 class CycloneDxReporter : Reporter {
     companion object {
-        const val REPORT_BASE_FILENAME = "CycloneDX-BOM"
+        const val REPORT_BASE_FILENAME = "bom.cyclonedx"
 
         const val OPTION_SINGLE_BOM = "single.bom"
         const val OPTION_OUTPUT_FILE_FORMATS = "output.file.formats"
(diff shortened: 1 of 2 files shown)


sschuberth:
This is just a proposal. Any thoughts @oss-review-toolkit/core-devs?

mnonnenmacher:
The reasoning makes sense to me, but I don't know if there are any (unwritten) conventions for filenames in any of those formats. Maybe @tsteenbe has a stronger opinion on this?

sschuberth:
> I don't know if there are any (unwritten) conventions for filenames in any of those formats.

At least for CycloneDX, the common naming [seems to be](https://github.com/CycloneDX/cyclonedx-core-java/tree/master/src/test/resources) `bom-*.xml`, to which the proposed `bom.cyclonedx.xml` is closer than the current `CycloneDX-BOM.xml`.

tsteenbe:
+1 on changing the name but I am thinking whether we should use `sbom` instead of `bom` as ORT only support generating BOM for software e.g. SBOMs

sschuberth:
> ORT only support generating BOM for software e.g. SBOMs

While currently probably mostly true, no one hinders an ORT user to generate an ORT analyzer result that's actually a hardware BOM, e.g. for (parts of) a car. This is actually something we're considering to look into. So I'm a bit hesitant to limit us to software BOMs at this point.

sschuberth:
I'm proposing to take this change as-is.

mnonnenmacher (approved):

PR #4374, opened 2021-08-18
Provide a function to remove licenses with resolved license violations

MarcelBochtler:
This provides a function to remove licenses from the list of ResolvedLicenses if a license rule violation exists for it.

Additionally to this change, it is required to remove the license from the list of packages. This will be provided in a followup PR.

This is intended to be used by the FreemarkerTemplateReporters to provide a possibility to resolve license findings locally and to remove [this limitation of resolutions mentioned in the documentation](https://github.com/oss-review-toolkit/ort/blob/master/docs/config-file-resolutions-yml.md#resolving-policy-rule-violations).

### DIFF
--- a/model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
+++ b/model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2017-2020 HERE Europe B.V.
+ * Copyright (C) 2021 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,10 +24,12 @@ import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.LicenseSource
 import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.RuleViolation
 import org.ossreviewtoolkit.model.TextLocation
 import org.ossreviewtoolkit.model.config.CopyrightGarbage
 import org.ossreviewtoolkit.model.config.LicenseFindingCuration
 import org.ossreviewtoolkit.model.config.PathExclude
+import org.ossreviewtoolkit.model.config.RuleViolationResolution
 import org.ossreviewtoolkit.spdx.SpdxExpression
 import org.ossreviewtoolkit.spdx.SpdxSingleLicenseExpression
 import org.ossreviewtoolkit.spdx.model.LicenseChoice
@@ -290,6 +293,24 @@ fun List<ResolvedLicense>.filterExcluded() =
         resolvedLicense.sources != setOf(LicenseSource.DETECTED) ||
                 resolvedLicense.locations.any { it.matchingPathExcludes.isEmpty() }
     }.map { it.filterExcludedCopyrights() }
+
+/**
+ * Filter licenses for which a resolution exists. Licenses are removed if all the conditions match:
+ *  * Any of the [resolutions] matches the [violations].
+ *  * The violation was caused by a [license][RuleViolation.license].
+ *  * The package with the [packageId] is licensed with the license from the violation.
+ */
+fun List<ResolvedLicense>.filterLicenseResolutions(
+    packageId: Identifier,
+    violations: List<RuleViolation>,
+    resolutions: List<RuleViolationResolution>
+) = filter { resolvedLicense ->
+    violations.none { ruleViolation ->
+        ruleViolation.pkg == packageId &&
+                resolvedLicense.license == ruleViolation.license &&
+                resolutions.any { it.matches(ruleViolation) }
+    }
+}
 
 private fun Collection<ResolvedCopyrightFinding>.toResolvedCopyrights(process: Boolean): List<ResolvedCopyright> {
     val allStatements = map { it.statement }
--- a/model/src/test/kotlin/licenses/ResolvedLicenseInfoTest.kt
+++ b/model/src/test/kotlin/licenses/ResolvedLicenseInfoTest.kt
@@ -20,6 +20,8 @@
 package org.ossreviewtoolkit.model.licenses
 
 import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.collections.haveSize
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
 
@@ -27,6 +29,10 @@ import io.mockk.mockk
 
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.LicenseSource
+import org.ossreviewtoolkit.model.RuleViolation
+import org.ossreviewtoolkit.model.Severity
+import org.ossreviewtoolkit.model.config.RuleViolationResolution
+import org.ossreviewtoolkit.model.config.RuleViolationResolutionReason
 import org.ossreviewtoolkit.model.licenses.TestUtils.containLicensesExactly
 import org.ossreviewtoolkit.spdx.SpdxSingleLicenseExpression
 import org.ossreviewtoolkit.spdx.model.LicenseChoice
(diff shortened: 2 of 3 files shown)


model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
```diff
@@ -291,6 +294,24 @@ fun List<ResolvedLicense>.filterExcluded() =
                 resolvedLicense.locations.any { it.matchingPathExcludes.isEmpty() }
     }.map { it.filterExcludedCopyrights() }
 
+/**
+ * Filter licenses for which a resolution exists. Licenses are removed if all the conditions match:
+ *  * Any of the [resolutions] matches the [violations].
+ *  * The violation was caused by a [license][RuleViolation.license].
+ *  * The package with the [packageId] is licensed with the license from the violation.
+ */
+fun List<ResolvedLicense>.filterLicenseResolutions(
```
oheger-bosch, line 303: Extend `Collection<ResolvedLicense>` rather than `List` to make this more generic.

model/src/test/kotlin/licenses/ResolvedLicenseInfoTest.kt
```diff
@@ -143,6 +149,177 @@ class ResolvedLicenseInfoTest : WordSpec() {
                 filteredResolvedLicenseInfo.licenses should containLicensesExactly(mit, bsd)
             }
         }
+
+        "filterLicenseResolutions()" should {
+            "remove license that is resolved by a license rule violation resolution" {
+                val resolvedLicenseInfo = ResolvedLicenseInfo(
```
oheger-bosch, line 155: The `ResolvedLicenseInfo` object is probably not needed; only the list of licenses is relevant.

model/src/test/kotlin/licenses/ResolvedLicenseInfoTest.kt
```diff
@@ -143,6 +149,177 @@ class ResolvedLicenseInfoTest : WordSpec() {
...
+                    id = Identifier.EMPTY,
+                    licenseInfo = mockk(),
+                    licenses = listOf(
+                        ResolvedLicense(
```
oheger-bosch, line 159: To remove clutter and allow focusing on essentials, I would extract the creation of a `ResolvedLicense` to a helper function that expects the license as parameter. This should shorten the test cases significantly.

model/src/test/kotlin/licenses/ResolvedLicenseInfoTest.kt
```diff
@@ -143,6 +149,177 @@ class ResolvedLicenseInfoTest : WordSpec() {
...
+                        "example message",
+                        "how to fix"
+                    )
+                )
+
+                val resolutions = listOf(
+                    RuleViolationResolution(
+                        "example message", RuleViolationResolutionReason.CANT_FIX_EXCEPTION, "example comment"
+                    ),
+                    RuleViolationResolution(
```
oheger-bosch, line 189: I think, this second resolution does not bring additional value to the test coverage, as there is a dedicated test case later to check whether only matching resolutions are taken into account.

model/src/test/kotlin/licenses/ResolvedLicenseInfoTest.kt
```diff
@@ -143,6 +149,177 @@ class ResolvedLicenseInfoTest : WordSpec() {
...
+                        "example message",
+                        "how to fix"
+                    ),
+                )
+
+                val resolutions = listOf(
+                    RuleViolationResolution(
+                        "example message", RuleViolationResolutionReason.CANT_FIX_EXCEPTION, "example comment"
+                    ),
+                    RuleViolationResolution(
```
oheger-bosch, line 250: This resolution could again be removed.

model/src/test/kotlin/licenses/ResolvedLicenseInfoTest.kt
```diff
@@ -143,6 +149,177 @@ class ResolvedLicenseInfoTest : WordSpec() {
...
+                        Identifier("Maven", "org.oss-review-toolkit", "example", "0.0.1"),
+                        SpdxSingleLicenseExpression.parse(gpl),
+                        null,
+                        Severity.ERROR,
+                        "example message",
+                        "how to fix"
+                    ),
+                    RuleViolation(
+                        "example rule",
+                        Identifier("Maven", "org.oss-review-toolkit", "example", "0.0.10"),
```
oheger-bosch, line 237: It is easy to overlook that there are two different identifiers involved, because the only difference in the coordinates is a digit in the version. Maybe use a completely different identifier for the second package to make this more explicit?

model/src/test/kotlin/licenses/ResolvedLicenseInfoTest.kt
```diff
@@ -143,6 +149,177 @@ class ResolvedLicenseInfoTest : WordSpec() {
...
+
+                val filteredLicenses = resolvedLicenses.filterLicenseResolutions(
+                    Identifier("Maven", "org.oss-review-toolkit", "example", "0.0.1"),
+                    violations,
+                    resolutions
+                )
+
+                filteredLicenses should haveSize(2)
+            }
+        }
```
oheger-bosch, line 322: You could test additionally the corner case that no licenses are filtered if the list of violations is empty.

mnonnenmacher:
@MarcelBochtler Could you explain the use case for this? In my understanding a `RuleViolationResolution` can be used to explain why it is acceptable in a certain case that a rule was violated, but it does not say that the license does not apply. For this we have concluded licenses and license finding curations already. Actually, I find the [linked docs](https://github.com/oss-review-toolkit/ort/blob/master/docs/config-file-resolutions-yml.md#resolving-policy-rule-violations) confusing as they do explain a shortcoming of resolutions, but not what the use case for resolutions is.

MarcelBochtler:
@mnonnenmacher 
The use case is:
Enable the user to write resolutions (also on a project level in the `.ort.yml`) that are related to license findings, and the possibility to reflect this in the reports.

> In my understanding a `RuleViolationResolution` can be used to explain why it is acceptable in a certain case that a rule was violated, but it does not say that the license does not apply.

That is correct. How about adding a separate `RuleViolationResolutionReason` that would explicitly state that and add it additionally to the filter logic?

> For this we have concluded licenses and license finding curations already.

The license finding curation is only for code in the project.
For findings in packages this does not apply according to the documentation.
concluded licenses cannot also not be set on a project level, but only on ORT instance level (in the `curations.yml`).
Therefore currently there is no possibility to resolve license rule violations in the project's scope. Only on an ORT instance scope.

model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
```diff
@@ -291,6 +294,24 @@ fun List<ResolvedLicense>.filterExcluded() =
                 resolvedLicense.locations.any { it.matchingPathExcludes.isEmpty() }
     }.map { it.filterExcludedCopyrights() }
 
+/**
+ * Filter licenses for which a resolution exists. Licenses are removed if all the conditions match:
```
fviernau, line 298: I find "Filter licenses for which a resolution exists" misleading, because a resolution corresponds to a rule violation, not to a license.

fviernau (changes requested):
> The license finding curation is only for code in the project.
For findings in packages this does not apply according to the documentation.

That's not true. License finding curations (as well as path excludes) can be setup also for packages, see the class `PackageConfiguration`.

I have doubts that adding this functionality is good, because I believe that the problem should be addressed using license finding curations / path excludes. That'd introduce a second way to achieve the same thing (remove a license from the effective license).
It also adds several hidden assumption, e.g. one of them being that for any given license id there can be only one rule flagging issues for it. ..and also that resolving an issue means that the corresponding license is not applicable.

@MarcelBochtler would `PackageConfiguration` work for you? e.g.

```
- path: "**"
  detected_license: "GPL-2.0-only"
  concluded_licenses: "NONE"
  reason: "INCORRECT"
```

MarcelBochtler:
@fviernau 

> > The license finding curation is only for code in the project.
> > For findings in packages this does not apply according to the documentation.
> 
> That's not true. License finding curations (as well as path excludes) can be setup also for packages, see the class `PackageConfiguration`.

I'm confused now. Please correct me where I'm wrong:
`LicenseFindingCurations` that can be added to the `.ort.yml` cannot be used for 3rd party dependencies.
[LicenseFindingCuration](https://github.com/oss-review-toolkit/ort/blob/master/model/src/main/kotlin/config/LicenseFindingCuration.kt) does not allow to specify an ID. Only a path. Therefore it can only be used for findings in the project itself.

`PackageConfiguration`s cannot be configured in `.ort.yml` and are therefore not configurable for the project, but only for the ORT instance.

I'd love to reuse a functionality to "remove" license policy violations from a config file within the project's repository (`.ort.yml`).
But I don't think that this functionality exists.

fviernau:
> `PackageConfiguration`s cannot be configured in `.ort.yml` and are therefore not configurable for the project, but only for the ORT instance.

Can you give an example use case, which requires a package configuration to apply only in the context of a single project / not globally?

MarcelBochtler:
> > `PackageConfiguration`s cannot be configured in `.ort.yml` and are therefore not configurable for the project, but only for the ORT instance.
> 
> @MarcelBochtler - Can you give an example use case, which requires a package configuration to apply only in the context of a single project / not globally?

Different responsibilities:
The team managing the project code does not have access to the ORT instance's configuration files.

fviernau:
> Different responsibilities:
> The team managing the project code does not have access to the ORT instance's configuration files.
> 

Am I right assuming that the package configurations the "team" would create would be applicable globally, but couldn't be applied globally just due to access reasons? (That'd be astonishing since the biggest redundancy in the license investigations lies within the dependency ususally, not in the project's sources)

> Edit: And using project local package configurations was not the scope of this PR. It was the ability to resolve license rule violations in the project scope.

Resolving rule violations in project scope was already possible way before this PR.

mnonnenmacher:
@MarcelBochtler @fviernau After reading that discussion my opinion is that the best solution would be to allow curations and license finding curations on project level, based on the assumption that there are cases where they cannot be made global. My main concerns regarding the proposed solution are what @fviernau mentioned above: the hidden assumptions that all licenses with resolved violations do not apply, and that it adds another approach to suppress detected licenses.

MarcelBochtler:
Closing after discussion. Will be replaced with local package curations and package configuraitons.

PR #4380, opened 2021-08-20
Vulnerability: Allow instances without references

oheger-bosch:
When using the VulnerableCode advisor implementation, results have
been encountered that actually contain only an ID, but no references.
Therefore, to represent such results in ORT's vulnerability model,
remove the check for an empty list of references from the
Vulnerability class. (It is probably safer to store such results -
even if they contain limited information - than to drop them.)

### DIFF
--- a/model/src/main/kotlin/Vulnerability.kt
+++ b/model/src/main/kotlin/Vulnerability.kt
@@ -45,13 +45,7 @@ data class Vulnerability(
      * A list with detailed information for this vulnerability obtained from different sources.
      */
     val references: List<VulnerabilityReference>
-) {
-    init {
-        require(references.isNotEmpty()) {
-            "A Vulnerability must have at least one reference."
-        }
-    }
-}
+)
 
 /**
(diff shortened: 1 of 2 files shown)


sschuberth:
Don't we also need to adjust https://github.com/oss-review-toolkit/ort/blob/e9204dc54768b07967873d7ecd3506d5a4c1a1dc/reporter/src/main/resources/templates/asciidoc/vulnerability_report.ftl#L46 then to show something meaningful?

sschuberth (approved):

oheger-bosch:
> 
> 
> Don't we also need to adjust
> 
> https://github.com/oss-review-toolkit/ort/blob/e9204dc54768b07967873d7ecd3506d5a4c1a1dc/reporter/src/main/resources/templates/asciidoc/vulnerability_report.ftl#L45-L48
> then to show something meaningful?
> 
> _Edit:_ Ah, no the whole list is then skipped, which should be fine.

Yes, just the vulnerability ID will be shown without any further details - which is exactly the information available.

PR #4381, opened 2021-08-20
Project-specific package curations

MarcelBochtler:
Enable package curations from the project's repository configuration file in order to enable project-specific curations.

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -40,6 +40,7 @@ import org.ossreviewtoolkit.analyzer.PackageManager
 import org.ossreviewtoolkit.analyzer.curation.ClearlyDefinedPackageCurationProvider
 import org.ossreviewtoolkit.analyzer.curation.FallbackPackageCurationProvider
 import org.ossreviewtoolkit.analyzer.curation.FilePackageCurationProvider
+import org.ossreviewtoolkit.analyzer.curation.SimplePackageCurationProvider
 import org.ossreviewtoolkit.analyzer.curation.Sw360PackageCurationProvider
 import org.ossreviewtoolkit.cli.GlobalOptions
 import org.ossreviewtoolkit.cli.concludeSeverityStats
@@ -170,17 +171,20 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
         val config = globalOptionsForSubcommands.config
         val analyzer = Analyzer(config.analyzer)
 
+        val repositoryConfiguration = actualRepositoryConfigurationFile?.readValueOrNull() ?: RepositoryConfiguration()
+
         val curationProvider = FallbackPackageCurationProvider(
             listOfNotNull(
-                FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir),
+                SimplePackageCurationProvider(
+                    FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations +
+                            repositoryConfiguration.curations.packages
+                ),
                 config.analyzer.sw360Configuration?.let {
                     Sw360PackageCurationProvider(it).takeIf { useSw360Curations }
                 },
                 ClearlyDefinedPackageCurationProvider().takeIf { useClearlyDefinedCurations }
             )
         )
-
-        val repositoryConfiguration = actualRepositoryConfigurationFile?.readValueOrNull() ?: RepositoryConfiguration()
 
         val ortResult = analyzer.analyze(
             inputDir, distinctPackageManagers, curationProvider, repositoryConfiguration
--- a/docs/config-file-ort-yml.md
+++ b/docs/config-file-ort-yml.md
@@ -1,6 +1,8 @@
 # The `.ort.yml` file
 
 The items below can be configured by adding an `.ort.yml` file to the root of the source code repository.
+All configurations in this file apply only to this Project's context. Usually the global context is preferred for an
+increased degree of automation and local configurations should only be done if there are good reasons.
 
 * [excludes](#excludes) - Mark [files, directories](#excluding-paths) or [package manager scopes](#excluding-scopes) as
   not included in released artifacts.
@@ -125,7 +127,7 @@ scopes defined in the examples below match the scopes in your project.
 
 ### When to Use Curations
 
-Project-specific curations should be used when you want to correct the licenses detected in the source code of the
+Curations should be used when you want to correct the licenses detected in the source code of the
 project. If you need to correct the license findings for a third-party dependency then add a curation to
 [curations.yml](config-file-curations-yml.md) or [package configuration](config-file-package-configuration-yml.md).
 
@@ -160,7 +162,7 @@ The list of available options for `reason` are defined in
 
 ### When to Use Resolutions
 
-Project-specific resolutions should be used if you are unable to solve an issue by other means.
+Resolutions should be used if you are unable to solve an issue by other means.
 
 If a resolution is not project-specific than add it to [resolutions.yml](./config-file-resolutions-yml.md) so that it is
 applied to each scan.
--- a/docs/config-file-ort-yml.md
+++ b/docs/config-file-ort-yml.md
@@ -6,7 +6,7 @@ increased degree of automation and local configurations should only be done if t
 
 * [excludes](#excludes) - Mark [files, directories](#excluding-paths) or [package manager scopes](#excluding-scopes) as
   not included in released artifacts.
-* [license finding curations](#curations) - Overwrite scan results to correct identified licenses.
+* [curations](#curations) - Overwrite package metadata, set a concluded license or correct license findings.
 * [resolutions](#resolutions) - Resolve any issues or policy rule violations.
 * [license choices](#License-Choices) - Select a license for packages which offer a license choice.
 
@@ -127,11 +127,11 @@ scopes defined in the examples below match the scopes in your project.
 
 ### When to Use Curations
 
-Curations should be used when you want to correct the licenses detected in the source code of the
-project. If you need to correct the license findings for a third-party dependency then add a curation to
-[curations.yml](config-file-curations-yml.md) or [package configuration](config-file-package-configuration-yml.md).
+License finding curations should be used when you want to correct the licenses detected in the source code of the
+project. To define path excludes on a global level for third-party packages, please use the
+[package configurations](config-file-package-configuration-yml.md).
 
-### Curating License Findings
+### Curating Project License Findings
 
 An `ort scan` result represents the detected licenses as a collection of license findings. A single `LicenseFinding` is
 represented as a tuple: `(license id, file path, start line, end line)`. Applying a `LicenseFindingCuration` changes the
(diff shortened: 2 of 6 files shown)


model/src/main/kotlin/config/Curations.kt
```diff
@@ -21,10 +21,18 @@ package org.ossreviewtoolkit.model.config
...
+
 /**
  * Curations for artifacts in a repository.
  */
 data class Curations(
+    /**
+     * Curations for third-party packages.
+     */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    val packageCurations: List<PackageCuration> = emptyList(),
```
fviernau, line 34: For consistency I believe it should be named `packages` which is a reserved word.
How about naming it `packagez` or `pkgs` and override the serialized name.

docs/config-file-ort-yml.md
```diff
@@ -4,7 +4,8 @@ The items below can be configured by adding an `.ort.yml` file to the root of th
 
 * [excludes](#excludes) - Mark [files, directories](#excluding-paths) or [package manager scopes](#excluding-scopes) as
   not included in released artifacts.
-* [license finding curations](#curations) - Overwrite scan results to correct identified licenses.
+* [curations](#curations) - Overwrite scan results to correct identified licenses or correct metadata of third-party
```
fviernau, line 7: How about stealing the first sentence from `config-file-curations-yml.md´? e.g. "Curations correct invalid or missing package metadata and set the concluded license for packages."

docs/config-file-ort-yml.md
```diff
@@ -125,11 +126,10 @@ scopes defined in the examples below match the scopes in your project.
 
 ### When to Use Curations
 
-Project-specific curations should be used when you want to correct the licenses detected in the source code of the
-project. If you need to correct the license findings for a third-party dependency then add a curation to
-[curations.yml](config-file-curations-yml.md) or [package configuration](config-file-package-configuration-yml.md).
+Project-specific license finding curations should be used when you want to correct the licenses detected in the source
```
fviernau, line 129: 'license finding curations' != curations

I believe this is out of scope of this PR.

docs/config-file-ort-yml.md
```diff
@@ -125,11 +126,10 @@ scopes defined in the examples below match the scopes in your project.
...
 ### When to Use Curations
 
-Project-specific curations should be used when you want to correct the licenses detected in the source code of the
-project. If you need to correct the license findings for a third-party dependency then add a curation to
-[curations.yml](config-file-curations-yml.md) or [package configuration](config-file-package-configuration-yml.md).
+Project-specific license finding curations should be used when you want to correct the licenses detected in the source 
+code of the project.
 
-### Curating License Findings
+### Curating Project License Findings
```
fviernau, line 132: Why add `Project`? License finding curations apply also to `Packages`.

docs/config-file-ort-yml.md
```diff
@@ -156,6 +156,25 @@ For details of the specification, see
 The list of available options for `reason` are defined in
 [LicenseFindingCurationReason.kt](../model/src/main/kotlin/config/LicenseFindingCurationReason.kt).
 
+### Curating Third-Party Metadata
```
fviernau, line 159: Curations are not limited to third-Party packages but could also curated your orgs' packages.

docs/config-file-ort-yml.md
```diff
@@ -156,6 +156,25 @@ For details of the specification, see
...
+
+Project-specific package curations can be added if you want to correct metadata of third-party dependencies.
```
fviernau, line 161: Instead of introducing multiple sections speaking about "project specific" things, how about explaining that once in the first section of this file. E.g. that all configuration in this file applies only to this Project's context. Usually global is preferred for increased degree of automation and local should only be done if there is a good reason.

fviernau (changes requested):

fviernau:
Note: the PR title says "package configurations" but it actually means "package curations"

model/src/main/kotlin/config/Curations.kt
MarcelBochtler: Only `package` is a reserved word. The plural is not. I'll change it.

docs/config-file-ort-yml.md
MarcelBochtler: > 'license finding curations' != curations

I am aware.

The removed sentence: `If you need to correct the license findings for a third-party dependency then add a curation to
[curations.yml](config-file-curations-yml.md) or [package configuration](config-file-package-configuration-yml.md).`

Is now no longer true. As third-party dependencies can now also be curated in the `.ort.yml`.
I think this should be part of this PR.

docs/config-file-ort-yml.md
MarcelBochtler: The documentation states:
> Project-specific curations should be used when you want to correct the licenses detected in the source code of the
project.

Which implicitly excludes packages.
The [used Model](https://github.com/oss-review-toolkit/ort/blob/e9204dc54768b07967873d7ecd3506d5a4c1a1dc/model/src/main/kotlin/config/LicenseFindingCuration.kt#L33-L78) doesn't define an `Identifier`.
How can the described `license_finding_curations` in this case apply to packages?

docs/config-file-ort-yml.md
MarcelBochtler: Removed `Third-Party`.

docs/config-file-ort-yml.md
MarcelBochtler: Done in a separate commit

docs/config-file-ort-yml.md
fviernau: I meant "Project-specific license finding curations" don't exist, not even in this PR.

docs/config-file-ort-yml.md
fviernau: You're right, my bad.

docs/config-file-ort-yml.md
fviernau: Thx!

docs/config-file-ort-yml.md
```diff
@@ -1,6 +1,8 @@
 # The `.ort.yml` file
 
 The items below can be configured by adding an `.ort.yml` file to the root of the source code repository.
+All configurations in this file apply only to this Project's context. Usually the global context is preferred for an
```
fviernau, line 4: Commit-msg: "...is used **for**..."

docs/config-file-ort-yml.md
```diff
@@ -6,7 +6,8 @@ increased degree of automation and local configurations should only be done if t
 
 * [excludes](#excludes) - Mark [files, directories](#excluding-paths) or [package manager scopes](#excluding-scopes) as
   not included in released artifacts.
-* [license finding curations](#curations) - Overwrite scan results to correct identified licenses.
+* [curations](#curations) - Curations correct invalid or missing package metadata and set the concluded license for
```
fviernau, line 9: The structure is changed now from

```
curations:
  license_findings:
```
to
```
curations:
  license_findings:
  packages:
```
The description now only describes package curations. I believe it should describe both.
Maybe with "Overwrite package meta-data, set a concluded license or correct license findings." or maybe separate sections for `package curations` and `license finding curations` would be more straight forward?

docs/config-file-ort-yml.md
```diff
@@ -127,11 +128,9 @@ scopes defined in the examples below match the scopes in your project.
 
 ### When to Use Curations
 
-Curations should be used when you want to correct the licenses detected in the source code of the
-project. If you need to correct the license findings for a third-party dependency then add a curation to
-[curations.yml](config-file-curations-yml.md) or [package configuration](config-file-package-configuration-yml.md).
+License finding curations should be used when you want to correct the licenses detected in the source code of the project.
```
fviernau, line 131: This section is about both, package curations and license finding curations. Is it? ..but it only mentions the latter.

fviernau (changes requested):

docs/config-file-ort-yml.md
MarcelBochtler: Yes. I'll add the part about package configurations.

docs/config-file-ort-yml.md
MarcelBochtler: > I meant "Project-specific license finding curations" don't exist, not even in this PR.

Project-specific license finding curations do exist as per the documentation: https://github.com/oss-review-toolkit/ort/blob/master/docs/config-file-ort-yml.md#curating-license-findings
Or at least a functionality is called like that in the documentation.

Are you talking about project-specific package configurations with a license finding curation?
Yes. This is out of the scope for this PR.

fviernau: > Are you talking about project-specific package configurations with a license finding curation?
Yes. This is out of the scope for this PR.

no, I'm talking about the the elements under
``
curations:
  license_findings:
``

MarcelBochtler: This was previously and still is documented:
`License finding curations should be used when you want to correct the licenses detected in the source code of the
project.`

Should I rephrase this to clarify this?

docs/config-file-ort-yml.md
```diff
@@ -127,11 +127,11 @@ scopes defined in the examples below match the scopes in your project.
 
 ### When to Use Curations
 
-Curations should be used when you want to correct the licenses detected in the source code of the
-project. If you need to correct the license findings for a third-party dependency then add a curation to
-[curations.yml](config-file-curations-yml.md) or [package configuration](config-file-package-configuration-yml.md).
+License finding curations should be used when you want to correct the licenses detected in the source code of the
+project. To define path excludes on a global level for third-party packages, please use the
+[package configurations](config-file-package-configuration-yml.md).
```
fviernau, line 132: Why is the second sentence explaining  path excludes? e.g. the previous section is about excludes.
I'm not saying it's wrong, just thinking out loud: Should this section only be above the types of curations which can be put into the ort.yml as this file is about the ort.yml?

If yes, then this section should explain (1) license finding curations (2) package curations.

I'd also be fine to leave it as is in this PR, but make a follow up PR to extract the "When to Use Curations" out of this file and move it to a more general place.  The concept of "package curations", "license finding curations" and "path excludes" is  or gets now implemented both 'project-local' and 'global'. The answer to the when to use question is same no matter if "local" or "global" so it shouldn't be in this documentation about "local" configuration.

What do you think?

fviernau (changes requested):

docs/config-file-ort-yml.md
fviernau: I believe you mixed up "package curations" with "package configurations".

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -170,18 +171,21 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
+        val repositoryConfiguration = actualRepositoryConfigurationFile?.readValueOrNull() ?: RepositoryConfiguration()
+
         val curationProvider = FallbackPackageCurationProvider(
             listOfNotNull(
                 FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir),
                 config.analyzer.sw360Configuration?.let {
                     Sw360PackageCurationProvider(it).takeIf { useSw360Curations }
                 },
+                SimplePackageCurationProvider(repositoryConfiguration.curations.packages).takeIf {
+                    repositoryConfiguration.curations.packages.isNotEmpty()
```
fviernau, line 183: This fallback provider is taking only the curations from the first provider in the list which has at least one.

I believe that the first item in the list should be a provider which return the union of local and global package curations.
Both, global and local configuration should always apply. BTW I believe that having mulitple curations per package is perfectly fine, we should not assume that the global configuration only contains one curation per package, nor should we assume the same for the local configuration and also not for their union.

fviernau (changes requested):

docs/config-file-ort-yml.md
MarcelBochtler: > I'd also be fine to merge this PR as-is, but make a follow up PR to extract the "When to Use Curations" out of this file and move it to a more general place.

As this documentation will be rewritten again as soon as _project-local_ package configurations are made possible, I'd prefer to do this in a separate PR.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
MarcelBochtler: Done.
Please also see the added description in the commit message.

Regarding the order. What should be the curation that takes precedence if it's conflicting with another one?
With the implemented change it is local > global because this is how it was discussed for a similar discussion with `license choices` with @mnonnenmacher.

fviernau: I agree with the order.

fviernau: Thanks!

fviernau (approved):

PR #4382, opened 2021-08-23
DependencyGraph: Rework test expectations to be in the dependency graph format

### LINKED ISSUE #4069: Improve handling of dependency graphs

oheger-bosch:
This PR is a final cleanup for the transition to the dependency graph format and to close #4069. 

Currently the functional tests for concrete `PackageManager` implementations store their expectations in the dependency tree format. So there is always some conversion required before results can be compared. This PR, among some other things, converts these expectations to the dependency graph format, so that tests can be simplified. Unfortunately, this causes a big diff.

### DIFF
--- a/analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
@@ -3,12 +3,12 @@ repository:
   vcs:
     type: "GitRepo"
     url: "https://github.com/oss-review-toolkit/ort-test-data-git-repo"
-    revision: "<REPLACE_REVISION>"
+    revision: "31588aa8f8555474e1c3c66a359ec99e4cd4b1fa"
     path: "manifest.xml"
   vcs_processed:
     type: "GitRepo"
     url: "https://github.com/oss-review-toolkit/ort-test-data-git-repo.git"
-    revision: "<REPLACE_REVISION>"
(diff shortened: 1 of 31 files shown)


sschuberth:
@oheger-bosch, would you mind providing an explanation for closing this (and eventually delete the remote branch)? Is #4480 supposed to be the successor?

oheger-bosch:
Closing this PR in favour of #4480. Changing the test expectations turned out to be problematic, as the exact representation of the dependency graph strongly depends on the order, in which dependencies are passed to the builder. Therefore, a strict comparison of a graph against a YAML file with expectations is prone to fail. (The resulting graphs are isomorphic, but not identical.)

PR #4386, opened 2021-08-24
Evaluator: Apply project-local package-configurations

MarcelBochtler:
Package-configurations must be unique per ID and provenance even
if one configuration is placed in the `.ort.yml` and the other one
is placed in an ORT-wide package-configurations directory.
See: https://github.com/oss-review-toolkit/ort/blob/f4c56d16b5cef2606ace63265816f607a81dba7c/model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt#L89-L100


model/src/test/kotlin/config/RepositoryConfigurationTest.kt
```diff
@@ -90,6 +90,25 @@ class RepositoryConfigurationTest : WordSpec({
                   - id: "vulnerability id"
                     reason: "INEFFECTIVE_VULNERABILITY"
                     comment: "vulnerability comment"
+                configurations:
+                  package_configurations:
```
MarcelBochtler, line 94: I'm not completely happy with the naming here, but wanted to reuse `package_configurations` as they are a known term within ORT.
Any suggestions for a better naming?

MarcelBochtler:
> Package-configurations must be unique per ID and provenance

Is this how it should work? For package curations it is different: ORT merges the curations from a single source and all curations are applied.

model/src/test/kotlin/config/RepositoryConfigurationTest.kt
fviernau: I'd personally either go for `package_configurations` as single top level entry, or
```
configurations:
  packages:
```
Tendency to the former. 
What's the reason for nesting it in `Configurations`? Consistency?

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -262,7 +263,12 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
             ortResultInput = ortResultInput.replacePackageCurations(curations)
         }
 
-        val packageConfigurationProvider = packageConfigurationOption.createProvider()
+        val repositoryPackageConfigurations = ortResultInput.repository.config.configurations.packageConfigurations
+        val optionPackageConfigurations = packageConfigurationOption.createProvider().getPackageConfigurations()
+
+        val packageConfigurationProvider = SimplePackageConfigurationProvider(
+            optionPackageConfigurations + repositoryPackageConfigurations
```
fviernau, line 270: Hint: The constructor can potentially throw an exception. Before it was possible to verify in PSV for the global configuration that an exception would not happen. Now it is not possible anymore, because the local configuration and global configuration may change independently.

I don't see a good solution for this, it's maybe just an aspect of this approach.

docs/config-file-ort-yml.md
```diff
@@ -128,8 +128,8 @@ scopes defined in the examples below match the scopes in your project.
 ### When to Use Curations
 
 License finding curations should be used when you want to correct the licenses detected in the source code of the
-project. To define path excludes on a global level for third-party packages, please use the
-[package configurations](config-file-package-configuration-yml.md).
+project. To define curations on global level for third-party packages, please use
```
fviernau, line 131: IIUC the term "repository package-curations " is new. 
I don't see need for introducing new terminology here, so I propose to not do that.

fviernau: Why does it talk about "package-**curations**"?

fviernau (changes requested):

fviernau:
> Is this how it should work? For package curations it is different: ORT merges the curations from a single source and all curations are applied.

Unique per ID and provenance: yes, that's how it's intended.s
That IMO should at least be the assumption for all the consumers of package configurations (providers).

It may make sense to enforce uniqueness within global and local configuration separately, and at the same time
merge two package configurations `a` and `b` if they correspond to the same package and one of them comes from local, the other from the global configuration. What do you think @MarcelBochtler ?

model/src/test/kotlin/config/RepositoryConfigurationTest.kt
MarcelBochtler: > What's the reason for nesting it in `Configurations`? Consistency?

Yes. That is the only reason. I'll change it to `package_configurations` as a single top level entry.

fviernau: Let's get @oss-review-toolkit/core-devs input on that first.

MarcelBochtler:
> It may make sense to enforce uniqueness within global and local configuration separately, and at the same time
> merge two package configurations `a` and `b` if they correspond to the same package and one of them comes from local, the other from the global configuration. What do you think @MarcelBochtler ?

What would be the reasoning to disallow merging of package-configurations from one source?
If merging from different sources is possible, it would also be possible for the same sources.

@fviernau Do you think that the decision / implementation of merging of package-configurations is blocking this PR?

model/src/test/kotlin/config/RepositoryConfigurationTest.kt
MarcelBochtler: Already changed it, as I didn't see your answer. However: I'd be fine with changing it back after some input from the other @oss-review-toolkit/core-devs.

mnonnenmacher: I also prefer a single top level entry `package_configurations`. Similarly `curation` could be added as another top level entry later.

fviernau: > Similarly `curation` could be added as another top level entry later.

I didn't get this bit. Do you mean refactoring as follows ?

```
curations:
  license_findings:
  packages:
```
to
```
package_curations:
license_finding_curations:
```

fviernau:
> What would be the reasoning to disallow merging of package-configurations from one source?

I think that's clearly bad practice as I cannot see any advantage of setting up multiple package configs for a single
tuple: package / scan result. It just adds complexity.

> If merging from different sources is possible, it would also be possible for the same sources.
> 
> @fviernau Do you think that the decision / implementation of merging package-configurations is blocking this PR?

Possible yes. As written above I believe it should not be implemented.  I'm find with the PR as-is: without the merging implemented.

fviernau (approved):

model/src/test/kotlin/config/RepositoryConfigurationTest.kt
MarcelBochtler: > I also prefer a single top level entry package_configurations. Similarly curation could be added as another top level entry later.

Curations were already added in this PR: https://github.com/oss-review-toolkit/ort/pull/4381
in this form:
```yml
curations:
  packages:
  - id: "Maven:id..."
    curations:
    ....
```

PR #4387, opened 2021-08-24
Miscellaneous improvements to FossID logging and issue handling

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -308,7 +308,7 @@ class FossId internal constructor(
                     }
                 } catch (e: IllegalStateException) {
                     e.showStackTrace()
-                    log.error("Package at url=$url cannot be scanned")
+                    log.error { "Package at URL $url cannot be scanned." }
 
                     val summary = createSingleIssueSummary(
                         pkg.id.toCoordinates(),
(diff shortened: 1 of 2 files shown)


sschuberth:
FYI @hanna-modica.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -261,8 +262,20 @@ class FossId internal constructor(
             filteredPackages.forEach { pkg ->
                 val startTime = Instant.now()
 
-                // TODO: Continue the processing of other packages and add an issue to the scan result.
-                require(pkg.vcsProcessed.type == VcsType.GIT) { "FossID only supports Git repositories." }
+                if (pkg.vcsProcessed.type != VcsType.GIT) {
```
oheger-bosch, line 265: +1 for the implementation. It would be great if a test case could be added for this scenario.

sschuberth: I've added a test. Please double-check that I'm not doing something stupid / superfluous there.

oheger-bosch: Thanks for adding the test. It looks already good. You could even improve it by passing a second package to the scanner and verify that it gets processed after the error with the unsupported VCS is handled. But I leave this to you.

oheger-bosch (approved):

sschuberth:
> You could even improve it by passing a second package to the scanner and verify that it gets processed after the error with the unsupported VCS is handled.

I'll do this in a follow-up.

PR #4389, opened 2021-08-25
Consistently spell it like "metadata" throughout the code base

sschuberth:
The "metadata" spelling seems to be way more common than "meta data" or
"meta-data".

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ It does so by orchestrating a _highly customizable_ pipeline of tools that _abst
 These tools are implemented as libraries (for programmatic use) and exposed via a command line interface (for scripted
 use):
 
-* [_Analyzer_](#analyzer) - determines the dependencies of projects and their meta-data, abstracting which package
+* [_Analyzer_](#analyzer) - determines the dependencies of projects and their metadata, abstracting which package
   managers or build systems are actually being used.
 * [_Downloader_](#downloader) - fetches all source code of the projects and their dependencies, abstracting which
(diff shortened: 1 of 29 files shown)


sschuberth:
Note that for yet more consistency this would require to also rename classes, functions and properties like `isMetaDataOnly`, but I'm not doing that here as it would break existing curations.

mnonnenmacher:
> Note that for yet more consistency this would require to also rename classes, functions and properties like `isMetaDataOnly`, but I'm not doing that here as it would break existing curations.

You could use `@JsonAlias` to make it backward compatible.

sschuberth:
> You could use `@JsonAlias` to make it backward compatible.

Right, but that would not work e.g. for rules using the `isMetaDataOnly` rule matcher. While I *really* like consistency, I'm not sure it's worth a breaking change here, and I also do not want to introduce a new `isMetadataOnly` rule matcher and deprecated the old one.

More opinions?

fviernau (dismissed):

fviernau:
> More opinions?

I'm fine with leaving `isMetaDataOnly` as-is.
I've found one more occurence of `meta-data` in `getting-started.md`. 
That could be fixed as well.

sschuberth:
> I've found one more occurence of `meta-data` in `getting-started.md`.
> That could be fixed as well.

Done, I've found even some more occurrences in non-code places.

fviernau (approved):

sschuberth:
Merging despite the unrelated failure.

PR #4390, opened 2021-08-25
Conan: Forward Java authenticator credentials

sschuberth:
Conan does not support parsing .netrc for credentials to authenticate
against remotes, see [1]. So forward any credentials for remote hosts
provided by the Java authenticator, which in this case is an instance of
OrtAuthenticator that supports getting credentails from a .netrc file.

[1] https://github.com/conan-io/conan/issues/3046#issuecomment-904516572

### DIFF
--- a/analyzer/src/main/kotlin/managers/Conan.kt
+++ b/analyzer/src/main/kotlin/managers/Conan.kt
@@ -26,6 +26,7 @@ import com.fasterxml.jackson.databind.JsonNode
 import com.vdurmont.semver4j.Requirement
 
 import java.io.File
+import java.net.Authenticator
 import java.util.SortedSet
 import java.util.Stack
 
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -116,6 +118,46 @@ class Conan(
...
+                                /* protocol = */ uri.scheme,
+                                /* prompt = */ null,
+                                /* scheme = */ null
+                            )
+
+                            if (auth != null) {
+                                val userAuth = run("user", "-r", remoteName, "-p", String(auth.password), auth.userName)
+                                if (userAuth.isError) {
+                                    log.warn { "Failed to configure user authentication for remote '$remoteName'." }
+                                }
```
MarcelBochtler, line 152: If the authentication fails, shouldn't this log an error?
As the followup analyser run would most likely fail as well?

MarcelBochtler (dismissed):

analyzer/src/main/kotlin/managers/Conan.kt
sschuberth: Agreed, done.

MarcelBochtler (approved):

PR #4406, opened 2021-08-27
More fixes for filtering empty paths in FossID scans

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -223,24 +223,25 @@ class FossId internal constructor(
                 else "Package authors filter is '${config.packageAuthorsFilter}'."
             }
 
+            fun addPackageWithSingleIssue(pkg: Package, issue: OrtIssue, provenance: Provenance) {
+                val time = Instant.now()
+                val summary = ScanSummary(time, time, "", sortedSetOf(), sortedSetOf(), listOf(issue))
+                val scanResult = ScanResult(provenance, details, summary)
+                results.getOrPut(pkg) { mutableListOf() } += scanResult
+            }
+
             val filteredPackages = packages
                 .filter { config.packageNamespaceFilter.isEmpty() || it.id.namespace == config.packageNamespaceFilter }
                 .filter { config.packageAuthorsFilter.isEmpty() || config.packageAuthorsFilter in it.authors }
                 .partition { it.vcsProcessed.type == VcsType.GIT }
                 .let { (packagesInsideGit, packagesOutsideGit) ->
                     packagesOutsideGit.forEach {
-                        val startTime = Instant.now()
-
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -135,11 +135,11 @@ class FossId internal constructor(
          */
         private fun queryAuthenticator(repoUrl: String): String {
             val repoUri = repoUrl.toUri().getOrElse {
-                log.warn { "Host cannot be extracted for $repoUrl." }
+                log.warn { "The repository '$repoUrl' in not a valid URI." }
```
fviernau, line 138: 'in not' -> 'is not'

Consider to reduce it to "The repository '$repoUrl' is not valid.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -278,7 +278,7 @@ class FossId internal constructor(
                 }
 
             if (filteredPackages.isEmpty()) {
-                log.warn { "There is no package to scan !" }
+                log.warn { "There is no package to scan." }
```
fviernau, line 281: Consider making this `.info`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -404,13 +409,13 @@ class FossId internal constructor(
...
         val scanCode = if (existingScan == null) {
-            log.info { "No scan found for $url and revision $revision. Creating scan ..." }
+            log.info { "No scan found for $url and revision $revision. Creating scan..." }
 
             val scanCode = namingProvider.createScanCode(projectName)
             val newUrl = if (config.addAuthenticationToUrl) queryAuthenticator(url) else url
             createScan(projectCode, scanCode, newUrl, revision)
 
-            log.info { "Initiating data download ..." }
+            log.info { "Initiating data download..." }
```
fviernau, line 418: out of scope, maybe: "Initiating the download..."

fviernau (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth: I went for yet another wording 😀

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth: I prefer to keep this a warning as this case should be rather unusual (there should be at least the project package).

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth: Done.

fviernau (approved):

PR #4408, opened 2021-08-30
Cargo: Add a code comment about Rust / Cargo package (non-)namespaces

sschuberth:


fviernau (approved):

sschuberth:
> Should it be "does" instead of "do"?

I'm not sure either 😀 I guess I meant to writes "does", but when rereading now I actually find "do" better to emphasize that this was not (only) a design decision by Cargo (and another package manager for Rust could implement it otherwise), but also by Rust as a language itself.

sschuberth:
@fviernau I've fixed *PipFunTest* (again! 😠) as part of this PR.

fviernau (approved):

sschuberth:
Sorry, this time I forgot the signoff... 😞

fviernau (approved):

PR #4409, opened 2021-08-30
NoticeTemplateReporter: Fix when a section header is printed

sschuberth:
Previously, when all license text had no content, still the section
header was printed in front of an empty list. Fix this by printing the
header on the first non-empty item instead.


reporter/src/main/resources/templates/notice/default.ftl
```diff
@@ -25,11 +25,6 @@ The notice file generated by this template consists of the following sections:
 Excluded projects and packages are ignored.
```
MarcelBochtler, line 25: Commit message:
`when all license text had no content`. Is this correct english?

I'd change it to `when none of the license texts had any content`.

sschuberth: > Is this correct english?

I believe it is, it's just inverse logic compared to yours 😁 Changed anyway.

fviernau (approved):

MarcelBochtler (approved):

PR #4410, opened 2021-08-30
Dockerfile: Do not hard-code $HOME to /root

sschuberth:
This is to allow running as a non-root user.


fviernau (approved):

sschuberth:
This does not work as expected as inside `--mount` environment variables like `$HOME` are not resolved.

sschuberth:
> This does not work as expected as inside `--mount` environment variables like `$HOME` are not resolved.

I've found a solution to this in https://github.com/oss-review-toolkit/ort/pull/4413.

PR #4415, opened 2021-08-31
[WIP]: Why are the tree traversal methods not equal?

maxhbr:
This was created as a result of a private discussion with @sschuberth .

This code implements a TraverseDepsFunTest.kt , that tests two different implementations for traversal of the dependency trees of all projects of a ortResult, but for the real world example

* they are not equal and
* neither catches all the dependencies 

At some point this code might be worth to be moved to a better location, to be added as an actual test.


maxhbr:
Type 1 tree traversal for example misses the following two packages:

maxhbr:
was resolved

PR #4416, opened 2021-08-31
DependencyGraphBuilder: Fix a potential exception when building the graph

fviernau:
When looking-up the node index of the "to node" of an edge, the look-up may fail since it happens that the map entry hasn't been created. Fix that issue and also eliminate some unclear things related to not using a proper unique key (package & fragment index) for the dependency references.


fviernau:
The bug this PR was targeting was fixed by #4422 instead.
Keep the branch around for a while as it provides indicators where assumptions about 'no cycles' are made in the code.

sschuberth:
> Keep the branch around for a while as it provides indicators where assumptions about 'no cycles' are made in the code.

Ping @oheger-bosch once you're back.

PR #4417, opened 2021-08-31
FossId: Explicitly check if a scan was added to scan queue

MarcelBochtler:
As FossID interprets queued scans as error, check for this and wait for
the completion without checking for further errors.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -537,8 +537,14 @@ class FossId internal constructor(
         if (response.data?.status in SCAN_STATE_FOR_TRIGGER) {
             log.info { "Triggering scan as it has not yet been started." }
 
-            service.runScan(config.user, config.apiKey, scanCode, *runOptions)
-                .checkResponse("trigger scan", false)
+            val scanResult = service.runScan(config.user, config.apiKey, scanCode, *runOptions)
+
+            // Scans that were added to the queue are interpreted as an error by FossID before version 2021.2.
+            // For older versions, `waitScanComplete()` is able to deal with queued scans. Therefore, not checking the
+            // response of queued scans.
+            if (version >= "2021.2" || scanResult.error != "Scan was added to queue") {
+                scanResult.checkResponse("trigger scan", false)
+            }
 
             waitScanComplete(scanCode)
         }


nnobelis (approved):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -537,8 +537,12 @@ class FossId internal constructor(
         if (response.data?.status in SCAN_STATE_FOR_TRIGGER) {
             log.info { "Triggering scan as it has not yet been started." }
 
-            service.runScan(config.user, config.apiKey, scanCode, *runOptions)
-                .checkResponse("trigger scan", false)
+            val scanResult = service.runScan(config.user, config.apiKey, scanCode, *runOptions)
+            // Scans that were added to the queue are interpreted as an error by FossID. However, `waitScanComplete()`
```
sschuberth, line 541: Nit: Blank line before this one to better group the comment to the code below than the code above?

nnobelis (approved):

nnobelis (changes requested):
I managed to reproduce the bug, and it seems this bug occurs only with FossID with version < 21.2. Could you mention it in a comment ?

sschuberth:
> I managed to reproduce the bug, and it seems this bug occurs only with FossID with version < 21.2. Could you mention it in a comment ?

Or maybe even only apply the work-around for those versions (as we can query the version by now)? Then the code would be the comment 😉

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -537,8 +537,14 @@ class FossId internal constructor(
...
             log.info { "Triggering scan as it has not yet been started." }
 
-            service.runScan(config.user, config.apiKey, scanCode, *runOptions)
-                .checkResponse("trigger scan", false)
+            val scanResult = service.runScan(config.user, config.apiKey, scanCode, *runOptions)
+
+            // Scans that were added to the queue are interpreted as an error by FossID before version 2021.1.
+            // For older versions, `waitScanComplete()` is able to deal with queued scans. Therefore, not checking the
+            // response of queued scans.
+            if (version >= "2021.1" || scanResult.error != "Scan was added to queue") {
```
sschuberth, line 545: Does `>=` on a string really do what we want here? Does it work because it's CalVer? Just want to double-check 😀 

Anyway, this probably also need to say `"2021.2"` (see above).

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
MarcelBochtler: Yes it does what we want, because it's CalVer.
This is also how it is done in https://github.com/oss-review-toolkit/ort/blob/cf5efc563cc5e4ede8e4c79d351d8d1d00621776/clients/fossid-webapp/src/main/kotlin/FossIdServiceWithVersion.kt#L36

Thanks for the hint with the wrong version.

sschuberth (approved):

sschuberth:
Merging despite the unrelated `reporter-web-app` failure.

Issue #2819, opened 2020-07-03
Allow to specify different scanners depending on package / project properties


sschuberth:
This is a generalization of the idea in https://github.com/oss-review-toolkit/ort/issues/2130. Instead of just distinguishing between packages and projects, we want to be able to use different scanners based on package / project properties. For example in a C/C++ project, if we know that a package is a third-party Open Source component *with modifications*, we might want to use a snippet scanner whereas *without* modifications a "normal" license scanner might suffice.

This ideally requires an extension of the package data model, with something like an "origin" (I'm open for a better name). "origin" could be a `enum class` with values like `COMPANY_INTERNAL`, `THIRD_PARTY_OPEN_SOURCE`, `THIRD_PARTY_COMMERCIAL`, `THIRD_PARTY_OPEN_SOURCE_WITH_MODIFICATIONS`, `THIRD_PARTY_COMMERCIAL_WITH_MODIFICATIONS`. As possibly more permutations might be needed, maybe it would make sense to instead come of with a "labeling system" where a package could have multiple labels, like `THIRD_PARTY` *and* `OPEN_SOURCE` *and* `WITH_MODIFICATIONS`.

nnobelis:
How a package label would be populated ?

Does the user have to manually edit analyzer_result.yml and add the labels for each package ?

sschuberth:
Ideally the analyzer would implement some heuristics to label packages, based on metadata the respective package manager provides. Or labels could be added via curations.

neubs-bsi:
Note from planning 02.11.2020:
To be cleared: Which properties should be diefined (potentially ideal: configurable)

KorSin:
Opened PR #3329 to solve this issue.

KorSin:
I actually prefer the suggestion by @sschuberth to have, e.g., a `Set<PackageClassification>`, where you decide for the scanner by the combinations inside the set. It should not be too difficult to implement an additional field for the `Package` class which can be curated. 

Then, the configuration could be expanded to configure which scanner to use for which combinations. The `ScannerCommand` should then be responsible to filter the packages accordingly.

However, multiple scanners are not implemented yet. Maybe, once #3638 is implemented, one can use these implementation details.

Looking at the structure of my thoughts, it would make sense to divide this issue into three PRs.

sschuberth:
A concrete use-case is: Use a snippet-scanner (as opposed to a regular license scanner) if a package has the `isModified` property set.

[...]

PR #4418, opened 2021-08-31
Allow to use a different scanner for projects than for packages

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -135,7 +135,7 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license 
                     it.isFile && it.name == transformedScanResultsFileName
                 }.count()
 
-                println("Local file storage has $fileCount scan results files.")
+                println("Local file storage has $fileCount scan results file(s).")
             }
         }
 
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -45,8 +45,6 @@ import org.ossreviewtoolkit.cli.utils.outputGroup
 import org.ossreviewtoolkit.cli.utils.readOrtResult
 import org.ossreviewtoolkit.cli.utils.writeOrtResult
 import org.ossreviewtoolkit.model.FileFormat
-import org.ossreviewtoolkit.model.config.DownloaderConfiguration
-import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.utils.mergeLabels
 import org.ossreviewtoolkit.scanner.LocalScanner
 import org.ossreviewtoolkit.scanner.ScanResultsStorage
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/Scanner.kt
```diff
@@ -52,6 +53,19 @@ fun scanOrtResult(
     ortResult: OrtResult,
     outputDirectory: File,
     skipExcluded: Boolean = false
+) = scanOrtResult(scanner, scanner, ortResult, outputDirectory, skipExcluded)
+
+/**
+ * Use the [scanner] and [projectScanner] to scan the [Project]s and [Package]s specified in the [ortResult],
+ * respectively. Scan results are stored in the [outputDirectory].  If [skipExcluded] is true, packages for which
```
MarcelBochtler, line 60: Nit: Double space before `If`.

cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -109,6 +110,12 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license
         help = "The scanner to use, one of ${Scanner.ALL}."
     ).convertToScanner().default(ScanCode.Factory())
 
+    private val projectScannerFactory by option(
```
fviernau, line 113: The commit message seems to assume that dependencies are always third-party dependencies.
Have you considered dependencies which come from the own ORG?
The rationale provided in the commit message suggests to me that a more sophisticated scanner selection mechanism might be needed to enable the use case. What do you think?

sschuberth: > Have you considered dependencies which come from the own ORG?

Yes, of course.

> The rationale provided in the commit message suggests to me that a more sophisticated scanner selection mechanism might be needed to enable the use case.

Indeed. This PR is only supposed to be a step into the right direction of a more general scanner selection mechanism as described in https://github.com/oss-review-toolkit/ort/issues/2819.

fviernau: Thanks!

MarcelBochtler (approved):

MarcelBochtler (approved):

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -86,7 +108,20 @@ fun scanOrtResult(
         }
 
     val scanResults = runBlocking {
-        scanner.scanPackages(packagesToScan, outputDirectory).mapKeys { it.key.id }
+        // Scan the projects from the ORT result.
+        val deferredProjectScan = async {
+            projectScanner.scanPackages(filteredProjectPackages, outputDirectory).mapKeys { it.key.id }
```
nnobelis, line 113: I think we need to consider some exception catching here: maybe it would be worthwhile to catch exceptions thrown by one `scanPackages`, to be able to collect the results of the other ?

I imagine the case where someone run Scancode + FossID on a big project and e.g. FossID has an exception, it would be cool to nevertheless have the Scancode results.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -86,7 +108,20 @@ fun scanOrtResult(
...
+
+        // Scan the packages from the ORT result.
+        val deferredPackageScan = async {
+            scanner.scanPackages(filteredPackages, outputDirectory).mapKeys { it.key.id }
+        }
+
+        val projectResults = deferredProjectScan.await()
+        val packageResults = deferredPackageScan.await()
+
+        projectResults + packageResults
```
nnobelis, line 124: Here you are mixing the two results together ?! Didn't you say you were going to separate them in a different `ScannerRun` ?

scanner/src/main/kotlin/Scanner.kt
sschuberth: I agree that would be cool. However, I'll leave that as a future exercise, as [exception handling with Kotlin coroutines is hard](https://www.lukaslechner.com/why-exception-handling-with-kotlin-coroutines-is-so-hard-and-how-to-successfully-master-it/).

scanner/src/main/kotlin/Scanner.kt
sschuberth: I don't remember whether I said "ScannerRun", but if so, that was wrong. We get separate `ScanResult`s, not `ScannerRun`s. In any case, the results are completely separate in the end, see this example result:

image

PR #4419, opened 2021-09-01
Dockerfile: Set `GRADLE_USER_HOME` instead of using a symlink to the default location

MarcelBochtler:
In some environments using a symlink to `$HOME/.gradle` won't work.
Instead, change the location by setting `GRADLE_HOME`.

For our Azure environment gradle is not able to create a lock file when symlinking the `.gradle` directory:
```
#15 [build 5/5] RUN --mount=type=cache,target=/tmp/.gradle/     ln -s /tmp/....
#15 1.381 Skipping the import of certificates as no HTTPS proxy is set.
#15 1.790 Exception in thread "main" java.lang.RuntimeException: Could not create parent directory for lock file /root/.gradle/wrapper/dists/gradle-7.2-bin/2dnblmf4td7x66yl1d74lt32g/gradle-7.2-bin.zip.lck
#15 1.791   at org.gradle.wrapper.ExclusiveFileAccessManager.access(ExclusiveFileAccessManager.java:43)
#15 1.792   at org.gradle.wrapper.Install.createDist(Install.java:48)
#15 1.792   at org.gradle.wrapper.WrapperExecutor.execute(WrapperExecutor.java:107)
#15 1.792   at org.gradle.wrapper.GradleWrapperMain.main(GradleWrapperMain.java:63)
#15 ERROR: executor failed running [/bin/sh -c ln -s /tmp/.gradle/ $HOME/.gradle &&     scripts/import_proxy_certs.sh &&     scripts/set_gradle_proxy.sh &&     sed -i -r 's,(^distributionUrl=)(.+)-all\.zip$,\1\2-bin.zip,' gradle/wrapper/gradle-wrapper.properties &&     ./gradlew --no-daemon --stacktrace -Pversion=$ORT_VERSION :cli:distTar :helper-cli:startScripts]: runc did not terminate sucessfully
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ WORKDIR /usr/local/src/ort
 
 # Gradle build.
 RUN --mount=type=cache,target=/tmp/.gradle/ \
-    ln -s /tmp/.gradle/ $HOME/.gradle && \
+    GRADLE_USER_HOME=/tmp/.gradle/ && \
     scripts/import_proxy_certs.sh && \
     scripts/set_gradle_proxy.sh && \
     sed -i -r 's,(^distributionUrl=)(.+)-all\.zip$,\1\2-bin.zip,' gradle/wrapper/gradle-wrapper.properties && \


Dockerfile
```diff
@@ -36,8 +36,8 @@ COPY . /usr/local/src/ort
 WORKDIR /usr/local/src/ort
 
 # Gradle build.
+ENV GRADLE_HOME=/tmp
```
sschuberth, line 39: This should be `GRADLE_USER_HOME`, not `GRADLE_HOME`, and it needs to contain the `.gradle` part.

Also, can we just do `GRADLE_USER_HOME=/tmp/.gradle` below (in original line 40) to avoid the creation of an intermediate layer due to `ENV`?

sschuberth (changes requested):

Dockerfile
```diff
@@ -37,7 +37,7 @@ WORKDIR /usr/local/src/ort
 
 # Gradle build.
 RUN --mount=type=cache,target=/tmp/.gradle/ \
-    ln -s /tmp/.gradle/ $HOME/.gradle && \
+    GRADLE_USER_HOME=/tmp/.gradle && \
```
sschuberth, line 40: Mini-nip (I did it wrong myself in the example): Should we end the path with a `/` to highlight it's a directory, and to resemble exactly the path from one line above? Approving anyway.

sschuberth (dismissed):

sschuberth (approved):

sschuberth:
Merging this already now as the relevant Docker build passed.

PR #4422, opened 2021-09-01
DependencyGraph: Fix a crash and a couple of subsequent issues

### LINKED ISSUE #4405: Analyzer crashes when 'Gradle' tries to build the dependency graph
Observed on 25th of aug.

```
Exception in thread "main" java.util.NoSuchElementException: Key org.ossreviewtoolkit.model.DependencyReference@26f77e00 is missing in the map.
	at kotlin.collections.MapsKt__MapsKt.getValue(Maps.kt:344)
	at org.ossreviewtoolkit.model.utils.DependencyGraphBuilderKt.toGraph$constructGraph(DependencyGraphBuilder.kt:385)
	at org.ossreviewtoolkit.model.utils.DependencyGraphBuilderKt.toGraph$constructGraph(DependencyGraphBuilder.kt:382)
	at org.ossreviewtoolkit.model.utils.DependencyGraphBuilderKt.toGraph(DependencyGraphBuilder.kt:391)
	at org.ossreviewtoolkit.model.utils.DependencyGraphBuilderKt.access$toGraph(DependencyGraphBuilder.kt:1)
	at org.ossreviewtoolkit.model.utils.DependencyGraphBuilder.build(DependencyGraphBuilder.kt:169)
	at org.ossreviewtoolkit.model.utils.DependencyGraphBuilder.build$default(DependencyGraphBuilder.kt:163)
	at org.ossreviewtoolkit.analyzer.managers.Gradle.createPackageManagerResult(Gradle.kt:125)
	at org.ossreviewtoolkit.analyzer.PackageManager.resolveDependencies(PackageManager.kt:273)
	at org.ossreviewtoolkit.analyzer.Analyzer$analyzeInParallel$1$1$1.invokeSuspend(Analyzer.kt:131)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

fviernau:
When `DependencyGraphBuilder.toGraph()` looks-up the node index of the "to node" of an edge, the look-up may fail since it happens that the map entry hasn't been created before. Fix that issue (crash) by making sure that the map entry is properly created.

The false logic which exhibited that bug also skipped edges which may mean actually eliminating circles. 
In fact when fixing the logic, ORT runs into a subsequent issue because several consumers of the graph assume the absence of any cycle. 

While in the long run it makes sense to allow cycle in the graph, implement a temporary solution to just break all cycle. 
Allowing circles is a bigger topic and needs proper planning and involvement of all relevant people. The goal of this PR is to fix the crash in the short term.

Fixes #4405.

### DIFF
--- a/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
+++ b/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
@@ -370,19 +370,19 @@ class DependencyGraphBuilder<D>(
 private fun Collection<DependencyReference>.toGraph(): Pair<List<DependencyGraphNode>, List<DependencyGraphEdge>> {
     val nodes = mutableSetOf<DependencyGraphNode>()
     val edges = mutableListOf<DependencyGraphEdge>()
-    val nodeIndexMapping = mutableMapOf<DependencyReference, Int>()
+    val nodeIndexMapping = mutableMapOf<NodeKey, Int>()
 
     fun constructGraph(dependencies: Collection<DependencyReference>) {
         dependencies.forEach { ref ->
             val node = DependencyGraphNode(ref.pkg, ref.fragment, ref.linkage, ref.issues)
             if (node !in nodes) {
-                val fromIndex = nodes.size.also { nodeIndexMapping[ref] = it }
+                val fromIndex = nodes.size.also { nodeIndexMapping[ref.key] = it }
                 nodes += node
 
                 constructGraph(ref.dependencies)
 
                 ref.dependencies.forEach { dep ->
-                    edges += DependencyGraphEdge(fromIndex, nodeIndexMapping.getValue(dep))
+                    edges += DependencyGraphEdge(fromIndex, nodeIndexMapping.getValue(dep.key))
                 }
             }
         }
@@ -391,3 +391,11 @@ private fun Collection<DependencyReference>.toGraph(): Pair<List<DependencyGraph
     constructGraph(this)
     return nodes.toList() to edges
 }
+
+private data class NodeKey(
+    val pkg: Int,
+    val fragment: Int
+)
+
+private val DependencyReference.key: NodeKey
+    get() = NodeKey(pkg, fragment)
--- a/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
+++ b/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
@@ -370,19 +370,19 @@ class DependencyGraphBuilder<D>(
 private fun Collection<DependencyReference>.toGraph(): Pair<List<DependencyGraphNode>, List<DependencyGraphEdge>> {
     val nodes = mutableSetOf<DependencyGraphNode>()
     val edges = mutableListOf<DependencyGraphEdge>()
-    val nodeIndexMapping = mutableMapOf<NodeKey, Int>()
+    val nodeIndices = mutableMapOf<NodeKey, Int>()
 
     fun constructGraph(dependencies: Collection<DependencyReference>) {
         dependencies.forEach { ref ->
             val node = DependencyGraphNode(ref.pkg, ref.fragment, ref.linkage, ref.issues)
             if (node !in nodes) {
-                val fromIndex = nodes.size.also { nodeIndexMapping[ref.key] = it }
+                val fromIndex = nodes.size.also { nodeIndices[ref.key] = it }
                 nodes += node
 
                 constructGraph(ref.dependencies)
 
                 ref.dependencies.forEach { dep ->
-                    edges += DependencyGraphEdge(fromIndex, nodeIndexMapping.getValue(dep.key))
+                    edges += DependencyGraphEdge(fromIndex, nodeIndices.getValue(dep.key))
                 }
             }
         }
--- a/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
+++ b/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
@@ -387,15 +387,18 @@ private fun Collection<DependencyReference>.toGraph(): Pair<List<DependencyGraph
     val edges = mutableListOf<DependencyGraphEdge>()
     val nodeIndices = mutableMapOf<NodeKey, Int>()
 
-    visitEach { fromRef ->
-        val node = DependencyGraphNode(fromRef.pkg, fromRef.fragment, fromRef.linkage, fromRef.issues)
-        if (node !in nodes) {
-            val fromIndex = nodes.size.also { nodeIndices[fromRef.key] = it }
-            nodes += node
+    fun getOrAddNodeIndex(ref: DependencyReference): Int =
+        nodeIndices.getOrPut(ref.key) {
+            nodes += DependencyGraphNode(ref.pkg, ref.fragment, ref.linkage, ref.issues)
+            nodes.size - 1
+        }
 
-            fromRef.dependencies.forEach { toRef ->
-                edges += DependencyGraphEdge(fromIndex, nodeIndices.getValue(toRef.key))
-            }
+    visitEach { fromRef ->
+        val fromNodeIndex = getOrAddNodeIndex(fromRef)
+
+        fromRef.dependencies.forEach { toRef ->
+            val toNodeIndex = getOrAddNodeIndex(toRef)
+            edges += DependencyGraphEdge(fromNodeIndex, toNodeIndex)
         }
     }
 
--- a/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
+++ b/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
@@ -387,14 +387,14 @@ private fun Collection<DependencyReference>.toGraph(): Pair<List<DependencyGraph
     val edges = mutableListOf<DependencyGraphEdge>()
     val nodeIndices = mutableMapOf<NodeKey, Int>()
 
-    visitEach { ref ->
-        val node = DependencyGraphNode(ref.pkg, ref.fragment, ref.linkage, ref.issues)
+    visitEach { fromRef ->
+        val node = DependencyGraphNode(fromRef.pkg, fromRef.fragment, fromRef.linkage, fromRef.issues)
         if (node !in nodes) {
-            val fromIndex = nodes.size.also { nodeIndices[ref.key] = it }
+            val fromIndex = nodes.size.also { nodeIndices[fromRef.key] = it }
             nodes += node
 
-            ref.dependencies.forEach { dep ->
-                edges += DependencyGraphEdge(fromIndex, nodeIndices.getValue(dep))
+            fromRef.dependencies.forEach { toRef ->
+                edges += DependencyGraphEdge(fromIndex, nodeIndices.getValue(toRef.key))
             }
         }
     }
--- a/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
+++ b/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
@@ -161,14 +161,18 @@ class DependencyGraphBuilder<D>(
      * the conditions do not hold then.
      */
     fun build(checkReferences: Boolean = true): DependencyGraph {
-        require(!checkReferences || resolvedPackages.keys.containsAll(validPackageDependencies)) {
-            "The following references do not actually refer to packages: " +
-                    "${validPackageDependencies - resolvedPackages.keys}."
-        }
+        if (checkReferences) checkReferences()
 
         val (nodes, edges) = directDependencies.toGraph()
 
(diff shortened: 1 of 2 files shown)


MarcelBochtler (changes requested):

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -161,14 +164,39 @@ class DependencyGraphBuilder<D>(
...
+        return DependencyGraph(dependencyIds, sortedSetOf(), scopeMapping, nodes, edges.removeAllCycles())
+    }
+
+    private fun Collection<DependencyGraphEdge>.removeAllCycles(): List<DependencyGraphEdge> {
+        val edges = mapTo(mutableSetOf()) { it.from to it.to }
+        val edgesToKeep = breakAllCycles(edges)
+        val edgesToRemove = edges - edgesToKeep
+
+        edgesToRemove.forEach {
+            log.warn { "Removing edge '${it.first} -> ${it.second}' to break a cycle." }
```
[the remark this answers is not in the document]
fviernau: Thanks!

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -173,6 +173,15 @@ class DependencyGraphBuilder<D>(
             "The following references do not actually refer to packages: " +
                     "${validPackageDependencies - resolvedPackages.keys}."
         }
+
+        val packageReferencesKeysWithMultipleDistintPackageReferences = directDependencies.groupBy { it.key }.filter {
```
MarcelBochtler, line 192: `packageReferencesKeysWithMultipleDistintPackageReferences`: Missing `c`:
`packageReferencesKeysWithMultipleDistinctPackageReferences`

sschuberth: Typo in "Distinct".

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -387,15 +387,18 @@ private fun Collection<DependencyReference>.toGraph(): Pair<List<DependencyGraph
     val edges = mutableListOf<DependencyGraphEdge>()
```
MarcelBochtler, line 400: Commit message: `guarateed` -> `guaranteed`

MarcelBochtler (dismissed):

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -173,6 +173,15 @@ class DependencyGraphBuilder<D>(
             "The following references do not actually refer to packages: " +
                     "${validPackageDependencies - resolvedPackages.keys}."
         }
+
+        val packageReferencesKeysWithMultipleDistintPackageReferences = directDependencies.groupBy { it.key }.filter {
+            it.value.distinct().size > 1
+        }.map { it.key }
```
sschuberth, line 194: Could be just `.keys`.

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
sschuberth: Also:

* "creates an edge" -> "created an edge"
* "but hasn't created" -> "but didn't create"
* "Note: the previous" -> "Note: The previous"

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -167,7 +168,19 @@ class DependencyGraphBuilder<D>(
 
         val (nodes, edges) = directDependencies.toGraph()
 
-        return DependencyGraph(dependencyIds, sortedSetOf(), scopeMapping, nodes, edges)
+        return DependencyGraph(dependencyIds, sortedSetOf(), scopeMapping, nodes, edges.removeAllCycles())
+    }
+
+    private fun Collection<DependencyGraphEdge>.removeAllCycles(): List<DependencyGraphEdge> {
+        val edges = mapTo(mutableSetOf()) { it.from to it.to }
```
sschuberth, line 175: Does it make things easier to use `associate { it.from to it.to }` here and work on a map rather than a set of pairs?

MarcelBochtler (commented):
LGTM, but I don't feel confident enough to be the one approving it. I'd appreciate a review from @sschuberth or @mnonnenmacher who are more familiar with the dependency graph.

Edit: This fixes the issue for us in a project, where it was reproducible before.

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
fviernau: IMO not, because `edges` are passed to `breakAllCycles(edges: Collection<Pair<Int, Int>>)`.
Making the parameter a `Map` would mean that each node can only have one outgoing edge which is not desired.

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -173,6 +173,15 @@ class DependencyGraphBuilder<D>(
             "The following references do not actually refer to packages: " +
                     "${validPackageDependencies - resolvedPackages.keys}."
         }
+
```
sschuberth, line 191: Commit message: "pkg-indx" -> "pkg-idx" (I guess)

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -19,6 +19,8 @@
 
 package org.ossreviewtoolkit.model.utils
 
+import java.util.LinkedList
```
sschuberth, line 22: In the commit message, I find it a bit misleading to refer to `visitEach()` which did not exist before. Also IMO the main point is not an "out factoring" of code, but the conversion of a recursive algorithm into a queue-based one. So, maybe better say: "DependencyGraphBuilder: Avoid recursion when constructing the graph".

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -429,3 +442,46 @@ private data class NodeKey(
...
 private val DependencyReference.key: NodeKey
     get() = NodeKey(pkg, fragment)
+
+private enum class NodeColor { WHITE, GRAY, BLACK }
+
+/**
+ * A depth-first-search (DFS)-based implementation which breaks all cycles in O(V + E).
+ * Finding a minimal solution is NP-complete.
+ */
+internal fun breakAllCycles(edges: Collection<Pair<Int, Int>>): Set<Pair<Int, Int>> {
```
sschuberth, line 452: Nit: Does the "All" as part of the function name add any value? Could it be just "breakCycles"?

sschuberth (changes requested):

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -385,26 +387,38 @@ private fun Collection<DependencyReference>.toGraph(): Pair<List<DependencyGraph
...
 
-    constructGraph(this)
     return nodes.toList() to edges
 }
 
+private fun Collection<DependencyReference>.visitEach(visit: (ref: DependencyReference) -> Unit) {
+    val visited = mutableSetOf<NodeKey>()
+    val queue = LinkedList(this)
+
+    while (queue.isNotEmpty()) {
```
nnobelis, line 425: Question: Instead of this queue management, would it not be more readable with `flatMap` ?

model/src/test/kotlin/utils/DependencyGraphBuilderTest.kt
```diff
@@ -249,6 +250,38 @@ class DependencyGraphBuilderTest : WordSpec({
...
+            val edges = listOf(
+                1 to 2,
+                2 to 3,
+                3 to 4,
+                4 to 1
+            )
+
+            val result = breakCycles(edges)
+
+            result.intersect(edges) shouldHaveSize 3
```
nnobelis, line 282: The test is a little bit too light. Shouldn't we test the content of `result` ?

nnobelis (changes requested):

model/src/test/kotlin/utils/DependencyGraphBuilderTest.kt
fviernau: The test does test the content.

I disagree with that, because any added strictness would make the test too strict:
The algorithm has the freedom to remove any 1 edge from the circle, and if the algorithm is changed to remove a different one the text shouldn't fail. 

What do you think?

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
fviernau: I find it actually readable and prefer to keep it as-is. 

I'm not exactly sure what you envision with the use of `flatMap`, so if you want to have it considered further mind sharing the code you have in mind?

model/src/test/kotlin/utils/DependencyGraphBuilderTest.kt
nnobelis: Ah yes this is a valid remark.
In this case, testing the amount of edges is the only thing you can do.

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -385,26 +387,38 @@ private fun Collection<DependencyReference>.toGraph(): Pair<List<DependencyGraph
...
+        val ref = queue.removeFirst()
+        val key = ref.key
+
+        if (key !in visited) {
+            visit(ref)
+            visited += ref.key
```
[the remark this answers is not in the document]
fviernau: I've inlined the variable

model/src/test/kotlin/utils/DependencyGraphBuilderTest.kt
fviernau: It's not testing only the amount of edges. The intersection implies that it does test the content.

MarcelBochtler (approved):

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
nnobelis: My bad... I thought it would be easy as:
```
flatMap {
        it.dependencies + it
    }.forEach {
        if (it.key !in visited) {
            visit(it)
            visited += it.key
        }
    }
```
but this code goes only to depth 1.
There are some technics with `flatMap` that allow any depth, but they are less elegant than yours with `Queue`.

nnobelis (approved):

sschuberth:
@oheger-bosch would be good if you could also take a look at this in retrospect once you're back.

oheger-bosch:
@fviernau this is interesting stuff, especially with the algorithm to break cycles.

Most of the changes look good to me, there is only one thing I am not sure about: The original bug occurred because a `to` reference of an edge might be encountered, for which no `DependencyGraphNode` has been created yet. This has been fixed in 143ed1b17de2053e97162a51cfc3eb7b12538437 by adding an additional check. So, is it still required to have the `NodeKey` class? As there is a 1:1 relation between `DependencyReference`s and `NodeKey` objects (in this PR a check has been added to verify this), it should be possible to use the `DependencyReference`s as keys of the `nodeIndices` map as before.

I am a bit concerned, because the `DependencyRef.key` property is queried frequently, which may cause the creation of many `NodeKey` objects.

fviernau:
Regarding the "key", I introduced it in a separate commit: 42aa2d1958427c79c72c8c158979bb69f5ed7130.
I believe that even though the requires check exists the argumentation in the commit message of that commit still applies.
What do you think ?

oheger-bosch:
@fviernau : `DependencyReference` is by intension no data class. Thus it uses the hash code inherited from `Object`, and comparisons are done on a reference basis. When constructing the graph `DependencyGraphBuilder` makes sure that all references are de-duplicated. So it is safe to use these references directly as keys.

fviernau:
@oheger-bosch so the remaining benefit of the "key" class is the removal of (hidden) assumptions (such as you mention) which IMO increases clarity and reduces complexity. I'd prefer to keep the key class also because I do not anticipate any performance issues. If you strongly feel it should be removed I won't block it.

oheger-bosch:
@fviernau The impact on performance is hard to predict, especially for huge dependency graphs. But let's keep it for now and do an optimization later if it should be necessary.

PR #4429, opened 2021-09-02
Flutter: Update to 2.2.3

MarcelBochtler:
Flutter 2.2.3 uses Dart version 2.13.4.
Change the used `pub` command because the `pub` binary is deprecated and
no longer contained as binary in the flutter packages from [1].
Update minimum dart version requirement to 2.10.x because this is the
first version providing the `dart pub` subcommand [2].

[1]: https://flutter.dev/docs/development/tools/sdk/releases
[2]: https://dart.dev/tools/pub/cmd


sschuberth:
FYI, this conflicts with https://github.com/oss-review-toolkit/ort/pull/3993.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -580,15 +581,27 @@ class Pub(
...
     }
 
+    override fun getVersion(workingDir: File?): String {
+        val result = ProcessCapture(workingDir, command(workingDir), getVersionArguments())
+        if (result.isError) {
+            throw IOException(result.errorMessage)
+        }
+
+        return transformVersion(result.stderr)
+    }
```
sschuberth, line 591: From a quick glance this looks pretty much like the default implementation, so why is this overridden?

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -580,15 +581,27 @@ class Pub(
...
+
     override fun command(workingDir: File?): String =
-        if (pubAbsolutePath.isDirectory) "$pubAbsolutePath${File.separator}$pubCommand" else pubCommand
+        if (pubAbsolutePath.isDirectory) "$pubAbsolutePath${File.separator}$dartCommand" else dartCommand
+
+    private fun commandPub(): String =
+        if (pubAbsolutePath.isDirectory) "$pubAbsolutePath${File.separator}$dartCommand pub" else "$dartCommand pub"
```
sschuberth, line 597: So, could this be just `"${command()} pub"`?

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -75,9 +76,9 @@ private const val GRADLE_VERSION = "5.6.4"
 private const val PUB_LOCK_FILE = "pubspec.lock"
 
 private val flutterCommand = if (Os.isWindows) "flutter.bat" else "flutter"
-private val pubCommand = if (Os.isWindows) "pub.bat" else "pub"
+private val dartCommand = if (Os.isWindows) "dart.bat" else "dart"
```
sschuberth, line 79: This actually needs to be `dart.exe` on Windows (which causes the test failure),

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -580,15 +581,27 @@ class Pub(
...
+    private fun commandPub(): String =
+        if (pubAbsolutePath.isDirectory) "$pubAbsolutePath${File.separator}$dartCommand pub" else "$dartCommand pub"
 
     private fun commandFlutter(): String =
         if (flutterAbsolutePath.isDirectory) "$flutterAbsolutePath${File.separator}$flutterCommand packages"
         else "$flutterCommand packages"
 
     override fun run(workingDir: File?, vararg args: String): ProcessCapture {
-        var result = ProcessCapture(workingDir, *command(workingDir).split(' ').toTypedArray(), *args)
+        var result = ProcessCapture(workingDir, *commandPub().split(' ').toTypedArray(), *args)
```
sschuberth, line 604: Can we inline `commandPub()` to just `ProcessCapture(workingDir, command(), "pub", *args)`?

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Pub.kt
MarcelBochtler: The downloaded package also contains the `dart.bat` file.
I'll try your suggestion, but I don't really understand why this is required.

sschuberth: At least the Windows package I downloaded only contains a `dart.exe` in the path where `dart.bat` is being looked for.

sschuberth: Here's the content that I see:

```
 ~/Downloads/flutter_windows_2.2.3-stable/flutter (stable)
$ find . -type f -name "dart.exe" -or -name "dart.bat"
./bin/cache/dart-sdk/bin/dart.exe
./bin/dart.bat
```

sschuberth: And because of

    private val pubAbsolutePath = flutterAbsolutePath.resolve("cache/dart-sdk/bin")

the path being used is "bin/cache/dart-sdk/bin", not just "bin".

analyzer/src/main/kotlin/managers/Pub.kt
MarcelBochtler: It would be possible, but I prefer to explicitly specify a separate method for the pub command, as this is the main package manager command.

sschuberth (dismissed):

sschuberth:
@MarcelBochtler I've taken the liberty to push the Windows fix (plus a few more trivial but rather unrelated changes). However, locally still "resolve dependencies for a project with flutter correctly" fails for me, but this time because the dependency tree differs from the expected one:

image

Somehow the Maven dependencies are not detected at all...

sschuberth:
> Somehow the Maven dependencies are not detected at all...

Hmm, as it works on Azure, probably a problem specific to my local setup.

sschuberth:
@fviernau would you be fine with merging this, as I assume you've customized your `FLUTTER_VERSION` anyway?

sschuberth:
@MarcelBochtler can you please rebase this once more as tests seem to got stuck?

sschuberth:
> @MarcelBochtler can you please rebase this once more as tests seem to got stuck?

Please wait and let's first see if the re-run that I just triggered works.

fviernau (approved):

fviernau:
> @fviernau would you be fine with merging this, as I assume you've customized your `FLUTTER_VERSION` anyway?

I'm fine with merging this. Yes, we use custom Flutter version in our org.

sschuberth:
Merging despite the unrelated test failures on Windows.

Issue #2579, opened 2020-05-05
Java Exception in scanner after proxy failures


NicolasToussaint:
Just tried to scan a Maven project with ORT Scanner with fresh sources, and made ORT crash with the 3 scanners (Licensee, LicenseFinder and ScanCode).

I seem to have qui a few networking problem before that (possibly both Proxy configuration and proxy responsiveness), but in all cases it ends wit the following Exception:

```
16:41:46.355 [main] INFO  org.ossreviewtoolkit.model.OrtResult - Computing excluded projects which may take a while...
Exception in thread "main" java.lang.IllegalArgumentException: The VcsInfo(type=Git, url=https://github.com/spring-projects/spring-boot.git, revision=, resolvedRevision=null, path=spring-boot-starter-parent/_proj
ect) of project 'Maven:com.orange.accessibility:_project:1.0.0' cannot be found in Repository(vcs=VcsInfo(type=, url=, revision=, resolvedRevision=null, path=), vcsProcessed=VcsInfo(type=, url=, revision=, resolv
edRevision=null, path=), nestedRepositories={}, config=RepositoryConfiguration(excludes=Excludes(paths=[], scopes=[]), resolutions=Resolutions(issues=[], ruleViolations=[]), curations=Curations(licenseFindings=[]
))).
        at org.ossreviewtoolkit.model.OrtResult.getFilePathRelativeToAnalyzerRoot(OrtResult.kt:259)
        at org.ossreviewtoolkit.model.OrtResult.getDefinitionFilePathRelativeToAnalyzerRoot(OrtResult.kt:245)
        at org.ossreviewtoolkit.model.config.Excludes.findPathExcludes(Excludes.kt:53)
        at org.ossreviewtoolkit.model.OrtResult$projects$2.invoke(OrtResult.kt:94)
        at org.ossreviewtoolkit.model.OrtResult$projects$2.invoke(OrtResult.kt:43)
        at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
        at org.ossreviewtoolkit.model.OrtResult.getProjects(OrtResult.kt)
        at org.ossreviewtoolkit.model.OrtResult.getProject(OrtResult.kt:331)
        at org.ossreviewtoolkit.scanner.Scanner.scanOrtResult(Scanner.kt:121)
        at org.ossreviewtoolkit.commands.ScannerCommand.run(ScannerCommand.kt:151)
[...]

PR #4430, opened 2021-09-03
Configuration for project-local package-configurations and curations

MarcelBochtler:
This disables the project-local package-configuration and curations by default, as it was discussed in the [developer meeting](https://github.com/oss-review-toolkit/ort/wiki/Developer-Meeting#2021-09-02).

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -59,6 +59,7 @@ import org.ossreviewtoolkit.utils.core.ORT_PACKAGE_CURATIONS_DIRNAME
 import org.ossreviewtoolkit.utils.core.ORT_PACKAGE_CURATIONS_FILENAME
 import org.ossreviewtoolkit.utils.core.ORT_REPO_CONFIG_FILENAME
 import org.ossreviewtoolkit.utils.core.ORT_RESOLUTIONS_FILENAME
+import org.ossreviewtoolkit.utils.core.log
 import org.ossreviewtoolkit.utils.core.ortConfigDirectory
 
 class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine dependencies of a software project.") {
@@ -184,12 +185,20 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
 
         val repositoryConfiguration = actualRepositoryConfigurationFile?.readValueOrNull() ?: RepositoryConfiguration()
 
+        val packageCurations =
+            FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations.toMutableList()
+
+        val repositoryPackageCurations = repositoryConfiguration.curations.packages
+
+        if (config.enableRepositoryPackageCurations) {
+            packageCurations += repositoryPackageCurations
+        } else if (repositoryPackageCurations.isNotEmpty()) {
+            log.warn { "Local package curation were not applied because the feature is not enabled." }
+        }
+
         val curationProvider = FallbackPackageCurationProvider(
             listOfNotNull(
-                SimplePackageCurationProvider(
-                    FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations +
-                            repositoryConfiguration.curations.packages
-                ),
+                SimplePackageCurationProvider(packageCurations.toSet()),
                 config.analyzer.sw360Configuration?.let {
                     Sw360PackageCurationProvider(it).takeIf { useSw360Curations }
                 },
--- a/cli/src/main/kotlin/commands/EvaluatorCommand.kt
+++ b/cli/src/main/kotlin/commands/EvaluatorCommand.kt
@@ -275,15 +275,21 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
             ortResultInput = ortResultInput.replacePackageCurations(curations)
         }
 
-        val repositoryPackageConfigurations = ortResultInput.repository.config.packageConfigurations
-        val optionPackageConfigurations = packageConfigurationOption.createProvider().getPackageConfigurations()
+        val config = globalOptionsForSubcommands.config
 
-        val packageConfigurationProvider = SimplePackageConfigurationProvider(
-            optionPackageConfigurations + repositoryPackageConfigurations
-        )
+        val packageConfigurations =
+            packageConfigurationOption.createProvider().getPackageConfigurations().toMutableList()
+        val repositoryPackageConfigurations = ortResultInput.repository.config.packageConfigurations
+
+        if (config.enableRepositoryPackageConfigurations) {
+            packageConfigurations += repositoryPackageConfigurations
+        } else if (repositoryPackageConfigurations.isNotEmpty()) {
+            log.warn { "Local package configurations were not applied because the feature is not enabled." }
+        }
+
+        val packageConfigurationProvider = SimplePackageConfigurationProvider(packageConfigurations.toSet())
         val copyrightGarbage = copyrightGarbageFile.takeIf { it.isFile }?.readValue<CopyrightGarbage>().orEmpty()
 
-        val config = globalOptionsForSubcommands.config
         val licenseInfoResolver = LicenseInfoResolver(
             provider = DefaultLicenseInfoProvider(ortResultInput, packageConfigurationProvider),
             copyrightGarbage = copyrightGarbage,
--- a/docs/config-file-ort-yml.md
+++ b/docs/config-file-ort-yml.md
@@ -153,7 +153,9 @@ curations:
     concluded_license: "Apache-2.0"
  ```
 
-For findings in third-party dependencies package-configurations can be used to replace findings:
+To correct identified licenses in a dependency you can use a package configuration to overwrite scanner findings.
+Note that this feature requires `enableRepositoryPackageConfigurations` to be enabled in the `ort.conf` see
+[reference.conf](../model/src/main/resources/reference.conf).
 ```yaml
 package_configurations:
 - id: 'Maven:com.example:package:1.2.3'
--- a/docs/config-file-ort-yml.md
+++ b/docs/config-file-ort-yml.md
@@ -178,6 +178,8 @@ The list of available options for `reason` are defined in
 Package curations can be added if you want to correct metadata of third-party dependencies.
 
 The following example corrects the source-artifact URL of the package with the id `Maven:com.example:dummy:0.0.1`.
+Note that this feature requires `enableRepositoryPackageCurations` to be enabled in the `ort.conf`, see
+[reference.conf](../model/src/main/resources/reference.conf).
 
 e.g.:
 ```yaml
--- a/model/src/main/kotlin/config/OrtConfiguration.kt
+++ b/model/src/main/kotlin/config/OrtConfiguration.kt
@@ -66,6 +66,13 @@ data class OrtConfiguration(
     val enableRepositoryPackageCurations: Boolean = false,
 
     /**
+     * Enable the usage of project-local package configurations from the [RepositoryConfiguration]. If set to true,
+     * apply package configurations from a local .ort.yml file before applying those specified via the command line i.e.
+     * configurations from the .ort.yml take precedence.
+     */
+    val enableRepositoryPackageConfigurations: Boolean = false,
+
+    /**
      * The configuration of the analyzer.
      */
     val analyzer: AnalyzerConfiguration = AnalyzerConfiguration(),
--- a/model/src/main/kotlin/config/OrtConfiguration.kt
+++ b/model/src/main/kotlin/config/OrtConfiguration.kt
@@ -59,6 +59,13 @@ data class OrtConfiguration(
     val severeRuleViolationThreshold: Severity = Severity.WARNING,
 
     /**
+     * Enable the usage of project-local package curations from the [RepositoryConfiguration]. If set to true, apply
+     * package curations from a local .ort.yml file before applying those specified via the command line i.e. curations
+     * from the.ort.yml take precedence.
+     */
+    val enableRepositoryPackageCurations: Boolean = false,
+
+    /**
      * The configuration of the analyzer.
      */
     val analyzer: AnalyzerConfiguration = AnalyzerConfiguration(),
--- a/model/src/main/resources/reference.conf
+++ b/model/src/main/resources/reference.conf
@@ -10,6 +10,8 @@ ort {
   severeIssueThreshold = ERROR
   severeRuleViolationThreshold = ERROR
 
+  enableRepositoryPackageCurations = true
+
   analyzer {
     ignoreToolVersions = true
     allowDynamicVersions = true
--- a/model/src/main/resources/reference.conf
+++ b/model/src/main/resources/reference.conf
@@ -11,6 +11,7 @@ ort {
   severeRuleViolationThreshold = ERROR
 
   enableRepositoryPackageCurations = true
+  enableRepositoryPackageConfigurations = true
 
   analyzer {
     ignoreToolVersions = true
--- a/model/src/test/kotlin/config/OrtConfigurationTest.kt
+++ b/model/src/test/kotlin/config/OrtConfigurationTest.kt
@@ -179,6 +179,7 @@ class OrtConfigurationTest : WordSpec({
             }
 
             ortConfig.enableRepositoryPackageCurations shouldBe true
+            ortConfig.enableRepositoryPackageConfigurations shouldBe true
         }
 
         "correctly prioritize the sources" {
--- a/model/src/test/kotlin/config/OrtConfigurationTest.kt
+++ b/model/src/test/kotlin/config/OrtConfigurationTest.kt
@@ -177,6 +177,8 @@ class OrtConfigurationTest : WordSpec({
                 patentFilenames shouldContainExactly listOf("patents")
                 rootLicenseFilenames shouldContainExactly listOf("readme*")
             }
+
+            ortConfig.enableRepositoryPackageCurations shouldBe true
         }
 
         "correctly prioritize the sources" {
--- a/model/src/test/kotlin/config/OrtConfigurationTest.kt
+++ b/model/src/test/kotlin/config/OrtConfigurationTest.kt
@@ -162,9 +162,7 @@ class OrtConfigurationTest : WordSpec({
                     useSsl shouldBe true
                     fromAddress shouldBe "no-reply@oss-review-toolkit.org"
                 }
-            }
 
-            with(ortConfig.notifier) {
                 jira shouldNotBeNull {
                     host shouldBe "localhost"
                     username shouldBe "user"


model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -43,5 +43,12 @@ data class AnalyzerConfiguration(
...
      */
-    val sw360Configuration: Sw360StorageConfiguration? = null
+    val sw360Configuration: Sw360StorageConfiguration? = null,
+
+    /**
+     * Enable the usage of project-local package-curations from the [RepositoryConfiguration]. If set to true apply
+     * package-curations from a local .ort.yml file. These package-curations are merged with package-curations from
+     * the command line option, while the one from .ort.yml take precedence.
+     */
+    val applyRepositoryPackageCurations: Boolean = false
```
fviernau, line 53: How about `enablePackageCurationsInRepositoryConfiguration`?

I believe that this option potentially applies to all components other than the analyzer.
IIRC the evaluator also has an option to re-apply curations.
So how about moving that to `OrtConfiguration`?

I presume in another PR you'll add a similar flag for `PackageConfigurations`.
These are not used by the analyzer, but evalutor / reporters. I guess these would fit into `OrtConfigurations` so
the option introduced here would fit right next to it?!

model/src/main/kotlin/config/EvaluatorConfiguration.kt
```diff
@@ -0,0 +1,37 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonInclude
+
+/**
+ * The configuration model of the evaluator.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class EvaluatorConfiguration(
```
fviernau, line 30: As proposed, move to `OrtConfiguration`...

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -263,15 +264,26 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
...
 
-        val packageConfigurationProvider = SimplePackageConfigurationProvider(
-            optionPackageConfigurations + repositoryPackageConfigurations
-        )
+        if (config.evaluator.applyRepositoryPackageConfigurations) {
+            packageConfigurations += repositoryPackageConfigurations
+        } else {
+            if (repositoryPackageConfigurations.isNotEmpty()) {
+                log.warn {
+                    "Package-configurations from '$ORT_REPO_CONFIG_FILENAME' not applied because the feature is not" +
```
fviernau, line 278: I believe that both assumptions about the filenames are wrong.
The `.ort.yml` could also be called `my-project.ort.yml`. 
IIRC also the ort config filename can be overridden.
How about dropping the filename?

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -263,15 +264,26 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
...
+            packageConfigurationOption.createProvider().getPackageConfigurations().toMutableList()
         val repositoryPackageConfigurations = ortResultInput.repository.config.packageConfigurations
-        val optionPackageConfigurations = packageConfigurationOption.createProvider().getPackageConfigurations()
 
-        val packageConfigurationProvider = SimplePackageConfigurationProvider(
-            optionPackageConfigurations + repositoryPackageConfigurations
-        )
+        if (config.evaluator.applyRepositoryPackageConfigurations) {
+            packageConfigurations += repositoryPackageConfigurations
+        } else {
```
fviernau, line 275: can be `else if {` for reduced nesting.

fviernau (changes requested):

model/src/main/kotlin/config/AnalyzerConfiguration.kt
MarcelBochtler: > How about `enablePackageCurationsInRepositoryConfiguration`?

I like the shorter `applyRepositoryPackageCurations` more.
However as the `.ort.yml` is called "repository configuration" in all the docs, I'll change it.

fviernau: The reason for my proposal was that I was afraid some pick-up "RepositoryPackageCuration" as new term.
Not sure if it's a valid concern.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -173,12 +174,20 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
 
+        val packageCurations =
+            FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations.toMutableList()
+
+        val repositoryPackageCurations = repositoryConfiguration.curations.packages
+
+        if (config.applyRepositoryPackageCurations) {
+            packageCurations += repositoryPackageCurations
+        } else if (repositoryPackageCurations.isNotEmpty()) {
+            log.warn { "Local package-curation were not applied because the feature is not enabled." }
```
fviernau, line 185: `package-curation` -> `package curations`?

fviernau (changes requested):

model/src/main/kotlin/config/AnalyzerConfiguration.kt
fviernau: Another way to avoid that could be naming the variable `ignoreLocalPackageCurations`. What do you think?

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -173,12 +174,20 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
 
         val repositoryConfiguration = actualRepositoryConfigurationFile?.readValueOrNull() ?: RepositoryConfiguration()
 
+        val packageCurations =
+            FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations.toMutableList()
```
sschuberth, line 178: How about using `.toMutableSet()` here and omitting `.toSet()` in line 190 below?

model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -43,5 +43,5 @@ data class AnalyzerConfiguration(
     /**
      * Configuration of the SW360 package curation provider.
      */
-    val sw360Configuration: Sw360StorageConfiguration? = null
+    val sw360Configuration: Sw360StorageConfiguration? = null,
```
sschuberth, line 46: Trailing comma introduced.

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -48,6 +48,13 @@ data class OrtConfiguration(
      */
     val severeIssueThreshold: Severity = Severity.WARNING,
 
+    /**
+     * Enable the usage of project-local package-curations from the [RepositoryConfiguration]. If set to true apply
+     * package-curations from a local .ort.yml file. These package-curations are merged with package-curations from
+     * the command line option, while the one from .ort.yml take precedence.
+     */
+    val applyRepositoryPackageCurations: Boolean = false,
```
sschuberth, line 56: Both the log statement in `AnalyzerCommand.kt` and the docs here use the term "enable". So how about also naming the property `enableRepositoryPackageCurations`?

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -263,15 +263,21 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
             ortResultInput = ortResultInput.replacePackageCurations(curations)
         }
 
+        val config = globalOptionsForSubcommands.config
+
+        val packageConfigurations =
+            packageConfigurationOption.createProvider().getPackageConfigurations().toMutableList()
```
sschuberth, line 269: Should this also be a (mutable) set for deduplication?

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -263,15 +263,21 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
...
         val repositoryPackageConfigurations = ortResultInput.repository.config.packageConfigurations
-        val optionPackageConfigurations = packageConfigurationOption.createProvider().getPackageConfigurations()
 
-        val packageConfigurationProvider = SimplePackageConfigurationProvider(
-            optionPackageConfigurations + repositoryPackageConfigurations
-        )
+        if (config.applyRepositoryPackageConfigurations) {
+            packageConfigurations += repositoryPackageConfigurations
+        } else if (repositoryPackageConfigurations.isNotEmpty()) {
+            log.warn { "Local package-configurations were not applied because the feature is not enabled." }
```
sschuberth, line 275: Also here: "package configurations" without a dash?

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -55,6 +55,13 @@ data class OrtConfiguration(
      */
     val applyRepositoryPackageCurations: Boolean = false,
 
+    /**
+     * Enable the usage of project-local package-configurations from the [RepositoryConfiguration]. If set to true apply
+     * package-configurations from a local .ort.yml file. These package-configurations are merged with
+     * package-configurations from the command line option, while the one from .ort.yml take precedence.
+     */
+    val applyRepositoryPackageConfigurations: Boolean = false,
```
sschuberth, line 63: Also here: "enable" vs "apply"?

sschuberth (changes requested):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
MarcelBochtler: AFAIK, a Set does not guarantee to retain the order of the elements, but this is required to ensure that the local curations will take precedence over the global ones.

See Line 183:
```kt
if (config.applyRepositoryPackageCurations) {
    packageCurations += repositoryPackageCurations
} else if (repositoryPackageCurations.isNotEmpty()) {
    log.warn { "Local package-curation were not applied because the feature is not enabled." }
}
```

model/src/main/kotlin/config/AnalyzerConfiguration.kt
MarcelBochtler: > Another way to avoid that could be naming the variable `ignoreLocalPackageCurations`. What do you think?

I don't like to turn around the logic when using `ignore`, because it will introduce double negative when turning this feature on.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
sschuberth: For a general set you're right, but `toMutableSet()` guarantees that "The returned set preserves the element iteration order of the original collection."

model/src/main/kotlin/config/AnalyzerConfiguration.kt
MarcelBochtler: > The reason for my proposal was that I was afraid some pick-up "RepositoryPackageCuration" as new term.
> Not sure if it's a valid concern.

`LocalPackageCurations` would also introduce a new term.
I think as `Repository Configuration` is already a term used in multiple places, `repository curation` is IMO self-explanatory.

What do you think about @sschuberth's suggestion `enableRepositoryPackageCurations` [here](https://github.com/oss-review-toolkit/ort/pull/4430#discussion_r702615271)?

fviernau: Ok with me. I don't have a suggestion which solves the issue I outlined :-(

fviernau: > What do you think about @sschuberth's suggestion `enableRepositoryPackageCurations` [here](https://github.com/oss-review-toolkit/ort/pull/4430#discussion_r702615271)?

Ok with me. I don't have a suggestion which solves the issue I outlined :-(

sschuberth: PS: I was also thinking about "local" vs. "repository", and while I like "local" for being the more generic term, and ORT *used* to work fine for analyzing local non-VCS directories, that [does not really true hold anymore](https://github.com/oss-review-toolkit/ort/issues/2579#issuecomment-913166262), and we implicitly assume local directories to always be VCS working trees. Also, the term "repository" is already used e.g. in `RepositoryConfiguration`, which otherwise should also be renamed to `LocalConfiguration` or `ProjectConfiguation` (the latter actually is what I wanted to discuss in the ORT Developer Meeting for quite some time, see the [wiki](https://github.com/oss-review-toolkit/ort/wiki/Developer-Meeting)).

cli/src/main/kotlin/commands/AnalyzerCommand.kt
MarcelBochtler: Yes in the initial transformation to a mutable map.
When adding other elements, which is being done in Line 183, the order is no longer guaranteed, right?

sschuberth: > When adding other elements, which is being done in Line 183, the order is no longer guaranteed, right?

Hmm, you're probably right, good point.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -56,6 +56,7 @@ import org.ossreviewtoolkit.utils.ORT_PACKAGE_CURATIONS_DIRNAME
 import org.ossreviewtoolkit.utils.ORT_PACKAGE_CURATIONS_FILENAME
 import org.ossreviewtoolkit.utils.ORT_REPO_CONFIG_FILENAME
 import org.ossreviewtoolkit.utils.expandTilde
+import org.ossreviewtoolkit.utils.log
```
sschuberth, line 59: The commit message still says "applyRepositoryPackageCurations".

BTW, my general recommendation to not mention newly introduced properties by name in the commit message (for exactly the reason happened here), but rather say something more general like "Introduce a configuration flag for repository package curations".

model/src/test/kotlin/config/OrtConfigurationTest.kt
```diff
@@ -146,9 +146,7 @@ class OrtConfigurationTest : WordSpec({
                     useSsl shouldBe true
                     fromAddress shouldBe "no-reply@oss-review-toolkit.org"
                 }
-            }
```
sschuberth, line 149: Nit: This commit is unrelated to the topics of the "surrounding" commits. While it's perfectly fine to still keep it as part of this PR, maybe just make it the last commit, so it does not "disrupt" the review flow for the "real" topic of this PR.

sschuberth (changes requested):

model/src/main/resources/reference.conf
```diff
@@ -9,6 +9,9 @@ ort {
 
   severeIssueThreshold = ERROR
 
+  enableRepositoryPackageCurations = true
```
tsteenbe, line 12: The commit message and PR message says this feature is disable by default but in the `reference.conf` you show it as enabled.  Should we set it to false and add a comment what these flags do.

MarcelBochtler: > The commit message and PR message says this feature is disable by default but in the `reference.conf` you show it as enabled. Should we set it to false and add a comment what these flags do.

The `reference.conf` does not reflect the default settings. See e.g. line 10 for `severeIssueThreshold`, which defaults to `WARNING`.

If the flag for `enableRepositoryPackageCurations` is not set at all, it is set to `false` and therefore disabled by default.
I did set it to `true` to check in a test if the changed value is taken into account.

> and add a comment what these flags do.

As of now the `reference.conf` does not document any of the other options.
I documented the feature in `docs/config-file-ort-yml.md`.

docs/config-file-ort-yml.md
```diff
@@ -153,7 +153,9 @@ curations:
     concluded_license: "Apache-2.0"
  ```
 
-For findings in third-party dependencies package-configurations can be used to replace findings:
+For findings in third-party dependencies package configurations can be used to replace findings if
+`enableRepositoryPackageConfigurations` is enabled in the evaluator section of `ort.conf`, see:
```
tsteenbe, line 157: The 'in the evaluator section' is incorrect as `enableRepositoryPackageCuration` and `enableRepositoryPackageConfigurations` both are set in the root level of `ort.conf` looking at the submitted change to `reference.conf`

tsteenbe: I would rephrase the sentence to align it more with [config-file-package-configuration-yml.md](https://github.com/oss-review-toolkit/ort/blob/master/docs/config-file-package-configuration-yml.md) wording e.g. "To overwrite scanner findings to correct identified licenses in a dependency a package configuration can be used. Note that this feature requires `enableRepositoryPackageConfigurations` to be enabled in the `ort.conf` see [reference.conf](../model/src/main/resources/reference.conf).

docs/config-file-ort-yml.md
```diff
@@ -178,6 +180,8 @@ The list of available options for `reason` are defined in
 Package curations can be added if you want to correct metadata of third-party dependencies.
 
 The following example corrects the source-artifact URL of the package with the id `Maven:com.example:dummy:0.0.1`.
+This requires to enable `enableRepositoryPackageCurations` in `ort.conf` like it is being done in
```
tsteenbe, line 183: Change to: Note that this feature requires `enableRepositoryPackageCurations` to be enabled in the `ort.conf`, see [reference.conf](../model/src/main/resources/reference.conf).

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -48,6 +48,20 @@ data class OrtConfiguration(
...
+    /**
+     * Enable the usage of project-local package curations from the [RepositoryConfiguration]. If set to true apply
+     * package curations from a local .ort.yml file. These package curations are merged with package curations from
+     * the command line option, while the one from .ort.yml take precedence.
+     */
+    val enableRepositoryPackageCurations: Boolean = false,
+
+    /**
+     * Enable the usage of project-local package configurations from the [RepositoryConfiguration]. If set to true apply
+     * package configurations from a local .ort.yml file. These package configurations are merged with
```
tsteenbe, line 60: Recommend to change to: If set to true, apply package configurations from a local .ort.yml file before applying those specified via the command line e.g. configurations from the.ort.yml take precedence.

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -48,6 +48,20 @@ data class OrtConfiguration(
      */
     val severeIssueThreshold: Severity = Severity.WARNING,
 
+    /**
+     * Enable the usage of project-local package curations from the [RepositoryConfiguration]. If set to true apply
+     * package curations from a local .ort.yml file. These package curations are merged with package curations from
+     * the command line option, while the one from .ort.yml take precedence.
```
tsteenbe, line 54: Recommend to change to: If set to true, apply package curations from a local .ort.yml file before applying those specified via the command line e.g. curations from the.ort.yml take precedence.

tsteenbe (changes requested):

model/src/main/kotlin/config/OrtConfiguration.kt
MarcelBochtler: Done. But replaced `e.g.` with `i.e.`.

model/src/main/kotlin/config/OrtConfiguration.kt
MarcelBochtler: Done. But again used `i.e.` instead of e.g. which I think is the correct term to use here.

docs/config-file-ort-yml.md
MarcelBochtler: Done. Added an `and` after findings.

docs/config-file-ort-yml.md
```diff
@@ -153,7 +153,9 @@ curations:
     concluded_license: "Apache-2.0"
  ```
 
-For findings in third-party dependencies package-configurations can be used to replace findings:
+To overwrite scanner findings by correcting identified licenses in a dependency a package configuration can be used.
```
tsteenbe: Simpler: "To correct identified licenses in a dependency you can use a package configuration to overwrite scanner findings."

docs/config-file-ort-yml.md
```diff
@@ -153,7 +153,9 @@ curations:
     concluded_license: "Apache-2.0"
  ```
 
-For findings in third-party dependencies package-configurations can be used to replace findings:
+To correct identified licenses in a dependency you can use a package configuration to overwrite scanner findings.
```
tsteenbe, line 156: Simpler: "To correct identified licenses in a dependency you can use a package configuration to overwrite scanner findings."

tsteenbe (changes requested):

MarcelBochtler:
@tsteenbe Do you want this merged? Otherwise I'll close this PR.

fviernau (approved):

MarcelBochtler:
Adding the breaking change label, as this change by default disables repository curations / package-configurations.

sschuberth (approved):

sschuberth:

PR #4432, opened 2021-09-06
 docs: Drop the "experimental" state from package configurations

fviernau:
Package configurations have been around for about 1.5 years already,
are actively being used and the implementation has remained stable. So,
they are not experimental anymore.


MarcelBochtler (dismissed):

MarcelBochtler (approved):

fviernau:
Merged despite the one unrelated failing analyzer test which is known to be flaky.

PR #4436, opened 2021-09-06
AnalyzerCommand: Output details about found project types

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -190,7 +190,10 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
             inputDir, distinctPackageManagers, curationProvider, repositoryConfiguration
         ).mergeLabels(labels)
 
-        println("Found ${ortResult.getProjects().size} project(s) in total.")
+        val analyzedProjects = ortResult.getProjects()
+        val countPerType = analyzedProjects.groupingBy { it.id.type }.eachCount()
+        val projectCountDetails = countPerType.toSortedMap().map { (type, count) -> "$type ($count)" }.joinToString()
+        println("Found ${analyzedProjects.size} project(s): $projectCountDetails")
 
         outputDir.safeMkdirs()
         writeOrtResult(ortResult, outputFiles, "analyzer")


cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -190,7 +190,10 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
             inputDir, distinctPackageManagers, curationProvider, repositoryConfiguration
         ).mergeLabels(labels)
 
-        println("Found ${ortResult.getProjects().size} project(s) in total.")
+        val analyzedProjects = ortResult.getProjects()
+        val countPerType = analyzedProjects.groupingBy { it.id.type }.eachCount()
+        val projectCountDetails = countPerType.toSortedMap().map { (type, count) -> "$type ($count)" }.joinToString()
```
fviernau, line 195: Would it make sense to pass the lambda `{ (type, count) -> "$type ($count)" }` to `joinToString()` instead of to `map()`?

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -190,7 +190,10 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
+        println("Found ${analyzedProjects.size} project(s): $projectCountDetails")
```
fviernau, line 196: As `projectCountDetails` might become a bit long in case of many types, how about using `\n` as separator?
(Also for readability)

fviernau (changes requested):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
sschuberth: Unfortunately, I cannot call `joinToString()` on a map directly, so

    countPerType.toSortedMap().joinToString { (type, count) -> "$type ($count)" }

does not work. However, I could do

    countPerType.toSortedMap().entries.joinToString { (type, count) -> "$type ($count)" }

Not sure if that's nicer, though.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
sschuberth: Hmm, I believe cases where more than, say, 2-3 package managers are used in a single repo are rather rare, or? And these should still fit into a single line.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
fviernau: Yeah, I kind of assumed already you'd add the `\n` as well, like:

`countPerType.toSortedMap().entries.joinToString(prefix = "  ", separator="\n  ") { (type, count) -> "$type ($count)" }`

I'll leave it up to you to decide, no strong opinion here.

fviernau (approved):

PR #4438, opened 2021-09-07
FossIdConfig: Align parsing of booleans for clarity and case-insensitivity

sschuberth:
toBooleanStrictOrNull() parses case-sensitively while toBoolean() is
case-insensitive. Align on the latter for more lenient parsing.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
@@ -135,9 +135,9 @@ internal data class FossIdConfig(
                 ?: throw IllegalArgumentException("No FossId User configuration found.")
             val packageNamespaceFilter = fossIdScannerOptions[NAMESPACE_FILTER_PROPERTY].orEmpty()
             val packageAuthorsFilter = fossIdScannerOptions[AUTHORS_FILTER_PROPERTY].orEmpty()
-            val addAuthenticationToUrl = fossIdScannerOptions[CREDENTIALS_IN_URL_PROPERTY].toBoolean()
-            val waitForResult = fossIdScannerOptions[WAIT_FOR_RESULT_PROPERTY]?.toBooleanStrictOrNull() ?: true
-            val deltaScans = fossIdScannerOptions[DELTA_SCAN_PROPERTY].toBoolean()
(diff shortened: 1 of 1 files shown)


fviernau (approved):

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -135,9 +135,9 @@ internal data class FossIdConfig(
                 ?: throw IllegalArgumentException("No FossId User configuration found.")
             val packageNamespaceFilter = fossIdScannerOptions[NAMESPACE_FILTER_PROPERTY].orEmpty()
             val packageAuthorsFilter = fossIdScannerOptions[AUTHORS_FILTER_PROPERTY].orEmpty()
-            val addAuthenticationToUrl = fossIdScannerOptions[CREDENTIALS_IN_URL_PROPERTY].toBoolean()
-            val waitForResult = fossIdScannerOptions[WAIT_FOR_RESULT_PROPERTY]?.toBooleanStrictOrNull() ?: true
-            val deltaScans = fossIdScannerOptions[DELTA_SCAN_PROPERTY].toBoolean()
+            val addAuthenticationToUrl = fossIdScannerOptions[CREDENTIALS_IN_URL_PROPERTY]?.toBoolean() ?: false
+            val waitForResult = fossIdScannerOptions[WAIT_FOR_RESULT_PROPERTY]?.toBoolean() ?: true
```
nnobelis, line 139: This is wrong. If the value is `null`, `toBoolean` returns `false`. And we want the default value to be `true`.

Suggestion: `?.lowercase()?.toBooleanStrictOrNull() ?: true`

nnobelis (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
nnobelis: Ah I see the `?.toBoolean ?: true`

sschuberth: > If the value is `null`, `toBoolean` returns `false`.

Note my use of `?.`. This now calls the non-null version of `String.toBoolean()`, not anymore `String?.toBoolean()`, to make the default value in case of `null` explicit after the `?:`.

nnobelis (approved):

PR #4440, opened 2021-09-07
PackageManager: Fix a broken assumption

### LINKED ISSUE #4441: Evaluator crashes with IllegalArgumentException
Reproduce: Analyze a source tree with just one definition file, for which `resolveDependencies()` throws an exception. Then scan and evaluate that analyzer result.

Stacktrace:
```
12:52:19 10:52:19.859 [main] PERFORMANCE org.ossreviewtoolkit.cli.commands.EvaluatorCommand - Read ORT result from 'scan-result.json' (0.02 MiB) in 1009ms.
12:52:35 10:52:35.412 [main] PERFORMANCE org.ossreviewtoolkit.model.OrtResult - Computing excluded packages...
12:52:35 Exception in thread "main" javax.script.ScriptException: java.lang.IllegalArgumentException: No dependency graph available to initialize DependencyGraphNavigator.
12:52:35 	at org.jetbrains.kotlin.cli.common.repl.KotlinJsr223JvmScriptEngineBase.asJsr223EvalResult(KotlinJsr223JvmScriptEngineBase.kt:101)
12:52:35 	at org.jetbrains.kotlin.cli.common.repl.KotlinJsr223JvmScriptEngineBase.compileAndEval(KotlinJsr223JvmScriptEngineBase.kt:63)
12:52:35 	at kotlin.script.experimental.jvmhost.jsr223.KotlinJsr223ScriptEngineImpl.compileAndEval(KotlinJsr223ScriptEngineImpl.kt:94)
12:52:35 	at org.jetbrains.kotlin.cli.common.repl.KotlinJsr223JvmScriptEngineBase.eval(KotlinJsr223JvmScriptEngineBase.kt:31)
12:52:35 	at java.scripting/javax.script.AbstractScriptEngine.eval(AbstractScriptEngine.java:264)
12:52:35 	at org.ossreviewtoolkit.utils.ScriptRunner.run(ScriptRunner.kt:68)
12:52:35 	at org.ossreviewtoolkit.evaluator.Evaluator.run(Evaluator.kt:68)
12:52:35 	at org.ossreviewtoolkit.cli.commands.EvaluatorCommand.run(EvaluatorCommand.kt:286)
12:52:35 	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:204)
12:52:35 	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:213)
12:52:35 	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:17)
12:52:35 	at com.github.ajalt.clikt.core.CliktCommand.parse(CliktCommand.kt:396)
12:52:35 	at com.github.ajalt.clikt.core.CliktCommand.parse$default(CliktCommand.kt:393)
12:52:35 	at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:411)
12:52:35 	at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:436)
12:52:35 	at org.ossreviewtoolkit.cli.OrtMainKt.main(OrtMain.kt:108)
12:52:35 Caused by: java.lang.IllegalArgumentException: No dependency graph available to initialize DependencyGraphNavigator.
12:52:35 	at org.ossreviewtoolkit.model.DependencyGraphNavigator.<init>(DependencyGraphNavigator.kt:29)
12:52:35 	at org.ossreviewtoolkit.model.CompatibilityDependencyNavigator$Companion.create(CompatibilityDependencyNavigator.kt:57)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult.createDependencyNavigator(OrtResult.kt:469)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult.access$createDependencyNavigator(OrtResult.kt:44)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult$dependencyNavigator$2.invoke(OrtResult.kt:100)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult$dependencyNavigator$2.invoke(OrtResult.kt:100)
12:52:35 	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult.getDependencyNavigator(OrtResult.kt:100)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult$packages$2.invoke(OrtResult.kt:147)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult$packages$2.invoke(OrtResult.kt:136)
12:52:35 	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult.getPackages(OrtResult.kt:136)
12:52:35 	at org.ossreviewtoolkit.model.OrtResult.getPackage(OrtResult.kt:357)
12:52:35 	at org.ossreviewtoolkit.model.licenses.DefaultLicenseInfoProvider.createConcludedLicenseInfo(DefaultLicenseInfoProvider.kt:53)
12:52:35 	at org.ossreviewtoolkit.model.licenses.DefaultLicenseInfoProvider.createLicenseInfo(DefaultLicenseInfoProvider.kt:47)
12:52:35 	at org.ossreviewtoolkit.model.licenses.DefaultLicenseInfoProvider.get(DefaultLicenseInfoProvider.kt:42)
12:52:35 	at org.ossreviewtoolkit.model.licenses.LicenseInfoResolver.createLicenseInfo(LicenseInfoResolver.kt:72)
12:52:35 	at org.ossreviewtoolkit.model.licenses.LicenseInfoResolver.resolveLicenseInfo(LicenseInfoResolver.kt:63)
12:52:35 	at org.ossreviewtoolkit.evaluator.RuleSet.packageRule(RuleSet.kt:53)
12:52:35 	at ScriptingHostfdeec12_Line_1.unhandledLicenseRule(ScriptingHostfdeec12_Line_1.kts:2836)
12:52:35 	at ScriptingHostfdeec12_Line_1.commonRules(ScriptingHostfdeec12_Line_1.kts:3531)
12:52:35 	at ScriptingHostfdeec12_Line_1$ruleSet$1.invoke(ScriptingHostfdeec12_Line_1.kts:3568)
12:52:35 	at ScriptingHostfdeec12_Line_1$ruleSet$1.invoke(ScriptingHostfdeec12_Line_1.kts:9)
12:52:35 	at org.ossreviewtoolkit.evaluator.RuleSetKt.ruleSet(RuleSet.kt:139)
12:52:35 	at ScriptingHostfdeec12_Line_1.<init>(ScriptingHostfdeec12_Line_1.kts:3567)
12:52:35 	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
12:52:35 	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
12:52:35 	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
12:52:35 	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
12:52:35 	at kotlin.script.experimental.jvm.BasicJvmScriptEvaluator.evalWithConfigAndOtherScriptsResults(BasicJvmScriptEvaluator.kt:96)
12:52:35 	at kotlin.script.experimental.jvm.BasicJvmScriptEvaluator.invoke$suspendImpl(BasicJvmScriptEvaluator.kt:41)
12:52:35 	at kotlin.script.experimental.jvm.BasicJvmScriptEvaluator.invoke(BasicJvmScriptEvaluator.kt)
12:52:35 	at kotlin.script.experimental.jvmhost.repl.JvmReplEvaluator$eval$1$res$1.invokeSuspend(legacyReplEvaluation.kt:63)
12:52:35 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
12:52:35 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
12:52:35 	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:274)
12:52:35 	at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:85)
12:52:35 	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59)
12:52:35 	at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source)
12:52:35 	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:38)
12:52:35 	at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source)
12:52:35 	at kotlin.script.experimental.jvmhost.repl.JvmReplEvaluator.eval(legacyReplEvaluation.kt:63)
12:52:35 	at org.jetbrains.kotlin.cli.common.repl.GenericReplCompilingEvaluatorBase.eval(GenericReplCompilingEvaluator.kt:76)
12:52:35 	at org.jetbrains.kotlin.cli.common.repl.GenericReplCompilingEvaluatorBase.compileAndEval(GenericReplCompilingEvaluator.kt:51)
12:52:35 	at org.jetbrains.kotlin.cli.common.repl.KotlinJsr223JvmScriptEngineBase$compileAndEval$1.invoke(KotlinJsr223JvmScriptEngineBase.kt:64)
12:52:35 	at org.jetbrains.kotlin.cli.common.repl.KotlinJsr223JvmScriptEngineBase$compileAndEval$1.invoke(KotlinJsr223JvmScriptEngineBase.kt:63)
12:52:35 	at org.jetbrains.kotlin.cli.common.repl.KotlinJsr223JvmScriptEngineBase.asJsr223EvalResult(KotlinJsr223JvmScriptEngineBase.kt:90)
12:52:35 	... 15 more
```

fviernau:
PackageMangaer: Fix a broken assumption
    
The evaluator calls `CompatibilityDependencyNavigator.create()` in order
to construct a dependency navigator. That function creates an instance
 of DependencyGraphNavigator if `Project.scopeNames` is not null [2].
The constructor of DependencyGraphNavigator in turn requires a
dependency graph to exist [3].
    
That requires check [3] fails if the project has an empty set as
`scopeNames` and at the same time there is no dependency graph available
for the type of the project. This happens e.g. if a source tree with
only one definition file is analyzed, whereas `resolvedDependencies()`
fails with an exception, see [4]. The failed requires check results in
a crash of the evaluator.
    
Guarantee that in case `project.usesTree() == false` a dependency graph
for the corresponding project type exists, to fix that problem.
    
[1] https://github.com/oss-review-toolkit/ort/blob/42845fa5eb5fbb1a9bc617aa778c2979ca0da26d/model/src/main/kotlin/CompatibilityDependencyNavigator.kt#L47-L60
[2] https://github.com/oss-review-toolkit/ort/blob/463a6ebb57cd3e9613b954a2094a509947c85dc3/model/src/main/kotlin/CompatibilityDependencyNavigator.kt#L96
[3] https://github.com/oss-review-toolkit/ort/blob/463a6ebb57cd3e9613b954a2094a509947c85dc3/model/src/main/kotlin/DependencyGraphNavigator.kt#L29-L31
[4] https://github.com/oss-review-toolkit/ort/blob/463a6ebb57cd3e9613b954a2094a509947c85dc3/analyzer/src/main/kotlin/PackageManager.kt#L255
   
Fixes #4441 .

### DIFF
--- a/analyzer/src/main/kotlin/PackageManager.kt
+++ b/analyzer/src/main/kotlin/PackageManager.kt
@@ -34,6 +34,7 @@ import org.apache.maven.project.ProjectBuildingException
 
 import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.downloader.VersionControlSystem
+import org.ossreviewtoolkit.model.DependencyGraph
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Project
 import org.ossreviewtoolkit.model.ProjectAnalyzerResult
@@ -272,7 +273,7 @@ abstract class PackageManager(
 
         afterResolution(definitionFiles)
 
-        return createPackageManagerResult(result)
+        return createPackageManagerResult(result).addDependencyGraphIfMissing()
     }
 
     /**
@@ -301,3 +302,15 @@ abstract class PackageManager(
  */
 fun parseAuthorString(author: String?, vararg delimiters: Char = charArrayOf('<')): String? =
     author?.split(*delimiters, limit = 2)?.firstOrNull()?.trim()?.ifEmpty { null }
+
+private fun PackageManagerResult.addDependencyGraphIfMissing(): PackageManagerResult {
+    // If the condition is true, then [CompatibilityDependencyNavigator] constructs a [DependencyGraphNavigator].
+    // That construction throws an exception if there is no dependency graph available.
+    val isGraphRequired = projectResults.values.flatten().any { it.project.scopeNames != null }
+
+    return if (isGraphRequired && dependencyGraph == null) {
+        copy(dependencyGraph = DependencyGraph())
+    } else {
+        this
+    }
+}
--- a/model/src/test/kotlin/CompatibilityDependencyNavigatorTest.kt
+++ b/model/src/test/kotlin/CompatibilityDependencyNavigatorTest.kt
@@ -217,10 +217,18 @@ class CompatibilityDependencyNavigatorTest : WordSpec() {
  */
 private fun createOrtResult(vararg projects: Project): OrtResult {
     val graph = DependencyGraph(packages = emptyList(), scopeRoots = sortedSetOf(), scopes = emptyMap())
-    val analyzerResult =
-        AnalyzerResult(sortedSetOf(*projects), sortedSetOf(), dependencyGraphs = mapOf("test" to graph))
-    val analyzerRun =
-        AnalyzerRun(result = analyzerResult, environment = Environment(), config = AnalyzerConfiguration())
+
+    val analyzerResult = AnalyzerResult(
+        projects = sortedSetOf(*projects),
+        packages = sortedSetOf(),
+        dependencyGraphs = mapOf("test" to graph)
+    )
+
+    val analyzerRun = AnalyzerRun(
+        result = analyzerResult,
+        environment = Environment(),
+        config = AnalyzerConfiguration()
+    )
 
     return OrtResult.EMPTY.copy(analyzer = analyzerRun)
 }


analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -34,6 +34,7 @@ import org.apache.maven.project.ProjectBuildingException
```
oheger-bosch, line 34: Commit message nit: In the title "PackageManagaer" -> "PackageManager"

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -272,7 +273,7 @@ abstract class PackageManager(
 
         afterResolution(definitionFiles)
 
-        return createPackageManagerResult(result)
+        return createPackageManagerResult(result).addDependencyGraphIfMissing()
```
oheger-bosch, line 276: It would be good to have a unit test to expose the bug; we could then easier check different approaches to solve it.

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -301,3 +302,15 @@ abstract class PackageManager(
  */
 fun parseAuthorString(author: String?, vararg delimiters: Char = charArrayOf('<')): String? =
     author?.split(*delimiters, limit = 2)?.firstOrNull()?.trim()?.ifEmpty { null }
+
+private fun PackageManagerResult.addDependencyGraphIfMissing(): PackageManagerResult {
```
oheger-bosch, line 306: Adding an empty `DependencyGraph` should fix the problem; however, I am not sure whether there is a more direct solution. The check in `CompatibilityDependencyNavigator` whether a project uses the tree format could be changed as well. Would it be sufficient to change it to something like

`private fun Project.usesTree(): Boolean = scopeNames?.isEmpty() ?: true` ?

So, the dependency graph format is only assumed if there are actually scope names available. This change does not break the unit tests; they are still green.

fviernau: > Adding an empty `DependencyGraph` should fix the problem;

I've verified that it does for the case where it happened.

> Would it be sufficient to change it to something like `private fun Project.usesTree(): Boolean = scopeNames?.isEmpty() ?: true ?`

IIUC it would, but I haven't testet it. I thought about this option as well and decided against it because

1. The function name wouldn't match anymore the return value
2. I didn't want to migrate the case from `Graph` to `Tree` because I assumed that our plan is to get rid of the `Tree` in the future. So I thought it moves into the wrong direction.

Anyhow, I can also try changing it if you still think it makes sense. @oheger-bosch do you want me to?

analyzer/src/main/kotlin/PackageManager.kt
fviernau: I've spent some time to add a (unit) test, but it's not easy because the invariants go across components (PackageManager, CompatibilityDependencyNavigator.create()). If the plan is to remove the Tree representation / Compatibility navigator, would it be ok to live without a test for this bug?

analyzer/src/main/kotlin/PackageManager.kt
oheger-bosch: I am a bit on the fence. IIUC, this is about a corner case, in which there are neither scope names nor scopes (in the old format). I would agree that the graph format is the new one and should be used as far as possible; so this is an argument for the proposed solution.

There are, however, two things I do not like with this approach:
* The fix is done at a different place. I mean, it is not directly obvious that a change in `PackageManager` is necessary to impact the selection of the dependency navigator.
* To determine whether a dummy dependency graph needs to be created, logic is executed that is slightly different from the one contained in the `usesTree()` function. Ideally, `PackageManager` should not know about these details at all.

Well, these are no strong objections, just a gut feeling. So I leave the decision, which way to follow, to you.

analyzer/src/main/kotlin/PackageManager.kt
oheger-bosch: My first idea was to construct an `OrtResult` with a single project that has an empty set of scope names and no dependency graphs. When then the `DependencyNavigator` is queried, an exception is currently thrown.

But I see now that this test already depends on a solution. With your approach, such a result would not be created by the `PackageManager`. - But wouldn't this be another argument for not solving the problem in `PackageManager`? `OrtResult` should not fail to return a navigator, no matter what projects it contains?

fviernau: > But I see now that this test already depends on a solution. With your approach, such a result would not be created by the `PackageManager`. - But wouldn't this be another argument for not solving the problem in `PackageManager`? `OrtResult` should not fail to return a navigator, no matter what projects it contains?

So what solution exactly do you have in mind?

oheger-bosch: Changing the `CompatibilityDependencyNavigator` to handle this corner case; namely to modify the `usesTree()` function as I have proposed. Maybe it should then be given a better name.

analyzer/src/main/kotlin/PackageManager.kt
fviernau: I am following you with having things in two places PackageManager and CompatibilityNavigator is not nice. 
However, the package manager currently does construct the OrtResult and I don't see a way to eliminate the construction from the package manager. On the other hand the navigator has some precondition checks. So the only flexibility / alternative I can see is to change the require condition [1] to tolerate the absence of the graph which implies the need for adding a graceful handling. 

So in total we have 3 options:
1. implementation as-is
2. change `usesTree()`
3. change the require condition [1] and add handling

My preference is still towards option 1, because I cannot see clearly much benefit in using the other opions.
Anyhow, If you want me to change it I will do it.

[1] https://github.com/oss-review-toolkit/ort/blob/463a6ebb57cd3e9613b954a2094a509947c85dc3/model/src/main/kotlin/DependencyGraphNavigator.kt#L29-L31

oheger-bosch: Option 3 would also be fine with me. For option 1 I see the drawback that an `OrtResult` can be constructed manually, for which no `DependencyNavigator` can be requested. So options 2 and 3 would be safer in this regard.

oheger-bosch (approved):

fviernau:
@oheger-bosch as discussed, I'll work on a follow up PR to add a requires check to fail early in case a problematic OrtResult is constructed by the analyzer.

PR #4442, opened 2021-09-07
Cleanups related to tracking authors

sschuberth:


analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -131,7 +131,7 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
                     if (!it.name.isNullOrEmpty()) add(it.name)
                 }
 
-                val developers = mavenProject.developers.mapNotNull { it.organization.orEmpty().ifEmpty { it.name } }
+                val developers = mavenProject.developers.mapNotNull { it.organization ?: it.name }
```
fviernau, line 134: Would it make sense to use `isNullOrBlank` in line 131?

fviernau (approved):

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: I didn't really want to change behavior in this PR, but I anyway believe `isNullOrEmpty()` is fine to check for a "not present" value there.

fviernau (approved):

PR #4448, opened 2021-09-09
Improvements to the NOTICE default FreeMarker template

sschuberth:


reporter/src/main/resources/templates/notice/default.ftl
```diff
@@ -62,10 +62,12 @@ ${exceptionText}
         [/#if]
     [/#if]
 [/#list]
-[#-- Add the licenses of all dependencies. --]
+[#-- Add the licenses of all packages. --]
 [#if packages?has_content]
+[#if hasNoticeProjectLicenses]
```
MarcelBochtler, line 64: I think the indentation is wrong for this element?!

sschuberth: Hah, good eyes, fixed!

reporter/src/main/resources/templates/notice/default.ftl
```diff
@@ -67,14 +67,18 @@ ${exceptionText}
     [/#if]
 This software depends on external packages and source code.
 The applicable license information is listed below:
-[/#if]
 
+[/#if]
```
MarcelBochtler, line 71: is this change intentional?

sschuberth: Yes, IIRC this new blank line inside the `[/#if]` is a substitute for the blank line in former line 69 outside the `[/#if]`. TBH, I also only found out by running the tests and tweaking the output.

MarcelBochtler (approved):

PR #4454, opened 2021-09-13
OrtResult: Enhance getRuleViolations to omit resolved violations

MarcelBochtler:
Provide the possibility to remove resolved violations and violations
below a specified threshold.
This is not used in the ORT code, but is intended to be used in the
`rules.kts` and `notifications.kts` scripts.

Signed-off-by: Marcel Bochtler <marcel.bochtler@bosch.io>

### DIFF
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2017-2021 HERE Europe B.V.
+ * Copyright (C) 2021 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -403,10 +404,30 @@ data class OrtResult(
     fun getAdvisorResultsForId(id: Identifier): List<AdvisorResult> = advisorResultsById[id].orEmpty()
 
     /**
-     * Return all [RuleViolation]s contained in this [OrtResult].
+     * Return all [RuleViolation]s contained in this [OrtResult]. Optionally exclude resolved violations with
+     * [omitResolved] and remove violations below the [minSeverity].
      */
     @JsonIgnore
-    fun getRuleViolations(): List<RuleViolation> = evaluator?.violations.orEmpty()
+    fun getRuleViolations(omitResolved: Boolean = false, minSeverity: Severity? = null): List<RuleViolation> {
+        val allViolations = evaluator?.violations.orEmpty()
+
+        val severeViolations = when (minSeverity) {
+            null -> allViolations
+            else -> allViolations.filter { it.severity >= minSeverity }
+        }
+
+        return if (omitResolved) {
+            val resolutions = getResolutions().ruleViolations
+
+            severeViolations.filter { violation ->
+                resolutions.none { resolution ->
+                    resolution.matches(violation)
+                }
+            }
+        } else {
+            severeViolations
+        }
+    }
 
     @JsonIgnore
     fun getExcludes(): Excludes = repository.config.excludes
--- a/model/src/test/kotlin/OrtResultTest.kt
+++ b/model/src/test/kotlin/OrtResultTest.kt
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Copyright (C) 2021 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +25,7 @@ import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.collections.containExactly
 import io.kotest.matchers.collections.containExactlyInAnyOrder
 import io.kotest.matchers.collections.haveSize
+import io.kotest.matchers.collections.shouldContainExactly
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.string.shouldMatch
@@ -32,6 +34,10 @@ import io.kotest.matchers.types.beInstanceOf
 import java.lang.IllegalArgumentException
 
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.model.config.Resolutions
+import org.ossreviewtoolkit.model.config.RuleViolationResolution
+import org.ossreviewtoolkit.model.config.RuleViolationResolutionReason
 import org.ossreviewtoolkit.utils.Environment
 import org.ossreviewtoolkit.utils.test.readOrtResult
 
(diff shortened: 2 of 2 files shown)


model/src/main/kotlin/OrtResult.kt
```diff
@@ -403,10 +403,27 @@ data class OrtResult(
     fun getAdvisorResultsForId(id: Identifier): List<AdvisorResult> = advisorResultsById[id].orEmpty()
 
     /**
-     * Return all [RuleViolation]s contained in this [OrtResult].
+     * Return all [RuleViolation]s contained in this [OrtResult]. Optionally exclude resolved violations with
+     * [omitResolved] and remove violations below the [threshold].
      */
     @JsonIgnore
-    fun getRuleViolations(): List<RuleViolation> = evaluator?.violations.orEmpty()
+    fun getRuleViolations(omitResolved: Boolean = false, threshold: Severity = Severity.HINT): List<RuleViolation> {
```
sschuberth, line 410: IMO the second argument should default to `OrtConfig.severeIssueThreshold` instead.

sschuberth (changes requested):

model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -159,4 +163,94 @@ class OrtResultTest : WordSpec({
...
+                    )
+                )
+            )
+
+            val violations = ortResult.getRuleViolations()
+
+            violations should haveSize(1)
+        }
+
+        "not return any resolved rule violations" {
```
fviernau, line 203: consider naming it: "drop resolved rule violations if omitDropped is true".
The test cases above could have similar name with the boolean swapped.

model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -159,4 +163,94 @@ class OrtResultTest : WordSpec({
...
+                            licenseSource = null,
+                            severity = Severity.HINT,
+                            message = "Message without any resolution",
+                            howToFix = "how to fix message"
+                        )
+                    )
+                )
+            )
+
+            val violations = ortResult.getRuleViolations(true, Severity.WARNING)
```
fviernau, line 251: `true` is a bit magic. How about using named parameter?

fviernau (dismissed):

model/src/main/kotlin/OrtResult.kt
MarcelBochtler: That would make this a breaking change and I think that `getRuleViolation()` without any parameters should return all violations and not implicitly apply the default 'severseIssueThreshold`.

sschuberth: Even if it's a breaking change, I believe that's what we need to do, to align with https://github.com/oss-review-toolkit/ort/pull/4207 and fix the general mess have have in this context.

fviernau: Why should the default value be `OrtConfig.severeIssueThreshold`? 

Above question is independent of this one: Is `OrtConfig.severeIssueThreshold` actually referring to just `OrtIssue`, no to rule violation? I'm asking this as rule violations are highly customizable (as opposed to ORT issues) and what severity actually is considered severe, in context of rule violations, is IMO up the user who implements the rules. What do you think?

model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -159,4 +163,94 @@ class OrtResultTest : WordSpec({
             ortResult.dependencyNavigator should beInstanceOf<DependencyGraphNavigator>()
         }
     }
+
+    "getRuleViolations" should {
+        "return all violations by default" {
```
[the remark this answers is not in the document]
MarcelBochtler: The duplicated code is mainly the creation of the `RuleViolationResolution`, and its message of this is very relevant to this test.
The extracted function would still require to get the `violations` from the `EvaluatorRun` as parameter, and would hide the Resolution's message.

I'm more in favour of keeping it as is.

model/src/main/kotlin/OrtResult.kt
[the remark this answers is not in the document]
MarcelBochtler: Added an example to the commit message.

model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -159,4 +164,94 @@ class OrtResultTest : WordSpec({
...
+                            licenseSource = null,
+                            severity = Severity.HINT,
+                            message = "Message without any resolution",
+                            howToFix = ""
+                        )
+                    )
+                )
+            )
+
+            val violations = ortResult.getRuleViolations(omitResolved = true, minSeverity = Severity.WARNING)
```
fviernau, line 252: The assertion could be made stronger with: 

```
ortResult.getRuleViolations(omitResolved = true, minSeverity = Severity.WARNING).map {it.rule } shouldContainExactlyInAnyOrder(
   "rule id",
   "another rule"
)
```
...same for the other introduced test further above.

model/src/main/kotlin/OrtResult.kt
sschuberth: > Is `OrtConfig.severeIssueThreshold` actually referring to just `OrtIssue`, not to `RuleViolation`?

Now that you name it, I have to agree that the name "severe*Issue*Threshold" is misleading. When I introduced that configuration option, I did not intent to limit it to issues / the `OrtIssue` class, but apply it to anything that has a `Severity` (I probably was under the misbelief that `RuleViolation` inhertis from `OrtIssue`).

That's why even for the Evaluator, which creates no `OrtIssue`s at all but only `RuleViolation`s, we currently have this code:

https://github.com/oss-review-toolkit/ort/blob/42845fa5eb5fbb1a9bc617aa778c2979ca0da26d/cli/src/main/kotlin/commands/EvaluatorCommand.kt#L304-L305

However, we also had already internal discussion about the multiple ways now to tweak the effect of a severity for rule violations: Either by changing the emitted `Severity` in the `rules.kts` directly, or by altering the `severeIssueThreshold`, and how this all should refer to the exit code of the Evaluator.

I guess we need some more brainstorming here.

MarcelBochtler: > However, we also had already internal discussion about the multiple ways now to tweak the effect of a severity for rule violations: Either by changing the emitted `Severity` in the `rules.kts` directly, or by altering the `severeIssueThreshold`, and how this all should refer to the exit code of the Evaluator.

While I agree that this is not ideal, I don't think this is strictly relevant for this PR.

After all I still think that the default behaviour of `getRuleViolations()` should return all violations, and not adhere to the configured `severeIssueThreshold`.

sschuberth: Hmm, I guess I'd be fine with not having a default value at all for the `severity` parameter (just like in the `concludeSeverityStats()` function), but I'd not like to have a default value now and eventually change that default value later (as that makes it hard for callers who rely on the default to check whether they still work as intended).

MarcelBochtler: How about making `minSeverity` nullable and defaulting to `null`?
Which would then mean that no filter at all is applied?

sschuberth: > After all I still think that the default behaviour of `getRuleViolations()` should return all violations

That makes sense. However, the current implementation assumes that `Severity.HINT` is the lowest severity, and that filtering with `it.severity >= minSeverity` actually keeps all violations.

How about instead making `minSeverity` a nullable parameter and not filter at all if it's `null`?

fviernau: > How about instead making `minSeverity` a nullable parameter and not filter at all if it's `null`?

I think in this case it's also ok to use `Severity.values()[0] `, just as alternative idea to avoid nullability.

sschuberth: But that would still filter potentially unnecessarily, and it makes the assumption that the least severe enum value is the first entry (the values could also be sorted exactly the opposite).

model/src/main/kotlin/OrtResult.kt
```diff
@@ -403,10 +404,27 @@ data class OrtResult(
...
-     * Return all [RuleViolation]s contained in this [OrtResult].
+     * Return all [RuleViolation]s contained in this [OrtResult]. Optionally exclude resolved violations with
+     * [omitResolved] and remove violations below the [minSeverity].
      */
     @JsonIgnore
-    fun getRuleViolations(): List<RuleViolation> = evaluator?.violations.orEmpty()
+    fun getRuleViolations(omitResolved: Boolean = false, minSeverity: Severity? = null): List<RuleViolation> {
+        val allViolations = evaluator?.violations.orEmpty()
+
+        val violationsWithMinSeverity = allViolations.filter { minSeverity == null || it.severity >= minSeverity }
```
sschuberth, line 414: Please avoid filter at all in case of `minSeverity == null` by doing something like

        val violationsWithMinSeverity = minSeverity?.let { allViolations.filter { it.severity >= minSeverity } }
            ?: allViolations

sschuberth (changes requested):

model/src/main/kotlin/OrtResult.kt
fviernau: > But that would still filter potentially unnecessarily

I don't think performance tweaks are needed here

> , and it makes the assumption that the least severe enum value is the first entry (the values could also be sorted exactly the opposite).

Assumption about the ordering is made `as-is` already, and IMO it's ok to make assumptions about the ordering.

model/src/main/kotlin/OrtResult.kt
sschuberth: Or maybe nicer:

        val violationsWithMinSeverity = when (minSeverity) {
            null -> allViolations
            else -> allViolations.filter { it.severity >= minSeverity }
        }

model/src/main/kotlin/OrtResult.kt
sschuberth: > I don't think performance tweaks are needed here

I agree they're not needed. But I believe it's still good practice to write code so that unnecessary work is avoided (if readability is not sacrificed).

model/src/main/kotlin/OrtResult.kt
```diff
@@ -403,10 +404,30 @@ data class OrtResult(
...
-     * Return all [RuleViolation]s contained in this [OrtResult].
+     * Return all [RuleViolation]s contained in this [OrtResult]. Optionally exclude resolved violations with
+     * [omitResolved] and remove violations below the [minSeverity].
      */
     @JsonIgnore
-    fun getRuleViolations(): List<RuleViolation> = evaluator?.violations.orEmpty()
+    fun getRuleViolations(omitResolved: Boolean = false, minSeverity: Severity? = null): List<RuleViolation> {
+        val allViolations = evaluator?.violations.orEmpty()
+
+        val violationsWithMinSeverity = when (minSeverity) {
```
sschuberth, line 414: Nit: `violationsWithMinSeverity` is not strictly correct, but `violationsWithAtLeastMinSeverity` is very long, so maybe just `severeViolations`?

sschuberth (dismissed):

sschuberth (approved):

fviernau (approved):

PR #4459, opened 2021-09-14
MailNotifier: Provide the possibility to send HTML emails

MarcelBochtler:
Enable the mail notifier to be able to send HTML mails.

### DIFF
--- a/notifier/src/main/kotlin/modules/MailNotifier.kt
+++ b/notifier/src/main/kotlin/modules/MailNotifier.kt
@@ -19,9 +19,10 @@
 
 package org.ossreviewtoolkit.notifier.modules
 
+import java.nio.charset.Charset
+
 import org.apache.commons.mail.DefaultAuthenticator
-import org.apache.commons.mail.Email
-import org.apache.commons.mail.SimpleEmail
+import org.apache.commons.mail.HtmlEmail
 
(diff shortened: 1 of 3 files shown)


notifier/src/main/kotlin/modules/MailNotifier.kt
```diff
@@ -19,34 +19,42 @@
...
 
     init {
-        client = SimpleEmail()
-
-        client.setHostName(config.hostName)
+        client.hostName = config.hostName
         client.setSmtpPort(config.port)
         client.setAuthenticator(DefaultAuthenticator(config.username, config.password))
-        client.setSSLOnConnect(config.useSsl)
+        client.isSSLOnConnect = config.useSsl
```
sschuberth, line 39: Nit: Unrelated change?

MarcelBochtler: This was not possible with the `Email` interface.
it is only possible with `HtmlEmail`. Should I move it to a separate commit?

sschuberth: Interesting that this is only possible with `HtmlEmail`. If you want, you could move this to a succeeding commit with an explanation that this is now possible but wasn't before. But maybe that's overdoing it; I'd also be fine to keep it as-is.

sschuberth (approved):

PR #3902, opened 2021-04-17
New multi docker

heliocastro:
Dockerfile: Multidocker based on Ubuntu 20.04

* Switchable build arg versions
* Scancode build from releases, not from python npip
* Python uses pyenv. Default python is 3.8.11
* Nodejs is using nvenv.  Default nodejs is the latest LTS
* Ruby uses rbenv. Bundler and Cocoapods now come by default
* Wrapper script to use bashrc on direct ort usage
* Use bash and pipefail on all operations as default. Exception for
Android sdkmanager that behaves differently on pipefail
* Silence gpg dearmor output on tee
* Using a single downloader, curl
* Go updated to last stable release

PR #4462, opened 2021-09-15
Dockerfile: Stop basing the image on the deprecated 'adoptopenjdk'

fviernau:
Use 'eclipse-temurin' instead as recommended by [1]. Note that there is
no alpine image, so use the Ubuntu focal based one also for the build.

[1] https://github.com/docker-library/docs/tree/074a723eee99a7e0999ddbe0cb080fffd5c650a2/adoptopenjdk#deprecation-notice


Dockerfile
```diff
@@ -17,20 +17,21 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
-
-FROM adoptopenjdk/openjdk11:alpine-slim AS build
+FROM eclipse-temurin:11-jdk-focal AS build
```
fviernau, line 20: Switching to focal seems to open a Pandora's box of Python questions as Python2 is no more the default.
Has anyone started looking into that already, maybe @sschuberth  @mnonnenmacher ?

sschuberth: @heliocastro did touch this topic in https://github.com/oss-review-toolkit/ort/pull/3902 (however, that's doing too many things at once for my taste). Maybe it's easier if we tackle the ScanCode upgrade first?

fviernau: Thanks @sschuberth , I'll close this draft then. Upgrading ScanCode first seems very reasonable.
Do we have plans to drop Python2 support from ORT after the ScanCode upgrade @oss-review-toolkit/core-devs  (so that switching to focal would become simpler)?

sschuberth: > Do we have plans to drop Python2 support from ORT after the ScanCode upgrade

I don't think we can (fully drop Python 2 from the Docker image), as we'd still need to be able to analyze Python 2 projects, or?

fviernau: From the complexity the Python2 support adds it would be great to get rid of it. However, if it's a requirement then we have to keep it. Would it make sense to discuss this in the dev meeting ?

sschuberth: Feel free to put it onto the agenda. However, I'm hoping we don't get "from the hundredth to the thousandth" in that ever-lasting Docker-ScanCode-Python vicious circle.

PR #4463, opened 2021-09-15
Make the mapping of authors to copyrights configurable

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/DownloaderCommand.kt
+++ b/cli/src/main/kotlin/commands/DownloaderCommand.kt
@@ -192,6 +192,7 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
     ).enum<DataEntity>().split(",").default(enumValues<DataEntity>().asList())
 
     override fun run() {
+        val config = globalOptionsForSubcommands.config
         val failureMessages = mutableListOf<String>()
 
         when (input) {
@@ -220,7 +221,7 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
                     }
                 }
 
-                val includedLicenseCategories = globalOptionsForSubcommands.config.downloader.includedLicenseCategories
+                val includedLicenseCategories = config.downloader.includedLicenseCategories
 
                 val packageDownloadDirs =
                     if (includedLicenseCategories.isNotEmpty() && licenseClassificationsFile.isFile) {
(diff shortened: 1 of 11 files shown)


sschuberth:
@porsche-rishisaxena, @nikpete, I'd appreciate if you could verify this works as you expect before the merge.

mnonnenmacher (approved):

porsche-rishisaxena:
@rbieniek: Would you be so kind to provide the feedback to this thread?

rbieniek:
The fix is fine to avoid having some author information 'leaking' into the final result as copyright holders.

However, it doesn't IMHO not the more general use case to be able to attach the the copright holder information to the package configuration information upfront. And have it passed through the system to be taken into account when generating the final report.

I'm currently working on a change to add copyright holder information to the package configurtations and have them passed through th processin chain.
Additionally, I will introduce some source information attached to the copyright holder to make it distinguishable where the copyright information originated from, which would by etiher from the package configuration or from the scancode run or sysnthesized useing the auther information

That change wouild complete the information in a way to fulfill the needs for the Porsche use case

sschuberth:
> The fix is fine to avoid having some author information 'leaking' into the final result as copyright holders.

That was the sole purpose of this PR. Adding a way to curate copyrights directly, without going over authors and enabling `addAuthorsToCopyrights`, will be added later.

> I'm currently working on a change to add copyright holder information to the package configurtations and have them passed through th processin chain.

Good to know.

rbieniek:
For clarification, I plan to use (and implement this) by adding the Copyright holder statements to the org.ossreviewtoolkit.model.PackageCurationData entity

sschuberth:
> adding the Copyright holder statements to the org.ossreviewtoolkit.model.PackageCurationData entity

FYI, I just pushed my local [`curate-copyrights` branch](https://github.com/oss-review-toolkit/ort/tree/curate-copyrights) which trivially starts doing that.

@rbieniek do you want to take over that branch of mine?

fviernau:
> For clarification, I plan to use (and implement this) by adding the Copyright holder statements to the org.ossreviewtoolkit.model.PackageCurationData entity

In case you are not aware of it, it may be worth taking a look at [1] to get an idea of how ORT currently assigns the copyright statements to licenses. ORT uses the term root licenses for licenses detected in special files. For copyright statements which have no nearby license text, that matcher assigns these statements to ALL root licenses. 

The result of that license matching is that any copyright statement is either discarded or assigned to at least one license ID. 
The statements are later on consumed in the notice reporter. The data model is such that a copyright statement needs to be associated with a license.

I'm mentioning that just to bring attention to it, so that it gets considered in the design and also to have a starting point for potential use cases. Quick ideas for semantics of the curation could be (not saying all this is needed):
1. Add a copyright statement for a specific license ID
2. Add a copyright statement to all root licenses (which are not specified but detected in the scanning phase)
3. Add a copyright statement to all license IDs



[1] https://github.com/oss-review-toolkit/ort/blob/master/model/src/main/kotlin/utils/RootLicenseMatcher.kt

sschuberth:
Please continue the discussion in https://github.com/oss-review-toolkit/ort/issues/4519 that I just created.

PR #4468, opened 2021-09-16
NotifierCommand: Add the label option to the Notifier

MarcelBochtler:
In contrast to other labels the Notifier labels are not persisted in a
resulting ort-result.yml, as the Notifier itself does not create an
ort-result.yml. However, this option is still useful if specific options
for any of the Notifier implementations are required, e.g. E-Mail
addresses or Jira Ticket numbers.

### DIFF
--- a/cli/src/main/kotlin/commands/NotifierCommand.kt
+++ b/cli/src/main/kotlin/commands/NotifierCommand.kt
@@ -22,6 +22,7 @@ package org.ossreviewtoolkit.cli.commands
 import com.github.ajalt.clikt.core.CliktCommand
 import com.github.ajalt.clikt.core.UsageError
 import com.github.ajalt.clikt.core.requireObject
+import com.github.ajalt.clikt.parameters.options.associate
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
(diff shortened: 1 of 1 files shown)


cli/src/main/kotlin/commands/NotifierCommand.kt
```diff
@@ -52,12 +54,20 @@ class NotifierCommand : CliktCommand(name = "notify", help = "Create notificatio
         .convert { it.absoluteFile.normalize() }
         .inputGroup()
 
+    private val labels by option(
+        "--label", "-l",
+        help = "Set a label in the ORT result, overwriting any existing label of the same name. This label can be " +
```
oheger-bosch, line 59: Nit: Should the help text explicitly state that the label is only transient and will not be persisted? The current statement sounds as if the result would be modified. Maybe something like "Temporarily update a label in the ORT result before it is passed to the notifier script, overwriting any existing label of the same name. This label can then be accessed by the notifier script, but the modified ORT result is not persisted. ..."

oheger-bosch (approved):

cli/src/main/kotlin/commands/NotifierCommand.kt
sschuberth: At first I also though such a hint would make sense, but then again, where should the label eventually be persisted to, as no output result is written anyway? And *if* an output would be written, it would also be persisted.

The current text matches the one in the other CLI tools except for the inserted "This label can be accessed in the used notifier script" sentence. As a compromise, how about:

> Set a label in the ORT result passed to the notifier script, overwriting any existing label of the same name. Can be used multiple times. For example: --label distribution=external

sschuberth (approved):

PR #4475, opened 2021-09-17
ResolvedLicenseInfo: Properly filterExcluded() detected licenses

fviernau:
The effective license (which is shown e.g. in the static HTML report) does not adhere to path excludes for detected licenses.
The required handling hasn't been implemented and is added by this PR. Therefore a dedicated type for original expressions is introduced and extended by the "excluded state", which is required for implementing the handling. Note that for notice templates using `filterExcluded()` the changes may have effect on the output notice file.

In order to simplify these changes, the very first commit simplifies how the excel reporter shows declared licenses.


model/src/main/kotlin/licenses/LicenseInfoResolver.kt
```diff
@@ -131,15 +131,23 @@ class LicenseInfoResolver(
...
 
         resolvedLocations.keys.forEach { license ->
             license.builder().apply {
                 resolvedLocations[license]?.let { locations.addAll(it) }
 
-                originalExpressions += detectedLicenses.filterTo(mutableSetOf()) { license in it.decompose() }.map {
-                    ResolvedOriginalExpression(expression = it, source = LicenseSource.DETECTED)
+                originalExpressions += detectedLicenses.entries.filter { (spdxExpression, _) ->
+                    license in spdxExpression.decompose()
+                }.map { (expression, isDetectedExcluded) ->
```
mnonnenmacher, line 149: For consistency either rename `spdxExpression` to `expression` in the `filter` call or `expression` to `spdxExpression` here.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #4476, opened 2021-09-17
LicenseInfoResolver: Make original expression curated instead of raw

fviernau:
The previous implementation was using raw licenses for original expressions and curated licenses for resolved locations which is inconsistent. As original expressions are the basis for applying license choices these have to be curated.

The PR illustrates and fixes a bug where the resolved license info lacks original expressions.

### DIFF
--- a/model/src/main/kotlin/licenses/LicenseInfoResolver.kt
+++ b/model/src/main/kotlin/licenses/LicenseInfoResolver.kt
@@ -128,7 +128,11 @@ class LicenseInfoResolver(
         val unmatchedCopyrights = mutableMapOf<Provenance, MutableSet<CopyrightFinding>>()
         val resolvedLocations = resolveLocations(filteredDetectedLicenseInfo, unmatchedCopyrights)
         val detectedLicenses = licenseInfo.detectedLicenseInfo.findings.flatMapTo(mutableSetOf()) { findings ->
-            findings.licenses.map { it.license }
+            FindingCurationMatcher().applyAll(
+                findings.licenses,
+                findings.licenseFindingCurations,
+                findings.relativeFindingsPath
+            ).mapNotNull { it.curatedFinding?.license }
         }
 
         resolvedLocations.keys.forEach { license ->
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/licenses/LicenseInfoResolver.kt
```diff
@@ -128,7 +128,13 @@ class LicenseInfoResolver(
...
         val resolvedLocations = resolveLocations(filteredDetectedLicenseInfo, unmatchedCopyrights)
         val detectedLicenses = licenseInfo.detectedLicenseInfo.findings.flatMapTo(mutableSetOf()) { findings ->
-            findings.licenses.map { it.license }
+            val curatedFindings = FindingCurationMatcher().applyAll(
+                findings.licenses,
+                findings.licenseFindingCurations,
+                findings.relativeFindingsPath
+            ).mapNotNull { it.curatedFinding }
+
+            curatedFindings.map { it.license }
```
mnonnenmacher, line 137: Could be merged into the above expression for less iterations: `mapNotNull { it.curatedFinding?.license }`

model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
```diff
@@ -459,6 +459,9 @@ class LicenseInfoResolverTest : WordSpec() {
                         )
                     )
                 )
+                result.licenses.flatMap {
```
mnonnenmacher, line 462: Commit message type: "orig**i**nal"

mnonnenmacher (changes requested):

model/src/main/kotlin/licenses/LicenseInfoResolver.kt
fviernau: IMO keeping the variable helps with readability. Do you agree?

mnonnenmacher: Yes, but I'd still like to avoid the duplicate iteration. So how about moving the `mapNotNull` down to `curatedFindings.map { it.curatedFinding?.license }` (and maybe renaming `curatedFindings` to `curationResults` then)?

fviernau: I ended up going for your original suggestion.

mnonnenmacher (approved):

PR #4483, opened 2021-09-21
Pub: Override the getVersion command

MarcelBochtler:
The default implementation uses the `run` method. In Pub's case this
uses `dart pub`, which does not offer the `--version` flag. Therefore,
using the `dart`'s version information.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -581,6 +581,12 @@ class Pub(
         return yamlMapper.readTree(definitionFile)
     }
 
+    override fun getVersion(workingDir: File?): String {
+        val result = ProcessCapture(workingDir, command(workingDir), getVersionArguments()).requireSuccess()
+
+        return transformVersion(result.stderr)
+    }
+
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -581,6 +581,15 @@ class Pub(
         return yamlMapper.readTree(definitionFile)
     }
 
+    override fun getVersion(workingDir: File?): String {
+        val result = ProcessCapture(workingDir, command(workingDir), getVersionArguments())
+        if (result.isError) {
+            throw IOException(result.errorMessage)
+        }
```
sschuberth, line 588: But `run` is implemented as

    ProcessCapture(workingDir, command(workingDir), *args).requireSuccess()

which does not look much different than your code. What am I missing?

sschuberth: Oops, sorry, I missed that `Pub` also overrides `run()`. But could we use `.requireSuccess()` instead of the custom `if`-clause?

sschuberth (approved):

PR #4484, opened 2021-09-21
scanner: Operate on (deduplicated) sets of packages

sschuberth:
Packages should be distinct for performance reasons, so ensure that
already via the function signatures.

This is, in a sense, a follow-up to 3fca5af.

### DIFF
--- a/scanner/src/main/kotlin/LocalScanner.kt
+++ b/scanner/src/main/kotlin/LocalScanner.kt
@@ -200,7 +200,7 @@ abstract class LocalScanner(
     }
 
     override suspend fun scanPackages(
-        packages: Collection<Package>,
+        packages: Set<Package>,
         outputDirectory: File
     ): Map<Package, List<ScanResult>> {
         val scannerCriteria = getScannerCriteria()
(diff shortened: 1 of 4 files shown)


mnonnenmacher (approved):

scanner/src/main/kotlin/LocalScanner.kt
```diff
@@ -200,7 +200,7 @@ abstract class LocalScanner(
     }
 
     override suspend fun scanPackages(
-        packages: Collection<Package>,
+        packages: Set<Package>,
```
fviernau, line 203: Is `Package` the right key to de-duplicate it. Or should it be `distinct by` the `Identifier` ?

fviernau (approved):

scanner/src/main/kotlin/LocalScanner.kt
sschuberth: I guess we *could* more aggressively de-duplicate based on the `id`, but actually we should never have different `Package`s with the same `id` here, so I believe we are fine.

PR #4488, opened 2021-09-22
Make the advisor module more generic

oheger-bosch:
This PR extends the data model for vulnerabilities by additional fields, so it can be used in a broader context. The new properties are from the domain of issue tracking systems. The goal is to represent more general findings related to software problems.

To reflect this broader scope, the model classes have been renamed - the general term *Finding* is now used. Renaming was also done for classes in the advisor module to indicate that an advisor is not limited to security vulnerabilities, but can produce arbitrary findings.

*Notes*
* Maybe some commits could be squashed, but as comments needed to be updated as well, I think smaller commits are easier to review.
* Some more renaming of client classes could be done. This is possible in a follow-up PR - I would like to get feedback about the proposed fields first.


tsteenbe (changes requested):
I see the use case for making the Advisor findings more generic but I don't like Vulnerability model being completely removed - as in SPDX 3.0 and CycloneDX report we only want to include vulnerabilities and not any of the other advisor findings. 

I would prefer a Advisor code refactoring where parent finding model that's generic but then have specific implementations such as Vulnerability., Code Analysis issues as each may require specific field A Vulnerability may only be reported to a package whilst an code analysis issue may be file and line level.  By implementing it in this way one should be able to also easily filter by type all Advisor findings to only get vulnerabilities, code analysis issues, etc.

oheger-bosch:
@tsteenbe I see your point. If there are significant differences between the properties of different findings, it probably does not make sense to have a one-fits-all data model class.

That, however, opens the question what is the best way to integrate different finding types (with different sets of properties) into ORT's data model. Using base classes with specialized implementations as you propose, seems not to be easy with Kotlin's data classes (see for instance [this StackOverflow question](https://stackoverflow.com/questions/26444145/extend-data-class-in-kotlin)). Also, serialization and deserialization of a polymorphic data model would be rather complex.

So would it be an option to define multiple classes to represent advisor findings in parallel (such as `Vulnerability`, `Defect`, etc.) and extend `AdvisorResult` to store a list for each of these classes? A concrete advisor implementation can then decide which of these lists it populates.

sschuberth:
> That, however, opens the question what is the best way to integrate different finding types

@oheger-bosch, have you considered [sealed classes](https://kotlinlang.org/docs/sealed-classes.html)?

mnonnenmacher:
> So would it be an option to define multiple classes to represent advisor findings in parallel (such as Vulnerability, Defect, etc.) and extend AdvisorResult to store a list for each of these classes? A concrete advisor implementation can then decide which of these lists it populates.

I have a tendency towards this approach, because polymorphism adds more complexity than having separate fields, even though Jackson supports (de-)serialization of sealed classes. If we need to add/remove/change any of the advise types I would expect that it is easier to manage if they are separate, for example to keep backward compatibility. On the other hand separate fields will likely require more code to handle them in places where the differences do not matter.

oheger-bosch:
> 
> 
> > That, however, opens the question what is the best way to integrate different finding types
> 
> @oheger-bosch, have you considered [sealed classes](https://kotlinlang.org/docs/sealed-classes.html)?

Yes, I thought about this, too; but there is still the increased complexity with serialization. Also, I have the impression that the single classes in the hierarchy will not have that much attributes in common; so I don't think there will be much benefit in introducing a base class.

fviernau:
> So would it be an option to define multiple classes to represent advisor findings in parallel (such as `Vulnerability`, `Defect`, etc.) and extend `AdvisorResult` to store a list for each of these classes? A concrete advisor implementation can then decide which of these lists it populates.

+1 for separate classes + separate lists for each instead of one class with all fields. Haven't thought too deep about polymorhic approach yet.

oheger-bosch:
Thanks for the feedback. According to the comments, I close this PR in favour of #4493, which adds a separate class to represent defects.

PR #4491, opened 2021-09-23
Add an experimental provenance based scanner implementation

mnonnenmacher:
Add a new experimental scanner implementation that aims to scan each provenance (meaning source code repository or source artifact) only once. This provides better performance and uses less storage for scan results. The new implementation can be tried by providing the `--experimental` switch to the `scan` command.

Simplified, the algorithm of the new scanner is:
* Resolve the provenance of all packages, including sub-repositories, to determine if there are any overlaps.
* Check if any stored scan results are available for the provenances.
* Scan the remaining provenances.
* Associate the scan results back to the packages.

The implementation is not yet feature complete, missing functionality will be added in later PRs. Please see the commit messages for details.

### DIFF
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Copyright (C) 2021 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -38,6 +39,10 @@ import com.github.ajalt.clikt.parameters.options.split
 import com.github.ajalt.clikt.parameters.types.enum
 import com.github.ajalt.clikt.parameters.types.file
 
+import java.io.File
+
+import kotlinx.coroutines.runBlocking
+
 import org.ossreviewtoolkit.cli.GlobalOptions
 import org.ossreviewtoolkit.cli.concludeSeverityStats
 import org.ossreviewtoolkit.cli.utils.OPTION_GROUP_INPUT
@@ -45,11 +50,20 @@ import org.ossreviewtoolkit.cli.utils.outputGroup
 import org.ossreviewtoolkit.cli.utils.readOrtResult
 import org.ossreviewtoolkit.cli.utils.writeOrtResult
 import org.ossreviewtoolkit.model.FileFormat
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.config.OrtConfiguration
 import org.ossreviewtoolkit.model.utils.mergeLabels
 import org.ossreviewtoolkit.scanner.LocalScanner
 import org.ossreviewtoolkit.scanner.ScanResultsStorage
 import org.ossreviewtoolkit.scanner.Scanner
+import org.ossreviewtoolkit.scanner.experimental.DefaultNestedProvenanceResolver
+import org.ossreviewtoolkit.scanner.experimental.DefaultPackageProvenanceResolver
+import org.ossreviewtoolkit.scanner.experimental.DefaultProvenanceDownloader
+import org.ossreviewtoolkit.scanner.experimental.ExperimentalScanner
 import org.ossreviewtoolkit.scanner.scanOrtResult
+import org.ossreviewtoolkit.scanner.scanners.Askalono
+import org.ossreviewtoolkit.scanner.scanners.BoyterLc
+import org.ossreviewtoolkit.scanner.scanners.Licensee
 import org.ossreviewtoolkit.scanner.scanners.scancode.ScanCode
 import org.ossreviewtoolkit.scanner.storages.FileBasedStorage
 import org.ossreviewtoolkit.scanner.storages.SCAN_RESULTS_FILE_NAME
@@ -118,6 +132,14 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license 
         help = "Do not scan excluded projects or packages. Works only with the '--ort-file' parameter."
     ).flag()
 
+    private val experimental by option(
+        "--experimental",
+        help = "Use a new experimental implementation of the scanner which scans by provenance instead of by " +
+                "package. This improves reuse of stored scan results and increases performance if multiple packages " +
+                "are coming from the same source code repository. The experimental scanner is work in progress and " +
+                "it is therefore not recommended to use it in production."
+    ).flag()
+
     private val globalOptionsForSubcommands by requireObject<GlobalOptions>()
 
     override fun run() {
@@ -140,6 +162,28 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license 
 
         val config = globalOptionsForSubcommands.config
 
+        val ortResult = if (experimental) {
+            runExperimental(config)
+        } else {
+            run(nativeOutputDir, config)
+        }.mergeLabels(labels)
+
+        // Write the result.
+        outputDir.safeMkdirs()
+        writeOrtResult(ortResult, outputFiles, "scan")
+
+        val scanResults = ortResult.scanner?.results
+
+        if (scanResults == null) {
+            println("There was an error creating the scan results.")
+            throw ProgramResult(1)
+        }
+
+        val counts = scanResults.collectIssues().flatMap { it.value }.groupingBy { it.severity }.eachCount()
+        concludeSeverityStats(counts, config.severeIssueThreshold, 2)
+    }
+
+    private fun run(nativeOutputDir: File, config: OrtConfiguration): OrtResult {
         // Configure the scan storage, which is common to all scanners.
         ScanResultsStorage.configure(config.scanner)
 
@@ -168,7 +212,7 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license 
         } ?: scanner
 
         // Perform the scan.
-        val ortResult = if (input.isFile) {
+        return if (input.isFile) {
             val ortResult = readOrtResult(input)
             scanOrtResult(scanner, projectScanner, ortResult, nativeOutputDir, skipExcluded)
         } else {
@@ -180,20 +224,39 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license 
                 inputPath = input,
                 outputDirectory = nativeOutputDir
             )
-        }.mergeLabels(labels)
+        }
+    }
 
-        // Write the result.
-        outputDir.safeMkdirs()
-        writeOrtResult(ortResult, outputFiles, "scan")
-
-        val scanResults = ortResult.scanner?.results
-
-        if (scanResults == null) {
-            println("There was an error creating the scan results.")
-            throw ProgramResult(1)
+    private fun runExperimental(config: OrtConfiguration): OrtResult {
+        // TODO: The experimental scanner supports using multiple scanner wrappers at once, for now use only one to stay
+        //       compatible with the existing scanner command. Once the experimental flag is removed this command can
+        //       support multiple scanners and use the proper ScannerWrapperFactories.
+        val scannerWrapper = when (scannerFactory.scannerName) {
+            "Askalono" -> Askalono.Factory().create(config.scanner, config.downloader)
+            "BoyterLc" -> BoyterLc.Factory().create(config.scanner, config.downloader)
+            "Licensee" -> Licensee.Factory().create(config.scanner, config.downloader)
+            "ScanCode" -> ScanCode.Factory().create(config.scanner, config.downloader)
+            else -> {
+                throw IllegalArgumentException(
+                    "The scanner ${scannerFactory.scannerName} is not supported by the experimental scanner."
+                )
+            }
         }
 
-        val counts = scanResults.collectIssues().flatMap { it.value }.groupingBy { it.severity }.eachCount()
-        concludeSeverityStats(counts, config.severeIssueThreshold, 2)
+        val scanner = ExperimentalScanner(
+            scannerConfig = config.scanner,
+            downloaderConfig = config.downloader,
+            provenanceDownloader = DefaultProvenanceDownloader(config.downloader),
+            storageReaders = emptyList(),
+            storageWriters = emptyList(),
+            packageProvenanceResolver = DefaultPackageProvenanceResolver(),
+            nestedProvenanceResolver = DefaultNestedProvenanceResolver(),
+            scannerWrappers = listOf(scannerWrapper)
+        )
+
+        val ortResult = readOrtResult(input)
+        return runBlocking {
+            scanner.scan(ortResult)
+        }
     }
 }
--- a/downloader/src/main/kotlin/Downloader.kt
+++ b/downloader/src/main/kotlin/Downloader.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Copyright (C) 2017-2021 HERE Europe B.V.
  * Copyright (C) 2021 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -177,10 +177,16 @@ class Downloader(private val config: DownloaderConfiguration) {
     /**
      * Download the source code of the [package][pkg] to the [outputDirectory] using its VCS information. The
      * [allowMovingRevisions] parameter indicates whether the download accepts symbolic names, like branches, instead of
-     * only fixed revisions. A [Provenance] is returned on success or a [DownloadException] is thrown in case of
-     * failure.
+     * only fixed revisions. If [recursive] is `true`, any nested repositories (like Git submodules or Mercurial
+     * subrepositories) are downloaded, too. A [Provenance] is returned on success or a [DownloadException] is thrown in
+     * case of failure.
      */
-    fun downloadFromVcs(pkg: Package, outputDirectory: File, allowMovingRevisions: Boolean): Provenance {
+    fun downloadFromVcs(
+        pkg: Package,
+        outputDirectory: File,
+        allowMovingRevisions: Boolean,
+        recursive: Boolean = true
+    ): Provenance {
         verifyOutputDirectory(outputDirectory)
 
         log.info {
@@ -240,7 +246,7 @@ class Downloader(private val config: DownloaderConfiguration) {
         }
 
         val workingTree = try {
-            applicableVcs.download(pkg, outputDirectory, allowMovingRevisions)
+            applicableVcs.download(pkg, outputDirectory, allowMovingRevisions, recursive)
         } catch (e: DownloadException) {
             // TODO: We should introduce something like a "strict" mode and only do these kind of fallbacks in
             //       non-strict mode.
@@ -256,7 +262,7 @@ class Downloader(private val config: DownloaderConfiguration) {
                 outputDirectory.safeMkdirs()
 
                 val fallbackPkg = pkg.copy(vcsProcessed = pkg.vcsProcessed.copy(url = vcsUrlNoCredentials))
-                applicableVcs.download(fallbackPkg, outputDirectory, allowMovingRevisions)
+                applicableVcs.download(fallbackPkg, outputDirectory, allowMovingRevisions, recursive)
             } else {
                 throw e
             }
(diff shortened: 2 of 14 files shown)


downloader/src/main/kotlin/Downloader.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Copyright (C) 2017-2021 HERE Europe B.V.
```
sschuberth, line 2: Commit message: s/chose/choose/

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.ossreviewtoolkit.model.ScanSummary
+import org.ossreviewtoolkit.model.ScannerDetails
+import org.ossreviewtoolkit.model.config.ScannerConfiguration
+import org.ossreviewtoolkit.scanner.ScanResultsStorage
+import org.ossreviewtoolkit.scanner.ScannerCriteria
+
+/**
+ * An interface to wrap source code scanners.
+ */
+sealed interface ScannerWrapper {
```
sschuberth, line 40: The `sealed` in here prevents us from adding scanner implementations via plugins as [all implementations of a sealed interface are known at compile time](https://kotlinlang.org/docs/whatsnew15.html#sealed-interfaces). That's a problem as we have concrete plans to add scanner plugins.

mnonnenmacher: Fortunately that's not an issue with a `sealed interface`. In this case you cannot make a class or interface inherit directly from `ScannerWrapper` in another module, but it is allowed to have classes or interfaces inherit from one of the sub-interfaces (like `LocalScannerWrapper`) in another module.

sschuberth: > but it is allowed to have classes or interfaces inherit from one of the sub-interfaces (like `LocalScannerWrapper`) in another module.

Interesting, do you have a reference for that? Or did you actually try it out?

mnonnenmacher: See: https://kotlinlang.org/docs/sealed-classes.html#location-of-direct-subclasses

> Direct subclasses of sealed classes and interfaces must be declared in the same package. [...] These restrictions don't apply to indirect subclasses. If a direct subclass of a sealed class is not marked as sealed, it can be extended in any ways that its modifiers allow.

This makes this a good use case for a `sealed interface`. We ensure that no new kind of scanner can be added by a plugin, because ORT would not know how to handle it, but new implementations of existing kinds can be added. However, while writing this I realized that `interface RemoteScannerWrapper : ScannerWrapper` also needs to be sealed, I will update the PR.

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.ScanResult
+import org.ossreviewtoolkit.model.ScanSummary
+import org.ossreviewtoolkit.model.ScannerDetails
+import org.ossreviewtoolkit.model.config.ScannerConfiguration
+import org.ossreviewtoolkit.scanner.ScanResultsStorage
+import org.ossreviewtoolkit.scanner.ScannerCriteria
+
+/**
+ * An interface to wrap source code scanners.
```
sschuberth, line 38: Nit: I wonder whether we should keep the wording a bit more generic / vague here, as we're planning to implement "fake" scanners that just extract / process files from binary artifacts / archives.

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
```diff
@@ -0,0 +1,86 @@
...
+ */
+sealed interface ScannerWrapper {
+    /**
+     * The name of the scanner.
+     */
+    val name: String
```
sschuberth, line 53: `name` is already part of `ScannerDetails`. Is there really a need to duplicate this here?

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Return a [ScannerCriteria] object to be used when looking up existing scan results from a [ScanResultsStorage].
+     * By default the properties of this object are initialized by the scanner implementation. These default can be
+     * overridden with the [ScannerConfiguration.options] property: Use properties of the form
+     * _scannerName.criteria.property_, where _scannerName_ is the name of the scanner and _property_ is the name of a
+     * property of the [ScannerCriteria] class. For instance, to specify that a specific minimum version of ScanCode is
+     * allowed, set this property: `options.ScanCode.criteria.minScannerVersion=3.0.2`.
+     */
+    fun getScannerCriteria(): ScannerCriteria
```
sschuberth, line 68: I'm probably missing something, but why is this not also a property, like the ones above?

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
```diff
@@ -0,0 +1,86 @@
...
+}
+
+/**
+ * A wrapper interface for remote scanners. A remote scanner is a scanner that is not executed locally but is running on
+ * a different machine and can be accessed by a remote interface.
```
sschuberth, line 73: Nit: Maybe "by" -> "via"?

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -0,0 +1,51 @@
...
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.scanner.experimental
+
+import org.ossreviewtoolkit.model.KnownProvenance
+import org.ossreviewtoolkit.model.ScanResult
+
+/**
+ * A class that contain all [ScanResult]s for a [NestedProvenance].
```
sschuberth, line 34: contain*s*

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * A map of [KnownProvenance]s from [nestedProvenance] associated with lists of [ScanResult]s.
+     */
+    val scanResults: Map<KnownProvenance, List<ScanResult>>,
+) {
+    /**
+     * Return a set of all [KnownProvenance]s contained in [nestedProvenance].
+     */
+    fun getProvenances(): Set<KnownProvenance> =
+        setOf(nestedProvenance.root, *nestedProvenance.subRepositories.values.toTypedArray())
```
sschuberth, line 51: I'm not sure about the performance of  `*` / `toTypedArray()`  here, even if there should not be a lot of entries. Maybe better `mutableSetOf(nestedProvenance.root) + nestedProvenance.subRepositories.values` (the returned set would still be seen as immutable)?

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Return true if [scanResults] contains at least on scan result for each of the [KnownProvenance]s contained in
```
sschuberth, line 54: on*e*

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -0,0 +1,51 @@
...
+     * [nestedProvenance].
+     */
+    fun isComplete(): Boolean = getProvenances().all { scanResults[it]?.isNotEmpty() == true }
```
sschuberth, line 57: Nit: Maybe inverse the logic to `getProvenances().none { scanResults[it]?.isEmpty() == true }`?

sschuberth (changes requested):
I stopped reviewing at the "scanner: Add a new experimental implementation" commit for now. Will continue later.

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
mnonnenmacher: So just scratch "source code"?

sschuberth: Maybe. I'm also struggling a bit with the use of "wrap" in here, as IMO it's not the interface itself that wraps the scanner, but the implementation (eventually). You *could* also implement the `ScannerWrapper` interface with a "real" scanner written in Kotlin, that does not "wrap" any CLI / remote scanner.

So how about "The base interface for all types of scanners"?

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
mnonnenmacher: It's not strongly required, but `details` depends on `getVersion()` which could be slow, e.g. if it depends on calling the CLI tool or a network call in the case of FossId.

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
mnonnenmacher: Mainly because the existing implementations already have this function, I thought that's something to better clean up once we reach a state were the old implementation can be removed.

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
mnonnenmacher: I think the performance impact is negligible. However, while the proposed solution avoids transforming the sub repositories twice, it still creates an unnecessary set, because the `+` operator always creates a new set and doesn't add to the existing mutable set. I went with this, because it creates only one set:
```kotlin
    fun getProvenances(): Set<KnownProvenance> =
        nestedProvenance.subRepositories.values.toMutableSet<KnownProvenance>().also { it.add(nestedProvenance.root) }
```

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
mnonnenmacher: That would be wrong in the case that `scanResults[it] == null` so we would have to use `!= false` in which case I prefer the existing code.

fviernau:
The PR description says "[..] > to scan each provenance (meaning source code repository or source artifact) only once [..]". Could you update the PR description to explain the idea how this is achieved for an overview?

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
sschuberth: Could this say `it +=nestedProvenance.root`?

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
sschuberth: So, are there use-cases where you want to get the scanner name (fast), and don't need the other scanner details?

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
sschuberth: I believe the new scanner implementation is so different already anyway that we don't have to be considerate of eventual limitations of the existing implementations. My fear simply is that we forget about this if we don't do it now.

mnonnenmacher:
> 
> 
> The PR description says "[..] > to scan each provenance (meaning source code repository or source artifact) only once [..]". Could you update the PR description to explain the idea how this is achieved for an overview?

I have added a simplified description of the algorithm, I hope that's what you were looking for.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -0,0 +1,406 @@
...
+private fun <K, V> Map<K, V?>.filterNotNullValues(): Map<K, V> = filterValues { it != null } as Map<K, V>
+
+/**
+ * Split this [ScanResult] into separate results for each [KnownProvenance] contained in the [nestedProvenance] by
+ * matching the paths of findings with the paths in the source tree.
+ */
+fun ScanResult.toNestedProvenanceScanResult(nestedProvenance: NestedProvenance): NestedProvenanceScanResult {
+    val provenanceByPath = listOf(
+        "" to nestedProvenance.root,
+        *nestedProvenance.subRepositories.entries.map { it.key to it.value }.toTypedArray()
```
sschuberth, line 374: What's the purpose of just mapping keys to values here, isn't this a no-op?

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -0,0 +1,406 @@
...
+
+    val copyrightFindingsByProvenance = summary.copyrightFindings.groupBy { copyrightFinding ->
+        provenanceByPath.first { copyrightFinding.location.path.startsWith(it.first) }.second
+    }
+
+    val licenseFindingsByProvenance = summary.licenseFindings.groupBy { licenseFinding ->
+        provenanceByPath.first { licenseFinding.location.path.startsWith(it.first) }.second
+    }
+
+    val provenances = setOf(nestedProvenance.root, *nestedProvenance.subRepositories.values.toTypedArray())
```
sschuberth, line 385: Can we apply the same optimization as before here?

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -48,4 +55,77 @@ data class NestedProvenanceScanResult(
...
+
+        return allScanners.map { scanner ->
+            val scanResultsForScanner = scanResults.mapValues { (_, results) ->
+                results.filter { it.scanner == scanner }
+            }
+
+            val allScanResults = scanResults.values.flatten()
+
+            val startTime = allScanResults.minByOrNull { it.summary.startTime }?.summary?.startTime ?: Instant.now()
+            val endTime = allScanResults.maxByOrNull { it.summary.endTime }?.summary?.endTime ?: Instant.now()
```
sschuberth, line 77: Nit: Shouldn't these use exactly the same Instant instance as a fallback?

cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -118,6 +132,14 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license
         help = "Do not scan excluded projects or packages. Works only with the '--ort-file' parameter."
     ).flag()
 
+    private val experimental by option(
+        "--experimental",
+        help = "Use a new experimental implementation of the scanner which scans by provenance instead of by " +
+                "package. This improves reuse of stored scan results and increases performance if multiple packages " +
+                "come from the same source code repository. The experimental scanner is work in progress and it is " +
```
sschuberth, line 139: s/come/coming/

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -68,7 +68,6 @@ class ExperimentalScanner(
         )
 
         val endTime = Instant.now()
-
```
sschuberth: Unintended change in this commit?

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
```diff
@@ -36,6 +38,15 @@ import org.ossreviewtoolkit.scanner.ScannerCriteria
  * The base interface for all types of scanners.
  */
 sealed interface ScannerWrapper {
+    companion object {
+        private val LOADER = ServiceLoader.load(ScannerWrapperFactory::class.java)!!
+
+        /**
+         * The list of all available scanners in the classpath.
+         */
+        val ALL by lazy { LOADER.iterator().asSequence().toList() }
```
sschuberth, line 47: To catchup with recent changes, this should now have `.sortedBy { it.scannerName }`.

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
mnonnenmacher: I'm not sure, the name is used in several log statements, for example, but on the other hand the `details` should be built only once. Shall I remove `name` for now and if we find any performance issues we re-introduce it?

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
mnonnenmacher: I have made it a property.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -0,0 +1,406 @@
...
+    suspend fun scan(ortResult: OrtResult): OrtResult {
+        val startTime = Instant.now()
+
+        val result = scan(
+            (ortResult.getProjects().map { it.toPackage() } + ortResult.getPackages().map { it.pkg }).toSet()
+        )
+
+        val endTime = Instant.now()
+
+        val scanResults = result.mapKeys { it.key.id }.mapValues { it.value.merge() }.toSortedMap()
```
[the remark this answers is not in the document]
mnonnenmacher: I'm using `associateTo(sortedMapOf())` now.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -0,0 +1,406 @@
...
+        }.filterValues { it.isNotEmpty() }
+
+    private fun createNestedProvenanceScanResults(
+        packages: Set<Package>,
+        nestedProvenances: Map<Package, NestedProvenance>,
+        scanResults: Map<ScannerWrapper, Map<KnownProvenance, List<ScanResult>>>
+    ): Map<Package, NestedProvenanceScanResult> =
+        packages.associateWith { pkg ->
+            val nestedProvenance = nestedProvenances.getValue(pkg)
+            val provenances = setOf(nestedProvenance.root, *nestedProvenance.subRepositories.values.toTypedArray())
```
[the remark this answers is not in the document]
mnonnenmacher: Yes, and I have moved it to the `NestedProvenance` class to reduce repetition.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: It was to transform entires to pairs, but I have replaced it with `toList()` now which does the same.

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
mnonnenmacher: I am using `startTime` as a fallback now, that should achieve this.

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
```diff
@@ -0,0 +1,103 @@
...
+ * The base interface for all types of scanners.
+ */
+sealed interface ScannerWrapper {
+    companion object {
+        private val LOADER = ServiceLoader.load(ScannerWrapperFactory::class.java)!!
+
+        /**
+         * The list of all available scanners in the classpath.
+         */
+        val ALL by lazy { LOADER.iterator().asSequence().toList().sortedBy { it.javaClass.simpleName } }
```
sschuberth, line 46: As discussed, probably drop the factories completely for now for the new scanner classes.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -0,0 +1,402 @@
...
+
+/**
+ * Split this [ScanResult] into separate results for each [KnownProvenance] contained in the [nestedProvenance] by
+ * matching the paths of findings with the paths in the source tree.
+ */
+fun ScanResult.toNestedProvenanceScanResult(nestedProvenance: NestedProvenance): NestedProvenanceScanResult {
+    val provenanceByPath = listOf(
+        "" to nestedProvenance.root,
+        *nestedProvenance.subRepositories.toList().toTypedArray()
+    ).sortedByDescending { it.first.length }
```
sschuberth, line 372: As discussed, please double-check whether this is still needed and / or can be simplified.

mnonnenmacher: It's still needed, but I have applied the same optimization as in other places before.

sschuberth (approved):

PR #4493, opened 2021-09-23
Add support for defects to ORT's data model

oheger-bosch:
This PR adds a `Defect` class to the data model that can be used in a similar way as the existing `Vulnerability` class. The goal is to support advisor implementations that can retrieve information about known defects for the packages involved. To make this possible, the result classes used by advisors have been extended to store defects as well.

To reflect this broader scope advisors can now have, some classes have been renamed to no longer refer to *vulnerabilities* exclusively.

### DIFF
--- a/advisor/src/main/kotlin/AdviceProvider.kt
+++ b/advisor/src/main/kotlin/AdviceProvider.kt
@@ -31,22 +31,22 @@ import org.ossreviewtoolkit.utils.collectMessagesAsString
 import org.ossreviewtoolkit.utils.showStackTrace
 
 /**
- * An abstract class that represents a service that can retrieve vulnerability information
- * for a list of given [Package]s.
+ * An abstract class that represents a service that can retrieve any kind of advice information
+ * for a list of given [Package]s. Examples of such information can be security vulnerabilities, known defects,
+ * or code analysis results.
  */
-abstract class VulnerabilityProvider(val providerName: String) {
+abstract class AdviceProvider(val providerName: String) {
     /**
-     * For a given list of [Package]s, retrieve vulnerability information and return a map
-     * that associates each package with a list of [AdvisorResult]s. Needs to be implemented
-     * by child classes.
+     * For a given list of [Package]s, retrieve  findings and return a map that associates each package with a list
+     * of [AdvisorResult]s. Needs to be implemented by child classes.
      */
-    abstract suspend fun retrievePackageVulnerabilities(
+    abstract suspend fun retrievePackageFindings(
         packages: List<Package>
     ): Map<Package, List<AdvisorResult>>
 
     /**
      * A generic method that creates a failed [AdvisorResult] for [Package]s if there was an issue
-     * during the retrieval of vulnerability information.
+     * constructing the provider-specific information.
      */
     protected fun createFailedResults(
         startTime: Instant,
@@ -67,7 +67,7 @@ abstract class VulnerabilityProvider(val providerName: String) {
                     issues = listOf(
                         createAndLogIssue(
                             source = providerName,
-                            message = "Failed to retrieve security vulnerabilities from $providerName: " +
+                            message = "Failed to retrieve findings from $providerName: " +
                                     t.collectMessagesAsString()
                         )
                     )
--- a/advisor/src/main/kotlin/Advisor.kt
+++ b/advisor/src/main/kotlin/Advisor.kt
@@ -36,22 +36,23 @@ import org.ossreviewtoolkit.utils.Environment
 import org.ossreviewtoolkit.utils.log
 
 /**
- * The class to manage [VulnerabilityProvider]s that retrieve security advisories.
+ * The class to manage [AdviceProvider]s. It invokes the configured providers and adds their findings to the current
+ * [OrtResult].
  */
 class Advisor(
-    private val providerFactories: List<VulnerabilityProviderFactory>,
+    private val providerFactories: List<AdviceProviderFactory>,
     private val config: AdvisorConfiguration
 ) {
     companion object {
-        private val LOADER = ServiceLoader.load(VulnerabilityProviderFactory::class.java)!!
+        private val LOADER = ServiceLoader.load(AdviceProviderFactory::class.java)!!
 
         /**
-         * The list of all available [VulnerabilityProvider]s in the classpath.
+         * The list of all available [AdviceProvider]s in the classpath.
          */
         val ALL by lazy { LOADER.iterator().asSequence().toList().sortedBy { it.providerName } }
     }
 
-    fun retrieveVulnerabilityInformation(ortResult: OrtResult, skipExcluded: Boolean = false): OrtResult {
+    fun retrieveFindings(ortResult: OrtResult, skipExcluded: Boolean = false): OrtResult {
         val startTime = Instant.now()
 
         if (ortResult.analyzer == null) {
@@ -72,7 +73,7 @@ class Advisor(
         runBlocking {
             providers.map { provider ->
                 async {
-                    provider.retrievePackageVulnerabilities(packages)
+                    provider.retrievePackageFindings(packages)
                 }
             }.forEach { providerResults ->
                 providerResults.await().forEach { (pkg, advisorResults) ->
--- a/advisor/src/main/kotlin/advisors/NexusIq.kt
+++ b/advisor/src/main/kotlin/advisors/NexusIq.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Bosch.IO GmbH
+ * Copyright (C) 2020-2021 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -97,9 +97,9 @@ class NexusIq(name: String, private val nexusIqConfig: NexusIqConfiguration) : V
                 componentDetails[pkg.id.toPurl()]?.let { details ->
                     pkg to listOf(
                         AdvisorResult(
-                            details.securityData.securityIssues.map { it.toVulnerability() },
                             AdvisorDetails(providerName),
-                            AdvisorSummary(startTime, endTime)
+                            AdvisorSummary(startTime, endTime),
+                            vulnerabilities = details.securityData.securityIssues.map { it.toVulnerability() }
                         )
                     )
                 }
--- a/advisor/src/main/kotlin/advisors/NexusIq.kt
+++ b/advisor/src/main/kotlin/advisors/NexusIq.kt
@@ -23,8 +23,8 @@ import java.io.IOException
 import java.net.URI
 import java.time.Instant
 
-import org.ossreviewtoolkit.advisor.AbstractVulnerabilityProviderFactory
-import org.ossreviewtoolkit.advisor.VulnerabilityProvider
+import org.ossreviewtoolkit.advisor.AbstractAdviceProviderFactory
+import org.ossreviewtoolkit.advisor.AdviceProvider
 import org.ossreviewtoolkit.clients.nexusiq.NexusIqService
 import org.ossreviewtoolkit.model.AdvisorDetails
 import org.ossreviewtoolkit.model.AdvisorResult
@@ -50,8 +50,8 @@ private const val REQUEST_CHUNK_SIZE = 128
 /**
  * A wrapper for [Nexus IQ Server](https://help.sonatype.com/iqserver) security vulnerability data.
  */
-class NexusIq(name: String, private val nexusIqConfig: NexusIqConfiguration) : VulnerabilityProvider(name) {
-    class Factory : AbstractVulnerabilityProviderFactory<NexusIq>("NexusIQ") {
+class NexusIq(name: String, private val nexusIqConfig: NexusIqConfiguration) : AdviceProvider(name) {
+    class Factory : AbstractAdviceProviderFactory<NexusIq>("NexusIQ") {
         override fun create(config: AdvisorConfiguration) = NexusIq(providerName, config.forProvider { nexusIq })
     }
 
@@ -64,7 +64,7 @@ class NexusIq(name: String, private val nexusIqConfig: NexusIqConfiguration) : V
         )
     }
 
-    override suspend fun retrievePackageVulnerabilities(packages: List<Package>): Map<Package, List<AdvisorResult>> {
+    override suspend fun retrievePackageFindings(packages: List<Package>): Map<Package, List<AdvisorResult>> {
         val startTime = Instant.now()
 
         val components = packages.map { pkg ->
--- a/advisor/src/main/kotlin/advisors/VulnerableCode.kt
+++ b/advisor/src/main/kotlin/advisors/VulnerableCode.kt
@@ -22,8 +22,8 @@ package org.ossreviewtoolkit.advisor.advisors
 import java.net.URI
 import java.time.Instant
 
-import org.ossreviewtoolkit.advisor.AbstractVulnerabilityProviderFactory
-import org.ossreviewtoolkit.advisor.VulnerabilityProvider
+import org.ossreviewtoolkit.advisor.AbstractAdviceProviderFactory
+import org.ossreviewtoolkit.advisor.AdviceProvider
 import org.ossreviewtoolkit.clients.vulnerablecode.VulnerableCodeService
 import org.ossreviewtoolkit.clients.vulnerablecode.VulnerableCodeService.PackagesWrapper
 import org.ossreviewtoolkit.model.AdvisorDetails
@@ -37,14 +37,14 @@ import org.ossreviewtoolkit.model.config.VulnerableCodeConfiguration
 import org.ossreviewtoolkit.utils.OkHttpClientHelper
 
 /**
- * A [VulnerabilityProvider] implementation that obtains security vulnerability information from a
+ * An [AdviceProvider] implementation that obtains security vulnerability information from a
  * [VulnerableCode][https://github.com/nexB/vulnerablecode] instance.
  */
 class VulnerableCode(
     name: String,
     private val vulnerableCodeConfiguration: VulnerableCodeConfiguration
-) : VulnerabilityProvider(name) {
-    class Factory : AbstractVulnerabilityProviderFactory<VulnerableCode>("VulnerableCode") {
+) : AdviceProvider(name) {
+    class Factory : AbstractAdviceProviderFactory<VulnerableCode>("VulnerableCode") {
         override fun create(config: AdvisorConfiguration) =
             VulnerableCode(providerName, config.forProvider { vulnerableCode })
     }
@@ -67,7 +67,7 @@ class VulnerableCode(
         VulnerableCodeService.create(vulnerableCodeConfiguration.serverUrl, OkHttpClientHelper.buildClient())
     }
 
-    override suspend fun retrievePackageVulnerabilities(packages: List<Package>): Map<Package, List<AdvisorResult>> {
+    override suspend fun retrievePackageFindings(packages: List<Package>): Map<Package, List<AdvisorResult>> {
         val startTime = Instant.now()
 
         @Suppress("TooGenericExceptionCaught")
--- a/advisor/src/main/kotlin/advisors/VulnerableCode.kt
+++ b/advisor/src/main/kotlin/advisors/VulnerableCode.kt
@@ -97,7 +97,7 @@ class VulnerableCode(
             packageMap[pv.purl]?.let { pkg ->
                 val vulnerabilities = pv.unresolvedVulnerabilities.map { it.toModel() }
                 val summary = AdvisorSummary(startTime, Instant.now())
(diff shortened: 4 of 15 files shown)


advisor/src/main/kotlin/advisors/NexusIq.kt
```diff
@@ -97,9 +97,9 @@ class NexusIq(name: String, private val nexusIqConfig: NexusIqConfiguration) : V
                 componentDetails[pkg.id.toPurl()]?.let { details ->
                     pkg to listOf(
                         AdvisorResult(
-                            details.securityData.securityIssues.mapNotNull { it.toVulnerability() },
                             AdvisorDetails(providerName),
-                            AdvisorSummary(startTime, endTime)
+                            AdvisorSummary(startTime, endTime),
+                            details.securityData.securityIssues.mapNotNull { it.toVulnerability() }
```
mnonnenmacher, line 102: Please use a named parameter for this row.

model/src/main/kotlin/AdvisorRecord.kt
```diff
@@ -67,7 +67,21 @@ data class AdvisorRecord(
      * from different advisors are merged if necessary.
      */
     fun getVulnerabilities(pkgId: Identifier): List<Vulnerability> =
-        advisorResults[pkgId].orEmpty().flatMap { it.vulnerabilities }.mergeVulnerabilities()
+        getFindings(pkgId) { it.vulnerabilities }.mergeVulnerabilities()
+
+    /**
+     * Return a list with all [Defect] objects that have been found for the given [package][pkgId]. If there are
+     * results from different advisors, a union list is constructed. (No merging is done, as it is expected that the
+     * results from different advisors cannot be combined.)
```
mnonnenmacher, line 75: I think you could just remove the parenthesis.

model/src/main/kotlin/AdvisorResult.kt
```diff
@@ -36,5 +34,15 @@ data class AdvisorResult(
...
+
+    /**
+     * The found vulnerabilities.
+     */
+    val vulnerabilities: List<Vulnerability> = emptyList(),
+
+    /**
+     * The found defects.
+     */
+    val defects: List<Defect> = emptyList()
```
mnonnenmacher, line 47: What about moving `defects` above `vulnerabilities` for alphabetic order?

advisor/src/main/kotlin/Advisor.kt
```diff
@@ -36,22 +36,23 @@ import org.ossreviewtoolkit.utils.Environment
 import org.ossreviewtoolkit.utils.log
 
 /**
- * The class to manage [VulnerabilityProvider]s that retrieve security advisories.
+ * The class to manage [AdviceProvider]s. It invokes the configured providers and adds their finding to the current
```
mnonnenmacher, line 39: Typo: "finding**s**"

mnonnenmacher (changes requested):

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+     * A state of the associated defect. The concrete meaning of this string depends on the source from where it was
+     * obtained, as different issue tracker systems use their specific terminology. Possible values could be *OPEN*,
+     * *IN PROGRESS*, *BLOCKED*, etc.
+     */
+    val state: String = UNKNOWN_STATE,
+
+    /**
+     * The severity assigned to the defect. The meaning of this string depends on the source system.
+     */
+    val severity: String = UNKNOWN_SEVERITY,
```
fviernau, line 62: What's the reason for not using ORTs severity? E.g. exposing the severity definition of the provider adds complexity of the code using this severity and makes the contract a bit imprecise.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+import java.net.URI
+import java.time.Instant
+
+/**
+ * A data class defining a model for software defects.
+ *
+ * Instances of this class are created by advisor implementations that retrieve information about known defects in
+ * packages.
+ */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
```
fviernau, line 33: Putting include non default makes it easy to overlook it when reading any of the properties.
If a code change adjusts the default value of any property, that is not backwards compatible.
So putting this annotation to each property would be a bit less risky IMO.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+
+package org.ossreviewtoolkit.model
+
+import com.fasterxml.jackson.annotation.JsonInclude
+
+import java.net.URI
+import java.time.Instant
+
+/**
+ * A data class defining a model for software defects.
```
fviernau, line 28: shorter/less details: A model for software defects.
or: A data model for software defects.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+     */
+    val description: String? = null,
+
+    /**
+     * The creation date of this defect if available.
+     */
+    val createdAt: Instant? = null,
+
+    /**
+     * Contains a date of last modification of this defect if available. This information can be useful for instance
```
fviernau, line 77: to me it's unclear what is a "modification of this defect" exactly?

advisor/src/main/kotlin/AdviceProvider.kt
```diff
@@ -31,22 +31,23 @@ import org.ossreviewtoolkit.utils.collectMessagesAsString
...
      */
-    abstract suspend fun retrievePackageVulnerabilities(
+    abstract suspend fun retrievePackageFindings(
         packages: List<Package>
     ): Map<Package, List<AdvisorResult>>
 
     /**
      * A generic method that creates a failed [AdvisorResult] for [Package]s if there was an issue
-     * during the retrieval of vulnerability information.
+     * during constructing of the provider-specific information.
```
fviernau, line 50: maybe just: an issue constructing the provider-specific information

advisor/src/main/kotlin/AdviceProvider.kt
```diff
@@ -31,22 +31,23 @@ import org.ossreviewtoolkit.utils.collectMessagesAsString
...
- * for a list of given [Package]s.
+ * An abstract class that represents a service that can retrieve any kind of advice information
+ * for a list of given [Package]s. Examples of such information can be security vulnerabilities, known defects,
+ * or results of some code analysis.
  */
-abstract class VulnerabilityProvider(val providerName: String) {
+abstract class AdviceProvider(val providerName: String) {
     /**
-     * For a given list of [Package]s, retrieve vulnerability information and return a map
+     * For a given list of [Package]s, retrieve the provider-specific findings and return a map
```
fviernau, line 40: Is "provider-specific" redundant, e.g. implied from the class context?

advisor/src/main/kotlin/AdviceProvider.kt
```diff
@@ -31,22 +31,23 @@ import org.ossreviewtoolkit.utils.collectMessagesAsString
 import org.ossreviewtoolkit.utils.showStackTrace
 
 /**
- * An abstract class that represents a service that can retrieve vulnerability information
- * for a list of given [Package]s.
+ * An abstract class that represents a service that can retrieve any kind of advice information
+ * for a list of given [Package]s. Examples of such information can be security vulnerabilities, known defects,
+ * or results of some code analysis.
```
fviernau, line 36: maybe: results of some code analysis -> code analysis results.

advisor/src/main/kotlin/advisors/VulnerableCode.kt
```diff
@@ -37,14 +37,14 @@ import org.ossreviewtoolkit.model.config.VulnerableCodeConfiguration
 import org.ossreviewtoolkit.utils.OkHttpClientHelper
 
 /**
- * A [VulnerabilityProvider] implementation that obtains security vulnerability information from a
+ * A [AdviceProvider] implementation that obtains security vulnerability information from a
```
fviernau, line 40: An

fviernau (changes requested):

model/src/test/kotlin/AdvisorRecordTest.kt
```diff
@@ -269,13 +296,20 @@ private fun createVulnerability(
...
+
+/**
+ * Create an [AdvisorResult] for an advisor with the given [advisorIndex] with the passed in [issues],
+ * [vulnerabilities], and [defects].
  */
 private fun createResult(
     advisorIndex: Int,
     issues: List<OrtIssue> = emptyList(),
-    vulnerabilities: List<Vulnerability> = emptyList()
+    vulnerabilities: List<Vulnerability> = emptyList(),
```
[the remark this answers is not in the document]
oheger-bosch: These parameters are passed through to other objects that require lists. I don't think, it is worth the effort in the test to declare them as collection and later convert them explicitly to lists.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+ * Instances of this class are created by advisor implementations that retrieve information about known defects in
+ * packages.
+ */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+data class Defect(
+    /**
+     * The (external) ID of this defect. This is a string used by a concrete issue tracker system to reference this
+     * defect, such as a bug ID or ticket number.
+     */
+    val id: String,
```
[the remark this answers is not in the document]
oheger-bosch: To uniquely identify defects, the `url` property is probably the correct one. It also references the source. The _id_ property alone is not appropriate for this purpose, as different tracker systems may use clashing IDs. So maybe _id_ is not a good name; perhaps better use `trackerId`? And mention the possibility of conflicting IDs in the comment?

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * The URL pointing to the source of this defect. This is typically a reference into the issue tracker system that
+     * contains this defect.
+     */
+    val url: URI,
```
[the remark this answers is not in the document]
oheger-bosch: IMHO, the URL should not be nullable, because - as mentioned before - it also allows tracking the source of the defect. A non-nullable URL is also in-line with the vulnerabilities model: here the URLs of references are mandatory, too.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+     * A title for this defect. This is a short summary describing the problem at hand.
+     */
+    val title: String,
+
+    /**
+     * A state of the associated defect. The concrete meaning of this string depends on the source from where it was
+     * obtained, as different issue tracker systems use their specific terminology. Possible values could be *OPEN*,
+     * *IN PROGRESS*, *BLOCKED*, etc.
+     */
+    val state: String = UNKNOWN_STATE,
```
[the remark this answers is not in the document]
oheger-bosch: My thinking here was that having non-null fields with a defined value for UNKNOWN would maybe slightly simplify client code because null checks are not needed. But if you prefer, nullable fields would be fine for me, too.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+     * An optional description of this defect. It can contain more detailed information about the defect and its
+     * impact. The field may be undefined if the [url] of this defect already points to a website with all this
+     * information.
+     */
+    val description: String? = null,
+
+    /**
+     * The creation date of this defect if available.
+     */
+    val createdAt: Instant? = null,
```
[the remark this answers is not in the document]
oheger-bosch: I had the idea that these timestamps could be an indicator how likely it is that a bug gets fixed in the near future. If it has been open for ages and exists for multiple releases, you probably cannot expect a solution soon. Also, these properties should be normally available in typical bug tracking systems, so it should be easy to provide them in a concrete advisor implementation.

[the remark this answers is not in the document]
sschuberth: Also the "time to fix", i.e. the span between created an resolved, could give an impression about how well the project is maintained / how active it is, and as a result play a role in eventually getting rid of it. That said, I'd also like to track a "resolvedAt" time.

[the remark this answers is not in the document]
oheger-bosch: This is of course an interesting measure. It is, however, not clear to me whether an advisor for defects would/should retrieve defects that are already resolved; they are no longer problematic.

But that said, there may be tracker systems, for which it is difficult to map the resolved date of a defect against a specific release of a package. For instance, in GitHub issues, AFAICT there is no "fix version" property; so an implementation would probably need to check release dates to find out whether a concrete package version is affected by a defect.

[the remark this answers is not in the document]
sschuberth: > it is difficult to map the resolved date of a defect against a specific release of a package.

Correct. That's why I was really only thinking about the "time to fix" (how long did it take to address the issue, as a community health metric), not about finding out which release actually comes with the fix.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * The URL pointing to the source of this defect. This is typically a reference into the issue tracker system that
+     * contains this defect.
+     */
+    val url: URI,
+
+    /**
+     * A title for this defect. This is a short summary describing the problem at hand.
+     */
+    val title: String,
```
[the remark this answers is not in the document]
oheger-bosch: Made it nullable.

model/src/main/kotlin/Defect.kt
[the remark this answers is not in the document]
oheger-bosch: Made the properties nullable and removed the default values.

model/src/main/kotlin/Defect.kt
oheger-bosch: If we used ORT's severity, there would be the problem of mapping a terminology used by a specific issue tracker to the values defined by ORT. This is not trivial, and may sometimes not even possible.

For vulnerabilities, we did not introduce an ORT-specific severity either, because we would then have to deal with all the different scoring systems for vulnerabilities. It is true that this makes it harder for client code to interpret these values; however, I doubt that ORT could provide a solution that covers all cases.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,100 @@
...
+         * Constant for a state whose value is not known. This string is used as value for the [state] property if the
+         * source system does not provide any state information.
+         */
+        const val UNKNOWN_STATE = "UNKNOWN_STATE"
+
+        /**
+         * Constant for a severity value to be used if the source system does not provide any severity information.
+         */
+        const val UNKNOWN_SEVERITY = "UNKNOWN_SEVERITY"
+    }
```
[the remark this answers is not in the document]
oheger-bosch: Added a check for the `id` field, which is the only remaining non-nullable string value.

model/src/main/kotlin/Defect.kt
oheger-bosch: Do you think this is an issue here? All default values are *null* or an empty collection. I doubt that these will change.

model/src/main/kotlin/Defect.kt
oheger-bosch: Reworded the comment.

model/src/main/kotlin/AdvisorRecord.kt
```diff
@@ -67,7 +67,21 @@ data class AdvisorRecord(
      * from different advisors are merged if necessary.
      */
     fun getVulnerabilities(pkgId: Identifier): List<Vulnerability> =
-        advisorResults[pkgId].orEmpty().flatMap { it.vulnerabilities }.mergeVulnerabilities()
+        getFindings(pkgId) { it.vulnerabilities }.mergeVulnerabilities()
+
+    /**
```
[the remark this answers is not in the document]
oheger-bosch: Which docs exactly do you propose to enhance?

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * The creation date of this defect if available.
+     */
+    val createdAt: Instant? = null,
+
+    /**
+     * Contains a date when this defect has been modified the last time in the tracker system it has been obtained
+     * from. This information can be useful for instance to find out how up-to-date this defect report might be.
+     */
+    val modifiedAt: Instant? = null,
```
sschuberth, line 80: Maybe `modifiedOn` (and `createdOn` above)? Or `modificationTime` and `creationTime`?

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * A set with labels assigned to this defect. Labels provide a means frequently used by issue tracker systems to
+     * classify defects based on defined criteria. The exact meaning of these labels is depending on the source system.
+     */
+    val labels: Set<String> = emptySet()
```
sschuberth, line 86: How about making this `val labels: Map<String, String> = emptyMap()` like in `OrtResult`, so we could also use it more easily for things like relations between defects, where the relationship type could be the key in the map, and the target URL the value?

model/src/main/kotlin/Defect.kt
oheger-bosch: Renamed to `creationTime` and `modificationTime`. BTW, the original names were taken from GitHub's issues; so I would argue that they could not be completely wrong.

model/src/main/kotlin/Defect.kt
oheger-bosch: Done.

model/src/main/kotlin/Defect.kt
sschuberth: > BTW, the original names were taken from GitHub's issues; so I would argue that they could not be completely wrong.

Whether "at" or "on" is correct [depends on](https://english.stackexchange.com/questions/182660/on-vs-at-with-date-and-time) whether you refer to the date (on) or time (at). So while not necessarily completely wrong, it's at least a bit ambiguous. I find you names new better :-)

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,91 @@
...
+    val modificationTime: Instant? = null,
+
+    /**
+     * A map with labels assigned to this defect. Labels provide a means frequently used by issue tracker systems to
+     * classify defects based on defined criteria. The exact meaning of these labels is depending on the source system.
+     */
+    val labels: Map<String, String> = emptyMap()
+) {
+    init {
+        require(id.isNotBlank()) { "Defect ID must not be empty." }
```
sschuberth, line 89: Nit: As you're testing for "blank", not for "empty", the message should also say "blank".

sschuberth (commented):
Just one minor nit, otherwise looks good to me now. But I'd also like others to approve.

oheger-bosch:
@fviernau I think I have addressed most of your comments. Could you please have a look what is still open?

model/src/main/kotlin/Defect.kt
fviernau: Agree. (Some string default values were different when I wrote this comment).

model/src/main/kotlin/Defect.kt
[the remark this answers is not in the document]
fviernau: I see. I thought there may be cases where there is no URL or where one does not want to share it because with the outside world because its internal. Anyhow, let's leave it as is. Can be changed later if needed.

model/src/main/kotlin/AdvisorRecord.kt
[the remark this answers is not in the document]
fviernau: The function docs where this comment is at, line 72-76. However, I guess form
the docs I get that the `ids` of the returned defects are not unique. If that's the case then I think it's better to keep it as-is.

model/src/main/kotlin/Defect.kt
```diff
@@ -0,0 +1,91 @@
...
+     * A state of the associated defect if available. The concrete meaning of this string depends on the source from
+     * where it was obtained, as different issue tracker systems use their specific terminology. Possible values could
+     * be *OPEN*, *IN PROGRESS*, *BLOCKED*, etc.
+     */
+    val state: String? = null,
+
+    /**
+     * The severity assigned to the defect if available. The meaning of this string depends on the source system.
+     */
+    val severity: String? = null,
```
fviernau, line 62: I cannot find my previous comment on this anymore. Using String is quite generic. 
This implies to me that there are no plans to use the severity in ORT. Is this the case?

fviernau (approved):

model/src/main/kotlin/Defect.kt
oheger-bosch: > 
> 
> I cannot find my previous comment on this anymore. Using String is quite generic.
> This implies to me that there are no plans to use the severity in ORT. Is this the case?

Yes, it is currently not planned to reuse ORT's severity, as this would imply some mapping logic, which might become rather complex. In the vulnerabilities model, there is also currently no unification of the several severity scores; we just provide the severity as a string and the scoring system it belongs to; I think, the situation is similar here.

mnonnenmacher (approved):

PR #4494, opened 2021-09-23
Support OSS Index as a vulnerability provider

sschuberth:

### DIFF
--- a/README.md
+++ b/README.md
@@ -613,6 +613,13 @@ ort {
 
 To enable this provider, pass `-a NexusIQ` on the command line.
 
+## OSS Index
+
+This vulnerability provider does not require any further configuration as it uses the public service at
+https://ossindex.sonatype.org/.
+
+To enable this provider, pass `-a OssIndex` on the command line.
+
 ## VulnerableCode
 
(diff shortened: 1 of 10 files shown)


oheger-bosch:
Note that there are conflicting parts with #4493.

sschuberth:
> Note that there are conflicting parts with #4493.

Yes, thanks for pointing that out. If not too much work for you, I'd prefer to get this merged first, as it had been pending for a long time, and could serve as another implementation to double-check the new model.

oheger-bosch (approved):
I would love to see a (Wiremock-) test for this new advisor implementation. This would also include sample responses from the service, helping to better understand the model used behind the scenes.

sschuberth:
> I would love to see a (Wiremock-) test for this new advisor implementation.

Can we do this together quickly next week as a follow-up?

sschuberth:
> > I would love to see a (Wiremock-) test for this new advisor implementation.
> 
> Can we do this together quickly next week as a follow-up?

Done in https://github.com/oss-review-toolkit/ort/pull/4507.

PR #4495, opened 2021-09-23
detekt-rules: Show a colored info if the Gradle plugin needs to reload

sschuberth:
Also see [[1]].

[1]: https://detekt.github.io/detekt/extensions.html#pitfalls

### DIFF
--- a/detekt-rules/build.gradle.kts
+++ b/detekt-rules/build.gradle.kts
@@ -17,6 +17,10 @@
  * License-Filename: LICENSE
  */
 
+import org.gradle.internal.logging.text.StyledTextOutput
+import org.gradle.internal.logging.text.StyledTextOutputFactory
+import org.gradle.kotlin.dsl.support.serviceOf
+
 val detektPluginVersion: String by project
 
 plugins {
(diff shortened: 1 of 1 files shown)


detekt-rules/build.gradle.kts
```diff
@@ -17,6 +17,10 @@
  * License-Filename: LICENSE
  */
 
+import org.gradle.internal.logging.text.StyledTextOutput
```
mnonnenmacher, line 20: Commit message: Are the double brackets intentional?

sschuberth: Yes, as that way it renders like:

image

mnonnenmacher: I see, so you want the brackets to show in the PR description. I would prefer to stick with the formatting we have used so far, for consistency, and because double brackets only look good if the commit message is directly used for the PR description for single commit PRs, but they look bad in all other places.

sschuberth: I've removed the double square brackets.

mnonnenmacher (approved):

detekt-rules/build.gradle.kts
mnonnenmacher: Thanks.

PR #4496, opened 2021-09-23
scanner: Prevent storing the same scan result twice

mnonnenmacher:

### DIFF
--- a/scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
+++ b/scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
@@ -159,6 +159,24 @@ abstract class AbstractStorageFunTest : WordSpec() {
                 readResult should beSuccess()
                 (readResult as Success).result should beEmpty()
             }
+
+            "not store a result for the same scanner and provenance twice" {
+                val summary1 = scanSummaryWithFiles
+                val summary2 = scanSummaryWithFiles.copy(packageVerificationCode = "anotherPackageVerificationCode")
+
+                val scanResult1 = ScanResult(provenanceWithSourceArtifact1, scannerDetails1, summary1)
+                val scanResult2 = ScanResult(provenanceWithSourceArtifact1, scannerDetails1, summary2)
+
+                val addResult1 = storage.add(id1, scanResult1)
+                val addResult2 = storage.add(id1, scanResult2)
+
+                addResult1 should beSuccess()
+                addResult2 should beFailure()
+
+                val readResult = storage.read(id1)
+                readResult should beSuccess()
+                (readResult as Success).result should containExactly(scanResult1)
+            }
         }
 
         "Reading a scan result" should {
--- a/scanner/src/main/kotlin/storages/FileBasedStorage.kt
+++ b/scanner/src/main/kotlin/storages/FileBasedStorage.kt
@@ -83,6 +83,15 @@ class FileBasedStorage(
             is Failure -> emptyList()
         }
 
+        if (existingScanResults.any { it.scanner == scanResult.scanner && it.provenance == scanResult.provenance }) {
+            val message = "Did not store scan result for '${id.toCoordinates()}' because a scan result for the same " +
+                    "scanner and provenance was already stored."
+
+            log.warn { message }
+
+            return Failure(message)
+        }
+
         val scanResults = existingScanResults + scanResult
 
         val path = storagePath(id)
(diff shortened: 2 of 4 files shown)


scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
```diff
@@ -159,6 +160,20 @@ abstract class AbstractStorageFunTest : WordSpec() {
...
             }
+
+            "not store a result for the same scanner and provenance twice" {
+                val scanResult = ScanResult(provenanceWithSourceArtifact1, scannerDetails1, scanSummaryWithFiles)
+
+                val addResult1 = storage.add(id1, scanResult)
+                val addResult2 = storage.add(id1, scanResult)
+
+                addResult1 should beSuccess()
+                addResult2 should beFailure()
```
fviernau, line 171: I wonder if returning failure in that case is good, because from that failure the caller cannot know whether a
succeeding `read(id1)` would return the stored value. What do you think?

Also would overwriting make sense, like with `Map.put` ?

scanner/src/main/kotlin/storages/FileBasedStorage.kt
```diff
@@ -83,6 +83,15 @@ class FileBasedStorage(
...
         }
 
+        if (existingScanResults.any { it.scanner == scanResult.scanner && it.provenance == scanResult.provenance }) {
+            val message = "Did not store scan result for '${id.toCoordinates()}' because a scan result for the same " +
+                    "scanner and provenance was already stored."
+
+            log.warn { message }
+
+            return Failure(message)
+        }
```
fviernau, line 93: Is this class used with concurrency? (seems not but wanted to double check)

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
```diff
@@ -159,6 +160,20 @@ abstract class AbstractStorageFunTest : WordSpec() {
                 readResult should beSuccess()
                 (readResult as Success).result should beEmpty()
             }
+
+            "not store a result for the same scanner and provenance twice" {
+                val scanResult = ScanResult(provenanceWithSourceArtifact1, scannerDetails1, scanSummaryWithFiles)
+
+                val addResult1 = storage.add(id1, scanResult)
+                val addResult2 = storage.add(id1, scanResult)
```
fviernau, line 168: Would it make sense to use a different summary (non-key attributes) for the second add call, and then assert which of the two scan results is in storage after execution. That'd document the behavior I asked about below.

fviernau (changes requested):

scanner/src/main/kotlin/storages/FileBasedStorage.kt
mnonnenmacher: Currently not.

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
mnonnenmacher: I'll wait for the outcome of the other discussion.

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
fviernau: > Makes sense, but on the other hand even though we expect scanners to be deterministic the new result might be different. It will at least have different timestamps. The current documentation of the add() function does not define the behavior for this situation. The reason I chose this implementation was that it was simpler, e.g. we don't have to check the SQL exception to determine the reason why the query has failed.

Makes sense. But still, would it make sense to keep the implementation of the storage as-is, but adjust the test so that it ensures that the second `add()` call does not overwrite anything? This would ensure that the different storage implementations behave in the same way.

> I wouldn't do that because it's less efficient to overwrite. For the file based storage we don't have to upload anything and for the Postgres storage we don't have to recover from the exception.

Make sense to me as well.

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
mnonnenmacher: I have updated the test.

fviernau (approved):

mnonnenmacher:
I have confirmed that the performance impact of this change is negligible. For a test database with ~60,000 entries it adds about 0.4ms to the duration of adding a single row to the table.

PR #4497, opened 2021-09-23

fviernau:


model/src/main/kotlin/licenses/LicenseInfoResolver.kt
```diff
@@ -138,7 +138,9 @@ class LicenseInfoResolver(
                     pathExclude.matches(licenseFinding.location.prependPath(findings.relativeFindingsPath))
                 }
             }
-        }.groupBy(keySelector = { it.first }, valueTransform = { it.second }).mapValues { it.value.all { it } }
+        }.groupBy(keySelector = { it.first }, valueTransform = { it.second }).mapValues { (_, isExcluded) ->
+            isExcluded.all { it }
```
sschuberth, line 142: Nit: By convention, the "is" prefix for variables should only be used for single booleans, not for a list of booleans. So maybe just call the list "excluded"?

sschuberth (approved):

sschuberth (approved):

PR #4500, opened 2021-09-24
FilePackageCurationProvider: Fail if curations file cannot be read

fviernau:
Fail if any curation file does not exist or cannot be deserialized.
Previously, it could easily go unnoticed if curations have not been
applied, because the only hint was a warning in the logs.

### DIFF
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -48,12 +48,12 @@ class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCura
     }
 
     val packageCurations: Set<PackageCuration> = run {
-        val allCurations = curationFiles.mapNotNull { curationFile ->
+        val allCurations = curationFiles.map { curationFile ->
             runCatching {
                 curationFile.readValueOrDefault(emptyList<PackageCuration>())
             }.onFailure {
                 log.warn { "Failed parsing package curation from '${curationFile.absoluteFile}'." }
-            }.getOrNull()
+            }.getOrThrow()
         }.flatten()
 
         val duplicates = allCurations.getDuplicates()
--- a/analyzer/src/test/assets/package-curations-not-deserializable.yml
+++ b/analyzer/src/test/assets/package-curations-not-deserializable.yml
@@ -1,0 +1,1 @@
+not deserializable
--- a/analyzer/src/test/kotlin/curation/FilePackageCurationProviderTest.kt
+++ b/analyzer/src/test/kotlin/curation/FilePackageCurationProviderTest.kt
@@ -19,6 +19,7 @@
 
 package org.ossreviewtoolkit.analyzer.curation
 
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.StringSpec
 import io.kotest.matchers.collections.beEmpty
 import io.kotest.matchers.collections.haveSize
@@ -110,6 +111,22 @@ class FilePackageCurationProviderTest : StringSpec() {
             }
 
             curationsOutVersion should beEmpty()
+        }
+
+        "Provider throws an exception if the curations file is not de-serializable" {
+            val curationsFile = File("src/test/assets/package-curations-not-deserializable.yml")
+
+            shouldThrow<Exception> {
+                FilePackageCurationProvider(curationsFile)
+            }
+        }
+
+        "Provider throws an exception if the curations file does not exist" {
+            val curationsFile = File("src/test/assets/package-curations-not-existing.yml")
+
+            shouldThrow<Exception> {
+                FilePackageCurationProvider(curationsFile)
+            }
         }
     }
 }


analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -53,6 +53,7 @@ class FilePackageCurationProvider(curationFiles: Collection<File>) : PackageCura
                 curationFile.readValueOrDefault(emptyList<PackageCuration>())
             }.onFailure {
                 log.warn { "Failed parsing package curation from '${curationFile.absoluteFile}'." }
+                throw it
```
sschuberth, line 56: If at all, you should use `getOrThrow()` in the line below instead. But that'd also render the surrounding `mapNotNull()` useless.

Didn't we have a discussion with @MarcelBochtler back then and agreed that we should not stop reading further (valid) curations just becuase there was an invalid curation in between?

sschuberth (changes requested):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
fviernau: > Didn't we have a discussion with @MarcelBochtler back then and agreed that we should not stop reading further (valid) curations just becuase there was an invalid curation in between?

It could be that there was such discussion, I remember even being involved there but not the details.
Anyhow, as we just faced the issue of curations not being applied unnoticed I believe this change is good, no matter what the discussion before was. What do you think?

fviernau: > If at all, you should use `getOrThrow()` in the line below instead. But that'd also render the surrounding `mapNotNull()` useless.

Can you provide a rational why getOrThrow()`  has to be used?

sschuberth: > Anyhow, as we just faced the issue of curations not being applied unnoticed I believe this change is good, no matter what the discussion before was. What do you think?

I'm not sure I agree. If you have multiple parties contributing to curations / have multiple places curations are read from, it's highly inconvenient to allow any broken curation from someone to cause curations from others being disregarded until the one broken curation is fixed.

> Can you provide a rational why getOrThrow()` has to be used?

Well, because I believe it matches your semantics. You want to either get the (valid) curation, or throw an exception, right?

fviernau: > I'm not sure I agree. If you have multiple parties contributing to curations / have multiple places curations are read from, it's highly inconvenient to allow any broken curation from someone to cause curations from others being disregarded until the one broken curation is fixed.

How about running the `FilePackageCurationProvider ` in the PSV for these contributions so that the problem is non-existent?

fviernau: > Well, because I believe it matches your semantics. You want to either get the (valid) curation, or throw an exception, right?

IIUC using `getOrThrow` would require removing the output of the log message.
The problem with that is, that the path of the problematic file is not output anymore (if it isn't contained in the exception, which is the case).

sschuberth: It might be that we do not have control over the PSV checks performed at all sources for curations. Anyway, I do not remember the details of the discussion anymore, and @MarcelBochtler is currently on vacation.

sschuberth: > IIUC using `getOrThrow` would require removing the output of the log message.

No, you can have both `onFailure` and `getOrThrow`. Just like I wrote above: You'd simply have to replace `getOrNull` with `getOrThrow` and replace `mapNotNull` with `map`...

fviernau: > It might be that we do not have control over the PSV checks performed at all sources for curations.

IMO maybe that's the problem which needs to be addressed. 

However, we can just leave this PR (discussion) alone until @MarcelBochtler is back if you prefer.

fviernau: > No, you can have both `onFailure` and `getOrThrow`. Just like I wrote above: You'd simply have to replace `getOrNull` with `getOrThrow` and replace `mapNotNull` with `map`...

Thanks!

rdesgroppes: too late to the party but better safe than sorry: I don't think it's a good idea to swallow such exceptions (i.e. denoting invalid inputs) in that it leads to:
1. draw wrong conclusions on the applicable licenses, s.a. allowing to release infringing software or preventing from releasing compliant software,
2. delays in realizing there's a problem (who would look for hypothetical warnings in a job output when the said job succeeds?),
3. delays in taking necessary countermeasures.

I therefore kindly request to have the tool [fail fast(er)](https://dzone.com/articles/fail-fast-principle-in-software-development) should it be passed an invalid input.

analyzer/src/test/assets/package-curations-not-deserializable.yml
```diff
@@ -0,0 +1,21 @@
...
+      hash: ""
+      hash_algorithm: ""
+    vcs:
+      type: ""
+      url: "http://url.git"
+      revision: ""
+      path: ""
+- id: "Maven:org.hamcrest:hamcrest-library:1.3"
+curations:
+homepage_url: "http://home.page"
```
mnonnenmacher, line 21: For the purpose of the test the file could be smaller, an empty file or a single "-" should already be sufficient.

fviernau: Single dash didn't work, but I reduced the size...

sschuberth (approved):
Looks like this is consistent now with what we do in `SimplePackageConfigurationProvider`:

https://github.com/oss-review-toolkit/ort/blob/c5ae7f53921ce6e7c466e3c8575ce74a17b3f262/model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt#L53-L57

So I'm fine with merging.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
MarcelBochtler: > > It might be that we do not have control over the PSV checks performed at all sources for curations.
> 
> IMO maybe that's the problem which needs to be addressed.
> 
> However, we can just leave this PR (discussion) alone until @MarcelBochtler is back if you prefer.


There was not much discussion in the PR introducing multiple curation files: https://github.com/oss-review-toolkit/ort/pull/4131#discussion_r647997224

If we want to increase the visibility of invalid curations, how about creating an ORT issue, so it is listed in the resulting report without breaking the ORT run?

Hint: Local curations currently also throw an exception.

PR #4502, opened 2021-09-27
Wiremock structure

sschuberth:


scanner/src/test/kotlin/storages/ClearlyDefinedStorageTest.kt
```diff
@@ -225,15 +228,15 @@ private fun assertCurrentTime(time: Instant) {
...
+    val templateFile = File("$TEST_FILES_ROOT/cd_definitions.json")
     return templateFile.readText()
 }
 
 class ClearlyDefinedStorageTest : WordSpec({
     val wiremock = WireMockServer(
         WireMockConfiguration.options()
             .dynamicPort()
-            .usingFilesUnderDirectory("src/test/assets/")
+            .usingFilesUnderDirectory(TEST_FILES_ROOT)
```
oheger-bosch, line 239: Is this configuration option needed here? It does not seem to be defined for the other tests.

oheger-bosch (approved):
Changes look good to me.

There is still some redundancy in the WireMock tests, e.g. to declare the server and start and stop it. Would it be possible (in a later PR) to extract this into a helper class or a new test base class?

scanner/src/test/kotlin/storages/ClearlyDefinedStorageTest.kt
sschuberth: I believe it's needed (as we use "assets", not "resources"), and I also see `usingFilesUnderDirectory` being called in other tests like `VulnerableCodeTest`, `JiraNotifierTest` etc.

sschuberth:
> Would it be possible (in a later PR) to extract this into a helper class or a new test base class?

Probably. I haven't looked at that, but just did a search-and-replace, more or less. Any improvements you see fit are welcome in the form of a PR 😉

PR #4514, opened 2021-09-29
Add support for unpacking Debian package archives

oheger-bosch:
Refer to the single commit messages for further details.


utils/src/test/assets/control-expected.txt
```diff
@@ -0,0 +1,11 @@
+Package: testpkg
+Source: test.src
+Version: 1.2.3.1-1+apertis1bv2021.0b1
```
nnobelis, line 3: Maybe change to a more normal version, to remove reference to Apertis ?

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -69,34 +70,40 @@ enum class ArchiveType(vararg val extensions: String) {
...
     }
 
 /**
- * Unpack the [File] assuming it is a 7-Zip archive. This implementation ignores empty directories and symbolic links.
+ * Unpack the [File] assuming it is a 7-Zip archive. This implementation ignores empty directories and symbolic links
+ * and all entries not matched by the given [filter].
  */
-fun File.unpack7Zip(targetDirectory: File) {
+fun File.unpack7Zip(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = { true }) {
+    val skip = skipFilter(
```
nnobelis, line 97: This code is really confusing. The first parameter is called `skip` but you name the full object also `skip` locally. Would it be not better to name the local variable `filter` ?
Same question everywhere.

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -112,6 +118,35 @@ fun File.unpack7Zip(targetDirectory: File) {
...
+ * A list with file names that are expected to be contained in a Debian package archive file.
+ */
+private val DEBIAN_ARCHIVE_CONTENT = listOf("data.tar.xz", "control.tar.xz")
+
+/**
+ * Unpack the [File] assuming it is a Debian archive. A Debian archive is an ar archive that contains itself two tar
+ * files with metadata and the actual content of the package. The top-level archive is deflated into a temporary
+ * directory. Then the tar files are extracted to the provided [targetDirectory].
+ */
+fun File.unpackDeb(targetDirectory: File) {
```
fviernau: Should this be made private?

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -112,6 +118,35 @@ fun File.unpack7Zip(targetDirectory: File) {
  */
 fun File.unpackZip(targetDirectory: File) = ZipFile(this).unpack(targetDirectory)
 
+/**
+ * A list with file names that are expected to be contained in a Debian package archive file.
+ */
+private val DEBIAN_ARCHIVE_CONTENT = listOf("data.tar.xz", "control.tar.xz")
+
+/**
+ * Unpack the [File] assuming it is a Debian archive. A Debian archive is an ar archive that contains itself two tar
```
fviernau, line 135: To me "itself" sounds odd, consider either dropping it or saying "..which in turn contains two tar.."
(Could be just my english)

Also maybe: "...files with metadata and the actual content of the package." -> 
"...files, meta data and the actual content of the package.

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -112,6 +118,35 @@ fun File.unpack7Zip(targetDirectory: File) {
...
+            tempDir,
+            { entry -> entry.isDirectory || File(entry.name).isAbsolute },
+            { entry -> (entry as ArArchiveEntry).mode }
+        )
+
+        DEBIAN_ARCHIVE_CONTENT.forEach { name ->
+            val file = tempDir.resolve(name)
+            file.unpack(targetDirectory)
+        }
+    } finally {
```
fviernau, line 154: Does the above loop create the dirs "$targetDirectory/data", "$targetDirectory/content"?

utils/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: I think the name `skip` is correct in this context, as the resulting function returns *true* when an entry should be skipped. This becomes obvious in the `if (skip(entry))` statement. Maybe the name of the function `skipFilter()` is a bit misleading.

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -112,6 +118,35 @@ fun File.unpack7Zip(targetDirectory: File) {
  */
 fun File.unpackZip(targetDirectory: File) = ZipFile(this).unpack(targetDirectory)
 
+/**
+ * A list with file names that are expected to be contained in a Debian package archive file.
+ */
+private val DEBIAN_ARCHIVE_CONTENT = listOf("data.tar.xz", "control.tar.xz")
```
fviernau, line 132: maybe rename to: DEBIAN_PACKAGE_SUBARCHIVES

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -76,5 +77,43 @@ class ArchiveUtilsTest : StringSpec() {
...
+
+        "Debian udeb archive can be unpacked" {
+            val archive = File("src/test/assets/testpkg.udeb")
+
+            archive.unpack(outputDir)
+
+            val extractedScriptFile = outputDir.resolve("usr/bin/test")
+            extractedScriptFile.exists() shouldBe true
+
+            val extractedControlFile = outputDir.resolve("control")
```
fviernau, line 152: The code seems to be copy and pasted but the empty line above is inconsistent.

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -76,5 +77,43 @@ class ArchiveUtilsTest : StringSpec() {
...
+            extractedControlFile.readText() shouldBe expectedControl
+
+            listOf("data.tar.xz", "control.tar.xz").forEach { tarFileName ->
+                val tarFile = outputDir.resolve(tarFileName)
+                tarFile.exists() shouldBe false
+            }
+        }
+
+        "Debian udeb archive can be unpacked" {
+            val archive = File("src/test/assets/testpkg.udeb")
```
fviernau, line 145: Instead of adding two separate identical assets, how about making a copy of the file in the test to eliminate the redundancy?

Also,  IIUC only one test is needed and for the other format it would suffice to mock the return value of the archive function.

utils/src/main/kotlin/ArchiveUtils.kt
nnobelis: I understand. But nevertheless when following the call chain, it gets confusing.

You can leave it as if if you want.

nnobelis (dismissed):

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -112,6 +118,35 @@ fun File.unpack7Zip(targetDirectory: File) {
...
+        ArArchiveInputStream(inputStream()).unpack(
+            tempDir,
+            { entry -> entry.isDirectory || File(entry.name).isAbsolute },
+            { entry -> (entry as ArArchiveEntry).mode }
+        )
+
+        DEBIAN_ARCHIVE_CONTENT.forEach { name ->
+            val file = tempDir.resolve(name)
+            file.unpack(targetDirectory)
+        }
```
sschuberth: I'm surprised that this is still needed, as I was reading [this comment](https://stackoverflow.com/questions/7432223/open-debian-package-with-java/7553295#comment9037704_7432828) so that Commons Compress would do it automatically... but then again, it probably can't, as it's seeing the `.deb` only as an `.ar`, not knowing about the "nested tar convention".

utils/src/test/kotlin/ArchiveUtilsTest.kt
fviernau: Can the redundancy be removed by merging both tests, using the following?
´``
listOf(x.deb, e.udeb).forAll {
  
}
´``

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -69,34 +70,40 @@ enum class ArchiveType(vararg val extensions: String) {
 }
 
 /**
- * Unpack the [File] to [targetDirectory].
+ * Unpack the [File] to [targetDirectory] using [filter] to select only the entries of interest.
```
fviernau, line 73: commit: Why is logic necessary? The filtering can be applied as well after extracting ALL files for simplicity.
The commit message rationale should at least consider this IMO.

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -76,5 +77,43 @@ class ArchiveUtilsTest : StringSpec() {
...
+
+            val extractedScriptFile = outputDir.resolve("usr/bin/test")
+            extractedScriptFile.exists() shouldBe true
+
+            val extractedControlFile = outputDir.resolve("control")
+            extractedControlFile.exists() shouldBe true
+            val expectedControl = File("src/test/assets/control-expected.txt").readText()
+            extractedControlFile.readText() shouldBe expectedControl
+
+            listOf("data.tar.xz", "control.tar.xz").forEach { tarFileName ->
```
sschuberth, line 138: Should we use `DEBIAN_ARCHIVE_CONTENT` here? Or should we exactly *not* to that for testing?

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -157,5 +159,94 @@ class ArchiveUtilsTest : StringSpec() {
...
+            val extractedControlFile = outputDir.resolve("control")
+            extractedControlFile.exists() shouldBe false
+        }
     }
 }
+
+/**
+ * A filter for archive entries that filters only for a file named "a".
+ */
+private val aFilter: (ArchiveEntry) -> Boolean = { it.name == "a" }
```
fviernau, line 252: All caps var name?

.gitattributes
```diff
@@ -31,3 +31,6 @@ reporter-web-app/public/index.html		linguist-generated=true
 spdx-utils/src/main/resources/exceptions/*	linguist-detectable=false
 spdx-utils/src/main/resources/licenserefs/*	linguist-detectable=false
 spdx-utils/src/main/resources/licenses/*	linguist-detectable=false
+
+# Use Unix line endings for the expected control file of a Debian package.
+utils/src/test/assets/control-expected.txt  text eol=lf
```
fviernau, line 36: commit message: Please provide a rationale why this new option is needed, which use case it targets.

fviernau (changes requested):

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -76,5 +77,43 @@ class ArchiveUtilsTest : StringSpec() {
...
             fileB.readText() shouldBe "b\n"
         }
+
+        "Debian deb archive can be unpacked" {
+            val archive = File("src/test/assets/testpkg.deb")
+
+            archive.unpack(outputDir)
+
+            val extractedScriptFile = outputDir.resolve("usr/bin/test")
+            extractedScriptFile.exists() shouldBe true
```
sschuberth, line 131: Please use matchers like `shouldBe aFile()` which provide nicer output on failure that just a mismatch to "true".

sschuberth (changes requested):

utils/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: I would say yes, but all the other format-specific `unpack()` functions are public as well.

sschuberth: IMO this should stay publish like the other functions to allow unpacking of concrete formats independent of the file extension.

utils/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: No, it just deflates the two archives to `targetDirectory`. There should not be any overlapping content in them.

utils/src/main/kotlin/ArchiveUtils.kt
fviernau: Ok with me.

utils/src/main/kotlin/ArchiveUtils.kt
fviernau: I'd vote for creating the dirs to not make assumptions about whether it overlaps.

oheger-bosch: Done.

utils/src/test/kotlin/ArchiveUtilsTest.kt
oheger-bosch: Merged the tests.

utils/src/test/kotlin/ArchiveUtilsTest.kt
oheger-bosch: Done

utils/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: Added a corresponding statement to the commit message, although IMHO the reasoning is pretty obvious.

utils/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: Yes, Commons Compress only extracts the two archive files from the original package.

utils/src/test/kotlin/ArchiveUtilsTest.kt
oheger-bosch: I opted for the latter. Also, the constant is private.

sschuberth: > Also, the constant is private.

Well, you could easily make it `internal` (which would be my personal preference).

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -143,12 +146,101 @@ class ArchiveUtilsTest : StringSpec() {
 
                     listOf("data.tar.xz", "control.tar.xz").forEach { tarFileName ->
                         val tarFile = outputDir.resolve(tarFileName)
-                        tarFile shouldBe aFile()
+                        tarFile shouldNotBe aFile()
```
sschuberth, line 149: Why this change here as no filter has been set?

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -143,12 +146,101 @@ class ArchiveUtilsTest : StringSpec() {
...
                     }
                 }
             } finally {
                 tempDir.safeDeleteRecursively(force = true)
             }
         }
+
+        "Tar archive can be unpacked with a filter" {
```
sschuberth, line 157: Can we please group non-filter / filter tests by start using a `WordSpec` like

```
        "Tar archives" should {
            "unpack" {}
            "unpack with a filter" {}
        }
```

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -269,14 +277,18 @@ private fun ArchiveInputStream.unpack(
  * Unpack the [ZipFile]. In contrast to [InputStream.unpackZip] this properly parses the ZIP's central directory, see
  * https://commons.apache.org/proper/commons-compress/zip.html#ZipArchiveInputStream_vs_ZipFile.
  */
-private fun ZipFile.unpack(targetDirectory: File) =
+private fun ZipFile.unpack(targetDirectory: File, filter: (ArchiveEntry) -> Boolean) =
     use { zipFile ->
         val entries = zipFile.entries
+        val skip = skipFilter(
+            { entry -> entry.isDirectory || (entry as ZipArchiveEntry).isUnixSymlink || File(entry.name).isAbsolute },
```
sschuberth, line 284: I think I'd prefer to not introduce the `skipFilter()` function, but inline the `filter()` call to where the original filtering already happens. At this example, this as the advantage of avoiding the `as ZipArchiveEntry` cast.

sschuberth (changes requested):

utils/src/test/kotlin/ArchiveUtilsTest.kt
oheger-bosch: Converted the test class to a `WordSpec`.

utils/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: Done. However, now detekt complains about complex conditions; so I had to add `@Suppress` annotations.

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -76,5 +79,29 @@ class ArchiveUtilsTest : StringSpec() {
...
             fileB.readText() shouldBe "b\n"
         }
+
+        "Debian deb archive can be unpacked" {
+            val tempDir = createTestTempDir(ORT_NAME)
+            val archiveDeb = File("src/test/assets/testpkg.deb")
+            val archiveUdepTemp = tempDir.resolve("testpkg.udeb")
+            val archiveUdep = archiveDeb.copyTo(archiveUdepTemp)
+
+            listOf(archiveDeb, archiveUdep).forEach { archive ->
```
fviernau: Please use `forAll` instead of `forEach` for better output in case of assertion failure.

utils/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -30,76 +30,132 @@ import java.io.File
...
     private lateinit var outputDir: File
 
     override fun beforeTest(testCase: TestCase) {
         outputDir = createTestTempDir()
     }
 
     init {
-        "Tar GZ archive can be unpacked" {
-            val archive = File("src/test/assets/test.tar.gz")
+        "Tar archives" should {
```
fviernau, line 43: Consider grouping it by function, like:

"unpack()" should { "unpack Tar archvies"

utils/src/main/kotlin/ArchiveUtils.kt
fviernau: Well, it was obvious to me as well that the filtering is faster. I wasn't after this.I hope I don't just misunderstand this.
I wanted to understand the use case making it necessary. So you say you are planning to filter for LICENSE files.
and NOTICE files. Can you give a hint (not necessarily in the commit message) what the use case is?

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -145,17 +148,18 @@ fun File.unpackDeb(targetDirectory: File) {
             subFolder.safeMkdirs()
 
             val file = tempDir.resolve(name)
-            file.unpack(subFolder)
+            file.unpack(subFolder, filter)
```
fviernau, line 151: I presume that the filter acts on the file path in the `deb` package.
The call in line 151 makes it unclear unclear how the filter works, because I guess the filter then now suddenly works on paths relative to the root of the `.xz` library. And there is no means for the caller to make it specific to one `.xz` library.

My preference would be to ditch this commit as it adds complexity and I don't really see the need for it right now. If you strongly think the filter should be done, I won't block it though.

fviernau (changes requested):

utils/src/test/kotlin/ArchiveUtilsTest.kt
oheger-bosch: The current grouping was suggested by @sschuberth. Could you two come to an agreement, please?

fviernau: With the "Consider" prefix a wanted to indicate that it's fully optional and only would like it to be considered, but leave the decision up to you. So no agreement with me needed. (Wasn't aware of @sschuberth 's suggestion)

utils/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: It is correct that the filter has no information about the subarchive the entry passed to it is contained. I do not see this as a problem in this specific case, because the paths in the subarchives are disjunct, and even if you do not want to rely on this, this is just a filter; so the worst thing that could happen is that too many files would be extracted.

Our use case is that we want to collect the license information from all the packages contained in a Debian repository. This information is available as a text file contained at a specific path in the package archive. The repository contains ~ 15k packages. So we do not want to extract all of them just to pick up the license files.

fviernau: > It is correct that the filter has no information about the subarchive the entry passed to it is contained. I do not see this as a problem in this specific case, because the paths in the subarchives are disjunct, and even if you do not want to rely on this, this is just a filter; so the worst thing that could happen is that too many files would be extracted.

I don't really see the problem with the overlap which is not happening in reality, but with the (1) broken encapsulation (developer needs to look into the details to understand how the filter actually works and (2) an unclear an misleading function contract, e.g. the developer wonders whereto the paths are relative or falsely assume that the filter paths are relative to the root of the archive. I'd appreciate to see how critical others see that, maybe @oss-review-toolkit/core-devs anyone?

> Our use case is that we want to collect the license information from all the packages contained in a Debian repository. This information is available as a text file contained at a specific path in the package archive. The repository contains ~ 15k packages. So we do not want to extract all of them just to pick up the license files.

How will that use case be integrated into ORT? What component are you planning to add which is responsible for scanning the contained Debian repository?

sschuberth: > How will that use case be integrated into ORT? What component are you planning to add which is responsible for scanning the contained Debian repository?

This will be implemented as a Bosch-internal scanner plugin.

utils/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -112,6 +118,40 @@ fun File.unpack7Zip(targetDirectory: File) {
...
+    val tempDir = createOrtTempDir("unpackDeb")
+
+    try {
+        ArArchiveInputStream(inputStream()).unpack(
+            tempDir,
+            { entry -> entry.isDirectory || File(entry.name).isAbsolute },
+            { entry -> (entry as ArArchiveEntry).mode }
+        )
+
+        DEBIAN_PACKAGE_SUBARCHIVES.forEach { name ->
```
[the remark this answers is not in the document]
sschuberth: That's IMO overdoing it a bit. Remember that we still talk specifically about `.deb` files, not `.ar` files in general, or even any sort of nested archives. And `.deb` files have a defined layout, which only allows those two archive names at the root. Thus, `name` was the correct choice here, IMO. But I do not want to stress @oheger-bosch's nerves by asking him to change it back again 😄

utils/src/test/kotlin/ArchiveUtilsTest.kt
sschuberth: Just as a remark: I was meaning to do the conversion to a `WordSpec` in the last commit when introducing the new tests, as before that the conversion does not really have any value, and interferes with the other changes. No need to change that now again, though.

utils/src/main/kotlin/ArchiveUtils.kt
sschuberth: > I'd appreciate to see how critical others see that, maybe @oss-review-toolkit/core-devs anyone?

I don't see it that critical, mostly because filtering can happen on any property of `ArchiveEntry`, not just the path. For the special case of `unpackDeb()` it would be enough for me to document that `filter()` applies to each nested archive.

fviernau: > This will be implemented as a Bosch-internal scanner plugin.

@oheger-bosch , @sschuberth  if some functionality is being added which is not even planned to be used in ORT, how shall the reviewer review that if he doesn't even know about that fact? (I'm a bit puzzled why this information hasn't been provided with the PR description or commit message)

sschuberth: > I'm a bit puzzled why this information hasn't been provided with the PR description or commit message

I was (and still am) convinced that the sole feature to be able to unpack `.deb` archives is valuable to have in ORT upstream. The use in an internal binary scanner is independent of that.

fviernau: > I was (and still am) convinced that the sole feature to be able to unpack `.deb` archives is valuable to have in ORT upstream. The use in an internal binary scanner is independent of that.

Yeah, with just unpacking `.deb` files I didn't see an issue, because it' simple.
With adding the filtering it smelled a bit like unneeded complexity and added redundancy in tests, so I wanted to know the details whether it is really needed. Given the information about use cases in ORT provided, I still cannot see the need for the filtering function. Anyhow, this single case is not a big issue as it''s isolated. But if repeatedly unused code would be added, I'd tend to question that. I'll leave it anyhow up to you guys to decide about this one, I can live with it being merged.

Otherwise LGTM now.

sschuberth: > Given the information about use cases in ORT provided, I still cannot see the need for the filtering function.

It's simply a performance optimization: File I/O is slow, esp. when dealing with many files in many archives, and if you're only interested in a small subset of the files, you may want to skip unpacking the other files in the first place rather than ignoring them in a later step.

oheger-bosch: > 
> I don't see it that critical, mostly because filtering can happen on any property of `ArchiveEntry`, not just the path. For the special case of `unpackDeb()` it would be enough for me to document that `filter()` applies to each nested archive.

I have enhanced the documentation of the `unpackDeb()` function accordingly.

fviernau (approved):

utils/src/main/kotlin/ArchiveUtils.kt
sschuberth: I believe the added docs mention some unnecessary implementation details. But let's get this merged now, I'll make a follow-up proposal afterwards.

sschuberth (approved):

PR #4515, opened 2021-09-29
Un-deprecate a path exclude reason

fviernau:
Undo a deprecation which should not have been done in the first place according to [1] and refactor an example policy rule accordingly.

[1] https://github.com/spdx/spdx-spec/blob/development/v2.2.2/chapters/relationships-between-SPDX-elements.md

### DIFF
--- a/model/src/main/kotlin/config/PathExcludeReason.kt
+++ b/model/src/main/kotlin/config/PathExcludeReason.kt
@@ -72,12 +72,5 @@ enum class PathExcludeReason {
      * The path only contains tools used for testing source code which are not included in distributed build
      * artifacts.
      */
-    @Deprecated(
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/config/PathExcludeReason.kt
```diff
@@ -72,12 +72,5 @@ enum class PathExcludeReason {
      * The path only contains tools used for testing source code which are not included in distributed build
      * artifacts.
      */
-    @Deprecated(
```
sschuberth, line 75: Commit message: Please use a tag instead of a branch for the link:

https://github.com/spdx/spdx-spec/blob/v2.2.2/chapters/relationships-between-SPDX-elements.md

sschuberth (approved):

PR #4521, opened 2021-09-30
DependencyNavigator: Handle results with empty dependency graphs

oheger-bosch:
CompatibilityDependencyNavigator.create() could fail with an exception
for an OrtResult that contains projects with an empty set of scope
names, but no dependency graph. This is a corner case that could
appear if a package manager had issues when constructing projects. To
prevent this exception, check first whether the result contains a
dependency graph, and if not, return the DependencyTreeNavigator.

### DIFF
--- a/model/src/main/kotlin/CompatibilityDependencyNavigator.kt
+++ b/model/src/main/kotlin/CompatibilityDependencyNavigator.kt
@@ -45,6 +45,10 @@ class CompatibilityDependencyNavigator internal constructor(
          * [CompatibilityDependencyNavigator].
          */
         fun create(ortResult: OrtResult): DependencyNavigator {
+            if (ortResult.analyzer?.result?.dependencyGraphs.orEmpty().isEmpty()) {
+                return DependencyTreeNavigator
+            }
+
             val (treeProjects, graphProjects) = ortResult.getProjects().partition(Project::usesTree)
 
             return when {
--- a/model/src/test/kotlin/CompatibilityDependencyNavigatorTest.kt
+++ b/model/src/test/kotlin/CompatibilityDependencyNavigatorTest.kt
@@ -78,6 +78,26 @@ class CompatibilityDependencyNavigatorTest : WordSpec() {
 
                 navigator shouldBe DependencyTreeNavigator
             }
+
+            "return a DependencyTreeNavigator for a result that does not contain any dependency graphs" {
+                val project = createProject("dummy", scopeNames = sortedSetOf())
+
+                val analyzerResult = AnalyzerResult(
+                    projects = sortedSetOf(project),
+                    packages = sortedSetOf(),
+                    dependencyGraphs = mapOf()
+                )
+                val analyzerRun = AnalyzerRun(
+                    result = analyzerResult,
+                    environment = Environment(),
+                    config = AnalyzerConfiguration()
+                )
+                val result = OrtResult.EMPTY.copy(analyzer = analyzerRun)
+
+                val navigator = CompatibilityDependencyNavigator.create(result)
+
+                navigator shouldBe DependencyTreeNavigator
+            }
         }
 
         "scopeNames" should {


sschuberth (approved):

model/src/main/kotlin/CompatibilityDependencyNavigator.kt
```diff
@@ -45,6 +45,10 @@ class CompatibilityDependencyNavigator internal constructor(
          * [CompatibilityDependencyNavigator].
          */
         fun create(ortResult: OrtResult): DependencyNavigator {
+            if (ortResult.analyzer?.result?.dependencyGraphs.orEmpty().isEmpty()) {
```
fviernau, line 48: To my understanding, this problem happens because the ORT result gets constructed in the wrong way.
Recently [1] was introduced to fix the broken construction. The rationale for [1] was that if the construction is broken, the construction should be fixed rather than adding handling for a falsely constructed OrtResult.

From that perspective, would it make sense to instead extend [1] to cover the problematic corner case from this context?

[1] https://github.com/oss-review-toolkit/ort/blob/5a2704eab8a47e431b0912f26c6bcebfc90e5067/analyzer/src/main/kotlin/PackageManager.kt#L306-L316

fviernau: So, this would be an adjustment of the `any` condition...

fviernau (changes requested):

model/src/main/kotlin/CompatibilityDependencyNavigator.kt
fviernau: FYI, I just stumbled across another similar bug in that area. Maybe this is interesting for this commit, I believe it's not fixed by the change as-is.

```
Exception in thread "main" java.lang.IllegalArgumentException: No dependency graph available to initialize DependencyGraphNavigator.
	at org.ossreviewtoolkit.model.DependencyGraphNavigator.<init>(DependencyGraphNavigator.kt:29)
	at org.ossreviewtoolkit.model.CompatibilityDependencyNavigator$Companion.create(CompatibilityDependencyNavigator.kt:53)
	at org.ossreviewtoolkit.model.OrtResult.createDependencyNavigator(OrtResult.kt:479)
	at org.ossreviewtoolkit.model.OrtResult.access$createDependencyNavigator(OrtResult.kt:45)
	at org.ossreviewtoolkit.model.OrtResult$dependencyNavigator$2.invoke(OrtResult.kt:101)
	at org.ossreviewtoolkit.model.OrtResult$dependencyNavigator$2.invoke(OrtResult.kt:101)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at org.ossreviewtoolkit.model.OrtResult.getDependencyNavigator(OrtResult.kt:101)
	at org.ossreviewtoolkit.model.OrtResult$packages$2.invoke(OrtResult.kt:148)
	at org.ossreviewtoolkit.model.OrtResult$packages$2.invoke(OrtResult.kt:137)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at org.ossreviewtoolkit.model.OrtResult.getPackages(OrtResult.kt:137)
	at org.ossreviewtoolkit.model.OrtResult.getPackage(OrtResult.kt:357)
	at org.ossreviewtoolkit.model.licenses.DefaultLicenseInfoProvider.createConcludedLicenseInfo(DefaultLicenseInfoProvider.kt:53)
	at org.ossreviewtoolkit.model.licenses.DefaultLicenseInfoProvider.createLicenseInfo(DefaultLicenseInfoProvider.kt:47)
	at org.ossreviewtoolkit.model.licenses.DefaultLicenseInfoProvider.get(DefaultLicenseInfoProvider.kt:42)
	at org.ossreviewtoolkit.model.licenses.LicenseInfoResolver.createLicenseInfo(LicenseInfoResolver.kt:70)
	at org.ossreviewtoolkit.model.licenses.LicenseInfoResolver.resolveLicenseInfo(LicenseInfoResolver.kt:61)
	at org.ossreviewtoolkit.helper.common.UtilsKt.processAllCopyrightStatements(Utils.kt:243)
	at org.ossreviewtoolkit.helper.common.UtilsKt.processAllCopyrightStatements$default(Utils.kt:228)
	at org.ossreviewtoolkit.helper.commands.ListCopyrightsCommand.run(ListCopyrightsCommand.kt:95)
	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:204)
	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:213)
	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:17)
	at com.github.ajalt.clikt.core.CliktCommand.parse(CliktCommand.kt:396)
	at com.github.ajalt.clikt.core.CliktCommand.parse$default(CliktCommand.kt:393)
	at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:411)
	at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:436)
	at org.ossreviewtoolkit.helper.HelperMainKt.main(HelperMain.kt:61)
```

fviernau (approved):

PR #4523, opened 2021-09-30
Refactor scancode license texts

mnonnenmacher:
Read the ScanCode license texts from the ScanCode installation instead of importing them to spdx-utils. Please see the commit messages for details.


sschuberth:
> Please have a look and provide feedback, no need for code review until we agree on the approach.

I agree with the approach.

>As an alternative we could also keep the imported texts and read license texts from the ScanCode installation only as a fallback, in case the installed version differs from the version for which we imported the files.

I like the fact that we can get rid of the license resources this way. After all, ScanCode is just one of the scanners that can be used, so I always felt a bit uncomfortable giving its license texts a special role in ORT.

utils/spdx/src/main/kotlin/SpdxUtils.kt
```diff
@@ -171,4 +171,11 @@ private fun getLicenseTextResource(id: String): URL? =
     object {}.javaClass.getResource("/licenserefs/$id")
 
 private fun getLicenseTextFile(id: String, dir: File): File? =
-    dir.resolve(id).takeIf { it.isFile }
+    sequenceOf(
+        id,
+        id.removePrefix("LicenseRef-"),
+        id.replaceFirst(Regex("LicenseRef-\\w*-"), ""),
+        "${id.replaceFirst(Regex("LicenseRef-\\w*-"), "")}.LICENSE"
```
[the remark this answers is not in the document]
sschuberth: You could keep it as a function expression like this:

```
private fun getLicenseTextFile(id: String, dir: File): File? =
    id.replace(Regex("^LicenseRef-\\w+-"), "").let { idWithoutLicenseRefNamespace ->
        sequenceOf(
            id,
            id.removePrefix("LicenseRef-"),
            idWithoutLicenseRefNamespace,
            "$idWithoutLicenseRefNamespace.LICENSE"
        ).mapNotNull { filename ->
            dir.resolve(filename).takeIf { it.isFile }
        }.firstOrNull()
    }
```

cli/src/main/kotlin/commands/ReporterCommand.kt
```diff
@@ -291,8 +295,10 @@ class ReporterCommand : CliktCommand(
         }
 
         val successCount = reportFormats.size - failureCount
-        println("Created $successCount of ${reportFormats.size} report(s) in " +
-                "${reportDurationMap.duration.inWholeSeconds}s.")
+        println(
+            "Created $successCount of ${reportFormats.size} report(s) in " +
+                    "${reportDurationMap.duration.inWholeSeconds}s."
+        )
```
sschuberth, line 301: Nit: Unrelated reformatting.

utils/spdx/src/main/kotlin/SpdxUtils.kt
```diff
@@ -179,3 +180,14 @@ private fun getLicenseTextFile(id: String, dir: File): File? =
     ).mapNotNull { filename ->
         dir.resolve(filename).takeIf { it.isFile }
     }.firstOrNull()
+
+/**
+ * Try to locate the directory which contains the license texts provided by ScanCode. This requires ScanCode to be
+ * installed and the `scancode` binary to be on the path.
+ */
+fun getScanCodeLicenseTextsDir(): File? {
+    val scanCodeDir = getPathFromEnvironment("scancode")?.parentFile?.absoluteFile
```
sschuberth, line 189: I believe we can drop `?.absoluteFile` here, these paths should already be absolute.

utils/spdx/src/main/kotlin/SpdxUtils.kt
```diff
@@ -179,3 +180,14 @@ private fun getLicenseTextFile(id: String, dir: File): File? =
...
+
+    return scanCodeDir?.resolve("src/licensedcode/data/licenses")?.takeIf { it.isDirectory }
+        ?: scanCodeDir?.resolve("../lib/site-packages/licensedcode/data/licenses")?.takeIf { it.isDirectory }
```
sschuberth, line 192: Maybe in preparation for potentially adding more candidate paths do:

    return listOf(
        "src/licensedcode/data/licenses",
        "../lib/site-packages/licensedcode/data/licenses"
    ).mapNotNull { relativePath ->
        scanCodeDir?.resolve(relativePath)?.takeIf { it.isDirectory }
    }.firstOrNull()

.azure-pipelines.yml
```diff
@@ -21,6 +21,7 @@ variables:
   GO_DEP_VERSION: 0.5.4
   PYTHON_PIPENV_VERSION: 2018.11.26
   RUST_VERSION: 1.35.0
+  SCANCODE_VERSION: 3.2.1rc2
```
sschuberth, line 24: Should this be squashed to the previous commit, so really each commit in this PR would pass tests on its own?

sschuberth: Also, is this change really necessary as ScanCode still would get bootstrapped, or?

sschuberth (changes requested):
I stopped the review for now at `spdx-utils: Remove the task to generate license ref text resources`.

.azure-pipelines.yml
mnonnenmacher: > 
> 
> Should this be squashed to the previous commit, so really each commit in this PR would pass tests on its own?

I have instead merged it with a new commit that removes the committed ScanCode license files which I forgot to do previously. FYI, I have also rewritten the commit that introduces the function to locate the ScanCode license dir to automatically do this in the license text functions. Otherwise all callers of those functions would have had to do this. This was actually the reason why this was put on hold until the utils refactoring was done.

mnonnenmacher: > 
> 
> Also, is this change really necessary as ScanCode still would get bootstrapped, or?

Only by the `ScanCodeScannerFunTest`, but this would not help for the license text related tests which do not call the `ScanCode` binary at all.

mnonnenmacher: It seems like detecting the ScanCode dir does not work on CI, I'll have to look into this next week.

utils/spdx/src/main/kotlin/Utils.kt
```diff
@@ -42,6 +43,28 @@ internal val PATH_STRING_COMPARATOR = compareBy<String>({ path -> path.count { i
...
 internal val yamlMapper = YAMLMapper().registerKotlinModule()
 
+/**
+ * The directory that contains the ScanCode license texts. This is located using a heuristic based on the path of the
+ * ScanCode binary.
+ */
+private val scanCodeLicenseTextDir by lazy {
+    val scanCodeDir = Os.getPathFromEnvironment("scancode")?.parentFile
+
+    // Locate directories that contain the python version in their name.
```
sschuberth, line 53: Can we maybe safe us all of this by simply searching for e.g. `apache-2.0.LICENSE` below the parent directory of `scanCodeDir`, and using the parent directory where the file was found?

mnonnenmacher: Yes, this should work as well, but it would be a bit slower. Given that the current algorithm finally works locally and on CI I'd rather merge it as is and do changes in a separate smaller PR.

fviernau:
I've made the following test with 5c54101a56dbe6715ea00afce2e5bd26ff66afd6:

1. Picked an arbitrary scan with 'LicenseRef-scancode-*' texts in the NOTICE file.
2. Re-scanned with 5c54101a56dbe6715ea00afce2e5bd26ff66afd6
3. Checked the NOTICE file for the license texts presence of the license texts and stopped after 3 license IDs as all were missing:
  - LicenseRef-scancode-unicode-mappings
  - LicenseRef-scancode-x11-opengl
  - LicenseRef-scancode-x11-opengroup

utils/spdx/src/main/kotlin/Utils.kt
```diff
@@ -164,3 +187,6 @@ private fun getLicenseTextFile(id: String, dir: File): File? =
             dir.resolve(filename).takeIf { it.isFile }
         }.firstOrNull()
     }
+
+private fun addScanCodeLicenseTextsDir(licenseTextDirectories: List<File>): List<File> =
```
sschuberth, line 191: IMO the need for this helper function is not nice. Could we get rid of it by changing `getLicenseTextReader()` to take varargs instead of a list of files for the directories?

utils/spdx/build.gradle.kts
```diff
@@ -66,8 +65,7 @@ val importScanCodeLicenseTexts by tasks.registering(SvnExport::class) {
     description = "Imports license texts from the ScanCode repository."
     group = "SPDX"
 
-    val versionWithoutHyphen = scancodeVersion.replace("-", "")
-    svnUrl = "https://github.com/nexB/scancode-toolkit/tags/v$versionWithoutHyphen/src/licensedcode/data/licenses"
+    svnUrl = "https://github.com/nexB/scancode-toolkit/trunk/src/licensedcode/data/licenses"
```
sschuberth, line 68: > Since the ScanCode license texts are now only used for the SPDX licenses

Huh? So ScanCode license texts are not also used for ScanCode licenses (licenses only found by ScanCode)?!

utils/spdx/src/main/kotlin/Utils.kt
mnonnenmacher: This would change the signature and we would still have to add the scancode dir in `hasLicenseText` and `getLicenseTextReader`. Would it be ok to improve that outside of this PR?

utils/spdx/build.gradle.kts
mnonnenmacher: Yes, as we do not import LicenseRef-scancode license texts anymore the texts imported from ScanCode are only used for SPDX licenses since the previous commits.

mnonnenmacher:
> 
> 
> I've made the following test with [5c54101](https://github.com/oss-review-toolkit/ort/commit/5c54101a56dbe6715ea00afce2e5bd26ff66afd6):
> 
>     1. Picked an arbitrary scan with 'LicenseRef-scancode-*' texts in the NOTICE file.
> 
>     2. Re-scanned with [5c54101](https://github.com/oss-review-toolkit/ort/commit/5c54101a56dbe6715ea00afce2e5bd26ff66afd6)
> 
>     3. Checked the NOTICE file for the license texts presence of the license texts and stopped after 3 license IDs as all were missing:
> 
> 
>     * LicenseRef-scancode-unicode-mappings
> 
>     * LicenseRef-scancode-x11-opengl
> 
>     * LicenseRef-scancode-x11-opengroup

Thanks for testing, I think the root cause was that in the Docker image the `scancode` binary was a symbolic link. I have changed the code to use the real path now, that should fix it. Could you test again?

utils/spdx/src/main/kotlin/Utils.kt
sschuberth: Yes, we can do that later. Maybe add a `TODO` to this function to remember?

fviernau:
> Thanks for testing, I think the root cause was that in the Docker image the `scancode` binary was a symbolic link. I have changed the code to use the real path now, that should fix it. Could you test again?

Done. This time it looks good. The only difference I found was some trailing white spaces, but that's good. Ok to merge from my side in this regard.

mnonnenmacher:
> 
> 
> > Thanks for testing, I think the root cause was that in the Docker image the `scancode` binary was a symbolic link. I have changed the code to use the real path now, that should fix it. Could you test again?
> 
> @mnonnenmacher -
> 
> Done. This time it looks good. The only difference I found was some trailing white spaces, but that's good. Ok to merge from my side in this regard.

Great, thanks again for testing!

utils/spdx/src/main/kotlin/Utils.kt
sschuberth: Nit: s/python/Python/

utils/spdx/src/main/kotlin/Utils.kt
```diff
@@ -164,3 +188,6 @@ private fun getLicenseTextFile(id: String, dir: File): File? =
...
+    (listOf(scanCodeLicenseTextDir) + licenseTextDirectories).filterNotNull().distinct()
```
sschuberth, line 193: Consider: s/listOf/listOfNotNull/ instead of `filterNotNull`.

.azure-pipelines.yml
sschuberth: Commit message nit: the "previous commit" reference does not seem right anymore.

utils/spdx/build.gradle.kts
```diff
@@ -297,59 +297,12 @@ fun generateLicenseTextResources(description: String, ids: Map<String, LicenseMe
...
-    }
-}
-
 val generateSpdxLicenseEnum by tasks.registering {
     description = "Generates the enum class of SPDX license ids and their associated texts as resources."
     group = "SPDX"
 
     dependsOn(importLicenseTexts)
-    finalizedBy("cleanImportLicenseTexts")
+    finalizedBy("cleanImportLicenseTexts", "cleanImportScanCodeLicenseTexts")
```
sschuberth, line 305: I guess this could be omitted, same below.

utils/spdx/src/test/kotlin/UtilsTest.kt
```diff
@@ -36,7 +36,7 @@ import java.io.File
 import org.ossreviewtoolkit.utils.common.VCS_DIRECTORIES
 import org.ossreviewtoolkit.utils.test.createTestTempDir
 
-class SpdxUtilsTest : WordSpec() {
+class UtilsTest : WordSpec() {
```
sschuberth, line 39: If you want, mention this to be a fixup for my d564080.

sschuberth (approved):

fviernau (approved):
Approve as I tested that it works with our internal setup. I didn't review it in detail.

mnonnenmacher:
Merged, ignoring the unrelated failing test.

PR #4527, opened 2021-09-30
SpdxDocumentFiles: Use `originator` as the author

### LINKED ISSUE #4058: Use `originator` as the "author" in SpdxDocumentFiles
I gave this some though again after the last dev meeting. To me, we should:
- Use `originator` as the "author", and make use of the existing new mechanism to deduce copyright holders from authors.
- In the future, treat `copyrightText` as yet another new potential source of copyright holder information in the context of solving https://github.com/oss-review-toolkit/ort/issues/3780.

_Originally posted by @sschuberth in https://github.com/oss-review-toolkit/ort/pull/3665#discussion_r597455543_

sschuberth:
Resolves #4058.

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -329,7 +329,8 @@ class SpdxDocumentFile(
     private fun SpdxPackage.toIdentifier() =
         Identifier(
             type = managerName,
-            namespace = originator?.withoutPrefix(SpdxConstants.ORGANIZATION).orEmpty(),
+            namespace = listOfNotNull(supplier, originator).firstOrNull()
+                ?.withoutPrefix(SpdxConstants.ORGANIZATION).orEmpty(),
             name = name,
             version = versionInfo
         )
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -263,6 +263,25 @@ private fun String.isExternalDocumentReferenceId(): Boolean = startsWith(SpdxCon
 private fun String.mapNotPresentToEmpty(): String = takeUnless { SpdxConstants.isNotPresent(it) }.orEmpty()
 
 /**
+ * Wrap any "present" SPDX value in a sorted set, or return an empty sorted set otherwise.
+ */
+private fun String?.wrapPresentInSortedSet(): SortedSet<String> {
+    if (SpdxConstants.isPresent(this)) {
+        withoutPrefix(SpdxConstants.PERSON)?.let { persons ->
+            // In case of a person, allow a comma-separated list of persons.
+            return persons.split(',').mapTo(sortedSetOf()) { it.trim() }
+        }
+
+        // Do not split an organization like "Acme, Inc." by comma.
+        withoutPrefix(SpdxConstants.ORGANIZATION)?.let {
+            return sortedSetOf(it)
+        }
+    }
+
+    return sortedSetOf()
+}
+
+/**
  * Return the [PackageLinkage] between [dependency] and [dependant] as specified in [relationships]. If no
  * relationship is found, return [PackageLinkage.DYNAMIC].
  */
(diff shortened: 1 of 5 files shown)


analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -271,6 +271,23 @@ private fun String.isExternalDocumentReferenceId(): Boolean = startsWith(SpdxCon
...
+
+    val persons = withoutPrefix(SpdxConstants.PERSON)?.let { persons ->
+        // In case of a person, allow a comma-separated list of persons.
+        persons.split(',').mapTo(sortedSetOf()) { it.trim() }
+    }
+
+    // Do not split an organization like "Acme, Inc." by comma.
+    val organization = withoutPrefix(SpdxConstants.ORGANIZATION)?.let { sortedSetOf(it) }
+
+    return persons ?: organization ?: sortedSetOf()
```
mnonnenmacher, line 288: What is the reason to ignore the organization if persons were found?

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -271,6 +271,23 @@ private fun String.isExternalDocumentReferenceId(): Boolean = startsWith(SpdxCon
...
+private fun String?.wrapPresentInSortedSet(): SortedSet<String> {
+    if (this == null || SpdxConstants.isNotPresent(this)) return sortedSetOf()
+
+    val persons = withoutPrefix(SpdxConstants.PERSON)?.let { persons ->
+        // In case of a person, allow a comma-separated list of persons.
+        persons.split(',').mapTo(sortedSetOf()) { it.trim() }
+    }
+
+    // Do not split an organization like "Acme, Inc." by comma.
+    val organization = withoutPrefix(SpdxConstants.ORGANIZATION)?.let { sortedSetOf(it) }
```
mnonnenmacher, line 286: Should this use the `extractOrganization()` function? Accordings to the docs is supposed to be used for an "originator" string.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: The [originator](https://spdx.github.io/spdx-spec/package-information/#761-description) is "Single line of text with *one* of the following keywords", so it cannot contain both "Person:" and "Organization:", and it's fine to only accept either one.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: Actually I decided to do somewhat the opposite and remove the `extractOrganization()` function. See the commit message for the rationale.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -329,7 +329,7 @@ class SpdxDocumentFile(
     private fun SpdxPackage.toIdentifier() =
         Identifier(
             type = managerName,
-            namespace = originator?.withoutPrefix(SpdxConstants.ORGANIZATION).orEmpty(),
+            namespace = supplier?.withoutPrefix(SpdxConstants.ORGANIZATION).orEmpty(),
```
mnonnenmacher, line 349: In the commit message it says "more likely to match the supplier than the originator", this sounds like it would make sense to use the originator as a fallback.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
mnonnenmacher: Then you could early return inside the `let` blocks and omit the `val`s.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: Indeed makes sense, done.

mnonnenmacher (approved):

PR #4528, opened 2021-10-01
FossId: Fix the extraction of a repository name from a Git URL

bs-ondem:
This fixes the regex for the extraction of the repository name from
a Git URL, because the old regex could not omit the .git suffix from
the Git URL. Additionally, it adds a new test case for testing the
function convertGitUrlToProjectName with several Git URLs.

### DIFF
--- a/scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
+++ b/scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
@@ -88,6 +88,7 @@ import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.scanner.scanOrtResult
+import org.ossreviewtoolkit.scanner.scanners.fossid.FossId.Companion.convertGitUrlToProjectName
 import org.ossreviewtoolkit.utils.test.shouldNotBeNull
 
 class FossIdTest : WordSpec({
@@ -100,6 +101,14 @@ class FossIdTest : WordSpec({
             val fossId = createFossId(createConfig())
 
             fossId.version shouldBe FOSSID_VERSION
+        }
+    }
+
+    "convertGitUrlToProjectName()" should {
+        "extract the repository name from the Git URL without the .git suffix" {
+            convertGitUrlToProjectName("https://github.com/jshttp/mime-types.git") shouldBe "mime-types"
+            convertGitUrlToProjectName("https://github.com/vdurmont/semver4j.git") shouldBe "semver4j"
+            convertGitUrlToProjectName("https://dev.azure.com/org/project/_git/repo") shouldBe "repo"
         }
     }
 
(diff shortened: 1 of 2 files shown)


scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -103,6 +103,18 @@ class FossIdTest : WordSpec({
         }
     }
 
+    "convertGitUrlToProjectName()" should {
+        "extract the repository name from the git URL without the .git suffix" {
+            val httpGitUrl = FossId.convertGitUrlToProjectName("http://github.com/path/to/example-repo.git")
```
fviernau, line 108: The variable names, e.g. `httpGitUrl` do not hold and URL but the project name as value.
How about inlining the variables? Maybe the assertions fit on one line if you add a static import for the convert function.

fviernau (changes requested):

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -103,6 +103,18 @@ class FossIdTest : WordSpec({
...
+            val httpsGitUrl = FossId.convertGitUrlToProjectName("https://github.com/path/to/example-repo.git")
+            val httpsGitUrlDottedRepoName = FossId.convertGitUrlToProjectName("https://github.com/example.repo.git")
```
sschuberth, line 110: I believe we need different test cases here. Instead of only using GitHub as a host, please use the GitHub URL mentioned in the function docs for `convertGitUrlToProjectName` ("https://github.com/jshttp/mime-types.git"), and additionally a URL to a Git repo hosted on Azure, like "https://dev.azure.com/org/project/_git/repo".

@nnobelis do you have more URLs / hosts in mind that should be tested?

sschuberth (changes requested):

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
nnobelis: These are the two main ones. Maybe one without `-` such as `https://github.com/vdurmont/semver4j.git`

BTW this is `https://dev.azure.com/example/_git/project` so the project name is what is coming after the `_git`.

sschuberth: > BTW this is `https://dev.azure.com/example/_git/project` so the project name is what is coming after the `_git`.

Well, we're using the (Azure) repo name as the (ORT) project name. The Azure project name, however, indeed is the part *before* "_git".

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -103,6 +104,14 @@ class FossIdTest : WordSpec({
         }
     }
 
+    "convertGitUrlToProjectName()" should {
+        "extract the repository name from the git URL without the .git suffix" {
```
sschuberth, line 108: Nit: s/git/Git/

sschuberth (dismissed):

sschuberth:
@bs-ondem could you please have a look why the FossId tests seem to fail? (The *ConanFunTest* is expected to fail currently.)

fviernau (approved):

bs-ondem:
> @bs-ondem could you please have a look why the FossId tests seem to fail? (The _ConanFunTest_ is expected to fail currently.)

@sschuberth Oh sorry, i had to remove .git suffix from `PROJECT` constant.

sschuberth (approved):

sschuberth:

PR #4535, opened 2021-10-04
EvaluatorCommand: Summary improvements

mnonnenmacher:
Improve the rule violation summary printed by the evaluator command. See the commit messages for details.

### DIFF
--- a/cli/src/main/kotlin/commands/EvaluatorCommand.kt
+++ b/cli/src/main/kotlin/commands/EvaluatorCommand.kt
@@ -55,6 +55,7 @@ import org.ossreviewtoolkit.cli.utils.readOrtResult
 import org.ossreviewtoolkit.cli.utils.writeOrtResult
 import org.ossreviewtoolkit.evaluator.Evaluator
 import org.ossreviewtoolkit.model.FileFormat
+import org.ossreviewtoolkit.model.RuleViolation
 import org.ossreviewtoolkit.model.config.CopyrightGarbage
 import org.ossreviewtoolkit.model.config.LicenseFilenamePatterns
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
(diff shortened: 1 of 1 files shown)


sschuberth (dismissed):

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -289,10 +289,8 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
 
         log.perf { "Executed the evaluator in ${duration.inWholeMilliseconds}ms." }
 
-        if (log.delegate.isErrorEnabled) {
-            evaluatorRun.violations.forEach { violation ->
-                log.error { violation.format() }
-            }
+        evaluatorRun.violations.forEach { violation ->
+            log.log(violation.severity.toLog4jLevel()) { violation.format() }
```
fviernau, line 293: I believe previous `error` log level didn't make sense, because I would regard a violation an error. 
It's just the output of a successfuly evaluator run. From that perspective I believe the same log level should be used for all violations, as otherwise one can easily get the false impression of viewing the whole summary instead of just the filtered one. 

What do you think about using log.info for all violations?

fviernau, line 293: I believe previous error log level didn't make sense, because I would regard a violation an error.
It's just the output of a successfuly evaluator run. From that perspective I believe the same log level should be used for all violations, as otherwise one can easily get the false impression of viewing the whole summary instead of just the filtered one.

What do you think about using log.info for all violations?

fviernau (changes requested):

cli/src/main/kotlin/commands/EvaluatorCommand.kt
mnonnenmacher: I agree, we could also go one step further and print the list directly to the console?

fviernau: would work for me.

mnonnenmacher: Done.

fviernau (approved):

Issue #2037, opened 2019-12-10
Conan: Get real provenance data


fviernau:
The VCS location of several Conan dependencies point to repositories which contain only recipes used to publish these packages on Conan Central, but not the actual source code of the packages.
As result the actual sources are not being scanned, see e.g. [1,2,3].

[1] https://bintray.com/conan/conan-center
[2] https://bincrafters.readthedocs.io/en/latest/
[3] https://github.com/bincrafters/conan-protobuf

sschuberth:
Some news about ConanCenter (not "Central"): https://blog.conan.io/2020/01/13/New-Conan-Center-Improves-Search-Discovery.html

sschuberth:
See the [Recipe and Sources in a Different Repo](https://docs.conan.io/en/latest/creating_packages/external_repo.html) docs for details.

PR #4541, opened 2021-10-04
Conan improvements

mnonnenmacher:
Various improvements for the Conan package manager:
* Fix the dependency tree.
* Improve performance.
* Fix resolving provenance information (#2037).


analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -272,8 +275,12 @@ class Conan(
     /**
      * Return the value `conan inspect` reports for the given [field].
      */
-    private fun runInspectRawField(pkgName: String, workingDir: File, field: String): String =
-        run(workingDir, "inspect", pkgName, "--raw", field).stdout
+    private fun inspectField(pkgName: String, workingDir: File, field: String): String =
+        pkgInspectResults.getOrPut(pkgName) {
+            val jsonFile = createOrtTempFile(managerName)
+            run(workingDir, "inspect", pkgName, "--json", jsonFile.absolutePath).requireSuccess()
+            jsonMapper.readTree(jsonFile).also { jsonFile.delete() }
```
sschuberth, line 280: Nit: Should we do `jsonFile.parentFile.safeDeleteRecursively(force = true)` instead, to delete the whole temporary directory?

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -272,8 +275,12 @@ class Conan(
...
+        }[field].textValueOrEmpty()
```
sschuberth, line 281: Nit: IMO this mix of curly braces and square brackets does not read so nice. Maybe use `.get(field)` instead? (I've verified that the IDE does not hint to use `[]` in this case.)

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -283,15 +276,19 @@ class Conan(
...
-    private fun removeProjectPackage(rootNode: JsonNode, definitionFile: File): List<JsonNode> =
-        rootNode.find {
+    private fun findProjectNode(pkgInfos: JsonNode, definitionFileName: String): JsonNode =
+        pkgInfos.first {
             // Contains because conanfile.py's reference string often includes other data.
-            it["reference"].textValueOrEmpty().contains(definitionFile.name)
-        }?.let { projectPackage ->
-            rootNode.minusElement(projectPackage)
-        } ?: rootNode.toList<JsonNode>()
+            it["reference"].textValueOrEmpty().contains(definitionFileName)
```
sschuberth, line 289: Nit: Maybe replace `contains` with `in` while at it? (And eventually adjust the comment.)

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -262,7 +262,7 @@ class Conan(
             homepageUrl = node["homepage"].textValueOrEmpty(),
             binaryArtifact = RemoteArtifact.EMPTY, // TODO: implement me!
             sourceArtifact = RemoteArtifact.EMPTY, // TODO: implement me!
-            vcs = parseVcsInfo(node)
+            vcs = processPackageVcs(VcsInfo.EMPTY, node["homepage"].textValueOrEmpty())
```
sschuberth, line 269: Nit: Can we extract this to a variable and also use it in line 262?

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -325,6 +330,28 @@ class Conan(
     private fun parsePackageField(node: JsonNode, workingDir: File, field: String): String =
         inspectField(node["display_name"].textValue(), workingDir, field)
 
+    /**
+     * Try to read the source artifact from the [conanStoragePath], if not possible return [RemoteArtifact.EMPTY].
+     */
+    private fun parseSourceArtifact(id: Identifier, conanStoragePath: File): RemoteArtifact {
+        val conanDataFile = conanStoragePath.resolve("${id.name}/${id.version}/_/_/export/conandata.yml")
+
+        return kotlin.runCatching {
```
sschuberth, line 339: Nit: Drop the `kotlin.` prefix.

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -325,6 +330,28 @@ class Conan(
...
+            val conanData = yamlMapper.readTree(conanDataFile)
+            val artifactEntry = conanData["sources"][id.version]
+
+            val url = artifactEntry["url"].let {
+                if (it.isTextual) it.textValueOrEmpty() else it.first().textValueOrEmpty()
```
sschuberth, line 344: Alternatively: `(it.takeIf { it.isTextual } ?: it.first()).textValueOrEmpty()`

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -325,6 +330,28 @@ class Conan(
...
+            }
+            val hash = artifactEntry["sha256"].textValueOrEmpty().takeIf { it.isNotBlank() }
+                ?.let { Hash(it, HashAlgorithm.SHA256) } ?: Hash.NONE
```
sschuberth, line 347: Could be just

    val  hash = Hash.create(artifactEntry["sha256"].textValueOrEmpty())

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -205,49 +205,42 @@ class Conan(
     }
 
     /**
-     * Return the dependency tree starting from a [rootNode] for given [scopeName].
+     * Return the dependency tree starting from a [pkgInfos] for given [scopeName].
      */
     private fun parseDependencyTree(
-        rootNode: JsonNode,
-        workingDir: File,
+        pkgInfos: JsonNode,
```
fviernau, line 213: "from a [pkgInfos]" reads a bit odd in terms of plural vs. singular. Looking at the code here and also in the caller to me singular form reads a bit better. What do you think?

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -283,16 +276,20 @@ class Conan(
             jsonMapper.readTree(jsonFile).also { jsonFile.parentFile.safeDeleteRecursively(force = true) }
         }.get(field).textValueOrEmpty()
 
+    /**
+     * Find the node that represents the project defined in the definition file.
+     */
+    private fun findProjectNode(pkgInfos: JsonNode, definitionFileName: String): JsonNode =
+        pkgInfos.first {
+            // Us "in" because conanfile.py's reference string often includes other data.
```
fviernau, line 290: Us -> Us**e**

fviernau (changes requested):

fviernau:
Out of scope: Seeing these VCS URLs pointing to Conan index to me raises the question to me whether it is ok to return VCS URLs which do not point to the source code of the artifact. Actually I believe the answer to that should be a clear NO, as it creates the false impression that sources have been scanned and is inconsistent with ORTs approach. How about making the URLs in a following task? Maybe @oss-review-toolkit/core-devs ?

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -49,7 +49,9 @@ import org.ossreviewtoolkit.model.jsonMapper
 import org.ossreviewtoolkit.utils.CommandLineTool
 import org.ossreviewtoolkit.utils.Os
 import org.ossreviewtoolkit.utils.ProcessCapture
+import org.ossreviewtoolkit.utils.createOrtTempFile
```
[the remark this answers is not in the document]
sschuberth: The command *does* support JSON output, just not to *stdout*, but only to a file. So maybe "does not support to output the JSON *to stdout*" (not even with `--json -`, I've checked).

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -81,6 +83,8 @@ class Conan(
         ) = Conan(managerName, analysisRoot, analyzerConfig, repoConfig)
     }
 
+    private val pkgInspectResults = mutableMapOf<String, JsonNode>()
```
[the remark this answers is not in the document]
sschuberth: Hint: *If* we want to do this, it could be implemented in `afterResolution()`.

analyzer/src/funTest/assets/projects/synthetic/conan-expected-output-txt.yml
```diff
@@ -117,6 +118,35 @@ packages:
...
+      value: ""
+      algorithm: ""
+  source_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  vcs:
+    type: "Git"
+    url: "https://github.com/conan-io/conan-center-index"
```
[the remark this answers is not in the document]
sschuberth: No, they're not, and they're being removed in the second-last commit.

sschuberth:
> Seeing these VCS URLs pointing to Conan index

Note that after the last commit in this PR, no VCS URL points to Conan index anymore.

analyzer/src/main/kotlin/managers/Conan.kt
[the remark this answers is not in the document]
fviernau: Shouldn't this be done with a try / finally? (and thus really be put into `resolveDependencies` e.g. like in `CocoaPods`?

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -243,35 +242,18 @@ class Conan(
         rootNode: JsonNode,
         scopeName: String,
         workingDir: File
-    ): SortedSet<PackageReference> {
-        val stack = Stack<JsonNode>().apply { addAll(rootNode) }
```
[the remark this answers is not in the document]
mnonnenmacher: I think the commit message already makes it clear that the stacks were useless.

analyzer/src/main/kotlin/managers/Conan.kt
mnonnenmacher: I have rewritten the function docs. For the variable I prefer plural because it contains information about all packages.

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -99,6 +105,16 @@ class Conan(
      * Primary method for resolving dependencies from [definitionFile].
      */
     override fun resolveDependencies(definitionFile: File): List<ProjectAnalyzerResult> {
+        return try {
```
fviernau, line 108: Consider making it an expression function.

fviernau (approved):

sschuberth (approved):

PR #4552, opened 2021-10-06
Environment: Add `PATH` to the relevant variables

fviernau:
Show the value of the `PATH` variable as part of the version header of
the cli, see also `getVersionHeader()` in `OrtMain.kt`. This can be
useful for debugging.

### DIFF
--- a/utils/src/main/kotlin/Environment.kt
+++ b/utils/src/main/kotlin/Environment.kt
@@ -86,6 +86,7 @@ data class Environment(
             "http_proxy",
             "https_proxy",
             "JAVA_HOME",
+            "PATH",
             "ANDROID_HOME",
             "GOPATH"
         )


sschuberth:
I deliberately didn't add `PATH` to begin with, as 1) it might become quite long, and 2) it might expose details about the system that users don't want to share (e.g. when copying & pasting ORT debug output to issue reports).

Any thoughts on that @oss-review-toolkit/core-devs?

mnonnenmacher:
> I deliberately didn't add `PATH` to begin with, as 1) it might become quite long, and 2) it might expose details about the system that users don't want to share (e.g. when copying & pasting ORT debug output to issue reports).
> 
> Any thoughts on that @oss-review-toolkit/core-devs?

On my system the path is 4 lines long, I wouldn't want ORT to print this everytime. It's also easy to just add an `echo $PATH` somewhere while testing changes on Docker/CI setups. Another option would be to change the `requirements` command to list not only the version of detected tools but also their path, and to add Java there.

PR #4553, opened 2021-10-06
Move `allowMovingRevisions` to the `DownloaderConfiguration`

sschuberth:
This way the option can not only be configured when running the
`DownloaderCommand` directly, but takes also effect when the downloader
is invoked by the scanner.


fviernau:
@sschuberth  Are you planning to merge this soon, so we can internally start aligning with it?

PR #4570, opened 2021-10-08
Add an advisor implementation to retrieve defects from GitHub

oheger-bosch:
Advisor: Add an implementation to retrieve defects from GitHub

This advisor can handle packages whose source code is hosted on
GitHub. It uses the GitHub GraphQL client to query information about
the issues reported for these packages.

### DIFF
--- a/advisor/build.gradle.kts
+++ b/advisor/build.gradle.kts
@@ -19,6 +19,7 @@
  */
 
 val kotlinxCoroutinesVersion: String by project
+val mockkVersion: String by project
 val wiremockVersion: String by project
 
 plugins {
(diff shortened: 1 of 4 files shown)


tsteenbe:
@oheger-bosch The commit message states "This advisor can handle packages whose source code is hosted on
GitHub" but should this advisor be able to retrieve defects from GitHub for given package id regardless of where their source code is hosted?

tsteenbe:
@oheger-bosch  I know we added `Defect`class was added to ORT  what I was trying to say is that it unclear to me whay consider a `Defect` in GitHub. Reading the code in this PR it looks to me like you consider every issue associated with a release a defect which is weird to me as issue as may implement new features.

PR #4571, opened 2021-10-08
VersionControlSystem: Fix an intended `break` which is `continue`

fviernau:
Previously, the loop always iterated over all revision candidates, no
matter whether they are successful or not and it seems that this was
by accident. If a package specifies a fixed revision in its meta-data,
that revision should take precendence over any revision guessed from
the package version or Git tags. So, always use the first successful
revision candidate.

Note that ORT's downloader in fact preferred the guessed revisions over
the provided fixed revision for projects (ORT speak), which always have
a correct fixed revision in their meta-data (disregarding the case of a
history re-write happening after the analysis but before the downloader
or scanner run repsectively). That issue is now fixed.

### DIFF
--- a/downloader/src/main/kotlin/VersionControlSystem.kt
+++ b/downloader/src/main/kotlin/VersionControlSystem.kt
@@ -219,10 +219,10 @@ abstract class VersionControlSystem {
         val revisionCandidates = getRevisionCandidates(workingTree, pkg, allowMovingRevisions)
         val results = mutableListOf<Result<String>>()
 
-        revisionCandidates.forEachIndexed { index, revision ->
+        for ((index, revision) in revisionCandidates.withIndex()) {
             log.info { "Trying revision candidate '$revision' (${index + 1} of ${revisionCandidates.size})..." }
             results += updateWorkingTree(workingTree, revision, pkg.vcsProcessed.path, recursive)
-            if (results.last().isSuccess) return@forEachIndexed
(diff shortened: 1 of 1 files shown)


sschuberth:
> Why "don't you think that's correct"?

Like I wrote, because "the last successful candidate was choosen from the list" is something different than "the last element in the list is checked whether it was successful (in each iteration)".

fviernau:
> The last statement in the original loop is `if (results.last().isSuccess) return@forEachIndexed`, which erroneously works like a `continue` (whereas it should have been a `break`). As a last statement in a loop a `continue` is a no-op. So the loop is always over _all_ revision candidates, whether they were successful or not. So if the last element in the list of candidates was _not_ successful, it would still be chosen. That's why I believe saying " the last _successful_ candidate is returned" is incorrect.

Ok, now I got it. Thanks.

fviernau:
> Thank _you_ for finding and fixing it! Quite an embarrassing mistake on my side (I believe it was me who introduced this code). And also a bit disgraceful that we seem to have no test coverage for this case...

Haven't done the Git blame. I'm surprised that this bug hasn't been discovered earlier. I wouldn't expect it to affect packages but for project it can mean that it's scanning an outdated commit instead of a recent one. So teams making changes to `ort.yml` should not that their changes do not take any effect. So, maybe the use case happens more rarely than my gut feeling suggests.

PR #4573, opened 2021-10-08
Make use of Kotlin 1.5.30's better `toString()` for `Duration`

sschuberth:
See [1].

[1]: https://blog.jetbrains.com/kotlin/2021/08/kotlin-1-5-30-released/#stdlib


fviernau:
> @fviernau does this somehow affect your performance log parsing or so?

Probably yes, I'm even though ok doing this change.

PR #4583, opened 2021-10-12
LocalScanner: Fix wrong property names in documentation

oheger-bosch:
In the documentation of the getScannerCriteria() method, wrong
property names are referenced to overwrite the properties of a
ScannerCriteria object in the configuration. Make the documentation
reflect reality.


sschuberth:
> Is there library support to interpret Ivy version ranges?

The Semver4j library we already use can do that... see the version ranges defined for various analyzer tools.

PR #4591, opened 2021-10-13
License: Use the HTTPS link to the Apache 2.0 license in the template

MarcelBochtler:
As it is done in the source files.


MarcelBochtler:
> This needs to update a bunch of tests that use the `LICENSE` file's hash, though.

Thanks I'll fix them.

PR #4593, opened 2021-10-14
OrtResultRule: Use default values for the how-to-fix strings

fviernau:
There are use cases where the error message is very clear, so that no
addition how-to-fix instructions are not needed.


sschuberth:
I guess the reason to *not* have default values here in the first place was to encourage writing "how to fix"-hints. Don't you think the setting of empty defaults will now discourage people to write "how to fix"-hints?

PR #4639, opened 2021-10-29
Add JSON schema for .ort.yml configuration

MarcelBochtler:
A JSON schema [1] is a description of a JSON document, which can also be
used for YAML files. This schema can be used to enable auto-completion
in multiple editors for the .ort.yml file [2]. E.g. IntelliJ IDEA has build-in support
for it [3]:
Screenshot 2021-10-29 at 11 21 44

As a next step, the URL to this schema can be contributed to https://github.com/schemastore/schemastore/ and many editors will then have autocompletion for .ort.yml out of the box.

It would be nice to autogenerate this file, however I didn't find a convenient way to do this.
Adding the test coverage to avoid regressions in the schema should be good enough for now.

Missing as of now is the documentation from the model classes. It would be nice to also get this documentation during auto completion.

I opted to use the JSON format instead of a YAML format, to ensure better compatibility. I was e.g. not able to use a YAML schema with VS Code.

[1]: https://json-schema.org
[2]: https://www.schemastore.org/json
[3]: https://www.jetbrains.com/help/ruby/yaml.html#remote_json

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -60,6 +60,10 @@ Files: clients/github-graphql/src/main/assets/*
 Copyright: 2021 Bosch.IO GmbH <osm@bosch.com>
 License: Apache-2.0
 
+Files: integrations/schemas/*
+Copyright: 2021 Bosch.IO GmbH <osm@bosch.com>
+License: Apache-2.0
+
 # Third-party files.
 
 Files: gradlew*
--- a/examples/curations.ort.yml
+++ b/examples/curations.ort.yml
@@ -1,0 +1,25 @@
+curations:
+  license_findings:
+  - path: "src/**.cpp"
+    start_lines: "3,8"
+    line_count: 1
+    detected_license: "GPL-2.0-only"
+    reason: "CODE"
+    comment: "The scanner matches a variable named `gpl`."
+    concluded_license: "Apache-2.0"
+  packages:
+  - id: "Maven:com.example:dummy:0.0.1"
+    curations:
+      comment: "An explanation why the curation is needed."
+      homepage_url: "https://example.com"
+      is_modified: false
+      vcs:
+        type: "Git"
+        url: "https://example.com/package.git"
+        revision: "main"
+      authors:
+      - "Name of the author"
+      is_meta_data_only: false
+      concluded_license: "BSD-2-Clause"
+      declared_license_mapping:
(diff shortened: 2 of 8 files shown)


MarcelBochtler:
> > It would be nice to autogenerate this file, however I didn't find a convenient way to do this.
> 
> Would you mind listing the tool you looked at? For example, did you try https://github.com/FasterXML/jackson-module-jsonSchema?

I tried https://github.com/mbknor/mbknor-jackson-jsonSchema.
This has some nice options to manipulate the generated schema, which might be handy to have. However, it has issues with the `required` option. [It uses nullable types](https://github.com/mbknor/mbknor-jackson-jsonSchema#code---using-java) to find the required parameters. As in ORT's models the fields are not nullable they are all generated as required.
I didn't find a way to work around this issue. (manually setting the parameters as not required [does also not work](https://github.com/mbknor/mbknor-jackson-jsonSchema/issues/142))

I also tried your linked https://github.com/FasterXML/jackson-module-jsonSchema.
Unfortunately the generate schema was not valid, or at least it was not usable within VS Code:
```
Problems loading reference 'urn:jsonschema/%3Aorg%3Aossreviewtoolkit%3Autils%3Aspdx%3Amodel%3ALicenseChoice': Unable to load schema from 'urn:jsonschema/%3Aorg%3Aossreviewtoolkit%3Autils%3Aspdx%3Amodel%3ALicenseChoice': No content.
```
And the `required` option is also not controllable and does not generate the correct scheme for the `RepositoryConfiguration`.

.reuse/dep5
```diff
@@ -60,6 +60,10 @@ Files: clients/github-graphql/src/main/assets/*
 Copyright: 2021 Bosch.IO GmbH <osm@bosch.com>
 License: Apache-2.0
 
+Files: integrations/schemas/*
```
sschuberth, line 63: Commit message nits:
- I think it's fine / better to drop the "JSON-schema" prefix; we usually use these prefixes to denote which *existing* code is touched, but here the code does not exist yet. And the part after ":" also talks about "JSON schema".
- - [3] links to RubyMine docs. Probably https://www.jetbrains.com/help/idea/json.html#ws_json_schema_add_custom is better.

sschuberth (changes requested):

sschuberth (approved):

PR #4640, opened 2021-10-29
Separate finding files from analysis

sschuberth:

### DIFF
--- a/analyzer/src/funTest/kotlin/GitRepoFunTest.kt
+++ b/analyzer/src/funTest/kotlin/GitRepoFunTest.kt
@@ -65,7 +65,10 @@ class GitRepoFunTest : StringSpec() {
     init {
         // Disabled on Azure Windows because it fails for unknown reasons.
         "Analyzer correctly reports VcsInfo for git-repo projects".config(enabled = !Ci.isAzureWindows) {
-            val ortResult = Analyzer(DEFAULT_ANALYZER_CONFIGURATION).analyze(outputDir)
+            val ortResult = Analyzer(DEFAULT_ANALYZER_CONFIGURATION).run {
+                analyze(findManagedFiles(outputDir))
+            }
+
             val actualResult = ortResult.withResolvedScopes().toYaml()
             val expectedResult = patchExpectedResult(
                 File("src/funTest/assets/projects/external/git-repo-expected-output.yml"),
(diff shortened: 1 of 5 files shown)


cli/src/main/kotlin/utils/SeverityStats.kt
```diff
@@ -77,17 +77,17 @@ internal class SeverityStats(
         val resolvedErrorCount = getResolvedCount(Severity.ERROR)
 
         println(
-            "Found $resolvedErrorCount resolved error(s), $resolvedWarningCount resolved warning(s), " +
-                    "$resolvedHintCount resolved hint(s)."
+            "Resolved issues: $resolvedErrorCount error(s), $resolvedWarningCount warning(s), " +
```
mnonnenmacher, line 80: I was avoiding the word "issue" here because this is also used by the evaluator where it relates to rule violations, not issues. I think we could have to give the `SeverityStats` some property to set the source of the severerities to be able to print the correct message here.

mnonnenmacher (changes requested):

cli/src/main/kotlin/utils/SeverityStats.kt
sschuberth: Good point. That reminds me of a [discussion](https://github.com/oss-review-toolkit/ort/pull/4454#discussion_r707358332) we had (and it's also an item on the dev meeting's [agenda](https://github.com/oss-review-toolkit/ort/wiki/Developer-Meeting#agenda)) whether the `severeIssueThreshold` should even apply to violations, as violations are not issues (which is something I tend to forget myself), and while the severity of issues in hard-coded in ORT, the severity of violations is user-defined by the rules.

So, how about addressing this by making `severeIssueThreshold` *not* to be used when the evaluator prints its stats?

sschuberth:
I've dropped the last two commits for now as I plan to rework them based on the [discussion above](https://github.com/oss-review-toolkit/ort/pull/4640#discussion_r740140267).

mnonnenmacher (approved):

cli/src/main/kotlin/utils/SeverityStats.kt
mnonnenmacher: I think it would be better to have a separate `severeRuleViolationThreshold`, but let's wait for the outcome of the dicussion in the dev meeting.

sschuberth: Let's continue the discussion at https://github.com/oss-review-toolkit/ort/issues/4642.

PR #4647, opened 2021-11-02
Start splitting :utils:core

mnonnenmacher:
Start to split `:utils:core` into two separate modules by creating the `:utils:common` module and moving the first code there. This will be finished in another PR. Please see the commit messages for details.


analyzer/src/funTest/kotlin/managers/CocoaPodsFunTest.kt
```diff
@@ -27,7 +27,7 @@ import java.time.Instant
 
 import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.ProjectAnalyzerResult
-import org.ossreviewtoolkit.utils.core.Os
+import org.ossreviewtoolkit.utils.common.Os
```
sschuberth, line 30: Commit message: Maybe say that the log statement was dropped for simplicity as the log helper is not available, and because the (eventually fixuped) user home it simplicitly dumped as part of printing `ORT_DATA_DIR`.

utils/common/build.gradle.kts
```diff
@@ -17,7 +17,13 @@
...
+val springCoreVersion: String by project
+
 plugins {
     // Apply core plugins.
     `java-library`
 }
+
+dependencies {
+    implementation("org.springframework:spring-core:$springCoreVersion")
+}
```
sschuberth, line 29: This should in turn be removed from `utils/core/build.gradle.kts`.

sschuberth (changes requested):

sschuberth (approved):

PR #4649, opened 2021-11-02
Pub: Create ORT issue if pubspec.yaml is missing for a package

MarcelBochtler:
Instead of throwing an IllegalArgumentException as the yamlMapper does
not handle nullable files.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -44,6 +44,7 @@ import org.ossreviewtoolkit.analyzer.PackageManager
 import org.ossreviewtoolkit.analyzer.parseAuthorString
 import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.downloader.VersionControlSystem
+import org.ossreviewtoolkit.model.EMPTY_JSON_NODE
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtIssue
 import org.ossreviewtoolkit.model.Package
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -578,6 +578,16 @@ class Pub(
...
     private fun readPackageInfoFromCache(packageInfo: JsonNode): JsonNode {
         val definitionFile = reader.findFile(packageInfo, "pubspec.yaml")
+        if (definitionFile == null) {
+            createAndLogIssue(
+                source = managerName,
+                message = "Could not find 'pubspec.yaml' for '${packageInfo["name"].textValueOrEmpty()}'.",
+                severity = Severity.WARNING
+            )
+
+            return yamlMapper.createObjectNode()
```
sschuberth, line 588: Maybe use our `EMPTY_JSON_NODE` instead?

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -578,6 +578,16 @@ class Pub(
 
     private fun readPackageInfoFromCache(packageInfo: JsonNode): JsonNode {
         val definitionFile = reader.findFile(packageInfo, "pubspec.yaml")
+        if (definitionFile == null) {
+            createAndLogIssue(
+                source = managerName,
+                message = "Could not find 'pubspec.yaml' for '${packageInfo["name"].textValueOrEmpty()}'.",
```
sschuberth, line 584: Does it make sense to move `"pubspec.yaml"` to a variable now, and also use it above?

MarcelBochtler: How about a `const` as this string is being used multiple times in this class?

sschuberth: > How about a `const` as this string is being used multiple times in this class?

:+1:

sschuberth: Now you still need to use that constant ;-)

MarcelBochtler: Done. Sorry for that :)

mnonnenmacher (approved):

PR #4652, opened 2021-11-03
Move more code from :utils:core to :utils:common

mnonnenmacher:
Continue moving code from `:utils:core` to `:utils:common` which was started in #4647. Also add a suffix to the names of the utility modules to avoid misunderstandings. See the commit messages for details.


settings.gradle.kts
```diff
@@ -69,6 +69,11 @@ include(":utils:core")
 include(":utils:spdx")
 include(":utils:test")
 
+project(":utils:common").name = "common-utils"
+project(":utils:core").name = "core-utils"
+project(":utils:spdx").name = "spdx-utils"
+project(":utils:test").name = "test-utils"
+
```
sschuberth, line 76: Not sure if it's worth it, but to group inclusion with setting the project name we could do something like

```
listOf("common", "core", "spdx", "test").forEach {
    val path = ":utils:$it"
    include(path)
    project(path).name = "$it-utils"
}
```

(And then maybe use similar code for the clients.)

utils/common/src/main/kotlin/ProcessCapture.kt
```diff
@@ -17,13 +17,17 @@
...
+package org.ossreviewtoolkit.utils.common
 
 import java.io.File
 import java.io.IOException
 
 import kotlin.io.path.createTempDirectory
 
+import org.apache.logging.log4j.kotlin.loggerOf
+
+private val log = loggerOf(ProcessCapture::class.java)
```
sschuberth, line 29: Please use `cachedLoggerOf` (which was recently added to Log4j's Kotlin wrapper).

utils/spdx/build.gradle.kts
```diff
@@ -53,6 +53,8 @@ tasks.withType<KotlinCompile>().configureEach {
 dependencies {
     antlr("org.antlr:antlr4:$antlrVersion")
 
+    implementation(project(":utils:common-utils"))
```
sschuberth, line 56: Commit message typo in "duplicaetd".

utils/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -51,20 +42,6 @@ fun <T> Collection<T>.getDuplicates(): Set<T> = getDuplicates { it }
...
-        // Note that we cannot use exists() to check beforehand whether a symbolic link exists to avoid a
-        // NoSuchFileException to be thrown as it returns "false" e.g. for dangling Windows junctions.
-        Files.readAttributes(toPath(), BasicFileAttributes::class.java, LinkOption.NOFOLLOW_LINKS).let {
-            it.isSymbolicLink || (isWindows && it.isOther)
-        }
-    }.getOrDefault(false)
-
 /**
  * Create an [SpdxExpression] by concatenating [this][SpdxLicense] and [other] using [SpdxOperator.AND].
  */
```
sschuberth, line 47: `fun String.titlecase()` can also be removed.

utils/common/src/main/kotlin/Utils.kt
```diff
@@ -0,0 +1,116 @@
...
+        prefix.commonPrefixWith(path)
+    }?.let {
+        val commonPrefix = File(it)
+        if (commonPrefix.isDirectory) commonPrefix else commonPrefix.parentFile
+    }
+
+/**
+ * Return the full path to the given executable file if it is in the system's PATH environment, or null otherwise.
+ */
+fun getPathFromEnvironment(executable: String): File? {
```
sschuberth, line 56: Did we want to move this (and `resolveWindowsExecutable`) to inside `Os`?

sschuberth (changes requested):

utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -0,0 +1,417 @@
...
+
+            val targetDir = targetPath.resolve(sourcePath.relativize(dir))
+            targetDir.toFile().safeMkdirs()
+
+            return FileVisitResult.CONTINUE
+        }
+
+        override fun visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult {
+            val targetFile = targetPath.resolve(sourcePath.relativize(file))
+            Files.copy(file, targetFile, *copyOptions)
```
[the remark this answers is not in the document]
mnonnenmacher: @sonatype-lift ignore

settings.gradle.kts
mnonnenmacher: I thought about it as well, but decided not to do something like this because the current way is simple and expressive and not much to write anyway.

utils/common/src/main/kotlin/Utils.kt
mnonnenmacher: Yes, I wanted to do this in the next PR because there were some open questions which I wanted to discuss with you first.

sschuberth (approved):

PR #4653, opened 2021-11-03
Apply VCS curation to `vcsProcessed` instead of to `vcs`

### LINKED ISSUE #4618: Applying package curations may lead to loosing data from VCS processed
`PackageCurationData.apply()` applies the VCS curations to `Package.vcs`, not `Package.vcsProcessed` [1]. So, the `vcsProcessed` gets set to `vcs.normalize()` [2]. So, if `vcsProcessed` was created in any non-default way, the information it represents always gets lost. For example several package managers incorporate the fallback URLs only into `vcsProcessed`, see e.g. [3]. So, a VCS curation for Maven artifacts always lead to loosing the the VCS info if it was created from the fallback URL.

**Example**

package:
```
- package:
      id: "Maven:org.aspectj:aspectjrt:1.9.6"
      ... 
       vcs:
          type: ""
          url: ""
          revision: ""
          path: ""
        vcs_processed:
          type: "Git"
          url: "https://github.com/eclipse/org.aspectj.git"
          revision: ""
          path: ""
      curations: []
```
package curation:
```
- id: "Maven:org.aspectj:aspectjrt"
  curations:
    comment: "Package resides in its own directory within repo."
    vcs:
      path: "runtime"
```
Applying above curation to above package leads to a package with:
```
       vcs:
          type: ""
          url: ""
          revision: ""
          path: "runtime"
        vcs_processed:
          type: ""
          url: ""
          revision: "runtime"
          path: ""
```

[1] https://github.com/oss-review-toolkit/ort/blob/971292669a7820e1d749dc24582e0c0aca76649d/model/src/main/kotlin/PackageCurationData.kt#L105-L116
[2] https://github.com/oss-review-toolkit/ort/blob/971292669a7820e1d749dc24582e0c0aca76649d/model/src/main/kotlin/Package.kt#L108
[3] https://github.com/oss-review-toolkit/ort/blob/971292669a7820e1d749dc24582e0c0aca76649d/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt#L647-L649

fviernau:
When a package lacks VCS meta-data several of ORT's package manager
integrations use fallback URLs to guess the VCS location. The result of
that guess ends up in `Package.vcsProcess` but not in `Package.vcs`.

Applying a VCS package curation works by patching `Package.vcs` and
setting `Package.vcsProcessed` to `Package.vcs.normalize()`. So, the
result of the above mentioned guess based on the fallback URLs gets
discarded.

For example, applying a VCS path curation to a package with empty `vcs`
but non-empty `vcsProcessed` yields a package wich has empty VCS info
(besides the path) for both, `vcs` and `vcsProcessed`.

Avoid loosing the guessed VCS by applying the curation to `vcsProcessed`
instead of to `vcs`. Note that only the former is considered by the
downloader or scanner respectively.

Apply the normalization on top of the curation to make this change
non-breaking, even though it might make sense to drop the normalization
later on.

Fixes #4618.

### DIFF
--- a/model/src/main/kotlin/Package.kt
+++ b/model/src/main/kotlin/Package.kt
@@ -103,7 +103,8 @@ data class Package(
     val vcs: VcsInfo,
 
     /**
-     * Processed VCS-related information about the [Package] that has e.g. common mistakes corrected.
+     * Processed VCS-related information about the [Package] in normalized form. The information is either derived from
+     * [vcs], guessed from additional data as a fallback, or empty.
      */
     val vcsProcessed: VcsInfo = vcs.normalize(),
 
--- a/model/src/main/kotlin/Package.kt
+++ b/model/src/main/kotlin/Package.kt
@@ -104,7 +104,8 @@ data class Package(
 
     /**
      * Processed VCS-related information about the [Package] in normalized form. The information is either derived from
-     * [vcs], guessed from additional data as a fallback, or empty.
+     * [vcs], guessed from additional data as a fallback, or empty. On top of that [PackageCuration]s may have been
+     * applied.
      */
     val vcsProcessed: VcsInfo = vcs.normalize(),
 
@@ -164,7 +165,7 @@ data class Package(
             homepageUrl = homepageUrl.takeIf { it != other.homepageUrl },
             binaryArtifact = binaryArtifact.takeIf { it != other.binaryArtifact },
             sourceArtifact = sourceArtifact.takeIf { it != other.sourceArtifact },
-            vcs = vcs.takeIf { it != other.vcs }?.toCuration(),
+            vcs = vcsProcessed.takeIf { it != other.vcsProcessed }?.toCuration(),
             isMetaDataOnly = isMetaDataOnly.takeIf { it != other.isMetaDataOnly }
         )
     }
--- a/model/src/main/kotlin/PackageCurationData.kt
+++ b/model/src/main/kotlin/PackageCurationData.kt
@@ -105,15 +105,15 @@ data class PackageCurationData(
 private fun applyCurationToPackage(targetPackage: CuratedPackage, curation: PackageCurationData): CuratedPackage {
     val base = targetPackage.pkg
 
-    val vcs = curation.vcs?.let {
+    val vcsProcessed = curation.vcs?.let {
         // Curation data for VCS information is handled specially so we can curate only individual properties.
         VcsInfo(
-            type = it.type ?: base.vcs.type,
-            url = it.url ?: base.vcs.url,
-            revision = it.revision ?: base.vcs.revision,
-            path = it.path ?: base.vcs.path
-        )
-    } ?: base.vcs
+            type = it.type ?: base.vcsProcessed.type,
+            url = it.url ?: base.vcsProcessed.url,
+            revision = it.revision ?: base.vcsProcessed.revision,
+            path = it.path ?: base.vcsProcessed.path
+        ).normalize()
+    } ?: base.vcsProcessed
 
     val authors = curation.authors ?: base.authors
     val declaredLicenseMapping = targetPackage.getDeclaredLicenseMapping() + curation.declaredLicenseMapping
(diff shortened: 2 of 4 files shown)


fviernau:
@sschuberth @mnonnenmacher @tsteenbe I'm aware we discussed the `normalize()` should be dropped. However, I had further doubts that this is a good idea. As-is vcsProcessed is always normalized, which simplifies things and I don't see the reason why to drop that simplification. E.g. the downloader would need to apply normalization before downloading and also the matching of package configurations need to adhere to the fact that URLs are not normalized anymore. I guess that also the scan storage would be affected. No, matter what we decide about dropping the normalization. I believe it should be done out of scope of this PR.

model/src/main/kotlin/PackageCurationData.kt
```diff
@@ -108,12 +108,12 @@ private fun applyCurationToPackage(targetPackage: CuratedPackage, curation: Pack
...
-            revision = it.revision ?: base.vcs.revision,
-            path = it.path ?: base.vcs.path
-        )
-    } ?: base.vcs
+            type = it.type ?: base.vcsProcessed.type,
+            url = it.url ?: base.vcsProcessed.url,
+            revision = it.revision ?: base.vcsProcessed.revision,
+            path = it.path ?: base.vcsProcessed.path
+        ).normalize()
+    } ?: base.vcsProcessed
```
mnonnenmacher, line 116: I think in line 132 below `vcs` now needs to stay unchanged and `vcsProcessed` needs to be set instead. This way `vcs` still represents the unmodified metadata of the package.

analyzer/src/funTest/assets/projects/synthetic/gradle-all-dependencies-expected-result-with-curations.yml
```diff
@@ -608,7 +608,7 @@ analyzer:
             algorithm: "SHA-1"
         vcs:
           type: "Git"
-          url: "git@github.com:hamcrest/JavaHamcrest.git"
+          url: "ssh://git@github.com/hamcrest/JavaHamcrest.git"
```
mnonnenmacher, line 611: Commit message: Typo in `Package.vcsProcess`

mnonnenmacher (changes requested):

mnonnenmacher:
> 
> 
> @sschuberth @mnonnenmacher @tsteenbe I'm aware we discussed the `normalize()` should be dropped. However, I had further doubts that this is a good idea. As-is vcsProcessed is always normalized, which simplifies things. E.g. the downloader would need to apply normalization before downloading and also the matching of package configurations need to adhere to the fact that URLs are not normalized anymore. Not sure what handling in the scan storage would be affected. Given these worries I prefer to not drop normalization.
> 
> No, matter what we decide at the end. I believe dropping normalization should be done out of scope of this PR, if at all.

IMO not normalizing curations is still the right idea but I understand that there is some migration effort involved as you describe, so I'm fine with doing it separately. To estimate the migration effort you could write a small function that goes through the curations and package configurations to check how many of them would be affected, maybe it's not that bad after all.

mnonnenmacher (approved):

model/src/main/kotlin/Package.kt
```diff
@@ -103,7 +103,8 @@ data class Package(
     val vcs: VcsInfo,
 
     /**
-     * Processed VCS-related information about the [Package] that has e.g. common mistakes corrected.
+     * Processed VCS-related information about the [Package] in normalized form. The information is either derived from
+     * [vcs], guessed from additional data as fallback or empty.
```
sschuberth: "[...] as *a* fallback, or empty."

model/src/main/kotlin/Package.kt
```diff
@@ -104,7 +104,8 @@ data class Package(
 
     /**
      * Processed VCS-related information about the [Package] in normalized form. The information is either derived from
-     * [vcs], guessed from additional data as fallback or empty.
+     * [vcs], guessed from additional data as fallback or empty. On top of that [PackageCuration]s may have been
```
sschuberth, line 107: The commit message should fix https://github.com/oss-review-toolkit/ort/issues/4618.

sschuberth (approved):

PR #4655, opened 2021-11-03
Allow curating the package URL

### LINKED ISSUE #4565: Allow curating the PURL for a package
As a users I would like to be able to amend the [PURLs](https://github.com/package-url/purl-spec) ORT currently generates. 

Use cases:
- Generate PURL is incorrect e.g. for internal packages
- Improve PURL with additional qualifiers or paths

Propose to add purl as a field to https://github.com/oss-review-toolkit/ort/blob/master/model/src/main/kotlin/PackageCurationData.kt so ORT users can simply set PURL via [curations.yml](https://github.com/oss-review-toolkit/ort/blob/master/docs/config-file-curations-yml.md)

fviernau:
For some packages the default value for the package URL generated by
ORT is invalid. That generated default value also does not cover some
of the parameters e.g. qualifiers or subpaths, see [1].

Allow curating the package URLs to enable fixing invalid or improving
valid default values.

[1] https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst

Fixes #4565.

### DIFF
--- a/docs/config-file-curations-yml.md
+++ b/docs/config-file-curations-yml.md
@@ -52,6 +52,7 @@ The structure of the curations file consist of one or more `id` entries:
 - id: "Maven:com.example.app:example:0.0.1"
   curations:
     comment: "An explanation why the curation is needed or the reasoning for a license conclusion"
+    purl: "pkg:Maven/com.example.app/example@0.0.1?arch=arm64-v8a#src/main"
     concluded_license: "Valid SPDX license expression to override the license findings."
     declared_license_mapping:
       "license a": "Apache-2.0"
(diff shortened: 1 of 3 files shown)


docs/config-file-curations-yml.md
```diff
@@ -52,6 +52,7 @@ The structure of the curations file consist of one or more `id` entries:
 - id: "package identifier."
   curations:
     comment: "An explanation why the curation is needed or the reasoning for a license conclusion"
+    purl: "[Valid](https://github.com/package-url/purl-spec) package URL."
```
sschuberth, line 55: Shouldn't you better curate to something that really is a valid PURL syntax-wise?

sschuberth (approved):

PR #4658, opened 2021-11-03
Analyzer: Pass the labels to the `resolveDependencies` functions

nnobelis:
This allows implementing a dependency resolution logic depending on the
labels given as parameters to the AnalyzerCommand. This is useful, for
instance, for custom analyzer plugins.

### DIFF
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -42,7 +42,7 @@ import org.ossreviewtoolkit.utils.core.log
 /**
  * The class to run the analysis. The signatures of public functions in this class define the library API.
  */
-class Analyzer(private val config: AnalyzerConfiguration) {
+class Analyzer(private val config: AnalyzerConfiguration, private val labels: Map<String, String> = emptyMap()) {
     data class ManagedFileInfo(
         val absoluteProjectPath: File,
         val managedFiles: Map<PackageManager, List<File>>,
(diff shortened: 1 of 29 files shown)


mnonnenmacher (approved):

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -42,7 +42,7 @@ import org.ossreviewtoolkit.utils.core.log
 /**
  * The class to run the analysis. The signatures of public functions in this class define the library API.
  */
-class Analyzer(private val config: AnalyzerConfiguration) {
+class Analyzer(private val config: AnalyzerConfiguration, val labels: Map<String, String>) {
```
sschuberth, line 45: Could `labels` be `private` as well?

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -42,7 +42,7 @@ import org.ossreviewtoolkit.utils.core.log
 /**
  * The class to run the analysis. The signatures of public functions in this class define the library API.
  */
-class Analyzer(private val config: AnalyzerConfiguration) {
+class Analyzer(private val config: AnalyzerConfiguration, private val labels: Map<String, String>) {
```
sschuberth, line 45: Sorry, one more thing: Should we default this to `= emptyMap()` like we do in `OrtResult` and `Defect` already, to avoid some of the trivial code changes?

sschuberth (approved):

PR #4660, opened 2021-11-03
scanner: Add a working tree cache for the experimental scanner

mnonnenmacher:
Add a working tree cache to avoid downloading the same sources multiple times. See the commit messages for details.

### DIFF
--- a/scanner/src/main/kotlin/experimental/ProvenanceDownloader.kt
+++ b/scanner/src/main/kotlin/experimental/ProvenanceDownloader.kt
@@ -29,16 +29,19 @@ import org.ossreviewtoolkit.model.KnownProvenance
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.utils.core.createOrtTempDir
 
 /**
  * An interface that provides functionality to download source code.
  */
 interface ProvenanceDownloader {
     /**
-     * Download the source code specified by the provided [provenance] to the [downloadDir]. Throws a
-     * [DownloadException] if the download fails.
+     * Download the source code specified by the provided [provenance] and return the path to the directory that
+     * contains the downloaded source code.
+     *
+     * Throws a [DownloadException] if the download fails.
      */
-    fun download(provenance: KnownProvenance, downloadDir: File)
+    fun download(provenance: KnownProvenance): File
 }
 
 /**
@@ -48,7 +51,9 @@ interface ProvenanceDownloader {
 class DefaultProvenanceDownloader(config: DownloaderConfiguration) : ProvenanceDownloader {
     private val downloader = Downloader(config)
 
-    override fun download(provenance: KnownProvenance, downloadDir: File) {
+    override fun download(provenance: KnownProvenance): File {
+        val downloadDir = createOrtTempDir()
+
         when (provenance) {
             // TODO: Add dedicated download functions for VcsInfo and RemoteArtifact to the Downloader.
             is ArtifactProvenance -> {
@@ -63,5 +68,7 @@ class DefaultProvenanceDownloader(config: DownloaderConfiguration) : ProvenanceD
                 downloader.downloadFromVcs(pkg, downloadDir, recursive = false)
             }
         }
+
+        return downloadDir
     }
 }
--- a/scanner/src/main/kotlin/experimental/ProvenanceDownloader.kt
+++ b/scanner/src/main/kotlin/experimental/ProvenanceDownloader.kt
@@ -22,6 +22,8 @@ package org.ossreviewtoolkit.scanner.experimental
 
 import java.io.File
 
+import kotlinx.coroutines.runBlocking
+
 import org.ossreviewtoolkit.downloader.DownloadException
 import org.ossreviewtoolkit.downloader.Downloader
 import org.ossreviewtoolkit.model.ArtifactProvenance
@@ -29,6 +31,8 @@ import org.ossreviewtoolkit.model.KnownProvenance
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.utils.common.safeCopyRecursively
+import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.core.createOrtTempDir
 
 /**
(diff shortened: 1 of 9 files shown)


scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    override suspend fun <T> use(vcsInfo: VcsInfo, block: (VersionControlSystem, WorkingTree) -> T): T =
+        getWorkingTreeMutex(vcsInfo).withLock {
+            val vcs = getVcs(vcsInfo)
+            val workingTree = getWorkingTree(vcsInfo)
+
+            block(vcs, workingTree)
+        }
+
+    private fun getKey(vcsInfo: VcsInfo) = "${vcsInfo.type}|${vcsInfo.url}|${vcsInfo.path}"
```
fviernau, line 66: Why is the path part of the key?

fviernau (changes requested):

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
```diff
@@ -0,0 +1,93 @@
...
+    private val mutex = Mutex()
+    private val workingTreeMutexes = mutableMapOf<String, Mutex>()
+    private val workingTrees = mutableMapOf<String, WorkingTree>()
+
+    override suspend fun <T> use(vcsInfo: VcsInfo, block: (VersionControlSystem, WorkingTree) -> T): T =
+        getWorkingTreeMutex(vcsInfo).withLock {
+            val vcs = getVcs(vcsInfo)
+            val workingTree = getWorkingTree(vcsInfo)
+
+            block(vcs, workingTree)
```
fviernau, line 63: 1. `use()` first aquires the `workingTreeMutex` and then the `mutex`
2. clear() first aquires the `mutex` and then each `workingTreeMutex`

...so this fullfills the necessary precondition for a deadlock. Any way to avoid that?

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
```diff
@@ -0,0 +1,93 @@
...
+            ?: VersionControlSystem.forUrl(vcsInfo.url)
+            ?: throw IOException("Could not determine VCS type for (type=$${vcsInfo.type}, url=${vcsInfo.url}).")
+
+    private fun getWorkingTree(vcsInfo: VcsInfo): WorkingTree {
+        val vcs = getVcs(vcsInfo)
+        val dir = createOrtTempDir()
+        return workingTrees.getOrPut(getKey(vcsInfo)) { vcs.initWorkingTree(dir, vcsInfo.copy(revision = "")) }
+    }
+
+    override suspend fun clear() {
```
fviernau, line 82: Would it make sense to turn `clear()` into `shutdown()`, in particular not allow any `use` call after `clear()` returns?

This way the caller of `clear()` can be sure that storage space is actually cleared as soon as his code execution continues after returning from this blocking call.

fviernau (changes requested):

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
mnonnenmacher: Two reasons:
* For git-repo the path represents the manifest and I don't want to reuse the same working tree for different manifests.
* For git the sparse checkout is configured durint `initWorkingTree()` so it is currently not possible to reuse the working tree for different paths.

fviernau: In the context of previous refactorings moving towards "scan by repository" it recall that it often times made sense to move away from sparse checkouts. So, would this make sense also here? e.g. drop the path from the vcs info so that no sparse checkout is made?

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
mnonnenmacher: No, `use` only locks with `mutex` while acquiring the `workingTreeMutex` in `getWorkingTreeMutex()`. When using the `workingTreeMutex` the lock on `mutex` is already released.

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
mnonnenmacher: It's another approach, but making `clear()` non-blocking also has benefits, it allows to continue operation while the data is deleted in the background (even though currently it's only called inside `runBlocking` anyway).

fviernau: What do you mean by "making `clear()` non-locking"?

To my understanding it is blocking because it acquires locks.

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
fviernau: Oh, your're right. I missed the release of the lock.

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
mnonnenmacher: I initially did this, but git-repo made it impossible because we need the path there to know which manifest to use. I would like to change this so that for git-repo the manifest path is given as part of the URL (like `http://path/to/repo?manifest.xml`), then we could drop the path here.

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
mnonnenmacher: It's still a suspend function, the mutex lock only blocks other coroutines from entering the same code block, but they can continue doing other stuff. See: https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#mutual-exclusion

mnonnenmacher: Currently this does not make a difference because all provenance resolution is done in order, but I want to change this to use coroutines later on for better performance.

oheger-bosch: I like the idea to prevent any `use()` calls after an invocation of `clear()`. Otherwise, there is always a potential for a race condition that a `use` is processed after a `clean` leaving some data on the disk. This could be achieved by setting a flag in `clear()`. `use()` would check this flag and fail it if was *true*.

scanner/src/main/kotlin/experimental/ProvenanceDownloader.kt
```diff
@@ -62,13 +69,26 @@ class DefaultProvenanceDownloader(config: DownloaderConfiguration) : ProvenanceD
...
         }
 
         return downloadDir
     }
+
+    private suspend fun downloadFromVcs(provenance: RepositoryProvenance, downloadDir: File) {
+        workingTreeCache.use(provenance.vcsInfo) { vcs, workingTree ->
+            vcs.updateWorkingTree(workingTree, provenance.resolvedRevision, recursive = false)
+
+            // Make sure that all nested repositories are removed. Even though we do not clone recursive above, nested
```
oheger-bosch, line 83: Grammar: we do not clone *recursively*.

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
mnonnenmacher: Ok, as both of you prefer that approach I have changed it.

scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
fviernau: >  for git-repo the manifest path is given as part of the URL (like http://path/to/repo?manifest.xml), then we could drop the path here.

nice idea!

oheger-bosch (approved):

PR #4662, opened 2021-11-04
utils-core: Make showStackTrace() an expression function

mnonnenmacher:
Use `let` to make the function an expression function.

### DIFF
--- a/utils/core/src/main/kotlin/Extensions.kt
+++ b/utils/core/src/main/kotlin/Extensions.kt
@@ -52,8 +52,4 @@ fun Any.createOrtTempFile(prefix: String? = null, suffix: String? = null): File 
 /**
  * Print the stack trace of the [Throwable] if [printStackTrace] is set to true.
  */
-fun Throwable.showStackTrace() {
-    // We cannot use a function expression for a single "if"-statement, see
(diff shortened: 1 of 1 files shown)


utils/core/src/main/kotlin/Extensions.kt
```diff
@@ -52,8 +52,4 @@ fun Any.createOrtTempFile(prefix: String? = null, suffix: String? = null): File
 /**
  * Print the stack trace of the [Throwable] if [printStackTrace] is set to true.
  */
-fun Throwable.showStackTrace() {
-    // We cannot use a function expression for a single "if"-statement, see
-    // https://discuss.kotlinlang.org/t/if-operator-in-function-expression/7227.
-    if (printStackTrace) printStackTrace()
-}
+fun Throwable.showStackTrace(): Unit = let { if (printStackTrace) printStackTrace() }
```
sschuberth, line 55: Nit: Not that it matters much, but IMO the semantics of `run` match a bit better here (e.g. as `it` is not used / required), and it arguably also reads a bit nicer. What do you think?

mnonnenmacher: Agreed.

fviernau (approved):

PR #4664, opened 2021-11-04
GitRepo: Improve the code comment for running "repo init"

sschuberth:
The command does not actually clone the referenced Git repositories yet.

### DIFF
--- a/downloader/src/main/kotlin/vcs/GitRepo.kt
+++ b/downloader/src/main/kotlin/vcs/GitRepo.kt
@@ -151,11 +151,15 @@ class GitRepo : VersionControlSystem(), CommandLineTool {
             "Initializing git-repo from ${vcs.url} with revision '$manifestRevision' and manifest '$manifestPath'."
         }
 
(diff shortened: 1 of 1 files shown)


downloader/src/main/kotlin/vcs/GitRepo.kt
```diff
@@ -151,7 +151,8 @@ class GitRepo : VersionControlSystem(), CommandLineTool {
             "Initializing git-repo from ${vcs.url} with revision '$manifestRevision' and manifest '$manifestPath'."
         }
 
-        // Clone all projects instead of only those in the "default" group until we support specifying groups.
+        // Prepare for cloning all projects instead of only those in the "default" group until we support specifying
+        // groups.
```
fviernau, line 155: Would it make sense to re-format the below code to one param per line and then move the comment next to `--groups=all` as it only refers to that option?

fviernau (dismissed):

fviernau (approved):

Issue #95, opened 2017-12-01
NPM/PNPM: Add support for peer dependencies


mnonnenmacher:
Add support for the peerDependencies scope to NPM. Be aware that the specific version of the peer dependency will only be resolved in the context of a project using the library.

npm-install-peers can be used to install peer dependencies, see this relevant ticket:
https://github.com/spatie/npm-install-peers/issues/13

tsteenbe:
@mnonnenmacher  Harp.gl has multiple peer dependency on three.js see https://github.com/heremaps/harp.gl/blob/e3110f1dc35a889059ba23faefc76e3a4c1cfe73/%40here/harp-debug-datasource/package.json#L42

Now ORT does not pick up the core library that powers harp.gl

sschuberth:
Good write-up about what peer dependencies actually are: https://nodejs.org/en/blog/npm/peer-dependencies/. Basically, peer dependencies define against what API of a "host" package a (plugin) package has been written.

PR #4665, opened 2021-11-04
analyzer: Remove `peerDependencies` from NPM / Yarn test assets

sschuberth:
ORT does not support `peerDependencies` yet [1]. That is mostly because
handling of peer dependencies varies with NPM versions: Some versions do
install them by default, some do not [2]. This results in ORT's functional
tests to fail depending on the NPM version, and as it turns out, they
would fail with the NPM version 7 currently used in ORT's `Dockerfile`.
However, tests succeed on Azure CI which uses NPM 6.  Avoid that
inconsistency by simply not using peer dependencies at all for testing for
now.

[1]: https://github.com/oss-review-toolkit/ort/issues/95
[2]: https://nodejs.org/en/blog/npm/peer-dependencies#using-peer-dependencies

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/npm/no-lockfile/package.json
+++ b/analyzer/src/funTest/assets/projects/synthetic/npm/no-lockfile/package.json
@@ -15,9 +15,6 @@
   "devDependencies": {
     "cson": "~4.1.0"
   },
-  "peerDependencies": {
-    "tea": "0.x"
-  },
   "optionalDependencies": {
     "promise": "~7.3.1"
   }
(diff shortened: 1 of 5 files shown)


analyzer/src/funTest/assets/projects/synthetic/npm/no-lockfile/package.json
```diff
@@ -15,9 +15,6 @@
   "devDependencies": {
     "cson": "~4.1.0"
   },
```
fviernau, line 17: commit-msg: as it turn -> as it turn**s** || as it turn**ed**

fviernau (dismissed):

MarcelBochtler (approved):

Issue #4631, opened 2021-10-28
Epic about using custom / newer ScanCode versions


sschuberth:
Currently, the ScanCode version to use is [hard-coded](https://github.com/oss-review-toolkit/ort/blob/5fc30626d26e1be1c618525a912a3da055460c15/gradle.properties#L88) in ORT. This tight coupling prevents mixing of arbitrary ORT / ScanCode versions, and hinders a fast / user-specific upgrade of ScanCode (also see #4296).

To solve this, the following things would need to be addressed:
- [x] Improve the boostrapping of scanners:
  - [x] Accept a scanner in any version, as long as it is present on the system (leaving it to the user to install a compatible version; also see #4582); see https://github.com/oss-review-toolkit/ort/pull/4669
  - [x] Bootstrap to a cached location for reuse (like done for Flutter; also see #1580).
  - [x] ~~Alternatively: Remove bootstrapping for scanners completely. However, that would requite CI to always install all supported scanners, and not only those whose tests are actually run.~~
- [x] Improve handling of license texts (also see [this comment](https://github.com/oss-review-toolkit/ort/issues/4296#issuecomment-952644016)):
  - [x] Refactor the `utils/core` module so that `utils/spdx` can depend on non-ORT-specific parts of it (the plan is to introduce `utils/common`; see https://github.com/oss-review-toolkit/ort/pull/4647, https://github.com/oss-review-toolkit/ort/pull/4652, https://github.com/oss-review-toolkit/ort/pull/4670, https://github.com/oss-review-toolkit/ort/pull/4671)
  - [x] Read license texts from the ScanCode version in use rather than embedding them, see #4523.
- [x] Verify that the [scanner criteria](https://github.com/oss-review-toolkit/ort/blob/master/scanner/src/main/kotlin/ScannerCriteria.kt) for looking up stored scan results can be configured flexible enough to e.g. declare ScanCode results of all versions from 3.2.1-rc2 to 30.1.0 as compatible:
  - [x] #3247
  - [x] ~~#3223~~
- [x] Make the ScanCode version in the Docker image easily configurable (see #4721)

### FAQ

Why not just always use the latest version of ScanCode in ORT?
- Different findings might render existing package configurations invalid.
- Expensive testing might be required.

maxhbr:
#1580 is closed. Does that mean that all boxes in "Improve the boostrapping of scanners" can be checked?

sschuberth:
[...]

Issue #4582, opened 2021-10-12
Make versions of local scanners configurable


oheger-bosch:
So far, for local scanners, ORT has the specific version it supports hard-coded; if a different version of the scanner is installed on the system, it is not accepted, but ORT tries to bootstrap this exact version (see https://github.com/oss-review-toolkit/ort/blob/4ab3679b1b654ab00953224e2c52b3f34a32b377/scanner/src/main/kotlin/LocalScanner.kt#L119).

This behaviour is rather strict, and it complicates updates to newer scanner versions (e.g. refer to #4296). To be more flexible, ORT could support setting the accepted scanner version (or even a version range) in the scanner configuration. For instance, for ScanCode, this could look as follows:

```
  scanner {
     ...
      ScanCode {
        commandLine = "--copyright --license --info --strip-root --timeout 300 --max-in-memory 5000"
        parseLicenseExpressions = true
	minVersion = 3.2.1
	maxVersion = 22.1.1
     }
   ...
}

```
_Note_: The properties _minVersion_ and _maxVersion_ are currently already used to configure the accepted versions when loading results from the scan result storage; so they would have to be renamed.

The `expectedVersion` hard-coded in ORT could be kept and used as default value if no version is configured for a scanner; this would also be the version that gets bootstrapped.

sschuberth:
Closing this in favor of the even more relaxed approach in https://github.com/oss-review-toolkit/ort/pull/4669, which accepts any scanner version already installed on the host, but informs about a version mismatch.

PR #4669, opened 2021-11-05
LocalScanner: Remove the version requirement for scanners

mnonnenmacher:
Do not require that the installed version of a scanner matches the
expected version. Instead print a message if the installed version is
not the expected version, but continue using it for scanning.


mnonnenmacher:
This depends on #4523 to be merged first. This should be the last required step to allow using custom ScanCode versions (#4631). Instead of making the scanner versions configurable as described in #4582 it simply removes the version requirement.

fviernau (approved):

sschuberth (dismissed):
Approving, though a rebase is needed to resolve conflicts.

sschuberth (approved):

PR #4670, opened 2021-11-05
Move remaining common code from spdx-utils to common-utils

sschuberth:


cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -71,7 +71,7 @@ import org.ossreviewtoolkit.utils.core.archive
 import org.ossreviewtoolkit.utils.core.log
 import org.ossreviewtoolkit.utils.core.ortConfigDirectory
 import org.ossreviewtoolkit.utils.core.showStackTrace
-import org.ossreviewtoolkit.utils.spdx.model.LicenseChoice
+import org.ossreviewtoolkit.utils.spdx.model.SpdxLicenseChoice
```
mnonnenmacher, line 74: The new name sounds a bit like the class can only be used for choosing SPDX licenses.

sschuberth: Which *is* the case, see the commit message: "[...] because the class operates exclusively on `SpdxExpression`s". Or am I missing something?

MarcelBochtler (approved):

cli/src/main/kotlin/commands/DownloaderCommand.kt
mnonnenmacher: Yes, SPDX expressions can contain non-SPDX licenses. Except if you count LicenseRef-* as SPDX licenses.

sschuberth: In this context I actually do... my main point is that `SpdxLicenseChoice` only works on licenses with IDs that follow the SPDX specification, incl. LicenseRefs.

PR #4671, opened 2021-11-05
Finalize the `utils` module refactoring

sschuberth:


analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -141,7 +139,7 @@ object PythonVersion : CommandLineTool {
             // Parse a line like " -2.7-32        C:\Python27\python.exe".
             versionAndPath?.split(' ', limit = 3)?.last()?.trimStart()
         } else {
-            getPathFromEnvironment("python$version")?.path
+            Os.getPathFromEnvironment("python$version")?.path
```
mnonnenmacher, line 142: I couldn't find a way to comment on the previous commit "core-utils: Move all storage tests under functional tests": The test classes should also be renamed to end with "FunTest".

utils/core/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -75,22 +77,17 @@ enum class ArchiveType(vararg val extensions: String) {
 
 /**
  * Archive the contents of [inputDir], omitting common [VCS_DIRECTORIES], to [zipFile] where an optional [prefix] is
- * added to each file. Return a [Result] wrapping the [zipFile] on success, or an exception of failure.
+ * added to each file. Return the [zipFile] on success, or throw an exception on failure.
  */
-fun archive(inputDir: File, zipFile: File, prefix: String = ""): Result<File> {
-    return runCatching {
+fun archive(inputDir: File, zipFile: File, prefix: String = ""): File =
```
oheger-bosch, line 77: From the commit message, it is not clear to me why the strategy for error handling needs to be changed here. Kotlin's `Result` class should be independent on other utility functions. Maybe this becomes obvious in later commits, but it would be helpful to have a hint in the commit message.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/Pip.kt
sschuberth: Good catch, thanks, fixed.

utils/core/src/main/kotlin/ArchiveUtils.kt
sschuberth: My reason for this change was that I needed to remove the dependency on `showStackTrace()`, which basically left the `onFailure` branch empty. So the function then only did an internal wrapping in `runCatching`, which I felt should be something that the caller should do, if desired. Also, it allowed to use a slightly more elegant syntax to return the `zipFile`.

oheger-bosch: Thanks for clarifying.

mnonnenmacher (approved):

PR #4672, opened 2021-11-08
Analyzer: Prepend the logger to log with ORT class

nnobelis:
This prevents an error when an Analyzer plugin is used which does not use the ORT package name.

```
12:13:17.890 [DefaultDispatcher-worker-3] INFO  org.ossreviewtoolkit.utils.core.OrtProxySelector - Proxy selector was successfully installed.
Exception in thread "main" java.lang.IllegalArgumentException: Logging is only allowed on ORT classes, but 'io.bosch.ort.plugins.MySuperPlugin' is used.
	at org.ossreviewtoolkit.analyzer.PackageManager.resolveDependencies(PackageManager.kt:389)
	at org.ossreviewtoolkit.analyzer.Analyzer$analyzeInParallel$1$1$1.invokeSuspend(Analyzer.kt:132)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```
I assume that this occurs because the `open fun` of `PackageManager` gets inlined in the plugin.

### DIFF
--- a/analyzer/src/main/kotlin/PackageManager.kt
+++ b/analyzer/src/main/kotlin/PackageManager.kt
@@ -235,7 +235,7 @@ abstract class PackageManager(
         definitionFiles.forEach { definitionFile ->
             val relativePath = definitionFile.relativeTo(analysisRoot).invariantSeparatorsPath
 
-            log.info { "Resolving $managerName dependencies for '$relativePath'..." }
+            PackageManager.log.info { "Resolving $managerName dependencies for '$relativePath'..." }
 
             val duration = measureTime {
                 @Suppress("TooGenericExceptionCaught")
@@ -271,7 +271,9 @@ abstract class PackageManager(
                 }
             }
 
-            log.info { "Resolving $managerName dependencies for '$relativePath' took ${duration.inWholeSeconds}s." }
+            PackageManager.log.info {
+                "Resolving $managerName dependencies for '$relativePath' took ${duration.inWholeSeconds}s."
+            }
         }
 
         afterResolution(definitionFiles)


fviernau (dismissed):

oheger-bosch:
For my understanding: Isn't this a general problem that can occur when external code extends ORT base classes? If so, should this better be fixed in the log implementation, which is also responsible for throwing this exception?

nnobelis:
Your comment makes sense because each ORT step can be implemented trough a plugin.
However how would your address it in the log implementation ?

oheger-bosch:
Somewhere in `Logger.kt`. A generic solution would probably require some reflection magic to find out whether the current class extends an ORT class.

sschuberth:
When I did the [final utils refactoring](https://github.com/oss-review-toolkit/ort/pull/4671) I discovered a new [helper interface in Log4j](https://github.com/apache/logging-log4j-kotlin/blob/master/log4j-api-kotlin/src/main/kotlin/org/apache/logging/log4j/kotlin/Logging.kt)'s Kotlin bindings. I was actually considering to use that throughout ORT, replacing `Logger.kt` completely. However, `logOnce` would then also need to be implemented differently. And I don't like `logger.` quite as much as `log.`.

nnobelis:
@oheger-bosch: Regarding @sschuberth remark, can we merge current PR without investing effort in changing `Logger.kt` logic, if it will be refactored anyway ?

oheger-bosch:
I am currently hit by a similar problem: I am working on a plugin implementation that calls a method of the super class, which contains a log statement. Here again this exception is thrown, but for the plugin implementation, there is no easy workaround.

As the refactoring will probably not happen immediately, I am investigating whether I can find a more generic quick-fix.

sschuberth:
> As the refactoring will probably not happen immediately, I am investigating whether I can find a more generic quick-fix.

Maybe a pragmatic approach would be to simply that using the described new approach only in the classes your plugins are using?

mnonnenmacher:
> When I did the [final utils refactoring](https://github.com/oss-review-toolkit/ort/pull/4671) I discovered a new [helper interface in Log4j](https://github.com/apache/logging-log4j-kotlin/blob/master/log4j-api-kotlin/src/main/kotlin/org/apache/logging/log4j/kotlin/Logging.kt)'s Kotlin bindings. I was actually considering to use that throughout ORT, replacing `Logger.kt` completely. However, `logOnce` would then also need to be implemented differently. And I don't like `logger.` quite as much as `log.`.

As the helper interface is quite simple we could make our own `OrtLogging` interface which uses `log` instead of `logger` and includes the log once functionality (maybe optionally).

oheger-bosch:
I have created #4675 with a quick fix that allows logging not only for ORT classes but also for classes extending an ORT class or interface. This would be a temporary fix until we decide how to proceed.

nnobelis:
Superseded by https://github.com/oss-review-toolkit/ort/pull/4675.

Please create an issue for the suggestions discussed above.

PR #4675, opened 2021-11-09
Utils: Support logging from classes that extend ORT base classes

oheger-bosch:
ORT's logging mechanism currently checks whether the current class
belongs to a package in ORT's namespace; otherwise, an exception is
thrown. This check can fail for external classes that extend ORT
functionality, e.g. plugins - even if the logging code is in a base
class.

As a work-around, make the check more lenient to accept classes that
extend a class or interface from the ORT namespace.

To prevent a detekt failure with the corresponding test case (that
uses a class in a different namespace), the OrtPackageNaming rule
class had to be tweaked.

### DIFF
--- a/detekt-rules/src/main/kotlin/OrtPackageNaming.kt
+++ b/detekt-rules/src/main/kotlin/OrtPackageNaming.kt
@@ -48,7 +48,9 @@ class OrtPackageNaming : Rule() {
     override fun visitPackageDirective(directive: KtPackageDirective) {
         super.visitPackageDirective(directive)
 
-        if (directive.qualifiedName.isEmpty()) return
+        // Exclusion of packages starting with "test." is required for OrtLogTestExtension used by LoggerTest, which
+        // simulates an ORT extension class in a different package.
+        if (directive.qualifiedName.isEmpty() || directive.qualifiedName.startsWith("test.")) return
 
         val path = directive.containingKtFile.toFilePath().relativePath.toString()
         if (!path.contains(pathPattern)) return
--- a/utils/core/src/main/kotlin/Logger.kt
+++ b/utils/core/src/main/kotlin/Logger.kt
@@ -41,12 +41,26 @@ val loggerOfClass = ConcurrentHashMap<Any, KotlinLogger>()
 val Any.log: KotlinLogger
     inline get() = loggerOfClass.getOrPut(this::class.java) {
         this::class.qualifiedName?.let { name ->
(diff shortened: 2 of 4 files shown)


nnobelis (dismissed):

detekt-rules/src/main/kotlin/OrtPackageNaming.kt
```diff
@@ -48,7 +48,7 @@ class OrtPackageNaming : Rule() {
     override fun visitPackageDirective(directive: KtPackageDirective) {
         super.visitPackageDirective(directive)
 
-        if (directive.qualifiedName.isEmpty()) return
+        if (directive.qualifiedName.isEmpty() || directive.qualifiedName.startsWith("test.")) return
```
sschuberth, line 51: Please add a code comment that this change is *only* required for `OrtLogTestExtension`.

utils/core/src/main/kotlin/Logger.kt
```diff
@@ -41,14 +41,28 @@ val loggerOfClass = ConcurrentHashMap<Any, KotlinLogger>()
...
+
+/**
+ * Check whether [cls] represents a class or interface that belongs to ORT. This is the case if [cls] belongs to a
+ * package in the ORT namespace (as defined by [ORT_PACKAGE]) or extends an ORT class or interface.
+ */
+fun isOrtClass(cls: Class<*>?): Boolean =
+    when {
+        cls == null -> false
+        cls.name.startsWith(ORT_PACKAGE) -> true
+        else -> cls.interfaces.find { it.name.startsWith(ORT_PACKAGE) } != null || isOrtClass(cls.superclass)
```
sschuberth, line 63: Could use `any {}` instead of `find {} != null`.

utils/core/src/test/kotlin/test/other/OrtLogTestExtension.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+abstract class OrtLogTestBaseExtension : CommandLineTool
+
+/**
+ * A test class simulating an extension of ORT that produces log output. This is used to test whether classes from
+ * different packages extending ORT classes can use logging.
+ */
+class OrtLogTestExtension : OrtLogTestBaseExtension() {
+    override fun command(workingDir: File?): String {
+        log.info("Test of the logger.")
```
sschuberth, line 39: Please also use curly-brace syntax here for consistency.

sschuberth (changes requested):

sschuberth (approved):

Issue #4486, opened 2021-09-22
Adding @JvmOverloads to Analyzer methods


LeelaChacha:
Hello ORT People,

I am trying to integrate the ORT Analyzer in a Java project using Jitpack. This use case of ORT is not at all documented or not placed prominently in the README or the Wiki. Therefore, it is sometimes difficult to figure out how one might go about doing that. There are many configuration possibilities. On top of that, the default arguments are not as helpful because Java cannot use them.

I propose adding the [@JvmOverloads annotation](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-overloads/) to such methods so that integrating such methods can be done easily.

JvmOverloads tells the Kotlin compiler to automatically generate overloaded methods for the target. This means that JVM languages without default parameters support can use the overloaded methods.

I would like to create a pull request for this from my fork and we can discuss this further in this issue.

Linked PR: #4487

PR #4677, opened 2021-11-09
Add JVM overloads for main API functions with default arguments

sschuberth:
This allows to benefit from default arguments also when ORT in called
from Java.

### DIFF
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -49,6 +49,7 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
         val repositoryConfiguration: RepositoryConfiguration
     )
 
+    @JvmOverloads
     fun findManagedFiles(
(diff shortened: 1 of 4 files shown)


analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -88,6 +89,7 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
         return ManagedFileInfo(absoluteProjectPath, managedFiles, repositoryConfiguration)
     }
 
+    @JvmOverloads
```
mnonnenmacher, line 92: You should add "Fixes #4486" to the commit message.

oheger-bosch (approved):
Changes look good to me. How can we know that we do not miss an important API function? Is there a definition what are the main API functions?

sschuberth:
> Is there a definition what are the main API functions?

We currently document classes containing such functions with the sentence "The signatures of public functions in this class define the library API".

PR #4691, opened 2021-11-11
fossid-webapp: Make the timeout for FossID connections configurable

nnobelis:
When ORT waits for FossID scan to complete, the current timeout is
hardcoded and is set to one hour.
We have some FossID scans lasting for longer therefore the timeout needs
to be configurable.

### DIFF
--- a/model/src/main/resources/reference.conf
+++ b/model/src/main/resources/reference.conf
@@ -119,6 +119,7 @@ ort {
 
         deltaScans = "true"
         deltaScanLimit = "10"
+        timeout = "60"
       }
     }
 
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -94,9 +94,6 @@ class FossId internal constructor(
         @JvmStatic
         private val WAIT_INTERVAL_MS = 10000L
 
-        @JvmStatic
-        private val WAIT_REPETITION = 360
-
         /**
          * The scan states for which a scan can be triggered.
          */
@@ -572,7 +569,7 @@ class FossId internal constructor(
      * Wait until the repository of a scan with [scanCode] has been downloaded.
      */
     private suspend fun waitDownloadComplete(scanCode: String) {
-        val result = wait(WAIT_INTERVAL_MS * WAIT_REPETITION, WAIT_INTERVAL_MS) {
+        val result = wait(config.timeout * 60000L, WAIT_INTERVAL_MS) {
             FossId.log.info { "Checking download status for scan '$scanCode'." }
 
             val response = service.checkDownloadStatus(config.user, config.apiKey, scanCode)
@@ -600,7 +597,7 @@ class FossId internal constructor(
      * Wait until a scan with [scanCode] has completed.
      */
     private suspend fun waitScanComplete(scanCode: String) {
-        val result = wait(WAIT_INTERVAL_MS * WAIT_REPETITION, WAIT_INTERVAL_MS) {
+        val result = wait(config.timeout * 60000L, WAIT_INTERVAL_MS) {
             FossId.log.info { "Waiting for scan '$scanCode' to complete." }
 
             val response = service.checkScanStatus(config.user, config.apiKey, scanCode)
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
@@ -80,6 +80,9 @@ internal data class FossIdConfig(
     /** A maximum number of delta scans to keep for a single repository. */
     val deltaScanLimit: Int,
 
+    /** Timeout in minutes for communication with FossID. */
+    val timeout: Int,
+
     /** Stores the map with FossID-specific configuration options. */
     private val options: Map<String, String>
 ) {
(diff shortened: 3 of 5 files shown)


scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -117,11 +120,20 @@ internal data class FossIdConfig(
         /** Name of the configuration property that limits the number of delta scans. */
         private const val DELTA_SCAN_LIMIT_PROPERTY = "deltaScanLimit"
 
+        /** Name of the configuration property defining the amount of 10s delays to wait as timeout for communication
+         * with FossID.*/
```
sschuberth, line 124: Broken formatting.

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -572,7 +569,7 @@ class FossId internal constructor(
      * Wait until the repository of a scan with [scanCode] has been downloaded.
      */
     private suspend fun waitDownloadComplete(scanCode: String) {
-        val result = wait(WAIT_INTERVAL_MS * WAIT_REPETITION, WAIT_INTERVAL_MS) {
+        val result = wait(WAIT_INTERVAL_MS * config.waitRepetition, WAIT_INTERVAL_MS) {
```
oheger-bosch, line 572: Should the download use the same timeout? I would assume that this would be much faster than a scan operation.

model/src/main/resources/reference.conf
```diff
@@ -117,6 +117,7 @@ ort {
 
         deltaScans = "true"
         deltaScanLimit = "10"
+        waitRepetition = "360"
```
oheger-bosch, line 120: I am confused what this value actually means and how it is related to the maximum amount of time we wait for a scan. For usability, wouldn't it be better to specify the timeout directly? The property would then have to be named accordingly.

nnobelis:
Do not merge please: I got an idea last evening. I will change the impl.

model/src/main/resources/reference.conf
nnobelis: This is exactly my new idea. I will change the implementation.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: You're right. However don't you think keeping only one timeout lower the complexity of the code ?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -572,7 +569,8 @@ class FossId internal constructor(
      * Wait until the repository of a scan with [scanCode] has been downloaded.
      */
     private suspend fun waitDownloadComplete(scanCode: String) {
-        val result = wait(WAIT_INTERVAL_MS * WAIT_REPETITION, WAIT_INTERVAL_MS) {
+        config.timeout
```
oheger-bosch, line 572: What does this line do?

model/src/main/resources/reference.conf
```diff
@@ -117,6 +117,7 @@ ort {
 
         deltaScans = "true"
         deltaScanLimit = "10"
+        timeout = "3600000"
```
oheger-bosch, line 120: As the timeout is a rather big value (in the range of minutes up to an hour), I would suggest using a different time unit than milliseconds. Maybe minutes would be more appropriate. WDYT?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Copy/Paste error :)

model/src/main/resources/reference.conf
nnobelis: Yes this is a good idea. I would go for minutes too.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -80,6 +80,9 @@ internal data class FossIdConfig(
     /** A maximum number of delta scans to keep for a single repository. */
     val deltaScanLimit: Int,
 
+    /** Timeout in minutes for communication with FossID. */
+    val timeout: Long,
```
oheger-bosch, line 84: Minor nit: As we now use minutes as unit for timeouts, this property should probably better of type `Int` rather than `Long`, since Java APIs typically use `Long` values to represent milliseconds. But I leave this to you.

oheger-bosch (approved):

oheger-bosch (dismissed):

scanner/src/test/kotlin/scanners/fossid/FossIdConfigTest.kt
```diff
@@ -52,6 +52,7 @@ class FossIdConfigTest : WordSpec({
                 "waitForResult" to "false",
                 "deltaScans" to "true",
                 "deltaScanLimit" to "42",
+                "timeout" to "300",
```
sschuberth, line 55: Is the timeout now really in minutes, so this is 300 minutes?

oheger-bosch (approved):

scanner/src/test/kotlin/scanners/fossid/FossIdConfigTest.kt
nnobelis: The timeout is now in minutes so 300 minutes -> 5 hours.
Or did I understood the question wrong ?

sschuberth: Yes, I was just wondering about that huge timeout.

sschuberth (approved):

PR #4695, opened 2021-11-12
Improvements to Gradle's system properties handling

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Gradle.kt
+++ b/analyzer/src/main/kotlin/managers/Gradle.kt
@@ -123,12 +123,25 @@ class Gradle(
     private val dependencyHandler = GradleDependencyHandler(managerName, maven)
     private val graphBuilder = DependencyGraphBuilder(dependencyHandler)
 
+    // The path to the root project. In a single-project, just points to the project path.
+    private lateinit var rootProjectDir: File
+
     override fun createPackageManagerResult(projectResults: Map<File, List<ProjectAnalyzerResult>>) =
         PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages())
 
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
         val gradleSystemProperties = mutableListOf<Pair<String, String>>()
         val gradleProperties = mutableListOf<Pair<String, String>>()
+
+        val projectDir = definitionFile.parentFile
+        val isRootProject = GRADLE_SETTINGS_FILES.any { projectDir.resolve(it).isFile }
+
+        // TODO: Improve the logic to work for independent projects that are stored in a directory below another
+        //       independent project.
+        val isIndependentProject = !this::rootProjectDir.isInitialized || !projectDir.startsWith(rootProjectDir)
+
+        // Do not reset the root project directory for subprojects.
+        if (isRootProject || isIndependentProject) rootProjectDir = projectDir
 
         // Usually, the Gradle wrapper's Java code handles applying system properties defined in a Gradle properties
         // file. But as we use the Gradle Tooling API instead of the wrapper to start the build, we need to manually
@@ -186,10 +199,9 @@ class Gradle(
             jvmArgs += "$JAVA_MAX_HEAP_SIZE_OPTION$JAVA_MAX_HEAP_SIZE_VALUE"
         }
 
-        val projectDir = definitionFile.parentFile
         val gradleConnection = gradleConnector.forProjectDirectory(projectDir).connect()
 
-        return temporaryProperties(*gradleSystemProperties.toTypedArray()) {
+        return temporaryProperties("user.dir" to rootProjectDir.path, *gradleSystemProperties.toTypedArray()) {
             gradleConnection.use { connection ->
                 val initScriptFile = File.createTempFile("init", ".gradle")
                 initScriptFile.writeBytes(javaClass.getResource("/scripts/init.gradle").readBytes())
--- a/analyzer/src/main/kotlin/managers/Gradle.kt
+++ b/analyzer/src/main/kotlin/managers/Gradle.kt
@@ -130,7 +130,6 @@ class Gradle(
         PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages())
 
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
-        val gradleSystemProperties = mutableListOf<Pair<String, String>>()
         val gradleProperties = mutableListOf<Pair<String, String>>()
 
         val projectDir = definitionFile.parentFile
@@ -143,36 +142,14 @@ class Gradle(
         // Do not reset the root project directory for subprojects.
         if (isRootProject || isIndependentProject) rootProjectDir = projectDir
 
-        // Usually, the Gradle wrapper's Java code handles applying system properties defined in a Gradle properties
-        // file. But as we use the Gradle Tooling API instead of the wrapper to start the build, we need to manually
-        // load any system properties from a Gradle properties file and set them in the process that uses the Tooling
-        // API. A typical use case for this is to apply proxy settings so that the Gradle distribution used by the build
-        // can be downloaded behind a proxy, see https://github.com/gradle/gradle/issues/6825#issuecomment-502720562.
-        // For simplicity, limit the search for system properties to the current user's Gradle properties file for now.
         val userPropertiesFile = GRADLE_USER_HOME.resolve("gradle.properties")
         if (userPropertiesFile.isFile) {
             userPropertiesFile.inputStream().use {
                 val properties = Properties().apply { load(it) }
 
-                properties.mapNotNullTo(gradleSystemProperties) { (key, value) ->
-                    val systemPropKey = (key as String).removePrefix("systemProp.")
-                    (systemPropKey to (value as String)).takeIf { systemPropKey != key }
-                }
-
                 properties.mapNotNullTo(gradleProperties) { (key, value) ->
                     ((key as String) to (value as String)).takeUnless { key.startsWith("systemProp.") }
                 }
-            }
-
-            log.debug {
-                "Will apply the following system properties defined in file '$userPropertiesFile':" +
-                        gradleSystemProperties.joinToString(separator = "\n\t", prefix = "\n\t") {
-                            "${it.first} = ${it.second}"
-                        }
-            }
-        } else {
-            log.debug {
-                "Not applying any system properties as no '$userPropertiesFile' file was found."
             }
         }
 
@@ -201,7 +178,7 @@ class Gradle(
 
         val gradleConnection = gradleConnector.forProjectDirectory(projectDir).connect()
 
-        return temporaryProperties("user.dir" to rootProjectDir.path, *gradleSystemProperties.toTypedArray()) {
+        return temporaryProperties("user.dir" to rootProjectDir.path) {
             gradleConnection.use { connection ->
                 val initScriptFile = File.createTempFile("init", ".gradle")
                 initScriptFile.writeBytes(javaClass.getResource("/scripts/init.gradle").readBytes())


fviernau (changes requested):

analyzer/src/main/kotlin/managers/Gradle.kt
```diff
@@ -186,7 +186,7 @@ class Gradle(
         val projectDir = definitionFile.parentFile
         val gradleConnection = gradleConnector.forProjectDirectory(projectDir).connect()
 
-        return temporaryProperties(*gradleSystemProperties.toTypedArray()) {
+        return temporaryProperties("user.dir" to projectDir.absolutePath, *gradleSystemProperties.toTypedArray()) {
```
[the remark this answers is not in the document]
sschuberth: I'm not sure, I'd need to check. Is this your indirect way of suggesting that this property should only be set to the *root* project directory, but not to the directory of the respective subprojects?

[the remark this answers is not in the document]
fviernau: > Is this your indirect way of suggesting that this property should only be set to the root project directory, but not to the directory of the respective subprojects?

It wasn't meant as suggestion. I wouldn't know right away what to suggest. So I asked in the hope that the answer is such that there is no need to think about that case any further. (I was afraid that it might break something)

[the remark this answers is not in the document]
sschuberth: > this property should only be set to the root project directory

But actually this is indeed the case, so thank for making me think about that. I've pushed a new iteration that addressed this.

analyzer/src/main/kotlin/managers/Gradle.kt
```diff
@@ -120,13 +120,25 @@ class Gradle(
...
         val gradleSystemProperties = mutableListOf<Pair<String, String>>()
         val gradleProperties = mutableListOf<Pair<String, String>>()
 
+        val projectDir = definitionFile.parentFile
+        val isRootProject = GRADLE_SETTINGS_FILES.any { projectDir.resolve(it).isFile }
+        if (isRootProject) {
+            currentRootProjectDir = projectDir
+        } else {
+            val isIndependentProject = currentRootProjectDir?.let { projectDir.relativeToOrNull(it) } == null
+            if (isIndependentProject) currentRootProjectDir = null
```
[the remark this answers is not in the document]
sschuberth: > an independent project doesn't have a root project

That was exactly my rationale.

analyzer/src/main/kotlin/managers/Gradle.kt
```diff
@@ -120,13 +120,24 @@ class Gradle(
...
     override fun createPackageManagerResult(projectResults: Map<File, List<ProjectAnalyzerResult>>) =
         PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages())
 
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
         val gradleSystemProperties = mutableListOf<Pair<String, String>>()
         val gradleProperties = mutableListOf<Pair<String, String>>()
 
+        val projectDir = definitionFile.parentFile
+        val isRootProject = GRADLE_SETTINGS_FILES.any { projectDir.resolve(it).isFile }
+        val isIndependentProject = currentRootProjectDir?.let { projectDir.relativeToOrNull(it) } == null
```
MarcelBochtler, line 137: According to the documentation of [relativeToOrNull](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/relative-to-or-null.html) it returns a File with an empty path if `projectDir` matches the base file (in this case `it`).
This would be the case for an independent project and evaluate to false with `== null`.

sschuberth: > Wouldn't this be the case for an independent project and evaluate to false with `== null`?

I don't think so. Before this line, it should never be the case that `currentRootProjectDir == projectDir` because:

1. `currentRootProjectDir` might be `null`, but `projectDir` can't be `null`
2. `currentRootProjectDir` might point to the project directory of the previously resolved project, but `projectDir` always points to the directory of the current project to be resolved

sschuberth: Somewhat unrelated, but I could also rewrite this to

    val isIndependentProject = currentRootProjectDir?.let { projectDir.startsWith(it) } != true

Is that any better?

sschuberth: Somewhat unrelated, but I could also rewrite this to

    val isIndependentProject = currentRootProjectDir?.let { projectDir.startsWith(it) } != true

Is that any better?

MarcelBochtler: Yes I think that is clearer.

MarcelBochtler (dismissed):

analyzer/src/main/kotlin/managers/Gradle.kt
```diff
@@ -120,13 +120,24 @@ class Gradle(
     private val dependencyHandler = GradleDependencyHandler(managerName, maven)
     private val graphBuilder = DependencyGraphBuilder(dependencyHandler)
 
+    // The path to the root project. In a single-project, just points to the project path. Is only null when not yet
```
fviernau, line 125: commit message: This might fix also the issue when you call the analyzer from outside the project dir and the build script uses git commands, assuming the working dir is inside the clone of the git repo. Parsing the results of git commands can have unexpected input and make the build script fail. 

Do you think that's worth mentioning?

analyzer/src/main/kotlin/managers/Gradle.kt
```diff
@@ -120,13 +120,24 @@ class Gradle(
...
+    // initialized.
+    private var rootProjectDir: File? = null
```
fviernau, line 127: Would it make sense to move the variable into `resolveDependencies()´ to avoid introducing a state?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Gradle.kt
sschuberth: Unfortunately, we *need* to introduce state in order to "remember" the current root project path across `resolveDependencies()` calls.

analyzer/src/main/kotlin/managers/Gradle.kt
sschuberth: > the build script uses git commands

Yes, this is indeed also fixed by this change. I've generalized the commit message a bit in that regard.

analyzer/src/main/kotlin/managers/Gradle.kt
fviernau: > Unfortunately, we _need_ to introduce state in order to "remember" the current root project path across `resolveDependencies()` calls.

I see, thanks.

I've researched a bit if there is any other way to determine the root project dir. After reading [1] I guess it may be possible to do something like the following (taken from [1]) but I haven't tried it:

`
ProjectConnection connection = GradleConnector.newConnector()
     .forProjectDirectory(getProjectDirPath().toFile())
     .connect();
project = connection.getModel(EclipseProject.class);
follow project.getParent() towards root
obtain the path
`

I'm not sure whether that above approach works at all and whether it would run itself into the issue you are trying to fix.

[1] https://www.programcreek.com/java-api-examples/?api=org.gradle.tooling.GradleConnector

Note: what I don't yet understand about this commit is what exact assumption about the ordering of the calls to `resolvedDependencies()` is introduced. And how it works for a repository with multiple root projects. Mind giving a hint?

sschuberth: > what I don't yet understand about this commit is what exact assumption about the ordering of the calls to resolvedDependencies() is introduced.

IMO there's no assumption being *introduced*; but the change relies on the fact that `PackageManager.findManagedFiles()` returns files in depth-first order.

fviernau: > relies on the fact that `PackageManager.findManagedFiles()` returns files in depth-first order.

Thanks. IMO it'd be helpful to mention it in the comment above `private var rootProjectDir: File? = null`, as it needs quite a bit digging outside of the class to figure out where the order comes from.

sschuberth: > `ProjectConnection connection = GradleConnector.newConnector()`

My fear is that we might run into a chicken-and-egg problem here: Connection to / evaluation / configuration of the project might fail if the `user.dir` property is not set as expected by the project, so we cannot use that approach to determine what `user.dir` needs to be set to.

fviernau: I still do not fully get whether the variable is always correctly reset. 

`
build.gradle
independent/build.gradle
child/gradle
`

If it processes in the order `build.gradle`, `independent/build.gradle`, `child/gradle`, would then the `rootProjectDir` be set to `independent` when processing `child`?

sschuberth: > would then the `rootProjectDir` be set to `independent` when processing `child`?

No, the `rootProjectDir` stays at `./build.gradle`, because the `rootProjectDir` is only ever "reset" if a project is a root project, or an independent project / non-subproject.

fviernau: > > would then the `rootProjectDir` be set to `independent` when processing `child`?
> 
> No, the `rootProjectDir` stays at `./build.gradle`, because the `rootProjectDir` is only ever "reset" if a project is a root project, or an independent project / non-subproject.

My assumption was of course that `independent/build.gradle` is an independent / non-suproject. Have you considered this when writing above reply?

sschuberth: > > No, the `rootProjectDir` stays at `./build.gradle`, because the `rootProjectDir` is only ever "reset" if a project is a root project, or an independent project / non-subproject.
> 
> My assumption was of course that `independent/build.gradle` is an independent / non-suproject. Have you considered this when writing above reply?

No.

sschuberth:
I've split out uncontroversial parts to https://github.com/oss-review-toolkit/ort/pull/4709 to move this forward.

analyzer/src/main/kotlin/managers/Gradle.kt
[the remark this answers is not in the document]
sschuberth: > even single-project Gradle project have a root project in the Gradle data model, which just points to itself.

That's indeed the case. I've changed the implementation accordingly.

analyzer/src/main/kotlin/managers/Gradle.kt
sschuberth: So, coming back to your original question, now that I got that the "independent" is not supposed to be a subproject of the root project:

> would then the `rootProjectDir` be set to `independent` when processing `child`?

No, but instead the `rootProjectDir` would be (wrongly) set to the directory containing the root "build.gradle" when processing "independent". So the project that gets the wrong `rootProjectDir` would be "independent", not "child". I don't currently see a feasible approach to improve this.

However, with the currently code "independent" would also get the wrong `rootProjectDir` (and all other projects, too). So from that perspective the current implementation still does the right thing for a sane project setup.

fviernau: > However, with the currently code "independent" would also get the wrong `rootProjectDir` (and all other projects, too). So from that perspective the current implementation still does the right thing for a sane project setup.

Do you mean, that the PR as-is fixes the issue in some case. In other cases it does not have any (negative) effect?

sschuberth: Yes. It's not worse than before in some edge cases, but better than before in most cases, IMO.

fviernau: I'm ok if this was merged. I'd prefer to have some small comment in the code to highlight that this intentionally does not work in all cases.

sschuberth: I just had added a `TODO`. Is its wording fine for you?

fviernau (dismissed):

analyzer/src/main/kotlin/managers/Gradle.kt
fviernau: Yes, thanks!

fviernau (approved):

sschuberth:

Issue #4699, opened 2021-11-15
analyzer/NPM: Add support for ignore-scripts parameter when `npm ci` is used


rockebee:
Introduced with #273, the "--ignore-scripts" option is used to prevent execution of any build scripts of the node.js project or its dependencies. However, this flag is not (yet) used for `npm ci` cases (introduced with #877), but IMHO it is totally valid to do so as same reasoning applies here (prevent execution of arbitrary build scripts + high risk of failing because the scripts could try to compile native code).

Caveat: `ignore-scripts` was introduced with NPM 5.7.0 (together with `npm ci` itself), but was not supported in NPM versions 7.0.0 < 7.4.0 ([link](https://github.com/npm/cli/pull/2455))

If there are no objections, I would be able to prepare a PR for this.

PR #4700, opened 2021-11-15
#4699: analyzer/npm :: Add usage of --ignore-scripts parameter to "npm ci" in order to prevent unnecessary and risky script execution

rockebee:
PR attempts to use `--ignore-scripts` parameter also in `npm ci` cases, not only for `npm install`.
(cmp. #4699 for respective issue description)

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -649,7 +649,7 @@ open class Npm(
 
         // Install all NPM dependencies to enable NPM to list dependencies.
         if (hasLockFile(workingDir) && this::class.java == Npm::class.java) {
-            run(workingDir, "ci")
+            run(workingDir, "ci", *installParameters)
         } else {
             run(workingDir, "install", *installParameters)
         }


mnonnenmacher (changes requested):
@rockebee Thank you for the contribution. Please sign your commit message (see the failing DCO check) and reformat it to match our standards, for example like so:

```
Npm: Add `--ignore-scripts` to `npm ci`

Add usage of `--ignore-scripts` to the `npm ci` call in order to prevent
unnecessary and risky script execution.

Fixes #4699.
```

sschuberth:
Quoting @rockebee from #4699:

> Caveat: `ignore-scripts` was introduced with NPM 5.7.0 (together with `npm ci` itself), but was not supported in NPM versions 7.0.0 < 7.4.0 ([link](https://github.com/npm/cli/pull/2455))

So, should we change the code to only pass `--ignore-scripts` to `npm ci` for version 5.7.0 to 7.0.0, or versions above 7.4.0?

mnonnenmacher:
> 
> 
> Quoting @rockebee from #4699:
> 
> > Caveat: `ignore-scripts` was introduced with NPM 5.7.0 (together with `npm ci` itself), but was not supported in NPM versions 7.0.0 < 7.4.0 ([link](https://github.com/npm/cli/pull/2455))
> 
> So, should we change the code to only pass `--ignore-scripts` to `npm ci` for version 5.7.0 to 7.0.0, or versions above 7.4.0?

I understood the PR so that the option was still present but simply not working from 7.0.0-7.4.0.

sschuberth:
> I understood the PR so that the option was still present but simply not working from 7.0.0-7.4.0.

Ah, ok. Then of course the code should be fine as-is.

rockebee:
@mnonnenmacher @sschuberth, thx for feedback. Commit message is now signed properly and commit message modified as suggested.
I agree that IMHO there is no need to exclude NPM versions 7.0.0 to < 7.4.0, as `npm ci` itself is properly working, just the `--ignore-scripts` flag was not working and ignored.

sschuberth:
Nit: There needs to be a blank line between

    Fixes #4699.
    Signed-off-by: Stefan Schwenkenbecher <38685542+rockebee@users.noreply.github.com>

to separate the body from the footer.

sschuberth (approved):

PR #4705, opened 2021-11-16
Disable running detekt as part of Lift

sschuberth:
Detekt is already run separately, and this way detekt's "SpreadOperator"
rule does not need to be disabled via Lift.


mnonnenmacher (approved):

sschuberth:

PR #4706, opened 2021-11-16
Remove the option to ignore tool versions in favor of only logging a warning

sschuberth:

### DIFF
--- a/model/src/main/kotlin/config/AnalyzerConfiguration.kt
+++ b/model/src/main/kotlin/config/AnalyzerConfiguration.kt
@@ -25,13 +25,6 @@ import com.fasterxml.jackson.annotation.JsonInclude
 @JsonInclude(JsonInclude.Include.NON_NULL)
 data class AnalyzerConfiguration(
     /**
-     * If set to true, ignore the versions of used command line tools. Note that this might lead to erroneous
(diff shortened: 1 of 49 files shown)


model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -24,13 +24,6 @@ import com.fasterxml.jackson.annotation.JsonInclude
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 data class AnalyzerConfiguration(
```
mnonnenmacher, line 26: Does Hoplite fail if the config file contains unexpected options? If so the PR needs the breaking change label.

mnonnenmacher (approved):

model/src/main/kotlin/config/AnalyzerConfiguration.kt
sschuberth: > Does Hoplite fail if the config file contains unexpected options?

No, it simply ignores them. We used that feature in the past to already introduce upcoming options to our config before they were implemented upstream, to have a smooth transition.

mnonnenmacher (dismissed):

mnonnenmacher (approved):

sschuberth:

PR #4714, opened 2021-11-17
CycloneDxReporter: Add component homepages as external references

sschuberth:


sschuberth:
FYI @schuerle.

reporter/src/funTest/kotlin/TestData.kt
```diff
@@ -109,7 +109,7 @@ val ORT_RESULT = OrtResult(
                     definitionFilePath = "project-without-findings/package.json",
                     declaredLicenses = sortedSetOf(),
                     vcs = VcsInfo.EMPTY,
-                    homepageUrl = "",
+                    homepageUrl = "https://github.com/oss-review-toolkit/ort",
```
oheger-bosch, line 112: Ideally, the URLs should be slightly different, so that they can be distinguished in tests. For now, they do not seem to be referenced in a test yet though (nudge).

oheger-bosch (approved):

reporter/src/funTest/kotlin/TestData.kt
sschuberth: > Ideally, the URLs should be slightly different, so that they can be distinguished in tests.

I agree, but failed to come up with a nice general scheme to deduce the URL from the project / package, and I also didn't want them to be completely random.

> For now, they do not seem to be referenced in a test yet though (nudge).

Also correct, I just needed them for debugging for now 😁

PR #4724, opened 2021-11-19
Align on using find instead of firstOrNull

mnonnenmacher:
`find` is an alias for `firstOrNull`. Since `find` is already used far
mor often in the codebase, align on using `find`.


analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -153,7 +153,7 @@ abstract class PackageManager(
 
             val fallbackVcs = fallbackUrls.mapTo(mutableListOf(VcsHost.toVcsInfo(normalizedVcsFromPackage.url))) {
                 VcsHost.toVcsInfo(normalizeVcsUrl(it))
-            }.firstOrNull {
+            }.find {
```
sschuberth, line 156: Commit message: s/mor/more/

sschuberth (changes requested):

sschuberth (approved):

PR #4727, opened 2021-11-22
OrtResult: Add helper function to get all open issues 

MarcelBochtler:
This helper is mainly required for rules and notifications.

### DIFF
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -442,6 +442,15 @@ data class OrtResult(
         }
 
     /**
+     * Retrieve all unresolved and non-excluded issues with severities equal to or over [minSeverity].
+     */
+    @JsonIgnore
+    fun getOpenIssues(minSeverity: Severity = Severity.WARNING) = collectIssues()
+        .mapNotNull { (id, issues) -> issues.takeUnless { isExcluded(id) } }
+        .flatten()
+        .filter { issue -> issue.severity >= minSeverity && getResolutions().issues.none { it.matches(issue) } }
+
+    /**
      * Return the [Resolutions] contained in the repository configuration of this [OrtResult].
      */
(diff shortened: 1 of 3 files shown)


model/src/main/kotlin/OrtResult.kt
```diff
@@ -441,6 +441,16 @@ data class OrtResult(
...
 
+    /**
+     * Retrieve all unresolved and non-excluded issues with severities equal to or over [minSeverity].
+     */
+    @JsonIgnore
+    fun getOpenIssues(minSeverity: Severity = Severity.WARNING) = collectIssues()
+        .filterNot { (id, _) -> isExcluded(id) }
+        .values
+        .flatten()
+        .filter { issue -> getResolutions().issues.none { it.matches(issue) } && issue.severity >= minSeverity }
```
oheger-bosch, line 452: Minor optimization: The filter may be slightly more efficient if you switch the conditions, i.e. check first for the severity threshold and then for the presence of resolutions. The severity check is just a comparison while the other check is more involved.

oheger-bosch (approved):

model/src/main/kotlin/OrtResult.kt
```diff
@@ -441,6 +441,16 @@ data class OrtResult(
             emptyList()
         }
 
+    /**
+     * Retrieve all unresolved and non-excluded issues with severities equal to or over [minSeverity].
+     */
+    @JsonIgnore
+    fun getOpenIssues(minSeverity: Severity = Severity.WARNING) = collectIssues()
+        .filterNot { (id, _) -> isExcluded(id) }
+        .values
```
sschuberth, line 450: These could be combined via

    .mapNotNull { (id, issues) -> issues.takeUnless { isExcluded(id) } }

oheger-bosch (approved):

PR #4730, opened 2021-11-22
Add more JGit / netrc related debugging

sschuberth:

### DIFF
--- a/utils/core/src/main/kotlin/OrtAuthenticator.kt
+++ b/utils/core/src/main/kotlin/OrtAuthenticator.kt
@@ -122,7 +122,7 @@ class OrtAuthenticator(private val original: Authenticator? = null) : Authentica
  * Parse the [contents] of a [.netrc](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html)
  * file for a login / password matching [machine].
  */
-fun getNetrcAuthentication(contents: String, machine: String): PasswordAuthentication? {
+internal fun getNetrcAuthentication(contents: String, machine: String): PasswordAuthentication? {
     val lines = contents.lines().mapNotNull { line ->
         line.trim().takeUnless { it.startsWith('#') }
     }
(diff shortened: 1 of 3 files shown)


utils/core/src/main/kotlin/OrtAuthenticator.kt
```diff
@@ -129,19 +130,22 @@ internal fun getNetrcAuthentication(contents: String, machine: String): Password
 
     val iterator = lines.joinToString(" ").split(Regex("\\s+")).iterator()
 
-    var machineFound = false
+    var machineFound: String? = null
```
fviernau, line 133: How about naming this variable 'machine'?

utils/core/src/main/kotlin/OrtAuthenticator.kt
```diff
@@ -129,19 +130,22 @@ internal fun getNetrcAuthentication(contents: String, machine: String): Password
...
+            "login" -> login = machineFound?.let { iterator.nextOrNull() }
+            "password" -> password = machineFound?.let { iterator.nextOrNull() }
+            "default" -> machineFound = token
         }
 
-        if (login != null && password != null) return PasswordAuthentication(login, password.toCharArray())
+        if (login != null && password != null) {
+            OrtAuthenticator.log.debug { "Found a '$machineFound' entry for '$machine'." }
+            return PasswordAuthentication(login, password.toCharArray())
+        }
```
fviernau, line 148: In order to allow a bit more detailed debugging, how instead logging in lines 140 / 141 like the following?

"Found login $login for machine $machine."
"Found a password for machine $machine."

fviernau (approved):

utils/core/src/main/kotlin/OrtAuthenticator.kt
sschuberth: That would conflict with the function parameter of the same name.

utils/core/src/main/kotlin/OrtAuthenticator.kt
sschuberth: I didn't want to expose that much detail yet. Might be, though, that I'm forced to in a follow-up, if the current additions do not suffice for my current debugging.

utils/core/src/main/kotlin/OrtAuthenticator.kt
fviernau: I missed that, thanks!

PR #4733, opened 2021-11-23
Vulnerability report: Improve error handling

oheger-bosch:
If no vulnerabilities can be retrieved due to errors, the vulnerability report is currently misleading, since it only lists packages without further information. ORT issues are not shown, so there can be the false impression that no vulnerabilities have been found.

This PR changes the template for the vulnerability report to show a prominent warning if there were errors during the advisor run, so that it is obvious that the report is incomplete. Also, the issues assigned to the single packages are now shown.

When deciding whether or not to display this warning, there is the problem that there are now advisor implementations retrieving different findings, e.g. defects. The warning is, however, relevant only if an advisor for vulnerabilities is affected. To handle this, the PR extends the `AdvisorDetails` class to contain the advisor's _capabilities_, i.e. the type of findings the advisor produces. That way it can be figured out whether errors of a specific advisor impact the vulnerabilities report or not.

### DIFF
--- a/model/src/main/kotlin/AdvisorDetails.kt
+++ b/model/src/main/kotlin/AdvisorDetails.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Bosch.IO GmbH
+ * Copyright (C) 2020-2021 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,10 @@
 
 package org.ossreviewtoolkit.model
 
+import java.util.EnumSet
+
+import org.ossreviewtoolkit.utils.common.enumSetOf
+
 /**
  * Details about the used provider of vulnerability information.
  */
@@ -26,7 +30,13 @@ data class AdvisorDetails(
     /**
      * The name of the used advisor.
      */
-    val name: String
+    val name: String,
+
+    /**
+     * The capabilities of the used advisor. This property indicates, which kind of findings are retrieved by the
+     * advisor.
+     */
+    val capabilities: EnumSet<AdvisorCapability> = enumSetOf()
 ) {
     companion object {
         val EMPTY = AdvisorDetails(
--- a/reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
+++ b/reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.reporter.reporters.freemarker
 
 import freemarker.cache.ClassTemplateLoader
+import freemarker.ext.beans.BeansWrapperBuilder
 import freemarker.template.Configuration
 import freemarker.template.DefaultObjectWrapper
 import freemarker.template.TemplateExceptionHandler
@@ -27,6 +28,7 @@ import freemarker.template.TemplateExceptionHandler
 import java.io.File
 import java.util.SortedMap
 
+import org.ossreviewtoolkit.model.AdvisorCapability
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtIssue
 import org.ossreviewtoolkit.model.OrtResult
@@ -122,7 +124,7 @@ class FreemarkerTemplateProcessor(
             "helper" to TemplateHelper(input),
             "projectsAsPackages" to projectsAsPackages,
             "vulnerabilityReference" to VulnerabilityReference
-        )
+        ) + enumModel()
 
         val freemarkerConfig = Configuration(Configuration.VERSION_2_3_30).apply {
             defaultEncoding = "UTF-8"
@@ -177,6 +179,18 @@ class FreemarkerTemplateProcessor(
         }
 
         return outputFiles
+    }
+
+    /**
+     * Return a map with wrapper beans for the enum classes that are relevant for templates.These enums can then be
+     * referenced directly by templates.
+     * See https://freemarker.apache.org/docs/pgui_misc_beanwrapper.html#jdk_15_enums.
+     */
+    private fun enumModel(): Map<String, Any> {
+        val beansWrapper = BeansWrapperBuilder(Configuration.VERSION_2_3_30).build()
+        val enumModels = beansWrapper.enumModels
+
+        return ENUM_CLASSES.associate { it.simpleName to enumModels.get(it.name) }
     }
 
     /**
@@ -334,9 +348,29 @@ class FreemarkerTemplateProcessor(
          */
         @Suppress("UNUSED") // This function is used in the templates.
         fun filterForUnresolvedVulnerabilities(vulnerabilities: List<Vulnerability>): List<Vulnerability> =
-                vulnerabilities.filterNot { input.resolutionProvider.isResolved(it) }
+            vulnerabilities.filterNot { input.resolutionProvider.isResolved(it) }
+
+        /**
+         * Return a flag indicating that issues have been encountered during the run of an advisor with the given
+         * [capability] with at least the given [severity]. This typically means that the report is incomplete;
+         * therefore, it should contain a corresponding warning.
+         */
+        fun hasAdvisorIssues(capability: AdvisorCapability, severity: Severity): Boolean =
+            input.ortResult.advisor?.results?.advisorResults.orEmpty()
+                .values
+                .flatten()
+                .filter { capability in it.advisor.capabilities }
+                .any { result -> result.summary.issues.any { it.severity >= severity } }
     }
 }
+
+/**
+ * A list of the enum classes that are made available to templates.
+ */
+private val ENUM_CLASSES = listOf(
+    AdvisorCapability::class.java,
+    Severity::class.java
+)
 
 private fun List<ResolvedLicense>.merge(): ResolvedLicense {
     require(isNotEmpty()) { "Cannot merge an empty list." }
--- a/advisor/src/main/kotlin/AdviceProvider.kt
+++ b/advisor/src/main/kotlin/AdviceProvider.kt
@@ -45,6 +45,11 @@ abstract class AdviceProvider(val providerName: String) {
     ): Map<Package, List<AdvisorResult>>
 
     /**
+     * An object with detail information about this [AdviceProvider].
+     */
+    abstract val details: AdvisorDetails
+
+    /**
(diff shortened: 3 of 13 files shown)


sschuberth:
> show a prominent warning if there were errors

Would it be possible to get a screenshot posted here?

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -335,6 +336,23 @@ class FreemarkerTemplateProcessor(
...
         fun filterForUnresolvedVulnerabilities(vulnerabilities: List<Vulnerability>): List<Vulnerability> =
                 vulnerabilities.filterNot { input.resolutionProvider.isResolved(it) }
+
+        /**
+         * Return a flag indicating that issues have been encountered during the run of an advisor with the given
+         * [capability][capabilityStr] with at least the given [severity][severityStr]. This typically means that
+         * the report is incomplete; therefore, it should contain a corresponding warning.
+         */
+        @JvmOverloads
+        fun hasAdvisorIssues(capabilityStr: String, severityStr: String = Severity.ERROR.name): Boolean {
```
sschuberth, line 346: Is it really necessary that this function takes the string representations instead of the typed values? IMO we should rather make the `enum` classes available to the template.

oheger-bosch:
Here are screenshots of the new warning section:
vulnerability_report_warning

oheger-bosch:
and the issues displayed for the single packages:
vulnerability_report_issues

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
sschuberth: Maybe this helps:
* https://freemarker.apache.org/docs/pgui_misc_beanwrapper.html#jdk_15_enums
* https://stackoverflow.com/a/44723453/1127485

sschuberth (changes requested):

model/src/main/kotlin/AdvisorDetails.kt
```diff
@@ -26,7 +26,13 @@ data class AdvisorDetails(
...
      * The name of the used advisor.
      */
-    val name: String
+    val name: String,
+
+    /**
+     * The capabilities of the used advisor. This property indicates, which kind of findings are retrieved by the
+     * advisor.
+     */
+    val capabilities: Set<AdvisorCapability> = emptySet()
```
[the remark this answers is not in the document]
oheger-bosch: The default was introduced, because otherwise old result files cannot be read any more. IMHO, this would be a high price to pay for this new feature.

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
oheger-bosch: The BeansWrapper mechanism is now used to make the enums available to the template.

model/src/main/kotlin/AdvisorDetails.kt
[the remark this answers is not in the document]
oheger-bosch: While I agree that each advisor should specify a capability, I think a default is needed to prevent that older result files cannot be loaded any more.

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -179,6 +183,20 @@ class FreemarkerTemplateProcessor(
...
+     * directly by templates.
+     * See https://freemarker.apache.org/docs/pgui_misc_beanwrapper.html#jdk_15_enums.
+     */
+    private fun makeEnumsAvailable(dataModel: Map<String, Any>): Map<String, Any> {
+        val beansWrapper = BeansWrapperBuilder(Configuration.VERSION_2_3_30).build()
+        val enumModels = beansWrapper.enumModels
+
+        val enumBeans = ENUM_CLASSES.associate { it.simpleName to enumModels.get(it.name) }
+
+        return dataModel + enumBeans
```
sschuberth, line 197: Is there a design reason why this already returns the union of `dataModel + enumBeans`? I think I'd prefer this to only return `enumBeans` (that variable could then be inlined) and write at the call site `val dataModel = mapOf(...) + enumModels()` (while renaming this function to `enumModels()` and making it take no arguments).

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -335,9 +353,30 @@ class FreemarkerTemplateProcessor(
...
         fun filterForUnresolvedVulnerabilities(vulnerabilities: List<Vulnerability>): List<Vulnerability> =
                 vulnerabilities.filterNot { input.resolutionProvider.isResolved(it) }
+
+        /**
+         * Return a flag indicating that issues have been encountered during the run of an advisor with the given
+         * [capability] with at least the given [severity]. This typically means that the report is incomplete;
+         * therefore, it should contain a corresponding warning.
+         */
+        @JvmOverloads
+        fun hasAdvisorIssues(capability: AdvisorCapability, severity: Severity = Severity.ERROR): Boolean =
```
sschuberth, line 363: Nit: Consider dropping the default argument for `severity`. IMO "hasAdvisorIssues" reads as if by default *any* issue should be returned, not just errors.

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -335,9 +353,30 @@ class FreemarkerTemplateProcessor(
...
+            input.ortResult.advisor?.results?.advisorResults.orEmpty()
+                .values
+                .flatten()
+                .filter { capability in it.advisor.capabilities }
+                .any { result -> result.summary.issues.any { it.severity.ordinal >= severity.ordinal } }
```
sschuberth, line 368: There's no need to compare ordinals, we have defined ordering as of https://github.com/oss-review-toolkit/ort/pull/3956.

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
sschuberth: Actually, I'd expect the hard-coded `ENUM_CLASSES` to be a parameter of this function instead, so it gets called like `enumModels(ENUM_CLASSES)`.

sschuberth (changes requested):

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -113,16 +115,17 @@ class FreemarkerTemplateProcessor(
...
-            "packages" to packages,
-            "ortResult" to input.ortResult,
-            "licenseTextProvider" to input.licenseTextProvider,
-            "LicenseView" to LicenseView,
-            "helper" to TemplateHelper(input),
-            "projectsAsPackages" to projectsAsPackages,
-            "vulnerabilityReference" to VulnerabilityReference
-        )
+        val dataModel =
+            mapOf(
```
sschuberth, line 119: Nit: Formatting should stay at `val dataModel = mapOf(` and following lines be de-intended (to keep the diff small for review convenience).

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -179,6 +182,18 @@ class FreemarkerTemplateProcessor(
         return outputFiles
     }
 
+    /**
+     * Return a map with wrapper beans for the specified [enumClasses], so that these enums can be referenced
+     * directly by templates.
+     * See https://freemarker.apache.org/docs/pgui_misc_beanwrapper.html#jdk_15_enums.
+     */
+    private fun makeEnumsAvailable(enumClasses: Collection<Class<*>>): Map<String, Any> {
```
sschuberth, line 190: Did you keep the function name intentionally as-is? Because this function does not "*make* enums available" by itself (applying the config does that). So maybe indeed rename this to my proposed `enumModel()` (to resemble `dataModel`), or to something else that comes to your mind.

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -408,7 +443,8 @@ private fun OrtResult.getRepositoryPath(provenance: RepositoryProvenance): Strin
     repository.nestedRepositories.forEach { (path, vcsInfo) ->
         if (vcsInfo.type == provenance.vcsInfo.type
             && vcsInfo.url == provenance.vcsInfo.url
-            && vcsInfo.revision == provenance.resolvedRevision) {
+            && vcsInfo.revision == provenance.resolvedRevision
+        ) {
```
sschuberth, line 447: Unrelated change.

sschuberth (changes requested):

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
oheger-bosch: Renamed the function to `enumModel()` as proposed. The function is now alone responsible for this model and uses the `ENUM_CLASSES` constant for this purpose, no longer expecting it as parameter.

sschuberth: > no longer expecting it as parameter.

Hmm, personally I would have kept it as a parameter and used the constant for it, but ok with me.

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -334,10 +348,30 @@ class FreemarkerTemplateProcessor(
          */
         @Suppress("UNUSED") // This function is used in the templates.
         fun filterForUnresolvedVulnerabilities(vulnerabilities: List<Vulnerability>): List<Vulnerability> =
-                vulnerabilities.filterNot { input.resolutionProvider.isResolved(it) }
+            vulnerabilities.filterNot { input.resolutionProvider.isResolved(it) }
```
sschuberth, line 351: Nit: Unrelated change (should have gone to a separate commit).

sschuberth (approved):
Approving now despite my optional comments.

PR #4734, opened 2021-11-23
JiraNotifier: Add function to create a comment in an issue 

MarcelBochtler:


notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -40,6 +40,16 @@ class JiraNotifier(
     private val restClient: JiraRestClient = AsynchronousJiraRestClientFactory()
         .createWithBasicHttpAuthentication(URI(config.host), config.username, config.password)
 ) {
+    /**
+     * Create a [comment] within the issue specified by the [issueKey].
+     */
+    @Suppress("UNUSED") // This is intended to be used via scripting.
+    fun createComment(issueKey: String, comment: String) {
+        val issue = restClient.issueClient.getIssue(issueKey).claim()
+
```
sschuberth, line 49: Nit: Remove the blank line?

sschuberth (approved):

notifier/src/main/kotlin/modules/JiraNotifier.kt
MarcelBochtler: `restClient.issueClient.addComment...` is the result of the method. To to be consistent with other code using an empty line before a `return` statement, I'd like to keep it as-is.

PR #4737, opened 2021-11-23
helper-cli: Add a command to generate a synthetic analyzer result

fviernau:
The command matches the given packages IDs against the database entries
of the 'scan_results' table, which belongs to the postgres storage, to
determine the provenance. Only the most recent provenance per id and
provenance type is used in case there are many matches, like it happens
e.g. for snapshot artifacts.

The command can be used to trigger scans for arbitrary sets of package
IDs by just passing the file to ORT's `scan` CLI command. That can be
useful for re-scanning with a different ScanCode or for any development
task which requires a synthetic analyzer result for any specific
package ID.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
@@ -1,0 +1,195 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.associate
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.file
+
+import java.sql.Connection
+import java.time.Instant
+
+import org.ossreviewtoolkit.helper.common.ORTH_NAME
+import org.ossreviewtoolkit.helper.common.writeOrtResult
+import org.ossreviewtoolkit.model.AnalyzerResult
+import org.ossreviewtoolkit.model.AnalyzerRun
+import org.ossreviewtoolkit.model.ArtifactProvenance
+import org.ossreviewtoolkit.model.CuratedPackage
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.RemoteArtifact
+import org.ossreviewtoolkit.model.RepositoryProvenance
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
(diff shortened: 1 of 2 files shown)


helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -0,0 +1,187 @@
...
+            projects = sortedSetOf(),
+            packages = packages.mapTo(sortedSetOf()) { CuratedPackage(it.toPackage()) }
+        )
+    )
+)
+
+private fun Collection<ScannedPackage>.filterMaxByDate(): List<ScannedPackage> {
+    fun filterMaxByDate(predicate: (ScannedPackage) -> Boolean) = filter(predicate).groupBy {
+        it.id
+    }.mapNotNull { (_, packages) -> packages.maxByOrNull { it.scanTime } }
```
mnonnenmacher, line 164: Nit - alternative formatting:
```
    fun filterMaxByDate(predicate: (ScannedPackage) -> Boolean) = 
        filter(predicate).groupBy { it.id }.mapNotNull { (_, packages) -> 
            packages.maxByOrNull { it.scanTime } 
        }
```

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -0,0 +1,187 @@
...
+    declaredLicenses = sortedSetOf(),
+    concludedLicense = null,
+    description = "",
+    homepageUrl = "",
+    binaryArtifact = RemoteArtifact.EMPTY,
+    sourceArtifact = if (provenance is ArtifactProvenance) {
+        provenance.sourceArtifact
+    } else {
+        RemoteArtifact.EMPTY
+    },
```
mnonnenmacher, line 181: If it fits in one line maybe:
```
sourceArtifact = provenance.sourceArtifact.takeIf { provenance is ArtifactProvenance } ?: RemoteArtifact.EMPTY
```

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -0,0 +1,187 @@
...
+    vcs = if (provenance is RepositoryProvenance) {
+        provenance.vcsInfo.copy(revision = provenance.resolvedRevision)
+    } else {
+        VcsInfo.EMPTY
+    }
```
mnonnenmacher, line 186: Maybe assign `vcsProcessed` as well to make sure that normalization does not mess with the URL.

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -0,0 +1,187 @@
...
+private fun getScannedPackages(connection: Connection, ids: Collection<Identifier>): List<ScannedPackage> {
+    val query = """
+        SELECT 
+            s.identifier as id,
+            s.scan_result->>'provenance' as provenance,
+            (s.scan_result->'summary'->>'start_time')::timestamp as start_time
+        FROM 
+            scan_results s 
+        WHERE 
+            s.identifier = ANY(?);
```
mnonnenmacher, line 127: Nit: If you add `ORDER BY start_time DESC` you could simplify `filterMaxByDate()`.

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -0,0 +1,187 @@
...
+        val ortResult = createAnalyzerResult(packages)
+
+        println("Writing analyzer result with ${packages.size} packages to '${ortFile.absolutePath}'.")
+        writeOrtResult(ortResult, ortFile)
+    }
+
+    private fun openDatabaseConnection(): Connection {
+        val storageConfig = OrtConfiguration.load(configArguments, configFile).scanner.storages
+            ?.get("postgres") as? PostgresStorageConfiguration
+            ?: throw IllegalArgumentException("postgresStorage not configured.")
```
mnonnenmacher, line 102: As the names in the configuration are arbitrary you could make this more generic with:
```
config.storages?.values?.filterIsInstance<PostgresStorageConfiguration>()?.firstOrNull() ?:  throw IllegalArgumentException("...")
```

mnonnenmacher (changes requested):

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
fviernau: This does not compile as `provenance.sourceArtifact` cannot be accessed before the `is`.

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
fviernau: I prefer the function as-is, e.g. without assuming any ordering.

mnonnenmacher (approved):

PR #4742, opened 2021-11-23
CreateAnalyzerResultCommand: Add the parameter `--scancode-version`

fviernau:
Allow to narrow down the matching of scan result entries in the postgres
storage to a given ScanCode version.

This enables the use case to re-scan a given set of package IDs which
have been scanned before with the specified ScanCode version.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
@@ -87,9 +87,14 @@ internal class CreateAnalyzerResultCommand : CliktCommand(
                 "-P ort.scanner.storages.postgres.schema=testSchema"
     ).associate()
 
+    private val scancodeVersion by option(
+        "--scancode-version",
+        help = "The ScanCode version to match for in the scan results."
+    )
+
     override fun run() {
         val ids = packageIdsFile.readLines().filterNot { it.isBlank() }.map { Identifier(it.trim()) }
-        val packages = openDatabaseConnection().use { getScannedPackages(it, ids) }.filterMaxByDate()
+        val packages = openDatabaseConnection().use { getScannedPackages(it, ids, scancodeVersion) }.filterMaxByDate()
         val ortResult = createAnalyzerResult(packages)
 
         println("Writing analyzer result with ${packages.size} packages to '${ortFile.absolutePath}'.")
@@ -115,7 +120,16 @@ private data class ScannedPackage(
     val scanTime: Instant
 )
 
-private fun getScannedPackages(connection: Connection, ids: Collection<Identifier>): List<ScannedPackage> {
+private fun getScannedPackages(
+    connection: Connection,
+    ids: Collection<Identifier>,
+    scanCodeVersion: String?
+): List<ScannedPackage> {
+    val whereClause = listOfNotNull(
+        "s.identifier = ANY(?)",
+        scanCodeVersion?.let { "s.scan_result->'scanner'->>'version' = '$it'" }
+    ).joinToString(" AND ")
+
     val query = """
         SELECT 
             s.identifier as id,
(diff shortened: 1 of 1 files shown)


helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -87,9 +87,14 @@ internal class CreateAnalyzerResultCommand : CliktCommand(
                 "-P ort.scanner.storages.postgres.schema=testSchema"
     ).associate()
 
+    private val scanCodeVersion by option(
+        "--scan-code-version",
```
sschuberth, line 91: How about naming this `--scancode-version`, like the `scancode` CLI?

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -87,9 +87,14 @@ internal class CreateAnalyzerResultCommand : CliktCommand(
...
+        help = "The scan code version to match for equality against the scan results."
```
sschuberth, line 92: s/scan code/ScanCode/

sschuberth: Also "to match for equality against" sounds a bit clumsy... how about just "to match for in the scan results"?

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -115,15 +120,20 @@ private data class ScannedPackage(
...
     val query = """
         SELECT 
             s.identifier as id,
             s.scan_result->>'provenance' as provenance,
             (s.scan_result->'summary'->>'start_time')::timestamp as start_time
         FROM 
             scan_results s 
-        WHERE 
+        WHERE
+            ${ scanCodeVersion?.let { "s.scan_result->'scanner'->>'version' = '$it' AND" }.orEmpty() }
```
sschuberth, line 136: Just an idea (typing blindly):

    val whereClause = listOfNotNull(
        scanCodeVersion?.let { "s.scan_result->'scanner'->>'version' = '$it'",
        "s.identifier = ANY(?)"
    ).joinToString(" AND ")

    // ....

    WHERE $whereClause;

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
sschuberth: The commit message title still has `--scan-code-version`.

sschuberth:
PS: A failing test will be fixed by my https://github.com/oss-review-toolkit/ort/pull/4743/commits/42f090d884e0e6b798968ac0d8ad9a5fa79f877e.

helper-cli/src/main/kotlin/commands/CreateAnalyzerResultCommand.kt
```diff
@@ -115,16 +120,25 @@ private data class ScannedPackage(
...
 
-private fun getScannedPackages(connection: Connection, ids: Collection<Identifier>): List<ScannedPackage> {
+private fun getScannedPackages(
+    connection: Connection,
+    ids: Collection<Identifier>,
+    scanCodeVersion: String?
+): List<ScannedPackage> {
+    val whereClause = listOfNotNull(
+        "s.identifier = ANY(?)",
+        scanCodeVersion?.let { "s.scan_result->'scanner'->>'version' = '$it'" },
```
sschuberth, line 130: Nit: Please remove the trailing comma.

sschuberth (dismissed):

sschuberth (approved):

sschuberth (changes requested):
Is that the right commit on that branch? It seems to mismatch the PR title at least...

fviernau:
> Is that the right commit on that branch? It seems to mismatch the PR title at least...

Good catch, it was not the right one.

sschuberth (approved):

fviernau:
Merged despite failing unrelated tests.

PR #4746, opened 2021-11-24
Update Dockerfile to use multi-stage builds

heliocastro:
* Uses multi-stage build to reduce final container
* Harmonize the base image to use same OpenJDK base from Eclipse Temurin and Ubuntu Focal LTS
* Make all tooling versions customizable through docker build args
* Each tooling is built in separate components, improving the  process to add or remove specific tools


docker/python.sh
```diff
@@ -0,0 +1,36 @@
...
+export PYENV_ROOT
+
+add_local_path "${PYENV_ROOT}/bin"
+
+eval "$(pyenv init --path)";
+#shellcheck disable=1091
+. "$(pyenv root)"/completions/pyenv.bash;
+
+
+# Needed if you want new pyhon version then default
```
tsteenbe, line 30: Spelling "Needed if you want new pyhon version then default" -> "Needed if you want a newer Python version then the default"

docker/ort-wrapper.sh
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+set -e
+
+#shellcheck disable=SC1091
```
tsteenbe, line 22: Change`#shellcheck disable=SC1091`to `# shellcheck disable=SC1091` aligned with usage in other files.

docker/python.sh
```diff
@@ -0,0 +1,36 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+PYENV_ROOT="/opt/python"
+export PYENV_ROOT
+
+add_local_path "${PYENV_ROOT}/bin"
+
+eval "$(pyenv init --path)";
+#shellcheck disable=1091
```
tsteenbe, line 26: Change`#shellcheck disable=SC1091`to `# shellcheck disable=SC1091` aligned with usage in other files.

scripts/docker_build.sh
```diff
@@ -17,8 +18,17 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
+# Docker from ORT uses buildkit feature
```
tsteenbe, line 21: Change buildkit to BuildKit which is the official name per https://docs.docker.com/develop/develop-images/build_enhancements/

Dockerfile
```diff
@@ -19,166 +20,291 @@
...
+
+# Haskell
+COPY --from=haskellbuild /usr/bin/stack /usr/bin
+
+# Repo and Android
+COPY --from=androidbuild /usr/bin/repo /usr/bin/
+COPY --from=androidbuild /opt/android-sdk /opt/android-sdk
+COPY --from=androidbuild /etc/ort/bash_modules/android.sh /etc/ort/bash_modules/
+
+# Scancode
```
tsteenbe, line 265: Change `Scancode` to `ScanCode` which is official spelling per https://github.com/nexB/scancode-toolkit

docker/android.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Copyright (C) 2021 Helio Chissini de Castro <helio@kde.org>
```
tsteenbe, line 3: Change copyright to `Copyright (C) 2021 BMW CarIT GmbH` so its aligned with other files in this PR.

docker/nodejs.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Copyright (C) 2021 Helio Chissini de Castro <helio@kde.org>
```
tsteenbe, line 3: Change copyright to `Copyright (C) 2021 BMW CarIT GmbH` so its aligned with other files in this PR.

tsteenbe (changes requested):

docker/python.sh
heliocastro: Done

Dockerfile
```diff
@@ -19,166 +20,291 @@
...
+COPY --from=scancodebuild /opt/scancode /opt/scancode
+RUN ln -s /opt/scancode/bin/scancode /usr/bin/scancode \
```
maxhbr, line 267: maybe also link `extractcode`, it is a potentially useful tool that could be exposed.

Dockerfile
```diff
@@ -19,166 +20,291 @@
...
+    && rm -rf /opt/scancode/thirdparty
+
+#------------------------------------------------------------------------
+# Main container
+FROM eclipse-temurin:11-jre
+
+# ORT
+COPY --from=ortbuild /opt/ort /opt/ort
+COPY docker/bash_bootstrap.sh /etc/ort/bash_bootstrap.sh
+COPY docker/ort-wrapper.sh /usr/bin/ort
```
maxhbr, line 238: Maybe also add a wrapper / symlink for `orth`.

maxhbr (commented):
just some comments

docker/bash_bootstrap.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
heliocastro: Can't be done as docker run not executes the bash profiling. 
Would need to sourcing anyway

docker/python.sh
```diff
@@ -0,0 +1,36 @@
...
+. "$(pyenv root)"/completions/pyenv.bash;
+
+
+# Needed if you want new pyhon version then default
+install_python_builddeps () {
+    apt update
+    apt install -y make build-essential libssl-dev zlib1g-dev \
```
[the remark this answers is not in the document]
heliocastro: I will remove this one. The idea was well intended, but i don't expect anyone will open the docker in interactive mode and install more than one versions.

Dockerfile
heliocastro: On it

Dockerfile
```diff
@@ -19,166 +20,291 @@
...
+RUN chmod 755 /usr/bin/ort
+
+# Python
+COPY --from=pythonbuild /opt/python /opt/python
```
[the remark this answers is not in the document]
heliocastro: Good question. 
pyenv take care of static python building, but i'm not sure about missing shared libraries.
Checking...

docker/python.sh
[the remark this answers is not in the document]
heliocastro: Done, removed

Dockerfile
heliocastro: Done

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
+ENV NVM_DIR=/opt/nodejs
+
+RUN git clone --depth 1 https://github.com/nvm-sh/nvm.git /opt/nodejs
+RUN . $NVM_DIR/nvm.sh && nvm install "${NODEJS_VERSION}"
+RUN . $NVM_DIR/nvm.sh \
+    && npm install --global npm@$NPM_VERSION bower@$BOWER_VERSION yarn@$YARN_VERSION
+COPY docker/nodejs.sh /etc/ort/bash_modules
+
+#------------------------------------------------------------------------
+# GOLANG - BUild as a separate component
```
tsteenbe, line 175: Change BUild to Build

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
+        unzip
+
+# External repositories for SBT
+ARG SBT_VERSION=1.3.8
+ENV KEYURL=https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823
+RUN echo "deb https://repo.scala-sbt.org/scalasbt/debian all main" | tee /etc/apt/sources.list.d/sbt.list
+RUN echo "deb https://repo.scala-sbt.org/scalasbt/debian /" | tee /etc/apt/sources.list.d/sbt_old.list
+RUN curl -ksS "$KEYURL" | gpg --dearmor | tee "/etc/apt/trusted.gpg.d/scala_ubuntu.gpg" > /dev/null
+
+# External repository for latest git
```
tsteenbe, line 289: Change git to Git

Dockerfile
```diff
@@ -19,166 +20,293 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
-# Set this to the version ORT should report.
-ARG ORT_VERSION="DOCKER-SNAPSHOT"
+#------------------------------------------------------------------------
+# OpenJDK Eclipe Temurin Ubuntu LTS based
```
tsteenbe, line 24: Suggestion: OpenJDK Eclipe Temurin Ubuntu LTS based -> Use OpenJDK Eclipe Temurin Ubuntu LTS

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
 
+# Set this to the version ORT should report.
+ARG ORT_VERSION="DOCKER-SNAPSHOT"
+
+COPY . /usr/local/src/ort
 WORKDIR /usr/local/src/ort
 
-# Gradle build.
-ARG ORT_VERSION
+# Prepare gradle
```
tsteenbe, line 88: gradle change to Gradle

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
+
+ARG CONAN_VERSION=1.40.3
+ARG PYTHON_VERSION=3.6.15
+ARG PYTHON_VIRTUALENV_VERSION=15.1.0
+ARG PIPTOOL_VERSION=21.3.1
+
+ENV PYENV_ROOT=/opt/python
+RUN curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
+ENV PATH=/opt/python/bin:$PATH
+# Python 3.6.x series has problems with alignment with modern gcc
```
tsteenbe, line 114: gcc -> GCC

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
+# As we don not want patch python versions we use clang as compiler
```
tsteenbe, line 115: python -> Python

tsteenbe, line 115: clang -> Clang

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
+    pip=="${PIPTOOL_VERSION}" \
+    pipenv \
+    Mercurial \
+    virtualenv=="${PYTHON_VIRTUALENV_VERSION}" \
+    wheel
+
+COPY docker/python.sh /etc/ort/bash_modules
+
+#------------------------------------------------------------------------
+# RUBY - Build ruby as a separate component with rbenv
```
tsteenbe, line 130: ruby -> Ruby

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
+RUN git clone --depth 1 https://github.com/rbenv/rbenv.git ${RBENV_ROOT}
+RUN git clone --depth 1 https://github.com/rbenv/ruby-build.git "$(rbenv root)"/plugins/ruby-build
+WORKDIR ${RBENV_ROOT}
+RUN src/configure \
+    && make -C src
+RUN rbenv install ${RUBY_VERSION} \
+    && rbenv global ${RUBY_VERSION} \
+    && gem install bundler cocoapods
+
+# Install 'CocoaPods'. As https://github.com/CocoaPods/CocoaPods/pull/10609 is needed but not yet released.
```
tsteenbe, line 146: https://github.com/CocoaPods/CocoaPods/pull/10609 was merged and I believe also released?

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
+    && (cd CocoaPods-9461b346aeb8cba6df71fd4e71661688138ec21b \
+        && gem build cocoapods.gemspec \
+        && gem install cocoapods-1.10.1.gem \
+        ) \
+    && rm -rf CocoaPods-9461b346aeb8cba6df71fd4e71661688138ec21b
+
+COPY docker/ruby.sh /etc/ort/bash_modules
+
+#------------------------------------------------------------------------
+# NODEJS - Build nodejs as a separate component with nvm
```
tsteenbe, line 158: nodejs -> NodeJS,

Dockerfile
```diff
@@ -19,166 +20,293 @@
...
+
+ARG SCANCODE_VERSION="3.2.1rc2"
+ENV SCANCODE_URL "https://github.com/nexB/scancode-toolkit/archive/refs/tags/"
+ENV PATH=/opt/python/shims:$PATH
+
+RUN mkdir -p /opt/scancode \
+    && curl -ksSL ${SCANCODE_URL}/v${SCANCODE_VERSION}.tar.gz | tar -C /opt/scancode -xz --strip-components=1 \
+    && cd /opt/scancode \
+    && PYTHON_EXE=python3 ./configure \
+    # cleanup unneeded installed binaries
```
tsteenbe, line 228: cleanup -> Clean up

tsteenbe (changes requested):

Dockerfile
heliocastro: Maybe, but i do not want change this since is this way on upstream Dockerfile. If we merge, then in later step i change to newer version of CocoaPods.

Dockerfile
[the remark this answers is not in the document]
heliocastro: Ok, now i got a confident answer.

With latest commit addition ( netbase ), here's the results after running Python tests ( Read to the end ):

```
== Tests result: FAILURE ==

379 tests OK.

1 test failed:
test_ssl

25 tests skipped:
test_curses test_dbm_ndbm test_devpoll test_gdb test_idle
test_kqueue test_msilib test_ossaudiodev test_smtpnet
test_socketserver test_startfile test_tcl test_timeout test_tix
test_tk test_ttk_guionly test_ttk_textonly test_turtle
test_urllib2net test_urllibnet test_winconsoleio test_winreg
test_winsound test_xmlrpc_net test_zipfile64

2 tests run no tests:
test_dtrace test_future4

```
The test_ssl fails is a upstream bug on the test itself:
https://bugs.python.org/issue40018

Since Ubuntu Focal is using openssl 1.1.1e, this test failed correctly following bug description.
So i can say we're good

docker/bash_bootstrap.sh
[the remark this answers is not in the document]
maxhbr: But using `/etc/profile.d/` would be easier and more consistent than implementing your own loading. You could place all scripts into `/etc/profile.d/` and replace every `source /etc/ort/bash_bootstrap.sh` with `source /etc/profile`  or replace the first line of the `docker/ort-wrapper.sh` shell script with `#!/bin/bash -l` to explicitly load the env.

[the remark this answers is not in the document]
maxhbr: See : https://github.com/heliocastro/ort/pull/18

maxhbr:
Is there a way to unresolve a conversation, I think https://github.com/oss-review-toolkit/ort/pull/4746#discussion_r757104461 (and https://github.com/heliocastro/ort/pull/18) is worth to think about

Dockerfile
```diff
@@ -19,166 +20,294 @@
...
+FROM build AS gobuild
+
+ARG GO_DEP_VERSION=0.5.4
+ARG GO_VERSION=1.17.3
+ENV GOPATH=/opt/go
+RUN curl -L https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /opt -xz
+ENV PATH=/opt/go/bin:$PATH
+RUN go version
+RUN curl -ksS https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | bash
+RUN echo "add_local_path /opt/go/bin:$PATH" > /etc/ort/bash_modules/go.sh
```
maxhbr, line 185: when is the `$PATH` variable resolved? On docker build time? Might need escaping:

```suggestion
RUN echo "add_local_path /opt/go/bin:\$PATH" > /etc/ort/bash_modules/go.sh
```

heliocastro: Really liked this approach. Many thanks @maxhbr 
I merged your PR

Dockerfile
```diff
@@ -53,133 +91,220 @@ RUN --mount=type=cache,target=/tmp/.gradle/ \
...
+    && PYTHON_EXE=python3 ./configure \
+    # Clean up unneeded installed binaries
+    && rm -rf /opt/scancode/thirdparty
+
+#------------------------------------------------------------------------
+# Main container
+FROM eclipse-temurin:11-jre
+
+# ORT
+COPY --from=ortbuild /opt/ort /opt/ort
```
maxhbr, line 234: This is the thing, that gets rebuilt nearly every time? Why not do that last so that caching of layers can be leveraged more?

Dockerfile
```diff
@@ -53,133 +91,220 @@ RUN --mount=type=cache,target=/tmp/.gradle/ \
...
+
+# External repository for latest Git
+ENV KEYURL=https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xa1715d88e1df1f24
+RUN echo "deb http://ppa.launchpad.net/git-core/ppa/ubuntu bionic main" > /etc/apt/sources.list.d/git-core-ubuntu-ppa-bionic.list
+RUN curl -ksS "$KEYURL" | gpg --dearmor | tee "/etc/apt/trusted.gpg.d/git-core_ubuntu_ppa.gpg"  > /dev/null
+
+ENV COMPOSER_VERSION=1.10.1-1 \
+    CONAN_VERSION=1.40.3
+
+ENV PATH="$PATH:$HOME/.local/bin"
```
maxhbr, line 295: If someone runs the container as other user, the directory `$HOME/.local/bin` does no longer point to the right folder or points to a folder that is not readable.

Dockerfile
```diff
@@ -53,133 +91,220 @@ RUN --mount=type=cache,target=/tmp/.gradle/ \
...
+RUN echo "deb https://repo.scala-sbt.org/scalasbt/debian /" | tee /etc/apt/sources.list.d/sbt_old.list
+RUN curl -ksS "$KEYURL" | gpg --dearmor | tee "/etc/apt/trusted.gpg.d/scala_ubuntu.gpg" > /dev/null
+
+# External repository for latest Git
+ENV KEYURL=https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xa1715d88e1df1f24
+RUN echo "deb http://ppa.launchpad.net/git-core/ppa/ubuntu bionic main" > /etc/apt/sources.list.d/git-core-ubuntu-ppa-bionic.list
+RUN curl -ksS "$KEYURL" | gpg --dearmor | tee "/etc/apt/trusted.gpg.d/git-core_ubuntu_ppa.gpg"  > /dev/null
+
+ENV COMPOSER_VERSION=1.10.1-1 \
+    CONAN_VERSION=1.40.3
```
maxhbr, line 293: CONAN_VERSION is never used after that point here?

Dockerfile
```diff
@@ -53,133 +91,220 @@ RUN --mount=type=cache,target=/tmp/.gradle/ \
...
+        gnupg \
+        libarchive-tools \
         netbase \
-        openssh-client \
-        unzip \
-        xz-utils \
-        zlib1g-dev \
-        # Install VCS tools (no specific versions required here).
+        openssl \
+        unzip
```
maxhbr, line 278: no `&& rm -rf /var/lib/apt/lists/* ` here?

Dockerfile
```diff
@@ -53,133 +91,220 @@ RUN --mount=type=cache,target=/tmp/.gradle/ \
...
+RUN echo "deb https://repo.scala-sbt.org/scalasbt/debian all main" | tee /etc/apt/sources.list.d/sbt.list
+RUN echo "deb https://repo.scala-sbt.org/scalasbt/debian /" | tee /etc/apt/sources.list.d/sbt_old.list
+RUN curl -ksS "$KEYURL" | gpg --dearmor | tee "/etc/apt/trusted.gpg.d/scala_ubuntu.gpg" > /dev/null
+
+# External repository for latest Git
+ENV KEYURL=https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xa1715d88e1df1f24
+RUN echo "deb http://ppa.launchpad.net/git-core/ppa/ubuntu bionic main" > /etc/apt/sources.list.d/git-core-ubuntu-ppa-bionic.list
+RUN curl -ksS "$KEYURL" | gpg --dearmor | tee "/etc/apt/trusted.gpg.d/git-core_ubuntu_ppa.gpg"  > /dev/null
+
+ENV COMPOSER_VERSION=1.10.1-1 \
```
maxhbr, line 292: Why Isn't `COMPOSER_VERSION` an ARG?

docker/bash_bootstrap.sh
```diff
@@ -1,40 +0,0 @@
...
-# SPDX-License-Identifier: Apache-2.0
-# License-Filename: LICENSE
-
-# We will parse the necessary shells modules for each language
-# independently
-
-set -e
-
-# Global functions
-add_local_path () {
```
maxhbr: my changes dropped `add_local_path` even though that is still used in the other .sh files

maxhbr: I am working on a fix

maxhbr: fix: https://github.com/heliocastro/ort/pull/19

Dockerfile
maxhbr: HOME seems to be unset on build time:
```
$ docker run -ti --entrypoint=/bin/bash ort
root@1cc91553aeff:/# echo $PATH
/opt/java/openjdk/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/.local/bin
```

docker/bash_bootstrap.sh
heliocastro: I merged your PR, thanks

Dockerfile
heliocastro: I was wondering the same. Looks like because we are applying a sed on runtime over gradle that invalidate the cache, and worst, everytme increase the cache build.
When developing this one sometime i need clean up >100Gb of caches on docker :-/

Dockerfile
heliocastro: Nope, we will need install another set of packages some lines below,m then this avoid a second update.
Second line removes the lists then.

Dockerfile
heliocastro: This is a mistake

Dockerfile
heliocastro: Actually, no .local/bin is needed anymore, we don't have any tool there anymore. 
 I just forgot to remove it

Dockerfile
maxhbr: But that increases the size of the layer, and I thought it was best practice to always clean up.

maxhbr: AFAIK every `RUN` / `ARG` / `ENV` results in its own layer.

Dockerfile
heliocastro: Solved

Dockerfile
heliocastro: Solved

Dockerfile
heliocastro: Solved

Dockerfile
heliocastro: Solved

docker/legacy/Dockerfile
```diff
@@ -0,0 +1,187 @@
...
+ARG ORT_VERSION
+RUN --mount=type=cache,target=/tmp/.gradle/ \
+    GRADLE_USER_HOME=/tmp/.gradle/ && \
+    scripts/import_proxy_certs.sh && \
+    scripts/set_gradle_proxy.sh && \
+    sed -i -r 's,(^distributionUrl=)(.+)-all\.zip$,\1\2-bin.zip,' gradle/wrapper/gradle-wrapper.properties && \
+    sed -i -r '/distributionSha256Sum=[0-9a-f]{64}/d' gradle/wrapper/gradle-wrapper.properties && \
+    ./gradlew --no-daemon --stacktrace -Pversion=$ORT_VERSION :cli:distTar :helper-cli:startScripts
+
+FROM adoptopenjdk:11-jre-hotspot-bionic
```
maxhbr, line 56: I am surprised that this PR keeps the ubuntu version at bionic. That is sadly to old for my use cases. I will have to continue to maintain a fork.

maxhbr: This base image also was last pushed more than a year ago. Might not be the best basis from a security point of view: https://hub.docker.com/layers/adoptopenjdk/library/adoptopenjdk/11-jre-hotspot-bionic/images/sha256-d27491ca50e6d00ca5f7331d5f2bdd136a2f8a68d57c87b9337f45a64d7f5362?context=explore

maxhbr: see https://github.com/oss-review-toolkit/ort/pull/4899

heliocastro: @maxhbr This is the current upstream image i moved to legacy. Should i move it to new openjdk as well ?

maxhbr: You should follow that, as soon as it is merged on master. But this also conflicts with the old Scancode version requiring python 3.6

maxhbr: It is now merged, you could now follow that update

sschuberth: I'd recommend to wait a bit, I'm planning for a few other updates to the current Docker image that we need ASAP.

sschuberth: > I'd recommend to wait a bit, I'm planning for a few other updates to the current Docker image that we need ASAP.

See https://github.com/oss-review-toolkit/ort/pull/4934.

heliocastro:
Up to date with upstream Docker changes, including new Java 17

sschuberth:
> Up to date with upstream Docker changes, including new Java 17

I just had to revert that 😢 See https://github.com/oss-review-toolkit/ort/pull/4948.

sschuberth:
Taking some notes from today's meeting with @heliocastro and @tsteenbe:

> * Uses multi-stage build to reduce final container

Now different package managers are installed in different layers of "helper images", for which the final images just copies required files. This optimizes for incremental build time, but it also reduces the image size, as only required binaries are included, and not the files to build the binaries.

> * Harmonize the base image to use same OpenJDK base from Eclipse Temurin and Ubuntu Focal LTS

This is good, but just as a remark, meanwhile also available in the "legacy" `Dockerfile`.

> * Make all tooling versions customizable through docker build args

Note to myself: Tool versions were customizable before, but only via environment variables, but via build arguments.

> * Each tooling is built in separate components, improving the  process to add or remove specific tools

Also see explanations for the first bullet item.

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
+    xz-utils \
+    zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+#------------------------------------------------------------------------
+# PYTHON - Build Python as a separate component with pyenv
+FROM build as pythonbuild
+
+ARG CONAN_VERSION=1.44.0
+ARG PYTHON_VERSION=3.9.9
```
sschuberth, line 81: Would it be possible to get a configurable Python 2 version *and* a configurable Python 3 version in the same image? If so, could you make that change?

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
+ARG PYTHON_VIRTUALENV_VERSION=20.13.0
+ARG PIPTOOL_VERSION=21.3.1
+
+ENV PYENV_ROOT=/opt/python
+RUN curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
+ENV PATH=/opt/python/bin:$PATH
+# Python 3.6.x series has problems with alignment with modern GCC
+# As we don not want patch Python versions we use Clang as compiler
+RUN CC=clang-9 CXX=clang++9 pyenv install "${PYTHON_VERSION}"
```
sschuberth, line 90: Note to myself: This is required as `pyenv` builds Python versions from source.

sschuberth: This is the line where to append an eventual `"${PYTHON2_VERSION}"`.

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+#------------------------------------------------------------------------
+# PYTHON - Build Python as a separate component with pyenv
+FROM build as pythonbuild
+
+ARG CONAN_VERSION=1.44.0
+ARG PYTHON_VERSION=3.9.9
+ARG PYTHON_VIRTUALENV_VERSION=20.13.0
+ARG PIPTOOL_VERSION=21.3.1
```
sschuberth, line 83: As a general idea for improvement: It would be nice of the installation of a tool could be skipped completely by setting the corresponding build argument to an empty string or so.

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
+ENV PATH=/opt/python/bin:$PATH
+# Python 3.6.x series has problems with alignment with modern GCC
+# As we don not want patch Python versions we use Clang as compiler
+RUN CC=clang-9 CXX=clang++9 pyenv install "${PYTHON_VERSION}"
+RUN pyenv global "${PYTHON_VERSION}"
+ENV PATH=/opt/python/shims:$PATH
+RUN pip install -U \
+    pip=="${PIPTOOL_VERSION}" \
+    wheel
+RUN pip install -U \
```
sschuberth, line 96: Please add a code comment why these need to be installed separately from the above.

Dockerfile
sschuberth: Please see again how this is now done in the "legacy" `Dockerfile`.

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
-# Set this to the ScanCode version to use.
 ARG SCANCODE_VERSION="30.1.0"
+ENV SCANCODE_URL "https://github.com/nexB/scancode-toolkit/archive/refs/tags/"
+ENV PATH=/opt/python/shims:$PATH
 
-FROM adoptopenjdk:11-jdk-hotspot-focal AS build
+RUN mkdir -p /opt/scancode \
+    && curl -ksSL ${SCANCODE_URL}/v${SCANCODE_VERSION}.tar.gz | tar -C /opt/scancode -xz --strip-components=1 \
+    && cd /opt/scancode \
+    && PYTHON_EXE=python3 ./configure \
```
sschuberth, line 203: I guess we should now also install this via `pip` like done in the `main` branch.

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
+    # Clean up unneeded installed binaries
+    && rm -rf /opt/scancode/thirdparty \
+    # Run scancode once to generate indexes as superuser
+    && /opt/scancode/bin/scancode --version
 
-COPY . /usr/local/src/ort
+#------------------------------------------------------------------------
+# ORT
+FROM nodebuild as ortbuild
```
sschuberth, line 211: FYI: This could probably be just `build` now as findings the bootstrapped `node` at Kotlin build time has been fixed as part of https://github.com/oss-review-toolkit/ort/pull/5040.

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
+FROM eclipse-temurin:11-jdk-focal
+
+ENV LANG=en_US.UTF-8
+ENV LANGUAGE=en_US:en
+ENV LC_ALL=en_US.UTF-8
+ENV HOME=/workspace
+
+# Run with non privileged user
+RUN groupadd --gid 1000 ort \
+    && useradd --uid 1000 --gid ort --shell /bin/bash --home-dir /workspace --create-home ort
```
sschuberth, line 248: Note to myself: Think about whether this eventually causes some problems when running the images with a custom user.

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
+ENV ANDROID_HOME=/opt/android-sdk
+ENV ANDROID_API_LEVEL=29
+COPY --from=androidbuild /usr/bin/repo /usr/bin/
+COPY --from=androidbuild /etc/profile.d/android.sh /etc/profile.d/
+COPY --chown=ort:ort --from=androidbuild /opt/android-sdk /opt/android-sdk
+
+# ScanCode
+COPY --chown=ort:ort --from=scancodebuild /opt/scancode /opt/scancode
+RUN ln -s /opt/scancode/bin/scancode /usr/bin/scancode \
+    && ln -s /opt/scancode/bin/pip /usr/bin/scancode-pip \
```
sschuberth, line 284: I don't think need need this in `PATH`.

Dockerfile
```diff
@@ -19,156 +20,330 @@
...
+    curl \
+    gnupg \
+    libarchive-tools \
+    locales \
+    netbase \
+    openssh-client \
+    openssl \
+    unzip \
+    && rm -rf /var/lib/apt/lists/* \
+    && sed -i '/en_US.UTF-8/s/^# //g' /etc/locale.gen \
```
sschuberth, line 302: Can this be simpler by just writing "en_US.UTF-8" to `/etc/locale.gen`? In any case please add a comment what this is needed for.

scripts/docker_build.sh
```diff
@@ -17,8 +18,17 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
+# Docker from ORT uses BuildKit feature
+DOCKER_BUILDKIT=1
+export DOCKER_BUILDKIT
```
sschuberth, line 23: FYI, this is now also in upstream; you'll see when you rebase.

scripts/docker_build.sh
```diff
@@ -17,8 +18,17 @@
...
+
 GIT_ROOT=$(git rev-parse --show-toplevel)
-GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty)
+GIT_VERSION=$(git describe --abbrev=7 --always --tags --dirty)
```
sschuberth, line 26: BTW, the 7 needs to stay a 10 (to match what JitPack uses).

sschuberth (changes requested):

Dockerfile
maxhbr: I would neither configure it yourself, nor use pip:

I think their preferred way of installing is "download and extract the release from github" https://github.com/nexB/scancode-toolkit/blob/develop/docs/source/getting-started/install.rst#installation-on-linux-and-mac

pip is just under the category "Installation as a library", so I would assume it is not the recommended way for usage.

scripts/docker_build.sh
heliocastro: Done on last commit

scripts/docker_build.sh
heliocastro: Done on last commit

Dockerfile
heliocastro: Removed. Now installed by pip on last commit

Dockerfile
heliocastro: Changed the way user is created. Now can be customized on build stage.

Dockerfile
heliocastro: For now we go with PIP. 
There's no guarantee that the binary will be 100% statically compatible ( yes, this kind of thing exists on Linux )
So, pip is good enough. After we merge the PR, we discuss with Phillipe properly

Dockerfile
heliocastro: Done in last commit

Dockerfile
heliocastro: Unfortunately IF/ELSE is something that docker can't handle.
The "solution" would be a external tool that dynamically create the Dockerfile, but again, will be too much.

Dockerfile
heliocastro: Now passing build-arg PYTHON2_VERSION=<somthing> will install python2 as well, otherwise only python3

Dockerfile
heliocastro: Yes, done on last commit

docker/bash_prompt
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
```
maxhbr, line 1: Under which license is that file?

I am thinking that this is going to far. convenience for interactive usage should be in another image that is added on top and not included by default.

heliocastro:
Removed.
You are correct

tsteenbe:
@fviernau @heliocastro @tsteenbe had a meeting on July 15th to discuss getting this PR merged.

@fviernau @tsteenbe agreed this can be merged once below action items 4,6 and 7 are done

**Action items:**

1. Fill issue to replace  virtualenv on Python analysis to use pyenv  (@heliocastro)
2. Fill issue to drop Ruby 2 in ORT (@heliocastro)
3. Fill issue to ORT docker image should also work on ARM (@tsteenbe)
4. Fix functional test failure  "Could not build image for container run.", see https://github.com/oss-review-toolkit/ort/runs/7354719297?check_suite_focus=true
How to debug this issue? (@heliocastro/ @sschuberth)
5. Nominate @heliocastro to become an ORT contributor
6. Update ort-gitlab-ci to enable running with legacy Docker image (@tsteenbe)
7. Test multi-stage Docker image with a variety of projects using different package managers (@fviernau / @tsteenbe)
8. Once 4 has been addressed, make announcement that PR will be merged by date X (1 week period) and that users dependend on legacy image need to take mitigation measures. (@tsteenbe)

heliocastro:
@tsteenbe [Ruby issue](https://github.com/oss-review-toolkit/ort/issues/5613)
@tsteenbe [Python vitualenv replacement proposal](https://github.com/oss-review-toolkit/ort/issues/5614)

nicorikken:
Android copyright information is present in the following files in the resulting container:
```
/opt/android-sdk/cmdline-tools/NOTICE.txt
/opt/android-sdk/platform-tools/NOTICE.txt
```

docker/android.sh
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+add_local_path ${ANDROID_HOME}/tools/bin
```
nicorikken, line 20:
```suggestion
add_local_path "${ANDROID_HOME}/tools/bin"
```

Dockerfile
```diff
@@ -19,151 +20,332 @@
...
 
+#------------------------------------------------------------------------
+# PYTHON - Build Python as a separate component with pyenv
+FROM build as pythonbuild
+
+ARG PYTHON_VERSION=3.10.2
+ARG PYTHON2_VERSION=""
+
+ENV PYENV_ROOT=/opt/python
+RUN curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
```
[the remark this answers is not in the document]
nicorikken: I haven't tested it yet, but this would be a more safe way to execute an online script:

- `-e` Exit immediately if a pipeline or compound command exists with a non-zero status.
- `-u` Treat unset variables and parameters other than the special parameters "@" and "*" as an error when performing parameter expansion.
- `-x` After expanding each simple command, for command, case command, select command, or arithmetic for command, display the expanded value of PS4, followed by the command and its expanded arguments or associated word list.
- `-o pipefail` If set, the return value of a pipeline is the value of the last (rightmost) command to exit with a non-zero status, or zero if all commands in the pipeline exit successfully.`

```suggestion
RUN bash -c "set -euxo pipefail && curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash"
```

Dockerfile
```diff
@@ -19,151 +20,332 @@
...
+ARG PYTHON_VIRTUALENV_VERSION=20.13.0
+ARG PIPTOOL_VERSION=21.3.1
+ARG SCANCODE_VERSION=30.1.0
+
+# Install pip and wheel ahead of regular modules as python try to generate
+# whl packages if they are not automatic available to download.
+RUN pip install -U \
+    pip=="${PIPTOOL_VERSION}" \
+    wheel
+RUN pip install -U \
```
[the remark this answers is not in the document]
nicorikken:
```suggestion
RUN pip install --no-cache-dir -U \
```

Dockerfile
```diff
@@ -19,151 +20,332 @@
...
+    && npm install --global npm@$NPM_VERSION bower@$BOWER_VERSION yarn@$YARN_VERSION
+
+#------------------------------------------------------------------------
+# GOLANG - Build as a separate component
+FROM build AS gobuild
+
+ARG GO_DEP_VERSION=0.5.4
+ARG GO_VERSION=1.17.3
+ENV GOPATH=/opt/go
+RUN curl -L https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /opt -xz
```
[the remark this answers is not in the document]
nicorikken:
```suggestion
RUN bash -c "set -euxo pipefail && curl -L https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /opt -xz"
```

Dockerfile
```diff
@@ -19,151 +20,332 @@
...
+ENV PATH=/opt/go/bin:$PATH
+RUN go version
+RUN curl -ksS https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | bash
```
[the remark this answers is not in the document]
nicorikken:
```suggestion
RUN bash -c "set -euxo pipefail && curl -ksS https://raw.githubusercontent.com/golang/dep/v${GO_DEP_VERSION}/install.sh | bash"
```

Dockerfile
```diff
@@ -19,151 +20,332 @@
...
+# REPO / ANDROID SDK
+FROM build AS androidbuild
+
+ARG ANDROID_CMD_VERSION=7583922
+ENV ANDROID_HOME=/opt/android-sdk
+
+RUN curl -ksS https://storage.googleapis.com/git-repo-downloads/repo > /usr/bin/repo \
+    && chmod a+x /usr/bin/repo
+
+RUN curl -Os https://dl.google.com/android/repository/commandlinetools-linux-${ANDROID_CMD_VERSION}_latest.zip \
```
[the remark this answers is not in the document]
nicorikken: Looked at sonatype comment but is a wont-fix.

Dockerfile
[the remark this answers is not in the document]
maxhbr: The following seems to be a lot easier to read and also works:
```suggestion
RUN set -o pipefail && curl -ksS https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | bash
```

Dockerfile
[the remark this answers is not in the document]
maxhbr: - `-e` -> there is just one command, makes no difference
- `u` -> there are no variables, makes no difference
- `x` docker also shows you what is executed, for a single line command that might just add noise

it is also not necessary to wrap that in bash, the following also works:

```suggestion
RUN set -o pipefail && curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
```

## your options are not passed on to the `pyenv-installer` script
Also, the bash options are not applied to the execution of `pyenv-installer`, as you can see in the following example:
```
> cat test.sh
───────┬
       │ File: test.sh
───────┼
   1   │ #!/usr/bin/env bash
   2   │
   3   │ set -e
   4   │ cat <<EOF | bash
   5   │ false
   6   │
   7   │ echo "command after false"
   8   │ EOF
───────┴
> ./test.sh
command after false
```

If you want to add these you have to do the following:

```suggestion
RUN curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash -euxo pipefail
```

### But you should not apply these options to scripts that are not written with these options in mind
This changes the behaviour of the script and if it was not developed with these options in mind it will behave differently. So you should not apply these options to a random script outside of your control

And that script already has `-e` set, and would set `-x` if the right env variable is provided:
https://github.com/pyenv/pyenv-installer/blob/49fba599e872bc761858ea6f700271fb6dcb5a97/bin/pyenv-installer#L3-L4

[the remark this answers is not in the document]
nicorikken: So, leave it all out then?

Dockerfile
```diff
@@ -19,159 +20,329 @@
...
+    && npm install --global npm@$NPM_VERSION bower@$BOWER_VERSION yarn@$YARN_VERSION
+
+#------------------------------------------------------------------------
+# GOLANG - Build as a separate component
+FROM build AS gobuild
+
+ARG GO_DEP_VERSION=0.5.4
+ARG GO_VERSION=1.18.3
+ENV GOPATH=/opt/go
+RUN curl -L https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /opt -xz
```
nicorikken, line 158:
```suggestion
RUN set -o pipefail && curl -L "https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz" | tar -C /opt -xz
```

docker/android.sh
```diff
@@ -0,0 +1,22 @@
...
+add_local_path ${ANDROID_HOME}/emulator
```
nicorikken, line 21:
```suggestion
add_local_path "${ANDROID_HOME}/emulator"
```

Dockerfile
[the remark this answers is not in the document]
maxhbr: I would suggest to do just the following change:
```suggestion
RUN set -o pipefail && curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
```

tsteenbe (approved):

sschuberth:
Cross-posting from Slack: Please don't merge this with a whopping 43 (!) commits in the history. I guess most of this should be squashable, or?

Dockerfile
```diff
@@ -202,7 +202,9 @@ ENV LANG=en_US.UTF-8
 ENV LANGUAGE=en_US:en
 ENV LC_ALL=en_US.UTF-8
 
-RUN echo $LANG > /etc/locale.gen
+RUN echo $LANG > /etc/locale.gen \
```
sschuberth, line 228: How is this related to "Make batect understand entrypoint"?

sschuberth (changes requested):
Sorry, the Git history in this PR is a mess. I'm blocking the merge until the TSC has decided on how to proceed.

sschuberth:
@mnonnenmacher @MarcelBochtler @fviernau could you live with the Git history in this PR?

mnonnenmacher:
> @mnonnenmacher @MarcelBochtler @fviernau could you live with the Git history in this PR?

@sschuberth I would not merge it like this, because just from a short look I see so many issues:

* 43 commits is far too much for a single PR. This is impossible to review.
* Many commits change multiple things at once. Just as one example, the "Revert to Java 11" commit changes the pyenv installation and there is no explanation in the commit message if this is related to the base image change or just completely unrelated.
* There are a lot of typos and hard to understand sentences in the commit messages.
* Several commits contain lists of changes which shows they should have been separate commits.
* There is one commit that creates a legacy Dockerfile, but this is not the current version of the file but an old version. Instead there is a separate commit that updates the legacy file.
* There are version updates all over the place, it's unfeasible to check if all of these are updates related to some old version of the original Dockerfile or to the current Dockerfile. For example, the first commit downgrades ScanCode only to upgrade it again in a later commit.
* Commit message prefixes are inconsistent, some start with "Docker fix:" or "Docker update:" or "Dockerfile:" or "docker:".
* It's unclear why some entries were added to the .gitignore.
* The link to the BuildKit docs was removed from the README.

And this is not the complete list.

I understand that this PR was worked on for almost a year now, and that this was probably a bit frustrating. So I want to thank @heliocastro for being to persistent with your efforts to improve our Docker setup. But nevertheless this PR should follow our standards. We have put a lot of effort into maintaining a clean Git history for this project because it brings a lot of benefits, and I would not like to ignore this now just because someone has an urge to merge these changes.

My suggestion to move this forward in a clean way would be to squash all changes into just a few commits, so that the PR can be reviewed properly, for example:

* One commit to copy the current Dockerfile to the legacy directory.
* A few commits that apply the relevant changes on top of the current Dockerfile, but without all the version updates that just repeat version updates which already happened in our current Dockerfile. Or even squashing all changes into a single commit would be better than the current state of the PR.
* One commit for each version update that changes a version compared to our current Dockerfile.

And as a general comment, if you work on a contribution that takes a long time, to a project with mandatory code review, it is super important to rebase your changes on top of the current main branch on a regular basis. Yes, this can be a lot of work, but it is the responsibility of the author to make sure that their changes are in a reviewable state.

sschuberth:
> Does one of you maybe have a better idea how to clean this up?

Looks like a clean-up was already done in https://github.com/oss-review-toolkit/ort/pull/5760. I'll close this PR in favor of that one. Let's continue the discussion over there.

PR #4751, opened 2021-11-24
VersionControlSystem: Try to determine the VCS type statically first 

sschuberth:

### DIFF
--- a/downloader/src/main/kotlin/VcsHost.kt
+++ b/downloader/src/main/kotlin/VcsHost.kt
@@ -254,7 +254,8 @@ enum class VcsHost(
          * Return all [VcsInfo] that can be parsed from [projectUrl] without actually making a network request.
          */
         fun toVcsInfo(projectUrl: String): VcsInfo {
-            val vcsInfoFromHost = projectUrl.toUri { toVcsHost(it)?.toVcsInfoInternal(it) }.getOrNull()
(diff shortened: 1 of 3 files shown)


mnonnenmacher (dismissed):

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -304,6 +304,18 @@ enum class VcsHost(
...
+                    val query = "?version=GB"
+                    val url = projectUrl.substringBefore(query)
+                    val revision = projectUrl.substringAfter(query, "")
+
+                    VcsInfo(
+                        type = VcsType.GIT,
+                        url = url,
+                        revision = revision
+                    )
+                }
```
fviernau, line 317: Would be nice to have a test which illustrates how the TFS URLs actually look like.

fviernau (changes requested):

downloader/src/main/kotlin/VcsHost.kt
sschuberth: Done.

oheger-bosch: The tests should cover all the conditions under which a TFS URL is detected, also the part with ".visualstudio.com". 
Could it be that the URL contains multiple query parameters? Should the parsing of the version be made more robust, e.g. by splitting the query string at `&` characters and inspecting the key-value pairs?

sschuberth: Done.

PR #4761, opened 2021-11-25
Bootstrapping improvements

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
+++ b/scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
@@ -21,18 +21,9 @@
 package org.ossreviewtoolkit.scanner.scanners.scancode
 
 import java.io.File
-import java.io.IOException
-import java.net.HttpURLConnection
 import java.time.Instant
 
-import kotlin.io.path.createTempDirectory
-import kotlin.io.path.createTempFile
 import kotlin.math.max
-
-import okhttp3.Request
-
-import okio.buffer
-import okio.sink
 
 import org.ossreviewtoolkit.model.ScanSummary
 import org.ossreviewtoolkit.model.ScannerDetails
@@ -49,7 +40,6 @@ import org.ossreviewtoolkit.utils.common.Os
 import org.ossreviewtoolkit.utils.common.ProcessCapture
 import org.ossreviewtoolkit.utils.common.isTrue
 import org.ossreviewtoolkit.utils.common.unpack
-import org.ossreviewtoolkit.utils.core.ORT_NAME
 import org.ossreviewtoolkit.utils.core.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.core.createOrtTempDir
 import org.ossreviewtoolkit.utils.core.log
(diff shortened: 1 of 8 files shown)


utils/core/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -140,43 +141,47 @@ object OkHttpClientHelper {
...
+        val request = Request.Builder()
+            .apply { acceptEncoding?.also { header("Accept-Encoding", it) } }
+            .get()
+            .url(url)
+            .build()
+
+        return runCatching { execute(request) }.mapCatching { response ->
+            val body = response.body
+
+            if (!response.isSuccessful || body == null) {
```
mnonnenmacher, line 159: This seems to alter the behavior of the `downloadText()` function, previously it returned `Success` if the body was `null`.

utils/core/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -140,43 +141,47 @@ object OkHttpClientHelper {
...
-
-        val body = response.body
-
-        if (!response.isSuccessful || body == null) {
-            return Result.failure(HttpDownloadError(response.code, response.message))
-        }
+        // Disable transparent gzip compression, as otherwise we might end up writing a tar file to disk while
+        // expecting to find a tar.gz file, and fail to unpack the archive. See
+        // https://github.com/square/okhttp/blob/parent-3.10.0/okhttp/src/main/java/okhttp3/internal/http/BridgeInterceptor.java#L79
+        val (response, body) = download(url, "identity").getOrElse { return Result.failure(it) }
```
mnonnenmacher, line 184: Using a named argument for `"identity"` would help with readability.

scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
```diff
@@ -162,37 +152,17 @@ class ScanCode(
         val url = "https://github.com/nexB/scancode-toolkit/archive/$archive"
 
         log.info { "Downloading $scannerName from $url... " }
+        val unpackDir = createOrtTempDir(expectedVersion).apply { deleteOnExit() }
```
mnonnenmacher: Any reason not to squash this with the previous commit?

mnonnenmacher (changes requested):

utils/core/src/main/kotlin/OkHttpClientHelper.kt
sschuberth: Well spotted :-) This actually is an alignment with the behavior of `downloadFile()`, and in `downloadText()` it was "wrong" from start, as I lazily just used `.orEmpty()`. I'll document this in the commit message.

scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
sschuberth: I just had a mental split between "bootstrapping scanners" and "bootstrapping anything else" in mind. I also would have trouble finding a common "prefix" for the commit message 😉 But I can squash if you prefer.

utils/core/src/main/kotlin/OkHttpClientHelper.kt
mnonnenmacher: Yes, that's why I asked, because you did the other scanners in the previous commit. But I don't care too much so I'll approve.

scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
mnonnenmacher: Yes, that's why I asked, because you did the other scanners in the previous commit. But I don't care too much so I'll approve.

utils/core/src/main/kotlin/OkHttpClientHelper.kt
mnonnenmacher: Ok, I just shortly checked the callers of `downloadText()` and it seems none of them could work with an empty string anyway.

mnonnenmacher (approved):

PR #4765, opened 2021-11-26
ListPackagesCommand: Allow limiting the output to projects or packages

fviernau:


helper-cli/src/main/kotlin/commands/ListPackagesCommand.kt
```diff
@@ -27,14 +27,16 @@ import com.github.ajalt.clikt.parameters.options.required
...
+import org.ossreviewtoolkit.helper.commands.Type.PROJECT
 import org.ossreviewtoolkit.helper.common.readOrtResult
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.licenses.LicenseView
 import org.ossreviewtoolkit.model.utils.createLicenseInfoResolver
 import org.ossreviewtoolkit.utils.common.expandTilde
 
 class ListPackagesCommand : CliktCommand(
-    help = "Lists the packages and projects contained in the given ORT result file."
+    help = "Lists the packages and optinally also projects contained in the given ORT result file."
```
sschuberth, line 39: "optionally"

helper-cli/src/main/kotlin/commands/ListPackagesCommand.kt
```diff
@@ -74,3 +81,5 @@ class ListPackagesCommand : CliktCommand(
         println(result)
     }
 }
+
+private enum class Type { PROJECT, PACKAGE }
```
sschuberth, line 85: How about moving

https://github.com/oss-review-toolkit/ort/blob/c60ee6232727025c8fb0db2aa1af2465923228a8/cli/src/main/kotlin/OrtMain.kt#L60-L73

to a common place for you to reuse it?

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/ListPackagesCommand.kt
fviernau: Hm, I'm struggling to find a good place for this. To me this class seems quite CLI specific, so I hesitate to move it out of the `cli` module. Do you have any place in mind where it can be put? (and how it could be named)

sschuberth: I'd simply put it in its own file under the `model` module, like the `Package` and `Project` it refers to.

helper-cli/src/main/kotlin/commands/ListPackagesCommand.kt
```diff
@@ -49,6 +52,11 @@ class ListPackagesCommand : CliktCommand(
         help = "Omit all packages not matching all licenses given by this comma separated list of license identifiers."
     ).split(",").default(emptyList())
 
+    private val type by option(
+        "--type",
+        help = "Filter the output by type which can be either 'projects' or 'packages'."
+    ).convert { DataEntity.valueOf(it.uppercase()) }
```
sschuberth, line 58: I'd recommend to reuse this code from `DownloadCommand`:

    .enum<DataEntity>().split(",").default(enumValues<DataEntity>().asList())

By setting a default value you do not have to handle the `null` case below in line 70 below.

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/ListPackagesCommand.kt
```diff
@@ -59,12 +68,11 @@ class ListPackagesCommand : CliktCommand(
                 .filter(LicenseView.ONLY_DETECTED)
                 .map { it.license.toString() }
 
-        val packages = ortResult
-            .collectProjectsAndPackages()
-            .filter {
+        val packages = ortResult.collectProjectsAndPackages().filter { id ->
+                (ortResult.isPackage(id) && PACKAGES in type) || (ortResult.isProject(id) && PROJECTS in type)
```
sschuberth, line 72: One more nit: For performance reasons, you should do the cheap `in type` checks first before the more expensive `isPackage()` / `isProject()` checks.

fviernau: Hm, I believe performance is really not relevant here.

sschuberth: I believe it'd be good practice anyway; if you don't always do it, it's easy to forget in cases where it really matters. But your call, I'll approve anyway.

sschuberth (approved):

helper-cli/src/main/kotlin/commands/ListPackagesCommand.kt
fviernau: I believe it's also valuable to code review practices to not discuss irrelevant things to keep things focused. IMO this should be weighed against each other. Anyhow, readability-wise the order IMO does not matter, so I'll just merge it as is.

PR #4767, opened 2021-11-28
Clean up Wiremock tests

sschuberth:

### DIFF
--- a/advisor/src/test/kotlin/advisors/OssIndexTest.kt
+++ b/advisor/src/test/kotlin/advisors/OssIndexTest.kt
@@ -21,11 +21,9 @@ package org.ossreviewtoolkit.advisor.advisors
 
 import com.github.tomakehurst.wiremock.WireMockServer
 import com.github.tomakehurst.wiremock.client.WireMock.aResponse
-import com.github.tomakehurst.wiremock.client.WireMock.configureFor
 import com.github.tomakehurst.wiremock.client.WireMock.equalTo
 import com.github.tomakehurst.wiremock.client.WireMock.equalToJson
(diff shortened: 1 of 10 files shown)


advisor/src/test/kotlin/advisors/OssIndexTest.kt
```diff
@@ -69,6 +68,7 @@ class OssIndexTest : WordSpec({
 
     beforeTest {
         server.resetAll()
+        configureFor(server.port())
```
mnonnenmacher, line 71: Doesn't this just reduce the risk of race conditions? It sounds like the correct solution would be not to use a global client instance but create local instances instead, I've never worked with WireMock though.

sschuberth: That's a good question that I cannot answer either, as I've also not worked much with WireMock yet. Maybe @oheger-bosch knows? Anyway, this changed solved [the concrete issue I was seeing with Kotest 5](https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1638023794193000).

sschuberth: Anyway, as this is at least better than what we had before, can we proceed with this for now as an immediate fix for the race conditions I currently see?

mnonnenmacher (dismissed):

advisor/src/test/kotlin/advisors/OssIndexTest.kt
sschuberth: Thanks to @oheger-bosch I've indeed found a better way that I decided to implement right now.

mnonnenmacher (approved):

Issue #4651, opened 2021-11-03
Analyzing a simple Flutter project fails


MarcelBochtler:
### To reproduce
Create a new project:
```bash
❯ flutter create flutter_example
Creating project flutter_example...
...
All done!
In order to run your application, type:

  $ cd flutter_example
  $ flutter run

Your application code is in flutter_example/lib/main.dart.
```

Setup Git repo:
```bash
git init && git add . && git commit -m "Initial commit"
```

Delete all gitignored files:
```bash
git clean -fdx
```

Run ORT analyzer:
```bash
ort --info --stacktrace analyze -i . -o ort/analyzer
```

This analyzer run fails with 2 Errors:
<details>
  <summary>ORT log</summary>

```sh
❯ ort --info --stacktrace analyze -i . -o ort/analyzer
10:04:01.204 [main] INFO  org.ossreviewtoolkit.model.config.OrtConfiguration - Using ORT configuration file '/Users/bom1wa3/.ort/config/ort.conf'.
________ _____________________
\_____  \\______   \__    ___/ the OSS Review Toolkit, version 6892f99.
 /   |   \|       _/ |    |
/    |    \    |   \ |    |    Running 'analyze' under Java 11.0.13 on Mac OS X with
\_______  /____|_  / |____|    12 CPUs and a maximum of 8192 MiB of memory.
        \/       \/
Environment variables:
ORT_CONFIG_DIR = /Users/bom1wa3/.ort/config
ORT_DATA_DIR = /Users/bom1wa3/.ort
SHELL = /opt/local/bin/zsh
TERM = xterm-256color
JAVA_HOME = /Library/Java/JavaVirtualMachines/openjdk11-zulu/Contents/Home
ANDROID_HOME = /Users/bom1wa3/.android
GOPATH = /Users/bom1wa3/work/go

The following configuration files and directories are used:
	/Users/bom1wa3/.ort/config/curations.yml
	/Users/bom1wa3/.ort/config/curations
The following package managers are activated:
	Bower, Bundler, Cargo, Carthage, CocoaPods, Composer, Conan, DotNet, GoDep, GoMod, Gradle, Maven, NPM, NuGet, PIP, Pipenv, Pub, SBT, SpdxDocumentFile, Stack, Yarn
Analyzing project path:
	/Users/bom1wa3/work/playground/flutter_example
[...]

PR #4770, opened 2021-11-29
Pub: Run flutter pub get to install dependencies in flutter projects

MarcelBochtler:
This partially solves https://github.com/oss-review-toolkit/ort/issues/4651.
However, it is important that `flutter pub get` is run before the `Gradle` project is analyzed this might still cause concurrency issues.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -575,8 +575,8 @@ class Pub(
     private fun commandPub(): String = "${command()} pub"
 
     private fun commandFlutter(): String =
-        if (flutterAbsolutePath.isDirectory) "$flutterAbsolutePath${File.separator}$flutterCommand packages"
-        else "$flutterCommand packages"
+        if (flutterAbsolutePath.isDirectory) "$flutterAbsolutePath${File.separator}$flutterCommand pub"
+        else "$flutterCommand pub"
 
     override fun run(workingDir: File?, vararg args: String): ProcessCapture {
         var result = ProcessCapture(workingDir, *commandPub().split(' ').toTypedArray(), *args)
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -594,9 +594,25 @@ class Pub(
     private fun installDependencies(workingDir: File) {
         requireLockfile(workingDir) { workingDir.resolve(PUB_LOCK_FILE).isFile }
 
-        // The "get" command creates a "pubspec.lock" file (if not yet present) except for projects without any
-        // dependencies, see https://dart.dev/tools/pub/cmd/pub-get.
-        run(workingDir, "get")
+        if (containsFlutterSdk(workingDir)) {
+            // For Flutter projects it is not enough to run `dart pub get`. Instead, use `flutter pub get` which
+            // installs the required dependencies and also creates the `local.properties` file which is required for
+            // the Android analysis.
+            ProcessCapture(workingDir, *commandFlutter().split(' ').toTypedArray(), "get").requireSuccess()
+        } else {
+            // The "get" command creates a "pubspec.lock" file (if not yet present) except for projects without any
+            // dependencies, see https://dart.dev/tools/pub/cmd/pub-get.
+            run(workingDir, "get")
+        }
+    }
+
+    /**
+     * Check the [PUBSPEC_YAML] within [workingDir] if the project contains the Flutter SDK.
+     */
+    private fun containsFlutterSdk(workingDir: File): Boolean {
+        val specFile = yamlMapper.readTree(workingDir.resolve(PUBSPEC_YAML))
+
+        return specFile?.get("dependencies")?.get("flutter")?.get("sdk")?.textValue() == "flutter"
     }
 }
 


analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -575,8 +575,8 @@ class Pub(
     private fun commandPub(): String = "${command()} pub"
 
     private fun commandFlutter(): String =
-        if (flutterAbsolutePath.isDirectory) "$flutterAbsolutePath${File.separator}$flutterCommand packages"
-        else "$flutterCommand packages"
+        if (flutterAbsolutePath.isDirectory) "$flutterAbsolutePath${File.separator}$flutterCommand pub"
```
sschuberth, line 578: Commit message: No dot at the end of the title.

More importantly, was "packages" renamed to "pub" already in Flutter version 2.2.3-stable, which is the version we bootstrap by default? If yes, why did the test work until now? If no, shouldn't we stick to "packages" for older version of Flutter, i.e. decide base on the Flutter version in use whether to use "packages" or "pub"?

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -594,9 +594,27 @@ class Pub(
     private fun installDependencies(workingDir: File) {
         requireLockfile(workingDir) { workingDir.resolve(PUB_LOCK_FILE).isFile }
 
-        // The "get" command creates a "pubspec.lock" file (if not yet present) except for projects without any
-        // dependencies, see https://dart.dev/tools/pub/cmd/pub-get.
-        run(workingDir, "get")
+        if (containsFlutter(workingDir)) {
```
sschuberth, line 597: Commit message:
- s/flutter/Flutter/ (at least if you mean the tool not the CLI command; if you do mean the CLI command, put it in backticks)
- No comma after "flutter"
- Add a comma after "SCM"

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -594,9 +594,27 @@ class Pub(
...
+            // For Flutter projects it is not enough to run `dart pub get`. Instead, use `flutter pub get` which
+            // installs the required dependencies and also creates a missing configuration file which is required for
+            // the Android analysis.
+            val result = ProcessCapture(workingDir, *commandFlutter().split(' ').toTypedArray(), "get")
+
+            if (result.isError) throw IOException(result.errorMessage)
```
sschuberth, line 603: Use `.requireSuccess()` instead.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -594,9 +594,27 @@ class Pub(
...
+            // The "get" command creates a "pubspec.lock" file (if not yet present) except for projects without any
+            // dependencies, see https://dart.dev/tools/pub/cmd/pub-get.
+            run(workingDir, "get")
+        }
+    }
+
+    /**
+     * Check the [PUBSPEC_YAML] if the project contains the Flutter SDK.
+     */
+    private fun containsFlutter(workingDir: File): Boolean {
```
sschuberth, line 614: Maybe "containsFlutterSdk"?

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -594,9 +594,27 @@ class Pub(
...
+        val specFile = yamlMapper.readTree(workingDir.resolve(PUBSPEC_YAML))
+
+        return specFile?.get("dependencies")?.get("flutter")?.get("sdk").textValueOrEmpty() == "flutter"
```
sschuberth, line 617: Could be just `textValue()`.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Pub.kt
MarcelBochtler: > was "packages" renamed to "pub" already in Flutter version 2.2.3-stable, which is the version we bootstrap by default?

Yes.

> If yes, why did the test work until now?

Because `flutterCommand()` / `flutter pub get` was never really used. Also the test called just `dart pub get` (via `run()`).
This does work in _some_ cases, but not in cases like I outlined in https://github.com/oss-review-toolkit/ort/issues/4651.

sschuberth: So what if users decide to bootstrap an older version of Flutter than the one we bootstrap by default? Should we support that case by still deciding based on the Flutter version which command to call?

sschuberth (dismissed):
Approving despite my comment.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -594,9 +594,25 @@ class Pub(
     private fun installDependencies(workingDir: File) {
         requireLockfile(workingDir) { workingDir.resolve(PUB_LOCK_FILE).isFile }
 
-        // The "get" command creates a "pubspec.lock" file (if not yet present) except for projects without any
-        // dependencies, see https://dart.dev/tools/pub/cmd/pub-get.
-        run(workingDir, "get")
+        if (containsFlutterSdk(workingDir)) {
+            // For Flutter projects it is not enough to run `dart pub get`. Instead, use `flutter pub get` which
+            // installs the required dependencies and also creates a missing configuration file which is required for
```
mnonnenmacher, line 599: Would be even better to be specific and say "creates the local.properties file".

analyzer/src/main/kotlin/managers/Pub.kt
MarcelBochtler: I downloaded Flutter 1.0.0 from 04.12.2018 to check. This version also includes `flutter pub` and the important `get` sub-command. So I think we should be completely backwards compatible.

sschuberth (approved):

PR #4771, opened 2021-11-29
Sanitize creating an `Identifier` from SPDX

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -263,6 +263,12 @@ private fun String.isExternalDocumentReferenceId(): Boolean = startsWith(SpdxCon
 private fun String.mapNotPresentToEmpty(): String = takeUnless { SpdxConstants.isNotPresent(it) }.orEmpty()
 
 /**
+ * Sanitize a string for use as an [Identifier] property where colons are not supported by replacing them with spaces,
+ * and finally collapsing multiple consecutive spaces.
+ */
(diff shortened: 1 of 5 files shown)


MarcelBochtler:
Adding `breaking change` label as this will break curations and package configuration that use the unsanitized ID

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -262,6 +262,13 @@ private fun String.isExternalDocumentReferenceId(): Boolean = startsWith(SpdxCon
  */
 private fun String.mapNotPresentToEmpty(): String = takeUnless { SpdxConstants.isNotPresent(it) }.orEmpty()
 
+/**
+ * Sanitize a string for use as an [Identifier] property where colons are not supported. While spaces theoretically are
+ * supported, replace them with dashes, too, for readability.
+ */
+private fun String.sanitize(): String =
+    replace(Regex("[: ]"), "-").replace(Regex("-{2,}"), "-")
```
MarcelBochtler, line 270: While I agree that replacing spaces with dashes is a good idea, it will break a lot of curations and I'm not sure if readability is worth it.

sschuberth: I don't believe there are many `SpdxDocumentFile`-type curations for packages with spaces in their names around, but I agree we should be conservative still, and I've changed the implementation.

sschuberth:
@oss-review-toolkit/core-devs can we move forward with this?

mnonnenmacher (approved):

PR #4772, opened 2021-11-29
MavenSupport: Handle Sprint Boot Starters as metadata-only packages

sschuberth:
Starters are a set of convenient dependency descriptors that you can
include in your application, see [1].

Starters usually have no sources published at all (see e.g. [2]), and if
sources are published, they only contain metadata file in the `META-INF`
directory that can also be retrieved from the POM.

[1]: https://github.com/spring-projects/spring-boot#spring-boot-starters
[2]: https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-starter-undertow/2.0.8.RELEASE/
[3]: https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-starter-security/2.6.0/

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -674,6 +674,10 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
             PackageManager.processProjectVcs(it, vcsFromPackage, *vcsFallbackUrls)
         } ?: PackageManager.processPackageVcs(vcsFromPackage, *vcsFallbackUrls)
 
+        val isSpringBootStarter = with(mavenProject) {
+            groupId == "org.springframework.boot" && artifactId.startsWith("spring-boot-starter-")
+        }
+
         return Package(
             id = Identifier(
                 type = "Maven",
@@ -690,7 +694,7 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
             sourceArtifact = sourceRemoteArtifact,
             vcs = vcsFromPackage,
             vcsProcessed = vcsProcessed,
-            isMetaDataOnly = mavenProject.packaging == "pom",
+            isMetaDataOnly = mavenProject.packaging == "pom" || isSpringBootStarter,
             isModified = isBinaryArtifactModified || isSourceArtifactModified
         )
     }


MarcelBochtler (commented):
LGTM, but I'll let someone else of the maintainers have a look if this is the correct approach to deal with this.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -674,6 +674,10 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
             PackageManager.processProjectVcs(it, vcsFromPackage, *vcsFallbackUrls)
         } ?: PackageManager.processPackageVcs(vcsFromPackage, *vcsFallbackUrls)
 
+        val isSpringBootStarter = with(mavenProject) {
+            groupId == "org.springframework.boot" && artifactId.startsWith("spring-boot-starter-")
```
oheger-bosch, line 678: I find it rather hacky to hard-code artifact coordinates directly in the `MavenSupport` class. Also, is this concept of starter poms unique to Spring or are there other frameworks using a similar approach? If so, it would probably make sense to have this somehow configurable outside of the code, so that it could be changed easily. Something like a configuration file listing patterns to identify well-known artifacts that are meta-data only.

sschuberth: > I find it rather hacky to hard-code artifact coordinates directly in the `MavenSupport` class.

I believe hard-coding is fine in this particular case. It's somewhat similar the hard-coding we already use in `MavenDependencyHandler`:

https://github.com/oss-review-toolkit/ort/blob/ea0851b6f47866453622e65581ca2a2480d7a92d/analyzer/src/main/kotlin/managers/MavenDependencyHandler.kt#L113

> Also, is this concept of starter poms unique to Spring or are there other frameworks using a similar approach?

I don't know. But if for other coordinates where we could tell for sure as well that these are metadata-only, we indeed should add similar code. I'd do that once we come across more cases, and then eventually extend the existing mechanism.

> If so, it would probably make sense to have this somehow configurable outside of the code, so that it could be changed easily. Something like a configuration file listing patterns to identify well-known artifacts that are meta-data only.

The whole point of hard-coding this is to not have any user-facing configuration. Users can already configure `is_metadata_only` via curations, so having another mechanism would be confusing. And in case you're talking about non-user-facing configuration: I don't see a value in this unless we have more cases to implement, like mentioned above.

mnonnenmacher: I would like to have a bit more detail in the commit message, for example that the JAR files of those packages contain only metadata with license and copyright information which can also be parsed from the POM files, and maybe also a code comment.

sschuberth: I've amended the commit message.

fviernau: I also find it a bit hacky to have package ID specific code in ORT. Have you considered extending the identifier matching of PackageCurations so that the same effect can be achieved with a single curation matching all the relevant Spring boot starter package IDs?

sschuberth: No, I have not considered that, as I believe ORT should be smart enough to set `isMetadataOnly` *without* any curation in clear cases for user convenience. Otherwise each separate user has to create the same curation internally again.

fviernau: I also find it a bit hacky to have package ID specific code in ORT. Have you considered extending the identifier matching of PackageCurations so that the same effect can be achieved with a single curation matching all the relevant Spring boot starter package IDs?

fviernau: > No, I have not considered that, as I believe ORT should be smart enough to set isMetadataOnly without any curation in clear cases for user convenience. Otherwise each separate user has to create the same curation internally again.

Along with a set of "default curations" packaged with ORT, we'd end-up with a single mechanism to achieve one thing instead of two mechanisms. However, the curations mechanism does not yet exist as the `id` matching is not yet powerful enough. 

I know it's getting a bit offtopic, maybe we should move the discussion elsewhere - I'm not blocking this change -
Would a default set of curations would actually make sense for you guys? That'd also allow turning the declared license mappings into curations, making another mechanism redundant. But I haven't tought through it yet.

oheger-bosch: > The whole point of hard-coding this is to not have any user-facing configuration.

Sorry, I did not mean a user-facing configuration. Rather a kind of resource file that lists patterns for artifact coordinates to be considered as meta-data only, with the aim to not have hard-coded IDs in the source code. Such a list would also be easier to extend if necessary.

sschuberth: > Would a default set of curations would actually make sense for you guys?

I though we agreed long ago that HERE would open-source (most of) its curations for Open Source packages... and I'm still waiting for that. IMO we should host these in a separate Git repo, but then package them with ORT, yes.

sschuberth: > with the aim to not have hard-coded IDs in the source code. Such a list would also be easier to extend if necessary.

I'm not really getting the difference between modifying a resource vs. modifying a single list in Kotlin code. Both are modifications in a single place that require recompilation. So how exactly does using a resource make things easier?

sschuberth: > However, the curations mechanism does not yet exist as the `id` matching is not yet powerful enough.

So, as that's the case, can we move forward with this as-is for the time being?

fviernau: > 

yes

mnonnenmacher (approved):

PR #4776, opened 2021-11-30
Miscellaneous code style and inspection fixes

fviernau:


reporter/src/test/kotlin/reporters/OpossumReporterTest.kt
```diff
@@ -176,8 +176,8 @@ class OpossumReporterTest : WordSpec({
         }
 
         "create signals with all uuids being assigned" {
-            opossumInput.pathToSignal.forEach { e ->
-                e.value.forEach { uuid ->
+            opossumInput.pathToSignal.forEach { (_, signal) ->
```
sschuberth: Would it make sense to only iterate over the `values` here, similar to like you did in a previous change for `keys`?

sschuberth: And actually you do a similar change in the test in the next commit.

sschuberth:
Nice improvements @fviernau! @maxhbr might be interested in these.

reporter/src/test/kotlin/reporters/OpossumReporterTest.kt
fviernau: Yeah, the change became necessary because `forAll` isn'T available for `Map`s.

maxhbr (dismissed):
LGTM

reporter/src/test/kotlin/reporters/OpossumReporterTest.kt
sschuberth: I don't see this being addressed. IMO it would still make sense to align on

    opossumInput.pathToSignal.values.forEach { signal ->

here.

fviernau: Of course it would make sense. But is it worth the effort?
I doubt it but will do anyway.

sschuberth (approved):

reporter/src/test/kotlin/reporters/OpossumReporterTest.kt
```diff
@@ -133,7 +133,7 @@ class OpossumReporterTest : WordSpec({
         }
 
         "create a file list that contains files from other lists" {
-            opossumInput.pathToSignal.forEach { e -> fileList shouldContain resolvePath(e.key) }
+            opossumInput.pathToSignal.keys.forEach { path -> fileList shouldContain resolvePath(path) }
```
sschuberth: Commit message nit: s/Oppossum/Opossum/

reporter/src/test/kotlin/reporters/OpossumReporterTest.kt
```diff
@@ -151,8 +151,7 @@ class OpossumReporterTest : WordSpec({
 
         "create a result that contains all packages in its signals" {
             result.analyzer!!.result.packages.forAll { pkg ->
-                val id = pkg.pkg.id
-                opossumInput.signals.find { it.id == id } shouldNot beNull()
+                opossumInput.signals.find { signal -> signal.id == pkg.pkg.id } shouldNot beNull()
```
sschuberth, line 154: Nit: This does more than just inlining `id` as it also introduces the (superfluous) `signal` lambda parameter.

sschuberth (approved):

Issue #4755, opened 2021-11-25
How to track CPEs in ORT?


sschuberth:
In addition to PURLs, from time to time there are inquiries to also track CPEs in ORT's data model, e.g. in order to let the advisor query vulnerability providers that only understand CPEs.

Open questions:
- Should we track CPEs at all?
- Should a CPE be a top-level `Package`-property like `purl`? Or should both be merged into an "external reference" concept similar to [SpdxExternalReference](https://github.com/oss-review-toolkit/ort/blob/master/utils/spdx/src/main/kotlin/model/SpdxExternalReference.kt)?
- Should the analyzer instead try to convert CPEs given in package metadata to PURL, e.g. via VulnerableCode like @pombredanne has mentioned?

heliocastro:
As middle of 2019, UE instituted a regulation for automotive industry related to cibersecurity. 
https://unece.org/sustainable-development/press/un-regulations-cybersecurity-and-software-updates-pave-way-mass-roll

Because of that, we need to track it down security related issues in our software anyway, and we are relying on CPE strins to track CVE advisories, as nothing else exists right now, and i don't see alternative initiatives.

So, even not tracking now, this questions will bite back again and again as Ort usage is increasing tenfold lately.

oheger-bosch:
Some thoughts:

I like the idea of having external references, this seems to be pretty flexible. I assume we would then have an enumeration like `ReferenceType`, and packages would contain a `Map` from `ReferenceType` to `String`, correct?

If a package manager can provide the CPE directly (as may be the case for SPDX), it should do so. It would be possible to have a post-processing step to generate/retrieve missing CPEs using an external source. As this may slow down the analyzer step, this should probably be an optional step that can be enabled if this information is needed.

pombredanne:
Some comments:
1. in general and IMHO tracking CPE is a good thing. For instance we track these as references in VulnerableCode.  They are far from great as identifiers especially for open source packages, but they have the merit to exist.
2. VulnerableCode was started in part because CPEs are weak when it comes to open source packages
3. Package URLs are an effective (emerging) alternative for  open source packages packages adopted in many places such as OWASP including more recently in the OSSF OSV schema and the CSAF Vex schema. So they are always a good choice IMHO.
4. VulnerableCode ends up being also a CPE<-> purl translator since we track both; we have a script that can effectively dump a translation between the two. You can see an example run in https://raw.githubusercontent.com/sbs2001/purl2cpe/master/2020.json

[...]

PR #4777, opened 2021-11-30
Model: Add support for CPE identifiers

oheger-bosch:
CPEs are an alternative mechanism to identify software components, and
in some areas, e.g. databases for security vulnerabilities, they are
in widespread use. To support use cases in these domains, enable ORT
to store these identifiers if they are available.

This PR is related to #4755.

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/spdx-project-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/spdx-project-expected-output.yml
@@ -64,6 +64,7 @@ packages:
     path: ""
 - id: "SpdxDocumentFile::zlib:1.2.11"
   purl: "pkg:spdxdocumentfile/zlib@1.2.11"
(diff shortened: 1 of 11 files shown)


sschuberth (dismissed):

sschuberth:
LGTM. Just a question: Shouldn't the first commit message already auto-close #4755?

oheger-bosch:
> LGTM. Just a question: Shouldn't the first commit message already auto-close #4755?

Yes, makes sense. I was not sure, however, to which commit message this should be added. Is the feature already complete in the first commit?

oheger-bosch:
> Well, as the issue is only about the general tracking in the data model, probably yes. But I could understand if you'd feel more comfortable to only add the auto-close to the last commit, as it concludes the series.

Ok, done in the last commit.

sschuberth (approved):

sschuberth:
Merging this without another test run as only the commit message changed.

PR #4779, opened 2021-11-30
Make the provenance storage configurable

mnonnenmacher:
Add configuration options to configure either a file storage or a
PostgreSQL storage to store provenance information. This is used only by
the experimental scanner.

By default a local file storage is used as prior to this commit.


fviernau (changes requested):

fviernau (approved):

mnonnenmacher:
Merged despite an unrelated failing Windows test.

PR #4783, opened 2021-11-30
spdx-utils: Add a constant for an empty package verification code

sschuberth:
This is useful as a "real looking" code that in fact describes no input.

### DIFF
--- a/utils/spdx/src/main/kotlin/SpdxConstants.kt
+++ b/utils/spdx/src/main/kotlin/SpdxConstants.kt
@@ -71,6 +71,11 @@ object SpdxConstants {
     const val LICENSE_LIST_URL = "https://spdx.org/licenses/"
 
     /**
+     * The package verification code for no input.
+     */
+    const val EMPTY_PACKAGE_VERIFICATION_CODE = "da39a3ee5e6b4b0d3255bfef95601890afd80709"
+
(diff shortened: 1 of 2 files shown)


utils/spdx/src/main/kotlin/SpdxConstants.kt
```diff
@@ -70,6 +70,11 @@ object SpdxConstants {
      */
     const val LICENSE_LIST_URL = "https://spdx.org/licenses/"
 
+    /**
+     * The package verification code for no input.
+     */
+    const val EMPTY_PACKAGE_VERIFICATION_CODE = "da39a3ee5e6b4b0d3255bfef95601890afd80709"
```
fviernau, line 76: The constant is only used by tests. What's the reason for putting it into production code?

fviernau (changes requested):

utils/spdx/src/main/kotlin/SpdxConstants.kt
sschuberth: The test verifies the correctness of the constant. The idea is to later use the constant in production code where (for whatever reason) having a package verification code is enforced, but there was no actual data to compute it on.

fviernau (approved):

PR #4787, opened 2021-12-01
purls: Extract 'toPurl' function's content

nnobelis:
This allows to customize the creation of purls in plugins i.e. create
[Package] instances with a special purl.
The extracted function offers also the possible to pass additional
query parameters to the purls that the previous implementation didn't.

### DIFF
--- a/model/src/main/kotlin/utils/Extensions.kt
+++ b/model/src/main/kotlin/utils/Extensions.kt
@@ -172,26 +172,51 @@ fun Identifier.getPurlType() =
  * [in the documentation](https://github.com/package-url/purl-spec/blob/master/README.rst#purl).
  * E.g. 'maven' for Gradle projects.
  */
-fun Identifier.toPurl() =
-    if (this == Identifier.EMPTY) {
-        ""
-    } else {
-        buildString {
-            append("pkg:")
-            append(getPurlType())
+fun Identifier.toPurl() = if (this == Identifier.EMPTY) "" else createPurl(getPurlType(), namespace, name, version)
 
-            if (namespace.isNotEmpty()) {
-                append('/')
-                append(namespace.percentEncode())
-            }
+/**
+ * Create the canonical [package URL](https://github.com/package-url/purl-spec) ("purl") based on given properties:
+ * [type] (which must be a String representation of a [PurlType] instance, [namespace], [name] and [version].
+ * Optional [qualifiers] may be given and will be appended to the purl as query parameters e.g.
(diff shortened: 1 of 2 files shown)


model/src/test/kotlin/IdentifierTest.kt
```diff
@@ -172,6 +173,18 @@ class IdentifierTest : WordSpec({
...
             purl shouldBe "pkg:type/namespace/name@release%20candidate"
         }
+
+        "allow query parameters" {
+            val purl = createPurl(
+                "type",
+                "namespace",
+                "name",
+                "version",
+                mapOf("argName" to "argValue")
```
oheger-bosch, line 183: Could you test with a map containing at least two parameters? So the different separator characters in the query string will be tested.

oheger-bosch (approved):

sschuberth (commented):
Sorry, looks like my review was not submitted...

model/src/main/kotlin/utils/Extensions.kt
```diff
@@ -172,27 +172,52 @@ fun Identifier.getPurlType() =
...
+ * pkg:golang/google.golang.org/genproto#googleapis/api/annotations
+ *
+ */
+fun createPurl(
+    type: String,
+    namespace: String,
+    name: String,
+    version: String,
+    qualifiers: Map<String, String> = emptyMap(),
+    subpath: List<String> = emptyList()
```
sschuberth, line 192: I believe this should just be a single `String`, because this is not a list of paths, but a single subpath. Note that you should *not* `percentEncode()` the whole subpath string at once then, as you'd want to keep the `/`, but do something like (typed blindly)

```
if (subpath.isNotEmpty()) {
    val value = subpath.split('/').joinToString("/", prefix = "#") { it.percentEncode() }
    append(value)
}
```

nnobelis: This is disputable as the specs says: `The subpath contains zero or more segments, separated by slash '/'`
https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst

We could have also renamed the parameter as `segments`.
However I am mildly opiniated on this so  I did the changes.

sschuberth (approved):

nnobelis:
It's seems the CI checks failure are flakiness: when I run `BabelFunTest` locally, it passes.

@sschuberth could you please check and if this the case, merge this PR ?

sschuberth:

PR #4796, opened 2021-12-02

mnonnenmacher:
Some improvements for the experimental scanner, please see the commit messages for details.

### DIFF
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -82,6 +82,7 @@ import org.ossreviewtoolkit.scanner.experimental.PostgresPackageProvenanceStorag
 import org.ossreviewtoolkit.scanner.experimental.ProvenanceBasedFileStorage
 import org.ossreviewtoolkit.scanner.experimental.ProvenanceBasedPostgresStorage
 import org.ossreviewtoolkit.scanner.experimental.ScanStorage
+import org.ossreviewtoolkit.scanner.experimental.ScannerWrapper
 import org.ossreviewtoolkit.scanner.scanOrtResult
 import org.ossreviewtoolkit.scanner.scanners.Askalono
 import org.ossreviewtoolkit.scanner.scanners.BoyterLc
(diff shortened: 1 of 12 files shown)


scanner/src/main/kotlin/experimental/ScannerWrapper.kt
```diff
@@ -54,6 +55,11 @@ sealed interface ScannerWrapper {
      * `options.ScanCode.criteria.minScannerVersion=3.0.2`.
      */
     val criteria: ScannerCriteria
+
+    /**
+     * Filter the scanner specific options to remove any secrets, like credentials.
+     */
+    fun filterSecretOptions(options: ScannerOptions) = options
```
sschuberth, line 62: Shouldn't we align the naming to `Scanner.filterOptionsForResult()` to avoid confusion? After all, it's up to the scanner implementor what this filters... it does not necessarily need to be secrets only.

sschuberth (changes requested):

cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -289,20 +290,27 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license
...
-        val scannerWrapper = when (scannerFactory.scannerName) {
-            "Askalono" -> Askalono.Factory().create(config.scanner, config.downloader)
-            "BoyterLc" -> BoyterLc.Factory().create(config.scanner, config.downloader)
-            "Licensee" -> Licensee.Factory().create(config.scanner, config.downloader)
-            "ScanCode" -> ScanCode.Factory().create(config.scanner, config.downloader)
-            else -> {
-                throw IllegalArgumentException(
-                    "The scanner ${scannerFactory.scannerName} is not supported by the experimental scanner."
-                )
+        // TODO: The experimental scanner supports using multiple scanner wrappers for projects and packages at once,
```
fviernau, line 293: commit message: Did you mean "To support that" instead of "To support this"

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -51,24 +52,49 @@ class ExperimentalScanner(
...
+        val packageScannerWrappers = scannerWrappers[DataEntity.PACKAGES].orEmpty()
+
+        val projectResults = if (projectScannerWrappers.isNotEmpty()) {
+            val packages = ortResult.getProjects().mapTo(mutableSetOf()) { it.toPackage() }
+
+            log.info { "Scanning ${packages.size} projects with ${projectScannerWrappers.size} scanners." }
+
+            scan(packages, DataEntity.PROJECTS)
+        } else {
+            log.info { "Skipping scan of projects because no project scanner wrapper is configured." }
```
fviernau, line 76: Would it make sense to drop the term "wrapper" from the log messages, because it's a technical detail?

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -51,24 +52,49 @@ class ExperimentalScanner(
...
         }
     }
 
     suspend fun scan(ortResult: OrtResult): OrtResult {
         val startTime = Instant.now()
 
-        val result = scan(
-            (ortResult.getProjects().map { it.toPackage() } + ortResult.getPackages().map { it.pkg }).toSet()
-        )
+        val projectScannerWrappers = scannerWrappers[DataEntity.PROJECTS].orEmpty()
```
fviernau, line 66: Have you considered following the Open-Closed principle here, so that we can later-on add more sophisticated methods for assigning a wrapper to a project or package without changing anything in the code of this class?

What I have in mind is injecting an instance of:
```
interface ScannerWrapperResolver {
       fun resolveWrapper(pkg: CuratedPackage, isProject: Boolean): List<ScannerWrapper>
}
```

What do you think?

fviernau (changes requested):

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
mnonnenmacher: We do not have any use case for filtering anything but secrets so I thought better be specific.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: Is there any real use case for this or is this more a theoretic question? Because I currently strive for feature parity with the default scanner implementation.

fviernau: I can imagine real use cases, but the design implications this would bring would be already worth it IMO even if there was no real use case. It becomes more modular and better testable and a bit less repetetive and slightly more generic. IIUC correctly it'd not add complexity, but maybe I missed something.

cli/src/main/kotlin/commands/ScannerCommand.kt
mnonnenmacher: I have rephrased the sentence to be more clear.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: I agree that it would be nice to have this, but it's too complex for me to implement right now, given that we don't have a use case at hand. Especially since it's unclear based on what the wrappers should be selected, should it be based on static configuration or custom code like you suggested on Slack? And what information should be available for the selection (just `pkg` and `isProject` as suggested above is certainly not enough)?
To me the priority is to cover the features we currently have to finally reach a state where we can replace the current scanner implementation.

fviernau: Let's keep it for now then.

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
sschuberth: But others may have other use-cases in the future, and then not dare to use this function because of its name... just thinking out aloud.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
sschuberth: BTW, this topic is basically documented as https://github.com/oss-review-toolkit/ort/issues/2819, and we actually do have a concrete use-case at hand by now as well: We'd like to use a snippet-scanner (as opposed to a regular license scanner) if a package has the `isModified` property set.

scanner/src/main/kotlin/experimental/ScannerWrapper.kt
sschuberth: Approving anyway.

sschuberth (approved):

fviernau (dismissed):

sschuberth (approved):

PR #4797, opened 2021-12-02
Add a defect report

oheger-bosch:
Analogously to the vulnerability report, a report for defects is now be generated if the ORT result contains corresponding information. For this purpose, a new template was added.

This PR also adapts the existing template for vulnerabilities to make sure that information and issues from the different advisors do not interfere with each other.

### DIFF
--- a/reporter/src/funTest/kotlin/TestData.kt
+++ b/reporter/src/funTest/kotlin/TestData.kt
@@ -19,9 +19,16 @@
 
 package org.ossreviewtoolkit.reporter
 
+import java.net.URI
 import java.time.Instant
 
 import org.ossreviewtoolkit.model.AccessStatistics
+import org.ossreviewtoolkit.model.AdvisorCapability
+import org.ossreviewtoolkit.model.AdvisorDetails
+import org.ossreviewtoolkit.model.AdvisorRecord
+import org.ossreviewtoolkit.model.AdvisorResult
+import org.ossreviewtoolkit.model.AdvisorRun
+import org.ossreviewtoolkit.model.AdvisorSummary
 import org.ossreviewtoolkit.model.AnalyzerResult
 import org.ossreviewtoolkit.model.AnalyzerRun
 import org.ossreviewtoolkit.model.ArtifactProvenance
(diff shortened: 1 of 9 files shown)


oheger-bosch:
Here is a screenshot how a defect is rendered in the report:

defects_report

reporter/src/funTest/kotlin/TestData.kt
```diff
@@ -19,9 +19,16 @@
 
 package org.ossreviewtoolkit.reporter
 
+import java.net.URI
```
sschuberth, line 22: > Prevent the generation of empty reports by enabling a specific advisor
report only if there is an advisor result with findings of this
specific type.

I'm not sure I agree to this rationale. How do you then, as a user, distinguish between the cases where a report was not generated because it was not enabled at all, or because there were no findings? In other words, I would image users to prefer a "written confirmation" that there were *no* findings.

sschuberth (changes requested):

sschuberth:
Thanks for the screenshot! I'm surprised to see closed defects being listed by default, though. What's the reason for that decision? Aren't users usually only interested in still open defects?

*Edit:* A, I think I got it: The defect is still "open" (i.e. exists) in version in version 0.4.4, but it was fixed ("closed") in version 0.4.9. Is that's the way how to read it? I believe we need to improve this...

oheger-bosch:
> Thanks for the screenshot! I'm surprised to see closed defects being listed by default, though. What's the reason for that decision? Aren't users usually only interested in still open defects?
> 
> _Edit:_ A, I think I got it: The defect is still "open" (i.e. exists) in version in version 0.4.4, but it was fixed ("closed") in version 0.4.9. Is that's the way how to read it? I believe we need to improve this...

Yes, this is the meaning. Do you have a proposal how this could be rendered?

reporter/src/funTest/kotlin/TestData.kt
oheger-bosch: Valid point. I think, what we actually want is to generate reports depending on the advisors that have run. But unfortunately, this information is not available yet in an ORT result; an advisor only occurs if it has generated findings.

So I think I will change this for now to always generate the defects report if there are any advisor results. Do you agree?

sschuberth:
> Do you have a proposal how this could be rendered?

Let's discuss this quickly by voice.

oheger-bosch:
As discussed, the information about the fix availability have been consolidated. The report now shows either the fix release URL or version, or a text that no fix is available:

defects_report

reporter/src/funTest/kotlin/TestData.kt
oheger-bosch: Changed accordingly.

sschuberth: > So I think I will change this for now to always generate the defects report if there are any advisor results. Do you agree?

I'm not sure. Wouldn't it be confusing to get a defects report if no advisor capable of providing defects is configured? What would be the contents of that report then? "No defects found" or could we do "No defect advisor configured" (which IMO would be preferable)?

oheger-bosch: Currently, the report says that no defects were found.

As I said, the ORT result does not contain the information which advisor was run. I think, it would make sense to change this, but this would be a bigger effort; so this could be a topic for a separate PR.

oheger-bosch:
Newest styling of the list of defects:

defects_report

sschuberth (dismissed):

sschuberth (approved):

sschuberth (approved):

PR #4800, opened 2021-12-03
Rename `DataEntity` to `PackageType`

mnonnenmacher:
`DataEntity` is a very generic name, replace it with a more specific
name that better describes the purpose of the enum. While at it, also
improve the documentation of the enum and rename the command line
parameters accordingly.

### DIFF
--- a/cli/src/main/kotlin/commands/DownloaderCommand.kt
+++ b/cli/src/main/kotlin/commands/DownloaderCommand.kt
@@ -196,7 +196,7 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
 
     private fun downloadFromOrtResult(ortFile: File, failureMessages: MutableList<String>) {
         println(
-            "Downloading ${packageTypes.joinToString(" and ")} from ORT result file at " +
+            "Downloading ${packageTypes.joinToString(" and ") { "${it}s" }} from ORT result file at " +
                     "'${ortFile.canonicalPath}'..."
         )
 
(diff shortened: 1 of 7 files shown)


fviernau (dismissed):
Out of scope: Would it be a good idea to turn the plural enum elements into singular ones in a separate PR ?

sschuberth:
Note that *detekt* analysis failed.

mnonnenmacher:
> Out of scope: Would it be a good idea to turn the plural enum elements into singular ones in a separate PR ?

I have added a commit for that.

cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -213,11 +213,11 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
         }
 
         val packages = mutableListOf<Package>().apply {
-            if (PackageType.PROJECTS in packageTypes) {
+            if (PackageType.PROJECT in packageTypes) {
```
sschuberth, line 216: Nit: I think this commit should then also adjust the `Downloading ${packageTypes.joinToString(" and ")} from ORT` code from the previous commit by adding a plural "s" after `}`.

cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -196,7 +196,7 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
 
     private fun downloadFromOrtResult(ortFile: File, failureMessages: MutableList<String>) {
         println(
-            "Downloading ${packageTypes.joinToString(" and ")} from ORT result file at " +
+            "Downloading ${packageTypes.joinToString(" and ") { "${it}s" }} from ORT result file at " +
```
fviernau, line 199: The curly braces around `it` are unnecessary.

fviernau (changes requested):

cli/src/main/kotlin/commands/DownloaderCommand.kt
sschuberth: I don't think so, as you need to separate from the "s".

sschuberth (approved):

cli/src/main/kotlin/commands/DownloaderCommand.kt
fviernau: OMG, you're right ^^

fviernau (approved):

PR #4804, opened 2021-12-03
ExperimentalScanner: Support the creation of file archives

mnonnenmacher:
Create file archives in a way compatible with the way they are created
in the `LocalScanner`. This allows to keep using the existing file
archives without changing the logic used to read those file archives in
the reporter.

### DIFF
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -20,8 +20,12 @@
 
 package org.ossreviewtoolkit.scanner.experimental
 
+import java.io.File
+import java.nio.file.StandardCopyOption
 import java.time.Instant
 
+import kotlin.io.path.moveTo
+import kotlin.time.measureTime
 import kotlin.time.measureTimedValue
 
 import org.ossreviewtoolkit.downloader.DownloadException
@@ -40,7 +44,9 @@ import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.config.ScannerOptions
+import org.ossreviewtoolkit.model.config.createFileArchiver
 import org.ossreviewtoolkit.utils.common.collectMessagesAsString
+import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.core.Environment
 import org.ossreviewtoolkit.utils.core.log
 
@@ -278,6 +284,8 @@ class ExperimentalScanner(
             }
         }
 
+        createFileArchives(nestedProvenances)
+
         return nestedProvenanceScanResults
     }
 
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -421,6 +428,50 @@ class ExperimentalScanner(
             ScanResult(provenance, scanner.details, summary)
         }
     }
+
+    private fun createFileArchives(nestedProvenances: Map<Package, NestedProvenance>) {
+        // TODO: The archives are currently created in a way compatible with the existing implementation in the
+        //       LocalScanner. This allows to keep using existing file archives without changing the logic used to
+        //       access those archives in the reporter. To achieve this nested provenances are downloaded recursively.
```
fviernau, line 435: Is the sentence "To achieve this nested provenances are downloaded recursively" intended like it is now? I think there's something missing...

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -20,8 +20,11 @@
 
 package org.ossreviewtoolkit.scanner.experimental
 
+import java.io.File
```
fviernau, line 23: commit(msg): Does this backwards compatibility come with a (negative) performance impact?

If so, what do you think how much effort it would take to implement it in a non-backwards compatible way with better performance?

fviernau (changes requested):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: I have improved the sentence a bit to make clear why they are downloaded recursively.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: For large repositories with may nested repositories creating the archives might take a while, but this is a temporary solution anyway to make the experimental scanner feature complete to be able to start testing. It will be replaced with the proper implementation later on.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -421,6 +428,51 @@ class ExperimentalScanner(
...
+            provenancesWithMissingArchives.forEach { (_, nestedProvenance) ->
+                val dir = downloadRecursively(nestedProvenance)
+                archiver.archive(dir, nestedProvenance.root)
+                dir.safeDeleteRecursively(force = true)
+            }
+        }
+
+        log.info {
+            "Created file archives for ${provenancesWithMissingArchives.size} packages in " +
+                    "${duration.inWholeMilliseconds} ms."
```
sschuberth, line 458: Please don't use an explicit unit anymore, but let the `Duration` class itself do the formatting. Also see https://github.com/oss-review-toolkit/ort/pull/4573.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -421,6 +428,51 @@ class ExperimentalScanner(
...
+    private fun downloadRecursively(nestedProvenance: NestedProvenance): File {
+        // Use the provenanceDownloader to download each provenance from nestedProvenance separately, because they are
+        // likely already cached if a local scanner wrapper is used.
+
+        val root = provenanceDownloader.download(nestedProvenance.root)
+
+        nestedProvenance.subRepositories.forEach { (path, provenance) ->
+            val tempDir = provenanceDownloader.download(provenance)
+            val targetDir = root.resolve(path)
+            tempDir.toPath().moveTo(targetDir.toPath())
```
sschuberth, line 471: Can we rely on source and target files to reside on the same filesystem? If so, please pass `StandardCopyOption.ATOMIC_MOVE` as a copy option.

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: Thanks, I didn't notice #4573. It's nice for the end-user but it makes log analytics for performance very difficult. I have changed it for consistency but I think we should reconsider this sometime.

sschuberth (approved):

mnonnenmacher:
Merged because failing tests were unrelated.

PR #4809, opened 2021-12-05
Authenticator refactoring

sschuberth:

### DIFF
--- a/utils/core/src/main/kotlin/EnvVarAuthenticator.kt
+++ b/utils/core/src/main/kotlin/EnvVarAuthenticator.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2020-2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.utils.core
+
+import java.net.Authenticator
+import java.net.PasswordAuthentication
+
+import org.ossreviewtoolkit.utils.common.Os
+
+/**
+ * A simple non-caching authenticator that reads credentials from the [ORT_HTTP_USERNAME_ENV_NAME] and
+ * [ORT_HTTP_PASSWORD_ENV_NAME] [environment variables][env].
+ */
+class EnvVarAuthenticator(env: Map<String, String> = Os.env) : Authenticator() {
+    companion object {
+        const val ORT_HTTP_USERNAME_ENV_NAME = "ORT_HTTP_USERNAME"
+        const val ORT_HTTP_PASSWORD_ENV_NAME = "ORT_HTTP_PASSWORD"
+    }
+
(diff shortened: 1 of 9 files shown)


utils/core/src/main/kotlin/EnvVarAuthenticator.kt
```diff
@@ -0,0 +1,41 @@
...
+import java.net.PasswordAuthentication
+
+import org.ossreviewtoolkit.utils.common.Os
+
+/**
+ * A simple non-caching authenticator that reads credentials from [environment variables][env].
+ */
+class EnvVarAuthenticator(env: Map<String, String> = Os.env) : Authenticator() {
+    private val usernameFromEnv = env["ORT_HTTP_USERNAME"]
+    private val passwordFromEnv = env["ORT_HTTP_PASSWORD"]
```
oheger-bosch, line 32: You could already create the `PasswordAuthentication` beforehand based on the values in the env variables and store it in a nullable field. Then `getPasswordAuthentication()` would just return the stored authentication or call the super method if it is *null*.

utils/core/src/main/kotlin/EnvVarAuthenticator.kt
```diff
@@ -0,0 +1,41 @@
...
+
+package org.ossreviewtoolkit.utils.core
+
+import java.net.Authenticator
+import java.net.PasswordAuthentication
+
+import org.ossreviewtoolkit.utils.common.Os
+
+/**
+ * A simple non-caching authenticator that reads credentials from [environment variables][env].
```
oheger-bosch, line 28: Suggestion: Maybe name the variables explicitly in the comment, as they are hard-coded.

utils/core/src/main/kotlin/OrtAuthenticator.kt
```diff
@@ -71,9 +68,9 @@ class OrtAuthenticator(private val original: Authenticator? = null) : Authentica
...
     }
 
-    // TODO: Add support for ".authinfo" files (which use the same syntax as .netrc files) once Git.kt does not call the
-    //       Git CLI anymore which only supports ".netrc" (and "_netrc") files.
-    private val netrcFileNames = listOf(".netrc", "_netrc")
+    // First look for (potentially machine-specific) credentials in a netrc-style file, then look for generic
+    // credentials passed as environment variables.
+    private val delegateAuthenticators = listOf(NetRcAuthenticator(), EnvVarAuthenticator())
 
     private val serverAuthentication = mutableMapOf<String, PasswordAuthentication>()
```
oheger-bosch, line 75: This has not been changed by this PR, but storing the credentials in a plain map is not thread-safe.

utils/core/src/main/kotlin/NetRcAuthenticator.kt
```diff
@@ -0,0 +1,83 @@
...
+
+        return super.getPasswordAuthentication()
+    }
+}
+
+/**
+ * Parse the [contents] of a [.netrc](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html)
+ * file for a login / password matching [machine].
+ */
+internal fun getNetrcAuthentication(contents: String, machine: String): PasswordAuthentication? {
```
oheger-bosch, line 57: Just wondering: Instead of parsing the netrc files for every machine, wouldn't it be more efficient to read them once and extract the credentials for all machines?

utils/core/src/main/kotlin/OrtAuthenticator.kt
sschuberth: Changed in a new commit.

utils/core/src/main/kotlin/NetRcAuthenticator.kt
sschuberth: My thinking was that for this non-caching low-level authenticator we might want to be able to consider changes to `.netrc` files during the runtime of the application (mainly for on-the-fly debugging). If that's not desired, a higher level authenticator can implement caching, like the `OrtAuthenticator` does.

oheger-bosch (approved):

PR #4811, opened 2021-12-06
SpdxDocumentFile: Support preemptive HTTP authentication

sschuberth:
Support preemptive authentication when resolving remote SPDX documents
via HTTP. This is required to download e.g. from Bitbucket which does
not trigger reactive authentication via a 401 "Not Authorized" response,
but forwards to a login page.

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -22,6 +22,7 @@
 package org.ossreviewtoolkit.analyzer.managers
 
 import java.io.File
+import java.net.Authenticator
 import java.net.URI
 import java.util.SortedSet
 
@@ -48,7 +49,9 @@ import org.ossreviewtoolkit.model.utils.toPurl
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.common.withoutPrefix
 import org.ossreviewtoolkit.utils.core.OkHttpClientHelper
+import org.ossreviewtoolkit.utils.core.addBasicAuthorization
 import org.ossreviewtoolkit.utils.core.createOrtTempDir
+import org.ossreviewtoolkit.utils.core.downloadFile
 import org.ossreviewtoolkit.utils.core.log
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -158,7 +162,31 @@ private fun SpdxExternalDocumentReference.resolve(definitionFile: File, issues:
...
+            addInterceptor { chain ->
+                val request = chain.request()
+                val requestBuilder = request.newBuilder()
+
+                // Use the authenticator also to request preemptive authentication.
+                val auth = Authenticator.requestPasswordAuthentication(
+                    /* host = */ uri.host,
+                    /* addr = */ null,
+                    /* port = */ uri.port,
+                    /* protocol = */ uri.scheme,
```
mnonnenmacher, line 176: Is it correct to use `uri.scheme` here and not two rows below?

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -158,7 +162,31 @@ private fun SpdxExternalDocumentReference.resolve(definitionFile: File, issues:
         }
     } else {
         tempDir = createOrtTempDir()
-        OkHttpClientHelper.downloadFile(uri.toString(), tempDir).getOrNull() ?: run {
+
+        val client = OkHttpClientHelper.buildClient {
+            addInterceptor { chain ->
```
oheger-bosch, line 167: As this interceptor is quite generic, it could be worth extracting it to an utility class, e.g. `PreemptiveAuthenticationInterceptor`.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: Yes. What `URI` calls "scheme" is the protocol, and what the `Authenticator` calls "scheme" is the "authorization scheme".

mnonnenmacher (dismissed):

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: I've extracted it to an extension function instead.

oheger-bosch (approved):
Looks good to me.

Just noticed that in contrast to the previous version, the credentials are now retrieved once when the HTTP client is setup and no longer dynamically for each request. Alternatively, the code to query the authenticator could also be moved into the interceptor.

PR #4815, opened 2021-12-07
AdvisorConfiguration: Disable serialization of generic options

oheger-bosch:
The map with generic advisor options (which is used to configure
external advisor plugins) may contain sensitive information like
credentials and API keys and should therefore not be serialized to an
ORT result.


model/src/test/kotlin/config/AdvisorConfigurationTest.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    "VulnerableCodeConfiguration" should {
+        "support a serialization round-trip via an ObjectMapper" {
+            val originalConfig = loadAdvisorConfig()
+            val rereadConfig = rereadAdvisorConfig(originalConfig)
+
+            val expectedVCConfig = originalConfig.vulnerableCode.shouldNotBeNull()
+            val actualVCConfig = rereadConfig.vulnerableCode.shouldNotBeNull()
+            actualVCConfig shouldBe expectedVCConfig
```
sschuberth, line 60: Nit: Blank line before this one.

model/src/test/kotlin/config/AdvisorConfigurationTest.kt
```diff
@@ -81,6 +81,12 @@ class AdvisorConfigurationTest : WordSpec({
             rereadConfig.token.shouldBeNull()
         }
     }
+
+    "generic advisor options" should {
+        "not be serialized, as they may contain sensitive information" {
```
sschuberth, line 86: Nits:
- drop the comma
- s/may/might/

sschuberth (changes requested):

sschuberth (approved):

PR #4816, opened 2021-12-07
Use `GENERATED_FROM` SPDX relations to distinguish between source and binary artifacts

sschuberth:


oheger-bosch (dismissed):
Changes look good to me. I am, however, not deep enough in the SPDX spec to fully verify the correctness.

fviernau (approved):

PR #4823, opened 2021-12-07
Merge SPDX project and package directories into a single hierarchy

sschuberth:


oheger-bosch (changes requested):
There are test failures. `SpdxDocumentFileTest` needs to be adapted to the changes as well.

Otherwise, looks good to me.

oheger-bosch (approved):

PR #4824, opened 2021-12-08
Npm: Performance and debug logging

fviernau:
Some `debug` log statements may trash the analysis execution performance as they are to frequent. So, reduce the level to `trace`. Apart from that save I/Os by a factor of about 1000 and add some alternative debug logging so that there is at least some status output on the `debug` log level.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -519,15 +519,12 @@ open class Npm(
             return null
         }
 
-        log.debug { "Building dependency tree for '${moduleInfo.name}' from directory '$moduleDir'." }
-
         val pathToRoot = listOf(moduleDir) + ancestorModuleDirs
         moduleInfo.dependencyNames.forEach { dependencyName ->
             val dependencyModuleDirPath = findDependencyModuleDir(dependencyName, pathToRoot)
 
             if (dependencyModuleDirPath.isNotEmpty()) {
                 val dependencyModuleDir = dependencyModuleDirPath.first()
-                log.debug { "Found module dir for '$dependencyName' at '$dependencyModuleDir'." }
 
                 getModuleInfo(
                     moduleDir = dependencyModuleDir,
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -356,43 +356,47 @@ open class Npm(
         val workingDir = definitionFile.parentFile
 
         stashDirectories(workingDir.resolve("node_modules")).use {
-            // Actually installing the dependencies is the easiest way to get the metadata of all transitive
-            // dependencies (i.e. their respective "package.json" files). As NPM uses a global cache, the same
-            // dependency is only ever downloaded once.
-            installDependencies(workingDir)
-
-            // Create packages for all modules found in the workspace and add them to the graph builder. They are
-            // reused when they are referenced by scope dependencies.
-            val packages = parseInstalledModules(workingDir)
-            graphBuilder.addPackages(packages.values)
-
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -519,15 +519,15 @@ open class Npm(
             return null
         }
 
-        log.debug { "Building dependency tree for '${moduleInfo.name}' from directory '$moduleDir'." }
+        log.trace { "Building dependency tree for '${moduleInfo.name}' from directory '$moduleDir'." }
```
sschuberth, line 530: We don't actually use the "trace" level in ORT yet (besides a single use in the contributed `OpossumReporter`; I'm going to remove that single use), and we don't even allow user to set the "trace" level in `OrtMain`. So How about simply dropping these?

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -355,8 +355,12 @@ open class Npm(
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
         val workingDir = definitionFile.parentFile
 
-        stashDirectories(workingDir.resolve("node_modules")).use {
-            return resolveDependenciesInternal(definitionFile)
+        try {
+            stashDirectories(workingDir.resolve("node_modules")).use {
+                return resolveDependenciesInternal(definitionFile)
```
sschuberth, line 360: Please move out the `return` to before `try` for visibility.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -591,7 +600,9 @@ open class Npm(
             version = version,
             dependencyNames = dependencyNames,
             packageJson = packageJsonFile
-        )
+        ).also {
+            rawModuleInfoCache[cacheKey] = it
```
sschuberth, line 605: Can we rather leverage `getOrPut()` here?

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -574,6 +574,7 @@ open class Npm(
         rawModuleInfoCache[cacheKey]?.let { return it }
 
         val packageJsonFile = moduleDir.resolve("package.json")
+        log.debug { "Parsing module info from '${packageJsonFile.absolutePath}'." }
```
sschuberth, line 577: I believe it'd be fine to squash this into the previous commit, esp. if the first commit is changed to simply drop the logs.

sschuberth (dismissed):

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: > I'm going to remove that single use

See https://github.com/oss-review-toolkit/ort/pull/4825.

analyzer/src/main/kotlin/managers/Npm.kt
fviernau: There is no reading involved: `get` would always return `null` so the "or put" code branch would always be executed. I'd find this a bit misleading. Ok if we keep it as-is?

sschuberth: I'm confused, doesn't `rawModuleInfoCache[cacheKey]?.let { return it }` in line 571 do the `get`?

fviernau: Ah, you want all the code go into the getOrPut lambda. That wasn't clear to me...

sschuberth: ...yes, as anything else wouldn't make sense 😉

sschuberth (approved):

PR #4826, opened 2021-12-08
SpdxExpression: Add a new ALLOW_LICENSEREF_EXCEPTIONS strictness level

sschuberth:

### DIFF
--- a/utils/spdx/src/main/kotlin/SpdxExpression.kt
+++ b/utils/spdx/src/main/kotlin/SpdxExpression.kt
@@ -58,7 +58,13 @@ sealed class SpdxExpression {
          * Only current SPDX license identifier strings and LicenseRefs are allowed. This excludes deprecated SPDX
          * license identifier strings and arbitrary license identifier strings.
          */
-        ALLOW_CURRENT
+        ALLOW_CURRENT,
+
+        /**
+         * This is the same as [ALLOW_CURRENT], but additionally allows LicenseRefs that contain the "exception" string
+         * to be used as license exceptions after the [WITH] operator.
+         */
+        ALLOW_LICENSEREF_EXCEPTIONS
     }
 
     companion object {
@@ -433,6 +439,8 @@ class SpdxLicenseWithExceptionExpression(
     val exception: String
 ) : SpdxSingleLicenseExpression() {
     companion object {
+        private val EXCEPTION_STRING_PATTERN = Regex("\\bexception\\b", RegexOption.IGNORE_CASE)
+
         /**
          * Parse a string into an [SpdxLicenseWithExceptionExpression]. Throws an [SpdxException] if the string cannot
          * be parsed. Throws a [ClassCastException] if the string is not an [SpdxLicenseWithExceptionExpression].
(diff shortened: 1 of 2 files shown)


utils/spdx/src/main/kotlin/SpdxExpression.kt
```diff
@@ -477,7 +477,7 @@ class SpdxLicenseWithExceptionExpression(
 
         val spdxException = SpdxLicenseException.forId(exception)
         when (strictness) {
-            Strictness.ALLOW_ANY -> exception // Return something non-null.
+            Strictness.ALLOW_ANY -> Unit // Return something non-null to not trigger the fallback below.
```
oheger-bosch, line 486: To be honest, I find the whole construct difficult to understand. Why is an exception constructed first, and then another one is thrown? I would rather expect something like

```
val spdxException = when(strictness) {
  ...
}
spdxException?.let { throw it }
```

sschuberth: Note that `spdxException` is not a JVM exception, but an SPDX license exception instance. I'll rename the variable to `spdxLicenseException` to make that clearer.

sschuberth: I decided to rewrite this completely, getting rid of the `when`.

utils/spdx/src/main/kotlin/SpdxExpression.kt
```diff
@@ -433,6 +433,8 @@ class SpdxLicenseWithExceptionExpression(
     val exception: String
 ) : SpdxSingleLicenseExpression() {
     companion object {
+        private val EXCEPTION_STRING_PATTERN = Regex("\\bexception\\b", RegexOption.IGNORE_CASE)
```
mnonnenmacher, line 436: Commit message: Explain what was broken that is fixed with this commit.

oheger-bosch (commented):
Looks good to me, but as I am not too familiar with SPDX validation, I would rather another pair of eyes to have a look at the changes.

sschuberth:
> Looks good to me, but as I am not too familiar with SPDX validation, I would rather another pair of eyes to have a look at the changes.

@mnonnenmacher maybe?

mnonnenmacher (changes requested):

utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
```diff
@@ -175,6 +175,46 @@ class SpdxExpressionTest : WordSpec() {
...
+                    licenseRefExpressionWithException2.toSpdx(Strictness.ALLOW_ANY)
+                    licenseRefExpressionWithLicenseRefExpression.toSpdx(Strictness.ALLOW_ANY)
+                }
+            }
+
+            "be valid in deprecated mode" {
+                assertSoftly {
+                    licenseRefExpression.toSpdx(Strictness.ALLOW_DEPRECATED)
+                    licenseRefExpressionWithException.toSpdx(Strictness.ALLOW_DEPRECATED)
+                    licenseRefExpressionWithException2.toSpdx(Strictness.ALLOW_DEPRECATED)
```
mnonnenmacher, line 197: I understood the code so that with `ALLOW_DEPRECATED` and `ALLOW_CURRENT` only SPDX exception identifiers are allowed. This would also be according with the spec (see [D.4.4](https://spdx.github.io/spdx-spec/SPDX-license-expressions/#d44-exception-with-operator)):

> If the applicable exception is not found on the SPDX License Exception List, then use a single <license-ref> to represent the entire license terms (including the exception).

mnonnenmacher (changes requested):

utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
mnonnenmacher: I don't really understand why custom exceptions are not allowed there, but it's also in the draft for SPDX v3: https://spdx.github.io/spdx-spec/v3-draft/SPDX-license-expressions/#d44-exception-with-operator

mnonnenmacher: There is an open issue to change that for v3: https://github.com/spdx/spdx-spec/issues/153

sschuberth: I've rewritten the code to introduce a new `ALLOW_LICENSEREF_EXCEPTIONS` strictness level instead.

mnonnenmacher (approved):

PR #4827, opened 2021-12-08
Reduce reflection

sschuberth:


cli/src/main/kotlin/commands/UploadCurationsCommand.kt
```diff
@@ -174,7 +173,7 @@ private fun PackageCuration.toContributionPatch(): ContributionPatch? {
     )
 
     val curation = Curation(
-        described = described.takeIf { it.hasNonNullProperty() },
+        described = described.takeIf { it != Described() },
```
oheger-bosch, line 176: There could be a constant `Described.Empty`, which would make this more readable and efficient.

sschuberth: I was indeed considering this, but decided against it. That's mostly because I found it to be inconsistent to have an `Empty` (or `EMPTY`) for this class but not the others. Also, I think the current code highlights nicely that we compare to the default instance of the class, as no arguments are passed.

oheger-bosch (approved):

PR #4829, opened 2021-12-09
SpdxDocumentFile: Introduce a helper class to load documents

oheger-bosch:
The class offers functionality to load SPDX documents from files,
cache the loaded documents, and verify their checksums. It is going to
be used by SpdxDocumentFile to prevent that local files are parsed
multiple times.

This PR is a preparation for upcoming improvements on the SPDX package manager to detect all transitive dependencies.

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -56,7 +56,6 @@ import org.ossreviewtoolkit.utils.core.downloadFile
 import org.ossreviewtoolkit.utils.core.log
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
-import org.ossreviewtoolkit.utils.spdx.SpdxModelMapper
 import org.ossreviewtoolkit.utils.spdx.model.SpdxDocument
 import org.ossreviewtoolkit.utils.spdx.model.SpdxExternalDocumentReference
 import org.ossreviewtoolkit.utils.spdx.model.SpdxExternalReference
@@ -106,15 +105,16 @@ internal fun SpdxDocument.projectPackage(): SpdxPackage? =
         ?.singleOrNull { it.packageFilename.isEmpty() || it.packageFilename == "." }
 
 /**
- * Get the [SpdxPackage] from the [SpdxExternalDocumentReference] that the [packageId] refers to, where [definitionFile]
- * is used to resolve local relative URIs to files.
+ * Use [loader] to get the [SpdxPackage] from the [SpdxExternalDocumentReference] that the [packageId] refers to, where
+ * [definitionFile] is used to resolve local relative URIs to files.
  */
 internal fun SpdxExternalDocumentReference.getSpdxPackage(
+    loader: SpdxDocumentLoader,
     packageId: String,
     definitionFile: File,
     issues: MutableList<OrtIssue>
 ): SpdxPackage? {
-    val externalSpdxDocument = resolve(definitionFile, issues) ?: return null
+    val externalSpdxDocument = resolve(loader, definitionFile, issues) ?: return null
 
     val spdxDocumentPath = URI(spdxDocument).path
     val spdxDocumentFile = definitionFile.resolveSibling(spdxDocumentPath).normalize()
@@ -134,7 +134,11 @@ internal fun SpdxExternalDocumentReference.getSpdxPackage(
 /**
  * Return the [SpdxDocument] this [SpdxExternalDocumentReference]'s [SpdxDocument] refers to.
  */
-private fun SpdxExternalDocumentReference.resolve(definitionFile: File, issues: MutableList<OrtIssue>): SpdxDocument? {
+private fun SpdxExternalDocumentReference.resolve(
+    loader: SpdxDocumentLoader,
+    definitionFile: File,
+    issues: MutableList<OrtIssue>
+): SpdxDocument? {
     val uri = runCatching { URI(spdxDocument) }.getOrElse {
         issues += createAndLogIssue(
             source = MANAGER_NAME,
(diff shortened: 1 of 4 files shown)


analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -106,15 +105,16 @@ internal fun SpdxDocument.projectPackage(): SpdxPackage? =
         ?.singleOrNull { it.packageFilename.isEmpty() || it.packageFilename == "." }
 
 /**
- * Get the [SpdxPackage] from the [SpdxExternalDocumentReference] that the [packageId] refers to, where [definitionFile]
- * is used to resolve local relative URIs to files.
+ * Get the [SpdxPackage] from the [SpdxExternalDocumentReference] using [loader] that the [packageId] refers to, where
```
mnonnenmacher, line 108: This makes the sentence hard to read, I'd rather add it to the beginning: "Use [load] to get..."

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/SpdxDocumentLoader.kt
```diff
@@ -0,0 +1,97 @@
...
+    val document: SpdxDocument,
+
+    /** A map with checksums that have been validated. */
+    val checksums: MutableMap<SpdxChecksum, Boolean>
+) {
+    /**
+     * Verify this document [file] against the given [checksum]. Record checksums that have been verified.
+     */
+    fun verify(file: File, checksum: SpdxChecksum): Boolean = checksums.getOrPut(checksum) {
+        val hash = Hash.create(checksum.checksumValue, checksum.algorithm.name)
```
sschuberth, line 94: See above.

analyzer/src/main/kotlin/managers/SpdxDocumentLoader.kt
```diff
@@ -0,0 +1,99 @@
...
+ * A data class combining an [SpdxDocument] with a flag whether it has been successfully validated against a checksum.
+ * Instances of this class are returned by [SpdxDocumentLoader.loadAndVerify] to represent the two return values: the
+ * document itself and the result of the checksum verification.
+ */
+internal data class VerifiedDocument(
+    /** The [SpdxDocument] that was loaded. */
+    val document: SpdxDocument,
+
+    /** Flag whether verification against the provided checksum was successful. */
+    val valid: Boolean
```
sschuberth, line 77: I'm not sure whether it's correct to bundle these properties in this way. In SPDX, the checksum is stored along with the *reference*, not with the document. Which means that different "parent" SPDX documents could refer to the same "child" SPDX document, but with different checksum, where one checksum is correct and the other is not. How would you model this here?

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -554,7 +563,7 @@ class SpdxDocumentFile(
 
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
         // For direct callers of this function mapDefinitionFiles() did not populate the map before, so add a fallback.
```
sschuberth, line 565: I believe this comment can go away now, too.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -420,7 +426,12 @@ class SpdxDocumentFile(
         }
 
         return packageForExternalDocumentId.getOrPut(externalDocumentReference.externalDocumentId) {
-            externalDocumentReference.getSpdxPackage(identifier.substringAfter(":"), definitionFile, issues)
+            externalDocumentReference.getSpdxPackage(
+                spdxDocumentLoader,
+                identifier.substringAfter(":"),
+                definitionFile,
+                issues
+            )
```
sschuberth, line 434: The formatting of the line below now also needs to be adjusted.

analyzer/src/main/kotlin/managers/SpdxDocumentLoader.kt
sschuberth: As discussed verbally, I'd prefer to get rid of this class in favor of returning `Pair<SpdxDocument, Boolean>` from `loadAndVerify()`.

analyzer/src/main/kotlin/managers/SpdxDocumentLoader.kt
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * An internal data class to store information about SPDX documents that have been loaded. In addition to the document
+ * itself, the verified checksums are recorded.
+ */
+private data class DocumentEntry(
+    /** The [SpdxDocument] that was loaded. */
+    val document: SpdxDocument,
+
+    /** A map with checksums that have been validated. */
+    val checksums: MutableMap<SpdxChecksum, Boolean>
```
sschuberth, line 89: As this parameter is never set to anything non-empty, I wonder whether it could become a private member property instead.

In general, this class caches checksum in a very similar way like `SpdxDocumentLoader` caches documents. So could this also become a plain `class` instead of a `data class`, like

```
private class DocumentEntry(
    /** The [SpdxDocument] that was loaded. */
    val document: SpdxDocument
) {
    /** A map with checksums that have been validated. */
    private val checksums = mutableMapOf<SpdxChecksum, Boolean>()

    /**
     * Verify this document [file] against the given [checksum]. Record checksums that have been verified.
     */
    fun verify(file: File, checksum: SpdxChecksum): Boolean =
        checksums.getOrPut(checksum) {
            val hash = Hash.create(checksum.checksumValue, checksum.algorithm.name)
            hash.verify(file)
        }
```

sschuberth (changes requested):

oheger-bosch:
Moving this back to draft state. On progressing further with supporting transitive dependencies in the SPDX package manager, it could be that parts of the functionality provided here are no longer needed or will change.

sschuberth:
> parts of the functionality provided here are no longer needed or will change.

Just FYI, IMO changing / extending this class as part of that work would be fine. I find it easier to initially review a simple version of the class on introduction, and then extending it as needed in future PRs. So from that perspective, I'm not opposed to merging this.

The case is different, of course, if the "no longer needed" case comes true.

oheger-bosch:
Closing this PR in favour of #4844, which partly includes these changes.

PR #4834, opened 2021-12-10
Gomod: Handle some more pseudo version formats

fviernau:


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -321,6 +321,7 @@ private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>
 }
 
 // See https://golang.org/ref/mod#pseudo-versions.
+// Format for no known base version, e.g. v0.0.0-20191109021931-daa7c04131f5:
```
sschuberth, line 324: Typo in commit message title: "pseudeo"

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -322,7 +322,12 @@ private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>
 
 // See https://golang.org/ref/mod#pseudo-versions.
 // Format for no known base version, e.g. v0.0.0-20191109021931-daa7c04131f5:
-private val PSEUDO_VERSION_REGEX = "^v0\\.0\\.0-[\\d]{14}-(?<sha1>[0-9a-f]+)$".toRegex()
+private val PSEUDO_VERSION_REGEXES = listOf(
+    // no known base version, e.g. v0.0.0-20191109021931-daa7c04131f5.
```
sschuberth, line 326: Shouldn't the comment from line 324 now be dropped? And these comments start with "Format for"?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -322,7 +322,12 @@ private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>
...
+    "^v0.0.0-[\\d]{14}-(?<sha1>[0-9a-f]+)$".toRegex(),
```
sschuberth, line 327: You seem to have dropped the proper `.`-escaping here again.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -327,6 +327,8 @@ private val PSEUDO_VERSION_REGEXES = listOf(
     "^v0.0.0-[\\d]{14}-(?<sha1>[0-9a-f]+)$".toRegex(),
     // base version is a release version, e.g. v1.2.4-0.20191109021931-daa7c04131f5.
     "^v[0-9]+\\.[0-9]+\\.[0-9]+-0\\.[\\d]{14}-(?<sha1>[0-9a-f]+)$".toRegex(),
+    // base version is a pre-release version, e.g. v1.2.4-pre.0.20191109021931-daa7c04131f5.
+    "^v[0-9]+\\.[0-9]+\\.[0-9]+-pre.0\\.[\\d]{14}-(?<sha1>[0-9a-f]+)$".toRegex(),
```
sschuberth, line 331: Can't we simply make the `pre.` part optional in the expression above?

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Yes we can. I however prefer it as-is, because to me it is easier to understand and also nice to have one separate entry per format in the spec. Ok with keeping it as-is?

sschuberth: How about then at least extracting some repeating parts like `[\\d]{14}-(?<sha1>[0-9a-f]+)$` to a common variable?

fviernau: Done in a separate commit on top.

sschuberth (approved):
I'm ok with the changes as they improve the current situation. However, as the second regex matches everything that the first matches, and the second regex could also easily be made to match the third one, I still believe we should consolidate these regexes into one. That is, unless we at some point need to now *which* regex matched.

fviernau:
> I'm ok with the changes as they improve the current situation. However, as the second regex matches everything that the first matches, and the second regex could also easily be made to match the third one, I still believe we should consolidate these regexes into one. That is, unless we at some point need to now _which_ regex matched.

I believe that "However, as the second regex matches everything that the first matches" does not hold.

sschuberth:
> I believe that "However, as the second regex matches everything that the first matches" does not hold.

Ah, right, I missed the additional "-0".

PR #4837, opened 2021-12-10
Gomod: Reduce over-reporting of packages

fviernau:
The analysis result may contain packages which are not referenced from the dependency tree.
Remove these package as they are not needed and find the rationale for it in the commit.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -155,7 +155,7 @@ class GoMod(
                 version = entry.substringAfter('@', "")
             )
 
-        val graph = Graph()
+        var graph = Graph()
 
         for (line in run("mod", "graph", workingDir = projectDir).requireSuccess().stdout.lines()) {
             if (line.isBlank()) continue
@@ -169,7 +169,14 @@ class GoMod(
             graph.addEdge(parent, child)
         }
 
-        return graph.subgraph(getUsedPackages(graph, projectDir, graph.projectId()))
+        val usedPackages = getUsedPackages(graph, projectDir, graph.projectId())
+        if (usedPackages.size < graph.size()) {
+            log.debug { "Removing ${graph.size() - usedPackages.size} unused packages from the dependency graph." }
+
+            graph = graph.subgraph(usedPackages)
+        }
+
+        return graph
     }
 
     /**
@@ -185,12 +192,7 @@ class GoMod(
                 parseWhyOutput(run(projectDir, "mod", "why", *pkgNames).requireSuccess().stdout)
         }
 
-        val usedPackages = graph.nodes().filter { it.name in usedPackageNames }
-        if (usedPackages.size < graph.size()) {
-            log.debug { "Removing ${graph.size() - usedPackages.size} unused packages from the dependency graph." }
-        }
-
-        return usedPackages.toSet()
+        return graph.nodes().filterTo(mutableSetOf()) { it.name in usedPackageNames }
     }
 
     private fun createPackage(id: Identifier): Package {
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -149,8 +149,6 @@ class GoMod(
             .toSet()
 
     private fun getDependencyGraph(projectDir: File): Graph {
-        val graph = run("mod", "graph", workingDir = projectDir).requireSuccess().stdout
-
         fun parsePackageEntry(entry: String) =
             Identifier(
                 type = managerName,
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -91,10 +91,8 @@ class GoMod(
         val projectDir = definitionFile.parentFile
 
         stashDirectories(projectDir.resolve("vendor")).use {
-            val fullGraph = getDependencyGraph(projectDir)
-            val projectId = fullGraph.projectId()
-
-            val graph = fullGraph.subgraph(getUsedPackages(fullGraph, projectDir, projectId))
+            val graph = getDependencyGraph(projectDir)
```
mnonnenmacher, line 94: Commit message:
* Typo in "repor~~i~~ting"
* I don't get the second sentence, do you mean something like "...returning unused dependencies and leaving their removal to the caller"?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -171,6 +169,8 @@ class GoMod(
             graph.addEdge(parent, child)
         }
 
+        graph = graph.subgraph(getUsedPackages(graph, projectDir, graph.projectId()))
```
mnonnenmacher: `return graph.subgraph(...)`, then there is no need to make it a `var`.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -169,7 +169,12 @@ class GoMod(
             graph.addEdge(parent, child)
         }
 
-        graph = graph.subgraph(getUsedPackages(graph, projectDir, graph.projectId()))
+        val usedPackages = getUsedPackages(graph, projectDir, graph.projectId())
+        if (usedPackages.size < graph.size()) {
+            log.debug { "Removing ${graph.size() - usedPackages.size} unused packages from the dependency graph." }
```
mnonnenmacher, line 174: "Removed" would be better because it already happened.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -176,6 +176,15 @@ class GoMod(
             graph = graph.subgraph(usedPackages)
         }
 
+        val referencedPackages = graph.getTransitiveDependencies(graph.projectId())
```
mnonnenmacher, line 179: To me it seems that this is a workaround for a bug in `subgraph()` and should have been fixed in the `subgraph()` function instead.

It's unfortunate that none of the GoMod tests reflect this fix, this would help to prevent regressions. If you know a case where this fix helps, would you mind adding it to the test project in a previous commit?

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Why do you think it's a bug in subgraph?
(I believe it isn't because it creates an overlay graph containing only the given set of nodes).

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Out of scope, so done it in a separate commit on top.

analyzer/src/main/kotlin/managers/GoMod.kt
mnonnenmacher: How I understand the commit message is that if I have the graph A -> B -> C and call `subgraph(A, C)` it returns two loose graphs `A` and `C` even though it knows that `A` is the root node and `C` is not reachable from there anymore. That sounds very much like a bug to me. Btw, there is also a call to `subgraph` in line 110, doesn't it suffer from the same issue?

fviernau: > How I understand the commit message is that if I have the graph A -> B -> C and call `subgraph(A, C)` it returns two loose graphs `A` and `C` even though it knows that `A` is the root node and `C` is not reachable from there anymore. That sounds very much like a bug to me

To the best of my knowledge a graph doesn't have the concept of a root  node, so I cannot make sense of this argumentation.

> Btw, there is also a call to subgraph in line 110, doesn't it suffer from the same issue?

I don't see how this could suffer from the same problem. Maybe I'm missing something?

mnonnenmacher: > To the best of my knowledge a graph doesn't have the concept of a root node, so I cannot make sense of this argumentation.

I think you can because you mention the concept of a root in your own commit message.

But after all, wouldn't it have been a much simpler solution to the issue to just call `collectDependencies()´ on the "all" scope to get the package list?

Anyway, I will approve this to unblock the fix.

mnonnenmacher (approved):

PR #4840, opened 2021-12-13
declared-license-mapping: Add EUPL-1-1 mapping

mmurto:
Found "European Union Public Licence (EUPL v.1.1)" in [1].

[1]: https://github.com/felleslosninger/efm-commons-libs/blob/master/commons-bdx/pom.xml#L41


fviernau (changes requested):
Please use a perma link in the commit messsage.

fviernau (approved):

PR #4843, opened 2021-12-14
fossid-webapp: Make Scan.id an integer

nnobelis:
Historically in ORT, the id was always a free nullable String even if
the FossID server in fact always returned the id as a non-nullable
integer.

This led to a subtle bug: when creating new delta scan,
FossID.recentScansForRepository was listing all the scans and sorting
them with .sortedByDescending { scan.id }, i.e. alphanumerically.
We didn't notice this behaviour because our scan ids were between 1000
and 9999 for a long time. As soon as the server went over 10000, the
lexicographic sorting order was wrong and led to cases where older scans
were used for delta scans basis, and also newer scans were automatically
deleted by ORT scanner.

### DIFF
--- a/scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
+++ b/scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
@@ -556,15 +556,20 @@ class FossIdTest : WordSpec({
             val scanCode = scanCode(PROJECT, FossId.DeltaTag.DELTA)
             val config = createConfig(deltaScanLimit = deltaScanLimit)
             val vcsInfo = createVcsInfo()
-            val originScan = createScan(vcsInfo.url, vcsInfo.revision, originCode, scanId = "${SCAN_ID}0")
-            val otherScan1 = createScan(vcsInfo.url, "${vcsInfo.revision}_other", "anotherCode")
-            val otherScan2 = createScan("someURL", "someRevision", "someCode", scanId = "zzz")
+            val originScan = createScan(vcsInfo.url, vcsInfo.revision, originCode, scanId = SCAN_ID)
+            val otherScan1 = createScan(
+                vcsInfo.url,
+                "${vcsInfo.revision}_other",
+                "anotherCode",
+                scanId = 0
+            )
+            val otherScan2 = createScan("someURL", "someRevision", "someCode", scanId = 999)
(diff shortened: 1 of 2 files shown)


scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -887,7 +887,7 @@ private fun FossIdServiceWithVersion.expectCreateScan(
 ): FossIdServiceWithVersion {
     coEvery {
         createScan(USER, API_KEY, projectCode, scanCode, vcsInfo.url, vcsInfo.revision)
-    } returns MapResponseBody(status = 1, data = mapOf("scan_id" to SCAN_ID))
+    } returns MapResponseBody(status = 1, data = mapOf("scan_id" to "$SCAN_ID"))
```
sschuberth, line 890: Nit: Maybe `SCAN_ID.toString()` would be nicer?

sschuberth (approved):

sschuberth (dismissed):

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -556,15 +556,15 @@ class FossIdTest : WordSpec({
             val scanCode = scanCode(PROJECT, FossId.DeltaTag.DELTA)
             val config = createConfig(deltaScanLimit = deltaScanLimit)
             val vcsInfo = createVcsInfo()
-            val originScan = createScan(vcsInfo.url, vcsInfo.revision, originCode, scanId = "${SCAN_ID}0")
+            val originScan = createScan(vcsInfo.url, vcsInfo.revision, originCode, scanId = SCAN_ID)
```
oheger-bosch, line 559: This slightly changes the logic in the test, since `originScan` and `otherScan1` now have the same `scanId`. Not sure whether this is a problem, but this change was probably not intendet.

nnobelis: This is a valid point.

oheger-bosch (approved):

sschuberth:
Merging despite the unrelated Python test failure which will get fixed separately by https://github.com/oss-review-toolkit/ort/pull/4848.

PR #4852, opened 2021-12-14
Advisor: Skip execution if there are no packages in the ORT result

oheger-bosch:
It does not make sense to invoke the advice providers if there are no
packages they can operate on. Therefore, return early in this case.

### DIFF
--- a/advisor/src/main/kotlin/Advisor.kt
+++ b/advisor/src/main/kotlin/Advisor.kt
@@ -65,21 +65,25 @@ class Advisor(
             return ortResult
         }
 
-        val providers = providerFactories.map { it.create(config) }
-
         val results = sortedMapOf<Identifier, List<AdvisorResult>>()
 
         val packages = ortResult.getPackages(skipExcluded).map { it.pkg }
+        if (packages.isEmpty()) {
+            log.info { "There are no packages to give advice for." }
+        } else {
 
-        runBlocking {
-            providers.map { provider ->
-                async {
-                    provider.retrievePackageFindings(packages)
-                }
-            }.forEach { providerResults ->
-                providerResults.await().forEach { (pkg, advisorResults) ->
-                    results.merge(pkg.id, advisorResults) { oldResults, newResults ->
-                        oldResults + newResults
+            val providers = providerFactories.map { it.create(config) }
+
+            runBlocking {
+                providers.map { provider ->
+                    async {
+                        provider.retrievePackageFindings(packages)
+                    }
+                }.forEach { providerResults ->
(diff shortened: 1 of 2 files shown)


advisor/src/main/kotlin/Advisor.kt
```diff
@@ -65,12 +65,16 @@ class Advisor(
             return ortResult
         }
 
+        val packages = ortResult.getPackages(skipExcluded).map { it.pkg }
+        if (packages.isEmpty()) {
+            log.info { "Skipping advisor run, as there are no packages to advise." }
```
sschuberth, line 70: Nits:
- I'd drop the common.
- Maybe better: "...no packages to give advice for."

sschuberth (changes requested):

advisor/src/main/kotlin/Advisor.kt
```diff
@@ -65,12 +65,16 @@ class Advisor(
...
+            return ortResult
```
mnonnenmacher, line 71: I would prefer if it would still add an `AdvisorRun` to show that the advisor tool was executed. Otherwise it might give the false impression in reports that no advisor was executed.

oheger-bosch: Good point. I changed the implementation to produce an empty result if there are no packages.

advisor/src/main/kotlin/Advisor.kt
```diff
@@ -65,21 +65,25 @@ class Advisor(
             return ortResult
         }
 
-        val providers = providerFactories.map { it.create(config) }
-
         val results = sortedMapOf<Identifier, List<AdvisorResult>>()
 
         val packages = ortResult.getPackages(skipExcluded).map { it.pkg }
+        if (packages.isEmpty()) {
+            log.info { "Skipping advisor run as there are no packages to give advice for." }
```
sschuberth, line 72: Is "skipping" now still the right term here? Because an `AdvisorRun` still gets added to the result. Maybe just simply "There are no packages to give advice for."

advisor/src/test/kotlin/AdvisorTest.kt
```diff
@@ -56,6 +57,23 @@ class AdvisorTest : WordSpec({
             }
         }
 
+        "return an ORT result with a dummy advisor result if there are no packages" {
```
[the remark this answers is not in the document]
oheger-bosch: I guess, in this case both terms would be okay. Changed as proposed.

sschuberth (approved):

PR #4854, opened 2021-12-14

mnonnenmacher:

### DIFF
--- a/cli/src/main/kotlin/commands/UploadResultToPostgresCommand.kt
+++ b/cli/src/main/kotlin/commands/UploadResultToPostgresCommand.kt
@@ -86,7 +86,17 @@ class UploadResultToPostgresCommand : CliktCommand(
         val ortResult = readOrtResult(ortFile)
 
         val postgresConfig = globalOptionsForSubcommands.config.scanner.storages?.values
-            ?.filterIsInstance<PostgresStorageConfiguration>()?.singleOrNull()
+            ?.filterIsInstance<PostgresStorageConfiguration>()?.let { configs ->
+                if (configs.size > 1) {
+                    val config = configs.first()
+                    println(
+                        "Multiple PostgreSQL storages are configured, using the first one which points to schema " +
+                                "${config.schema} at ${config.url}."
+                    )
+                }
+
+                configs.firstOrNull()
+            }
 
         requireNotNull(postgresConfig) {
             "No PostgreSQL storage is configured for the scanner."
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -347,7 +347,7 @@ class ExperimentalScanner(
         return scannerWrappers.associateWith { scanner ->
             val scannerCriteria = scanner.criteria ?: return@associateWith mutableMapOf()
 
-            val result = mutableMapOf<KnownProvenance, List<ScanResult>>()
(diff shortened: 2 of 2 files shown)


scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -392,7 +392,7 @@ class ExperimentalScanner(
                 }
             }
 
-            result
+            result.mapValuesTo(mutableMapOf()) { it.value.toList() }
```
sschuberth, line 395: This is quite ugly, but I currently cannot think of a better solution either.

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: Me neither, I tried a few things but nothing worked.

sschuberth (approved):

cli/src/main/kotlin/commands/UploadResultToPostgresCommand.kt
```diff
@@ -86,7 +86,17 @@ class UploadResultToPostgresCommand : CliktCommand(
         val ortResult = readOrtResult(ortFile)
 
         val postgresConfig = globalOptionsForSubcommands.config.scanner.storages?.values
-            ?.filterIsInstance<PostgresStorageConfiguration>()?.singleOrNull()
+            ?.filterIsInstance<PostgresStorageConfiguration>()?.let { configs ->
+                if (configs.size > 1) {
+                    val config = configs.first()
```
oheger-bosch, line 91: I feel a bit uneasy with this change. Selecting the first storage configuration seems rather arbitrary. The use case to have multiple database connections configured in the scanner options does not seem to be really supported currently. So, isn't the current behavior, to fail if there is ambiguity, safer than using a connection that might not be intended?

For a long-term solution, it may be necessary to refactor the way external connections are declared in `ort.conf`; e.g. have a separate section for this and allow assigning them a name. This would probably make the configuration clearer. I remember, in the past we had used SW360 for different purposes, and the connection had to be defined at multiple places.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
oheger-bosch: To get rid of this, would it be possible to generate the result map in a more functional way, e.g. something in the lines of:

```
provenances.associateWith { provenance ->
    storageReaders.filterIsInstance(...)
      .flatMap { reader ->
         reader.read(...)
      }
}
```

cli/src/main/kotlin/commands/UploadResultToPostgresCommand.kt
mnonnenmacher: I agree that it's not a nice solution because it's arbitrary but I need a quick fix for this without refactoring the configuration. After all, if the first connection is not the one a user wants they could always create a separate configuration file for the command.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: The problem is that there are two different types of storage readers, combining both in a single loop would be very difficult.

oheger-bosch: Thanks, no objections from my side.

PR #4855, opened 2021-12-14
docs: Miscellaneous minor grammar fixes

fviernau:


docs/getting-started.md
```diff
@@ -8,7 +8,7 @@ ORT:
 * Analyze the dependencies of `mime-types` using the _analyzer_.
 * Scan the source code of `mime-types` and its dependencies using the `scanner`.
 * Run the evaluator to find any rule violations.
-* Generate reports to show the results.
+* Generate reports showing the results.
```
sschuberth, line 11: Grammar-wise I don't think it's actually *wrong*... it's probably just a bit unusual. Maybe it was done that way to resemble the "to find" one sentence above.

docs/getting-started.md
```diff
@@ -27,7 +27,7 @@ For the full list of supported package managers and Version Control Systems see
 
 ## 2. Download & Install ORT
 
-In future we will provide binaries of the ORT tools, but currently you have to build the tools on your own. First
+In future, we will provide binaries of the ORT tools, but currently you have to build the tools on your own. First
```
sschuberth, line 30: While at it, maybe yet better "In *the* future, ..."? Also, we usually use passive voice in docs and should not talk about "we" at all. Thinking about it, I'd actually drop the first sentence completely and start saying "You currently have to build the tools on your own". However, that's also not quite correct, as you could use the CLI fat JAR from JitPack...

docs/getting-started.md
fviernau: I dropped the commit.

sschuberth (approved):

fviernau:
Merge even though the liftBot hangs. The other checks went ok.

PR #4857, opened 2021-12-15
Clean up on the local vs remote scanner topic

sschuberth:


nnobelis:
I am fine with the changes. I find `PathScanner` to be a tad too generic....but I see no better alternatives !

sschuberth:
> I am fine with the changes. I find `PathScanner` to be a tad too generic....but I see no better alternatives !

I guess this name will make more sense in the context of the upcoming `PackageScannerWrapper` and `ProvenanceScannerWrapper` renamings 😉 Stay tuned!

oheger-bosch:
Well, I also do not like the new name `PathScanner` that much. IMHO, this is misleading as well, as it sounds as if you point the scanner to a path containing the sources and execute it there. But this is not the case, the scanner is even responsible for downloading source code (to a temporary path) before it can be processed.

I think, what makes these types of scanners special is the fact that they wrap a command line tool installed on the local machine, and the base class provides some functionality to manage this tool. This is in contrast to the FossID scanner, which interacts with a remote service.

sschuberth:
> Well, I also do not like the new name `PathScanner` that much. IMHO, this is misleading as well, as it sounds as if you point the scanner to a path containing the sources and execute it there.

That's actually exactly what's supported: If you call the `scanPath()` function, any *existing* source code in the given directory is scanned. The download only happens if you call `scanPackage()`.

> I think, what makes these types of scanners special is the fact that they wrap a command line tool installed on the local machine

Unfortunately that's not correct: E.g. the upcoming SCANOSS scanner would also be a "LocalScanner" / "PathScanner" as it requires access to the source code in order to calculate the fingerprints. But it's executed via an API to which we sent the fingerprints.

oheger-bosch:
> > I think, what makes these types of scanners special is the fact that they wrap a command line tool installed on the local machine
> 
> Unfortunately that's not correct: E.g. the upcoming SCANOSS scanner would also be a "LocalScanner" / "PathScanner" as it requires access to the source code in order to calculate the fingerprints. But it's executed via an API to which we sent the fingerprints.

So would the SCANOSS implementation extend "PathScanner"? Does it use a tool to calculate the fingerprints?

sschuberth:
> So would the SCANOSS implementation extend "PathScanner"?

Yes. That's partly why this renaming should happen, as it would be confusing to call SCANOSS a "LocalScanner".

> Does it use a tool to calculate the fingerprints?

Not a CLI tool, no. We have a Java implementation for that, and we'll eventually have a ported Kotlin implementation. So no external tools would be called from a SCANOSS scanner implementation.

oheger-bosch:
> > So would the SCANOSS implementation extend "PathScanner"?
> 
> Yes. That's partly why this renaming should happen, as it would be confusing to call SCANOSS a "LocalScanner".
> 
> > Does it use a tool to calculate the fingerprints?
> 
> Not a CLI tool, no. We have a Java implementation for that, and we'll eventually have a ported Kotlin implementation. So no external tools would be called from a SCANOSS scanner implementation.

I am trying to understand the current hierarchy and why certain scanner implementations extend specific base classes or not.

So by extending the "local scanner" base class, SCANOSS would also implement the `CommandLineTool` interface, but actually there is no underlying tool. I assume, from the base class, it would use the functionality to download source code, so that it can be processed locally. Then it's this functionality what is special on these scanners?

The FossID scanner does not need access to the source code, since it is loaded by FossID itself. However, there was the idea to upload a zipped archive to FossID instead. If this was implemented, would the FossID scanner then also be considered a local scanner?

sschuberth:
> I am trying to understand the current hierarchy

Now that you name it, I believe the hierarchy to still be a bit broken even after this PR...

> So by extending the "local scanner" base class, SCANOSS would also implement the CommandLineTool interface

That would currently be the case, yes, but of course it shouldn't. We probably should stop `PathScanner` to implement `CommandLineTool` and instead have a new `CommandLineScanner` that inherits from `PathScanner` and implements `CommandLineTool`. What do you think?

> However, there was the idea to upload a zipped archive to FossID instead.

Not "instead", but as an alternate mode of operation. That way of passing code to FossID would probably be implemented as its own scanner class. And that class would then be a `PathScanner`, yes.

nnobelis:
> That would currently be the case, yes, but of course it shouldn't. We probably should stop `PathScanner` to implement `CommandLineTool` and instead have a new `CommandLineScanner` that inherits from `PathScanner` and implements `CommandLineTool`. What do you think?

This looks like the right solution. And ScanOSS would inherit `PathScanner`,
Good finding @oheger-bosch !

sschuberth:
Ok, I'll make the agreed changes in a follow-up PR. But for now I'd like to merge this, as the follow-up PR has already grown quite large. Approvals, anyone?

oheger-bosch (approved):
Ok, I think we had a good discussion :-)

sschuberth:
> > I am fine with the changes. I find `PathScanner` to be a tad too generic....but I see no better alternatives !
> 
> I guess this name will make more sense in the context of the upcoming `PackageScannerWrapper` and `ProvenanceScannerWrapper` renamings 😉 Stay tuned!

FYI, see @mnonnenmacher's https://github.com/oss-review-toolkit/ort/pull/4862.

sschuberth:
> Ok, I'll make the agreed changes in a follow-up PR. But for now I'd like to merge this, as the follow-up PR has already grown quite large. Approvals, anyone?

See https://github.com/oss-review-toolkit/ort/pull/4871.

PR #4859, opened 2021-12-15
Gradle: Use the most recent file in case of multiple cache matches

sschuberth:
The Gradle cache might contain the "same" artifact multiple times, that
is, different files for the same Maven coordinates, deduplicated by the
cache path which contains the file hash. This can happen if an artifact
was silently updated on a Maven server, or if different Maven servers
serve different files for the same artifact coordinates and servers were
switched in a Gradle project. Instead of choosing the first artifact the
directory walk comes across, deterministically use the file with the
most recent modification date.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Gradle.kt
+++ b/analyzer/src/main/kotlin/managers/Gradle.kt
@@ -99,17 +99,24 @@ class Gradle(
                 "${artifact.groupId}/${artifact.artifactId}/${artifact.version}"
             )
 
-            val artifactFile = artifactRootDir.walk().find {
+            val artifactFiles = artifactRootDir.walk().filter {
                 val classifier = if (artifact.classifier.isNullOrBlank()) "" else "${artifact.classifier}-"
                 it.isFile && it.name == "${artifact.artifactId}-$classifier${artifact.version}.${artifact.extension}"
+            }.sortedByDescending {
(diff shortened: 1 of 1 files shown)


sschuberth:
Fun fact screenshots:

image

MarcelBochtler (approved):

analyzer/src/main/kotlin/managers/Gradle.kt
```diff
@@ -99,17 +99,22 @@ class Gradle(
                 "${artifact.groupId}/${artifact.artifactId}/${artifact.version}"
             )
 
-            val artifactFile = artifactRootDir.walk().find {
+            val artifactFiles = artifactRootDir.walk().filter {
```
mnonnenmacher, line 102: Could you add to the commit message or as a code comment why the most recent edited file is chosen / is better than an arbitrary one? I assume it is because it is most likely the one downloaded during the current dependency resolution?

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/Gradle.kt
sschuberth: I've added a code comment.

mnonnenmacher (approved):

sschuberth:

Issue #4519, opened 2021-09-30
Allow to curate copyrights directly (not via authors)


sschuberth:
> adding the Copyright holder statements to the org.ossreviewtoolkit.model.PackageCurationData entity

FYI, I just pushed my local [`curate-copyrights` branch](https://github.com/oss-review-toolkit/ort/tree/curate-copyrights) which trivially starts doing that.

@rbieniek do you want to take over that branch of mine?

_Originally posted by @sschuberth in https://github.com/oss-review-toolkit/ort/issues/4463#issuecomment-931138189_

sschuberth:
> @rbieniek do you want to take over that branch of mine?

Ping @porsche-rbieniek and @porsche-rishisaxena as a reminder to move this forward.

porsche-rbieniek:
Porsche solution submitted as https://github.com/oss-review-toolkit/ort/pull/5315

sschuberth:
> Porsche solution submitted as #5315

Please associate issues with PRs by using one of the respective [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) in one of the commits in the PR instead of manually adding comments.

tsteenbe:
Copy-paste from [July 14th, 2022 ORT developer meeting minutes](https://github.com/oss-review-toolkit/ort/wiki/Developer-Meeting#2022-07-14) in which we had a discussion on how to implement curating declared copyrights 

**A. Use curations to fix-up curate copyrights**

The declared copyrights comes package metadata collected by ORT _analyzer_ which can be fixed up using curations.
Note that a lot of package managers do not have copyrights fields only author, contributors, developers which in some case are "mis-used" to convey copyright information. We could implement a parseCopyrights() to from parseAuthors() find copyright statements e.g. entries starting with "copyright" or "(c)"

Below several ideas for how curating declared licenses in curations.yml could look like.

Remove a declared copyright

```
declared_copyright_mapping:
   "MIT": ""
```

Add a declared copyright

```
declared_copyright_mapping:
   "": "Copyright (C) 2022 John Doe"
```

Overwrite all declared copyrights with a single one:

```
declared_copyright_mapping:
   "*": ""
   "": "Copyright (C) 2022 John Doe"
```

Overwrite all declared copyrights with a more than one:

```
declared_copyright_mapping:
   "*": ""
[...]

PR #4860, opened 2021-12-15
ExperimentalScanner: Support skip options

mnonnenmacher:
Support the optional skipping of excluded packages and packages with concluded license.

### DIFF
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -86,7 +86,7 @@ class ExperimentalScanner(
         }
 
         val packageResults = if (packageScannerWrappers.isNotEmpty()) {
-            val packages = ortResult.getPackages().mapTo(mutableSetOf()) { it.pkg }
+            val packages = ortResult.getPackages().map { it.pkg }.filterNotConcluded().toSet()
 
             log.info { "Scanning ${packages.size} packages with ${packageScannerWrappers.size} scanners." }
 
@@ -325,6 +325,16 @@ class ExperimentalScanner(
 
         return nestedProvenanceScanResults
     }
+
+    private fun Collection<Package>.filterNotConcluded(): Collection<Package> =
+        takeUnless { scannerConfig.skipConcluded }
+            ?: partition { it.concludedLicense != null && it.authors.isNotEmpty() }.let { (skip, keep) ->
+                if (skip.isNotEmpty()) {
+                    log.debug { "Not scanning the following packages with concluded licenses: $skip" }
+                }
+
+                keep
+            }
 
     private fun getPackageProvenances(packages: Set<Package>): Map<Package, Provenance> =
         packages.associateWith { pkg ->
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -67,14 +67,14 @@ class ExperimentalScanner(
         }
     }
 
-    suspend fun scan(ortResult: OrtResult): OrtResult {
+    suspend fun scan(ortResult: OrtResult, skipExcluded: Boolean): OrtResult {
         val startTime = Instant.now()
 
         val projectScannerWrappers = scannerWrappers[PackageType.PROJECT].orEmpty()
         val packageScannerWrappers = scannerWrappers[PackageType.PACKAGE].orEmpty()
 
         val projectResults = if (projectScannerWrappers.isNotEmpty()) {
-            val packages = ortResult.getProjects().mapTo(mutableSetOf()) { it.toPackage() }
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -326,6 +326,16 @@ class ExperimentalScanner(
         return nestedProvenanceScanResults
     }
 
+    private fun List<Package>.filterConcluded(): List<Package> =
```
sschuberth, line 329: Nit: The extension function could be on `Collection<Package>`.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -86,7 +86,7 @@ class ExperimentalScanner(
         }
 
         val packageResults = if (packageScannerWrappers.isNotEmpty()) {
-            val packages = ortResult.getPackages().mapTo(mutableSetOf()) { it.pkg }
+            val packages = ortResult.getPackages().map { it.pkg }.filterConcluded().toSet()
```
sschuberth: In Kotlin semantics, `filter { true }` defines what to keep (the "filtrate"), not what to "filter out". With that in mind, `filterConcluded()` sounds as if you'd keep packages with a concluded license, whereas the opposite is the case. So maybe call this `filterNotConcluded()` to resemble `filterNot`?

sschuberth (changes requested):

sschuberth (approved):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -326,6 +326,16 @@ class ExperimentalScanner(
         return nestedProvenanceScanResults
     }
 
+    private fun Collection<Package>.filterNotConcluded(): Collection<Package> =
+        takeUnless { scannerConfig.skipConcluded }
+            ?: partition { it.concludedLicense != null && it.authors.isNotEmpty() }.let { (skip, keep) ->
```
fviernau, line 331: Why did you decide to check for authors?
To me this functionality seems unexpected from the name of the configuration option.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -326,6 +326,16 @@ class ExperimentalScanner(
         return nestedProvenanceScanResults
     }
 
+    private fun Collection<Package>.filterNotConcluded(): Collection<Package> =
```
sschuberth, line 329: Ah, I overlooked this: Input types should be as generic as possible (`Collection`) but output types as specific as possible, so the return type should stay a `List`.

sschuberth (changes requested):

fviernau (changes requested):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
sschuberth: This matches the current implementation in

https://github.com/oss-review-toolkit/ort/blob/8f4e9c18f5fa9f54dd076866351b0c084a411fcb/scanner/src/main/kotlin/Scanner.kt#L47-L56

until we have

https://github.com/oss-review-toolkit/ort/issues/4519

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: That was deliberate, because if I change the input type to `Collection` I would have to call `toList()` on the input if `skipConcluded == false` which would copy the list to a new `ArrayList` for no reason. And it's a private function after all, I think for those it's totally fine to be tailored to the use case of the class, that's why I didn't use `Collection` in the first place.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
fviernau: @mnonnenmacher how about adding to the commit message that "authors" is only checked for consistency with the existing impl?

mnonnenmacher: That wasn't my decision, it's how the option was designed, see:

https://github.com/oss-review-toolkit/ort/blob/0ed3b248ae30f17bdebf330b71fe6357b33b4c12/model/src/main/kotlin/config/ScannerConfiguration.kt#L37-L41

fviernau (approved):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
sschuberth: Ok, in this case I'll leave it to you whether to use `Collection` or `List` in both places. Approving in any case.

sschuberth (approved):

PR #4871, opened 2021-12-16
Introduce a `CommandLineScanner`

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/CommandLineScanner.kt
+++ b/scanner/src/main/kotlin/CommandLineScanner.kt
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2017-2021 HERE Europe B.V.
+ * Copyright (C) 2021 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.scanner
+
+import com.vdurmont.semver4j.Requirement
+
+import java.io.File
+import java.io.IOException
+
+import kotlin.time.measureTimedValue
+
+import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.model.config.ScannerConfiguration
+import org.ossreviewtoolkit.utils.common.CommandLineTool
+import org.ossreviewtoolkit.utils.common.Os
+import org.ossreviewtoolkit.utils.core.log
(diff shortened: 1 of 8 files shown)


scanner/src/main/kotlin/CommandLineScanner.kt
```diff
@@ -0,0 +1,110 @@
...
+import kotlin.time.measureTimedValue
+
+import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.model.config.ScannerConfiguration
+import org.ossreviewtoolkit.utils.common.CommandLineTool
+import org.ossreviewtoolkit.utils.common.Os
+import org.ossreviewtoolkit.utils.core.log
+import org.ossreviewtoolkit.utils.core.perf
+
+abstract class CommandLineScanner(
```
oheger-bosch, line 37: It would be nice to have some documentation here; what makes this base class special in the hierarchy of scanner classes.

scanner/src/main/kotlin/PathScanner.kt
```diff
@@ -182,7 +182,7 @@ abstract class PathScanner(
      */
     open fun getScannerCriteria(): ScannerCriteria {
         val options = scannerConfig.options?.get(scannerName).orEmpty()
-        val minVersion = parseVersion(options[PROP_CRITERIA_MIN_VERSION]) ?: Semver(normalizeVersion(expectedVersion))
+        val minVersion = parseVersion(options[PROP_CRITERIA_MIN_VERSION]) ?: Semver(normalizeVersion(version))
```
oheger-bosch, line 114: Just asking: Can this cause any issues if the local executable cannot be resolved and hence the version is not available? Would then the database already be queried with a wrong version before scanning of local sources fails?

scanner/src/main/kotlin/CommandLineScanner.kt
sschuberth: Done, I simply forgot about this.

scanner/src/main/kotlin/PathScanner.kt
sschuberth: In the worst case an `IOException` could be thrown when querying the version fails if the tool is not available *and* bootstrapping failed. But I guess throwing should be fine in this case, or?

scanner/src/test/kotlin/PathScannerTest.kt
```diff
@@ -100,12 +100,8 @@ private fun createScanner(
     downloaderConfig: DownloaderConfiguration
 ): PathScanner =
     object : PathScanner(SCANNER_NAME, scannerConfig, downloaderConfig) {
+        override val version = SCANNER_VERSION
```
mnonnenmacher, line 103: You could update the copyright statement in this file as well.

scanner/src/main/kotlin/PathScanner.kt
oheger-bosch: It is just because the documentation of the `version` property says: "The actual version of the scanner, or an empty string in case of failure." If the string was empty, could it be the case that `Scanner` already starts processing before a failure is thrown when the external tool is tried to be invoked?

sschuberth: The docs are simply wrong / misleading, as it depends on the *kind* of failure: If the tool could be executed but the version could not be found, and empty string is returned. If the tool could not be executed in the first place, and `IOException` is thrown.

oheger-bosch: Thanks for clarifying.

oheger-bosch (approved):

PR #4876, opened 2021-12-17
ExperimentalScanner: Fix logging issue

mnonnenmacher:
Call the `log` helper on the `ExperimentalScanner` class in extension
functions.


scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -331,7 +331,9 @@ class ExperimentalScanner(
         takeUnless { scannerConfig.skipConcluded }
             ?: partition { it.concludedLicense != null && it.authors.isNotEmpty() }.let { (skip, keep) ->
                 if (skip.isNotEmpty()) {
-                    log.debug { "Not scanning the following packages with concluded licenses: $skip" }
+                    this@ExperimentalScanner.log.debug {
```
sschuberth, line 334: You should be able to drop the `this@` prefix here and below, or?

mnonnenmacher: No, because the `log` extension property requires an instance of the class.

sschuberth (approved):

PR #4885, opened 2021-12-21
Identifier: Allow Maven namespaces without TLD

MarcelBochtler:
While it is a good practice to use the dot notation in the Maven
groupId, it is not required [1]. See [2] for a popular example.

[1]: https://maven.apache.org/pom.html#maven-coordinates
[2]: https://github.com/junit-team/junit4/blob/main/pom.xml

### DIFF
--- a/model/src/main/kotlin/Identifier.kt
+++ b/model/src/main/kotlin/Identifier.kt
@@ -102,7 +102,7 @@ data class Identifier(
             val lowerName = name.lowercase()
             val vendorNamespace = when (type) {
                 "NPM" -> "@$lowerName"
-                "Gradle", "Maven", "SBT" -> "(com|io|net|org)\\.$lowerName(\\..+)?"
+                "Gradle", "Maven", "SBT" -> "((com|io|net|org)\\.)?$lowerName(\\..+)?"
                 else -> ""
             }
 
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/Identifier.kt
```diff
@@ -102,7 +102,7 @@ data class Identifier(
             val lowerName = name.lowercase()
             val vendorNamespace = when (type) {
                 "NPM" -> "@$lowerName"
-                "Gradle", "Maven", "SBT" -> "(com|io|net|org)\\.$lowerName(\\..+)?"
+                "Gradle", "Maven", "SBT" -> "(com|io|net|org)?\\.?$lowerName(\\..+)?"
```
sschuberth, line 105: This would also allow a namespace with a leading dot. I guess that's not allowed in Maven. So you probably need to write this as `((com|io|net|org)\\.)?$lowerName(\\..+)?`

sschuberth (changes requested):

model/src/main/kotlin/Identifier.kt
sschuberth: Also the commit message is a bit misleading, IMO. It's not that the `Identifier` class did not allow namespaces in non-TLD form, it's just that the `isFromOrg` did not work as expected in that case.

sschuberth (approved):

sschuberth (approved):

PR #4887, opened 2021-12-21
Rules scripting

sschuberth:

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,9 +299,9 @@ can be used to populate a directory with template package configuration files.
 
 The file containing any policy rule implementations to be used with the _evaluator_.
 
-| Format | Scope | Default location | Default value |
-| ------ | ----- | ---------------- | ------------- |
-| Kotlin script (DSL) | Evaluator | `$ORT_CONFIG_DIR/rules.kts` | Empty (n/a) |
+| Format | Scope | Default location                | Default value |
+| ------ | ----- |---------------------------------| ------------- |
+| Kotlin script (DSL) | Evaluator | `$ORT_CONFIG_DIR/evaluator.rules.kts` | Empty (n/a) |
 
 # Details on the tools
 
@@ -649,7 +649,7 @@ To enable this provider, pass `-a VulnerableCode` on the command line.
 
 The _evaluator_ is used to perform custom license policy checks on scan results. The rules to check against are
 implemented as scripts (currently Kotlin scripts, with a dedicated DSL, but support for other scripting can be added as
-well. See [rules.kts](./examples/rules.kts) for an example file.
+well. See [example.rules.kts](./examples/example.rules.kts) for an example file.
 
 <a name="reporter">&nbsp;</a>
 
--- a/cli/src/funTest/kotlin/ExamplesFunTest.kt
+++ b/cli/src/funTest/kotlin/ExamplesFunTest.kt
@@ -133,7 +133,7 @@ class ExamplesFunTest : StringSpec() {
             howToFixText shouldContain "Manually verify that the file does not contain any license information."
         }
 
-        "rules.kts can be compiled and executed" {
+        "example.rules.kts can be compiled and executed" {
             val resultFile = File("src/funTest/assets/semver4j-analyzer-result.yml")
             val licenseFile = File("../examples/license-classifications.yml")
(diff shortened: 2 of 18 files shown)


sschuberth:
@oss-review-toolkit/kotlin-devs, while this is still a draft, could some of you maybe confirm that this offers proper syntax hightlighting and auto-completion in IntelliJ for `evaluator/src/main/resources/rules/no_gpl.rules.kts` now (but not for `examples/example.rules.kts`; that is expected as that script is not part of any module)?

README.md
```diff
@@ -297,9 +297,9 @@ can be used to populate a directory with template package configuration files.
 
 The file containing any policy rule implementations to be used with the _evaluator_.
 
-| Format | Scope | Default location | Default value |
-| ------ | ----- | ---------------- | ------------- |
-| Kotlin script (DSL) | Evaluator | `$ORT_CONFIG_DIR/rules.kts` | Empty (n/a) |
+| Format | Scope | Default location                | Default value |
+| ------ | ----- |---------------------------------| ------------- |
+| Kotlin script (DSL) | Evaluator | `$ORT_CONFIG_DIR/ort.rules.kts` | Empty (n/a) |
```
mnonnenmacher, line 302: I have a tendency towards `evaluator.rules.kts` as the default name.

sschuberth: Fine with me, as "ort" is already part of the parent directory, and it opens the possibility to have more (non-evaluator) rules in the future.

mnonnenmacher:
> @oss-review-toolkit/kotlin-devs, while this is still a draft, could some of you maybe confirm that this offers proper syntax hightlighting and auto-completion in IntelliJ for `evaluator/src/main/resources/rules/no_gpl.rules.kts` now (but not for `examples/example.rules.kts`; that is expected as that script is not part of any module)?

Does not work for me. After Gradle sync I got:
"No script runtime was found in the classpath: class 'kotlin.script.templates.standard.ScriptTemplateWithArgs' not found. Please add kotlin-script-runtime.jar to the module dependencies."

If I apply the suggested fix it adds `implementation(kotlin("script-runtime"))` to the evaluator module but then I get compilation errors like `Unresolved reference: PackageRule`.

sschuberth:
> Does not work for me.

Could you additionally try to build from the command line once, then do a Gradle sync from the IDE, and then open the script file again?

> If I apply the suggested fix it adds `implementation(kotlin("script-runtime"))` to the evaluator module but then I get compilation errors like `Unresolved reference: PackageRule`.

That's an odd suggestion... none of the examples at https://github.com/Kotlin/kotlin-script-examples has that dependency.

MarcelBochtler:
> Could you additionally try to build from the command line once (I usually do this via ./gradlew :cli:run --args="--version"), then do a Gradle sync from the IDE, and then open the script file again?

Works very well for me after building it from command line.

mnonnenmacher:
> > Could you additionally try to build from the command line once (I usually do this via ./gradlew :cli:run --args="--version"), then do a Gradle sync from the IDE, and then open the script file again?
> 
> Works very well for me after building it from command line.

Same here, very nice.

sschuberth:
I'd like to move `examples.rules.kts` to a Kotlin module so that completion works. What do you guys think about simply replacing the overly simplistic `no_gpl.rules.kts` with `examples.rules.kts` as a resource? Should we then symlink this to the `examples` directory, even if that might break things for Windows users?

mnonnenmacher:
> I'd like to move `examples.rules.kts` to a Kotlin module so that completion works. What do you guys think about simply replacing the overly simplistic `no_gpl.rules.kts` with `examples.rules.kts` as a resource? Should we then symlink this to the `examples` directory, even if that might break things for Windows users?

Does this solution only work within this Gradle project, or is it possible to create a minimal Gradle project that depends on the evaluator to have working completion? If it is possible I would prefer such a project and put it in the examples folder.
And I would not like a symlink, I would rather have a file that contains something like "// See: path/to/evalutor.rules.kts".

sschuberth:
> is it possible to create a minimal Gradle project that depends on the evaluator to have working completion?

It should indeed be possible to create a "rules-only" project by depending on the `scripting-utils` artifact, and I like the idea to create such an example. However, I'll do that in a separate PR then.

mnonnenmacher (approved):

PR #4893, opened 2021-12-22
Scanner: Rename `filterOptionsForResult()` to `filterSecretOptions()`

sschuberth:
This is to match the `ScannerWrapper`s function name so both can be
implemented by the same code. While at it, also align function docs.

This also requires to remove the `ScannerWrapper`s default
implementation in the interface, as otherwise it would clash with the
`Scanner`s default implementation.

### DIFF
--- a/scanner/src/test/kotlin/experimental/ExperimentalScannerTest.kt
+++ b/scanner/src/test/kotlin/experimental/ExperimentalScannerTest.kt
@@ -56,6 +56,7 @@ import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
+import org.ossreviewtoolkit.model.config.ScannerOptions
 import org.ossreviewtoolkit.model.yamlMapper
 import org.ossreviewtoolkit.scanner.ScannerCriteria
 import org.ossreviewtoolkit.utils.core.createOrtTempDir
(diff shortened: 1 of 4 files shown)


scanner/src/test/kotlin/experimental/ExperimentalScannerTest.kt
```diff
@@ -636,7 +639,7 @@ private class FakePackageScannerWrapper(
 ) : PackageScannerWrapper {
     override val details = ScannerDetails(name, "1.0.0", "config")
     override val name = details.name
-    override val criteria: ScannerCriteria? = ScannerCriteria.forDetails(details)
+    override val criteria = ScannerCriteria.forDetails(details)
```
mnonnenmacher, line 642: I find the type declaration a lot less annoying than the cast and suppress statement in the test code.

sschuberth: Yeah, at first I thought it'd just work without the cast or suppress statement, but then I had to add the cast, and then also the suppress statement. While the code proposed here does not trigger any inspection hints anymore, I agree it's a bit clumsy.

I just tested and I can suppress the original hint instead. Will do that.

scanner/src/test/kotlin/experimental/ExperimentalScannerTest.kt
```diff
@@ -629,6 +629,7 @@ class ExperimentalScannerTest : WordSpec({
 /**
  * An implementation of [PackageScannerWrapper] that creates empty scan results.
  */
+@Suppress("RedundantNullableReturnType")
```
fviernau, line 632: Commit-msg: Maybe turn "here" -> "there"

fviernau (approved):

fviernau (approved):

PR #4897, opened 2021-12-29
CommandLineScanner: Do not require the expected version

mnonnenmacher:
Do not override `getVersionRequirement()` and instead use the default
implementation which allows any version.

Since 9af81d8 any version of the command line scanners is accepted, the
expected version is only used to bootstrap a scanner if it is not found
on the path. Therefore returning the expected version as a requirement
was wrong.

### DIFF
--- a/scanner/src/main/kotlin/CommandLineScanner.kt
+++ b/scanner/src/main/kotlin/CommandLineScanner.kt
@@ -20,8 +20,6 @@
 
 package org.ossreviewtoolkit.scanner
 
-import com.vdurmont.semver4j.Requirement
-
 import java.io.File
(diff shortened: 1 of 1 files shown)


sschuberth (approved):

maxhbr (approved):
This solved my issue (https://ort-talk.slack.com/archives/C9NNJ54B1/p1640710709193800)

mnonnenmacher:
Merged as test failures were unrelated.

mnonnenmacher:
> This solved my issue (https://ort-talk.slack.com/archives/C9NNJ54B1/p1640710709193800)

Thanks for verifying.

PR #4912, opened 2022-01-04
Gradle: Target the latest version supported by the current JVM

sschuberth:
To benefit from bytecode optimizations, always use the latest supported
JVM target. Note that at a minimum still Java 11 is required.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -238,7 +238,7 @@ subprojects {
 
         kotlinOptions {
             allWarningsAsErrors = true
-            jvmTarget = "11"
+            jvmTarget = javaVersion.majorVersion
             apiVersion = "1.6"
             freeCompilerArgs = freeCompilerArgs + customCompilerArgs
         }


build.gradle.kts
```diff
@@ -238,7 +238,7 @@ subprojects {
 
         kotlinOptions {
             allWarningsAsErrors = true
-            jvmTarget = "11"
+            jvmTarget = javaVersion.majorVersion
```
mnonnenmacher, line 241: That would mean that the JitPack packages would be published with the bytecode level of whatever version of the JVM JitPack is running, right?

sschuberth: Correct, but as I've [configured JitPack to use `openjdk11`](https://github.com/oss-review-toolkit/ort/blob/16fc1c2428b15db32f8b4a0037700a2a5d95c9b2/.jitpack.yml#L2), it would still be Java 11 bytecode.

mnonnenmacher (approved):

PR #4918, opened 2022-01-05
Simplify custom Kotest matchers

mnonnenmacher:

### DIFF
--- a/model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
+++ b/model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
@@ -19,13 +19,14 @@
 
 package org.ossreviewtoolkit.model.licenses
 
-import io.kotest.assertions.print.print
 import io.kotest.assertions.withClue
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.Matcher
-import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.be
 import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.collections.contain
 import io.kotest.matchers.collections.containExactly
+import io.kotest.matchers.collections.containExactlyInAnyOrder
 import io.kotest.matchers.collections.haveSize
 import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
 import io.kotest.matchers.neverNullMatcher
@@ -68,6 +69,8 @@ import org.ossreviewtoolkit.utils.spdx.getLicenseText
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 import org.ossreviewtoolkit.utils.test.createDefault
 import org.ossreviewtoolkit.utils.test.shouldNotBeNull
+import org.ossreviewtoolkit.utils.test.transformingCollectionEmptyMatcher
+import org.ossreviewtoolkit.utils.test.transformingCollectionMatcher
 
 class LicenseInfoResolverTest : WordSpec() {
     init {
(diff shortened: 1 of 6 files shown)


utils/test/src/main/kotlin/Utils.kt
```diff
@@ -99,3 +100,13 @@ fun patchActualResultObject(result: OrtResult, patchStartAndEndTime: Boolean = f
...
+
+/**
+ * A helper function to create a custom matcher that compares an [expected] collection to a collection obtained by
+ * [transform] using the provided [matcher].
+ */
+fun <T, U> transformingCollectionMatcher(
+    expected: Collection<U>,
+    matcher: (Collection<U>) -> Matcher<Collection<U>>,
+    transform: (T) -> Collection<U>
+): Matcher<T?> = neverNullMatcher { value -> matcher(expected).test(transform(value)) }
```
sschuberth, line 113: When talking about "expected" in tests, I like to also talk about "actual". So, would it be appropriate to rename `value` to `actual` here?

model/src/test/kotlin/licenses/LicenseViewTest.kt
```diff
@@ -55,23 +51,12 @@ class LicenseViewTest : WordSpec() {
...
-                    },
-                    { "List should not contain exactly licenses ${expectedLicenses.print().value}" }
-                )
-            }
+    ): Matcher<List<Pair<SpdxSingleLicenseExpression, LicenseSource>>?> =
+        transformingCollectionMatcher(
+            expected = licenses.toList(),
+            matcher = ::containAll
+        ) { list ->
+            list.map { (expression, source) -> expression.toString() to source }
```
sschuberth, line 59: I'm a bit confused where the `toSpdx()` conversion from before did go... and why we now have `toString()` instead. Could you please elaborate and / or explain in the commit message?

model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
```diff
@@ -19,13 +19,14 @@
 
 package org.ossreviewtoolkit.model.licenses
 
-import io.kotest.assertions.print.print
 import io.kotest.assertions.withClue
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.Matcher
-import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.be
```
sschuberth, line 25: Commit message: "this allows remove" -> "this allows to remove"

utils/test/src/main/kotlin/Utils.kt
```diff
@@ -110,3 +111,10 @@ fun <T, U> transformingCollectionMatcher(
     matcher: (Collection<U>) -> Matcher<Collection<U>>,
     transform: (T) -> Collection<U>
 ): Matcher<T?> = neverNullMatcher { value -> matcher(expected).test(transform(value)) }
+
+/**
+ * A helper function to create custom matchers that assert that the collection obtained by [transform] is empty.
+ */
+fun <T, U> transformingCollectionEmptyMatcher(
+    transform: (T) -> Collection<U>
+): Matcher<T?> = neverNullMatcher { value -> beEmpty<U>().test(transform(value)) }
```
sschuberth, line 120: `value` -> `actual`?

utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
```diff
@@ -24,11 +24,10 @@ import com.fasterxml.jackson.dataformat.yaml.YAMLMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 
 import io.kotest.assertions.assertSoftly
-import io.kotest.assertions.print.print
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.Matcher
-import io.kotest.matchers.MatcherResult
+import io.kotest.matchers.be
```
sschuberth, line 30: Commit message: "be reusing" -> "by reusing"

sschuberth (changes requested):

utils/test/src/main/kotlin/Utils.kt
mnonnenmacher: I'm not sure about this in this case, because the "actual" that is compared is derived from "value", so calling it "actual" could also create confusion in that regard.

model/src/test/kotlin/licenses/LicenseViewTest.kt
mnonnenmacher: I have added an explanation to the commit message.

sschuberth (approved):

mnonnenmacher:
Merged because failing tests are unrelated to this change.

Issue #3909, opened 2021-04-19
Provide more information about ORT-specific LicenseRefs


mnonnenmacher:
The [declared license mapping](https://github.com/oss-review-toolkit/ort/blob/master/utils/spdx/src/main/resources/declared-license-mapping.yml) contains a few mappings to licenses that start with "LicenseRef-ort-". For the user there is no way to find out more information about those licenses, for example the license text, without investigating the text that was mapped to these licenses and searching the Git log for the commits that introduced those mappings.
We should either stop using ORT-specific LicenseRefs or provide an easy way for users to get more information about those licenses.

Also see the discussion in this PR: https://github.com/oss-review-toolkit/ort/pull/3898

sschuberth:
FYI, at the example of `LicenseRef-ort-oracle-futc`, the license text is at https://www.oracle.com/downloads/licenses/oracle-free-license.html.

mnonnenmacher:
This is partly fixed by #4920, the only left license is "LicenseRef-ort-google-maps-tos".
@sschuberth @tsteenbe I'm not sure if we would want to include the license text of a commercial license in our repository. Shall we keep this open until https://github.com/nexB/scancode-toolkit/issues/2322 is implemented? Or remove the LicenseRef even before that?

sschuberth:
> include the license text of a commercial license in our repository.

I don't have a problem with that. And we I believe we already have rules depending on "LicenseRef-ort-google-maps-tos", I'd prefer to keep the ID and also add the text.

> Shall we keep this open until nexB/scancode-toolkit#2322 is implemented?

The issue is from 2020. So wouldn't hold my breath for it...

PR #4920, opened 2022-01-06
Handle part of the LicenseRef-ort license texts

mnonnenmacher:
Start addressing the missing license texts for LicenseRef-ort licenses from the declared license mapping (#3909). See the commit messages for details.

### DIFF
--- a/utils/spdx/src/main/resources/declared-license-mapping.yml
+++ b/utils/spdx/src/main/resources/declared-license-mapping.yml
@@ -164,7 +164,7 @@
 "Creative Commons Public Domain": CC-PDDC
 "Creative Commons Zero": CC0-1.0
 "Creative Commons": CC-BY-3.0
-"DBAD": LicenseRef-ort-dbad
+"DBAD": LicenseRef-scancode-dbad
 "DFSG approved": LicenseRef-scancode-free-unknown
 "Dual License: CDDL 1.0 and GPL V2 with Classpath Exception": CDDL-1.0 AND GPL-2.0-only
 "Dual license consisting of the CDDL v1.1 and GPL v2": CDDL-1.1 AND GPL-2.0-only
--- a/utils/spdx/src/main/resources/declared-license-mapping.yml
+++ b/utils/spdx/src/main/resources/declared-license-mapping.yml
@@ -377,7 +377,7 @@
 "Ruby's": Ruby
 "SIL OPEN FONT LICENSE Version 1.1": OFL-1.1
 "SIL Open Font License 1.1 (OFL-1.1)": OFL-1.1
-"SOLACE CORPORATION LICENSE AGREEMENT": LicenseRef-ort-Solace-Software-EULA
+"SOLACE CORPORATION LICENSE AGREEMENT": LicenseRef-scancode-solace-software-eula-2020
 "Scala License": Apache-2.0 AND BSD-3-Clause AND MIT
 "Sequence Library License (BSD-like)": BSD-3-Clause
 "Similar to Apache License but with the acknowledgment clause removed": LicenseRef-scancode-jdom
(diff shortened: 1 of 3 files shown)


utils/spdx/src/main/resources/licenserefs/LicenseRef-ort-oracle-futc
```diff
@@ -0,0 +1,65 @@
+License Rights and Restrictions
```
mnonnenmacher, line 1: At the reviewer: Please check if you agree with the part of the text I have taken from the source.

sschuberth: I actually believe we should really take all of the text from that page, starting with the "Oracle Free Use Terms and Conditions" title. Esp. the definitions are required IMO, as the "License Rights and Restrictions" make use of the defined terms.

utils/spdx/src/test/kotlin/UtilsTest.kt
```diff
@@ -166,6 +166,13 @@ class UtilsTest : WordSpec() {
             "return null for an unknown SPDX LicenseRef" {
                 getLicenseText("LicenseRef-foo-bar") should beNull()
             }
+
+            "return the full license text for a LicenseRef-ort license" {
+                val text = getLicenseText("LicenseRef-ort-oracle-futc")?.trim()
```
sschuberth, line 171: I believe we should not trim (at least from the end) here, but ensure that files are committed without trailing spaces. Also trimming from the start is not required as there is not leading space before the asserted "License Rights and Restrictions".

utils/spdx/src/main/resources/declared-license-mapping.yml
```diff
@@ -164,7 +164,7 @@
 "Creative Commons Public Domain": CC-PDDC
 "Creative Commons Zero": CC0-1.0
 "Creative Commons": CC-BY-3.0
-"DBAD": LicenseRef-ort-dbad
+"DBAD": LicenseRef-scancode-dbad
```
sschuberth, line 167: I agree with the change, but it's a breaking one in case users have rules / classifications that use `LicenseRef-ort-dbad`.

utils/spdx/src/main/resources/declared-license-mapping.yml
```diff
@@ -377,7 +377,7 @@
 "Ruby's": Ruby
 "SIL OPEN FONT LICENSE Version 1.1": OFL-1.1
 "SIL Open Font License 1.1 (OFL-1.1)": OFL-1.1
-"SOLACE CORPORATION LICENSE AGREEMENT": LicenseRef-ort-Solace-Software-EULA
+"SOLACE CORPORATION LICENSE AGREEMENT": LicenseRef-scancode-solace-software-eula-2020
```
sschuberth, line 380: See my comment to the previous commit.

sschuberth (changes requested):

utils/spdx/src/main/resources/licenserefs/LicenseRef-ort-oracle-futc
mnonnenmacher: I have added the full text.

utils/spdx/src/test/kotlin/UtilsTest.kt
mnonnenmacher: This was done for consistency with other similar tests, see lines 139 and 150.

utils/spdx/src/main/resources/declared-license-mapping.yml
mnonnenmacher: I have added the breaking change label to the PR.

sschuberth (approved):

PR #4934, opened 2022-01-11
Use JDK 17 LTS in the `Dockerfile`

sschuberth:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,16 +28,7 @@ ARG CRT_FILES=""
 # Set this to the ScanCode version to use.
 ARG SCANCODE_VERSION="30.1.0"
 
-FROM adoptopenjdk/openjdk11:alpine-slim AS build
-
-# Apk install commands.
-RUN apk add --no-cache \
-        # Required for Node.js to build the reporter-web-app.
-        libstdc++ \
-        # Required to allow to download via a proxy with a self-signed certificate.
-        ca-certificates \
-        coreutils \
-        openssl
+FROM eclipse-temurin:17-jdk-focal AS build
 
 COPY . /usr/local/src/ort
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:1.2
+# syntax=docker/dockerfile:1.3
```
MarcelBochtler, line 1: While this works, I'd prefer to use `1.3.0` like it is done in the documentation.

sschuberth: That apparently depends on the docs you look at, e.g. https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#build-mounts-run---mount has `# syntax=docker/dockerfile:1.3`. Also, if it works like for Docker image tags, `1.3` actually means `1.3.x`, so you'll get patch-level updates automatically, which probably is a good thing.

Dockerfile
```diff
@@ -28,16 +28,7 @@ ARG CRT_FILES=""
 # Set this to the ScanCode version to use.
```
MarcelBochtler, line 28: Commit message:

> While "Alpine is not in a supported release by OpenJDK" [3],
eclipse-temurin [4] (which supersedes the deprecated adoptopenjdk [5])
does offer an Alpine image [6]. However, prefer to use the slightly
larger "17-jdk-focal" image instead as that JDK image can also be easily
used to *run* ORT in order to supersede #4178, so building and running
ORT share the same image.

I agree with the change, but I understand your argumentation as follows:
You use the slightly larger "17-jdk-focal" image instead of using an alpine base image.
What I think you mean is that you use the JDK image instead of the JRE image which was used before.

sschuberth: No, I indeed mean "I'm using an Ubuntu (Focal) instead of an Alpine image" (independently of JRE vs. JDK) although it's larger, because I can then use the same image for running ORT, so only a single Ubuntu image needs to be downloaded instead of Ubuntu + Alpine images.

sschuberth: I.e., there would also be an Alpine *JDK* image (see `eclipse-temurin:17-jdk-alpine`), and this is the explanation why this is *not* used.

sschuberth: I've reworded the commit message a bit to hopefully make that clearer.

MarcelBochtler:
LGTM

sschuberth:
Approvals anyone @oss-review-toolkit/kotlin-devs?

Dockerfile
mnonnenmacher: > ... so you'll get patch-level updates automatically, which probably is a good thing.

Unless the next release is called 1.3.33-liberty-2 ;)

mnonnenmacher (approved):

PR #4937, opened 2022-01-11
Evaluator: Pass the current time to the policy rules script

fviernau:
Policy rules may require the current date and time as input, such as
e.g. checks for copyright year ranges. Constructor inject the date and
time into the script to avoid a direct dependency on the system time in
order not to sacrifice testability.

### DIFF
--- a/evaluator/src/main/kotlin/Evaluator.kt
+++ b/evaluator/src/main/kotlin/Evaluator.kt
@@ -39,7 +39,8 @@ import org.ossreviewtoolkit.utils.scripting.ScriptRunner
 class Evaluator(
     ortResult: OrtResult = OrtResult.EMPTY,
     licenseInfoResolver: LicenseInfoResolver = OrtResult.EMPTY.createLicenseInfoResolver(),
-    licenseClassifications: LicenseClassifications = LicenseClassifications()
+    licenseClassifications: LicenseClassifications = LicenseClassifications(),
+    time: Instant = Instant.now()
 ) : ScriptRunner() {
     override val compConfig = createJvmCompilationConfigurationFromTemplate<RulesScriptTemplate> {
         defaultImports(
@@ -49,7 +50,7 @@ class Evaluator(
     }
 
     override val evalConfig = ScriptEvaluationConfiguration {
-        constructorArgs(ortResult, licenseInfoResolver, licenseClassifications)
+        constructorArgs(ortResult, licenseInfoResolver, licenseClassifications, time)
         scriptsInstancesSharing(true)
     }
 
--- a/utils/scripting/src/main/kotlin/RulesScriptConfiguration.kt
+++ b/utils/scripting/src/main/kotlin/RulesScriptConfiguration.kt
@@ -21,6 +21,7 @@
 package org.ossreviewtoolkit.utils.scripting
 
 import java.security.MessageDigest
+import java.time.Instant
 
 import kotlin.script.experimental.annotations.KotlinScript
 import kotlin.script.experimental.api.ScriptAcceptedLocation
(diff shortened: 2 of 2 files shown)


sschuberth:
> Constructor inject the date and time into the script to avoid a direct dependency on the system time in order not to sacrifice testability.

That's the rationale why you're not simply calling `Instant.now()` in the script itself, correct?

fviernau:
> That's the rationale why you're not simply calling `Instant.now()` in the script itself, correct?

Yes.

evaluator/src/main/kotlin/Evaluator.kt
```diff
@@ -39,7 +39,8 @@ import org.ossreviewtoolkit.utils.scripting.ScriptRunner
 class Evaluator(
     ortResult: OrtResult = OrtResult.EMPTY,
     licenseInfoResolver: LicenseInfoResolver = OrtResult.EMPTY.createLicenseInfoResolver(),
-    licenseClassifications: LicenseClassifications = LicenseClassifications()
+    licenseClassifications: LicenseClassifications = LicenseClassifications(),
+    evaluationTime: Instant = Instant.now()
```
sschuberth, line 43: I'm wondering whether we should somehow deduplicate this with the `startTime` in `run()`... it would somehow be nice if the start time written to the evaluator result would match the time the script was seeing.

fviernau: I thought about this as well and decided against it, because of the following
- measurement of execution time IMO becomes less clear from the code and duration is no more limited to the execution time of the `run()` function but involves the time from construction to the return of `run()`.
- introduces the assumption that the current time is passed to the constructor of the evaluator
- single value is used for multiple purposes, exec duration + parameter for the rules

What do you think?

sschuberth: > * measurement of execution time IMO becomes less clear from the code and duration is no more limited to the execution time of the `run()` function but involves the time from construction to the return of `run()`.

As a side note, IIRC the *main* purpose of capturing `startTime` and `endTime` was *not* to capture the duration (otherwise we could have simply serialized the calculated duration), but to store the absolute time when then run was performed to e.g. ease debugging of network issues at that time. And only as a *bonus*, also the duration can be calculated.

> introduces the assumption that the current time is passed to the constructor of the evaluator

That coupling *might* indeed be a problem.

> single value is used for multiple purposes, exec duration + parameter for the rules

Using a single value for multiple purposes is not a problem per se, IMO, if the purposes are strongly related, as they are here.

But esp. for the case of testing I can see how you want to capture the correct `startTime` / `endTime` but set an arbitrary `evaluationTime`, so I'm fine with the code as-is.

However, please add these considerations to the commit message, so we can later still comprehend the design decision.

mnonnenmacher: The purpose of the `evaluationTime` should also be documented at the constructor argument so that you don't have to check the Git history to find out what it is used for.

evaluator/src/main/kotlin/Evaluator.kt
```diff
@@ -39,7 +39,12 @@ import org.ossreviewtoolkit.utils.scripting.ScriptRunner
 class Evaluator(
     ortResult: OrtResult = OrtResult.EMPTY,
     licenseInfoResolver: LicenseInfoResolver = OrtResult.EMPTY.createLicenseInfoResolver(),
-    licenseClassifications: LicenseClassifications = LicenseClassifications()
+    licenseClassifications: LicenseClassifications = LicenseClassifications(),
+
+    /**
+     * The date and time of the evaluation to be used by policy rules as the current date and time.
```
sschuberth, line 45: The repetition of "data and time" sounds a bit odd, IMO.

Actually, I'm thinking about renaming the parameter to just `time` as the "evaluation" bit in `evaluationTime` really sounds very much like what `startTime` is.

And if it's just generically called "time", I think we could even drop the docs, as it's then completely up to the implementation *what* time is it and how to use it in the rules.

evaluator/src/main/kotlin/Evaluator.kt
```diff
@@ -39,7 +39,8 @@ import org.ossreviewtoolkit.utils.scripting.ScriptRunner
 class Evaluator(
     ortResult: OrtResult = OrtResult.EMPTY,
     licenseInfoResolver: LicenseInfoResolver = OrtResult.EMPTY.createLicenseInfoResolver(),
-    licenseClassifications: LicenseClassifications = LicenseClassifications()
+    licenseClassifications: LicenseClassifications = LicenseClassifications(),
+    time: Instant = Instant.now()
```
sschuberth, line 43: The commit message now needs to be adjusted accordingly.

sschuberth (approved):

Issue #4663, opened 2021-11-04
How to map vulnerability scores to EvaluatedModel statistics? 


tsteenbe:
Vulnerabilities use different scoring systems CVSSv3 or CVSSv2 which I would like to add to EvaluatedModel [Statistics.kt](https://github.com/oss-review-toolkit/ort/blob/master/reporter-web-app/src/models/Statistics.js) but the questions is how should the mapping be done? Do we map it to error, warning, hints same as issues/rule violations?

tsteenbe:
Discussed in the ORT developer meeting of November 4th, 2020 - we won't add any mapping for vulnerability as we agreed one should throw policy violations if one wants scanned project to address vulnerabilities. WebApp will only list the vulnerabilities.

PR #4939, opened 2022-01-11
Statistics: Add the number of open vulnerabilities

mnonnenmacher:

### DIFF
--- a/reporter/src/main/kotlin/reporters/evaluatedmodel/StatisticsCalculator.kt
+++ b/reporter/src/main/kotlin/reporters/evaluatedmodel/StatisticsCalculator.kt
@@ -45,6 +45,7 @@ internal class StatisticsCalculator {
         repositoryConfiguration = getRepositoryConfigurationStatistics(ortResult),
         openIssues = getOpenIssues(ortResult, resolutionProvider),
         openRuleViolations = getOpenRuleViolations(ortResult, resolutionProvider),
+        openVulnerabilities = getOpenVulnerabilities(ortResult, resolutionProvider),
         dependencyTree = DependencyTreeStatistics(
             includedProjects = ortResult.getProjects().count { !ortResult.isExcluded(it.id) },
             excludedProjects = ortResult.getProjects().count { ortResult.isExcluded(it.id) },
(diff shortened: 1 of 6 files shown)


tsteenbe:
@mnonnenmacher  Didn't we agree in a last November/December ORT developer meeting that we would not have an open vulnerabilities stat number as what is "open" should be controlled by the ORT user via policy violations?

reporter/src/main/kotlin/reporters/evaluatedmodel/StatisticsCalculator.kt
```diff
@@ -78,6 +79,10 @@ internal class StatisticsCalculator {
         )
     }
 
+    private fun getOpenVulnerabilities(ortResult: OrtResult, resolutionProvider: ResolutionProvider): Int =
+        ortResult.getVulnerabilities(omitExcluded = true).values.flatten()
```
oheger-bosch, line 83: Just a question for my understanding: In this implementation resolved vulnerabilities are filtered out using the `resolutionProvider`. It would also be possible to call `getVulnerabilities()` with `omitResolved` set to *true*, right?. What is the difference? And isn't there some redundancy with the different ways to apply resolutions?

mnonnenmacher: The difference is that for `omitResolved` the function only takes vulnerability resolutions into account that are configured in the repository configuration. The `resolutionProvider` also knows about global vulnerability resolutions. I'm not super happy with that, but I have included the option to be consistent with the `getRuleViolations()` function.

mnonnenmacher:
> @mnonnenmacher Didn't we agree in a last November/December ORT developer meeting that we would not have an open vulnerabilities stat number as what is "open" should be controlled by the ORT user via policy violations?

I did not remember that, but here is the related issue: https://github.com/oss-review-toolkit/ort/issues/4663

However, I think that decision was not well thought through, because we have vulnerability resolutions after all which contradicts the idea of using rules for that. To be more consistent with how we handle issues and rule violations we should have a configuration property to configure if unresolved vulnerabilities should be reported by the web app reporter, similar to the severity thresholds for issues and rule violations. If a user decides to use rules instead, they could set that property to false.

reporter/src/main/kotlin/reporters/evaluatedmodel/StatisticsCalculator.kt
oheger-bosch: Thanks for explaining.

oheger-bosch (approved):

PR #4940, opened 2022-01-12
JiraNotifier: Add function to change the state of an issue

bs-ondem:


notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -47,6 +48,31 @@ class JiraNotifier(
...
 
+    /**
+     * Change the [state] of the given issue specified by the [issueKey].
+     */
+    fun changeState(issueKey: String, state: String): Result<Boolean> {
+        val issue = restClient.issueClient.getIssue(issueKey).claim()
+        val possibleTransitions = restClient.issueClient.getTransitions(issue).claim()
+
+        possibleTransitions.forEach { transition ->
+            if (state == transition.name) {
```
sschuberth, line 59: Maybe first `filter` for this?

notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -47,6 +48,31 @@ class JiraNotifier(
...
     }
 
+    /**
+     * Change the [state] of the given issue specified by the [issueKey].
+     */
+    fun changeState(issueKey: String, state: String): Result<Boolean> {
+        val issue = restClient.issueClient.getIssue(issueKey).claim()
+        val possibleTransitions = restClient.issueClient.getTransitions(issue).claim()
+
+        possibleTransitions.forEach { transition ->
```
MarcelBochtler, line 58: How about this to get rid of the loop?:
```kt
val transition = possibleTransitions.singleOrNull { it.name == state } ?: return Result.failure(
    IllegalArgumentException("The transition to the state '$state' is not possible.")
)
```

notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -47,6 +48,31 @@ class JiraNotifier(
...
+                return try {
+                    restClient.issueClient.transition(issue, TransitionInput(transition.id)).claim()
+
+                    Result.success(true)
+                } catch (e: RestClientException) {
+                    log.error { "The transition to state '$state' for the issue '$issueKey' is not possible: " +
+                            e.collectMessagesAsString() }
+
+                    Result.failure(e)
+                }
```
sschuberth, line 69: You should be able to slightly simplify this, similar to like I just did in https://github.com/oss-review-toolkit/ort/pull/4944/commits/85fcab0c62080fb1858008962202fb4a3091fe93.

sschuberth (changes requested):

notifier/src/main/kotlin/modules/JiraNotifier.kt
bs-ondem: Ah nice! Due to the changes from https://github.com/oss-review-toolkit/ort/commit/f24fa74edb96056e2d7199016ca32cadb8df6829 we can use `runCatching` now 👍

notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -47,6 +48,26 @@ class JiraNotifier(
...
+        val issue = restClient.issueClient.getIssue(issueKey).claim()
+        val possibleTransitions = restClient.issueClient.getTransitions(issue).claim()
+
+        val transition = possibleTransitions.singleOrNull { it.name == state } ?: return Result.failure(
+            IllegalArgumentException("The transition to the state '$state' is not possible.")
+        )
+
+        return runCatching {
+            restClient.issueClient.transition(issue, TransitionInput(transition.id)).claim()
+            true
```
MarcelBochtler, line 64: There is no advantage of using `Result<Boolean>` as a return value. The success of the creation is already shown by `Result` itself.
It should be possible to use `Result<Void>` and to remote the `true` as a return value.

MarcelBochtler (changes requested):

notifier/src/main/kotlin/modules/JiraNotifier.kt
sschuberth: If `changeState` is not required to return a `Result` object, let it simply return `Boolean`, drop the `true` here, and use `.isSuccess` in line 68.

bs-ondem: I prefer using `Result<Void>` otherwise https://github.com/oss-review-toolkit/ort/blob/54c206ba4629849ebe542c7cfdd74d0d49698668/notifier/src/main/kotlin/modules/JiraNotifier.kt#L58-L59 would crash the notifier if a transition is not possible for one issue.

notifier/src/main/kotlin/modules/JiraNotifier.kt
```diff
@@ -47,6 +48,25 @@ class JiraNotifier(
...
+            IllegalArgumentException("The transition to the state '$state' is not possible.")
+        )
+
+        return runCatching {
+            restClient.issueClient.transition(issue, TransitionInput(transition.id)).claim()
+        }.onFailure {
+            log.error { "The transition to state '$state' for the issue '$issueKey' is not possible: " +
+                    it.collectMessagesAsString() }
+        }
+    }
```
sschuberth, line 68: So phrase out my previous comment, how about:

    fun changeState(issueKey: String, state: String): Boolean {
        val issue = restClient.issueClient.getIssue(issueKey).claim()
        val possibleTransitions = restClient.issueClient.getTransitions(issue).claim()

        return runCatching {
            val transition = possibleTransitions.single { it.name == state }
            restClient.issueClient.transition(issue, TransitionInput(transition.id)).claim()
        }.onFailure {
            log.error { "The transition to state '$state' for the issue '$issueKey' is not possible: " +
                    it.collectMessagesAsString() }
        }.isSuccess
    }

MarcelBochtler: Wouldn't it be better to use `Result<>` for two reasons:
1. Consistency with the `createIssue()` method.
2. Explicitly stating what the return value of the method represents?

sschuberth: > Wouldn't it be better to use `Result<>` for two reasons:

Personally, I don't have a clear opinion on that, and will leave it to you guys. I'll approve in any case.

> Explicitly stating what the return value of the method represents?

What's not explicit about the `Boolean` return value?

sschuberth (approved):

MarcelBochtler (approved):

notifier/src/main/kotlin/modules/JiraNotifier.kt
MarcelBochtler: > What's not explicit about the Boolean return value?

`Result<Void>` specifies that a http request is being made.
But I'm okay with either, and won't block.

sschuberth: > `Result<Void>` specifies that a http request is being made.

Uhm, does it? How? Because you know by heart which functions return `Void`?

bs-ondem: I think we can leave it as it is for now.

MarcelBochtler: Yes, you're right. It's not http request related, but `Result` explicitly states that it can fail or succeed.
A Boolean return value only implicitly states the same thing.

PR #4941, opened 2022-01-12
SpdxExpression: Simplify the DNF for compound expressions with equal operands

sschuberth:
In contrast to the argumentation in fe1f1ff, this should not cause any
severe performance issues with large expressions as "equals()" does not
call `toString()`.

### DIFF
--- a/utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
+++ b/utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
@@ -349,6 +349,11 @@ class SpdxExpressionTest : WordSpec() {
         }
 
         "disjunctiveNormalForm()" should {
+            "simplify compound expressions with equal operands" {
+                "a AND a".toSpdx().disjunctiveNormalForm() should beString("a")
+                "a OR a".toSpdx().disjunctiveNormalForm() should beString("a")
+            }
+
             "not change an expression already in DNF" {
                 "a AND b OR c AND d".toSpdx().disjunctiveNormalForm() should beString("a AND b OR c AND d")
             }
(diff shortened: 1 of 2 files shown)


utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
```diff
@@ -407,8 +412,7 @@ class SpdxExpressionTest : WordSpec() {
             "not contain duplicate valid choice for a complex expression" {
                 "(a OR b) AND (a OR b)".toSpdx().validChoices() should containExactlyInAnyOrder(
                     "a".toSpdx(),
-                    "b".toSpdx(),
-                    "a AND b".toSpdx()
+                    "b".toSpdx()
```
sschuberth, line 415: Esp. @MarcelBochtler, to double-check, do you agree with this to have been an invalid choice?

mnonnenmacher: Why? `(a OR b) AND (a OR b) == (a OR b)`, so `b` should be a valid choice.

mnonnenmacher (approved):

utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
mnonnenmacher: Sorry, i completely misunderstood the comment and now it's too late because it's already merged. My interpretation would be that if there are two different locations which offer the choice `(a OR b)` it would technically be possible to choose `a` for the first and `b` for the second. However, I'm not sure if this is a scenario that we would want to support.

sschuberth: > two different locations which offer the choice `(a OR b)`

You're right: The important bit here is that you need to take the *location* of the licenses into account (e.g. outbound vs inbound). I'll create a revert PR.

sschuberth: See https://github.com/oss-review-toolkit/ort/pull/4946.

PR #4942, opened 2022-01-12
Pip: Extend the work-around for declared license parsing

fviernau:
When the classifier format is accidentally used in the `license` property, it happens that just "OSI Approved :: " is used accidentally as the prefix instead of "License :: OSI Approved :: ". Add a work-around so this gets properly parsed.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pip.kt
+++ b/analyzer/src/main/kotlin/managers/Pip.kt
@@ -438,10 +438,13 @@ class Pip(
             it.isBlank() || it == "UNKNOWN"
         }
 
-    private fun getLicenseFromClassifier(classifier: String): String? =
+    private fun getLicenseFromClassifier(classifier: String): String? {
         // Example license classifier:
         // "License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)"
-        classifier.split(" :: ").takeIf { it.first() == "License" }?.last()?.takeUnless { it == "OSI Approved" }
+        val classifiers = classifier.split(" :: ").map { it.trim() }
+        val license = classifiers.takeIf { it.first() == "License" }?.last()
+        return license?.takeUnless { it in listOf("License", "OSI Approved") }
+    }
 
     private fun setupVirtualEnv(workingDir: File, definitionFile: File): File {
         // Create an out-of-tree virtualenv.
--- a/analyzer/src/main/kotlin/managers/Pip.kt
+++ b/analyzer/src/main/kotlin/managers/Pip.kt
@@ -442,8 +442,9 @@ class Pip(
         // Example license classifier:
         // "License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)"
         val classifiers = classifier.split(" :: ").map { it.trim() }
-        val license = classifiers.takeIf { it.first() == "License" }?.last()
-        return license?.takeUnless { it in listOf("License", "OSI Approved") }
+        val licenseClassifiers = listOf("License", "OSI Approved")
+        val license = classifiers.takeIf { it.first() in licenseClassifiers }?.last()
+        return license?.takeUnless { it in licenseClassifiers }
     }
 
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -438,10 +438,15 @@ class Pip(
...
+    private fun getLicenseFromClassifier(classifier: String): String? {
         // Example license classifier:
         // "License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)"
-        classifier.split(" :: ").takeIf { it.first() == "License" }?.last()?.takeUnless { it == "OSI Approved" }
+        return if (classifier.startsWith("License :: OSI Approved :: ")) {
+             classifier.substringAfterLast(" :: ", missingDelimiterValue = "").trim().ifEmpty { null }
+        } else {
+            null
+        }
+    }
```
sschuberth, line 454: I'd propose instead:

    private fun getLicenseFromClassifier(classifier: String): String? {
        // Example license classifier:
        // "License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)"
        val classifiers = classifier.split(" :: ").map { it.trim() }
        val license = classifiers.takeIf { it.first() == "License" }?.last()
        return license?.takeUnless { it in listOf("License", "OSI Approved") }
    }

analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -433,19 +433,24 @@ class Pip(
...
-            null
+        val prefixes = listOfNotNull("License :: OSI Approved :: ", extraPrefix)
+
+        prefixes.forEach { prefix ->
+            if (classifier.startsWith(prefix)) {
+                return classifier.substringAfterLast(" :: ", missingDelimiterValue = "").trim().ifEmpty { null }
+            }
         }
+
+        return null
```
sschuberth, line 453: Based on my previous suggestion, I'd extend it like

    private fun getLicenseFromClassifier(classifier: String): String? {
        // Example license classifier:
        // "License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)"
        val classifiers = classifier.split(" :: ").map { it.trim() }
        val licenseClassifiers = listOf("License", "OSI Approved")
        val license = classifiers.takeIf { it.first() in licenseClassifiers }?.last()
        return license?.takeUnless { it in licenseClassifiers }
    }

instead.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Pip.kt
fviernau: Are you aware that the work-around should only be applied for the caller `getLicenseFromLicenseField()` but not for all other callers?

sschuberth: I realize your current code is written like that, but actually I do not see any risk with making the code a bit simpler and apply the work-around in any case.

fviernau: Looked again at the classifiers and I agree there is not much risk, https://pypi.org/classifiers/.
Want me to squash so that we just have one commit ?

sschuberth: I'd be fine with squashing.

analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -438,10 +438,13 @@ class Pip(
             it.isBlank() || it == "UNKNOWN"
         }
 
-    private fun getLicenseFromClassifier(classifier: String): String? =
+    private fun getLicenseFromClassifier(classifier: String): String? {
```
sschuberth, line 441: Commit message: s/rebust/robust/

sschuberth (changes requested):

sschuberth (approved):

PR #4943, opened 2022-01-12
Merge curations before uploading to ClearlyDefined

sschuberth:

### DIFF
--- a/model/src/main/kotlin/PackageCurationData.kt
+++ b/model/src/main/kotlin/PackageCurationData.kt
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.annotation.JsonInclude
 
 import java.util.SortedSet
 
+import org.ossreviewtoolkit.utils.common.zip
 import org.ossreviewtoolkit.utils.core.DeclaredLicenseProcessor
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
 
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/PackageCurationData.kt
```diff
@@ -161,4 +162,27 @@ data class PackageCurationData(
...
         return CuratedPackage(pkg, curations)
     }
+
+    /**
+     * Merge with [other] curation data. If in question, data in this instance has precedence over data in the other
+     * instance.
+     */
+    fun merge(other: PackageCurationData) =
+        PackageCurationData(
+            comment = comment ?: other.comment,
```
mnonnenmacher, line 172: The comment is supposed to "contain information about how and why the curation was created", this means important information might be lost if only one of the comments is kept. But also merging both comments might be confusing if one curation completely replaces another. So I'm not sure if merging is the right approach in general.

sschuberth: Good point. I still believe joining existing comments is the only meaningful thing to do here. Esp. in the (currently only) context of creating ClearlyDefined curations, this is only something like a template after all... as part of the PR review the comment could still be reviewed / amended.

So I'll change the code to join comments.

mnonnenmacher (changes requested):

PR #4947, opened 2022-01-12
helper-cli: Add a command to save a result with resolved scopes

mnonnenmacher:
Add a helper-cli command to save an existing ORT result file with
resolved scopes. This is useful if the dependency tree in the file needs
to be investigated which is very difficult with the optimized format.

### DIFF
--- a/helper-cli/src/main/kotlin/HelperMain.kt
+++ b/helper-cli/src/main/kotlin/HelperMain.kt
@@ -43,6 +43,7 @@ import org.ossreviewtoolkit.helper.commands.ListPackagesCommand
 import org.ossreviewtoolkit.helper.commands.ListStoredScanResultsCommand
 import org.ossreviewtoolkit.helper.commands.MapCopyrightsCommand
 import org.ossreviewtoolkit.helper.commands.MergeRepositoryConfigurationsCommand
+import org.ossreviewtoolkit.helper.commands.SaveResultWithResolvedScopesCommand
 import org.ossreviewtoolkit.helper.commands.SetDependencyRepresentationCommand
 import org.ossreviewtoolkit.helper.commands.SetLabelsCommand
 import org.ossreviewtoolkit.helper.commands.SubtractScanResultsCommand
@@ -92,6 +93,7 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
             PackageConfigurationCommand(),
             PackageCurationsCommand(),
             RepositoryConfigurationCommand(),
+            SaveResultWithResolvedScopesCommand(),
             ScanStorageCommand(),
             SetDependencyRepresentationCommand(),
             SetLabelsCommand(),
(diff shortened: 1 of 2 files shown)


helper-cli/src/main/kotlin/commands/SaveResultWithResolvedScopesCommand.kt
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (C) 2021 Bosch.IO GmbH
```
sschuberth, line 2: Should be 2022.

helper-cli/src/main/kotlin/HelperMain.kt
```diff
@@ -92,6 +93,7 @@ internal class HelperMain : CliktCommand(name = ORTH_NAME, epilog = "* denotes r
             PackageConfigurationCommand(),
             PackageCurationsCommand(),
             RepositoryConfigurationCommand(),
+            SaveResultWithResolvedScopesCommand(),
```
sschuberth, line 96: Doesn't the existing `SetDependencyRepresentationCommand` do the same thing?

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/SaveResultWithResolvedScopesCommand.kt
mnonnenmacher: Actually, I wrote this in 2021, but fixed :)

helper-cli/src/main/kotlin/HelperMain.kt
mnonnenmacher: Maybe, at least the [`TargetFormat`](https://github.com/oss-review-toolkit/ort/blob/268cf7f127ccfcabf03b0085abe879042a18a7e1/helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt#L62-L73) there only supports `GRAPH`. @oheger-bosch Can you tell us more about this? Was this command supposed to also be able to convert to the tree format, or does it do this already and I miss something?

helper-cli/src/main/kotlin/commands/SaveResultWithResolvedScopesCommand.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val outputFile by option(
+        "--output-file", "-o",
+        help = "The file to write the ORT result with resolved scopes to."
+    ).convert { it.expandTilde() }
+        .file(mustExist = false, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+
+    override fun run() {
+        val ortResult = inputFile.readValue<OrtResult>()
```
sschuberth, line 56: Note that the helper-cli's `readOrtResult()` utility function already calls `withResolvedScopes()`; so maybe use it here?

helper-cli/src/main/kotlin/HelperMain.kt
oheger-bosch: The `SetDependencyRepresentationCommand` was indeed intended to do conversions in both directions - therefore the generic name and the `TargetFormat` enumeration. The conversion to the tree format just has not yet been implemented. I think, I started once with an implementation, but did not finish it; I will see whether I can activate this old branch.

oheger-bosch: See #4949.

mnonnenmacher:
Replaced by #4949 .

PR #4949, opened 2022-01-13
helper-cli: Support conversions to the dependency tree format

oheger-bosch:
SetDependencyRepresentationCommand currently only supports conversions
to the dependency graph format. Conversions to the tree format are
useful as well, as this format is more readable and is better suited
to analyze dependencies. So, extend the command to support conversions
in this direction, too.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Bosch.IO GmbH
+ * Copyright (C) 2021-2022 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


helper-cli/src/main/kotlin/commands/SetDependencyRepresentationCommand.kt
```diff
@@ -64,6 +64,11 @@ class SetDependencyRepresentationCommand : CliktCommand(
         GRAPH {
             override fun convert(result: AnalyzerResult): AnalyzerResult =
                 DependencyGraphConverter.convert(result)
+        },
+
+        /** The classic dependency tree format. */
+        TREE {
+            override fun convert(result: AnalyzerResult): AnalyzerResult = result.withScopesResolved()
```
sschuberth, line 71: Uhm, do we have both `OrtResult.withResolvedScopes()` and `AnalyzerResult.withScopesResolved()` functions in ORT? If so, can we align this by add a commit to this PR? Personally, I'd prefer the "withResolvedScopes" name.

sschuberth (approved):

PR #4953, opened 2022-01-13
FossID: Fix and improve handling of `Scan.isArchived`

nnobelis:
Please refer to individual commit message.


clients/fossid-webapp/src/main/kotlin/model/Scan.kt
```diff
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.clients.fossid.model
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
```
sschuberth, line 23: Commit message nits:
- If "String" in the title refers to the Java class, put it also in backticks (`String`).
- In 'either null, either "0"' drop the second "either".
- "This commit fixes InvalidFormatException" -> "This commit fixes an `InvalidFormatException`"

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -398,10 +398,11 @@ class FossId internal constructor(
         url: String,
         revision: String? = null
     ): List<Scan> = filter {
+        val isArchived = it.isArchived ?: false
```
sschuberth, line 401: Does it make sense to instead modify the `IntBooleanDeserializer` to deserialize to `false` in case of a `null` value? Is that the right thing to do for all users of `IntBooleanDeserializer`?

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -398,10 +398,11 @@ class FossId internal constructor(
...
         // The scans in the server contain the url with the credentials so we have to remove it for the
         // comparison. If we don't, the scans won't be matched if the password changes!
         val urlWithoutCredentials = it.gitRepoUrl?.replaceCredentialsInUri()
-        urlWithoutCredentials == url && (revision == null || (it.gitBranch == revision))
+        !isArchived && urlWithoutCredentials == url && (revision == null || (it.gitBranch == revision))
```
sschuberth, line 405: Unrelated, but mind adding another commit that removes the superfluous parentheses around `(it.gitBranch == revision)`?

sschuberth (approved):

nnobelis:
@sschuberth pushed new changes because of CI checks error.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -402,7 +402,7 @@ class FossId internal constructor(
         // The scans in the server contain the url with the credentials so we have to remove it for the
         // comparison. If we don't, the scans won't be matched if the password changes!
         val urlWithoutCredentials = it.gitRepoUrl?.replaceCredentialsInUri()
-        !isArchived && urlWithoutCredentials == url && (revision == null || (it.gitBranch == revision))
+        !isArchived && urlWithoutCredentials == url && (revision == null || it.gitBranch == revision)
```
sschuberth, line 405: Commit message nit: It should be "parenthes*e*s" (plural).

sschuberth (changes requested):

sschuberth (approved):

PR #4960, opened 2022-01-14
Implement auto-completion support for Notifier scripts

sschuberth:


tsteenbe (commented):
@sschuberth  Am I right this auto-completion works similar to how you changed rules.kts, if you then shouldn't this PR also update https://github.com/oss-review-toolkit/ort/blob/master/examples/notifications.kts?

sschuberth:
Yes, I'll also update the example once the details of this PR (like the chosen double-extension of notification scripts) are set. But I'll do this separately as this PR already has 8 commits.

sschuberth:
> But I'll do this separately as this PR already has 8 commits.

Ok, I changed my mind an added 2 additional commits as I had to make a fixup anyway.

README.md
```diff
@@ -54,7 +54,7 @@ use):
 * [_Reporter_](#reporter) - presents results in various formats such as visual reports, Open Source notices or
   Bill-Of-Materials (BOMs) to easily identify dependencies, licenses, copyrights or policy rule violations.
 * [_Notifier_](./notifier) - sends result notifications via different channels (like
-  [emails](./examples/notifications.kts) and / or JIRA tickets).
+  [emails](./examples/notifications/src/main/resources/example.notifications.kts) and / or JIRA tickets).
```
mnonnenmacher, line 57: Commit message typo: "zu" -> "to"

mnonnenmacher (changes requested):

README.md
sschuberth: Fixed.

mnonnenmacher:
> Yes, I'll also update the example once the details of this PR (like the chosen double-extension of notification scripts) are set.

I do agree with the chosen name.

mnonnenmacher (approved):

sschuberth:
> I do agree with the chosen name.

I was considering to rather use `*.<toolname>.kts`, like `*.evaluator.kts` instead of `*.rules.kts`. That would al least "scale" if more tools got scripting. On the other hand it would get difficult again once a single tool supports multiple types of scripts...

PR #4961, opened 2022-01-14
GoDep: Fix retrieval of VCS metadata

MarcelBochtler:
Please have a look at the commit messages for more details.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoDep.kt
+++ b/analyzer/src/main/kotlin/managers/GoDep.kt
@@ -28,6 +28,7 @@ import java.nio.file.Paths
 
 import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
 import org.ossreviewtoolkit.analyzer.PackageManager
+import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtIssue
@@ -39,6 +40,7 @@ import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.RemoteArtifact
 import org.ossreviewtoolkit.model.Scope
 import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
@@ -275,7 +277,14 @@ class GoDep(
 
         val repoRoot = Paths.get(gopath.path, "src", importPath).toFile()
 
+        // The "processProjectVcs()" function should always be able to deduce VCS information from the working tree
+        // created by "go get". However, if that fails for whatever reason, fall back to guessing VCS information from
+        // the "importPath" (which usually resembles a URL).
+        val fallbackVcsInfo = VcsHost.toVcsInfo("https://$importPath").takeIf {
+            it.type != VcsType.UNKNOWN
+        } ?: VcsInfo.EMPTY
+
         // We want the revision recorded in Gopkg.lock contained in "vcs", not the one "go get" fetched.
-        return processProjectVcs(repoRoot).copy(revision = revision)
+        return processProjectVcs(repoRoot, fallbackVcsInfo).copy(revision = revision)
     }
 }
--- a/analyzer/src/main/kotlin/managers/GoDep.kt
+++ b/analyzer/src/main/kotlin/managers/GoDep.kt
@@ -253,7 +253,10 @@ class GoDep(
     }
 
     private fun resolveVcsInfo(importPath: String, revision: String, gopath: File): VcsInfo {
-        val pc = ProcessCapture("go", "get", "-d", importPath, environment = mapOf("GOPATH" to gopath.path))
+        val pc = ProcessCapture(
+            "go", "get", "-d", importPath,
+            environment = mapOf("GOPATH" to gopath.path, "GO111MODULE" to "off")
+        )
 
         // HACK Some failure modes from "go get" can be ignored:
         // 1. repositories that don't have .go files in the root directory


analyzer/src/main/kotlin/managers/GoDep.kt
```diff
@@ -272,7 +273,11 @@ class GoDep(
 
         val repoRoot = Paths.get(gopath.path, "src", importPath).toFile()
 
+        // Create fallback VcsInfo for known VCS hosts.
+        val fallbackVcsInfo =
+            VcsHost.toVcsHost(URI("https://$importPath"))?.toVcsInfo("https://$importPath") ?: VcsInfo.EMPTY
```
MarcelBochtler, line 278: I explicitly did not use `VcsHost.toVcsInfo(importPath)`, as this would also try to create `VcsInfo`s for non-VCS URLs.
E.g. 
```yml
        vcs_processed:
          type: ""
          url: "go.uber.org/atomic"
          revision: "1ea20fb1cbb1cc08cbd0d913a96dead89aa18289"
          path: ""
```

analyzer/src/main/kotlin/managers/GoDep.kt
```diff
@@ -28,6 +28,7 @@ import java.nio.file.Paths
 
 import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
 import org.ossreviewtoolkit.analyzer.PackageManager
+import org.ossreviewtoolkit.downloader.VcsHost
```
sschuberth, line 31: Commit message: I might be mistaken, but I though it's anyway not directly the `go get -d` call that return the VCS information, but it only creates a local working tree that we inspect for VCS information later. Is that not the case?

In any case, could you please provide more detail under which circumstances VCS information might not be available?

analyzer/src/main/kotlin/managers/GoDep.kt
sschuberth: Hmm, that might even be a bug in `VcsHost.toVcsInfo()`. I guess I'd use it anyway as it's shorter, but only take the return value if it's not a `VcsType.UNKNOWN`.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/GoDep.kt
MarcelBochtler: I did more research and the issue is that from Go 1.16 the behavior of `go get` changed.
I added a commit to move back to the old behavior.

For more information see: https://maelvls.dev/go111module-everywhere/

MarcelBochtler: Regarding the fallback URL.
I'm not sure if it is required anymore. In my tests all sources downloaded via `go get` contained valid VCS metadata.
What do you think?

analyzer/src/main/kotlin/managers/GoDep.kt
```diff
@@ -253,7 +253,13 @@ class GoDep(
     }
 
     private fun resolveVcsInfo(importPath: String, revision: String, gopath: File): VcsInfo {
-        val pc = ProcessCapture("go", "get", "-d", importPath, environment = mapOf("GOPATH" to gopath.path))
+        val pc = ProcessCapture(
+            "go",
+            "get",
+            "-d",
+            importPath,
```
sschuberth, line 262: Nit: Even if it somehow violates our formatting habits, maybe put all of

    "go", "get", "-d", importPath

in a single line, as it all belongs to the same CLI call?

analyzer/src/main/kotlin/managers/GoDep.kt
sschuberth: I'm also a bit undecided. On the one hand it's always safer to have a fallback; on the other hand, if we can't for the love of God argue under which circumstances we'd run into that fallback, it's a bit weird to add.

I'd probably be pragmatic here and say, now that we have it, we keep it. But we should make clear in a code comment that this should actually never happen, and eventually log a warning in case it does.

analyzer/src/main/kotlin/managers/GoDep.kt
```diff
@@ -278,7 +280,11 @@ class GoDep(
 
         val repoRoot = Paths.get(gopath.path, "src", importPath).toFile()
 
+        // Create fallback VcsInfo for known VCS hosts.
+        val fallbackVcsInfo =
+            VcsHost.toVcsInfo("https://$importPath").takeIf { it.type != VcsType.UNKNOWN } ?: VcsInfo.EMPTY
```
sschuberth, line 285: Formatting nit:

```
val fallbackVcsInfo = VcsHost.toVcsInfo("https://$importPath").takeIf {
    it.type != VcsType.UNKNOWN
} ?: VcsInfo.EMPTY
```

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/GoDep.kt
MarcelBochtler: I changed the code comment for the fallback code.
I did not add the log message as it would require more logic which is currently hidden in `processProjectVcs()`.

But I can also add it if you insist.

analyzer/src/main/kotlin/managers/GoDep.kt
```diff
@@ -275,7 +277,13 @@ class GoDep(
 
         val repoRoot = Paths.get(gopath.path, "src", importPath).toFile()
 
+        // `go get` should always provide VCS information. However, if it doesn't the importPath can be used to create
```
sschuberth, line 280: Again, it's not `go get` itself that provides the VCS information. Proposal:

The `processProjectVcs()` function should always be able to deduce VCS information from the working tree created by `go get`. However, if that fails for whatever reason, fall back to guessing VCS information from the `importPath` (which usually resembles a URL).

analyzer/src/main/kotlin/managers/GoDep.kt
sschuberth: That's finer with me, I do not insist on the warning.

sschuberth (changes requested):

sschuberth (approved):

sschuberth:
Merging despite the unrelated test timeout.

PR #4962, opened 2022-01-15
Allow to deactivate package managers

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -148,7 +148,7 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",").default(PackageManager.ALL)
+    }.split(",").default(PackageManager.ALL.toList())
 
     private val globalOptionsForSubcommands by requireObject<GlobalOptions>()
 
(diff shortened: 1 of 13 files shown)


cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -142,14 +142,24 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
-        help = "The comma-separated package managers to activate, any of ${allPackageManagersByName.keys}."
+        help = "The comma-separated package managers to activate, any of ${allPackageManagersByName.keys}. Note that " +
+                "deactivation overrides activation."
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
     }.split(",").default(PackageManager.ALL)
 
+    private val deactivatedPackageManagers by option(
+        "--no-package-managers", "-n",
```
tsteenbe, line 155: `--no-package-managers` sound weird to me, using `--disable-package-managers` or `--deactivate-package-managers` make more sense to me.

tsteenbe (dismissed):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
sschuberth: I chose a long option name that starts with "n" as I liked the short option name to be `-n`, which sounds like a "not"-operator, which actually makes some sense. And I liked that "n" alphabetically comes after "m" to resemble that the `-m` and `-n` options are related here 😀

tsteenbe: Understood but is it then better to change to `--not-package-managers`?

sschuberth: I'd be fine with that, will change.

oheger-bosch (approved):

PR #4966, opened 2022-01-17
ort-config package curation provider

mnonnenmacher:
Enable loading curations from the ort-config repository, see commit messages for details.


cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -212,6 +218,7 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
                 config.analyzer.sw360Configuration?.let {
                     Sw360PackageCurationProvider(it).takeIf { useSw360Curations }
                 },
+                OrtConfigPackageCurationProvider().takeIf { useOrtCurations },
```
fviernau, line 221: What's the rational for using the curations only as a fallback, instead of merging these into `packageCurations.toSet()`?

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+        return if (file.exists()) {
+            yamlMapper.readValue(file)
+        } else {
+            emptyList()
+        }
+    }
+}
+
+private fun Identifier.toCurationPath() =
+    "${type.encodeOr("_")}/${namespace.encodeOr("_")}/${name.encodeOr("_")}.yml"
```
fviernau, line 67: Has there already been a decision made to use this file structure for that repository?

In order to fail early I believe it'd be good to load all the curations upfront and keep them in memory with a `map`. Therefore `SimplePackageConfigurationProvider.forDirectory()` could be re-used if slighly refactored. 

This would avoid enforcing a directory structure, so that all files under a given directory are considered.

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+            updateOrtConfig(it)
+        }
+    }
+
+    override fun getCurationsFor(pkgIds: Collection<Identifier>) =
+        pkgIds.mapNotNull { pkgId ->
+            getCurationsFor(pkgId).takeIf { it.isNotEmpty() }?.let { pkgId to it }
+        }.toMap()
+
+    private fun getCurationsFor(pkgId: Identifier): List<PackageCuration> {
```
fviernau, line 56: This is a bit redundant and inconsistent with `SimplePackageConfigurationProvider.forDirectory()`.

If serialization fails, it should be obvious from the logged text in which file the problem resides.

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
fviernau: I wonder if you can drop this class in favor of a static function in SimplePackageCurationProvider, e.g. 
`forOrtConfigRespository()`

fviernau (changes requested):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
mnonnenmacher: Simply because that is the existing mechanism for the other providers and I did not want to change that now. I agree that it should not be only a fallback, but the same argument applies to the other providers and I'd rather discuss this with the team first before introducing a new mechanism.

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
mnonnenmacher: Yes, the repository enforces a certain structure which makes sense for a public repository. The exact structure is subject to further discussion of course, but for now it is that. Given that, it is just unnecessary to load all curations into memory, especially since hopefully a LOT of curations will be added to that repository.

fviernau: > Yes, the repository enforces a certain structure which makes sense for a public repository. 

It's out of question that it makes sense to enforce a certain structure in the repository. That was not my point. I was rather after whether it makes sense to make this ORT code reflect all these details or whether it should be kept a bit more generic.

> Given that, it is just unnecessary to load all curations into memory, especially since hopefully a LOT of curations will be added to that repository.

As you say "unnecessary" this implies to me you mean that it does not provide any value. From my comments above I was arguing that it does provide value, in particular it (1) implements fail early and (2) allows flexibility to the file structure
which might also make sense with changing it because: 

> The exact structure is subject to further discussion of course,

cli/src/main/kotlin/commands/AnalyzerCommand.kt
fviernau: Would it make sense to provide that rationale in the commit message?

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
mnonnenmacher: That's right, but I don't expect that the structure changes regularly given the effort required to migrate all files, I would more expect that it changes maybe once or twice at all, so it wouldn't be hard to keep in sync. The fail early part can also be bad, if it fails because of curations which are not required for your project at all. However, as the config repo verifies that curation files can be parsed it should actually never fail, except if there are breaking changes in the file format, but I think we would always want to implement those in a backward compatible way.

mnonnenmacher: Btw, I don't consider this to be the final solution for integrating the ort-config repo, it's just a quick implementation to be able to start using it. Once more files types like package configurations or reporter templates are added we might have to think about a more sophisticated integration anyway.

fviernau: I still tend to rather consider assuming a file organization is "unnecessary" compared to parsing all `yaml` files within the curations directory, in particular I don't see any performance or memory consumption issues with holding these in memory. Anyhow, let's just disagree on that one and continue.

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
```diff
@@ -0,0 +1,85 @@
...
+        val file = curationsDir.resolve("curations").resolve(pkgId.toCurationPath())
+        return if (file.exists()) {
```
fviernau, line 59: Should `isFile` be used instead of `exists` ?

fviernau (changes requested):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
mnonnenmacher: FYI, I have added the fallback topic to the agenda of the next dev meeting.

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
```diff
@@ -0,0 +1,85 @@
...
+            getCurationsFor(pkgId).takeIf { it.isNotEmpty() }?.let { pkgId to it }
+        }.toMap()
+
+    private fun getCurationsFor(pkgId: Identifier): List<PackageCuration> {
+        val file = curationsDir.resolve("curations").resolve(pkgId.toCurationPath())
+        return if (file.isFile) {
+            runCatching {
+                yamlMapper.readValue<List<PackageCuration>>(file)
+            }.onFailure {
+                log.warn { "Failed parsing package curation from '${file.absolutePath}'." }
```
fviernau, line 63: Sorry, one more nit. I believe this should be error log level to ensure that it's output and because the exception makes ORT quite the process.

fviernau (approved):

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
mnonnenmacher: Just saw this now, I'll make another PR.

PR #4967, opened 2022-01-17
RulesScriptTemplate: Use a more specific name for a variable

fviernau:
The name `time` is too generic given the large size of it's visibility
scope. Use a more specific name to avoid naming conflicts and reduce the
risk of undesired variable name shadowing.

### DIFF
--- a/utils/scripting/src/main/kotlin/RulesScriptConfiguration.kt
+++ b/utils/scripting/src/main/kotlin/RulesScriptConfiguration.kt
@@ -55,7 +55,7 @@ open class RulesScriptTemplate(
     val ortResult: OrtResult = OrtResult.EMPTY,
     val licenseInfoResolver: LicenseInfoResolver = OrtResult.EMPTY.createLicenseInfoResolver(),
     val licenseClassifications: LicenseClassifications = LicenseClassifications(),
-    val time: Instant
+    val executionTime: Instant
 ) {
     val ruleViolations = mutableListOf<RuleViolation>()
 }


sschuberth:
Can we please merge https://github.com/oss-review-toolkit/ort/pull/4960 first so I don't have to rebase it (again)?

sschuberth:
> The name `time` is too generic given the large size of it's visibility
> scope.

As you probably remember, [the generic name was on purpose](https://github.com/oss-review-toolkit/ort/pull/4937#discussion_r783057699). What exactly changed your mind?

> Use a more specific name to avoid naming conflicts and reduce the
> risk of undesired variable name shadowing.

Where does such a shadowing currently occur? If it occurs in your custom rules, why can't you rename the variable in the rules instead?

fviernau:
> As you probably remember, [the generic name was on purpose](https://github.com/oss-review-toolkit/ort/pull/4937#discussion_r783057699). 

I do remember.

> What exactly changed your mind?

That's outlined in the commit message.

> Where does such a shadowing currently occur? 

The shadowing occured in:

```
fun getCurrentYear(): Int = Calendar.getInstance().apply { time = Date.from(time) }.get(Calendar.YEAR)
```

> If it occurs in your custom rules, why can't you rename the variable in the rules instead?

Of course I can rename the variable. I believe the discussion should be focused around which name is better rather than whether a different name can be chosen in the rules.

Anyhow, I'll drop this PR and go for 

```
fun getCurrentYear(): Int {
    val calendar = Calendar.getInstance()
    calendar.time = Date.from(time)
    return calendar.get(Calendar.YEAR)
}
```

PR #4969, opened 2022-01-17
Fix the dependency version check

sschuberth:


gradle.properties
```diff
@@ -87,6 +87,7 @@ org.gradle.parallel = true
 buildCacheRetentionDays = 7
 
 kotlin.code.style = official
+kotlin.stdlib.default.dependency=false
```
mnonnenmacher, line 89: I think this is worth a code comment.

mnonnenmacher (changes requested):

gradle.properties
sschuberth: I was hoping "Git blame" would be enough, but I can repeat the info from there in a code comment if you want.

mnonnenmacher: Yes, maybe git blame is enough.

mnonnenmacher (approved):

gradle.properties
sschuberth: I had actually added a comment but forgot to push. See https://github.com/oss-review-toolkit/ort/pull/4974.

PR #4970, opened 2022-01-17
AnalyzerCommand: Fix calculating the total project count

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -228,16 +228,19 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
             println("No project found.")
         } else {
             val filesPerManager = info.managedFiles.mapKeysTo(sortedMapOf()) { it.key.managerName }
+            var count = 0
 
             filesPerManager.forEach { (manager, files) ->
+                count += files.size
                 println("Found ${files.size} $manager project(s) at:")
+
                 files.forEach { file ->
                     val relativePath = file.toRelativeString(inputDir).takeIf { it.isNotEmpty() } ?: "."
                     println("\t$relativePath")
                 }
             }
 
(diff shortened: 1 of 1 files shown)


cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -228,16 +228,19 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
                 println("Found ${files.size} $manager project(s) at:")
+
                 files.forEach { file ->
                     val relativePath = file.toRelativeString(inputDir).takeIf { it.isNotEmpty() } ?: "."
                     println("\t$relativePath")
                 }
             }
 
-            println("Found ${filesPerManager.size} project(s) in total.")
+            println("Found $count project(s) from ${filesPerManager.size} package manager(s) in total.")
```
mnonnenmacher, line 243: You could use `filesPerManager.values.sumOf { it.size }` here to get rid of the `count` variable.

sschuberth: Correct, but I didn't like so much to iterate over `filesPerManager` twice.

mnonnenmacher: I think performance wise this is completely irrelevant, and I don't like the `count` variable :) But I'll leave it up to you.

mnonnenmacher (approved):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
mnonnenmacher: Too late now anyway, since I missed that auto-merge was on.

sschuberth: It's never too late for a follow-up 😉 I agree performance-wise it does not matter, but it just does not seem right to me to recalculate something that is already partly know.

PR #4976, opened 2022-01-19
fossid-webapp: Carry existing ignore rules to the new delta scan

nnobelis:
FossID offers the possibility to exclude files from the scan. To do so,
the user has to write 'ignore rules' in FossID UI.
This commit adapts the integration to takes these rules in account: when
a new delta scan is created, all 'ignore rules' from the previous latest
scan are carried to the new scan.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -34,18 +34,22 @@ import kotlinx.coroutines.withTimeoutOrNull
 
 import org.ossreviewtoolkit.clients.fossid.checkDownloadStatus
 import org.ossreviewtoolkit.clients.fossid.checkResponse
+import org.ossreviewtoolkit.clients.fossid.createIgnoreRule
 import org.ossreviewtoolkit.clients.fossid.createProject
 import org.ossreviewtoolkit.clients.fossid.createScan
 import org.ossreviewtoolkit.clients.fossid.deleteScan
 import org.ossreviewtoolkit.clients.fossid.downloadFromGit
 import org.ossreviewtoolkit.clients.fossid.getProject
 import org.ossreviewtoolkit.clients.fossid.listIdentifiedFiles
+import org.ossreviewtoolkit.clients.fossid.listIgnoreRules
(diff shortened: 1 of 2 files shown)


mnonnenmacher (dismissed):

sschuberth:
> to takes these rules

to *take*

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -470,6 +474,23 @@ class FossId internal constructor(
                 "The code for an existing scan must not be null."
             }
 
+            log.info { "Loading the ignore rules from '$existingScanCode'." }
```
sschuberth, line 477: Nits: Drop "the". Blank line after this one.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -470,6 +474,23 @@ class FossId internal constructor(
...
+                log.info { "${rules.size} ignore rule(s) have been found." }
+
+                // When a scan is created with the optional property 'git_repo_url', the server automatically creates
+                // an 'ignore rule' to exclude the '.git' directory.
+                // Therefore, this rule will be created automatically and does not need to be carried from the old scan.
+                rules.filterNot { it.type == RuleType.DIRECTORY && it.value == ".git" }.forEach {
+                    service.createIgnoreRule(config.user, config.apiKey, scanCode, it.type, it.value, RuleScope.SCAN)
+                        .checkResponse("create ignore rules", false)
+                    log.info { "Ignore rule of type '${it.type}' and value '${it.value}' " +
+                            "has been carried to the new scan." }
```
sschuberth, line 490: Nit: Looks like this line could be wrapped at a later column.

nnobelis: No it doesn't. Have I wrong IDE settings ?

image

sschuberth: By "later" I meant it could be wrapped after the word "new" only. Or be reformatted like

```
log.info {
    "Ignore ..."
}
```

sschuberth (approved):

sschuberth (approved):

PR #4977, opened 2022-01-19
Implement the OSADL license compliance matrix as rules

sschuberth:

~*Note to the reviewers:* Please do not put too much effort into reviewing the (intermediate) CSV-based implementation, as I'm switching to a JSON-based implementation in later commits (the JSON file was just published within the last day, that's why I didn't use that approach earlier; but I'd also like to keep the CSV implementation in the Git history for reference).~


evaluator/src/main/resources/osadl/osadl.rules.kts
```diff
@@ -0,0 +1,112 @@
...
+    dependencyRule("OSADL compliance matrix") {
+        // Requirements for the rule to trigger a violation.
+        require {
+            -isExcluded()
+        }
+
+        // Define a rule that is executed for each license of the dependency.
+        val licenseView = LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED
+        licenseRule("OSADL compliance matrix", licenseView) {
+            // Use all licenses of all direct dependencies as inbound licenses.
```
mnonnenmacher, line 74: Just wondering, why is it sufficient to check only direct dependencies here?

evaluator/src/main/resources/osadl/osadl.rules.kts
```diff
@@ -70,7 +56,7 @@ fun PackageRule.LicenseRule.isOsadlMatrixCompliant(inboundLicenses: Collection<S
 
         override fun matches(): Boolean {
             val outbound = license.toString()
-            return inboundLicenses.all { matrix[it to outbound] in Compatibility.COMPLIANT }
+            return inboundLicenses.all { matrix.getValue(it).getValue(outbound) in Compatibility.COMPLIANT }
```
mnonnenmacher, line 59: This will fail if one of the licenses is not contained in the OSADL matrix which is pretty likely to happen in a real life project with ScanCode results. Maybe it would be a better option to return false in this case, and optionally add another rule to check if all licenses are covered by the matrix.

evaluator/src/main/resources/osadl/osadl.rules.kts
sschuberth: Because for each dependency its directing dependencies are checked, so the whole tree is covered in the end. I believe that should be fine, or?

evaluator/src/main/resources/osadl/osadl.rules.kts
sschuberth: Or simply leave the rule as-is, but only create a warning instead of an error?

sschuberth: Or yet better (maybe that's what you mean): Only raise a violation if a license is explicitly defined as non-compliant. Yeah, that seems to make sense, will change.

evaluator/src/main/resources/osadl/osadl.rules.kts
```diff
@@ -0,0 +1,98 @@
...
+
+        override fun matches(): Boolean {
+            val outbound = license.toString()
+            return matrix.getValue(inbound).getValue(outbound) == Compatibility.NO
+        }
+    }
+
+val ruleSet = ruleSet(ortResult, licenseInfoResolver) {
+    // Define a rule that is executed for each dependency.
+    dependencyRule("OSADL compliance matrix") {
```
tsteenbe, line 63: We have always used all caps sperate by underscoring in the names of rules, see e.g. [1](https://github.com/oss-review-toolkit/ort/blob/master/examples/evaluator-rules/src/main/resources/example.rules.kts#L98) or [2](https://github.com/oss-review-toolkit/ort/blob/master/evaluator/src/main/resources/rules/no_gpl.rules.kts#L36) so for consistency and to make easier for user understand what this rule does I propose to rename "OSADL compliance matrix" t"OSADL_LICENSE_COMPLIANCE"

tsteenbe (changes requested):
Showing how one can implement license compatibility is a great new feature for ORT - why are we hiding this? Why is there not documentation update in this PR?

evaluator/src/main/resources/osadl/osadl.rules.kts
```diff
@@ -0,0 +1,98 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
```
tsteenbe, line 19: As we re-using work of OSADL I would like to give some credits e.g. link to their website where they make this dataset we use here available and add a note on the license of the JSON dataset.

tsteenbe (changes requested):

evaluator/src/main/resources/osadl/osadl.rules.kts
sschuberth: I was considering to drop the word "matrix", but then to me it sounded too much like "compliance *with* OSADL" (as if OSADL itself was a license or so). So maybe `OSADL_MATRIX_COMPLIANCE`?

sschuberth:
> Why is there not documentation update in this PR?

Simply because I already spent way too much time on this side-project due to the DSL / scripting improvements done as a preparation. I currently have no energy left for docs.

evaluator/src/main/resources/osadl/osadl.rules.kts
sschuberth: I believe such attribution should be better done as part of Markdown docs that explain how to use the evaluator and its built-in rules (which yet need to be created).

evaluator/src/main/resources/osadl/osadl.rules.kts
mnonnenmacher: I haven't looked at the matrix, but if there are situations where license A is compatible with B, B is compatible with C, but A is not compatible with C, it would not be fine.

dgutson:
How can we contribute to get this merged?

sschuberth:
I meanwhile was able to clarify how transitive dependencies are supposed to be looked up in the matrix and will revive this a PR after my vacation, in the week from Aug. 15.

utils/test/src/main/kotlin/OrtResultDsl.kt
```diff
@@ -0,0 +1,107 @@
...
+    private val packages = mutableListOf<Package>()
+
+    @OrtResultDsl
+    inner class PkgBuilder(private val id: String) {
+        @OrtResultDsl
+        var license = "Apache-2.0"
+
+        @OrtResultDsl
+        fun pkg(id: String, setup: PkgBuilder.() -> Unit): Package {
+            this@OrtResultBuilder.parentChildIds.getOrPut(this@PkgBuilder.id) { mutableListOf() } += id
```
nnobelis, line 54: Why not put `parentChildIds` as constructor parameter of `PkgBuilder` to avoid doing such property access ?
You could then make `PkgBuilder` a private class in this file (not an inner class).

This has benefits that `OrtResultBuilder` can be garbage collected even if `PkgBuilder` instances are still in the memory (here its not so important but it's best practice to avoid reference to the enclosing class).

mnonnenmacher:
@sschuberth Is this up for review again?

sschuberth:
> @sschuberth Is this up for review again?

Not yet. I realized I first need to implement a rule that operates on each project (and it's transitive dependencies).

evaluator/src/main/resources/osadl/osadl.rules.kts
sschuberth: I've changed this now according to the recommendations of @LeChasseur in the "osadl.rules: Only use the project's license as the outbound license" commit.

sschuberth:
> > @sschuberth Is this up for review again?
> 
> Not yet. I realized I first need to implement a rule that operates on each project (and it's transitive dependencies).

Now it's up for review again.

evaluator/src/test/kotlin/EvaluatorTest.kt
```diff
@@ -116,4 +117,51 @@ class EvaluatorTest : WordSpec({
...
+                }
+            }
+            val script = javaClass.getResource("/rules/osadl.rules.kts").readText()
+
+            val result = Evaluator(compatibleOrtResult).run(script)
+
+            result.violations should beEmpty()
+        }
+
+        "return a violation for incompatible licenses" {
```
fviernau, line 143: Would it make sense to add a test which verifies that a violation is just reported once per dependency / child dependency tuple, e.g. if the dependency appears mutliple times in the dependency tree (maybe also in different scopes)?

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -48,50 +48,60 @@ val matrix = javaClass.getResourceAsStream("/rules/osadl-matrix.json").use {
...
+    val licenseView = LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED
+
     // Define a rule that is executed for each dependency.
     dependencyRule("OSADL compliance matrix") {
         // Requirements for the rule to trigger a violation.
         require {
             -isExcluded()
         }
 
+        val projectLicenseInfo = licenseInfoResolver.resolveLicenseInfo(project.id).filter(licenseView)
```
fviernau, line 74: I don't understand where `project.id` / `project` is defined. Can you give a hint?

sschuberth: It's part of the `DependencyRule` itself:

https://github.com/oss-review-toolkit/ort/blob/22640c2bfe89692bec1a23660ec006469d51f4a0/evaluator/src/main/kotlin/DependencyRule.kt#L63-L66

So a dependency rules iterates over all dependencies *of all projects*, and we can use `project` to determinie what the "current" project is.

evaluator/src/test/kotlin/EvaluatorTest.kt
sschuberth: I basically don't have a problem with adding such a test, but I'm wondering whether this isn't testing too much of an implementation detail of `dependencyRule`:

https://github.com/oss-review-toolkit/ort/blob/22640c2bfe89692bec1a23660ec006469d51f4a0/evaluator/src/main/kotlin/RuleSet.kt#L87-L92

So should the "If the same dependency appears multiple times in the same scope, the rule will be applied only once" bit rather be tested as part of `DependencyRuleTest`?

evaluator/src/main/resources/rules/osadl.rules.kts
fviernau: Thanks!

evaluator/src/test/kotlin/EvaluatorTest.kt
fviernau: My intention for asking this was to see / point out that this rule flags redundant errors.
In each project, if a dependency with identifier `ID` occurs in `N` scopes I believe the dependency rule is exectued `N` times for `ID`.

Independently of that, I believe the amount of violations for repositories with many definition files, in particular for projects with a fine grained submodule structure can become quite "redundant". So, would it make sense to eliminate the outer "for loop-ish iteration through projects", and replace it by compution the applicable licenses for all (non-excluded) projects, and then using that as outbound license?

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,114 @@
...
+        val inbound = cellIterator.next()
+
+        columnHeaders.forEach { outbound ->
+            val value = cellIterator.next()
+            matrix[inbound to outbound] = Compatibility.fromString(value)
+        }
+    }
+}
+
+fun PackageRule.LicenseRule.isOsadlMatrixNonCompliant(inbound: SpdxSingleLicenseExpression) =
```
mnonnenmacher: I would rather name this `isOsadlMatrixCompliant` and use the `-` operator in the `require` block.

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,114 @@
...
+    object : RuleMatcher {
+        override val description = "isOsadlMatrixCompliant($license)"
```
mnonnenmacher, line 53: Wrong name, except if you change it above.

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,114 @@
...
+            // Use all licenses of all direct dependencies as inbound licenses.
+            directDependencies().forEach { dependency ->
+                val resolveLicenseInfo = licenseInfoResolver.resolveLicenseInfo(dependency.id).filter(licenseView)
+                val inboundLicenses = resolveLicenseInfo.licenses.map { it.license }
+
+                inboundLicenses.forEach { inboundLicense ->
+                    // Requirements for the rule to trigger a violation.
+                    require {
+                        +isOsadlMatrixNonCompliant(inboundLicense)
+                    }
```
mnonnenmacher: I would be surprised if this works as intended. The requirements are not magically only applied to the code block below, instead you add a new matcher to the dependency rule with each iteration of the loop, and they are all concatenated with `and`, and the code below also adds a new violation for each iteration. So if all matchers match you will get all violations, but if a single matcher does not match you will get no violations. To see this in the tests you would have to use test data with multiple dependencies on the same level.

If you want to use loops inside the rule simply use `if` instead of `require`.

evaluator/src/test/kotlin/EvaluatorTest.kt
mnonnenmacher: The `RuleSet` collects all violations in a set, so all duplicates should be eliminated. I think even across projects because it seems that the data in the violations will be identical.

evaluator/build.gradle.kts
```diff
@@ -33,6 +33,7 @@ dependencies {
 
     implementation("org.jetbrains.kotlin:kotlin-scripting-common")
     implementation("org.jetbrains.kotlin:kotlin-scripting-jvm-host")
+    implementation(libs.kotlinxSerialization)
```
mnonnenmacher, line 36: This commit should be squashed into the previous commit, why not use the easier approach right away?

mnonnenmacher (changes requested):

evaluator/src/test/kotlin/EvaluatorTest.kt
sschuberth: > In each project, if a dependency with identifier `ID` occurs in `N` scopes I believe the dependency rule is exectued `N` times for `ID`. What do you think?

I think you're right. IIRC I briefly discussed this with @mnonnenmacher, and we agreed that these "redundant" checks actually make sense, as different scopes might have different distribution channels, and scope excludes might apply.

evaluator/build.gradle.kts
sschuberth: Out of sentimentality, I guess 😆 I started this PR when there was no JSON but only the CSV version, and I did put some effort into the correct implementation that I wanted to keep, just in case someone ever needs to process and old CSV version again.

But I agree it's a bit inconvenient to review this way, also see my [initial note](https://github.com/oss-review-toolkit/ort/pull/4977#issue-1108001840) again. Do you prefer me to squash the JSON implementation into the CSV one?

sschuberth: > Do you prefer me to squash the JSON implementation into the CSV one?

I've done that.

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,110 @@
...
+        // Define a rule that is executed for each license of the dependency.
+        licenseRule("OSADL compliance matrix", licenseView) {
+            outboundLicenses.forEach { outboundLicense ->
+                if (isOsadlMatrixCompliant(outboundLicense).matches()) return@forEach
+
+                val inboundLicenseString = if (license.exception() != null) {
+                    "'${license.simpleLicense()}' (simplified from '$license')"
+                } else {
+                    "'$license'"
+                }
```
fviernau, line 88: Turning the license into this license string should be factored out into a function.

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,110 @@
...
+            // resulting false-positives are better than an incompatibility to be missed.
+            val inbound = license.simpleLicense()
+            return matrix.getValue(inbound).getValue(outbound.simpleLicense()) in compatibleValues
+        }
+    }
+
+val ruleSet = ruleSet(ortResult, licenseInfoResolver) {
+    val licenseView = LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED
+
+    // Define a rule that is executed for each dependency.
```
fviernau, line 69: Could you add a comment explaining why the licenses of the dependencies only need to be checked against the project licenses. E.g to me it is unclear why the matrix is checked like this: e.g. why don't the licenses of the dependencies need to be compatile with each other, or at least with their ancestors in the dependency tree.

fviernau (changes requested):

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,110 @@
...
+            }
+    }
+}
+
+@ExperimentalSerializationApi
+val matrix = javaClass.getResourceAsStream("/rules/osadl-matrix.json").use {
+    Json.Default.decodeFromStream<Map<String, Map<String, String>>>(it)
+}.mapValues { (_, outbounds) -> outbounds.mapValues { (_, value) -> Compatibility.fromString(value) } }
+
+fun PackageRule.LicenseRule.isOsadlMatrixCompliant(outbound: SpdxSingleLicenseExpression) =
```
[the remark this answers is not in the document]
sschuberth: I'd prefer to keep it implemented as a `RuleMatcher` to exemplify writing one, and for easy copy & paste into other similar code.

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,110 @@
...
+        require {
+            -isExcluded()
+        }
+
+        val projectLicenseInfo = licenseInfoResolver.resolveLicenseInfo(project.id).filter(licenseView)
+        val outboundLicenses = projectLicenseInfo.licenses.map { it.license }
+
+        // Define a rule that is executed for each license of the dependency.
+        licenseRule("OSADL compliance matrix", licenseView) {
+            outboundLicenses.forEach { outboundLicense ->
```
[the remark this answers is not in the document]
sschuberth: > e.g can licenses be incompatible with an empty project license?

I believe so, yes. The rule of thumb is not use *nothing* that does not come with a license, and that should also apply to own projects, which should come at least with a proprietary license.

evaluator/src/main/resources/rules/osadl.rules.kts
sschuberth: > E.g to me it is unclear why the matrix is checked like this: e.g. why don't the licenses of the dependencies need to be compatile with each other, or at least with their ancestors in the dependency tree.

@LeChasseur, could you chime in here and repeat what you told me during our meeting about that?

LeChasseur: It depends on the purpose of the compatibility check:
1.
If I want to check compatibility with the project license, it is sufficient to check compatibility of the licenses of the dependencies (one-way) with the project license.
2.
If I want to check if the involved licenses are compatible in general (independent of the outbound license), it is necessary to check the compatibility of all licenses among each other. One use case is to verify if I can combine code under several licenses in one new project.

sschuberth:
FYI, I'm waiting for a bug in the OSADL matrix to get fixed before moving this again out of draft state.

dgutson:
> FYI, I'm waiting for a bug in the OSADL matrix to get fixed before moving this again out of draft state.

@sschuberth could you please provide the link if it is in the issue tracker?

sschuberth:
> > FYI, I'm waiting for a bug in the OSADL matrix to get fixed before moving this again out of draft state.
> 
> @sschuberth could you please provide the link if it is in the issue tracker?

The bug we discovered was reported via @LeChasseur and I don't have a link.

sschuberth:
> > > FYI, I'm waiting for a bug in the OSADL matrix to get fixed before moving this again out of draft state.
> > 
> > 
> > @sschuberth could you please provide the link if it is in the issue tracker?
> 
> The bug we discovered was reported via @LeChasseur and I don't have a link.

The bugs (there were actually multiple related ones) were fixed in the OSADL matrix dated "2022-10-18 16:03". I've updated my PR accordingly and will move it out of draft state again.

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,76 @@
...
+
+        val projectLicenseInfo = licenseInfoResolver.resolveLicenseInfo(project.id).filter(licenseView)
+        val outboundLicenses = projectLicenseInfo.licenses.map { it.license }
+
+        // Define a rule that is executed for each license of the dependency.
+        licenseRule("OSADL project license compatibility", licenseView) {
+            outboundLicenses.forEach { outboundLicense ->
+                val compatibilityInfo = matrix
+                    // Be conservative and use the simple license string without the exception string for lookup.
+                    // Eventually resulting false-positives are better than an incompatibility to be missed.
```
mnonnenmacher, line 45: Do you mean "Potentially"?

mnonnenmacher (dismissed):

evaluator/src/main/resources/rules/osadl.rules.kts
fviernau: Thanks @LeChasseur  @sschuberth !

evaluator/src/main/resources/osadl/osadl.rules.kts
tsteenbe: IMO it should be in the source code as well OSADL is not known internationally that well and we as ORT maintainers should make clear on whose work we are building.

evaluator/src/main/kotlin/osadl/CompatibilityMatrix.kt
```diff
@@ -0,0 +1,51 @@
...
+)
+
+@Serializable
+data class CompatibilityInfo(
+    val name: String,
+    val compatibility: Compatibility,
+    val explanation: String
+)
+
+val matrix by lazy {
```
fviernau, line 44: I believe it would be nice to introduce a class for a compatibility matrix which internally 
holds a map, which provides one function to access the information of any cell.

That class could the have a companion object constant for the osadl matrix, which has the code for deserializing the `osadl-matrix-with-explanations.json`. So, the de-serialization / data classes would become private. That would minimize the API expose to the rules a bit.

Would that make sense to you as well?

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,76 @@
...
+// $ ort evaluate -i scanner/src/funTest/assets/dummy-expected-output-for-analyzer-result.yml --rules-resource /rules/osadl.rules.kts
+
+import org.ossreviewtoolkit.evaluator.osadl.Compatibility
+import org.ossreviewtoolkit.evaluator.osadl.matrix
+
+val ruleSet = ruleSet(ortResult, licenseInfoResolver) {
+    val licenseView = LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED
+
+    // Define a rule that is executed for each dependency.
+    dependencyRule("OSADL compliance matrix") {
```
tsteenbe, line 31: We always have used rules names as all caps with underscores (kind of enums) see https://github.com/oss-review-toolkit/ort/blob/main/examples/evaluator-rules/src/main/resources/example.rules.kts. Would like us to be consistent across the ORT project

evaluator/src/main/resources/rules/osadl.rules.kts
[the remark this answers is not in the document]
fviernau: Are you saying it should be handled, but not as part of this change or not as part of this rule?

Same for the case where `inbound` licenses is empty?

tsteenbe (changes requested):
Update to ORT's documentation is missing so this would basically be a hidden ORT feature. Is this intentional?

evaluator/src/main/resources/rules/osadl.rules.kts
sschuberth: No, I indeed mean "eventually" in the German sense of "schließlich".

evaluator/src/main/resources/rules/osadl.rules.kts
sschuberth: THB the convention to use as all caps with underscores here never made sense to me. IMO this should be more of a speaking value, similar to the rule violation messages themselves, so it also reads nicely which rule actually failed. I'd rather change the other existing rules to have speaking names, too.

sschuberth: > Update to ORT's documentation is missing so this would basically be a hidden ORT feature. Is this intentional?

This PR is pending already for way too long, and I don't want to hold it back further due to missing docs. I'll add them later.

fviernau: Using all caps (with underscores) has the advantage of being able to use the string as-is as input to a (helper-CLI command). Personally, I also like all caps more than free text from, because the textual representation is the "message". That message does not require further text as it stands on its own. So, in my view - the 'name' needs to be as easily / quickly recognizable as possible by the human brain. In my case this works better with all caps. So I vote for it.

evaluator/src/main/kotlin/osadl/CompatibilityMatrix.kt
sschuberth: I've introduced such a new class. Its lookup function now also handles the case of licenses that are not present in the matrix, incl. empty ones.

fviernau: Thanks!

evaluator/src/main/resources/rules/osadl.rules.kts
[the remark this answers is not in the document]
sschuberth: Empty licenses (and in general licenses that are not present in the matrix) are now gracefully handled by the new lookup function.

evaluator/src/main/resources/osadl/osadl.rules.kts
sschuberth: This conversation is continued [here](https://github.com/oss-review-toolkit/ort/pull/4977#discussion_r1000201431).

.reuse/dep5
```diff
@@ -106,6 +106,10 @@ Files: clients/github-graphql/src/main/assets/schema.graphql
 Copyright: 2017 Gregor Martynus
 License: MIT
 
+Files: evaluator/src/main/resources/osadl/osadl-matrix.csv
```
priv-kweihmann, line 109: I don't see this file in the change set anymore - it seems like it was replaced by osadl-matrix-with-explanations.json - so t guess this has to be updated here as well

evaluator/src/main/resources/rules/osadl.rules.kts
sschuberth: I've changed it to be able to move forward.

evaluator/src/main/resources/osadl/osadl.rules.kts
sschuberth: I've added a code co0mment also to the rules, beyond what's already documented as part of the newly added `CompatibilityMatrix` class.

.reuse/dep5
sschuberth: Thanks for pointing this out! Done.

fviernau (approved):

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,94 @@
...
+                    Compatibility.CONTEXTUAL -> warning(
+                        "Whether The outbound license $outboundLicenseString of project '$projectCoords' is " +
+                                "incompatible with the inbound license $inboundLicenseString of its dependency " +
+                                "'$depCoords' depends on the context. " + compatibilityInfo.explanation,
+                        "Get legal advice and eventually create a rule violation resolution."
+                    )
+
+                    Compatibility.UNKNOWN -> warning(
+                        "It is unknown whether the outbound license $outboundLicenseString of project " +
+                                "'$projectCoords' is incompatible with the inbound license $inboundLicenseString of " +
```
fviernau, line 76: nit: a tad simpler could be to say "is compatible" instead of "is incompatible"

evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -0,0 +1,94 @@
...
+                                "its dependency '$depCoords'. " + compatibilityInfo.explanation,
+                        "Get legal advice and eventually create a rule violation resolution."
```
fviernau, line 78: I believe using project specific rule violation resolutions is an unnecessarily expensive way to address this issue.
Ideally, the matrix would be updated but then it wouldn't be the OSADL matrix anymore. 

Can we do something to point the user in the right directing of trying to solve this globally?
Maybe it already helps a bit to say: "create a global rule violation resolution"?

fviernau (changes requested):

sschuberth:
@fviernau @mnonnenmacher @tsteenbe please have another look. Things I'd like to do in a follow.up PR:
1. Implement the [second form of check](https://github.com/oss-review-toolkit/ort/pull/4977#discussion_r997947353).
2. Add dedicated docs for this new feature.

fviernau (approved):

dgutson:
@sschuberth sorry I was confused with this issue. How can we use this?
CC @franco0700 again...

sschuberth:
See: https://github.com/oss-review-toolkit/ort/blob/218c55fd43854850e094066ca4869c984aee4ccc/evaluator/src/main/resources/rules/osadl.rules.kts#L20-L25

PR #4979, opened 2022-01-20
NuGetSupport: Use the homepage URL as fallback for the VCS URL

mnonnenmacher:
Use the homepage URL of a project as fallback URL for the VCS URL like
it is already done for other package managers.


sschuberth:
There's an empty file at `cli/src/main/kotlin/Curations.kt` being added...

sschuberth (approved):

PR #4981, opened 2022-01-20
Fix `ClearlyDefined`'s `Curation` class layout

sschuberth:


analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
```diff
@@ -145,7 +145,10 @@ class ClearlyDefinedPackageCurationProvider(
 
         contributedCurations.forEach { (_, contributions) ->
             contributions.curations.forEach inner@{ (coordinates, curation) ->
-                val pkgId = coordinatesToIds[coordinates] ?: return@inner
+                val pkgId = coordinatesToIds[coordinates] ?: run {
```
oheger-bosch, line 148: Just out of curiosity: Did this happen in the past that unexpected coordinates where returned?

oheger-bosch (dismissed):

analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
sschuberth: It started to happen when using bulk instead of single requests for package curations. So, doing multiple single requests vs. a single bulk request does not return the same information from ClearlyDefined 😨 See https://github.com/clearlydefined/service/issues/901.

oheger-bosch (approved):

PR #4982, opened 2022-01-21
Gradle: Apply the `gradle-graal` plugin to experiment with native images

sschuberth:
Run it like `./gradlew :cli:nativeImage`. While this does not succeed yet,
there seem to only be a few blockers to get it running, so still make
this publicly available for people to play with.


MarcelBochtler (approved):
LGTM and works apart from the unsupported elements mentioned in the commit message.

PR #4985, opened 2022-01-23
Introduce Batect as an easy way to run ORT functional test inside a Docker container

sschuberth:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.batect/
 .gradle/
 .idea/
 !.idea/icon*.png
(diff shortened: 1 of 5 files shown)


oheger-bosch (commented):
Looks promising. It is, however, not clear to me, how mature this already is. The latest commit message seems to indicate that only a subset of functional tests can be run. Should it describe the current state, i.e. what is supported and what not?

A documentation how to use this new method would be great, but I assume, this will come later.

sschuberth:
> The latest commit message seems to indicate that only a subset of functional tests can be run.

I just realized the last commit message is completely wrong, probably an oversight when squashing commits. I'll update.

oheger-bosch (approved):

PR #4987, opened 2022-01-24
GoMod: Strip a major version suffix from host path URLs

sschuberth:
Improve guessing the VCS information from package IDs like

    GoMod::github.com/lithammer/shortuuid/v3:v3.0.4

by stripping the "/v3" suffix from the path before passing the URL to
`VcsHost.toVcsInfo()`.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -347,6 +347,8 @@ private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>
     private fun dependencies(id: Identifier): Set<Identifier> = nodeMap[id].orEmpty()
 }
 
+private val GITHUB_NAME_REGEX = "(github\\.com/[^/]+/[^/]+)/v\\d".toRegex()
+
 private const val DATE_REVISION_PATTERN = "[\\d]{14}-(?<sha1>[0-9a-f]+)"
 
 // See https://golang.org/ref/mod#pseudo-versions.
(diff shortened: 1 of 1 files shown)


mnonnenmacher (dismissed):

MarcelBochtler (approved):

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -347,6 +347,8 @@ private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>
     private fun dependencies(id: Identifier): Set<Identifier> = nodeMap[id].orEmpty()
 }
 
+private val GITHUB_NAME_REGEX = "(github\\.com/[^/]+/[^/]+)/v\\d".toRegex()
+
```
fviernau, line 351: Why does this commit improve the guessing?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
sschuberth: Do you really mean *why*? Well, in order to avoid the need to write VCS curations in those cases.

If you meant *how*, well I hope to have detailed that in the commit message: Some GoMod packages have an additional "/v<version>" suffix in the GitHub path, which needs to gets stripped in order to make it a regular GitHub project path.

sschuberth: > but for some packages makes the situation worse.

For *which* kind of packages does the situation become worse? So far, the source code any id like `GoMod::github.com/lithammer/shortuuid/v3:v3.0.4` could *never* be found. With this change, the source code can be found. Maybe we're scanning too much in case of a monorepo as we're scanning the whole repo instead of only a subpath, but I believe that's still better than scanning nothing like before (which also would have required a curation).

At the example of https://github.com/lithammer/shortuuid/tree/v3.0.4 exactly the right thing gets scanned with this change.

fviernau: > For which kind of packages does the situation become worse?

For the ones which do have a `vx` directory the situation becomes worse, as you just confirmed. However, I do agree this change is good. What do you think about just mentioning that "for some packages a VCS path curation will become necessary" in the commit message? (and probably also mark it as a breaking change).

sschuberth: I've amended the commit message.

fviernau: Thanks!

fviernau (approved):

PR #4992, opened 2022-01-25
SpdxResolvedDocument: Copy over query strings when resolving URIs

sschuberth:
Resolving a path against a URI drops the original query string. However,
the query string might be required to get the right file, so copy an
existing query string over on resolution.


oheger-bosch:
The change looks good to me, but I would love to see a test case.

sschuberth:
> The change looks good to me, but I would love to see a test case.

I would have added a test case if you had already added a test spec for `SpdxExternalDocumentReference.resolve` 😉 Let me see what I can do...

oheger-bosch:
`SpdxResolvedDocumentTest` has test cases that also cover loading of documents from URIs. It should be possible to extend one of these.

oheger-bosch (approved):

PR #4993, opened 2022-01-25
ScanCode: Work around an underscore in a SPDX license key

fviernau:
ScanCode has one SPDX license key containing an underscore characters
which is not allwed, see [1]. This results in ORT's scanner crashing
due to an SpdxException when it tries to parse the SPDX license key.

This issue has first occured in 2020 and been fixed by [2]. It got
re-introduced recently by [3].

Deliberatly don't fix the general problem with underscores in
`getLicenseTextFile()` in favor of a license ID specific work around,
because this can be implemented efficiently without doing a refactoring
first.

[1] nexB/scancode-toolkit#2813
[2] fb0370f
[3] #4523

### DIFF
--- a/utils/spdx/src/main/kotlin/Utils.kt
+++ b/utils/spdx/src/main/kotlin/Utils.kt
@@ -162,11 +162,15 @@ private val LICENSE_REF_FILENAME_REGEX by lazy { Regex("^LicenseRef-\\w+-") }
 
 private fun getLicenseTextFile(id: String, dir: File): File? =
     id.replace(LICENSE_REF_FILENAME_REGEX, "").let { idWithoutLicenseRefNamespace ->
-        sequenceOf(
+        listOfNotNull(
             id,
             id.removePrefix("LicenseRef-"),
             idWithoutLicenseRefNamespace,
-            "$idWithoutLicenseRefNamespace.LICENSE"
+            "$idWithoutLicenseRefNamespace.LICENSE",
+            "x11-xconsortium_veillard.LICENSE".takeIf {
+                // Work around for https://github.com/nexB/scancode-toolkit/issues/2813.
+                id == "LicenseRef-scancode-x11-xconsortium-veillard"
+            }
         ).mapNotNull { filename ->
             dir.resolve(filename).takeIf { it.isFile }
         }.firstOrNull()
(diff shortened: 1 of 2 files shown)


sschuberth:
> Delierately don't refactor for efficiency because this fix can be reverted as soon as [3] is fixed.

Is that so? As we're now always using the licenses from the installed ScanCode version, I guess we'd need to keep this work-around even if we'd upgrade to a newer version of ScanCode, unless we're saying that we only support the version of ScanCode that ORT uses by default.

fviernau:
> Is that so? As we're now always using the licenses from the installed ScanCode version, I guess we'd need to keep this work-around even if we'd upgrade to a newer version of ScanCode, unless we're saying that we only support the version of ScanCode that ORT uses by default.

Oh, right - missed that. Do you think it's worth to refactor it for performance now? Or would you be fine keeping it as-is and do the refactoring in a following PR. (At least the scanner crash is fixed than a bit earlier)

sschuberth:
> Do you think it's worth to refactor it for performance now? Or would you be fine keeping it as-is and do the refactoring in a following PR. (At least the scanner crash is fixed than a bit earlier)

Hmm, given the search at https://scancode-licensedb.aboutcode.org/?search=_, `LicenseRef-scancode-x11-xconsortium_veillard` really is the *only* ID with an underscore. Which means the crash should happen only rarely, and thus there's no rush to fix it. But is also means we can probably do a simple & performant work-around that only hard-code this particular ID.

utils/spdx/src/main/kotlin/Utils.kt
```diff
@@ -160,17 +160,23 @@ private fun getLicenseTextResource(id: String): URL? =
...
-private fun getLicenseTextFile(id: String, dir: File): File? =
-    id.replace(LICENSE_REF_FILENAME_REGEX, "").let { idWithoutLicenseRefNamespace ->
-        sequenceOf(
-            id,
-            id.removePrefix("LicenseRef-"),
-            idWithoutLicenseRefNamespace,
-            "$idWithoutLicenseRefNamespace.LICENSE"
-        ).mapNotNull { filename ->
-            dir.resolve(filename).takeIf { it.isFile }
-        }.firstOrNull()
```
sschuberth, line 172: Maybe we could simply write this as

```
private fun getLicenseTextFile(id: String, dir: File): File? =
    id.replace(LICENSE_REF_FILENAME_REGEX, "").let { idWithoutLicenseRefNamespace ->
        listOfNotNull(
            id,
            id.removePrefix("LicenseRef-"),
            idWithoutLicenseRefNamespace,
            "$idWithoutLicenseRefNamespace.LICENSE",
            "LicenseRef-scancode-x11-xconsortium_veillard".takeIf {
                id == "LicenseRef-scancode-x11-xconsortium-veillard"
            }
        ).firstNotNullOfOrNull { filename ->
            dir.resolve(filename).takeIf { it.isFile }
        }
    }
```

fviernau: Nice idea, I've used it.

sschuberth (approved):

PR #4997, opened 2022-01-26
Improve the prioritization in `getRevisionCandidates()`

sschuberth:


mnonnenmacher (commented):
It's a bit unfortunate that we have no test case for this very important function that would show the difference your PR makes. Would you mind adding one?

sschuberth:
> It's a bit unfortunate that we have no test case for this very important function that would show the difference your PR makes. Would you mind adding one?

Done, finally, @mnonnenmacher.

mnonnenmacher (approved):

PR #5000, opened 2022-01-27
Small SPDX document resolution improvements

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/SpdxResolvedDocument.kt
+++ b/analyzer/src/main/kotlin/managers/utils/SpdxResolvedDocument.kt
@@ -272,8 +272,8 @@ internal fun SpdxExternalDocumentReference.resolve(
             baseUri,
             createAndLogIssue(
                 source = managerName,
-                message = "'$spdxDocument' defined in '$baseUri' as part of '$externalDocumentId' is not a valid " +
-                        "URI."
+                message = "The SPDX document at '$spdxDocument' cannot be resolved as a URI (referred from $baseUri " +
+                        "as part of '$externalDocumentId')."
             )
         )
     }
@@ -319,7 +319,7 @@ private fun SpdxExternalDocumentReference.resolveFromDownload(
     baseUri: URI,
     managerName: String
 ): ResolutionResult {
-    log.info { "Downloading '$uri' referred to by '$baseUri' in reference '$externalDocumentId'." }
+    log.info { "Downloading SPDX document from $uri (referred from $baseUri as part of '$externalDocumentId')." }
 
     val tempDir = createOrtTempDir()
     return try {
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/managers/utils/SpdxResolvedDocument.kt
```diff
@@ -272,8 +272,8 @@ internal fun SpdxExternalDocumentReference.resolve(
             baseUri,
             createAndLogIssue(
                 source = managerName,
-                message = "'$spdxDocument' defined in '$baseUri' as part of '$externalDocumentId' is not a valid " +
-                        "URI."
+                message = "The SPDX document '$spdxDocument' defined in '$baseUri' as part of '$externalDocumentId' " +
```
oheger-bosch, line 275: I find the resulting sentence a bit strange, since an SPDX document is not a URI. Should it better be something like "The SPDX document reference ... is not a valid URI."?

utils/spdx/src/main/kotlin/model/SpdxExternalDocumentReference.kt
```diff
@@ -48,6 +48,10 @@ data class SpdxExternalDocumentReference(
             "The external document ID must start with '${SpdxConstants.DOCUMENT_REF_PREFIX}'."
         }
 
-        require(spdxDocument.isNotBlank()) { "The SPDX document must not be blank." }
+        require(spdxDocument.isNotEmpty()) { "The SPDX document must not be empty." }
+
+        require(spdxDocument.trim() == spdxDocument) {
+            "The SPDX document must not contain any leading or trailing whitespace."
```
oheger-bosch, line 54: Ideally, this change should be accompanied by a test case.

analyzer/src/main/kotlin/managers/utils/SpdxResolvedDocument.kt
sschuberth: > since an SPDX document is not a URI.

Well, the `spdxDocument` field *is* defined to be a URI according to the spec. The field name also is not called `spdxDocumentReference` or so.

But if you prefer, I could write this as "The SPDX document *at* '$spdxDocument'...".

utils/spdx/src/main/kotlin/model/SpdxExternalDocumentReference.kt
sschuberth: We have a lot of such assertions in almost all SPDX model classes, and none of them is covered by explicit tests so far. So I don't feel like adding a test for this specific assertion right now, TBH.

oheger-bosch: Somewhere we have to start with improving test coverage. But I won't insist.

oheger-bosch (approved):

PR #5003, opened 2022-01-28
scanner: Add scannerDir to PATH after bootstrapping

### LINKED ISSUE #5002: Test cases of ScanCodeScannerFunTest failed
```
$ ./gradlew :scanner:funTest --tests org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest

> Configure project :
Building ORT version 64bfa05d8d.

> Task :scanner:funTest

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > Scanning a single file succeeds STARTED

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > Scanning a single file succeeds PASSED

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > Scanning a directory succeeds STARTED

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > Scanning a directory succeeds PASSED

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > return the full license text for the HERE proprietary license STARTED

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > return the full license text for the HERE proprietary license FAILED
    java.lang.AssertionError: Expecting actual not to be null
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$1.invokeSuspend(ScanCodeScannerFunTest.kt:43)
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$1.invoke(ScanCodeScannerFunTest.kt)
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$1.invoke(ScanCodeScannerFunTest.kt)

        Caused by:
        java.lang.AssertionError: Expecting actual not to be null
            at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$1.invokeSuspend(ScanCodeScannerFunTest.kt:43)
            ... 2 more

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > return the full license text for a known SPDX LicenseRef STARTED

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > return the full license text for a known SPDX LicenseRef FAILED
    java.lang.AssertionError: Expecting actual not to be null
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$2.invokeSuspend(ScanCodeScannerFunTest.kt:50)
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$2.invoke(ScanCodeScannerFunTest.kt)
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$2.invoke(ScanCodeScannerFunTest.kt)

        Caused by:
        java.lang.AssertionError: Expecting actual not to be null
            at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$2.invokeSuspend(ScanCodeScannerFunTest.kt:50)
            ... 2 more

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > return the ort license text for a license ID known by ort and also in custom dir STARTED

org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest > return the ort license text for a license ID known by ort and also in custom dir FAILED
    java.lang.AssertionError: Expected null but actual was "x
    y
    "
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$3.invokeSuspend(ScanCodeScannerFunTest.kt:60)
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$3.invoke(ScanCodeScannerFunTest.kt)
        at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$3.invoke(ScanCodeScannerFunTest.kt)

        Caused by:
        java.lang.AssertionError: Expected null but actual was "x
        y
        "
            at org.ossreviewtoolkit.scanner.scanners.scancode.ScanCodeScannerFunTest$3.invokeSuspend(ScanCodeScannerFunTest.kt:60)
            ... 2 more

5 tests completed, 3 failed
```

ldang0:
These test cases of ScanCodeScannerFunTest failed because ScanCode is bootstrapped and is not in PATH, so `getLicenseText` cannot find license files in directory under ScanCode.  Adding `scannerDir` to PATH after bootstrapped ScanCode can fix #5002

### DIFF
--- a/utils/common/src/main/kotlin/Os.kt
+++ b/utils/common/src/main/kotlin/Os.kt
@@ -119,6 +119,21 @@ object Os {
     }
 
     /**
+     * Prepend [path] to the system's PATH environment.
+     */
+    fun prependToPathEnvironment(path: File) {
+        val paths = env["PATH"]?.split(File.pathSeparatorChar)
+        val absolutePath = path.absolutePath
+        if (paths != null) {
+            if (absolutePath !in paths) {
+                env["PATH"] = absolutePath + File.pathSeparatorChar + env["PATH"]
+            }
+        } else {
+            env["PATH"] = absolutePath
+        }
+    }
+
+    /**
      * Resolve the Windows [executable] to its full name including the optional extension.
      */
     fun resolveWindowsExecutable(executable: File): File? {
--- a/scanner/src/main/kotlin/CommandLineScanner.kt
+++ b/scanner/src/main/kotlin/CommandLineScanner.kt
@@ -69,6 +69,7 @@ abstract class CommandLineScanner(
                                         "$expectedVersion."
                             )
                         }
+                        Os.prependToPathEnvironment(it)
                     }
                 }
 


sschuberth:
Thanks for the PR! But actually, it should not be necessary to add a bootstrapped ScanCode installation to `PATH`, because if ScanCode was already bootstrapped, the path to the bootstrapped installation will be returned:

https://github.com/oss-review-toolkit/ort/blob/64bfa05d8d90a00814919d73e659184ba0622d39/scanner/src/main/kotlin/scanners/scancode/ScanCode.kt#L147-L150

ldang0:
> Thanks for the PR! But actually, it should not be necessary to add a bootstrapped ScanCode installation to `PATH`, because if ScanCode was already bootstrapped, the path to the bootstrapped installation will be returned:
> 

However, if we did not prepend that bootstrapped ScanCode to `PATH`, then we need to find another way to let `scanCodeLicenseTextDir` find installation directory of this ScanCode, because currently it just search `scancode` in `PATH`, and cannot find it. That's why these test cases in `ScanCodeScannerFunTest` failed.

sschuberth:
> However, if we did not prepend that bootstrapped ScanCode to `PATH`, then we need to find another way to let `scanCodeLicenseTextDir` find installation directory of this ScanCode

Yeah, that's a good point that this is not about *running* ScanCode, but finding the license text in a ScanCode installation.

I vaguely remember that this has indeed been an open point, correct @mnonnenmacher?

It's not super-nice, but maybe we could extend `scanCodeLicenseTextDir` to not only look in `PATH`, but also in the path ORT would bootstrap to? What I don't like about this is that the independent `spdx-utils` then needs to know implementation details from the (unrelated) `scanner` module... from that perspective, adding the scanner's path to the `Os` classes' copy of the environment indeed does not sound like a too weird approach. We just have to keep in mind that this doe snot really modify the processes environment, but only the map which holds our internal copy.

More opinions @oss-review-toolkit/core-devs?

ldang0:
> > However, if we did not prepend that bootstrapped ScanCode to `PATH`, then we need to find another way to let `scanCodeLicenseTextDir` find installation directory of this ScanCode
> 
> Yeah, that's a good point that this is not about _running_ ScanCode, but finding the license text in a ScanCode installation.
> 
> I vaguely remember that this has indeed been an open point, correct @mnonnenmacher?
> 
> It's not super-nice, but maybe we could extend `scanCodeLicenseTextDir` to not only look in `PATH`, but also in the path ORT would bootstrap to? What I don't like about this is that the independent `spdx-utils` then needs to know implementation details from the (unrelated) `scanner` module

Another approach to this issue, which better IMHO, is to move all these ScanCode related code from `org.ossreviewtoolkit.utils.spdx` to `org.ossreviewtoolkit.scanner.scanners.scancode`, and call `getLicenseText` properly in `ScanCodeScannerFunTest` (and maybe other places).

utils/common/src/main/kotlin/Os.kt
```diff
@@ -118,6 +118,13 @@ object Os {
         }
     }
 
+    /**
+     * Prepend [path] to the system's PATH environment.
+     */
+    fun prependToPathEnvironment(path: File) {
+        env["PATH"] = path.toString() + env["PATH"]?.let { File.pathSeparatorChar + it }
```
sschuberth, line 125: Won't this append a literal "null" string if "PATH" was unset before?

Also, should we maybe add a sanity check to only prepend if `path` is not already the first item?

sschuberth (changes requested):

utils/common/src/main/kotlin/Os.kt
ldang0: > Won't this append a literal "null" string if "PATH" was unset before?

Unfortunately, yes.

> 
> Also, should we maybe add a sanity check to only prepend if `path` is not already the first item?

Good point. 

How about this ?

```
    fun prependToPathEnvironment(path: File) {
        val paths = env["PATH"]?.split(File.pathSeparatorChar)
        val absolutePath = path.absolutePath
        if (paths != null) {
            if (absolutePath !in paths) {
                env["PATH"] = absolutePath + File.pathSeparatorChar + env["PATH"]
            }
        } else {
            env["PATH"] = absolutePath
        }
    }
```

ldang0:
This commit has nothing to do with this error
```
org.ossreviewtoolkit.analyzer.managers.NpmVersionUrlFunTest > NPM should > org.ossreviewtoolkit.analyzer.managers.NpmVersionUrlFunTest.resolve dependencies with URLs as versions correctly
```

https://dev.azure.com/oss-review-toolkit/fa9881bf-9fc2-4a4f-8305-803f4455d3c6/_apis/build/builds/6973/logs/65

PR #5005, opened 2022-01-28
Improve startup performance

mnonnenmacher:
Reduce the duration of running `ort --help` from ~13s to ~2s, see the commit messages for details.

### DIFF
--- a/reporter/src/main/kotlin/reporters/freemarker/asciidoc/AsciiDocTemplateReporter.kt
+++ b/reporter/src/main/kotlin/reporters/freemarker/asciidoc/AsciiDocTemplateReporter.kt
@@ -58,7 +58,7 @@ abstract class AsciiDocTemplateReporter(private val backend: String, override va
         ASCII_DOC_FILE_EXTENSION,
         ASCII_DOC_TEMPLATE_DIRECTORY
     )
-    private val asciidoctor = Asciidoctor.Factory.create()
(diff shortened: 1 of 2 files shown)


sschuberth (approved):

reporter/src/main/kotlin/reporters/freemarker/asciidoc/AsciiDocTemplateReporter.kt
```diff
@@ -58,7 +58,7 @@ abstract class AsciiDocTemplateReporter(private val backend: String, override va
         ASCII_DOC_FILE_EXTENSION,
         ASCII_DOC_TEMPLATE_DIRECTORY
     )
-    private val asciidoctor = Asciidoctor.Factory.create()
+    private val asciidoctor by lazy { Asciidoctor.Factory.create() }
```
oheger-bosch, line 61: A short Google search seems to indicate that _Asciidoctor_ is thread-safe, e.g. [1]. So, do you think it would be possible to use a single shared instance for all reporter instances?

[1] https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/326

mnonnenmacher: That might be an option, I think we would have to give it a try.

PR #5010, opened 2022-01-31
FossIdScanResults: Replace problematic FossID licenses with valid expressions

MarcelBochtler:
Parenthesis are not allowed in license expressions, resulting in an
exception when trying to parse these with `toSpdx()`.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
@@ -26,6 +26,34 @@ import org.ossreviewtoolkit.clients.fossid.model.summary.Summarizable
 import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.TextLocation
+import org.ossreviewtoolkit.utils.spdx.toSpdx
+
+/**
+ * Collection of problematic FossID license expressions. Compiled from FossID's list of licenses. May be updated if
+ * FossID adds new licenses.
+ */
+private val fossIdLicenseMappings = mapOf(
+    "BSD (Three Clause License)" to "BSD-3-clause".toSpdx(),
+    "BSD-Acknowledgment-(Carrot2)" to "LicenseRef-scancode-bsd-ack-carrot2".toSpdx(),
+    "Freeware-Public-(FPL)" to "LicenseRef-scancode-fpl".toSpdx(),
+    "MediaInfo(Lib)" to "LicenseRef-scancode-mediainfo-lib".toSpdx(),
+    "Oracle-Master-Agreement-(OMA)" to "LicenseRef-scancode-oracle-master-agreement".toSpdx(),
+    "Things-I-Made-(TIM)-Public" to "LicenseRef-scancode-things-i-made-public-license".toSpdx(),
+    "X11-Style-(Adobe)" to "LicenseRef-scancode-x11-adobe".toSpdx(),
+    "X11-Style-(Adobe-DEC)" to "LicenseRef-scancode-x11-adobe-dec".toSpdx(),
+    "X11-Style-(Bitstream-Charter)" to "LicenseRef-scancode-x11-bitstream".toSpdx(),
+    "X11-Style-(DEC-1)" to "LicenseRef-scancode-x11-dec1".toSpdx(),
+    "X11-Style-(DEC-2)" to "LicenseRef-scancode-x11-dec2".toSpdx(),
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -26,6 +26,34 @@ import org.ossreviewtoolkit.clients.fossid.model.summary.Summarizable
...
+    "MediaInfo(Lib)" to "LicenseRef-scancode-mediainfo-lib".toSpdx(),
+    "Oracle-Master-Agreement-(OMA)" to "LicenseRef-scancode-oracle-master-agreement".toSpdx(),
+    "Things-I-Made-(TIM)-Public" to "LicenseRef-scancode-things-i-made-public-license".toSpdx(),
+    "X11-Style-(Adobe)" to "LicenseRef-scancode-x11-adobe".toSpdx(),
+    "X11-Style-(Adobe-DEC)" to "LicenseRef-scancode-x11-adobe-dec".toSpdx(),
+    "X11-Style-(Bitstream-Charter)" to "LicenseRef-scancode-x11-bitstream".toSpdx(),
+    "X11-Style-(DEC-1)" to "LicenseRef-scancode-x11-dec1".toSpdx(),
+    "X11-Style-(DEC-2)" to "LicenseRef-scancode-x11-dec2".toSpdx(),
+    "X11-Style-(DSC-Technologies)" to "LicenseRef-scancode-x11-dsc".toSpdx(),
+    "X11-Style-(David-R.-Hanson)" to "LicenseRef-scancode-x11-hanson".toSpdx(), // TODO
```
sschuberth, line 48: *What's* to do here?

mnonnenmacher (approved):

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
MarcelBochtler: removed

sschuberth:
The commit message is a bit misleading IMO, as in general parentheses *are* allowed in SPDX (compound) expressions.

MarcelBochtler:
> The commit message is a bit misleading IMO, as in general parentheses _are_ allowed in SPDX (compound) expressions.

I rephrased the message a bit to be more precise.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -26,6 +26,34 @@ import org.ossreviewtoolkit.clients.fossid.model.summary.Summarizable
 import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.TextLocation
+import org.ossreviewtoolkit.utils.spdx.SpdxExpression
+import org.ossreviewtoolkit.utils.spdx.toSpdx
+
+/**
+ * Collection of problematic FossID license expressions.
+ */
+private val fossIdLicenseMappings: HashMap<String, SpdxExpression> = hashMapOf(
```
sschuberth, line 35: Why not just 'mapOf<String, SpdxExpression>(...)'?

sschuberth (dismissed):

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -26,6 +26,33 @@ import org.ossreviewtoolkit.clients.fossid.model.summary.Summarizable
 import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.TextLocation
+import org.ossreviewtoolkit.utils.spdx.toSpdx
+
+/**
+ * Collection of problematic FossID license expressions.
```
oheger-bosch, line 32: It would be helpful to have some information in the comment from where these licenses come (how was the list compiled) and whether it might have to be updated for new FossID versions (if this is the case).

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -26,6 +26,33 @@ import org.ossreviewtoolkit.clients.fossid.model.summary.Summarizable
 import org.ossreviewtoolkit.model.CopyrightFinding
```
oheger-bosch, line 26: Commit message: I would not mention parantheses at all, but just state that FossID can return a number of licenses whose identifiers cannot be parsed to SPDX expressions, and, therefore, a mapping is needed.

oheger-bosch (approved):

PR #5013, opened 2022-02-01
NuGet: Parse more information

mnonnenmacher:
Parse more information from DotNet and NuGet definition files, as a preparation for improving the NuGet dependency resolution. See the commit messages for details.

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/dotnet-expected-output-with-nuspec.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/dotnet-expected-output-with-nuspec.yml
@@ -24,11 +24,45 @@ project:
   scopes:
   - name: "dependencies"
     dependencies:
+    - id: "NuGet::System.Globalization:4.3.0"
+      dependencies:
+      - id: "NuGet::Microsoft.NETCore.Platforms:1.1.0"
+      - id: "NuGet::Microsoft.NETCore.Targets:1.1.0"
+      - id: "NuGet::System.Runtime:4.3.0"
+    - id: "NuGet::System.Threading:4.0.11"
+      dependencies:
+      - id: "NuGet::System.Runtime:4.1.0"
+        dependencies:
+        - id: "NuGet::Microsoft.NETCore.Platforms:1.0.1"
+        - id: "NuGet::Microsoft.NETCore.Targets:1.0.1"
+      - id: "NuGet::System.Threading.Tasks:4.0.11"
+        dependencies:
+        - id: "NuGet::Microsoft.NETCore.Platforms:1.0.1"
+        - id: "NuGet::Microsoft.NETCore.Targets:1.0.1"
+        - id: "NuGet::System.Runtime:4.1.0"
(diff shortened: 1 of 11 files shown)


analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -286,8 +286,15 @@ private fun resolveLocalSpec(definitionFile: File): File? =
 private fun getIdentifier(name: String, version: String) =
     Identifier(type = "NuGet", namespace = "", name = name, version = version)
 
+data class NuGetDependency(
```
oheger-bosch, line 289: It would be nice to have a comment on this class about its purpose; that it serves as a NuGet-specific version of the `Identifier` class.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -304,7 +311,8 @@ fun PackageManager.resolveNuGetDependencies(
     val issues = mutableListOf<OrtIssue>()
 
     val references = reader.getPackageReferences(definitionFile)
-    support.buildDependencyTree(references, dependencies, packages, issues)
+    val ids = references.map { Identifier(type = "NuGet", namespace = "", name = it.name, version = it.version) }
```
oheger-bosch, line 314: Maybe implement the creation of an `Identifier` from a `NutGetDependency` as an extension function on the latter?

analyzer/src/funTest/assets/projects/synthetic/dotnet-expected-output-with-nuspec.yml
```diff
@@ -24,11 +24,45 @@ project:
   scopes:
```
oheger-bosch, line 24: Just a bit confused that the last commit in this PR is a demonstration of issues. But looking forward to the follow-up PR with the fix.

oheger-bosch (approved):

mnonnenmacher:
Merged despite failing Azure pipeline, all relevant tests were verified locally.

analyzer/src/main/kotlin/managers/DotNet.kt
```diff
@@ -70,33 +70,61 @@ class DotNet(
...
-        return itemGroups.flatMapTo(mutableSetOf()) { itemGroup ->
-            itemGroup.packageReference?.map { packageReference ->
-                NuGetDependency(
-                    name = packageReference.include,
-                    version = packageReference.version,
-                    targetFramework = "",
-                    developmentDependency = false
-                )
-            }.orEmpty()
+        val targetFrameworks = propertyGroups.find { it.targetFramework != null }?.targetFramework?.let { listOf(it) }
```
sschuberth, line 109: @mnonnenmacher is it really correct here that only the first non-null `targetFramework` from the list of `propertyGroups` is taken into account, and any `targetFramework` from remaining `propertyGroups` is ignored?

mnonnenmacher: Yes, it can only be defined once. If multiple target frameworks they have to be separated by ";".

PR #5016, opened 2022-02-02
Disable Sonatype Lift's JavaScript checks

tsteenbe:
ESLint is already run as part of ORT's build and Lift is returning tens of
false-positives in pull requests see for example [1].

[1]: https://github.com/oss-review-toolkit/ort/pull/5012

### DIFF
--- a/.lift.toml
+++ b/.lift.toml
@@ -2,7 +2,7 @@
 
 jdk11 = true
 
-disableTools = [ "detekt" ]
+disableTools = [ "detekt", "eslint" ]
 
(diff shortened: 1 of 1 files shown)


.lift.toml
```diff
@@ -2,7 +2,7 @@
 
 jdk11 = true
 
-disableTools = [ "detekt" ]
+disableTools = [ "detekt", eslint ]
```
sschuberth, line 5: Please put the string in quotes.

PR #5035, opened 2022-02-10
spdx-utils: Ignore the '+' operator for SPDX license texts

### LINKED ISSUE #5004: License texts for `+` variants of SPDX licenses
ORT currently provides license text for license identifiers on the [SPDX license list](https://spdx.org/licenses/) for notice generation with Reporter. No license text is provided for SPDX licenses with `+` identifier; for example, a license expression `LPPL-1.0+` does not produce any license text for the notice files.

Should the non-plus variant of the license be used for the plus variants?

mnonnenmacher:
Ignore the '+' operator when getting the license text of an SPDX license.
For example, for the id 'Apache-2.0+' return the license text for
'Apache-2.0'. This is valid, because in the example 'Apache-2.0' is a
valid license for the expression.

Choosing a later version of the license should be implemented as an
extension of the license choice feature.

Fixes #5004.

### DIFF
--- a/utils/spdx/src/test/kotlin/UtilsTest.kt
+++ b/utils/spdx/src/test/kotlin/UtilsTest.kt
@@ -142,6 +142,13 @@ class UtilsTest : WordSpec() {
                 text should endWith("limitations under the License.")
             }
 
+            "return the full license text for a valid SPDX license id with the '+' operator" {
+                val text = getLicenseText("Apache-2.0+")?.trim()
+
+                text should startWith("Apache License")
+                text should endWith("limitations under the License.")
+            }
+
             "return null for an invalid SPDX license id" {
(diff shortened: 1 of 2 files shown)


utils/spdx/src/test/kotlin/UtilsTest.kt
```diff
@@ -142,6 +142,13 @@ class UtilsTest : WordSpec() {
                 text should endWith("limitations under the License.")
             }
 
+            "return the full license text for a valid SPDX license id with the or later flag set" {
```
sschuberth, line 145: Nits:
- Maybe single-quote `'or later'`?
- SPDX itself calls `+` an "operator", not a "flag", so we should use the same wording here.

sschuberth (changes requested):

sschuberth (approved):

PR #5044, opened 2022-02-11
helper-cli: Add a command to download results from Postgres

fviernau:
The `cli` module provides the UploadResultToPostgresCommand. Add a
command to retrieve the previously uploaded commands. This can be useful
to synchronize a local directory with the database incrementally, in
order to play around with the latest data on the local machine.

Use md5sums to detect broken files, so that the download is gracefully
resumed after a hard exit or a connectivity issue. The need for md5sums
prevented the re-use of the LocalFileStorage class.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
@@ -1,0 +1,189 @@
+/*
+ * Copyright (C) 2022 HERE Europe B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.associate
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.file
+
+import java.io.File
+import java.sql.Connection
+
+import kotlin.time.measureTime
+
+import org.apache.commons.compress.compressors.xz.XZCompressorOutputStream
+
+import org.ossreviewtoolkit.helper.common.ORTH_NAME
+import org.ossreviewtoolkit.model.HashAlgorithm
+import org.ossreviewtoolkit.model.config.OrtConfiguration
+import org.ossreviewtoolkit.model.config.PostgresStorageConfiguration
+import org.ossreviewtoolkit.model.utils.DatabaseUtils
+import org.ossreviewtoolkit.utils.common.expandTilde
+import org.ossreviewtoolkit.utils.core.ORT_CONFIG_FILENAME
+import org.ossreviewtoolkit.utils.core.ortConfigDirectory
+
+class DownloadResultsFromPostgresCommand : CliktCommand(
+    name = "download-results-from-postgres",
+    help = "Download an ORT result from a PostgreSQL database. The symmetric command to ORT's " +
+            " upload-result-to-postgres command."
+) {
+    private val outputDir by option(
+        "--output-dir", "-o",
+        help = "The directory to write the ORT results to."
+    ).convert { it.expandTilde() }
+        .file(mustExist = false, canBeFile = false, canBeDir = true, mustBeWritable = false, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+
+    private val tableName by option(
+        "--table-name",
+        help = "The name of the table which holds the ORT results."
+    ).required()
+
+    private val columnName by option(
+        "--column-name",
+        help = "The name of the JSONB column containing the ORT result."
+    ).required()
+
+    private val configFile by option(
+        "--config",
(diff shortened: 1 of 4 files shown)


helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
```diff
@@ -0,0 +1,204 @@
...
+        // the gzip extension can help speed this up https://github.com/pramsey/pgsql-gzip.
+
+        val query = """
+        SELECT 
+            t.$columnName::TEXT AS result
+        FROM 
+            $tableName t 
+        WHERE
+            id = ?;
+        """.trimMargin()
```
sschuberth, line 126: You're not using any margin marker, so should this be `trimIndent` instead? Also, the query string (lines 120 to 125) should probably be indented.

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
```diff
@@ -0,0 +1,204 @@
...
+
+private class OrtResultStorage(private val storageDir: File) {
+    fun getIds(): Set<Int> {
+        val ids = storageDir.listFiles { file, _ -> file.isDirectory }.mapNotNull {
+            try {
+                it.name.toInt()
+            } catch (e: NumberFormatException) {
+                null
+            }
+        }
```
sschuberth, line 176: Maybe better use `runCatching` / `getOrNull`?

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
fviernau: Just noticed that it fits in one line...

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
```diff
@@ -0,0 +1,193 @@
...
+            resultSet.getString("result")
+        } else {
+            null
+        }
+    }
+
+    private fun fetchRowIds(connection: Connection): Set<Int> {
+        val query = """
+            SELECT t.id FROM $tableName t WHERE id >= ?;
+        """.trimIndent()
```
sschuberth, line 133: I guess this then also fits into a single line...

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
mnonnenmacher: Or even better use `toIntOrNull()`.

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
```diff
@@ -0,0 +1,189 @@
...
+    override fun run() {
+        if (!outputDir.exists()) outputDir.mkdirs()
+
+        val localStorage = OrtResultStorage(outputDir)
+        val localRowIds = localStorage.getIds()
+
+        openDatabaseConnection().use { connection ->
+            val remoteRowIds = fetchRowIds(connection)
+            val missingRowIds = remoteRowIds - localRowIds
+            println("Requested the download of ${remoteRowIds.size} of which ${missingRowIds.size} are missing.")
```
mnonnenmacher, line 99: "results" or "rows" is missing after `${remoteRowIds.size}`.

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
```diff
@@ -0,0 +1,189 @@
...
+            missingRowIds.forEachIndexed { i, rowId ->
+                println("[${i + 1} / ${missingRowIds.size}] downloading row $rowId")
+
+                val duration = measureTime {
+                    val ortResultJson = fetchOrtResult(connection, rowId)
+
+                    if (ortResultJson != null) localStorage.add(rowId, ortResultJson)
+                }
+
+                println("[${i + 1} / ${missingRowIds.size}] took ${duration.toString(DurationUnit.SECONDS)}s.")
```
mnonnenmacher, line 110: If there is no requirement to always print the duration in seconds you could use the default `toString()` which chooses matching units automatically.

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
```diff
@@ -0,0 +1,189 @@
...
+            }
+        }
+    }
+
+    private fun fetchOrtResult(connection: Connection, rowId: Int): String? {
+        // TODO: The transmission is relatively slow currently due to lack of compression. Check whether
+        // the gzip extension can help speed this up https://github.com/pramsey/pgsql-gzip.
+
+        val query = "SELECT t.$columnName::TEXT AS result FROM $tableName t WHERE id = ?;"
```
mnonnenmacher, line 119: If you would reuse the `OrtResults` table from the upload command you could get rid of the SQL related boilerplate code.

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    private fun openDatabaseConnection(): Connection {
+        val storageConfig = OrtConfiguration.load(configArguments, configFile).scanner.storages?.values
+            ?.filterIsInstance<PostgresStorageConfiguration>()?.firstOrNull()
+            ?: throw IllegalArgumentException("postgresStorage not configured.")
+
+        return DatabaseUtils.createHikariDataSource(
+            config = storageConfig,
+            applicationNameSuffix = ORTH_NAME,
```
mnonnenmacher, line 153: You could make a separate commit and align this in the upload command.

fviernau: The upload command is a sub command of `ort`, while this one is a sub command `orth`.

helper-cli/src/main/kotlin/commands/DownloadResultsFromPostgresCommand.kt
fviernau: I'd keep this for now, in particular as I hope to make some PG specific optimizations, see comment above.

mnonnenmacher (dismissed):

sschuberth:
Sorry @fviernau, I clicked on the wrong button and updated this via a merge instead of a rebase. I'll fix this.

sschuberth (dismissed):

sschuberth (approved):

PR #5047, opened 2022-02-11
ExperimentalScanner: Always scan the full repository

mnonnenmacher:
Always scan the full repository, even if a VCS path is set. This makes
the initial scan slower if a VCS path is set, because more files are
scanned, but it also has several advantages:

* There is exactly one scan result in the storage for each repository
  and revision, not multiple for different VCS paths.
* Changing the VCS path via a curation does not require a new scan,
  instead the existing result can be reused and filtered appropriately.
* Only one scan of a repository revision is required, no matter how many
  packages where published from it.

### DIFF
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -167,7 +167,7 @@ class ExperimentalScanner(
         log.info { "Resolving nested provenances for ${controller.packages.size} packages." }
 
         val duration = measureTime {
-            controller.getPackageProvenances().forEach { provenance ->
+            controller.getPackageProvenancesWithoutVcsPath().forEach { provenance ->
                 runCatching {
                     nestedProvenanceResolver.resolveNestedProvenance(provenance)
                 }.onSuccess { nestedProvenance ->
@@ -298,7 +298,7 @@ class ExperimentalScanner(
 
     private fun readStoredResults(controller: ScanController) {
         log.info {
-            "Reading stored scan results for ${controller.getPackageProvenances().size} packages with " +
+            "Reading stored scan results for ${controller.getPackageProvenancesWithoutVcsPath().size} packages with " +
                     "${controller.getAllProvenances().size} provenances."
         }
 
--- a/scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
+++ b/scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
@@ -27,6 +27,7 @@ import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.KnownProvenance
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
 import org.ossreviewtoolkit.model.ScanSummary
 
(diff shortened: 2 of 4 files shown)


sschuberth:
> * packages where published from it.

Nit: "were"

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -594,3 +604,15 @@ private fun Map<ScannerWrapper, Map<KnownProvenance, List<ScanResult>>>.hasResul
...
 ) = getValue(scanner)[provenance].let { it != null && it.isNotEmpty() }
+
+private fun Map<Package, NestedProvenanceScanResult>.filterByVcsPath(
+    packageProvenances: Map<Package, Provenance>
+): Map<Package, NestedProvenanceScanResult> =
+    mapValues { (pkg, nestedProvenanceScanResult) ->
+        val path = when (val packageProvenance = packageProvenances.getValue(pkg)) {
+            is RepositoryProvenance -> packageProvenance.vcsInfo.path
+            else -> ""
+        }
```
sschuberth, line 615: Maybe instead:

    val path = (packageProvenances.getValue(pkg) as? RepositoryProvenance)?.vcsInfo?.path.orEmpty()

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -141,15 +142,24 @@ class ExperimentalScanner(
             "Resolved provenance for ${packages.size} packages in $packageProvenanceDuration."
         }
 
+        // Remove all VCS paths from the provenances so that always the full repositories are scanned. The VCS paths
+        // will be used after scanning to correctly filter the scan results.
+        val packageProvenancesWithoutVcsPath = packageProvenances.mapValues { (_, provenance) ->
+            when (provenance) {
+                is RepositoryProvenance -> provenance.copy(vcsInfo = provenance.vcsInfo.copy(path = ""))
```
fviernau, line 149: Does this need a case distinction for 'Git-Repo'? (Or are the Git repositories of Git-Repo projects scanned already separately?)

fviernau (changes requested):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: Yes, but I don't want to add more special handling for git-repo, instead I have created #5142 as discussed some time ago.

fviernau: Nice!

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -136,4 +137,70 @@ data class NestedProvenanceScanResult(
...
+     * This also updates the VCS paths for all [RepositoryProvenance]s accordingly to make clear that they are not
+     * results for the whole repository.
+     *
+     * Throws an [IllegalArgumentException] if any provenance in [nestedProvenance] already has a VCS path set.
+     */
+    fun filterByVcsPath(path: String): NestedProvenanceScanResult {
+        if (path.isBlank()) return this
+
+        require(getProvenances().all { it !is RepositoryProvenance || it.vcsInfo.path.isBlank() }) {
+            "filterByVcsPath() requires that all provenances have no VCS path set."
```
sschuberth, line 153: I wouldn't quote the function name here. Maybe better:

    "Can only filter repository / generic provenances by VCS path that were not created for a specific VCS path."

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -136,4 +137,70 @@ data class NestedProvenanceScanResult(
...
+        }
+
+        val pathsWithinProvenances = getProvenances().filter {
+            val provenancePath = getPath(it)
+            provenancePath.isBlank() || provenancePath.startsWith("$path/")
```
sschuberth, line 158: Can you please add a code comment why you allow blank paths here? (I guess because root license files need to be considered in any case.)

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -136,4 +137,70 @@ data class NestedProvenanceScanResult(
...
+        }.associateWith { provenance ->
+            val provenancePath = getPath(provenance)
+            path.removePrefix("$provenancePath/")
+        }
+
+        fun KnownProvenance.withVcsPath() =
+            when (this) {
+                is RepositoryProvenance -> {
+                    val pathWithinProvenance = pathsWithinProvenances.getValue(this)
+                    this.copy(vcsInfo = vcsInfo.copy(path = pathWithinProvenance))
```
sschuberth, line 168: Drop `this.`.

scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -46,20 +47,31 @@ class ScanController(
...
 
     /**
-     * A map of package [Identifier]s to their resolved [KnownProvenance]s.
+     * A map of package [Identifier]s to their resolved [KnownProvenance]s. These provenances are used to filter the
+     * scan results for a package based on the VCS path.
      */
     private val packageProvenances = mutableMapOf<Identifier, KnownProvenance>()
 
+    /**
+     * A map of package [Identifier]s to their resolved [KnownProvenance]s with VCS path removed. These provenances are
```
sschuberth, line 56: Nit: "the VCS path"

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
mnonnenmacher: The code was actually wrong, with the fixed code it's hopefully more obvious that the blank path is treated separately, because paths of findings are not prefixed with "/":
```provenancePath.isBlank() || path.startsWith("$provenancePath/")```

File: scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt [hunk 7]
```diff
@@ -136,4 +137,75 @@ data class NestedProvenanceScanResult(
...
+     * Throws an [IllegalArgumentException] if any provenance in [nestedProvenance] already has a VCS path set.
+     */
+    fun filterByVcsPath(path: String): NestedProvenanceScanResult {
+        if (path.isBlank()) return this
+
+        val provenances = getProvenances()
+        val provenancesWithVcsPath = provenances.filter { it is RepositoryProvenance && it.vcsInfo.path.isNotBlank() }
+
+        require(provenancesWithVcsPath.isEmpty()) {
+            "Cannot filter scan results by VCS path that have a repository provenance with a non-blank VCS path, " +
```
sschuberth, line 156: I'd drop the comma here.

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -136,4 +137,75 @@ data class NestedProvenanceScanResult(
...
+                    "because the partial scan result might not contain the path to filter by, but the following " +
```
sschuberth, line 157: Maybe "their partial scan result" to emphasize you're referring to "repository provenance"?

sschuberth: And to not let the sentence too long, I'd simply start a new one "The following provenances...".

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -136,4 +137,75 @@ data class NestedProvenanceScanResult(
...
+     *
+     * Throws an [IllegalArgumentException] if any provenance in [nestedProvenance] already has a VCS path set.
+     */
+    fun filterByVcsPath(path: String): NestedProvenanceScanResult {
+        if (path.isBlank()) return this
+
+        val provenances = getProvenances()
+        val provenancesWithVcsPath = provenances.filter { it is RepositoryProvenance && it.vcsInfo.path.isNotBlank() }
+
+        require(provenancesWithVcsPath.isEmpty()) {
```
fviernau, line 155: Would it make sense to make this an invariant of `NestedProvenance` instead? e.g. move the check into that class?

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -136,4 +137,75 @@ data class NestedProvenanceScanResult(
...
+                else -> this
+            }
+
+        // Set the VCS path of all provenances according to the provided path and filter provenances which are not
+        // within path.
+        val newNestedProvenance = NestedProvenance(
+            root = nestedProvenance.root.withVcsPath(),
+            subRepositories = nestedProvenance.subRepositories.filterValues { it in pathsWithinProvenances.keys }
+                .mapValues { (_, provenance) -> provenance.withVcsPath() as RepositoryProvenance }
+        )
```
fviernau, line 185: Can you help me understanding? ...why it is necessary to set the VCS path in the returned NestedProvenanceScanResult?

If the VCS path of the root provenance points into a subdirectory of a submodule, this is the case when it is done IIUC. If the goal is to scan the full repo no matter of the VCS path, why is setting this path needed?

fviernau (changes requested):

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
mnonnenmacher: All applied.

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
mnonnenmacher: Maybe later at some point, but currently the `NestedProvenance` must be allowed, see my response to your next comment.

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
fviernau: Thanks!

fviernau (approved):

sschuberth (approved):

PR #5057, opened 2022-02-14
Add note for using subdirectories with docker

tafli:


docs/hints-for-use-with-docker.md
```diff
@@ -12,7 +12,9 @@ docker run \
   analyze [...] # Insert further arguments for the command.
 ```
 
-**Note:** The single forward slash `/` between the environment variable `$PWD` and the `:` is required for PowerShell compatibility, as PowerShell otherwise interprets `:` as part of the environment variable. 
+**Note:** The single forward slash `/` between the environment variable `$PWD` and the `:` is required for PowerShell compatibility, as PowerShell otherwise interprets `:` as part of the environment variable.
+
+**Note:** When using Yarn/NPM as a projects package manager and mounting a directory relative to the root directory, make sure to **not** have a [repository](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#repository) defined in your `package.json`. As the subdirectory is not recogniced as a VCS managed directory, the repository information cannot be written to the analyzer result. Generating a report will lead to an error.
```
sschuberth, line 17: I believe this note goes exactly into the wrong direction: Instead of omitting VCS information from `package.json`, you should always mount the root of the project into docker, and eventually adjust the input directory passed to the analyzer to match the subdirectory instead. The point is that ORT needs to be able to "see" the `.git` directory, even if analysis is limited to a subdirectory of the Git working tree.

sschuberth (changes requested):

docs/hints-for-use-with-docker.md
tafli: Absolutely. In this case the title is misleading and the "or directory relative to working directory" part must be removed. In this case my note is obsolete.

With "adjust the input directory" you mean using the `-o`flag?

sschuberth: No, as this is about the *input* directory, you need to adjust the `-i` option, not the `-o` option.

Let me try to give an example. If on the host you have this directory structure:

```
development
|
+-project-root
  |
  +-.git
  |
  +-sub-project
```

And what you want to get analyzed is only `sub-project`, then you still need to mount `project-root` to Docker, like e.g. `-v /development/project-root:/workdir`, and the analyzer needs to be invoked with `ort analyze -i /workdir/sub-project -o ...`.

tafli: > No, as this is about the _input_ directory, you need to adjust the `-i` option, not the `-o` option.

Haha, my bad. Sure it's the `-i`option. Small typo, huge effect.

tafli: Remove note and clarification for working on subprojects

docs/hints-for-use-with-docker.md
```diff
@@ -12,7 +12,18 @@ docker run \
...
 
-**Note:** The single forward slash `/` between the environment variable `$PWD` and the `:` is required for PowerShell compatibility, as PowerShell otherwise interprets `:` as part of the environment variable. 
+Make sure to mount the project root directory so ORT is able to detect VCS information. To work on a subproject use the option `-i` to specify the actual working directory:
+
+```shell
+docker run \
+  -v $PWD/:/project \ # Mount current working directory into /project to use as input.
+  ort --info \
+  -c /project/ort/config.hocon \ # Use file from "<workingdirectory>/ort" as config.
+  -i /project/subproject \ # Specify working directory
```
sschuberth, line 22: The `-i` option is specific to the `analyze` sub-command and needs to be moved after it.

docs/hints-for-use-with-docker.md
```diff
@@ -12,7 +12,18 @@ docker run \
...
+  analyze [...] # Insert further arguments for the command.
+```
```
sschuberth, line 24: In general, I wonder whether there's a nice way to combine this example with the previous one. For example, if we extended the first example's line 12 by writing

    analyze -i /project [...] # Insert further arguments for the command.

we could simply add this sentence instead of a second mostly identical example:

> If only a subproject shall be analyzed, change `-i /project` to `-i /project/subproject`. Note that still the project root directory needs to be mounted to Docker for ORT to detect VCS information.

Or something like that.

sschuberth (changes requested):

docs/hints-for-use-with-docker.md
tafli: Good idea, changed the docuumentation accordingly.

sschuberth: Thanks! One (or actually three) last wish(es): Please squash your commits (we prefer commits to be amended to address review comment, not new commits to be added to a PR), rebase onto `origin/main` (note that we've changed the default branch from `master` to `main` recently), and force-push your branch.

tafli:
Well that didn't come out as expected. I'll delete this PR and create a new one... sry.

PR #5060, opened 2022-02-14
Conan improvements and updates

sschuberth:


analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -222,6 +180,50 @@ class Conan(
         }
     }
 
+    private fun configureRemoteAuthentication(conanConfig: File) {
```
oheger-bosch, line 183: Good idea to put this code in a new function with a meaningful name, that way it is by far easier to understand! Maybe, as a further improvement and to support audience not familiar with Conan, would it make sense to add comments to the single command executions, e.g. what they are doing and why they are necessary in this context? Or just add links to the official documentation?

oheger-bosch (approved):

analyzer/src/main/kotlin/managers/Conan.kt
sschuberth: I've created a follow-up: https://github.com/oss-review-toolkit/ort/pull/5061

PR #5077, opened 2022-02-19
Create database connections lazily

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -39,6 +39,8 @@ import com.github.ajalt.clikt.parameters.options.split
 import com.github.ajalt.clikt.parameters.types.enum
 import com.github.ajalt.clikt.parameters.types.file
 
+import com.zaxxer.hikari.HikariDataSource
+
 import kotlinx.coroutines.runBlocking
 
 import org.ossreviewtoolkit.cli.GlobalOptions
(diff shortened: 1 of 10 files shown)


mnonnenmacher:
Creating the connection early has the benefit that it fails immediately if the configuration is wrong. You would not want to find out that the archiver configuration is wrong only after a potentially very long lasting scan.

sschuberth:
> Creating the connection early has the benefit that it fails immediately if the configuration is wrong. You would not want to find out that the archiver configuration is wrong only after a potentially very long lasting scan.

FYI, the reason I was doing this change is to make the workbench work for me: Right now, the workbench creates the `OrtApi` incl. a `fileArchiver` right at the start when loading a file. This means I cannot even load a file if the archiver configuration is wrong.

Do you have a better proposal how to solve this? Maybe we could simply add come `testConnection()` function that could be used to explicitly test the connection when wanted?

mnonnenmacher:
> FYI, the reason I was doing this change is to make the workbench work for me: Right now, the workbench creates the `OrtApi` incl. a `fileArchiver` right at the start when loading a file. This means I cannot even load a file if the archiver configuration is wrong.
>
> Do you have a better proposal how to solve this? Maybe we could simply add come `testConnection()` function that could be used to explicitly test the connection when wanted?

To me it's two things that should be done, first the workbench should handle such errors better and probably allow to ignore them or disable certain parts of the configuration, and second the data sources should be initialized lazily as you propose with an optional function to test the configuration which the ORT CLI should call on start.

However, what I don't like about the current implementation is that it makes the Hikari config a dependency of the archiver class. I don't have a better idea right know, but let's discuss this offline tomorrow.

sschuberth:
> However, what I don't like about the current implementation is that it makes the Hikari config a dependency of the archiver class. I don't have a better idea right know, but let's discuss this offline tomorrow.

I've found a different solution, please see https://github.com/oss-review-toolkit/ort/pull/5091. Closing in favor of that one.

PR #5078, opened 2022-02-20
Associate stand-alone ScanCode exceptions to licenses

sschuberth:


scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -245,6 +249,52 @@ internal fun replaceLicenseKeys(licenseExpression: String, replacements: Collect
         result
     }
 
+private const val LICENSES_WITH_EXCEPTIONS_TOLERANCE_LINES = 5
+
+internal fun associateLicensesWithExceptions(findings: List<LicenseFinding>): List<LicenseFinding> {
```
tsteenbe, line 254: Prefer not to have LLVM-exception mapping to Apache-2.0 hard-coded in ORT's Kotlin code as IMO it does not really scale well - SPDX has 40+ exxceptions alone.
Propose we introduce a file called `exception-mapping.yml` in `utils/spdx/src/main/resources` to map exceptions to license ids.
I created such a `exception-mapping.yml` in the below list based on `notes`/`text` fields or links for each offical [SPDX exception](https://spdx.org/licenses/exceptions-index.html)

```
"389-exception": GPL-2.0-only, GPL-2.0-or-later
"Autoconf-exception-2.0": GPL-2.0-only, GPL-2.0-or-later
"Autoconf-exception-3.0": GPL-3.0-only, GPL-3.0-or-later
"Bison-exception-2.2": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"Bootloader-exception": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"CLISP-exception-2.0": GPL-2.0-only, GPL-2.0-or-later
"Classpath-exception-2.0": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"DigiRule-FOSS-exception": GPL-2.0-only, GPL-2.0-or-later
"FLTK-exception": LGPL-2.0-only
"Fawkes-Runtime-exception": GPL-2.0-only
"Font-exception-2.0": GPL-2.0-only, GPL-2.0-or-later
"GCC-exception-2.0": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"GCC-exception-3.1": GPL-3.0-only, GPL-3.0-or-later
"GPL-3.0-linking-exception": GPL-3.0-only, GPL-3.0-or-later
"GPL-3.0-linking-source-exception": GPL-3.0-only, GPL-3.0-or-later
"GPL-CC-1.0": GPL-2.0-only, GPL-2.0-or-later, LGPL-2.0-only, LGPL-2.0-or-later, LGPL-2.1-only, LGPL-2.1-or-later
"LGPL-3.0-linking-exception": LGPL-3.0-only, LGPL-3.0-or-later
"LLVM-exception": Apache-2.0
"LZMA-exception": CPL-1.0
"Libtool-exception": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"Linux-syscall-note": GPL-2.0-only
"Nokia-Qt-exception-1.1": LGPL-2.1-only
"OCCT-exception-1.0": LGPL-2.1-only
"OCaml-LGPL-linking-exception": LGPL-2.0-only, LGPL-2.0-or-later, LGPL-2.1-only, LGPL-2.1-or-later,  LGPL-3.0-only, LGPL-3.0-or-later
"OpenJDK-assembly-exception-1.0": GPL-2.0-only
"PS-or-PDF-font-exception-20170817": AGPL-3.0
"Qt-GPL-exception-1.0": GPL-3.0-only
"Qt-LGPL-exception-1.1": LGPL-2.1-only, LGPL-2.1-or-later
"Qwt-exception-1.0": LGPL-2.1-only, LGPL-2.1-or-later
"SHL-2.0": Apache-2.0
"SHL-2.1": Apache-2.0
"Swift-exception": Apache-2.0
"Universal-FOSS-exception-1.0": GPL-2.0-only, GPL-2.0-or-later
"WxWindows-exception-3.1": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"eCos-exception-2.0": GPL-2.0-only, GPL-2.0-or-later
"freertos-exception-2.0": GPL-2.0-only
"gnu-javamail-exception": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"i2p-gpl-java-exception": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"mif-exception": GPL-2.0-only
"openvpn-openssl-exception": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
"u-boot-exception-2.0": GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-only, GPL-3.0-or-later
```

tsteenbe (changes requested):

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: Actually, I consider the code in this PR to be a work-around for a [bug / limitation in ScanCode](https://github.com/nexB/scancode-toolkit/issues/2873). So my hope is that the issue will soon be fixed upstream in ScanCode, and we don't need this mechanism at all anymore in ORT. With that in mind I'm unsure if it's worth to put in more effort to generalize the approach. *Maybe* I find the time to try, but if not, I'd be grateful if this could be merged as-is and eventually improved later.

sschuberth: Also, @tsteenbe, what do you think about adding the licenses an exception is applicable for to `SpdxLicenseException` (as a forth argument to the `enum`s constructor) instead of a separate YAML file? I mean, these mappings should change far less often than the `declared-license-mapping.yml`.

sschuberth:
I've split out some changes to https://github.com/oss-review-toolkit/ort/pull/5080.

sschuberth:
I've generalized the solution in additional commits, please have another look.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
tsteenbe: I thought of `exception-mapping.yml` as 1) it's scanner independent mapping solution,  2) using a YAML file makes it less scary to contribute than Kotlin file for say a license clearance team, 3) would allow for validation (exception -> right license?) and 4) ScanCode has about 80 [custom exceptions](https://scancode-licensedb.aboutcode.org/?search=exception) and the list is growing

sschuberth: It's now implemented using a YAML file, please have a look.

tsteenbe: Looks good to me, happy to approve but would love one of @oss-review-toolkit/kotlin-devs to also review the Kotlin code.

tsteenbe: Recommend we add the label 'release-notes' as this change how ORT handles exceptions so likely ORT users will have to update exception handling in their rules.kts

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -252,47 +253,58 @@ internal fun replaceLicenseKeys(licenseExpression: String, replacements: Collect
...
-    val llvmExceptions = mutableListOf<LicenseFinding>()
+    val remainingExceptions = exceptions.toMutableList()
+    val fixedLicenses = licenses.toMutableList()
 
-    findings.forEach {
-        when (it.license.toString()) {
-            "Apache-2.0" -> apacheLicenses += it
-            "LLVM-exception" -> llvmExceptions += it
-            else -> otherFindings += it
+    val e = remainingExceptions.iterator()
```
mnonnenmacher, line 261: Using `e` here makes the code look a bit like it would handle actual throwable exceptions, not license exceptions.

scanner/src/test/kotlin/scanners/scancode/ScanCodeResultParserTest.kt
```diff
@@ -654,12 +654,13 @@ class ScanCodeResultParserTest : FreeSpec({
...
-                    LicenseFinding("LLVM-exception", TextLocation("file", 100))
+                    LicenseFinding("GPL-2.0-only", TextLocation("file", 1)),
+                    LicenseFinding("389-exception", TextLocation("file", 100))
                 )
             ) should containExactlyInAnyOrder(
-                LicenseFinding("Apache-2.0", TextLocation("file", 1)),
-                LicenseFinding("Apache-2.0 WITH LLVM-exception", TextLocation("file", 100))
+                LicenseFinding("GPL-2.0-only", TextLocation("file", 1)),
+                LicenseFinding("GPL-2.0-only WITH 389-exception", TextLocation("file", 100)),
+                LicenseFinding("GPL-2.0-or-later WITH 389-exception", TextLocation("file", 100))
```
mnonnenmacher, line 663: This makes me wonder if it would be better to associate any matching license finding regardless of the line to the exception. Otherwise this mechanism introduces licenses which were not detected anywhere, e.g. GPL-2.0-or-later in this case, or when I look at the mapping in the previous commit it could also introduce different versions of GPL licenses. And if no matching license is detected in the whole file it might be better to create an issue than to make up an association, at least if it is ambiguous, because it needs a manual review anyway.

mnonnenmacher (changes requested):

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -245,6 +249,52 @@ internal fun replaceLicenseKeys(licenseExpression: String, replacements: Collect
...
+            "LLVM-exception" -> llvmExceptions += it
+            else -> otherFindings += it
+        }
+    }
+
+    val fixedApacheLicenses = apacheLicenses.map { apache ->
+        val llvm = llvmExceptions.filter { it.location.path == apache.location.path }
+            .map { it to it.location.distanceTo(apache.location) }
+            .sortedBy { it.second }
+            .firstOrNull { it.second <= LICENSES_WITH_EXCEPTIONS_TOLERANCE_LINES }
```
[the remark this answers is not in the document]
mnonnenmacher: Already done in the last commit.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -245,6 +249,52 @@ internal fun replaceLicenseKeys(licenseExpression: String, replacements: Collect
...
+
+        if (llvm != null) {
+            llvmExceptions.remove(llvm.first)
```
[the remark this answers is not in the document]
sschuberth: I guess that wouldn't work easily as I'm later handling the remaining exceptions in `llvmExceptions`.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: Renamed to `i` instead 😁 (Unless you come up with a better name, but I'd like to keep it short.)

scanner/src/test/kotlin/scanners/scancode/ScanCodeResultParserTest.kt
sschuberth: To simplify things, how about associating remaining exceptions with just `NOASSERTION` in general, instead of all applicable licenses?

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: ...although IMO `e.remove()` read nicer than `i.remove()`.

scanner/src/test/kotlin/scanners/scancode/ScanCodeResultParserTest.kt
tsteenbe: Was first thinking associating "orphan" exceptions to the root/main license but then I realized that finding them in scans would be difficult. 👍 from me to associate "orphan" exceptions with `NOASSERTION` - this should make it easy for ORT users to write a Evaluator rule requiring all exceptions in-combination with a `NOASSERTION` need to be investigated and resolved.

sschuberth:
I've made a (hopefully) final refacting to put the (now) generic `associateLicensesWithExceptions()` function into the `FindingsMatcher` file, which IMO is a better fit.

mnonnenmacher (approved):

PR #5085, opened 2022-02-22
Pub: Extract PubCacheReader to a separate file

fviernau:
Reduce the file size a bit and apply a minor simplification.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -31,6 +31,7 @@ import java.util.SortedSet
 
 import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
 import org.ossreviewtoolkit.analyzer.PackageManager
+import org.ossreviewtoolkit.analyzer.managers.utils.PubCacheReader
 import org.ossreviewtoolkit.analyzer.parseAuthorString
 import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.downloader.VersionControlSystem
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/managers/PubCacheReader.kt
```diff
@@ -0,0 +1,104 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.analyzer.managers
```
sschuberth, line 21: I believe this should be moved to `package org.ossreviewtoolkit.analyzer.managers.utils` instead.

analyzer/src/main/kotlin/managers/PubCacheReader.kt
```diff
@@ -82,20 +82,18 @@ internal class PubCacheReader {
...
-        if (path != null) {
-            pubCacheRoot.resolve(path).let {
-                if (it.isDirectory) {
-                    return it
-                }
+        pubCacheRoot.resolve(path).let {
+            if (it.isDirectory) {
+                return it
             }
+        }
```
sschuberth, line 92: How about instead (using `also` instead of `let` as the return value from the lambda is irrelevant):

    pubCacheRoot.resolve(path).takeIf { it.isDirectory }?.also { return it }

Similar below.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/PubCacheReader.kt
fviernau: I was undecided between `utils` and `managers`. The `*DependencyHandler` classes are in the `managers` package as well. Would you agree the `Handler` classes should be moved to `utils` too?

analyzer/src/main/kotlin/managers/PubCacheReader.kt
fviernau: I've made a separate commit for this on top based on your idea.

analyzer/src/main/kotlin/managers/PubCacheReader.kt
sschuberth: Yes, I always wanted to move them, but forgot about it, so I did it now: https://github.com/oss-review-toolkit/ort/pull/5086

PR #5103, opened 2022-02-25
Improve logging when archiving files

sschuberth:
Only log (the performance) in the downstream `archive()` function. The
package coordinates are still visible from the logged directory name.

### DIFF
--- a/scanner/src/main/kotlin/PathScanner.kt
+++ b/scanner/src/main/kotlin/PathScanner.kt
@@ -120,7 +120,7 @@ abstract class PathScanner(
     ): Map<Package, List<ScanResult>> {
         val scannerCriteria = getScannerCriteria()
 
-        log.info { "Searching scan results for ${packages.size} package(s)." }
+        log.info { "Searching scan results for ${packages.size} package(s)..." }
 
         val remainingPackages = packages.filterTo(mutableListOf()) { pkg ->
             !pkg.isMetaDataOnly.also {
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/PathScanner.kt
```diff
@@ -138,7 +138,10 @@ abstract class PathScanner(
 
         remainingPackages.removeAll { it in resultsFromStorage.keys }
 
-        log.info { "Scanning ${remainingPackages.size} package(s) for which no stored scan results were found." }
+        log.info {
+            "Scanning ${remainingPackages.size} package(s) for which no stored scan results were found out of total " +
+                    "${packages.size} package(s)."
```
mnonnenmacher, line 143: "... out of **a total of** x package(s)."

mnonnenmacher (changes requested):

scanner/src/main/kotlin/PathScanner.kt
mnonnenmacher: The sentence sounds a bit clumsy, how about changing the log above instead to "Found stored scan results for x of y packages matching $scannerCriteria", which would also be more correct, as technically `resultsFromStorage.size` is the number of packages for which results were found, not the number of scan results in the map.

sschuberth: Wuold you then remove the

    log.info { "Found ${resultsFromStorage.size} stored scan result(s) matching $scannerCriteria." }

from a few lines above?

mnonnenmacher (approved):

scanner/src/main/kotlin/PathScanner.kt
mnonnenmacher: Resolved offline.

PR #5106, opened 2022-02-25
ExperimentalScanner refactoring

mnonnenmacher:
Several refactorings of the experimental scanner code, see the commit messages for details. This is only the first part of the refactoring, another refactoring will follow which will add handling of issues during provenance resolution and might also include a cleaner implementation of #5047.

### DIFF
--- a/scanner/src/funTest/kotlin/experimental/DefaultPackageProvenanceResolverFunTest.kt
+++ b/scanner/src/funTest/kotlin/experimental/DefaultPackageProvenanceResolverFunTest.kt
@@ -20,9 +20,12 @@
 
 package org.ossreviewtoolkit.scanner.experimental
 
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.Spec
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.shouldBe
+
+import java.io.IOException
 
 import kotlinx.coroutines.runBlocking
 
@@ -33,7 +36,6 @@ import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.RemoteArtifact
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.SourceCodeOrigin
-import org.ossreviewtoolkit.model.UnknownProvenance
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
 
@@ -71,7 +73,7 @@ class DefaultPackageProvenanceResolverFunTest : WordSpec() {
                     )
                 )
 
-                resolver.resolveProvenance(pkg, listOf(SourceCodeOrigin.ARTIFACT)) shouldBe UnknownProvenance
+                shouldThrow<IOException> { resolver.resolveProvenance(pkg, listOf(SourceCodeOrigin.ARTIFACT)) }
             }
         }
 
@@ -117,7 +119,7 @@ class DefaultPackageProvenanceResolverFunTest : WordSpec() {
                     )
                 )
 
-                resolver.resolveProvenance(pkg, listOf(SourceCodeOrigin.VCS)) shouldBe UnknownProvenance
+                shouldThrow<IOException> { resolver.resolveProvenance(pkg, listOf(SourceCodeOrigin.VCS)) }
             }
 
             "Guess the correct tag for a package" {
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -25,7 +25,6 @@ import java.nio.file.StandardCopyOption
 import java.time.Instant
 
 import kotlin.io.path.moveTo
-import kotlin.time.TimeSource
 import kotlin.time.measureTime
 
 import org.ossreviewtoolkit.downloader.DownloadException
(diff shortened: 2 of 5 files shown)


scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -178,6 +171,22 @@ class ExperimentalScanner(
         return controller.getNestedScanResultsByPackage()
     }
 
+    private fun resolvePackageProvenances(controller: ScanController) {
+        log.info { "Resolving provenance for ${controller.packages.size} packages." }
+
+        val mark = TimeSource.Monotonic.markNow()
```
sschuberth, line 154: Why not `measureTime`?

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -187,6 +176,22 @@ class ExperimentalScanner(
         log.info { "Resolved provenance for ${controller.packages.size} packages in ${mark.elapsedNow()}." }
     }
 
+    private fun resolveNestedProvenances(controller: ScanController) {
+        log.info { "Resolving nested provenances for ${controller.packages.size} packages." }
+
+        val mark = TimeSource.Monotonic.markNow()
```
sschuberth, line 170: Why not `measureTime`?

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -309,7 +297,14 @@ class ExperimentalScanner(
...
         }
 
-    private fun getStoredResults(controller: ScanController) {
+    private fun readStoredResults(controller: ScanController) {
+        log.info {
+            "Reading stored scan results for ${controller.getPackageProvenances().size} packages with " +
+                    "${controller.getAllProvenances().size} provenances."
+        }
+
+        val mark = TimeSource.Monotonic.markNow()
```
sschuberth: Why not `measureTime`?

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -170,7 +170,10 @@ class ExperimentalScanner(
             "Read stored scan results in $readDuration:"
         }
         storedResults.entries.forEach { (scanner, results) ->
-            log.info { "\t${scanner.name}: Results for ${results.size} of ${allKnownProvenances.size} provenances." }
+            log.info {
+                "\t${scanner.name}: Results for ${results.filter { it.value.isNotEmpty() }.size} of " +
```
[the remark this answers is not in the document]
mnonnenmacher: This has been squashed into the later refactoring as it was replaced anyway.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
```diff
@@ -82,14 +85,28 @@ class DefaultPackageProvenanceResolver(
...
                 log.info {
                     "Could not resolve $sourceCodeOrigin for ${pkg.id.toCoordinates()}: ${it.collectMessagesAsString()}"
                 }
             }
         }
 
-        return UnknownProvenance
+        val message = buildString {
+            append(
+                "Could not resolve provenance for ${pkg.id.toCoordinates()} for source code origins " +
```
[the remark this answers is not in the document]
mnonnenmacher: I will make a separate commit in the next PR to add single quotes to all logs messages, if that's fine with you.

scanner/src/funTest/kotlin/experimental/DefaultPackageProvenanceResolverFunTest.kt
```diff
@@ -71,7 +73,7 @@ class DefaultPackageProvenanceResolverFunTest : WordSpec() {
                     )
                 )
 
-                resolver.resolveProvenance(pkg, listOf(SourceCodeOrigin.ARTIFACT)) shouldBe UnknownProvenance
+                shouldThrow<IOException> { resolver.resolveProvenance(pkg, listOf(SourceCodeOrigin.ARTIFACT)) }
```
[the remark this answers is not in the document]
mnonnenmacher: I think it's ok to just assert the exception, we do this in many other places as well.

scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt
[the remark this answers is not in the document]
sschuberth: Sure!

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: To reduce nesting. I find `measureTime` is often very invasive if the block is longer than 1-2 lines because it looks like control flow but is just measuring a duration.

sschuberth: I see. On the upside, `measureTime` nicely groups *what*'s being measured, which is esp. useful for more than 1-2 lines, IMO. I'm not entirely happy about the use of something else than `measureTime` here (mostly for consistency reasons), but I also won't insist on changing it.

sschuberth: Alternatively, the caller could now measure the time for the single function call.

scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -188,7 +193,7 @@ class ScanController(
     /**
      * Return true if [ScanResult]s for the complete [NestedProvenance] of the package are available.
      */
-    fun hasCompleteScanResult(scanner: PackageScannerWrapper, pkg: Package): Boolean =
+    fun hasCompleteScanResult(scanner: ScannerWrapper, pkg: Package): Boolean =
```
sschuberth, line 196: Just double-checking while skimming through the code: Is this change intended?

mnonnenmacher: Yes, but I have moved it to the commit that introduced the function because `ScannerWrapper` made more sense from the beginning.

sschuberth (approved):

mnonnenmacher:
Merged despite tests failing with cache issues. I have manually verified all related tests.

PR #5117, opened 2022-03-01
EvaluatedModelMapper: Do not decompose a package's licenses

sschuberth:
Since a while, ScanCode can report license expressions, which are also
used by ORT if the `parseLicenseExpressions` option is enabled. Display
such expressions in the webapp report without decomposing them, as these
expressions are what is being used in the matcher logic of package
configurations / license findings curations. Otherwise, the user would
have no chance to know from the web app report which string to use as a
detected license within a license findings curation.

It has been verified that this change as no negative impact on the
"Detected Licenses" summary in the web app report: In there, the
license counts still display decomposed licenses.


sschuberth:
Before:

image

After:

image

mnonnenmacher:
This kind of reverts 05dca02. I do not recall the reason behind that commit, @fviernau could you verify that this does not break anything on your side?

fviernau (approved):

reporter/src/funTest/assets/evaluated-model-reporter-test-expected-output.json
```diff
@@ -270,20 +270,6 @@
       "start_line" : 1,
       "end_line" : 1,
       "scan_result" : 0
```
fviernau, line 272: Any idea why entries for the file locations 'analyzer/src/funTest/assets/projects/synthetic/gradle/lib/src/code.cpp' get removed?

sschuberth: There's still one location left in that file, and I believe the other locations get consolidated into that one, as it's now not 3 separate license findings anymore, but one compound license expression finding.

sschuberth:
Merging despite the unrelated funTest failures.

PR #5123, opened 2022-03-01
Add declared license mapping for "Apache-2.0 license"

mnonnenmacher:
Add a few declared license mappings, see the commit messages for details.

### DIFF
--- a/utils/spdx/src/main/resources/declared-license-mapping.yml
+++ b/utils/spdx/src/main/resources/declared-license-mapping.yml
@@ -58,6 +58,7 @@
 "Apache version 2.0": Apache-2.0
 "Apache, Version 2.0": Apache-2.0
 "Apache-2.0 */ &#39; &quot; &#x3D;end --\n": Apache-2.0
+"Apache-2.0 license": Apache-2.0
 "Apple Public Source License": APSL-1.0
 "Artistic 2.0": Artistic-2.0
 "Artistic License": Artistic-2.0


utils/spdx/src/main/resources/declared-license-mapping.yml
```diff
@@ -79,6 +79,7 @@
 "BSD 3-clause New License": BSD-3-Clause
 "BSD 4 Clause": BSD-4-Clause
 "BSD Four Clause License": BSD-4-Clause
+"BSD License": BSD-3-Clause
```
sschuberth, line 82: We deliberately removed this in 391fe7144d05e7c41e2040ca4aa72ad6841c15f6 as it depends on the context / concrete package whether "BSD License" actually mean 3-Clause or 2-Clause; so we cannot have a generic mapping for it here.

I acknowledge there are probably more cases on non-generic mapping in this file which should be removed eventually, but we should not add more of these.

utils/spdx/src/main/resources/declared-license-mapping.yml
```diff
@@ -241,6 +241,7 @@
 "GNU General Public License, version 2": GPL-2.0-only
 "GNU General Public License, version 2, with the Classpath Exception": GPL-2.0-only WITH Classpath-exception-2.0
 "GNU LESSER GENERAL PUBLIC LICENSE V3.0": LGPL-3.0-only
+"GNU Lesser General Public License v2 (LGPLv2)": LGPL-2.1-or-later
```
sschuberth, line 244: From purely looking at the mapped text (without knowing anything about your concrete package), this should map to `LGPL-2.0-or-later` IMO.

sschuberth (changes requested):

utils/spdx/src/main/resources/declared-license-mapping.yml
fviernau: Could be contributed to the 'ort-config' repo...

sschuberth: > Could be contributed to the 'ort-config' repo...

As a concluded license for that particular package, yes, maybe.

mnonnenmacher: If you agree I will create a commit to remove the other ambiguous mappings.

sschuberth: > we should remove all of them immediately and not only prevent new ones from being added.

I agree. So far just a lack of time / priorities and the fear that such a PR would break too many things for our users prevented us from doing it. But if the consensus now is to do it, I can prepare a PR.

sschuberth: > I can prepare a PR.

Your comment and mine overlapped. Yes, please do prepare a PR.

fviernau: > Rather as a declared license mapping for that package, concluding a license based only on the declared license should not be a default.

+1

mnonnenmacher:
I have removed the two commits with ambiguous mappings.

sschuberth (approved):

PR #5124, opened 2022-03-01
Add shell autocompletion scripts

MarcelBochtler:


MarcelBochtler:
> As CLI options might change over time, I wonder how feasible it is to maintain these committed scripts.

I thought about that, too.
My decision to add the scripts was from a developer's perspective: I'd like to source the script automatically from my `~/.zshrc`. If the scripts are committed, I can be sure, that I'm always using the correct ones for the branch I'm currently working with.

However, I agree that this would add maintenance overhead, and at some point the scripts might be out of sync.

> or use e.g. https://github.com/test-room-7/action-update-file to auto-commit these files if CLI options change.

This might be a good solution, but I doubt that the overhead of such an Action (which would need to have access to the ORT binary), is worth it.

All facts considered, I tend to not committing the autocompletion scripts at all, and the user can manually generate them.

sschuberth:
> I'd like to source the script automatically from my `~/.zshrc`.

That's actually a good point: Being able to use the scripts directly from this repo. However, ...

> If the scripts are committed, I can be sure, that I'm always using the correct ones for the branch I'm currently working with.

... that's not the case. Unless committed scripts are updated when changes to the CLI modules occur, using them directly from the repo is no guarantee that they match the code.

> I doubt that the overhead of such an Action (which would need to have access to the ORT binary), is worth it.

There's no strict need to have access to the ORT binary. After the `build` job, we could simply run `./gradlew :cli:run --args="..."`.

Also the "overhead" of taking the action into use probably gets smaller when using the same action to auto-update an `AUTHORS` file, which is something I'm looking into in the context of https://github.com/oss-review-toolkit/ort-governance/issues/6.

> All facts considered, I tend to not committing the autocompletion scripts at all, and the user can manually generate them.

Ok, let's merge this as-is for now, and eventually add the automation later.

PR #5126, opened 2022-03-02
Properly capture project dependencies when writing SPDX documents

sschuberth:


reporter/src/main/kotlin/reporters/spdx/SpdxDocumentModelMapper.kt
```diff
@@ -234,6 +212,21 @@ private fun Package.toSpdxExternalReferences(): List<SpdxExternalReference> {
     return externalRefs
 }
 
+private fun Package.toSpdxPackage(licenseInfoResolver: LicenseInfoResolver, isProject: Boolean = false) =
```
oheger-bosch, line 225: I would rather not define a default value for the `isProject` parameter here, so that the caller is forced to specify it explicitly (and think about the correct value).

oheger-bosch (approved):

reporter/src/main/kotlin/reporters/spdx/SpdxDocumentModelMapper.kt
sschuberth: I had actually considered that, but as the function is called "toSpdx*Package*()" and will likely be mostly used for ORT *Packages* and rather rarely for ORT *Projects*, I decided to have that default argument for convenience.

PR #5127, opened 2022-03-02
declared-license-mapping: Remove abiguous mappings

mnonnenmacher:
Remove mappings where the version of a license is not obvious from the
source string. For these declared licenses package specific declared
license mappings should be created instead.

This is a draft for discussion, and to be able to test the impact of this change. Ideally package specific declared license mappings should be created in the [ort-config repository](https://github.com/oss-review-toolkit/ort-config) before the change is merged, at least for those removed mappings where the source package can be found, for example from the message of the commit that introduced the mapping.

### DIFF
--- a/utils/spdx/src/main/resources/declared-license-mapping.yml
+++ b/utils/spdx/src/main/resources/declared-license-mapping.yml
@@ -14,7 +14,6 @@
 "ASL 2": Apache-2.0
 "ASL 2.0": Apache-2.0
 "ASL, version 2": Apache-2.0
-"Academic Free License (AFL)": AFL-2.1
 "Academic Free License (AFL-2.1)": AFL-2.1
 "Affero General Public License (AGPL) v. 3": AGPL-3.0-only
 "Aladdin Free Public License (AFPL)": Aladdin
@@ -32,7 +31,6 @@
 "Apache License Version 2.0": Apache-2.0
 "Apache License v2": Apache-2.0
 "Apache License v2.0": Apache-2.0
-"Apache License": Apache-2.0
 "Apache License, 2.0": Apache-2.0
 "Apache License, ASL Version 2.0": Apache-2.0
 "Apache License, V2 or later": Apache-2.0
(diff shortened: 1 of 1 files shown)


utils/spdx/src/main/resources/declared-license-mapping.yml
```diff
@@ -32,7 +31,6 @@
 "Apache License Version 2.0": Apache-2.0
 "Apache License v2": Apache-2.0
 "Apache License v2.0": Apache-2.0
-"Apache License": Apache-2.0
```
tsteenbe, line 35: Apache license is 99,99% of the cases Apache 2.0 as previous version of this licenses (Apache-1.0, Apache-1.1) are super rare. I was thinking maybe we could implement a mapping for aboguous licenses based on year. Apache-2.0 was released in 2004 so could we say "if a project is from 2004 or younger map Apache License to Apache-2.0" Thoughts?

tsteenbe:
I would like to bring this PR to a vote by ORT TSC as  'as-is' it would cause serious disruption to ORT users and tens/hundreds of hours of additional work.

utils/spdx/src/main/resources/declared-license-mapping.yml
sschuberth: Even taking the year into account this way does not give 100% confidence. And we currently do not capture a project's / package's release date, which would be a significant change (but a good one anyway; I'd also like to be able to warn users about old dependencies in their tree that haven't seen a new release for years).

mnonnenmacher:
@tsteenbe FYI, this is the discussion that this PR originated from: https://github.com/oss-review-toolkit/ort/pull/5123#discussion_r816937603

utils/spdx/src/main/resources/declared-license-mapping.yml
```diff
@@ -14,7 +14,6 @@
 "ASL 2": Apache-2.0
 "ASL 2.0": Apache-2.0
 "ASL, version 2": Apache-2.0
```
fviernau, line 16: abiguous -> abmiguous

sschuberth:
Closing this in favor of the approach discussed in the [community meeting on 2022-11-24](https://github.com/oss-review-toolkit/ort/wiki/Community-Meeting#2022-11-24).

PR #5154, opened 2022-03-15
Upgrade SPDX licenses to 3.16

mnonnenmacher:
Upgrade the SPDX enums and license texts to version 3.16, see the commit messages for details.


sschuberth (approved):
> This is a fixup for https://github.com/oss-review-toolkit/ort/commit/6580509f749a7f949379653d778e405c38799e54, the changes done in this commit should have
been added to the generator script instead, otherwise they would be lost
if the `:utils:spdx:generateSpdxEnums` task is executed.

Ouch. Thanks for fixing this!

PR #5160, opened 2022-03-17
scanoss: Use the new Retrofit-based client

nnobelis:

The two sample Java classes are coming from https://github.com/vdurmont/semver4j.

### DIFF
--- a/scanner/build.gradle.kts
+++ b/scanner/build.gradle.kts
@@ -56,6 +56,7 @@ dependencies {
 
     implementation(project(":clients:clearly-defined"))
     implementation(project(":clients:fossid-webapp"))
+    implementation(project(":clients:scanoss"))
     implementation(project(":downloader"))
     implementation(project(":utils:core-utils"))
 
--- a/scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
+++ b/scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
@@ -20,22 +20,34 @@
 
 package org.ossreviewtoolkit.scanner.scanners.scanoss
 
-import com.scanoss.scanner.ScanFormat
-import com.scanoss.scanner.Scanner
-import com.scanoss.scanner.ScannerConf
+import com.scanoss.scanner.BlacklistRules
+import com.scanoss.scanner.Winnowing
 
 import java.io.File
 import java.time.Instant
+import java.util.UUID
 
+import kotlinx.coroutines.runBlocking
+
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.MultipartBody
+import okhttp3.RequestBody.Companion.toRequestBody
+
+import org.ossreviewtoolkit.clients.scanoss.ScanOssService
 import org.ossreviewtoolkit.model.ScanSummary
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
+import org.ossreviewtoolkit.model.jsonMapper
 import org.ossreviewtoolkit.model.readJsonFile
 import org.ossreviewtoolkit.scanner.AbstractScannerFactory
 import org.ossreviewtoolkit.scanner.BuildConfig
 import org.ossreviewtoolkit.scanner.PathScanner
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.core.createOrtTempDir
+import org.ossreviewtoolkit.utils.core.log
+
+// An arbitrary name to use for the multipart body being sent.
+private const val FAKE_WFP_FILE_NAME = "fake.wfp"
 
 class ScanOss internal constructor(
     name: String,
@@ -47,30 +59,75 @@ class ScanOss internal constructor(
             ScanOss(scannerName, scannerConfig, downloaderConfig)
     }
 
+    private val config = ScanOssConfig.create(scannerConfig).also {
+        log.info { "The $scannerName API URL is ${it.apiUrl}." }
+    }
+
+    private val service = ScanOssService.create(config.apiUrl)
+
+    // TODO: Find out the best / cheapest way to query the SCANOSS server for its version.
     override val version = BuildConfig.SCANOSS_VERSION
+
     override val configuration = ""
+
+    /**
+     * The name of the file corresponding to the fingerprints can be sent to SCANOSS for more precise matches.
+     * However, for anonymity, a unique identifier should be generated and used instead. This property holds the
+     * mapping between the file paths and the unique identifiers. When receiving the response, the UUID will be
+     * replaced by the actual file path.
+     *
+     * TODO: This behavior should be driven by a configuration parameter enabled by default.
+     */
+    private val fileNamesAnonymizationMapping = mutableMapOf<UUID, String>()
 
     override fun scanPathInternal(path: File): ScanSummary {
         val resultFile = createOrtTempDir().resolve("result.json")
         val startTime = Instant.now()
 
-        // TODO: Support API configurations other than OSSKB.
-        val scannerConf = ScannerConf.defaultConf()
-        val scanner = Scanner(scannerConf)
-
-        val scanType = null
-        val sbomPath = ""
-
-        if (path.isDirectory) {
-            scanner.scanDirectory(path.absolutePath, scanType, sbomPath, ScanFormat.plain, resultFile.absolutePath)
-        } else if (path.isFile) {
-            scanner.scanFileAndSave(path.absolutePath, scanType, sbomPath, ScanFormat.plain, resultFile.absolutePath)
+        val wfpString = buildString {
+            path.walk()
+                // TODO: Consider not applying the (somewhat arbitrary) blacklist.
+                .filter { !it.isDirectory && !BlacklistRules.hasBlacklistedExt(it.name) }
+                .forEach {
+                    this@ScanOss.log.info { "Computing fingerprint for file ${it.absolutePath}..." }
+                    append(createWfpForFile(it.path))
+                }
         }
+
+        val response = runBlocking {
+            val wfpBody = wfpString.toRequestBody("application/octet-stream".toMediaType())
+            val wfpFile = MultipartBody.Part.createFormData(FAKE_WFP_FILE_NAME, FAKE_WFP_FILE_NAME, wfpBody)
+
+            service.scan(wfpFile)
+        }
+
+        // Replace the anonymized UUIDs by their file paths.
+        val resolvedResponse = response.map { entry ->
+            val uuid = UUID.fromString(entry.key)
+
+            require(uuid in fileNamesAnonymizationMapping) {
+                "The $scannerName server returned an UUID not present in the mapping."
+            }
+
+            fileNamesAnonymizationMapping[uuid] to entry.value
+        }.toMap()
+
+        jsonMapper.writeValue(resultFile, resolvedResponse)
 
         val endTime = Instant.now()
         val result = readJsonFile(resultFile)
         resultFile.parentFile.safeDeleteRecursively(force = true)
 
         return generateSummary(startTime, endTime, path, result)
+    }
+
+    internal fun generateRandomUUID() = UUID.randomUUID()
+
+    internal fun createWfpForFile(filePath: String): String {
+        generateRandomUUID().let { uuid ->
+            // TODO: Let's keep the original file extension to give SCANOSS some hint about the mime type.
+            fileNamesAnonymizationMapping[uuid] = filePath
+            return Winnowing.wfpForFile(uuid.toString(), filePath)
+        }
     }
 }
--- a/scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
+++ b/scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
@@ -37,7 +37,7 @@ import org.ossreviewtoolkit.scanner.PathScanner
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.core.createOrtTempDir
 
-class ScanOss(
+class ScanOss internal constructor(
     name: String,
     scannerConfig: ScannerConfiguration,
     downloaderConfig: DownloaderConfiguration
--- a/scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
+++ b/scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
@@ -58,7 +58,6 @@ class ScanOss(
         val scannerConf = ScannerConf.defaultConf()
         val scanner = Scanner(scannerConf)
 
-        // TODO: Implement support for scanning with SBOM.
         val scanType = null
         val sbomPath = ""
 
--- a/scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
+++ b/scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2022 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.scanner.scanners.scanoss
+
+import org.ossreviewtoolkit.model.config.ScannerConfiguration
+
+/**
+ * A data class that holds the configuration options supported by the [ScanOss] scanner. An instance of this class is
+ * created from the options contained in a [ScannerConfiguration] object under the key _ScanOss_. It offers the
+ * following configuration options:
+ */
+internal data class ScanOssConfig(
+    /** URL of the ScanOSS server. */
+    val apiUrl: String,
+
+    /** API Key required to authenticate with the ScanOSS server. */
+    val apiKey: String
+) {
+    companion object {
+        /** Name of the configuration property for the API URL. */
+        const val API_URL_PROPERTY = "apiUrl"
+
+        /** Name of the configuration property for the API key. */
+        const val API_KEY_PROPERTY = "apiKey"
+
+        fun create(scannerConfig: ScannerConfiguration): ScanOssConfig {
+            val scanOssIdScannerOptions = scannerConfig.options?.get("ScanOss")
+
+            requireNotNull(scanOssIdScannerOptions) { "No ScanOSS Scanner configuration found." }
+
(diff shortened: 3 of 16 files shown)


scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -37,14 +37,15 @@ import org.ossreviewtoolkit.scanner.PathScanner
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.core.createOrtTempDir
 
-class ScanOss(
+class ScanOss internal constructor(
```
sschuberth, line 49: Why make the constructor internal? So far, `FossId` is the only other scanner with an internal constructor. And I tend to rather make it public there, too.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -37,14 +37,15 @@ import org.ossreviewtoolkit.scanner.PathScanner
...
     name: String,
     scannerConfig: ScannerConfiguration,
-    downloaderConfig: DownloaderConfiguration
+    downloaderConfig: DownloaderConfiguration,
+    private val config: ScanOssConfig
 ) : PathScanner(name, scannerConfig, downloaderConfig) {
     class Factory : AbstractScannerFactory<ScanOss>("SCANOSS") {
         override fun create(scannerConfig: ScannerConfiguration, downloaderConfig: DownloaderConfiguration) =
-            ScanOss(scannerName, scannerConfig, downloaderConfig)
+            ScanOss(scannerName, scannerConfig, downloaderConfig, ScanOssConfig.create(scannerConfig))
```
sschuberth, line 57: IMO it's a bit odd to pass both `scannerConfig` and `ScanOssConfig.create(scannerConfig)` to `ScanOss` here, and I'd rather only pass `scannerConfig` and let `ScanOss` internally call `ScanOssConfig.create(scannerConfig)`.

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
```diff
@@ -0,0 +1,95 @@
...
+            val apiKey = scanOssIdScannerOptions[API_KEY_PROPERTY].orEmpty()
+
+            return ScanOssConfig(apiURL, apiKey)
+        }
+    }
+
+    /**
+     * Create the [FossIdServiceWithVersion] to interact with the FossID instance described by this configuration.
+     */
+    fun createService(): ScanossService {
```
sschuberth, line 72: I realized only just now that the service is called `ScanossService`, but in all other classes we seem to be using `ScanOss...`. Could you please add a commit to align this?

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
```diff
@@ -0,0 +1,95 @@
...
+                init(null, trustAllCerts, SecureRandom())
+            }.socketFactory
+
+            sslSocketFactory(insecureSocketFactory, naiveTrustManager)
+            hostnameVerifier(HostnameVerifier { _, _ -> true })
+            return this
+        }
+
+        return ScanossService.create(apiURL, OkHttpClientHelper.buildClient() {
+            ignoreAllSSLErrors()
```
sschuberth, line 92: I believe we shouldn't do this unconditionally. Maybe also add a SCANOSS-specific option for this?

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -20,22 +20,31 @@
 
 package org.ossreviewtoolkit.scanner.scanners.scanoss
 
-import com.scanoss.scanner.ScanFormat
-import com.scanoss.scanner.Scanner
-import com.scanoss.scanner.ScannerConf
+import com.scanoss.scanner.BlacklistRules
```
sschuberth, line 23: Commit message: Remove the conflict marker.

sschuberth (changes requested):
Note: I haven't review the last commit fully yet, but wanted to already give some feedback.

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.ossreviewtoolkit.utils.core.log
+
+/**
+ * A data class that holds the configuration options supported by the [ScanOss] scanner. An instance of this class is
+ * created from the options contained in a [ScannerConfiguration] object under the key _ScanOss_. It offers the
+ * following configuration options:
+ */
+internal data class ScanOssConfig(
+    /** URL of the ScanOSS API server. */
+    val apiURL: String,
```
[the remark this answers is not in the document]
nnobelis: Non, this was the internal name in the legacy client.
There is no "browse URL" but I like the name `apiUrl` as it qualifies the nature of the URL. Can we keep it ?

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
```diff
@@ -0,0 +1,95 @@
...
+
+        /** Name of the configuration property for the API key. */
+        private const val API_KEY_PROPERTY = "apiKey"
+
+        fun create(scannerConfig: ScannerConfiguration): ScanOssConfig {
+            val scanOssIdScannerOptions = scannerConfig.options?.get("ScanOss")
+
+            requireNotNull(scanOssIdScannerOptions) { "No ScanOSS Scanner configuration found." }
+
+            val apiURL = scanOssIdScannerOptions[API_URL_PROPERTY] ?: "https://osskb.org/api/"
```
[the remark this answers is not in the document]
nnobelis: Nope it crashes retrofit:
```
baseUrl must end in /: https://osskb.org/api
java.lang.IllegalArgumentException: baseUrl must end in /: https://osskb.org/api
	at retrofit2.Retrofit$Builder.baseUrl(Retrofit.java:563)
	at retrofit2.Retrofit$Builder.baseUrl(Retrofit.java:506)
	at org.ossreviewtoolkit.clients.scanoss.ScanossService$Companion.create(ScanossService.kt:51)
```

scanner/build.gradle.kts
```diff
@@ -56,6 +56,7 @@ dependencies {
 
     implementation(project(":clients:clearly-defined"))
     implementation(project(":clients:fossid-webapp"))
+    implementation(project(":clients:scanoss"))
```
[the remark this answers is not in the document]
sschuberth: Ah, I misse that. But I'm wondering whether there should bet any coupling between `ScanossService` and `ScanOssConfig`. Shouldn't the just the `ScanOss` class later create a `ScanossService` from `ScanOssConfig`?

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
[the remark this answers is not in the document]
sschuberth: I'm ok with keeping the variable name, but can we then change "ScanOSS API server" in the code comments to just "ScanOSS server"?

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
nnobelis: Isn't it because you should not instantiate the scanner directly and go through the factory instead ?
So I understand the design of Oliver (who did it for FossID).

sschuberth: Well, ORT itself should use the factory (except maybe in tests), yes. But I don't see a reason why people who use ORT's libraries programmatically (if there are any 😀) instantiate the class directly.

If we make this constructor internal, it would be internal for all scanner implementations, and even going father, for all ORT classes that have factories.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
nnobelis: Fixed but please look at it again if it is to your taste because it has consequence on the property initialization.

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
nnobelis: It should not be done at all ! :D
I needed to intercept HTTPS calls with Wiremock.

sschuberth: Oh! If this is only needed for testing, we should probably also only configure the client this way for testing... but I'll have a look, maybe having this option for production code is also valuable.

nnobelis: You only need it at the start of the testing, we you record the requests by putting Wiremock between the client and the server.
After that, the Wiremock replays the server responses but the connection between the test and Wiremock is plain HTTP.
Therefore I removed this code.
If you need it again, I have a Gist :)

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -20,24 +20,33 @@
...
 import org.ossreviewtoolkit.model.readJsonFile
 import org.ossreviewtoolkit.scanner.AbstractScannerFactory
 import org.ossreviewtoolkit.scanner.BuildConfig
 import org.ossreviewtoolkit.scanner.PathScanner
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.core.createOrtTempDir
+import org.ossreviewtoolkit.utils.core.log
 
-class ScanOss internal constructor(
+class ScanOss constructor(
```
sschuberth, line 49: You should not remove the `internal` that we just added here again 😉

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -47,30 +56,79 @@ class ScanOss internal constructor(
...
     }
 
+    private val config = ScanOssConfig.create(scannerConfig).also {
+        log.info { "The $scannerName API URL is ${it.apiUrl}." }
+    }
+
+    private val service = ScanOssService.create(config.apiUrl)
+
+    companion object {
+        // The name of the multipart body being sent. Has no importance.
```
sschuberth, line 66: Maybe: s/Has no importance./Can be arbitrary./

Or yet more compact: "An arbitrary name to use for the multipart body being sent."

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -47,30 +56,79 @@ class ScanOss internal constructor(
...
+        private const val FAKE_WFP_FILE_NAME = "anon.wfp"
```
sschuberth, line 67: For consistency with the variable name, how about using "fake.wfp" as the file name?

sschuberth: Also, at some point we IIRC decided with @fviernau that only public constants should become part of companion objects for discoverability. Private constants should become top-level in the file.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -47,30 +56,79 @@ class ScanOss internal constructor(
...
+
     override val configuration = ""
 
+    /**
+     * The name of the file corresponding to the fingerprints can be sent to SCANOSS for more precise matches.
+     * However, for anonymity, a unique identifier should be generated and used instead. This property holds the
+     * mapping between the file paths and the unique identifiers. When receiving the response, the UUID will be
+     * replaced by the actual file path.
+     *
+     * TODO: This behaviour should be driven by a configuration parameter enabled by default.
```
sschuberth, line 81: Stick to American instead of British English: s/behaviour/behavior/

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -47,30 +56,79 @@ class ScanOss internal constructor(
...
+    // TODO: Find out the best / cheapest way to query the SCANOSS server for its version.
     override val version = BuildConfig.SCANOSS_VERSION
+
     override val configuration = ""
 
+    /**
+     * The name of the file corresponding to the fingerprints can be sent to SCANOSS for more precise matches.
+     * However, for anonymity, a unique identifier should be generated and used instead. This property holds the
+     * mapping between the file paths and the unique identifiers. When receiving the response, the UUID will be
+     * replaced by the actual file path.
```
sschuberth, line 79: Proposal:

> SCANOSS uses the file's path to increase confidence in snippet matches / to prioritize them. However, as exposing the file path from private repositories might be undesired, this map provides a way to anonymize paths before sending them to SCANOSS, and restore the original paths them afterwards.

@eeisegn, does that wording sound good to you, too?

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -47,30 +56,79 @@ class ScanOss internal constructor(
...
-        val scannerConf = ScannerConf.defaultConf()
-        val scanner = Scanner(scannerConf)
+        val wfpString = buildString {
+            path.walk()
+                // TODO: Consider not applying the (somewhat arbitrary) blacklist.
+                .filter { !it.isDirectory && !BlacklistRules.hasBlacklistedExt(it.name) }
+                .forEach {
+                    ScanOss.log.info { "Computing fingerprint for file ${it.absolutePath}..." }
+                    val wfpString = createWfpForFile(it.path)
+                    append(wfpString)
```
sschuberth, line 96: Maybe inline the `wfpString` variable here, esp. as the outer variable is called the same?

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -47,30 +56,79 @@ class ScanOss internal constructor(
...
-            scanner.scanDirectory(path.absolutePath, scanType, sbomPath, ScanFormat.plain, resultFile.absolutePath)
-        } else if (path.isFile) {
-            scanner.scanFileAndSave(path.absolutePath, scanType, sbomPath, ScanFormat.plain, resultFile.absolutePath)
+            service.scan(wfpFile)
         }
 
+        // Replace the anonymized UUIDs by their file paths.
+        val resolvedResponse = response.map { entry ->
+            val uuid = UUID.fromString(entry.key)
+            require(fileNamesAnonymizationCache.contains(uuid)) {
```
sschuberth, line 110: Put blank lines around the `require` block and use `uuid in ...` syntax.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -47,30 +56,79 @@ class ScanOss internal constructor(
...
+                "The $scannerName server returned an UUID not present in the cache."
```
sschuberth, line 111: It's rather a "map" than a "cache", IMO.

sschuberth: So actually the variable should also be renamed.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -47,30 +56,79 @@ class ScanOss internal constructor(
...
         val result = readJsonFile(resultFile)
         resultFile.parentFile.safeDeleteRecursively(force = true)
 
         return generateSummary(startTime, endTime, path, result)
     }
+
+    internal fun generateRandomUUID() = UUID.randomUUID()
+
+    internal fun createWfpForFile(filePath: String): String {
+        generateRandomUUID().let {
```
sschuberth, line 128: Please use `uuid ->` here for clarity.

scanner/src/test/assets/scanoss/filesToScan/Semver4jRange.java
```diff
@@ -0,0 +1,92 @@
+package com.vdurmont.semver4j;
```
sschuberth, line 1: It makes me feel a bit uneasy with committing third-party code here. Can't we simply copy (some old state of) some ORT code itself?

sschuberth (changes requested):

scanner/src/test/assets/scanoss/filesToScan/Semver4jRange.java
nnobelis: Unfortunately, you are right :(
I will change the examples and adapt the code, it will take some time because I need to recapture the requests with wiremock.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
nnobelis: Should I put at least the extension after the UUID ?

scanner/src/test/assets/scanoss/scanSingle/mappings/scanoss-semver4j-response.json
```diff
@@ -0,0 +1,23 @@
...
+    "headers" : {
+      "Server" : "nginx/1.14.2",
+      "Date" : "Wed, 16 Mar 2022 13:07:04 GMT",
+      "Content-Type" : "application/json"
+    }
+  },
+  "uuid" : "b28995c4-70e0-4fda-a1a8-c8a0ea1f1c85",
+  "persistent" : true,
+  "insertionIndex" : 1
+}
```
sschuberth, line 23: Missing final newline.

scanner/src/test/kotlin/scanners/scanoss/ScanOssScannerDirectoryTest.kt
```diff
@@ -0,0 +1,99 @@
...
+    afterSpec {
+        server.stop()
+    }
+
+    beforeTest {
+        server.resetAll()
+    }
+
+    "The scanner must scan a directory" {
+        // manipulate the UUID generation to have the same IDs as in the response
```
sschuberth, line 71: Please capitalize the sentence and end it with a dot.

scanner/src/test/kotlin/scanners/scanoss/ScanOssScannerFileTest.kt
```diff
@@ -0,0 +1,95 @@
...
+    afterSpec {
+        server.stop()
+    }
+
+    beforeTest {
+        server.resetAll()
+    }
+
+    "The scanner must scan a single file" {
+        // manipulate the UUID generation to have the same IDs as in the response
```
sschuberth, line 71: Please capitalize the sentence and end it with a dot.

scanner/src/test/kotlin/scanners/scanoss/ScanOssScannerDirectoryTest.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    afterSpec {
+        server.stop()
+    }
+
+    beforeTest {
+        server.resetAll()
+    }
+
+    "The scanner must scan a directory" {
```
sschuberth, line 70: s/must/should/

scanner/src/test/kotlin/scanners/scanoss/ScanOssScannerFileTest.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    afterSpec {
+        server.stop()
+    }
+
+    beforeTest {
+        server.resetAll()
+    }
+
+    "The scanner must scan a single file" {
```
sschuberth, line 70: s/must/should/

sschuberth (changes requested):
Just some final nits.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
sschuberth: We should certainly do that, but I'd be fine with leaving it as a TODO for this PR and address it in a follow-up.

sschuberth (approved):

PR #5171, opened 2022-03-21
DownloaderCommand: Add an option to filter by package ids

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/DownloaderCommand.kt
+++ b/cli/src/main/kotlin/commands/DownloaderCommand.kt
@@ -223,27 +223,24 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
         }
 
         val includedLicenseCategories = globalOptionsForSubcommands.config.downloader.includedLicenseCategories
+        if (includedLicenseCategories.isNotEmpty() && licenseClassificationsFile.isFile) {
+            val licenseCategorizations = licenseClassificationsFile.readValue<LicenseClassifications>().categorizations
+            val licenseInfoResolver = ortResult.createLicenseInfoResolver()
 
-        val packageDownloadDirs =
-            if (includedLicenseCategories.isNotEmpty() && licenseClassificationsFile.isFile) {
-                val licenseCategorizations =
-                    licenseClassificationsFile.readValue<LicenseClassifications>().categorizations
-                val licenseInfoResolver = ortResult.createLicenseInfoResolver()
+            packages.retainAll { pkg ->
+                // A package is only downloaded if its license is part of a category that is part of the
+                // DownloaderConfiguration's includedLicenseCategories.
+                getLicenseCategoriesForPackage(
+                    pkg,
+                    licenseCategorizations,
+                    licenseInfoResolver,
+                    ortResult.getRepositoryLicenseChoices(),
+                    ortResult.getPackageLicenseChoices(pkg.id)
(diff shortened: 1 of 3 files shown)


cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -223,27 +223,24 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
         }
 
         val includedLicenseCategories = globalOptionsForSubcommands.config.downloader.includedLicenseCategories
+        if (includedLicenseCategories.isNotEmpty() && licenseClassificationsFile.isFile) {
+            val licenseCategorizations = licenseClassificationsFile.readValue<LicenseClassifications>().categorizations
+            val licenseInfoResolver = ortResult.createLicenseInfoResolver()
```
mnonnenmacher, line 247: Unrelated to this commit, but this makes me realize that we do not provide the correct package configuration provider here which can lead to wrong results. It should be initialized like in `EvaluatorCommand`, just in case you want to fix this. Otherwise let me know, then I will make a separate PR later.

cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -173,6 +173,12 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
                 PackageType.values().joinToString { it.name }
     ).enum<PackageType>().split(",").default(enumValues<PackageType>().asList())
 
+    private val packageIds by option(
+        "--package-ids",
+        help = "A comma-separated list of regular expressions for matching package ids from the ORT file's analyzer " +
+                "result to limit downloads to. If not specified, all package types are downloaded."
```
mnonnenmacher, line 179: Nit: "packages types" -> "packages"

cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -222,6 +228,11 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
             }
         }
 
+        packageIds?.also {
```
mnonnenmacher, line 231: I would consider `if(packageIds != null)` more straightforward to read.

cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -249,6 +257,10 @@ class DownloaderCommand : CliktCommand(name = "download", help = "Fetch source c
                     ortResult.getPackageLicenseChoices(pkg.id)
                 ).any { it in includedLicenseCategories }
             }
+
+            if (isModified) {
+                log.info { "Filtered $originalCount to ${packages.size} package(s) by license classification." }
```
mnonnenmacher, line 262: This can be misleading if some of those packages have been filtered by the id filter already. Maybe a sequential log like this would be better:
```
Found 100 packages.
Removing 10 packages which do not match the id filter.
Removing 20 packages which do not match the license classification filter.
Downloading remaining 70 packages.
```

cli/src/main/kotlin/commands/DownloaderCommand.kt
sschuberth: That's what I had at first, but then got a hint that the smart-cast on a Clikt option (which is a delegated property) would not work, so I went for this syntax.

mnonnenmacher (approved):

PR #5172, opened 2022-03-21
JiraNotifier refactoring

mnonnenmacher:
Some refactorings of the `JiraNotifier` and the `JiraNotifierTest`, see the commit messages for details.


notifier/src/test/kotlin/modules/JiraNotifierTest.kt
```diff
@@ -19,257 +19,195 @@
...
-        server.resetAll()
-    }
-
     "createIssue" should {
         "succeed for valid input" {
-            val projectKey = "TEST"
-            val notifier = JiraNotifier(
-                JiraConfiguration("http://localhost:${server.port()}", "testuser", "testpassword")
-            )
+            val project = createProject()
```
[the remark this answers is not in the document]
oheger-bosch: Looks good to me. Testing the notifier's interaction with the Jira client should be sufficient.

notifier/src/test/kotlin/modules/JiraNotifierTest.kt
```diff
@@ -19,257 +19,195 @@
...
+            val issueClient = mockk<IssueRestClient>()
```
oheger-bosch, line 53: Just a suggestion: There is some redundancy in the tests for creating and preparing the mocks for the different clients. In such cases, I find it useful to have a test helper class that manages the mock objects involved and offers some basic functionality tests can make use of. For instance, here a class like `JiraClientHelper` could be defined with properties for the REST client, the issue client, and the project client. It could already prepare the REST client mock to return the other mocks. It could further offer convenience functions to prepare the mocks for creating issues or querying projects.

oheger-bosch (approved):

notifier/src/test/kotlin/modules/JiraNotifierTest.kt
```diff
@@ -64,8 +64,8 @@ class JiraNotifierTest : WordSpec({
         server.resetAll()
     }
 
-    "JiraNotifier" should {
-        "create a Jira ticket" {
+    "createIssue" should {
```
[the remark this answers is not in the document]
mnonnenmacher: I don't have a preference, camel-case and the verb are also a good indication it's a function. We could discuss offline if we want to align the code base here.

notifier/src/test/kotlin/modules/JiraNotifierTest.kt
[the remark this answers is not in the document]
mnonnenmacher: Yes, that's the point, not testing the library but interaction with the library. And making it easier to add tests because I will add more functions to the `JiraNotifier`.

notifier/src/test/kotlin/modules/JiraNotifierTest.kt
mnonnenmacher: I will consider this when adding tests in an upcoming PR. But it's always a trade-off, if the test code becomes too complex it increases the risk that it introduces it's own bugs.

sschuberth: > if the test code becomes too complex it increases the risk that it introduces it's own bugs.

Agreed, but I believe @oheger-bosch's main point was about *reducing* complexity by removing redundancies.

mnonnenmacher: At least partly done: https://github.com/oss-review-toolkit/ort/pull/5179

PR #5178, opened 2022-03-22
JiraNotifier: Add function to change the assignee

mnonnenmacher:
Add a function to change the assignee of an issue.


notifier/src/test/kotlin/modules/JiraNotifierTest.kt
```diff
@@ -216,6 +217,34 @@ class JiraNotifierTest : WordSpec({
         }
     }
 
+    "changeAssignee" should {
+        "succeed for valid input" {
+            val issueClient = mockk<IssueRestClient>()
```
fviernau, line 222: Consider factoring out the redundant common setup in lines 222 - 225.

fviernau (approved):

notifier/src/test/kotlin/modules/JiraNotifierTest.kt
mnonnenmacher: Done: https://github.com/oss-review-toolkit/ort/pull/5179

PR #5181, opened 2022-03-22
Introduce and use a `CompositePackageCurationProvider`

sschuberth:


analyzer/src/main/kotlin/curation/CompositePackageCurationProvider.kt
```diff
@@ -0,0 +1,41 @@
+/*
```
mnonnenmacher, line 1: Commit message: "provide**r**s", "whic**h**"

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -211,19 +212,20 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
+            val repositoryPackageCurations = repositoryConfiguration.curations.packages
 
-        if (config.enableRepositoryPackageCurations) {
-            packageCurations += repositoryPackageCurations
-        } else if (repositoryPackageCurations.isNotEmpty()) {
-            log.warn { "Local package curation were not applied because the feature is not enabled." }
+            if (config.enableRepositoryPackageCurations) {
+                add(SimplePackageCurationProvider(repositoryPackageCurations))
+            } else if (repositoryPackageCurations.isNotEmpty()) {
+                log.warn { "Local package curation were not applied because the feature is not enabled." }
```
mnonnenmacher, line 223: "curation**s**"
"Local" could be confusing, given that the other provider above is also a "local" provider. Maybe "Package curations from the .ort.yml are not applied...".

mnonnenmacher (changes requested):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
sschuberth: Please note that I just indented this existing code. So I'll change this in a separate commit.

mnonnenmacher (approved):

fviernau (approved):

PR #5162, opened 2022-03-18
Npm: Fix querying of package details from private / scoped registries 

sschuberth:
Please have a look at the individual commit messages for the details.

PR #5188, opened 2022-03-24
Npm vs Yarn fixes

sschuberth:

### DIFF
--- a/analyzer/src/funTest/kotlin/managers/NpmFunTest.kt
+++ b/analyzer/src/funTest/kotlin/managers/NpmFunTest.kt
@@ -44,7 +44,7 @@ class NpmFunTest : WordSpec() {
                 val workingDir = projectsDir.resolve("shrinkwrap")
                 val packageFile = workingDir.resolve("package.json")
 
-                val result = createNPM().resolveSingleProject(packageFile, resolveScopes = true)
+                val result = createNpm().resolveSingleProject(packageFile, resolveScopes = true)
                 val vcsPath = vcsDir.getPathToRoot(workingDir)
                 val expectedResult = patchExpectedResult(
                     projectsDir.parentFile.resolve("npm-expected-output.yml"),
@@ -62,7 +62,7 @@ class NpmFunTest : WordSpec() {
                 val workingDir = projectsDir.resolve("package-lock")
                 val packageFile = workingDir.resolve("package.json")
 
-                val result = createNPM().resolveSingleProject(packageFile, resolveScopes = true)
+                val result = createNpm().resolveSingleProject(packageFile, resolveScopes = true)
                 val vcsPath = vcsDir.getPathToRoot(workingDir)
                 val expectedResult = patchExpectedResult(
                     projectsDir.parentFile.resolve("npm-expected-output.yml"),
@@ -80,7 +80,7 @@ class NpmFunTest : WordSpec() {
                 val workingDir = projectsDir.resolve("no-lockfile")
                 val packageFile = workingDir.resolve("package.json")
 
-                val result = createNPM().resolveSingleProject(packageFile)
+                val result = createNpm().resolveSingleProject(packageFile)
                 val vcsPath = vcsDir.getPathToRoot(workingDir)
                 val expectedResult = patchExpectedResult(
                     projectsDir.parentFile.resolve("npm-expected-output-no-lockfile.yml"),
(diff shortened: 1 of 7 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -416,6 +309,115 @@ open class Npm(
         return modulesDir.takeIf { it.name == "node_modules" }
     }
 
+    /**
+     * Construct a [Package] by parsing its _package.json_ file and - if applicable - querying additional
+     * content via the `npm view` command. The result is a [Pair] with the raw identifier and the new package.
+     */
+    @Suppress("HttpUrlsUsage")
+    internal fun parsePackage(packageFile: File): Pair<String, Package> {
```
fviernau, line 312: Is this function just being moved or also modified?

Would it make sense to factor out moving of code to a separate commit for easier review?

fviernau (dismissed):

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: It was moved out of static context, but in the same run needed to be modified (code was factored out to ``) to do the right thing for Yarn. I considered splitting the move from the modifications, but could not figure out how in this case (without leaving things intermediately in a broken state).

sschuberth: Ok, I found a way to improve this.

analyzer/src/main/kotlin/managers/utils/NpmDependencyHandler.kt
```diff
@@ -49,13 +49,13 @@ data class NpmModuleInfo(
 /**
  * A specialized [DependencyHandler] implementation for NPM.
  */
-class NpmDependencyHandler : DependencyHandler<NpmModuleInfo> {
+class NpmDependencyHandler(private val npm: Npm) : DependencyHandler<NpmModuleInfo> {
```
oheger-bosch, line 52: This bidirectional dependency between `Npm` and `NpmDependencyHandler` is a bit problematic IMHO. Is there a way to move functionality, so that the relation becomes unidirectional again? For instance, could the `parsePackage()` function (and maybe further helper functions) be moved to `NpmDependencyHandler`?

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -71,12 +71,6 @@ import org.ossreviewtoolkit.utils.common.withoutPrefix
 import org.ossreviewtoolkit.utils.core.log
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 
-object NpmCli : CommandLineTool {
```
oheger-bosch, line 74: Not sure whether it makes a big difference, but you could keep the objects implementing `CommandLineTool`, and instead of extending the `CommandLineTool` interface in `Npm`, introduce a function (or property) like `getCommandLineTool()` that returns the tool for this concrete implementation.

analyzer/src/funTest/kotlin/managers/NpmFunTest.kt
```diff
@@ -44,7 +44,7 @@ class NpmFunTest : WordSpec() {
                 val workingDir = projectsDir.resolve("shrinkwrap")
```
oheger-bosch, line 44: Commit message nit: Should it be "adhere *to* Java naming conventions"?

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: In the current architecture, it's actually more convenient if really the package manager classes themselves implement the `CommandLineTool` interface, as they then show up in the correct section for `ort requirements`. Also see my remark in [this commit message](https://github.com/oss-review-toolkit/ort/pull/5162/commits/0c8cd71e6ef741a3816fd17fb165c55ad27b5bec).

analyzer/src/main/kotlin/managers/utils/NpmDependencyHandler.kt
sschuberth: Unfortunately, `parsePackage()` needs to call a package-manager-specific function. Currently, the only way I see to avoid this is to also introduce a `YarnDependencyHandler` which inherits from `NpmDependencyHandler`. But I'm not sure yet whether that's much nicer.

analyzer/src/main/kotlin/managers/Npm.kt
fviernau: Thanks!

oheger-bosch (dismissed):

oheger-bosch (approved):

sschuberth:
Merging despite the unrelated `DotNetFunTest` socket timeout.

PR #5193, opened 2022-03-28
Cargo: Remove a possible suffix from the version string

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Cargo.kt
+++ b/analyzer/src/main/kotlin/managers/Cargo.kt
@@ -75,7 +75,10 @@ class Cargo(
 
     override fun command(workingDir: File?) = "cargo"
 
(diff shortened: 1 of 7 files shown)


analyzer/src/main/kotlin/managers/Cargo.kt
```diff
@@ -75,7 +75,10 @@ class Cargo(
 
     override fun command(workingDir: File?) = "cargo"
 
-    override fun transformVersion(output: String) = output.removePrefix("cargo ")
+    override fun transformVersion(output: String) =
```
mnonnenmacher, line 78: Commit message: I think you mean "possible" or "potential", not "eventual".

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5197, opened 2022-03-28
Improvements to the `helper-cli`'s scan storage "delete" subcommand

sschuberth:

### DIFF
--- a/helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
@@ -113,6 +113,10 @@ internal class DeleteCommand : CliktCommand(
             }
 
             println("Would delete $count scan result(s).")
+
+            if (log.delegate.isDebugEnabled) {
+                ScanResults.slice(ScanResults.identifier).select { condition }.forEach(log::debug)
+            }
         } else {
             val count = database.transaction {
                 ScanResults.deleteWhere { condition }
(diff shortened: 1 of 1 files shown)


helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
```diff
@@ -99,11 +106,20 @@ internal class DeleteCommand : CliktCommand(
...
 
         val condition = conditions.compoundAnd()
-        database.transaction {
-            ScanResults.deleteWhere { condition }
-        }
 
-        log.info { "Successfully deleted stored scan results." }
+        if (dryRun) {
+            val results = database.transaction {
+                ScanResults.select { condition }.toList()
```
mnonnenmacher, line 112: Use `ScanResults.select { condition }.count()` instead for better performance.

helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
```diff
@@ -92,7 +92,7 @@ internal class DeleteCommand : CliktCommand(
             }
         }
 
-        val typeCondition = packageType?.let { ScanResults.identifier like "$it:%" }
+        val typeCondition = packageId?.let { ScanResults.identifier regexp it }
```
mnonnenmacher, line 95: Rename `typeCondition` to `identifierCondition`.

helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
```diff
@@ -113,6 +113,10 @@ internal class DeleteCommand : CliktCommand(
             }
 
             println("Would delete ${results.size} scan result(s).")
+
+            if (log.delegate.isDebugEnabled) {
+                results.map { it[ScanResults.identifier] }.forEach(log::debug)
+            }
```
mnonnenmacher, line 119: This would not work anymore with the `count()` from my previous comment. Instead you could use this query to not download the scan results but only the identifiers: `ScanResults.slice(ScanResults.identifier).select { condition }`

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5198, opened 2022-03-28
Slightly increase the default findings matcher tolerance lines to 7

sschuberth:


sschuberth:
Not needed anymore as line counts came from wrong findings.

PR #5205, opened 2022-03-30
FindingsMatcher: Omit contained exceptions

sschuberth:

### DIFF
--- a/model/src/test/kotlin/utils/FindingsMatcherTest.kt
+++ b/model/src/test/kotlin/utils/FindingsMatcherTest.kt
@@ -238,6 +238,19 @@ class FindingsMatcherTest : WordSpec() {
                 )
             }
 
+            "omit exceptions that are already contained in existing findings" {
+                associateLicensesWithExceptions(
+                    listOf(
+                        LicenseFinding("GPL-1.0-or-later", TextLocation("pom.xml", 45)),
+                        LicenseFinding("GPL-2.0-only WITH Classpath-exception-2.0", TextLocation("pom.xml", 46, 48)),
+                        LicenseFinding("Classpath-exception-2.0", TextLocation("pom.xml", 47))
+                    )
+                ) should containExactlyInAnyOrder(
+                    LicenseFinding("GPL-1.0-or-later", TextLocation("pom.xml", 45)),
+                    LicenseFinding("GPL-2.0-only WITH Classpath-exception-2.0", TextLocation("pom.xml", 46, 48))
+                )
+            }
+
             "associate orphan exceptions by NOASSERTION" {
                 associateLicensesWithExceptions(
                     listOf(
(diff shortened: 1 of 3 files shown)


model/src/test/kotlin/utils/FindingsMatcherTest.kt
```diff
@@ -238,6 +238,31 @@ class FindingsMatcherTest : WordSpec() {
...
+                ) should containExactlyInAnyOrder(
+                    LicenseFinding("EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0",
+                        TextLocation("pom.xml", 6, 16)),
+                    LicenseFinding("Apache-2.0 OR EPL-2.0", TextLocation("pom.xml", 39, 41)),
+                    LicenseFinding("LGPL-2.1-only OR Apache-2.0", TextLocation("pom.xml", 44, 46)),
+                    LicenseFinding("GPL-2.0-only WITH Classpath-exception-2.0", TextLocation("pom.xml", 46, 47)),
+                    LicenseFinding("GPL-1.0-or-later", TextLocation("pom.xml", 46)),
+                    LicenseFinding("Apache-2.0", TextLocation("pom.xml", 53)),
+                    LicenseFinding("Apache-2.0", TextLocation("pom.xml", 53, 58))
+                )
```
mnonnenmacher, line 263: It's difficult to see what the actual test is here, do all the other findings actually contribute to the test? How about something this to demonstrate that one exception is omitted and the other is not:

```
            "omit exceptions that are already contained in existing findings" {
                associateLicensesWithExceptions(
                    listOf(
                        LicenseFinding("GPL-1.0-or-later", TextLocation("pom.xml", 45)),
                        LicenseFinding("GPL-2.0-only WITH Classpath-exception-2.0", TextLocation("pom.xml", 46, 48)),
                        LicenseFinding("Classpath-exception-2.0", TextLocation("pom.xml", 45)),
                        LicenseFinding("Classpath-exception-2.0", TextLocation("pom.xml", 47))
                    )
                ) should containExactlyInAnyOrder(
                        LicenseFinding("GPL-1.0-or-later", TextLocation("pom.xml", 45)),
                        LicenseFinding("GPL-2.0-only WITH Classpath-exception-2.0", TextLocation("pom.xml", 46, 48)),
                        LicenseFinding("Classpath-exception-2.0", TextLocation("pom.xml", 45))
                )
```

mnonnenmacher (changes requested):

model/src/test/kotlin/utils/FindingsMatcherTest.kt
sschuberth: But I'll remove the case of the orphan exception, as that's already covered by "associate orphan exceptions by NOASSERTION".

mnonnenmacher (approved):

PR #5208, opened 2022-03-30
`ExperimentalScanner`: Plugin support

mnonnenmacher:
Add support for plugins for the experimental scanner, and for using multiple scanners at once. See the commit messages for details.

### DIFF
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2017-2019 HERE Europe B.V.
- * Copyright (C) 2021 Bosch.IO GmbH
+ * Copyright (C) 2021-2022 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import com.github.ajalt.clikt.core.CliktCommand
 import com.github.ajalt.clikt.core.ProgramResult
 import com.github.ajalt.clikt.core.UsageError
 import com.github.ajalt.clikt.core.requireObject
+import com.github.ajalt.clikt.parameters.groups.default
 import com.github.ajalt.clikt.parameters.groups.mutuallyExclusiveOptions
 import com.github.ajalt.clikt.parameters.groups.required
 import com.github.ajalt.clikt.parameters.groups.single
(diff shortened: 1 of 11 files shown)


cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -145,16 +161,36 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license
...
-
-    private val projectScannerFactory by option(
-        "--project-scanner",
-        help = "The scanner to use for scanning the source code of projects. By default, projects and packages are " +
-                "scanned with the same scanner as specified by '--scanner'."
-    ).convertToScanner()
+    private val scannerOption by mutuallyExclusiveOptions(
+        option(
+            "--scanner", "-s",
+            help = "The scanner to use, one of ${Scanner.ALL}."
```
sschuberth, line 167: While at it, should we also align this to the "Possible values are" phrase?

cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -145,16 +161,36 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license
...
+                    "it is therefore not recommended to use it in production.\n" +
+                    "Possible values are: ${ScannerWrapper.ALL}"
+        ).convertToScannerWrapperFactories()
+    ).single().default(ScannerOption.Stable(ScanCode.Factory()))
+
+    private val projectScannerOption by mutuallyExclusiveOptions(
+        option(
+            "--project-scanner",
+            help = "The scanner to use for scanning the source code of projects. By default, projects and packages " +
+                    "are scanned with the same scanner as specified by '--scanner'."
```
sschuberth, line 184: Nit: While at it, should we also add the "Possible values are" phrase here? Or drop it from below. I didn't add the list of scanners here initially as this option already refers to the general option which does have the list.

cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -201,10 +229,22 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license
 
         val config = globalOptionsForSubcommands.config
 
-        val ortResult = if (experimental) {
-            runExperimental(config)
-        } else {
-            run(config)
+        val ortResult = when (scannerOption) {
```
sschuberth, line 232: You could use `when (val scanner = scannerOption) {` here and then `scanner` instead of `scannerOption` below to make the smart-cast work (and probably inline the `scannerFactory` variable).

sschuberth (changes requested):

cli/src/main/kotlin/commands/ScannerCommand.kt
mnonnenmacher: I decided to be more verbose and list the values for each option, to make it easier to understand for the user.

sschuberth (dismissed):

sschuberth:
PS: If you use GitHub's new "Update with rebase" on this, existing approvals do not get lost.

mnonnenmacher:
> PS: If you use GitHub's new "Update with rebase" on this, existing approvals do not get lost.

Nice!

sschuberth (approved):

PR #5209, opened 2022-03-30
ScanCodeResultParser: Fix the license mapping to NOASSERTION

sschuberth:

### DIFF
--- a/utils/spdx/src/main/kotlin/Extensions.kt
+++ b/utils/spdx/src/main/kotlin/Extensions.kt
@@ -86,3 +86,37 @@ fun String.isSpdxExpressionOrNotPresent(): Boolean =
  */
 fun String.toSpdx(strictness: Strictness = Strictness.ALLOW_ANY): SpdxExpression =
     SpdxExpression.parse(this, strictness)
+
+/**
+ * Convert a [String] to an SPDX "idstring" (like license IDs, package IDs, etc.) which may only contain letters,
+ * numbers, ".", and / or "-". If [allowPlusSuffix] is enabled, a "+" (as used in license IDs) is kept as the suffix.
+ */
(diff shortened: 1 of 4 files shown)


utils/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -86,3 +86,30 @@ fun String.isSpdxExpressionOrNotPresent(): Boolean =
...
+            // Take allowed chars as-is.
+            in special, in '0'..'9', in 'A'..'Z', in 'a'..'z' -> id + c
+
+            // Replace colons and underscores with dashes, but do not allow consecutive special chars for readability.
+            ':', '_' -> if (id.lastOrNull() !in special) "$id-" else id
+
+            // Replace anything else with dots, but do not allow consecutive special chars for readability.
+            else -> if (id.lastOrNull() !in special) "$id." else id
+        }
+    }
```
mnonnenmacher, line 109: As this function is probably called very often it might be worth to optimize it using a string builder. For example:
```kotlin
val converted = buildString {
  var lastChar: Char? = null
  forEachIndexed { index, c ->
    when (c) {
      in special, in '0'..'9', in 'A'..'Z', in 'a'..'z' -> c
      ':', '_' -> if (lastChar !in special) '-' else null
      else -> if (lastChar !in special) '.' else null
    }?.let { c ->
      append(c)
      lastChar = c
    }
  }
}
```

mnonnenmacher (changes requested):

mnonnenmacher (approved):

sschuberth:
Merging despite the unrelated failure in `PipFunTest` (which will get fixed by https://github.com/oss-review-toolkit/ort/pull/5211).

PR #5219, opened 2022-04-04
MavenFunTest: Update the binary artifact URL for "version-number"

sschuberth:
It is unclear why this has changed.


sschuberth:
The need for this seems to have magically disappeared.

PR #5220, opened 2022-04-04
OrtResult: Refactor the `includeSubProjects` parameter to `getProjects()`

sschuberth:
The filtering for sub-projects should already happen in `getProjects()`
itself as that is more fitting.

### DIFF
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -211,18 +211,9 @@ data class OrtResult(
      */
     fun collectProjectsAndPackages(includeSubProjects: Boolean = true): Set<Identifier> {
         val projectsAndPackages = mutableSetOf<Identifier>()
-        val projects = getProjects()
+        val projects = getProjects(includeSubProjects = includeSubProjects)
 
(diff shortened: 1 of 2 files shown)


mnonnenmacher (approved):

model/src/main/kotlin/OrtResult.kt
```diff
@@ -211,18 +211,9 @@ data class OrtResult(
      */
     fun collectProjectsAndPackages(includeSubProjects: Boolean = true): Set<Identifier> {
         val projectsAndPackages = mutableSetOf<Identifier>()
-        val projects = getProjects()
+        val projects = getProjects(includeSubProjects)
```
mnonnenmacher, line 214: You need to use the named argument, otherwise you are using the `omitExcluded` parameter.

mnonnenmacher (changes requested):

model/src/main/kotlin/OrtResult.kt
sschuberth: Doh, thanks 😆

mnonnenmacher (approved):

PR #5222, opened 2022-04-05
Add a root `.ort.yml` to sparse checkouts

sschuberth:

### DIFF
--- a/downloader/src/main/kotlin/VersionControlSystem.kt
+++ b/downloader/src/main/kotlin/VersionControlSystem.kt
@@ -145,9 +145,9 @@ abstract class VersionControlSystem {
         }
 
         /**
-         * Return glob patterns matching all potential license or patent files.
+         * Return glob patterns for files that should be checkout out in addition to explicit sparse checkout paths.
          */
-        internal fun getLicenseFileGlobPatterns(): List<String> =
+        internal fun getSparseCheckoutGlobPatterns(): List<String> =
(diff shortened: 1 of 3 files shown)


downloader/src/main/kotlin/VersionControlSystem.kt
```diff
@@ -147,8 +148,11 @@ abstract class VersionControlSystem {
         /**
          * Return glob patterns for files that should be checkout out in addition to explicit sparse checkout paths.
          */
-        internal fun getSparseCheckoutGlobPatterns(): List<String> =
-            LicenseFilenamePatterns.getInstance().allLicenseFilenames.generateCapitalizationVariants().map { "**/$it" }
+        internal fun getSparseCheckoutGlobPatterns(): List<String> {
+            val globPatterns = mutableListOf(ORT_REPO_CONFIG_FILENAME)
```
mnonnenmacher, line 152: For cases where multiple .ort.yml files are used we could use a pattern like `"*$ORT_REPO_CONFIG_FILENAME"`, then it would match all files as long as they follow the convention to end with `.ort.yml`.

mnonnenmacher (changes requested):

downloader/src/main/kotlin/VersionControlSystem.kt
sschuberth: Good idea, will do.

mnonnenmacher (approved):

PR #5223, opened 2022-04-05
Fix the evaluator's `hasLicense()` for "not present" values 

sschuberth:

### DIFF
--- a/utils/spdx/src/main/kotlin/SpdxExpression.kt
+++ b/utils/spdx/src/main/kotlin/SpdxExpression.kt
@@ -555,7 +555,7 @@ class SpdxLicenseIdExpression(
     val id: String,
     val orLaterVersion: Boolean = false
 ) : SpdxSimpleExpression() {
-    private val spdxLicense = SpdxLicense.forId(id)
+    private val spdxLicense by lazy { SpdxLicense.forId(id) }
 
     override fun decompose() = setOf(this)
 
--- a/utils/spdx/src/main/kotlin/SpdxExpression.kt
+++ b/utils/spdx/src/main/kotlin/SpdxExpression.kt
@@ -158,6 +158,12 @@ sealed class SpdxExpression {
     protected open fun validChoicesForDnf(): Set<SpdxExpression> = setOf(this)
 
     /**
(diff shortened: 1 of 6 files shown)


utils/spdx/src/main/kotlin/SpdxExpression.kt
```diff
@@ -157,6 +157,12 @@ sealed class SpdxExpression {
      */
     protected open fun validChoicesForDnf(): Set<SpdxExpression> = setOf(this)
 
+    /**
+     * Return whether this expression contains present licenses, i.e. not all licenses in this expression are "not
```
mnonnenmacher, line 161: Use `"contains [present][SpdxConstants.isPresent] licenses"`, to help users that do not know what a "present" license is.

utils/spdx/src/main/kotlin/SpdxExpression.kt
```diff
@@ -571,8 +569,9 @@ class SpdxLicenseIdExpression(
     override fun validate(strictness: Strictness) {
         val isValid = SpdxConstants.isNotPresent(id) || when (strictness) {
             Strictness.ALLOW_ANY -> true
-            Strictness.ALLOW_DEPRECATED -> spdxLicense != null
-            Strictness.ALLOW_CURRENT, Strictness.ALLOW_LICENSEREF_EXCEPTIONS -> spdxLicense?.deprecated == false
+            Strictness.ALLOW_DEPRECATED -> SpdxLicense.forId(id) != null
+            Strictness.ALLOW_CURRENT, Strictness.ALLOW_LICENSEREF_EXCEPTIONS ->
+                SpdxLicense.forId(id)?.deprecated == false
```
mnonnenmacher, line 574: A lazy property has the benefit that it is only executed once if `validate` is called multiple times.

mnonnenmacher (changes requested):

utils/spdx/src/main/kotlin/SpdxExpression.kt
sschuberth: It's probably not very likely that `validate()` is called multiple times for the same instance, but still done.

mnonnenmacher (approved):

PR #5225, opened 2022-04-05
SpdxDocumentModelMapper: Make SPDX "idstring" generation predictable

sschuberth:
Being able to predict the SPDX "idstring" for a given package is a
prerequisite for upcoming changes that will maintain the transitive
package relationships.

The SPDX ID is now derived from the coordinate representation of a
project's / package's `Identifier`. As the `Identifier` is unique within
an `OrtResult`, the derived SPDX ID is very likely unique, too, except
for cases where coordinate representations differ only in special
characters that get mapped to the same valid character for an SPDX ID.

### DIFF
--- a/reporter/src/main/kotlin/reporters/spdx/SpdxDocumentModelMapper.kt
+++ b/reporter/src/main/kotlin/reporters/spdx/SpdxDocumentModelMapper.kt
@@ -49,6 +49,7 @@ import org.ossreviewtoolkit.utils.spdx.model.SpdxExtractedLicenseInfo
 import org.ossreviewtoolkit.utils.spdx.model.SpdxPackage
 import org.ossreviewtoolkit.utils.spdx.model.SpdxPackageVerificationCode
 import org.ossreviewtoolkit.utils.spdx.model.SpdxRelationship
+import org.ossreviewtoolkit.utils.spdx.toSpdxId
 
 /**
  * A class for mapping [OrtResult]s to [SpdxDocument]s.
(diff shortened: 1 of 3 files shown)


sschuberth:
@fviernau I hope this aligns with what we discussed.

reporter/src/main/kotlin/reporters/spdx/SpdxDocumentModelMapper.kt
```diff
@@ -66,12 +67,12 @@ object SpdxDocumentModelMapper {
...
         params: SpdxDocumentParams
     ): SpdxDocument {
-        val spdxPackageIdGenerator = SpdxPackageIdGenerator()
         val packages = mutableListOf<SpdxPackage>()
         val relationships = mutableListOf<SpdxRelationship>()
 
+        val rootPackages = ortResult.getProjects(omitExcluded = true, includeSubProjects = false)
         val rootPackage = SpdxPackage(
-            spdxId = spdxPackageIdGenerator.nextId("root-package"),
+            spdxId = rootPackages.first().id.toSpdxId("Project"),
```
mnonnenmacher, line 75: It looks like only packages are added to the SPDX document, not projects. If this is correct it would be better to use a generic name here like just "root-package" to not give the impression that the packages are all dependencies of the randomly selected project. But then also previously the `nextId` call would not have been required.

sschuberth: > It looks like only packages are added to the SPDX document, not projects.

That will actually change with https://github.com/oss-review-toolkit/ort/pull/5126 that I need to rebase onto this one after merge.

mnonnenmacher (approved):

tsteenbe:
Am I correct that we now write use "first" instead of "1"

tsteenbe (approved):

sschuberth:
> Am I correct that we now write use "first" instead of "1"

This has been clarified in today's ORT developer meeting. The short answer is "no" 😀

PR #5226, opened 2022-04-05
Parse NPM packages in parallel

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -30,6 +30,11 @@ import java.io.File
 import java.io.IOException
 import java.util.SortedSet
 
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.runBlocking
+
 import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
 import org.ossreviewtoolkit.analyzer.PackageManager
 import org.ossreviewtoolkit.analyzer.PackageManagerResult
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -267,11 +271,20 @@ open class Npm(
...
+        runBlocking(Dispatchers.IO) {
+            nodeModulesFiles.map { file ->
+                this@Npm.log.debug { "Starting to parse '$file'..." }
+                async { file to parsePackage(rootDirectory, file) }
+            }.forEach { deferredResult ->
+                val (file, result) = deferredResult.await()
+                val (id, pkg) = result
+                this@Npm.log.debug { "Finished parsing '$file' to '$id'." }
+                packages[id] = pkg
+            }
```
mnonnenmacher, line 287: This could be simplified a bit:

```kotlin
        return runBlocking {
            nodeModulesDir.walk().filter {
                it.name == "package.json" && isValidNodeModulesDirectory(nodeModulesDir, nodeModulesDirForPackageJson(it))
            }.map { file ->
                async {
                    this@Npm.log.debug { "Starting to parse '$file'..." }
                    parsePackage(rootDirectory, file).also { (id, _) ->
                        this@Npm.log.debug { "Finished parsing '$file' to '$id'." }
                    }
                }
            }.toList().awaitAll().toMap()
        }
```

sschuberth: I basically had it like that before, but was bewildered by not seeing any concurrency at all according to log output, so I wanted to put the "Finished" log after the `await()` for the respective file to see if that makes a change. But it didn't 😟

mnonnenmacher (approved):

oheger-bosch:
Could you shortly comment about the performance gain caused by this change?

sschuberth:
For only the "resolve shrinkwrap dependencies correctly" `NpmFunTest` I saw the runtime going down from ~40s to ~15s.

PR #5229, opened 2022-04-06
scanoss: Replace usages of raw JSON by model classes for ScanResponse

nnobelis:


clients/scanoss/src/main/kotlin/model/Model.kt
```diff
@@ -67,6 +67,17 @@ data class License(
     val osadlUpdated: String? = null
 )
 
+/**
```
sschuberth, line 70: Commit message nits:
- "of *the* scanoss-junit-4.12.json` test file"
- "by *the* SCANOSS team"
- "in a new commit" is clear and could be dropped IMO

clients/scanoss/src/main/kotlin/ScanOssService.kt
```diff
@@ -34,6 +34,8 @@ import retrofit2.http.Multipart
 import retrofit2.http.POST
 import retrofit2.http.Part
 
+typealias FullScanResponse = Map<String, List<ScanResponse>>
```
sschuberth, line 37: Please move the introduction of this `typealias` to a separate preceding commit, so that this commit really only touches files in the "scanner" directory.

scanner/build.gradle.kts
```diff
@@ -85,3 +90,11 @@ buildConfig {
     buildConfigField("String", "SCANCODE_VERSION", "\"$scancodeVersion\"")
     buildConfigField("String", "SCANOSS_VERSION", "\"$scanossVersion\"")
 }
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
```
sschuberth, line 94: Please use an import instead of a full qualified name (also see #5234).

scanner/build.gradle.kts
```diff
@@ -85,3 +90,11 @@ buildConfig {
...
+    if (name.contains("Test")) {
```
sschuberth, line 95: I believe it's fine to unconditionally do this for all tasks. That would also align with existing similar code.

sschuberth: Speaking of aligning code, please also use the `customCompilerArgs` approach here.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -120,16 +115,11 @@ class ScanOss internal constructor(
                 "The $scannerName server returned an UUID not present in the mapping."
             }
 
-            fileNamesAnonymizationMapping[uuid] to entry.value
+            fileNamesAnonymizationMapping[uuid]!! to entry.value
```
sschuberth, line 118: Instead of introducing `!!`, how about (in a preceeding commit) replacing the code incl. `require()` above with

            val fileName = fileNamesAnonymizationMapping[uuid] ?: throw IllegalArgumentException(
                "The $scannerName server returned an UUID not present in the mapping."
            )

            fileName to entry.value

scanner/src/test/assets/scanoss-junit-4.12.json
```diff
@@ -2,6 +2,22 @@
   "/tmp/ort-ScanOss2759786101559527642/Maven/junit/junit/4.12/to-do.txt": [
     {
       "id": "none",
+      "status": "pending",
```
sschuberth, line 5: Why was this file modified? Where do these changes come from? Maybe do this in a separate preceding commit with more explanations in the commit message?

scanner/src/test/kotlin/scanners/scanoss/ScanOssResultParserTest.kt
```diff
@@ -28,17 +28,23 @@ import io.kotest.matchers.should
...
+import kotlinx.serialization.json.decodeFromStream
+
+import org.ossreviewtoolkit.clients.scanoss.FullScanResponse
 import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.TextLocation
-import org.ossreviewtoolkit.model.readJsonFile
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 
+private val json = Json { ignoreUnknownKeys = true }
```
sschuberth, line 40: Please use `ScanOssService.JSON` instead.

sschuberth (changes requested):

scanner/build.gradle.kts
nnobelis: > I believe it's fine to unconditionally do this for all tasks. That would also align with existing similar code.

It does not work if I remove the conditional: the `kotlinx` dependency being a test dependency (no release runtime code uses it), it tries to build non-test classes with the `optIn` and fails.

sschuberth: Interesting. Then please at lest use `"test" in name.lowerCase()` instead.

nnobelis: Done but Intellij does not recognize `lowerCase`: I am forced to use the Java version. Any idea ?

sschuberth: Ah, might be that for Gradle KTS files we're stuck with a Kotlin version where it's still called `toLowerCase()`.

nnobelis: Can we merge this ?

sschuberth (approved):

PR #5233, opened 2022-04-06
PackageConfiguration: Support matching any `revision`

sschuberth:

### DIFF
--- a/model/src/main/kotlin/config/PackageConfiguration.kt
+++ b/model/src/main/kotlin/config/PackageConfiguration.kt
@@ -98,17 +98,19 @@ data class VcsMatcher(
     val url: String,
 
     /**
-     * The [revision] to match for equality against [RepositoryProvenance.resolvedRevision].
+     * The [revision] to match for equality against [RepositoryProvenance.resolvedRevision], or null to match any
+     * revision.
      */
-    val revision: String
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    val revision: String? = null
 ) {
     init {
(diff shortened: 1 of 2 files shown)


sschuberth:
@fviernau, after another discussion with @mnonnenmacher we decided to propose this change *without* hiding it behind a feature toggle. The reason mostly is that any such conventions like requiring a fixed `revision` could (and probably should) be implemented via CI checks on package configuration changes. That way, we avoid an increasing number of feature toggles being introduced, esp. before have a proper mechanism to share global configuration throughout all of ORT without passing config options everywhere.

So, what do you think @fviernau?

model/src/main/kotlin/config/PackageConfiguration.kt
```diff
@@ -98,17 +98,19 @@ data class VcsMatcher(
     val url: String,
 
     /**
-     * The [revision] to match for equality against [RepositoryProvenance.resolvedRevision].
+     * The [revision] to match for equality against [RepositoryProvenance.resolvedRevision], or null to match any
+     * revision.
      */
-    val revision: String
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    val revision: String? = null
```
mnonnenmacher, line 105: Please also update `docs/config-file-package-configuration-yml.md`, maybe just add "(optional)" before the one mention of the revision property.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

tsteenbe (approved):

PR #5239, opened 2022-04-08
SpdxDocumentCache: Log the file name before loading it

oheger-bosch:
If a project contains multiple SPDX files and one of these is invalid
(so that it cannot be loaded), there is no indication in the log,
which file is affected. Therefore, add a log statement allowing to
keep track on the files that are processed.


sschuberth (approved):

sschuberth:
Merging despite the unrelated `NuGetFunTest` timeout failure.

PR #5241, opened 2022-04-08
ExperimentalScanner fixes

mnonnenmacher:
Some fixes for the experimental scanner, see the commit messages for details.

### DIFF
--- a/scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
+++ b/scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
@@ -130,13 +130,7 @@ data class NestedProvenanceScanResult(
         return findings
     }
 
-    private fun getPath(provenance: KnownProvenance): String {
-        if (provenance == nestedProvenance.root) return ""
-
-        nestedProvenance.subRepositories.forEach { if (provenance == it.value) return it.key }
-
-        throw IllegalArgumentException("Could not find entry for $provenance.")
-    }
+    private fun getPath(provenance: KnownProvenance) = nestedProvenance.getPath(provenance)
 
     fun filterByIgnorePatterns(ignorePatterns: List<String>): NestedProvenanceScanResult =
         copy(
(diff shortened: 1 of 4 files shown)


scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
```diff
@@ -167,10 +167,15 @@ data class NestedProvenanceScanResult(
...
             val provenancePath = getPath(it)
-            provenancePath.isEmpty() || path.startsWith("$provenancePath/")
+            provenancePath.isEmpty() || provenancePath == path || provenancePath.startsWith("$path/") || path.startsWith("$provenancePath/")
         }.associateWith { provenance ->
             val provenancePath = getPath(provenance)
-            path.removePrefix("$provenancePath/")
+
+            when {
+                provenancePath.isEmpty() -> path
+                path.startsWith("$provenancePath/") -> path.removePrefix("$provenancePath/")
```
sschuberth, line 171: Just checking: We don't need the the same logic for `"$path/"`, right?

sschuberth (dismissed):

scanner/src/main/kotlin/experimental/NestedProvenanceScanResult.kt
mnonnenmacher: No, above the check is reciprocal to figure out if the subrepository is on the same branch as the provided filter path, but for getting the relative path within the subrepository we don't need it.

sschuberth: Thanks for confirming. Would it make sense to put that sentence of yours into a code comment above the `when`?

mnonnenmacher: I have added a few explanatory code comments.

sschuberth (approved):

sschuberth:
FYI @mnonnenmacher, we currently have issues with the Markdown check / GitHub's checkout action, so that test is expected to fail.

PR #5243, opened 2022-04-08
Work around Artifactory returning API URLs instead of download URLs for NPM

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -318,7 +318,6 @@ open class Npm(
      * Construct a [Package] by parsing its _package.json_ file and - if applicable - querying additional
      * content via the `npm view` command. The result is a [Pair] with the raw identifier and the new package.
      */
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -196,6 +196,7 @@ open class Npm(
         }
     }
 
+    private val artifactoryApiPathPattern = Regex("(.*artifactory.*)(?:/api/npm/)(.*)")
```
MarcelBochtler, line 199: What do you think about adding `.*jfrog.*` as an alternative match?

MarcelBochtler (approved):

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: Do we have a real-life example for that?

MarcelBochtler: No.

PR #5249, opened 2022-04-12
Miscellaneous minor formatting stuff

sschuberth:

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -24,8 +24,28 @@ Files: mlc_config.json
 Copyright: 2017-2020 HERE Europe B.V. <opensource@here.com>
 License: Apache-2.0
 
+Files: */src/main/resources/*
+Copyright: 2017-2020 HERE Europe B.V. <opensource@here.com>
+License: Apache-2.0
+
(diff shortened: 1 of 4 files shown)


.reuse/dep5
```diff
@@ -28,31 +28,35 @@ Files: ci/*
 Copyright: 2017-2020 HERE Europe B.V. <opensource@here.com>
 License: Apache-2.0
 
-Files: docs/*
-Copyright: 2017-2020 HERE Europe B.V. <opensource@here.com>
+Files: clients/fossid-webapp/src/test/assets/*
```
MarcelBochtler, line 31: As the order does change the semantics, do you know if there is a way to compare the assigned licenses before and after your change using the reuse-tool?

sschuberth: > As the order does change the semantics

Very good point, thanks for paying attention here!

> do you know if there is a way to compare the assigned licenses before and after your change using the reuse-tool?

I just simply ran `reuse spdx` before and afterwards, and compared the output. And I indeed discovered and fixed an issue (also see the updated commit message).

PR #5256, opened 2022-04-13
Notifier: Add `--resolutions-file` option

MarcelBochtler:


notifier/src/main/kotlin/Notifier.kt
```diff
@@ -31,17 +31,22 @@ import kotlin.script.experimental.jvmhost.createJvmCompilationConfigurationFromT
...
+import org.ossreviewtoolkit.model.utils.ResolutionProvider
 import org.ossreviewtoolkit.notifier.modules.JiraNotifier
 import org.ossreviewtoolkit.notifier.modules.MailNotifier
 import org.ossreviewtoolkit.utils.scripting.ScriptRunner
 
 class Notifier(
     ortResult: OrtResult = OrtResult.EMPTY,
-    config: NotifierConfiguration = NotifierConfiguration()
+    config: NotifierConfiguration = NotifierConfiguration(),
+    resolutionProvider: ResolutionProvider = DefaultResolutionProvider.create(ortResult, null)
```
sschuberth, line 43: You can omit the `null` here. And I'm wondering wether this should be just `DefaultResolutionProvider()` to align with https://github.com/oss-review-toolkit/ort/blob/5e3817234d08775de27f99353009daea6687a5d9/reporter/src/main/kotlin/ReporterInput.kt#L59. 

But then again, I guess we should consider dropping *all* these default values from here (and `ReporterInput`, and probably more places) to force the caller to think about the arguments. This could be done in a follow-up, though.

sschuberth (changes requested):

sschuberth (approved):

sschuberth (approved):

PR #5272, opened 2022-04-21
MavenSupport: Fixup VCS paths that contain the project name as a prefix

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -236,7 +236,18 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
                             }
                         }
 
-                        else -> VcsInfo(type = VcsType(type), url = url, revision = tag)
+                        else -> {
+                            VcsHost.fromUrl(url)?.let { host ->
+                                host.toVcsInfo(url)?.let { vcsInfo ->
+                                    // Fixup paths that are specified as part of the URL and contain the project name as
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -236,7 +236,18 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
                             }
                         }
 
-                        else -> VcsInfo(type = VcsType(type), url = url, revision = tag)
+                        else -> {
+                            VcsHost.fromUrl(url)?.let { host ->
```
sschuberth, line 240: FYI @MarcelBochtler and @mnonnenmacher, I didn't see a need to limit the logic to GitHub in the end has `VcsHost` allows to implement this nicely in a generic way, this could very well happen also e.g. on Bitbucket, and most of the code is on GitHub anyway (so not handling the rest would not really shield us against the logic falsely kicking in).

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -236,7 +236,18 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
                             }
                         }
 
-                        else -> VcsInfo(type = VcsType(type), url = url, revision = tag)
```
MarcelBochtler, line 239: Can you please add the reason why this is required to the commit message?

MarcelBochtler (approved):

PR #5282, opened 2022-04-25
Utils: Allow `filterVersionNames()` to strip some prefix and suffix strings

sschuberth:

### DIFF
--- a/utils/core/src/test/kotlin/UtilsTest.kt
+++ b/utils/core/src/test/kotlin/UtilsTest.kt
@@ -22,6 +22,8 @@ package org.ossreviewtoolkit.utils.core
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.inspectors.forAll
 import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.collections.shouldContainExactly
+import io.kotest.matchers.collections.shouldHaveSingleElement
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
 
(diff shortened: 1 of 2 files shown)


utils/core/src/test/kotlin/UtilsTest.kt
```diff
@@ -75,6 +75,35 @@ class UtilsTest : WordSpec({
...
+                "REL_3_17_BETA1",
+                "REL_3_16_FINAL",
+                "REL_3_16_BETA2",
+                "REL_3_15_BETA1",
+                "REL_3_16_BETA1",
+                "REL_3_15_FINAL",
+                "REL_3_15_BETA3"
+            )
+
+            filterVersionNames("4.0.0", names) shouldHaveSingleElement "REL_4_0_0_FINAL"
```
mnonnenmacher, line 102: For a version like 3.16 it would still fail to find a tag because of the multiple suffixes?

utils/core/src/test/kotlin/UtilsTest.kt
```diff
@@ -75,6 +75,35 @@ class UtilsTest : WordSpec({
...
+                "REL_4_0_1",
+                "REL_4_0_0_FINAL",
+                "REL_3_17_FINAL",
+                "REL_3_17_BETA1",
+                "REL_3_16_FINAL",
+                "REL_3_16_BETA2",
+                "REL_3_15_BETA1",
+                "REL_3_16_BETA1",
+                "REL_3_15_FINAL",
+                "REL_3_15_BETA3"
```
mnonnenmacher, line 99: You could remove many entries from that list without making the test worse.

mnonnenmacher (changes requested):

utils/core/src/test/kotlin/UtilsTest.kt
sschuberth: Ok, I've done that.

utils/core/src/test/kotlin/UtilsTest.kt
sschuberth: No, it would actually succeed (as not both prefix and suffix are part of `ignorablePrefixSuffixPattern`). I've added a test to demonstrate that.

mnonnenmacher (approved):

PR #5286, opened 2022-04-26
Improve mapping of declared licenses

sschuberth:

### DIFF
--- a/utils/common/src/test/kotlin/ExtensionsTest.kt
+++ b/utils/common/src/test/kotlin/ExtensionsTest.kt
@@ -403,6 +403,33 @@ class ExtensionsTest : WordSpec({
         }
     }
 
+    "String.unquote" should {
+        "remove surrounding quotes" {
+            "'single'".unquote() shouldBe "single"
+            "\"double\"".unquote() shouldBe "double"
+        }
+
+        "remove nested quotes" {
+            "'\"single-double\"'".unquote() shouldBe "single-double"
+            "\"'double-single'\"".unquote() shouldBe "double-single"
+        }
+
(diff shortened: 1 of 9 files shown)


utils/common/src/test/kotlin/ExtensionsTest.kt
```diff
@@ -403,6 +403,33 @@ class ExtensionsTest : WordSpec({
...
+            "\"double-unmatched".unquote() shouldBe "double-unmatched"
+            "'\"broken-nesting'\"".unquote() shouldBe "broken-nesting"
+        }
+
+        "remove whitespace by default" {
+            "  '  \"  single-double  \"  '  ".unquote() shouldBe "single-double"
+        }
+
+        "keep whitespace optionally" {
+            "  '  \"  single-double  \"  '  ".unquote(trimWhitespace = false) shouldBe "  '  \"  single-double  \"  '  "
```
mnonnenmacher, line 428: I don't understand why quotes are kept in this case?

mnonnenmacher (changes requested):

utils/common/src/test/kotlin/ExtensionsTest.kt
sschuberth: Whitespaces are the outermost / surrounding characters here. As `unquote` uses `trim` internally, and `trim` only looks at outermost characters, trimming basically immediately stops / does nothing because the outermost characters are not in the list of characters to trim.

I was actually a bit undecided how `unquote` *should* behave in this case, and just wanted to document via this test case how it *does* currently behave. But if you find this highly counter-intuitive, I could change the implementation.

PR #5291, opened 2022-04-27
fossid-webapp: Add a generateReport function to the service

nnobelis:
Refer to individual commit for details.


clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -302,3 +309,54 @@ suspend fun FossIdRestService.createIgnoreRule(
...
+            apiKey,
+            "scan_code" to scanCode,
+            "report_type" to reportType.name.lowercase(),
+            "selection_type" to selectionType.name.lowercase()
+        )
+    )
+
+    require(response.isSuccessful) {
+        """ ${response.code()} - ${response.message()}
+            ${response.errorBody()?.string()}
```
sschuberth, line 334: Formatting looks a bit odd, why is there a single space after `"""`? Also, could you please add something about the context? Like "Report generation failed with ...".

clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -302,3 +309,54 @@ suspend fun FossIdRestService.createIgnoreRule(
...
+        """.trimIndent()
+    }
+
+    // The API is quirky here: if the report type is HTML, the result is returned as a standard HTML response, without
```
sschuberth, line 338: Nit: Capitalized "If" after the colon.

clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -302,3 +309,54 @@ suspend fun FossIdRestService.createIgnoreRule(
...
+    // (even for report types generating HTML such as DYNAMIC).
+    val fileName = if (reportType == ReportType.HTML) "${UUID.randomUUID()}-report.html"
+    else {
+        response.headers()["Content-disposition"]?.split(';')?.firstNotNullOfOrNull {
+            it.trim().withoutPrefix("filename=")?.removeSurrounding("\"")
+        } ?: throw IllegalStateException("Cannot determine name of the report")
+    }
+
+    val tempDir = createTempDirectory("fossid_report").toFile()
+        .apply { deleteOnExit() }
```
sschuberth, line 349: That's dirty. IMO we should rather add a parameter for the parent directory to this function.

clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -302,3 +309,54 @@ suspend fun FossIdRestService.createIgnoreRule(
...
+            it?.inputStream()?.copyTo(outputStream())
+        }
+    }
+}
+
+/**
+ * If this string starts with [prefix], return the string without the prefix, otherwise return [missingPrefixValue].
+ */
+fun String?.withoutPrefix(prefix: String, missingPrefixValue: () -> String? = { null }): String? =
+    this?.removePrefix(prefix)?.takeIf { it != this } ?: missingPrefixValue()
```
sschuberth, line 362: I wonder whether we should start to allow a dependency on ORT's `common-utils` in `clients` to avoid such code duplication...

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/Extensions.kt
nnobelis: I notice you removed the dependency in a previous commit, hence the duplication....

sschuberth: Do you mean 5423682cb5a9221b1b5d0b1c06cb214c81dfdf05? That code shouldn't even have compiled, as there was (and is) no dependency on ORT's libraries in `fossid-webapp`'s `build.gradle.kts`. And in the past [we wanted it to be like that](https://github.com/oss-review-toolkit/ort/blob/main/clients/README.md). However, since the split of ORT's utility modules into `common` and `core`, I'm less against adding a dependency from the clients on ORT `common-utils`. What do you think @mnonnenmacher?

mnonnenmacher: I would be fine with adding that dependency. However, it could also be done after this PR to not block it.

sschuberth: Yes, agreed, we should do that afterwards and remove duplicate code from *all* client libraries at once in favor of introducing that dependency.

clients/fossid-webapp/src/test/kotlin/FossIdReportTest.kt
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    "A dynamic HTML report can be generated for a scan" {
+        service.generateReport("", "", SCAN_CODE, ReportType.DYNAMIC, SelectionType.INCLUDE_FOSS, directory)
+            .let { result ->
+                result.shouldBeSuccess {
+                    it shouldHaveName
+                       "fossid-semver4j_semver4j__20220119_094101_delta-cc1267688905d7493df292786a245297c9fd36ee.html"
+                    val lines = it.readLines()
+                    lines shouldContain "<html>"
```
sschuberth, line 72: Do we really need to read all lines here? How about something like `file.useLines { it shouldContain "<html>" }` instead, which uses a sequence?

clients/fossid-webapp/src/test/kotlin/FossIdReportTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        server.stop()
+    }
+
+    beforeTest {
+        server.resetAll()
+    }
+
+    "A dynamic HTML report can be generated for a scan" {
+        service.generateReport("", "", SCAN_CODE, ReportType.DYNAMIC, SelectionType.INCLUDE_FOSS, directory)
+            .let { result ->
```
sschuberth, line 67: For readability (and to maintain the Arrange/Act/Assert pattern), I think I'd prefer to reduce indentation and not use `let` here, like

```
val result = service.generateReport(...)

result.shouldBeSuccess {
...
```

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -302,3 +310,62 @@ suspend fun FossIdRestService.createIgnoreRule(
...
+                    response.body()?.use { target.writeAll(it.source()) }
+                }
+            }
+        )
+    } else {
+        Result.failure(
+            IllegalStateException(
+                """Report generation failed with error code ${response.code()}: ${response.message()}.
+                Details: ${response.errorBody()?.string()}
+                """.trimIndent()
```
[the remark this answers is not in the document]
nnobelis: That's why I originally put a line return after the opening quotes
```
    """
    Report ....
```
But you complained that this looked wrong. What would you do here ?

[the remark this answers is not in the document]
sschuberth: Please simply have a look at existing code like

https://github.com/oss-review-toolkit/ort/blob/099677ef14d62cf6f55a73e4384fa66de1cd24f8/analyzer/src/test/kotlin/managers/GoModTest.kt#L95-L104

As blank lines are ignored when calculating the minimum indent, this should work as expected.

clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -302,3 +310,62 @@ suspend fun FossIdRestService.createIgnoreRule(
...
+            "report_type" to reportType.name.lowercase(),
+            "selection_type" to selectionType.name.lowercase()
+        )
+    )
+
+    return if (response.isSuccessful) {
+        // The API is quirky here: If the report type is HTML, the result is returned as a standard HTML response,
+        // without filename. For any other report types, the result is returned as an attachment with a filename in the
+        // headers (even for report types generating HTML such as DYNAMIC).
+        val fileName = if (reportType == ReportType.HTML) "${UUID.randomUUID()}-$scanCode-report.html"
```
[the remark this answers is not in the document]
nnobelis: To have a unique report name ?
If you query different reports for the same scan, they would overwrite themselves ?
(but in really they don't because for non-HTML reports, FossID generates a unique identifier in the file name).
So you could see this as a safeguard.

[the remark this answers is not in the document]
sschuberth: I'm not sure what you mean by "different reports". Do you mean different *types* of reports? If so, they should be deduplicated by the file extension.

Or you you saying that even when sticking to a HTML report for the same `scanCode`, the report could vary over time? I'm sorry if my FossID knowledge is lacking here, but I thought a `scanCode` uniquely identified a fixed scan and its result, thus requesting a report for a given `scanCode` should always yield the same result. Is that not the case?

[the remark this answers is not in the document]
sschuberth: Anyway, can we somehow reverse-engineer how the unique ID for the other report formats (which seems to be a SHA1) is calculated by FossID, and do the same for HTML reports? I'd strongly prefer that over a random UUID, because requesting the same report for the same `scanCode` without any changes to the scan should create exactly the same report, including the filename.

clients/fossid-webapp/src/test/kotlin/FossIdReportTest.kt
```diff
@@ -0,0 +1,97 @@
...
+                result.shouldBeSuccess {
+                    it shouldHaveName
+                       "fossid-semver4j_semver4j__20220119_094101_delta-cc1267688905d7493df292786a245297c9fd36ee.html"
```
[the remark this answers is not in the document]
nnobelis: I replaced with a string concat, Please advise if it not ok.

clients/fossid-webapp/src/main/kotlin/Extensions.kt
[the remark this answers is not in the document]
nnobelis: > I'm not sure what you mean by "different reports". Do you mean different _types_ of reports? If so, they should be deduplicated by the file extension.

Yes I mean report type but unfortunately extension is not enough because both `HTML` and `DYNAMIC` types are html.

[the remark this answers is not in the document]
nnobelis: However I agree with the SHA1 this is a go idea.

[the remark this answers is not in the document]
nnobelis: This would require too much changes to the current logic because we get the file contents only after the fine name has been calculated.
Therefore I removed the unique ID altogether.

clients/fossid-webapp/src/main/kotlin/model/report/Model.kt
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/**
+ * License types to be included in report.
+ */
+enum class SelectionType {
+    INCLUDE_ALL_LICENSES,
+    INCLUDE_COPYLEFT,
+    INCLUDE_FOSS,
+    INCLUDE_MARKED_LICENSES
```
sschuberth, line 39: Please also add docs about these, even if there are no official ones, based on your experience / assumptions what these mean.

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/model/report/Model.kt
```diff
@@ -0,0 +1,40 @@
...
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.clients.fossid.model.report
+
+/**
+ * Available FossID report types.
+ */
+enum class ReportType {
+    DYNAMIC,
```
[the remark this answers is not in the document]
nnobelis: In the end I noticed that the report types in the API matched 1:1 the ones in the UI, hence I could copy the documentation.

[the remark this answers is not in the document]
nnobelis: Please note that, due to the way the parameters are passed to FossID, I used the `toString?  technique instead of JSON annotations.

clients/fossid-webapp/src/main/kotlin/model/report/Model.kt
```diff
@@ -0,0 +1,75 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.clients.fossid.model.report
+
+/**
+ * Available FossID report types.
+ */
+enum class ReportType(private val jsonRepresentation: String) {
```
sschuberth, line 25: I believe we shouldn't bee too specific here that this is being serialized to JSON. What about just `value`, `apiName` or the like?

clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -302,3 +309,63 @@ suspend fun FossIdRestService.createIgnoreRule(
...
+            "selection_type" to selectionType.name.lowercase()
+        )
+    )
+
+    return if (response.isSuccessful) {
+        // The API is quirky here: If the report type is HTML, the result is returned as a standard HTML response,
+        // without filename. For any other report types, the result is returned as an attachment with a filename in the
+        // headers (even for report types generating HTML such as DYNAMIC).
+        val fileName = if (reportType == ReportType.HTML_STATIC) "fossid-$scanCode-report.html"
+        else {
```
sschuberth, line 342: One more style nit:

```
val fileName = if (reportType == ReportType.HTML_STATIC) {
    "fossid-$scanCode-report.html"
} else {
...
```

sschuberth (changes requested):

sschuberth (approved):

clients/fossid-webapp/src/test/assets/report/mappings/xlsx_report.json
```diff
@@ -0,0 +1,30 @@
...
+      "ignoreArrayOrder" : true,
+      "ignoreExtraElements" : true
+    } ]
+  },
+  "response" : {
+    "status" : 200,
+    "bodyFileName" : "generate_report-1650976369669.xlsx",
+    "headers" : {
+      "Cache-Control" : "no-cache",
+      "Content-Disposition" : "attachment; filename=semver4j_semver4j__20220119_094101_delta-4d39359cd861c314020ead401d95ca20dc179495.xlsx",
```
sschuberth, line 18: BTW, do you have any idea how the `filename` here relates to the `bodyFileName` in line 15?

nnobelis: The `filename` is the special header returned by FossID.
`bodyFileName` is Wiremock specify trick to reference a file payload to the response your are stubbing. This is mainly used to avoid having binary in the JSON stub.

sschuberth: Thanks for the explanation. So the `bodyFileName` part does not come from FossID itself, that's what I was after.

nnobelis:
@sschuberth The CI failure is unrelated. Can we please merge this ?

sschuberth:
Merging despite the unrelated `PipFunTest` failure.

PR #5299, opened 2022-04-29
Add support for Azure DevOps VcsHost

MarcelBochtler:


utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -464,3 +464,12 @@ fun Throwable.collectMessages(): List<String> {
  * Recursively collect the messages of this [Throwable] and all its causes and join them to a single [String].
  */
 fun Throwable.collectMessagesAsString() = collectMessages().joinToString("\nCaused by: ")
+
+/**
+ * Retrieve query parameters specified by its [key].
+ */
+fun URI.getQueryParameterValue(key: String) = query.split("&").flatMap {
```
sschuberth, line 471: Formatting: Please start a new line after `=`.

sschuberth: Please use the single-char version of split for clarity when possible: `split('&')`

utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -464,3 +464,12 @@ fun Throwable.collectMessages(): List<String> {
  * Recursively collect the messages of this [Throwable] and all its causes and join them to a single [String].
  */
 fun Throwable.collectMessagesAsString() = collectMessages().joinToString("\nCaused by: ")
+
+/**
+ * Retrieve query parameters specified by its [key].
```
sschuberth, line 469: You should probably document that this returns an empty list (instead of e.g. null) in case of an absent key.

sschuberth: But actually, we probably *do* want to distinguish between a key being absent, and its value being empty.

utils/common/src/test/kotlin/ExtensionsTest.kt
```diff
@@ -452,4 +453,22 @@ class ExtensionsTest : WordSpec({
...
+            URI("https://oss-review-toolkit.org?key=value1,value2,value3").getQueryParameterValue("key") shouldBe
+                    listOf("value1", "value2", "value3")
+        }
+
+        "work with multiple query parameters" {
+            val uri = URI("https://oss-review-toolkit.org?key1=value1&key2=value2")
+
+            uri.getQueryParameterValue("key1") shouldBe listOf("value1")
+            uri.getQueryParameterValue("key2") shouldBe listOf("value2")
+        }
```
sschuberth, line 472: Please add test cases for absent / empty keys, like in

    URI("https://oss-review-toolkit.org")
    URI("https://oss-review-toolkit.org?key=")

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -44,6 +46,52 @@ enum class VcsHost(
      */
     vararg supportedTypes: VcsType
 ) {
+    AZURE_DEVOPS("dev.azure.com", VcsType.GIT) {
+        private val commitPrefix = "GC"
```
sschuberth, line 50: Maybe name this `gitCommitPrefix`, as I guess that's what "GC" (= "Git Commit") stands for?

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -44,6 +46,52 @@ enum class VcsHost(
...
+
+        override fun getUserOrOrgInternal(projectUrl: URI) = projectUrlToUserOrOrgAndProject(projectUrl)?.first
+
+        override fun getProjectInternal(projectUrl: URI) = projectUrl.path.substringAfterLast("/")
+
+        override fun toVcsInfoInternal(projectUrl: URI): VcsInfo {
+            val url = projectUrl.toString().replace(projectUrl.query.toString(), "").dropLast(1)
+            val revision =
+                projectUrl.getQueryParameterValue("version").firstOrNull()?.withoutPrefix(commitPrefix).orEmpty()
```
sschuberth, line 59: Personally, I'd rather wrap this at the end of the line, grouping chained calls that somewhat belong together, like

            val revision = projectUrl.getQueryParameterValue("version").firstOrNull()
                ?.withoutPrefix(commitPrefix).orEmpty()

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -44,6 +46,52 @@ enum class VcsHost(
...
 ) {
+    AZURE_DEVOPS("dev.azure.com", VcsType.GIT) {
+        private val commitPrefix = "GC"
+
+        override fun getUserOrOrgInternal(projectUrl: URI) = projectUrlToUserOrOrgAndProject(projectUrl)?.first
+
+        override fun getProjectInternal(projectUrl: URI) = projectUrl.path.substringAfterLast("/")
+
+        override fun toVcsInfoInternal(projectUrl: URI): VcsInfo {
+            val url = projectUrl.toString().replace(projectUrl.query.toString(), "").dropLast(1)
```
sschuberth, line 57: So, this simply removes any query string, right? If so, I'd prefer to make the more clear by simply reconstructing an URI like

    URI(it.scheme, it.userInfo, it.host, it.port, it.path, null, it.fragment)

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -44,6 +46,52 @@ enum class VcsHost(
...
+        }
+
+        override fun toPermalinkInternal(vcsInfo: VcsInfo, startLine: Int, endLine: Int): String {
+            val actualEndLine = if (endLine != -1) endLine else startLine + 1
+
+            val lineQueryParam = "line=$startLine&lineEnd=$actualEndLine&lineColumn=1&lineEndColumn=1"
+            val pathQueryParam = if (vcsInfo.path.isNotEmpty()) "&path=/${vcsInfo.path}" else ""
+            val revisionQueryParam = if (vcsInfo.revision.isNotEmpty()) {
+                "&version=$commitPrefix${vcsInfo.revision}"
+            } else ""
```
sschuberth, line 72: Could use a `takeUnless` / `orEmpty` construct.

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -44,6 +46,52 @@ enum class VcsHost(
...
+
+            return "${vcsInfo.url}?$lineQueryParam$pathQueryParam$revisionQueryParam"
+        }
+
+        override fun toArchiveDownloadUrlInternal(userOrOrg: String, project: String, vcsInfo: VcsInfo): String {
+            val team = URI(vcsInfo.url).path.split("/")[2]
```
sschuberth, line 78: Would it make sense to guard against index-out-of-bounds by using a path iterator, maybe? Similar below.

sschuberth (changes requested):

utils/common/src/main/kotlin/Extensions.kt
MarcelBochtler: I changed this method, so it can be used in a more generic way. I think your comment is obsolete in this new version.

utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -464,3 +464,16 @@ fun Throwable.collectMessages(): List<String> {
...
+ * Retrieve query parameters of an [URI]. Multiple values of a single key are supported if they are split by a comma, or
+ * if keys are repeated as defined in RFC6570 section 3.2.9, see https://datatracker.ietf.org/doc/rfc6570.
+ */
+fun URI.getQueryParameters(): Map<String, List<String>> {
+    if (query == null) return emptyMap()
+
+    return query.split('&')
+        .map { it.split('=') }
+        .groupBy({ it.first() }, { it.last().split(",") })
+        .mapValues { it.value.flatten() }
```
sschuberth, line 478: Alternatively:

    return query.split('&')
        .groupBy({ it.substringBefore('=') }, { it.substringAfter('=').split(',') })
        .mapValues { (_, v) -> v.flatten() }

utils/common/src/test/kotlin/ExtensionsTest.kt
```diff
@@ -452,4 +453,29 @@ class ExtensionsTest : WordSpec({
...
+                    mapOf("key" to listOf("value1", "value2", "value3"))
+
+            URI("https://oss-review-toolkit.org?key=value1&key=value2").getQueryParameters() shouldBe
+                    mapOf("key" to listOf("value1", "value2"))
+        }
+
+        "work with empty values" {
+            URI("https://oss-review-toolkit.org?key=").getQueryParameters() shouldBe
+                    mapOf("key" to listOf(""))
+        }
```
sschuberth, line 479: A simple test for an URI without a query string is missing.

sschuberth (changes requested):

downloader/src/main/kotlin/VcsHost.kt
MarcelBochtler: Not sure if this is the nicest way to do this, but I tried.

utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -464,3 +464,15 @@ fun Throwable.collectMessages(): List<String> {
  * Recursively collect the messages of this [Throwable] and all its causes and join them to a single [String].
  */
 fun Throwable.collectMessagesAsString() = collectMessages().joinToString("\nCaused by: ")
+
+/**
+ * Retrieve query parameters of an [URI]. Multiple values of a single key are supported if they are split by a comma, or
```
sschuberth, line 469: Nits:
- Commit message: "Add *a* function"
- Docs: "an URI" -> "this URL" (as it's an extension function)

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -44,6 +46,57 @@ enum class VcsHost(
...
 ) {
+    AZURE_DEVOPS("dev.azure.com", VcsType.GIT) {
+        private val gitCommitPrefix = "GC"
+
+        override fun getUserOrOrgInternal(projectUrl: URI) = projectUrlToUserOrOrgAndProject(projectUrl)?.first
+
+        override fun getProjectInternal(projectUrl: URI) = projectUrl.path.substringAfterLast("/")
+
+        override fun toVcsInfoInternal(projectUrl: URI): VcsInfo {
+            val uri = with(projectUrl) { URI(scheme, authority, path, null, fragment) }
```
sschuberth, line 57: Nice idea to use `with` here!

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -44,6 +46,57 @@ enum class VcsHost(
...
+                vcsInfo.revision.isEmpty()
+            }.orEmpty()
+            return "${vcsInfo.url}?$lineQueryParam$pathQueryParam$revisionQueryParam"
+        }
+
+        override fun toArchiveDownloadUrlInternal(userOrOrg: String, project: String, vcsInfo: VcsInfo): String {
+            val pathIterator = Paths.get(URI(vcsInfo.url).path).iterator()
+            val team = if (pathIterator.hasNext() && pathIterator.next().toString() == userOrOrg) {
+                if (pathIterator.hasNext()) pathIterator.next() else ""
+            } else ""
```
sschuberth, line 80: Going a bit wild alternatively 😉 

            val team = pathIterator.nextOrNull().takeIf { it.toString() == userOrOrg }?.let {
                pathIterator.nextOrNull()?.toString()
            }.orEmpty()

Similar below.

sschuberth (changes requested):

omarnabss (approved):

PR #5306, opened 2022-05-02
StaticHtmlReporter: Use unsafe rendering for issue and violation messages 

bennati:

### DIFF
--- a/reporter/src/main/kotlin/reporters/StaticHtmlReporter.kt
+++ b/reporter/src/main/kotlin/reporters/StaticHtmlReporter.kt
@@ -305,7 +305,7 @@ class StaticHtmlReporter : Reporter {
                 }
             }
             td {
-                p { +ruleViolation.violation.message }
+                p { unsafe { +ruleViolation.violation.message } }
                 if (ruleViolation.isResolved) {
                     p { +ruleViolation.resolutionDescription }
                 } else {
@@ -357,7 +357,7 @@ class StaticHtmlReporter : Reporter {
                     issues.forEach { issue ->
                         li {
                             issueDescription(issue)
-                            p { +issue.resolutionDescription }
+                            p("ort-message") { +issue.resolutionDescription }
                         }
 
                         if (!issue.isResolved && issue.howToFix.isNotBlank()) {
--- a/reporter/src/main/resources/static-html-reporter.css
+++ b/reporter/src/main/resources/static-html-reporter.css
@@ -193,6 +193,10 @@ table.ort-excluded tr.ort-excluded td li.ort-excluded {
     overflow: scroll;
 }
 
+.ort-message {
+    white-space: pre-wrap
+}
+
 @media all and (max-width: 1000px) {
     .ort-report-table th:nth-child(2), .ort-report-table td:nth-child(2) {
         display: none;


reporter/src/main/kotlin/reporters/StaticHtmlReporter.kt
```diff
@@ -305,7 +305,7 @@ class StaticHtmlReporter : Reporter {
                 }
             }
             td {
-                p { +ruleViolation.violation.message }
+                markdown(ruleViolation.violation.message)
```
tsteenbe, line 308: @bennati Per https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/kotlin/RuleViolation.kt#L53 message is a plain text string and not a MarkDown string as RuleViolation.HowToFix. 

If I remember it correctly we made message plain text so 1) can be easily ingested by tools and 2) printing to logs but @mnonnenmacher should weigh as he has worked the most on the Evaluator.

tsteenbe (changes requested):

reporter/src/main/kotlin/reporters/StaticHtmlReporter.kt
mnonnenmacher: I am a bit undecided, it would be nice to be able to format the messages, but it also has benefits to keep them simple strings. In any way, if this is changed, it should also be changed for the web app reporter for consistency. I think this would be a good topic to discuss in the dev meeting to get more opinions.

bennati: I think you misunderstand my request: I am not interested in enabling markdown in the violation message, e.g. displaying links and such. I want to fix an issue I am having with normal text not being displayed correctly.

For example the following HTML source text
```
line 1

line 2
```

is currently displayed as follows: `line 1 line 2`.

I would expect newlines not to be ignored.

sschuberth: > In any way, if this is changed, it should also be changed for the web app reporter for consistency.

Agreed. And shouldn't we support Markdown then also for issues?

bennati: I've updated the commit as I have found that marking the message as 'unsafe' allows to add new lines without the need to enable markdown.
We can still discuss whether to enable markdown or not, but that should be independent from this PR, I think.

mnonnenmacher: > > In any way, if this is changed, it should also be changed for the web app reporter for consistency.
> 
> Agreed. And shouldn't we support Markdown then also for issues?

@bennati As you are on it, could you do the same change for issue messages as part of this PR?

tsteenbe (commented):
Commit message English can be improved: This allows to add html tags such as <br/> to display a message on many line => Allow usage of HTML tags such as <br/> to break a message into multiple lines.

reporter/src/main/kotlin/reporters/StaticHtmlReporter.kt
```diff
@@ -305,7 +305,7 @@ class StaticHtmlReporter : Reporter {
                 }
             }
             td {
-                p { +ruleViolation.violation.message }
+                p { unsafe { +ruleViolation.violation.message } }
```
mnonnenmacher, line 308: After a discussion with @sschuberth I think using `unsafe` here is not a good idea, as it could easily break the HTML if special characters like `<` are used in the message. A better option would be to add a CSS class for `p` with [`white-space: pre-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#values).

mnonnenmacher (changes requested):

reporter/src/main/kotlin/reporters/StaticHtmlReporter.kt
bennati: Where is `p` defined?

sschuberth: Just `Ctrl`+Click on it, it's built into *kotlinx-html*.

bennati: I tried but it does not work on my machine, would you mind pasting here the link to the file in the public `kotlinx-html` repo? Thank you!

sschuberth: Here you go: https://github.com/Kotlin/kotlinx.html/blob/c2f8ee0df425e5188681ac821034c8781ee2a146/src/commonMain/kotlin/generated/gen-tag-groups.kt#L125

bennati: I implemented what (I think) you requested, but I do not see that CSS class being used anywhere in the HTML sources

bennati: any ideas how to fix this?

bennati: I implemented all suggested changes but the solution does not work, can someone please suggest a way forward? Thank you!

sschuberth: > I do not see that CSS class being used anywhere in the HTML sources

I've just tested this and can confirm the CSS to *be* embedded the generated HTML. So, do you want to move this forward by also changing this line to use `p("ort-message")`, @bennati?

bennati: Using `p("ort-message")` visualizes big indentations every time a multi-line string has a new line or any time two strings are joined with +. I tried with `trimMargins()` and `replaceIndentByMargin()` but it did not help.
I think I won't spend any more time on this PR, so we can just close this.

PR #5308, opened 2022-05-02
Improve maven checksum handling

mnonnenmacher:

### DIFF
--- a/model/src/main/kotlin/Hash.kt
+++ b/model/src/main/kotlin/Hash.kt
@@ -97,6 +97,11 @@ data class Hash(
 
         return algorithm.calculate(file).equals(value, ignoreCase = true)
     }
+
+    /**
+     * Verify that the provided [hash] matches this hash.
+     */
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/Hash.kt
```diff
@@ -97,6 +97,11 @@ data class Hash(
 
         return algorithm.calculate(file).equals(value, ignoreCase = true)
     }
+
+    /**
+     * Verify that the provided [hash] matches this hash.
+     */
+    fun verify(hash: Hash): Boolean = algorithm == hash.algorithm && value.equals(hash.value, ignoreCase = true)
```
sschuberth, line 104: Nit: Maybe rename the `hash` parameter to `other`?

sschuberth (approved):

model/src/main/kotlin/Hash.kt
mnonnenmacher: I thought about it, but it is also nice if the options in code completion show up as `file: File`, `hash: Hash`.

PR #5314, opened 2022-05-04
OkHttpClientHelper: Improve exception handling in downloadText()

oheger-bosch:
Make sure that exceptions thrown when accessing the response body are
handled by replacing map with mapCatching.

### DIFF
--- a/utils/core/src/main/kotlin/OkHttpClientHelper.kt
+++ b/utils/core/src/main/kotlin/OkHttpClientHelper.kt
@@ -219,7 +219,7 @@ fun OkHttpClient.downloadFile(url: String, directory: File): Result<File> =
  * [Result] wrapping an [IOException] (which might be a [HttpDownloadError]) on failure.
  */
 fun OkHttpClient.downloadText(url: String): Result<String> =
-    download(url).map { (_, body) ->
+    download(url).mapCatching { (_, body) ->
         body.use { it.string() }
     }
 
--- a/utils/core/src/test/kotlin/OkHttpClientHelperTest.kt
+++ b/utils/core/src/test/kotlin/OkHttpClientHelperTest.kt
@@ -23,6 +23,7 @@ import io.kotest.core.spec.style.StringSpec
 import io.kotest.engine.spec.tempdir
 import io.kotest.matchers.result.shouldBeFailure
 import io.kotest.matchers.should
+import io.kotest.matchers.shouldBe
 import io.kotest.matchers.types.beInstanceOf
 import io.kotest.matchers.types.shouldBeSameInstanceAs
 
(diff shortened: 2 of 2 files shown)


MarcelBochtler (dismissed):

utils/core/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -219,7 +219,7 @@ fun OkHttpClient.downloadFile(url: String, directory: File): Result<File> =
  * [Result] wrapping an [IOException] (which might be a [HttpDownloadError]) on failure.
  */
 fun OkHttpClient.downloadText(url: String): Result<String> =
-    download(url).map { (_, body) ->
+    download(url).mapCatching { (_, body) ->
```
sschuberth, line 222: Can you elaborate a bit more (in the commit message) why this is necessary? What exception could occur when accessing the response body, though the download succeeded? Because `map` is only called on success of the download.

oheger-bosch: Added some more details to the commit message.

sschuberth: Thanks. However this:

> At the time the response is available, the download of the body is not necessarily complete

Sounds to me as if we should rather block on the download of the body to be complete, or? Is there such a functionality in OkHttp?

sschuberth (approved):
Approving despite my comments as the code certainly becomes more robust this way.

utils/core/src/main/kotlin/OkHttpClientHelper.kt
oheger-bosch: > Sounds to me as if we should rather block on the download of the body to be complete, or? Is there such a functionality in OkHttp?

Well, this is exactly what the `ResponseBody.string()` method does (which is called in the `map` block). The actual problem was that the original code did not take into account that this function can throw an `IOException` [1].

[1] https://square.github.io/okhttp/3.x/okhttp/okhttp3/ResponseBody.html#string--

sschuberth: So, on the one hand you're saying that the `IOException` can only occur if the response body is not complete, on the other hand you're saying that the response body will always be complete as `string()` blocks. What am I missing here?

oheger-bosch: Probably a misunderstanding. An `IOException` is not thrown because the response body is not complete, but while loading the body (triggered by the `string()` function), an `IOException` may occur.

Your assumption that the (full) download is successful when `map` is called is not correct. Rather, at that time, the client may have only downloaded the part of the response with the status code. When the body is actually accessed, more data is loaded, and this can cause exceptions.

sschuberth: > Your assumption that the (full) download is successful when `map` is called is not correct.

Thanks for the explanation. The word "full" was crucial to me here: The download of the response (header) is complete and succeeded, but the body might not have been fully downloaded before `map` is called.

Issue #4364, opened 2021-08-16
Clarification / Best Practice required on how to handle multiple analyzer-result.yml files


nikpete:
All,

We do have several use cases where an overarching project might consists of multiple services and product teams. Therefore, multiple analyzer-result.yml file will be submitted (let's say 10+) because the project is simply not able to generate ONE .yml file (i.e. technical restrictions, mono repo)

1. I would like to understand if there is an automatic way (best practice) how we can merge (in ORT) multiple yml. files into ONE?
2. The merge of multiple yml. files should also remove possibe duplicates (FOSS-componentes that are reused in all the repositories and are part of the different .yml files) because currently we are doing it manually which is to be honest a "pain in the ass".

Thanks,
Nik

sschuberth:
Sounds like a good fit a new `helper-cli` subcommand, or @fviernau?

sschuberth:
@porsche-rishisaxena, FYI, I can only assign @nikpete as he's the reporter (and none of you are project members, yet).

porsche-rbieniek:
Porsche solution submitted as https://github.com/oss-review-toolkit/ort/pull/5315

sschuberth:
> Porsche solution submitted as #5315

Please associate issues with PRs by using one of the respective [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) in one of the commits in the PR instead of manually adding comments.

mnonnenmacher:
For reference, if all code is located in Git repositories, a possible workaround for this issue is to create a new Git repository with all the other Git repositories as submodules. You can then use the new Git repository as input for the analyzer to get a single analyzer result with correct provenance information for all projects.

sschuberth:
> I would like to understand if there is an automatic way (best practice) how we can merge (in ORT) multiple yml. files into ONE?

I have a hunch that we're jumping to conclusions too quickly with implying that *merging* (analyzer) result files is the right approach.

Let's step back a bit and ask ourselves what we want to achieve. From re-reading @nikpete's OP my guess is that in the end only a single / combined *report* (i.e. attribution document) is needed. As attribution documents contain far less information than full-blown ORT result files, maybe the better approach is to not merge ORT result files beforehand at all, but to create a special reporter  that can take multiple ORT result files and simply attributes to the union of all contained projects / packages in a single report?

[...]

PR #5317, opened 2022-05-04
Helper CLI function for merging muliple analyzer results into one

porsche-rbieniek:
This helper CLI function is used by Porsche to solve the issue https://github.com/oss-review-toolkit/ort/issues/4364

The rationale behind this is that some projects at Porsche deliver individual analyzer-results for each subproject in a large monorepo. The FOSS analysts need to see a condensed form of the individual dependency graphs across the project monorep. We solve this issue by merging all individual analyzer results into one.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
@@ -1,0 +1,291 @@
+/*
+ * Copyright (C) 2021 Porsche AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.options.split
+import com.github.ajalt.clikt.parameters.types.file
+
+import java.io.File
+import java.time.Clock
+import java.time.Instant
+import java.time.ZonedDateTime
+import java.util.LinkedList
+import java.util.SortedMap
+import java.util.SortedSet
+import java.util.TreeMap
+import java.util.TreeSet
+
+import org.ossreviewtoolkit.model.AnalyzerResult
+import org.ossreviewtoolkit.model.AnalyzerRun
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.OrtIssue
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.PackageReference
+import org.ossreviewtoolkit.model.Project
+import org.ossreviewtoolkit.model.Repository
+import org.ossreviewtoolkit.model.Scope
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.Curations
+import org.ossreviewtoolkit.model.config.Excludes
+import org.ossreviewtoolkit.model.config.LicenseChoices
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.model.config.Resolutions
+import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.common.expandTilde
+import org.ossreviewtoolkit.utils.core.Environment
+
+class MergeAnalyzerResultsCommand : CliktCommand(
+    help = "Read multiple analyzer result files and merge them into one combined analyzer result file."
+) {
+    companion object {
+        private val utcClock = Clock.systemUTC()
+        private fun now(): Instant = ZonedDateTime.now(utcClock).toInstant()
+    }
+
+    private val inputAnalyzerResultFiles by option(
+        "--input-analyzer-result-files", "-i",
+        help = "A comma separated list of analyzer result files to be merged."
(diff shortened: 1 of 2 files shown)


helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
```diff
@@ -0,0 +1,291 @@
...
+            startTime = now(),
+            endTime = now(),
+            environment = Environment(),
+            config = aggregateAnalyzerConfiguration(inputOrtResults),
+            result = aggregrateAnalyzerRun(inputOrtResults)
+        )
+
+        val repository = Repository(
+            vcs = analyzerRun.result.projects.first().vcs,
+            vcsProcessed = analyzerRun.result.projects.first().vcsProcessed,
```
mnonnenmacher, line 102: ORT requires in several places that it can find the relative path of a project to the analyzer root repository. This requirement cannot be fulfilled when merging multiple analyzer result files from independent repositories, because the model was never designed to support this. Taking the VCS of the arbitrary first project here will cause issues with the handling of license findings and repository configurations in later steps, this should at least be mentioned in the documentation.

helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
```diff
@@ -0,0 +1,291 @@
...
+        val outputFile = OrtResult(
+            repository = repository,
+            analyzer = analyzerRun
+        )
+
+        outputAnalyzerResultFile.writeValue(outputFile)
+    }
+
+    private fun aggregateRepositoryConfigurations(repositories: List<Repository>): RepositoryConfiguration {
+        fun mergeExcludes(leftExlcudes: Excludes, rightExcludes: Excludes): Excludes = Excludes(
```
mnonnenmacher, line 115: Typo in `leftExlcudes`.

helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
```diff
@@ -0,0 +1,291 @@
...
+            paths = (leftExlcudes.paths + rightExcludes.paths).distinct(),
+            scopes = (leftExlcudes.scopes + rightExcludes.scopes).distinct()
```
mnonnenmacher, line 117: Merging excludes can lead to unexpected results, if excludes from on analyzer result accidentally match content from another analyzer result.

helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
```diff
@@ -0,0 +1,291 @@
...
+        fun mergeLicenseChoices(leftChoices: LicenseChoices, rightChoices: LicenseChoices): LicenseChoices =
+            LicenseChoices(
+                repositoryLicenseChoices = (
+                        leftChoices.repositoryLicenseChoices
+                                + rightChoices.repositoryLicenseChoices
+                        ).distinct(),
+                packageLicenseChoices = (
+                        leftChoices.packageLicenseChoices
+                                + rightChoices.packageLicenseChoices
+                        ).distinct()
```
mnonnenmacher, line 141: Merging license choices can lead to unexpected results, if they are contradictory, or if license choices from one analyzer result apply to findings from another analyzer result.
There are more similar consistency issues in the code below where data is lost because the model was not designed for this use case.

helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
```diff
@@ -0,0 +1,291 @@
...
+                    excludes = mergeExcludes(leftConfig.excludes, rightConfig.excludes),
+                    resolutions = mergeResolutions(leftConfig.resolutions, rightConfig.resolutions),
+                    curations = mergeCurations(leftConfig.curations, rightConfig.curations),
+                    licenseChoices = mergeLicenseChoices(leftConfig.licenseChoices, rightConfig.licenseChoices)
+                )
+            }
+            .orElse(RepositoryConfiguration())
+    }
+
+    private fun aggregateAnalyzerConfiguration(inputOrtResults: List<AnalyzerRun>): AnalyzerConfiguration {
```
mnonnenmacher, line 157: Consistently use either plural or singular for the aggregate function names, but don't mix them.

helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    private fun <T> mergeSortedSets(left: SortedSet<T>, right: SortedSet<T>): SortedSet<T> {
+        val set = TreeSet<T>()
+
+        set.addAll(left)
+        set.addAll(right)
+
+        return set
+    }
```
mnonnenmacher, line 197: This could be shorter:
```kotlin
sortedSetOf().apply {
  addAll(left)
  addAll(right)
}
```

mnonnenmacher (changes requested):
Commit message: Please stick to the line length limit.

If have further explained the reason behind @fviernau's concerns from https://github.com/oss-review-toolkit/ort/pull/4698#issuecomment-971505128 in the code review. Based on the rationale from the commit message and the discussion in #4698 I think merging analyzer results is not the correct approach, because it is technically not possible to do it correctly. If all the mentioned limitations of this approach are no issue for your use case and all you need is a list of projects and packages, the better approach would be to let this helper-cli command output a custom format that fits your needs. Another alternative would be to implement this as a reporter, because these kind of reports are exactly what the reporter module is there for.

helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
porsche-rbieniek: This limitation is well-known and therefore the merge code should assert that all merged analyzer results orignate from the same 'root' VCS.
I just realized that I have not implemented that in the first pull request

The helper command was by the need to handle the case of a multiple product teams working on individual project directories in a shared repository with each team delivering its own analyzer result.
From a logical perspective, it would not make any sense to merge analyzer results from arbitary sources.

helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
porsche-rbieniek: Point taken. The excludes will not be merged then

helper-cli/src/main/kotlin/commands/MergeAnalyzerResultsCommand.kt
porsche-rbieniek: This might be true but we chose to leave that problem to the analyst to sort it out.

Anyway, it is bad sign if project team choose contradicting license statements

porsche-rbieniek:
This pull requests supersedes https://github.com/oss-review-toolkit/ort/pull/4698 
The earlier PR had major drawbacks with the most problematic being that it completely flattened the  dependency graph.

This new version is a rewrite which tries to keep the dependency trees intact among other things. Consequently, I closed https://github.com/oss-review-toolkit/ort/pull/4698

sschuberth:
> This pull requests supersedes #4698

I believe this PR in turn is superseded by https://github.com/oss-review-toolkit/ort/pull/5519 now.

PR #5319, opened 2022-05-04
Pip: Add a few sanity checks when parsing license information

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pip.kt
+++ b/analyzer/src/main/kotlin/managers/Pip.kt
@@ -419,13 +419,12 @@ class Pip(
         return declaredLicenses
     }
 
-    private fun getLicenseFromLicenseField(value: String?): String? =
-        value?.let {
-            // Work-around for projects that declare licenses in classifier-style syntax.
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -420,7 +422,10 @@ class Pip(
     }
 
     private fun getLicenseFromLicenseField(value: String?): String? {
-        if (value.isNullOrBlank() || value == "UNKNOWN" || value.lines().size > 1) return null
+        if (value.isNullOrBlank() || value == "UNKNOWN") return null
+
+        val isShortString = value.length <= SHORT_STRING_MAX_CHARS && value.lines().size == 1
```
mnonnenmacher, line 427: It would nice to have the documentation link from the commit message also here as a code comment, as reference why this check is done.

mnonnenmacher (changes requested):

djrising1234:
Thanks for the help. Still learning. And will get my shi together

PR #5320, opened 2022-05-04
Additional reports for Volkswagen group

porsche-rbieniek:
This pull request implements additional ORT report implementations for types of reports being used within the Volkswagen group.

All 3rd party vendors are required to submit FOSS compliance reports in a specfic format. The rationale for sharing these report implementation is to ease the report delivery process.

On a technical level, there are to different types of report in use:

- An Excel-based reporting format which is currently in use as the main report format
- A JSON-based report which will eventually become the standard FOSS compliance reprt format (within the Volkswagen group)
- 


sschuberth:
Superseded by https://github.com/oss-review-toolkit/ort/pull/5518.

PR #5322, opened 2022-05-04
Add a FossId reporter

nnobelis:
This Reporter asks FossID to generate some reports and returns them.


reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
```diff
@@ -0,0 +1,104 @@
...
+            readTimeout(Duration.ofSeconds(60))
+        }
+        return FossIdRestService.create(serverUrl, client).let { FossIdServiceWithVersion.instance(it) }.also {
+            if (it.version.isEmpty()) {
+                log.warn { "The FossID server is running an unknown version." }
+            } else {
+                log.info { "The FossID server is running version ${it.version}." }
+            }
+        }
+    }
```
sschuberth, line 78: This code is only slightly different from the code in `scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt`. Can we align / extract it? E.g. by creating a client with increased timeout already in

https://github.com/oss-review-toolkit/ort/blob/c8c18090f49018e55b2095fd1be202eac3fe95bf/clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt#L125

if no other client is given, and also doing the logging already over there?

reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
```diff
@@ -0,0 +1,104 @@
...
+        val service = createService(serverUrl)
+
+        return input.ortResult.scanner?.results?.scanResults?.values?.flatten()?.mapNotNull { result ->
+            runBlocking {
+                result.additionalData[SCAN_CODE_KEY]?.let { scanCode ->
+                    this@FossIdReporter.log.info { "Generating report for scan $scanCode." }
+                    service.generateReport(user, apiKey, scanCode, reportType, selectionType, outputDir).getOrNull()
+                }
+            }
+        } ?: emptyList()
```
sschuberth, line 102: Use `orEmpty()`.

reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
```diff
@@ -0,0 +1,104 @@
...
+
+        log.info { "The FossID server URL is $serverUrl." }
+
+        val service = createService(serverUrl)
+
+        return input.ortResult.scanner?.results?.scanResults?.values?.flatten()?.mapNotNull { result ->
+            runBlocking {
+                result.additionalData[SCAN_CODE_KEY]?.let { scanCode ->
+                    this@FossIdReporter.log.info { "Generating report for scan $scanCode." }
+                    service.generateReport(user, apiKey, scanCode, reportType, selectionType, outputDir).getOrNull()
```
sschuberth, line 99: Can we actually generate these reports in parallel and await them?

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -117,17 +121,33 @@ interface FossIdRestService {
...
+
             val retrofit = Retrofit.Builder()
-                .apply { if (client != null) client(client) }
+                .apply {
+                    client(
+                        client ?: OkHttpClientHelper.buildClient {
+                            readTimeout(Duration.ofSeconds(60))
+                        }
+                    )
+                }
```
sschuberth, line 137: Ok, now see the problem why you're using `OkHttpClientHelper.buildClient`: When used from ORT, we'd like to get all the custom settings done in `OkHttpClientHelper`, but we also want to get the adjusted timeout. I'd propose to solve the issue like this:

            val minReadTimeout = Duration.ofSeconds(60)
            val retrofit = Retrofit.Builder()
                .apply {
                    client(
                        client?.run {
                            takeUnless { readTimeoutMillis < minReadTimeout.toMillis() }
                                ?: newBuilder().readTimeout(minReadTimeout).build()
                        } ?: OkHttpClient.Builder().readTimeout(minReadTimeout).build()
                    )
                }
                .baseUrl(url)
                .addConverterFactory(JacksonConverterFactory.create(JSON_MAPPER))
                .build()

sschuberth (changes requested):

clients/fossid-webapp/build.gradle.kts
```diff
@@ -30,6 +30,8 @@ plugins {
 dependencies {
     api("com.squareup.retrofit2:retrofit:$retrofitVersion")
 
+    implementation(project(":utils:core-utils"))
```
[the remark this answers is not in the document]
nnobelis: > But I'm not sure yet what to do about the logging...

Shouldn't the logging be part of `common-utils` ?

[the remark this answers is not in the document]
nnobelis: Or use directly in the client log4j.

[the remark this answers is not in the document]
sschuberth: > Or use directly in the client log4j.

Yes, I tend to agree. However, I was wondering whether we should now introduce Log4j here, while there is [an effort going on to get rid of Log4j](https://github.com/oss-review-toolkit/ort/pull/5323). But probably we can still do it, and then get rid of Log4j in a single blow everywhere, instead of creating fragmentation now.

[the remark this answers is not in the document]
nnobelis: @sschuberth the PR has been updated.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -58,6 +62,9 @@ import retrofit2.http.POST
 
 interface FossIdRestService {
     companion object {
+        @JvmStatic
+        val logger = loggerOf(FossIdRestService::class.java)
```
sschuberth, line 66: Nit: Would have been nice to also use just `log` here as in ORT to ease code search for logging statements. But provided that it's unsure how long we'll keep it, it's also ok for me to stick to this.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -103,6 +105,18 @@ class FossIdTest : WordSpec({
...
+        // See https://lifesaver.codes/answer/cannot-mockkstatic-for-kotlin-companion-object-static-method-136
+        mockkObject(FossIdRestService)
+
+        every { FossIdRestService.createService(any()) } returns createServiceMock()
+    }
+
+    afterTest {
+        unmockkObject(FossIdRestService)
+    }
+
```
sschuberth, line 119: Just wondering: How are the new test assets being used, as I don't see any (relevant) changes to the test code?

reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    override val reporterName = "FossId"
+
+    override fun generateReport(input: ReporterInput, outputDir: File, options: Map<String, String>): List<File> {
+        val serverUrl = options[SERVER_URL_PROPERTY]
+            ?: throw IllegalArgumentException("No FossID server URL configuration found.")
+        val apiKey = options[API_KEY_PROPERTY]
+            ?: throw IllegalArgumentException("No FossID API Key configuration found.")
+        val user = options[USER_PROPERTY]
+            ?: throw IllegalArgumentException("No FossID User configuration found.")
```
sschuberth, line 72: These should be replaced with

    val serverUrl = requireNotNull(options[SERVER_URL_PROPERTY]) {
        "No FossID server URL configuration found."
    }

etc.

reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
```diff
@@ -0,0 +1,90 @@
...
+        val reportType = options[REPORT_TYPE_PROPERTY]?.let { ReportType.valueOf(it) } ?: ReportType.HTML_DYNAMIC
+        val selectionType =
+            options[SELECTION_TYPE_PROPERTY]?.let { SelectionType.valueOf(it) } ?: SelectionType.INCLUDE_ALL_LICENSES
```
sschuberth, line 75: These would crash if the options were non-null but invalid values. So maybe better:

    val reportType = runCatching { ReportType.valueOf(options[REPORT_TYPE_PROPERTY]) }
        .getOrDefault(ReportType.HTML_DYNAMIC)

reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        val service = FossIdRestService.createService(serverUrl)
+
+        return runBlocking {
```
sschuberth, line 79: This should probably use `runBlocking(Dispatchers.IO)`.

reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
```diff
@@ -0,0 +1,90 @@
...
+            input.ortResult.scanner?.results?.scanResults?.values?.flatten()?.mapNotNull { result ->
+                result.additionalData[SCAN_CODE_KEY]?.let { scanCode ->
+                    async {
+                        this@FossIdReporter.log.info { "Generating report for scan $scanCode." }
+                        service.generateReport(user, apiKey, scanCode, reportType, selectionType, outputDir).getOrNull()
```
sschuberth, line 84: This silently swallows failures in report generation. Shouldn't we at least log them?

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
nnobelis: I thought about exactly the opposite :)
I would like the people to see that this code is special and does not use standard ORT logger, hence the different name.

The day we migrate from log4j, we will start by looking at all dependencies in `build.gradle.kts` files anyway...

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
nnobelis: This is because we stopped using `FossIdRestService` and use `FossIdServiceWithVersion` instead.
The latter does a `getFossIdVersion` called that should be mocked with an asset:
https://github.com/bosch-io/oss-review-toolkit/blob/445df9a1354b678a64c5e5956c42efb5ace5bb17/clients/fossid-webapp/src/main/kotlin/FossIdServiceWithVersion.kt#L33

sschuberth: Thanks for the explanation!

reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
nnobelis: hmmm this code is not working. It should be:
```
val reportType = runCatching {           
           ReportType.valueOf(options[REPORT_TYPE_PROPERTY].orEmpty())
}.getOrDefault(ReportType.HTML_DYNAMIC)
```
however, although you have a point, I don't like this solution: if the parameter is null, it will nevertheless be parsed with `valueOf` throw an exception, then return the default value.
```
val reportType =  options[REPORT_TYPE_PROPERTY]?.let { 
            kotlin.runCatching { 
                ReportType.valueOf(it)}.getOrNull()
        } ?: ReportType.HTML_DYNAMIC
```

sschuberth: > if the parameter is null, it will nevertheless be parsed with `valueOf` throw an exception, then return the default value.

But isn't this exactly the behavior that we want?

> hmmm this code is not working.

That's odd, as `valueOf()` is documented to throw an NPE if the argument is *null*, so the wrapping `runCatching` *should* catch it...

nnobelis: > But isn't this exactly the behavior that we want?

You didn't understand what I meant. Yes, this is what we want but trying to parse a _null_ value is just wasting unnecessary CPU cycle. My code skips the call to `valueOf` if the option is `null`.

> > hmmm this code is not working.
> 
> That's odd, as `valueOf()` is documented to throw an NPE if the argument is _null_, so the wrapping `runCatching` _should_ catch it...

I could not get an online confirmation but it seems `valueOf` of Kotlin is NOT the `valueOf` of java. Therefore the parameter is of type `String` not `String?`.

sschuberth: > You didn't understand what I meant.

Indeed 😁 Because I thought by "hmmm this code is not working" you mean that the code is not behaving as expected at runtime. But now I understand you mean that it does not compile.

reporter/src/main/kotlin/reporters/fossid/FossIdReporter.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        return runBlocking(Dispatchers.IO) {
+            input.ortResult.scanner?.results?.scanResults?.values?.flatten()?.mapNotNull { result ->
+                result.additionalData[SCAN_CODE_KEY]?.let { scanCode ->
+                    async {
+                        this@FossIdReporter.log.info { "Generating report for scan $scanCode." }
+                        service.generateReport(user, apiKey, scanCode, reportType, selectionType, outputDir)
+                            .onFailure {
+                                it.showStackTrace()
+                                this@FossIdReporter.log.info { "Error during report generation: ${it.message}." }
```
sschuberth, line 100: Please use `it.collectMessages()` here.

reporter/src/test/kotlin/reporters/fossid/FossIdReporterTest.kt
```diff
@@ -0,0 +1,294 @@
...
+    beforeSpec {
+        mockkStatic("org.ossreviewtoolkit.clients.fossid.ExtensionsKt")
+    }
+
+    afterTest {
+        clearAllMocks()
+    }
+
+    "generateReport of FossIdReport " should {
+        "checks if a server URL was given" {
```
sschuberth, line 89: "check"

reporter/src/test/kotlin/reporters/fossid/FossIdReporterTest.kt
```diff
@@ -0,0 +1,294 @@
...
+                reporter.generateReport(
+                    input,
+                    DIRECTORY_SAMPLE,
+                    DEFAULT_OPTIONS.filterNot { it.key == FossIdReporter.SERVER_URL_PROPERTY }
+                )
+            }
+            exception shouldHaveMessage "No FossID server URL configuration found."
+        }
+
+        "checks if an API key was given" {
```
sschuberth, line 102: "check"

reporter/src/test/kotlin/reporters/fossid/FossIdReporterTest.kt
```diff
@@ -0,0 +1,294 @@
...
+                reporter.generateReport(
+                    input,
+                    DIRECTORY_SAMPLE,
+                    DEFAULT_OPTIONS.filterNot { it.key == FossIdReporter.API_KEY_PROPERTY }
+                )
+            }
+            exception shouldHaveMessage "No FossID API Key configuration found."
+        }
+
+        "checks if a user was given" {
```
sschuberth, line 115: "check"

sschuberth (changes requested):

sschuberth (approved):

PR #5323, opened 2022-05-04
Replace log4j2 with logback

porsche-rbieniek:
This pull request replaces the Apache Log4J logging framework with Logback framework.

The Log4J framework had some major security issue revealed over the course of the last few months. Although these issues have been fixed, the log4j framework has gained some attention and the use of it may even be prohibited in security-sensitive environments.
 
From a technical standpoint, only the use of the log4j API remains and the log output is funneled through SLF4J to the logback logging backend. Same applies for other java logging approaches like java commons logging and java util logging


sschuberth:
This PR does not touch e.g. `utils/core/src/main/kotlin/Logger.kt` at all it seems, so I wonder how it could ever work.

Anyway, while I stick to my statement that I'm *basically* fine with switching from Log4j to something else, I doubt that "bad publicity" is a good rationale to do so. IMO the Log4j team has done a good job at fixing the infamous issues under high pressure.

That said, a good rationale could be that we should exchange the JVM-only Log4j library with a Kotlin multi-platform (preferably incl. native) framework. Candidates could be

* https://github.com/MicroUtils/kotlin-logging
* https://github.com/AAkira/Napier
* https://github.com/Kodein-Framework/Kodein-Log

porsche-rbieniek:
I admit that the log4j-Team has spent an enormous amount of effort to fix the poblems but the damage is done. 

The use of log4j has become critical in some areas. And I'm part in projects from another company where it has become a corporate policy to banish log4j due to the number of CVEs which have been popping up.

This PR does not need to touch  utils/core/src/main/kotlin/Logger.kt to work. Looking into the file reveals that it is using only the stuff from Log4J2-API. 

The logback logging implementation provides an implementation for the SLF4J-API, as well as backend implementations for other logging frameworks as well like java.util.logging

In the PR, these dependencies are pulled in:

- implementation("ch.qos.logback:logback-classic:$logbackClassicVersion")
-  implementation("org.slf4j:slf4j-api:$slf4jVersion")
-  implementation("org.slf4j:jcl-over-slf4j:$slf4jVersion")
-  implementation("org.slf4j:jul-to-slf4j:$slf4jVersion")

The logback-classic dependency provides the appenders and other mechanics to make logging work. The other three dependencies are simply facades to make APIs from other java logging frameworks work:

- log4j --> org.slf4j:slf4j-api
- apache commons logging --> org.slf4j:jcl-over-slf4j
- java.util.logging --> org.slf4j:jul-to-slf4j

I could have migrated the Logger.kt to natie logback but I didn't bother to, simply because you need to have log4j available anyway because so many projects use at least the API.

Another reason to use logback and not something else is that it is a well-known and battle-tested framework, given that it is basically used in every spring boot-based microservice.
Given the capabilty to extend ORT to use the logback.xml configuration file with its flexibility in terms of logging formats and destinations makes it a quite strong pick when it comes to have ORT running in a server environment. Which is a long-term goal anyway.

sschuberth:
> That said, a good rationale could be that we should exchange the JVM-only Log4j library with a Kotlin multi-platform (preferably incl. native) framework.

However, I acknowledge the downside of this is that logging done in dependencies of ORT would stop to work.

sschuberth:
As a side note, maybe using the [Java-built-in `System.Logger`](https://blog.frankel.ch/system-logger/) could be used as the API, and as the implementation also `System.Logger` or the proposed Logback.

sschuberth:
Discussion points from ORT developer meeting:
* Kotlin multi-platform logging might not work with Java dependencies
* We should get rid of Log4j completely, not just the implementation, but also the API
* In order to reduce dependencies, it might make sense to prefer `System.Logger` over SLF4J
* We can get rid of our custom `perf` log level, and simply log to `info` instead
* Should look into lazy evaluation of log statements (e.g. don't interpolate strings unless the log level is active)

sschuberth:
> * We can get rid of our custom `perf` log level, and simply log to `info` instead

See https://github.com/oss-review-toolkit/ort/pull/5393.

PR #5326, opened 2022-05-05

mnonnenmacher:

### DIFF
--- a/scanner/src/main/kotlin/experimental/ScanController.kt
+++ b/scanner/src/main/kotlin/experimental/ScanController.kt
@@ -140,7 +140,7 @@ class ScanController(
      * Get the [NestedProvenance] for the provided [id], or null if no nested provenance for the [id] is available.
      */
     fun getNestedProvenance(id: Identifier): NestedProvenance? =
-        nestedProvenances[packageProvenancesWithoutVcsPath.getValue(id)]
+        packageProvenancesWithoutVcsPath[id]?.let { nestedProvenances[it] }
 
     /**
      * Get all [NestedProvenance]s by [Package].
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -140,7 +140,7 @@ class ScanController(
      * Get the [NestedProvenance] for the provided [id], or null if no nested provenance for the [id] is available.
      */
     fun getNestedProvenance(id: Identifier): NestedProvenance? =
-        nestedProvenances[packageProvenancesWithoutVcsPath.getValue(id)]
+        nestedProvenances[packageProvenancesWithoutVcsPath[id]]
```
sschuberth, line 143: Hmm, your intended "Instead return null" seems to be achieved like this: `packageProvenancesWithoutVcsPath[id]` returns null, then that null value is being looked in `nestedProvenances[packageProvenancesWithoutVcsPath`, which fails and finally also returns null. Is this "indirection" intended? And / or should this maybe be written for clarify like

    packageProvenancesWithoutVcsPath[id]?.let { nestedProvenances[it] }

mnonnenmacher: I thought that's just the simplest way to achieve that without adding more code. I'm fine with changing it to the suggestion as well, if that's what you prefer?

sschuberth: I'm fine with the code, but I'd appreciate some explanation in the commit message about this "indirection". Or keep the commit message and change the code 😉 But I'll approve anyway and leave it to you.

sschuberth (approved):

scanner/src/main/kotlin/experimental/ScanController.kt
mnonnenmacher: Changed as requested.

sschuberth (approved):

PR #5327, opened 2022-05-06
FossId: Store the scan code of a scan in the ScanSummary

bs-ondem:
Check commit messages for details.

### DIFF
--- a/model/src/main/kotlin/ScanResult.kt
+++ b/model/src/main/kotlin/ScanResult.kt
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.model
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonInclude
 
 import org.ossreviewtoolkit.model.utils.RootLicenseMatcher
 
@@ -41,7 +42,14 @@ data class ScanResult(
     /**
      * A summary of the scan results.
      */
-    val summary: ScanSummary
+    val summary: ScanSummary,
+
+    /**
+     * A map for scanner specific data that cannot be mapped into any generalized property, but still needs to be
+     * stored in the scan result.
+     */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    val additionalData: Map<String, String> = emptyMap()
 ) {
     /**
      * Filter all detected licenses and copyrights from the [summary] which are underneath [path], and set the [path]
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -110,6 +110,9 @@ class FossId internal constructor(
         @JvmStatic
         private val WAIT_DELAY = 10.seconds
 
+        @JvmStatic
+        internal val SCAN_CODE_KEY = "scancode"
+
         /**
          * The scan states for which a scan can be triggered.
          */
@@ -326,7 +329,7 @@ class FossId internal constructor(
 
                     if (config.waitForResult) {
                         val rawResults = getRawResults(scanCode)
-                        val resultsSummary = createResultSummary(startTime, provenance, rawResults)
+                        val resultsSummary = createResultSummary(startTime, provenance, rawResults, scanCode)
 
                         results.getOrPut(pkg) { mutableListOf() } += resultsSummary
                     } else {
@@ -340,7 +343,7 @@ class FossId internal constructor(
                             startTime, Instant.now(), "", sortedSetOf(), sortedSetOf(), listOf(issue)
                         )
 
-                        val scanResult = ScanResult(provenance, details, summary)
+                        val scanResult = ScanResult(provenance, details, summary, mapOf(SCAN_CODE_KEY to scanCode))
                         results.getOrPut(pkg) { mutableListOf() } += scanResult
                     }
                 } catch (e: IllegalStateException) {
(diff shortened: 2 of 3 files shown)


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -110,6 +110,9 @@ class FossId internal constructor(
         @JvmStatic
         private val WAIT_DELAY = 10.seconds
 
+        @JvmStatic
+        private val SCAN_CODE_KEY = "scancode"
```
nnobelis, line 114: Rather plural `scancodes` ?

model/src/main/kotlin/ScanSummary.kt
```diff
@@ -72,7 +72,14 @@ data class ScanSummary(
...
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
-    val issues: List<OrtIssue> = emptyList()
+    val issues: List<OrtIssue> = emptyList(),
+
+    /**
+     * A map for scanner specific data that cannot be mapped into any property of a scan summary, but still needs to be
+     * stored in it.
+     */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    val additionalData: Map<String, String> = sortedMapOf()
```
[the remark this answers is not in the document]
nnobelis: @bs-ondem If we use the `ScanContext`, then we have to store the scan codes as CSV like I did with the labels, right ?

Why do you think it isn't a good fit ?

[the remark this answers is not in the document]
bs-ondem: > @bs-ondem If we use the `ScanContext`, then we have to store the scan codes as CSV like I did with the labels, right ?

Yes, if we use the `labels` in the `ScanContext` then we have to store the scan codes CSV like or in a comma separated list of scan codes.

> Why do you think it isn't a good fit ?

`ScanContext` is only used in relation with the experimental scanners and can't be used currently with the `--scanner` option of the ORT scan command. In addition, a scan code belongs to a specific scan of package which should be mapped in the `ScanSummary` of the package.

model/src/main/kotlin/ScanResult.kt
```diff
@@ -41,7 +42,14 @@ data class ScanResult(
     /**
      * A summary of the scan results.
      */
-    val summary: ScanSummary
+    val summary: ScanSummary,
+
+    /**
+     * A map for scanner specific data that cannot be mapped into any generalized property, but still needs to be
+     * stored for in the scan result.
```
nnobelis, line 49: - for

model/src/main/kotlin/ScanResult.kt
```diff
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.model
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonInclude
```
sschuberth, line 23: Commit message nits:
- check line wrapping
- quote `additionalData`
- don't only say it's called `additionalData` now, but also why (to emphasize this is about optional data that cannot be captures in the common data model).
- let's say "*will be used" instead of "can be used", so make clear that there is a real need for this *now*

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -337,10 +340,20 @@ class FossId internal constructor(
                             severity = Severity.HINT
                         )
                         val summary = ScanSummary(
-                            startTime, Instant.now(), "", sortedSetOf(), sortedSetOf(), listOf(issue)
+                            startTime,
+                            Instant.now(),
+                            "",
+                            sortedSetOf(),
+                            sortedSetOf(),
+                            listOf(issue)
```
sschuberth, line 348: Nit: Unrelated / unnecessary change.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -337,10 +340,20 @@ class FossId internal constructor(
...
                         )
 
-                        val scanResult = ScanResult(provenance, details, summary)
+                        val scanResult = ScanResult(
+                            provenance,
+                            details,
+                            summary,
+                            mapOf(SCAN_CODE_KEY to scanCode)
+                        )
```
sschuberth, line 356: Nit: This would still fit into a single line, and probably still is readable enough that way.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
sschuberth, line 114: Consider making this `internal` so it can be used in the test below.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -686,6 +688,33 @@ class FossIdTest : WordSpec({
...
+                .expectCheckScanStatus(scanCode, ScanStatus.FINISHED)
+                .expectCreateScan(projectCode, scanCode, vcsInfo)
+                .expectDownload(scanCode)
+                .mockFiles(scanCode, identifiedRange = 1..2, markedRange = 1..2)
+
+            val fossId = createFossId(config)
+
+            val results = fossId.scan(listOf(createPackage(pkgId, vcsInfo))).scanResults(pkgId)
+
+            val expectedAdditionalData = mapOf("scancode" to scanCode)
```
sschuberth, line 712: Use the constant for `"scancode"`, see my comment above.

sschuberth (changes requested):

model/src/main/kotlin/ScanResult.kt
sschuberth: > quote `additionalData`

My fault, with "quote" I meant "add backticks" to get formatting exactly as in my example; but let's keep it as-is now.

sschuberth (approved):

Issue #5329, opened 2022-05-06
NPM analyzer cannot handle URLs pointing to the GitHub npm registry


oheger-bosch:
Such URLs have the form https://npm.pkg.github.com/download/@organization/project/version/sha1-hash. The NPM analyzer processes them incorrectly, resulting in a source artifact URL like https://github.com/download/@organization/archive/.tar.gz. The scanner is then not able to download those packages.

PR #5331, opened 2022-05-06
VcsHost: Add special handling for URLs to the GitHub npm registry

oheger-bosch:
Such URLs are interpreted as VCS URLs of type GITHUB (based on their
host name), but they must not be processed like regular GitHub
repository URLs, as this would yield incorrect results.

This PR partly solves #5329. It makes sure that the URL is not changed in an incorrect way. There is, however, still a problem with the download in the scanner, because the deflation of the archive fails due to the missing file extension.


downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -158,29 +158,33 @@ enum class VcsHost(
             projectUrlToUserOrOrgAndProject(projectUrl)?.second?.removeSuffix(".git")
 
         override fun toVcsInfoInternal(projectUrl: URI) =
-            gitProjectUrlToVcsInfo(projectUrl) { baseUrl, pathIterator ->
-                var revision = ""
-                var path = ""
+            if (isGitHubNpmRegistryUrl(projectUrl.toString()))
```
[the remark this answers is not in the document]
oheger-bosch: Yes, I was also thinking about this; this is probably a cleaner solution. I will give it a try.

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -87,8 +87,12 @@ fun archive(inputDir: File, zipFile: File, prefix: String = ""): File =
 /**
  * Unpack the [File] to [targetDirectory] using [filter] to select only the entries of interest.
  */
-fun File.unpack(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = { true }) =
-    when (ArchiveType.getType(name)) {
+fun File.unpack(
```
sschuberth, line 90: Commit message nits:
- No comma in "To support cases, in which"
- "there is valid file name" -> "there is *a* valid file name"
- "to derive the archive type" -> "to derive the archive type *from*"

Maybe better after all: "To support cases where no archive type can be derived from the file name, allow to pass an explicit archive type fallback."

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -87,8 +87,12 @@ fun archive(inputDir: File, zipFile: File, prefix: String = ""): File =
...
+    targetDirectory: File,
+    fallbackArchiveType: ArchiveType = ArchiveType.NONE,
+    filter: (ArchiveEntry) -> Boolean = { true }
+) =
+    when (ArchiveType.getType(name).takeUnless { it == ArchiveType.NONE } ?: fallbackArchiveType) {
```
sschuberth, line 95: I wonder whether we should really implement this as a fallback, or rather an override. An override would have the advantage that we could unpack wrong named archives. While that rarely happens, it seems like the more generic approach to always use the `ArchiveType` unless it is `NONE`. What do you think?

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -163,7 +167,7 @@ fun File.unpackDeb(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = {
             subDirectory.safeMkdirs()
 
             val file = tempDir.resolve(name)
-            file.unpack(subDirectory, filter)
+            file.unpack(subDirectory, filter = filter)
```
sschuberth, line 170: Unnecessary change.

downloader/src/main/kotlin/Downloader.kt
```diff
@@ -328,7 +329,7 @@ class Downloader(private val config: DownloaderConfiguration) {
                     gemDirectory.safeDeleteRecursively(force = true)
                 }
             } else {
-                sourceArchive.unpack(outputDirectory)
+                sourceArchive.unpack(outputDirectory, fallbackArchiveType = ArchiveType.TAR_GZIP)
```
sschuberth, line 332: I don't feel comfortable with hard-coding this at this central place. I haven't looked at the details of the issue you're trying to address, but maybe we could instead set a correct file extension based on a mime-type in a HTTP response?

sschuberth: I have some more ideas. Can we maybe have a look at this together?

utils/common/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: Reworded and adapted to the override logic.

utils/common/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: Yes, this is better. Switched to override logic.

utils/common/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: This is actually needed because of the new optional parameter in `unpack()`.

downloader/src/main/kotlin/Downloader.kt
oheger-bosch: Sure. I have pushed my current state.

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -85,10 +85,16 @@ fun archive(inputDir: File, zipFile: File, prefix: String = ""): File =
     }
 
 /**
- * Unpack the [File] to [targetDirectory] using [filter] to select only the entries of interest.
+ * Unpack the [File] to [targetDirectory] using [filter] to select only the entries of interest. If aan archive type
```
sschuberth, line 88: Typo: "aan" -> "an"

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -85,10 +85,16 @@ fun archive(inputDir: File, zipFile: File, prefix: String = ""): File =
...
+ * other than [ArchiveType.NONE] is specified in [useArchiveType], use this one; otherwise, try to detect the type
+ * based on the file extension.
  */
-fun File.unpack(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = { true }) =
-    when (ArchiveType.getType(name)) {
+fun File.unpack(
+    targetDirectory: File,
+    useArchiveType: ArchiveType = ArchiveType.NONE,
```
sschuberth, line 94: Nit: Maybe better `forceArchiveType`?

sschuberth (changes requested):

PR #5334, opened 2022-05-07
Prepare ort-config PackageConfigurationProvider

mnonnenmacher:
Several refactorings to prepare the implementation of a `PackageConfigurationProvider` for the ort-config repository, see the commit messages for details.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2021 HERE Europe B.V.
+ * Copyright (C) 2022 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -88,12 +89,11 @@ class GetPackageLicensesCommand : CliktCommand(
         val packageConfigurationProvider = packageConfigurationOption.createProvider()
 
         val result = scanResults.firstOrNull()?.let { scanResult ->
-            val packageConfiguration = packageConfigurationProvider.getPackageConfiguration(
-                packageId, scanResult.provenance
-            )
+            val packageConfigurations =
+                packageConfigurationProvider.getPackageConfigurations(packageId, scanResult.provenance)
 
-            val licenseFindingCurations = packageConfiguration?.licenseFindingCurations.orEmpty()
-            val pathExcludes = packageConfiguration?.pathExcludes.orEmpty()
+            val licenseFindingCurations = packageConfigurations.flatMap { it.licenseFindingCurations }
+            val pathExcludes = packageConfigurations.flatMap { it.pathExcludes }
 
             val nonExcludedLicenseFindings = scanResult.summary.licenseFindings.filter { licenseFinding ->
                 pathExcludes.none { it.matches(licenseFinding.location.path) }
--- a/model/src/main/kotlin/utils/CompositePackageConfigurationProvider.kt
+++ b/model/src/main/kotlin/utils/CompositePackageConfigurationProvider.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2022 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model.utils
+
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.config.PackageConfiguration
+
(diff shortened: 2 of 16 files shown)


model/src/main/kotlin/utils/CompositePackageConfigurationProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.config.PackageConfiguration
+
+/**
+ * A [PackageConfigurationProvider] that combines the provided [providers] into a single provider. The order of the
+ * [providers] determines the order in which they are accessed when requesting a [PackageConfiguration] in
+ * [getPackageConfiguration].
+ */
+class CompositePackageConfigurationProvider(
```
sschuberth, line 31: I wonder whether we should somehow delimitate this more from `CompositePackageCurationProvider` to emphasize that this implementation does not accumulate / merge multiple `PackageConfiguration`s (which would be possible as long as they are not contradicting). This *could* simply be done by choosing another class name that does not include "composite". Because in the end, this is more like a `FallbackPackageCurationProvider` in that is uses the first match only.

On the other hand, package curations (in contrast to package configurations) were from the start designed to be cumulative, so maybe being more clear here is not needed after all. What do you think?

model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt
```diff
@@ -88,6 +60,31 @@ class SimplePackageConfigurationProvider(
...
+    SimplePackageConfigurationProvider(file.readValue<List<PackageConfiguration>>())
+
+private fun readDirectory(directory: File) =
+    FileFormat.findFilesWithKnownExtensions(directory).map { file ->
+        runCatching {
+            file.readValue<PackageConfiguration>()
+        }.getOrElse { e ->
+            throw IOException("Error reading package configuration from '${file.absolutePath}'.", e)
+        }
+    }
```
sschuberth, line 78: Nit: For grouping to its (currently) only caller, maybe move this to a companion object inside `DirectoryPackageConfigurationProvider`?

sschuberth (changes requested):

model/src/main/kotlin/utils/CompositePackageConfigurationProvider.kt
mnonnenmacher: Actually, in a later PR I want to change the interface to return all package configurations for a package, instead of only one. Now that we have multiple sources (global config and .ort.yml now, ort-config to be added) it makes a lot of sense to me, that you want to combine them. E.g. to amend the configurations from the ort-config repository with internal ones that can not be published.

model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt
mnonnenmacher: I thought we agreed a while ago in a dev meeting that we only move functions into the companion object if there is a technical or design reason to do so, to reduce overall indentation, but to put small helper functions like this on the top level.

sschuberth: You're right. It's just situations like this (multiple class declarations in a single file, resulting in unclear "ownership" of the private function) where I personally find our rule to not be very convenient to follow. But let's stick to what we agreed. However, maybe more the `readDirectory` function directly below `DirectoryPackageConfigurationProvider`?

model/src/main/kotlin/utils/CompositePackageConfigurationProvider.kt
sschuberth: Ok. So maybe put in a `TODO` for now for clarity? Or is that upcoming PR so close that it's not worth it?

mnonnenmacher: It will come very soon. I first wanted to include the change in the PR that introduces the ort-config provider, but the latter requires some design discussion first so I will make them separately.

model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt
mnonnenmacher: Done.

sschuberth (approved):

model/src/main/kotlin/utils/CompositePackageConfigurationProvider.kt
mnonnenmacher: The change was smaller than expected so I decided to add the commit to this PR as the reporter-web-app build was failing anyway for an unrelated reason.

model/src/main/kotlin/utils/CompositePackageConfigurationProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A [PackageConfigurationProvider] that combines the provided [providers] into a single provider. The order of the
+ * [providers] determines the order in which they are queried when requesting a [PackageConfiguration] in
+ * [getPackageConfiguration].
+ */
+class CompositePackageConfigurationProvider(
+    private val providers: List<PackageConfigurationProvider>
+) : PackageConfigurationProvider {
+    constructor(vararg providers: PackageConfigurationProvider) : this(providers.toList())
```
sschuberth, line 33: Sorry, one more nit: `toList` could become `asList` for efficiency. But this could also be done as a follow-up.

helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2021 HERE Europe B.V.
+ * Copyright (C) 2022 Bosch.IO GmbH
```
sschuberth, line 3: Commit message nits:
- "ammend" -> "amend"
- "important, once" -> no comma here

sschuberth (changes requested):

sschuberth (approved):

PR #5336, opened 2022-05-09
Allow empty definition files again

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Composer.kt
+++ b/analyzer/src/main/kotlin/managers/Composer.kt
@@ -44,8 +44,9 @@ import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
+import org.ossreviewtoolkit.model.jsonMapper
 import org.ossreviewtoolkit.model.orEmpty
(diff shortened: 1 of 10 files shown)


analyzer/src/main/kotlin/managers/Composer.kt
```diff
@@ -45,7 +45,7 @@ import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.orEmpty
-import org.ossreviewtoolkit.model.readJsonFile
+import org.ossreviewtoolkit.model.readTree
```
mnonnenmacher, line 48: Commit message nit: Remove "a" before "FileFormat" in the title.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/Composer.kt
sschuberth: The "a" is deliberately in there to indicate that `readTree()` is a newly added function, in contrast to "*the* ... `readJsonFile()`" as a existing function.

mnonnenmacher: Still seems wrong to me. What about "Replace `Mapper.readJsonFile()` with the new `FileFormat.readTree()`" instead?

mnonnenmacher (approved):

Issue #4668, opened 2021-11-05
Consider replacing `ProcessCapture` with something else


sschuberth:
Regarding the `getVersion()` invocation mentioned by @mnonnenmacher and @sschuberth, I did some testing. I ran a simple test that executes this method in a loop. What I found out is that there is indeed a small memory leak caused by `ProcessCapture` calling `deleteOnExit()` on the temporary files it creates to capture the process output. `deleteOnExit()` causes the name of the file affected to be recorded in a static Set managed by the internal `DeleteOnExitHook` system class, so that it can be deleted later.

So each usage of the `ProcessCapture` class adds three file names to this set (for stdout, stderr, and the temporary directory). This could be cured by ensuring manually that the files are cleaned up, e.g. using a try-finally construct. However, my gut feeling is that the amount of memory wasted here is not really significant.

_Originally posted by @oheger-bosch in https://github.com/oss-review-toolkit/ort/issues/3091#issuecomment-740462805_

sschuberth:
Possible alternatives (alphabetic order):
- https://github.com/05nelsonm/kmp-process
- https://github.com/brettwooldridge/NuProcess
- https://github.com/fleipold/jproc
- https://github.com/jbangdev/jbang-jash
- https://github.com/jnr/jnr-process
- https://github.com/kgit2/kommand
- https://github.com/lordcodes/turtle
- https://github.com/pgreze/kotlin-process
- https://github.com/zeroturnaround/zt-exec

*Edit:*

According to some AI-base analysis regarding features and maintenance, the candidates to look at first should be
1. https://github.com/zeroturnaround/zt-exec
2. https://github.com/brettwooldridge/NuProcess
3. https://github.com/05nelsonm/kmp-process
4. https://github.com/pgreze/kotlin-process

fviernau:
Maybe slightly related: Some of the `pip` commands the analyzer runs may take quite long. The process capture swallows all output during execution so that no progress can be observed at all and only provides the output after the execution has finished. This makes it IMO harder to debug. Maybe this can be solved together with migrating to something else.

mnonnenmacher:
Another alternative: https://github.com/lordcodes/turtle

heliocastro:
This last one opens a Pandora's box, no? 
If you decide to use one tool to manipulate CLI utils, then is like modifying all manipulation tools ?

sschuberth:
Some requirements that come to my mind:
- It should be possible to "define" a command line tool so that
[...]

PR #5337, opened 2022-05-09
ProcessCapture: Make `{stdout,stderr}File` private

sschuberth:
The fact that `ProcessCapture` uses temporary files to work around the
64k buffering limit is an implementation detail that should not be
exposed to ease switching to another process builder alternative
eventually. Also see [1].

[1]: https://github.com/oss-review-toolkit/ort/issues/4668

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -425,7 +425,7 @@ open class Npm(
 
     protected open fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
         val process = run(workingDir, "view", "--json", packageName)
-        return jsonMapper.readTree(process.stdoutFile)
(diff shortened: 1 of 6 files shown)


oheger-bosch (commented):
Agreed that the temporary files used internally should not be exposed to clients. However, to process potentially large results, it would be good to have some streaming abstraction, such as an input stream or (on a higher level) a sequence of lines (e.g. as provided by [useLines](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/use-lines.html)).

mnonnenmacher (approved):

sschuberth:
> However, to process potentially large results, it would be good to have some streaming abstraction

I agree, but that should be done as part of addressing https://github.com/oss-review-toolkit/ort/issues/4668, and not by adding that functionality ourselves.

PR #5339, opened 2022-05-11
Support unpacking unknown archives

### LINKED ISSUE #5329: NPM analyzer cannot handle URLs pointing to the GitHub npm registry

oheger-bosch:
This PR fixes #5329. Refer to the single commits for more information.

### DIFF
--- a/utils/common/src/main/kotlin/ArchiveUtils.kt
+++ b/utils/common/src/main/kotlin/ArchiveUtils.kt
@@ -111,6 +111,23 @@ fun File.unpack(
     }
 
 /**
+ * Try to unpack this [File] of an unknown archive type to [targetDirectory] using [filter] to select only the entries
+ * of interest. Use the file name to guess the archive type. If this is not possible due to a missing or unknown file
+ * extension, try out the supported archive types one after the other.
+ */
+fun File.unpackTryAllTypes(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = { true }) {
+    val typeFromName = ArchiveType.getType(name)
+
+    enumValues<ArchiveType>().mapNotNullTo(mutableListOf(typeFromName)) { type ->
+        type.takeUnless { type == typeFromName || type == ArchiveType.NONE }
+    }.find { archiveType ->
+        runCatching {
+            unpack(targetDirectory, forceArchiveType = archiveType, filter)
+        }.isSuccess
+    } ?: throw IOException("Unable to unpack '$this'. This file is not a supported archive type.")
+}
+
+/**
  * Unpack the [File] assuming it is a 7-Zip archive. This implementation ignores empty directories and symbolic links
  * and all entries not matched by the given [filter].
  */
--- a/utils/common/src/main/kotlin/ArchiveUtils.kt
+++ b/utils/common/src/main/kotlin/ArchiveUtils.kt
@@ -87,7 +87,7 @@ fun archive(inputDir: File, zipFile: File, prefix: String = ""): File =
 /**
  * Unpack the [File] to [targetDirectory] using [filter] to select only the entries of interest. If an archive type
  * other than [ArchiveType.NONE] is specified in [forceArchiveType], use this one; otherwise, try to detect the type
- * based on the file extension.
+ * based on the file extension. Throw an [IOException] if unpacking fails.
  */
 fun File.unpack(
     targetDirectory: File,
@@ -274,8 +274,11 @@ private fun ArchiveInputStream.unpack(
     mode: (ArchiveEntry) -> Int
 ) =
     use { input ->
+        var processed = false
+
         while (true) {
             val entry = input.nextEntry ?: break
+            processed = true
 
             if (shouldSkip(entry)) continue
 
@@ -291,6 +294,8 @@ private fun ArchiveInputStream.unpack(
 
             copyExecutableModeBit(target, mode(entry))
         }
+
+        if (!processed) throw IOException("Unsupported archive format or empty archive.")
     }
 
 /**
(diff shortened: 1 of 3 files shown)


utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -87,13 +87,13 @@ fun archive(inputDir: File, zipFile: File, prefix: String = ""): File =
 /**
  * Unpack the [File] to [targetDirectory] using [filter] to select only the entries of interest. If an archive type
  * other than [ArchiveType.NONE] is specified in [forceArchiveType], use this one; otherwise, try to detect the type
- * based on the file extension.
+ * based on the file extension. Return a flag whether the archive was not empty and could be processed.
```
sschuberth, line 90: Hmm. I guess at least for some archive types an empty archive is still valid archive, so I'd expect `true` to be returned in that case.

sschuberth: Also, what if the archive is corrupt? According to the docs I'd assume `false` to be returned in that case, whereas instead probably an exception is thrown by the downstream function.

So, what about simply determining the return value by wrapping the `when` in `runCatching` and returning `true` on success, and `false` on failure / if an exception was thrown? That would avoid the changes to the type-specific `unpack` functions.

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -110,6 +110,22 @@ fun File.unpack(
...
+ * extension, try out the supported archive types one after the other.
+ */
+fun File.unpackTryAllTypes(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = { true }) {
+    (listOf(ArchiveType.getType(this.name)) + ArchiveType.values().toList()).distinct().asSequence()
+        .filterNot { it == ArchiveType.NONE }
+        .firstOrNull { archiveType ->
+            val result = runCatching {
+                unpack(targetDirectory, forceArchiveType = archiveType, filter)
+            }
+            result.getOrDefault(false)
```
sschuberth, line 125: Chain these calls. However, when moving the `runCatching` to `unpack` as proposed in the previous commit you wouldn't need this at all.

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -110,6 +110,22 @@ fun File.unpack(
...
 
+/**
+ * Try to unpack this [File] of an unknown archive type to [targetDirectory] using [filter] to select only the entries
+ * of interest. Use the file name to guess the archive type. If this is not possible due to a missing or unknown file
+ * extension, try out the supported archive types one after the other.
+ */
+fun File.unpackTryAllTypes(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = { true }) {
+    (listOf(ArchiveType.getType(this.name)) + ArchiveType.values().toList()).distinct().asSequence()
+        .filterNot { it == ArchiveType.NONE }
+        .firstOrNull { archiveType ->
```
sschuberth, line 121: Use `find` instead of `firstOrNull`.

sschuberth: Overall proposal:

    val typeFromName = ArchiveType.getType(name)

    enumValues<ArchiveType>().mapNotNullTo(mutableListOf(typeFromName)) { type ->
        type.takeUnless { type == typeFromName || type == ArchiveType.NONE }
    }.find { type ->
        // ...
    }

sschuberth (changes requested):

utils/common/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: The problem I want to solve with this approach is that for some archive types no exception is thrown, even if the archive has a different format. So, the only indication whether the archive type is correct in such cases is that at least one entry can be extracted.

Would you prefer throwing an exception in these cases instead of returning a flag?

sschuberth: > Would you prefer throwing an exception in these cases instead of returning a flag?

As discussed separately, yes.

utils/common/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: This is no longer necessary.

utils/common/src/main/kotlin/ArchiveUtils.kt
oheger-bosch: Changed accordingly.

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -87,7 +87,7 @@ fun archive(inputDir: File, zipFile: File, prefix: String = ""): File =
 /**
  * Unpack the [File] to [targetDirectory] using [filter] to select only the entries of interest. If an archive type
  * other than [ArchiveType.NONE] is specified in [forceArchiveType], use this one; otherwise, try to detect the type
- * based on the file extension.
+ * based on the file extension. Throw an [IOException] if unpacking fails.
```
sschuberth, line 90: Commit message nit: No comma at beginning of ", this should not matter".

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -272,10 +272,13 @@ private fun ArchiveInputStream.unpack(
     targetDirectory: File,
     shouldSkip: (ArchiveEntry) -> Boolean,
     mode: (ArchiveEntry) -> Int
-) =
-    use { input ->
+) {
+    return use { input ->
```
sschuberth, line 293: Looks like this could stay a function expression, or?

sschuberth (changes requested):

utils/common/src/main/kotlin/ArchiveUtils.kt
sschuberth: > further tests have shown that only archive formats operating on `ArchiveInputStream` are affected

Thanks for finding this out!

sschuberth (approved):

sschuberth:

Issue #5236, opened 2022-04-07
Analyzer: WARNING: A connection to https://repo.maven.apache.org/ was leaked.


Alphavax:
The analyzer scan for a maven project throws a lot of warnings regarding leaked connections to maven repo:

08:44:09.264 [DefaultDispatcher-worker-3] INFO  org.ossreviewtoolkit.analyzer.managers.utils.MavenSupport - Fixed up invalid SCM connection 'scm:git://github.com/lihaoyi/fastparse.git' without a provider to VcsInfo(type=Git, url=git://github.com/lihaoyi/fastparse.git, revision=, path=).
Apr 07, 2022 8:44:11 AM okhttp3.internal.platform.Platform log
WARNING: A connection to https://repo.maven.apache.org/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
Apr 07, 2022 8:44:11 AM okhttp3.internal.platform.Platform log
WARNING: A connection to https://repo.maven.apache.org/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
Apr 07, 2022 8:44:11 AM okhttp3.internal.platform.Platform log

sschuberth:
I've also seen this, and did https://github.com/oss-review-toolkit/ort/pull/5026 a while back to set OkHttp's `FINE` when ORT's `DEBUG` is set, but so far had no lock seeing this again with a `--debug` run of ORT.

@Alphavax can you try to rerun with ORT's `--debug` i order to see OkHttp's `FINE` log output to hunt down the root cause?

Alphavax:
Hello Sebastian,

Not sure from where to start putting the debug output, but here are the lines before and after the warning:

[...]

PR #5343, opened 2022-05-12
Analyzer: Close Response Body on failure

janderssonse:
A Resource leak was partially fixed in
https://github.com/oss-review-toolkit/ort/issues/5240.
However the Response would still not be closed if unsuccessful, and the same warnings as in #5240 would appear if so.

By also closing (as recommended by OKHttp) before throwing the exception, it will not show the warn messages.
(and the potential resource leak).

Note: It can be argued it could be solved in less imperative ways by some overall refactoring,
but for a first PR I would like it to be as simple and straightforward as possible.

Signed-off-by: Josef Andersson <josef.andersson@gmail.com>

### DIFF
--- a/utils/core/src/main/kotlin/OkHttpClientHelper.kt
+++ b/utils/core/src/main/kotlin/OkHttpClientHelper.kt
@@ -241,6 +241,7 @@ fun OkHttpClient.download(url: String, acceptEncoding: String? = null): Result<P
         val body = response.body
 
         if (!response.isSuccessful || body == null) {
+            body?.close()
             throw HttpDownloadError(response.code, response.message)
         }
 


sschuberth:
> A Resource leak was partially fixed in #5236.

Nit: This text should actually link to the PR: https://github.com/oss-review-toolkit/ort/pull/5240

> Note: It can be argued it could be solved in less imperative ways by some overall refactoring,
but for a first PR I would like it to be as simple and straightforward as possible.

Welcome as a contributor, and thank you!

sschuberth:
> as recommended by OKHttp

Would you have any reference for that?

utils/core/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -241,7 +241,12 @@ fun OkHttpClient.download(url: String, acceptEncoding: String? = null): Result<P
         val body = response.body
 
         if (!response.isSuccessful || body == null) {
-            throw HttpDownloadError(response.code, response.message)
+            
+            val code=response.code 
+            val message=response.message
+            response.close()
```
sschuberth, line 247: As [according to OkHttp docs](https://github.com/square/okhttp/blob/0a27e9ff30c6129f4fdeb1f99c07c9b009bcc537/okhttp/src/main/kotlin/okhttp3/Response.kt#L295-L303) `close()` on a `Response` does nothing else than calling `body().close()` we IMO should

1. not need to save the `code` and `message` before closing,
2. verify that the `body` is not null to not run into the check inside OkHttp (which throws an exception).

So, could you verify whether simply adding the line `body?.close()` would also work?

janderssonse: Note: Will also fix the detect issues.:)

janderssonse: Tested suggest solution IRL project (and no longer get the leak messages on failed downloads), updated PR, fixed detekt issues.

sschuberth: Thanks. Would you mind squashing your two commits into one, and maybe change the commit message to say "failure" instead of "fail"?

PR #5344, opened 2022-05-12
FossID: Extract mapping of licenses to scanner specific configurations

MarcelBochtler:


nnobelis:
Maybe explain in the commit message why this change is needed ?

PR #5349, opened 2022-05-13
Do not store empty ZIPs

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
+++ b/cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
@@ -29,6 +29,8 @@ import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.types.file
 
+import java.io.IOException
+
 import org.eclipse.sw360.clients.adapter.AttachmentUploadRequest
 import org.eclipse.sw360.clients.adapter.SW360ProjectClientAdapter
 import org.eclipse.sw360.clients.adapter.SW360ReleaseClientAdapter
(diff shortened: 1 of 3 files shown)


oheger-bosch (commented):
Changes look good to me in principal; I have two points:
* Returning *null* from `packZip()` if no files are matched by the filter seems a bit, well, arbitrary. As an alternative, to make the meaning of the result more explicit, we could introduce a data class `PackResult` or something that contains the result archive file and the number of files it contains.
* Tests are missing for the new functionality.

sschuberth:
I've split out all but the last commit to https://github.com/oss-review-toolkit/ort/pull/5352. Please review that first.

sschuberth:
> Returning _null_ from `packZip()` if no files are matched by the filter seems a bit, well, arbitrary.

I absolutely agree it's arbitrary 😁 But as long as we don't have anything meaningful more to return, I believe it's fine.

> As an alternative, to make the meaning of the result more explicit, we could introduce a data class `PackResult` or something that contains the result archive file and the number of files it contains.

What I don't like about this is that in case of 0 matching file still an archive would be created (as I guess you'd want to the the result archive file non-nullable, so it as to point at at existing file). Or would you let the `File` instance point to a non-existing file if the number of files is 0? But that doesn't seem very clean either.

Should we maybe just throw an `IllegalArgumentException` on no matching files?

oheger-bosch:
> 
> What I don't like about this is that in case of 0 matching file still an archive would be created (as I guess you'd want to the the result archive file non-nullable, so it as to point at at existing file). Or would you let the `File` instance point to a non-existing file if the number of files is 0? But that doesn't seem very clean either.
> 
> Should we maybe just throw an `IllegalArgumentException` on no matching files?

Then, to be more aligned to the functional style, how about returning a `Result<File>`, which contains an Exception in case no files are matched by the filter?

sschuberth:
> Then, to be more aligned to the functional style, how about returning a `Result<File>`

That would in turn create some misalignment with the `unpack()` function that throw, but I'll give it a try and see how it looks.

sschuberth:
Can anyone reproduce the failure in `ExperimentalScannerTest` locally? Because I can't, neither on Linux nor on Windows.

oheger-bosch (approved):

mnonnenmacher:
> Can anyone reproduce the failure in `ExperimentalScannerTest` locally? Because I can't, neither on Linux nor on Windows.

I also cannot reproduce it, I have restarted the tests again just in case it was some temporary issue.

sschuberth:
> I have restarted the tests again just in case it was some temporary issue.

I also had tried that several times already 😞 But let's see...

sschuberth:
I was finally able to reproduce the issue (caused by a non-empty file archiver storage) and will rebase this once the fix in https://github.com/oss-review-toolkit/ort/pull/5363 got merged.

sschuberth:
I'll drop the change to throw an exception on empty archives again, as after a discussion with @mnonnenmacher it turned out that we (mis-)use empty archives as a "marker" in the archive if no matching license files could be found.

sschuberth:
Replaced by https://github.com/oss-review-toolkit/ort/pull/5388.

PR #5354, opened 2022-05-16
DependencyGraphBuilder: Handle cyclic dependencies

oheger-bosch:
If a dependency to a package was added to the builder which contains
another reference to this package in its dependencies, the builder
created multiple DependencyReference objects with the same index and
fragment. This caused a corrupted dependency graph which could not be
deserialized. Fix this by checking the fragment again before creating
a new DependencyReference.

Note that this issue was encountered with a real Gradle project where
the analyzer actually produced such cyclic dependencies.

### DIFF
--- a/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
+++ b/model/src/main/kotlin/utils/DependencyGraphBuilder.kt
@@ -363,6 +363,13 @@ class DependencyGraphBuilder<D>(
         }
 
         val fragmentMapping = referenceMappings[index.fragment]
+        if (index.root in fragmentMapping) {
+            // If this point is reached, the package has already been inserted when processing its dependencies.
+            // This means that there is a cyclic dependency. To handle this case correctly, the insert operation has
+            // to be started anew.
+            return addDependencyToGraph(scopeName, dependency, transitive)
+        }
+
         val ref = DependencyReference(
             pkg = index.root,
             fragment = index.fragment,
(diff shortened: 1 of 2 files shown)


sschuberth:
> If a dependency to a package

"*of* a package ..."?

> to this package in its dependencies

"to the *same* package ..."?

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -362,21 +362,26 @@ class DependencyGraphBuilder<D>(
...
-        val fragmentMapping = referenceMappings[index.fragment]
-        val ref = DependencyReference(
-            pkg = index.root,
-            fragment = index.fragment,
-            dependencies = transitiveDependencies.toSortedSet(),
-            linkage = dependencyHandler.linkageFor(dependency),
-            issues = issues
-        )
-        fragmentMapping[index.root] = ref
+        return referenceMappings[index.fragment].takeIf { index.root !in it }?.let { fragmentMapping ->
```
sschuberth, line 365: Probably just a matter of taste, but instead of more nesting I'd prefer to return early in the `index.root in it` case, like

    val fragmentMapping = referenceMappings[index.fragment]
    if (index.root in fragmentMapping) {
        // ...
        return addDependencyToGraph(scopeName, dependency, transitive)
    }

sschuberth (changes requested):

sschuberth (approved):

PR #5358, opened 2022-05-17
Raise a warning if 'yarn --info' returns empty data

bennati:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Yarn.kt
+++ b/analyzer/src/main/kotlin/managers/Yarn.kt
@@ -32,6 +32,7 @@ import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.jsonMapper
 import org.ossreviewtoolkit.utils.common.Os
+import org.ossreviewtoolkit.utils.core.log
 
 /**
  * The [Yarn](https://classic.yarnpkg.com/) package manager for JavaScript.
@@ -69,6 +70,8 @@ class Yarn(
 
     override fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
         val process = run(workingDir, "info", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)["data"]
+        return jsonMapper.readTree(process.stdout)["data"] ?: jsonMapper.readTree(process.stderr)["data"].also {
+            log.warn { "Error running '${process.commandLine}' in directory $workingDir: $it" }
+        }
     }
 }


analyzer/src/main/kotlin/managers/Yarn.kt
```diff
@@ -69,6 +70,12 @@ class Yarn(
...
     override fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
         val process = run(workingDir, "info", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)["data"]
+        val yarnInfo = jsonMapper.readTree(process.stdout)["data"]
+        if (yarnInfo == null) {
+            val yarnErr = jsonMapper.readTree(process.stderr)["data"]
+            log.warn("Error running `yarn info --json $packageName` in directory $workingDir: $yarnErr")
+            return yarnErr
+        }
+        return yarnInfo
```
mnonnenmacher, line 79: You can write this a bit more compact:
```kotlin
return jsonMapper.readTree(process.stdout)["data"] ?: jsonMapper.readTree(process.stderr)["data"].also {
  log.warn("Error running `yarn info --json $packageName` in directory $workingDir: $yarnErr")
}
```

And does the error message make sense? Does `yarnErr` contain the actual error message? Also you should use `process.commandLine` instead of manually repeating the command line in the message.

Regarding the commit message, is the VCS information really the only thing missing in case of the fallback?

mnonnenmacher (dismissed):

analyzer/src/main/kotlin/managers/Yarn.kt
sschuberth: And in any case please the curly-brace syntax for log statements: `log.warn { ... }`

analyzer/src/main/kotlin/managers/Yarn.kt
```diff
@@ -69,6 +70,9 @@ class Yarn(
 
     override fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
         val process = run(workingDir, "info", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)["data"]
+        val yarnErr = jsonMapper.readTree(process.stderr)["data"]
+        return jsonMapper.readTree(process.stdout)["data"] ?: yarnErr.also {
```
sschuberth, line 74: Doing it this way instead of how @mnonnenmacher proposed has the drawback that it's now always tried to deserialize from `stderr`, which might fail if there is no error.

bennati: fixed

sschuberth: But now you need drop the `yarnErr` variable 😉

bennati: done

analyzer/src/main/kotlin/managers/Yarn.kt
```diff
@@ -69,6 +70,8 @@ class Yarn(
 
     override fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
         val process = run(workingDir, "info", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)["data"]
+        return jsonMapper.readTree(process.stdout)["data"] ?: jsonMapper.readTree(process.stderr)["data"].also {
+            log.warn { "Error running `${process.commandLine}` in directory $workingDir: ${jsonMapper.readTree(process.stderr)["data"]}" }
```
sschuberth, line 74: `also` makes its context available as `it` inside the lamda, so `${jsonMapper.readTree(process.stderr)["data"]}` should simply become `$it` here.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Yarn.kt
sschuberth: Or you do `...also { yarnErr ->`, which is what @mnonnenmacher probably forgot in his example.

bennati: done

sschuberth (dismissed):

analyzer/src/main/kotlin/managers/Yarn.kt
```diff
@@ -69,6 +70,8 @@ class Yarn(
 
     override fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
         val process = run(workingDir, "info", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)["data"]
+        return jsonMapper.readTree(process.stdout)["data"] ?: jsonMapper.readTree(process.stderr)["data"].also { yarnErr ->
```
sschuberth, line 73: But with `yarnErr ->` the line becomes too long and you need to wrap it. So maybe stick to the implicit `it`after all...

bennati: done

sschuberth (dismissed):

sschuberth (approved):

sschuberth (approved):

sschuberth:

PR #5360, opened 2022-05-18
GenerateScopeExcludesCommand: exclude Yarn bundled dependencies

bennati:

### DIFF
--- a/helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
@@ -257,6 +257,18 @@ private fun getScopeExcludesForPackageManager(packageManagerName: String): List<
                 pattern = "devDependencies",
                 reason = ScopeExcludeReason.DEV_DEPENDENCY_OF,
                 comment = "Packages for development only."
+            ),
+            ScopeExclude(
+                pattern = "bundleDependencies",
+                reason = ScopeExcludeReason.PROVIDED_DEPENDENCY_OF,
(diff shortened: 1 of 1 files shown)


helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
```diff
@@ -257,6 +257,18 @@ private fun getScopeExcludesForPackageManager(packageManagerName: String): List<
...
+                pattern = "bundleDependencies",
+                reason = ScopeExcludeReason.PROVIDED_DEPENDENCY_OF,
+                comment = "Bundled dependencies are delivered with the " +
+                "dependent package, and they are unsupported by Yarn"
+            ),
+            ScopeExclude(
+                pattern = "bundledDependencies",
+                reason = ScopeExcludeReason.PROVIDED_DEPENDENCY_OF,
+                comment = "Bundled dependencies are delivered with the " +
+                "dependent package, and they are unsupported by Yarn"
```
sschuberth, line 271: BTW, I believe "provided" and "bundled" mean exactly opposite things here:
- "provided" dependencies are those that *need to be provided* by the user.
- "bundled" dependencies are those that *come with* the parent.

So, "bundles" dependencies should actually never be excluded (unless their parent also is), and using "provided" as the reason is wrong.

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
bennati: tbh, I forgot the context of this PR so we can close it

PR #5363, opened 2022-05-19
Allow to disable the file archiver

sschuberth:

### DIFF
--- a/model/src/main/kotlin/config/FileArchiverConfiguration.kt
+++ b/model/src/main/kotlin/config/FileArchiverConfiguration.kt
@@ -38,6 +38,11 @@ import org.ossreviewtoolkit.utils.core.storage.LocalFileStorage
 @JsonIgnoreProperties(value = ["patterns"])
 data class FileArchiverConfiguration(
     /**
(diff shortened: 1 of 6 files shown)


model/src/main/kotlin/config/FileArchiverConfiguration.kt
```diff
@@ -37,6 +37,11 @@ import org.ossreviewtoolkit.utils.core.storage.LocalFileStorage
  */
 @JsonIgnoreProperties(value = ["patterns"])
 data class FileArchiverConfiguration(
+    /**
+     * Toggle to enable or disable the file archiver functionality altogether.
+     */
+    val enabled: Boolean = true,
```
mnonnenmacher, line 43: Please add this also to `reference.conf`.

mnonnenmacher (changes requested):

model/src/main/kotlin/config/FileArchiverConfiguration.kt
sschuberth: Thanks for reminding me, done.

mnonnenmacher (approved):

PR #5366, opened 2022-05-20
Npm: Disable audit submission during `npm ci` or `npm install`

nnobelis:
According to the documentation ([1] and [2]), NPM submits by default a
dependency description to the configured registry and asks for a report
of known vulnerabilities.
Since:
* ORT does not use this information (security analysis is done
separately by the Advisor)
* this auditing should be done regularly and not only at install time
this commit deactivates this option for now.

[1]: https://docs.npmjs.com/cli/v8/commands/npm-ci
[2]: https://docs.npmjs.com/cli/v8/commands/npm-audit

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -645,5 +645,5 @@ open class Npm(
     }
 
     protected open fun runInstall(workingDir: File) =
-        run(workingDir, if (hasLockFile(workingDir)) "ci" else "install", "--ignore-scripts")
+        run(workingDir, if (hasLockFile(workingDir)) "ci" else "install", "--ignore-scripts", "--no-audit")
 }


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -645,5 +645,5 @@ open class Npm(
     }
 
     protected open fun runInstall(workingDir: File) =
-        run(workingDir, if (hasLockFile(workingDir)) "ci" else "install", "--ignore-scripts")
+        run(workingDir, if (hasLockFile(workingDir)) "ci" else "install", "--ignore-scripts --no-audit")
```
sschuberth, line 648: `"--no-audit"` should be passed as a separate argument to `run`.

sschuberth: And just checking, is there a similar logic in Yarn that we should disable, too?

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Npm.kt
nnobelis: For yarn:

>  Unlike npm, which automatically runs an audit on every install, yarn will only do so when requested.

https://classic.yarnpkg.com/en/docs/cli/install

So nothing is needed.

sschuberth (approved):

Issue #5365, opened 2022-05-20
Scanning packages should not claim the same file for different packages (of the same type)


sschuberth:
Currently, scanning in ORT is package-based, and (project-)packages are identified by "definition files" (like "pom.xml", "build.gradle", "package.json" etc.) in the directory tree. So all files and directories below a definition file are regarded as belonging to the project that the definition file defines. See:

```
ROOTDIR
|
+-SUBPROJDIR_A
| |
| +-pom-a.xml
| |
| +-license-a.txt
|
+-SUBPROJDIR_B
| |
| +-pom-b.xml
| |
| +-license-b.txt
|
+-WEBDIR
| |
| +-package.json
| |
| +-license-w.txt
|
+-pom.xml
|
+-license.txt
```

So, the project spanned by `pom.xml` in the root directory is considered to "own" all files below the root directory, including `SUBPROJDIR_A/license-a.txt`, `SUBPROJDIR_B/license-b.txt` and `WEBDIR/license-w.txt`. This means that scanner findings in those file will get associated to the roo project.

However, when the scanner's view shifts to the (project-)packages in the subdirectories, the project spanned by `SUBPROJDIR_A/pom-a.xml` *also* gets the scan result for `SUBPROJDIR_A/license-a.txt` assigned (similar for the other subprojects).

This is historically so because ORT not really understand the semantics of a project's directory tree. However, the result can be really confusing, as scan findings (and potential violations) might show up multiple times in the reports, although they all stem from the same single file.

As a solution to this, one idea is to associate files always only to the nearest enclosing project when walking up the directory tree to the root. Maybe this logic should be limited to projects of the same type / package manager only; however, in the example above this would result in the scanner findings from `WEBDIR/license-w.txt` to still be associated to the root project spanned by `pom.xml`.

I have some hopes that the required filtering logic would be easier to implement once https://github.com/oss-review-toolkit/ort/issues/2668 is merged, as it implements some similar filtering to associate provenance-based scan results to individual packages IIUC.

sschuberth:
And as mentioned by @mnonnenmacher over [here](https://github.com/oss-review-toolkit/ort/pull/3145#discussion_r530331646), in the context of this it should also be ensured that *Unmanaged* project only get the "left-overs" associated that cannot be accosicated to any other project according to the logic.

PR #5369, opened 2022-05-23
Analyzer: Remove projects that are also dependencies

bennati:
Local directories that are dependencies can appear in the
analyzer results both as packages, e.g. Pub, and projects.
After building the results, projects are removed if they match
with packages on VCS information and package manager type.

This is caused by issue #5365. This is only a workaround which
should be replaced by a comprehensive solution for #5365.

### DIFF
--- a/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
+++ b/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
@@ -42,6 +42,7 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
     private val dependencyGraphs = sortedMapOf<String, DependencyGraph>()
 
     fun build(): AnalyzerResult {
+        deduplicateProjects()
         val duplicateIds = (projects.map { it.id } + packages.map { it.pkg.id }).getDuplicates()
         require(duplicateIds.isEmpty()) {
             "AnalyzerResult contains packages that are also projects. Duplicates: '$duplicateIds'."
@@ -114,5 +115,31 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
     fun addDependencyGraph(packageManagerName: String, graph: DependencyGraph): AnalyzerResultBuilder {
         dependencyGraphs[packageManagerName] = graph
         return this
+    }
+
+    fun deduplicateProjects() {
+        val packages = this.packages.filter { p ->
+            p.pkg.vcsProcessed.path != ""
+        }
+        val duplicatedProjects = this.projects.filter { prj ->
+            prj.vcsProcessed.path != ""
+        }.filter { prj ->
+            val matchingPackages = packages.filter { p ->
+                prj.id.type == p.pkg.id.type &&
+                prj.vcsProcessed == p.pkg.vcsProcessed
+            }
+            if (matchingPackages.size > 0) {
+                matchingPackages.map { p ->
+                    log.warn {
+                        "Project ${prj.id.type}:${prj.id.namespace}:${prj.id.name} " +
+                        "resembles package ${p.pkg.id.type}:${p.pkg.id.namespace}:${p.pkg.id.name}"
+                    }
+                }
+            }
+            matchingPackages.size != 0
+        }
+        duplicatedProjects.map { prj ->
+            this.projects.remove(prj)
+        }
     }
 }


bennati:
We have a project where ~20 packages are mistakenly identified as projects, I'm not even sure how to solve these issues using curations.
A proper fix is likely too far away for our release schedule.
Happy to evaluate alternatives.

mnonnenmacher:
@sschuberth What do you mean by per-package-manager? This seems to be a generic workaround that affects all package managers where the described issue could appear.

@bennati I would rather know more details about the issue and fix the root cause instead of adding this workaround. Can you provide more details in which case projects also appear as packages? This should not happen in general. I also don't understand how this realtes to #5365, as this PR is about project/package duplicates while the issue addresses the association of files to projects.

bennati:
@mnonnenmacher sure, let's take an example directory structure like the following:
```
project_root/
|--- lib_dependency/
|--- app/
       |--- pubspec.yaml
```

And `pubspec.yaml` contains the following
```
dependencies:
  here_lib_dependency:
    path: ../lib_dependency
```

The scanner will first identify `lib_dependency` as a Project when traversing the file system, then the same will be identified as Package when analyzing `pubspec.yaml`. The VCS info will be the same, but the name will differ, and this raises issues as the folder is in one case identified as HERE proprietary and in the other not.

Hope this helps

bennati:
https://github.com/oss-review-toolkit/ort/commit/cea7f66ea43c4ca61cebbed2835d5681acd8d03a 
breaks this patch.
I would like clarity on whether this contribution is actually wanted before spending time to adapt it.

mnonnenmacher:
> [cea7f66](https://github.com/oss-review-toolkit/ort/commit/cea7f66ea43c4ca61cebbed2835d5681acd8d03a) breaks this patch. I would like clarity on whether this contribution is actually wanted before spending time to adapt it.

@bennati If anything is removed it should be the duplicate packages, not projects. Because if a `pubspec.yaml` is in the file tree it is a project from the ORT perspective.
But as mentioned above I would prefer to fix the root cause instead of adding a work-around. As the problem seems to be that the Pub implementation does not recognize that the dependency is also a local project, the real fix should solve this somehow. I won't have time to look into this anytime soon, but if you want to work on this, my suggestion would be to add a similar project dependency to our synthetic pub test project first to document the issue, and then find a way to detect project dependencies in the Pub implementation.
The solution could use a similar approach as the Maven implementation, where all projects are detected in the `beforeResolution()` function to later resolve project dependencies without having to build the Maven project.

bennati:
@mnonnenmacher that sounds a good approach.
Given that I won't have time to work on this anytime soon either, could we go for this approach and then revert it once the full fix is implemented?

mnonnenmacher:
> @mnonnenmacher that sounds a good approach. Given that I won't have time to work on this anytime soon either, could we go for this approach and then revert it once the full fix is implemented?

If we implement a workaround it should remove the duplicate packages, not projects. Note that a similar change was just done in #5509 for `SpdxDocumentFile`.

mnonnenmacher:
> @mnonnenmacher that sounds a good approach. Given that I won't have time to work on this anytime soon either, could we go for this approach and then revert it once the full fix is implemented?

If we implement a workaround it should remove the duplicate packages, not projects. Note that a similar change was just done in #5509 for `SpdxDocumentFile`.

bennati:
The reason why I removed duplicate projects instead of duplicate packages is that the correct name for the dependency is specified in `pubspec.yaml`, while the name of the corresponding project is given by the path it's in.

mnonnenmacher:
@bennati I'm not sure what you mean by "the project was generated from the directory structure", but if you mean that the project name is the name of the folder of the definition file, that means that the fallback in this line is used:
https://github.com/oss-review-toolkit/ort/blob/1ca71eb3c2ab8b8379eb9f387b13d0b66e276b83/analyzer/src/main/kotlin/managers/Pub.kt#L445

If this is the case, it seems that the name is not properly set in the pubspec.yaml, or that the linked code needs to be improved.

mnonnenmacher:
> If this is the case, it seems that the name is not properly set in the pubspec.yaml, or that the linked code needs to be improved.

The code was indeed wrong which means that project names were never parsed correctly, I have added a fix to #5524.

sschuberth:
> The code was indeed wrong which means that project names were never parsed correctly, I have added a fix to #5524.

@bennati is this PR still needed with the above-mentioned PR?

bennati:
No, we can close this

PR #5370, opened 2022-05-23
Add JSON Schema for package configuration

mmurto:
Add a JSON Schema for package configuration for IDE validation. 

Can be tested with VSCode by adding the following in `.vscode/settings.json` of a workspace containing package configurations and copying the schema to `schemas/package-configuration-schema.json` of the workspace:

```yaml
{
  "yaml.schemas": {
    "./schemas/package-configuration-schema.json": [
      "source-artifact.yml",
      "vcs.yml"
    ]
  },
}
```

### DIFF
--- a/integrations/schemas/package-configuration-schema.json
+++ b/integrations/schemas/package-configuration-schema.json
@@ -1,0 +1,158 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$id": "https://oss-review-toolkit.org/package-configuration.yml",
+  "title": "ORT package configuration",
+  "description": "The OSS-Review-Toolkit (ORT) provides a possibility to define path excludes and license finding curations for a specific package (dependency) and provenance in a package configuration file. A full list of all available options can be found at https://github.com/oss-review-toolkit/ort/blob/main/docs/config-file-package-configuration-yml.md.",
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string"
+    },
+    "license_finding_curations": {
+      "items": {
+        "$ref": "#/definitions/licenseFindingCurations"
+      },
+      "type": "array"
+    },
+    "path_excludes": {
+      "items": {
+        "properties": {
+          "comment": {
+            "type": "string"
+          },
+          "pattern": {
+            "type": "string"
+          },
+          "reason": {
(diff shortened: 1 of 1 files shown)


integrations/schemas/package-configuration-schema.json
```diff
@@ -0,0 +1,158 @@
...
+      "required": [
+        "vcs"
+      ]
+    },
+    {
+      "required": [
+        "source_artifact_url"
+      ]
+    }
+  ]
```
mmurto, line 157: Note: this should be correct according to the JSON Schema spec, but vscode-yaml plugin currently [does not correctly handle it](https://github.com/redhat-developer/vscode-yaml/issues/769). The plugin does not produce error if both `vcs` and `source_artifact_url` are provided.

sschuberth:
LGTM, but maybe @MarcelBochtler can chime in as he's got the most experience here.

Also, can we submit this to a schema registry, too?

mmurto:
> Also, can we submit this to a schema registry, too?

That would be great!

integrations/schemas/package-configuration-schema.json
```diff
@@ -0,0 +1,158 @@
...
+}
```
MarcelBochtler, line 158: Empty line missing at the end of the file.

MarcelBochtler (changes requested):

MarcelBochtler:
I also hoped that it would be possible to use this package-configuration schema and somehow include it as a reference in the `repository-configuration-schema.json`.
But as I didn't find a way to do that some months ago, I'd be okay, to duplicate the data.

integrations/schemas/package-configuration-schema.json
sschuberth: When addressing this @mmurto, please also rebase onto latest `main` to get rid of a test failure.

mmurto:
> I also hoped that it would be possible to use this package-configuration schema and somehow include it as a reference in the `repository-configuration-schema.json`. But as I didn't find a way to do that some months ago, I'd be okay, to duplicate the data.

I haven't tried it myself, but according to [this](https://json-schema.org/understanding-json-schema/structuring.html), it should ne possible to refer to schema from inside of another.

PR #5379, opened 2022-05-24
Gradle: Rename core-utils to ort-utils

mnonnenmacher:
In an upcoming change the `model` module will be renamed to `core`.
Rename `core-utils` to `ort-utils` to avoid confusion, and to emphasize
that the module contains ORT specific utilities.


sschuberth:
> In an upcoming change the `model` module will be renamed to `core`.

Did we really want to rename, or split? I don't recall anymore 😆

mnonnenmacher:
> > In an upcoming change the `model` module will be renamed to `core`.
> 
> Did we really want to rename, or split? I don't recall anymore laughing

Actually both, but I didn't want to go into detail with the split because we still need to decide how exactly we want to do it.

sschuberth:
Sorry, my recent merge caused a conflict here now.

sschuberth (approved):

mnonnenmacher:
@sschuberth  Need another review because the GitHub actions did not start and I made a force push to trigger them again.

sschuberth (approved):

PR #5382, opened 2022-05-25
.ort.yml: Update configuration for license mappings

mnonnenmacher:
Exclude the whole resources directory of spdx-utils to also exclude the
license and exception mapping files. Conclude the license for the YAML
files containing the mappings instead of the Kotlin files which do not
contain the mappings anymore since 69fa315 and 87127d8.


sschuberth (approved):
Just a though: Maybe after all necessary fixes we could change the comment style in this file to `comment: |` (like in the curations) for readability.

PR #5386, opened 2022-05-25
SpdxDocumentFile: Warn about artifact download locations that look like Git

sschuberth:
It is a common mistake by SPDX document authors to omit the VCS-specific
prefix in download locations that refer to VCS locations instead of
artifacts, so warn about this at least in the Git case.

[1]: https://spdx.github.io/spdx-spec/package-information/#77-package-download-location-field

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -113,7 +113,16 @@ private fun SpdxPackage.getRemoteArtifact(): RemoteArtifact? =
     when {
         SpdxConstants.isNotPresent(downloadLocation) -> null
         SPDX_VCS_PREFIXES.any { (prefix, _) -> downloadLocation.startsWith(prefix) } -> null
-        else -> RemoteArtifact(downloadLocation, Hash.NONE)
+        else -> {
+            if (downloadLocation.endsWith(".git")) {
+                log.warn {
+                    "The download location $downloadLocation of SPDX package '$spdxId' looks like a Git repository " +
+                            "URL but it lacks the 'git+' prefix and thus will be treated as an artifact URL."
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -113,7 +113,16 @@ private fun SpdxPackage.getRemoteArtifact(): RemoteArtifact? =
...
+        else -> {
+            if (downloadLocation.endsWith(".git")) {
+                log.warn {
+                    "The download location $downloadLocation of SPDX package '$spdxId' looks like a Git repository " +
+                            "URL but it lacks the 'git+' prefix and thus will be treated as an artifact URL."
+                }
+            }
+
+            RemoteArtifact(downloadLocation, Hash.NONE)
+        }
```
mnonnenmacher, line 125: It would require a bigger refactoring, but I think it would be very useful to also create an `OrtIssue` for this, because it will make the problem more visible for users.

sschuberth: I absolutely agree. However, the bigger factoring was the reason why I didn't do any better than the current proposal. In the end, it would be nice if an "issues registry" would be available "everywhere" via DI or something, so `OrtIssue`s could be created anytime from anywhere...

mnonnenmacher: Indeed, but that would be an even bigger refactoring.
I found one commit message nit: Typo in "locations".

mnonnenmacher (changes requested):

sschuberth:
> I found one commit message nit: Typo in "locations".

Fixed.

mnonnenmacher (approved):

PR #5389, opened 2022-05-25
PackageRule: Provide possibility to ignore resolved vulnerabilities

MarcelBochtler:


sschuberth:
I'm ok with this, but would also like @mnonnenmacher to comment.

In the end, we should rethink the way we handle vulnerabilities in ORT. Right now I believe it was wrong to allow vulnerabilities to get resolved without them being policy violations. Advisors should just be "fact providers", not much different from metadata as determined by the Analyzer. The fact that a vulnerability is present should not yet force you to resolve it; only if also a policy triggers on that vulnerability that should be possible. However, that would first requite to refactor our policy violations to be more generic and not license-centric. If we had policy violations that could be pure security violations caused by vulnerabilities without any license issues, then we could exclusively use policy violations resolutions to address vulnerabilities.

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -84,15 +86,22 @@ open class PackageRule(
...
      * A [RuleMatcher] that checks whether any vulnerability for the [package][pkg] has a score that equals or is
-     * greater than [threshold] according to the [scoringSystem] and the belonging [severityComparator].
+     * greater than [threshold] according to the [scoringSystem] and the belonging [severityComparator]. Resolved
+     * vulnerabilities are ignored if a resolution exists in the [resolutionProvider].
      */
-    fun hasVulnerability(threshold: String, scoringSystem: String, severityComparator: (String, String) -> Boolean) =
+    fun hasVulnerability(
+        threshold: String,
+        scoringSystem: String,
+        resolutionProvider: ResolutionProvider = DefaultResolutionProvider.create(),
```
mnonnenmacher, line 95: Do we need a default value here? It introduces the risk that the parameter is forgotten.
If you add the `resolutionProvider` to the `PackageRule` as well in the previous commit, all matchers could access it without requiring an argument.

mnonnenmacher (changes requested):

mnonnenmacher:
> I'm ok with this, but would also like @mnonnenmacher to comment.
> 
> In the end, we should rethink the way we handle vulnerabilities in ORT. Right now I believe it was wrong to allow vulnerabilities to get resolved without them being policy violations. Advisors should just be "fact providers", not much different from metadata as determined by the Analyzer. The fact that a vulnerability is present should not yet force you to resolve it; only if also a policy triggers on that vulnerability that should be possible. However, that would first requite to refactor our policy violations to be more generic and not license-centric. If we had policy violations that could be pure security violations caused by vulnerabilities without any license issues, then we could exclusively use policy violations resolutions to address vulnerabilities.

I see advantages in both approaches. For flexibility it is nice that only the analyzer + advisor can be used, without any need for writing evaluator rules. This could be amended with the option to define thresholds for each scoring system, comparable to the `severeIssueThreshold` and `severeRuleViolationThreshold`. On the other hand rules are much more versatile and can implement more complex requirements and having only a single system for resolving problems makes it simpler. I think this needs more discussion, and until we come to a conclusion this change at least allows to use package rules for vulnerabilities correctly.

evaluator/src/main/kotlin/PackageRule.kt
MarcelBochtler: The default is intended.
Using `DefaultResolutionProvider.create()` means that no resolution provider is being used and therefore no resolutions are applied.
I could also use `resolutionProvider: ResolutionProvider? = null` to make this more explicit.
What do you think?

MarcelBochtler: Added the `resolutionProvider` to the `PackageRule`.
However, I opted to still use a default:
```kt
fun packageRule(
        name: String,
        resolutionProvider: ResolutionProvider = DefaultResolutionProvider.create(),
        configure: PackageRule.() -> Unit
    )
```

Otherwise, this would break most existing `rules.kts`.

evaluator/src/main/kotlin/RuleSet.kt
```diff
@@ -33,7 +35,8 @@ import org.ossreviewtoolkit.utils.ort.log
  */
 class RuleSet(
     val ortResult: OrtResult,
-    val licenseInfoResolver: LicenseInfoResolver = ortResult.createLicenseInfoResolver()
+    val licenseInfoResolver: LicenseInfoResolver = ortResult.createLicenseInfoResolver(),
+    val resolutionProvider: ResolutionProvider = DefaultResolutionProvider.create()
```
MarcelBochtler, line 39: Does this default make sense? Or should it be `DefaultResolutionProvider.create(ortResult)`?

If the user doesn't provide the `resolutionProvider` should no resolutions be applied at all, or should only the resolutions from the OrtResult be applied?

mnonnenmacher (approved):

sschuberth (approved):

PR #5392, opened 2022-05-25
Gradle: Improve the detection of unresolvable configurations

mnonnenmacher:
Align the logic to detect unresolvable configurations with the logic
used by Gradle [1]. This prevents ORT from trying to resolve some
configurations that shall not be resolved, for example the "default"
configuration in many projects, and thereby avoids issues during the
resolution which are impossible to resolve. For example, in such
configurations Gradle can sometimes not determine the correct variant of
a dependency.

[1]: https://github.com/gradle/gradle/blob/0109a3c028c6ac8c6040c5dfcbdec2e5a9c51539/subprojects/diagnostics/src/main/java/org/gradle/api/tasks/diagnostics/internal/ConfigurationDetails.java#L40-L43

### DIFF
--- a/analyzer/src/main/resources/scripts/init.gradle
+++ b/analyzer/src/main/resources/scripts/init.gradle
@@ -168,9 +168,7 @@ class AbstractDependencyTreePlugin<T> implements Plugin<T> {
             }
 
             List<Configuration> configurations = project.configurations.findResults { configuration ->
-                // For versions of Gradle before the "canBeResolved" property was introduced, consider any
-                // configuration to be resolvable.
-                if (!configuration.hasProperty('canBeResolved') || configuration.canBeResolved) {
+                if (canBeResolved(configuration)) {
                     ResolutionResult result = configuration.getIncoming().getResolutionResult()
                     Set<ResolvedArtifact> resolvedArtifacts = []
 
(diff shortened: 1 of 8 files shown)


analyzer/src/main/resources/scripts/init.gradle
```diff
@@ -353,6 +351,22 @@ class AbstractDependencyTreePlugin<T> implements Plugin<T> {
             }
         }
 
+        private static boolean canBeResolved(org.gradle.api.artifacts.Configuration configuration) {
+            // The DeprecatableConfiguration class is available since Gradle 6.0.
+            boolean isDeprecatedForResolving = false
+            if (configuration.metaClass.respondsTo(configuration, "getResolutionAlternatives")) {
+                isDeprecatedForResolving = configuration."getResolutionAlternatives"() != null
```
sschuberth, line 358: Nit: Maybe use single quotes like in `hasProperty('canBeResolved')`?

sschuberth: Also, just wondering, are the quotes in the second line even required? Line 364 also seems to require no quotes after the check that the property is present.

sschuberth: Also semantically, I would have preferred to check whether `configuration` is a `DeprecatableConfiguration`, and if so call `getResolutionAlternatives()`. That would make more clear, why you're referring to that class in the comment.

analyzer/src/main/resources/scripts/init.gradle
```diff
@@ -353,6 +351,22 @@ class AbstractDependencyTreePlugin<T> implements Plugin<T> {
...
+            }
+
+            // The canBeResolved property is available since Gradle 3.3.
+            boolean canBeResolved = true
```
sschuberth, line 362: Couldn't you stick to the original one-liner?

    boolean canBeResolved = !configuration.hasProperty('canBeResolved') || configuration.canBeResolved

analyzer/src/main/resources/scripts/init.gradle
```diff
@@ -353,6 +351,22 @@ class AbstractDependencyTreePlugin<T> implements Plugin<T> {
...
+            if (configuration.hasProperty('canBeResolved')) {
+                canBeResolved = configuration.canBeResolved
+            }
+
+            return canBeResolved && !isDeprecatedForResolving;
```
sschuberth, line 367: You should be able to drop the semicolon.

sschuberth (changes requested):

analyzer/src/main/resources/scripts/init.gradle
mnonnenmacher: I have adapted the comment to make clear why I check for the existence of the function.

sschuberth: > I have adapted the comment to make clear why I check for the existence of the function.

Did that happen yet?

sschuberth (approved):

analyzer/src/main/resources/scripts/init.gradle
sschuberth: Ah, sorry, I was looking for this in the commit message, not in the code comment.

sschuberth (approved):

PR #5395, opened 2022-05-26
JSON Schemas: Use references in repository configuration schema

MarcelBochtler:
To test this, you can use this configuration in VSCode `settings.json`:

For the .ort.yml:
```json
  "yaml.schemas": {
    "file:///Users/marcel/work/oss/ort/integrations/schemas/repository-configuration-schema.json": "*.ort.yml"
  }
```

PR #5396, opened 2022-05-26
Schemas: Add JSON schema for curations

MarcelBochtler:
Moving this from https://github.com/oss-review-toolkit/ort/pull/5395 to a separate PR, as the test won't pass without this being merged first.

This can be tested with this setting in VSCode:
```json
  "yaml.schemas": {
    "file:///Users/marcel/work/oss/ort/integrations/schemas/curations-schema.json": "*.yml"
  }
```

### DIFF
--- a/model/src/test/kotlin/JsonSchemaTest.kt
+++ b/model/src/test/kotlin/JsonSchemaTest.kt
@@ -47,7 +47,7 @@ class JsonSchemaTest : StringSpec() {
             errors should beEmpty()
         }
 
-        ".ort.yml examples validates successfully" {
+        ".ort.yml examples validate successfully" {
             val examplesDir = File("../examples")
             val exampleFiles =
                 examplesDir.walk().filterTo(mutableListOf()) { it.isFile && it.name.endsWith(".ort.yml") }
--- a/model/src/test/kotlin/JsonSchemaTest.kt
+++ b/model/src/test/kotlin/JsonSchemaTest.kt
@@ -60,6 +60,19 @@ class JsonSchemaTest : StringSpec() {
                 errors should beEmpty()
             }
         }
+
+        "curation.yml example validates successfully" {
+            val curationsSchema = JsonSchemaFactory
+                .builder(JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7))
+                .objectMapper(mapper)
+                .build()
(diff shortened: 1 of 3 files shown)


model/src/test/kotlin/JsonSchemaTest.kt
```diff
@@ -60,6 +60,19 @@ class JsonSchemaTest : StringSpec() {
                 errors should beEmpty()
             }
         }
+
+        "curation.yml example validates" {
```
sschuberth, line 64: Nit: All other tests say "validates successfully", so we should probably do so here, too.

model/src/test/kotlin/JsonSchemaTest.kt
```diff
@@ -73,6 +73,19 @@ class JsonSchemaTest : StringSpec() {
 
             errors should beEmpty()
         }
+
+        "package-configuration.yml validates" {
```
sschuberth, line 77: Nit: All other tests say "validates successfully", so we should probably do so here, too.

sschuberth (changes requested):

model/src/test/kotlin/JsonSchemaTest.kt
```diff
@@ -73,6 +73,19 @@ class JsonSchemaTest : StringSpec() {
 
             errors should beEmpty()
         }
+
+        "package-configuration.yml validates successfully" {
+            val packageConfigurationSchema = JsonSchemaFactory
+                .builder(JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4))
```
sschuberth, line 79: BTW, any idea why we're using V4 here, but V7 for `curations.yml`? I mean, yes these flags correspond to the *draft-XX* in the schema files, but why are we using different version after all?

sschuberth (approved):

model/src/test/kotlin/JsonSchemaTest.kt
MarcelBochtler: V7 in the `curations-schema.yml` is required because I'm using if-then-else there which is only available for draft-07.
I explained it in this commit in the follow up PR: https://github.com/oss-review-toolkit/ort/pull/5395/commits/cb9da0a2d7d6e7c66253580590485c25ca7d9c62

PR #5397, opened 2022-05-27
FossId: Explicitly install the ORT authenticator

oheger-bosch:
The authenticator is needed to add credentials to the Git URL passed
to FossID. Recent changes in the scanner infrastructure have obviously
caused that the ORT authenticator has not always been installed when a
new FossID scan is created. The scan then fails due to missing
credentials.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -76,6 +76,7 @@ import org.ossreviewtoolkit.scanner.experimental.ScanContext
 import org.ossreviewtoolkit.utils.common.enumSetOf
 import org.ossreviewtoolkit.utils.common.replaceCredentialsInUri
 import org.ossreviewtoolkit.utils.common.toUri
+import org.ossreviewtoolkit.utils.ort.installAuthenticatorAndProxySelector
 import org.ossreviewtoolkit.utils.ort.log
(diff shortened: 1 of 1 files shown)


sschuberth (dismissed):

sschuberth:
FYI @mnonnenmacher, that looks like an interesting side-effect of switching to scan-by-repo. However, I guess it was always wrong to implicitly rely on the ORT authenticator to be installed.

sschuberth:
@oheger-bosch the commit lacks a signed-off-by.

oheger-bosch:
> @oheger-bosch the commit lacks a signed-off-by.

Sorry about that. Fixed now.

oheger-bosch:
> FYI @mnonnenmacher, that looks like an interesting side-effect of switching to scan-by-repo. However, I guess it was always wrong to implicitly rely on the ORT authenticator to be installed.

Agreed. If the authenticator is invoked, the caller needs to ensure that it is installed. Maybe, in addition to the `installAuthenticatorAndProxySelector()` helper function, there could be another function to query credentials for a given URL that calls the authenticator and ensures that it is installed?

sschuberth (approved):

sschuberth:
> there could be another function to query credentials for a given URL that calls the authenticator and ensures that it is installed?

Yeah, maybe a thin ORT-specific wrapper around `Authenticator.requestPasswordAuthentication` could make sense...

Issue #5345, opened 2022-05-12
Licenses mapped from scanner results should be user configurable


MarcelBochtler:
At the moment, ORT hard-codes multiple license mappings for FossID:
https://github.com/oss-review-toolkit/ort/blob/99e51faa30b890353f674bafdeb0da9153e45d0a/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt#L38-L82

and for ScanCode:
https://github.com/oss-review-toolkit/ort/blob/99e51faa30b890353f674bafdeb0da9153e45d0a/scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt#L70-L93

These mappings might not fit / be sufficient for all ORT users, therefore the mappings should be configurable in the `ort.conf` file.

MarcelBochtler:
The current plan for the implementation is to store the scan_results in the `ScanStorage` without applying the mapping, but when reading the results either from any `ScanStorage` or from the Scanner directly, apply the mapping.

PR #3731, opened 2021-03-05
Improvements to ORT's configuration management

sschuberth:
Please have a look at the individual commit messages for the details.

Issue #3522, opened 2021-01-21
Move hard-coded defaults to `default.conf`


sschuberth:
Yes, because so far there are either no defaults for the config options or they have not been moved to this file.
Putting them in both places means to places to maintain so I'd prefer just one of them, but we could keep the defaults here for now and then make a separate change to move all defaults to `default.conf` later.

_Originally posted by @mnonnenmacher in https://github.com/oss-review-toolkit/ort/pull/3510#r561082591_

fviernau:
Another option would be to have the defaults in the Kotlin code (configuration) and generate a 'default.conf' from it if that file is actually needed. This would ensure that ORT does actually have a reasonable default behavior without requiring the user to copy and paste the default configuration. Further it would be fully supported by the IDE including refactoring. Also constants defined in Kotlin code can be re-used. Also, for the user it might be simpler to revert parts of the configuration back to default, because its simpler to delete a line compared to finding the line in the default config and copy pasting it. Finally the user would remain up to date with default values automatically.

...just as an idea to consider

oheger-bosch:
As `default.conf` is part of the ORT binary and the file is automatically included as a configuration source when constructing the global `OrtConfiguration`, the values contained in this file are always available.

This approach is quite similar to the approach taken by the [Lightbend Config library](https://github.com/lightbend/config). In the section about [defaults](https://github.com/lightbend/config#how-to-handle-defaults), they state that the API of the library does not support passing default values for missing configuration properties from application code. The philosophy is rather to have a reference configuration file (that corresponds to `default.conf` in our case) with defaults for all supported properties, and an application configuration (`ort.conf`), in which the user can override values that should be different from defaults.

oheger-bosch:
While looking into this, I spotted a problem with the default value for the `commandLineNonConfig` property: It is defined as `"--processes", max(1, Runtime.getRuntime().availableProcessors() - 1).toString()`, so it contains a dynamic computation. How can we deal with this?

There are multiple options:
* Place a hard-coded value for the number of processes into `default.conf`.
* Do not place this property into `default.conf`, but have the value in Kotlin code.
* Introduce a placeholder, which is evaluated when reading the configuration. (This is a bit messy because there is already a mechanism to resolve environment variables; unfortunately, there is no portable environment variable for the number of processors we could use.)
* ?

@sschuberth @fviernau @mnonnenmacher, what are your preferences?

fviernau:
[...]

PR #5406, opened 2022-05-31
ScannerConfiguration: Add option to provide detected license mappings

MarcelBochtler:
Scanners might return invalid SPDX license expressions. This causes
issues when reading the scanner results. Provide an option that users
can map detected licenses in the scanner configuration.

This functionality is not limited to invalid SPDX expressions. The user
can also map unwanted licenses, like `LicenseRef-scancode-unknown` to
`NOASSERTION`.

The mapping is being done when creating the LicenseFinding after
receiving the results from the scanners. This has the advantage that
internally, ORT handles only valid SPDX expressions. The disadvantage
of that approach is, that mapped licenses are stored in the
ScanResultStorage which might become invalid when the detected license
mapping in the scanner configuration is changed.

Resolves https://github.com/oss-review-toolkit/ort/issues/5345.

This PR removes some hard coded license mappings to make them user configurable, this is the configuration to reproduce the previous behavior:
```
ort {
  scanner {
    detectedLicenseMapping {
      # https://scancode-licensedb.aboutcode.org/?search=generic
      "LicenseRef-scancode-agpl-generic-additional-terms" = "NOASSERTION"
      "LicenseRef-scancode-generic-cla" = "NOASSERTION"
      "LicenseRef-scancode-generic-exception" = "NOASSERTION"
      "LicenseRef-scancode-generic-export-compliance" = "NOASSERTION"
      "LicenseRef-scancode-generic-tos" = "NOASSERTION"
      "LicenseRef-scancode-generic-trademark" = "NOASSERTION"
      "LicenseRef-scancode-gpl-generic-additional-terms" = "NOASSERTION"
      "LicenseRef-scancode-patent-disclaimer" = "NOASSERTION"
      "LicenseRef-scancode-warranty-disclaimer" = "NOASSERTION"

      # https://scancode-licensedb.aboutcode.org/?search=other
      "LicenseRef-scancode-other-copyleft" = "NOASSERTION"
      "LicenseRef-scancode-other-permissive" = "NOASSERTION"

      # https://scancode-licensedb.aboutcode.org/?search=unknown
      "LicenseRef-scancode-free-unknown" = "NOASSERTION"
      "LicenseRef-scancode-unknown" = "NOASSERTION"
      "LicenseRef-scancode-unknown-license-reference" = "NOASSERTION"
      "LicenseRef-scancode-unknown-spdx" = "NOASSERTION"

      # FossID
      "BSD (Three Clause License)" = "BSD-3-clause"
      "BSD-Acknowledgment-(Carrot2)" = "LicenseRef-scancode-bsd-ack-carrot2"
      "Freeware-Public-(FPL)" = "LicenseRef-scancode-fpl"
      "MediaInfo(Lib)" = "LicenseRef-scancode-mediainfo-lib"
      "Oracle-Master-Agreement-(OMA)" = "LicenseRef-scancode-oracle-master-agreement"
      "Things-I-Made-(TIM)-Public" = "LicenseRef-scancode-things-i-made-public-license"
      "X11-Style-(Adobe)" = "LicenseRef-scancode-x11-adobe"
      "X11-Style-(Adobe-DEC)" = "LicenseRef-scancode-x11-adobe-dec"
      "X11-Style-(Bitstream-Charter)" = "LicenseRef-scancode-x11-bitstream"
      "X11-Style-(DEC-1)" = "LicenseRef-scancode-x11-dec1"
      "X11-Style-(DEC-2)" = "LicenseRef-scancode-x11-dec2"
      "X11-Style-(DSC-Technologies)" = "LicenseRef-scancode-x11-dsc"
      "X11-Style-(David-R.-Hanson)" = "LicenseRef-scancode-x11-hanson"
      "X11-Style-(Keith-Packard)" = "HPND-sell-variant"
      "X11-Style-(Lucent)" = "LicenseRef-scancode-x11-lucent"
      "X11-Style-(OAR)" = "LicenseRef-scancode-x11-oar"
      "X11-Style-(Open-Group)" = "MIT-open-group"
      "X11-Style-(Quarterdeck)" = "LicenseRef-scancode-x11-quarterdeck"
      "X11-Style-(Realmode)" = "LicenseRef-scancode-x11-realmode"
      "X11-Style-(Silicon-Graphics)" = "LicenseRef-scancode-x11-sg"
      "X11-Style-(Tektronix)" = "LicenseRef-scancode-x11-tektronix"
      "u_Agere-BSD" = "LicenseRef-scancode-agere-bsd"
      "u_BSD-3-Clause-Sun" = "LicenseRef-scancode-bsd-3-clause-sun"
      "u_CC-BY-2.0-UK" = "LicenseRef-scancode-cc-by-2.0-uk"
      "u_GLUT-License" = "LicenseRef-scancode-glut"
      "u_HACOS-1.2" = "LicenseRef-scancode-hacos-1.2"
      "u_Jscheme-License" = "LicenseRef-scancode-jscheme"
      "u_MongoDB-SSPL-1.0" = "SSPL-1.0"
      "u_Philips-Proprietary-Notice-2000" = "LicenseRef-scancode-philips-proprietary-notice-2000"
      "u_Sun-BCL-11-plus-6" = "LicenseRef-scancode-sun-bcl-11-06"
      "u_Sun-Communications-API" = "LicenseRef-scancode-sun-communications-api"
      "u_Sun-EJB-Specification-3.0" = "LicenseRef-scancode-sun-ejb-spec-3.0"
      "u_Sun-Entitlement-3-plus-15" = "LicenseRef-scancode-sun-entitlement-03-15"
      "u_Sun-Entitlement-JAF" = "LicenseRef-scancode-sun-entitlement-jaf"
      "u_Sun-Java-Web-Services-Developer-Pack-1.6" = "Xerox"
      "u_Sun-JavaMail" = "LicenseRef-scancode-sun-javamail"
      "u_Sun-Project-X" = "LicenseRef-scancode-sun-project-x"
      "u_Taligent-JDK-Proprietary-Notice" = "LicenseRef-scancode-taligent-jdk"
      "u_Trolltech-GPL-Exception-1.2" = "LicenseRef-scancode-trolltech-gpl-exception-1.2"
      "u_Vuforia-2013-07-29" = "LicenseRef-scancode-vuforia-2013-07-29"
      "u_Xerox" = "Xerox"
      "u_nexB-EULA-for-SaaS-1.1.0" = "LicenseRef-scancode-nexb-eula-saas-1.1.0"
      "u_nexB-SSLA-1.1.0" = "LicenseRef-scancode-nexb-ssla-1.1.0"
    }
  }
}
```

### DIFF
--- a/model/src/main/kotlin/LicenseFinding.kt
+++ b/model/src/main/kotlin/LicenseFinding.kt
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.annotation.JsonInclude
 
 import org.ossreviewtoolkit.model.config.LicenseFindingCuration
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
+import org.ossreviewtoolkit.utils.spdx.isSpdxExpression
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 
 /**
@@ -51,9 +52,41 @@ data class LicenseFinding(
 ) : Comparable<LicenseFinding> {
     companion object {
         private val COMPARATOR = compareBy<LicenseFinding>({ it.license.toString() }, { it.location })
+
+        /**
+         * Create a [LicenseFinding] with [detectedLicenseMapping]s applied.
+         */
+        fun createAndMap(
+            license: String,
+            location: TextLocation,
+            score: Float? = null,
+            detectedLicenseMapping: Map<String, String>
+        ): LicenseFinding = LicenseFinding(
+            license = if (license.isSpdxExpression() || detectedLicenseMapping.isEmpty()) {
+                license
+            } else {
+                license.applyDetectedLicenseMapping(detectedLicenseMapping)
+            }.toSpdx(),
+            location = location,
+            score = score
+        )
     }
 
     constructor(license: String, location: TextLocation, score: Float? = null) : this(license.toSpdx(), location, score)
 
     override fun compareTo(other: LicenseFinding) = COMPARATOR.compare(this, other)
+}
+
+/**
+ * Apply [detectedLicenseMapping] from the [org.ossreviewtoolkit.model.config.ScannerConfiguration] to any license
+ * String.
+ */
+private fun String.applyDetectedLicenseMapping(detectedLicenseMapping: Map<String, String>): String {
+    var result = this
+    detectedLicenseMapping.forEach { (from, to) ->
+        // Replace a license restricted by whitespace boundaries.
+        result = result.replace("""(?<!\S)${Regex.escape(from)}""".toRegex(), to)
+    }
+
+    return result
 }
--- a/model/src/main/kotlin/config/ScannerConfiguration.kt
+++ b/model/src/main/kotlin/config/ScannerConfiguration.kt
@@ -26,6 +26,7 @@ import com.fasterxml.jackson.annotation.JsonInclude
 
 import org.ossreviewtoolkit.model.utils.FileArchiver
 import org.ossreviewtoolkit.utils.ort.storage.FileStorage
+import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 
 typealias ScannerOptions = Map<String, String>
 
@@ -50,6 +51,33 @@ data class ScannerConfiguration(
      * Create archives for packages that have a stored scan result but no license archive yet.
      */
     val createMissingArchives: Boolean = false,
+
+    /**
+     * Mappings from licenses returned by the scanner to valid SPDX licenses. Note that these mappings are only applied
+     * in new scans, stored scan results are not affected.
+     */
+    val detectedLicenseMapping: Map<String, String> = mapOf(
+        // https://scancode-licensedb.aboutcode.org/?search=generic
+        "LicenseRef-scancode-agpl-generic-additional-terms" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-generic-cla" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-generic-exception" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-generic-export-compliance" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-generic-tos" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-generic-trademark" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-gpl-generic-additional-terms" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-patent-disclaimer" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-warranty-disclaimer" to SpdxConstants.NOASSERTION,
+
+        // https://scancode-licensedb.aboutcode.org/?search=other
+        "LicenseRef-scancode-other-copyleft" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-other-permissive" to SpdxConstants.NOASSERTION,
+
+        // https://scancode-licensedb.aboutcode.org/?search=unknown
+        "LicenseRef-scancode-free-unknown" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-unknown" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-unknown-license-reference" to SpdxConstants.NOASSERTION,
+        "LicenseRef-scancode-unknown-spdx" to SpdxConstants.NOASSERTION
+    ),
 
     /**
      * Scanner specific configuration options. The key needs to match the name of the scanner class, e.g. "ScanCode"
--- a/scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
+++ b/scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
@@ -106,6 +106,7 @@ internal fun generateSummary(
     endTime: Instant,
     scanPath: File,
     result: JsonNode,
+    detectedLicenseMapping: Map<String, String> = emptyMap(),
     parseExpressions: Boolean = true
 ) =
     generateSummary(
@@ -113,6 +114,7 @@ internal fun generateSummary(
         endTime,
         calculatePackageVerificationCode(scanPath),
         result,
+        detectedLicenseMapping,
         parseExpressions
     )
 
@@ -126,13 +128,14 @@ internal fun generateSummary(
     endTime: Instant,
     verificationCode: String,
     result: JsonNode,
+    detectedLicenseMapping: Map<String, String> = emptyMap(),
     parseExpressions: Boolean = true
 ) =
     ScanSummary(
         startTime = startTime,
         endTime = endTime,
         packageVerificationCode = verificationCode,
-        licenseFindings = getLicenseFindings(result, parseExpressions).toSortedSet(),
+        licenseFindings = getLicenseFindings(result, detectedLicenseMapping, parseExpressions).toSortedSet(),
         copyrightFindings = getCopyrightFindings(result).toSortedSet(),
         issues = getIssues(result)
     )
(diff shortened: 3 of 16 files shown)


model/src/main/kotlin/LicenseFinding.kt
```diff
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.annotation.JsonInclude
 
 import org.ossreviewtoolkit.model.config.LicenseFindingCuration
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
+import org.ossreviewtoolkit.utils.spdx.isSpdxExpression
```
sschuberth, line 26: Commit message nits:
- "Add *an* option ..."
- "Scanners might return licenses that cannot be parsed as SPDX license expressions."
- "Provide a configuration option for user to map ..."

> This functionality is not limited to invalid SPDX expressions. The user
can also map unwanted licenses, like `LicenseRef-scancode-unknown` to
`NOASSERTION`.

I believe this paragraph is misleading, as it sounds as if mapping to `NOASSERTION` would get rid of the detected license (similar to concluding a declared license to `NONE`), which is not the case.

- "... ORT has to handle only valid SPDX expressions internally."

> The disadvantage of this approach is, that mapped licenses are stored in the
ScanResultStorage

Hmm, wasn't one point of doing this to *not* store the mapped licenses, so no rescan is required when the mapping changes?

fviernau:
The PR description illustrates a mapping which reproduces the previous behavior.
The `keys` for scancode license all have `LicenseRef-scancode-` as prefix while the FOSS-ID ones lack such prefix.
Is this intended? / should prefixes be added to FOSS-ID licenses?

model/src/main/kotlin/LicenseFinding.kt
MarcelBochtler: > > The disadvantage of this approach is, that mapped licenses are stored in the
> > ScanResultStorage
> 
> Hmm, wasn't one point of doing this to _not_ store the mapped licenses, so no rescan is required when the mapping changes?

This was my / our initial plan (See: https://github.com/oss-review-toolkit/ort/issues/5345#issuecomment-1125130550). However, implementing this would require to basically remove the `.toSpdx()` calls everywhere before storing anything in the ScanStorage and only handle Strings within ORT.
E.g. `BSD (3-clause)` is not a valid SPDX expression and even uses characters which are reserved in SPDX: `(` and `)` for indicating order of operations in license choices.

That's why I chose to implement it differently.

MarcelBochtler:
> The PR description illustrates a mapping which reproduces the previous behavior.
> The `keys` for scancode license all have `LicenseRef-scancode-` as prefix while the FOSS-ID ones lack such prefix.
> Is this intended? / should prefixes be added to FOSS-ID licenses?

Yes, this is intended, as the hard-coded mapping before also used the `LicenseRef-scancode` prefix.
With this change, users are free to use any valid prefix for FossID LicenseRefs.

model/src/main/kotlin/config/ScannerConfiguration.kt
```diff
@@ -51,6 +51,11 @@ data class ScannerConfiguration(
      */
     val createMissingArchives: Boolean = false,
 
+    /**
+     * Mappings from licenses returned by the scanner to valid SPDX licenses.
```
mnonnenmacher, line 55: You could add here the information from the discussion above, for example: "Note that these mappings are only applied in new scans, stored scan results are not affected."

MarcelBochtler: Added the note to the JavaDoc and additionally to the comment in the `reference.conf`.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -71,31 +70,6 @@ private val TIMEOUT_ERROR_REGEX = Pattern.compile(
...
-    "LicenseRef-scancode-other-copyleft",
-    "LicenseRef-scancode-other-permissive",
-
-    // https://scancode-licensedb.aboutcode.org/?search=unknown
-    "LicenseRef-scancode-free-unknown",
-    "LicenseRef-scancode-unknown",
-    "LicenseRef-scancode-unknown-license-reference",
-    "LicenseRef-scancode-unknown-spdx"
-)
-
```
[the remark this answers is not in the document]
MarcelBochtler: I added the previous ScanCode license mappings as defaults, to avoid a breaking change. For FossID this doesn't really make sense, as the licenses are different for every FossID instance.

I think that @sschuberth's proposal makes sense, as it would be more transparent for the user which defaults are being used.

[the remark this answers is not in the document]
mnonnenmacher: @sschuberth I'm glad you reconsider this after #3522 / #3731 :) I would still prefer a `default.conf` over hardcoded values in Kotlin, because I think it's beneficial to the user, but I wouldn't agree that it should replace the `reference.conf`. The purpose of `default.conf` is to provide a meaningful default configuration, while the purpose of `reference.conf` is to showcase all possible configuration options.

mnonnenmacher (approved):

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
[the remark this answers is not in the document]
sschuberth: > The purpose of `default.conf` is to provide a meaningful default configuration, while the purpose of `reference.conf` is to showcase all possible configuration options.

Right, but IMO with `default.conf` there should be no implicit default anymore. So if everything is explicit, `default.conf` basically *is* a `reference.conf`. The only question would be about examples for credentials etc. that should not go into a `default.conf`. Maybe these could be added as comments into `default.conf`, and the existing `ort config --show-default` could simply display them?

model/src/main/kotlin/LicenseFinding.kt
```diff
@@ -51,9 +52,41 @@ data class LicenseFinding(
...
+ * String.
+ */
+private fun String.applyDetectedLicenseMapping(detectedLicenseMapping: Map<String, String>): String {
+    var result = this
+    detectedLicenseMapping.forEach { (from, to) ->
+        // Replace a license restricted by whitespace boundaries.
+        result = result.replace("""(?<!\S)${Regex.escape(from)}""".toRegex(), to)
+    }
+
+    return result
```
sschuberth, line 91: Nit: Looks like this could be simplified via `apply {}`.

MarcelBochtler: Not sure how to simplify this with `apply {}`. Can you give me a hint what you have in mind?

sschuberth: It doesn't work as I thought it would, please disregard.

sschuberth: But I have another idea 😁 Maybe use `fold` similar to

https://github.com/oss-review-toolkit/ort/blob/787b903c760deba18ae311c3c17a3e132d1166f9/scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt#L264-L271

MarcelBochtler: IMO, this is less readable. Is it ok, if I keep it as-is?

MarcelBochtler: `fold` is not available for a `Map`.

MarcelBochtler: Is it ok if I keep it as-is?

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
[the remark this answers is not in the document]
mnonnenmacher: It's more than just credentials, for example the `default.conf` would likely configure a local file based storage for scan results, but we also need to demonstrate how the other storage types can be configured. Of course this could also be done with comments in `default.conf`, but then our current testing approach wouldn't work. Let's discuss the details once anyone wants to pick this topic up.

model/src/main/kotlin/LicenseFinding.kt
sschuberth: Sure, I'm just commenting from the off 😉 Your code just remembered me of that code, which mostly does the same thing.

mnonnenmacher (approved):

PR #5408, opened 2022-06-02
Add helper functions to request password authentication

oheger-bosch:
Refer to the single commits for further information.

### DIFF
--- a/utils/ort/src/test/kotlin/UtilsTest.kt
+++ b/utils/ort/src/test/kotlin/UtilsTest.kt
@@ -27,6 +27,13 @@ import io.kotest.matchers.collections.shouldHaveSingleElement
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
 
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.mockkStatic
+import io.mockk.verify
+
+import java.net.Authenticator
+import java.net.PasswordAuthentication
 import java.nio.file.Paths
 
 class UtilsTest : WordSpec({
@@ -377,6 +384,28 @@ class UtilsTest : WordSpec({
             val url = "https://github.com/oss-review-toolkit/ort-test-data-git-repo.git?manifest=manifest.xml"
 
             normalizeVcsUrl(url) shouldBe url
+        }
+    }
+
+    "requestPasswordAuthentication" should {
+        "return a correct authentication object for a host, port, and scheme combination" {
+            val host = "www.example.org"
+            val port = 442
+            val scheme = "https"
+
+            mockkStatic("org.ossreviewtoolkit.utils.ort.UtilsKt")
+            mockkStatic(Authenticator::class)
+            val passwordAuth = mockk<PasswordAuthentication>()
+
+            every {
+                Authenticator.requestPasswordAuthentication(host, null, port, scheme, null, null)
+            } returns passwordAuth
+
+            requestPasswordAuthentication(host, port, scheme) shouldBe passwordAuth
+
+            verify {
+                installAuthenticatorAndProxySelector()
+            }
         }
     }
 })
--- a/analyzer/src/main/kotlin/managers/Conan.kt
+++ b/analyzer/src/main/kotlin/managers/Conan.kt
@@ -28,7 +28,6 @@ import com.fasterxml.jackson.module.kotlin.readValue
 import com.vdurmont.semver4j.Requirement
 
 import java.io.File
-import java.net.Authenticator
 import java.util.SortedSet
 
 import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
@@ -57,6 +56,7 @@ import org.ossreviewtoolkit.utils.common.textValueOrEmpty
 import org.ossreviewtoolkit.utils.common.toUri
 import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 import org.ossreviewtoolkit.utils.ort.log
+import org.ossreviewtoolkit.utils.ort.requestPasswordAuthentication
 
 /**
  * The [Conan](https://conan.io/) package manager for C / C++.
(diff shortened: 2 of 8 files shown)


oheger-bosch:
Not sure about the failing `StackFunTest`. I don't think, this is related to the changes introduced here.

sschuberth:
> Not sure about the failing `StackFunTest`. I don't think, this is related to the changes introduced here.

I've retriggered only the failing tests, let's see.

oheger-bosch:
The failures are reproducible. The error message says:

```
message: "Resolving Stack dependencies for path 'src/funTest/assets/projects/external/quickcheck-state-machine/stack.yaml'\
        \ failed with: IOException: Running 'stack dot --global-hints --external' in '/home/ort/src/analyzer/src/funTest/assets/projects/external/quickcheck-state-machine'\
        \ failed with exit code 1:\nCabal file info not found for base-orphans-0.8@sha256:4885caa82ab94fa5cb841d6d06a43a97cdc7e835f346be4f11c1cc5707b886db,\
        \ updating\nSelected mirror [https://s3.amazonaws.com/hackage.fpcomplete.com/\n\](https://s3.amazonaws.com/hackage.fpcomplete.com//n/)
        Downloading root\nSelected mirror [https://s3.amazonaws.com/hackage.fpcomplete.com/\n\](https://s3.amazonaws.com/hackage.fpcomplete.com//n/)
        Downloading timestamp\nDownloading snapshot\nDownloading mirrors\nVerification\
        \ error: <repo>/mirrors.json is expired\nDownloading root\nDownloading timestamp\n\
        Downloading snapshot\nDownloading mirrors\nVerification error: <repo>/mirrors.json\
        \ is expired\nDownloading root\nDownloading timestamp\nDownloading snapshot\n\
        Downloading mirrors\nVerification error: <repo>/mirrors.json is expired\nDownloading\
        \ root\nDownloading timestamp\nDownloading snapshot\nDownloading mirrors\nVerification\
        \ error: <repo>/mirrors.json is expired\nDownloading root\nDownloading timestamp\n\
        Downloading snapshot\nDownloading mirrors\nVerification error: <repo>/mirrors.json\
        \ is expired\nDownloading root\nVerification loop. Errors in order:\n  <repo>/mirrors.json\
        \ is expired\n  <repo>/mirrors.json is expired\n  <repo>/mirrors.json is expired\n\
        \  <repo>/mirrors.json is expired\n  <repo>/mirrors.json is expired\n\n"
      severity: "ERROR"
    "
```

utils/ort/src/test/kotlin/UtilsTest.kt
```diff
@@ -379,4 +387,51 @@ class UtilsTest : WordSpec({
...
+
+            mockkStatic("org.ossreviewtoolkit.utils.ort.UtilsKt")
+            mockkStatic(Authenticator::class)
+            val proxySelector = mockk<OrtProxySelector>()
+            val passwordAuth = mockk<PasswordAuthentication>()
+
+            every { installAuthenticatorAndProxySelector() } returns proxySelector
+            every {
+                Authenticator.requestPasswordAuthentication(host, null, port, scheme, null, null)
+            } returns passwordAuth
```
sschuberth, line 405: Hmm. I'm having difficulties appreciating the value of these tests. You're basically mocking everything that's being called in `requestPasswordAuthentication()` away, effectively reimplementing the helper function with mocks only, leaving no original code to test.

Wouldn't it make more sense to simply "set" environment variables via `Os.env` and reply on `EnvVarAuthenticator` to do its job?

sschuberth (changes requested):

utils/ort/src/test/kotlin/UtilsTest.kt
sschuberth: > not more and not less.

I believe the test *does* more, in the sense that it mocks unnecessary things. For example, I believe lines 399 and 402 could simply be dropped without sacrificing anything.

utils/ort/src/test/kotlin/UtilsTest.kt
```diff
@@ -379,4 +387,51 @@ class UtilsTest : WordSpec({
...
+            verify {
+                installAuthenticatorAndProxySelector()
+            }
+        }
+
+        "return a correct authentication object for a URI" {
+            val host = "www.example.org"
+            val port = 442
+            val scheme = "https"
+            val uri = URI(scheme, "$host:$port", "/some/path", "someQuery", "someFragment")
```
sschuberth, line 418: IMO it would read nicer / fit better to the title of test if this URI was constructed from a string, and line 427 would refer to the URI components like `uri.host` etc.

oheger-bosch: If it helps to get the change in, I am fine with removing these tests. Just want to mention that this unremarkable two-liner function actually has some impact. The implicit assumption that the `OrtAuthenticator` was available caused us a lot of trouble recently.

sschuberth (approved):

sschuberth:
BTW, another possible solution that just came to my mind: Couldn't we simply assign the return value of `installAuthenticatorAndProxySelector()` to a global variable to ensure it always gets called at application startup?

oheger-bosch:
> BTW, another possible solution that just came to my mind: Couldn't we simply assign the return value of `installAuthenticatorAndProxySelector()` to a global variable to ensure it always gets called at application startup?

What I like at the current solution is that it actually simplifies the usage of the `Authenticator` API - one does not have to deal with the unneeded parameters.

I had been thinking about making `installAuthenticatorAndProxySelector()` an internal function, so that it cannot be called from other parts of the codebase. However, currently the return value is used on one call side. But maybe this could be fixed by managing a shared instance of `OrtProxySelector` in the companion object?

sschuberth:
> What I like at the current solution is that it actually simplifies the usage of the `Authenticator` API - one does not have to deal with the unneeded parameters.

Yes, that's indeed nice.

> I had been thinking about making `installAuthenticatorAndProxySelector()` an internal function, so that it cannot be called from other parts of the codebase.

That basically sounds like a good idea, but the function is currently also used in `MavenSupport` and `OkHttpClientHelper`; if it was `internal`, you'd need to find a way to get rid of those calls, too.

> However, currently the return value is used on one call side.

Yeah, it has always bothered me a bit that `installAuthenticatorAndProxySelector()` somewhat arbitrarily returns the proxy selector... users could ask why it doesn't return the authenticator instead, or both.

Actually, now that we have your helper function, I'd probably just inline `installAuthenticatorAndProxySelector()` at the callers to get rid of that helper function completely. Then we wouldn't need to bother making it `internal`, or care about its return value. What do you think?

oheger-bosch:
> 
> Actually, now that we have your helper function, I'd probably just inline `installAuthenticatorAndProxySelector()` at the callers to get rid of that helper function completely. Then we wouldn't need to bother making it `internal`, or care about its return value. What do you think?

Yes, that's fine with me.

sschuberth:
> Yes, that's fine with me.

I'll create a PR then.

sschuberth:
> I'll create a PR then.

See https://github.com/oss-review-toolkit/ort/pull/5413.

PR #5409, opened 2022-06-02
Add ADOPTERS.md

tsteenbe:
Inspired by [1].

[1]: https://github.com/backstage/backstage/blob/e487242d70e52aa7b6a48266a09caebbcc433080/ADOPTERS.md


ADOPTERS.md
```diff
@@ -0,0 +1,9 @@
+# Adopters
+
+_If you're using ORT in your organization, please add your organization's name to the table below. This really helps the project to gain momentum and credibility. It's a small contribution back to the project with a big impact._
```
fviernau, line 3: Would it make sense to say: please add your organization's name to the **bottom of** the table below? So that the ordering is defined?

fviernau (dismissed):

ADOPTERS.md
mnonnenmacher: I think I would prefer alphabetical sorting, to not imply that the order has any meaning. Also I don't like that the list is not sorted at all in the linked example.

fviernau: Agree with @mnonnenmacher. So, @tsteenbe let's mention the ordering in the comment then?!

tsteenbe: @mnonnenmacher  @fviernau Changed the comment to say "to the alphabetically sorted table below."

fviernau (approved):

PR #5412, opened 2022-06-02
ExperimentalScanner: Consolidate packages for package scanners

mnonnenmacher:
When scanning with package scanners, consolidate packages by their provenance, to scan the same provenance only once. See the commit messages for details.

### DIFF
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2021 HERE Europe B.V.
- * Copyright (C) 2021 Bosch.IO GmbH
+ * Copyright (C) 2021-2022 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -35,6 +35,7 @@ import org.ossreviewtoolkit.model.OrtIssue
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageType
+import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanRecord
 import org.ossreviewtoolkit.model.ScanResult
 import org.ossreviewtoolkit.model.ScanSummary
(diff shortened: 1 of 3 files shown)


scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -180,6 +180,20 @@ class ScanController(
...
+     */
+    fun getPackagesConsolidatedByProvenance(): Map<KnownProvenance, List<Package>> =
+        packages.mapNotNull { pkg ->
+            packageProvenances[pkg.id]?.let { pkg to it }
+        }.groupBy { (_, provenance) ->
+            when (provenance) {
+                is RepositoryProvenance -> provenance.copy(vcsInfo = provenance.vcsInfo.copy(path = ""))
+                else -> provenance
+            }
+        }.mapValues { it.value.map { (pkg, _) -> pkg } }
```
sschuberth, line 195: Alternatively (using only a single loop):

    fun getPackagesConsolidatedByProvenance(): Map<KnownProvenance, List<Package>> {
        val packagesByProvenance = mutableMapOf<KnownProvenance, MutableList<Package>>()

        packages.forEach { pkg ->
            val consolidatedProvenance = when (val provenance = packageProvenances[pkg.id]) {
                null -> return@forEach
                is RepositoryProvenance -> provenance.copy(vcsInfo = provenance.vcsInfo.copy(path = ""))
                else -> provenance
            }

            packagesByProvenance.getOrPut(consolidatedProvenance) { mutableListOf() } += pkg
        }

        return packagesByProvenance
    }

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -213,33 +214,69 @@ class ExperimentalScanner(
...
-            val nestedProvenance = controller.getNestedProvenance(pkg.id) ?: return@forEach
-
-            // TODO: Use coroutines to execute scanners in parallel.
+        controller.getPackagesConsolidatedByProvenance().forEach { (provenance, packages) ->
             controller.getPackageScanners().forEach scanner@{ scanner ->
-                if (controller.hasCompleteScanResult(scanner, pkg)) {
-                    log.debug {
-                        "Skipping scan of '${pkg.id.toCoordinates()}' with package scanner '${scanner.name}' as " +
-                                "stored results are available."
+                val packagesWithIncompleteScanResult = packages
```
sschuberth, line 219: How about filtering only once like:

                val packagesWithIncompleteScanResult = packages.filter { pkg ->
                    val hasNestedProvenance = controller.getNestedProvenance(pkg.id) != null
                    if (!hasNestedProvenance) {
                        log.debug {
                            "Skipping scan of '${pkg.id.toCoordinates()}' with package scanner ${scanner.name} as no " +
                                    "nested provenance for the package could be resolved."
                        }
                    }

                    val hasCompleteScanResult = controller.hasCompleteScanResult(scanner, pkg)
                    if (hasCompleteScanResult) {
                        log.debug {
                            "Skipping scan of '${pkg.id.toCoordinates()}' with package scanner '${scanner.name}' as " +
                                    "stored results are available."
                        }
                    }

                    hasNestedProvenance && !hasCompleteScanResult
                }

Note that the above code also improves the log messages slightly.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -213,33 +214,69 @@ class ExperimentalScanner(
...
+                                            "'${scanner.name}' as stored results are available."
+                                }
+                            }
+                        }
                     }
 
-                    return@scanner
+                if (packagesWithIncompleteScanResult.isEmpty()) return@scanner
+
+                // Create a reference package with any VCS path removed, to ensure the full repository is scanned.
```
sschuberth, line 244: Is this block still necessary, as `getPackagesConsolidatedByProvenance()` should already do that, or?

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -213,33 +214,69 @@ class ExperimentalScanner(
...
+                val scanResult = scanner.scanPackage(referencePackage, context)
 
-                val nestedProvenanceScanResult = scanResult.toNestedProvenanceScanResult(nestedProvenance)
+                log.info {
+                    "Scan of '${referencePackage.id.toCoordinates()}' with package scanner '${scanner.name}' finished."
+                }
 
-                controller.addNestedScanResult(scanner, nestedProvenanceScanResult)
+                packagesWithIncompleteScanResult.forEach processResults@{ pkg ->
+                    val nestedProvenance = controller.getNestedProvenance(pkg.id) ?: return@processResults
```
sschuberth, line 273: Would it make sense to try reusing the already determined nested provenance from above?

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -213,33 +214,69 @@ class ExperimentalScanner(
...
+                    val nestedProvenanceScanResult = scanResult.toNestedProvenanceScanResult(nestedProvenance)
+                    controller.addNestedScanResult(scanner, nestedProvenanceScanResult)
 
-                // TODO: Run in coroutine.
-                if (scanner.criteria != null) {
-                    storeNestedScanResult(pkg, nestedProvenanceScanResult)
+                    if (scanner.criteria != null) {
```
sschuberth, line 277: Is the TODO to run in a coroutine obsolete now?

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: The function does only group the packages, it does not create a reference package.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: I didn't want to create another map of packages to nested provenances in this function as the controller has that map already.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
mnonnenmacher: Recovered.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
sschuberth: Oh right, the function removes the VCS path from the provenance, not from the package.

mnonnenmacher: Yes, the packages need to keep the path to correctly filter the scan results later on.

sschuberth (approved):

PR #5414, opened 2022-06-03
Dockerfile: Avoid copying the distribution TAR by mounting it instead

sschuberth:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -155,7 +155,7 @@ RUN /opt/ort/bin/import_proxy_certs.sh && \
 ARG SCANCODE_VERSION
 RUN pip install --no-cache-dir scancode-toolkit==$SCANCODE_VERSION
 
-FROM run
+FROM run AS dist
 
 COPY --from=build /usr/local/src/ort/cli/build/distributions/ort-*.tar /opt/ort.tar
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -158,10 +158,8 @@ RUN pip install --no-cache-dir scancode-toolkit==$SCANCODE_VERSION
 FROM run AS dist
 
 ARG ORT_VERSION
-COPY --from=build /usr/local/src/ort/cli/build/distributions/ort-$ORT_VERSION.tar /opt/ort.tar
-
-RUN tar xf /opt/ort.tar -C /opt/ort --exclude="*.bat" --strip-components 1 && \
-    rm /opt/ort.tar && \
+RUN --mount=type=bind,from=build,source=/usr/local/src/ort/cli/build/distributions/ort-$ORT_VERSION.tar,target=/opt/ort.tar \
+    tar xf /opt/ort.tar -C /opt/ort --exclude="*.bat" --strip-components 1 && \
```
mnonnenmacher, line 162: Does it not have to be unmounted after extracting it?

sschuberth: No, the mount is only valid during the `RUN` step, and I've verified that the mounted `ort.tar` does not end up in the Docker image.

mnonnenmacher (approved):

PR #5424, opened 2022-06-09

fviernau:


MarcelBochtler (approved):

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -155,7 +155,7 @@ class NuGetSupport(serviceIndexUrls: List<String> = listOf(DEFAULT_SERVICE_INDEX
         // Note: The package name in the URL is case-sensitive and must be lower-case!
         val lowerId = id.name.lowercase()
 
-        val data = registrationsBaseUrls.asSequence().firstNotNullOfOrNull { baseUrl ->
+        val data = registrationsBaseUrls.firstNotNullOfOrNull { baseUrl ->
```
sschuberth, line 158: BTW, does any of you understand *why* `asSequence()` is redundant here? In theory, `firstNotNullOfOrNull` could exit early without evaluating all entries of the collection, so using a sequence could make sense. The only reason I can imagine is that in this particular case `registrationsBaseUrls` is a list of strings, and for plain strings objects there is no (expensive) evaluation happening anyway, so it probably also does not matter whether the collection is lazily evaluated via a sequence or not.

sschuberth (approved):

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: I believe that [1] is used in both cases, so if there is a matching item then it returns early after ´k´  iterations if the first matching item is at position ´k´.

[1] https://github.com/JetBrains/kotlin/blob/d8b0dbe71bbeb84ba0fbdac87801c8c6f4eed4c9/libraries/stdlib/common/src/generated/_Collections.kt#L247-L255

sschuberth: No, the original code actually called `public inline fun <T, R : Any> Sequence<T>.firstNotNullOfOrNull(transform: (T) -> R?): R?`. But as `registrationsBaseUrls` is a *list* of strings, turning it into a sequence afterwards does make make its already known elements evaluate lazily, so there's really no point in using a sequence here. It would have made sense if there was e.g. some kind of mapping in between.

PR #5427, opened 2022-06-10
Analyzer: Log the repository configuration as YAML

mnonnenmacher:
Log the repository configuration as YAML instead of using the generated
`toString()` function of the data class. This improves readability and
also equals the format that is used in the .ort.yml file.


mnonnenmacher:
Merged as the test failure in `DotNetFunTest` is unrelated to the change.

PR #5429, opened 2022-06-11
PostresStorage: Make the number of storage threads configurable

MarcelBochtler:
The database pool size is implicitly configured by the number of storage
threads. Before, this value was hard-coded to 5 threads. As this might
not fit the user's database setup, make the value
configurable.

### DIFF
--- a/model/src/main/kotlin/config/ScanStorageConfiguration.kt
+++ b/model/src/main/kotlin/config/ScanStorageConfiguration.kt
@@ -119,6 +119,11 @@ data class PostgresStorageConfiguration(
     val sslrootcert: String? = null,
 
     /**
+     * The number of parallel transactions to use for the storage dispatcher.
+     */
+    val parallelTransactions: Int = 5,
+
+    /**
      * The way that scan results are stored, defaults to [StorageType.PACKAGE_BASED].
      */
     val type: StorageType = StorageType.PACKAGE_BASED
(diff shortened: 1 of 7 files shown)


scanner/src/main/kotlin/storages/PostgresStorage.kt
```diff
@@ -66,7 +65,8 @@ class PostgresStorage(
     /**
      * The JDBC data source to obtain database connections.
      */
-    private val dataSource: Lazy<DataSource>
+    private val dataSource: Lazy<DataSource>,
+    private val storageThreads: Int
```
sschuberth, line 69: Nit: Please at least add a blank like here to make clear that the above docs only apply to line 68, or ideally also add brief docs for line 69.

sschuberth (changes requested):

sschuberth (dismissed):

model/src/main/kotlin/config/ScanStorageConfiguration.kt
```diff
@@ -118,6 +118,11 @@ data class PostgresStorageConfiguration(
      */
     val sslrootcert: String? = null,
 
+    /**
+     * The number of threads to use for the storage dispatcher.
+     */
+    val storageThreads: Int = 5,
```
oheger-bosch, line 124: The name "storage threads" was used before, but it is actually incorrect; coroutines are not directly mapped to threads, and the property rather defines the number of transactions spawned in parallel. So, while at it, should it be renamed to something like `parallelTransactions` ?

oheger-bosch (approved):

oheger-bosch (approved):

PR #5437, opened 2022-06-13
model: Add a global typealias for options

mnonnenmacher:
Replace the different typealiases for `Map<String, String>` used in
several configuration classes with one global typealias for options.

### DIFF
--- a/advisor/src/main/kotlin/AdviceProviderFactory.kt
+++ b/advisor/src/main/kotlin/AdviceProviderFactory.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020-2021 Bosch.IO GmbH
+ * Copyright (C) 2020-2022 Bosch.IO GmbH
  *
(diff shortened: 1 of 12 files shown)


model/src/main/kotlin/config/Options.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model.config
+
+typealias Options = Map<String, String>
```
sschuberth, line 22: Nit: I'm not 100% sure whether it's worth to introduce a separate file for this. How about just adding the `typealias` to `OrtConfiguration.kt`?

mnonnenmacher: Fine with me.

PR #5438, opened 2022-06-13
model: Allow to enable and disable package managers in the configuration

mnonnenmacher:
Allow to enable and disable specific package managers in the analyzer
configuration. This can still be overridden by the respective command
line options of the analyzer command.

The refactoring in this change is also a preparation for adding more
package manager specific configuration options later on.

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -28,6 +28,7 @@ import com.github.ajalt.clikt.core.requireObject
 import com.github.ajalt.clikt.parameters.options.associate
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.deprecated
 import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
@@ -159,7 +160,11 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",")
+    }.split(",").deprecated(
+        message = "--package-managers is deprecated, use -P ort.analyzer.enabledPackageManagers=... on the ort " +
+                "command instead.",
+        tagValue = "use -P ort.analyzer.enabledPackageManagers=... on the ort command instead"
+    )
 
     private val deactivatedPackageManagers by option(
         "--not-package-managers", "-n",
@@ -169,7 +174,11 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",")
+    }.split(",").deprecated(
+        message = "--not-package-managers is deprecated, use -P ort.analyzer.disabledPackageManagers=... on the ort " +
+                "command instead.",
+        tagValue = "use -P ort.analyzer.disabledPackageManagers=... on the ort command instead"
+    )
 
     private val globalOptionsForSubcommands by requireObject<GlobalOptions>()
 
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -37,6 +37,7 @@ import com.github.ajalt.clikt.parameters.types.file
 
 import org.ossreviewtoolkit.analyzer.Analyzer
 import org.ossreviewtoolkit.analyzer.PackageManager
+import org.ossreviewtoolkit.analyzer.PackageManagerFactory
 import org.ossreviewtoolkit.analyzer.curation.ClearlyDefinedPackageCurationProvider
 import org.ossreviewtoolkit.analyzer.curation.CompositePackageCurationProvider
 import org.ossreviewtoolkit.analyzer.curation.FallbackPackageCurationProvider
@@ -51,6 +52,7 @@ import org.ossreviewtoolkit.cli.utils.inputGroup
 import org.ossreviewtoolkit.cli.utils.outputGroup
 import org.ossreviewtoolkit.cli.utils.writeOrtResult
 import org.ossreviewtoolkit.model.FileFormat
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.readValueOrNull
 import org.ossreviewtoolkit.model.utils.DefaultResolutionProvider
@@ -64,10 +66,10 @@ import org.ossreviewtoolkit.utils.ort.ORT_RESOLUTIONS_FILENAME
 import org.ossreviewtoolkit.utils.ort.log
 import org.ossreviewtoolkit.utils.ort.ortConfigDirectory
 
-class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine dependencies of a software project.") {
-    private val allPackageManagersByName = PackageManager.ALL.associateBy { it.managerName }
-        .toSortedMap(String.CASE_INSENSITIVE_ORDER)
+private val allPackageManagersByName = PackageManager.ALL.associateBy { it.managerName }
+    .toSortedMap(String.CASE_INSENSITIVE_ORDER)
 
+class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine dependencies of a software project.") {
     private val inputDir by option(
         "--input-dir", "-i",
         help = "The project directory to analyze. As a special case, if only one package manager is activated, this " +
@@ -152,20 +154,22 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
     private val activatedPackageManagers by option(
         "--package-managers", "-m",
         help = "The comma-separated package managers to activate, any of ${allPackageManagersByName.keys}. Note that " +
-                "deactivation overrides activation."
+                "deactivation overrides activation. If set, package manager activation settings from configuration " +
+                "files are ignored."
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",").default(PackageManager.ALL.toList())
+    }.split(",")
 
     private val deactivatedPackageManagers by option(
         "--not-package-managers", "-n",
         help = "The comma-separated package managers to deactivate, any of ${allPackageManagersByName.keys}. Note " +
-                "that deactivation overrides activation."
+                "that deactivation overrides activation. If set, package manager activation settings from " +
+                "configuration files are ignored."
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",").default(emptyList())
+    }.split(",")
 
     private val globalOptionsForSubcommands by requireObject<GlobalOptions>()
 
@@ -200,13 +204,19 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
         println("Looking for analyzer-specific configuration in the following files and directories:")
         println("\t" + configurationInfo)
 
-        val distinctPackageManagers = activatedPackageManagers.toSet() - deactivatedPackageManagers.toSet()
-        println("The following package managers are activated:")
-        println("\t" + distinctPackageManagers.joinToString())
+        val config = globalOptionsForSubcommands.config
+
+        val enabledPackageManagers = if (activatedPackageManagers != null || deactivatedPackageManagers != null) {
+            (activatedPackageManagers?.toSet() ?: PackageManager.ALL) - deactivatedPackageManagers.orEmpty().toSet()
+        } else {
+            config.analyzer.determineEnabledPackageManagers()
+        }
+
+        println("The following package managers are enabled:")
+        println("\t" + enabledPackageManagers.joinToString())
 
         println("Analyzing project path:\n\t$inputDir")
 
-        val config = globalOptionsForSubcommands.config
         val analyzer = Analyzer(config.analyzer, labels)
 
         val repositoryConfiguration = actualRepositoryConfigurationFile.takeIf { it.isFile }?.readValueOrNull()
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -36,10 +36,17 @@ data class AnalyzerConfiguration(
...
 
     /**
-     * Package manager specific configuration options. The key needs to match the name of the package manager class,
-     * e.g. "NuGet" for the NuGet package manager. See the documentation of the respective class for available options.
+     * A flag to indicate whether all package managers should be enabled by default. If true, all package managers that
+     * are not explicitly disabled are enabled. If false, all package managers that are not explicitly enabled are
+     * disabled.
      */
-    val options: Map<String, Options>? = null,
+    val enablePackageManagersByDefault: Boolean = true,
```
sschuberth, line 43: Do we really need this new flag? I have a tendency to not introduce it to reduce complexity. On CLI, also all PMs are enabled by default without the possibility to change that.

sschuberth (dismissed):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -152,20 +154,22 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
         "--package-managers", "-m",
         help = "The comma-separated package managers to activate, any of ${allPackageManagersByName.keys}. Note that " +
-                "deactivation overrides activation."
+                "deactivation overrides activation. If set, package manager activation settings from configuration " +
+                "files are ignored."
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",").default(PackageManager.ALL.toList())
+    }.split(",")
```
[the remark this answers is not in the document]
mnonnenmacher: I think that's ok as previously the list of package managers was printed twice, so it's still printed once now.

model/src/main/kotlin/config/AnalyzerConfiguration.kt
mnonnenmacher: It's there for the same reason as the `-n` command line parameter. If you want to enable only one package manager, you would otherwise have to disable all others.

sschuberth: I believe on the CLI being able to negate the logic made sense, as listing e.g. all but one package manager clutters the command line, and having to type the list of PMs over and over again is inconvenient. However, with the PM options in `ort.conf` I don't think that's a real issue anymore. You could simply search & replace if you'd want to "mass toggle" PMs.

That's why I'd still be in favor of removing that option completely. However, maybe there's an alternative: I know you already mentioned that you like your way of per-PM options. But many we could extract the `enabled` flag out of it, and still have my proposed lists of enabled / disabled PMs in `ort.conf`. This would be similar to e.g. the scanner storage configuration, where the storages themselves have their options, but whether they are actually used as readers / writers is configured separately.

sschuberth: > I don't like the way it's implemented for scan storages either, because the configuration is scattered around.

It's maybe a question of encapsulation: Should a storage / PM itself know whether it's enabled, or is that the responsibility of an outer "manager"? I tend to say it's the latter.

> I would actually prefer if there were `read` and `write` flags in the storage configuration itself.

Even if getting a bit off topic: How would you then define the priority of storages?

Actually that question is semi-related to PMs: We'll soon have the need to configure the (relative) order in which PMs are run. I believe that requires a high-order concept anyway, similar to the mentioned enabled / disabled lists. Did you think about a syntax for that in `ort.conf` yet?

mnonnenmacher: > It's maybe a question of encapsulation: Should a storage / PM itself know whether it's enabled, or is that the responsibility of an outer "manager"? I tend to say it's the latter.

I think the focus for the config file should be usability, not software design patterns, and how the values of the configuration classes are mapped to other classes is an implementation detail. Also for a value like `licenseFilePatterns` in the top-level configuration the location does not tell anything about in which classes this value is used internally, only that it is a global configuration that is not specific to one ORT tool.

> > I would actually prefer if there were `read` and `write` flags in the storage configuration itself.
> 
> Even if getting a bit off topic: How would you then define the priority of storages?

I would say the order in which they are defined. That would work because the priority is only relevant for reading, not for writing.

> Actually that question is semi-related to PMs: We'll soon have the need to configure the (relative) order in which PMs are run. I believe that requires a high-order concept anyway, similar to the mentioned enabled / disabled lists. Did you think about a syntax for that in `ort.conf` yet?

What we discussed separately was the we will need a way to configure that a package manager must run after (or before) another, so it would be like:
```
packageManagers {
  Gradle.mustRunAfter = Maven, NPM
}
```
Or in a more complete example it could look like:
```
enablePackageManagersByDefault = false

packageManagers {
  NPM.enabled = true

  NuGet {
    enabled = true
    mustRunAfter = NPM
    options {
      directDependenciesOnly = true
    }
  }
}
```
Or alternatively a `mustRunBefore` option. From the functionality they would be equivalent, so I'm not sure if we should have both or only one of them, and then which.

model/src/main/resources/reference.conf
```diff
@@ -15,12 +15,18 @@ ort {
 
   analyzer {
     allowDynamicVersions = true
+    enablePackageManagersByDefault = true
```
sschuberth, line 18: I'm still not convinced that having this feature flag is a good idea, but I won't block it either. However, I'd like us to come up with a better name, as with `enablePackageManagersByDefault = false` it's not clear enough for *me* that it means *all* PMs are disabled by default.

It would be much easier with an *enum*, like `defaultPackageManagerState = {enabled,disabled}`, but if we'd like to stick with a *boolean*, maybe `disableAllPackageManagers` is better, as it both inverts the logic and explicitly mentions "All"?

mnonnenmacher: I'm fine with the suggestions but I also like the one I proposed.
@oss-review-toolkit/core-devs As this is a breaking configuration change it would be nice to hear another opinion as well (to the change in general, not just the option name).

mnonnenmacher: So I thought about the name again, if there is an option called `enablePackageManagersByDefault` and it is set to `false`, I don't understand why a user should not assume that this means that all package managers are disabled if not explicitly enabled.

sschuberth: > I don't understand why a user should not assume that this means that all package managers are disabled if not explicitly enabled.

As mentioned before, I'm mostly missing an explicit "all".

But also, what this option should emphasize is that the default of enabling all PMs gets changed to disabling all PMs. So the emphasis is on "disable". That's why I'd prefer to also have "disable" in the name.

But like I said, I won't block this and will discard my change request; the naming doesn't really matter too much if I'm still having problems with the feature per se. I'll approve today if there's no other approval / further comments.

model/src/main/kotlin/config/AnalyzerConfiguration.kt
fviernau:
```
enablePackageManagersByDefault = false

packageManagers {
  NPM.enabled = true

  NuGet {
    enabled = true
    mustRunAfter = NPM
    options {
      directDependenciesOnly = true
    }
  }
}
``` 

Maybe one thing which hasn't been considered yet in above more complete example. If a package manager has configuration attributes without default values, then disabling a package manager (assuming `enablePackageManagersByDefault = true`) requires adding further configuration attributes. Probably this case isn't relevant as there are always default values, just wanted to bring it to consideration.

Apart from that we could add one more alternative implementation:
´´´
enabledPackageManagers = NPM, Yarn, Gradle
disabledPackageManager = GoMod
``` 
1. No default values for each package managers enabled state
2. Behavior: in case enabledPackageManagers + disabledPackageManager != allPackageManager: stop with an error sayinfg that the enabled state must be defined
3. This will enforce the user to make a conscious decision whether to enable any package manager after an upgrade of ORT which added a package manager

Comment (mnonnenmacher) [reply model/src/main/kotlin/config/AnalyzerConfiguration.kt:43]: > Maybe one thing which hasn't been considered yet in above more complete example. If a package manager has configuration attributes without default values, then disabling a package manager (assuming `enablePackageManagersByDefault = true`) requires adding further configuration attributes. Probably this case isn't relevant as there are always default values, just wanted to bring it to consideration.

I'm not sure if I understand that correctly, but maybe it's a misunderstanding. If the config contains a line like `NuGet.enabled = false` it does not mean that an instance of `NuGet` is created and therefore options without default values would have to be provided. The actual class that is instantiated is `PackageManagerConfiguration`.

> Apart from that we could add one more alternative implementation:
> 
> ```
> enabledPackageManagers = NPM, Yarn, Gradle
> disabledPackageManager = GoMod
> ```
> 
>     1. No default values for each package managers enabled state
> 
>     2. Behavior: in case enabledPackageManagers + disabledPackageManager != allPackageManager: stop with an error saying that the enabled state must be defined
> 
>     3. This will enforce the user to make a conscious decision whether to enable any package manager after an upgrade of ORT which added a package manager

That's a good point, but as my intention is to allow this configuration also in the `.ort.yml` in a later step, it would not be manageable.

File: cli/src/main/kotlin/commands/AnalyzerCommand.kt (hunk 2 above)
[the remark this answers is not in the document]
Comment (mnonnenmacher) [reply cli/src/main/kotlin/commands/AnalyzerCommand.kt:162]: The technical reason why I removed the default value is that I could not find a way to find out if the value was provided on the command line or not, but this is a requirement to decide if the values from the configuration file or from the command line should be used.

**Comment by mnonnenmacher** [comment]
I have updated the PR based on the feedback on Slack and removed the breaking change label, as the breaking config change will be done in another PR now.

File: cli/src/main/kotlin/commands/AnalyzerCommand.kt (hunk 2 above)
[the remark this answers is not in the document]
Comment (sschuberth) [reply cli/src/main/kotlin/commands/AnalyzerCommand.kt]: For the record, such a check could probably be done via something like

    listOf("--package-managers", "-m").any { it in currentContext.originalArgv }

[but that's not really nice as it repeats the option names](https://github.com/ajalt/clikt/issues/359).

File: cli/src/main/kotlin/commands/AnalyzerCommand.kt [hunk 4]
```diff
@@ -286,3 +296,10 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
         severityStats.print().conclude(config.severeIssueThreshold, 2)
     }
 }
+
+private fun AnalyzerConfiguration.computeEnabledPackageManagers(): Set<PackageManagerFactory> {
```
sschuberth, line 307: Nit: "compute" is maybe a bit pretentious. Maybe just one of these: "get", "determine".

model/src/main/resources/reference.conf
```diff
@@ -16,6 +16,13 @@ ort {
   analyzer {
     allowDynamicVersions = true
 
+    # A list of disabled package managers. Disabling a package manager here overrides enabling it in
+    # `enabledPackageManagers`.
+    disabledPackageManagers = [Maven, NPM]
```
sschuberth, line 21: Nit: Due to precedence explained in the comment, should we maybe make `disabledPackageManagers` go after `enabledPackageManagers` to emphasize that "the last word counts"?

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -159,7 +160,10 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",")
+    }.split(",").deprecated(
+        message = "--package-managers is deprecated, use -P ort.analyzer.enabledPackageManagers=[] instead.",
```
sschuberth, line 164: Is `[]` correct here? Don't we expect a comma-separated list without surrounding `[]` on the command line?

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -159,7 +160,10 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
+        tagValue = "use -P ort.analyzer.enabledPackageManagers=[] instead."
```
sschuberth, line 165: I haven't checked, but should this sentence start capitalized to look good in the CLI help?

cli/src/main/kotlin/commands/AnalyzerCommand.kt
mnonnenmacher: Used `determine` because `get` would override the generated getter for the property with the same name.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
mnonnenmacher: This it what it looks like:
```
  -m, --package-managers VALUE     The comma-separated package managers to
                                   activate, any of [Bower, Bundler, Cargo,
                                   Carthage, CocoaPods, Composer, Conan,
                                   DotNet, GoDep, GoMod, Gradle, Maven, NPM,
                                   NuGet, PIP, Pipenv, Pub, SBT,
                                   SpdxDocumentFile, Stack, Unmanaged, Yarn,
                                   Yarn2]. Note that deactivation overrides
                                   activation. If set, package manager
                                   activation settings from configuration
                                   files are ignored. (deprecated: use -P
                                   ort.analyzer.enabledPackageManagers=[] on
                                   the ort command instead)
```

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -159,7 +160,11 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",")
+    }.split(",").deprecated(
+        message = "--package-managers is deprecated, use -P ort.analyzer.enabledPackageManagers on the ort command " +
+                "instead.",
+        tagValue = "use -P ort.analyzer.enabledPackageManagers=[] on the ort command instead"
```
sschuberth, line 166: This still has `=[]`. Also, I wonder whether instead of dropping it completely, we should write something like `=...` (I haven't checked how we might already be doing this in other places).

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -169,7 +174,11 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
     ).convert { name ->
         allPackageManagersByName[name]
             ?: throw BadParameterValue("Package managers must be one or more of ${allPackageManagersByName.keys}.")
-    }.split(",")
+    }.split(",").deprecated(
+        message = "--not-package-managers is deprecated, use -P ort.analyzer.disabledPackageManagers on the ort " +
+                "command instead.",
+        tagValue = "use -P ort.analyzer.disabledPackageManagers=[] on the ort command instead"
```
sschuberth, line 180: See above.

sschuberth (changes requested):

sschuberth (approved):

PR #5448, opened 2022-06-15
LicenseFinding: Apply detected license mapping to valid SPDX expressions

MarcelBochtler:
The functionality to map detected licenses can be used to map non-SPDX
licenses returned from scanners. However, it is also intended to be used
to map unwanted but valid licenses. Therefore, apply the mapping also to
valid SPDX expressions.

This is a fixup for https://github.com/oss-review-toolkit/ort/commit/0b8377dafdcae2cf48958876ff848c8c1d1b9d07.

### DIFF
--- a/model/src/main/kotlin/LicenseFinding.kt
+++ b/model/src/main/kotlin/LicenseFinding.kt
@@ -23,7 +23,6 @@ import com.fasterxml.jackson.annotation.JsonInclude
 
 import org.ossreviewtoolkit.model.config.LicenseFindingCuration
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
-import org.ossreviewtoolkit.utils.spdx.isSpdxExpression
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/LicenseFinding.kt
```diff
@@ -62,7 +61,7 @@ data class LicenseFinding(
             score: Float? = null,
             detectedLicenseMapping: Map<String, String>
         ): LicenseFinding = LicenseFinding(
-            license = if (license.isSpdxExpression() || detectedLicenseMapping.isEmpty()) {
+            license = if (detectedLicenseMapping.isEmpty()) {
```
mnonnenmacher, line 64: Commit message: Please explain why the change is necessary.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5449, opened 2022-06-15
fossid-webapp: Improve FossID reporter error logging

nnobelis:
When the report cannot be generated, log the headers of the response
received from the FossID server for better error logging.


clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -34,6 +36,8 @@ import org.ossreviewtoolkit.clients.fossid.model.rules.RuleType
 internal const val SCAN_GROUP = "scans"
 private const val PROJECT_GROUP = "projects"
 
+private val logger = loggerOf(FossIdRestService::class.java)
```
sschuberth, line 39: Can't you use `FossIdRestService.logger` directly? Or first move it from the companion object to a top-level property for more convenient use over here?

sschuberth (approved):

PR #5452, opened 2022-06-15
DeleteCommand: Add option to delete scan results with a specific license

MarcelBochtler:
The `--license` option can be used to delete stored scan results, if the
specified license is contained in the detected licenses.
This can be useful if the `detectedLicenseMapping` is changed and the
user wants to ensure that the stored scan results no longer returns the
old mapping.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Bosch.IO GmbH
+ * Copyright (C) 2021-2022 Bosch.IO GmbH
  *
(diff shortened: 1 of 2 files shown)


helper-cli/src/main/kotlin/commands/scanstorage/DeleteCommand.kt
```diff
@@ -93,11 +100,28 @@ internal class DeleteCommand : CliktCommand(
         }
 
         val identifierCondition = packageId?.let { ScanResults.identifier regexp it }
+
+        val licenseCondition = license?.let {
+            log.info { "Searching for stored scan results with license '$license'. This might take some time." }
```
sschuberth, line 105: Nit: Maybe even use `println` as this is a CLI tool?

sschuberth (approved):

Issue #5021, opened 2022-02-04
GoMod analysis seems to get package versions and graph pruning wrong


andreasdotorg:
I've been trying to make the ORT Analyzer work for a golang project, with less than satisfactory results. (If you want to follow along, you can find the code at https://github.com/impfen/services-inoeg). Main effects are: confusion about the right versions, missed dependencies.

After looking at the code, I think the main issue stems from the fact that the Analyzer looks at dependencies on a package level, not on a module level, and ignores minimum version resolution. Let me try to outline what I think is going wrong and why.

First, let's look at the definition of things in golang, what exactly are modules and packages? Here's a [definition](https://levelup.gitconnected.com/using-modules-and-packages-in-go-36a418960556):

> In Go, a package is a directory of .go files, and packages form the basic building blocks of a Go program. Using packages, you organize your code into reusable units. A module, on the other hand, is a collection of Go packages, with dependencies and versioning built-in.

I think the key here is the dependencies and versioning aspect. 

If we look at how the Analyzer works at the moment, it pulls a graph of module dependencies using `go mod graph`, and then tries to prune that graph to actually used modules by calling `go mod why`.  Let's look at how this fails with the example linked above. First issue: no resolution of versions. Let's find out who depends on github.com/golang/protobuf:

```
$ go mod graph|grep " github.com/golang/protobuf"
github.com/prometheus/client_golang@v1.11.0 github.com/golang/protobuf@v1.4.3
github.com/onsi/gomega@v1.16.0 github.com/golang/protobuf@v1.5.2
github.com/prometheus/client_model@v0.2.0 github.com/golang/protobuf@v1.2.0
github.com/prometheus/common@v0.26.0 github.com/golang/protobuf@v1.4.2
github.com/onsi/gomega@v1.10.1 github.com/golang/protobuf@v1.4.2
google.golang.org/protobuf@v1.23.0 github.com/golang/protobuf@v1.4.0
github.com/prometheus/client_golang@v1.7.1 github.com/golang/protobuf@v1.4.2
google.golang.org/protobuf@v1.26.0 github.com/golang/protobuf@v1.5.0
github.com/prometheus/common@v0.10.0 github.com/golang/protobuf@v1.3.2
google.golang.org/appengine@v1.4.0 github.com/golang/protobuf@v1.2.0
github.com/onsi/gomega@v1.7.1 github.com/golang/protobuf@v1.2.0
[...]

PR #5453, opened 2022-06-15
GoMod: Account for `go mod graph` not resolving module versions

fviernau:
The documentation [1] states that "Each vertex in the module graph
represents a specific version of a module. Each edge in the graph
represents a requirement on a minimum version of a dependency."

As ORT's dependency tree representation is about resolved version, not
about version requirements, using the graph as-is is wrong. Turn the
version requirements graph into a dependency graph with resolved version
by utilizing the command ´go list -m´, which outputs the resolved
version for each respective module [1][2].

[1] https://go.dev/ref/mod#go-mod-graph
[2] https://go.dev/ref/mod#minimal-version-selection

This fixes one of the issues outlined in #5021.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -150,13 +150,17 @@ class GoMod(
      * Return the module graph output from `go mod graph` with unused dependencies removed.
      */
     private fun getModuleGraph(projectDir: File): Graph {
-        fun parseModuleEntry(entry: String) =
-            Identifier(
-                type = managerName,
-                namespace = "",
-                name = entry.substringBefore('@'),
-                version = entry.substringAfter('@', "")
-            )
+        val moduleVersionForModuleName = getAllModules(projectDir).associateBy({ it.name }, { it.version })
+
+        fun parseModuleEntry(entry: String): Identifier =
+            entry.substringBefore('@').let { moduleName ->
+                Identifier(
+                    type = managerName,
+                    namespace = "",
+                    name = moduleName,
+                    version = moduleVersionForModuleName[moduleName].orEmpty()
+                )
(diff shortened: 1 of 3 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -191,6 +194,22 @@ class GoMod(
         return graph
     }
 
+    /**
+     * Return the list of all modules contained in the dependency tree with resolved versions.
+     */
+    private fun getAllModules(projectDir: File): Set<Identifier> {
+        val result = mutableSetOf<Identifier>()
+
+        for (line in run("list", "-m", "all", workingDir = projectDir).requireSuccess().stdout.lines()) {
```
sschuberth, line 203: `run` already calls `requireSuccess` so you don't need to do that explicitly.

Style-wise, I'd prefer to introduce a variable and use `forEach`:

```
val list = run("list", "-m", "all", workingDir = projectDir)
list.stdout.lines().forEach { line ->
  // ...
}
```

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -150,13 +150,16 @@ class GoMod(
      * Return the module graph output from `go mod graph` with unused dependencies removed.
      */
     private fun getModuleGraph(projectDir: File): Graph {
-        fun parseModuleEntry(entry: String) =
+        val moduleVersionForModuleName = getAllModules(projectDir).associateBy({ it.name }, { it.version })
+
+        fun parseModuleEntry(entry: String): Identifier = entry.substringBefore('@').let { moduleName ->
```
sschuberth, line 155: Style: Please break the line after `=`.

sschuberth (approved):

PR #5455, opened 2022-06-16
Analyzer: Prepare making the package manager order configurable

mnonnenmacher:
Prepare the analyzer code to support that package managers can be configured to wait for other package managers. The required configuration options will be added in another PR, after #5438 is merged.


analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -172,3 +175,73 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
         return analyzerResultBuilder.build()
     }
 }
+
+private class PackageManagerRunner(
```
MarcelBochtler, line 203: IIUC, this class will become a quite central component in the `Analyzer`. Can you please add some documentation?

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -137,43 +142,62 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
...
+            }.forEach { launch { it.start() } }
+        }
+
+        return state.buildResult()
+    }
+}
+
+private class AnalyzerState(curationProvider: PackageCurationProvider) {
+    private val builder = AnalyzerResultBuilder(curationProvider)
+    private val scope = CoroutineScope(Dispatchers.Default)
```
MarcelBochtler, line 166: What is the reason for switching from `Dispatchers.IO` to `Dispatchers.Default`?

analyzer/src/main/kotlin/Analyzer.kt
mnonnenmacher: I added only a short description, because I think the documentation of the constructor arguments explains the rest.

analyzer/src/main/kotlin/Analyzer.kt
mnonnenmacher: I'm using `Dispatchers.Default` here because the state itself does not do IO.

MarcelBochtler (approved):

PR #5458, opened 2022-06-17
Use Gradle's new version catalog for dependency management

sschuberth:


scanner/build.gradle.kts
```diff
@@ -20,23 +20,11 @@
...
-val retrofitVersion: String by project
-val scanossVersion: String by project
-val sw360ClientVersion: String by project
-val wiremockVersion: String by project
-
 val askalonoVersion: String by project
 val boyterLcVersion: String by project
 val licenseeVersion: String by project
 val scancodeVersion: String by project
+val scanossVersion: String by project
```
mnonnenmacher, line 27: Couldn't you get this version from the `libs` object like you did in some places in the previous commit? To avoid that the version is declared in two places.

mnonnenmacher (changes requested):

PR #5459, opened 2022-06-17
GoMod: Distinguish main module from development-only dependencies

fviernau:
See individual commits.

These are improvements on top of recent work done for fixing #5021.

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
@@ -16,6 +16,20 @@ project:
     path: "<REPLACE_PATH>"
   homepage_url: ""
   scopes:
+  - name: "main"
+    dependencies:
+    - id: "GoMod::github.com/fatih/color:v1.13.0"
+      linkage: "PROJECT_STATIC"
+    - id: "GoMod::github.com/google/uuid:v1.0.0"
+      linkage: "PROJECT_STATIC"
+    - id: "GoMod::github.com/mattn/go-colorable:v0.1.12"
+      linkage: "PROJECT_STATIC"
+    - id: "GoMod::github.com/mattn/go-isatty:v0.0.14"
+      linkage: "PROJECT_STATIC"
+    - id: "GoMod::github.com/pborman/uuid:v1.2.1"
+      linkage: "PROJECT_STATIC"
+    - id: "GoMod::golang.org/x/sys:v0.0.0-20220610221304-9f5ed59c137d"
+      linkage: "PROJECT_STATIC"
   - name: "vendor"
     dependencies:
     - id: "GoMod::github.com/davecgh/go-spew:v1.1.0"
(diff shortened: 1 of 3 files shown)


analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
```diff
@@ -16,6 +16,20 @@ project:
     path: "<REPLACE_PATH>"
   homepage_url: ""
   scopes:
+  - name: "dependencies"
```
sschuberth, line 19: Commit message: Typo in "developmenti-only".

analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
```diff
@@ -30,7 +30,7 @@ project:
       linkage: "PROJECT_STATIC"
     - id: "GoMod::golang.org/x/sys:v0.0.0-20220610221304-9f5ed59c137d"
       linkage: "PROJECT_STATIC"
-  - name: "vendor"
+  - name: "devDependencies"
```
sschuberth, line 33: I'm not sure whether this renaming is a good idea. Classically, "vendored" dependencies are dependencies that were copied to the actual project's source code. At least that's how I know the term from other languages, like Ruby. And isn't this also the case for Go? Or are these dependencies *not* copied to the project, and really retrieved from a remote, like "regular" dependencies?

fviernau: The "vendor" package can be copied to a "vendor" directory under the project root using the vendor command.
Looking at the following links, in my view it's ok to interpret it as the set of packages which actually do get copies. These are exactly the set of packages / modules needed to build and test the main module. Note that these are exactly the set of dependencies which should be included into ORT's analysis result.

Maybe "all" could be a better name to use for that "vendor" / "devDependencies" scope.
Anyhow, I suggest we just try to find good names for these two scopes. I'm open for changing these (have no strong opinion).

1. https://go.dev/ref/mod#vendoring
2. https://github.com/golang/go/blob/2a78e8afc0994f5b292bc9a5a7258c749e43032f/src/cmd/go/internal/modcmd/why.go#L26-L28

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -199,6 +211,26 @@ class GoMod(
...
+        )
+
+        list.stdout.lines().forEach { line ->
+            val columns = line.split(' ')
+            if (columns.size != 2) return@forEach
+
+            result += columns[0]
+        }
+
+        return result
```
mnonnenmacher, line 231: You can make it a bit more compact:
```kotlin
return list.stdout.lineSequence()
    .map { it.split(' ') }
    .filter { it.size == 2 }
    .map { it[0] }
```

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: I believe as-is is more consistent with local style. So in this commit I prefer to keep it. Fine?

mnonnenmacher (approved):

PR #5461, opened 2022-06-17
Reference.conf consistency improvements

tsteenbe:
Please see individual commits for details.

### DIFF
--- a/model/src/main/resources/reference.conf
+++ b/model/src/main/resources/reference.conf
@@ -134,21 +134,21 @@ ort {
 
       FossId {
         serverUrl = "https://fossid.example.com/instance/"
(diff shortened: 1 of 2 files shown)


model/src/main/resources/reference.conf
```diff
@@ -248,7 +248,7 @@ ort {
     mail {
       hostName = "localhost"
       username = user
-      password = secret
+      password = password
```
mnonnenmacher, line 255: If you change values here you also need to update `OrtConfigurationTest` accordingly.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5465, opened 2022-06-17
Analyzer: Make it possible to configure that one package manager must run after another

mnonnenmacher:

### DIFF
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -150,7 +150,7 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
                     manager = manager,
                     definitionFiles = files,
                     labels = labels,
-                    mustRunAfter = emptySet(),
+                    mustRunAfter = config.packageManagers?.get(manager.managerName)?.mustRunAfter?.toSet().orEmpty(),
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -36,14 +36,14 @@ data class AnalyzerConfiguration(
     val allowDynamicVersions: Boolean = false,
 
     /**
-     * A list of the case-insensitive names of package managers that are enabled. Disabling a package manager in
-     * [disabledPackageManagers] overrides enabling it here.
+     * A list of names of package managers that are enabled. Disabling a package manager in [disabledPackageManagers]
```
sschuberth, line 39: On the CLI these *are* case-insensitive, so should we rather change the implementation to make them so here, too? I haven't checked how we do it in other places where we refer to e.g. scanners or storages; in any case, we should align the case-sensitivity across all these places.

mnonnenmacher: I have dropped that commit, as these option are actually case-insensitive. The configuration map for package managers is not, but I will address that separately.

sschuberth (approved):

sschuberth (approved):

PR #5467, opened 2022-06-20
scanoss: Replace the hard-coded "Blacklist" with a configurable one

sschuberth:
For reference, the original hard-coded values from the ScanOss library's
`BlacklistRules` [1] are used.

This is a step towards getting rid of the dependency on the deprecated
ScanOss Java library in favor of only using ORT's own client library.

[1]: https://github.com/scanoss/scanner.java/blob/350c211/src/main/java/com/scanoss/scanner/BlacklistRules.java#L32-L40

### DIFF
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -150,6 +150,15 @@ ort:
 
         timeout: 60
 
+      ScanOss:
+        ignoredFileSuffixes: >-
+          .1,.2,.3,.4,.5,.6,.7,.8,.9,.ac,.am,.bmp,.build,.cfg,.chm,.changelog,.class,.cmake,.conf,.config,.contributors,
+          .copying,.csproj,.css,.csv,.cvsignore,.dat,.data,.dtd,.dts,.dtsi,.eps,.geojson,.gif,.gitignore,.glif,.gmo,
+          .guess,.hex,.html,.htm,.ico,.in,.inc,.info,.ini,.ipynb,.jpg,.jpeg,.json,.license,.log,.m4,.map,.markdown,.md,
+          .md5,.mk,.makefile,.meta,.mxml,.notice,.out,.pdf,.pem,.phtml,.png,.po,.prefs,.properties,.readme,.result,.rst,
+          .scss,.sha,.sha1,.sha2,.sha256,.sln,.spec,.sub,.svg,.svn-base,.tab,.template,.test,.tex,.todo,.txt,.utf-8,
+          .version,.vim,.wav,.xht,.xhtml,.xml,.xpm,.xsd,.xul,.yaml,.yml
+
     storages:
       local:
         backend:
--- a/scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
+++ b/scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
@@ -47,7 +47,7 @@ internal data class ScanOssConfig(
         const val IGNORED_FILE_SUFFIXES_PROPERTY = "ignoredFileSuffixes"
 
         /** The default API URL to use. */
-        private const val DEFAULT_API_URL = "https://osskb.org/api/"
+        internal const val DEFAULT_API_URL = "https://osskb.org/api/"
 
         /** The default list of ignored file extension. */
         private val DEFAULT_IGNORED_FILE_EXTENSIONS = listOf(
@@ -66,7 +66,7 @@ internal data class ScanOssConfig(
         )
 
         /** The default list of ignored file suffixes. */
-        private val DEFAULT_IGNORED_FILE_SUFFIXES = DEFAULT_IGNORED_FILE_EXTENSIONS + listOf(
+        internal val DEFAULT_IGNORED_FILE_SUFFIXES = DEFAULT_IGNORED_FILE_EXTENSIONS + listOf(
             "-doc", "authors", "changelog", "config", "copying", "ignore", "license", "licenses", "manifest", "news",
             "notice", "readme", "sqlite", "sqlite3", "swiftdoc", "texidoc", "todo", "version"
         )
--- a/scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
+++ b/scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
@@ -31,7 +31,10 @@ internal data class ScanOssConfig(
     val apiUrl: String,
 
     /** API Key required to authenticate with the ScanOSS server. */
-    val apiKey: String
+    val apiKey: String,
+
+    /** A list of suffixes of files to be ignored during the scan. */
+    val ignoredFileSuffixes: List<String>
 ) {
     companion object {
         /** Name of the configuration property for the API URL. */
@@ -40,6 +43,31 @@ internal data class ScanOssConfig(
         /** Name of the configuration property for the API key. */
         const val API_KEY_PROPERTY = "apiKey"
 
+        /** Name of the configuration property for the list of ignored file suffixes. */
+        const val IGNORED_FILE_SUFFIXES_PROPERTY = "ignoredFileSuffixes"
+
+        /** The default list of ignored file extension. */
+        private val DEFAULT_IGNORED_FILE_EXTENSIONS = listOf(
+            ".1", ".2", ".3", ".4", ".5", ".6", ".7", ".8", ".9", ".ac", ".adoc", ".am", ".asciidoc", ".bmp", ".build",
+            ".c9", ".c9revisions", ".cache", ".cfg", ".chm", ".class", ".cmake", ".cnf", ".conf", ".config",
+            ".contributors", ".copying", ".cover", ".coverage", ".crt", ".csproj", ".css", ".csv", ".dat", ".data",
+            ".doc", ".docx", ".dotcover", ".dtd", ".dts", ".dtsi", ".dump", ".editorconfig", ".egg", ".eot", ".eps",
+            ".gdoc", ".gem", ".geojson", ".gif", ".glif", ".gml", ".gmo", ".gradle", ".guess", ".hex", ".htm", ".html",
+            ".ico", ".iml", ".in", ".inc", ".info", ".ini", ".ipynb", ".iws", ".jpeg", ".jpg", ".json", ".jsonld",
+            ".lcov", ".lock", ".log", ".lst", ".m4", ".manifest", ".map", ".markdown", ".md", ".md5", ".meta", ".mk",
+            ".mxml", ".o", ".otf", ".out", ".pbtxt", ".pdb", ".pdf", ".pem", ".phtml", ".pickle", ".pid", ".plist",
+            ".plt", ".png", ".po", ".pot", ".ppt", ".prefs", ".properties", ".pyc", ".qdoc", ".result", ".rgb", ".rst",
+            ".scss", ".sha", ".sha1", ".sha2", ".sha256", ".sln", ".spec", ".sql", ".sub", ".svg", ".svn-base", ".tab",
+            ".template", ".test", ".tex", ".tiff", ".toml", ".ttf", ".txt", ".utf-8", ".vim", ".wav", ".wfp", ".whl",
+            ".woff", ".xht", ".xhtml", ".xls", ".xlsx", ".xml", ".xpm", ".xsd", ".xul", ".yaml", ".yml"
+        )
+
+        /** The default list of ignored file suffixes. */
+        private val DEFAULT_IGNORED_FILE_SUFFIXES = DEFAULT_IGNORED_FILE_EXTENSIONS + listOf(
+            "-doc", "authors", "changelog", "config", "copying", "ignore", "license", "licenses", "manifest", "news",
+            "notice", "readme", "sqlite", "sqlite3", "swiftdoc", "texidoc", "todo", "version"
+        )
+
         fun create(scannerConfig: ScannerConfiguration): ScanOssConfig {
             val scanOssIdScannerOptions = scannerConfig.options?.get("ScanOss")
 
@@ -47,8 +75,12 @@ internal data class ScanOssConfig(
 
             val apiURL = scanOssIdScannerOptions[API_URL_PROPERTY] ?: "https://osskb.org/api/"
             val apiKey = scanOssIdScannerOptions[API_KEY_PROPERTY].orEmpty()
+            val ignoredFileSuffixes = scanOssIdScannerOptions[IGNORED_FILE_SUFFIXES_PROPERTY]
+                ?.split(',')
+                ?.map { it.trim() }
+                ?: DEFAULT_IGNORED_FILE_SUFFIXES
 
-            return ScanOssConfig(apiURL, apiKey)
+            return ScanOssConfig(apiURL, apiKey, ignoredFileSuffixes)
         }
     }
 }
--- a/scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
+++ b/scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
@@ -46,6 +46,9 @@ internal data class ScanOssConfig(
         /** Name of the configuration property for the list of ignored file suffixes. */
         const val IGNORED_FILE_SUFFIXES_PROPERTY = "ignoredFileSuffixes"
 
+        /** The default API URL to use. */
+        private const val DEFAULT_API_URL = "https://osskb.org/api/"
+
         /** The default list of ignored file extension. */
         private val DEFAULT_IGNORED_FILE_EXTENSIONS = listOf(
             ".1", ".2", ".3", ".4", ".5", ".6", ".7", ".8", ".9", ".ac", ".adoc", ".am", ".asciidoc", ".bmp", ".build",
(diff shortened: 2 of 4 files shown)


fviernau (dismissed):
For ScanCode it's been decided to stop using scanner specific ignores but move to a scanner agnostic way for ignoring files [1]. This has the advantage that changing the config does not invalidate the cache key. Also with doing this transformation on-the-fly, AFTER pulling a scan result from storage has the advantage that the transformation logic can be changed at any time without invalidating the cache, e.g. bugs in the transformation become less severe and less expensive to do.

I believe this change would need to provide a rationale why it makes sense to do the ignores in a scanner specific way and also why the ignore patterns for different scanners should be different.

[1] https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/kotlin/config/ScannerConfiguration.kt#L109-L118

sschuberth:
> I believe this change would need to provide a rationale why it makes sense to do the ignores in a scanner specific way and also why the ignore patterns for different scanners should be different.

I guess the key to this question is that ScanCode ans SCANOSS are fundamentally different scanners: While ScanCode is a license / copyright scanner whose results are findings per file, SCANOSS is a snippet scanner whose results are other files that are similar to the scanned files (and from these known similar files "detected" licenses are returned).

So, to avoid the list of similar files to expode (as may oSS projects share similar project templates), SCANOSS tried to limit its scan to "real" source code files only. That's a little bit of a different idea than the existing global `ignorePatterns` in the `ScannerConfiguration`.

Anyway, @eeisegn one thing I just wondered about is whether the (deprecated) (Java clients' blacklist](https://github.com/scanoss/scanner.java/blob/c901f60718ee0557e4e7c19cbe4f68bb6c09af54/src/main/java/com/scanoss/scanner/BlacklistRules.java#L32-L40) is still applicable, or whether SCANOSS is maybe now doing the filtering on the server side. I just found some [similar code in the current Python client](https://github.com/scanoss/scanoss.py/blob/1aaaea6a1f4224c1ae049a7498ed30727515a892/src/scanoss/winnowing.py#L55-L61), but the list of extension is different. So it really seems as if SCANOSS should apply these ignores on the server side, so it's client agnostic.

sschuberth:
> So it really seems as if SCANOSS should apply these ignores on the server side, so it's client agnostic.

Or actually, the Python list of extension to ignore seems to mainly target binary files. So what if we replaced the blacklist completely with a check for binary files only (which is something the deprecated Java client does [*in addition* to the extension check](https://github.com/scanoss/scanner.java/blob/c901f60718ee0557e4e7c19cbe4f68bb6c09af54/src/main/java/com/scanoss/scanner/Winnowing.java#L165)).

fviernau:
> I guess the key to this question is that ScanCode ans SCANOSS are fundamentally different scanners: While ScanCode is a license / copyright scanner whose results are findings per file, SCANOSS is a snippet scanner whose results are other files that are similar to the scanned files (and from these known similar files "detected" licenses are returned).

I missed that it's a sniplet scanner - which changes things a bit in fact.
So, I'm fine with this approach for now. For future iterations we may consider refactoring ScannerConfig to contain both,
(1) ignore patterns for license copyright scanners and (2) ignore patterns for sniplet scanners. So that ignores aren't configured separately for any sniplet scanner implementation but once for all sniplet scanners.

model/src/main/resources/reference.conf
```diff
@@ -154,6 +154,24 @@ ort {
...
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",in,inc,info,ini,ipynb,jpg,jpeg"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",json,license,log,m4,map,markdown"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",md,md5,mk,makefile,meta,mxml"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",notice,out,pdf,pem,phtml,png,po"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",prefs,properties,readme,result"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",rst,scss,sha,sha1,sha2,sha256,sln"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",spec,sub,svg,svn-base,tab"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",template,test,tex,todo,txt,utf-8"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",version,vim,wav,xht,xhtml,xml,xpm"
+        ignoredFileExtensions = ${ort.scanner.options.ScanOss.ignoredFileExtensions}",xsd,xul,yaml,yml"
```
mnonnenmacher, line 173: Is there really no better way to configure this long list? The boilerplate is so massive that I would actually prefer one very long line with just the list in this case. Or would it maybe work if you configure it like one of the ideas below? The [HOCON documentation](https://github.com/lightbend/config/blob/main/HOCON.md#note-arrays-without-commas-or-newlines) is not very clear in that regard.

```
ignoredFileExtensions = [ 1,2,3,4,5,
    6,7,8,9,ac,
    am,...
]

ignoredFileExtensions = [ 1,2,3,4,5 ]
    [6,7,8,9,ac]
    ...
```

sschuberth: > Is there really no better way to configure this long list?

Not that I'm aware of, also see [this answer](https://stackoverflow.com/a/31387467/1127485).

But I'm now hoping to be able to get rid of the blacklist completely, so I'll convert this back to a draft.

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
```diff
@@ -31,7 +31,10 @@ internal data class ScanOssConfig(
     val apiUrl: String,
 
     /** API Key required to authenticate with the ScanOSS server. */
-    val apiKey: String
+    val apiKey: String,
+
+    /** A list of extensions for files (without the dot) to be ignored during the scan. */
+    val ignoredFileExtensions: List<String>
```
mnonnenmacher, line 37: This should default to the list from `reference.yml`, otherwise every user has to set it now.

mnonnenmacher (changes requested):

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
mnonnenmacher: This actually shows that we seem to be missing a test for ignored file extensions.

sschuberth: > This should default to the list from `reference.yml`, otherwise every user has to set it now.

Thanks for noticing. I've used the opportunity to also rework the code to work with suffixes instead of extensions, like in the maintain [SCANOSS Python client](https://github.com/scanoss/scanoss.py/blob/main/src/scanoss/scanner.py#L54-L75).

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
```diff
@@ -40,15 +43,42 @@ internal data class ScanOssConfig(
...
+            ".plt", ".png", ".po", ".pot", ".ppt", ".prefs", ".properties", ".pyc", ".qdoc", ".result", ".rgb", ".rst",
+            ".scss", ".sha", ".sha1", ".sha2", ".sha256", ".sln", ".spec", ".sql", ".sub", ".svg", ".svn-base", ".tab",
+            ".template", ".test", ".tex", ".tiff", ".toml", ".ttf", ".txt", ".utf-8", ".vim", ".wav", ".wfp", ".whl",
+            ".woff", ".xht", ".xhtml", ".xls", ".xlsx", ".xml", ".xpm", ".xsd", ".xul", ".yaml", ".yml"
+        )
+
+        /** The default list of ignored file suffixes. */
+        val IGNORED_FILE_SUFFIXES = IGNORED_FILE_EXTENSIONS + listOf(
+            "-doc", "authors", "changelog", "config", "copying", "ignore", "license", "licenses", "manifest", "news",
+            "notice", "readme", "sqlite", "sqlite3", "swiftdoc", "texidoc", "todo", "version"
```
mnonnenmacher, line 68: Should these be contained in the `reference.yml` as well? Or maybe go the other direction and reduce the list in `reference.yml` to just a few entries, as we don't use it to showcase default configuration but only configuration options.

model/src/main/resources/reference.yml
```diff
@@ -150,6 +150,15 @@ ort:
 
         timeout: 60
 
+      ScanOss:
```
fviernau, line 153: Commit-message: 

There is a generic mechanism for ignoring files [1].
This raises the question to me whether that mechanism should be used, for the same reasoning it is being used for `ScanCode` currently. What do you think?

Apart from that, the below list of default values raises also the question whether the actual values are actually desired.
E.g. why should we exclude a build script from scanning, only for Scan OSS?

[1] https://github.com/oss-review-toolkit/ort/blob/2d27a2244f660e43b192058221805721033aa7ce/scanner/src/main/kotlin/Scanner.kt#L182-L184

fviernau (changes requested):

fviernau:
> I guess the key to this question is that ScanCode ans SCANOSS are fundamentally different scanners: While ScanCode is a license / copyright scanner whose results are findings per file, SCANOSS is a snippet scanner whose results are other files that are similar to the scanned files (and from these known similar files "detected" licenses are returned).
> 
> So, to avoid the list of similar files to expode (as may OSS projects share similar project templates), SCANOSS tries to limit its scan to "real" / relevant source code files only. That's a little bit of a different idea than the existing global `ignorePatterns` in the `ScannerConfiguration`.
> 
> Anyway, @eeisegn one thing I just wondered about is whether the (deprecated) (Java clients' blacklist](https://github.com/scanoss/scanner.java/blob/c901f60718ee0557e4e7c19cbe4f68bb6c09af54/src/main/java/com/scanoss/scanner/BlacklistRules.java#L32-L40) is still applicable, or whether SCANOSS is maybe now doing the filtering on the server side. I just found some [similar code in the current Python client](https://github.com/scanoss/scanoss.py/blob/1aaaea6a1f4224c1ae049a7498ed30727515a892/src/scanoss/winnowing.py#L55-L61), but the list of extension is different. So it really seems as if SCANOSS should apply these ignores on the server side, so it's client agnostic.

I gave this some though again. To my understanding one of the main reasons for inventing the existing `irgnorePatterns` and applying it `on-the-fly` was to allow for changing the configuration (patterns) without invalididating the scan cache. I believe, this is also very desirable for a sniplet scanner. So, I propose 

1. to extend the `ScannerConfiguration` with a property for ignore patterns for sniplet scanners
2. and most importantly, provide an empty list of ignore patterns to ScanOss and then filter the things on the fly.

Would that make sense for you?

scanner/src/main/kotlin/scanners/scanoss/ScanOssConfig.kt
sschuberth: In the latest iteration I went for using the deprecated Java client's values as reference values, and the maintained Python client's values as default values. That way there is at least some sense in the reference values for those users who want to stick to the original values from the Java client, as they could easily copy them over.

sschuberth:
> allow for changing the configuration (patterns) without invalididating the scan cache. I believe, this is also very desirable for a sniplet scanner.

I'm actually not sure yet whether caching / storing scan results from snippet scanners has much benefit after all, because snippet scanners usually return results *much* faster than regular license scanners.

> 1. extend the `ScannerConfiguration` with a property for ignore patterns for sniplet scanners

I don't think we can lump together all snippet scanners to use a single configuration option. In fact, I was very surprised to see SCANOSS to ignore some files in first place, which is something I haven't seen with any snippet scanner before. After I had a chat with @eeisegn my understanding is that this is mainly a measure to avoid too many false-positives that would otherwise be triggered due to SCANOSS's specific way of calculating fingerprints (without taking also e.g. the directory structure into account). So these ignore patterns are a SCANOSS-specific "work-around" to reduce the number of false positives, which also reduces the load on SCANOSS's server.

> 2. and most importantly, provide an empty list of ignore patterns to ScanOss and then filter the things on the fly.

By "on the fly" you mean to filter after retrieving the scan result? While that makes sense WRT caching / the scan storage in general, it defeats the intention of the original SCANOSS client library to reduce the load on the server.

sschuberth:
> > I don't think we can lump together all snippet scanners to use a single configuration option.
> 
> I don't think a single configuration file is necessary or makes sense. Filtering the scan result on ORT side and disabling the scanner internal filtering does not require moving to a single configuration file at all.

Note that I wasn't talking about a "configuration *file*" but about a "configuration *option*" (in ORT's main configuration file, where scanner options are configured). Because you were saying:

> 1. extend the ScannerConfiguration with a property for ignore patterns for sniplet scanners

I'm reading "*a* property" like you're suggesting to introduce a single new configuration *option* that configures ignore patterns for *all* snippet scanners that may be used with ORT. And my argument simply was that due to the internals of each snippet scanner, it probably makes no sense to have a single ignore pattern for all of them.

sschuberth:
> I believe caching the results has in general more positive impact on the server load compared to using these exclude patterns without caching.

You're probably right here. In any case, we should never scan thise ignored files with SCANOSS, even if we'd cache the scan results.

> All in all I believe ORT's approach so far was to scan everything, no matter if it's a build script or whatever.

Right. This PR basically just reveals that this was not always the case: As soon as we use third-party libraries to communicate with a scanner, we cannot be sure what filtering takes place.

Originally, I was just trying to replicate the current behavior while removing a dependency on a class from the SCANOSS client library. So from a user perspective nothing has changed. And out of a sudden we're again in a debate on principles.

So, the question probably boils down to: How should be deal with scanner-specific ignore patterns, if we want to allow them at all?

fviernau:
> So, the question probably boils down to: How should be deal with scanner-specific ignore patterns, if we want to allow them at all?

Yes, fully agree. To my understanding, if we would answer that question with "we don't allow them at all", then this PR would be tremendously simplified - which is why I'd vote for answering that question first. Maybe this works good in slack with involving further devs?

sschuberth:
> To my understanding, if we would answer that question with "we don't allow them at all", then this PR would be tremendously simplified

For sure that would simplify things. However, I didn't even consider that option as 1) [as said](https://github.com/oss-review-toolkit/ort/pull/5467#issuecomment-1243808668) the goal of this PR was to not change behavior, and 2) we IMO can't achieve what we want with the single ignore patters list that we have, as there is no intersection between the files we want to ignore in general, and those that we want / need to ignore on a per-scanner basis.

For example, I believe we for sure want to scan files with a `.license` extension with ScanCode, but the SCANOSS client suggests to not scan these files. What's your proposal to deal with this then? If we don't have per-scanner ignore patterns, then the only thing we can do is to still scan `.license` files with SCANOSS, but then we can't filter out the results for these files are they should not be part of the general ignore patterns. Meaning we'll get SCANOSS scan results for those files, probably leading to false-positives (I mean, there must be a reason why the SCANOSS client ignores them).

sschuberth:
> I admit you're right that it does not make sense to scan certain file types, like e.g. `.license` files for sniplets.
> On the other hand I do like that ignore patterns can be changed easily, but maybe there isn't much value in that.
> Probably for sniplet scanners it's way less critical to not allow changing the patterns easily.

I just had another chat with @eeisegn on the general topic. Turns out there's actually two kinds of ignore patterns in the (Python) SCANOSS client:

1. There's [FILTERED_EXT](https://github.com/scanoss/scanoss.py/blob/f27ed92/src/scanoss/scanner.py#L54-L75) and [FILTERED_FILES](https://github.com/scanoss/scanoss.py/blob/f27ed92/src/scanoss/scanner.py#L76-L80). These files are ignored completely by default, however the client has an option to not make use of these filters and still [scan all extensions](https://github.com/scanoss/scanoss.py/blob/f27ed92/src/scanoss/scanner.py#L106). @eeisegn confirmed that no "harm" is done (on the server side or so) by *not* filtering out these files, except for probably an increased number of false positives.

2. There's a list of [extensions for supposedly binary files](https://github.com/scanoss/scanoss.py/blob/f27ed92fa4d54a63b2947a02311522e1ab6820d4/src/scanoss/winnowing.py#L55-L61). According to @eeisegn this list is core to the Winnowing algorithm and *should* be adhered to.

Taking all of this into account I'd propose the following: Do *not* implement any of 1. in the ORT scanner code that calls our own SCANOSS client library, but *do* implement 2., however not as part of the ORT scanner code that calls our own SCANOSS client library, but as part of the our own SCANOSS client library itself.

Would that make sense to you @eeisegn and @fviernau?

sschuberth:
I have moved out uncontroversial stuff to https://github.com/oss-review-toolkit/ort/pull/5790.

eeisegn:
> > I admit you're right that it does not make sense to scan certain file types, like e.g. `.license` files for sniplets.
> > On the other hand I do like that ignore patterns can be changed easily, but maybe there isn't much value in that.
> > Probably for sniplet scanners it's way less critical to not allow changing the patterns easily.
> 
> I just had another chat with @eeisegn on the general topic. Turns out there's actually two kinds of ignore patterns in the (Python) SCANOSS client:
> 
> 1. There's [FILTERED_EXT](https://github.com/scanoss/scanoss.py/blob/f27ed92/src/scanoss/scanner.py#L54-L75) and [FILTERED_FILES](https://github.com/scanoss/scanoss.py/blob/f27ed92/src/scanoss/scanner.py#L76-L80). These files are ignored completely by default, however the client has an option to not make use of these filters and still [scan all extensions](https://github.com/scanoss/scanoss.py/blob/f27ed92/src/scanoss/scanner.py#L147). @eeisegn confirmed that no "harm" is done (on the server side or so) by _not_ filtering out these files, except for probably an increased number of false positives.
> 2. There's a list of [extensions for supposedly binary files](https://github.com/scanoss/scanoss.py/blob/f27ed92fa4d54a63b2947a02311522e1ab6820d4/src/scanoss/winnowing.py#L55-L61). According to @eeisegn this list is core to the Winnowing algorithm and _should_ be adhered to.
> 
> Taking all of this into account I'd propose the following: Do _not_ implement any of 1. in the ORT scanner code that calls our own SCANOSS client library, but _do_ implement 2., however not as part of the ORT scanner code that calls our own SCANOSS client library, but as part of the our own SCANOSS client library itself.
> 
> Would that make sense to you @eeisegn and @fviernau?

Makes sense to me! Thanks!

sschuberth:
Superseded by https://github.com/oss-review-toolkit/ort/pull/8978.

PR #5468, opened 2022-06-20
GoMod: Handle the replace directive

### LINKED ISSUE #4445: GoMod: ORT should respect the `replace` directive
As described in this draft PR: https://github.com/oss-review-toolkit/ort/pull/4435

ORT does not respect the `replace` directive as it should.
Using a `go.mod` file like this:

```
module github.com/oss-review-toolkit/ort/gomod-synthetic-test-project

go 1.13

require (
	github.com/fatih/color v1.7.0
	github.com/hashicorp/go-multierror v1.0.0 // indirect
	github.com/mattn/go-colorable v0.1.4 // indirect
	github.com/mattn/go-isatty v0.0.10 // indirect
	golang.org/x/oauth2 v0.0.0-20190226205417-e64efc72b421 // indirect
	rsc.io/quote/v3 v3.1.0
)

replace rsc.io/quote/v3 => rsc.io/quote/v3 v3.0.0
```

Will still end up with `rsc.io/quote/v3 v3.1.0` listed in ORT's result file.

The reason for this is because `go mod vendor -v` and `go mod graph` is being used to find the dependencies.
Like @tsteenbe mentioned (https://github.com/oss-review-toolkit/ort/issues/3934#issuecomment-852299217), I think the correct solution would be to use `go list -deps -json ./...` to determine the dependencies, which might also enable us to remove the usage of `go mod why`.

fviernau:
Extend the logic for parsing the output for `go list -m all` to properly
handle lines for replaced modules [1]. Adjust the test asset to cover a
replacement. Also remove the mention of this limitation from the
documentation as replacements are now properly handled.

[1] https://go.dev/ref/mod#go-mod-file-replace

Fixes #4445.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -171,7 +171,9 @@ class GoMod(
 
         val list = run("list", "-m", "all", workingDir = projectDir)
         list.stdout.lines().forEach { line ->
-            val columns = line.split(' ')
+            // For replaced modules the line is formatted as "${orig-module} => ${replaced-module}", see also
+            // https://go.dev/ref/mod#go-mod-file-replace.
+            val columns = line.substringAfterLast(" => ").split(' ')
             if (columns.size != 2) return@forEach
 
             result += Identifier(type = managerName, namespace = "", name = columns[0], version = columns[1])
--- a/README.md
+++ b/README.md
@@ -326,8 +326,7 @@ supported:
   * [dep](https://golang.github.io/dep/)
   * [Glide](https://github.com/Masterminds/glide)
   * [Godep](https://github.com/tools/godep)
-  * [GoMod](https://github.com/golang/go/wiki/Modules) (limitations:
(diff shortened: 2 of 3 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -171,7 +171,9 @@ class GoMod(
 
         val list = run("list", "-m", "all", workingDir = projectDir)
         list.stdout.lines().forEach { line ->
-            val columns = line.split(' ')
+            // For replaced modules the line is formatted as "${orig-module} => ${replaced-module}", see also
+            // https://go.dev/ref/mod#go-mod-file-replace.
+            val columns = line.substringAfterLast(" => ").split(' ')
```
mnonnenmacher, line 176: While at it, you could make this more compact, similar to what I suggested in your other PR:
```kotlin
return list.stdout.lineSequence()
    .map { it.substringAfterLast(" => ").split(' ') }
    .filter { it.size == 2 }
    .map { Identifier(...) }
```
But I'll approve anyway as it's unrelated to the change itself.

mnonnenmacher (approved):

Issue #4127, opened 2021-06-08
Improve handling of module declaration for Golang packages


tsteenbe:
ORT currently throws a DownloadException for module declaration suffixes such as  `v2` or `v10` as it assumes them to be a part of the VCS path which they are not. Related Golang docs (https://golang.org/doc/modules/gomod-ref#module-syntax) is ambigious on whether module version should be a directory in the repo or not.

We can't just simply drop the `v[digit]` from the VCS path as their may be repositories that do use directories for each version module.

**Example go-playground/assert**

Id: GoMod::github.com/go-playground/assert/v2:v2.0.1
Package URL: pkg:golang/github.com%2Fgo-playground%2Fassert%2Fv2@v2.0.1
Repository Declared: https://github.com/go-playground/assert.git
Repository Processed: https://github.com/go-playground/assert.git

```
Could not download 'GoMod::github.com/go-playground/assert/v2:v2.0.1': DownloadException: Download failed for 'GoMod::github.com/go-playground/assert/v2:v2.0.1'. Suppressed: DownloadException: The Git working directory at '/tmp/ort-ScanCode13095570840845971637/GoMod/unknown/github.com%2Fgo-playground%2Fassert%2Fv2/v2.0.1' does not contain the requested path 'v2'. Suppressed: DownloadException: No source artifact URL provided for 'GoMod::github.com/go-playground/assert/v2:v2.0.1'.
```

**Example go-playground/validator**

Id: GoMod::github.com/go-playground/validator/v10:v10.4.1
Package URL: pkg:golang/github.com%2Fgo-playground%2Fvalidator%2Fv10@v10.4.1
Repository Declared: https://github.com/go-playground/validator.git
Repository Processed: https://github.com/go-playground/validator.git

```
Could not download 'GoMod::github.com/go-playground/validator/v10:v10.4.1': DownloadException: Download failed for 'GoMod::github.com/go-playground/validator/v10:v10.4.1'. Suppressed: DownloadException: The Git working directory at '/tmp/ort-ScanCode13095570840845971637/GoMod/unknown/github.com%2Fgo-playground%2Fvalidator%2Fv10/v10.4.1' does not contain the requested path 'v10'. Suppressed: DownloadException: No source artifact URL provided for 'GoMod::github.com/go-playground/validator/v10:v10.4.1'.
```

tsteenbe:
Workaround for this issue is to create a VCS curations in `curations.yml` to simply set the VCS path to empty e.g.

[...]

PR #5469, opened 2022-06-20
GoModFunTest: Illustrate how the VCS path is set for modules with '/vx' suffix

fviernau:
The source code may or may not be placed in a directory 'vx'. Illustrated both cases.
One of them can be improved later on by setting the VCS path to 'v2'.

See #4127.


analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
```diff
@@ -22,6 +22,8 @@ project:
       linkage: "PROJECT_STATIC"
     - id: "GoMod::github.com/fatih/color:v1.13.0"
       linkage: "PROJECT_STATIC"
+    - id: "GoMod::github.com/go-playground/assert/v2:v2.0.1"
```
mnonnenmacher, line 25: Question regarding the commit message, when you say "which illustrates the former" it means that for this library the source code is in a "v2" folder, and respectively that for the library added in the next commit it is not?
I got a bit confused, because in both cases ORT does not set the VCS path to "v2", so maybe you could make it a bit clearer, that what you illustrate is that ORT ignores that suffix (probably because you plan to fix this later on?).

mnonnenmacher (dismissed):

analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
```diff
@@ -38,8 +38,43 @@ project:
       linkage: "PROJECT_STATIC"
     - id: "GoMod::github.com/go-playground/assert/v2:v2.0.1"
       linkage: "PROJECT_STATIC"
-    - id: "GoMod::github.com/google/uuid:v1.0.0"
+    - id: "GoMod::github.com/google/uuid:v1.1.2"
```
mnonnenmacher, line 41: Commit message: Typo in "dependency" in title.

mnonnenmacher (changes requested):

fviernau:
The context of the PR was #4127 which has been superseeded by #5484. So, closing this PR for now as the approach for testing this case may change.

PR #5470, opened 2022-06-21
fossid-webapp: Fix deserialization error on `deleteScan` response

nnobelis:

This PR fixes the following exception:
```
Exception in thread "main" com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.lang.Integer` from Array value (token `JsonToken.START_ARRAY`)
 at [Source: (okhttp3.ResponseBody$BomAwareReader); line: 1, column: 49] (through reference chain: org.ossreviewtoolkit.clients.fossid.EntityResponseBody["data"]) 
```

### DIFF
--- a/clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
+++ b/clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
@@ -34,3 +34,5 @@ typealias MapResponseBody<T> = EntityResponseBody<Map<String, T>>
 typealias PolymorphicResponseBody<T> = EntityResponseBody<PolymorphicList<T>>
 
 class PolymorphicList<T>(data: List<T> = listOf()) : List<T> by data
+
+class PolymorphicInt(val value: Int?)
--- a/clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
+++ b/clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
@@ -23,7 +23,7 @@ class EntityResponseBody<T>(
     val operation: String? = null,
     val status: Int? = null,
     val message: String? = null,
-    val messageParameters: Array<String>? = null,
+    val messageParameters: Map<String, String> = emptyMap(),
     val error: String? = null,
 
     val data: T? = null
--- a/clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
+++ b/clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
@@ -75,6 +75,7 @@ interface FossIdRestService {
             enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
             addModule(
                 kotlinModule().addDeserializer(PolymorphicList::class.java, PolymorphicListDeserializer())
+                    .addDeserializer(PolymorphicInt::class.java, PolymorphicIntDeserializer())
             )
         }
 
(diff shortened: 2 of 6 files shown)


clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
```diff
@@ -23,7 +23,7 @@ class EntityResponseBody<T>(
     val operation: String? = null,
     val status: Int? = null,
     val message: String? = null,
-    val messageParameters: Array<String>? = null,
+    val messageParameters: Map<String, String> = emptyMap(),
```
sschuberth, line 26: Commit message: Please mention in which version of FossID this has changed in the API. (And why it's not necessary to support the older API anymore.)

clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
```diff
@@ -34,3 +34,5 @@ typealias MapResponseBody<T> = EntityResponseBody<Map<String, T>>
 typealias PolymorphicResponseBody<T> = EntityResponseBody<PolymorphicList<T>>
 
 class PolymorphicList<T>(data: List<T> = listOf()) : List<T> by data
+
```
sschuberth, line 37: Commit message:
- "When deleting *a* scan, ..."
- "... as *a* String ..."

clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
nnobelis: > Commit message: Please mention in which version of FossID this has changed in the API. (And why it's not necessary to support the older API anymore.)

There is a reason I didn't put this information: I am not sure when the change was done... I think version 21 but this is a guess, and FossID don't generally mention API changes in their release notes....

As for why this is necessary to drop the support for the older API, I can answer that our support of old versions is a "best effort" and this one in particular would be much effort for few returns...
This property comes in only special cases, for instance here when the scan does not exist.

What is important is to support current version IMO.

nnobelis: Ah I found the version in a internal release notes.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -123,6 +124,37 @@ interface FossIdRestService {
             }
         }
 
+        /**
+         * A class to modify the standard Jackson deserialization to deal with inconsistencies in responses
+         * sent by the FossID server.
+         * When deleting scan, FossId returns the scan id as String in the 'data' property of the response. If no scan
```
mnonnenmacher, line 130: "deleting **a** scan"

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -123,6 +124,37 @@ interface FossIdRestService {
...
+         * could be found, it returns an empty array.
+         * This custom deserializer streamLines the result: everything is converted to Int and empty array is converted
```
mnonnenmacher, line 132: "stream**l**ines"

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -123,6 +124,37 @@ interface FossIdRestService {
...
+         * to `null`. This deserializer also accepts primitive integers and array containing integers, which will
```
mnonnenmacher, line 133: "array**s**"

clients/fossid-webapp/src/test/kotlin/FossIdClientReturnTypeTest.kt
```diff
@@ -196,4 +197,10 @@ class FossIdClientReturnTypeTest : StringSpec({
             }
         }
     }
+
+    "When the scan to delete does not exist, no exception is thrown" {
+        service.deleteScan("", "", SCAN_CODE_1).shouldNotBeNull().run {
+            error shouldBe "Classes.TableRepository.row_not_found"
+        }
```
mnonnenmacher, line 204: You should not need run:
```kotlin
service.deleteScan("", "", SCAN_CODE_1) shouldNotBeNull {
  error shouldBe "Classes.TableRepository.row_not_found"
}
```

mnonnenmacher (changes requested):

clients/fossid-webapp/src/main/kotlin/EntityResponseBody.kt
sschuberth: > I put the version in the commit message.

Great, thank you.

> There is a reason I didn't put this information: I am not sure when the change was done...

If that's the case again in the future. you should explicitly mention that the version that introduced is unknown. That's just honest and better than saying nothing about it.

clients/fossid-webapp/src/test/kotlin/FossIdClientReturnTypeTest.kt
nnobelis: If you look at @sschuberth changes in the same file, this is intentional to prevent creating a dependency.

nnobelis:
@mnonnenmacher I fixed the test compile error.

PR #5471, opened 2022-06-21
Evaluated model: Strip dependency trees

oheger-bosch:
This PR addresses a problem with out-of-memory errors we face during report generation for large projects: For complex dependency graphs, the generation of the evaluated model for the WebApp reporter consumes a huge amount of memory, since the condensed graph gets unwrapped again.

The reporters dealing with the evaluated model now accept a new _deduplicateDependencyTree_ parameter. When set to *true* (the default is *false*) the dependency tree does no longer contain duplicate subtrees; when a duplicate subtree occurs, only the root node is inserted. This is similar to the way dependency trees are displayed by Gradle.

### DIFF
--- a/reporter/src/main/kotlin/reporters/WebAppReporter.kt
+++ b/reporter/src/main/kotlin/reporters/WebAppReporter.kt
@@ -33,7 +33,18 @@ import org.ossreviewtoolkit.reporter.reporters.evaluatedmodel.EvaluatedModel
 
 private const val PLACEHOLDER = "ORT_REPORT_DATA_PLACEHOLDER"
 
+/**
+ * A [Reporter] that generates a web application that allows browsing an ORT result interactively.
+ *
+ * This reporter supports the following options:
+ * - *deduplicateDependencyTree*: Controls whether subtrees occurring multiple times in the dependency tree are
+ *   stripped.
+ */
 class WebAppReporter : Reporter {
+    companion object {
+        const val OPTION_DEDUPLICATE_DEPENDENCY_TREE = "deduplicateDependencyTree"
+    }
+
     override val reporterName = "WebApp"
 
     private val reportFilename = "scan-report-web-app.html"
@@ -44,7 +55,7 @@ class WebAppReporter : Reporter {
         options: Map<String, String>
     ): List<File> {
         val template = javaClass.getResource("/scan-report-template.html").readText()
-        val evaluatedModel = EvaluatedModel.create(input)
+        val evaluatedModel = EvaluatedModel.create(input, options[OPTION_DEDUPLICATE_DEPENDENCY_TREE].toBoolean())
 
         val index = template.indexOf(PLACEHOLDER)
         val prefix = template.substring(0, index)
--- a/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
+++ b/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
@@ -148,7 +148,8 @@ data class EvaluatedModel(
             YAMLMapper().apply(mapperConfig).registerModule(IntIdModule(INT_ID_TYPES))
         }
 
-        fun create(input: ReporterInput): EvaluatedModel = EvaluatedModelMapper(input).build()
+        fun create(input: ReporterInput, deduplicateDependencyTree: Boolean = false): EvaluatedModel =
+            EvaluatedModelMapper(input).build(deduplicateDependencyTree)
     }
 
     /**
--- a/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
+++ b/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
@@ -24,6 +24,7 @@ import org.ossreviewtoolkit.model.CuratedPackage
 import org.ossreviewtoolkit.model.DependencyNode
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtIssue
+import org.ossreviewtoolkit.model.PackageLinkage
 import org.ossreviewtoolkit.model.Project
 import org.ossreviewtoolkit.model.Provenance
 import org.ossreviewtoolkit.model.RemoteArtifact
@@ -79,7 +80,11 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
 
     private val packageExcludeInfo = mutableMapOf<Identifier, PackageExcludeInfo>()
 
-    fun build(): EvaluatedModel {
+    /**
+     * Build an [EvaluatedModel] instance. If [deduplicateDependencyTree] is *true*, remove duplicate subtrees from
+     * the dependency tree. This may be necessary for huge projects to avoid excessive memory consumption.
+     */
+    fun build(deduplicateDependencyTree: Boolean = false): EvaluatedModel {
         createExcludeInfo()
         createScopes()
 
@@ -99,7 +104,7 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
 
         input.ortResult.analyzer?.result?.projects?.forEach { project ->
             val pkg = packages.getValue(project.id)
-            addDependencyTree(project, pkg)
+            addDependencyTree(project, pkg, deduplicateDependencyTree)
         }
 
         input.ortResult.analyzer?.result?.projects?.forEach { project ->
@@ -423,7 +428,29 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
         return actualScanResult
     }
 
-    private fun addDependencyTree(project: Project, pkg: EvaluatedPackage) {
+    private fun addDependencyTree(
+        project: Project,
+        pkg: EvaluatedPackage,
+        deduplicateDependencyTree: Boolean
+    ) {
+        val visitedNodes = mutableMapOf<Any, DependencyTreeNode>()
+
+        fun createDependencyNode(
+            dependency: EvaluatedPackage,
+            linkage: PackageLinkage,
+            issues: List<EvaluatedOrtIssue>,
+            children: List<DependencyTreeNode> = emptyList()
+        ) =
+            DependencyTreeNode(
+                linkage = linkage,
+                pkg = dependency,
+                scope = null,
+                children = children,
+                pathExcludes = emptyList(),
+                scopeExcludes = emptyList(),
+                issues = issues
+            )
+
         fun DependencyNode.toEvaluatedTreeNode(
             scope: EvaluatedScope,
             path: List<EvaluatedPackage>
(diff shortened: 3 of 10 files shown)


mnonnenmacher (changes requested):

reporter/src/main/kotlin/reporters/WebAppReporter.kt
```diff
@@ -34,6 +34,10 @@ import org.ossreviewtoolkit.reporter.reporters.evaluatedmodel.EvaluatedModel
 private const val PLACEHOLDER = "ORT_REPORT_DATA_PLACEHOLDER"
 
 class WebAppReporter : Reporter {
```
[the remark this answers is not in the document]
oheger-bosch: Done.  Since the class did not have a comment before, I needed to come up with a sentence describing the purpose of this reporter.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
```diff
@@ -148,7 +148,8 @@ data class EvaluatedModel(
             YAMLMapper().apply(mapperConfig).registerModule(IntIdModule(INT_ID_TYPES))
         }
 
-        fun create(input: ReporterInput): EvaluatedModel = EvaluatedModelMapper(input).build()
+        fun create(input: ReporterInput, deduplicateDependencyTree: Boolean = false): EvaluatedModel =
```
fviernau, line 151: Can you help me understanding this change with the following questions?

1. Does it prune redundant subtrees if they are from same scope?
2. Does it prune redundant subtrees if they are in different scope?
3. Is the logical tree affected or is it just the representation?
4. How fast is this code to execute?

fviernau (changes requested):

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
oheger-bosch: Hi Frank,

ad 1: Yes, redundant subtrees are pruned from the same scope.
ad 2: No, subtrees are not pruned over different scopes. So if a subtree appears multiple times in different scopes, each scope includes one copy.
ad 3: I am not exactly sure what you mean by "logical tree". Nodes are actually removed from the tree to reduce its size. Users must be aware that they do not see all levels of the tree in the WebApp report when they set this option.
ad 4: With this option enabled, the report generation may even run faster, since only parts of the dependency graph are traversed. For the projects that caused us problems, the reporter step failed after 30 to 40 minutes. With the option it is successful after < 3 minutes.

mnonnenmacher: To add one example where this change is required: We saw one Android project with many dependencies on Android core libraries where each spans a very large and highly redundant dependency tree. IIRC the largest scope had a dependency tree with more than 300,000 nodes, and over all scopes there were about 10,000,000 nodes. So even if the reporter would not eventually run out of memory, it would be impossible to load the generated web app report in any browser and render the dependency tree.

mnonnenmacher (dismissed):

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelReporter.kt
```diff
@@ -32,10 +32,14 @@ import org.ossreviewtoolkit.reporter.ReporterInput
  *
  * This reporter supports the following options:
  * - *output.file.formats*: The list of [FileFormat]s to generate, defaults to [FileFormat.JSON].
+ * - *deduplicateDependencyTree*: Controls whether subtrees occurring multiple times in the dependency tree are
+ *   .stripped
```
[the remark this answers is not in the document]
oheger-bosch: Copy and Paste error :-( Fixed now.

sschuberth (approved):
I'm fine with the change, but I wonder whether the feature to deduplicate the dependency tree should be moved to the `DependencyGraphNavigator` instead, so all report formats (or other navigators of the dependency tree) *could* make use of it.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
fviernau: Thanks @oheger-bosch for the clarifications!

With #3 I intended to ask whether it would make sense to actually do the propose change to the evaluated model while making the webapp still show the full tree. What do you think?

oheger-bosch:
> I'm fine with the change, but I wonder whether the feature to deduplicate the dependency tree should be moved to the `DependencyGraphNavigator` instead, so all report formats (or other navigators of the dependency tree) _could_ make use of it.

Interesting. How would this look like API-wise? Do you envision some kind of _visitor_ implementation that supports traversing either the full graph or the deduplicated one?

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
sschuberth: > while making the webapp still show the full tree.

Didn't @mnonnenmacher basically [already answer that](https://github.com/oss-review-toolkit/ort/pull/5471#discussion_r903077866)?

> it would be impossible to load the generated web app report in any browser and render the dependency tree.

I'm not sure how "smart" the rendering in the web app reporter is, though. Could be that collapsed parts of the tree / graph do not really take up much memory, and / or that duplicate subtrees could be rendered as "references" instead of "copies".

sschuberth:
> Do you envision some kind of _visitor_ implementation that supports traversing either the full graph or the deduplicated one?

Yes, something like that (a visitor or iterator), although I haven't thought about it thoroughly yet.

sschuberth:
> Yes, something like that (a visitor or iterator), although I haven't thought about it thoroughly yet.

Also other features, like navigating the graph only up to a certain depth, could then be implemented via such a pattern / arguments to a visitor / iterator.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
oheger-bosch: So you mean, the JavaScript code would be modified, so that it can render the full tree even if the model stores a compressed representation? This would be a good solution; however, I have no clue how much effort this would mean.

I had the idea to use a representation similar to the dependency graph. Maybe there are JavaScript libraries available that can render a graph with such a representation. This would probably look rather fancy. But I am no JavaScript expert and do not know which libraries could be used for this purpose.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
```diff
@@ -445,13 +455,27 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
...
             }
 
+            val children = visitDependencies { dependencies ->
+                dependencies.map { node ->
+                    if (!deduplicateDependencyTree || node.getInternalId() !in visitedNodes) {
+                        val childNode = node.toEvaluatedTreeNode(scope, path + dependency)
+                        visitedNodes += node.getInternalId() to childNode
+                        childNode
+                    } else {
+                        // If this node has already been visited, cut the subtree here. So only return the node
```
fviernau, line 465: I'm not sure how much this is just personal preference: to me it seems that swapping the code of ìf`and `else` blocks, inverting the 'if´-condition would be easier to read. IMO this would also allow dropping the comment "If this node has already been visited" as it's obvious from the condition then.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
sschuberth: > Maybe there are JavaScript libraries available that can render a graph with such a representation.

In this context it could make sense to look at https://github.com/snyk/dep-graph again. I don't see any rendering happening in there, though, but maybe there's a companion library to do so.

oheger-bosch:
> > Yes, something like that (a visitor or iterator), although I haven't thought about it thoroughly yet.
> 
> Also other features, like navigating the graph only up to a certain `maxDepth`, could then be implemented via such a pattern / arguments to a visitor / iterator.

Yes, I was thinking about this as well; since there is already now some duplication in code which traverses the dependency graph for various purposes.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
```diff
@@ -445,13 +455,27 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
...
+                        // without its children.
+                        val duplicateRoot = visitedNodes.getValue(node.getInternalId())
+                        duplicateRoot.associateWithScope(scope)
+                        duplicateRoot.copy(children = emptyList())
+                    }
+                }.toList()
+            }
```
fviernau, line 472: As I understand the deduplication is intended to not work across scopes (from another comment on the PR of yours). The state of `visitedNodes` seems to be kept across scopes. Is there need to clear the map inbetween processing the scopes?

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
```diff
@@ -445,13 +455,27 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
                 issues += addIssues(this.issues, EvaluatedOrtIssueType.ANALYZER, dependency, null, packagePath)
             }
 
+            val children = visitDependencies { dependencies ->
+                dependencies.map { node ->
+                    if (!deduplicateDependencyTree || node.getInternalId() !in visitedNodes) {
```
fviernau, line 460: How about introducing a variable for `node.getInternalId()`? (Currently, it get's called either once or twice.)

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
```diff
@@ -445,13 +455,27 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
...
+                        val childNode = node.toEvaluatedTreeNode(scope, path + dependency)
+                        visitedNodes += node.getInternalId() to childNode
```
fviernau, line 462: Do I understand it correctly that subtrees in the graph are already deduplicated (as the logic uses `node.getInternalId()` for de-duplication), so that it only prevents duplicating subtrees from the graph but does not de-duplicate subtrees from the graph?

fviernau (changes requested):

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
fviernau: > Didn't @mnonnenmacher basically [already answer that](https://github.com/oss-review-toolkit/ort/pull/5471#discussion_r903077866)?

@sschuberth - No, I was referring to what @oheger-bosch  writes below, e.g. "the JavaScript code would be modified, so that it can render the full tree even if the model stores a compressed representation"

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
oheger-bosch: Reverted the condition, but I decided to keep a part of the comment.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
oheger-bosch: Good catch! This has been fixed.

I also spotted one additional problem: Since the node was added to the `visitedNodes` map only after its children have been processed, it could be that duplicate nodes were missed. This has been changed as well, and the test has been adapted.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
oheger-bosch: Yes, in the graph all subtrees have already been deduplicated.

oheger-bosch:
Testing with a larger project revealed some problems. I need to investigate further...

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
fviernau: optional: maybe this would be an interresting note to add to the commit message?

oheger-bosch:
Tests have been successful with the current approach.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
oheger-bosch: The commit message has been extended accordingly.

fviernau: Thanks!

fviernau (approved):

PR #5473, opened 2022-06-21
AnalyzerConfiguration: Make the package managers case-insensitive

mnonnenmacher:
Make sure that package manager names used within `AnalyzerConfiguration` are case-insensitive. See the commit messages for details.

### DIFF
--- a/model/src/main/kotlin/config/AnalyzerConfiguration.kt
+++ b/model/src/main/kotlin/config/AnalyzerConfiguration.kt
@@ -57,4 +57,25 @@ data class AnalyzerConfiguration(
      * Configuration of the SW360 package curation provider.
      */
     val sw360Configuration: Sw360StorageConfiguration? = null
-)
+) {
+    /**
+     * A copy of [packageManagers] with case-insensitive keys.
+     */
+    private val packageManagersCaseInsensitive: Map<String, PackageManagerConfiguration>? =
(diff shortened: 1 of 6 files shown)


model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -51,10 +53,14 @@ data class AnalyzerConfiguration(
      * Package manager specific configurations. The key needs to match the name of the package manager class, e.g.
      * "NuGet" for the NuGet package manager.
      */
-    val packageManagers: Map<String, PackageManagerConfiguration>? = null,
+    @ConfigAlias("packageManagers")
+    private val packageManagersInternal: Map<String, PackageManagerConfiguration>? = null,
```
fviernau, line 57: Would it make sense to fail in case there are multiple configurations for the same package manager provided?
e.g. add a require block checking `packageManagersInternal` ?

fviernau (changes requested):

model/src/main/kotlin/config/AnalyzerConfiguration.kt
mnonnenmacher: Good idea, done.

fviernau (approved):

oheger-bosch (approved):

PR #5475, opened 2022-06-22
FreemarkerTemplateProcessor: Add rule violation filtering for the templates

MarcelBochtler:
See commit messages


reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -346,6 +346,14 @@ class FreemarkerTemplateProcessor(
                 violation.severity >= threshold && !isResolved && !isExcluded
             } ?: false
 
+        /**
+         * Return a list of [RuleViolation]s for which no
+         * [RuleViolationResolution][org.ossreviewtoolkit.model.config.RuleViolationResolution] is provided.
```
mnonnenmacher, line 351: Please add an import for `RuleViolationResolution` instead of adding the package here.

mnonnenmacher (changes requested):

Issue #5479, opened 2022-06-22
GoMod: Dependency tree is missing edges


fviernau:
The synthetic test project under './analyzer/src/funTest/assets/projects/synthetic/gomod' is missing the edges
1. ´github.com/mattn/go-colorable -> github.com/mattn/go-isatty´
2. ´github.com/mattn/go-isatty -> golang.org/x/sys:v0.0.0-20220610221304-9f5ed59c137d´

See also
´´´
cd ./analyzer/src/funTest/assets/projects/synthetic/gomod
go mod graph
```

sschuberth:
Before you're maybe hunting down another path: Could it be that these are OS-specific dependencies?

fviernau:
> Before you're maybe hunting down another path: Could it be that these are OS-specific dependencies?

I've updated the ticket description, as I just realized could have been mis-leading. Anyhow, any indication why these could be OS specific?

sschuberth:
> Anyhow, any indication why these could be OS specific?

It was just a wild guess from the words "tty" and "color" in the package names, as I've seen something similar in other projects before. Windows terminals cannot natively display ANSI colors, so CLI programs might need additional packages that simulate these on Windows.

PR #5482, opened 2022-06-23
GoMod: Fix the conversion function turning the graph into a tree

fviernau:
See individual commits.

Fixes: #5479.


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -370,23 +370,23 @@ private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>
...
-            val dependencyReferences = currentDependencies.mapTo(sortedSetOf()) {
-                getPackageReference(it, nextPredecessorNodes)
-            }
+        fun getPackageReference(id: Identifier, predecessorNodes: Set<Identifier> = emptySet()): PackageReference {
+            val dependencies = nodeMap.getValue(id).mapNotNull {
+                if (predecessorNodes.contains(it)) {
+                    null
+                } else {
+                    getPackageReference(it, predecessorNodes + id)
+                }
```
sschuberth, line 379: Could be something like

    mapNotNull { ids ->
    // ...
    predecessorNodes.takeUnless { it in ids }?.let { getPackageReference(ids, it + id) }

fviernau: I've made it a separate commit on top, to keep this commit only doing the revert.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -370,7 +370,7 @@ private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>
      * of the given [root] package.
      */
     fun toPackageReferenceForest(root: Identifier): SortedSet<PackageReference> {
-        fun getPackageReference(id: Identifier, predecessorNodes: Set<Identifier> = mutableSetOf()): PackageReference {
+        fun getPackageReference(id: Identifier, predecessorNodes: Set<Identifier> = emptySet()): PackageReference {
```
sschuberth, line 373: Nit: This could be squashed into the previous commit, the commit message would still fit.

mnonnenmacher:
@oheger-bosch Could you review this change?

oheger-bosch (approved):
It's been a while since I touched this code for the last time, so I am no longer deep in the logic here. But the changes look good to me; especially, since the adapted tests show that accidentally dropped dependencies are now correctly detected.

PR #5488, opened 2022-06-24
OrtConfiguration: Extract the storages from the `ScannerConfiguration`

bs-ondem:
Check commit messages for details.

### DIFF
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -332,13 +332,14 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license 
             .associateWith(OrtConfiguration::resolveStorage)
             .mapValues { createStorage(it.value) }
 
-        fun resolve(name: String): ScanStorage = requireNotNull(storages[name]) { "Could not resolve storage '$name'." }
+        fun resolve(name: String): List<ScanStorage> =
+            requireNotNull(storages[name]) { "Could not resolve storage '$name'." }
 
         val defaultStorage = createDefaultStorage()
 
-        val readers = config.scanner.storageReaders.orEmpty().map { resolve(it) }
+        val readers = config.scanner.storageReaders.orEmpty().flatMap { resolve(it) }
             .takeIf { it.isNotEmpty() } ?: listOf(defaultStorage)
-        val writers = config.scanner.storageWriters.orEmpty().map { resolve(it) }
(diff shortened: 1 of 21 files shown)


cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -329,15 +328,18 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license
...
+            .mapValues { createStorage(it.value) }
 
-        fun resolve(name: String): ScanStorage = requireNotNull(storages[name]) { "Could not resolve storage '$name'." }
+        fun resolve(name: String): List<ScanStorage> =
+            requireNotNull(storages[name]) { "Could not resolve storage '$name'." }
 
         val defaultStorage = createDefaultStorage()
 
-        val readers = config.scanner.storageReaders.orEmpty().map { resolve(it) }
+        val readers = config.scanner.storageReaders.orEmpty().map { resolve(it) }.flatten()
```
sschuberth, line 340: Would `flatMap` work here and below?

model/src/main/kotlin/utils/DatabaseUtils.kt
```diff
@@ -36,6 +36,11 @@ import org.ossreviewtoolkit.utils.ort.ORT_FULL_NAME
 import org.ossreviewtoolkit.utils.ort.log
 
 object DatabaseUtils {
+    /**
+     * This map holds the [HikariDataSource] based on the [PostgresConfiguration].
+     */
+    private val dataSources: MutableMap<PostgresConfiguration, Lazy<DataSource>> = mutableMapOf()
```
oheger-bosch, line 42: Does this have to be concurrent map, so that it can be used safely from different threads?

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -98,7 +105,22 @@ data class OrtConfiguration(
      */
     val notifier: NotifierConfiguration = NotifierConfiguration()
 ) {
+    init {
+        Companion.storages = storages
```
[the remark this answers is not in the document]
bs-ondem: Yes, I agree with you that this is a bit hacky, but if I implement the `resolveStorage()` as an instance method, I have to pass through the whole `OrtConfiguration`. This should be not the reason to hold storages in the companion objects, but it was the easiest one. What do you think about introducing a new data class, which has the `storages` and the `ScannerConfiguration` as properties?

bs-ondem:
I'm closing this for now, because I will separate the changes from this PR into different ones.

PR #5491, opened 2022-06-24
Gradle: Apply the version-catalog-update-plugin

sschuberth:
See [1].

[1]: https://github.com/littlerobots/version-catalog-update-plugin

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -42,6 +42,7 @@ plugins {
     alias(libs.plugins.ideaExt)
     alias(libs.plugins.kotlin)
     alias(libs.plugins.taskInfo)
+    alias(libs.plugins.versionCatalogUpdate)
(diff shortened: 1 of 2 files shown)


clients/github-graphql/src/main/kotlin/GitHubService.kt
```diff
@@ -73,7 +72,7 @@ class GitHubService private constructor(
          * Optionally, the [url] for the GitHub GraphQL endpoint can be configured.
          */
         fun create(token: String, url: URI = ENDPOINT): GitHubService {
-            val client = HttpClient(engineFactory = OkHttp) {
+            val client = HttpClient {
```
mnonnenmacher, line 75: What's the effect of this change, does it still use the OkHttp client or the Ktor client now?

PR #5492, opened 2022-06-24
fossid-webapp: Improvements to FossID scanner logging

nnobelis:


scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -444,6 +444,11 @@ class FossId internal constructor(
     ): String {
         // we ignore the revision because we want to do a delta scan
         val recentScans = scans.recentScansForRepository(url)
+
+        log.info {
+            "${recentScans.size} recent scans have been found."
+        }
```
mnonnenmacher, line 453: This would fit in one line.

mnonnenmacher (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -442,6 +442,9 @@ class FossId internal constructor(
         projectCode: String,
         projectName: String
     ): String {
+        if (url.contains("@")) {
```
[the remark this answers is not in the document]
sschuberth: Use Kotlin style: `"@" in url`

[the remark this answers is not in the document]
sschuberth: Also, could we simply use `String.isValidUri()` or even ` String.replaceCredentialsInUri()` here to strip credentials?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -442,6 +442,9 @@ class FossId internal constructor(
...
+            log.warn { "The URL should not contain credentials as its interaction with delta scans is unpredictable." }
+        }
```
sschuberth, line 447: Blank line after this one.

nnobelis: @sschuberth  Somehow I cannot reply to the question above...

I wanted to warn the user that it should not give credentials, because FossID scanner will add its own.

`isValidUri` is too generic, it won't allow me to display a specific warning message to this case I want to address.

Same for `replaceCredentialsInUri`: It will silently digest this case, leaving a URL that could be not valid without the given credentials, and won't output a warning making this case hard to debug.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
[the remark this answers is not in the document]
nnobelis: I wanted to warn the user that it should not give credentials, because FossID scanner will add its own.

isValidUri is too generic, it won't allow me to display a specific warning message to this case I want to address.

Same for replaceCredentialsInUri: It will silently digest this case, leaving a URL that could be not valid without the given credentials, and won't output a warning making this case hard to debug.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -444,6 +444,9 @@ class FossId internal constructor(
     ): String {
         // we ignore the revision because we want to do a delta scan
         val recentScans = scans.recentScansForRepository(url)
+
+        log.info { "${recentScans.size} recent scans have been found." }
```
fviernau, line 452: Consider (optional): "Found ${recentScans.size} scans."

It's a not shorter and doesn't start the sentence with a number.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -442,6 +442,10 @@ class FossId internal constructor(
         projectCode: String,
         projectName: String
     ): String {
+        if ("@" in url) {
```
fviernau, line 445: Would it make sense to add an issue to [ScanSummary](https://github.com/oss-review-toolkit/ort/blob/9ce597968e1fb270e0aa8c68f7b16b5f632d6606/model/src/main/kotlin/ScanSummary.kt#L75) in this case?

fviernau (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
fviernau: Oh just read @sschuberth's proposal to strip credentials. +1 for that instead

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: This would make sense.
However, this would complicate the implementation because we are in a function deep in the call tree, and it would be cumbersome to bubble the issues up the tree.
If it was an error case,  we could short circuit the logic with an exception.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -442,22 +442,31 @@ class FossId internal constructor(
         projectCode: String,
         projectName: String
     ): String {
+        val urlWithoutCredentials = if ("@" in url) {
```
sschuberth, line 445: Alternatively:

```
val urlWithoutCredentials = url.replaceCredentialsInUri()
if (urlWithoutCredentials != url) {
    log.warn { "The URL should not contain credentials as its interaction with delta scans is unpredictable." }
}
```

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -442,22 +442,31 @@ class FossId internal constructor(
...
         } else {
-            log.info { "Scan found for $url and revision $revision. Creating delta scan..." }
+            log.info { "Scan found for $urlWithoutCredentials and revision $revision. Creating delta scan..." }
             namingProvider.createScanCode(projectName, DeltaTag.DELTA)
         }
 
-        val newUrl = if (config.addAuthenticationToUrl) queryAuthenticator(url) else url
+        val newUrl = if (config.addAuthenticationToUrl) {
+            queryAuthenticator(urlWithoutCredentials)
+        } else urlWithoutCredentials
```
sschuberth, line 468: Formatting: `if` /  `else` should either be both in a single line, or both be blocks. I.e use

```
} else {
    urlWithoutCredentials
}
```

sschuberth (approved):

PR #5493, opened 2022-06-24
Analyzer: Automatically determine package manager order

mnonnenmacher:
This is the first step in generalizing the way that Pub uses to get Gradle and Cocoapods dependencies so that it can also be used for other package managers. To keep the diff small I will do it in several smaller PRs.

### DIFF
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -144,14 +144,18 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
     ): AnalyzerResult {
         val state = AnalyzerState(curationProvider)
 
+        val packageManagerDependencies = determinePackageManagerDependencies(managedFiles)
+
         runBlocking {
             managedFiles.entries.map { (manager, files) ->
+                val mustRunAfter = config.getPackageManagerConfiguration(manager.managerName)?.mustRunAfter?.toSet()
+                    ?: packageManagerDependencies[manager]?.mapTo(mutableSetOf()) { it.managerName }.orEmpty()
+
                 PackageManagerRunner(
                     manager = manager,
                     definitionFiles = files,
                     labels = labels,
-                    mustRunAfter = config.getPackageManagerConfiguration(manager.managerName)?.mustRunAfter?.toSet()
-                        .orEmpty(),
+                    mustRunAfter = mustRunAfter,
                     finishedPackageManagersState = state.finishedPackageManagersState,
                     onResult = { result -> state.addResult(manager, result) }
                 )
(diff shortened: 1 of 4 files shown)


analyzer/src/main/kotlin/PackageManagerDependencyResult.kt
```diff
@@ -0,0 +1,32 @@
...
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.analyzer
+
+data class PackageManagerDependencyResult(
+    /**
+     * The names of package managers that the package manager that provided this result must run before.
+     */
+    val mustRunBefore: Set<String>,
```
sschuberth, line 26: Why the decision to introduce `mustRunBefore` in addition to `mustRunAfter`? "A mustRunBefore B" can also be modeled as "B mustRunAfter A", but introduces a(nother) way to shoot yourself in the foot by creating contradicting dependencies.

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -144,14 +144,18 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
     ): AnalyzerResult {
         val state = AnalyzerState(curationProvider)
 
+        val packageManagerDependencies = determinePackageManagerDependencies(managedFiles)
+
         runBlocking {
             managedFiles.entries.map { (manager, files) ->
+                val mustRunAfter = config.getPackageManagerConfiguration(manager.managerName)?.mustRunAfter?.toSet()
+                    ?: packageManagerDependencies[manager]?.map { it.managerName }?.toSet().orEmpty()
```
sschuberth, line 152: Could be `mapTo(mutableSetOf())`.

analyzer/src/main/kotlin/PackageManagerDependencyResult.kt
mnonnenmacher: It's to give package managers the option to request both, e.g. in the only current use case Pub knows that it must run before Gradle, but Gradle should not have to know about Pub. This is because the requirement is Pub specific.
In the end `mustRunBefore` is translated to `mustRunAfter`, as can be configured by the user.
Note that I plan to introduce cycle detection in a later change to not end up in a deadlock, that's also the reason that the user can overwrite the detected order.

sschuberth: I still dislike somewhat that this now doesn't match the `ort.conf` configuration anymore 1:1. Should we then also introduce `mustRunBefore` in `ort.conf` to align?

mnonnenmacher: IMO no, because it's possible to configure all possible dependencies with the existing property, I don't see a reason to align the config with the internal API. I initially thought about letting the function just return a `Map<String, String>` which would represent `mustRunAfter`, but I preferred the explicit return type to not have yet another map that is open to interpretation.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -188,6 +189,22 @@ class Pub(
...
+        // the path to the Android SDK.
+        val gradle = managedFiles.keys.find { it.managerName == Gradle.Factory().managerName }
+        val mustRunBefore = if (gradle != null && findGradleDefinitionFiles(managedFiles.getValue(this)).isNotEmpty()) {
+            setOf(gradle.managerName)
+        } else {
+            emptySet()
+        }
+
+        return PackageManagerDependencyResult(mustRunBefore = mustRunBefore, mustRunAfter = emptySet())
+    }
```
sschuberth, line 206: @MarcelBochtler could you please review esp. this bit?

MarcelBochtler (approved):

PR #5494, opened 2022-06-24

mnonnenmacher:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Unmanaged.kt
+++ b/analyzer/src/main/kotlin/managers/Unmanaged.kt
@@ -69,17 +69,15 @@ class Unmanaged(
             vcsInfo == VcsInfo.EMPTY -> {
                 // This seems to be an analysis of a local directory that is not under version control, i.e. that is not
                 // a VCS working tree. In this case we have no change to get a version.
-                val projectDir = definitionFile
-
                 log.warn {
-                    "Analysis of local directory '$projectDir' which is not under version control will produce " +
+                    "Analysis of local directory '$definitionFile' which is not under version control will produce " +
                             "non-cacheable results as no version for the cache key can be determined."
                 }
 
                 Identifier(
                     type = managerName,
                     namespace = "",
-                    name = projectDir.name,
+                    name = definitionFile.name,
                     version = ""
                 )
             }
--- a/analyzer/src/main/kotlin/managers/Unmanaged.kt
+++ b/analyzer/src/main/kotlin/managers/Unmanaged.kt
@@ -68,7 +68,7 @@ class Unmanaged(
         val id = when {
             vcsInfo == VcsInfo.EMPTY -> {
                 // This seems to be an analysis of a local directory that is not under version control, i.e. that is not
-                // a VCS working tree. In this case we have no change to get a version.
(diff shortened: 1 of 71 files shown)


clients/github-graphql/src/test/kotlin/GitHubServiceTest.kt
```diff
@@ -48,6 +48,9 @@ import java.net.ServerSocket
 import java.net.URI
 import java.util.regex.Pattern
 
+import kotlinx.coroutines.Dispatchers
```
sschuberth, line 51: Commit message: Typo in "exectues"

analyzer/src/main/kotlin/managers/Unmanaged.kt
```diff
@@ -69,17 +69,15 @@ class Unmanaged(
             vcsInfo == VcsInfo.EMPTY -> {
                 // This seems to be an analysis of a local directory that is not under version control, i.e. that is not
                 // a VCS working tree. In this case we have no change to get a version.
-                val projectDir = definitionFile
-
                 log.warn {
-                    "Analysis of local directory '$projectDir' which is not under version control will produce " +
+                    "Analysis of local directory '$definitionFile' which is not under version control will produce " +
```
sschuberth, line 73: FYI: The variable was actually introduced intentionally to make it explicit that in this case not a *file* but a *dir* is being handled. But maybe the mention of "directory" in the warn log is enough.

clients/nexus-iq/src/main/kotlin/NexusIqService.kt
```diff
@@ -62,7 +62,7 @@ interface NexusIqService {
         /**
          * The JSON (de-)serialization object used by this service.
          */
-        val JSON = Json { ignoreUnknownKeys = true }
+        private val JSON = Json { ignoreUnknownKeys = true }
```
sschuberth, line 65: By convention these should stay public / be part of the public API, so consumers of the clients can use exactly the same serializer as the client code itself.

clients/oss-index/src/main/kotlin/OssIndexService.kt
```diff
@@ -45,7 +45,7 @@ interface OssIndexService {
         /**
          * The JSON (de-)serialization object used by this service.
          */
-        val JSON = Json.Default
+        private val JSON = Json.Default
```
sschuberth, line 48: See above.

clients/vulnerable-code/src/main/kotlin/VulnerableCodeService.kt
```diff
@@ -41,7 +41,7 @@ interface VulnerableCodeService {
         /**
          * The JSON (de-)serialization object used by this service.
          */
-        val JSON = Json { ignoreUnknownKeys = true }
+        private val JSON = Json { ignoreUnknownKeys = true }
```
sschuberth, line 44: See above.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Unmanaged.kt
mnonnenmacher: I would think so, as it's also explained in the code comment.

sschuberth (approved):

PR #5509, opened 2022-06-29
SPDX: Prevent duplicate packages for projects

oheger-bosch:
Refer to the single commits for further details.

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/spdx/project-xyz-with-sub-project-dependencies.spdx.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/spdx/project-xyz-with-sub-project-dependencies.spdx.yml
@@ -1,0 +1,43 @@
+SPDXID: "SPDXRef-DOCUMENT"
+spdxVersion: "SPDX-2.2"
+creationInfo:
+  created: "2020-07-23T18:30:22Z"
+  creators:
+  - "Organization: Example Inc."
+  - "Person: Thomas Steenbergen"
+  licenseListVersion: "3.9"
+name: "xyz-0.1.0"
+dataLicense: "CC0-1.0"
+documentNamespace: "http://spdx.org/spdxdocs/spdx-document-xyz"
+documentDescribes:
+- "SPDXRef-Package-xyz"
+externalDocumentRefs:
+- externalDocumentId: "DocumentRef-my-lib-8.88.8"
+  spdxDocument: "./libs/my-lib/package.spdx.yml"
+  checksum:
+    algorithm: "SHA1"
+    checksumValue: "2fb02799a2486623af054418736f6a30957e2fd1"
+- externalDocumentId: "DocumentRef-sub-xyz"
+  spdxDocument: "./projects/sub-project-xyz-with-inline-packages.spdx.yml"
+  checksum:
+    algorithm: "SHA1"
+    checksumValue: "6ffefacb845132a6425fd2006fb93e46f320d020"
+packages:
+- SPDXID: "SPDXRef-Package-xyz"
+  description: "Awesome product created by Example Inc."
+  copyrightText: "Copyright (C) 2020 Example Inc."
+  downloadLocation: "git+ssh://gitlab.example.com:3389/products/xyz.git@b2c358080011af6a366d2512a25a379fbe7b1f78"
+  filesAnalyzed: false
+  homepage: "https://example.com/products/xyz"
+  licenseConcluded:  "NOASSERTION"
+  licenseDeclared: "Apache-2.0 AND curl AND LicenseRef-Proprietary-ExampleInc"
+  name: "xyz"
+  versionInfo: "0.1.0"
+  originator: "Person: Thomas Steenbergen"
+relationships:
+- spdxElementId: "DocumentRef-my-lib-8.88.8:SPDXRef-Package-my-lib"
+  relationshipType: "RUNTIME_DEPENDENCY_OF"
+  relatedSpdxElement: "SPDXRef-Package-xyz"
+- spdxElementId: "DocumentRef-sub-xyz:SPDXRef-Package-sub-xyz"
+  relationshipType: "RUNTIME_DEPENDENCY_OF"
+  relatedSpdxElement: "SPDXRef-Package-xyz"
--- a/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
+++ b/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
@@ -32,6 +32,7 @@ import org.ossreviewtoolkit.model.Project
 import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.utils.DependencyGraphConverter
+import org.ossreviewtoolkit.utils.common.getDuplicates
 import org.ossreviewtoolkit.utils.ort.log
 
 class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvider = PackageCurationProvider.EMPTY) {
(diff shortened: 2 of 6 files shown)


analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -475,4 +476,28 @@ class SpdxDocumentFile(
...
     }
+
+    /**
+     * Create the final [PackageManagerResult] by making sure that packages are removed from [projectResults] that
+     * are also referenced as project dependencies.
+     */
+    override fun createPackageManagerResult(
+        projectResults: Map<File, List<ProjectAnalyzerResult>>
+    ): PackageManagerResult {
+        val projectIds = projectResults.flatMap { projResult -> projResult.value.map { it.project.id } }.toSet()
```
sschuberth, line 487: Could be `flatMapTo(mutableSetOf())`.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -475,4 +476,28 @@ class SpdxDocumentFile(
...
+
+        val filteredResults = projectResults.mapValues { entry ->
+            entry.value.map { projectResult ->
+                val projectReferences = projectResult.packages.filter { it.id in projectIds }.toSet()
```
sschuberth, line 491: Could be `filterTo(mutableSetOf())`.

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -40,7 +40,13 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
     private val issues = sortedMapOf<Identifier, List<OrtIssue>>()
     private val dependencyGraphs = sortedMapOf<String, DependencyGraph>()
 
-    fun build() = DependencyGraphConverter.convert(AnalyzerResult(projects, packages, issues, dependencyGraphs))
+    fun build(): AnalyzerResult {
+        require(projects.map { it.id }.toSet().intersect(packages.map { it.pkg.id }.toSet()).isEmpty()) {
```
sschuberth, line 44: Could be `mapTo(mutableSetOf())` in both places.

sschuberth: Maybe use `(projects.map { it.id } + packages.map { it.pkg.id }).getDuplicates()` to also get the actual set of packages that are also projects.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -475,4 +476,29 @@ class SpdxDocumentFile(
...
+
+    /**
+     * Create the final [PackageManagerResult] by making sure that packages are removed from [projectResults] that
+     * are also referenced as project dependencies.
+     */
+    override fun createPackageManagerResult(
+        projectResults: Map<File, List<ProjectAnalyzerResult>>
+    ): PackageManagerResult {
+        val projectIds =
+            projectResults.flatMapTo(mutableSetOf()) { projResult -> projResult.value.map { it.project.id } }
```
mnonnenmacher, line 488: We usually don't use abbreviations like `projResult` but always use the full names like `projectResult`. Also the naming in this case is a bit misleading as it's actually a map entry, I would suggest to use:
`{ (_, projectResult) -> projectResult.map { it.project.id } }`

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -475,4 +476,29 @@ class SpdxDocumentFile(
...
+
+        val filteredResults = projectResults.mapValues { entry ->
+            entry.value.map { projectResult ->
+                val projectReferences = projectResult.packages.filterTo(mutableSetOf()) { it.id in projectIds }
+                projectResult.takeIf { projectReferences.isEmpty() }
+                    ?: projectResult.copy(packages = (projectResult.packages - projectReferences).toSortedSet())
+                        .also {
+                            log.info { "Removing ${projectReferences.size} packages that are projects." }
+                            log.debug { projectReferences.joinToString { it.id.toCoordinates() } }
```
mnonnenmacher, line 497: IMO this could be merged into one debug statement, but I'm also fine if you want to keep it as is.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -40,7 +41,14 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
     private val issues = sortedMapOf<Identifier, List<OrtIssue>>()
     private val dependencyGraphs = sortedMapOf<String, DependencyGraph>()
 
-    fun build() = DependencyGraphConverter.convert(AnalyzerResult(projects, packages, issues, dependencyGraphs))
+    fun build(): AnalyzerResult {
+        val duplicateIds = (projects.map { it.id } + packages.map { it.pkg.id }).getDuplicates()
+        require(duplicateIds.isEmpty()) {
+            "AnalyzerResult contains packages that are also projects. Duplicates: '$duplicateIds'."
```
fviernau, line 47: Why is it guaranteed that the conflict is always between project and package. E.g. why could it not be between entities of same type?

analyzer/src/funTest/assets/projects/synthetic/spdx/project-xyz-with-sub-project-dependencies.spdx.yml
```diff
@@ -0,0 +1,43 @@
+SPDXID: "SPDXRef-DOCUMENT"
```
fviernau, line 1: commit-message: How is that "referencing" between projects relevant to this commit? Could you amend the commit message to explain the intended behavior?

(Is it so that if a SPDXDocument is referenced, then it has to be a package?)

fviernau (changes requested):

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
fviernau: De-duplication with sorted set relies solely on the Comparable interface? (not equals / hashcode?)

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
oheger-bosch: I would prefer keeping it this way. With out default log level being INFO, this makes sure that we get some hint about a potential problem in SPDX files, but the log is not too verbose.

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
oheger-bosch: Interesting point. Could it be a problem that the `compareTo()` implementation is not fully aligned with `equals()`? (Since it may return *true* for objects that are not equal.)

analyzer/src/funTest/assets/projects/synthetic/spdx/project-xyz-with-sub-project-dependencies.spdx.yml
oheger-bosch: I have reworded the commit message, no longer pointing out a specific reason for the problem, but also stating that there might be changes in the heuristic to detect packages or projects.

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
fviernau: TBH I'd need to research this for Kotlin, I only know the Java contracts of the functions...

sschuberth: > De-duplication with sorted set relies solely on the Comparable interface? (not equals / hashcode?)

Our `getDuplicates()` relies on `groupBy()` which in turn replies on deduplication of keys in a map, which AFAIK relies on `equals()`, not on `compareTo()`.

> TBH I'd need to research this for Kotlin, I only know the Java contracts of these functions...

There should be no different semantics between Kotlin and Java here...

mnonnenmacher: From the docs of `SortedSet`: 

> This is so because the Set interface is defined in terms of the equals operation, but a sorted set performs all element comparisons using its compareTo (or compare) method, so two elements that are deemed equal by this method are, from the standpoint of the sorted set, equal.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/SortedSet.html

sschuberth: But why are you referring to `SortedSet` at all here? The `map` operation turns the `SortedSet`s into `List`s.

mnonnenmacher: Because the initial question was if `projects` and `packages` could each contain duplicate identifiers. But since they are both `SortedSet`s, they can't. Therefore, if both are mapped to one list of identifiers, all duplicates must be between projects and packages.

oheger-bosch: So, if this issue is now clarified, is there anything else missing for an approval?

fviernau (approved):

mnonnenmacher (approved):

PR #5513, opened 2022-07-01
Implement mechanism for dependencies between package managers

mnonnenmacher:
Introduce a mechanism for dependencies between package managers. Please see the commit messages for details.

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/PackageManagerDependencyHandler.kt
+++ b/analyzer/src/main/kotlin/managers/utils/PackageManagerDependencyHandler.kt
@@ -1,0 +1,178 @@
+/*
+ * Copyright (C) 2022 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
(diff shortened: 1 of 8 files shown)


analyzer/src/main/kotlin/managers/utils/PackageManagerDependencyHandler.kt
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+
+    override fun identifierFor(dependency: ResolvableDependencyNode): Identifier = dependency.id
+
+    override fun issuesForDependency(dependency: ResolvableDependencyNode): Collection<OrtIssue> = dependency.issues
+
+    override fun linkageFor(dependency: ResolvableDependencyNode): PackageLinkage = dependency.linkage
+
+    fun resolvePackageManagerDependency(dependency: DependencyNode): List<ResolvableDependencyNode> =
```
oheger-bosch, line 100: This algorithm should be able to deal with nested dependencies to other projects/scopes recursively. How about cycles? Should there be a detection for cyclic dependencies between projects? Is this practically relevant?

oheger-bosch: > > This algorithm should be able to deal with nested dependencies to other projects/scopes recursively.
> 
> I think this should already be the case, as to my understanding the nodes returned by `dependenciesFor` will be handled recursively by the `DependencyGraphBuilder`.

Yes, this is my understanding as well.

> 
> > How about cycles? Should there be a detection for cyclic dependencies between projects? Is this practically relevant?
> 
> Isn't there already cycle prevention built into the `DependencyGraphBuilder`? In general, I currently cannot imagine a situation where such a cycle would be created between different package managers, so maybe we can wait until this actually becomes an issue. I think adding test cases for both scenarios would make sense, but I would prefer to do this in a separate PR.

Not sure whether the cycle detection in the builder is effective here, as `PackageManagerDependencyHandler` also calls itself recursively. But I agree that this is currently a pure theoretically construct and am fine with deferring this to later PRs.

oheger-bosch (dismissed):

oheger-bosch (approved):

PR #5514, opened 2022-07-02
PackageCuration: Fix version range check for versions without a patch

MarcelBochtler:
Using `isSatisfiedBy(String)` defaults to `SemverType.STRICT`, which
doesn't allow versions without a patch. Specify the `SemverType`
explicitly to avoid this issue.

While at it, also print a warning if a version or a version range cannot
be parsed.

### DIFF
--- a/model/src/main/kotlin/PackageCuration.kt
+++ b/model/src/main/kotlin/PackageCuration.kt
@@ -22,6 +22,10 @@ package org.ossreviewtoolkit.model
 import com.fasterxml.jackson.annotation.JsonProperty
 
 import com.vdurmont.semver4j.Requirement
+import com.vdurmont.semver4j.Semver
+
+import org.ossreviewtoolkit.utils.ort.log
+import org.ossreviewtoolkit.utils.ort.showStackTrace
 
 /**
  * Return true if this string equals the [other] string, or if either string is blank.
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/PackageCuration.kt
```diff
@@ -57,7 +61,18 @@ data class PackageCuration(
      * package with the given [identifier][pkgId].
      */
     private fun isApplicableIvyVersion(pkgId: Identifier) =
-        runCatching { Requirement.buildIvy(id.version).isSatisfiedBy(pkgId.version) }.getOrDefault(false)
+        runCatching {
+            // TODO: This check does not completely comply to the ivy version-matchers specification. E.g. the version
```
sschuberth, line 65: Nit: "Ivy"

model/src/main/kotlin/PackageCuration.kt
```diff
@@ -57,7 +61,18 @@ data class PackageCuration(
...
+            //       '1.0' does not satisfy the version range [1.0,2.0], see
+            //       https://github.com/vdurmont/semver4j/issues/67.
+            Requirement.buildIvy(id.version).isSatisfiedBy(Semver(pkgId.version, Semver.SemverType.LOOSE))
+        }.onFailure {
+            log.warn {
+                "Failed to check if package curation version '${id.version}' is applicable to package version " +
+                        "'${pkgId.version}' of package '${pkgId.toCoordinates()}'."
+            }
+            it.showStackTrace()
```
sschuberth, line 74: Nit: Blank line before this one.

mnonnenmacher (approved):

PR #5515, opened 2022-07-02
config-file-curations-yml: Document curations that cause the scanner to skip packages

MarcelBochtler:
Make this option more visible, as it can be used to fix other scanner
issues.

### DIFF
--- a/docs/config-file-curations-yml.md
+++ b/docs/config-file-curations-yml.md
@@ -46,6 +46,10 @@ These curations can be configured in a YAML file that is passed to the _analyzer
 amends the metadata provided by the packages themselves. This way, it is possible to fix broken VCS URLs or provide the
 location of source artifacts.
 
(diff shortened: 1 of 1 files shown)


docs/config-file-curations-yml.md
```diff
@@ -46,6 +46,9 @@ These curations can be configured in a YAML file that is passed to the _analyzer
 amends the metadata provided by the packages themselves. This way, it is possible to fix broken VCS URLs or provide the
 location of source artifacts.
 
+Hint: If the `concluded_license` _and_ the `authors` are curated, this package will be skipped during the `scan` step,
+as no more information from the scanner is required.
```
sschuberth, line 50: However, this is only true if [skipConcluded](https://github.com/oss-review-toolkit/ort/blob/01c598a04c5745dc7408f6f84b619bc62d13592e/model/src/main/resources/reference.conf#L100) is enabled.

sschuberth (changes requested):

PR #5519, opened 2022-07-04
helper-cli: Add helper merge command

porsche-rbieniek:
This pull request adds a helper CLI command for merging multiple analyzer results into one.

We have a use case where a larger monorepo is maintained by multiple project teams, each responsible for a specific subdirectory of the monorepo.

Each product team delivers a seperate analyzer result for its subdirectory and the analysts in charge need a way to merge the individual analyzer results into one analyzer result.


sschuberth:
Closing this because as once more described [here](https://github.com/oss-review-toolkit/ort/issues/4364#issuecomment-2809778242), it is technically not possible to merge different ORT (analyzer) result files while maintaining data from both sources in a meaningful / generic way.

PR #5521, opened 2022-07-04
AnalyzerResultBuilder: Extend check for packages that are projects

bennati:
Check also duplicates that have different IDs butt same VCS information.

Relates-to: #5369

### DIFF
--- a/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
+++ b/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
@@ -43,8 +43,8 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
 
     fun build(): AnalyzerResult {
         val duplicateIds = (projects.map { it.id } + packages.map { it.pkg.id }).getDuplicates()
-        val duplicateVcs = (projects.map { listOf(it.id.type,it.vcsProcessed) } +
-                            packages.map { listOf(it.pkg.id.type,it.pkg.vcsProcessed) }
+        val duplicateVcs = (projects.mapTo(mutableSetOf()) { listOf(it.id.type,it.vcsProcessed) } +
+                            packages.mapTo(mutableSetOf()) { listOf(it.pkg.id.type,it.pkg.vcsProcessed) }
         ).getDuplicates()
         require(duplicateIds.isEmpty() && duplicateVcs.isEmpty()) {
             "AnalyzerResult contains packages that are also projects. Duplicates: '{$duplicateVcs ?: duplicateVcs}'."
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -43,8 +43,11 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
 
     fun build(): AnalyzerResult {
         val duplicateIds = (projects.map { it.id } + packages.map { it.pkg.id }).getDuplicates()
-        require(duplicateIds.isEmpty()) {
-            "AnalyzerResult contains packages that are also projects. Duplicates: '$duplicateIds'."
+        val duplicateVcs = (projects.map { listOf(it.id.type,it.vcsProcessed) } +
+                            packages.map { listOf(it.pkg.id.type,it.pkg.vcsProcessed) }
```
mnonnenmacher, line 47: There are some situations where duplicate VCS are actually expected. For example, with some package managers like Gradle multiple projects can be defined in the same definition file (note that this is not yet supported, but the package manager API already allows it), so all projects would have the same VCS info. With the current implementation this would fail here. One solution could be to use `project.mapTo(mutableSetOf()) { ... }`, then only duplicates between projects and packages would be reported. Also for packages duplicate VCS are allowed, so this would have to be changed accordingly.

What is actually the reason to add this change? Can you provide an example where this check is required?

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -43,8 +43,11 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
 
     fun build(): AnalyzerResult {
         val duplicateIds = (projects.map { it.id } + packages.map { it.pkg.id }).getDuplicates()
```
fviernau, line 45: commit-msg: butt -> but

fviernau (changes requested):

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
fviernau: I believe that for duplicate VCSes there exists valid cases for each, packages and projects: So, IMO there should be no duplicate check for VCS, but curious about the example / use case of @bennati ...

fviernau (changes requested):

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
bennati: The use case is to find occurrences of https://github.com/oss-review-toolkit/ort/pull/5369

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
bennati: fixed

sschuberth:
> Relates-to: #5369

Similar to my question [over there](https://github.com/oss-review-toolkit/ort/pull/5369#issuecomment-1344136867), is this PR still needed @bennati?

bennati:
No, we can close this

Issue #3599, opened 2021-02-05
Add Google Open Source Vulnerabilities as an advisor


sschuberth:
See https://osv.dev/. Note that https://github.com/google/osv is ~not live yet at the time of writing~ live now.

sschuberth:
On a related note, we could also add a reporter for the [Open Source Vulnerability format](https://ossf.github.io/osv-schema/).

tsteenbe:
Related https://github.com/goneall/spdx-to-osv which is a tool to produce an Open Source Vulnerability JSON file based on information in an SPDX document

fviernau:
Some issues I found during development:

1. https://github.com/google/osv.dev/issues/497
2. https://github.com/google/osv.dev/issues/502

fviernau:
I've filed a follow-up ticket to extend ORT's vulnerability model: #5593.

fviernau:
Besides the linked changes above I've improved the identifiers of Go packages in:

1. https://github.com/oss-review-toolkit/ort/pull/5586
2. https://github.com/oss-review-toolkit/ort/pull/5546

Topics which came out of working on this ticket are further general advisor (and related areas) improvements:

1. https://github.com/oss-review-toolkit/ort/issues/5593
2. https://github.com/oss-review-toolkit/ort/issues/5597

I consider the initial implementation to be done, with https://github.com/oss-review-toolkit/ort/pull/5595 be the last small fix, so closing this ticket as done.

PR #5522, opened 2022-07-04
clients: Implement the "Open Source Vulnerability format" data model

fviernau:
Add a new client module which implements a client for the Google Open
Source Vulnerabilities API. The implementation corresponds to [1],
version 1.3.0 (March 24, 2022), and also [2] which is referenced from
[1].

The test assets are exact copies of all the examples contained in [1].
The test verify that serialization and deserialization works for these
examples.

[1] https://ossf.github.io/osv-schema/
[2] https://github.com/ossf/osv-schema/blob/11524982426be469795b9c684ba340c5c90895d0/validation/schema.json

See also #3599.


tsteenbe (dismissed):
@fviernau  Please include in this change an update to https://github.com/oss-review-toolkit/ort/blob/main/.reuse/dep5 to add assets/vulnerability/examples with "Copyright 2021-2022 OSV Schema Authors".

In the commit messages I would personally also change "Version 1.3.0" to "version 1.3.0" as it's used in a sentence.

Why do you write "which is supposed to implement"? Why not simply "implements"?

clients/osv/src/test/assets/vulnerability/examples/3.json
```diff
@@ -0,0 +1,75 @@
...
+        }
+    ],
+    "database_specific": {
+        "cwe_ids": [
+            "CWE-327"
+        ],
+        "severity": "MODERATE",
+        "github_reviewed": true
+    }
+}
```
tsteenbe, line 75: Missing newline

clients/osv/src/test/assets/vulnerability/examples/4.json
```diff
@@ -0,0 +1,36 @@
...
+                        },
+                        {
+                            "fixed": "c43455749b914feef56b178b256f29b3016146eb"
+                        }
+                    ]
+                }
+            ]
+        }
+    ]
+}
```
tsteenbe, line 36: Missing newline at end of file.

clients/osv/src/test/assets/vulnerability/examples/7.json
```diff
@@ -0,0 +1,78 @@
+{
```
tsteenbe, line 1: Change indentation of file to 2 spaces. This is what seems to used in ORT based on search for JSON files, see for example 

- https://github.com/oss-review-toolkit/ort/blob/main/integrations/schemas/package-configuration-schema.json
- https://github.com/oss-review-toolkit/ort/blob/main/scanner/src/test/assets/scancode-with-issues.json

fviernau:
> Why do you write "which is supposed to implement"? Why not simply "implements"?

because the implementation isn't finished yet. I'll change it anyway.

clients/osv/src/test/assets/vulnerability/examples/3.json
fviernau: Done.

clients/osv/src/test/assets/vulnerability/examples/4.json
fviernau: Done.

clients/osv/src/test/assets/vulnerability/examples/7.json
fviernau: I deliberately choose to keep these files in the original state, so that it's easy to update them in the future and to see that they are identical with the original ones. Fine?

tsteenbe (approved):

clients/osv/src/main/kotlin/Serializers.kt
```diff
@@ -0,0 +1,81 @@
...
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.jsonPrimitive
+
+/**
+ * Use a custom serializer in order to map the original data structure to a more strict and simple, enum based, model.
+ */
+internal object EventSerializer : KSerializer<Event> {
+    override val descriptor: SerialDescriptor
+        get() = PrimitiveSerialDescriptor(javaClass.canonicalName, PrimitiveKind.STRING)
```
sschuberth, line 43: You should be able to use the `@Serializer(Event::class)` annotation on the class to avoid the need for these lines.

clients/osv/src/test/kotlin/ModelTest.kt
```diff
@@ -0,0 +1,64 @@
...
+
+            normalizeJson(serializedVulnerabilityJson) shouldBe normalizeJson(vulnerabilityJson)
+        }
+    }
+})
+
+private val JSON = Json {}
+
+private fun getVulnerabilityExamplesJson(): List<String> =
+    (1..1).map { i -> File("src/test/assets/vulnerability/examples/$i.json").readText() }
```
sschuberth, line 49: The range should probably go up to 7, right?

sschuberth (dismissed):

clients/osv/src/main/kotlin/Serializers.kt
```diff
@@ -0,0 +1,81 @@
...
+        require(element.entries.size == 1)
+        val (key, value) = element.entries.first()
+        val type = enumValueOf<Event.Type>(key.uppercase())
+
+        return Event(type, value.jsonPrimitive.content)
+    }
+
+    override fun serialize(encoder: Encoder, value: Event) {
+        val output = encoder as? JsonEncoder
+            ?: throw SerializationException("This class can be encoded only by Json format")
```
mnonnenmacher, line 60: The message makes it sound like the problem is with the value that shall be encoded. Do you mean "This serializer can only work with the JSON format"?

clients/osv/src/main/kotlin/Serializers.kt
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Use a custom serializer in order to map the original data structure to a more strict and simple, enum based, model.
+ */
+internal object EventSerializer : KSerializer<Event> {
+    override val descriptor: SerialDescriptor
+        get() = PrimitiveSerialDescriptor(javaClass.canonicalName, PrimitiveKind.STRING)
+
+    override fun deserialize(decoder: Decoder): Event {
+        require(decoder is JsonDecoder)
```
mnonnenmacher, line 46: Add a message similar to the one below in `serialize`.

mnonnenmacher (changes requested):

clients/osv/src/main/kotlin/Model.kt
```diff
@@ -0,0 +1,148 @@
...
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonObject
+
+/**
+ * Implementation of the "Open Source Vulnerability format" according to schema version 1.3.0 (March 24, 2022), see
+ * https://ossf.github.io/osv-schema/ which links to
+ * https://github.com/ossf/osv-schema/blob/11524982426be469795b9c684ba340c5c90895d0/validation/schema.json which was
+ * used as a reference.
```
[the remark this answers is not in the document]
fviernau: It was manually created.

clients/osv/src/test/kotlin/ModelTest.kt
fviernau: Good catch!

clients/osv/src/test/kotlin/ModelTest.kt
```diff
@@ -0,0 +1,64 @@
...
+
+            normalizeJson(serializedVulnerabilityJson) shouldBe normalizeJson(vulnerabilityJson)
+        }
+    }
+})
+
+private val JSON = Json {}
+
+private fun getVulnerabilityExamplesJson(): List<String> =
+    (1..7).map { i -> File("src/test/assets/vulnerability/examples/$i.json").readText() }
```
sschuberth, line 49: In order to not having to maintain this, how about just listing and iterating over all JSON files in the "src/test/assets/vulnerability/examples" directory?

fviernau: I had originally implemented what you propose, but then changed to this approach.
It has a little bit duplicate book keeping and ensures that tests don't get disabled by accident unnoticed.
MINOR: Furthermore it was easier to hard code a fixed asset (1..1) for development.

Can we keep it?

mnonnenmacher (approved):

clients/osv/src/test/kotlin/ModelTest.kt
sschuberth: > It has a little bit duplicate book keeping which ensures that tests don't get disabled by accident unnoticed.

Well, as you saw during the review using a number range does *not* prevent tests from getting disabled by accident, so I'm not convinced by that argument.

> Can we keep it?

TBH I just don't like it, and it's not in line with how we do it e.g. in [ExamplesFunTest](https://github.com/oss-review-toolkit/ort/blob/1ca71eb3c2ab8b8379eb9f387b13d0b66e276b83/cli/src/funTest/kotlin/ExamplesFunTest.kt#L66-L70), but I won't block it, esp. as you already have @mnonnenmacher's approval.

PR #5526, opened 2022-07-06
clients/osv: Implement querying the OSV API

fviernau:
Add a retrofit based REST API client and a wrapper for simplicity and
information hiding. The implementation is based on [1], even though it
has some inconsistencies with the real behavior, see [2].

[1] https://osv.dev/docs/
[2] https://github.com/google/osv.dev/issues/466

This is an enabler for #3599.

### DIFF
--- a/clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
+++ b/clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
@@ -1,0 +1,117 @@
+/*
+ * Copyright (C) 2022 EPAM Systems, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import io.kotest.assertions.json.shouldEqualJson
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
+import io.kotest.matchers.result.shouldBeSuccess
+import io.kotest.matchers.shouldBe
+
+import java.io.File
+
+import kotlinx.serialization.encodeToString
+
+import org.ossreviewtoolkit.clients.osv.OsvApiClient
+import org.ossreviewtoolkit.clients.osv.OsvService
+import org.ossreviewtoolkit.clients.osv.Package
+import org.ossreviewtoolkit.clients.osv.VulnerabilitiesForPackageRequest
+
+private val VULNERABILITY_FOR_PACKAGE_BY_COMMIT_REQUEST = VulnerabilitiesForPackageRequest(
+    commit = "6879efc2c1596d11a6a6ad296f80063b558d5e0f"
+)
+private val VULNERABILITY_FOR_PACKAGE_BY_NAME_AND_VERSION = VulnerabilitiesForPackageRequest(
+    pkg = Package(
+        name = "jinja2",
+        ecosystem = "PyPI"
+    ),
+    version = "2.4.1"
+)
+private val VULNERABILITY_FOR_PACKAGE_BY_INVALID_COMMIT_REQUEST = VulnerabilitiesForPackageRequest(
+    commit = "6879efc2c1596d11a6a6ad296f80063b558d5e0c"
+)
+
+class OsvServiceFunTest : StringSpec({
+    "getVulnerabilitiesForPackage() returns the expected vulnerability when queried by commit" {
+        val expectedResult = getAssetAsString("vulnerabilities-by-commit-expected-result.json")
+
+        val result = OsvService().getVulnerabilitiesForPackage(VULNERABILITY_FOR_PACKAGE_BY_COMMIT_REQUEST)
+
+        result.shouldBeSuccess {
+            JSON.encodeToString(it) shouldEqualJson expectedResult
+        }
+    }
+
+    "getVulnerabilitiesForPackage() returns the expected vulnerability when queried by name and version" {
+        val expectedResult = getAssetAsString("vulnerabilities-by-name-and-version-expected-result.json")
+
+        val result = OsvService().getVulnerabilitiesForPackage(VULNERABILITY_FOR_PACKAGE_BY_NAME_AND_VERSION)
+
+        result.shouldBeSuccess {
+            JSON.encodeToString(it) shouldEqualJson expectedResult
+        }
+    }
+
+    "getVulnerabilityIdsForPackages() return the vulnerability IDs for the given batch request" {
+        val requests = listOf(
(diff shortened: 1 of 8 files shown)


clients/osv/src/main/kotlin/OsvService.kt
```diff
@@ -0,0 +1,124 @@
...
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+
+/**
+ * This class wraps the OSV Rest API client in order to make its use simpler and less error-prone.
+ */
+class OsvService(val serverUrl: String = SERVER_URL_DEFAULT_VALUE, httpClient: OkHttpClient? = null) {
+    companion object {
+        const val SERVER_URL_DEFAULT_VALUE = OsvApiClient.SERVER_URL_PRODUCTION
```
sschuberth, line 38: Why introduce this alias here?

clients/osv/src/main/kotlin/OsvService.kt
```diff
@@ -0,0 +1,124 @@
...
+import okhttp3.OkHttpClient
+
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+
+/**
+ * This class wraps the OSV Rest API client in order to make its use simpler and less error-prone.
+ */
+class OsvService(val serverUrl: String = SERVER_URL_DEFAULT_VALUE, httpClient: OkHttpClient? = null) {
```
sschuberth, line 36: `val` can be dropped here.

clients/osv/src/main/kotlin/OsvService.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    private val client = OsvApiClient.create(serverUrl, httpClient)
+
+    /**
+     * Get the vulnerabilities for the package matching the given [request].
+     */
+    fun getVulnerabilitiesForPackage(
+        request: VulnerabilitiesForPackageRequest
+    ): Result<List<Vulnerability>> {
+        val response = client.getVulnerabilitiesForPackage(request).execute()
```
sschuberth, line 49: BTW, in this line it occurred to me that it probably would have been nicer if you had implemented the API by making use of [Retrofit's suspend functions](https://proandroiddev.com/suspend-what-youre-doing-retrofit-has-now-coroutines-support-c65bd09ba067#1cb3), also see the other client implementations.

sschuberth (changes requested):

clients/osv/src/main/kotlin/OsvService.kt
fviernau: Can you explain in what way that would have been nicer?

clients/osv/src/funTest/kotlin/OsvServiceTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        ecosystem = "PyPI"
+    ),
+    version = "2.4.1"
+)
+private val VULNERABILITY_FOR_PACKAGE_BY_INVALID_COMMIT_REQUEST = VulnerabilitiesForPackageRequest(
+    commit = "6879efc2c1596d11a6a6ad296f80063b558d5e0c"
+)
+
+class OsvServiceTest : StringSpec({
+    "getVulnerabilitiesForPackage()  returns the expected vulnerability when queried by commit" {
```
sschuberth, line 51: Collapse the double-space into one.

clients/osv/src/funTest/kotlin/OsvServiceTest.kt
```diff
@@ -0,0 +1,113 @@
+/*
```
sschuberth, line 1: This file should be called "...*Fun*Test...".

clients/osv/src/funTest/kotlin/OsvServiceTest.kt
```diff
@@ -0,0 +1,113 @@
...
+)
+
+class OsvServiceTest : StringSpec({
+    "getVulnerabilitiesForPackage()  returns the expected vulnerability when queried by commit" {
+        val expectedResult = getAssetAsString("vulnerabilities-by-commit-expected-result.json")
+
+        val result = OsvService().getVulnerabilitiesForPackage(VULNERABILITY_FOR_PACKAGE_BY_COMMIT_REQUEST)
+
+        result should beSuccess()
+        JSON.encodeToString(result.getOrNull()) shouldEqualJson expectedResult
```
sschuberth, line 57: Now you could combine these to

        result.shouldBeSuccess {
            JSON.encodeToString(it) shouldEqualJson expectedResult
        }

clients/osv/src/funTest/kotlin/OsvServiceTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        val ids = setOf("GHSA-xvch-5gv4-984h", "PYSEC-2014-82")
+
+        val result = OsvService().getVulnerabilitiesForIds(ids)
+
+        result should beSuccess()
+        result.getOrThrow().map { it.id } shouldContainExactlyInAnyOrder ids
+    }
+})
+
+private val JSON = Json { prettyPrint = true }
```
sschuberth, line 111: With the Kotest JSON matchers you don't need the pretty printing anymore and can straight use `OsvApiClient.JSON`.

sschuberth: Same goes for the previous commit, BTW.

clients/osv/src/main/kotlin/OsvApiClient.kt
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * A rest API client for the Google Open Source Vulnerabilities API (OSV.dev), see https://osv.dev/.
+ */
+internal interface OsvApiClient {
+    companion object {
+        const val SERVER_URL_PRODUCTION = "https://api.osv.dev"
+        const val SERVER_URL_STAGING = "https://api-staging.osv.dev"
+        const val BATCH_REQUEST_MAX_SIZE = 1000
+        val JSON = Json { }
```
sschuberth, line 49: Use `val JSON = Json.Default` instead to resolve and inspection hint. And I'd also add a blank line before this one.

clients/osv/src/main/kotlin/OsvService.kt
sschuberth: So how about making `OsvApiClient` public then instead? Consumers of the client library might be interested in using that low-level API directly.

sschuberth (changes requested):

clients/osv/src/test/kotlin/ModelTest.kt
```diff
@@ -41,7 +41,7 @@ class ModelTest : StringSpec({
     }
 })
 
-private val JSON = Json {}
+private val JSON = Json.Default
```
sschuberth, line 44: I'd prefer to have this inlined now that this doesn't create a new instance.

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+        val result = OsvService().getVulnerabilitiesForIds(ids)
+
+        result.shouldBeSuccess {
+            it.map { it.id } shouldContainExactlyInAnyOrder ids
+        }
+    }
+})
+
+private val JSON = OsvApiClient.JSON
```
sschuberth, line 115: I'd prefer to have this inlined.

sschuberth (approved):

clients/osv/src/test/kotlin/ModelTest.kt
fviernau: Would you agree moving the JSON constant from ´OsvApiClient` to `Model.kt` ?

sschuberth: Actually no, I would not agree, because IMO the model should not know anything about its concrete representation in a file format.

PR #5527, opened 2022-07-06
PackageCurationData: Do not apply already applied curations

mnonnenmacher:
If a curation has already been applied to a package, do not apply it
again.

### DIFF
--- a/model/src/main/kotlin/PackageCurationData.kt
+++ b/model/src/main/kotlin/PackageCurationData.kt
@@ -111,6 +111,8 @@ data class PackageCurationData(
      * non-null values of the curation data, and return the curated package.
      */
     fun apply(targetPackage: CuratedPackage): CuratedPackage {
+        if (this in targetPackage.curations.map { it.curation }) return targetPackage
+
         val original = targetPackage.pkg
 
         val vcsProcessed = vcs?.let {


model/src/main/kotlin/PackageCurationData.kt
```diff
@@ -111,6 +111,8 @@ data class PackageCurationData(
      * non-null values of the curation data, and return the curated package.
      */
     fun apply(targetPackage: CuratedPackage): CuratedPackage {
+        if (this in targetPackage.curations.map { it.curation }) return targetPackage
```
sschuberth, line 114: The rationale for this change might seem straight forward at first, but then I was wondering whether there *are* actually valid use cases for applying the same curation multiple times. Constructing a hypothetical case: Maybe a curation is done in a source you don't have control over, then the curation is undone in another source you also don't have control over, but then you want to redo the original curation in a source you *do* have control over. Does such a scenario make sense?

mnonnenmacher: I think this is very hypothetical, if we ever have this issue it might the right time to introduce a source property for curations so that they would not be equal anymore. However, I figured out that the root cause for the duplicated curations I saw is an issue with ClearlyDefined and I will create a separate fix for this. So we could either merge this as is or close it, depending on how relevant you think the described use case is.

sschuberth: I don't believe my use-case is very relevant, but I also wouldn't use that as the basis for the decision. Instead, I'd ask how relevant *your* use-case to get rid of duplicates is. And my understanding is now that it's not relevant at all, as it will be fixed differently. Which means having this code is (currently) of no use. Thus I'd close this PR just for simplicity.

mnonnenmacher:
Replaced by #5537.

PR #5528, opened 2022-07-07
Analyzer: Rewrite support for Yarn2+

nnobelis:
The previous design, while working on simple projects, was not
compatible with complex projects relying on Yarn workspaces : The
'node_modules' linker was creating symbolic links pointing to symbolic
links, therefore walking such file hierarchies crashed ORT.
This commit changes completely the design of the package manager. It
tries to use as much as possible Yarn wrapper commands to build the
dependency graph.
First, `yarn info` is called to get the dependency tree of the project.
While providing a lot of information, it unfortunately does not contain
package details and dependency scope information.
To alleviate the former problem, `yarn npm info` is then called with all
packages and details are fetched from the registry.
Regarding the lack of dependency scope information, the project
'package.json' files are parsed to find out the scope of project
dependencies. Please note that only project dependencies have their
scope qualified. Dependencies of dependencies are always falling in the
'dependencies' scope.
Another shortcoming is that Yarn 2+ does not provide details of a
project : thus, `npm list` has to be run to get this information.

A functional test showcasing support of workspaces will be added in a
future commit.


sschuberth:
> The 'node_modules' linker was creating symbolic links pointing to symbolic links, therefore walking such file hierarchies crashed ORT.

Can you elaborate a bit more on this? Walking symlinks that point to other symlinks should not crash ORT, unless maybe cycles are created.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -63,20 +93,56 @@ class Yarn2(
      */
     private val yarn2ExecutablesByPath: MutableMap<File, String> = mutableMapOf()
 
-    /**
-     * A NPM package manager, used to call NPM when Yarn 2+ does not provide enough information.
-     */
-    private val delegateNpmAnalyzer = Npm("npm", analysisRoot, analyzerConfig, repoConfig)
+    private val disableRegistryCertificateVerification = analyzerConfig
+        .getPackageManagerConfiguration(managerName)?.options
+        ?.get(OPTION_DISABLE_REGISTRY_CERTIFICATE_VERIFICATION).toBoolean()
```
sschuberth, line 98: Nit: Proposed wrapping

```
    private val disableRegistryCertificateVerification = analyzerConfig.getPackageManagerConfiguration(managerName)
        ?.options
        ?.get(OPTION_DISABLE_REGISTRY_CERTIFICATE_VERIFICATION)
        .toBoolean()
```

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -63,20 +93,56 @@ class Yarn2(
...
         const val YARN2_RESOURCE_FILE = ".yarnrc.yml"
 
         // The property in `.yarnrc.yml`containing the path to the Yarn2+ executable.
         const val YARN_PATH_PROPERTY_NAME = "yarnPath"
+
+        // The name of the scope with the regular dependencies.
+        private enum class DependencyType(val type: String) {
+            DEPENDENCIES("dependencies"),
+            DEV_DEPENDENCIES("devDependencies")
+        }
```
sschuberth, line 111: I guess there's no reason to have this `enum class` inside a (static) companion object. It should be moved to the top level, like the other private properties, too.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -63,20 +93,56 @@ class Yarn2(
...
+
+        // The pattern to extract rawName, type and version from a Yarn 2+ locator e.g. @babel/preset-env@npm:7.11.0.
+        private val EXTRACT_FROMLOCATOR_PATTERN = Regex("(.+)@(\\w+):(.+)")
+
+        // The pattern to match Artifactory API path.
+        private val ARTIFACTORY_API_PATH_PATTERN = Regex("(.*artifactory.*)/api/npm/(.*)")
+
+        private const val BULK_DETAILS_SIZE = 1000
```
sschuberth, line 119: Please also add docs to this property.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -63,20 +93,56 @@ class Yarn2(
...
     }
 
-    override fun hasLockFile(projectDir: File) = hasYarnLockFile(projectDir)
+    /**
+     * A builder to build the dependency graph of the project.
+     */
```
sschuberth, line 124: These could all use single-line `//` comments as they are not part of the public API anyway.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -112,9 +178,9 @@ class Yarn2(
     }
 
     override fun getVersion(workingDir: File?): String =
-        // `getVersion` with a `null` parameter is called once by the Analyzer to get the version of the global tool.
-        // For Yarn2+, the version is specific to each definition file being scanned therefore a global version doesn't
-        // apply.
+    // `getVersion` with a `null` parameter is called once by the Analyzer to get the version of the global tool.
+    // For Yarn2+, the version is specific to each definition file being scanned therefore a global version doesn't
+    // apply.
```
sschuberth, line 183: Broken indentation.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
-    override fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
-        // Yarn2+ 'info' command does not provide enough information so the default behavior of 'npm view --json' is
-        // kept. Note that it is mandatory to override this function as the super function calls `run()` which would
-        // call Yarn.
-        val process = delegateNpmAnalyzer.run(workingDir, "view", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)
+    override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
+        val workingDir = definitionFile.parentFile
+
+        // Running `yarn install` before `yarn info` allow to get the real local package version. Otherwise, it will be
```
sschuberth, line 198: "allows"

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+        // a generic one such as '0.0.0-use.local'.
+        log.info { "Running 'Yarn install'..." }
+
+        run(
+            "install",
+            workingDir = workingDir
+        )
```
sschuberth, line 205: Use a single line for compactness.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+            "--cache",
+            "--json",
+            workingDir = workingDir,
+            environment = mapOf("YARN_NODE_LINKER" to "pnp")
+        )
+
+        // First pass: Parse the headers and query each package details.
+        val parser = jsonMapper.createParser(process.stdout)
+        parser.use {
+            val iterator = jsonMapper.readValues(parser, ObjectNode::class.java)
```
sschuberth, line 224: Use the Kotlin-version: `jsonMapper.readValues<ObjectNode>(it)`

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+
+            queryPackageDetails(iterator, workingDir)
+        }
+
+        // Second pass: Parse the packages.
+        val parser2 = jsonMapper.createParser(process.stdout)
+        parser2.use {
+            val iterator = jsonMapper.readValues(parser2, ObjectNode::class.java)
```
sschuberth, line 232: Use `jsonMapper.createParser(process.stdout).use {` and `jsonMapper.readValues<ObjectNode>(it)`.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+            "--manifest",
+            "--cache",
+            "--json",
+            workingDir = workingDir,
+            environment = mapOf("YARN_NODE_LINKER" to "pnp")
+        )
+
+        // First pass: Parse the headers and query each package details.
+        val parser = jsonMapper.createParser(process.stdout)
+        parser.use {
```
sschuberth, line 223: Inline the "parser" variable: jsonMapper.createParser(process.stdout).use {

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+     * hood, NPM does a request per package. However, if a solution to batch these requests arise, the code is ready
+     * for it. From the response to `npm file`, package details are extracted and put in [packagesDetails].
+     */
+    private fun queryPackageDetails(iterator: MappingIterator<ObjectNode>, workingDir: File) {
+        log.info { "Parsing packages headers..." }
+
+        iterator.asSequence().forEach { json ->
+            val value = json["value"].textValue()
+            val nameMatcher = requireNotNull(EXTRACT_FROMLOCATOR_PATTERN.matchEntire(value)) {
+                "Name of package $value cannot be parsed"
```
sschuberth, line 257: Missing dot at end of sentence.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+                        *chunk.toTypedArray(),
+                        workingDir = workingDir,
+                        environment = mapOf("NODE_TLS_REJECT_UNAUTHORIZED" to "0")
+                            .takeIf { disableRegistryCertificateVerification }
+                            .orEmpty()
+                    )
+                    val parser = jsonMapper.createParser(process.stdout)
+                    parser.use {
+                        val detailsIterator =
+                            jsonMapper.readValues(parser, ObjectNode::class.java)
```
sschuberth, line 291: Again inline "parser" and use "it", or use `.use { parser ->` (the style should be aligned with the other cases in this file).

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+     */
+    private fun parseAllPackages(
+        iterator: MappingIterator<ObjectNode>,
+        definitionFile: File
+    ): Map<Identifier, Project> {
+        val allDependencies = mutableMapOf<DependencyType, MutableMap<Identifier, List<Identifier>>>()
+        iterator.forEach { node ->
+            val dependencyMapping = parsePackage(node, definitionFile)
+            dependencyMapping.forEach {
+                val mapping = allDependencies.computeIfAbsent(it.key) { mutableMapOf() }
```
sschuberth, line 315: Kotlin-style: Use `getOrPut` instead of `computeIfAbsent`.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+                            // the scope of this package.
+                            YarnModuleInfo(
+                                packageDependency.id,
+                                packageDependency,
+                                packageDependency.collectDependencies(allDependencies[DependencyType.DEPENDENCIES]!!)
+                            )
+                        }
+                    }
+                }?.toSet().orEmpty()
+                dependenciesInfo.forEach {
```
sschuberth, line 363: Blank line before this one.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+
+        val manifest = childrenNode["Manifest"] as ObjectNode
+        // Yarn's manifests contain licenses in an element with an uppercase 'L'. To leverage existing license parsing
+        // code, an extra property with lowercase 'L' is added.
+        manifest["License"].takeUnless {
+            it is NullNode
+        }?.also {
+            manifest.set<ObjectNode>("license", it)
+        }
+        val declaredLicenses = Npm.parseLicenses(manifest)
```
sschuberth, line 399: Blank line before this one, but in exchange remove them between the single-line properties below.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+            val versionFromLocator = cleanYarn2VersionString(header.version)
+            val details: AdditionalData = requireNotNull(packagesDetails["${header.rawName}@$versionFromLocator"]) {
+                "No package details found for '${header.rawName}' at version '$versionFromLocator'."
+            }
+
+            @Suppress("MaxLineLength")
+            // TODO : It is unclear how hashes coming from Yarn are created. They are present in the .yarn.lock file
+            //  under the name "checksum" and they are 130 characters long. It seems these are SHA512 hashes with two
+            //  extra characters. For instance :
+            //  8/b296c92c4b969e973260e47523207cd5769abd27c245a68c26dc7a0fe8053c55bb04360237cb51cab1df52be939da77150ace99ad331fb7fb13b3423ed73ff3d
```
sschuberth, line 432: I'm not seeing those extra 2 chars in the committed Yarn2 `yarn.lock` files...

nnobelis:
> > The 'node_modules' linker was creating symbolic links pointing to symbolic links, therefore walking such file hierarchies crashed ORT.
> 
> Can you elaborate a bit more on this? Walking symlinks that point to other symlinks should not crash ORT, unless maybe cycles are created.

Yes I meant cycles. I edited commit message for disambiguation.

analyzer/src/main/kotlin/managers/Yarn2.kt
nnobelis: I think it happens only if you use NPM package with SHA512 hashes.
None of the one in the test data packages has such hash....

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +192,530 @@ class Yarn2(
...
+        var result = rawVersion.substringAfter("version=")
+            .substringBefore("&")
+
+        // Remove the archive URLs that can be present due to private registries.
+        // See https://github.com/yarnpkg/berry/issues/2192.
+        result = result.substringBefore("::__archiveUrl")
+
+        // Rewrite some dependencies to make them compatible with Identifier.
+        // E.g. typescript@patch:typescript@npm%3A4.0.2#~builtin<compat/typescript>::version=4.0.2&hash=ddd1e8
+        return result.replace(":", "%3A")
```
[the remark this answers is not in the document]
nnobelis: Well I don't want to encode the full string to avoid incompatibility between Yarn2+ locator and NPM version strings.
I am only doing the minimal change which is getting rid of the `:` for the `Identifier` class.

analyzer/src/main/kotlin/managers/Yarn2.kt
nnobelis: I checked again: you get such hash only in the output of `yarn info` e.g.:
```
{
    "value": "@types/node@npm:12.6.8",
    "children": {
        "Version": "12.6.8",
        "Manifest": {
            "License": "MIT",
            "Homepage": null
        },
        "Cache": {
            "Checksum": "8/b0079ab6c10c6f62e9f6dc75b15446a0c21e476b9ff23977639a96407c34fb1354fbfa8edd6f18cc80b1f21e21d393eefbf5a2d48034f74c4c770d4ce2608d07",
            "Path": "analyzer/src/funTest/assets/projects/synthetic/yarn2/.yarn/cache/@types-node-npm-12.6.8-edf14989b6-b0079ab6c1.zip",
            "Size": 103602
        },
        "Dependents": [
            "parse5@npm:3.0.3"
        ]
    }
}
```
`Yarn.lock` indeed contains only the hash. This `8/` is a mystery to me.

nnobelis: I rephrased the TODO.

nnobelis: I adapted the code.

BTW after looking at your pointer I found this:
https://github.com/yarnpkg/berry/blob/1a489d1aed51a447022277b22695f783b4065e29/packages/yarnpkg-core/sources/Project.ts#L354

Could it be that this `8/` is the cache key ?

nnobelis: I am starting to wonder if this cache field of `yarn info` is not the checksum of the package but the checksum of the Yarn cache entry.
Therefore, the correct hash would be the one from `yarn npm info` (the `details` in code).
What do you think ?

sschuberth: > Could it be that this `8/` is the cache key ?

Looks like so, very weird.

> Therefore, the correct hash would be the one from `yarn npm info` (the `details` in code).

Well, that's what I mean with "will the remaining string be the correct SHA512 for the package". If the hashes from `yarn npm info` and the hash with stripped "8/" do not match (although being of the same length), I'd only use the hash from `yarn npm info` to be on the safe side.

nnobelis: Changed to use the hashes from `yarn npm info`.
Also removed fetching the yarn cache information altogether.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +181,503 @@ class Yarn2(
...
-        val process = delegateNpmAnalyzer.run(workingDir, "view", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)
+    override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
+        val workingDir = definitionFile.parentFile
+
+        // Running `yarn install` before `yarn info` allows to get the real local package version. Otherwise, it will be
+        // a generic one such as '0.0.0-use.local'.
+        log.info { "Running 'Yarn install'..." }
+
+        run("install", workingDir = workingDir)
```
MarcelBochtler, line 191: `run("install", workingDir = workingDir)` already logs running yarn install on debug.
I think you can remove the info log.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +181,503 @@ class Yarn2(
...
+
+        log.info { "Running 'Yarn info'..." }
```
MarcelBochtler, line 193: `run` already logs the commands that are run on `debug`. I think you can remove the info logs.

nnobelis: Well, we go again in the debate we had for `FossID` scanner :)

I think a software running several steps after one another should log these steps, so that the user doesn't think the application is hanging.
Naturally, this logging should be sporadic and "high level". For instance here each package requested with `npm list` is NOT logged.
If desired, this is a task for the `DEBUG` level.

MarcelBochtler: Yes, but as of now, the info log prints duplicated information:
```
14:57:32.293 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.analyzer.managers.Yarn2 - Resolving Yarn2 dependencies for path 'package.json'...
14:57:32.295 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.analyzer.managers.Yarn2 - Running 'Yarn install'...
14:57:32.299 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.utils.common.ProcessCapture - Running '/Users/bom1wa3/work/playground/yarn-V2-workspaces-simple-monorepo/.yarn/releases/yarn-2.2.2.cjs install' in '/Users/bom1wa3/work/playground/yarn-V2-workspaces-simple-monorepo'...
14:57:58.218 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.analyzer.managers.Yarn2 - Running 'Yarn info'...
14:57:58.223 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.utils.common.ProcessCapture - Running '/Users/bom1wa3/work/playground/yarn-V2-workspaces-simple-monorepo/.yarn/releases/yarn-2.2.2.cjs info -A -R --manifest --json' in '/Users/bom1wa3/work/playground/yarn-V2-workspaces-simple-monorepo'...
```

nnobelis: You have a point I removed them.

fviernau:
@nnobelis I remember past (fixed) performannce issues with Yarn workspaces projects with large amounts of nodes and, in particular, even larger amounts of edges in the dependency tree. Have you run the implementation against such large project already and can you give a hint on the impact on execution time?

nnobelis:
@fviernau I tested with a big monorepo from one of our customers with around ~1500 nodes and ~2500 edges and the implementation is holding up : Around 8 minutes for the Analyzer run.
If you look at the implementation, you can see that the JSON output of Yarn commands is streamed to improve memory usage.
Maybe you have a pointer to such a big project ? I could then do a run against it.

fviernau:
> @fviernau I tested with a big monorepo from one of our customers with around ~1500 nodes and ~2500 edges and the implementation is holding up : Around 8 minutes for the Analyzer run. If you look at the implementation, you can see that the JSON output of Yarn commands is streamed to improve memory usage. Maybe you have a pointer to such a big project ? I could then do a run against it.

Unfortunately the project I've been using for performance issues is not public. However, I just revisited my performance fixes and these are heavily implementation specific. So, with a complete re-write / new approach, performance characteristics change completely. I believe the monorepo performance you stated above is good enough, thanks!

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +181,500 @@ class Yarn2(
...
+
+            val identifierType = if ("workspace" in locatorType) "Yarn2" else "NPM"
+            when {
+                // Prevent @virtual dependencies because they are internal to Yarn.
+                // See https://yarnpkg.com/features/protocols
+                "virtual" in locatorType -> null
+
+                else -> {
+                    kotlin.runCatching {
+                        Identifier(identifierType, locatorNamespace, locatorName, version)
```
oheger-bosch, line 552: Question: Is this done in a `runCatching` block because the single identifier components could contain illegal characters, i.e. a ':' ?
Please remove the `kotlin.` prefix before `runCatching`.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +181,500 @@ class Yarn2(
...
+        // See https://github.com/yarnpkg/berry/issues/2192.
+        result = result.substringBefore("::__archiveUrl")
+
+        // Rewrite some dependencies to make them compatible with Identifier.
+        // E.g. typescript@patch:typescript@npm%3A4.0.2#~builtin<compat/typescript>::version=4.0.2&hash=ddd1e8
+        return result.replace(":", "%3A")
+    }
+
+    /**
+     * With Yarn 2+, it is currently no possible with a native command to get the repository information for a project
```
oheger-bosch, line 598: "... it is currently *not* possible..."

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -63,20 +99,39 @@ class Yarn2(
...
+        .toBoolean()
 
     companion object {
         // The name of Yarn 2+ resource file.
         const val YARN2_RESOURCE_FILE = ".yarnrc.yml"
 
         // The property in `.yarnrc.yml`containing the path to the Yarn2+ executable.
         const val YARN_PATH_PROPERTY_NAME = "yarnPath"
+
+        // The pattern to extract rawName, type and version from a Yarn 2+ locator e.g. @babel/preset-env@npm:7.11.0.
```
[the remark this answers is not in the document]
nnobelis: Well according to @sschuberth comment (https://github.com/oss-review-toolkit/ort/pull/5528#discussion_r916251480), non public properties should not have KDoc comment but single-line. Or do I understand it wrong ?

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +181,500 @@ class Yarn2(
...
+
+        // First pass: Parse the headers and query each package details.
+        jsonMapper.createParser(process.stdout).use {
+            val iterator = jsonMapper.readValues<ObjectNode>(it)
+
+            queryPackageDetails(iterator, workingDir)
+        }
+
+        // Second pass: Parse the packages.
+        jsonMapper.createParser(process.stdout).use { parser ->
```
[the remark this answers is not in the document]
nnobelis: Well there is a reason to it : if there was only one pass, it would be for each package:
1. parse the json from the header
2. query the details for this package
3. build the package

The problem would be that each package details would be queried one by one, with each time a new process spawn of `yarn npm info`.
The current implementation fires only (Npackages/BULK_DETAILS_SIZE) processes. Even if under the hood NPM still triggers one HTTP request per package, the whole execution is more performant.

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -126,21 +181,500 @@ class Yarn2(
...
+                dependenciesInfo.forEach {
+                    graphBuilder.addDependency(qualifiedScopeName, it)
+                }
+            }
+        }
+        return allProjects
     }
+
+    /**
+     * Construct a Package or a Project by parsing it [json] representation outputted by `yarn info`, run for the given
```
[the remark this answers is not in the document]
nnobelis: It does exist:
https://forum.wordreference.com/threads/is-outputted-a-word.2707379/post-13665821

I will change it by 'generated' nevertheless.

analyzer/src/main/kotlin/managers/Yarn2.kt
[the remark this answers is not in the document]
oheger-bosch: I was rather thinking of parsing the relevant information into temporary data structures once, which could then be iterated over as often as needed. But this means certainly more effort, so I am fine with keeping it as is.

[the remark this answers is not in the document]
nnobelis: Ah yes I see what you meant. I didn't use temporary data structures on purpose to avoid keeping the full parsed JSON in memory. I wanted to stream the JSON file and parse and process one package at a time, to minimize memory usage.
Thanks for your understanding !

analyzer/src/main/kotlin/managers/Yarn2.kt
oheger-bosch: Thanks.

analyzer/src/main/kotlin/managers/Yarn2.kt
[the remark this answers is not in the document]
oheger-bosch: Hm, would be new to me. Maybe @sschuberth can clarify? But nothing to block this PR.

oheger-bosch (approved):

analyzer/src/main/kotlin/managers/Yarn2.kt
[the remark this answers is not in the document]
sschuberth: My comment @nnobelis is referring to was about private properties, but these seem to be public constants. For private properties I was assuming that they will anyway not end up in the public API docs generated by Dokka, and as such also regular code comments could be used. But I might be mistaken, and also docs for private properties could end up being in the API docs.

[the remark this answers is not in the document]
nnobelis: Thanks @sschuberth for the explanation.

I changed the comment types and added a commit to align it for existing properties.

oheger-bosch (approved):

MarcelBochtler:
Merging despite the unrelated failing test.

PR #5534, opened 2022-07-08
Integrate OSV.dev as advisor

fviernau:
See individual commits.

A first iteration for #3599 with a couple of things left as TODO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -644,6 +644,24 @@ ort {
 
 To enable this provider, pass `-a VulnerableCode` on the command line.
 
+## OSV
+
+This provider obtains information about security vulnerabilities from Google [OSV](https://osv.dev/), a distributed
+vulnerability database for Open Source. The database aggregates data from different sources for various ecosystems. The
+configuration is optional and limited to overriding the server URL.
+
+```hocon
+ort {
+  advisor {
+    osv {
+      serverUrl = "https://api-staging.osv.dev"
+    }
+  }
+}
+```
+
+To enable this provider, pass `-a Osv` on the command line.
+
 <a name="evaluator">&nbsp;</a>
 
 [![Evaluator](./logos/evaluator.png)](./evaluator/src/main/kotlin)
--- a/advisor/build.gradle.kts
+++ b/advisor/build.gradle.kts
@@ -26,6 +26,7 @@ plugins {
 dependencies {
     api(project(":clients:nexus-iq"))
     api(project(":clients:oss-index"))
+    api(project(":clients:osv"))
     api(project(":clients:vulnerable-code"))
     api(project(":clients:github-graphql"))
     api(project(":model"))
@@ -35,4 +36,7 @@ dependencies {
 
     testImplementation(libs.mockk)
     testImplementation(libs.wiremock)
+
+    funTestImplementation(libs.jacksonModuleKotlin)
+    funTestApi(libs.jacksonDatabind)
 }
--- a/advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
+++ b/advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
@@ -1,0 +1,53 @@
+{
+  "NPM::glob-parent:5.1.1" : [ {
+    "advisor" : {
+      "name" : "Osv",
+      "capabilities" : [ "VULNERABILITIES" ]
+    },
+    "summary" : {
+      "start_time" : "2022-07-08T09:17:44.773547238Z",
+      "end_time" : "2022-07-08T09:17:45.267886427Z"
+    },
+    "defects" : [ ],
+    "vulnerabilities" : [ {
+      "id" : "GHSA-ww39-953v-wcq6",
+      "references" : [ {
+        "url" : "https://nvd.nist.gov/vuln/detail/CVE-2020-28469",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://github.com/gulpjs/glob-parent/pull/36",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://github.com/gulpjs/glob-parent/blob/6ce8d11f2f1ed8e80a9526b1dc8cf3aa71f43474/index.js%23L9",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://github.com/gulpjs/glob-parent/releases/tag/v5.1.2",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://snyk.io/vuln/SNYK-JAVA-ORGWEBJARSBOWERGITHUBES128-1059093",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://snyk.io/vuln/SNYK-JAVA-ORGWEBJARSNPM-1059092",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://snyk.io/vuln/SNYK-JS-GLOBPARENT-1016905",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://www.oracle.com/security-alerts/cpujan2022.html",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://github.com/gulpjs/glob-parent",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      } ]
+    } ]
+  } ]
+}
--- a/advisor/src/funTest/kotlin/OsvFunTest.kt
+++ b/advisor/src/funTest/kotlin/OsvFunTest.kt
@@ -1,0 +1,91 @@
+/*
+ * Copyright (C) 2022 EPAM Systems, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.inspectors.forAll
+import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNot
+
+import java.io.File
+import java.time.Instant
+
+import org.ossreviewtoolkit.advisor.advisors.Osv
+import org.ossreviewtoolkit.model.AdvisorResult
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.config.AdvisorConfiguration
+import org.ossreviewtoolkit.model.readValue
+
+class OsvFunTest : StringSpec({
+    "retrievePackageFindings() returns vulnerabilities for the supported ecosystems" {
+        val osv = createOsv()
+        val packages = listOf(
+            "PyPI::Plone:3.2",
+            "Maven:com.jfinal:jfinal:1.4",
+            "NPM::rebber:1.0.0",
+            "GoMod::github.com/nats-io/nats-server/v2:v2.1.0",
+            "GoDep::github.com/nats-io/nats-server/v2:v2.1.0",
+            "Crate::sys-info:0.7.0",
+            "NuGet::Microsoft.ChakraCore:1.10.0"
+            // TODO: Add an identifier for a composer package after https://github.com/google/osv.dev/issues/497 got
+            // fixed. That issue causes queries for vulnerabilities of the "Packagist" ecosystem to always return
+            // an empty result.
+        ).map { identifierToPackage(it) }
+
+        val packageFindings = osv.retrievePackageFindings(packages)
+
+        packageFindings.keys shouldContainExactlyInAnyOrder packages
+        packageFindings.forAll { (_, advisorResult) ->
+            advisorResult shouldNot beEmpty()
+        }
+    }
+
+    "retrievePackageFindings() returns the expected result for the given package(s)" {
+        val expectedResult = File("src/funTest/assets").resolve("retrieve-package-findings-expected-result.json")
+            .readValue<Map<Identifier, List<AdvisorResult>>>()
+        val osv = createOsv()
+        val packages = listOf(
+            "NPM::glob-parent:5.1.1"
+        ).map { identifierToPackage(it) }
+
+        val packageFindings = osv.retrievePackageFindings(packages).mapKeys { it.key.id }
+
+        packageFindings.patchTimes() shouldBe expectedResult.patchTimes()
+    }
+})
+
+private fun identifierToPackage(id: String): Package =
+    Package.EMPTY.copy(id = Identifier(id))
+
+private fun createOsv(): Osv =
+    Osv("Osv", AdvisorConfiguration())
+
+private fun Map<Identifier, List<AdvisorResult>>.patchTimes(): Map<Identifier, List<AdvisorResult>> =
+    mapValues { (_, advisorResults) ->
+        advisorResults.map { advisorResult ->
+            advisorResult.copy(
+                summary = advisorResult.summary.copy(
+                    startTime = Instant.MIN,
+                    endTime = Instant.MIN
+                )
+            )
+        }
+    }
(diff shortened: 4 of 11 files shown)


clients/osv/src/main/kotlin/Model.kt
```diff
@@ -75,6 +75,25 @@ data class Credit(
     val contact: List<String> = emptyList()
 )
 
+/**
+ * Defined package ecosystem values, see https://ossf.github.io/osv-schema/.
+ */
+object Ecosystem {
```
sschuberth, line 81: Could you please explain in the commit message why this is not an enum?

I *guess* this because you don't want to break deserialization on unknown values, and don't want to put effort into a custom deserializer...

advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
```diff
@@ -0,0 +1,53 @@
...
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      }, {
+        "url" : "https://github.com/gulpjs/glob-parent",
+        "scoring_system" : "CVSS_V3",
+        "severity" : "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"
+      } ]
+    } ]
+  } ]
+}
```
sschuberth, line 53: Nit: Missing newline at end of file.

advisor/src/funTest/kotlin/OsvFunTest.kt
```diff
@@ -0,0 +1,92 @@
...
+        val packageFindings = osv.retrievePackageFindings(packages)
+
+        packageFindings.keys shouldContainExactlyInAnyOrder packages
+        packageFindings.forAll { (_, advisorResult) ->
+            advisorResult shouldNot beEmpty()
+        }
+    }
+
+    "retrievePackageFindings() returns the expected result for the given package(s)" {
+        val expectedResult: Map<Identifier, List<AdvisorResult>> = File("src/funTest/assets").resolve(
```
sschuberth, line 62: Move the type to the `readValue<...>` call.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
+import org.ossreviewtoolkit.utils.ort.log
+
+/**
+ * An advice provider that obtains vulnerability from OSV.dev https://osv.dev/.
+ */
+class Osv(name: String, advisorConfiguration: AdvisorConfiguration) : AdviceProvider(name) {
+    class Factory : AbstractAdviceProviderFactory<Osv>("Osv") {
+        override fun create(config: AdvisorConfiguration): Osv =
+            Osv(providerName, config)
```
sschuberth, line 52: Nit: Drop the explicit return type and make this a one-liner.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    override val details: AdvisorDetails = AdvisorDetails(providerName, enumSetOf(AdvisorCapability.VULNERABILITIES))
```
sschuberth, line 55: Nit: Drop the explicit type for this override.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+        serverUrl = advisorConfiguration.osv?.serverUrl ?: OsvApiClient.SERVER_URL_PRODUCTION,
+        httpClient = OkHttpClientHelper.buildClient()
+    )
+
+    override suspend fun retrievePackageFindings(packages: List<Package>): Map<Package, List<AdvisorResult>> {
+        val startTime = Instant.now()
+
+        val vulnerabilitiesForPackage = getVulnerabilitiesForPackage(packages)
+
+        log.info { "Found ${vulnerabilitiesForPackage.values.flatten().distinct().size} distinct vulnerabilities." }
```
sschuberth, line 67: Shouldn't such a kind of information be logged at a higher level for all configured advisors?

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+
+        val affectedPackages = vulnerabilitiesForPackage.filter { it.value.isNotEmpty() }.keys
```
sschuberth, line 69: This makes me wonder why there might be package keys with empty values at all. Can't we simply not add such keys in the first place?

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+                )
+            } else {
+                emptyList()
+            }
+        }
+    }
+
+    private fun getVulnerabilitiesForPackage(packages: List<Package>): Map<Identifier, List<Vulnerability>> {
+        val vulnerabilityIdsForPackageId = getVulnerabilityIdsForPackages(packages)
+        val allVulnerabilityIds = vulnerabilityIdsForPackageId.values.flatten().toSet()
```
sschuberth, line 98: Nit: Use `flattenTo(mutableSet())`.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+            packages.forEach { pkg ->
+                putIfAbsent(pkg.id, emptyList())
+            }
+        }
+    }
+
+    private fun getVulnerabilitiesForIds(ids: Set<String>): List<Vulnerability> {
+        val result = service.getVulnerabilitiesForIds(ids)
+
+        return result.getOrNull() ?: run {
```
sschuberth, line 133: Use `getOrElse {` instead.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+            log.error {
+                "Requesting vulnerabilities IDs for packages failed: ${result.exceptionOrNull()!!.collectMessages()}"
+            }
+            emptyList()
+        }
+    }
+}
+
+private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
+    val name: String? = when {
```
sschuberth, line 143: `name` is always non-null, and you can drop the whole type anyway.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+        pkg.id.namespace.isEmpty() -> pkg.id.name
+        pkg.id.type == "Composer" -> "${pkg.id.namespace}/${pkg.id.name}"
+        else -> "${pkg.id.namespace}:${pkg.id.name}"
+    }
+
+    val ecosystem: String? = when (pkg.id.type) {
```
sschuberth, line 149: Drop the explicit type.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+        "GoMod" -> Ecosystem.GO
+        "NPM" -> Ecosystem.NPM
+        "NuGet" -> Ecosystem.NUGET
+        "Maven" -> Ecosystem.MAVEN
+        "PyPI" -> Ecosystem.PYPI
+        else -> null
+    }
+
+    // Strip the version prefix for GoMod / GoDep packages. That prefix should not be assigned by the package manager
+    // in the first place, see https://github.com/oss-review-toolkit/ort/issues/5532.
```
sschuberth, line 164: How about fixing that issue first, so we don't need to temporarily add this code?

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,199 @@
...
+    return null
+}
+
+private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulnerability {
+    val (scoringSystem, severity) = severity.firstOrNull()?.let {
+        it.type.name to it.score
+    } ?: null to null
+
+    // TODO: Improve parsing the severity properties. Parse vectors, e.g. CVSS:3.1 and consider using the severity
+    // in the database specific property as a fallback.
```
sschuberth, line 188: FYI: `VulnerabilityReference.getSeverityString()` already has (very) basic parsing for vectors.

sschuberth (changes requested):

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: This would result in a hint from IntelliJ asking to make it explicit for API stability.

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: As this is a breaking change which requires our user to make alignments in their internal setup for which the PR would need to wait, I prefer to do it afterwards. Will you accept that?

advisor/src/funTest/kotlin/OsvFunTest.kt
sschuberth: Yes, and IMO it (arguably, of course) simply looks nicer / more Kotlin-ish.

advisor/src/main/kotlin/advisors/Osv.kt
sschuberth: Oh, I actually tried, and that hint didn't show up for me (which made sense to me as this is an overridden property). But maybe I was just too impatient. Or the Kotlin 1.7.10 IDE plugin that I just installed behaves differently in this regard.

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: Dropped it and added log output to `Advisor.kt` instead.

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: I cannot find that function, can you give a hint please?

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: Of course we "can", it's however not obvious why we "should". I believe as-is is a tad simpler and also consistent with similar functions, see e.g. https://github.com/oss-review-toolkit/ort/blob/470e2f323b516389bd96a63a81c67b74891a36f8/advisor/src/main/kotlin/advisors/GitHubDefects.kt#L123.

advisor/src/main/kotlin/advisors/Osv.kt
sschuberth: Sorry, it would be `flatMapTo(mutableSetOf()) { it }`, but that's not nicer than the current code. So let's keep the current code even if it might be a tad less performant.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        return buildMap {
+            result.getOrNull().orEmpty().forEachIndexed { i, vulnerabilities ->
+                put(requests[i].first.id, vulnerabilities)
+            }
+
+            packages.forEach { pkg ->
+                putIfAbsent(pkg.id, emptyList())
+            }
```
sschuberth, line 116: Is this the code that requires you to filter out packages with a empty list of vulnerabilities at another place? I can't currently find that code anymore, but I remember to have commented on it during a previous review.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,190 @@
...
+        if (result.isFailure) {
+            log.error {
+                "Requesting vulnerabilities IDs for packages failed: ${result.exceptionOrNull()!!.collectMessages()}"
+            }
+        }
+
+        return buildMap {
+            result.getOrNull().orEmpty().forEachIndexed { i, vulnerabilities ->
+                put(requests[i].first.id, vulnerabilities)
+            }
```
sschuberth, line 112: I have a hunch that this code could be simplified by using `Result.map` which only transforms the result on success. Maybe something like `result.map { ... }.onFailure { ... }`.

model/src/main/resources/reference.conf
```diff
@@ -73,6 +73,10 @@ ort {
       parallelRequests = 5
     }
 
+    osv {
```
sschuberth, line 76: Commit message: "advice"

README.md
```diff
@@ -644,6 +644,24 @@ ort {
 
 To enable this provider, pass `-a VulnerableCode` on the command line.
 
+## OSV
```
sschuberth, line 647: Commit message: No dot at the end of the title. And we usually just use "README" as the prefix here.

sschuberth: In general, this could probably squashed with the previous commit. It seems a bit overly detailed to split these changes.

advisor/src/main/kotlin/Advisor.kt
```diff
@@ -79,7 +79,20 @@ class Advisor(
             runBlocking {
                 providers.map { provider ->
                     async {
-                        provider.retrievePackageFindings(packages)
+                        val providerResults = provider.retrievePackageFindings(packages)
+
+                        this@Advisor.log.info {
+                            "Found ${providerResults.values.flatten().distinct().size} distinct vulnerabilities via " +
+                            "${provider.providerName}. "
```
sschuberth, line 85: This line should be indented.

advisor/src/main/kotlin/Advisor.kt
```diff
@@ -79,7 +79,20 @@ class Advisor(
...
+                        }
+
+                        providerResults.filter { it.value.isNotEmpty() }.keys.takeIf { it.isNotEmpty() }?.let { pkgs ->
+                            this@Advisor.log.info {
+                                "Affected packages:\n\n${pkgs.joinToString("\n") { it.id.toCoordinates() }}\n"
+                            }
```
sschuberth, line 91: I guess this list could grow large. Maybe better use `debug` level?

advisor/src/test/kotlin/AdvisorTest.kt
```diff
@@ -147,3 +147,8 @@ private fun createOrtResultWithPackages(packages: List<Package>): OrtResult =
  */
 private fun createPackage(index: Int): Package =
     Package.EMPTY.copy(id = Identifier.EMPTY.copy(name = "test-package$index"))
+
+private fun mockkAdviseProvider(): AdviceProvider =
```
sschuberth, line 151: "Advice"

sschuberth (changes requested):

sschuberth:
@oheger-bosch I'd appreciate if you could also review this.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,190 @@
...
+
+        return packages.associate { pkg ->
+            pkg.id to vulnerabilityIdsForPackageId[pkg.id].orEmpty().map { vulnerabilityForId.getValue(it) }
+        }
+    }
+
+    private fun getVulnerabilityIdsForPackages(packages: List<Package>): Map<Identifier, List<String>> {
+        val requests = packages.mapNotNull { pkg ->
+            val request = createRequest(pkg) ?: return@mapNotNull null
+            pkg to request
```
sschuberth, line 99: Could be `createRequest(pkg)?.let { pkg to it }`.

README.md
fviernau: IIRC that's quite common practice in past PRs in ORT, so I prefer to just leave this as-is.

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: With this change, the returned map does not contain entries anymore for keys which would be associated with an empty list. There is already a discussion started about it, so let's finish this discussion first.

advisor/src/main/kotlin/advisors/Osv.kt
sschuberth: @oheger-bosch, do you remember whether returning an empty list of defects / vulnerabilities for unaffected packages was a deliberate design decision, or is it an oversight and should we simply have no entries for such packages (which is what I'd personally strongly prefer)?

fviernau: FYI @sschuberth: I just changed it according to your ask. For the private function this is very minor as visibility scope is very small, so I just changed it to shortcut the discussion.

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: I've made the change you suggested, see also discussion above.

advisor/src/main/kotlin/advisors/Osv.kt
oheger-bosch: > @oheger-bosch, do you remember whether returning an empty list of defects / vulnerabilities for unaffected packages was a deliberate design decision, or is it an oversight and should we simply have no entries for such packages (which is what I'd personally strongly prefer)?

As far as I remember, it was a deliberate decision. As the result is a `Map` keyed by `Package`, the thinking was that there should be an entry for each package that was passed as input.

clients/osv/src/main/kotlin/Model.kt
```diff
@@ -75,6 +75,25 @@ data class Credit(
     val contact: List<String> = emptyList()
```
oheger-bosch, line 75: Commit message nit: In the first line of the body "the changing" -> "then changing".

I was a bit irritated that the whole body of the commit message is a reasoning why no enum class is used, while I was expecting some background about the new class. To make clearer that you mainly refer to this implementation detail, maybe start with something like: "Note that for this purpose no enum is used because ..." ?

advisor/src/funTest/kotlin/OsvFunTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        val packages = listOf(
+            "PyPI::Plone:3.2",
+            "Maven:com.jfinal:jfinal:1.4",
+            "NPM::rebber:1.0.0",
+            "GoMod::github.com/nats-io/nats-server/v2:v2.1.0",
+            "GoDep::github.com/nats-io/nats-server/v2:v2.1.0",
+            "Crate::sys-info:0.7.0",
+            "NuGet::Microsoft.ChakraCore:1.10.0"
+            // TODO: Add an identifier for a composer package after https://github.com/google/osv.dev/issues/497 got
+            // fixed. That issues causes queries for vulnerabilities of the "Packagist" ecosystem to always return
```
oheger-bosch, line 49: Typo: "That issues causes" -> "That issue causes"

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -0,0 +1,187 @@
...
+import org.ossreviewtoolkit.model.VulnerabilityReference
+import org.ossreviewtoolkit.model.config.AdvisorConfiguration
+import org.ossreviewtoolkit.utils.common.collectMessages
+import org.ossreviewtoolkit.utils.common.enumSetOf
+import org.ossreviewtoolkit.utils.common.withoutPrefix
+import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
+import org.ossreviewtoolkit.utils.ort.log
+
+/**
+ * An advice provider that obtains vulnerability from OSV.dev https://osv.dev/.
```
oheger-bosch, line 47: Should this be "that obtains vulnerability *information*" or "vulnerabilities" ?

advisor/build.gradle.kts
```diff
@@ -26,6 +26,7 @@ plugins {
 dependencies {
     api(project(":clients:nexus-iq"))
     api(project(":clients:oss-index"))
+    api(project(":clients:osv"))
```
sschuberth, line 29: One more thing: Shouldn't this commit be set to auto-close https://github.com/oss-review-toolkit/ort/issues/3599 via the commit message?

oheger-bosch (approved):

advisor/build.gradle.kts
sschuberth: > One more thing: Shouldn't this commit be set to auto-close #3599 via the commit message?

Scratch that, I overlooked

> A first iteration for https://github.com/oss-review-toolkit/ort/issues/3599 with a couple of things left as TODO.

in the OP.

PR #5536, opened 2022-07-08
ExperimentalScanner: Catch `DownloadException` when creating archives

bs-ondem:
The `ExperimentalScanner` downloads missing archives for packages and whenever,
the `Downloader` fails and throws a `DownloadException` the scanner stopped
the execution. This commit moves the downloader call in a runCatching block
and creates an `OrtIssue` on failure.

Signed-off-by: Onur Demirci <onur.demirci@bosch.io>

### DIFF
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -153,7 +153,7 @@ class ExperimentalScanner(
         runProvenanceScanners(controller, context)
         runPathScanners(controller, context)
 
-        createMissingArchives(controller.getNestedProvenancesByPackage())
+        createMissingArchives(controller)
 
         val results = controller.getNestedScanResultsByPackage().entries.associateTo(sortedMapOf()) {
             it.key.id to it.value.merge()
@@ -548,7 +548,7 @@ class ExperimentalScanner(
         }
     }
 
-    private fun createMissingArchives(nestedProvenances: Map<Package, NestedProvenance>) {
+    private fun createMissingArchives(controller: ScanController) {
         // TODO: The archives are currently created in a way compatible with the existing implementation in the
         //       PathScanner. This allows to keep using existing file archives without changing the logic used to
         //       access those archives in the reporter. To achieve this nested provenances are downloaded recursively,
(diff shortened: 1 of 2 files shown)


sschuberth (changes requested):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -145,7 +145,7 @@ class ExperimentalScanner(
         runProvenanceScanners(controller, context)
         runPathScanners(controller, context)
 
-        createMissingArchives(controller.getNestedProvenancesByPackage())
+        createMissingArchives(controller)
```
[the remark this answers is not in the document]
bs-ondem: Yes, thank you. My commit msg was probably a little bit confusing.

sschuberth (dismissed):

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -541,17 +541,28 @@ class ExperimentalScanner(
...
-                archiver.archive(dir, nestedProvenance.root)
-                dir.safeDeleteRecursively(force = true)
+            provenancesWithMissingArchives.forEach { (pkg, nestedProvenance) ->
+                runCatching {
+                    downloadRecursively(nestedProvenance)
+                }.onSuccess { dir ->
+                    archiver.archive(dir, nestedProvenance.root)
+                    dir.safeDeleteRecursively(force = true)
+                }.onFailure {
+                    controller.addProvenanceResolutionIssue(
```
mnonnenmacher, line 557: Provenance resolution issues happen only for packages which cannot be scanned. Therefore the implementation expects that no scan results exist for those packages and later on creates empty scan results with only these issues.
Using this function here would result in duplicate scan results for those packages.

Instead, I would recommend to the the folowing:
* Add a new `addIssue()` function to the `ScanController` backed by its own map.
* Add these issues to the scan results created in `getNestedScanResultsByPackage()`, probably best by accessing above map in `buildNestedProvenanceScanResult()`.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -58,6 +58,11 @@ class ScanController(
      */
     private val provenanceResolutionIssues = mutableMapOf<Identifier, MutableList<OrtIssue>>()
 
+    /**
+     * TODO: Docs
```
sschuberth, line 62: Uhm, shouldn't this have been filled before merging?

bs-ondem: My bad, I forgot to add the docs -> https://github.com/oss-review-toolkit/ort/pull/5769

PR #5537, opened 2022-07-08
ClearlyDefinedPackageCurationProvider: Filter duplicate curations

mnonnenmacher:
When requesting curations for a certain coordinate the ClearlyDefined
API [1] sometimes also return curations for other coordinates. For
example, the response for "maven/mavencentral/org.ow2.asm/asm/7.0" also
contains curations for other versions of the same package, but also for
other packages that start with "asm", like "asm-tree". This means, that
the same curations can be contained multiple times in a response, if
multiple coordinates are sent at once.

As a result, it can happen that ORT adds the same curation multiple
times to a package. To prevent this, filter duplicate curations before
returning the result.

[1]: https://api.clearlydefined.io/api-docs/#/curations/post_curations_

### DIFF
--- a/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
@@ -172,6 +172,6 @@ class ClearlyDefinedPackageCurationProvider(
             }
         }
 
-        return curations
+        return curations.mapValues { (_, curations) -> curations.distinct() }
     }
 }
--- a/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
@@ -32,7 +32,6 @@ import org.ossreviewtoolkit.analyzer.PackageCurationProvider
 import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService
 import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService.Server
 import org.ossreviewtoolkit.clients.clearlydefined.ComponentType
-import org.ossreviewtoolkit.clients.clearlydefined.ContributedCurations
 import org.ossreviewtoolkit.clients.clearlydefined.Coordinates
 import org.ossreviewtoolkit.clients.clearlydefined.SourceLocation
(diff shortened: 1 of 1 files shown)


fviernau (changes requested):

sschuberth:
> This means, that
> the same curations can be contained multiple times in a response, if
> multiple coordinates are sent at once.

Hmm, I was pretty sure we're already accounting for this. Let me try to dig out some old references.

sschuberth:
> Let me try to dig out some old references

Ok, what I remembered is only semi-related: You might [get back curations for other *versions*](https://github.com/clearlydefined/service/issues/901#issuecomment-1040635510) than you requested. The filtering for the correct version happens more or less implicitly in https://github.com/oss-review-toolkit/ort/blob/ed022a8e9d493db040afb388b3a2dafb7b0790e0/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt#L148

However, I acknowledge that this filtering does not help in the case described in the commit message.

analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
```diff
@@ -172,6 +172,6 @@ class ClearlyDefinedPackageCurationProvider(
             }
         }
 
-        return curations
+        return curations.mapValues { (_, curations) -> curations.distinct() }
```
sschuberth, line 174: I'm wondering whether we should instead extend the check inline 166 / insert a new check between lines 166 and 167 to not add the `PackageCuration` if it's already present. But it's a bit tricky due to the `data.copy(comment = "Provided by ClearlyDefined.")`. *Could* be something like

                if (data != PackageCurationData()) {
                    val curation = PackageCuration(
                        id = pkgId,
                        data = data.copy(comment = "Provided by ClearlyDefined.")
                    )

                    curations.getOrPut(pkgId) { mutableListOf() }.also { if (curation !in it) it += curation }
                }

But your variant is probably more efficient.

mnonnenmacher: Another option would be to use a set for the curations and convert it to a list in the end, but I think with regards to efficiency this should be about the same as filtering duplicates at the end.

sschuberth (approved):

fviernau (approved):

PR #5545, opened 2022-07-12
Osv: Add a fallback for the severity

fviernau:
Please see the individual commits.

Part of #3599.

### DIFF
--- a/advisor/build.gradle.kts
+++ b/advisor/build.gradle.kts
@@ -33,6 +33,7 @@ dependencies {
 
     implementation(libs.cvssCalculator)
     implementation(libs.kotlinxCoroutines)
+    implementation(libs.kotlinxSerialization)
     implementation(libs.ktorClientOkHttp)
 
     testImplementation(libs.mockk)
--- a/advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
+++ b/advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
@@ -65,15 +65,15 @@
       "references" : [ {
         "url" : "https://github.com/amark/gun/security/advisories/GHSA-886v-mm6p-4m66",
         "scoring_system" : null,
-        "severity" : null
+        "severity" : "HIGH"
       }, {
         "url" : "https://github.com/advisories/GHSA-886v-mm6p-4m66",
         "scoring_system" : null,
-        "severity" : null
+        "severity" : "HIGH"
       }, {
         "url" : "https://github.com/amark/gun",
         "scoring_system" : null,
-        "severity" : null
+        "severity" : "HIGH"
       } ]
     } ]
   } ]
--- a/advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
+++ b/advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
@@ -49,5 +49,32 @@
         "severity" : "7.5"
       } ]
     } ]
+  } ],
+  "NPM::gun:0.7.0" : [ {
+    "advisor" : {
+      "name" : "Osv",
+      "capabilities" : [ "VULNERABILITIES" ]
+    },
+    "summary" : {
+      "start_time" : "2022-07-11T11:43:59.561110209Z",
+      "end_time" : "2022-07-11T11:44:00.397553119Z"
+    },
+    "defects" : [ ],
+    "vulnerabilities" : [ {
+      "id" : "GHSA-886v-mm6p-4m66",
+      "references" : [ {
+        "url" : "https://github.com/amark/gun/security/advisories/GHSA-886v-mm6p-4m66",
+        "scoring_system" : null,
+        "severity" : null
+      }, {
+        "url" : "https://github.com/advisories/GHSA-886v-mm6p-4m66",
+        "scoring_system" : null,
+        "severity" : null
+      }, {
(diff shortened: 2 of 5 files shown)


advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -168,19 +171,26 @@ private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
 }
 
 private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulnerability {
-    val (scoringSystem, severity) = severity.firstOrNull()?.let {
+    var (scoringSystem, score) = severity.firstOrNull()?.let {
```
sschuberth, line 175: IMO the variable should stay called "severity", esp. as that's also the field name of `databaseSpecific` you're reading out below.

clients/osv/src/main/kotlin/Model.kt
```diff
@@ -25,6 +25,15 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.JsonObject
 
+// FIXME: Remove all JsonElement subtypes as property types from the model in favor of raw strings holding JSON.
```
sschuberth, line 28: Can't you simply use a custom deserializer as outlined [here](https://github.com/Kotlin/kotlinx.serialization/issues/1058#issuecomment-983269638) that calls `encodeToString()` on the `JsonElement`? Performance should not be an issue here, as I do not expect these model element to be large.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -171,6 +171,7 @@ private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
 }
 
 private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulnerability {
+    // FIXME: Clarify which entry is relevant and pick the right one accordingly.
```
sschuberth, line 174: So far, we basically only use `TODO` in the code base (modulo a few places that you introduced). As I don't see a relevant difference between `TODO` and `FIXME`, can we stich to `TODO` so we'd only need to search for one term instead of two?

advisor/src/funTest/kotlin/OsvFunTest.kt
```diff
@@ -63,7 +63,8 @@ class OsvFunTest : StringSpec({
             .readValue<Map<Identifier, List<AdvisorResult>>>()
         val osv = createOsv()
         val packages = listOf(
-            "NPM::glob-parent:5.1.1"
+            "NPM::glob-parent:5.1.1",
+            "NPM::gun:0.7.0" // Package without severity, but with severity inside the databaseSpecific JSON object.
```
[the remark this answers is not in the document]
fviernau: yes.

clients/osv/src/main/kotlin/Model.kt
fviernau: I've seen this comment and tried to write code based on it. I haven't figured it out though. As you say "simply", maybe you know how to use it to reach the goal. In particular, it was unclear what this `polymorphicSerlializer` variable is and how to modify `BaseClass` in that code snippet to make it compile. Do you have a hint?

advisor/build.gradle.kts
```diff
@@ -33,6 +33,7 @@ dependencies {
 
     implementation(libs.cvssCalculator)
     implementation(libs.kotlinxCoroutines)
+    implementation(libs.kotlinxSerialization)
```
sschuberth, line 36: IMO this change plus the comments in `clients/osv/src/main/kotlin/Model.kt` should go to a separate commit as they fix an issue that was present before the other changes in this commit.

clients/osv/src/main/kotlin/Model.kt
```diff
@@ -33,7 +42,6 @@ import kotlinx.serialization.json.JsonObject
  *
  * For the documentation of all entities and properties please refer to above links.
  */
-
```
sschuberth, line 36: Nit: The removal of this blank line should also go to a separate commit.

advisor/build.gradle.kts
fviernau: What issue do you think this fixes?

sschuberth: Sorry, I was mislead by the earlier discussion of (not) using `kotlinx.serialization` classes in the API. Please ignore this comment of mine.

advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
```diff
@@ -49,5 +49,32 @@
         "severity" : "7.5"
       } ]
     } ]
+  } ],
```
sschuberth, line 52: I'm not sure if this was a misunderstanding of one of my earlier comments, but IMO this commit should be squashed with the next one, so test data and implementation are added in the same commit.

fviernau: My intention was to make the effect of the parsing obvious in the commits. Your proposal would defeat that idea. But I do actually like it and prefer to not squash. Ok?

sschuberth: Makes sense, let's keep it as-is.

PR #5546, opened 2022-07-12
Use "Go" as value for `Identifier.type` for all Go packages

### LINKED ISSUE #5533: GoMod and GoDep packages should use the same Identifier.type
Currently there are two ways to refer to the same thing:

1. `GoMod::github.com/nats-io/nats-server/v2:v2.1.0`
2. `GoDep::github.com/nats-io/nats-server/v2:v2.1.0`

Use `Go` as `Identifier.type` for packages of the managers `GoMod` and `GoDep` to fix that issue.

fviernau:
See individual commits.

Note: The label "breaking" has been set because any ORT configuration referring to Go packages by ID need alignment, like e.g. curations for Go packages. See also https://github.com/oss-review-toolkit/ort-config/pull/43.

Fixes #5533.

### DIFF
--- a/advisor/src/main/kotlin/advisors/Osv.kt
+++ b/advisor/src/main/kotlin/advisors/Osv.kt
@@ -143,8 +143,6 @@ private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
         "Bower" -> null
         "Composer" -> Ecosystem.PACKAGIST
         "Crate" -> Ecosystem.CRATES_IO
-        // TODO: The entries for Go should be merged, see https://github.com/oss-review-toolkit/ort/issues/5533.
-        "GoDep" -> Ecosystem.GO
         "Go" -> Ecosystem.GO
         "NPM" -> Ecosystem.NPM
         "NuGet" -> Ecosystem.NUGET
--- a/advisor/src/main/kotlin/advisors/Osv.kt
+++ b/advisor/src/main/kotlin/advisors/Osv.kt
@@ -145,7 +145,7 @@ private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
         "Crate" -> Ecosystem.CRATES_IO
         // TODO: The entries for Go should be merged, see https://github.com/oss-review-toolkit/ort/issues/5533.
         "GoDep" -> Ecosystem.GO
-        "GoMod" -> Ecosystem.GO
+        "Go" -> Ecosystem.GO
         "NPM" -> Ecosystem.NPM
         "NuGet" -> Ecosystem.NUGET
         "Maven" -> Ecosystem.MAVEN
--- a/analyzer/src/funTest/assets/projects/synthetic/glide-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/glide-expected-output.yml
@@ -18,12 +18,12 @@ project:
   scopes:
   - name: "default"
     dependencies:
-    - id: "GoDep::github.com/mattn/go-isatty:v0.0.12"
+    - id: "Go::github.com/mattn/go-isatty:v0.0.12"
       linkage: "STATIC"
-    - id: "GoDep::golang.org/x/sys:05986578812163b26672dabd9b425240ae2bb0ad"
+    - id: "Go::golang.org/x/sys:05986578812163b26672dabd9b425240ae2bb0ad"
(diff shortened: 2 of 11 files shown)


analyzer/src/funTest/assets/projects/synthetic/glide-expected-output.yml
```diff
@@ -18,13 +18,13 @@ project:
...
+    - id: "Go::github.com/mattn/go-isatty:v0.0.12"
       linkage: "STATIC"
-    - id: "GoDep::golang.org/x/sys:05986578812163b26672dabd9b425240ae2bb0ad"
+    - id: "Go::golang.org/x/sys:05986578812163b26672dabd9b425240ae2bb0ad"
       linkage: "STATIC"
 packages:
-- id: "GoDep::github.com/mattn/go-isatty:v0.0.12"
-  purl: "pkg:golang/github.com%2Fmattn%2Fgo-isatty@v0.0.12"
+- id: "Go::github.com/mattn/go-isatty:v0.0.12"
+  purl: "pkg:go/github.com%2Fmattn%2Fgo-isatty@v0.0.12"
```
MarcelBochtler, line 27: The purl specification lists `golang` in their examples, I think we should comply to this specification and continue to use `golang` for PURLs.
https://github.com/package-url/purl-spec#some-purl-examples

MarcelBochtler (changes requested):

maxhbr (changes requested):
in purls, it should be `golang` instead of `go`

analyzer/src/funTest/assets/projects/synthetic/glide-expected-output.yml
fviernau: This is IMO out of scope of this commit, because it wasn't golang before and isn't golang after. 
I've discovered this issue with the types we used in PURL as well already. In today's dev meeting I promised to file a GitHub issue to address it (the problem is not just for Go packages). Agreed?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -152,11 +152,13 @@ class GoMod(
 
         fun parseModuleEntry(entry: String): Identifier =
             entry.substringBefore('@').let { moduleName ->
+                val version = moduleInfo(moduleName).version
+
                 Identifier(
-                    type = managerName,
+                    type = managerName.takeIf { version.isBlank() } ?: "Go",
```
MarcelBochtler, line 158: I don't completely get this. Why is `GoMod` used as `type` if the version is blank?
Why not use `"Go"` all the time?

fviernau: There is just one project, all other entries are packages. Only the project has an empty version string. For the project GoMod shall be used.

MarcelBochtler (approved):

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -143,9 +143,7 @@ private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
         "Bower" -> null
         "Composer" -> Ecosystem.PACKAGIST
         "Crate" -> Ecosystem.CRATES_IO
-        // TODO: The entries for Go should be merged, see https://github.com/oss-review-toolkit/ort/issues/5533.
-        "GoDep" -> Ecosystem.GO
-        "GoMod" -> Ecosystem.GO
+        "Go" -> Ecosystem.GO
```
maxhbr, line 146: I am not sure whether this is used in the context of parsing of old reports, but just to mention it: you could also keep the "old" mapping for more backwards compatibility

model/src/main/kotlin/utils/Extensions.kt
```diff
@@ -157,7 +157,7 @@ fun Identifier.getPurlType() =
         "composer" -> PurlType.COMPOSER
         "conan" -> PurlType.CONAN
         "crate" -> PurlType.CARGO
-        "godep", "gomod" -> PurlType.GOLANG
+        "go" -> PurlType.GOLANG
```
maxhbr, line 160: same here, maybe it would be worth to keep the old mapping for backwards compatibility

fviernau: I'll make a follow-up PR so this can be discussed separately.

PR #5551, opened 2022-07-13
SpdxDocumentFileFunTest: Reorder test assets

mnonnenmacher:
Move the test projects into separate folders as it became difficult to
keep an overview of which files belong together, and more files will be
added to this directory in upcoming commits.

PR #5564, opened 2022-07-14
SPDX: Definition file references

mnonnenmacher:
Please see the commit messages for details. This PR depends on #5551 to be merged first.

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/spdx/subproject-conan/subproject/subproject-with-conan-reference.spdx.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/spdx/subproject-conan/subproject/subproject-with-conan-reference.spdx.yml
@@ -1,0 +1,21 @@
+SPDXID: "SPDXRef-DOCUMENT"
+spdxVersion: "SPDX-2.2"
+creationInfo:
+  created: "2022-07-06T12:00:00Z"
+  creators:
+    - "Organization: Example Inc."
+  licenseListVersion: "3.9"
+name: "sub-conan-0.1.0"
+dataLicense: "CC0-1.0"
+documentNamespace: "http://spdx.org/spdxdocs/spdx-document-xyz"
+documentDescribes:
+  - "SPDXRef-Package-subproject-conan"
+packages:
+  - SPDXID: "SPDXRef-Package-subproject-conan"
+    copyrightText: "Copyright (c) 2020 Example Inc."
+    downloadLocation: "git+ssh://gitlab.example.com:3389/products/conan-project.git@b2c358080011af6a366d2512a25a379fbe7b1f78"
+    licenseConcluded: "NOASSERTION"
+    licenseDeclared: "NOASSERTION"
+    name: "ConanProject"
+    versionInfo: "1.0.0"
+    packageFileName: "./conanfile.py?scope=requires"
--- a/analyzer/src/funTest/assets/projects/synthetic/spdx/subproject-conan/subproject/subproject-with-conan-reference.spdx.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/spdx/subproject-conan/subproject/subproject-with-conan-reference.spdx.yml
@@ -18,4 +18,8 @@ packages:
     licenseDeclared: "NOASSERTION"
     name: "ConanProject"
     versionInfo: "1.0.0"
-    packageFileName: "./conanfile.py?scope=requires"
+    packageFileName: "./conanfile.py"
+    externalRefs:
+    - referenceCategory: "PACKAGE_MANAGER"
+      referenceLocator: "pkg:conan/example.org/sub-conan@0.1.0?scope=requires"
+      referenceType: "purl"
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -48,6 +48,8 @@ import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.orEmpty
 import org.ossreviewtoolkit.model.utils.toPurl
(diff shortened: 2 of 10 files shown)


analyzer/src/funTest/assets/projects/synthetic/spdx/subproject-conan/subproject/subproject-with-conan-reference.spdx.yml
```diff
@@ -0,0 +1,21 @@
...
+  - "SPDXRef-Package-subproject-conan"
+packages:
+  - SPDXID: "SPDXRef-Package-subproject-conan"
+    copyrightText: "Copyright (c) 2020 Example Inc."
+    downloadLocation: "git+ssh://gitlab.example.com:3389/products/conan-project.git@b2c358080011af6a366d2512a25a379fbe7b1f78"
+    licenseConcluded: "NOASSERTION"
+    licenseDeclared: "NOASSERTION"
+    name: "ConanProject"
+    versionInfo: "1.0.0"
+    packageFileName: "./conanfile.py?scope=requires"
```
sschuberth, line 21: I fear this "made-up" syntax renders this file incompatible with any other SPDX parsers. What's the reason for trying it like this rather than relying on an `externalRefs` entry with a "purl" for this package, which *can* easily contain a query string / qualifier as part of the URI?

oheger-bosch: I like the idea. So, `packageFileName` would be a plain file name, and there would be an additional `externalRef` of type `Purl`, from which only the `scope` query parameter would be evaluated, right?

sschuberth: Yes, exactly, that's what I'd propose (and how I believe it was discussed Bosch-internally before; that's why I was asking whether there's maybe a reason to deviate from that proposal, but probably it was just forgotten / misunderstood).

sschuberth: FYI, here's an example (by @jsteinhofff from Bosch) how PURLs for COnan should look like: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#conan

oheger-bosch: I have changed the implementation accordingly in a new commit.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -414,6 +446,10 @@ class SpdxDocumentFile(
             )
         }
 
+    override fun beforeResolution(definitionFiles: List<File>) {
+        super.beforeResolution(definitionFiles)
+    }
+
```
sschuberth, line 464: This doesn't actually override anything and should be removed.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -100,6 +103,16 @@ internal fun SpdxDocument.projectPackage(): SpdxPackage? =
...
 
+/**
+ * Try to find an [SpdxExternalReference] in this [SpdxPackage] of type purl from which the scope of a
+ * package manager dependency can be extracted. Return this scope or *null* if cannot be determined.
+ */
+internal fun SpdxPackage.extractScopeFromExternalReferences(): String? =
+    externalRefs.filter { it.referenceType == SpdxExternalReference.Type.Purl }
+        .mapNotNull { REG_SCOPE.matchEntire(it.referenceLocator) }
+        .map { it.groupValues[1] }
+        .firstOrNull()
```
sschuberth, line 114: I propose to instead implement this by creating a `URI` from the PURL and use our `URI.getQueryParameters()` extension function.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: There's also the `String.toUri` extension function which should make this rather readable, like

    externalRefs.filter { it.referenceType == SpdxExternalReference.Type.Purl }
        .mapNotNull { it.referenceLocator.toUri { uri -> uri.getQueryParameters() }.getOrNull() }
        .firstNotNullOfOrNull { it["scope"]?.singleOrNull() }

oheger-bosch: I tried this, but found an unexpected problem with this approach: The PURL to reference the definition file looks like this: `pkg:conan/test.org@1.2.3?scope=someScope`. According to the [Javadocs of the URI class](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html), this is not a hierarchical URI because the scheme-specific part does not start with a slash. Therefore, this part is not further parsed, and the query string is not processed. For this reason, the `getQueryParameters()` function returns an empty map.

sschuberth: Crap, that looks like a design-flaw of PURLs to me... anyway, I'd probably just add another mapping then:

    externalRefs.filter { it.referenceType == SpdxExternalReference.Type.Purl }
        .map { it.referenceLocator.replace("pkg:", "pkg://") }
        .mapNotNull { it.toUri { uri -> uri.getQueryParameters() }.getOrNull() }
        .firstNotNullOfOrNull { it["scope"]?.singleOrNull() }

oheger-bosch: Changed to something close to your proposal.

sschuberth (changes requested):

sschuberth:
> I will try to rerun the failing tests.

I've already triggered a re-run.

PR #5566, opened 2022-07-14
OSV test fixes

sschuberth:

### DIFF
--- a/clients/osv/src/main/kotlin/OsvApiClient.kt
+++ b/clients/osv/src/main/kotlin/OsvApiClient.kt
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.clients.osv
 
 import com.jakewharton.retrofit2.converter.kotlinx.serialization.asConverterFactory
 
+import java.time.Instant
 import java.util.concurrent.Executors
 
 import kotlinx.serialization.SerialName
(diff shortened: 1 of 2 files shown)


clients/osv/src/main/kotlin/OsvApiClient.kt
```diff
@@ -114,12 +115,14 @@ data class VulnerabilitiesForPackageBatchResponse(
...
     data class IdList(
         @SerialName("vulns")
-        val vulnerabilities: List<Id> = emptyList()
+        val vulnerabilities: List<Vulnerability> = emptyList()
     )
 
     @Serializable
-    data class Id(
-        val id: String
+    data class Vulnerability(
```
fviernau, line 122: How about naming this differently, because we have a class of same name in `Model.kt` ?
Maybe `VulnerabilityRef` ? Or `IdWithTimestamp`?

fviernau (changes requested):

sschuberth:
This is (by mistake) also contained in https://github.com/oss-review-toolkit/ort/pull/5565, which got merged already.

clients/osv/src/main/kotlin/OsvApiClient.kt
sschuberth: You already approved this as part of https://github.com/oss-review-toolkit/ort/pull/5565 😆

fviernau:
> This is (by mistake) also contained in #5565, which got merged already.

I reviewed in the wrong order and overlooked the new class name :-( sorry.

sschuberth:
> Anyhow, what do you think about renaming the class in a follow up?

Hmm, I'd refer `VulnerabilityRef` then. Fine with you?

PR #5570, opened 2022-07-15
fossid-webapp: Add a property to keep failed scans

nnobelis:
When there is an error during a FossID scan, the default behavior is to
delete the scan to avoid cluttering the FossID instance. While useful,
this makes error analysis of runs difficult. Therefore, this commit adds
an option to keep failed scans.


nnobelis:
@oheger-bosch FYI

oheger-bosch (dismissed):

MarcelBochtler (approved):

PR #5577, opened 2022-07-18
Osv: Add fault tolerance for invalid URIs

fviernau:
See individual commits.

Context: #3599.

### DIFF
--- a/advisor/src/main/kotlin/advisors/Osv.kt
+++ b/advisor/src/main/kotlin/advisors/Osv.kt
@@ -19,7 +19,6 @@
 
 package org.ossreviewtoolkit.advisor.advisors
 
-import java.net.URI
 import java.time.Instant
 
 import kotlinx.serialization.json.JsonPrimitive
@@ -42,6 +41,7 @@ import org.ossreviewtoolkit.model.VulnerabilityReference
 import org.ossreviewtoolkit.model.config.AdvisorConfiguration
 import org.ossreviewtoolkit.utils.common.collectMessages
 import org.ossreviewtoolkit.utils.common.enumSetOf
+import org.ossreviewtoolkit.utils.common.toUri
 import org.ossreviewtoolkit.utils.common.withoutPrefix
 import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.log
@@ -198,14 +198,17 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
         }
     }
 
-    return org.ossreviewtoolkit.model.Vulnerability(
-        id = id,
-        references = references.map {
+    val references = references.mapNotNull { reference ->
+        reference.url.trim().toUri().onFailure {
+            log.debug { "Could not parse reference URL for vulnerability '$id': ${it.message}." }
+        }.map {
             VulnerabilityReference(
-                url = URI.create(it.url.trim()),
+                url = it,
                 scoringSystem = scoringSystem,
                 severity = severity,
             )
-        }
-    )
+        }.getOrNull()
+    }
+
+    return org.ossreviewtoolkit.model.Vulnerability(id = id, references = references)
 }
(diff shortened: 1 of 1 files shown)


advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -198,14 +198,20 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
...
 
-    return org.ossreviewtoolkit.model.Vulnerability(
-        id = id,
-        references = references.map {
-            VulnerabilityReference(
-                url = URI.create(it.url.trim()),
-                scoringSystem = scoringSystem,
-                severity = severity,
-            )
+    val references = references.mapNotNull {
```
sschuberth, line 201: Commit message: "ingore" -> "ignore"

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -198,14 +198,20 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
...
+        val url = try {
+            URI.create(it.url.trim())
```
sschuberth, line 203: I propose to make use of our `String.toUri()` extension function instead. Something along the line of

    it.url.trim().toUri().onFailure { log... }.getOrNull()

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -202,7 +202,7 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
         id = id,
         references = references.map {
             VulnerabilityReference(
-                url = URI.create(it.url),
+                url = URI.create(it.url.trim()),
```
sschuberth: The quoted issue also shows that sometimes the scheme is missing from the URI. Have you considered to also add "https" if the string does not start with "http" (but has "://")?

sschuberth (changes requested):

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: I've had this considered and ended up opting in only for the super trivial work around of trimming, deciding against since since there are only two instance of the problem with the hope of having it fixed upstream. But as you asked for it I've just added a commit on top to implement that.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -198,14 +198,17 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
         }
     }
 
-    return org.ossreviewtoolkit.model.Vulnerability(
-        id = id,
-        references = references.map {
+    val references = references.mapNotNull { reference ->
+        reference.url.trim().toUri().onFailure { e ->
+            log.debug { "Could not parse reference URL for vulnerability '$id': ${e.message}." }
+        }.getOrNull()?.let {
```
sschuberth, line 210: To be even more fluent, you'd model this with `.map { ... }.getOrNull()`.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -198,14 +198,17 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
         }
     }
 
-    return org.ossreviewtoolkit.model.Vulnerability(
-        id = id,
-        references = references.map {
+    val references = references.mapNotNull { reference ->
+        reference.url.trim().toUri().onFailure { e ->
```
sschuberth: Nit: Could keep using `it` as you also do so below.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -199,7 +199,13 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
     }
 
     val references = references.mapNotNull { reference ->
-        reference.url.trim().toUri().onFailure { e ->
+        reference.url.trim().let { url ->
```
sschuberth, line 202: Alternatively:

    val url = reference.url.trim()
    url.takeUnless { it.startsWith("://") } ?: "https$url"

sschuberth (approved):

PR #5578, opened 2022-07-18
Extensions: Add backwards compatibilites for renamed 'Go' packages

fviernau:
As of [1] Go packages use "Go" as value for `Identifier.type`. So, add
entries for backwards compatibility with old analyzer results. Note that
the entries are only ok for packages, but invalid for projects.

[1] https://github.com/oss-review-toolkit/ort/pull/5546

Note: This change was requested https://github.com/oss-review-toolkit/ort/pull/5546#discussion_r923048309. Personally I'd have the tendency to not merge it.

### DIFF
--- a/model/src/main/kotlin/utils/Extensions.kt
+++ b/model/src/main/kotlin/utils/Extensions.kt
@@ -157,7 +157,7 @@ fun Identifier.getPurlType() =
         "composer" -> PurlType.COMPOSER
         "conan" -> PurlType.CONAN
         "crate" -> PurlType.CARGO
-        "go" -> PurlType.GOLANG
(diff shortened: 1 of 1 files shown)


sschuberth:
This backward compatibility would only ever be needed when using a old analyzer result, and (re-)running the advisor / reporter on it, right?

If so, I also have a tendency to not merge this, as recreating the analyzer result should be rather cheap.

fviernau:
> This backward compatibility would only ever be needed when using a old analyzer result, and (re-)running the advisor / reporter on it, right?

right.

> 
> If so, I also have a tendency to not merge this, as recreating the analyzer result should be rather cheap.

Agreed, I'll close the PR.

PR #5580, opened 2022-07-18
Osv: Avoid using `var` by using `let`

sschuberth:
It is hard to track the current value of non-final `var`s, which is why
their use is discouraged. So use `let` instead.

### DIFF
--- a/advisor/src/main/kotlin/advisors/Osv.kt
+++ b/advisor/src/main/kotlin/advisors/Osv.kt
@@ -200,8 +200,7 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
     }
 
     val references = references.mapNotNull { reference ->
-        var url = reference.url.trim()
-        url = url.takeUnless { it.startsWith("://") } ?: "https$url"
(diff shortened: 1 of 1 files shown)


tsteenbe (dismissed):
Please fix 'uing' in the commit message "Osv: Avoid using var by uing let"

fviernau (approved):

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -200,8 +200,7 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
     }
 
     val references = references.mapNotNull { reference ->
-        var url = reference.url.trim()
-        url = url.takeUnless { it.startsWith("://") } ?: "https$url"
+        val url = reference.url.trim().let { url -> url.takeUnless { it.startsWith("://") } ?: "https$url" }
```
mnonnenmacher, line 203: Unrelated, but while at it you could also replace `takeUnless` with a normal `if`, because it's shorter and easier to understand:
```
if (url.startsWith("://")) "https$url" else url
```

fviernau (approved):

fviernau (approved):

PR #5586, opened 2022-07-19
GoMod/GoDep: Normalize package versions

### LINKED ISSUE #5532: Fix identifiers assigned to 'Go' packages by 'GoMod' and 'GoDep'
The following things should be fixed
1. The `Identifier.version` property has `v` as prefix which should be stripped
2. The `Identifier.type` property for packages is either `GoMod` or `GoDep` and should be changed to `Go`, as currently the e.g. following IDs refer to the same package:

"GoMod::github.com/nats-io/nats-server/v2:v2.1.0",
"GoDep::github.com/nats-io/nats-server/v2:v2.1.0",

fviernau:
Map the Go package / module version strings to semver compatible version strings. See individual commits.
Note: A reason for doing so, besides consistency, is to allow for passing the version strings as-is to any third-party advisor backend API. OSV can handle the `v` prefixes, but maybe not all backends can.

Fixes #5532.

_edit: The change is breaking, because any ORT configuration which uses version strings for Go packages need to be adjusted._

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/godep-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/godep-expected-output.yml
@@ -18,17 +18,17 @@ project:
   scopes:
   - name: "default"
     dependencies:
-    - id: "Go::github.com/fatih/color:v1.9.0"
+    - id: "Go::github.com/fatih/color:1.9.0"
       linkage: "STATIC"
-    - id: "Go::github.com/mattn/go-colorable:v0.1.6"
+    - id: "Go::github.com/mattn/go-colorable:0.1.6"
       linkage: "STATIC"
-    - id: "Go::github.com/mattn/go-isatty:v0.0.12"
+    - id: "Go::github.com/mattn/go-isatty:0.0.12"
       linkage: "STATIC"
     - id: "Go::golang.org/x/sys:05986578812163b26672dabd9b425240ae2bb0ad"
       linkage: "STATIC"
 packages:
-- id: "Go::github.com/fatih/color:v1.9.0"
-  purl: "pkg:golang/github.com%2Ffatih%2Fcolor@v1.9.0"
+- id: "Go::github.com/fatih/color:1.9.0"
+  purl: "pkg:golang/github.com%2Ffatih%2Fcolor@1.9.0"
   declared_licenses: []
   declared_licenses_processed: {}
   description: ""
@@ -53,8 +53,8 @@ packages:
     url: "https://github.com/fatih/color.git"
     revision: "daf2830f2741ebb735b21709a520c5f37d642d85"
     path: ""
-- id: "Go::github.com/mattn/go-colorable:v0.1.6"
-  purl: "pkg:golang/github.com%2Fmattn%2Fgo-colorable@v0.1.6"
+- id: "Go::github.com/mattn/go-colorable:0.1.6"
+  purl: "pkg:golang/github.com%2Fmattn%2Fgo-colorable@0.1.6"
   declared_licenses: []
   declared_licenses_processed: {}
   description: ""
@@ -79,8 +79,8 @@ packages:
     url: "https://github.com/mattn/go-colorable.git"
     revision: "68e95eba382c972aafde02ead2cd2426a8a92480"
     path: ""
-- id: "Go::github.com/mattn/go-isatty:v0.0.12"
-  purl: "pkg:golang/github.com%2Fmattn%2Fgo-isatty@v0.0.12"
+- id: "Go::github.com/mattn/go-isatty:0.0.12"
+  purl: "pkg:golang/github.com%2Fmattn%2Fgo-isatty@0.0.12"
   declared_licenses: []
   declared_licenses_processed: {}
   description: ""
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
@@ -18,66 +18,66 @@ project:
   scopes:
   - name: "main"
     dependencies:
-    - id: "Go::github.com/fatih/color:v1.13.0"
+    - id: "Go::github.com/fatih/color:1.13.0"
       linkage: "PROJECT_STATIC"
       dependencies:
-      - id: "Go::github.com/mattn/go-colorable:v0.1.12"
+      - id: "Go::github.com/mattn/go-colorable:0.1.12"
         linkage: "PROJECT_STATIC"
         dependencies:
-        - id: "Go::github.com/mattn/go-isatty:v0.0.14"
+        - id: "Go::github.com/mattn/go-isatty:0.0.14"
           linkage: "PROJECT_STATIC"
           dependencies:
-          - id: "Go::golang.org/x/sys:v0.0.0-20220610221304-9f5ed59c137d"
+          - id: "Go::golang.org/x/sys:0.0.0-20220610221304-9f5ed59c137d"
             linkage: "PROJECT_STATIC"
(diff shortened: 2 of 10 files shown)


sschuberth:
@fviernau could you please rebase this / fix the *detekt* issues before we start a review?

fviernau:
> @fviernau could you please rebase this / fix the _detekt_ issues before we start a review?

Done.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -145,8 +145,7 @@ class GoMod(
      * Return the module graph output from `go mod graph` with non-vendor dependencies removed.
      */
     private fun getModuleGraph(projectDir: File): Graph {
-        val moduleInfoForModuleName = getModuleInfos(projectDir).map { it.replace ?: it }
-            .associateBy({ it.path }, { it })
+        val moduleInfoForModuleName = getModuleInfos(projectDir).associateBy({ it.path }, { it })
```
sschuberth: Commit message:
- Maybe "for the" -> "of the" (in the title)
- "all use cases want to view" sounds a bit strange IMO, as if use cases were persons. Maybe "all callers assume the 'replace' directive to have been applied" or so?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -229,7 +228,7 @@ class GoMod(
                 }
             }
 
-            return result
+            return result.map { it.replace ?: it }
```
sschuberth, line 235: Maybe adjust the function docs to say "... and the replace directive applied"?

analyzer/src/test/kotlin/managers/GoModTest.kt
```diff
@@ -25,68 +25,71 @@ import io.kotest.matchers.collections.containExactlyInAnyOrder
...
-import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.VcsType
 
 class GoModTest : WordSpec({
     "toVcsInfo" should {
         "return the VCS type 'Git'" {
-            val id = Identifier("Go::github.com/chai2010/gettext-go:v1.0.0")
-
-            id.toVcsInfo().type shouldBe VcsType.GIT
+            GoMod.ModuleInfo(
```
sschuberth, line 33: Mind introducing a `moduleInfo` variable here (like below) even if only a single properly is asserted? That's also how it was before with `id`.

analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
```diff
@@ -18,66 +18,66 @@ project:
   scopes:
   - name: "main"
     dependencies:
-    - id: "Go::github.com/fatih/color:v1.13.0"
+    - id: "Go::github.com/fatih/color:1.13.0"
```
sschuberth, line 21: Commit message: Missing dot at the end of the last sentence.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -76,6 +76,10 @@ class GoMod(
 
     companion object {
         const val DEFAULT_GO_PROXY = "https://proxy.golang.org"
+        fun normalizeModuleVersion(moduleVersion: String): String =
```
sschuberth, line 79: Missing blank line before this one.

sschuberth: What's the reason to put this function inside `GoMod`, i.e. why does it fit better here than in `GoDep`? Can we find a more "neutral" place?

sschuberth: And the code comment should now ideally become function docs.

analyzer/src/funTest/assets/projects/synthetic/godep-expected-output.yml
```diff
@@ -18,17 +18,17 @@ project:
   scopes:
   - name: "default"
     dependencies:
-    - id: "Go::github.com/fatih/color:v1.9.0"
+    - id: "Go::github.com/fatih/color:1.9.0"
+nglide-expected-output.yml      linkage: "STATIC"
```
sschuberth, line 22: This looks broken.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Can you recommend a better place?

fviernau: I ended up creating a GoSupport file.

sschuberth: If that function would be the only content of a new `GoSupport` file, I'd also be fine with following your rationale to put / keep it in `GoMod` as that's the most modern / current package manager for Go.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -211,7 +210,8 @@ class GoMod(
     )
 
     /**
-     * Return the list of all modules contained in the dependency tree with resolved versions.
+     * Return the list of all modules contained in the dependency tree with resolved versions, with the 'replace'
```
sschuberth, line 214: I deliberately proposed to use "and" because the "with ... with" word duplication reads odd, IMO, but I won't insist on a change.

analyzer/src/main/kotlin/managers/utils/GoSupport.kt
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.analyzer.managers.utils
+
+object GoSupport {
```
sschuberth, line 22: We don't wrap functions that could be top-level in an `object` in the other support files, so I wouldn't do so here either. The "namespace" for top-level functions simply is the file name itself.

sschuberth: But like I wrote in the other comment, if this is the only function to introduce this file for, I'd be fine with keeping it at `GoMod`.

analyzer/src/main/kotlin/managers/utils/GoSupport.kt
```diff
@@ -0,0 +1,27 @@
...
+    fun normalizeModuleVersion(moduleVersion: String): String =
+    // Strip any suffix starting with '+', because build metadata is not involved in version comparison, see
+        // https://go.dev/ref/mod#incompatible-versions.
```
sschuberth, line 25: Wrong indentation, and it should still become the function documentation now.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/utils/GoSupport.kt
fviernau: > We don't wrap functions that could be top-level in an `object` in the other support files, so I wouldn't do so here either. 

Hm, just because we pollute the namespace elsewhere we have to follow that approach.
What advantage do you see in putting these into an object? ..and why do you think the advantage is worth less than (global) consistency in this case?

>The "namespace" for top-level functions simply is the file name itself.

Isn't the namespace for non-private top level functions the Java / Kotlin package?

fviernau: Done anyway.

analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
MarcelBochtler, line 21: Commit message nit: You're never referencing the [1].

MarcelBochtler (approved):

PR #5587, opened 2022-07-20
MavenDependencyHandler: Improve the checks for project dependencies

### LINKED ISSUE #5581: Maven analyzer fails due to duplicate identifiers for packages and projects
Recently, a check has been added to `AnalyzerResultBuilder` to verify that a `PackageManager` does not produce projects and packages with the same identifier. This check failed for one of our projects using `Maven`.

The constellation is a multi-module Maven build. All projects in the build are assigned a SNAPSHOT version number (0-SNAPSHOT to be precise). In the dependency graph, one of these projects is referenced, but with a specific, non-SNAPSHOT version number. What happens when constructing the dependency graph is the following:

* `DependencyGraphBuilder` is called for each dependency and delegates to `DependencyHandler` to construct a corresponding `Package` instance.
* `MavenDependencyHandler` has a set of identifiers that represent local projects. When called for the dependency to the project with the specific version, it recognizes that this is not a local project, since the version does not match. It therefore delegates to `MavenSupport` to actually create the `Package`.
* In `MavenSupport.parsePackage()` Maven functionality is used to construct a Maven project for the dependency. This is successful in this case; however, the version number of the generated Maven project is not the one of the dependency, but the one of the local project.
* Based on the Maven project constructed in the previous step, a `Package` is now created with the wrong version causing the duplicate to the local project.

I assume the change of the version number is due to some conflict resolution done by Maven. Since the declared dependencies in the multi-module build are obviously legal, this special case should be handled by the Maven analyzer.

oheger-bosch:
Under certain circumstances, the Maven analyzer could create a package
with the same identifier as a local project. This caused a sanity
check in AnalyzerResultBuilder to fail. To prevent this, improve the
tests whether a package corresponds to a local project in
MavenDependencyHandler. Also make sure that for the affected package
the correct linkage is returned.

Resolves #5581.

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
@@ -48,13 +48,19 @@ class MavenDependencyHandler(
      * A map with information about the local projects in the current Maven build. Dependencies pointing to projects
      * sometimes need to be treated in a special way.
      */
-    private val localProjects: Map<String, MavenProject>,
+    localProjects: Map<String, MavenProject>,
 
     /**
      * A flag whether [SBT compatibility mode][Maven.enableSbtMode] is enabled.
      */
     private val sbtMode: Boolean
 ) : DependencyHandler<DependencyNode> {
+    /**
+     * A set of identifiers that are known to point to local projects. This is updated for packages that are resolved
+     * to projects.
+     */
+    private val localProjectIds = localProjects.keys.toMutableSet()
+
     override fun identifierFor(dependency: DependencyNode): Identifier =
         Identifier(
             type = if (isLocalProject(dependency.identifier())) managerName else "Maven",
@@ -78,11 +84,26 @@ class MavenDependencyHandler(
     override fun linkageFor(dependency: DependencyNode): PackageLinkage =
         if (isLocalProject(dependency)) PackageLinkage.PROJECT_DYNAMIC else PackageLinkage.DYNAMIC
 
+    /**
+     * Create a [Package] representing a [dependency] if possible, recording any [issues]. Inter-project
+     * dependencies are skipped.
+     */
     override fun createPackage(dependency: DependencyNode, issues: MutableList<OrtIssue>): Package? {
         if (isLocalProject(dependency)) return null
 
         return runCatching {
-            support.parsePackage(dependency.artifact, dependency.repositories, localProjects, sbtMode)
+            support.parsePackage(dependency.artifact, dependency.repositories, sbtMode = sbtMode).let { pkg ->
+                // There is the corner case that a dependency references a project, but in a different version than
+                // the one used by the local build. Then, this dependency is actually a package, but Maven's
+                // resolution mechanism might prefer using the project. Therefore, the check whether the dependency
+                // is a project must be done after the package resolution again.
+                if (isLocalProject(pkg.id)) {
+                    localProjectIds += dependency.identifier()
+                    null
+                } else {
+                    pkg
+                }
+            }
         }.onFailure { e ->
             e.showStackTrace()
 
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
```diff
@@ -78,11 +84,23 @@ class MavenDependencyHandler(
     override fun linkageFor(dependency: DependencyNode): PackageLinkage =
         if (isLocalProject(dependency)) PackageLinkage.PROJECT_DYNAMIC else PackageLinkage.DYNAMIC
 
+    /**
+     * Create a [Package] representing [dependency] if possible, adding potential issues to [issues]. Inter-project
```
sschuberth, line 88: Nits:
- Maybe "the" / "a" before "dependency"?
- Maybe try to avoid the "issues" word duplication, like "... recording any [issues]." (Or "tracking", "reporting".)

analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
```diff
@@ -78,11 +84,23 @@ class MavenDependencyHandler(
...
+     * dependencies are skipped. There is the corner case that a dependency references a project, but in a different
+     * version than the one used by the local build. Then, this dependency is actually a package, but Maven's
+     * resolution mechanism might prefer using the project. Therefore, the check whether [dependency] is a project
```
sschuberth, line 91: Again "the" / "a" before "dependency"?

analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
```diff
@@ -78,11 +84,23 @@ class MavenDependencyHandler(
...
+     * must be done before and after the package resolution.
```
sschuberth, line 92: Why is not only after resolution enough?

analyzer/src/test/kotlin/managers/utils/MavenDependencyHandlerTest.kt
```diff
@@ -169,6 +169,38 @@ class MavenDependencyHandlerTest : WordSpec({
             issues should beEmpty()
         }
 
+        "return null for a package that is resolved to a project dependency" {
+            val dependency = createDependency(IDENTIFIER)
+            val artifact = dependency.artifact
+            val repos = listOf(createRepository(), createRepository())
+            val pkg = createPackage(PROJECT_ID)
+            val issues = mutableListOf<OrtIssue>()
```
fviernau, line 177: maybe, issues could be inlined.

fviernau (dismissed):

analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
```diff
@@ -78,11 +84,23 @@ class MavenDependencyHandler(
...
     override fun createPackage(dependency: DependencyNode, issues: MutableList<OrtIssue>): Package? {
         if (isLocalProject(dependency)) return null
 
         return runCatching {
-            support.parsePackage(dependency.artifact, dependency.repositories, localProjects, sbtMode)
+            support.parsePackage(dependency.artifact, dependency.repositories, sbtMode = sbtMode).let { pkg ->
+                if (isLocalProject(pkg.id)) {
+                    localProjectIds += dependency.identifier()
+                    null
+                } else pkg
```
sschuberth, line 102: Nit: When `if` is a block, `else` should be a block, too, for symmetry / readability.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
oheger-bosch: As proposed by @fviernau, this part of the comment has been moved to a code comment; it has also been slightly reworded.

PR #5588, opened 2022-07-20
OsvFunTest: Update expected test results

sschuberth:
Add more references and reorder results as expected.

### DIFF
--- a/advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
+++ b/advisor/src/funTest/assets/retrieve-package-findings-expected-result.json
@@ -10,6 +10,41 @@
     },
     "defects" : [ ],
     "vulnerabilities" : [ {
+      "id" : "GHSA-cj88-88mr-972w",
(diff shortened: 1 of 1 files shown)


oheger-bosch (approved):
Can it be that this test requires updates very frequently?

sschuberth:
> Can it be that this test requires updates very frequently?

At least recently, yes. Maybe it's not a good idea to use NPM packages for vulnerability tests, as these have new findings all the way long 😉

fviernau:
> Yes, different package could help. Any recommendations for the ecosystem / package to use?

Or refactoring the test so that it checks that the expected vulnerabilities are contained in the result instead of an equality check.  I'll take care of making it less flaky soon. IIRC I've chosen older versions of the package with the hope that vulnerabilities don't change too often.

sschuberth:
> Any recommendations for the ecosystem / package to use?

Anything Rust / Cargo maybe?

PR #5590, opened 2022-07-20
OSV: Enable ruby gems

fviernau:
See individual commits.

Context: #3599.


advisor/src/funTest/kotlin/OsvFunTest.kt
```diff
@@ -38,12 +38,12 @@ class OsvFunTest : StringSpec({
     "retrievePackageFindings() returns vulnerabilities for the supported ecosystems" {
         val osv = createOsv()
         val packages = listOf(
-            "PyPI::Plone:3.2",
+            "Crate::sys-info:0.7.0",
```
sschuberth, line 41: Can you explain in the commit message why? Is this just for readability / consistency, or a preparation for making the test order-agnostic?

sschuberth (changes requested):

PR #2128, opened 2020-01-17
Draft: Implement support for PNPM

adambhere:
Relates-to: #2024

Signed-off-by: Adam Bukolt <adam.bukolt@here.com>

PR #5594, opened 2022-07-20
Add Analyzer for PNPM using `pnpm list` approach

MarcelBochtler:
This PR adds support for PNPM.
PNPM is a package manager for node projects, see https://pnpm.io/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,7 @@ ENV \
     GO_VERSION=1.18.3 \
     HASKELL_STACK_VERSION=2.1.3 \
     NPM_VERSION=8.5.0 \
+    PNPM_VERSION=7.5.1 \
     PYTHON_PIPENV_VERSION=2018.11.26 \
     PYTHON_POETRY_VERSION=1.1.13 \
     PYTHON_VIRTUALENV_VERSION=15.1.0 \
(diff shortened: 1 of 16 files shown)


MarcelBochtler:
Unfortunately, this implementation does not work for real-life projects due to this bug: https://github.com/pnpm/pnpm/issues/3518

I will have to look into another way how to get the dependencies incl. workspaces support.

sschuberth:
> I will have to look into another way how to get the dependencies incl. workspaces support.

Is it maybe easier to contribute the fix to *pnpm* rather than finding a work-around on our side?

MarcelBochtler:
> > I will have to look into another way how to get the dependencies incl. workspaces support.
> 
> Is it maybe easier to contribute the fix to _pnpm_ rather than finding a work-around on our side?

Yes, that might be possible. However, in a project that I was still able to analyze, using `pnpm list` resulted in a ~300 MB JSON. So this approach might not scale for multiple thousands of NPM dependencies.

The approach I'm currently implementing is comparable to the workspaces approach used for `yarn`, which analyzes the installed packages by their `node_modules`.
I already started that approach while trying to reuse the implementation in #2128. The issues I encountered may be solvable.
Before, I stopped, because the approach with using the tool itself to analyze its dependency seemed better, than analyzing the `node_modules` manually.

MarcelBochtler:
This PR has some advantages, compared to the approach in #5611:
- It uses PNPM itself to analyze the dependencies instead of parsing the `node_modules` directory
- Submodules are shown as projects and the packages are correctly assigned to the projects. I.e. this PR uses the same approach as Yarn 2+ does, while #5611 produces the same result as Yarn (classic) workspace projects.

However, this implementation is currently not usable for real-world projects due to the PNPM bug: https://github.com/pnpm/pnpm/issues/3518.

Therefore, closing this in favor of #5611.

Issue #5593, opened 2022-07-20
advisor: Extend the `Vulnerability` model


fviernau:
The Vulnerability data model is currently quite minimalistic. This can make a lot of sense, things information can
probably obtained by following the links, so it avoids redundancy.

However, the following properties may add value when reflected in ORT's model:
1. summary: To quickly get an idea what it is about
2. description: To get some more idea what it is about
3. credits: To be a good citizen (currently there is no data for that field available)
4. published / modified / withdrawn: to get an idea from when this is

A bit more complicated:
1. aliases (and handling of deduplication), e.g. which IDs refer to the same thing

PR #5599, opened 2022-07-21
advisor: Extend `Vulnerability` by the properties `summary` and `description`

fviernau:
Add the properties and implement the mapping for all advise providers where applicable.
See individual commits.

Context: #5593.


fviernau:
@oheger-bosch due to the revert commit, your approval should be mendatory to merge this as per our discussion about Bosch's internal setup.

model/src/main/kotlin/Vulnerability.kt
```diff
@@ -31,6 +34,18 @@ data class Vulnerability(
      */
     val id: String,
 
+    /**
+     * A short description of this [Vulnerability].
```
sschuberth, line 38: I believe we shouldn't repeat the word "description" here to differentiate from the below. Just say "short summary".

model/src/main/kotlin/Vulnerability.kt
```diff
@@ -31,6 +34,18 @@ data class Vulnerability(
...
+     */
+    @JsonInclude(Include.NON_NULL)
+    val summary: String? = null,
+
+    /**
+     * A description of this [Vulnerability].
```
sschuberth, line 44: To complement the above, I'd write "full description" here.

reporter/src/funTest/kotlin/TestData.kt
```diff
@@ -452,8 +452,8 @@ val ORT_RESULT = OrtResult(
 )
 
 val VULNERABILITY = Vulnerability(
-    "CVE-2021-1234",
-    listOf(
+    id = "CVE-2021-1234",
```
oheger-bosch, line 455: Would it make sense to add the new attributes to the test data?

fviernau: This test data is only used for testing the evaluator currently. 
I wouldn't want to add a test to the evaluator which covers `summary` or `description` as I believe these attributes are irrelevant for the rules, e.g. they are free text. Without adding a test case to the evaluator I believe it doesn't make sense to add to the test data. Are you ok with leaving it as-is?

oheger-bosch: Ok.

oheger-bosch (approved):

PR #5605, opened 2022-07-22
MavenSupport: Fix a potential NPE in a log statement

oheger-bosch:
A ProjectBuildingException obviously does not always contain a
defined pomFile. This causes a NullPointerException in the log
output.

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -101,6 +101,12 @@ import org.ossreviewtoolkit.utils.spdx.SpdxOperator
 
 fun Artifact.identifier() = "$groupId:$artifactId:$version"
 
+/**
+ * Return the path to this file or a corresponding message if the file is unknown.
+ */
+private val File?.safePath: String
+    get() = this?.invariantSeparatorsPath ?: "<unknown file>"
+
 class MavenSupport(private val workspaceReader: WorkspaceReader) {
     companion object {
         private const val MAX_DISK_CACHE_SIZE_IN_BYTES = 1024L * 1024L * 1024L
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -446,15 +446,16 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
...
                             "Still continuing with the incompletely built project '${resultForPomFile.projectId}' at " +
-                            "'${resultForPomFile.pomFile.invariantSeparatorsPath}': ${e.collectMessages()}"
+                            "'${resultForPomFile.pomFile?.invariantSeparatorsPath}': ${e.collectMessages()}"
                 }
 
                 resultForPomFile
             } else {
                 log.error {
-                    "Failed to build project '${e.projectId}' at '${e.pomFile.invariantSeparatorsPath}': " +
+                    "Failed to build project '${e.projectId}' at '${e.pomFile?.invariantSeparatorsPath}': " +
```
sschuberth, line 458: Could / should we do something nicer and maybe say something like "unknown file" instead of "null"? Because in case of "null" the user does not know what entity "null" refers to.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -101,6 +101,18 @@ import org.ossreviewtoolkit.utils.spdx.SpdxOperator
...
+ * Return the path to this file or a corresponding message if the file is unknown.
+ */
+private val File?.safePath: String
+    get() = this?.invariantSeparatorsPath ?: "<unknown file>"
+
+/**
+ * Return the path to the pom file affected by this exception handling an undefined file gracefully.
+ */
+private val ProjectBuildingException.pomPath: String
+    get() = pomFile.safePath
```
sschuberth, line 114: I believe this helper function is a bit overkill, though, and I'd simply use `e.pomFile.safePath` below for clarity that the path is *safe*.

sschuberth (approved):

PR #5608, opened 2022-07-25
Bundler improvements

sschuberth:


sschuberth:
@itrich this should fix your issue.

analyzer/src/main/kotlin/managers/Bundler.kt
```diff
@@ -73,6 +73,11 @@ private const val ROOT_DEPENDENCIES_SCRIPT = "scripts/bundler_root_dependencies.
  */
 private const val RESOLVE_DEPENDENCIES_SCRIPT = "scripts/bundler_resolve_dependencies.rb"
 
+/**
```
fviernau, line 76: Commit-message: Would it be worth mentioning what issue the previous implementation had which the new hasn't anymore ?

analyzer/src/main/kotlin/managers/Bundler.kt
```diff
@@ -163,7 +169,15 @@ class Bundler(
                 scopeDependencies = scopes.toSortedSet()
             )
 
-            val packages = gemSpecs.values.mapTo(sortedSetOf()) { getPackageFromGemspec(it) }
+            val allProjectDeps = groupedDeps.values.flatten().toSet()
+            val uniqueHelperDeps = HELPER_SCRIPT_DEPENDENCIES.filterNot { it in allProjectDeps }
```
fviernau, line 173: When reading `uniqueHelperDeps` I first thought `unque` means `distinct`. But, I believe that it actually means, dependencies which are only helper dependencies but not project dependencies.
How about renaming this to `helperOnlyDependencies` or `dependenciesToRemove`.

analyzer/src/main/kotlin/managers/Bundler.kt
```diff
@@ -163,7 +169,15 @@ class Bundler(
...
+
+            val packages = gemSpecs.values.mapNotNullTo(sortedSetOf()) { gemSpec ->
+                // Bundler itself always shows up as a package because the helper script requires it, but it should be
```
fviernau, line 176: This comment assumes that the hard-coded list only contains bundler which is in fact true, but on the other hand the commit is about not hard-coding it. So, how about making the comment more generic?

Further on, would it make sense to avoid "shows up as a package"? 
However, I propose to consider dropping the comment without replacement, as it is quite redundant / obvious from the code IMO.

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Bundler.kt
sschuberth: I was hoping that's already implied by the current commit message:

> Do not hard-code "bundler"

The previous implementation hard-code (only) the "bundler" implementation, which obviously does not work anymore if the dependencies required by the helper scripts ever change.

> consider all project scopes, not just the runtime scope.

So previously, only the "runtime" scope was considered.

What information exactly are you missing from here?

analyzer/src/main/kotlin/managers/Bundler.kt
sschuberth: I'll go for `helperOnlyDeps`.

analyzer/src/main/kotlin/managers/Bundler.kt
sschuberth: I'll drop the comment altogether then.

analyzer/src/main/kotlin/managers/Bundler.kt
fviernau: > @itrich this should fix your issue.

I was wondering what was mentioned here.

sschuberth: See https://ort-talk.slack.com/archives/C9NNJ54B1/p1658731298911489?thread_ts=1658512604.176269&cid=C9NNJ54B1.

fviernau (approved):

PR #5611, opened 2022-07-26
Add Analyzer for PNPM

MarcelBochtler:
This PR adds support for [PNPM](https://pnpm.io/).

The implementation is loosely based on #2128.
In contrast to #5594 it does not use `pnpm list` to avoid an out of memory issue for large projects, see: https://github.com/pnpm/pnpm/issues/3518.

This implementation relies on the `pnpm install` `--shamefully-hoist` option to create a similar `node_modules` directory structure as NPM and Yarn create it, and then reuses most of NPM's analyzer implementation.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -225,7 +225,6 @@ open class Npm(
         analyzerConfig.getPackageManagerConfiguration(managerName)?.options?.get(OPTION_LEGACY_PEER_DEPS)
             .toBoolean()
 
-    private val artifactoryApiPathPattern = Regex("(.*artifactory.*)(?:/api/npm/)(.*)")
     private val graphBuilder = DependencyGraphBuilder(NpmDependencyHandler(this))
 
     protected open fun hasLockFile(projectDir: File) = hasNpmLockFile(projectDir)
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -29,6 +29,7 @@ import com.vdurmont.semver4j.Requirement
 import java.io.File
 import java.io.IOException
 import java.util.SortedSet
+import java.util.concurrent.ConcurrentHashMap
 
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
@@ -228,6 +229,22 @@ open class Npm(
     private val graphBuilder = DependencyGraphBuilder(NpmDependencyHandler(this))
 
     protected open fun hasLockFile(projectDir: File) = hasNpmLockFile(projectDir)
+
+    /**
+     * Load the submodule directories of the project defined in [moduleDir].
+     */
+    protected open fun loadWorkspaceSubmodules(moduleDir: File): List<File> {
+        val nodeModulesDir = moduleDir.resolve("node_modules")
+        if (!nodeModulesDir.isDirectory) return emptyList()
+
+        val searchDirs = nodeModulesDir.walk().maxDepth(1).filter {
+            it.isDirectory && it.name.startsWith("@")
+        }.toList() + nodeModulesDir
+
+        return searchDirs.flatMap { dir ->
+            dir.walk().maxDepth(1).filter { it.isDirectory && it.isSymbolicLink() }.toList()
+        }
+    }
 
     override fun command(workingDir: File?) = if (Os.isWindows) "npm.cmd" else "npm"
 
@@ -448,6 +465,17 @@ open class Npm(
         return jsonMapper.readTree(process.stdout)
     }
 
+    /** Cache for submodules identified by its moduleDir absolutePath */
+    private val submodulesCache: ConcurrentHashMap<String, List<File>> = ConcurrentHashMap()
+
+    /**
+     * Find the directories which are defined as submodules of the project within [moduleDir].
+     */
+    protected fun findWorkspaceSubmodules(moduleDir: File): List<File> =
+        submodulesCache.getOrPut(moduleDir.absolutePath) {
+            loadWorkspaceSubmodules(moduleDir)
+        }
+
     /**
      * Retrieve all the dependencies of [project] from the given [scopes] and add them to the dependency graph under
      * the given [targetScope]. Return the target scope name if dependencies are found; *null* otherwise.
@@ -481,19 +509,6 @@ open class Npm(
             id = Identifier(managerName, namespace, name, ""),
             issues = listOf(issue)
         )
-    }
-
-    private fun findWorkspaceSubmodules(moduleDir: File): List<File> {
-        val nodeModulesDir = moduleDir.resolve("node_modules")
-        if (!nodeModulesDir.isDirectory) return emptyList()
-
-        val searchDirs = nodeModulesDir.walk().maxDepth(1).filter {
-            it.isDirectory && it.name.startsWith("@")
-        }.toList() + nodeModulesDir
-
-        return searchDirs.flatMap { dir ->
-            dir.walk().maxDepth(1).filter { it.isDirectory && it.isSymbolicLink() }.toList()
-        }
     }
 
     private fun getModuleDependencies(moduleDir: File, scopes: Set<String>): Set<NpmModuleInfo> {
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -228,7 +228,18 @@ open class Npm(
 
     private val graphBuilder = DependencyGraphBuilder(NpmDependencyHandler(this))
 
+    /**
+     * Search depth in the `node_modules` directory for `package.json` files used for collecting all packages of the
+     * projects.
+     */
+    protected open val modulesSearchDepth = Int.MAX_VALUE
+
     protected open fun hasLockFile(projectDir: File) = hasNpmLockFile(projectDir)
+
+    /**
+     * Check if [this] represents a workspace within a `node_modules` directory.
+     */
+    protected open fun File.isWorkspaceDir() = isSymbolicLink()
 
     /**
      * Load the submodule directories of the project defined in [moduleDir].
@@ -317,7 +328,7 @@ open class Npm(
 
         logger.info { "Searching for 'package.json' files in '$nodeModulesDir'..." }
 
-        val nodeModulesFiles = nodeModulesDir.walk().filter {
+        val nodeModulesFiles = nodeModulesDir.walk().maxDepth(modulesSearchDepth).filter {
             it.name == "package.json" && isValidNodeModulesDirectory(nodeModulesDir, nodeModulesDirForPackageJson(it))
         }
 
@@ -395,7 +406,7 @@ open class Npm(
 
         val id = Identifier("NPM", namespace, name, version)
 
-        if (packageDir.isSymbolicLink()) {
+        if (packageDir.isWorkspaceDir()) {
             val realPackageDir = packageDir.realFile()
 
             logger.debug { "The package directory '$packageDir' links to '$realPackageDir'." }
--- a/analyzer/src/main/kotlin/managers/Pnpm.kt
+++ b/analyzer/src/main/kotlin/managers/Pnpm.kt
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2019-2020 HERE Europe B.V.
+ * Copyright (C) 2022 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.analyzer.managers
+
+import com.fasterxml.jackson.databind.JsonNode
+
+import com.vdurmont.semver4j.Requirement
+
+import java.io.File
+
+import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
+import org.ossreviewtoolkit.analyzer.managers.utils.hasPnpmLockFile
+import org.ossreviewtoolkit.analyzer.managers.utils.mapDefinitionFilesForPnpm
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.model.jsonMapper
+import org.ossreviewtoolkit.utils.common.Os
+import org.ossreviewtoolkit.utils.common.realFile
+
+/**
+ * The [fast, disk space efficient package manager](https://pnpm.io/).
+ */
+class Pnpm(
+    name: String,
+    analysisRoot: File,
+    analyzerConfig: AnalyzerConfiguration,
+    repoConfig: RepositoryConfiguration
+) : Npm(name, analysisRoot, analyzerConfig, repoConfig) {
+    class Factory : AbstractPackageManagerFactory<Pnpm>("PNPM") {
+        override val globsForDefinitionFiles = listOf("package.json", "pnpm-lock.yaml")
+
+        override fun create(
+            analysisRoot: File,
+            analyzerConfig: AnalyzerConfiguration,
+            repoConfig: RepositoryConfiguration
+        ) = Pnpm(managerName, analysisRoot, analyzerConfig, repoConfig)
+    }
+
+    /**
+     * PNPM symlinks workspace modules in the `node_modules` directory, which will result in cyclic symlinks of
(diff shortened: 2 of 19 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -449,6 +449,19 @@ open class Npm(
         return jsonMapper.readTree(process.stdout)
     }
 
+    protected open fun findWorkspaceSubmodules(moduleDir: File): List<File> {
```
oheger-bosch, line 458: Since the method is now *protected*, it becomes part of the API of the class and should therefore be documented - especially if it is designed for extension. It would be helpful to describe when this function is called as part of the resolving processes and how its results are further processed.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -301,7 +301,10 @@ open class Npm(
 
         log.info { "Searching for 'package.json' files in '$nodeModulesDir'..." }
 
-        val nodeModulesFiles = nodeModulesDir.walk().filter {
+        // PNPM symlinks workspace modules in the `node_modules` directory, which will result in cyclic symlinks of
+        // `node_module` directories. Limit the search depth to '2' in this case.
+        val searchDepth = if (managerName == "PNPM") 2 else Int.MAX_VALUE
```
oheger-bosch, line 306: I am not sure how much effort it means, but it may be possible to avoid these special treatments for PNPM by designing the `Npm` base class accordingly for being extended. For instance, `Npm` could define a function/property `modulesSearchDepth`, which could be overridden by `Pnpm`.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -504,7 +510,18 @@ open class Npm(
             addAll(getModuleInfo(moduleDir, scopes)!!.dependencies)
 
             workspaceModuleDirs.forEach { workspaceModuleDir ->
-                addAll(getModuleInfo(workspaceModuleDir, scopes, listOf(moduleDir))!!.dependencies)
+                val childNodeModulesDir = workspaceModuleDir.resolve("node_modules")
+                if (childNodeModulesDir.isDirectory && managerName == "PNPM") {
```
oheger-bosch, line 514: Similar to my other comment about extensibility of `Npm`. If `Npm` provided a `processModuleDir()` function (or similar), `Pnpm` could override it and apply the stashing before calling the super function.

analyzer/src/main/kotlin/managers/Pnpm.kt
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (C) 2019-2020 HERE Europe B.V.
```
oheger-bosch, line 2: Question: Is the copyright correct? Since the class is newly created, a date range starting with 2019 seems strange.

analyzer/src/main/kotlin/managers/utils/NodeSupport.kt
```diff
@@ -213,3 +245,46 @@ private fun getWorkspaceMatchers(definitionFile: File): List<PathMatcher> {
...
+    val result = mutableSetOf<File>()
+
+    definitionFiles.forEach { definitionFile ->
+        val pnpmWorkspacesFile = definitionFile.parentFile.resolve("pnpm-workspace.yaml")
+
+        if (pnpmWorkspacesFile.isFile) {
+            val pathMatchers = getPnpmWorkspaceMatchers(definitionFile)
+
+            pathMatchers.forEach { matcher ->
+                definitionFiles.forEach inner@{ other ->
```
oheger-bosch, line 276: Nit: I find these `return` statements to labels very hard to read - they are a bit like the `goto` style in BASIC. Is there an alternative to express this logic?

analyzer/src/main/kotlin/managers/Npm.kt
fviernau, line 306: I believe that adding case distinctions for child classes to the parent class is at the minimum a code smell. 
I believe it should at least follow the open closed principle, e.g. that the behavior is injected by the child class, so that the parent class doesn't know about it. The same applies to other places below.

Apart from that, have you considered using aggregation instead of inheritance? E.g. factor out a couple of functions from NPM to NpmSupport and then make use of them from (1) Npm (2) Yarn and (3) Pnpm. This would IMO reduce complexity quite a bit and make it easy to make changes to e.g. Npm, without the need to understand all the Pnpm and Yarn details.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -301,7 +301,10 @@ open class Npm(
 
         log.info { "Searching for 'package.json' files in '$nodeModulesDir'..." }
 
-        val nodeModulesFiles = nodeModulesDir.walk().filter {
+        // PNPM symlinks workspace modules in the `node_modules` directory, which will result in cyclic symlinks of
+        // `node_module` directories. Limit the search depth to '2' in this case.
```
fviernau, line 305: Why is `2` a reasonable search depth? e.g. why doesn't it miss anything?

analyzer/src/main/kotlin/managers/utils/NodeSupport.kt
```diff
@@ -125,35 +142,50 @@ fun expandNpmShortcutUrl(url: String): String {
...
     val hasYarn2ResourceFile: Boolean = false,
     val hasNpmLockfile: Boolean = false,
+    val hasPnpmLockfile: Boolean = false,
+    val isPnpmWorkspaceRoot: Boolean = false,
+    val isPnpmWorkspaceSubmodule: Boolean = false,
     val isYarnWorkspaceRoot: Boolean = false,
     val isYarnWorkspaceSubmodule: Boolean = false
 )
 
+private fun isHandledByPnpm(entry: PackageJsonInfo) =
```
fviernau, line 160: previously, isHandledByYarn and isHandledByNpm have been mutually exclusive.
Now, with the three functions, they are not mutually excusive anymore. 
Have you considered that, e.g. isHandledByPnpm and isHandledByYarn return true both? Can it lead to problems?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Pnpm.kt
MarcelBochtler: I took _some_ inspiration for this class from https://github.com/oss-review-toolkit/ort/pull/2128. That's why I decided it would be appropriate to keep the copyright.

analyzer/src/main/kotlin/managers/Npm.kt
MarcelBochtler: Because the `pnpm install` option `--shamefully-hoist` lifts all dependencies to the top level `node_modules` directory.
That means that all `package.json` files will be located in a direct subdirectory of `node_modules` which results in a required search depth of `2` in the case of `PNPM`.

fviernau: Could you please update the code comment to reflect that?

MarcelBochtler: Did that in the newly introduced method.

analyzer/src/main/kotlin/managers/utils/NodeSupport.kt
MarcelBochtler: This is decided based on the content of this object:
```kt
PackageJsonInfo(
    definitionFile = definitionFile,
    isPnpmWorkspaceRoot = isPnpmWorkspaceRoot(definitionFile.parentFile),
    isYarnWorkspaceRoot = isYarnWorkspaceRoot(definitionFile) && !isPnpmWorkspaceRoot(definitionFile.parentFile),
    hasYarnLockfile = hasYarnLockFile(definitionFile.parentFile),
    hasNpmLockfile = hasNpmLockFile(definitionFile.parentFile),
    hasPnpmLockfile = hasPnpmLockFile(definitionFile.parentFile),
    hasYarn2ResourceFile = hasYarn2ResourceFile(definitionFile.parentFile),
    isPnpmWorkspaceSubmodule = definitionFile in pnpmWorkspaceSubmodules,
    isYarnWorkspaceSubmodule = definitionFile in yarnWorkspaceSubmodules && definitionFile !in pnpmWorkspaceSubmodules
)
```

There are three cases when it is unclear if the project is managed by Yarn or by PNPM:
1. If the PNPM project also stores the workspaces in the root package.json, then `yarnWorkspaceSubmodules` will contain actual PNPM submodules. I worked around that by checking both:
`isYarnWorkspaceSubmodule = definitionFile in yarnWorkspaceSubmodules && definitionFile !in pnpmWorkspaceSubmodules`
2. Same as (1) for the `workspaceRoot`:
`isYarnWorkspaceRoot = isYarnWorkspaceRoot(definitionFile) && !isPnpmWorkspaceRoot(definitionFile.parentFile),`
3. If the directory contains both, a `yarn.lock` file and a `pnpm-lock.yaml` file. In this case, ORT would try to analyze the project with both package managers. Do you have a suggestion how to handle that case?

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -379,7 +393,9 @@ open class Npm(
 
         val id = Identifier("NPM", namespace, name, version)
 
-        if (packageDir.isSymbolicLink()) {
+        // While Yarn uses symbolic links only for workspace modules, PNPM uses symbolic links for all dependencies.
+        // Do not handle symbolic links differently when PNPM is using this method.
+        if (packageDir.isSymbolicLink() && managerName != "PNPM") {
```
oheger-bosch, line 398: A potential way to handle this special treatment of PNPM could be to introduce a function `isWorkspaceDir(dir: File): Boolean` (or similar) in `Npm`. This could be overridden in `Pnpm`.

analyzer/src/main/kotlin/managers/Pnpm.kt
```diff
@@ -0,0 +1,119 @@
...
+    override fun hasLockFile(projectDir: File) = hasPnpmLockFile(projectDir)
+
+    override fun processWorkspaceModuleDirs(
+        workspaceModuleDir: File,
+        scopes: Set<String>,
+        moduleDir: File
+    ): Set<NpmModuleInfo> {
+        val childNodeModulesDir = workspaceModuleDir.resolve("node_modules")
+
+        return buildSet {
```
fviernau, line 74: I believe moving childNodeModulesDir into `buildSet` and then turning `processWorkspaceModuleDirs()` into an expression function would read a tad nicer.

analyzer/src/main/kotlin/managers/Pnpm.kt
```diff
@@ -0,0 +1,119 @@
...
+            if (childNodeModulesDir.isDirectory) {
+                // PNPM creates a separate `node_modules` directory in all workspace modules. As these dependencies
+                // are already known to the dependency tree through the `--shamefully-hoist` option, this behavior
+                // results in duplicated package entries. Stash the child `node_modules` directory to not analyze
+                // it.
+                stashDirectories(childNodeModulesDir).use {
```
fviernau, line 80: Would it make sense to move this logic into the NPM class, e.g. make the NPM class always stash the directory, without checking whether the instance is PNPM, Yarn, NPM or anything else?
So that the function does not need to be overridden?

analyzer/src/main/kotlin/managers/utils/NodeSupport.kt
```diff
@@ -213,3 +245,46 @@ private fun getWorkspaceMatchers(definitionFile: File): List<PathMatcher> {
...
     }.orEmpty()
 }
+
+private fun getPnpmWorkspaceMatchers(definitionFile: File): List<PathMatcher> {
+    data class PnpmWorkspaces(val packages: List<String>)
+
+    val pnpmWorkspaceFile = definitionFile.parentFile.resolve("pnpm-workspace.yaml")
+
+    return if (pnpmWorkspaceFile.isFile) {
+        val workspaceMatchers = pnpmWorkspaceFile.readValue<PnpmWorkspaces>().packages
```
fviernau, line 255: Is `packages` the only attribute possible? If not, I believe we should configure that unknown attributes are to be ignored in order to not run into an exception when parsing the file.

analyzer/src/test/kotlin/PackageManagerTest.kt
```diff
@@ -51,9 +51,9 @@ class PackageManagerTest : WordSpec({
         "gradle-kotlin/build.gradle.kts",
         "maven/pom.xml",
 
-        // Note that the NPM and Yarn implementations share code. Internal logic decides dynamically whether to process
-        // "package.json" with NPM or Yarn.
-        "npm-and-yarn/package.json",
+        // Note that the NPM, PNPM and Yarn implementations share code. Internal logic decides dynamically whether to
+        // process "package.json" with NPM or Yarn.
```
fviernau, line 55: "with NPM or Yarn." <-- PNPM needs to be added there as well.

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Pnpm.kt
MarcelBochtler: While I tried to implement this, I found that the PNPM implementation had an issue with the following project setup:
```
root
- package-a
  - chalk@^5.0.0
- package-b
  - chalk@^4.0.0
```
Only one `chalk` version is hoisted to the root `node_modules` directory.
Therefore, I had to remove the stashing of the child `node_modules` directories, and solve the issue with duplicated dependencies in the dependency graph by overriding `equals` in `NpmModuleInfo`.

analyzer/src/main/kotlin/managers/Pnpm.kt
MarcelBochtler: No longer relevant with https://github.com/oss-review-toolkit/ort/pull/5611#discussion_r936594137.

analyzer/src/main/kotlin/managers/Pnpm.kt
MarcelBochtler: I also enhanced the workspaces test to test this project setup.

analyzer/src/main/kotlin/managers/Npm.kt
MarcelBochtler: Done. This also fixed a separate handling a few lines down:
```
val isWorkspaceSubmodule = packageFile.parentFile.realFile() in findWorkspaceSubmodules(workingDir)
if (hasIncompleteData && !isWorkspaceSubmodule) {
```

analyzer/src/main/kotlin/managers/utils/NodeSupport.kt
fviernau: The handling make sense to me, thanks! Regarding point #3 I don't have any proposal either.

analyzer/src/main/kotlin/managers/Pnpm.kt
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * PNPM symlinks workspace modules in the `node_modules` directory, which will result in cyclic symlinks of
+     * `node_module` directories. Limit the search depth to '2' in this case as all packages are in a direct
+     * subdirectory of the `node_modules` directory, thanks to the `--shamefully-hoist` install option.
+     */
+    override val modulesSearchDepth = 2
+
+    override fun hasLockFile(projectDir: File) = hasPnpmLockFile(projectDir)
+
+    override fun File.isWorkspaceDir() = realFile() in findWorkspaceSubmodules(analysisRoot)
```
oheger-bosch, line 66: A call to `findWorkspaceSubmodules()` could be expensive; so if this function is called for multiple files, the detected submodules should better be cached.

analyzer/src/main/kotlin/managers/Pnpm.kt
```diff
@@ -0,0 +1,101 @@
...
+        checkVersion()
+
+    override fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
+        val process = run(workingDir, "view", "--json", packageName)
+
+        return jsonMapper.readTree(process.stdout)
+    }
+
+    override fun findWorkspaceSubmodules(moduleDir: File): List<File> =
+        submodulesCache.getOrPut(moduleDir.absolutePath) {
```
oheger-bosch, line 96: Ideally, the subclass does not need to know about caching. This can be achieved as follows:
The `findWorkspaceSubmodules()` function in `Npm` is made non-open again. It checks whether the requested path is already in the cache. If not, it delegates to a new function `loadWorkspaceSubmodules()` (or whatever name fits) that does the actual retrieving of modules. The result is then stored in the cache and returned.

`Pnpm` only overrides this `load` function. The map with the cache can then be private in `Npm`. Then only care has to be taken that `Pnpm` never invokes the `load` function itself, but accesses all submodules via the `findWorkspaceSubmodules()` function.

analyzer/src/main/kotlin/managers/utils/NpmDependencyHandler.kt
```diff
@@ -47,7 +47,26 @@ data class NpmModuleInfo(
...
     /** A set with information about the modules this module depends on. */
     val dependencies: Set<NpmModuleInfo>
-)
+) {
+    /**
+     * [workingDir] and [packageFile] are not relevant when adding this [NpmModuleInfo] to the dependency graph.
+     * However, if these values differ the same dependencies are added as duplicates to the set which is used to create
+     * the dependency graph. Therefore, remove them from the equals check.
+     */
+    override fun equals(other: Any?): Boolean {
```
oheger-bosch, line 56: One idea as an alternative to overriding `equals()` could be to not use a `Set` in `Npm.getModuleInfo()` when determining the dependencies, but a `Map` using a combined key with only the relevant properties (the identifier and the dependencies). I am not sure, however, whether this is really a significant improvement over the current approach.

oheger-bosch (approved):

PR #5615, opened 2022-07-27
A couple of code inspection fixes

fviernau:
See individual commits.

### DIFF
--- a/clients/osv/src/main/kotlin/OsvApiClient.kt
+++ b/clients/osv/src/main/kotlin/OsvApiClient.kt
@@ -87,7 +87,7 @@ interface OsvApiClient {
 }
 
 @Serializable
-data class VulnerabilitiesForPackageRequest private constructor(
+class VulnerabilitiesForPackageRequest private constructor(
     val commit: String? = null,
     @SerialName("package")
     val pkg: Package? = null,
--- a/downloader/src/main/kotlin/vcs/GitRepo.kt
+++ b/downloader/src/main/kotlin/vcs/GitRepo.kt
@@ -194,7 +194,7 @@ class GitRepo : VersionControlSystem(), CommandLineTool {
             // Switching manifest branches / revisions requires running "init" again.
             runRepo(workingTree.workingDir, "init", *manifestOptions.toTypedArray())
 
-            // Repo allows to checkout Git repositories to nested directories. If a manifest is badly configured, a
+            // Repo allows to check out Git repositories to nested directories. If a manifest is badly configured, a
             // nested Git checkout overwrites files in a directory of the upper-level Git repository. However, we still
             // want to be able to download such projects, so specify "--force-sync" to work around that issue.
             val syncArgs = mutableListOf("sync", "-c", "--force-sync")
--- a/model/src/main/kotlin/AdvisorRecord.kt
+++ b/model/src/main/kotlin/AdvisorRecord.kt
@@ -65,12 +65,14 @@ data class AdvisorRecord(
          * encountered problems. Therefore, support an optional filter for a [capability] of the advisor that produced
          * a result.
          */
-        fun resultsWithIssues(minSeverity: Severity = Severity.HINT, capability: AdvisorCapability? = null):
-                    AdvisorResultFilter =
-            { result ->
-                (capability == null || capability in result.advisor.capabilities) &&
-                        result.summary.issues.any { it.severity >= minSeverity }
+        fun resultsWithIssues(
+            minSeverity: Severity = Severity.HINT,
+            capability: AdvisorCapability? = null
+        ): AdvisorResultFilter = { result ->
+            (capability == null || capability in result.advisor.capabilities) && result.summary.issues.any {
+                it.severity >= minSeverity
             }
+        }
     }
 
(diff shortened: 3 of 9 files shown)


scanner/src/main/kotlin/PathScanner.kt
```diff
@@ -212,7 +212,7 @@ abstract class PathScanner(
         val missingArchives = mutableSetOf<Pair<Package, KnownProvenance>>()
 
         scanResults.forEach { (pkg, results) ->
-            results.forEach { (provenance, _, _) ->
+            results.forEach { (provenance, _, _, _) ->
```
sschuberth, line 215: Nit: I believe this could be squashed to the previous commit, so that the same type of inspection hint is addressed in a single commit even if it occurs in multiple files.

clients/osv/src/main/kotlin/OsvApiClient.kt
```diff
@@ -87,7 +87,7 @@ interface OsvApiClient {
 }
 
 @Serializable
-data class VulnerabilitiesForPackageRequest private constructor(
+class VulnerabilitiesForPackageRequest private constructor(
```
sschuberth, line 90: > Use a normal `class` instead of a `data class` to not generate the
`copy()` function.

Please explain in the commit message *why* not generating the `copy()` function is a desirable thing here.

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -130,7 +130,7 @@ class ExperimentalScanner(
         return ortResult.copy(scanner = scannerRun)
     }
 
-    suspend fun scan(packages: Set<Package>, context: ScanContext): Map<Identifier, List<ScanResult>> {
+    fun scan(packages: Set<Package>, context: ScanContext): Map<Identifier, List<ScanResult>> {
```
sschuberth, line 133: Commit message: Please explain briefly *why* the modifier is redundant here.

model/src/main/kotlin/AdvisorRecord.kt
```diff
@@ -65,12 +65,14 @@ data class AdvisorRecord(
...
-                (capability == null || capability in result.advisor.capabilities) &&
-                        result.summary.issues.any { it.severity >= minSeverity }
-            }
+        fun resultsWithIssues(
+            minSeverity: Severity = Severity.HINT,
+            capability: AdvisorCapability? = null
+        ): AdvisorResultFilter = { result ->
+            (capability == null || capability in result.advisor.capabilities) && result.summary.issues.any {
+                it.severity >= minSeverity
+            }W
```
sschuberth, line 74: Drop "W".

sschuberth (changes requested):

clients/osv/src/main/kotlin/OsvApiClient.kt
sschuberth, line 90: Commit message nits:
- "with invalid parameter is still" -> "with invalid parameter*s* *was* still"
- "So, a normal" -> "So, *use* a normal"

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
sschuberth, line 133: Commit message proposal: "The function does no *need* to be suspendable as it does not call any other `suspend` functions."

downloader/src/main/kotlin/vcs/GitRepo.kt
```diff
@@ -194,7 +194,7 @@ class GitRepo : VersionControlSystem(), CommandLineTool {
             // Switching manifest branches / revisions requires running "init" again.
             runRepo(workingTree.workingDir, "init", *manifestOptions.toTypedArray())
 
-            // Repo allows to checkout Git repositories to nested directories. If a manifest is badly configured, a
+            // Repo allows to check out Git repositories to nested directories. If a manifest is badly configured, a
```
sschuberth, line 197: Nit (sorry for not mentioning this earlier): Similar to the other commits that fix the same type of inspection hint, this could also be squashed to the previous commit as a general "typo fixes in docs" commit.

sschuberth (changes requested):

downloader/src/main/kotlin/vcs/GitRepo.kt
fviernau: Would you agree that this nit is actually quite a bit subjective? E.g. it could be squashed, but it also could remain as is? So, is your comment optional or a must fix?

sschuberth: I agree it's subjective, but it would follow the idea to fix the same kind of inspection hint in one commit throughout the code base. As you followed that rationale earlier, I thought you're fine with doing it here, too. But it's up to you.

fviernau: I feared if I follow it again I may end-up with some unspoken agreement to always do this in the future, e.g. go through my commit stash and see what typo fixes can be squashed. So, I feared that this might introduce additional effort where I don't see always a value in it.

sschuberth (approved):

Issue #5621, opened 2022-07-28
Evaluator: Add support for "repolinter" style rules


tsteenbe:
Goal: Have ORT better support common contribution process checks 

Usage scenarios:
-  Check community file such as CONTRIBUTING.md, Code of Conduct, CODEOWNERS, pull request templates are present in your open source projects
- Check if a project has a 'docs' and 'example' directories
- Check if each C/C++ project follow agreed up ways of working scheme such as
   - All open source C/C++ code is included in an 'external' directory.
   - Within the external directory each package is contain within its own directory e.g. '/external/curl'
   - Each package directory has package.spdx.yml and LICENSE
   - Package known only for testing such as GoogleTest have been excluded via path exclude
   - Throw deprecation policy violation for deprecated package /external/legacy-sdk
- Checking Git commits for organization's secrets or naming
- Check if a CI is used

Came up with below API,  inspired by https://github.com/todogroup/repolinter/blob/main/docs/rules.md. 

- `hasCI()` checks for presence of known CI files within project being scanned such as .gitlab-ci.yml or .github/workflows 
-  `hasMoreContributorsThan(integer)` checks the number of unique names in version control log (Git log) of the project being scanned. 
-  `hasDirectory([glob/regex?])` checks for presence of directory within project being scanned 
-  `hasFile([glob/regex?])` to check for the presence of LICENSE, CONTRIBUTING.md, CHANGELOG.md, SECURITY.md, SUPPORT.md, NOTICE, CODE-OF-CONDUCT and README e.g. make sure the FOSS project has minimum files present to clarify its governance.  I would like to be able to write a policy rule to check if project’s license is APACHE-2.0 then a NOTICE file must be present 
-  `hasFileWithContents([filepath glob/regex?, contents regex?])` checks if the contents of a file matches a given regular expression. Useful for searching for internal names or urls. 
- `hasFileWithExtension([glob/regex?])` checks for presence of file with certain extensions within project being scanned e.g. hasFileWithExtension(‘.ttf’) to search for font files. See also https://github.com/todogroup/repolinter/blob/main/docs/rules.md#file-starts-with  
- `hasFileWithLicense([filepath glob/regex, spdx license id])` to check for presence of file with certain license within project being scanned 
- `usesGit()` checks whether the project being scanned is managed by Git. 
- `usesGitSubModules()` checks whether the project being scanned uses Git sub modules. 
- `usesSVN()` checks whether the project being scanned is managed by Subversion. 
[...]

PR #5623, opened 2022-07-28
RuleSet: Eliminate redundant default values for constructor paramaters

fviernau:
Currently, instances of `RuleSet` get constructed in two ways:

1. Tests call the constructor directly.
2. `rules.kts` calls the functions `ruleSet()` which in turn creates the
   instance.

Both, the constructor as well as the mentioned function, have
(redundant) default parameter values. Change the construction in tests
to also utilize `ruleSet()` and eliminate the now unnecessary default
values for the constructor parameters.

Context of #5621.

### DIFF
--- a/evaluator/src/test/kotlin/DependencyRuleTest.kt
+++ b/evaluator/src/test/kotlin/DependencyRuleTest.kt
@@ -26,7 +26,7 @@ import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageReference
 
 class DependencyRuleTest : WordSpec() {
-    private val ruleSet = RuleSet(ortResult)
+    private val ruleSet = ruleSet(ortResult) { }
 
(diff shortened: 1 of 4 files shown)


evaluator/src/test/kotlin/DependencyRuleTest.kt
```diff
@@ -26,7 +26,7 @@ import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageReference
 
 class DependencyRuleTest : WordSpec() {
-    private val ruleSet = RuleSet(ortResult)
+    private val ruleSet = ruleSet(ortResult) { }
```
sschuberth, line 29: How about moving `{ }` to the `ruleSet` function as a default argument for the `configure` parameter?

tsteenbe (approved):

PR #5635, opened 2022-08-04
helper-cli: Add a command for converting an ORT file

fviernau:


helper-cli/src/main/kotlin/commands/ConvertOrtFileCommand.kt
```diff
@@ -0,0 +1,55 @@
+/*
```
oheger-bosch, line 1: It took me a while to figure out that "convert" in this context means the transformation to another file format. Maybe this should be explicitly mentioned in the commit message.

helper-cli/src/main/kotlin/commands/ConvertOrtFileCommand.kt
```diff
@@ -0,0 +1,55 @@
...
+        "--input-ort-file", "-i",
+        help = "The ORT file to convert."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+
+    private val outputOrtFile by option(
+        "--output-ort-file", "-o",
+        help = "The ORT file to write to."
```
oheger-bosch, line 44: It probably makes sense to state here explicitly that the file extension used for the output file determines the output format.

I assume the use case to have untypical file extensions is not worth to be supported, right? This would be possible with a `--target-format` option that takes precedence over the file extension.

sschuberth: This reminds me that we already have a generic transform helper command, or? Would it make sense to instead extend that command to use an identity transformation by default, and to determine the target format via the file extension?

helper-cli/src/main/kotlin/commands/ConvertOrtFileCommand.kt
fviernau: I've applied @sschuberth proposal to use the transform command.

helper-cli/src/main/kotlin/commands/ConvertOrtFileCommand.kt
fviernau: This didn't quite work out. My use case was converting 'ort/cli/src/funTest/assets/semver4j-analyzer-result.yml' to 'json' and then reading it via `org.ossreviewtoolkit.helper.common.UtilsKt.readOrtResult()`. It runs into the following issue, while using the mapper for (de-)serializing it works.

``` 
Exception in thread "main" com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Could not resolve subtype of [simple type, class org.ossreviewtoolkit.model.config.ScanStorageConfiguration]: missing type id property '@c' (for POJO property 'storages')
 at [Source: (File); line: 234, column: 9]
```

fviernau: I ended up reverting back to the previous approach with a separate command as the transform command seems not to utilize ORT's custom serialization code which does not fit the needs.

oheger-bosch: ISTR that the transform command had some limitations anyway regarding the support of enhanced YAML features, but don't know any more for sure. So I am fine with reverting.

oheger-bosch (approved):

Issue #4637, opened 2021-10-28
Improve resolution of Python / PIP dependencies


sschuberth:
ORT's analyzer has various problems with resolving Python / PIP dependencies

- [x] Dependencies on native packages require native system tools to be installed, see #4578.
- [x] The Python-3-compatibility check might fail, see #4289.
- [x] Any specifically requested Python version is not adhered to, see #3671.
- [x] We use some rather obscure helper scripts based on abandoned projects, see #2816.
- [x] We might have some general problems with retrieving metadata:
  - [x] #812
  - [x] #509
  - [x] #485
  - [x] #5159
  - [x] #5740

sschuberth:
> or leveraging / extending existing tools like https://github.com/ddelange/pipgrip.

See in particular https://github.com/ddelange/pipgrip/issues/40.

sschuberth:
Also maybe worth a look as a helper tool is https://github.com/trailofbits/it-depends which claims to

> Finds native dependencies for high level languages like Python

pombredanne:
@sschuberth re:

> Also maybe worth a look as a helper tool is https://github.com/trailofbits/it-depends which claims to
>> Finds native dependencies for high level languages like Python

From a quick look  they seem to:
1. create a docker image in https://github.com/trailofbits/it-depends/blob/8f8988330239c6d3eb39f05988fdbe6802f4bbbe/it_depends/pip.py#L35
2. run pip directly https://github.com/trailofbits/it-depends/blob/8f8988330239c6d3eb39f05988fdbe6802f4bbbe/it_depends/pip.py#L176 or through https://github.com/wimglenn/johnnydep/blob/master/johnnydep/pipper.py

sschuberth:
Also see the [difficulties in finding Python 2 example projects](https://github.com/oss-review-toolkit/ort/issues/4901#issuecomment-1002644154).

sschuberth:
We could also take a deeper look at [component-detection's approach for PIP](https://github.com/microsoft/component-detection/blob/main/docs/detectors/pip.md).

sschuberth:
Some [interesting insights](https://github.com/dependabot/dependabot-core/pull/2281#discussion_r457554686) on the general topic from a Python maintainer.

sschuberth:
And yet another interesting [discussion](https://discuss.python.org/t/list-dependencies-of-a-package/12341) with links to:
* https://github.com/spack/spack
* https://github.com/thoth-station/solver
[...]

Issue #3671, opened 2021-02-23
Improve Python support to adhere to Python versions


fviernau:
For Python projects it is hard (impossible) to determine to which Python version they apply.
ORT has some mechanism which tries (1) Python 2.x and (2) Python 3. In particular, for the latter Python 3.6 is used.

When installing dependencies via `pip install -r requirements.txt` from _pypi.org_, pip considers only the dependencies which are compatible with the used Python version, 3.6.

For example in a Python 3.7 project this can lead to the following two issues:

1. A dependency cannot be resolved by ORT, while it is actually resolvable using the appropriate Python version. This
    happens if   none of the versions allowed by the constraints is compatible with Python 3.7.
2. Even worse, the wrong dependency version is resolved. This happens e.g. if the youngest 3.6 compatible version allowed by   
    the constraints does not equal the youngest 3.7 compatible version. In the worst case it can happen that the project uses the 
    latest version of a dependency while ORT resolves a really old (Python 3.6 compatible) version of that dependency.

sschuberth:
At least for `setup.py`-based projects, there *might* be *hints* which Python version to use. For example, the classifiers may contain something like

    classifiers=[
        "Programming Language :: Python",
        "Programming Language :: Python :: 3.6",
        "Programming Language :: Python :: 3.7",
        "Programming Language :: Python :: 3.8",
    ],

This at least tells you the minimum version to use. Similarly, a line like

    python_requires=">=3.6",

might be contained. However, both types of entries are a bit inconvenient to parse.

But even if the Python version to use would be known, another problem then is that you actually require *that* version to be installed. At least for our `Dockerfile` it's impractical to install all sorts of Python versions beforehand, just in case they might be needed. Maybe bootstrapping the right Python version on the fly would be an option, but doing that right for all supported platform is a lot of effort.

A completely different idea that I have in mind is to generally switch over to using GraalVM for ORT, and leverage its [Python support](https://www.graalvm.org/reference-manual/python/) to [run pip programmatically](https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program), and somehow "pretending" to be the right Python version.

sschuberth:
Ping @pombredanne, being the Python guy here, do you have any good idea how to solve this problem?

pombredanne:
@sschuberth a way is to rely on `python_requires=">=3.6"` in modern setup.cg or setup.py. Classifiers are weak information otherwise. 

[...]

PR #5645, opened 2022-08-11
Replace pipdeptree with python-inspector #4637 #3671

pombredanne:
This PR replaces pipdeptree with python-inspector to resolve Python packages dependencies found in requirement files. 

python-inspector can resolve dependencies for any target Python version and OS (and not only the one running the tool).
In this integration in ORT, it replaces pipdeptree pretty much in place as python-inspector implements a similar output data structure by design to ease the integration.

Reference: https://github.com/nexB/python-inspector
Reference: https://github.com/oss-review-toolkit/ort/issues/4637
Reference: https://github.com/oss-review-toolkit/ort/issues/3671
Signed-off-by: Tushar Goel <tushar.goel.dav@gmail.com>
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,9 @@ ARG CRT_FILES=""
 # Set this to the ScanCode version to use.
 ARG SCANCODE_VERSION="30.1.0"
 
+# Set this to the Python Inspector version to use.
+ARG PYTHON_INSPECTOR_VERSION="0.6.4"
+
 FROM eclipse-temurin:11-jdk-jammy AS build
 
 COPY . /usr/local/src/ort
@@ -155,6 +158,9 @@ RUN /opt/ort/bin/import_proxy_certs.sh && \
 # Add scanners (in versions known to work).
 ARG SCANCODE_VERSION
 RUN pip install --no-cache-dir scancode-toolkit==$SCANCODE_VERSION
+
+ARG PYTHON_INSPECTOR_VERSION
+RUN pip install --no-cache-dir python-inspector==$PYTHON_INSPECTOR_VERSION
 
 FROM run AS dist
 
--- a/analyzer/src/funTest/assets/projects/synthetic/python-inspector-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/python-inspector-expected-output.yml
@@ -1,0 +1,226 @@
+---
+project:
+  id: "PIP::src/funTest/assets/projects/synthetic/python-inspector/requirements.txt:<REPLACE_REVISION>"
+  definition_file_path: "analyzer/src/funTest/assets/projects/synthetic/python-inspector/requirements.txt"
+  declared_licenses: []
+  declared_licenses_processed: {}
+  vcs:
+    type: ""
+    url: ""
+    revision: ""
+    path: ""
+  vcs_processed:
+    type: "Git"
+    url: "<REPLACE_URL>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  homepage_url: ""
+  scopes:
+  - name: "install"
+    dependencies:
+    - id: "PyPI::flask:1.0"
+      dependencies:
+      - id: "PyPI::click:6.7"
+      - id: "PyPI::itsdangerous:0.24"
+      - id: "PyPI::jinja2:2.11.3"
+        dependencies:
+        - id: "PyPI::markupsafe:1.0"
+      - id: "PyPI::werkzeug:0.15.3"
+packages:
+- id: "PyPI::click:6.7"
+  purl: "pkg:pypi/click@6.7"
+  authors:
+  - "Armin Ronacher"
+  declared_licenses:
+  - "BSD License"
+  declared_licenses_processed:
+    unmapped:
+    - "BSD License"
+  description: "A simple wrapper around optparse for powerful command line utilities."
+  homepage_url: "http://github.com/mitsuhiko/click"
+  binary_artifact:
+    url: "https://files.pythonhosted.org/packages/34/c1/8806f99713ddb993c5366c362b2f908f18269f8d792aff1abfd700775a77/click-6.7-py2.py3-none-any.whl"
+    hash:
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -28,6 +28,9 @@ ARG CRT_FILES=""
 # Set this to the ScanCode version to use.
 ARG SCANCODE_VERSION="30.1.0"
 
+# Set this to the Python Inspector version to use.
```
mnonnenmacher, line 31: Commit message: Please add the details from the PR description to the commit message as well, to make sure it is recorded in the Git log. Also, please prefix the title with "Pip: Replace ...".

analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -183,6 +178,23 @@ class Pip(
 
     override fun transformVersion(output: String) = output.removePrefix("pip ").substringBefore(' ')
 
+    private fun runPythonInspector(
```
mnonnenmacher, line 181: Please add a class that overrides `CommandLineTool` instead, like for example `VirtualEnv` in the same file. This allows it to define supported versions (if required) and will make the tool appear in the output of the `RequirementsCommand`.

analyzer/src/funTest/kotlin/managers/PipFunTest.kt
```diff
@@ -99,6 +99,21 @@ class PipFunTest : WordSpec() {
 
                 result.toYaml() shouldBe expectedResult
             }
+
+            "capture metadata using python-inspector" {
```
mnonnenmacher, line 103: Does this test add anything compared to the other tests from the same class? Because python-inspector is now used for the other tests as well.

analyzer/src/funTest/assets/projects/synthetic/python-inspector-expected-output.yml
```diff
@@ -0,0 +1,226 @@
+---
```
mnonnenmacher, line 1: Commit message: Please use a proper title like "PipFunTest: Add a test for python-inspector".

mnonnenmacher (changes requested):

oheger-bosch:
There are two (trivial) issues in the static code analysis: Multiple blank lines at lines 64 and 197.

analyzer/src/funTest/kotlin/managers/PipFunTest.kt
oheger-bosch: It would be good to have a test scenario that cannot be handled by the original pip-based implementation, but is correctly resolved by python-inspector.

pombredanne: And we are updating the test to represent something different and uniquely handled by this new tool

Dockerfile
pombredanne: Done

oheger-bosch:
The changes look good to me, however, the commits probably need to be reworked. As stated in the [contributor guide](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md), requested changes should not be added in new commits, but the existing commits should be amended. Could you please do this?

pombredanne:
> The changes look good to me, however, the commits probably need to be reworked. As stated in the [contributor guide](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md), requested changes should not be added in new commits, but the existing commits should be amended. Could you please do this?

@oheger-bosch Thanks! My mistake there.

analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -318,27 +339,41 @@ class Pip(
...
-        if (pipdeptree.isSuccess) {
-            val fullDependencyTree = jsonMapper.readTree(pipdeptree.stdout)
+        if (pythonInspector.isSuccess) {
+            val fullDependencyTree = jsonMapper.readTree(jsonFile)
+            jsonFile.parentFile.safeDeleteRecursively(force = true)
 
             val projectDependencies = if (definitionFile.name == "setup.py") {
-                // The tree contains a root node for the project itself and pipdeptree's dependencies are also at the
-                // root next to it, as siblings.
+                // The tree contains a root node for the project itself.
```
mnonnenmacher, line 368: @pombredanne This assumption is not correct anymore with the python-inspector. I tested it locally and the python-inspector output is:

```json
[
  {
    "key": "ply",
    "package_name": "ply",
    "installed_version": "3.11",
    "dependencies": []
  },
  {
    "key": "rdflib",
    "package_name": "rdflib",
    "installed_version": "6.2.0",
    "dependencies": [
      {
        "key": "isodate",
        "package_name": "isodate",
        "installed_version": "0.6.1",
        "dependencies": [
          {
            "key": "six",
            "package_name": "six",
            "installed_version": "1.16.0",
            "dependencies": []
          }
        ]
      },
      {
        "key": "pyparsing",
        "package_name": "pyparsing",
        "installed_version": "3.0.9",
        "dependencies": []
      },
      {
        "key": "setuptools",
        "package_name": "setuptools",
        "installed_version": "65.2.0",
        "dependencies": []
      }
    ]
  }
]
```

So it does not contain a root node with the project name. Therefore the function returns `EMPTY_JSON_NODE` and the spdx-tools-python test is failing. The solution is to simplify lines 367-379 to:

```kotlin
            val projectDependencies = fullDependencyTree.filterNot {
                isPhonyDependency(it["package_name"].textValue(), it["installed_version"].textValueOrEmpty())
            }
```

(and to remove the `projectName` parameter of the function because it is unused now)

The test will still fail because "38" is always passed to python-inspector as Python version and therefore different versions of the dependencies are found. I would suggest to ignore this issue for now and just update the expected results file, and I will address this topic in a separate PR as discussed in the meeting today.

Dockerfile
mnonnenmacher: Please also remove the ticket ids from the commit message title, it is enough to mention them in the body.

mnonnenmacher (approved):

PR #5649, opened 2022-08-13
Switch to Gradle's "bin" distribution to reduce the download size

### LINKED ISSUE #5589: Switch from the "all" to the standard "bin" Gradle distribution
According to [this comment](https://discuss.gradle.org/t/gradle-upgrade-from-v4-7-to-v7-3-3-gives-failed-to-notify-project-evaluation-listener-error/41844/2) from @Vampire there is no value in using Gradle's "all" vs. the standard "bin" distribution when using Kotlin DSL for the build scripts, which is what we do.

Thus I propose to switch from "all" to "bin"; but first want to check esp. with @mnonnenmacher, who in the past advocated for using "all". So, objections anyone @oss-review-toolkit/core-devs?

sschuberth:
Resolves #5589.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,8 +40,6 @@ RUN --mount=type=cache,target=/tmp/.gradle/ \
     export GRADLE_USER_HOME=/tmp/.gradle/ && \
     scripts/import_proxy_certs.sh && \
     scripts/set_gradle_proxy.sh && \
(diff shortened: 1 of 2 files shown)


MarcelBochtler (approved):

sschuberth:

PR #5653, opened 2022-08-15
Conan: Pass the definition file name when calling `conan info`

mnonnenmacher:
If "conanfile.txt" and "conanfile.py" both exist in the same repository,
calling `conan.info` defaults to using "conanfile.py". This causes
issues if the currently analyzed file is "conanfile.txt". To fix this,
pass the name of the currently analyzed definition file to "conan info".

### DIFF
--- a/analyzer/src/main/kotlin/managers/Conan.kt
+++ b/analyzer/src/main/kotlin/managers/Conan.kt
@@ -145,7 +145,7 @@ class Conan(
             conanConfig?.also { configureRemoteAuthentication(it) }
 
             val jsonFile = createOrtTempDir().resolve("info.json")
-            run(workingDir, "info", ".", "--json", jsonFile.absolutePath, *DUMMY_COMPILER_SETTINGS)
+            run(workingDir, "info", definitionFile.name, "--json", jsonFile.absolutePath, *DUMMY_COMPILER_SETTINGS)
 
             val pkgInfos = jsonMapper.readTree(jsonFile)
             jsonFile.parentFile.safeDeleteRecursively(force = true)


MarcelBochtler (approved):

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -145,7 +145,7 @@ class Conan(
             conanConfig?.also { configureRemoteAuthentication(it) }
 
             val jsonFile = createOrtTempDir().resolve("info.json")
-            run(workingDir, "info", ".", "--json", jsonFile.absolutePath, *DUMMY_COMPILER_SETTINGS)
+            run(workingDir, "info", definitionFile.name, "--json", jsonFile.absolutePath, *DUMMY_COMPILER_SETTINGS)
```
sschuberth, line 148: Hmm, won't this code now always use `conanfile.txt` if both `conanfile.txt` and `conanfile.py` are present? Because our `globsForDefinitionFiles` are prioritized, and we list `conanfile.txt` before `conanfile.py`.

There's probably nothing wrong with this, as it makes the `conan info` call aligned with the semantics of `globsForDefinitionFiles`. But if that's what you intended, I find the commit message a bit misleading.

mnonnenmacher: Yes, if both are present `PackageManager.findManagedFiles()` will only keep the first file, in this case "conanfile.txt". However, from the point of view of the function this does not matter, it is only important, that it handles the passed `definitionFile` correctly. Programmatically you could call `resolveDependencies(definitionFile: File, ...)` for any file.

PR #5657, opened 2022-08-16
Stop polluting `Any` object with a `logger` property

sschuberth:


nnobelis:
As a sidenote, theses changes will most likely break any existing ORT plugins.

sschuberth:
> As a sidenote, theses changes will most likely break any existing ORT plugins.

I guess they / most of them would still compile (as the `logger` of the base class would implicitly be used), but plugins would need to be adjusted (have `companion object : Logging` added) to list the correct class name in the output.

nnobelis:
> > As a sidenote, theses changes will most likely break any existing ORT plugins.
> 
> I guess they / most of them would still compile (as the `logger` of the base class would implicitly be used), but plugins would need to be adjusted (have `companion object : Logging` added) to list the correct class name in the output.

True, but I was referring to the package renaming.

sschuberth:
> True, but I was referring to the package renaming.

Hmm, if I'm not mistaken the only package renaming happened in the `helper-cli` module, which should not used programmatically anyway.

mnonnenmacher (approved):

PR #5662, opened 2022-08-17
reporter: Add advisor issues to evaluated model

mnonnenmacher:

### DIFF
--- a/reporter/src/funTest/assets/static-html-reporter-test-input.yml
+++ b/reporter/src/funTest/assets/static-html-reporter-test-input.yml
@@ -39,10 +39,10 @@ repository:
       - given: "GPL-2.0-only OR MIT"
         choice: "MIT"
       package_license_choices:
-        - package_id: "Maven:com.h2database:h2:1.4.200"
-          license_choices:
-            - given: "MPL-2.0 OR EPL-1.0"
-              choice: "MPL-2.0"
+      - package_id: "Maven:com.h2database:h2:1.4.200"
+        license_choices:
+        - given: "MPL-2.0 OR EPL-1.0"
+          choice: "MPL-2.0"
 analyzer:
   start_time: "1970-01-01T00:00:00Z"
   end_time: "1970-01-01T00:00:00Z"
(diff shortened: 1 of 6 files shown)


reporter/src/funTest/assets/static-html-reporter-test-input.yml
```diff
@@ -39,23 +39,43 @@ repository:
...
         choice: "MIT"
       package_license_choices:
-        - package_id: "Maven:com.h2database:h2:1.4.200"
-          license_choices:
-            - given: "MPL-2.0 OR EPL-1.0"
-              choice: "MPL-2.0"
+      - package_id: "Maven:com.h2database:h2:1.4.200"
+        license_choices:
+        - given: "MPL-2.0 OR EPL-1.0"
+          choice: "MPL-2.0"
```
sschuberth, line 45: Nit: This seems to be a pure formatting change unrelated to the model changes, but I'm ok with keeping it as part of this commit.

reporter/src/funTest/assets/static-html-reporter-test-input.yml
```diff
@@ -566,8 +584,8 @@ scanner:
           start_time: "1970-01-01T00:00:00Z"
           end_time: "1970-01-01T00:00:00Z"
           package_verification_code: "0000000000000000000000000000000000000000"
-          licenses: [ ]
-          copyrights: [ ]
+          licenses: []
+          copyrights: []
```
sschuberth, line 588: Nit: This seems to be a pure formatting change unrelated to the model changes, but I'm ok with keeping it as part of this commit.

sschuberth (approved):

reporter/src/funTest/assets/static-html-reporter-test-input.yml
mnonnenmacher: That's correct, but I didn't want to spend time finding out which part of the diff is because of manual changes and which is because of model changes (the different sorting is probably also caused by manually added content), so I just regenerated the whole file all at once.

PR #5664, opened 2022-08-17
ExperimentalScanner: Explain why a superfluous `suspend` modifier is kept

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
+++ b/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
@@ -70,6 +70,8 @@ class ExperimentalScanner(
 
     private val archiver = scannerConfig.archive.createFileArchiver()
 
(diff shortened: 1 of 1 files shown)


mnonnenmacher (commented):
I have that modifier there deliberately, because the function will start using coroutines, and so callers already have to be implemented accordingly.

sschuberth:
> I have that modifier there deliberately, because the function will start using coroutines, and so callers already have to be implemented accordingly.

Ok, should I then suppress this inspection hint instead for now?

sschuberth:
> Ok, should I then suppress this inspection hint instead for now?

Done that.

mnonnenmacher:
> > Ok, should I then suppress this inspection hint instead for now?
> 
> Done that.

I have just created #5667 which I wanted to do for a long time, this would make the comment obsolete. I just need to do some testing.

PR #5665, opened 2022-08-17
Minor NodeSupport clean-ups

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -595,4 +595,5 @@ open class Npm(
     }
 }
 
+/** Name of the configuration option to toggle legacy peer dependency support. */
 private const val OPTION_LEGACY_PEER_DEPS = "legacyPeerDeps"
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -105,17 +105,6 @@ open class Npm(
         ) = Npm(managerName, analysisRoot, analyzerConfig, repoConfig)
     }
 
-    companion object {
-        /** Name of the scope with the regular dependencies. */
-        private const val DEPENDENCIES_SCOPE = "dependencies"
-
-        /** Name of the scope with optional dependencies. */
-        private const val OPTIONAL_DEPENDENCIES_SCOPE = "optionalDependencies"
-
-        /** Name of the scope with development dependencies. */
-        private const val DEV_DEPENDENCIES_SCOPE = "devDependencies"
-    }
-
     private val legacyPeerDeps =
         analyzerConfig.getPackageManagerConfiguration(managerName)?.options?.get(OPTION_LEGACY_PEER_DEPS)
             .toBoolean()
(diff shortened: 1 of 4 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -42,9 +41,13 @@ import org.ossreviewtoolkit.analyzer.PackageManagerResult
...
 import org.ossreviewtoolkit.analyzer.managers.utils.NpmModuleInfo
 import org.ossreviewtoolkit.analyzer.managers.utils.expandNpmShortcutUrl
+import org.ossreviewtoolkit.analyzer.managers.utils.fixDownloadUrl
 import org.ossreviewtoolkit.analyzer.managers.utils.hasNpmLockFile
 import org.ossreviewtoolkit.analyzer.managers.utils.mapDefinitionFilesForNpm
-import org.ossreviewtoolkit.analyzer.parseAuthorString
+import org.ossreviewtoolkit.analyzer.managers.utils.parseAuthors
+import org.ossreviewtoolkit.analyzer.managers.utils.parseLicenses
+import org.ossreviewtoolkit.analyzer.managers.utils.parseVcsInfo
+import org.ossreviewtoolkit.analyzer.managers.utils.splitNamespaceAndName
```
mnonnenmacher, line 50: I don't like that these function names have no NPM reference, although the functions are NPM specific. Some, but not all of them, at least mention NPM in the documention. I think this could easily lead to accidental use in the wrong context, for example `fixDownloadUrl` sounds like a generic function.
I would prefer to rename them to e.g. `fixNpmDownloadUrl`, note that the other already imported from `NpmSupport` all have an NPM reference in their name.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -597,3 +586,12 @@ open class Npm(
 
 /** Name of the configuration option to toggle legacy peer dependency support. */
 private const val OPTION_LEGACY_PEER_DEPS = "legacyPeerDeps"
+
```
mnonnenmacher, line 589: Commit message: "compan~~t~~ion"
I am not aware of that convention, do we have it documented somewhere?

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: > I am not aware of that convention, do we have it documented somewhere?

Not in written form. It was a proposal that @fviernau came up with and that we verbally agreed on, AFAIR.

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: > the other already imported from `NpmSupport` all have an NPM reference in their name.

I guess by `NpmSupport` you mean `NodeSupport`? But this actually raises the question on my side whether we should rename `NodeSupport` to `NpmSupport`, as these functions are indeed NPM (= package manager) related, and not generically Node(.js) related. Also, other support classes / files also use the exact name of the package manager as the prefix.

mnonnenmacher: Yes, I agree that `NodeSupport` is the wrong name.

analyzer/src/main/kotlin/managers/Npm.kt
mnonnenmacher: I think we really need to start documenting those, let's talk about it in the next TSC meeting.

mnonnenmacher (approved):

analyzer/src/main/kotlin/managers/Npm.kt
sschuberth: > I think we really need to start documenting those

Would `CONTRIBUTING.md` be a good place for that?

mnonnenmacher: Yes. Except if we want different rules for different repositories, but it would be a start.

PR #5671, opened 2022-08-22
Dependency updates

sschuberth:


gradle/libs.versions.toml
```diff
@@ -14,7 +14,7 @@ versionsPlugin = "0.42.0"
 
 antlr = "4.10.1"
 asciidoctorj = "2.5.5"
-asciidoctorjPdf = "2.1.6"
+asciidoctorjPdf = "2.3.0"
```
MarcelBochtler, line 17: The changelog states:
```
asciidoctorj-pdf 2.3.0 brings the asciidoctor-pdf 2.3.0 gem to AsciidoctorJ.
```

Does this mean that `asciidoctorjPdf` can be removed as a separate dependency?

sschuberth: No. That's just their standard-phrasing to say that they're wrapping the Gem in a plugin for use by AsciidoctorJ.

MarcelBochtler (approved):

PR #5516, opened 2022-07-04
WIP: OrtConfiguration: Introduce connections map to centralize StorageConnections

bs-ondem:
Please ensure that your pull request adheres to our [contribution guidelines](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md). Thank you!

PR #5678, opened 2022-08-23
Change the default ORT configuration file format from Hocon to YAML

mnonnenmacher:


model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -156,3 +156,8 @@ internal data class OrtConfigurationWrapper(
  * A typealias for key-value pairs, used in several configuration classes.
  */
 typealias Options = Map<String, String>
+
+/**
+ * The filename of the reference configuration file.
+ */
+const val REFERENCE_CONFIG_FILENAME = "reference.conf"
```
sschuberth: We have `ORT_CONFIG_FILENAME` in `Constants.kt`. So should we put `REFERENCE_CONFIG_FILENAME` there, too? Or move `ORT_CONFIG_FILENAME` to here?

model/src/main/resources/reference.yml
```diff
@@ -0,0 +1,234 @@
+# A reference configuration file containing all possible ORT configuration options. Some of those options are mutually
```
sschuberth, line 1: The addition of the YAML file raises the question how you converted the file, and how our users are supposed to convert their files. Have you considered to add a `migrate` subcommand to the `helper-cli` that can do the conversion (and possibly assist with other migration tasks in the future)?

README.md
```diff
@@ -188,9 +188,9 @@ The main configuration file for the operation of ORT. This configuration is main
 the ORT instance. In contrast to the configuration files in the following, this file rarely changes once ORT is
 operational.
 
-| Format | Scope  | Default location           |
-|--------|--------|----------------------------|
-| HOCON  | Global | `$ORT_CONFIG_DIR/ort.conf` |
+| Format | Scope  | Default location             |
```
sschuberth, line 191: > The new name "config.yml" is chosen over "ort.yml" to avoid confusion with the repository configuration file ".ort.yml".

Actually, one could also argue that we *do* want to name these files similarly, as they may in parts contain the same configuration options... I just wanted to bring this up to, I'm curious about other opinions from @oss-review-toolkit/core-devs.

docs/config-file-curations-yml.md
```diff
@@ -48,7 +48,7 @@ location of source artifacts.
 
 Hint: If the `concluded_license` _and_ the `authors` are curated, this package will be skipped during the `scan` step,
 as no more information from the scanner is required. This requires the `skipConcluded` scanner option to be enabled in
-the [ort.conf](https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/resources/reference.yml).
+the [config.yml](https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/resources/reference.yml).
```
sschuberth, line 51: Only semi-related: We should probably say "see `reference.yml`" here like in other similar places and not directly link `config.yml` to `reference.yml`, which might be confusing.

Or, thinking about it, do the exact opposite and always link to `config.yml` to `reference.yml` in the other places; there might not be much of a confusion after all, as there *is* no `config.yml` to link to.

docs/hints-for-use-with-docker.md
```diff
@@ -8,7 +8,7 @@
 docker run \
   -v $PWD/:/project \ # Mount current working directory into /project to use as input.
   ort --info \
-  -c /project/ort/config.hocon \ # Use file from "<workingdirectory>/ort" as config.
+  -c /project/ort/config.yml \ # Use file from "<workingdirectory>/ort" as config.
```
sschuberth, line 11: Huh, interesting, the original `config.hocon` file was already wrong. Good that you found this!

README.md
```diff
@@ -467,92 +465,59 @@ By default, the _scanner_ stores scan results on the local file system in the cu
...
+ort:
+  scanner:
+    storages:
+      fileBasedStorage:
+        backend:
+          localFileStorage:
+            directory: "/tmp/ort/scan-results"
+            compression: false
+
+    storageReaders: ["fileBasedStorage"]
```
sschuberth, line 478: Shouldn't this read as

    storageReaders:
    - "fileBasedStorage"

Similar below.

cli/src/main/kotlin/commands/ConfigCommand.kt
```diff
@@ -50,21 +49,25 @@ class ConfigCommand : CliktCommand(name = "config", help = "Show different ORT c
     ).flag()
 
     private val globalOptionsForSubcommands by requireObject<GlobalOptions>()
-    private val renderOptions = ConfigRenderOptions.defaults().setJson(false).setOriginComments(false)
 
-    private fun OrtConfiguration.renderHocon() = toConfig("ort").root().render(renderOptions)
+    private val mapper = YAMLMapper().apply {
+        registerKotlinModule()
+    }
```
sschuberth, line 55: Can't / shouldn't we just use `org.ossreviewtoolkit.model.yamlMapper`?

cli/src/main/kotlin/commands/ConfigCommand.kt
```diff
@@ -74,3 +77,5 @@ class ConfigCommand : CliktCommand(name = "config", help = "Show different ORT c
         }
     }
 }
+
+private data class OrtWrapper(val ort: OrtConfiguration)
```
sschuberth, line 81: How about making `OrtConfigurationWrapper` public and use it instead?

model/src/main/resources/reference.yml
sschuberth: Ok, in the last commit you mention rightfully that we already have `ort config --show-active` which can be used for the conversion. Maybe already mention `ort config --show-reference` in this commit's message, as I assume that's what you've used for the conversion?

sschuberth (changes requested):

model/src/main/resources/reference.yml
mnonnenmacher: I did the conversion by hand to keep the comments and empty lines. It's basically just replacing ` {` with `:` and ` = ` with `: `, and removing `}`.

README.md
mnonnenmacher: YAML allows both and I used the shorter syntax to keep the examples more compact.

cli/src/main/kotlin/commands/ConfigCommand.kt
mnonnenmacher: That's a difficult question to me: On the one hand using the existing `yamlMapper` would be more consistent with other configuration files because it uses snake case for property names. On the other hand I prefer camel case because I like that the names in the configuration file exactly match the names in Kotlin.

model/src/main/resources/reference.yml
sschuberth: Maybe still worth mentioning this in the commit message?

README.md
sschuberth: Interesting. I wonder whether we then can / should use that more compact representation also in ORT result serialization...

cli/src/main/kotlin/commands/ConfigCommand.kt
sschuberth: As I believe we use `yamlMapper` for `.ort.yml` deserialization, and as noted `.ort.yml` and `config.yml` may partly contain the same configuration options, I guess we need to stay consistent and use the same (de-)serializer for both configuration files, so users could copy over part of the global config to the repo config and alter it.

mnonnenmacher: Unfortunately I tend to agree. I'd rather reconsider if we should switch all other file to camel case, but that's too big a change for now...
I need to check if hoplite-yaml can be configured to deserialize snake case, but I would expect that this is possible.

README.md
mnonnenmacher: Jackson does not seem to provide an option for this:

https://github.com/FasterXML/jackson-dataformats-text/issues/43
https://fasterxml.github.io/jackson-dataformats-text/javadoc/yaml/2.13/com/fasterxml/jackson/dataformat/yaml/YAMLGenerator.Feature.html

model/src/main/kotlin/config/OrtConfiguration.kt
mnonnenmacher: I thought about it, but this constant references a file which is part of the model module while the other constants refer to user files, so it made more sense to me to keep it here.

nnobelis:
IMO the fact that ORT main config file has its own format was putting the emphasis about how central this file was.
Additionally, this file was often called `ort.conf` because, hey, this is the central ORT config file. Now we cannot call it `ort.yml` otherwise everyone will be confused with `.ort.yml` and we will need to come up with a new name.

So yeah, this commit is for alignment. Any other benefits of such format change ? If yes, can you list them in the commit message ?

mnonnenmacher:
> IMO the fact that ORT main config file has its own format was putting the emphasis about how central this file was. Additionally, this file was often called `ort.conf` because, hey, this is the central ORT config file. Now we cannot call it `ort.yml` otherwise everyone will be confused with `.ort.yml` and we will need to come up with a new name.

The new proposed default name is `config.yml`, I think that's still specific enough because the file will by default be in a folder called `.ort/config`.

> So yeah, this commit is for alignment. Any other benefits of such format change ? If yes, can you list them in the commit message ?

The main benefit is of course usability, for example some parts of the global configuration can be overwritten in the `.ort.yml` and it is inconvenient that both files use different formats. But this would also unblock #5516 which depends on preserving the order of maps in the configuration which the HOCON parser does not (and likely never will).

cli/src/main/kotlin/commands/ConfigCommand.kt
mnonnenmacher: @sschuberth and me discussed this offline and came to the conclusion that the file will continue using camel case, because there are some places where it cannot use snake case, for example in the scanner options map where the keys need to match the class names exactly. Instead, we would indeed prefer to convert the other files from snake case to camel case, I will create an issue for that with the reasoning.

model/src/main/resources/reference.yml
mnonnenmacher: I have added a comment to the commit message mentioning the config command.

docs/config-file-curations-yml.md
mnonnenmacher: I have added a separate commit that changes those links to point to the README section about the config file instead, because that makes more sense to me.

model/src/main/kotlin/config/OrtConfiguration.kt
sschuberth: I'd still argue it belongs close to `const val ORT_CONFIG_FILENAME = "ort.conf"`, but I won't insist on it.

README.md
sschuberth: > YAML allows both and I used the shorter syntax to keep the examples more compact.

Note to myself: This is called [flow sequences](https://yaml.org/spec/1.2.2/#flow-sequences) in YAML. At least in SnakeYAML directly, which is what Jackson uses, there seems to be a (hacky) way to [configure flow vs. block style](https://stackoverflow.com/a/48281824/1127485).

README.md
sschuberth, line 191: FYI, the file at `integrations/completions/ort-completion.fish` still mentions `ort.conf`. Maybe @MarcelBochtler can comment?

sschuberth (approved):
Approving despite my one comment.

README.md
mnonnenmacher: I have added a commit that updates the scripts.

sschuberth (approved):

PR #5679, opened 2022-08-24
Evaluator: Allow making arbitrary file checks on the project's source tree

fviernau:
See individual commits.

Note: The project's source tree will be cloned in the evaluator only if the rules needed it / if the file checks are actually used.

This implements parts of #5621.

### DIFF
--- a/cli/src/main/kotlin/commands/EvaluatorCommand.kt
+++ b/cli/src/main/kotlin/commands/EvaluatorCommand.kt
@@ -55,6 +55,7 @@ import org.ossreviewtoolkit.cli.utils.outputGroup
 import org.ossreviewtoolkit.cli.utils.readOrtResult
 import org.ossreviewtoolkit.cli.utils.writeOrtResult
 import org.ossreviewtoolkit.evaluator.Evaluator
+import org.ossreviewtoolkit.evaluator.SourceTree
 import org.ossreviewtoolkit.model.FileFormat
 import org.ossreviewtoolkit.model.RuleViolation
 import org.ossreviewtoolkit.model.config.CopyrightGarbage
@@ -310,7 +311,15 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
         val resolutionProvider = DefaultResolutionProvider.create(ortResultInput, resolutionsFile)
         val licenseClassifications =
             licenseClassificationsFile.takeIf { it.isFile }?.readValue<LicenseClassifications>().orEmpty()
-        val evaluator = Evaluator(ortResultInput, licenseInfoResolver, resolutionProvider, licenseClassifications)
+        val projectSourceTree = SourceTree.forRemoteRepository(ortResultInput.repository.vcsProcessed)
+
+        val evaluator = Evaluator(
+            ortResultInput,
+            licenseInfoResolver,
+            resolutionProvider,
+            licenseClassifications,
+            projectSourceTree
+        )
 
         val (evaluatorRun, duration) = measureTimedValue { evaluator.run(script) }
 
--- a/evaluator/src/main/kotlin/SourceTree.kt
+++ b/evaluator/src/main/kotlin/SourceTree.kt
@@ -28,6 +28,7 @@ import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.utils.common.FileMatcher
 import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 
 class SourceTree private constructor(
@@ -52,9 +53,26 @@ class SourceTree private constructor(
                     downloadDir
                 }
             )
+
+        fun forLocalDir(dir: File, vcsType: VcsType) =
+            SourceTree(
+                vcsType = vcsType,
+                getRepositoryRoot = { dir }
+            )
     }
 
     val rootDir by lazy {
         getRepositoryRoot()
+    }
+
+    /**
+     * Return true if any of the given [glob expressions][patterns] match an existing file.
+     */
+    fun hasFile(vararg patterns: String): Boolean {
+        val matcher = FileMatcher(patterns.toSet())
+
+        return rootDir.walkBottomUp().filter {
+            it.isFile && matcher.matches(it.relativeTo(rootDir).path)
+        }.count() > 0
     }
 }
--- a/evaluator/src/main/kotlin/SourceTree.kt
+++ b/evaluator/src/main/kotlin/SourceTree.kt
@@ -98,4 +98,16 @@ class SourceTree private constructor(
             file.takeIf { it.isFile }?.let { it.relativeTo(rootDir).path }
         }.filterTo(mutableSetOf()) { matcher.matches(it) }
     }
+
+    /**
+     * Return the file paths matching any of the given [glob expressions][patterns] with its file content matching
+     * [contentPattern].
+     */
+    fun findFilesWithContent(contentPattern: String, vararg patterns: String) =
+        findFiles(*patterns).filterTo(mutableSetOf()) { path ->
+            val content = rootDir.resolve(path).readText()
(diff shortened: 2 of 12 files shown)


cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -310,7 +311,15 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
         val resolutionProvider = DefaultResolutionProvider.create(ortResultInput, resolutionsFile)
         val licenseClassifications =
             licenseClassificationsFile.takeIf { it.isFile }?.readValue<LicenseClassifications>().orEmpty()
-        val evaluator = Evaluator(ortResultInput, licenseInfoResolver, resolutionProvider, licenseClassifications)
+        val projectSourceTree = SourceTree.forRemoteRepository(ortResultInput.repository.vcsProcessed)
```
mnonnenmacher, line 314: This feature should be optional, because downloading the project source code can add significant time to the evaluator run which is wasted if no rule checks any project files. This should be an option in the ORT configuration like `evaluator.downloadProjectSources`, and in my opinion it should default to false.
An alternative solution would be to make the `SourceTree` lazy in the way that it only downloads the source code once it is accessed.

examples/evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -264,6 +264,14 @@ fun RuleSet.deprecatedScopeExcludeReasonInOrtYmlRule() = ortResultRule("DEPRECAT
     }
 }
 
+fun RuleSet.missingContributingFileRule() = ortResultRule("MISSING_CONTRIBUTING_FILE") {
+    require {
+        -sourceTreeHasFile("CONTRIBUTING.md")
+    }
+
+    error("The project's code repository does not contain the file 'CONTRIBUTING.md' in the root directory.")
```
mnonnenmacher, line 314: Maybe just "... contain the file './CONTRIBUTING.md'." (similar in a later commit)?

mnonnenmacher (changes requested):
You use different prefixes for the example commits: "example.rules", "examples.rules", "example.rules.kts", please align.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
fviernau: >  An alternative solution would be to make the `SourceTree` lazy in the way that it only downloads the source code once it is accessed.

This is actually exactly what I implemented. See `SourceTree.kt` for
```
    val rootDir by lazy {
        getRepositoryRoot()
    }
```

evaluator/src/main/kotlin/SourceTree.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.logging.log4j.kotlin.Logging
+
+import org.ossreviewtoolkit.downloader.Downloader
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.utils.ort.createOrtTempDir
+
+class SourceTree private constructor(
```
sschuberth, line 34: Do we really need this new class? Can't we simply use the existing `WorkingTree` instead?

evaluator/src/main/kotlin/SourceTree.kt
```diff
@@ -52,9 +53,25 @@ class SourceTree private constructor(
                     downloadDir
                 }
             )
+        fun forLocalDir(dir: File, vcsType: VcsType) =
```
sschuberth, line 56: Missing blank line before this one.

evaluator/src/test/kotlin/OrtResultRuleTest.kt
```diff
@@ -0,0 +1,109 @@
...
+
+private fun tempDir() = TempDir()
+
+private class TempDir : File(createOrtTempDir().absolutePath), Closeable {
+    private val dir = createOrtTempDir()
+
+    override fun close() {
+        dir.safeDeleteRecursively()
+    }
+}
```
sschuberth, line 135: You shouldn't need this when using `createSpecTempDir` / `createTestTempDir`.

sschuberth (changes requested):

evaluator/src/main/kotlin/SourceTree.kt
fviernau: The reason I added this class will be more obvious when you look at what get's added in the following commits.
That's all IMO evaluator specific logic. Basically helper functions to use for implementing RuleMatchers and / or policy rules.

sschuberth: I saw the upcoming changes, but I'm still not convinced. Looks like the helper functions would as well operate on a `WorkingTree`.

fviernau: > I saw the upcoming changes, but I'm still not convinced.

let me just try on more time to convince you:

I believe even if these helper functions could all be implemented in WorkingTree encapsulation does make sense, because

1. It is impossible that API change in the analyzer (working tree) breaks a policy rule implementation. So, the rules API is  
    independent from working tree.
2. The API can be designed to reflect exactly the requirements from the rules. This is exactly what is needed to arrive at easy to read rules. In particular from my experience it's hard to foresee the exact API needs when implementing new policy rules use cases, and my gut feeling is that this encapsulation will make more and more sense the more functions are being added.
3. The functions can be implement based on the working tree, but it is not required. The encapsulation allows
    changing the implementation. For example a file existence check could be changed to not work on the cloned source but
    on the ScanResult, if that contained a full list of all files.
4. The helper functions are basically factored out logic from the rule matchers added to OrtResult. Exposing that logic is 
   important to expose because the logic inside rule matchers is not re-usable. For example the rule matcher `hasFile()` 
  contains logic to find the actual files. If you want to re-use the logic to find the files, the logic cannot be put into `hasFile()`   
  matcher but needs to be exposed somewhere. That somewhere is the `SourceTree` class I've added. I don't see why that somewhere should be `WorkingTree`.

Is any of the above points somewhat more convincing?

BTW.: I'm planning to add matchers for the commit history, and therefore create (not expose) a working tree instance inside the sourcetree class.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
mnonnenmacher: Thanks, I missed that.

evaluator/src/main/kotlin/SourceTree.kt
fviernau: Anyhow, as I plan to make a couple of further PRs on top, can we just keep it for now and make that decision when that work is done (guess we know more by then)? If it's really not needed then I'll refactor it away.

sschuberth: > can we just keep it for now and make that decision when that work is done

I'd actually prefer to get that sorted out now. IMO we had to much of "I need this urgently now so let's merge it"-style of changes recently, and ORT's code base is starting to suffer from suboptimal code design decisions (with only a single use-case in mind).

fviernau: Ok, then let's sort it out. What do you propose to do?

sschuberth: > So, the code base wouldn't suffer from it in the long run. Did you get that?

I did get that. But such promises were made in the past, and then that planned refactoring never happened, or only after a *very* long time.

> I prefer a rather iterative approach

Iterative is fine, but iterative should also mean that it's already going into the right direction, and not introducing stuff that gets removed later on (where foreseeable).

What I'd like to discuss first with you and at least also @mnonnenmacher is whether providing access to the source / working tree should really become a "first class citizen" of the evaluator as currently sketched. When I first read about https://github.com/oss-review-toolkit/ort/issues/5621, I was hoping that would be mostly implemented by helper functions implemented in EPAM's `rules.kts` itself, and not so much in ORT upstream.

fviernau:
I've created a new PR which is more minimal: https://github.com/oss-review-toolkit/ort/pull/5754.
Let's continue with that PR to agree on the concept and get it merged.

PR #5680, opened 2022-08-24
model: Introduce a `concludedCopyrights` field to package curations

### LINKED ISSUE #4519: Allow to curate copyrights directly (not via authors)

sschuberth:
In German law, the author and the copyright holder can be two separate
legal entities and therefore also need to be tracked separately.

Introduce a new field for declared copyrights that is now the primary
source for copyright holder information. Authors are still only used as
copyright holders if the `addAuthorsToCopyrights` option is enabled.

For now, all package manager implementations set empty declared
copyrights. Filling the declared copyrights is left as a future exercise
(also see [1]). Currently, the only way to add declared copyrights is via
curations.

This change resolves #4519.

[1]: https://github.com/oss-review-toolkit/ort/pull/5504#issuecomment-1191168485


fviernau:
@sschuberth why is this a new PR? How does it differ from #5504?

sschuberth:
> @sschuberth why is this a new PR? How does it differ from #5504?

As I'm taking over @porsche-rbieniek's PRs it's more convenient for me to work on my own branches than his. I'll document the differences once this moves out of draft state.

wkl3nk:
I like this PR. Do you consider to bring it into effect?

sschuberth:
> I like this PR. Do you consider to bring it into effect?

Probably not any time soon, as I've not found the time / priority to do this in more than 2 years...

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -302,7 +302,6 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
         val licenseInfoResolver = LicenseInfoResolver(
             provider = DefaultLicenseInfoProvider(ortResultInput, packageConfigurationProvider),
             copyrightGarbage = copyrightGarbage,
-            addAuthorsToCopyrights = config.addAuthorsToCopyrights,
```
sschuberth, line 305: It was decided during PR grooming by the TSC that an implementation of the general feature to conclude copyrights would at first *not* remove the feature to map authors to copyrights. This is to allow users for a transition period to migrate configuration from one feature to the other smoothly.

sschuberth (commented):
Needs to be reworked.

PR #5686, opened 2022-08-24
Improve license field docs

sschuberth:

### DIFF
--- a/model/src/main/kotlin/Package.kt
+++ b/model/src/main/kotlin/Package.kt
@@ -63,8 +63,8 @@ data class Package(
     val authors: SortedSet<String> = sortedSetOf(),
 
     /**
-     * The list of licenses the authors have declared for this package. This does not necessarily correspond to the
-     * licenses as detected by a scanner. Both need to be taken into account for any conclusions.
+     * The list of licenses declared for this package. This does not necessarily correspond to the licenses as detected
+     * by a scanner. Both need to be taken into account for any conclusions.
      */
     val declaredLicenses: SortedSet<String>,
 
--- a/model/src/main/kotlin/Package.kt
+++ b/model/src/main/kotlin/Package.kt
@@ -75,8 +75,8 @@ data class Package(
     val declaredLicensesProcessed: ProcessedDeclaredLicense = DeclaredLicenseProcessor.process(declaredLicenses),
 
     /**
-     * The concluded license as an [SpdxExpression]. It can be used to correct the license of a package in case the
-     * [declaredLicenses] found in the packages metadata or the licenses detected by a scanner do not match reality.
+     * The concluded license as an [SpdxExpression]. It can be used to override the [declared][declaredLicenses] /
+     * [detected][LicenseFinding.license] licenses of a package.
      *
      * ORT itself does not set this field, it needs to be set by the user using a [PackageCuration].
      */
--- a/model/src/main/kotlin/Project.kt
+++ b/model/src/main/kotlin/Project.kt
@@ -63,8 +63,8 @@ data class Project(
     val authors: SortedSet<String> = sortedSetOf(),
 
     /**
-     * The list of licenses the authors have declared for this project. This does not necessarily correspond to the
-     * licenses as detected by a scanner. Both need to be taken into account for any conclusions.
+     * The list of licenses declared for this project. This does not necessarily correspond to the licenses as detected
+     * by a scanner. Both need to be taken into account for any conclusions.
      */
     val declaredLicenses: SortedSet<String>,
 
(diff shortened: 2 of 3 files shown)


model/src/main/kotlin/Package.kt
```diff
@@ -75,8 +75,9 @@ data class Package(
     val declaredLicensesProcessed: ProcessedDeclaredLicense = DeclaredLicenseProcessor.process(declaredLicenses),
 
     /**
-     * The concluded license as an [SpdxExpression]. It can be used to correct the license of a package in case the
-     * [declaredLicenses] found in the packages metadata or the licenses detected by a scanner do not match reality.
+     * The concluded license as an [SpdxExpression]. It can be used to override the [declared licenses]
+     * [Package.declaredLicenses] and / or the [detected license expression][LicenseFinding.license] of a package in
```
mnonnenmacher, line 79: * `Package.` is not required for a property of the same class.
* I think just "and" instead of "and / or" is good enough, the "or" could be confusing. Also, why not "detected licenses" instead of "detected license expression" (as it can be multiple)? (same below)

mnonnenmacher (changes requested):

fviernau (approved):

model/src/main/kotlin/Package.kt
```diff
@@ -75,8 +75,9 @@ data class Package(
...
+     * case either are incorrect.
```
fviernau, line 80: Maybe "in case either are incorrect." can be dropped, because overriding a correct license with identity or something incorrect obviously does not make any sense.

model/src/main/kotlin/Package.kt
sschuberth: > Also, why not "detected licenses" instead of "detected license expression" (as it can be multiple)?

I wanted to emphasize the different data types: While `declaredLicenses` is a set of strings (where plural IMO fits better), `LicenseFinding.license` is a single SPDX expression.

model/src/main/kotlin/Package.kt
sschuberth: Agreed, esp. as I was a bit unhappy with the term "incorrect" (I simply had "wrong" locally before) as it left unclear in what regard the license is "incorrect" (completely wrong, or just missing additional licenses). So it's probably better to drop that bit completely.

model/src/main/kotlin/Package.kt
sschuberth: Still, in the context of addressing [@fviernau's comment](https://github.com/oss-review-toolkit/ort/pull/5686#discussion_r954311592), I went for a yet simpler sentence that uses plural for both license origins.

fviernau (approved):

model/src/main/kotlin/Package.kt
```diff
@@ -63,8 +63,8 @@ data class Package(
     val authors: SortedSet<String> = sortedSetOf(),
 
     /**
-     * The list of licenses the authors have declared for this package. This does not necessarily correspond to the
-     * licenses as detected by a scanner. Both need to be taken into account for any conclusions.
+     * The list of licenses declared for this package. This does not necessarily correspond to the licenses as detected
```
tsteenbe, line 66: I would change "declared for this package" to "declared in the metadata for this package" to make it clear what we mean with "declared" as in the industry declared can also mean the main license e.g. LICENSE, COPYING, etc. [SPDX Declared License 7.15.2](https://spdx.github.io/spdx-spec/v2.3/package-information/#715-declared-license-field): "This is simply the license identified in text in one or more files (for example COPYING file) in the source code package."

model/src/main/kotlin/Project.kt
```diff
@@ -63,8 +63,8 @@ data class Project(
     val authors: SortedSet<String> = sortedSetOf(),
 
     /**
-     * The list of licenses the authors have declared for this project. This does not necessarily correspond to the
-     * licenses as detected by a scanner. Both need to be taken into account for any conclusions.
+     * The list of licenses declared for this project. This does not necessarily correspond to the licenses as detected
```
tsteenbe, line 66: Same as above, recommend to replace "declared for this project" to "declared in the metadata for this project"

tsteenbe (changes requested):

model/src/main/kotlin/Package.kt
sschuberth: Is this really necessary as the `Package`'s class docs already say to "contain[s] all relevant *metadata* about a package"? I mean, we'd need to add that to each and every field then. Finally, this is not end-user documentation, but developer documentation.

sschuberth: @tsteenbe would you be fine with merging this as-is, because I have a few changes upcoming that build upon these docs?

tsteenbe (approved):

model/src/main/kotlin/Package.kt
tsteenbe: @sschuberth  Change this PR to approved.

sschuberth:

PR #5688, opened 2022-08-24
Pip: Improve python inspector integration

mnonnenmacher:
Some improvements for the python-inspector integration, see the commit messages for details. A follow-up PR will address some issues with regards to selecting the correct Python version.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pip.kt
+++ b/analyzer/src/main/kotlin/managers/Pip.kt
@@ -145,16 +145,24 @@ object PythonInspector : CommandLineTool {
         setupPyFile: String?,
         pythonVersion: String = "38",
     ): ProcessCapture {
-        var commandLineOptions = listOf(
-            "python-inspector",
-            "--python-version", pythonVersion,
-            "--json-pdt", outputFile,
-        )
+        val commandLineOptions = buildList {
+            add("python-inspector")
 
-        if (requirementsFile != null) {
-            commandLineOptions += listOf("--requirement", requirementsFile)
-        } else if (setupPyFile != null) {
-            commandLineOptions += listOf("--setup-py", setupPyFile)
+            add("--python-version")
+            add(pythonVersion)
+
+            add("--json-pdt")
+            add(outputFile)
+
+            if (requirementsFile != null) {
+                add("--requirement")
+                add(requirementsFile)
+            }
+
+            if (setupPyFile != null) {
+                add("--setup-py")
+                add(setupPyFile)
+            }
         }
 
         val process = ProcessCapture(workingDir, *commandLineOptions.toTypedArray())
--- a/analyzer/src/main/kotlin/managers/Pip.kt
+++ b/analyzer/src/main/kotlin/managers/Pip.kt
@@ -138,7 +138,7 @@ object PythonVersion : CommandLineTool, Logging {
 object PythonInspector : CommandLineTool {
     override fun command(workingDir: File?) = "python-inspector"
 
-    fun runPythonInspector(
+    fun run(
         workingDir: File,
         outputFile: String,
         requirementsFile: String?,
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -348,14 +348,14 @@ class Pip(
         val jsonFile = createOrtTempDir().resolve("python-inspector.json")
 
         val pythonInspector = if (definitionFile.name == "setup.py") {
-            PythonInspector.runPythonInspector(
+            PythonInspector.run(
```
fviernau: out of scope: Can `PythonInspector`  be dropped from `PythonInspector.run`?

analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -154,15 +153,13 @@ object PythonInspector : CommandLineTool {
             add("--json-pdt")
             add(outputFile)
 
-            if (requirementsFile != null) {
+            if (definitionFile.name == "setup.py") {
```
fviernau, line 158: Should "setup.py" be extracted to a constant to avoid redundancy?

analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -48,13 +48,15 @@ import org.ossreviewtoolkit.model.jsonMapper
...
 import org.ossreviewtoolkit.utils.common.ProcessCapture
+import org.ossreviewtoolkit.utils.common.collectMessages
 import org.ossreviewtoolkit.utils.common.normalizeLineBreaks
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.common.textValueOrEmpty
 import org.ossreviewtoolkit.utils.ort.DeclaredLicenseProcessor
 import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 import org.ossreviewtoolkit.utils.ort.createOrtTempFile
+import org.ossreviewtoolkit.utils.ort.showStackTrace
```
fviernau, line 59: commit message: 
- typo in "reuqireSuccess"
- maybe comma after "therefore"?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Pip.kt
mnonnenmacher: No, the call comes from another class.

analyzer/src/main/kotlin/managers/Pip.kt
mnonnenmacher: I think at least most other occurrences will be removed in follow-up changes, so I'd rather wait for that first.

fviernau (approved):

analyzer/src/main/kotlin/managers/Pip.kt
```diff
@@ -145,16 +145,24 @@ object PythonInspector : CommandLineTool {
...
-        if (requirementsFile != null) {
-            commandLineOptions += listOf("--requirement", requirementsFile)
-        } else if (setupPyFile != null) {
-            commandLineOptions += listOf("--setup-py", setupPyFile)
+            if (requirementsFile != null) {
+                add("--requirement")
+                add(requirementsFile)
+            }
+
+            if (setupPyFile != null) {
```
sschuberth: This slight changes semantics if both `requirementsFile` and `setupPyFile` were `!= null`. As that should never happen I'm fine with the change, but it should probably be noted in the commit message. Or an according `require()` check should be added.

BTW, how does `python-inspector` behave if both `--requirement` and `--setup-py` are passed?

sschuberth: Ok, I just saw this gets indirectly addressed in a later commit.

PR #5690, opened 2022-08-25
OrtMain: Add a print statement when the command terminates

oheger-bosch:
This is a rather trivial change, but can be useful when ORT is run in
CI environments to monitor commands and find out the commands that did
not terminate normally.

In our setup, a pipeline sometimes runs multiple ORT commands in a single step. In case of failures, it can be difficult from the existing log output to find out the command causing the failure. The new log line helps in this regard.

### DIFF
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -42,6 +42,8 @@ import java.io.File
 
 import kotlin.system.exitProcess
 
+import org.apache.logging.log4j.kotlin.Logging
+
 import org.ossreviewtoolkit.cli.commands.*
 import org.ossreviewtoolkit.cli.utils.logger
 import org.ossreviewtoolkit.model.config.LicenseFilenamePatterns
@@ -80,11 +82,19 @@ data class GlobalOptions(
  */
 fun main(args: Array<String>) {
     Os.fixupUserHomeProperty()
-    OrtMain().main(args)
+
+    try {
+        OrtMain().main(args)
+    } finally {
+        OrtMain.logger.debug { "ORT command execution terminated." }
+    }
+
     exitProcess(0)
 }
 
 class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
+    companion object : Logging
+
     private val configFile by option("--config", "-c", help = "The path to a configuration file.")
         .convert { it.expandTilde() }
         .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)


cli/src/main/kotlin/OrtMain.kt
```diff
@@ -80,7 +81,11 @@ data class GlobalOptions(
  */
 fun main(args: Array<String>) {
     Os.fixupUserHomeProperty()
-    OrtMain().main(args)
+    try {
+        OrtMain().main(args)
+    } finally {
+        println("ORT command execution terminated at ${LocalDateTime.now()}.")
```
fviernau, line 87: This is know printed only right before `exitProcess(0)`.
It is not printed if ORT exits in a different place. There must be some `exitProcess(1)` or `exitProcess(2)` calls elsewhere. So, could there be a better place to move the printing to?

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -58,6 +58,7 @@ import org.ossreviewtoolkit.utils.ort.ortDataDirectory
 import org.ossreviewtoolkit.utils.ort.printStackTrace
 
 import org.slf4j.LoggerFactory
+import java.time.LocalDateTime
```
sschuberth: Wrong import grouping.

fviernau, line 61: missing empty line above.

fviernau (changes requested):

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -80,7 +81,11 @@ data class GlobalOptions(
  */
 fun main(args: Array<String>) {
     Os.fixupUserHomeProperty()
-    OrtMain().main(args)
+    try {
```
sschuberth, line 84: Please add blank lines before and after the block to make it stand out.

cli/src/main/kotlin/OrtMain.kt
sschuberth, line 87: Can we make this a `debug` log instead? Like

    val ortMain = OrtMain()
    try {
        ortMain.main(args)
    } finally {
        ortMain.logger.debug { ... }
    }

sschuberth (changes requested):

cli/src/main/kotlin/OrtMain.kt
sschuberth: > There must be some `exitProcess(1)` or `exitProcess(2)` calls elsewhere.

No, there aren't. *clikt* never calls `exit(Process)` itself but throws `ProgramResult` exceptions to terminate and return to the caller (it's IMO bad that *clikt* uses exception for programm flow here, but that's a different story).

oheger-bosch: Done. Then it is not necessary to add the time to the message, since this should be part of the log output anyway.

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -80,11 +82,19 @@ data class GlobalOptions(
  */
 fun main(args: Array<String>) {
     Os.fixupUserHomeProperty()
-    OrtMain().main(args)
+
+    try {
+        OrtMain().main(args)
+    } finally {
+        OrtMain.logger.debug { "ORT command execution terminated." }
```
sschuberth, line 89: The commit message needs to be adjust now to not talk about `print`.

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -80,11 +82,19 @@ data class GlobalOptions(
...
+    }
+
     exitProcess(0)
 }
 
 class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
+    companion object : Logging
```
sschuberth, line 96: This is not needed thanks to some extension-property magic specific to `CliktCommand`.

sschuberth (changes requested):

cli/src/main/kotlin/OrtMain.kt
oheger-bosch: Do you have any proposals?

I currently don't have the time to invest deeper here. I would also be fine to redraw this PR if you think it is not useful in its current state

cli/src/main/kotlin/OrtMain.kt
oheger-bosch: Ah, but then I need an instance of `OrtMain` right? An expression like `OrtMain.logger` does not work.

cli/src/main/kotlin/OrtMain.kt
fviernau: > No, there aren't. _clikt_ never calls `exit(Process)` itself but throws `ProgramResult` exceptions to terminate and return to the caller

That's not exactly correct, see https://github.com/ajalt/clikt/blob/c9e7757e846f1f75fbe15488863c20b481116f1e/clikt/src/commonMain/kotlin/com/github/ajalt/clikt/core/CliktCommand.kt#L414-L437.

So, in case a `ProgramResult` is thrown, the added print statement won't be executed.

cli/src/main/kotlin/OrtMain.kt
sschuberth: See https://github.com/oss-review-toolkit/ort/pull/5690#discussion_r954597355.

cli/src/main/kotlin/OrtMain.kt
fviernau: From the above, maybe the whole code inside `run()` should be moved into a ´try` and then `finally` the statement should be printed inside `run()`?!

sschuberth: > That's not exactly correct

You're right.

> Do you have any proposals?

The only thing that comes to my mind is to call `CliktCommand.parse()` ourtselves directly but that's also not feasible

> I would also be fine to redraw this PR if you think it is not useful in its current state

I indeed fear the PR does not do anything meaningful (on application error) in its current state.

sschuberth: > From the above, maybe the whole code inside `run()` should be moved into a `try` and then `finally` the statement should be printed inside `run()`?!

IIUC you'd have to to this for all subcommands then, not just `OrtMain`, which IMO is not nice...

oheger-bosch: It helped me at least to debug cases when Kubernetes has killed a pod. But as said, I am fine to close this; it won't be possible to cover all scenarios.

oheger-bosch:
Closing based on review comments.

cli/src/main/kotlin/OrtMain.kt
sschuberth: > It helped me at least to debug cases when Kubernetes has killed a pod.

Ah, then I might have misunderstood: I thought the main purpose was to add the output when ORT exists with code != 0. If that's not the case, and this output had helped you in the past, I'm still fine with merging this.

fviernau: > maybe the whole code inside `run()` should be moved into a `try` and then `finally` the statement should be printed inside `run()`?!

Wouldn't that approach work? (e.g. only in the run function of the main command)

sschuberth: > Wouldn't that approach work? (e.g. only in the run function of the main command)

I guess not, because a subcommand's `run` is not called from the main command's `run`, so if an exception occurs in a subcommand's `run` it's not propagated back to the main command's `run`.

fviernau: I see, thanks for explaining.

fviernau: I just found https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#addShutdownHook-java.lang.Thread-, not sure if it's a good idea to use it for that purpose though.

PR #5691, opened 2022-08-25
Use pinned dependencies for ScanCode installation

mnonnenmacher:


.github/workflows/build-and-test.yml
```diff
@@ -83,7 +83,8 @@ jobs:
           chmod a+x $HOME/.local/bin/repo
 
           # Install ScanCode for license texts.
-          pip install --user scancode-toolkit==$SCANCODE_VERSION
+          curl -Os https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt
+          pip install --no-cache-dir --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION
```
sschuberth, line 87: Nit: Should we also do the `rm requirements.txt` here, for consistency? But we can also omit it as it does not count towards the Docker image size here.

sschuberth (approved):

.github/workflows/build-and-test.yml
mnonnenmacher: That's what I thought, currently I just hope that the tests pass now so that we can unblock our pipeline.

PR #5692, opened 2022-08-25
ScanCode: use pinned dependency constraints

pombredanne:
This PR fetches the requirements.txt from ScanCode repo
to use them as constraints in a pip installation command.
This guarantees that the dependency versions are tested
to work with the ScanCode version

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -157,7 +157,8 @@ RUN /opt/ort/bin/import_proxy_certs.sh && \
 
 # Add scanners (in versions known to work).
 ARG SCANCODE_VERSION
-RUN pip install --no-cache-dir scancode-toolkit==$SCANCODE_VERSION
+RUN wget https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt
+RUN pip install --no-cache-dir --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION
 
 ARG PYTHON_INSPECTOR_VERSION
 RUN pip install --no-cache-dir python-inspector==$PYTHON_INSPECTOR_VERSION


Dockerfile
```diff
@@ -157,7 +157,8 @@ RUN /opt/ort/bin/import_proxy_certs.sh && \
 
 # Add scanners (in versions known to work).
 ARG SCANCODE_VERSION
-RUN pip install --no-cache-dir scancode-toolkit==$SCANCODE_VERSION
+RUN wget https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt
+RUN pip install --no-cache-dir --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION
```
fviernau, line 161: commit-msg: Singed off twice.

fviernau (changes requested):

Dockerfile
```diff
@@ -157,7 +157,8 @@ RUN /opt/ort/bin/import_proxy_certs.sh && \
 
 # Add scanners (in versions known to work).
 ARG SCANCODE_VERSION
-RUN pip install --no-cache-dir scancode-toolkit==$SCANCODE_VERSION
+RUN wget https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt
```
sschuberth, line 160: Nit: We try to stick to `curl` in this file, so please probably use `curl -Os` instead.

sschuberth: And please combine this into the below `RUN` statement to create fewer layers. Like in total

```
RUN curl -Os https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt && \
    pip install --no-cache-dir scancode-toolkit==$SCANCODE_VERSION
```

sschuberth (changes requested):

sschuberth:
Unfortunately, this [does not seem to solve the issue](https://github.com/oss-review-toolkit/ort/runs/8012561837?check_suite_focus=true#step:5:1596).

mnonnenmacher:
> Unfortunately, this [does not seem to solve the issue](https://github.com/oss-review-toolkit/ort/runs/8012561837?check_suite_focus=true#step:5:1596).

@pombredanne  The same change needs to be done in `.github/workflows/build-and-test.yml` because the failing test does not run in Docker.

Dockerfile
mnonnenmacher: More commit message nits:
- "Fetch" instead of "Fetches" (imperative)
- Missing dot at the end.

mnonnenmacher:
@pombredanne Due to urgency I have updated #5691 with your changes and also tried to address the review comments from this PR.

sschuberth:
Superseded by https://github.com/oss-review-toolkit/ort/pull/5691.

PR #5695, opened 2022-08-25
Stop using Jackson in the CLI module

sschuberth:


scanner/src/main/kotlin/storages/Sw360Storage.kt
```diff
@@ -57,10 +57,14 @@ class Sw360Storage(
     configuration: Sw360StorageConfiguration
 ) : ScanResultsStorage() {
     companion object : Logging {
-        fun createConnection(config: Sw360StorageConfiguration, jsonMapper: ObjectMapper): SW360Connection {
+        val JSON_MAPPER: ObjectMapper = jsonMapper.copy()
```
fviernau, line 60: `JSON_MAPPER` could be private.

fviernau (approved):

scanner/src/main/kotlin/storages/Sw360Storage.kt
sschuberth: Right, but I chose to keep it public for now to eventually reuse it in either of the SW360 classes. But actually, I'd like to see https://github.com/eclipse/sw360/issues/1622 addressed to be able to also use *kotlinx-serialization* and `Json.Default` here.

PR #5699, opened 2022-08-26
CreateCommand: Allow for generating path excludes

fviernau:
Add a simple heuristic which generates path excludes (which likely
apply). The heuristic for now only maps directory names to path excludes
and will be extended in upcoming changes.

### DIFF
--- a/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
+++ b/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
@@ -1,0 +1,85 @@
+/*
+ * Copyright (C) 2022 EPAM Systems, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.utils
+
+import java.io.File
+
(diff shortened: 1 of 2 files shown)


helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+        while (dir != null) {
+            result += dir
+            dir = dir.parentFile
+        }
+    }
+
+    return result
+}
+
+private val PATH_EXCLUDES_REASON_FOR_DIR_NAME = mapOf(
```
tsteenbe, line 75: Could we make below mapping a config file in YAML?

tsteenbe (changes requested):

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
fviernau: Yes. Would you be fine doing this in a separate PR later on?

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+
+import java.io.File
+
+import org.ossreviewtoolkit.model.config.PathExclude
+import org.ossreviewtoolkit.model.config.PathExcludeReason.BUILD_TOOL_OF
+import org.ossreviewtoolkit.model.config.PathExcludeReason.DOCUMENTATION_OF
+import org.ossreviewtoolkit.model.config.PathExcludeReason.TEST_OF
+
+/**
+ * This class generates path excludes based on the set file paths present in the source tree.
```
mnonnenmacher, line 30: Nit: "set **of**"

mnonnenmacher (approved):

mnonnenmacher (changes requested):
Forgot I had a change request when clicking approve...

mnonnenmacher (approved):

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
fviernau: @tsteenbe let's sort this out later on, I'll experiment with the more complex heuristic first and try out having a fun test. Then we can still extract the patterns to a config file.

fviernau:
Merge despite the failing unrelated tests.

PR #5700, opened 2022-08-26
ScanCodeResultParser: Drop support for ScanCode < 3

sschuberth:


scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -162,6 +162,11 @@ private fun generateScannerOptions(options: JsonNode?): String {
     }.orEmpty()
 }
 
+private fun getInputPath(result: JsonNode): String {
```
fviernau, line 158: commit-msg: Shoud "being" be dropped?

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -257,9 +261,11 @@ internal fun replaceLicenseKeys(licenseExpression: String, replacements: Collect
 private fun getCopyrightFindings(result: JsonNode): List<CopyrightFinding> {
     val copyrightFindings = mutableListOf<CopyrightFinding>()
 
+    val input = getInputPath(result)
```
fviernau, line 252: how about naming this `inputPath` instead of just `input`?

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -162,6 +162,11 @@ private fun generateScannerOptions(options: JsonNode?): String {
     }.orEmpty()
 }
 
+private fun getInputPath(result: JsonNode): String {
+    val header = result["headers"]?.singleOrNull()
+    return header?.get("options")?.get("input")?.singleOrNull()?.textValue()?.let { "$it/" }.orEmpty()
```
fviernau: Can `it` be an empty string?

scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
```diff
@@ -72,7 +73,7 @@ class ClearlyDefinedStorageFunTest : StringSpec({
                 summary = ScanSummary(
                     startTime = Instant.parse("2020-02-14T00:36:14.000335513Z"),
                     endTime = Instant.parse("2020-02-14T00:36:37.000492119Z"),
-                    packageVerificationCode = "",
+                    packageVerificationCode = SpdxConstants.NONE,
```
fviernau, line 76: commit-msg: Please provide a rationale

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -72,48 +72,50 @@ private val TIMEOUT_ERROR_REGEX = Pattern.compile(
 )
 
 /**
- * Generate a summary from the given raw ScanCode [result], using [startTime] and [endTime] metadata. From the
- * [scanPath] the package verification code is generated. If [parseExpressions] is true, license findings are preferably
- * parsed as license expressions.
+ * Generate a summary from the given raw ScanCode [result]. From the [scanPath] the package verification code is
+ * generated. If [parseExpressions] is true, license findings are preferably parsed as license expressions.
```
fviernau, line 76: Please provide a rationale why it makes sense to align with the clearly defined implementation.

I believe using the system time is less error prone and simpler.

fviernau (changes requested):

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: The sentence sounds less proper to me when omitted it.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: I'd like to stick to the same name as in `getLicenseFindings` (where `input` was already used before; the variable is named after the field name in the raw scan result).

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: Not that I'm aware of. Also please note that this code was just moved to here.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: > I believe using the system time is less error prone and simpler.

But it's additional code instead of making reuse of existing code. I've amended the commit message, which now also refers to 2d0038d39a39606f6163a60f4e82ac9116c08e95 for some background information.

fviernau (approved):

PR #5711, opened 2022-08-29
FileMatcher: Reuse instances of `AntPathMatcher`

mnonnenmacher:
Create only instance of `AntPathMatcher` for case-sensitive and
case-insensitive matching each. For convenience, also create static
matcher functions for cases where the patterns are not reused and
creating an instance of `FileMatcher` would be overhead.

### DIFF
--- a/utils/common/src/main/kotlin/FileMatcher.kt
+++ b/utils/common/src/main/kotlin/FileMatcher.kt
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Copyright (C) 2022 Bosch.IO GmbH
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -39,13 +40,36 @@ class FileMatcher(
 ) {
     constructor(vararg patterns: String, ignoreCase: Boolean = false) : this(patterns.asList(), ignoreCase)
 
-    private val matcher = AntPathMatcher().apply {
-        setCaseSensitive(!ignoreCase)
+    companion object {
+        private val matchCaseInsensitive = AntPathMatcher().apply { setCaseSensitive(false) }::match
+        private val matchCaseSensitive = AntPathMatcher().apply { setCaseSensitive(true) }::match
+
+        /**
+         * Return true if [path] is matched by [pattern], false otherwise. The [path] must use '/' as separators, if it
+         * contains any.
+         */
+        fun match(pattern: String, path: String, ignoreCase: Boolean = false) =
+            when (ignoreCase) {
+                true -> matchCaseInsensitive(pattern, path)
+                false -> matchCaseSensitive(pattern, path)
+            }
+
+        /**
+         * Return true if [path] is matched by any of [patterns], false otherwise. The [path] must use '/' as
+         * separators, if it contains any.
+         */
+        fun match(patterns: Collection<String>, path: String, ignoreCase: Boolean = false) =
+            when (ignoreCase) {
+                true -> patterns.any { pattern -> matchCaseInsensitive(pattern, path) }
+                false -> patterns.any { pattern -> matchCaseSensitive(pattern, path) }
+            }
     }
 
+    private val match = if (ignoreCase) matchCaseInsensitive else matchCaseSensitive
+
     /**
-     * Return true if and only if the given [path] is matched by any of the file globs passed to the
-     * constructor. The [path] must use '/' as separators, if it contains any.
+     * Return true if [path] is matched by any of [patterns], false otherwise. The [path] must use '/' as separators,
+     * if it contains any.
      */
-    fun matches(path: String): Boolean = patterns.any { pattern -> matcher.match(pattern, path) }
+    fun matches(path: String): Boolean = patterns.any { pattern -> match(pattern, path) }
 }


utils/common/src/main/kotlin/FileMatcher.kt
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2017-2019 HERE Europe B.V.
+ * Copyright (C) 2022 Bosch.IO GmbH
```
sschuberth, line 3: Commit message: "Create only *one* instance"

utils/common/src/main/kotlin/FileMatcher.kt
```diff
@@ -39,13 +40,42 @@ class FileMatcher(
 ) {
     constructor(vararg patterns: String, ignoreCase: Boolean = false) : this(patterns.asList(), ignoreCase)
 
-    private val matcher = AntPathMatcher().apply {
-        setCaseSensitive(!ignoreCase)
+    companion object {
```
sschuberth, line 43: For the fun of it, here's an alternative:

```
package org.ossreviewtoolkit.utils.common

import org.springframework.util.AntPathMatcher

private val matcherMap = mapOf(
    false to AntPathMatcher().apply { setCaseSensitive(true) }::match,
    true to AntPathMatcher().apply { setCaseSensitive(false) }::match
)

/**
 * A class to determine whether a path is matched by any of the given globs.
 */
class FileMatcher(
    /**
     * The collection of [glob patterns][1] to consider for matching.
     *
     * [1]: https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob
     */
    val patterns: Collection<String>,

    /**
     * Toggle the case-sensitivity of the matching.
     */
    ignoreCase: Boolean = false
) {
    constructor(vararg patterns: String, ignoreCase: Boolean = false) : this(patterns.asList(), ignoreCase)

    private val matcher = matcherMap.getValue(ignoreCase)

    /**
     * Return true if [path] is matched by any of [patterns], false otherwise. The [path] must use '/' as separators,
     * if it contains any.
     */
    fun matches(path: String): Boolean = patterns.any { pattern -> matcher(pattern, path) }
}
```

utils/common/src/main/kotlin/FileMatcher.kt
```diff
@@ -39,13 +40,42 @@ class FileMatcher(
...
+        private val caseInsensitiveMatcher = AntPathMatcher().apply { setCaseSensitive(false) }
+        private val caseSensitiveMatcher = AntPathMatcher().apply { setCaseSensitive(true) }
+
+        /**
+         * Return true if [path] is matched case-insensitively by [pattern], false otherwise. The [path] must use '/' as
+         * separators, if it contains any.
+         */
+        fun matchCaseInsensitive(pattern: String, path: String) = caseInsensitiveMatcher.match(pattern, path)
```
fviernau, line 51: Just an idea for a bit shorter function names and for potentially avoiding case distinctions in the caller:`match(pattern: String, path: String, ignoreCase: Boolean = false)`?

utils/common/src/main/kotlin/FileMatcher.kt
```diff
@@ -39,13 +40,42 @@ class FileMatcher(
 ) {
     constructor(vararg patterns: String, ignoreCase: Boolean = false) : this(patterns.asList(), ignoreCase)
 
-    private val matcher = AntPathMatcher().apply {
-        setCaseSensitive(!ignoreCase)
+    companion object {
+        private val caseInsensitiveMatcher = AntPathMatcher().apply { setCaseSensitive(false) }
+        private val caseSensitiveMatcher = AntPathMatcher().apply { setCaseSensitive(true) }
```
fviernau, line 45: The above vars should be all caps, shouldn't they?

fviernau (changes requested):

utils/common/src/main/kotlin/FileMatcher.kt
mnonnenmacher: Yes, but I have changed them to function references now, I think in this case I prefer lowercase.

utils/common/src/main/kotlin/FileMatcher.kt
mnonnenmacher: I wanted to avoid an if, but that was not necessary.

utils/common/src/main/kotlin/FileMatcher.kt
mnonnenmacher: I made use of the function references but I prefer two properties for the matchers over a map.

fviernau (approved):

utils/common/src/main/kotlin/FileMatcher.kt
```diff
@@ -39,13 +40,36 @@ class FileMatcher(
...
-        setCaseSensitive(!ignoreCase)
+    companion object {
+        private val matchCaseInsensitive = AntPathMatcher().apply { setCaseSensitive(false) }::match
+        private val matchCaseSensitive = AntPathMatcher().apply { setCaseSensitive(true) }::match
+
+        /**
+         * Return true if [path] is matched by [pattern], false otherwise. The [path] must use '/' as separators, if it
+         * contains any.
+         */
+        fun match(pattern: String, path: String, ignoreCase: Boolean) =
```
sschuberth, line 51: How about defaulting `ignoreCase` to `false` like in Kotlin's `equals`? Similar below.

fviernau (approved):

sschuberth (approved):

PR #5713, opened 2022-08-30
OrtResultExtensions: Remove unused functions

sschuberth:
The last use of `collectConcludedLicenses` and `collectDeclaredLicenses`
has been removed in 6146b42.

### DIFF
--- a/model/src/main/kotlin/utils/OrtResultExtensions.kt
+++ b/model/src/main/kotlin/utils/OrtResultExtensions.kt
@@ -20,36 +20,11 @@
 
 package org.ossreviewtoolkit.model.utils
 
-import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.config.CopyrightGarbage
 import org.ossreviewtoolkit.model.config.LicenseFilenamePatterns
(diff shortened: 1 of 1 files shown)


fviernau:
Fine from my side. But I'm not sure whether these are used by any notice reporter templates in some ORT internal setup, maybe at HERE (@bennati) or at Bosch (@MarcelBochtler). Maybe you can leave a note or approval as well?

fviernau (approved):

mnonnenmacher (approved):

sschuberth:
> These function were used by the `AbstractNoticeReporter` which was removed in [6146b42](https://github.com/oss-review-toolkit/ort/commit/6146b42aecb0e6c9b7f6602b480ef7a4f0d13295).

Right, see the commit message.

> I think all uses of these functions should be gone since the migration to the `NoticeTemplateReporter` which uses the `LicenseInfoResolver` for license information.

IIUC @fviernau's concern was that meanwhile some users might have started to use these functions also in their rule scripts. Is that something we should be worried about, or can we indeed merge?

mnonnenmacher:
> IIUC @fviernau's concern was that meanwhile some users might have started to use these functions also in their rule scripts. Is that something we should be worried about, or can we indeed merge?

I think we should merge because it is easy to revert in the worst case and if anyone is using those functions they should move to the license resolver API instead.

PR #5714, opened 2022-08-30
Scanner: Remove a superfluous package filter

sschuberth:
`OrtResult.getPackages()` only returns packages from an analyzer result,
and as of https://github.com/oss-review-toolkit/ort/commit/01c598a04c5745dc7408f6f84b619bc62d13592e package ids cannot clash with project ids, so there is
no need to filter out projects here.


mnonnenmacher (changes requested):
I think this code was required until 01c598a fixed the issue that there could be projects and packages with the same identifier, please reference that in the commit message.

mnonnenmacher (approved):

Issue #5637, opened 2022-08-05
CuratedPackage: `toUncuratedPackage()` not working as expected


fviernau:
The data model of the curated package does not store the uncurated package explicitly. It stores a list of package diffs, computed by `Package.diff()`. The idea was that the original package can always be derived from the curated package and these diffs.
However, that idea does not work for nullable values in the diff: If the diff has let's say `concludedLicense=null` it could mean both

1. Applying the curation hasn't had any effect on the value of `concludedLicense`
2. The value of `concludedLicense` was null before applying the curation.

So, the data model (`PackageCurationResult`) does only provide ambigous information and thus the data model needs to change in order to fix that problem.

PR #5719, opened 2022-08-30
evaluator: Make `PackageRules` take a `CuratedPackage`

sschuberth:

### DIFF
--- a/evaluator/src/main/kotlin/DependencyRule.kt
+++ b/evaluator/src/main/kotlin/DependencyRule.kt
@@ -19,10 +19,9 @@
 
 package org.ossreviewtoolkit.evaluator
 
+import org.ossreviewtoolkit.model.CuratedPackage
 import org.ossreviewtoolkit.model.DependencyNode
 import org.ossreviewtoolkit.model.Identifier
-import org.ossreviewtoolkit.model.Package
-import org.ossreviewtoolkit.model.PackageCurationResult
 import org.ossreviewtoolkit.model.PackageLinkage
 import org.ossreviewtoolkit.model.Project
 import org.ossreviewtoolkit.model.licenses.ResolvedLicenseInfo
@@ -34,8 +33,7 @@ import org.ossreviewtoolkit.utils.common.enumSetOf
 class DependencyRule(
     ruleSet: RuleSet,
     name: String,
-    pkg: Package,
-    curations: List<PackageCurationResult>,
+    pkg: CuratedPackage,
     resolvedLicenseInfo: ResolvedLicenseInfo,
 
     /**
@@ -64,13 +62,13 @@ class DependencyRule(
      * The [Project] that contains the [dependency].
      */
     val project: Project
-) : PackageRule(ruleSet, name, pkg, curations, resolvedLicenseInfo) {
+) : PackageRule(ruleSet, name, pkg, resolvedLicenseInfo) {
     override val description =
         "Evaluating rule '$name' for dependency '${dependency.id.toCoordinates()}' " +
                 "(project=${project.id.toCoordinates()}, scope=$scopeName, level=$level)."
 
     override fun issueSource() =
-        "$name - ${pkg.id.toCoordinates()} (dependency of ${project.id.toCoordinates()} in scope $scopeName)"
+        "$name - ${pkg.pkg.id.toCoordinates()} (dependency of ${project.id.toCoordinates()} in scope $scopeName)"
 
     /**
      * Return the direct dependencies of the [dependency].
--- a/evaluator/src/main/kotlin/DependencyRule.kt
+++ b/evaluator/src/main/kotlin/DependencyRule.kt
@@ -68,7 +68,7 @@ class DependencyRule(
                 "(project=${project.id.toCoordinates()}, scope=$scopeName, level=$level)."
 
     override fun issueSource() =
-        "$name - ${pkg.pkg.id.toCoordinates()} (dependency of ${project.id.toCoordinates()} in scope $scopeName)"
+        "$name - ${pkg.metadata.id.toCoordinates()} (dependency of ${project.id.toCoordinates()} in scope $scopeName)"
 
     /**
      * Return the direct dependencies of the [dependency].
--- a/examples/evaluator-rules/src/main/resources/example.rules.kts
+++ b/examples/evaluator-rules/src/main/resources/example.rules.kts
@@ -108,7 +108,7 @@ fun RuleSet.unhandledLicenseRule() = packageRule("UNHANDLED_LICENSE") {
         error(
             "The license $license is currently not covered by policy rules. " +
                     "The license was ${licenseSource.name.lowercase()} in package " +
-                    "${pkg.pkg.id.toCoordinates()}",
+                    "${pkg.metadata.id.toCoordinates()}",
             howToFixDefault()
         )
     }
@@ -122,7 +122,7 @@ fun RuleSet.unmappedDeclaredLicenseRule() = packageRule("UNMAPPED_DECLARED_LICEN
     resolvedLicenseInfo.licenseInfo.declaredLicenseInfo.processed.unmapped.forEach { unmappedLicense ->
         warning(
             "The declared license '$unmappedLicense' could not be mapped to a valid license or parsed as an SPDX " +
-                    "expression. The license was found in package ${pkg.pkg.id.toCoordinates()}.",
+                    "expression. The license was found in package ${pkg.metadata.id.toCoordinates()}.",
             howToFixDefault()
         )
     }
@@ -141,9 +141,9 @@ fun RuleSet.copyleftInSourceRule() = packageRule("COPYLEFT_IN_SOURCE") {
 
         val message = if (licenseSource == LicenseSource.DETECTED) {
             "The ScanCode copyleft categorized license $license was ${licenseSource.name.lowercase()} " +
-                    "in package ${pkg.pkg.id.toCoordinates()}."
+                    "in package ${pkg.metadata.id.toCoordinates()}."
         } else {
-            "The package ${pkg.pkg.id.toCoordinates()} has the ${licenseSource.name.lowercase()} ScanCode copyleft " +
+            "The package ${pkg.metadata.id.toCoordinates()} has the ${licenseSource.name.lowercase()} ScanCode copyleft " +
                     "catalogized license $license."
         }
 
@@ -164,15 +164,15 @@ fun RuleSet.copyleftInSourceLimitedRule() = packageRule("COPYLEFT_LIMITED_IN_SOU
 
         val licenseSourceName = licenseSource.name.lowercase()
         val message = if (licenseSource == LicenseSource.DETECTED) {
-            if (pkg.pkg.id.type == "Unmanaged") {
+            if (pkg.metadata.id.type == "Unmanaged") {
                 "The ScanCode copyleft-limited categorized license $license was $licenseSourceName in package " +
-                        "${pkg.pkg.id.toCoordinates()}."
+                        "${pkg.metadata.id.toCoordinates()}."
             } else {
                 "The ScanCode copyleft-limited categorized license $license was $licenseSourceName in package " +
-                        "${pkg.pkg.id.toCoordinates()}."
+                        "${pkg.metadata.id.toCoordinates()}."
             }
         } else {
-            "The package ${pkg.pkg.id.toCoordinates()} has the $licenseSourceName ScanCode copyleft-limited " +
+            "The package ${pkg.metadata.id.toCoordinates()} has the $licenseSourceName ScanCode copyleft-limited " +
                     "categorized license $license."
         }
 
@@ -208,7 +208,7 @@ fun RuleSet.vulnerabilityInPackageRule() = packageRule("VULNERABILITY_IN_PACKAGE
 
     issue(
         Severity.WARNING,
-        "The package ${pkg.pkg.id.toCoordinates()} has a vulnerability",
+        "The package ${pkg.metadata.id.toCoordinates()} has a vulnerability",
         howToFixDefault()
     )
 }
@@ -226,7 +226,7 @@ fun RuleSet.highSeverityVulnerabilityInPackageRule() = packageRule("HIGH_SEVERIT
 
     issue(
         Severity.ERROR,
-        "The package ${pkg.pkg.id.toCoordinates()} has a vulnerability with $scoringSystem severity > " +
+        "The package ${pkg.metadata.id.toCoordinates()} has a vulnerability with $scoringSystem severity > " +
                 "$maxAcceptedSeverity",
         howToFixDefault()
     )
--- a/examples/evaluator-rules/src/main/resources/example.rules.kts
+++ b/examples/evaluator-rules/src/main/resources/example.rules.kts
@@ -108,7 +108,7 @@ fun RuleSet.unhandledLicenseRule() = packageRule("UNHANDLED_LICENSE") {
         error(
             "The license $license is currently not covered by policy rules. " +
                     "The license was ${licenseSource.name.lowercase()} in package " +
-                    "${pkg.id.toCoordinates()}",
+                    "${pkg.pkg.id.toCoordinates()}",
             howToFixDefault()
         )
     }
@@ -122,7 +122,7 @@ fun RuleSet.unmappedDeclaredLicenseRule() = packageRule("UNMAPPED_DECLARED_LICEN
     resolvedLicenseInfo.licenseInfo.declaredLicenseInfo.processed.unmapped.forEach { unmappedLicense ->
         warning(
             "The declared license '$unmappedLicense' could not be mapped to a valid license or parsed as an SPDX " +
-                    "expression. The license was found in package ${pkg.id.toCoordinates()}.",
+                    "expression. The license was found in package ${pkg.pkg.id.toCoordinates()}.",
             howToFixDefault()
         )
     }
@@ -141,9 +141,9 @@ fun RuleSet.copyleftInSourceRule() = packageRule("COPYLEFT_IN_SOURCE") {
 
         val message = if (licenseSource == LicenseSource.DETECTED) {
             "The ScanCode copyleft categorized license $license was ${licenseSource.name.lowercase()} " +
-                    "in package ${pkg.id.toCoordinates()}."
+                    "in package ${pkg.pkg.id.toCoordinates()}."
         } else {
-            "The package ${pkg.id.toCoordinates()} has the ${licenseSource.name.lowercase()} ScanCode copyleft " +
+            "The package ${pkg.pkg.id.toCoordinates()} has the ${licenseSource.name.lowercase()} ScanCode copyleft " +
                     "catalogized license $license."
         }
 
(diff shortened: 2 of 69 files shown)


evaluator/src/main/kotlin/DependencyRule.kt
```diff
@@ -34,8 +33,7 @@ import org.ossreviewtoolkit.utils.common.enumSetOf
 class DependencyRule(
     ruleSet: RuleSet,
     name: String,
-    pkg: Package,
-    curations: List<PackageCurationResult>,
+    pkg: CuratedPackage,
```
fviernau, line 36: - I believe that consuming curations in the rules is an edge case. So, I believe there is hardly a rule interrested in the
  curations
- This change makes rules depend on curated package, so it may become harder to change. There is already one potential future change known [1]

I'm not saying this change shouldn't be done. The argumentation that simplifying passing the parameter to the constructor satisfies the change seems not convincing to me, because there is only few constructor calls, but many places consuming the package (which get a bit more verbose).

Anyhow, would it make sense to first implement [1]?

[1] https://github.com/oss-review-toolkit/ort/issues/5637

fviernau (changes requested):

evaluator/src/main/kotlin/DependencyRule.kt
sschuberth: > I believe that consuming curations in the rules is an edge case. So, I believe there is hardly a rule interrested in the curations

Probably. But passing the whole `CuratedPackage` will become necessary anyway due to the upcoming changes in order to access the concluded license in rules.

sschuberth: > Anyhow, would it make sense to first implement #5637?

Doesn't this PR also implicitly resolve that issue (at least for the mentioned `concludedLicense` field) as it moves out that field from the `Package` class?

fviernau: > Doesn't this PR also implicitly resolve that issue (at least for the mentioned `concludedLicense` field) as it moves out that field from the `Package` class?

Yes, it implicitly solves that just for concluded license - but my point was about introducing dependencies from rules to a data model which we already know would change. For that, it doesn't matter whether the problem is partially solved for the concluded license IMO.

examples/evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -141,9 +141,9 @@ fun RuleSet.copyleftInSourceRule() = packageRule("COPYLEFT_IN_SOURCE") {
 
         val message = if (licenseSource == LicenseSource.DETECTED) {
             "The ScanCode copyleft categorized license $license was ${licenseSource.name.lowercase()} " +
-                    "in package ${pkg.id.toCoordinates()}."
+                    "in package ${pkg.pkg.id.toCoordinates()}."
         } else {
-            "The package ${pkg.id.toCoordinates()} has the ${licenseSource.name.lowercase()} ScanCode copyleft " +
+            "The package ${pkg.pkg.id.toCoordinates()} has the ${licenseSource.name.lowercase()} ScanCode copyleft " +
```
fviernau: If the rule class would still have separate properties (or just getters) for `pkg` and `curations` the rules.kts code IMO would be cleaner. Would you agree?

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -33,6 +33,7 @@ import org.ossreviewtoolkit.analyzer.managers.utils.PackageManagerDependencyHand
 import org.ossreviewtoolkit.analyzer.managers.utils.SpdxDocumentCache
 import org.ossreviewtoolkit.analyzer.managers.utils.SpdxResolvedDocument
 import org.ossreviewtoolkit.downloader.VersionControlSystem
+import org.ossreviewtoolkit.model.CuratedPackage
```
fviernau, line 36: I have the following thoughts about this change which is why I tend towards not agreeing

pros:
1. Step towards cleanly separating legal conclusion from package meta-data.
  - If this is just a step, what is the end goal? What other steps are planned.
  - Is it really more separation, as there has already been a dedicated property which IMO achives separation already.
  - `CuratedPackage` is not just about legal conclusions, but about curations in general. It may curate any field from     `Package`, not just legal conclusions. So, why should this class be used to separate legal conclusions?

cons:
1. Most of the code does not care about which curations have been applied, but just about curated package meta-data.
    All that code now becomes more complicated / verbose, including the test.
3. Code has more dependencies than needed and so becomes less flexible and harder to test.
4. Writing code which is agnostic of whether the input data is curated or not becomes harder / impossible.
5. Other properties from the package class may also get `curated` so `CuratedPackage` is not just about legal   conclusion
6. Law of Demeter gets violated more often
7. I don't see any problem with having all attributes for a `Package` in the `Package` class. It seems reasonable as these are closely related and usually needed altogether.

What do you think about that? 
Also, further opinions would be interresting, e.g. @oss-review-toolkit/core-devs ?

fviernau (dismissed):

examples/evaluator-rules/src/main/resources/example.rules.kts
sschuberth: No, I would actually not agree. `curations` is not an entity that's usually dealt with on its own in ORT's code base, they always come with an accompanying `Package` in the context of a `CuratedPackage`. As such I see no value in splitting something up here that belongs together, and that is always dealt with together in the rest of the code base.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: > If this is just a step, what is the end goal? What other steps are planned?

Ultimately, I'd like what I start to call "metadata corrections" (i.e. changes required to package metadata in order for the downloader to be able to download the source code) to also be configured in a different file than "legal conclusions" (i.e. concluding licenses or copyrights). That's because in most real-world workflows I've seen so far, these very different kinds of "curations" are performed by different people, and ORT should do a better job at providing role-specific configuration.

> Is it really more separation, as there has already been a dedicated property which IMO achives separation already.

With the above in mind, that's not enough separation. As we (de-)serialize on a per-class basis, things that should ultimately go to different files also need to go to different classes.

> `CuratedPackage` is not just about legal conclusions, but about curations in general.

That's exactly the problem I'd like to solve. IMO it has been a severe data model design flaw to allow curating "everything" (even semantically totally unrelated stuff) via a single curations mechanism.

fviernau: > With the above in mind, that's not enough separation. As we (de-)serialize on a per-class basis, things that should ultimately go to different files also need to go to different classes.

Yes, that's right. However, that argumentation does not apply as the data models changed in this PR are actually not part of ORT's configuration IIUC. So, these classes do not affect the way curations are written. As these classes are not part of the configuration, I believe the design discussion can remain independent of ORT's configuration interface.

sschuberth: > the data models changed in this PR are actually not part of ORT's configuration

They are not part of ORT's *configuration*, right, but `CuratedPackage` is serialized as part of ORT's results. And IMO the mentioned (planned) separation makes sense to be also reflected there. For the configuration, *way* more refactoring would be needed, and I didn't want to dive into that rabbit hole just now before find a general agreement on the way forward.

fviernau: I still have problems following the argumentation, let me try to explain in other words what I still believe I don't get yet:

> Ultimately, I'd like what I start to call "metadata corrections" (i.e. changes required to package metadata in order for the downloader to be able to download the source code) to also be configured in a different file than "legal conclusions" (i.e. concluded licenses or copyrights). That's because in most real-world workflows I've seen so far, these very different kinds of "curations" are performed by different people, and ORT should do a better job at providing role-specific configuration.

This section is about ORT's configuration interface, but this PR is not about ORT's configuration interface.

> With the above in mind, that's not enough separation. As we (de-)serialize on a per-class basis, things that should ultimately go to different files also need to go to different classes.

This again is about ORT's configuration interface, but this PR is not about ORT's configuration interface.

 > That's exactly the problem I'd like to solve. IMO it has been a severe data model design flaw to allow curating "everything" (even semantically totally unrelated stuff) via a single curations mechanism.

This PR does not change any curation mechanism. So, there is still a single one.

I'd like to understand what exactly you mean by this "severe design" flow.
Is it in ORT's configuration interface, or in the OrtResult? 

> They are not part of ORT's configuration, right, but CuratedPackage is serialized as part of ORT's results. And IMO the mentioned (planned) separation makes sense to be also reflected there. For the configuration, way more refactoring would be needed, and I didn't want to dive into that rabbit hole just now before find a general agreement on the way forward.

What I understand is that the proposal is to separate legal conclusions from other meta-data curations in ORT's configuration interface (the user facing API), so that the different "roles" involved in the process only need looking into the config which is relevant to them.

What I still don't get is why the problem isn't solved by introducing or finding a way to separate things 
at ORT's configuration API and limiting the solution to the configuration API. So, is there need for touching the rules API as well as the OrtResult? Note: I believe historically OrtResult was intended to be human readable, but nowadays it's moving more towards a more redundancy free file format, like with the introduction of the dependency graph. Also the rules API to me seems to target more technical persons, but not all roles involved in the process. In short, why does OrtResult and rules API need to be refactored? To make it readable by legal persons?

Apart from that, have you considered using ORT's configuration API (almost) as-is to solve the problem?
I'm talking about an approach like the following, which is using a single curation mechanism but allows users to separate things if they want to.

```
curations/provenances.yml
curations/concluded-licenses.yml
curations/declared-license-mappings.yml
curations/other.yml
```

sschuberth: > This section is about ORT's configuration interface, but this PR is not about ORT's configuration interface.

Correct. I drifted off-topic a bit due to your questions, trying to explain the motivation for these changes by some upcoming (IMO) related changes that *will* affect the configuration.

> I'd like to understand what exactly you mean by this "severe design" flow. Is it in ORT's configuration interface, or in the OrtResult or in the Rules API?

In all of these we have design flaws, IMO. Going into details here would probably again be off-topic, but briefly:

- In the ORT data model classes we have several problems IMO, esp. around the general topic of curated data. The `Package` cass for example should only capture package metadata as declared by the package managers. It's a superset of the metadata that a package manager providers about a package. Consequently, the `concludedLicense` should not be part of that class, as package managers do not provide a concluded license (with the exception of the SPDX "fake" package manager). Also, alone from the pure naming of the class, it's clear that the `concludedLicense` property should go to the `CuratedPackage` class. This is also backed by this [comment](https://github.com/oss-review-toolkit/ort/blob/a8ef0a09150ae5b0f304661cd82e2867a649f17e/model/src/main/kotlin/Package.kt#L81) which basically admits that the `concludedLicense` property is wrong at the current place.
- In the configuration, we have the concepts of (package) curations, package configurations, license finding curations etc. The main problem with these is that they were modeled on a very technical model around ORT's (partly flawed) data model in this area. Instead, we should have modeled the configuration after real-world use-cases and user's roles. For example, in the workflows I've seen implemented in companies so far, hardly ever the same people deal with fixing package metadata so that the downloader is able to download a package vs. making legal conclusions (e.g. by setting a concluded license). Due to these very different audiences, these things should also not be configured in the same file, so that files can more easily be edited independently.
- Regarding the rules API, we already hab talked several times in the developer meeting about it's current limitations, like assuming that every reported violation is somehow license-related, and @mnonnenmacher's refactoring plans.

> What I understand is that the proposal is to separate legal conclusions from other meta-data curations in ORT's configuration interface (the user facing API), so that the different "roles" involved in the process only need looking into the config which is relevant to them.

Correct, also see what I've written above.

> So, is there need for touching the rules API as well as the OrtResult?

Touching the rules API in this PR was just a by-product of the refactoring. When starting with the refactoring to replace usages of `Package` with `CuratedPackage` where needed, because the `concludedLicense` field has moved, I realized the independent use of `pkg` and `curations` in the rules, which just did not make any sense to me (neither before nor after my changes). Looking at the current code base only, I could have also just dropped `curations` from the rules (as I actually agree with you when you say that rules hardly access curations), but that wouldn't have worked with the upcoming changes, as rules may want to access the concluded license.

However, now that I think about it again, maybe the `curations` could *still* be dropped from the rules, as the rules would have access to the concluded license via the `ResolvedLicenseInfo`. Correct @mnonnenmacher?

> In short, why does OrtResult and rules API need to be refactored? To make it readable by legal persons?

I don't think that (purely) legal persons will every be able to write (bigger) rules, as it'll still be too complicated even after a refactoring. I mean, these people hardly ever have IDEs installed to get auto-completion etc.

But yes, we need to make writing rules easier in any case.

Regarding the `OrtResult`, I don't see *so* much need for refactoring there; but still, e.g. the things that can be curated / configured separately, should still stay separate in the `OrtResult` for easy resemblance. For example, I'd liek to be able to quickly show which consecute lines of data in the `OrtResult` are affected by a metadata correct vs. a legal conclusion.

> Apart from that, have you considered using ORT's configuration API (almost) as-is to solve the problem?

I believe you're confusing the (programmatic) API (i.e. functions) a bit with data models and / or serialized files. But in the end, yes, a split on the file basis similar to what you suggest is also on my mind. But that discussion should not be held as part of this PR, IMO.

mnonnenmacher:
I am not going to comment on all points of the already very long discussion, I just want to add some thoughts about the three commits in this PR:

1.  Changing `PackageRule` to get a `CuratedPackage`: That makes sense to me, if we have a class that contains both the package metadata and the curations, why not use it and instead pass both separately? This makes even more sense if more properties are going to be added to `CuratedPackage`, like there are plans to add concluded copyrights as well. But because it's a breaking change I have added this label to the PR.
2. Move `concludedLicense` to `CuratedPackage`: That makes sense to me as well, especially in the context of the previously mentioned new properties which might be added to `CuratedPackage`. A better separation of package metadata and user provided values like legal conclusions is a better design in my opinion.
3. Rename `pkg` to `metadata`: Also makes sense to me, since I suggested that renaming ;)

mnonnenmacher:
> However, now that I think about it again, maybe the curations could still be dropped from the rules, as the rules would have access to the concluded license via the ResolvedLicenseInfo. Correct @mnonnenmacher?

That is correct, I'm undecided about the question if package rules need access to the list of applied curations. But if more properties are added to `CuratedPackage` there will also be more uses cases where a `PackageRule` might want to have access to them.

fviernau:
> I am not going to comment on all points of the already very long discussion, I just want to add some thoughts about the three commits in this PR:
> 
>     1. Changing `PackageRule` to get a `CuratedPackage`: That makes sense to me, if we have a class that contains both the package metadata and the curations, why not use it and instead pass both separately? This makes even more sense if more properties are going to be added to `CuratedPackage`, like there are plans to add concluded copyrights as well. But because it's a breaking change I have added this label to the PR.
> 
>     2. Move `concludedLicense` to `CuratedPackage`: That makes sense to me as well, especially in the context of the previously mentioned new properties which might be added to `CuratedPackage`. A better separation of package metadata and user provided values like legal conclusions is a better design in my opinion.
> 
>     3. Rename `pkg` to `metadata`: Also makes sense to me, since I suggested that renaming ;)

IIUC the whole above argumentation is based on the underlying assumption(s) that it's good to move `concludedLicense` to `CuratedPackage` and that it's good to place `concludedCopyrights` in `CuratedPackage` not in `Package`. This underlying assumption was exactly what I was challenging in above lengthy discussion.

sschuberth:
> So, can we discuss pros and cons of (1) having all properties in `Package` vs. (2) having some properties (which exactly?) in `Package` and some in `CuratedPackage`?

To be honest, I'm a little bit exhausted by this surprisingly lengthy discussion about something that I (and probably also @mnonnenmacher) regard as straight-forward changes.

But sure @fviernau, please go ahead and provide your pros for keeping / putting properties in `Package` that don't actually come from the package manager, esp. given the fact that we already *have*  the `CuratedPackage` class. Also, please provide cons for having a clean separation between package metadata as provided by the manager, and data that can only come from curations.

mnonnenmacher:
> IIUC the whole above argumentation is based on the underlying assumption(s) that it's good to move `concludedLicense` to `CuratedPackage` and that it's good to place `concludedCopyrights` in `CuratedPackage` not in `Package`. This underlying assumption was exactly what I was challenging in above lengthy discussion, and if that underlying assumption would not hold the argumentation would fall apart completely.

There is no underlying assumption, in (2) I explain that I think it is a good idea to separate package metadata from user provided values like the concluded license, and (1) is mostly independent of the refactoring of the concluded license.

> So, can we discuss pros and cons of (1) having all properties in `Package` vs. (2) having some properties (which exactly btw.?) in `Package` and some in `CuratedPackage`?

@sschuberth Maybe you could provide a draft of what the final classes should look like in your opinion if legal conclusions are separated from metadata and also concluded copyrights are added? I think this might help to understand the motivation behind the change.

Just some independent ideas about the general refactoring that we could discuss later:
If this change is implemented we could consider renaming `CuratedPackage` to `Package` and `Package` to `PackageMetadata`. Then it would probably also be good to move the `id` to the main `Package` class for accessibility, and we could also consider to reuse `PackageMetadata` inside `Project` to reduce duplication.

sschuberth:
> @sschuberth Maybe you could provide a draft of what the final classes should look like in your opinion if legal conclusions are separated from metadata and also concluded copyrights are added? I think this might help to understand the motivation behind the change.

Not sure if it help, but in the context of #4519 / #5680 I envision the `CuratedPackage` class to get an additional `concludedCopyrights` poroperty so that it looks like

```
data class CuratedPackage(
    /**
     * The curated package after applying the [curations].
     */
    @JsonAlias("package")
    val metadata: Package,

    /**
     * The concluded license as an [SpdxExpression]. It can be used to override the [declared][declaredLicenses] /
     * [detected][LicenseFinding.license] licenses of a package.
     */
    val concludedLicense: SpdxExpression? = null,

    /**
     * The set of concluded copyright statements for this package. It can be used to override the
     * [detected copyright statements][CopyrightFinding.statement].
     */
    @JsonInclude(JsonInclude.Include.NON_DEFAULT)
    val concludedCopyrights: SortedSet<String> = sortedSetOf(),

    /**
     * The curations in the order they were applied.
     */
    val curations: List<PackageCurationResult> = emptyList()
)
```

> If this change is implemented we could consider renaming `CuratedPackage` to `Package` and `Package` to `PackageMetadata`. Then it would probably also be good to move the `id` to the main `Package` class for accessibility, and we could also consider to reuse `PackageMetadata` inside `Project` to reduce duplication.

I believe these are some good ideas. Esp. extracting common stuff of `Package` and `Project`.

reporter-web-app/src/models/Metadata.js
```diff
@@ -18,7 +18,7 @@
  * License-Filename: LICENSE
  */
 
-class MetaData {
+class Metadata {
     #analyzerStartTime;
```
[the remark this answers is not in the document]
sschuberth: @tsteenbe do you have any clue what Lift is complaining about here?

sschuberth:
Please have a look again (esp. @mnonnenmacher) as I've dropped the commit that moves `concludedLicense` from `Package` to `CuratedPackage` for now.

mnonnenmacher:
> Please have a look again (esp. @mnonnenmacher) as I've dropped the commit that moves `concludedLicense` from `Package` to `CuratedPackage` for now.

Looks good to me, but I'll wait with the approval until the next workday because of the required adaptations in our rules.

sschuberth:
> > Please have a look again (esp. @mnonnenmacher) as I've dropped the commit that moves `concludedLicense` from `Package` to `CuratedPackage` for now.
> 
> Looks good to me, but I'll wait with the approval until the next workday because of the required adaptations in our rules.

@bennati here's another heads-up for a PR with breaking changes you probably need to adapt your rules to. Your changes probably would need to look similar to [these](https://github.com/oss-review-toolkit/ort-config/pull/57/commits/fa4f1553ac3cf1f3bc6cba2aceb2e9c920501f3b#diff-974de7809631d34f6f6faebfb72ac052fafb8659162c6aabca8c4031814ea58e).

bennati:
Thanks, please go ahead

sschuberth:
> Thanks, please go ahead

@mnonnenmacher, @fviernau if you're also ok with moving forward with this, please approve. I'd like to merge this before I need to catch up with new changes.

sschuberth:
> please wait with the merge until tomorrow because I don't have time to do a synchronized merge today.

Are we good to go @mnonnenmacher?

mnonnenmacher:
> > please wait with the merge until tomorrow because I don't have time to do a synchronized merge today.
> 
> Are we good to go @mnonnenmacher?

Yes, go ahead.

PR #5727, opened 2022-09-01
ProcessCapture: Fix temporary directory

nnobelis:
Currently, the temporary directory is the `toString` of the vararg
parameter `command`, resulting in temporary directory name such as:
[Ljava.lang.String;@44864536-process5910638297610510207.
This commit fixes the temporary directory name.


utils/common/src/main/kotlin/ProcessCapture.kt
```diff
@@ -58,8 +58,8 @@ class ProcessCapture(vararg command: String, workingDir: File? = null, environme
         }
     }
 
-    private val tempDir = createTempDirectory("$command-process").toFile().apply { deleteOnExit() }
     private val tempPrefix = command.first().substringAfterLast(File.separatorChar)
+    private val tempDir = createTempDirectory("$tempPrefix-process").toFile().apply { deleteOnExit() }
```
fviernau, line 62: There is also `createOrtTempDir` which has more usages than `createTempDir`. 
Have you considered using that one? (I'm not 100% sure if `createOrtTempDir()` should actually be used, so maybe get a second approval)

fviernau (dismissed):

utils/common/src/main/kotlin/ProcessCapture.kt
nnobelis: Pinging @sschuberth for advice.

sschuberth:
> Currently, the temporary directory is the `toString` of the vararg
> parameter `command`, resulting in temporary directory name such as:
> [Ljava.lang.String;@44864536-process5910638297610510207.

Is that actually so? I've seen several temporary directory names than seemed actually correct. And this [code snippet](https://pl.kotl.in/VwKmYnmVS) seems to prove that `command.first()` returns the expected string.

nnobelis:
> > Currently, the temporary directory is the `toString` of the vararg
> > parameter `command`, resulting in temporary directory name such as:
> > [Ljava.lang.String;@44864536-process5910638297610510207.
> 
> Is that actually so? I've seen several temporary directory names than seemed actually correct. And this [code snippet](https://pl.kotl.in/VwKmYnmVS) seems to prove that `command.first()` returns the expected string.

Yes `command.first` works but I you look at the code before my changes, `command.toString` was implicitly called to create the temp folder.

The sample name I put in the commit message is what I saw on our production server.

sschuberth:
> Yes `command.first` works but I you look at the code before my changes, `command.toString` was implicitly called to create the temp folder.

Sorry, I misread the diff a you also swapped the lines.

utils/common/src/main/kotlin/ProcessCapture.kt
sschuberth: As `common-utils` must not depend on `ort-utils` we cannot use `createOrtTempDir()` here.

However, one might argue that we don't need / want `tempPrefix` in both the directory *and* the file names.

Maybe do

    private val tempDir = createTempDirectory(javaClass.simpleName).toFile().apply { deleteOnExit() }

and while at it, in another commit rename `tempPrefix` to `commandName` so that it all reads like

    private val tempDir = createTempDirectory(javaClass.simpleName).toFile().apply { deleteOnExit() }
    private val commandName = command.first().substringAfterLast(File.separatorChar)
    private val stdoutFile = tempDir.resolve("$commandName.stdout").apply { deleteOnExit() }
    private val stderrFile = tempDir.resolve("$commandName.stderr").apply { deleteOnExit() }

nnobelis: Will `javaClass.simpleName` not always return `ProcessCapture` ?
Is is not better to have `commandName` as part of the temp file ?

sschuberth: > Will `javaClass.simpleName` not always return `ProcessCapture` ?

Yes.

> Is is not better to have commandName as part of the temp file ?

`commandName` is still used for the names of the *files*, but not anymore for the name of the parent *directory*, which is fine, IMO.

utils/common/src/main/kotlin/ProcessCapture.kt
```diff
@@ -59,10 +59,10 @@ class ProcessCapture(vararg command: String, workingDir: File? = null, environme
     }
 
     private val tempDir = createTempDirectory(javaClass.simpleName).toFile().apply { deleteOnExit() }
-    private val tempPrefix = command.first().substringAfterLast(File.separatorChar)
 
-    private val stdoutFile = tempDir.resolve("$tempPrefix.stdout").apply { deleteOnExit() }
-    private val stderrFile = tempDir.resolve("$tempPrefix.stderr").apply { deleteOnExit() }
+    private val commandName = command.first().substringAfterLast(File.separatorChar)
```
sschuberth, line 63: Nit: Let's also drop empty line 62 while at it, as all 4 lines actually belong together.

sschuberth (approved):

PR #5728, opened 2022-09-01
ScanResult: Preserve additional data when filtering scan results

nnobelis:
This solves a bug where the [FossIdReporter] could not generate the
reports because the additional data was missing.

### DIFF
--- a/model/src/main/kotlin/ScanResult.kt
+++ b/model/src/main/kotlin/ScanResult.kt
@@ -56,15 +56,14 @@ data class ScanResult(
      * for [provenance]. Findings which [RootLicenseMatcher] assigns as root license files for [path] are also kept.
      */
     fun filterByPath(path: String): ScanResult =
-        takeIf { path.isBlank() } ?: ScanResult(
-            provenance = if (provenance is RepositoryProvenance) {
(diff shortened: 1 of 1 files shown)


sschuberth:
> This solves a bug where the [FossIdReporter] could not generate the

Please use backticks instead of square brackets here.

model/src/main/kotlin/ScanResult.kt
```diff
@@ -64,6 +64,7 @@ data class ScanResult(
             },
             scanner = scanner,
             summary = summary.filterByPath(path),
+            additionalData = additionalData
```
sschuberth, line 67: How about instead leveraging `copy` so that fields cannot be forgotten? Like

        takeIf { path.isBlank() } ?: if (provenance is RepositoryProvenance) {
            copy(
                provenance = provenance.copy(vcsInfo = provenance.vcsInfo.copy(path = path)),
                summary = summary.filterByPath(path)
            )
        } else {
            copy(summary = summary.filterByPath(path))
        }

sschuberth (approved):

nnobelis:
@sschuberth Can we please merge this ?

sschuberth:
> @sschuberth Can we please merge this ?

I've retriggered the CI jobs that failed. Auto-merge is enabled.

PR #5731, opened 2022-09-02
Improve the generation of path excludes

fviernau:
Fix an issue, add some tests and handle further directories.

### DIFF
--- a/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
+++ b/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
@@ -57,16 +57,20 @@ internal object PathExcludeGenerator {
     }
 }
 
-private fun getAllDirs(files: Set<File>): Set<File> {
-    val result = mutableSetOf<File>()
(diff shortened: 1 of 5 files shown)


helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -38,22 +39,26 @@ internal object PathExcludeGenerator {
...
+            if (dir.getAncestorFiles().intersect(dirsToExclude.keys).isEmpty()) {
+                result += PathExclude(
+                      pattern = "${dir.path}/**",
+                      reason = reason
+                )
+            }
         }
 
-        return greedySetCover(filesForPathExcludes).toList()
+        return result.toList()
```
sschuberth, line 61: Is `greedySetCover` now still used, or can it be removed?

fviernau: This is still used by another helper-cli command. So, it cannot be removed.
Note: I'll likely need to use it again for the path exclude generation.

sschuberth (approved):

PR #5732, opened 2022-09-02
PathExcludeGenerator: Use globs for matching directory names

fviernau:


sschuberth:
Isn't `helper-cli/expected-excluded-dirs.txt` being added a the wrong path?

fviernau:
> Isn't `helper-cli/expected-excluded-dirs.txt` being added a the wrong path?

Good catch, fixed it.

helper-cli/src/test/assets/path-exclude-gen/expected-excluded-dirs.txt
```diff
@@ -0,0 +1,981 @@
+.github/**
+.idea/**
+.mvn/**
+.teamcity/.mvn/**
+.teamcity/src/test/**
```
sschuberth, line 5: There's a lot of `/src/test/` paths in here... how about first changing the path exclude generator to exclude the `**/src/test/**` pattern?

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -42,8 +42,10 @@ internal object PathExcludeGenerator {
         val dirsToExclude = mutableMapOf<File, PathExcludeReason>()
 
         dirs.forEach { dir ->
-            PATH_EXCLUDES_REASON_FOR_DIR_NAME[dir.name]?.let { reason ->
-                dirsToExclude += dir to reason
+            PATH_EXCLUDES_REASON_FOR_DIR_NAME.forEach { (dirname, reason) ->
```
sschuberth, line 45: Commit message:
- Typo in "cleaar"
- What does the "just" in "Decided to use just wildcards" refer to? Why "just"? What's the alternative?

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -62,6 +64,19 @@ internal object PathExcludeGenerator {
     }
 }
 
+private val REGEX_SPECIAL_CHARACTERS = listOf('.', '+', '*', '?', '^', '$', '(', ')', '[', ']', '{', '}', '|', '\\')
+private val REGEX_IGNORE_CASE_OPTION = setOfNotNull(RegexOption.IGNORE_CASE)
+private fun String.matchWildcard(value: String, ignoreCase: Boolean = false): Boolean {
+    val pattern = REGEX_SPECIAL_CHARACTERS.filterNot { it == '*' }.fold(
```
sschuberth, line 70: If you're always filtering "*" from the constant `REGEX_SPECIAL_CHARACTERS`, why not just omit it from the list to begin with?

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -62,6 +64,19 @@ internal object PathExcludeGenerator {
     }
 }
 
+private val REGEX_SPECIAL_CHARACTERS = listOf('.', '+', '*', '?', '^', '$', '(', ')', '[', ']', '{', '}', '|', '\\')
+private val REGEX_IGNORE_CASE_OPTION = setOfNotNull(RegexOption.IGNORE_CASE)
+private fun String.matchWildcard(value: String, ignoreCase: Boolean = false): Boolean {
```
sschuberth, line 69: Can't you just treat the string as a path and then use the `FileMatcher` to match the string against a wildcard / glob pattern?

helper-cli/src/test/assets/path-exclude-gen/expected-excluded-dirs.txt
fviernau: If at all, I wanted to do patterns starting with `**/` in a later state, making it toggle-able by the user and configurable by a threshold. So, then it still makes sense to have a test with that option disabled. Note: The impact of not having `**/` in most of the cases should be small, as the generation is supposed to be run only against one package at a time, so against way less file paths.

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
fviernau: I found the code cleaner this way, but will change it according to your preference.

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
fviernau: I've updated the commit message in response to your above comment, which explains why I decided against that.

sschuberth: Did you forget to push? I'm still seeing the old commit message (also incl. the "cleaar" typo).

fviernau: Pushed again, should be up-to-date now.

sschuberth: Sorry, I'm having trouble understanding what you're trying to say in the commit message:

> Using wildcards has the advantage of matching more relevant cases

More relevant uses cases than *what*? What do you compare to?

> but it also introduces some false positives.

Why? How?

> as it is not yet clear whether these patterns should also end-up in the `PathExclude.pattern`

Why is that not clear yet? Who decides that?

The `pattern` property is documented as "A glob to match the path of the project definition file", so I always assumed it's perfectly valid to use full-blown globs here, not just wildcards.

I'm missing the overall picture of what you're trying to do. Are you trying to limit yourself *by convention* to just using wildcards instead of globs? And if so, why? Are you afraid that users might not be able to deal with the additional syntax that globs support over wildcards?

fviernau: > More relevant uses cases than _what_? What do you compare to?

compared to without wild cards.

> Why? How?

The matcher becomes broader so matches more dir names, maybe unintendedly some which should not be excluded.

> Why is that not clear yet? Who decides that?


Currently, the generated path excludes don't contain any `*` besides the `/**` suffix.
I plan to experiment with using the `*` also in generated path excludes. 
From these experiments I expect to gain some learnings, which will make it clearer to me.

fviernau: I went for glob expressions as they already give good results for the test data and are simpler. 
This may resolve the discussion, please take a look at the updated changes.

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -26,6 +26,7 @@ import org.ossreviewtoolkit.model.config.PathExcludeReason
 import org.ossreviewtoolkit.model.config.PathExcludeReason.BUILD_TOOL_OF
 import org.ossreviewtoolkit.model.config.PathExcludeReason.DOCUMENTATION_OF
 import org.ossreviewtoolkit.model.config.PathExcludeReason.TEST_OF
+import org.ossreviewtoolkit.utils.common.FileMatcher
```
sschuberth, line 29: Commit message:

> Compared to regular expression

It still confuses me to refer to regular expressions here, because to the reader there never was the alternative to use regular expressions instead of globs; you straight introduce globs without having had regular expressions before.

How about something like this:

> Make the search in the project's source tree for paths to be excluded more flexible by using globs. Any matching directory paths are then added to the path excludes with a `/**` suffix.

Additionally, it would be nice if you could explain why you're not simply writing out these globs as-is to the path excludes. My *guess* is that you want to keep path excludes to a minimum and only list patterns that actually match something in the project's source tree instead of always listing a (big) list of default excludes. So maybe also say:

> Note that these globs are expanded at the runtime of the helper command in order to create a minimal set of path excludes that actually match paths in the project's source tree. So, while these globs could also haven been used as path exclude patterns as-is, this is not done to avoid having patterns that do not match anything.

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -121,4 +121,10 @@ private val PATH_EXCLUDES_REASON_FOR_DIR_NAME = mapOf(
     "tools" to BUILD_TOOL_OF,
     "tutorial" to DOCUMENTATION_OF,
     "winbuild" to BUILD_TOOL_OF
-)
+).apply {
+    val duplicatePatterns = groupBy { it.first }.filter { it.value.size > 1 }.map { it.key }
```
sschuberth, line 125: Can you use one of our existing `getDuplicates()` extension functions?

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -43,11 +43,11 @@ internal object PathExcludeGenerator {
         val dirsToExclude = mutableMapOf<File, PathExcludeReason>()
 
         dirs.forEach { dir ->
-            PATH_EXCLUDES_REASON_FOR_DIR_NAME.forEach { (pattern, reason) ->
-                if (FileMatcher.match(pattern, dir.name, ignoreCase = true)) {
-                    dirsToExclude += dir to reason
-                }
-            }
+            val (_, reason) = PATH_EXCLUDES_REASON_FOR_DIR_NAME.find { (pattern, _) ->
```
sschuberth, line 46: Please simply squash this to the commit in this PR where you introduced this code.

sschuberth (changes requested):

sschuberth (approved):

PR #5745, opened 2022-09-06
GitWorkingTree: Properly close the repository after each operation

sschuberth:
Previously, the `Repository` instance remained open during the lifetime
of the `GitWorkingTree`. That is a problem if some clean-up code (like
in tests) wants to delete a cloned Git repository while there is still
a GitWorkingTree instance.

To avoid that, provide a getter for the `repo` property which allows the
caller to auto-close the `Repository` instance after use as on the next
access of the `repo` property it will be re-evaluated. The added runtime
overhead due to this should be neglible compared to the duration of the
Git operation.

### DIFF
--- a/downloader/src/main/kotlin/vcs/GitWorkingTree.kt
+++ b/downloader/src/main/kotlin/vcs/GitWorkingTree.kt
@@ -53,11 +53,12 @@ private fun findGitOrSubmoduleDir(workingDirOrFile: File): Repository {
 open class GitWorkingTree(workingDir: File, vcsType: VcsType) : WorkingTree(workingDir, vcsType) {
     companion object : Logging
 
-    private val repo = findGitOrSubmoduleDir(workingDir)
+    private val repo: Repository
+        get() = findGitOrSubmoduleDir(workingDir)
 
-    override fun isValid(): Boolean = repo.objectDatabase?.exists() == true
+    override fun isValid(): Boolean = repo.use { it.objectDatabase?.exists() == true }
 
-    override fun isShallow(): Boolean = repo.directory?.resolve("shallow")?.isFile == true
+    override fun isShallow(): Boolean = repo.use { it.directory?.resolve("shallow")?.isFile == true }
 
     private fun listSubmodulePaths(repo: Repository): List<String> {
         fun listSubmodules(parent: String, repo: Repository, paths: MutableList<String>) {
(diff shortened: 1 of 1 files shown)


downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -53,11 +53,12 @@ private fun findGitOrSubmoduleDir(workingDirOrFile: File): Repository {
 open class GitWorkingTree(workingDir: File, vcsType: VcsType) : WorkingTree(workingDir, vcsType) {
     companion object : Logging
 
-    private val repo = findGitOrSubmoduleDir(workingDir)
+    private val repo: Repository
```
fviernau, line 56: IIUC the constructor threw an exception previously, and as of now not anymore.
Shall we ensure it fails early again?

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -53,11 +53,12 @@ private fun findGitOrSubmoduleDir(workingDirOrFile: File): Repository {
...
+        get() = findGitOrSubmoduleDir(workingDir)
 
-    override fun isValid(): Boolean = repo.objectDatabase?.exists() == true
+    override fun isValid(): Boolean = repo.use { it.objectDatabase?.exists() == true }
```
fviernau, line 59: Commit-msg: The Repository class / interface seems to have an internal `useCount`, with the initial value of 1.
The class seems to release the resources when the count reaches 0. 

From the commit message I do not understand why the releasing of the resources is now triggered.
In particular what the actual trigger now is. Can you re-phrase it to explain that?
In particular, it should cover why the trigger is guaranteed to take place before the mentioned test clean-ups.

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -89,41 +90,55 @@ open class GitWorkingTree(workingDir: File, vcsType: VcsType) : WorkingTree(work
     }
 
     override fun getNested(): Map<String, VcsInfo> =
-        listSubmodulePaths(repo).associateWith { GitWorkingTree(repo.workTree.resolve(it), vcsType).getInfo() }
+        repo.use {
```
fviernau, line 93: Have you considered exposing the same mechanism as the Repository class, e.g. `open` and `close()` functions instead? I believe with such approach it would be easier to move towards thread safety (hm, may we need that in the future? if not irrelevant). Note that not all subclasses of `Repository` are thread safe but the `FileRepository` is.

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -89,41 +90,55 @@ open class GitWorkingTree(workingDir: File, vcsType: VcsType) : WorkingTree(work
...
 
-    override fun getRootPath(): File = repo.workTree
+    override fun getRootPath(): File = repo.use { it.workTree }
 
     override fun listRemoteBranches(): List<String> =
-        runCatching {
-            LsRemoteCommand(repo).setHeads(true).call().map { it.name.removePrefix("refs/heads/") }
-        }.getOrElse {
-            throw IOException("Unable to list the remote branches.", it)
+        repo.use {
```
fviernau, line 124: The diffs of this and below function contain a couple of unrelated changes.

fviernau (changes requested):

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -53,11 +53,12 @@ private fun findGitOrSubmoduleDir(workingDirOrFile: File): Repository {
...
 
-    private val repo = findGitOrSubmoduleDir(workingDir)
+    private val repo: Repository
+        get() = findGitOrSubmoduleDir(workingDir)
 
-    override fun isValid(): Boolean = repo.objectDatabase?.exists() == true
+    override fun isValid(): Boolean = repo.use { it.objectDatabase?.exists() == true }
 
-    override fun isShallow(): Boolean = repo.directory?.resolve("shallow")?.isFile == true
+    override fun isShallow(): Boolean = repo.use { it.directory?.resolve("shallow")?.isFile == true }
```
fviernau, line 61: Consider factoring out `repo.use` to a function.

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
sschuberth: > Have you considered exposing the same mechanism as the Repository class, e.g. incrementOpen and close() functions instead?

No, I'd prefer to not reimplement that logic here.

> Note that not all subclasses of `Repository` are thread safe but the `FileRepository` is.

Aren't we good then already in terms of thread safety? Anyway, the goal of this PR is not thread-safety, but to avoid a resource leak. Let's do one topic at a time.

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
sschuberth: Sorry, no. I'm not going to put an 8-letters-call to a function (whose name would probably longer).

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
sschuberth: Unfortunately, you're not very specific which "unrelated" changes you refer to. So once again, I can only guess that you mean the renaming of lambda arguments.

This first renaming to `branch` becomes necessary in order to not shadow the `it` from `use`. The second renaming to `e` was indeed not complained about by the IDE, but IMO it should, as keeping the original code would also shadow the outer `it`.

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
sschuberth: I actually think it's even better the new way: There might be cases where you'd like to create a `GitWorkingTree` instance that points to a `workingDir` that yet needs to be initialized (maybe even in the scope of file existence checks in rules).

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
fviernau: I didn't notice the parameter shadowing, so I'm fine with the changes.

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
fviernau: > Anyway, the goal of this PR is not thread-safety, but to avoid a resource leak. Let's do one topic at a time.

I was a bit unclear. My goal was not to ask you to put thread safety in scope of the PR. My intention was rather to bring it to attention and to clarify if we think it will be needed in future. Anyhow, I'm convinced now that it won't make moving towards thread safety (much) more difficult...

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
fviernau: Hm, I believe failing early is good by default unless there is a particular case which satisfies not doing so.
I'd prefer keeping fail early until we come across a case which requires it's removal. But am not going to block this.

fviernau (approved):

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
sschuberth: > I believe failing early is good by default

I agree, but what I'm trying to say is that creating a `GitWorkingTree` instance wrapping a directory that is no Git tree yet is not necessarily a failure. It only has to be a Git tree once a Git operation is executed on it.

That's BTW also how the other (purely CLI-based) `WorkingTree` implementations work: They will only throw an exception at the time a CLI tool is run and fails, not at instance construction time. So this commits also aligns the behavior across `WorkingTree` implementations.

PR #5751, opened 2022-09-07
Test change python version 3.9

bennati:


mnonnenmacher:
@bennati It is not necessary anymore to change the installed Python version. With the switch to python-inspector you can now configure the Python version that shall be used for dependency resolution either in the .ort.yml file or in the global configuration, without having that version installed, see: https://github.com/oss-review-toolkit/ort/blob/4730602b8fd6b6dcf7c329f4da2f3bb7f8e1b46d/analyzer/src/main/kotlin/managers/Pip.kt#L67-L70

There are only a few `python setup.py` calls left which will be removed very soon, I think you can close this PR.

PR #5753, opened 2022-09-07
Add the effective license to the web app reporter

mnonnenmacher:

### DIFF
--- a/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
+++ b/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
@@ -41,6 +41,7 @@ import org.ossreviewtoolkit.model.config.PathExclude
 import org.ossreviewtoolkit.model.config.RuleViolationResolution
 import org.ossreviewtoolkit.model.config.ScopeExclude
 import org.ossreviewtoolkit.model.config.VulnerabilityResolution
+import org.ossreviewtoolkit.model.licenses.LicenseView
 import org.ossreviewtoolkit.model.utils.FindingCurationMatcher
 import org.ossreviewtoolkit.model.utils.FindingsMatcher
 import org.ossreviewtoolkit.model.utils.RootLicenseMatcher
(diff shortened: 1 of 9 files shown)


tsteenbe (approved):
Did a quick review of JavaScript code, all looks good to me.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
```diff
@@ -306,6 +307,11 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
             detectedLicenses = detectedLicenses,
             detectedExcludedLicenses = detectedExcludedLicenses,
             concludedLicense = pkg.concludedLicense,
+            effectiveLicense = input.licenseInfoResolver.resolveLicenseInfo(pkg.id).filterExcluded().effectiveLicense(
+                LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED,
```
sschuberth, line 311: BTW, I just checked all callers of `effectiveLicense()`, and we currently sometimes use `LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED` (like here), but sometimes `LicenseView.ALL`. Would you mind double-checking whether that makes sense?

fviernau, line 311: Have license choices been already applied and if so, does the view need to be consistent with what was supplied to `applyChoices()` as view?

mnonnenmacher: We only use `LicenseView.ALL` in the `DownloaderCommand` which is a special case I believe, and in tests. `LicenseView.CONCLUDED_OR_DECLARED_AND_DETECTED` is what we consider the default and use in several other places, especially the static HTML reporter. We might want to consider making the default configurable later on.

mnonnenmacher: Yes, the choices are applied inside the `effectiveLicense` function based on the supplied view.

mnonnenmacher:
> Did a quick review of JavaScript code, all looks good to me.

I have implemented it exactly the same as the concluded license and have manually verified that it works.

PR #5754, opened 2022-09-07
Evaluator: Allow making arbitrary file checks on the project's source, take 2

fviernau:
See individual commits.

Note: The project's source tree will be cloned in the evaluator only if the rules needed it / if the file checks are actually used.

This implements parts of https://github.com/oss-review-toolkit/ort/issues/5621.

### DIFF
--- a/evaluator/src/main/kotlin/ProjectSourceRule.kt
+++ b/evaluator/src/main/kotlin/ProjectSourceRule.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2022 EPAM Systems, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.evaluator
+
+import java.io.File
+
+import org.ossreviewtoolkit.model.OrtResult
+
+/**
+ * An [OrtResultRule] which allows downloading the project's source code if needed.
+ */
+open class ProjectSourceRule(
+    ruleSet: RuleSet,
+    name: String,
+    projectSourceResolver: SourceTreeResolver = ruleSet.ortResult.createResolver()
+) : OrtResultRule(ruleSet, name) {
+    /**
+     * The directory containing the source code of the project. Accessing the property for the first time triggers a
+     * clone and may take a while.
+     */
+    val projectSourcesDir: File by lazy { projectSourceResolver.rootDir }
+}
+
+private fun OrtResult.createResolver() =
+    SourceTreeResolver.forRemoteRepository(repository.vcsProcessed)
--- a/evaluator/src/main/kotlin/ProjectSourceRule.kt
+++ b/evaluator/src/main/kotlin/ProjectSourceRule.kt
@@ -22,6 +22,7 @@ package org.ossreviewtoolkit.evaluator
 import java.io.File
 
 import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.utils.common.FileMatcher
 
 /**
  * An [OrtResultRule] which allows downloading the project's source code if needed.
@@ -36,6 +37,24 @@ open class ProjectSourceRule(
      * clone and may take a while.
      */
     val projectSourcesDir: File by lazy { projectSourceResolver.rootDir }
+
+    /**
+     * Return all files from the project's source tree which match any of the provided [glob expressions][patterns].
+     */
+    fun projectSourceFindFiles(vararg patterns: String): List<File> =
+        projectSourcesDir.walkBottomUp().filterTo(mutableListOf()) {
+            it.isFile && FileMatcher.match(patterns.toList(), it.relativeTo(projectSourcesDir).path)
+        }
+
+    /**
+     * A [RuleMatcher] that checks whether the project's source tree contains at least one file matching any of the
+     * provided [glob expressions][patterns].
+     */
+    fun projectSourceHasFile(vararg patterns: String): RuleMatcher =
+        object : RuleMatcher {
+            override val description = "projectSourceHasFile('${patterns.joinToString()}')"
+            override fun matches(): Boolean = projectSourceFindFiles(*patterns).isNotEmpty()
+        }
 }
 
 private fun OrtResult.createResolver() =
--- a/evaluator/src/main/kotlin/SourceTreeResolver.kt
+++ b/evaluator/src/main/kotlin/SourceTreeResolver.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2022 EPAM Systems, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.evaluator
+
+import java.io.File
+
+import org.apache.logging.log4j.kotlin.Logging
+
+import org.ossreviewtoolkit.downloader.Downloader
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.SourceCodeOrigin
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.utils.ort.createOrtTempDir
+
+class SourceTreeResolver private constructor(getRepositoryRoot: () -> File) {
+    companion object : Logging {
+        fun forRemoteRepository(vcsInfo: VcsInfo) =
+            SourceTreeResolver {
+                val downloadDir = createOrtTempDir()
+                val downloaderConfiguration = DownloaderConfiguration(sourceCodeOrigins = listOf(SourceCodeOrigin.VCS))
+                val downloader = Downloader(downloaderConfiguration)
+                val pkg = Package.EMPTY.copy(vcsProcessed = vcsInfo)
+
+                downloader.download(pkg, downloadDir)
+
+                downloadDir
+            }
+    }
+
+    val rootDir: File by lazy {
+        logger.info {
+            "The evaluator rules attempt to access the project's source code for the first time. Fetching sources..."
+        }
+
+        getRepositoryRoot()
+    }
+}
--- a/evaluator/src/main/kotlin/SourceTreeResolver.kt
+++ b/evaluator/src/main/kotlin/SourceTreeResolver.kt
@@ -43,6 +43,8 @@ class SourceTreeResolver private constructor(getRepositoryRoot: () -> File) {
 
                 downloadDir
             }
+
+        fun forLocalDirectory(dir: File) = SourceTreeResolver { dir }
     }
 
     val rootDir: File by lazy {
--- a/evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
+++ b/evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
@@ -1,0 +1,91 @@
+/*
+ * Copyright (C) 2022 EPAM Systems, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.evaluator
+
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.shouldBe
+
+import java.io.File
+
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.utils.test.createSpecTempDir
+
+class ProjectSourceRuleTest : WordSpec({
+    "projectSourceHasFile()" should {
+        "return true if at least one file matches the given glob pattern" {
+            val dir = createSpecTempDir().apply {
+                addFiles(
+                    "README.md",
+                    "module/docs/LICENSE.txt"
+                )
+            }
+            val rule = createOrtResultRule(dir)
+
+            with(rule) {
+                projectSourceHasFile("README.md").matches() shouldBe true
+                projectSourceHasFile("**/README.md").matches() shouldBe true
+                projectSourceHasFile("**/LICENSE*").matches() shouldBe true
+                projectSourceHasFile("**/*.txt").matches() shouldBe true
+            }
+        }
+
+        "return false if only a directory matches the given glob pattern" {
+            val dir = createSpecTempDir().apply {
+                addDirs("README.md")
+            }
+            val rule = createOrtResultRule(dir)
+
+            rule.projectSourceHasFile("README.md").matches() shouldBe false
+        }
+
+        "return false if neither any file nor directory matches the given glob pattern" {
+            val dir = createSpecTempDir()
+            val rule = createOrtResultRule(dir)
+
+            rule.projectSourceHasFile("README.md").matches() shouldBe false
+        }
+    }
+})
+
+private fun createOrtResultRule(projectSourcesDir: File): ProjectSourceRule =
+    ProjectSourceRule(
+        ruleSet = ruleSet(ortResult = OrtResult.EMPTY),
+        name = "RULE_NAME",
+        projectSourceResolver = SourceTreeResolver.forLocalDirectory(projectSourcesDir)
+    )
+
+private fun File.addFiles(vararg paths: String) {
+    require(isDirectory)
+
+    paths.forEach { path ->
+        resolve(path).apply {
+            parentFile.mkdirs()
+            createNewFile()
+        }
+    }
+}
+
+private fun File.addDirs(vararg paths: String) {
+    require(isDirectory)
+
+    paths.forEach { path ->
+        resolve(path).mkdirs()
+    }
+}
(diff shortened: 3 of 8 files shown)


cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -310,7 +310,13 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
         val resolutionProvider = DefaultResolutionProvider.create(ortResultInput, resolutionsFile)
         val licenseClassifications =
             licenseClassificationsFile.takeIf { it.isFile }?.readValue<LicenseClassifications>().orEmpty()
-        val evaluator = Evaluator(ortResultInput, licenseInfoResolver, resolutionProvider, licenseClassifications)
+
```
sschuberth, line 332: Commit message nits:
- "for dependency source" -> "for dependency source*s*" or "for the source code of dependencies" or similar
- "not have obvious need" -> "not have obvious need*s*" or just "is usually not required"
- "exection time" -> "exec*u*tion time"

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -310,7 +310,13 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
...
+        val evaluator = Evaluator(
+            ortResultInput,
+            licenseInfoResolver,
+            resolutionProvider,
+            licenseClassifications
+        )
```
sschuberth, line 339: Unrelated formatting change.

evaluator/src/main/kotlin/Evaluator.kt
```diff
@@ -40,12 +40,21 @@ class Evaluator(
     licenseInfoResolver: LicenseInfoResolver = ortResult.createLicenseInfoResolver(),
     resolutionProvider: ResolutionProvider = DefaultResolutionProvider.create(),
     licenseClassifications: LicenseClassifications = LicenseClassifications(),
+    projectSourceTree: SourceTreeResolver = SourceTreeResolver.forRemoteRepository(ortResult.repository.vcsProcessed),
```
sschuberth, line 43: So, the whole script evaluation still gets a `SourceTreeResolver`... however, I was hoping to see a solution that does not change any host-side scripting classes at all, but only adds a new rule class derived from `OrtResultRule` which creates a `SourceTreeResolver` from the `OrtResult`.

Am I overlooking something why that's not possible? Do you want me to make a proposal based on your code?

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -203,6 +204,11 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
         help = "Do not evaluate the script but only check its syntax. No output is written in this case."
     ).flag()
 
+    private val allowSourceCodeDownload by option(
+        "--allow-source-code-download",
```
sschuberth, line 208: I don't think we need this if the API was better, by moving the source code download to a dedicated rule.

evaluator/src/main/kotlin/Evaluator.kt
fviernau: > Am I overlooking something why that's not possible? Do you want me to make a proposal based on your code?



This code is using constructor injection of dependencies for testability. E.g. the test can inject an arbitrary directory as source code directory. I understand that you dislike constructor injection in this case. So, can you tell which way of injecting a directory from the test would be acceptable for you? (Ideally along with an argumentation why)

Having the above question answered would unblock the situation. Note that I understood from previous talks that you didn't like the construct, however I didn't get why and also not what alternative you want - which is why I didn't change it. It just didn't make sense to me.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
fviernau: I just had the idea of turning
```
fun ortResultRule(name: String, configure: OrtResultRule.() -> Unit) {
```
into 
```
fun ortResultRule(name: String, configure: OrtResultRule.() -> Unit, allowSourceCodeDownload: Boolean = false) {
```
Would this eliminate the need for a dedicated rule class? I'd still prefer not introducing a dedicated class but using OrtResultRule. However, if this is the only way to unblock this pull request then I will do it.  So if you insist on this dedicated class, maybe you have a good idea for a name? (I don't)

fviernau: Dropped this.

evaluator/src/main/kotlin/Evaluator.kt
fviernau: I changed things so that it fits what I understand you have in mind. I believe all should be covered now, except for the following two things for which I was lacking your input, so I did my best guess as follows:

1. I used injection via setter override, as to my understanding you seem to dislike injecting the resolver into the constructor or the corresponding factory method. Please see if you like that more.
2. I used the name `OrtResultRule2`. Please propose a better name as you pushed for introducing this class.

sschuberth: > So, can you tell which way of injecting a directory from the test would be acceptable for you? (Ideally along with an argumentation why)

If there was a rule that inherits from `OrtResultRule` and that takes a `projectSourceTree` as a constructor argument, wouldn't that work to run the tests on that new rule, passing in "fake" sources to run the test on?

> Note that I understood from previous talks that you didn't like the construct, however I didn't get why

My main arguments for not passing a `projectSourceTree` to the `Evaluator`'s constructor are:
- Not every rule needs this, and evaluating it / cloning the project source code (accidentally) is expensive.
- The information for a `SourceTree` can be derived from the `ortResult` which already gets passed. IMO there is no need (and it's bad design) to extract some more or less arbitrary bit of information from the `ortResult` and pass a class building upon it independently.

fviernau: > Passing a `SourceTree` to the constructor of `OrtResultRule2` would be of course fine! Doesn't that work?

Hm, I see. In the previous versions of my change it was a problem that this `SourceTreeResolver` was part of the user facing API. Did I just misunderstand that? (No worries, I'll change the code now to pass it into the constructor)

> I believe the proposal I had mentioned in our call earlier this week was ProjectSourceRule.

Let me just rename it quickly.

evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -0,0 +1,40 @@
...
+    name: String,
+    projectSourceTreeResolver: SourceTreeResolver = SourceTreeResolver.forRemoteRepository(
+        ruleSet.ortResult.repository.vcsProcessed
+    )
+) : OrtResultRule(ruleSet, name) {
+    /**
+     * The directory containing the source code of the project. Accessing the property for the first time triggers a
+     * clone and may take a while.
+     */
+    @Suppress("UnusedPrivateMember")
```
sschuberth: This should be `@Suppress("unused")` instead. But why are you introducing the property in this commit already if it's unused?

evaluator/src/main/kotlin/SourceTreeResolver.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.ossreviewtoolkit.downloader.Downloader
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.utils.ort.createOrtTempDir
+
+fun interface SourceTreeResolver {
+    fun getRootDir(): File
+
+    companion object : Logging {
```
sschuberth, line 35: Nit: Static content like `companion object`s should go first.

evaluator/src/main/kotlin/SourceTreeResolver.kt
```diff
@@ -0,0 +1,50 @@
...
+        fun forRemoteRepository(vcsInfo: VcsInfo) = SourceTreeResolver {
```
sschuberth, line 36: Formatting: The function body should start in the next line.

evaluator/src/main/kotlin/SourceTreeResolver.kt
```diff
@@ -0,0 +1,50 @@
...
+                "Downloading ${vcsInfo.type} source code repository: '${vcsInfo.url}'..."
+            }
+
+            val downloadDir = createOrtTempDir()
+            val downloader = Downloader(DownloaderConfiguration())
+            val pkg = Package.EMPTY.copy(vcsProcessed = vcsInfo)
+
+            downloader.download(pkg, downloadDir)
+
+            downloadDir
```
sschuberth, line 47: Maybe wrap this like

            createOrtTempDir().also { downloadDir ->
                val downloader = Downloader(DownloaderConfiguration())
                val pkg = Package.EMPTY.copy(vcsProcessed = vcsInfo)

                downloader.download(pkg, downloadDir)
            }

evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -0,0 +1,40 @@
...
+    val projectSourcesDir: File by lazy { projectSourceTreeResolver.getRootDir() }
```
sschuberth, line 40: The explicit type declaration could be dropped.

evaluator/src/main/kotlin/SourceTreeResolver.kt
```diff
@@ -46,5 +46,7 @@ fun interface SourceTreeResolver {
 
             downloadDir
         }
+
+        fun forLocalDir(dir: File) = SourceTreeResolver { dir }
```
sschuberth, line 50: Nit: As you use "Repository" instead of "Repo" in the function name above, maybe also spell out "Directory" as part of the function name here?

evaluator/src/main/kotlin/ProjectSourceRule.kt
sschuberth: BTW, even with the later commits, it looks like this property indeed can be private.

evaluator/src/main/kotlin/SourceTreeResolver.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.utils.ort.createOrtTempDir
+
+fun interface SourceTreeResolver {
+    fun getRootDir(): File
+
+    companion object : Logging {
+        fun forRemoteRepository(vcsInfo: VcsInfo) = SourceTreeResolver {
+            logger.info {
+                "Downloading ${vcsInfo.type} source code repository: '${vcsInfo.url}'..."
```
sschuberth, line 38: Nit: I'd replace the ":" with "from" and not quote the URL to align with our logging style.

sschuberth (dismissed):

evaluator/src/main/kotlin/ProjectSourceRule.kt
fviernau: The property is public intentionally, so that user can used it to write custom matchers.

evaluator/src/main/kotlin/SourceTreeResolver.kt
fviernau: I'll keep it as is.

evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * A [OrtResultRule] which allows downloading the project's source code if needed.
+ */
+open class ProjectSourceRule(
+    ruleSet: RuleSet,
+    name: String,
+    projectSourceTreeResolver: SourceTreeResolver = SourceTreeResolver.forRemoteRepository(
+        ruleSet.ortResult.repository.vcsProcessed
+    )
```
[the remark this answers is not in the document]
fviernau: As this is public, I kept the more specific name and solved the mentioned issue in an alternative way.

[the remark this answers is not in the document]
sschuberth: > As this is public, I kept the more specific name

I'm not getting that argument. `projectSourceTreeResolver` is a constructor argument, not a public property. And it's in newly added code. So renaming `projectSourceTreeResolver` to `resolver` would not break the experience for any user.

> and solved the mentioned issue in an alternative way.

Personally, I don't like the "trick" with the private extension function. IMO this additional code makes it harder to read, as you need to look at more places. The idea was to reduce the code / use fewer "characters", not more.

sschuberth:
As the rules expert, I'd like @mnonnenmacher to approve.

evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import org.ossreviewtoolkit.model.OrtResult
+
+/**
+ * An [OrtResultRule] which allows downloading the project's source code if needed.
+ */
+open class ProjectSourceRule(
+    ruleSet: RuleSet,
+    name: String,
+    projectSourceTreeResolver: SourceTreeResolver = ruleSet.ortResult.createProjectSourceTreeResolver()
```
mnonnenmacher, line 33: What about `sourceTreeResolver` (like the type) or `projectSourceRule` (like the rule class) to make it shorter?

evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -0,0 +1,42 @@
...
+) : OrtResultRule(ruleSet, name) {
+    /**
+     * The directory containing the source code of the project. Accessing the property for the first time triggers a
+     * clone and may take a while.
+     */
+    val projectSourcesDir: File by lazy { projectSourceTreeResolver.getRootDir() }
+}
+
+private fun OrtResult.createProjectSourceTreeResolver() =
```
mnonnenmacher, line 60: Just `createSourceTreeResolver` would be fine IMO in this context.

evaluator/src/main/kotlin/SourceTreeResolver.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.ossreviewtoolkit.model.config.DownloaderConfiguration
+import org.ossreviewtoolkit.utils.ort.createOrtTempDir
+
+fun interface SourceTreeResolver {
+    companion object : Logging {
+        fun forRemoteRepository(vcsInfo: VcsInfo) =
+            SourceTreeResolver {
+                logger.info {
+                    "Downloading ${vcsInfo.type} source code repository from ${vcsInfo.url}..."
+                }
```
mnonnenmacher, line 38: The `Downloader` already prints a very similar log:
https://github.com/oss-review-toolkit/ort/blob/2d27a2244f660e43b192058221805721033aa7ce/downloader/src/main/kotlin/Downloader.kt#L187-L189

evaluator/src/main/kotlin/SourceTreeResolver.kt
```diff
@@ -0,0 +1,51 @@
...
+
+                val downloadDir = createOrtTempDir()
+                val downloader = Downloader(DownloaderConfiguration())
```
mnonnenmacher, line 41: You should set `DownloaderConfiguration.sourceCodeOrigin = listOf(SourceCodeOrigin.VCS)` to be explicit, otherwise if the download fails there will be additional logs about the failed source artifact download which could be confusing as projects never have a source artifact.

evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -36,6 +37,24 @@ open class ProjectSourceRule(
      * clone and may take a while.
      */
     val projectSourcesDir: File by lazy { projectSourceTreeResolver.getRootDir() }
+
+    /**
+     * Return all files from the project's source tree which match any of the provided [glob expressions][patterns].
+     */
+    fun projectSourceFindFiles(vararg patterns: String): List<File> =
```
mnonnenmacher, line 44: Is this function supposed to be public?

evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
```diff
@@ -0,0 +1,93 @@
...
+
+class ProjectSourceRuleTest : WordSpec({
+    "projectSourceHasFile()" should {
+        "return true if at least one file matches the given glob pattern" {
+            val dir = createSpecTempDir().apply {
+                addFiles(
+                    "README.md",
+                    "module/docs/LICENSE.txt"
+                )
+            }
```
mnonnenmacher, line 38: Nit: It would actually be shorter and less nested to just write:
```kotlin
val dir = createSpecTempDir()
dir.addFiles(
  "README.md",
   "module/docs/LICENSE.txt"
)
```

Or if `addFiles` would return the directory it could just be:
```kotlin
val dir = createSpecTempDir().addFiles(
  "README.md",
  "module/docs/LICENSE.txt"
)
```

(same for `addDirs`)

evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -36,6 +37,24 @@ open class ProjectSourceRule(
...
+        projectSourcesDir.walkBottomUp().filterTo(mutableListOf()) {
+            it.isFile && FileMatcher.match(patterns.toList(), it.relativeTo(projectSourcesDir).path)
+        }
+
+    /**
+     * A [RuleMatcher] that checks whether the project's source tree contains at least one directory matching any of the
```
mnonnenmacher, line 50: This should say "at least one **file**".

evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
```diff
@@ -0,0 +1,93 @@
...
+            rule.projectSourceHasFile("README.md").matches() shouldBe false
+        }
+    }
+})
+
+private fun createOrtResultRule(projectSourcesDir: File): ProjectSourceRule =
+    ProjectSourceRule(
+        ruleSet = ruleSet(
+            ortResult = OrtResult.EMPTY
+        ),
```
mnonnenmacher, line 71: Nit: Could be a one-liner.

mnonnenmacher (changes requested):

evaluator/src/main/kotlin/ProjectSourceRule.kt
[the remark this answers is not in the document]
fviernau: > I'm not getting that argument. `projectSourceTreeResolver` is a constructor argument, not a public property. And it's in newly added code. So renaming `projectSourceTreeResolver` to `resolver` would not break the experience for any user.

My bad, I accidentally thought it was a public property as it was one before the refactoring requested by you.

> Personally, I don't like the "trick" with the private extension function. IMO this additional code makes it harder to read, as you need to look at more places. The idea was to reduce the code / use fewer "characters", not more.

evaluator/src/main/kotlin/SourceTreeResolver.kt
fviernau: I thought it was good to at least print one log statement when doing something exotic like downloading the sources from the evaluator. If you agree, maybe we can re-phrase the log comment like e.g. 

"The evaluator rules require the project source code. Cloning $url."

What do you think?

evaluator/src/main/kotlin/ProjectSourceRule.kt
fviernau: Yes: The use cases are

1. Use it from the policy rule directly (instead of the matcher)
2. Write custom matchers, re-using logic from ORT. Not that any code implemented in `Matcher.matches()` is 
    not really re-usable, so I minimize it. 

For 1, note that when implementing a policy rule often times using the matchers with their binary result is ok. But often times when one wants to write good error messages or how-to-fix, the binary result is not sufficient anymore.

evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
fviernau: Are you ok with keeping this as is. In a following PR further tests will be added which call both `addFiles()` and `addDirs()`, so it will be consistent.

If you still prefer to change it, can I do this in a follow up PR, to avoid the rebase hazzle with my following changes?

mnonnenmacher: Fleunt API could still be useful but I'll leave that up to you.

fviernau: I'll keep it in mind and probably implement it in a following PR.

mnonnenmacher (approved):

PR #5757, opened 2022-09-07
Json schema ort config

MarcelBochtler:
Add schema that validates and enables autocompletion for well typed
configuration options. However loosely typed configuration options like
the analyzer's `packageMangers` options which require a `Map`, are
currently omitted by allowing `"additionalProperties": { "type":
"object" }`, which results in allowing all keys and values in the
validation.

### DIFF
--- a/integrations/schemas/ort-configuration-schema.json
+++ b/integrations/schemas/ort-configuration-schema.json
@@ -1,0 +1,579 @@
+{
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "$id": "https://oss-review-toolkit.org/ort-configuration.yml",
+    "title": "ORT configuration",
+    "description": "The main configuration file for the OSS-Review-Toolkit (ORT). A full list of all available options can be found at https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/resources/reference.yml.",
+    "type": "object",
+    "properties": {
+        "ort": {
+            "properties": {
+                "licenseFilePatterns": {
+                    "$ref": "#/definitions/LicenseFilePatterns"
+                },
+                "severeIssueThreshold": {
+                    "$ref": "#/definitions/Severity"
+                },
+                "severeRuleViolationThreshold": {
+                    "$ref": "#/definitions/Severity"
+                },
+                "enableRepositoryPackageCurations": {
+                    "type": "boolean"
+                },
+                "enableRepositoryPackageConfigurations": {
+                    "type": "boolean"
+                },
+                "analyzer": {
+                    "$ref": "#/definitions/Analyzer"
+                },
+                "advisor": {
+                    "$ref": "#/definitions/Advisor"
+                },
+                "downloader": {
+                    "$ref": "#/definitions/Downloader"
+                },
+                "scanner": {
+                    "$ref": "#/definitions/Scanner"
+                },
+                "reporter": {
+                    "$ref": "#/definitions/Reporter"
+                },
+                "notifier": {
+                    "$ref": "#/definitions/Notifier"
+                }
+            }
+        }
+    },
+    "required": [
+        "ort"
+    ],
+    "definitions": {
+        "Advisor": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "nexusIq": {
+                    "$ref": "#/definitions/NexusIq"
+                },
+                "vulnerableCode": {
+                    "$ref": "#/definitions/VulnerableCode"
+                },
+                "gitHubDefects": {
+                    "$ref": "#/definitions/GitHubDefects"
+                },
+                "osv": {
+                    "$ref": "#/definitions/Osv"
+                },
+                "options": {
+                    "$ref": "#/definitions/AdvisorOptions"
+                }
+            }
+        },
+        "GitHubDefects": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "token": {
+                    "type": "string"
+                },
+                "labelFilter": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "maxNumberOfIssuesPerRepository": {
+                    "type": "integer"
+                },
+                "parallelRequests": {
+                    "type": "integer"
+                }
+            },
+            "required": [
+                "labelFilter"
+            ]
+        },
+        "NexusIq": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "serverUrl": {
+                    "type": "string",
+                    "format": "uri"
+                },
+                "browseUrl": {
+                    "type": "string",
+                    "format": "uri"
+                },
+                "username": {
+                    "type": "string"
+                },
+                "password": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "serverUrl"
+            ],
+            "title": "NexusIq"
+        },
+        "AdvisorOptions": {
+            "type": "object",
+            "additionalProperties": { "type": "object" }
+        },
+        "CustomAdvisor": {
+            "type": "object",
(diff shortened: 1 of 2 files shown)


sschuberth:
Nit: Both commit message titles lack an "a" after "Add".

integrations/schemas/ort-configuration-schema.json
```diff
@@ -0,0 +1,572 @@
+{
```
nnobelis, line 1: Commit typo `packageManAgers`.

model/src/test/kotlin/JsonSchemaTest.kt
```diff
@@ -99,6 +99,19 @@ class JsonSchemaTest : StringSpec() {
...
+
+        "reference.yml validates successfully" {
+            val ortConfigurationSchema = JsonSchemaFactory
+                .builder(JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V6))
+                .objectMapper(mapper)
+                .build()
+                .getSchema(File("../integrations/schemas/ort-configuration-schema.json").toURI())
+            val referenceConfigFile = File("src/main/resources/reference.yml").toJsonNode()
+
+            val errors = ortConfigurationSchema.validate(referenceConfigFile)
```
nnobelis, line 111: I would inline `errors`

model/src/test/kotlin/JsonSchemaTest.kt
```diff
@@ -99,6 +99,19 @@ class JsonSchemaTest : StringSpec() {
```
nnobelis, line 99: Commit message nit: quote ` the file names.

model/src/test/kotlin/JsonSchemaTest.kt
MarcelBochtler: I try to maintain the `given, when, then` structure, which requires the execution to be split from the verification.
It also follows the style of all other tests in this file.

model/src/test/kotlin/JsonSchemaTest.kt
```diff
@@ -99,6 +99,19 @@ class JsonSchemaTest : StringSpec() {
 
             errors should beEmpty()
         }
+
+        "reference.yml validates successfully" {
+            val ortConfigurationSchema = JsonSchemaFactory
+                .builder(JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V6))
```
sschuberth, line 105: Why is this `V6` instead of `V7` like the others?

sschuberth: Which boils down why we use `"$schema": "http://json-schema.org/draft-06/schema#"` in the previous commit.

MarcelBochtler: The other schemas use features which are only available in `V7` (if-then-else: https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else).
As the other schemas are split and reuse common components, all of them have to be `V7`.

In this case, no features from `V7` are (yet) used, and therefore I did use `V6`, as per the [best practices](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#best-practices) the "lowest possible version should be used to ensure interoperability with as many supported editors, IDEs and parsers as possible."

But thanks to you for pointing this out, as we actually support `V4` which I now switched to.

sschuberth: > as per the [best practices](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#best-practices) the "lowest possible version should be used

Makes sense. Is it worth to mention this in the commit message?

MarcelBochtler:
Added validation for the analyzer's `packageManager` configuration. Now, only supported package managers can be configured:
image

MarcelBochtler:
Split the schema into `ort-configuration-schema.json` and `package-managers-schema.json`. That latter will also be added to be used in the `repository-configuration-schema.json`.
This also means requiring the schema version `draft-07`.

MarcelBochtler:
I plan to also contribute this schema to [schemastore](https://github.com/SchemaStore/schemastore) to get automatic validation and autocompletion in some editors like for the already [existing schemas](https://github.com/SchemaStore/schemastore/blob/b5ca142cd05a8d61208d621a5c6edffb60b37341/src/api/json/catalog.json#L2215-L2238).

The file name should be unique for all supported schemas within the schemastore.
IMO, `config.yml` is a too generic name for this.
My plan is to use the default location as `fileMatch` which would be `.ort/config/config.yml`. This would mean that the autocompletion and validation will only work for this directory out-of-the-box.
`config.yml` files from a different location will require manual configuration.

Any alternative suggestions how to deal with this?

mnonnenmacher:
> I plan to also contribute this schema to [schemastore](https://github.com/SchemaStore/schemastore) to get automatic validation and autocompletion in some editors like for the already [existing schemas](https://github.com/SchemaStore/schemastore/blob/b5ca142cd05a8d61208d621a5c6edffb60b37341/src/api/json/catalog.json#L2215-L2238).
> 
> The file name should be unique for all supported schemas within the schemastore. IMO, `config.yml` is a too generic name for this. My plan is to use the default location as `fileMatch` which would be `.ort/config/config.yml`. This would mean that the autocompletion and validation will only work for this directory out-of-the-box. `config.yml` files from a different location will require manual configuration.
> 
> Any alternative suggestions how to deal with this?

I don't have a good idea, but maybe just `**/config/config.yml` is specific enough? It looks like other schemas also use pretty generic names.

How difficult is it to manually select the schema, for example when editing `reference.yml` in VS Code?

MarcelBochtler:
> How difficult is it to manually select the schema, for example when editing `reference.yml` in VS Code?

During testing in VS Code, I use something like this at the beginning of a file: 
```
# yaml-language-server: $schema=https://raw.githubusercontent.com/bosch-io/oss-review-toolkit/json-schema-ort-config/integrations/schemas/ort-configuration-schema.json
```
But this is VS Code specific.
It can also be configured in the settings of the `yaml-language-server` plugin.
(This will only work after this PR is merged, as the references point to `oss-review-toolkit/ort/main`.)

In IntelliJ it can also be configured in the settings.

However, this defeats the purpose of using `schemastore` as it should provide validation / autocompletion out of the box in all supported editors.
I think we have two options:
1. I could try to contribute it as `config.yml` as this is not yet taken. I also don't know of any guideline to use very specific file names.
2. I use the default location `.ort/config/config.yml`.

I'd suggest the second one, as the configuration file name is configurable by the ORT user, and something like `ort-config.yml` won't be recognized by the editors anyway.

We can also move this discussion to the PR in `schemastore` when I created it.

sschuberth:
> We could also consider renaming the default file to `.ort/config/ort-config.yml`, it's a bit redundant but more unique.
> @oss-review-toolkit/core-devs Opinions?

I don't like the redundancy, but if it's the only way to be able to contribute the schema, I'm fine with it.

MarcelBochtler:
> I don't like the redundancy, but if it's the only way to be able to contribute the schema, I'm fine with it.

It is not the only way. `**/.ort/config/config.yml` would work perfectly fine.
The autocompletion would then just no longer work if the user uses the `-c` flag to point to a separate configuration file.
And in this case, the filename isn't guaranteed to be `config.yml` anyway.

Therefore, renaming it to something like `ort-config.yml` would only provide _some_ benefit, but not solve all issues.

Can this PR btw. be merged? As this discussion is not really relevant for it.

sschuberth:
> Therefore, renaming it to something like `ort-config.yml` would only provide _some_ benefit, but not solve all issues.

Ok, then I'd *not* do the renaming.

> Can this PR btw. be merged? As this discussion is not really relevant for it.

I'm approving.

sschuberth (approved):

PR #5760, opened 2022-09-08
Docker: Component based Multi stage docker

heliocastro:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,9 @@ out/
 
 # Mac
 .DS_Store
+
+# Test generated
+**/node_modules
+**/lockfile
+**/bower_components
+.pnpm-store
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@
 # Copyright (C) 2020 Bosch Software Innovations GmbH
 # Copyright (C) 2021-2022 Bosch.IO GmbH
 # Copyright (C) 2021 Alliander N.V.
+# Copyright (C) 2021-2022 BMW CarIT GmbH
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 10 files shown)


tsteenbe:
@heliocastro  As discussed on ORT Slack made some commit message improvement such as spelling, grammar, etc. Believe we are now good to go.

sschuberth:
> Believe we are now good to go.

Sorry, still "no" from my side. Many of @mnonnenmacher's [very good points](https://github.com/oss-review-toolkit/ort/pull/4746#issuecomment-1241122895) still hold with this PR. Even with the reduced number of commits, unrelated stuff if done in  single commit, formatting does not adhere to our standards, etc.

I believe the way forward would be as @mnonnenmacher described:

> * One commit to copy the current Dockerfile to the legacy directory.
> 
> * A few commits that apply the relevant changes on top of the current Dockerfile, but without all the version updates that just repeat version updates which already happened in our current Dockerfile. Or even squashing all changes into a single commit would be better than the current state of the PR.
> 
> * One commit for each version update that changes a version compared to our current Dockerfile.

Dockerfile
```diff
@@ -19,41 +20,112 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
-# Set this to the version ORT should report.
-ARG ORT_VERSION="DOCKER-SNAPSHOT"
+#------------------------------------------------------------------------
+# Use OpenJDK Eclipe Temurin Ubuntu LTS
+FROM eclipse-temurin:11-jdk AS build
```
sschuberth: We've already used `eclipse-temurin:11-jdk-jammy` before, so "Base main Docker on OpenJDK 11 Eclipse Temurin" in the commit message is misleading.

Dockerfile
```diff
@@ -153,6 +153,23 @@ RUN curl -ksSL https://github.com/CocoaPods/CocoaPods/archive/9461b346aeb8cba6df
 
 COPY docker/ruby.sh /etc/ort/bash_modules
 
+#------------------------------------------------------------------------
+# SCANCODE - Build scancode as a separate component
+FROM pythonbuild AS scancodebuild
+
+ARG SCANCODE_VERSION="3.2.1rc2"
```
sschuberth: This is a step backwards. We use version 30.1.0 already in the current `Dockerfile`.

Dockerfile
```diff
@@ -257,15 +279,7 @@ RUN /opt/ort/bin/import_proxy_certs.sh && \
...
-        # Trigger ScanCode configuration for Python 3 and reindex licenses initially.
-        cd /usr/local/scancode-toolkit-$SCANCODE_VERSION && \
-        PYTHON_EXE=/usr/bin/python3 /usr/local/scancode-toolkit-$SCANCODE_VERSION/scancode --reindex-licenses && \
-        chmod -R o=u /usr/local/scancode-toolkit-$SCANCODE_VERSION && \
-        ln -s /usr/local/scancode-toolkit-$SCANCODE_VERSION/scancode /usr/local/bin/scancode
-
+    yes | $ANDROID_HOME/cmdline-tools/bin/sdkmanager $SDK_MANAGER_PROXY_OPTIONS --sdk_root=$ANDROID_HOME "platform-tools"
 
 ENTRYPOINT ["/usr/bin/ort"]
+
```
sschuberth: Superfluous change.

Dockerfile
```diff
@@ -67,8 +139,8 @@ ENV \
     # SDK versions.
     ANDROID_SDK_VERSION=6858069 \
     # Installation directories.
-    ANDROID_HOME=/opt/android-sdk \
-    GOPATH=/tmp/go
+    ANDROID_HOME=/opt/android-sDockerfiledk \
```
sschuberth: Typo in the path. This could have never worked.

.dockerignore
```diff
@@ -1,6 +1,5 @@
 # Ignore the Docker files themselves to avoid that all layers are invalidated if
 # the Docker files are changed during development.
-.dockerignore
```
sschuberth, line 3: This should not be removed. See the comment above.

Dockerfile
```diff
@@ -34,67 +34,77 @@ ARG CRT_FILES=""
 COPY "$CRT_FILES" /tmp/certificates/
 RUN /etc/scripts/import_proxy_certs.sh \
     && if [ -n "$CRT_FILES" ]; then \
-        /etc/scripts/import_certificates.sh /tmp/certificates/; \
-       fi
+    /etc/scripts/import_certificates.sh /tmp/certificates/; \
```
sschuberth, line 37: Indentation is still broken after this change.

Dockerfile
```diff
@@ -34,67 +34,77 @@ ARG CRT_FILES=""
...
-        libssl-dev \
-        make \
-        netbase \
-        tk-dev \
-        tzdata \
-        uuid-dev \
-        unzip \
-        xz-utils \
-        zlib1g-dev \
+    build-essential \
```
sschuberth, line 46: Broken indentation.

.gitignore
```diff
@@ -12,3 +12,7 @@ out/
 
 # Mac
 .DS_Store
+
+# Ignore node_modules
```
sschuberth, line 16: These changes seem unrelated to anything mentioned in the commit message.

scripts/docker_build.sh
```diff
@@ -23,10 +23,12 @@
 GIT_ROOT=$(git rev-parse --show-toplevel)
 GIT_VERSION=$(git describe --abbrev=10 --always --tags --dirty)
 ORT_DOCKER=${ORT_DOCKER:-ort}
+ARCH_TRANSLATION=""
 
 echo "Setting ORT_VERSION to $GIT_VERSION."
-DOCKER_BUILDKIT=1 docker build -f "$GIT_ROOT"/Dockerfile \
+docker buildx build -f "$GIT_ROOT"/Dockerfile \
```
sschuberth, line 29: Switching to `buildx` is a rather big under-the-hood change that deserves its own commit with a thorough explanation.

sschuberth (changes requested):
Sorry, still way too messy.

Dockerfile
tsteenbe: No longer applicable as commit was squashed.

docker/legacy/Dockerfile
```diff
@@ -0,0 +1,177 @@
+# syntax=docker/dockerfile:1.3
```
sschuberth, line 1: Commit message nits:

- "to new multi stage Dockerfile" lacks a "to *the* new" or "to *a* new".
- please don't emphasize "multi stage"; that's confusing because the legacy `Dockerfile` also is multi-stage already.

I propose: "This is a preparation for allowing ORT users to stick to the current Dockerfile instead of having to directly upgrade to the upcoming rewritten Dockerfile."

scripts/docker_build.sh
```diff
@@ -20,5 +21,10 @@
 GIT_ROOT=$(git rev-parse --show-toplevel)
 GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty)
 
-echo "Setting ORT_VERSION to $GIT_REVISION."
-DOCKER_BUILDKIT=1 docker build -f $GIT_ROOT/Dockerfile -t ort --build-arg ORT_VERSION=$GIT_REVISION $GIT_ROOT
+echo "Setting ORT_VERSION to $GIT_VERSION."
```
sschuberth, line 28: This should still say `GIT_REVISION`.

scripts/docker_build.sh
```diff
@@ -20,5 +21,10 @@
...
+docker buildx build -f "$GIT_ROOT"/Dockerfile \
```
sschuberth, line 29: Nit: I'd just quote all of the path for clarity.

scripts/docker_build.sh
```diff
@@ -20,5 +21,10 @@
...
+    -t "$ORT_DOCKER" \
```
sschuberth, line 30: There is no `ORT_DOCKER` variable; so this should probably stay at "ort".

scripts/docker_build.sh
```diff
@@ -20,5 +21,10 @@
...
+    --build-arg ORT_VERSION="$GIT_VERSION" \
```
sschuberth, line 31: Again, this should still say GIT_REVISION.

scripts/docker_build.sh
```diff
@@ -20,5 +21,10 @@
...
+    --platform linux/amd64 \
+    "$@" \
+    "$GIT_ROOT"
```
sschuberth, line 34: Ideally, all the quoting an wrapping that you do here should go to separate commits, so it becomes apparent that basically the only thing that this commit does is to add "buildx" and "--platform linux/amd64".

scripts/docker_build.sh
```diff
@@ -20,5 +21,10 @@
...
 GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty)
 
-echo "Setting ORT_VERSION to $GIT_REVISION."
-DOCKER_BUILDKIT=1 docker build -f $GIT_ROOT/Dockerfile -t ort --build-arg ORT_VERSION=$GIT_REVISION $GIT_ROOT
+echo "Setting ORT_VERSION to $GIT_VERSION."
+docker buildx build -f "$GIT_ROOT"/Dockerfile \
+    -t "$ORT_DOCKER" \
+    --build-arg ORT_VERSION="$GIT_VERSION" \
+    --platform linux/amd64 \
+    "$@" \
```
sschuberth, line 33: This is new. Why is it needed?

scripts/docker_build.sh
```diff
@@ -1,6 +1,7 @@
-#!/bin/sh
+#!/bin/bash
```
sschuberth, line 1: What Bash feature is being used now?

docker/00-add_local_path.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
sschuberth, line 1: Commit message: Again, please don't emphasize "multi stage". It has been multi-stage before.

scripts/docker_build.sh
```diff
@@ -18,8 +18,12 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
+# Docker from ORT uses BuildKit feature
+
 GIT_ROOT=$(git rev-parse --show-toplevel)
-GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty)
+GIT_VERSION=$(git describe --abbrev=10 --always --tags --dirty)
+ORT_DOCKER=${ORT_DOCKER:-ort}
+ARCH_TRANSLATION=""
```
sschuberth, line 26: These changes (renaming the `GIT_REVISION` variable, making the Docker tag to use configurable, etc.) are unrelated to the topic described in the commit message.

sschuberth (changes requested):

scripts/docker_build.sh
heliocastro: If someone pass ORT_DOCKER=ort_my_custom the container will be created  as ort_my_custom, otherwise will be used as ort

scripts/docker_build.sh
heliocastro: Build on Arm
Ask @tsteenbe

scripts/docker_build.sh
heliocastro: So let's say you want a custom image and want specia version of python and want to keep the pristine docker image ?
`PYTHON_VERSION=3.9.9 ORT_DOCKER=ort_python399 ./scripts/docker_build.sh`

scripts/docker_build.sh
tsteenbe: The `--platform linux/amd64`  forces Docker to use linux/amd64 platform by default on Apple Silicon.

batect.yml
```diff
@@ -94,4 +94,5 @@ tasks:
       - buildFunTest
     run:
       container: run
-      command: ./gradlew funTest
+      command: "'./gradlew funTest jacocoFunTestReport -p analyzer'"
```
sschuberth, line 97: Can you please explain what this does / why it is necessary? The commit message unfortunately does not give me any clue. In particular:

- Why are JaCoCo reports being created now?
- Why is running the tests now limited to functional tests in the analyzer?

tsteenbe: @heliocastro removed `jacocoFunTestReport -p analyzer` but now there is below error.

@sschuberth  Any clue what is causing this?

```

org.ossreviewtoolkit.model.utils.PostgresFileArchiverStorageFunTest > Before Spec Error FAILED
    java.lang.IllegalStateException: Could not find a valid Docker environment. Please see logs and check configuration
        at org.testcontainers.dockerclient.DockerClientProviderStrategy.lambda$getFirstValidStrategy$4(DockerClientProviderStrategy.java:156)
        at java.base/java.util.Optional.orElseThrow(Optional.java:408)
        at org.testcontainers.dockerclient.DockerClientProviderStrategy.getFirstValidStrategy(DockerClientProviderStrategy.java:148)
        at org.testcontainers.DockerClientFactory.getOrInitializeStrategy(DockerClientFactory.java:146)
        at org.testcontainers.DockerClientFactory.client(DockerClientFactory.java:188)
        at org.testcontainers.DockerClientFactory$1.getDockerClient(DockerClientFactory.java:101)
        at com.github.dockerjava.api.DockerClientDelegate.authConfig(DockerClientDelegate.java:107)
        at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:316)
        at com.opentable.db.postgres.embedded.EmbeddedPostgres.<init>(EmbeddedPostgres.java:99)
        at com.opentable.db.postgres.embedded.EmbeddedPostgres$Builder.start(EmbeddedPostgres.java:353)
        at org.ossreviewtoolkit.utils.test.PostgresListener.beforeSpec(PostgresListener.kt:42)
        at io.kotest.engine.spec.SpecExtensions.beforeSpec-gIAlu-s(SpecExtensions.kt:49)
        at io.kotest.engine
```

fviernau:
I've reviewed this again (not everything in most possible detail), haven't found any blockers.
I've compared the tool versions and the upgrades now  do happen in separate commits, e.g. the large commit should keep the versions identical. The names of some variables change, like e.g. `ANDROID_SDK_VERSION` -> `ANDROID_CMD_VERSION`, but in this case I believe this isn't an issue at all.

batect.yml
sschuberth: I don't have a concrete idea, but it looks like some of the changes broke running Docker-in-Docker, as our `PostgresFileArchiverStorageFunTest` relies on `org.testcontainers`.

sschuberth:
> I've reviewed this again (not everything in most possible detail), haven't found any blockers.

To me the last commit is still problematic as it touches the `batect.yml` file, which is exactly the bit I'd expect *not* to get touched in order to prove that the new `Dockerfile` does not break any test run via Batect WRT the current `Dockerfile`. Also, to me the commit message is not clear enough about what the change is actually needed for.

tsteenbe (approved):

scripts/docker_build.sh
sschuberth: Why was this resolved while being unanswered?

scripts/docker_build.sh
sschuberth: Still broken.

sschuberth (changes requested):

scripts/docker_build.sh
heliocastro: Is already fixed.

tsteenbe:
@sschuberth "Why was this resolved while being unanswered?" The current version of docker_build.sh is using  "#!/bin/sh" and not "#!/bin/bash" - you looking at outdated version for your code review I guess.

heliocastro:
> > I've reviewed this again (not everything in most possible detail), haven't found any blockers.
> 
> To me the last commit is still problematic as it touches the `batect.yml` file, which is exactly the bit I'd expect _not_ to get touched in order to prove that the new `Dockerfile` does not break any test run via Batect WRT the current `Dockerfile`. Also, to me the commit message is not clear enough about what the change is actually needed for.

Container has several tool version managers that install base is not on standard PATH.
To made this available, was necessary add bash profiles to made visible to Ort. Profiles are only parsed if you run in login environment, and container by default not do that.

For Ort binary we use a wrapper where it calls bash -l so pushing in login environment

In batect, the case is that we calling gradle, not ort wrapper, meaning not bash  login shell, though failing to access tooling.
As documentation of batect https://batect.dev/docs/reference/config/containers/#entrypoint, this is correct way to execute a command from entry point shell.

As the jacoco move to container call, before was working due be executed as root user, and ignoring the fact that the second call passed was running as github actions user.
With the new container running things as a regular user **ort**, the process ends up as a ghaction user, which is different from **ort** user trying to write in a non permissive space

So moving the jacoco to inside container execution and not outside produce the correct results, as entire build and test are placed in same environment, same user and non privileged user.

scripts/docker_build.sh
sschuberth: If so, then not in the correct commit. See how the current https://github.com/oss-review-toolkit/ort/pull/5760/commits/02787e58c054c34fe1ca07f2017131e0c2347212 still changes this line.

sschuberth: Yeah, you're changing it back only in https://github.com/oss-review-toolkit/ort/pull/5760/commits/07c68b664a7d2b9b1d7c367e4812feadafe8d677. What a mess...

tsteenbe: I have fixed it now the "docker_build: Enable builds on Apple Silicon" no longer changes line 1.

Dockerfile
```diff
@@ -19,159 +20,335 @@
...
+COPY --from=rubybuild /etc/profile.d/ruby.sh /etc/profile.d/
+RUN chmod o+rwx ${RBENV_ROOT}
+
+# NodeJS
+ARG NODEJS_VERSION=16.17.0
+ARG NVM_DIR=/opt/nvm
+ENV NODE_PATH $NVM_DIR/v$NODEJS_VERSION/lib/node_modules
+ENV PATH $NVM_DIR/versions/node/v$NODEJS_VERSION/bin:$PATH
+COPY --chown=$USERNAME:$USERNAME --from=nodebuild ${NVM_DIR} ${NVM_DIR}
+RUN chmod o+rwx ${RBENV_ROOT}
```
nnobelis, line 298: This should be `NVM_DIR`.

nnobelis (changes requested):

Dockerfile
heliocastro: Fixed

sschuberth:
Superseded by https://github.com/oss-review-toolkit/ort/pull/5830.

PR #5768, opened 2022-09-10
model: Remove annotations for backward compatibility

mnonnenmacher:
Remove most annotations that were introduced for backward compatibility when renaming or removing properties to reduce technical debt. The introduction of those annotations was long enough ago, so that we can assume that it is not required anymore that ORT can deserialize such old files.


model/src/test/assets/analyzer-result-with-dependency-graph.yml
```diff
@@ -144,14 +144,6 @@ analyzer:
           url: "ssh://git@github.com/hamcrest/JavaHamcrest.git"
           revision: ""
           path: ""
-      curations:
```
sschuberth, line 147: I currently don't understand why this is being dropped. Can you help me here?

mnonnenmacher: Declared licenses can only be curated with declared licenses mappings now, since then the property is ignored anyway and because it is unrelated to the license graph test I decided to drop it.

sschuberth (approved):

PR #5769, opened 2022-09-10
ScanController: Add missing documentation for the `issues` map

bs-ondem:
The `issues` map was introduced with 4ba16b6, but was merged without the documentation.

### DIFF
--- a/scanner/src/main/kotlin/experimental/ScanController.kt
+++ b/scanner/src/main/kotlin/experimental/ScanController.kt
@@ -59,7 +59,8 @@ class ScanController(
     private val provenanceResolutionIssues = mutableMapOf<Identifier, MutableList<OrtIssue>>()
 
     /**
-     * TODO: Docs
+     * A map of [Identifier]s associated with a list of [OrtIssue]s that occurred during a scan besides the issues
+     * created by the scanners themselves as part of the [ScanSummary].
      */
     private val issues = mutableMapOf<Identifier, MutableList<OrtIssue>>()
 


scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -59,7 +59,8 @@ class ScanController(
     private val provenanceResolutionIssues = mutableMapOf<Identifier, MutableList<OrtIssue>>()
 
     /**
-     * TODO: Docs
+     * A map of package [Identifier]s to a list of [OrtIssue]s, which can be used for occurring errors apart from the
```
sschuberth, line 62: By "apart" do you mean "independently"?

bs-ondem: Yes, exactly. Should I change it to make it clear?

sschuberth: I'd personally appreciate that, yes, but also then I'm still having slight difficulties understanding the sentence (but maybe that's just me): So is the map there in order to "share" `OrtIssues` between scanners on a per-`Identifier` basis? If that's the case maybe use something like:

> A map of `Identifer`s associated with a list of `OrtIssue`s that occurred during a scan with any scanner.

bs-ondem: Do you mean without any scanners?

sschuberth: No, did not mean "without". I wanted to emphasize "*any* scanner" to explain that a contained issue might stem from any active scanner, not just a single one. I.e. it's a shared list that any scanner can add to. Or am I again misunderstanding something?

bs-ondem: > I.e. it's a shared list that any scanner can add to.

A shared list of `OrtIssue`s. That's exactly what the list is for. However, "_any scanner_" confuses me a bit, because there are more sources of error in the [`scan`](https://github.com/oss-review-toolkit/ort/blob/main/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt#L141-L165) function and not only from "_any scanner_".

sschuberth: Ok, so how about this then?

> A map of [Identifer]s associated with a list of [OrtIssue]s that occurred during a scan besides the issues created by the scanners themselves as part of the [ScanSummary].

Does that nail it?

sschuberth (approved):

PR #5776, opened 2022-09-12
Migrate more Git code to JGit

sschuberth:

### DIFF
--- a/downloader/src/main/kotlin/vcs/Git.kt
+++ b/downloader/src/main/kotlin/vcs/Git.kt
@@ -179,8 +179,8 @@ class Git : VersionControlSystem(), CommandLineTool {
         path: String,
         recursive: Boolean
     ): Result<String> =
-        (workingTree as GitWorkingTree).repo.use { repo ->
-            Git(repo).use { git ->
+        (workingTree as GitWorkingTree).useRepo {
+            Git(this).use { git ->
                 updateWorkingTreeWithoutSubmodules(git, revision).mapCatching {
                     // In case this throws the exception gets encapsulated as a failure.
                     if (recursive) updateSubmodules(workingTree)
--- a/downloader/src/main/kotlin/vcs/Git.kt
+++ b/downloader/src/main/kotlin/vcs/Git.kt
@@ -38,10 +38,12 @@ import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.api.LsRemoteCommand
 import org.eclipse.jgit.api.errors.GitAPIException
 import org.eclipse.jgit.errors.UnsupportedCredentialItem
+import org.eclipse.jgit.lib.Constants
 import org.eclipse.jgit.lib.SymbolicRef
 import org.eclipse.jgit.transport.CredentialItem
 import org.eclipse.jgit.transport.CredentialsProvider
 import org.eclipse.jgit.transport.SshSessionFactory
+import org.eclipse.jgit.transport.TagOpt
 import org.eclipse.jgit.transport.URIish
 import org.eclipse.jgit.transport.ssh.jsch.JschConfigSessionFactory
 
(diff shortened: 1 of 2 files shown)


sschuberth:
I'd appreciate some feedback esp. from @oss-review-toolkit/bosch for testing this against private repos that require authentication. I'd like to be sure to not break anything in that regard.

mnonnenmacher:
@sschuberth Does the sparse-checkout option also influence what is fetched? So could it be that migrating fetch to JGit which does not support sparse checkouts somehow has a performance impact for sparse checkouts?

sschuberth:
> Does the sparse-checkout option also influence what is fetched?

No. Sparse checkout is a purely client-side feature (as opposed to a *shallow* fetch).

sschuberth:
BTW, I've actually verified that this change finally makes the `"Git does not prompt for credentials for non-existing repositories"` test pass on Windows for me, which previously triggered this dialog

image

downloader/src/main/kotlin/vcs/Git.kt
```diff
@@ -38,10 +38,12 @@ import org.eclipse.jgit.api.Git
...
 import org.eclipse.jgit.api.errors.GitAPIException
 import org.eclipse.jgit.errors.UnsupportedCredentialItem
+import org.eclipse.jgit.lib.Constants
 import org.eclipse.jgit.lib.SymbolicRef
 import org.eclipse.jgit.transport.CredentialItem
 import org.eclipse.jgit.transport.CredentialsProvider
 import org.eclipse.jgit.transport.SshSessionFactory
+import org.eclipse.jgit.transport.TagOpt
 import org.eclipse.jgit.transport.URIish
 import org.eclipse.jgit.transport.ssh.jsch.JschConfigSessionFactory
```
fviernau, line 48: commit-msg: "in not required" -> "**is** not required"

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -57,7 +57,7 @@ open class GitWorkingTree(
 ) : WorkingTree(workingDir, vcsType) {
     companion object : Logging
 
-    private val repo: Repository
+    val repo: Repository
```
fviernau, line 60: In the PR which introduced `repo.use` I asked whether it makes sense to extract it to a function.
Now, as you make the property public I think that would indeed make sense. So, instead of exposing `repo` expose a function to run something using `repo`. This would ensure that the contract (to use `use`) does not get violated.

fviernau (changes requested):

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
sschuberth: > I asked whether it makes sense to extract it to a function.
> Now, as you make the property public I think that would indeed make sense.

How would you envision such a function to look like? Should it take a lambda that is applied inside `repo.use`?

fviernau: > How would you envision such a function to look like? Should it take a lambda that is applied inside `repo.use`?

Yes, that was what I had in mind. Does it make sense?

sschuberth: Yes, I agree that makes sense now. I've implemented it in a new commit.

fviernau (approved):

sschuberth:
@mnonnenmacher, do you want to give this some internal testing before merging?

MarcelBochtler (approved):
I tested this branch with multiple authenticated Git repositories, and haven't seen any issues.

sschuberth:
> I tested this branch with multiple authenticated Git repositories, and haven't seen any issues.

Thank you!

PR #5777, opened 2022-09-12
Some fixups for recent changes

sschuberth:

### DIFF
--- a/evaluator/src/main/kotlin/OrtResultRule.kt
+++ b/evaluator/src/main/kotlin/OrtResultRule.kt
@@ -19,6 +19,7 @@
 
 package org.ossreviewtoolkit.evaluator
 
+import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Severity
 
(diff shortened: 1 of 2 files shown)


evaluator/src/main/kotlin/SourceTreeResolver.kt
```diff
@@ -47,7 +47,7 @@ class SourceTreeResolver private constructor(getRepositoryRoot: () -> File) {
         fun forLocalDirectory(dir: File) = SourceTreeResolver { dir }
     }
 
-    val rootDir: File by lazy {
+    val rootDir by lazy {
```
fviernau, line 50: I'd prefer to keep the explicit type for readability.

sschuberth: Fine 🤷🏻‍♂️ I've dropped that commit.

fviernau (approved):

fviernau:
Thanks for the fixes @sschuberth !

PR #5778, opened 2022-09-12
evaluator: Extend the API for making file checks against the project's source code

fviernau:
Extend the `ProjectSourceRule` to allow making further types of file checks and illustrate the use of each respective new API
with a policy rule derived from a real world use case.

Implements parts of #5621.

### DIFF
--- a/evaluator/src/main/kotlin/ProjectSourceRule.kt
+++ b/evaluator/src/main/kotlin/ProjectSourceRule.kt
@@ -57,6 +57,18 @@ open class ProjectSourceRule(
         }
 
     /**
+     * Return the file paths matching any of the given [glob expressions][patterns] with its file content matching
+     * [contentPattern].
+     */
+    fun projectSourceFindFilesWithContent(contentPattern: String, vararg patterns: String): List<File> =
+        projectSourceFindFiles(*patterns).filter { path ->
+            val content = projectSourcesDir.resolve(path).readText()
+            val regex = contentPattern.toRegex(setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.MULTILINE))
+
+            content.matches(regex)
+        }
+
+    /**
      * A [RuleMatcher] that checks whether the project's source tree contains at least one directory matching any of the
      * provided [glob expressions][patterns].
      */
@@ -75,6 +87,19 @@ open class ProjectSourceRule(
         object : RuleMatcher {
             override val description = "projectSourceHasFile('${patterns.joinToString()}')"
             override fun matches(): Boolean = projectSourceFindFiles(*patterns).isNotEmpty()
+        }
+
+    /**
+     * A [RuleMatcher] that checks whether the project's source tree contains at least one file matching any of the
+     * given [glob expressions][patterns] with its file content matching [contentPattern].
+     */
+    fun projectSourceHasFileWithContent(contentPattern: String, vararg patterns: String): RuleMatcher =
+        object : RuleMatcher {
+            override val description =
+                "projectSourceHasFileWithContents('$contentPattern', '${patterns.joinToString()}')"
+
+            override fun matches(): Boolean =
+                projectSourceFindFilesWithContent(contentPattern, *patterns).isNotEmpty()
         }
 }
 
--- a/evaluator/src/main/kotlin/ProjectSourceRule.kt
+++ b/evaluator/src/main/kotlin/ProjectSourceRule.kt
@@ -60,13 +60,15 @@ open class ProjectSourceRule(
      * Return the file paths matching any of the given [glob expressions][patterns] with its file content matching
      * [contentPattern].
      */
-    fun projectSourceFindFilesWithContent(contentPattern: String, vararg patterns: String): List<File> =
-        projectSourceFindFiles(*patterns).filter { path ->
+    fun projectSourceFindFilesWithContent(contentPattern: String, vararg patterns: String): List<File> {
+        val regex = contentPattern.toRegex(setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.MULTILINE))
+
+        return projectSourceFindFiles(*patterns).filter { path ->
             val content = projectSourcesDir.resolve(path).readText()
-            val regex = contentPattern.toRegex(setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.MULTILINE))
 
             content.matches(regex)
         }
+    }
 
     /**
      * A [RuleMatcher] that checks whether the project's source tree contains at least one directory matching any of the
--- a/evaluator/src/main/kotlin/ProjectSourceRule.kt
+++ b/evaluator/src/main/kotlin/ProjectSourceRule.kt
@@ -40,11 +40,31 @@ open class ProjectSourceRule(
     val projectSourcesDir: File by lazy { projectSourceResolver.rootDir }
 
     /**
+     * Return all directories from the project's source tree which match any of the
+     * provided [glob expressions][patterns].
+     */
+    fun projectSourceFindDirectories(vararg patterns: String): List<File> =
+        projectSourcesDir.walkBottomUp().filterTo(mutableListOf()) {
+            it.isDirectory && FileMatcher.match(patterns.toList(), it.relativeTo(projectSourcesDir).path)
+        }
+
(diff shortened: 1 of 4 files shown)


evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -38,6 +38,15 @@ open class ProjectSourceRule(
...
     val projectSourcesDir: File by lazy { projectSourceResolver.rootDir }
 
+    /**
+     * Return all directories from the project's source tree which match any of the
+     * provided [glob expressions][patterns].
+     */
+    fun projectSourceFindDirectories(vararg patterns: String): List<File> =
+        projectSourcesDir.walkBottomUp().filterTo(mutableListOf()) {
+            it.isDirectory && FileMatcher.match(patterns.toList(), it.relativeTo(projectSourcesDir).path)
+        }
```
mnonnenmacher, line 48: Unless the function is used in a different context in later commits it is inefficient to scan the whole file tree if the first match already gives the required result.

evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
```diff
@@ -62,6 +62,38 @@ class ProjectSourceRuleTest : WordSpec({
...
+        "return false if only a file matches the given glob pattern" {
+            val dir = createSpecTempDir().apply {
+                addFiles("a")
+            }
+            val rule = createOrtResultRule(dir)
+
+            rule.projectSourceHasDirectory("a").matches() shouldBe false
+        }
+
+        "return false if neither any file nor directory matches the given glob pattern" {
```
mnonnenmacher, line 90: The test should either create files and directories that do not match or be renamed to "return false if the directory is empty".
Currently you have no test that verifies that the pattern does not match any path.

examples/evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -264,6 +264,28 @@ fun RuleSet.deprecatedScopeExcludeReasonInOrtYmlRule() = ortResultRule("DEPRECAT
...
+            ),
+            projectSourceHasDirectory(
+                ".circleci",
+                ".github"
+            )
+        )
+    }
+
+    error(
+        message = "This project does not seem to have any known CI configuration files.",
```
mnonnenmacher, line 304: Remove "seem to".

examples/evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -264,6 +264,28 @@ fun RuleSet.deprecatedScopeExcludeReasonInOrtYmlRule() = ortResultRule("DEPRECAT
...
+        howToFix = "Please setup a CI. If you already have setup a CI and the error persists, please contact support."
```
mnonnenmacher, line 305: Mentioning support is a bit weird in an example rules file, better suggest to check if the CI file is contained in the file patterns.

examples/evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -264,6 +264,28 @@ fun RuleSet.deprecatedScopeExcludeReasonInOrtYmlRule() = ortResultRule("DEPRECAT
...
+        -AnyOf(
+            projectSourceHasFile(
+                ".appveyor.yml",
+                ".bitbucket-pipelines.yml",
+                ".gitlab-ci.yml",
+                ".travis.yml"
+            ),
+            projectSourceHasDirectory(
+                ".circleci",
+                ".github"
```
mnonnenmacher, line 298: Better check for ".github/workflows", the ".github" directory can contain files unrelated to CI.

examples/evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -180,6 +180,26 @@ fun RuleSet.copyleftInSourceLimitedRule() = packageRule("COPYLEFT_LIMITED_IN_SOU
...
+
+    denyDirPatterns.forEach { (pattern, packageManagers) ->
+        val offendingDirs = projectSourceFindDirectories(pattern)
+
+        if (offendingDirs.isNotEmpty()) {
+            issue(
+                Severity.ERROR,
+                "The directories ${offendingDirs.joinToString()} belong to the package manager(s) " +
+                        "${packageManagers.joinToString()} and must not be committed.",
+                "Please delete the directories: ${offendingDirs.joinToString()}."
```
mnonnenmacher, line 197: Nit: "the" -> "these"

evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -55,6 +55,18 @@ open class ProjectSourceRule(
...
         }
 
+    /**
+     * Return the file paths matching any of the given [glob expressions][patterns] with its file content matching
+     * [contentPattern].
+     */
+    fun projectSourceFindFilesWithContent(contentPattern: String, vararg patterns: String): List<File> =
+        projectSourceFindFiles(*patterns).filter { path ->
+            val content = projectSourcesDir.resolve(path).readText()
+            val regex = contentPattern.toRegex(setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.MULTILINE))
```
mnonnenmacher, line 65: The regex should be created outside the loop.

evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
```diff
@@ -103,13 +121,14 @@ private fun createOrtResultRule(projectSourcesDir: File): ProjectSourceRule =
...
 
-private fun File.addFiles(vararg paths: String) {
+private fun File.addFiles(vararg paths: String, content: String = "") {
     require(isDirectory)
 
     paths.forEach { path ->
         resolve(path).apply {
             parentFile.mkdirs()
             createNewFile()
+            writeText(content)
```
mnonnenmacher, line 131: Maybe: `if (content.isNotEmpty()) writeText(content)`

examples/evaluator-rules/src/main/resources/example.rules.kts
```diff
@@ -314,6 +314,26 @@ fun RuleSet.missingContributingFileRule() = projectSourceRule("MISSING_CONTRIBUT
...
+}
+
+fun RuleSet.missingReadmeFileLicenseSectionRule() = projectSourceRule("MISSING_README_FILE_LICENSE_SECTION") {
+    require {
+        +projectSourceHasFile("README.md")
+        -projectSourceHasFileWithContent(".*^#{1,2} License$.*", "README.md")
+    }
+
+    error(
+        message = "This project does not seem to have any known CI configuration files.",
```
mnonnenmacher, line 332: Wrong message.

mnonnenmacher (changes requested):

examples/evaluator-rules/src/main/resources/example.rules.kts
fviernau: If someone has CI setup, then the rule reports a false positive. 
In that case I believe that the devs responsible for the rules have to look into it anyway.
So, IMO it does not make sense to ask the user to look into any further details of the heuristic first.

Can we keep it, also for simplicity and for illustrating how the process could work?

evaluator/src/main/kotlin/ProjectSourceRule.kt
fviernau: It is used in a later commit. It is pubic so that it can be used from the user's `rules.kt` file, either directly from the rule or for composing a new matcher. Can we leave this performance optimization out of scope here?

examples/evaluator-rules/src/main/resources/example.rules.kts
mnonnenmacher: This implies a context that the reader of this file is likely unaware of, but ok.

examples/evaluator-rules/src/main/resources/example.rules.kts
mnonnenmacher: I think "these" can refer to anything in context, not necessarily mentioned before. I'm no English expert either but I would have used "diese" instead of "die" in German, but both are fine of course.

mnonnenmacher (approved):

PR #5784, opened 2022-09-13
Curation provider alignments

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -27,7 +27,7 @@ import org.apache.logging.log4j.kotlin.Logging
 import org.ossreviewtoolkit.analyzer.PackageCurationProvider
 import org.ossreviewtoolkit.model.FileFormat
 import org.ossreviewtoolkit.model.PackageCuration
-import org.ossreviewtoolkit.model.readValueOrDefault
+import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.utils.common.getDuplicates
 
 /**
@@ -53,7 +53,7 @@ class FilePackageCurationProvider(
 
             curationFiles.map { curationFile ->
                 val curations = runCatching {
-                    curationFile.readValueOrDefault(emptyList<PackageCuration>())
+                    curationFile.readValue<List<PackageCuration>>()
                 }.onFailure {
                     logger.warn { "Failed parsing package curation from '${curationFile.absoluteFile}'." }
                 }.getOrThrow()
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -21,6 +21,7 @@
 package org.ossreviewtoolkit.analyzer.curation
 
 import java.io.File
+import java.io.IOException
 
 import org.apache.logging.log4j.kotlin.Logging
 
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -55,8 +55,8 @@ class FilePackageCurationProvider(
                 val curations = runCatching {
                     curationFile.readValueOrDefault(emptyList<PackageCuration>())
                 }.onFailure {
-                    logger.warn { "Failed parsing package curation from '${curationFile.absoluteFile}'." }
-                }.getOrThrow()
+                    logger.error { "Failed parsing package curation from '${curationFile.absoluteFile}'." }
+                }.getOrDefault(emptyList())
```
sschuberth, line 59: @fviernau and @MarcelBochtler, IIRC we had discussions (and different opinions) about "failing early" in this case before, but the current behavior just seems inconsistent to me. Please have a look at the commit message for the details.

mnonnenmacher: I'm for changing the log level but don't like to continue with only printing an error. Consistent results are very important, and no one is checking the log after every run if it might contain an error message somewhere. IMO if there is an error ORT should either fail or create an `OrtIssue` so that you can at least see the issue in the report, otherwise issues like this will often go unnoticed.
And for this specific case, for a repository that contains curations it is easy to setup a CI job that verifies that the curations are at least parseable, so failing hard seems ok to me.

fviernau: I'm also for failing hard, as I don't see a point in continue running with a broken configuration.
The price of the issue going unnoticed can be very high. I've experienced this myself 
in the past, where I filed https://github.com/oss-review-toolkit/ort/pull/4500.
THB I don't see any benefits in continuing execution.

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
```diff
@@ -61,7 +59,7 @@ open class OrtConfigPackageCurationProvider : PackageCurationProvider {
         val file = curationsDir.resolve("curations").resolve(pkgId.toCurationPath())
         return if (file.isFile) {
             runCatching {
-                yamlMapper.readValue<List<PackageCuration>>(file).filter { it.isApplicable(pkgId) }
+                file.readValueOrDefault(emptyList<PackageCuration>()).filter { it.isApplicable(pkgId) }
```
fviernau, line 62: Please provide a rationale why tolerating empty files is better.
The argument of consistency would also allow aligning the other way round.

fviernau (changes requested):

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
sschuberth: > The argument of consistency would also allow aligning the other way round.

Well, given that you guys are for failing hard, it indeed makes more sense to align the other way around.

fviernau: Ups sorry. I just deleted my cited comment as I thought is was too minor to bring it up. 
I hoped it was fast enough but wasn't.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
sschuberth: > no one is checking the log after every run if it might contain an error message somewhere.

So why do we then log at all here, if we anyway throw? Isn't throwing only then enough?

fviernau: > So why do we then log at all here, if we anyway throw? Isn't throwing only then enough?

Throwing is not enough, because the user won't be able to figure out the problematic file.

mnonnenmacher: If you remove the log, just make sure that at least the exception message is always printed even without `--stacktrace`.

fviernau: > Ok, so you basically just want to amend the exception message, and could live without logging anything? I'll make a proposal for that.

I don't really care as long as it is possible to figure out the problematic file from the output.
Even though I think it's good to make that task as easy as possible for the user. If it's cryptic 
then the curation author may ask devs for help, which may be an avoidable ping pong.

fviernau (approved):

PR #5788, opened 2022-09-13
Schemas: Create separate analyzer configuration schema 

MarcelBochtler:
Analyzer configurations can be used in the main ORT configurations file as well as in the repository configuration. Move the analyzer configurations to a separate schema to be able to reuse them.

The referencing of this extracted schema will be done in a follow-up commit.


sschuberth:
Nit: Because of

> The referencing of this extracted schema will be done in a follow-up commit.

it sounds a but strange to talk about a "move" in 

> Move the analyzer configurations to a separate schema to be able to reuse them.

and in the title, as you're not actually moving anything yet, but rather copying.

Maybe still do the referencing in this commit already? Or at least in the same PR, so the unified diff for all commits nicely shows the code deduplication?

sschuberth (approved):

PR #5792, opened 2022-09-13
Minor FossId improvements

fviernau:
The original motivation of changing something was to eliminate the redundant spread operators in all callers of `runScan()`.
When dropping the redundant spread operators the callers which provided multiple pairs as varargs became a tad hard to read.
So, this PR refactors to use the type `Map` for the options, and thereby eliminates the rundant spread operators.

### DIFF
--- a/clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
+++ b/clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
@@ -26,5 +26,5 @@ class PostRequestBody(
     apiKey: String,
     vararg options: Pair<String, String>
 ) {
-    val data: MutableMap<String, String> = mutableMapOf("username" to user, "key" to apiKey, *options)
+    val data = mapOf("username" to user, "key" to apiKey, *options)
 }
--- a/clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
+++ b/clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
@@ -24,7 +24,7 @@ class PostRequestBody(
     val group: String,
     user: String,
     apiKey: String,
-    vararg pairs: Pair<String, String>
+    vararg options: Pair<String, String>
 ) {
-    val data: MutableMap<String, String> = mutableMapOf("username" to user, "key" to apiKey, *pairs)
+    val data: MutableMap<String, String> = mutableMapOf("username" to user, "key" to apiKey, *options)
 }
--- a/clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
+++ b/clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
@@ -24,7 +24,11 @@ class PostRequestBody(
     val group: String,
     user: String,
     apiKey: String,
-    vararg options: Pair<String, String>
+    options: Map<String, String> = emptyMap()
 ) {
-    val data = mapOf("username" to user, "key" to apiKey, *options)
+    val data = buildMap {
+        put("username", user)
+        put("key", apiKey)
+        putAll(options)
+    }
 }
--- a/clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
+++ b/clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
@@ -19,16 +19,24 @@
 
 package org.ossreviewtoolkit.clients.fossid
 
-class PostRequestBody(
+data class PostRequestBody private constructor(
     val action: String,
     val group: String,
-    user: String,
-    apiKey: String,
-    options: Map<String, String> = emptyMap()
+    val data: Map<String, String>
 ) {
-    val data = buildMap {
-        put("username", user)
-        put("key", apiKey)
-        putAll(options)
-    }
+    constructor(
+        action: String,
+        group: String,
+        user: String,
+        apiKey: String,
+        options: Map<String, String> = emptyMap()
+    ) : this(
+        action,
+        group,
+        buildMap {
+            put("username", user)
+            put("key", apiKey)
+            putAll(options)
+        }
+    )
 }
--- a/scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
+++ b/scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
@@ -384,7 +384,7 @@ class FossIdTest : WordSpec({
                     USER,
                     API_KEY,
                     scanCode,
-                    *arrayOf(
+                    mapOf(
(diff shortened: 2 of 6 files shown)


clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
```diff
@@ -19,16 +19,24 @@
 
 package org.ossreviewtoolkit.clients.fossid
 
-class PostRequestBody(
+data class PostRequestBody private constructor(
```
sschuberth, line 22: What's the benefit of turning this into a data class? Please provide a rationale in the commit message.

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
sschuberth: > The generated functions may be useful.

In what context might the generated functions be useful? I mean, we'll probably never need a proper `equals` for a `PostRequestBody`.

Don't get me wrong, I'm not completely against turning this into a data class. I'm just curious about the direction behind that idea, if any. Because standing on its own, this just increases the number of code lines without an obvious benefit currently.

clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
```diff
@@ -19,16 +19,24 @@
...
+        user: String,
+        apiKey: String,
+        options: Map<String, String> = emptyMap()
+    ) : this(
+        action,
+        group,
+        buildMap {
+            putAll(options)
+            put("username", user)
+            put("key", apiKey)
```
sschuberth, line 39: Here the order is still swapped.

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/PostRequestBody.kt
fviernau: I didn't have any specific case in mind. The class just seemed like a perfect fit for a data class, and I thought the client library might be used also from outside ort. Maybe some test compares a constructed test via `equals` maybe someone adds it to a `Set` / `Map` or uses `copy()` to change a single attribute value. 

I wouldn't have an issue at all with dropping this commit. What do you prefer?

sschuberth: Ok, thanks for your thoughts. Let's keep the code then now that it's there.

sschuberth (approved):

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -646,6 +646,18 @@ class FossIdTest : WordSpec({
...
+
+            coVerify(exactly = 1) {
+                service.runScan(
+                    USER,
+                    API_KEY,
+                    scanCode,
+                    *arrayOf(
+                        "auto_identification_detect_declaration" to "0",
+                        "auto_identification_detect_copyright" to "0"
+                    )
```
sschuberth, line 658: Sorry, I'm not good at reading MockK code. While I've verified that the test passes, I do not understand why. Doesn't this array also match the `any()` from line 647? Because if so, how can line 647 not be called at all, but this code be called once?

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -643,7 +643,7 @@ class FossIdTest : WordSpec({
                 service.checkDownloadStatus(USER, API_KEY, scanCode)
             }
 
-            coVerify(exactly = 0) {
+            coVerify(exactly = 1) {
```
sschuberth, line 646: Doesn't this now check exactly the same function signature as the check you added in the previous commit? And if so, can't we just drop this check completely?

fviernau: This now checks that there is one invocation in total. It matches the parameters exactly, but matches arbitrary values for the options. So, to put the assertion in words: "Assert that there is no invocation besides the one asserted exactly below." E.g. the one with:

```
 "auto_identification_detect_declaration" to "0",
 "auto_identification_detect_copyright" to "0"`
```

The reason why I added the assertion was that I also had a hard time figuring out why the test started failing.
And I thought the added assertion brings at least a bit of clarity / understanding.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
fviernau: > doesn't this array also match the any() from line 647? 

I thought that too initially, but I found it doesn't.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
fviernau: I read up a bit on this topic and it seems that what I've been doing with two assertions should be done with just one, using `verifyAll`, see https://github.com/mockk/mockk/issues/74.

fviernau: Hm, `verifyAll` also is not really what fits as that would verify all interactions with the mock, while my assertion is about all calls to the `runScan` function of the mock, but doesn't care about all the other function invocations.

fviernau: Any preference what to do @sschuberth ? I tend to change amend my commit which introduces the new assertion to remove the existing one above. (Rationale: we verify that the created scan is run, but don't check that it is not run twice. This seems reasonable as, like the test name suggests, the test is about creation of the scan - not about executing (running). Makes sense to you as well?

fviernau: I've updated the commits according to my previous comment.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
sschuberth: Apparently one has to use [`*anyVararg()`](https://mockk.io/#varargs) instead.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
sschuberth: > Makes sense to you as well?

Yes. However, in the commit message shouldn't "without empty `options`" say "with empty `options`"?

fviernau: ...right!

sschuberth (approved):

PR #5794, opened 2022-09-14
OsvServiceFunTest: Be lenient about the array order in all tests

sschuberth:
So far this was only done for "getVulnerabilityForId() returns the expected vulnerability for the given ID". Consistently do this in all tests to avoid breakages due changes in array order only.

### DIFF
--- a/clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
+++ b/clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
@@ -54,7 +54,10 @@ class OsvServiceFunTest : StringSpec({
         val result = OsvService().getVulnerabilitiesForPackage(VULNERABILITY_FOR_PACKAGE_BY_COMMIT_REQUEST)
 
         result.shouldBeSuccess {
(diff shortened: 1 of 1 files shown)


clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
```diff
@@ -54,7 +54,10 @@ class OsvServiceFunTest : StringSpec({
         val result = OsvService().getVulnerabilitiesForPackage(VULNERABILITY_FOR_PACKAGE_BY_COMMIT_REQUEST)
 
         result.shouldBeSuccess {
-            OsvApiClient.JSON.encodeToString(it) shouldEqualJson expectedResult
+            OsvApiClient.JSON.encodeToString(it).shouldEqualJson(
```
mnonnenmacher, line 57: Commit message nit: "due **to**"

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5797, opened 2022-09-14
FossIdTest improvements

mnonnenmacher:
See the commit messages for details. This is preparation for removing the legacy scanner.

### DIFF
--- a/scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
+++ b/scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
@@ -148,7 +148,7 @@ class FossIdTest : WordSpec({
 
             val fossId = createFossId(config)
 
-            val summary = fossId.scan(listOf(createPackage(pkgId, vcsInfo))).summary(pkgId)
+            val summary = fossId.scan(createPackage(pkgId, vcsInfo)).summary(pkgId)
 
             summary.issues shouldHaveSize 1
             with(summary.issues.first()) {
@@ -175,7 +175,7 @@ class FossIdTest : WordSpec({
 
             val fossId = createFossId(config)
 
-            fossId.scan(listOf(createPackage(createIdentifier(index = 1), vcsInfo)))
+            fossId.scan(createPackage(createIdentifier(index = 1), vcsInfo))
 
             coVerify {
                 service.createScan(USER, API_KEY, projectCode, scanCode, vcsInfo.url, vcsInfo.revision)
@@ -208,7 +208,7 @@ class FossIdTest : WordSpec({
                     EntityResponseBody(status = 1, data = DownloadStatus.FAILED)
 
             val fossId = createFossId(config)
-            fossId.scan(listOf(createPackage(createIdentifier(index = 1), vcsInfo)))
+            fossId.scan(createPackage(createIdentifier(index = 1), vcsInfo))
 
             coVerify {
                 service.createScan(USER, API_KEY, projectCode, scanCode, vcsInfo.url, vcsInfo.revision)
@@ -238,7 +238,7 @@ class FossIdTest : WordSpec({
 
             val fossId = createFossId(config)
 
-            val summary = fossId.scan(listOf(createPackage(pkgId, vcsInfo))).summary(pkgId)
+            val summary = fossId.scan(createPackage(pkgId, vcsInfo)).summary(pkgId)
 
             val expectedCopyrightFindings = listOf(
                 CopyrightFinding("copyrightMarked1", textLocation(1)),
@@ -269,7 +269,7 @@ class FossIdTest : WordSpec({
 
             val fossId = createFossId(config)
 
-            val summary = fossId.scan(listOf(createPackage(pkgId, vcsInfo))).summary(pkgId)
+            val summary = fossId.scan(createPackage(pkgId, vcsInfo)).summary(pkgId)
 
             val expectedCopyrightFindings = listOf(
                 CopyrightFinding("copyrightMarked1", textLocation(1))
@@ -299,7 +299,7 @@ class FossIdTest : WordSpec({
 
             val fossId = createFossId(config)
 
-            val summary = fossId.scan(listOf(createPackage(pkgId, vcsInfo))).summary(pkgId)
+            val summary = fossId.scan(createPackage(pkgId, vcsInfo)).summary(pkgId)
 
             val expectedCopyrightFindings = listOf(
                 CopyrightFinding("copyright1", textLocation(1)),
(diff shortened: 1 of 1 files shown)


scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -676,9 +676,10 @@ class FossIdTest : WordSpec({
 
             val fossId = createFossId(config)
 
-            val scannerRun = fossId.scan(listOf(pkg1, pkg2))
+            fossId.scan(listOf(pkg1))
```
sschuberth: Does this line then need to be kept at all? I currently don't see an assertion that would depend on it.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -148,7 +148,7 @@ class FossIdTest : WordSpec({
 
             val fossId = createFossId(config)
 
-            val summary = fossId.scan(listOf(createPackage(pkgId, vcsInfo))).summary(pkgId)
+            val summary = fossId.scan(createPackage(pkgId, vcsInfo)).summary(pkgId)
```
sschuberth: Commit message nit: Maybe drop "also" from "which also scans".

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
mnonnenmacher: How I understand the test the last check `coVerify { service.deleteScan(USER, API_KEY, scanCode) }` verifies that the scan of `pkg1` gets deleted as well if the scan of `pkg2` fails. But to be honest, I had a hard time coming to that conclusion and the whole test class seems a bit too complex.

nnobelis: @sschuberth this is required by `scannerRun.results.scanResults.keys shouldHaveSize 1`

If one scan fails, other successful scans should be deleted hence there should be no result for them.

nnobelis: That being said, I think theses changes are wrong @mnonnenmacher : now there a two scanner runs.
The first one is successful. The second one fails. However since these are two separate runs, the first scan is not deleted anymore.

nnobelis (changes requested):

nnobelis:
Please see my comment, I think changing the signature of `scan` is wrong here, because it removes the current feature of deleting all scans if one fails.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
mnonnenmacher: The test does not fail which shows the other scan is deleted. The reason is that the same instance of `FossId` is used and `createdScans` (which is used as input when deleting scans) is a class variable. However, if the scans would be triggered the other way round it would probably not work. It seems an interface change is required, but I'm also a bit surprised that this comes up only now since it's running like this for several months now on our side.

nnobelis: Ah right, we instantiate one `FossID` instance per scanner run so this is working as intended.

Yes, the test is dependent on the scan ordering.

However I don't get the second part of your sentence

> It seems an interface change is required, but I'm also a bit surprised that this comes up only now since it's running like this for several months now on our side.

Can you elaborate ?

mnonnenmacher: @nnobelis So I had a look at the current implementation, especially these to places:

https://github.com/oss-review-toolkit/ort/blob/1258f27bd2f9450f855acfcaa3343a6964f3330b/scanner/src/main/kotlin/scanners/fossid/FossId.kt#L314-L318

https://github.com/oss-review-toolkit/ort/blob/1258f27bd2f9450f855acfcaa3343a6964f3330b/scanner/src/main/kotlin/scanners/fossid/FossId.kt#L351-L355

Both are inside a loop over the packages to scan, so if I read the code correctly the behavior was always like this, that scans started before a failed scan would be deleted, but other scans could still be started after a failed scan. Is this an issue that needs to be fixed or is the current code working correctly?

mnonnenmacher: > > It seems an interface change is required, but I'm also a bit surprised that this comes up only now since it's running like this for several months now on our side.
> 
> Can you elaborate ?

The experimental scanner already calls the scanner once per package since its introduction. But see my previous comment, either the behavior is already broken or the change is fine. But this needs to be clarified now, because if it is broken either the `PackageScannerWrapper` interface needs to be adapted to take a list of packages or the issue needs to be fixed inside the `FossId` class somehow.

sschuberth: > the `PackageScannerWrapper` interface needs to be adapted to take a list of packages

I was wondering whether that doesn't make sense anyway, as it could allow for some significant scanner-implementation-internal performance improvements. A bit similar to how I changed `PackageCurationProvider` to take collection of packages in 0592a68f.

nnobelis: > @nnobelis So I had a look at the current implementation, especially these two places:
> 
> https://github.com/oss-review-toolkit/ort/blob/1258f27bd2f9450f855acfcaa3343a6964f3330b/scanner/src/main/kotlin/scanners/fossid/FossId.kt#L314-L318
> 
> https://github.com/oss-review-toolkit/ort/blob/1258f27bd2f9450f855acfcaa3343a6964f3330b/scanner/src/main/kotlin/scanners/fossid/FossId.kt#L351-L355
> 
> Both are inside a loop over the packages to scan, so if I read the code correctly the behavior was always like this, that scans started before a failed scan would be deleted, but other scans could still be started after a failed scan. Is this an issue that needs to be fixed or is the current code working correctly?

Yes I think you find a bug. Originally, the logic was to delete the scan in the `catch` and let the exception bubble up. Then somebody came with "it should not throw an exception, it should create an `ortIssue` instead.

But you are right, if a scan fails all successful scans should be deleted (if the option is set !) and no further scan should be created.
If the option is not set, successful scans should be kept and further scans should be run.
Makes sense ?

mnonnenmacher: Ok, I will propose a fix in a separate PR and update this PR afterwards.

mnonnenmacher: @nnobelis For now I have only added a commit that changes the test to illustrate the issue which also shows that this refactoring does not change the behavior.

mnonnenmacher: > > the `PackageScannerWrapper` interface needs to be adapted to take a list of packages
> 
> I was wondering whether that doesn't make sense anyway, as it could allow for some significant scanner-implementation-internal performance improvements. A bit similar to how I changed `PackageCurationProvider` to take collection of packages in [0592a68](https://github.com/oss-review-toolkit/ort/commit/0592a68fde252f95f730727755ddf889843d0822).

This might be, but this requires changes in `ExperimentalScanner` and because our only `PackageScannerWrapper` implementation `FossId` does not do any such optimizations I would not want to work on this before we have an actual use case.

mnonnenmacher:
@sschuberth @nnobelis Would one of you mind approving this or is there something left to do?

nnobelis (approved):

Issue #5681, opened 2022-08-24
ScanResult: Add a list of all files with checksum and directories


fviernau:
A complete list of all file and directory names in the repository along with the checksum can be valueable for various use cases.
For example:

1. Compute the package verification code on-the-fly
2. Implement basic file / dir existence checks without the need for cloning the source code
3. Auto generate excludes based on directory names

Moreover, the serialized scan result is redundant in terms of the file path. Each file path can be listed zero to multiple times, as it is listed once per license and copyright finding. That redundancy can be removed by changing the way license and copyright findings are serialized.

### Goal: 

A scan result should:
1. contain a full list of all files with checksums 
2. A full list of all (empty) dirs, so that directory existance checks can be implemented on top
3. have the redundancy in terms of file paths eliminated

sschuberth:
Some remarks from my side:
- How do we know what "all" files are? Scanners might only report files with findings (this is also true for ScanCode when run with the [`--only-findings` option](https://scancode-toolkit.readthedocs.io/en/stable/cli-reference/output-filters-and-control.html#only-findings-plugin)), so we cannot generally just use the files from the scanner's raw result.
  - If ORT should itself determine the list of "all" files, then ORT should probably *not* also calculate the hashes for all these files, as e.g. ScanCode already does that for files contained in its raw result. So as a performance optimization, ORT should only calculate the hashes of the remaining files.
- We should be prepared to calculate multiple different hashes per file. SHA1 obviously makes sense, but so does SHA1GIT (for lookup at Software Heritage).
- In general, starting to list *all* scanned files in the scan result, including those without findings, will make the scan result grow quite a bit. Does it still make sense to only add files with findings to the result, and only the hashes for those?

fviernau:
We've discussed my proposal with the core devs, here's the proposed changed to the OrtResult model,
and some notes from the meeting.

### Planned changes in ORT result model

The following things I believe are somewhat on the agenda and will impact
the model, so these should be considered to avoid conflicts or going into the wrong direction: 

1. Have scan results per provenance to drop redundancy, e.g. map package IDs to provenances.
2. Have a file list with hashes for all files for each resolved provenance
3. Have sniplet findings

### Current data structure

```
OrtResult {
  - scanner: ScannerRun { 
    - startTime
    - endTime
    - environment
    - config
    - results: ScanRecord {
      - storageStats
      - scanResults: Map<Id, List<ScanResult>>
    }
  }     
}    
   
ScanResult {
[...]

PR #5799, opened 2022-09-14
Replace legacy scanner with experimental scanner

mnonnenmacher:
Remove the legacy scanner and replace it with the experimental scanner. This is still work in progress.


mnonnenmacher:
Some tests might still be failing, but please review already.

scanner/src/main/kotlin/CommandLinePathScannerWrapper.kt
```diff
@@ -0,0 +1,131 @@
...
+
+import org.ossreviewtoolkit.model.ScannerDetails
+import org.ossreviewtoolkit.scanner.experimental.PathScannerWrapper
+import org.ossreviewtoolkit.utils.common.CommandLineTool
+import org.ossreviewtoolkit.utils.common.Os
+
+/**
+ * A [PathScannerWrapper] that is executed as a [CommandLineTool] on the local machine.
+ */
+abstract class CommandLinePathScannerWrapper(
```
sschuberth, line 36: Nit: Why not just "CommandLineScannerWrapper"?

scanner/src/main/kotlin/scanners/Askalono.kt
```diff
@@ -50,23 +49,18 @@ import org.ossreviewtoolkit.utils.spdx.calculatePackageVerificationCode
...
-
-    class Factory : AbstractScannerFactory<Askalono>("Askalono") {
-        override fun create(scannerConfig: ScannerConfiguration, downloaderConfig: DownloaderConfiguration) =
-            Askalono(scannerName, scannerConfig, downloaderConfig)
+            Askalono(scannerName, scannerConfig)
     }
 
     override val name = "Askalono"
     override val criteria by lazy { ScannerCriteria.fromConfig(details, scannerConfig) }
+
```
sschuberth, line 62: Nit: Unintended change? At least this is not also done for all other scanners.

scanner/src/main/kotlin/CommandLinePathScannerWrapper.kt
```diff
@@ -0,0 +1,131 @@
...
+     * The configuration used by the scanner, should contain command line options that influence the scan result.
+     */
+    abstract val configuration: String
+
+    /**
+     * The actual version of the scanner, or an empty string in case of failure.
+     */
+    val version by lazy { getVersion(scannerDir) }
+
+    override val details: ScannerDetails by lazy { ScannerDetails(name, version, configuration) }
```
sschuberth, line 104: Nit: Could drop the explicit return type.

scanner/src/main/kotlin/scanners/Askalono.kt
```diff
@@ -105,7 +99,9 @@ class Askalono internal constructor(
         return unpackDir
     }
 
-    override fun scanPathInternal(path: File): ScanSummary {
+    override fun filterSecretOptions(options: Options): Options = options
```
sschuberth: I was about to suggest this should become a default implementation, but then saw you do this later. Maybe move the commit with the default implementation to earlier, to reduce the diff here (and for the other scanners)?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -200,9 +193,7 @@ class FossId internal constructor(
 
     override val criteria: ScannerCriteria? = null
     override val name: String = "FossId"
-    override val version: String = service.version
-
-    override val configuration = ""
+    override val details: ScannerDetails = ScannerDetails(name, service.version, "")
```
sschuberth, line 196: Nit: Could drop the explicit type declaration.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -479,9 +479,12 @@ class FossIdTest : WordSpec({
 
             shouldThrow<TimeoutCancellationException> {
                 withTimeout(1000) {
-                    fossId.scanPackages(
-                        setOf(createPackage(createIdentifier(index = 1), vcsInfo)), emptyMap()
-                    )
+                    launch {
```
sschuberth, line 482: Note to myself: I need to think about whether `launch` is the correct thing to use here.

sschuberth (commented):
Just a few comments in this first iteration. Will do a more thorough second review tomorrow.

scanner/src/main/kotlin/scanners/Askalono.kt
mnonnenmacher: This wasn't possible earlier, because then there would have been two default implementations, one from `Scanner` and one from `ScannerWrapper` so all implementing classes would have had to re-implement anyway. It was easier to just wait until no class inherits `Scanner` anymore.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
mnonnenmacher: The reason to use `launch` here was that the code must be non-blocking for the timeout to work correctly.

mnonnenmacher: The surrounding `withTimeout(1000)` cancels the coroutine running inside it afer the given milliseconds and if it is still running it throws a `TimeoutCancellationException`. If it would be finished already, it would not throw an exception.

scanner/src/main/kotlin/CommandLinePathScannerWrapper.kt
mnonnenmacher: I guess to be more explicit. I'm fine with either name but if it's renamed I'd prefer to do this after this PR is merged.

sschuberth: I'm also ok with sticking to the longer name. I was just curious whether you'd like to emphasize that there might be "CommandLine" scanners that are not "PathScanners".

cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -162,37 +137,17 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license
...
-            help = "A comma-separated list of experimental scanners to use for scanning the source code of projects. " +
-                    "By default, projects and packages are scanned with the same scanner as specified by " +
-                    "'--experimental-scanner'.\n" +
-                    "Possible values are: ${ScannerWrapper.ALL}"
-        ).convertToScannerWrapperFactories()
-    ).single()
+    private val scanners by option(
+        "--scanners", "-s",
+        help = "A comma-separated list of scanners to use.\nPossible values are: ${ScannerWrapper.ALL}"
+    ).convertToScannerWrapperFactories().required()
```
sschuberth, line 135: Is this intentionally now `required` instead of providing a default scanner? IMO we should stick to ScanCode as a default scanner.

scanner/src/funTest/kotlin/scanners/ScannerIntegrationFunTest.kt
```diff
@@ -58,22 +93,28 @@ class ScannerIntegrationFunTest : StringSpec() {
...
+                    location = TextLocation(file.relativeTo(path).invariantSeparatorsPath, TextLocation.UNKNOWN_LINE)
+                )
+            }
+
             return ScanSummary(
                 startTime = time,
                 endTime = time,
-                packageVerificationCode = calculatePackageVerificationCode(path),
-                licenseFindings = sortedSetOf(),
+                packageVerificationCode = "",
```
sschuberth, line 113: > As the experimental scanner does not support the calculation of the package verification code, create a list of fake license findings instead to verify that the correct source code was scanned.

I'm not sure I can follow you here: Couldn't you just have the left the call to `calculatePackageVerificationCode()` in here?

scanner/src/main/kotlin/CommandLinePathScannerWrapper.kt
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright (C) 2022 Bosch.IO GmbH
```
sschuberth, line 2: Nit: New files could already use the new Copyright statement.

scanner/src/main/kotlin/scanners/scanoss/ScanOss.kt
```diff
@@ -51,35 +51,26 @@ import org.ossreviewtoolkit.scanner.experimental.ScanContext
...
     private val service = ScanOssService.create(config.apiUrl)
 
-    override val name = scannerName
     override val criteria by lazy { ScannerCriteria.fromConfig(details, scannerConfig) }
 
     // TODO: Find out the best / cheapest way to query the SCANOSS server for its version.
-    override val version = BuildConfig.SCANOSS_VERSION
-
-    override val configuration = ""
+    override val details: ScannerDetails = ScannerDetails(name, BuildConfig.SCANOSS_VERSION, "")
```
sschuberth, line 72: The explicit type is there there.

sschuberth (changes requested):

scanner/src/funTest/kotlin/scanners/ScannerIntegrationFunTest.kt
mnonnenmacher: No, the experimental scanner discards any package verification codes from the scan summaries returned by the scanners, so providing one here would be useless. Removing the package verification code from the model is one of the cleanup tasks for after this PR, and they will only be available again once #5681 is implemented.

cli/src/main/kotlin/commands/ScannerCommand.kt
mnonnenmacher: No, this was an oversight.

sschuberth (approved):

mnonnenmacher:
If tests pass I plan to merge this next Tuesday.

sschuberth:
@mnonnenmacher I'll fix the `ExcelReporterFunTest` failure which is unrelated to your changes.

sschuberth:
> @mnonnenmacher I'll fix the `ExcelReporterFunTest` failure which is unrelated to your changes.

Turns out it *was* related to your PR; I've taken the liberty to push a fixup commit for you to squash.

sschuberth (approved):

PR #5801, opened 2022-09-15
scanner: Move code to create criteria from config to `ScannerCriteria`

mnonnenmacher:
Move the code from `PathScanner` to `ScannerCriteria` as a preparation for removing the `PathScanner` class.


scanner/src/main/kotlin/ScannerCriteria.kt
```diff
@@ -108,6 +124,25 @@ data class ScannerCriteria(
...
+        /**
+         * Return a [ScannerCriteria] object to be used when looking up existing scan results from a
+         * [ScanResultsStorage]. Per default, the properties of this object are initialized to match this scanner
+         * implementation. It is, however, possible to override these defaults from the configuration, in the
+         * [ScannerConfiguration.options] property: Use properties of the form `scannerName.property`, where
+         * `scannerName` is the name of the scanner the configuration applies to, and `property` is the name of a
+         * property of the [ScannerCriteria] class. For instance, to specify that a specific minimum version of ScanCode
+         * is allowed, set this property: `options.ScanCode.minVersion=3.0.2`.
+         */
+        fun fromConfig(details: ScannerDetails, config: ScannerConfiguration): ScannerCriteria {
```
sschuberth, line 137: Now that this function has parameters, would it make sense to refer to them from the docs?

sschuberth (approved):
I have only an optional comment.

scanner/src/main/kotlin/ScannerCriteria.kt
mnonnenmacher: Good point, I have improved the docs a bit.

sschuberth (approved):

PR #5806, opened 2022-09-15
possibility to call cli with --warn option - compability with ort-gitlab-ci

unclejay80:
in https://github.com/oss-review-toolkit/ort-gitlab-ci/blob/main/scripts/set-defaults.sh#L12 the default log level is set to info.
if you set it to warn the ort cli will report an error because --warn is not allowed on the command line

To have more compability with the ort-gitlab-ci project this change will help a lot

Signed-off-by: Jürgen Wischer <33184634+unclejay80@users.noreply.github.com>

### DIFF
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -92,7 +92,8 @@ class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
 
     private val logLevel by option(help = "Set the verbosity level of log output.").switch(
         "--info" to Level.INFO,
-        "--debug" to Level.DEBUG
+        "--debug" to Level.DEBUG,
+        "--warn" to Level.WARN
     ).default(Level.WARN)
 
     private val stacktrace by option(help = "Print out the stacktrace for all exceptions.").flag()


cli/src/main/kotlin/OrtMain.kt
```diff
@@ -93,6 +93,7 @@ class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
     private val logLevel by option(help = "Set the verbosity level of log output.").switch(
         "--info" to Level.INFO,
         "--debug" to Level.DEBUG
+        "--warn" to Level.WARN
```
sschuberth, line 96: There's a comma missing in the line above.

sschuberth (changes requested):

sschuberth:
Thanks for the contribution @unclejay80! In this project, we put an eye on sophisticated commit messages that properly explain the rationale. I propose something like this:

```
OrtMain: Allow to explicitly set the `warn` log level

Even if `warn` already is the default, wrappers around the ORT CLI might
explicitly want to use `--warn` for convenience, e.g. because the log level
is passed as-is from a UI form. So add support for `--warn.`
```

Would you mind amending your commit message accordingly?

unclejay80:
> Thanks for the contribution @unclejay80! In this project, we put an eye on sophisticated commit messages that properly explain the rationale. I propose something like this:
> 
> ```
> OrtMain: Allow to explicitly set the `warn` log level
> 
> Even if `warn` already is the default, wrappers around the ORT CLI might
> explicitly want to use `--warn` for convenience, e.g. because the log level
> is passed as-is from a UI form. So add support for `--warn.`
> ```
> 
> Would you mind amending your commit message accordingly?

done

sschuberth:
> Would you mind amending your commit message accordingly?

Hi @unclejay80, sorry, while I was omitting your `Signed-off-by` from my example, you of course still need to add it! Please amend your commit via `git commit --amend --no-edit -s` (for example) and force-push once more.

sschuberth (approved):

sschuberth:
Merging despite the unrelated `OsvServiceFunTest` failure.

sschuberth:
Thanks for your contribution @unclejay80!

PR #5807, opened 2022-09-16
Fix `OrtResult.collectIssues()`

fviernau:
See individual commits.

### DIFF
--- a/model/src/main/kotlin/AdvisorRecord.kt
+++ b/model/src/main/kotlin/AdvisorRecord.kt
@@ -73,7 +73,9 @@ data class AdvisorRecord(
 
         advisorResults.forEach { (id, results) ->
             results.forEach { result ->
-                collectedIssues.getOrPut(id) { mutableSetOf() } += result.summary.issues
+                if (result.summary.issues.isNotEmpty()) {
+                    collectedIssues.getOrPut(id) { mutableSetOf() } += result.summary.issues
+                }
             }
         }
 
--- a/model/src/main/kotlin/AnalyzerRun.kt
+++ b/model/src/main/kotlin/AnalyzerRun.kt
@@ -41,4 +41,16 @@ data class AnalyzerRun(
     val environment: Environment,
     val config: AnalyzerConfiguration,
     val result: AnalyzerResult
-)
+) {
+    companion object {
+        /**
+         * A constant for an [AnalyzerRun] where all properties are empty.
+         */
+        @JvmField
+        val EMPTY = AnalyzerRun(
+            environment = Environment(),
+            config = AnalyzerConfiguration(),
+            result = AnalyzerResult.EMPTY
+        )
+    }
+}
--- a/model/src/main/kotlin/ScanRecord.kt
+++ b/model/src/main/kotlin/ScanRecord.kt
@@ -53,7 +53,9 @@ data class ScanRecord(
 
         scanResults.forEach { (id, results) ->
             results.forEach { result ->
-                collectedIssues.getOrPut(id) { mutableSetOf() } += result.summary.issues
+                if (result.summary.issues.isNotEmpty()) {
+                    collectedIssues.getOrPut(id) { mutableSetOf() } += result.summary.issues
+                }
             }
         }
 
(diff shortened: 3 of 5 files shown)


model/src/main/kotlin/AnalyzerRun.kt
```diff
@@ -41,4 +41,18 @@ data class AnalyzerRun(
...
-)
+) {
+    companion object {
+        /**
+         * A constant for an [AnalyzerResult] where all properties are empty.
+         */
+        @JvmField
+        val EMPTY = AnalyzerRun(
+            startTime = Instant.MIN,
+            endTime = Instant.MIN.plusSeconds(1),
```
sschuberth, line 52: In the past, we used `Instant.EPOCH` as "dummy" start / end times. Shouldn't we stick to that convention here, too? (I believe it's fine to set start and end time to the same value here.)

model/src/main/kotlin/AdvisorRecord.kt
```diff
@@ -80,7 +80,9 @@ data class AdvisorRecord(
 
         advisorResults.forEach { (id, results) ->
             results.forEach { result ->
-                collectedIssues.getOrPut(id) { mutableSetOf() } += result.summary.issues
+                if (result.summary.issues.isNotEmpty()) {
```
sschuberth, line 83: Commit message nit: No comma after "Both".

model/src/main/kotlin/ScanRecord.kt
```diff
@@ -53,7 +53,9 @@ data class ScanRecord(
 
         scanResults.forEach { (id, results) ->
             results.forEach { result ->
-                collectedIssues.getOrPut(id) { mutableSetOf() } += result.summary.issues
+                if (result.summary.issues.isNotEmpty()) {
+                    collectedIssues.getOrPut(id) { mutableSetOf() } += result.summary.issues
+                }
```
sschuberth, line 58: Only semi-related: We should probably add a new test somewhere (but I don't know yet where) that verifies that the values returned by `OrtResult.collectIssues()` are never empty. But that's out of the scope of this PR.

sschuberth (approved):

model/src/main/kotlin/ScanRecord.kt
sschuberth: > but I don't know yet where

Probably in `OrtResultTest`.

fviernau: Hm, I just noticed that `AdvisorRecordTest` expects `collectIssues()` to return empty lists. 
So, this seems intended in the advisor.

sschuberth: > So, this seems intended in the advisor.

I believe I had discussed this with @oheger-bosch once; IIRC he was in favor of keeping the empty lists as the caller can then rely on every `id` to be present as a key, but I was in favor of removing it, to more quickly be able to see which `id`s don't have issues.

Also, I already checked that for the `AnalyerResult` also only non-empty issues are added, so I'd prefer to align on that. Any more thoughts about this @oheger-bosch?

fviernau: @sschuberth @oheger-bosch, I've updated the test to the new behavior.

sschuberth: @oheger-bosch will only be back on Monday to chime in to this discussion.

mnonnenmacher: > > So, this seems intended in the advisor.
> 
> I believe I had discussed this with @oheger-bosch once; IIRC he was in favor of keeping the empty lists as the caller can then rely on every `id` to be present as a key, but I was in favor of removing it, to more quickly be able to see which `id`s don't have issues.
> 
> Also, I already checked that for the `AnalyerResult` also only non-empty issues are added, so I'd prefer to align on that. Any more thoughts about this @oheger-bosch?

I would also prefer that, because the benefit of being able to use `collectIssues().getValue(id).isEmpty()` instead of `collectIssues()[id]?.isEmpty() == true` is minimal. And the first relies on the context knowledge that the map really contains keys for all packages which is already wrong if `skipExcluded` is set to `true` and there are excluded packages.

fviernau: Or even `id !in collectIssues().keys()`.

oheger-bosch: These are good arguments, and I have no strong feelings about it. So, no objections from my side.

sschuberth (approved):

PR #5808, opened 2022-09-16
AdvisorRecord: Drop deserialization backwards compatibility

fviernau:
The old format has become legacy in march 2021. So drop the support for deserializing it to simplify the code. The value of such old advisor data is questionable anyway.

Also note that vulnerabilities also had a legacy format from 2021 for which the support already has been dropped.

[1] https://github.com/oss-review-toolkit/ort/commit/d297c58f8afe76fd6567ec6fda610b2461c19b1e

### DIFF
--- a/model/src/main/kotlin/AdvisorRecord.kt
+++ b/model/src/main/kotlin/AdvisorRecord.kt
@@ -22,12 +22,6 @@ package org.ossreviewtoolkit.model
 
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
-import com.fasterxml.jackson.core.JsonParser
-import com.fasterxml.jackson.core.JsonToken
-import com.fasterxml.jackson.databind.DeserializationContext
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize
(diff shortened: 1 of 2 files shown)


sschuberth (changes requested):
Our inline unused import hints now work again 😁

sschuberth (approved):
I'm fine with this, but probably someone from @oss-review-toolkit/bosch should also approve.

model/src/main/kotlin/AdvisorRecord.kt
```diff
@@ -22,12 +22,6 @@ package org.ossreviewtoolkit.model
 
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
-import com.fasterxml.jackson.core.JsonParser
```
mnonnenmacher, line 25: Commit message nit: "deserialization" is usually written without hyphen.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

fviernau:
Merge despite the failing unrelated test.

fviernau:
Added the "breaking" label as I'm only 99% certain that this is not breaking (because it has been broken before already, see linked commit).

PR #5809, opened 2022-09-16
AdvisorRecord: Some test improvements

fviernau:
Reduce some test setup, reduce some repetition and eliminate some magic values.

### DIFF
--- a/model/src/test/kotlin/AdvisorRecordTest.kt
+++ b/model/src/test/kotlin/AdvisorRecordTest.kt
@@ -47,7 +47,7 @@ class AdvisorRecordTest : WordSpec({
             val record = advisorRecordOf(
                 queryId to listOf(
                     createResult(
-                        2,
+                        1,
                         issues = listOf(OrtIssue(source = "Advisor", message = "Failure"))
                     )
                 )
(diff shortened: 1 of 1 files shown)


model/src/test/kotlin/AdvisorRecordTest.kt
```diff
@@ -36,30 +36,22 @@ import org.ossreviewtoolkit.utils.common.enumSetOf
 class AdvisorRecordTest : WordSpec({
     "hasIssues" should {
         "return false given the record has no issues" {
-            val vul1 = createVulnerability("CVE-2021-1")
-            val vul2 = createVulnerability("CVE-2021-2")
             val record = AdvisorRecord(
                 sortedMapOf(
-                    langId to listOf(createResult(1)),
-                    queryId to listOf(createResult(1, vulnerabilities = listOf(vul1, vul2)))
+                    langId to listOf(createResult(1))
```
sschuberth: Commit message:
- Spell "misleading" as one word.
- "This makes it" -> maybe "That makes this test"

model/src/test/kotlin/AdvisorRecordTest.kt
```diff
@@ -37,7 +37,7 @@ class AdvisorRecordTest : WordSpec({
     "hasIssues" should {
         "return false given the record has no issues" {
             val record = advisorRecordOf(
-                langId to listOf(createResult(1))
+                langId to listOf(createResult())
```
sschuberth, line 40: Commit message: "Use a default value of 1 for the index in tests for which its value does not matter." (This better explains that you're acutally introducing a default value and not using an existing one.)

sschuberth (changes requested):

sschuberth (approved):

PR #5814, opened 2022-09-16
Rescripts not found fix

sschuberth:

### DIFF
--- a/reporter-web-app/build.gradle.kts
+++ b/reporter-web-app/build.gradle.kts
@@ -50,7 +50,7 @@ tasks.addRule("Pattern: yarn<Command>") {
             commandLine = listOf(nodeExecutable.path, yarnJs.path, command)
 
             // Prepend the directory of the bootstrapped Node.js to the PATH environment.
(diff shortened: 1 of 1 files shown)


reporter-web-app/build.gradle.kts
```diff
@@ -53,6 +53,8 @@ tasks.addRule("Pattern: yarn<Command>") {
             val newPath = listOf(
                 // Prepend the directory of the bootstrapped Node.js to the PATH environment.
                 nodeBinDir.path,
+                // Prepend the directory of additional tools like "rescripts" to the PATH environment.
+                projectDir.resolve("node_modules/.bin").path,
```
fviernau, line 57: Would it make sense to use `.absolutePath` instead of just `.path`? (Same for previous commit)

fviernau (approved):

reporter-web-app/build.gradle.kts
sschuberth: Gradle's `projectDir` is already absolute, so there's no need for that.

PR #5817, opened 2022-09-17
VcsInfo: Remove resolved revision

mnonnenmacher:

### DIFF
--- a/model/src/main/kotlin/VcsInfo.kt
+++ b/model/src/main/kotlin/VcsInfo.kt
@@ -105,8 +105,7 @@ fun VcsInfo?.orEmpty(): VcsInfo = this ?: VcsInfo.EMPTY
 private class VcsInfoDeserializer : StdDeserializer<VcsInfo>(VcsInfo::class.java) {
     companion object {
         val KNOWN_FIELDS by lazy {
-            VcsInfo::class.memberProperties.map { PROPERTY_NAMING_STRATEGY.translate(it.name) } +
-                    PROPERTY_NAMING_STRATEGY.translate("resolvedRevision")
+            VcsInfo::class.memberProperties.map { PROPERTY_NAMING_STRATEGY.translate(it.name) }
         }
     }
 
@@ -123,9 +122,7 @@ private class VcsInfoDeserializer : StdDeserializer<VcsInfo>(VcsInfo::class.java
         return VcsInfo(
             VcsType(node["type"].textValueOrEmpty()),
             node["url"].textValueOrEmpty(),
-            // For backward compatibility, if resolved_revision is set, prefer it over revision because it is more
-            // specific.
-            node["resolved_revision"]?.textValue() ?: node["revision"].textValueOrEmpty(),
+            node["revision"].textValueOrEmpty(),
             node["path"].textValueOrEmpty()
         )
     }
--- a/model/src/main/kotlin/VcsInfo.kt
+++ b/model/src/main/kotlin/VcsInfo.kt
@@ -79,10 +79,10 @@ data class VcsInfo(
         if (this == EMPTY) return other
 
         return VcsInfo(
-            type.takeUnless { it == EMPTY.type } ?: other.type,
-            url.takeUnless { it == EMPTY.url } ?: other.url,
-            revision.takeUnless { it == EMPTY.revision } ?: other.revision,
-            path.takeUnless { it == EMPTY.path } ?: other.path
+            type = type.takeUnless { it == EMPTY.type } ?: other.type,
+            url = url.takeUnless { it == EMPTY.url } ?: other.url,
+            revision = revision.takeUnless { it == EMPTY.revision } ?: other.revision,
+            path = path.takeUnless { it == EMPTY.path } ?: other.path
         )
     }
 
@@ -120,10 +120,10 @@ private class VcsInfoDeserializer : StdDeserializer<VcsInfo>(VcsInfo::class.java
         }
 
         return VcsInfo(
-            VcsType(node["type"].textValueOrEmpty()),
-            node["url"].textValueOrEmpty(),
-            node["revision"].textValueOrEmpty(),
-            node["path"].textValueOrEmpty()
+            type = VcsType(node["type"].textValueOrEmpty()),
+            url = node["url"].textValueOrEmpty(),
+            revision = node["revision"].textValueOrEmpty(),
+            path = node["path"].textValueOrEmpty()
         )
     }
 }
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/VcsInfo.kt
```diff
@@ -105,8 +105,7 @@ fun VcsInfo?.orEmpty(): VcsInfo = this ?: VcsInfo.EMPTY
 private class VcsInfoDeserializer : StdDeserializer<VcsInfo>(VcsInfo::class.java) {
     companion object {
         val KNOWN_FIELDS by lazy {
-            VcsInfo::class.memberProperties.map { PROPERTY_NAMING_STRATEGY.translate(it.name) } +
-                    PROPERTY_NAMING_STRATEGY.translate("resolvedRevision")
+            VcsInfo::class.memberProperties.map { PROPERTY_NAMING_STRATEGY.translate(it.name) }
```
sschuberth: BTW, do you remember why we introduced the check for unknown properties via `KNOWN_FIELDS`? Isn't it Jackson's default to fail on unknown properties if `@JsonIgnoreProperties(ignoreUnknown = true)` is not specified?

model/src/main/kotlin/VcsInfo.kt
```diff
@@ -79,10 +79,10 @@ data class VcsInfo(
         if (this == EMPTY) return other
 
         return VcsInfo(
-            type.takeUnless { it == EMPTY.type } ?: other.type,
-            url.takeUnless { it == EMPTY.url } ?: other.url,
-            revision.takeUnless { it == EMPTY.revision } ?: other.revision,
-            path.takeUnless { it == EMPTY.path } ?: other.path
+            type = type.takeUnless { it == EMPTY.type } ?: other.type,
```
sschuberth, line 82: Nit: The commit message could provide a rationale... is readability really and issue here, as each line already has the field name?

model/src/test/kotlin/VcsInfoTest.kt
```diff
@@ -100,7 +99,7 @@ class VcsInfoTest : WordSpec({
 
             with(exception) {
                 propertyName shouldBe "unknown"
-                knownPropertyIds should containAll<Any>("type", "url", "revision", "resolved_revision", "path")
+                knownPropertyIds should containAll<Any>("type", "url", "revision", "path")
```
sschuberth, line 102: Any idea why the test still passed? Again a Gradle task caching issue?

sschuberth: Ah, scratch that, I had overlooked that `resolved_revision` was still present a few lines above in the test data.

model/src/main/kotlin/VcsInfo.kt
sschuberth: I just verified that the `"fail if the input contains unknown fields"` test actually passes even if the `VcsInfoDeserializer` is removed. So maybe you could remove it completely as part of this PR?

model/src/main/kotlin/VcsInfo.kt
```diff
@@ -123,9 +122,7 @@ private class VcsInfoDeserializer : StdDeserializer<VcsInfo>(VcsInfo::class.java
         return VcsInfo(
             VcsType(node["type"].textValueOrEmpty()),
             node["url"].textValueOrEmpty(),
-            // For backward compatibility, if resolved_revision is set, prefer it over revision because it is more
-            // specific.
-            node["resolved_revision"]?.textValue() ?: node["revision"].textValueOrEmpty(),
+            node["revision"].textValueOrEmpty(),
```
fviernau: Would it make sense to fail hard if user is still using the old (now unhandled) format? That would be nice, in particular if it was possible to remove the custom serializer as per @sschuberth 's comment above.

fviernau (changes requested):

sschuberth:
@mnonnenmacher my proposal would be to drop "VcsInfo: Make KNOWN_FIELDS a set" (and maybe "VcsInfo: Use named arguments") from this PR, and then cherry-pick @fviernau's single commit from https://github.com/oss-review-toolkit/ort/pull/5827. That would make it nicely clear from the Git history why the custom deserializer can be removed.

fviernau:
> @mnonnenmacher my proposal would be to drop "VcsInfo: Make KNOWN_FIELDS a set" (and maybe "VcsInfo: Use named arguments") from this PR, and then cherry-pick @fviernau's single commit from #5827. That would make it nicely clear from the Git history why the custom deserializer can be removed.

@sschuberth, doesn't my commit message in #5827 not make it sufficiently clear? e.g. backwards compatibility for that time (first part of 2021) has been broken anyway by recent changes. Partial backwards compatibility is of no value as deserialization fails anyway. What am I missing?

sschuberth:
> What am I missing?

To me, @mnonnenmacher's `resolved_revision`-related commits in this PR are somewhat fixups / follow-ups to abd41579bd1e882870350dbcc7db781e22ec5c56 only, and having these committed separately nicely creates an intermediate state with

        val KNOWN_FIELDS by lazy {
            VcsInfo::class.memberProperties.map { PROPERTY_NAMING_STRATEGY.translate(it.name) }
        }

which makes it super-clear that there is no value in keeping the custom deserializer. This way the Git history simply tells a nicer story, IMO. But I'm also ok with just superseding this PR with yours, if @mnonnenmacher agrees.

mnonnenmacher:
@sschuberth @fviernau This change is not super urgent, why couldn't you just have waited a few days to give me a chance to respond to the initial code review instead of creating a separate PR and starting this discussion? I think this is just a waste of time for everyone involved.

With that being said, I think the reason for the custom deserializer was also to allow only a single value of `VcsInfo` while not providing default values in code. @fviernau Would this still work after your PR, e.g. to set only the VCS revision in a curation?

fviernau:
> @sschuberth @fviernau This change is not super urgent, why couldn't you just have waited a few days to give me a chance to respond to the initial code review instead of creating a separate PR and starting this discussion? I think this is just a waste of time for everyone involved.

That was unintended. When I created my PR I didn't recall that it overlaps with your PR. 
Only @sschuberth brought this to my attention. I decided to not close my PR, as in my view I could not see anything wrong with it and thought it could be merged as-is to save time for everyone involved in this discussion here. If that's not the case then let's just ditch my PR, as the last thing I wanted with this is wasting your time.

> With that being said, I think the reason for the custom deserializer was also to allow only a single value of `VcsInfo` while not providing default values in code. @fviernau Would this still work after your PR, e.g. to set only the VCS revision in a curation?

`PackageCurationData` is using `VcsInfoCurationData`, not `VcsInfo`. So, the requirement to work with not present values is not existing to my understanding. Am I missing something?

mnonnenmacher:
> `PackageCurationData` is using `VcsInfoCurationData`, not `VcsInfo`. So, the requirement (from curations) to work with absent values is not existing to my understanding. Am I missing something?

I just remembered that this was the initial reason for the deserializer, maybe it could already have been removed in 550e208 a long time ago. So yes, go ahead and merge your PR.

mnonnenmacher:
Obsolete with #5827.

PR #5819, opened 2022-09-19
Hash: Drop support for deserializing the legacy format

fviernau:
The model for the `Hash` class has been changed in 2019.
Use the new format throughout the code base and drop handling of the legacy format to reduce technical debt.

### DIFF
--- a/integrations/schemas/curations-schema.json
+++ b/integrations/schemas/curations-schema.json
@@ -50,30 +50,10 @@
               },
               "hash": {
                 "$ref": "#/definitions/hash"
-              },
-              "hash_algorithm": {
-                "type": "string"
               }
-            },
-            "if": {
-              "properties": {
-                "hash": {
-                  "type": "object"
-                }
-              }
-            },
-            "then": {
-              "required": [
-                "hash"
-              ]
-            },
-            "else": {
-              "required": [
-                "hash",
-                "hash_algorithm"
-              ]
             },
             "required": [
+              "hash",
               "url"
             ]
           },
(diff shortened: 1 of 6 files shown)


integrations/schemas/curations-schema.json
```diff
@@ -85,30 +65,10 @@
...
-              ]
-            },
-            "else": {
-              "required": [
-                "hash",
-                "hash_algorithm"
-              ]
-            },
             "required": [
+              "hash",
```
sschuberth, line 71: I believe this commit should also drop the `"string"` type from https://github.com/oss-review-toolkit/ort/blob/ca7beff23342d5c1b9a2bdbaa6c79c9e6cadadeb/integrations/schemas/curations-schema.json#L133.

sschuberth (changes requested):

sschuberth (dismissed):

fviernau:
@oss-review-toolkit/bosch, @bennati is this ok to merge from Bosch and HERE side?

nnobelis:
> @oss-review-toolkit/bosch, @bennati is this ok to merge from Bosch and HERE side?

Hey @fviernau I found some usage of the old syntax by us. Let me fix them first please, I will then give you an heads-up.

bennati:
please go ahead

fviernau:
> Hey @fviernau I found some usage of the old syntax by us. Let me fix them first please, I will then give you an heads-up.

Thanks @nnobelis !

mnonnenmacher (approved):

mnonnenmacher:
@fviernau Fixes on our side are done.

MarcelBochtler:
Unfortunately, this breaks all projects, which use the legacy format in their `.ort.yml` :/

MarcelBochtler:
Additionally, the JsonSchema for the curations still shows the legacy format as valid.

mnonnenmacher:
I guess the problem is that the documentation was not updated long ago when the format was changed in 2019.

PR #5820, opened 2022-09-19
ScanRecord: Drop support for deserializing the legacy format

fviernau:
The format has become legacy 1.5 years ago, see [1].

[1] https://github.com/oss-review-toolkit/ort/commit/17bdbe6c77e673fac84eddf0854bd263bbf3bf3a


cli/src/funTest/assets/semver4j-analyzer-result.yml
```diff
@@ -13,8 +13,8 @@ repository:
   config:
     excludes:
       scopes:
-      - pattern: "testClasspath"
-        reason: "TEST_TOOL_OF"
+        - pattern: "testClasspath"
```
sschuberth, line 16: I'm not sure whether we should align in *that* direction: While I'm aware that many YAML examples do have the leading spaces for [block sequence](https://yaml.org/spec/1.2.2/#821-block-sequences) entries, our serialized files actually seem not to use them...

sschuberth: BTW, "consistent" to what? Other block sequences in the same file, or to other files?

cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -134,7 +134,7 @@ class ExamplesFunTest : StringSpec() {
         }
 
         "example.rules.kts can be compiled and executed" {
-            val resultFile = File("src/funTest/assets/semver4j-analyzer-result.yml")
+            val resultFile = File("src/funTest/assets/semver4j-advisor-result.yml")
```
sschuberth, line 137: As the scanner could also be run after the advisor, using "advisor" does not really make clear that a scanner result is included. Maybe be super-explicit and name it `semver4j-analyzer-scanner-advisor-result.yml`, or just `semver4j-ort-result.yml`?

sschuberth (changes requested):

cli/src/funTest/assets/semver4j-analyzer-result.yml
fviernau: > I'm not sure whether we should align in that direction: While I'm aware that many YAML examples do have the leading spaces for [block sequence](https://yaml.org/spec/1.2.2/#821-block-sequences) entries, our serialized files actually seem not to use them...

As most part's of the file have been using that identation before, I believe it's ok to fix the few inconsistencies.
I agree that it should be consistent also with other files, but in scope of this PR I believe local consistency is ok.

> BTW, "consistent" to what? Other block sequences in the same file, or to other files?

local consistency.

Note: I was planning to make an upcoming change which aligns the formatting with other files in a separate PR.

cli/src/funTest/kotlin/ExamplesFunTest.kt
fviernau: I think it would be valuable to assume an order, at least for the test assets in order to not making our lives unnecessarily hard finding names for files. Changed according to your comment regardlessly.

sschuberth (approved):

fviernau:
@benati , @oheger-bosch  it'd be great to get your ACK for this as well.

cli/src/funTest/assets/semver4j-analyzer-result.yml
mnonnenmacher: If you plan to align with the "correct" formatting anyway, why do this commit in the first place? I don't see the benefit over just fixing it right away.

fviernau: The otherway around it is way more effort, and now the problem is already fixed locally. I don't see the benefit of discussing this any further TBH, it is making the formatting consistent within the file. So, what's your aim in continuing this discussion?

nnobelis:
> @bennati, @oss-review-toolkit/bosch it'd be great to get your ACK for this as well. @bennati

I checked for Bosch and it seems we are not affected by this breaking change. Also for use this PR can be merged.

sschuberth (approved):

fviernau:
Got thumbs up from HERE as well, so merging it now.

PR #5823, opened 2022-09-19
ProjectSourceRule: Add a matcher for the project's VCS type

fviernau:
See individual commits.

Part of https://github.com/oss-review-toolkit/ort/issues/5621.


tsteenbe (changes requested):
nit "The matcher can be used to enable rule only for certain VCS types." change to "The matcher can be used to enable a rule only for certain VCS types."

tsteenbe (approved):

sschuberth (approved):

PR #5825, opened 2022-09-19
Minor `ToStringSerializer` improvements

sschuberth:


clients/clearly-defined/src/main/kotlin/Serializers.kt
```diff
@@ -32,6 +32,9 @@ import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
 import kotlinx.serialization.json.jsonPrimitive
 
+/**
+ * A (de-)serializer for ClearlyDefined [Coordinates] instances to / from strings, or from a nested structure.
```
mnonnenmacher, line 36: Very nit pick: If you read "(de-)serializer" as "deserialzer/serializer" it should  be "from/to" instead of "to/from".

fviernau (approved):

PR #5826, opened 2022-09-19
DependencyRule: Add a helper function to get the direct dependencies

sschuberth:
Some rules might only need to look at the direct dependencies.

### DIFF
--- a/evaluator/src/main/kotlin/DependencyRule.kt
+++ b/evaluator/src/main/kotlin/DependencyRule.kt
@@ -78,7 +78,8 @@ class DependencyRule(
     fun directDependencies() = dependency.visitDependencies { it.toList() }
 
     /**
-     * A [RuleMatcher] that checks if the level of the [dependency] inside the dependency tree equals [level].
+     * A [RuleMatcher] that checks if the level of the [dependency] inside the dependency tree equals [level], which
+     * starts with 0 for a direct dependency.
      */
     fun isAtTreeLevel(level: Int) =
         object : RuleMatcher {
--- a/evaluator/src/main/kotlin/DependencyRule.kt
+++ b/evaluator/src/main/kotlin/DependencyRule.kt
@@ -73,6 +73,11 @@ class DependencyRule(
         "$name - ${pkg.id.toCoordinates()} (dependency of ${project.id.toCoordinates()} in scope $scopeName)"
 
     /**
+     * Return the direct dependencies of the [dependency].
+     */
+    fun directDependencies() = dependency.visitDependencies { it.toList() }
+
+    /**
      * A [RuleMatcher] that checks if the level of the [dependency] inside the dependency tree equals [level].
      */
     fun isAtTreeLevel(level: Int) =


evaluator/src/main/kotlin/DependencyRule.kt
```diff
@@ -72,6 +72,11 @@ class DependencyRule(
     override fun issueSource() =
         "$name - ${pkg.id.toCoordinates()} (dependency of ${project.id.toCoordinates()} in scope $scopeName)"
 
+    /**
+     * Return the direct dependencies of the [dependency].
+     */
+    fun dependencies() = dependency.visitDependencies { it.toList() }
```
fviernau, line 78: I don't understand how this is supposed to be used.
There is already the matcher `DependencyRule.isAtTreeLevel` which can be used to limit a rule to 
a certain tree level, e.g. direct dependencies. How is the use case different from that one?

fviernau (changes requested):

evaluator/src/main/kotlin/DependencyRule.kt
sschuberth: > How is the targeted use case different from that one?

I don't want to iterate over all dependencies and filter those at level 0 only, but straight away only get the direct dependencies for efficiency.

> Yes, could you provide an example use case for a rule that would use this?

This commit was extracted from https://github.com/oss-review-toolkit/ort/pull/4977 where also an example use is given [here](https://github.com/oss-review-toolkit/ort/pull/4977/commits/12ebf5523f6a3ab285797a0d6e512a004ac82f32#diff-826eafbbb8a3949e640571ae685f52f399c36e9b39e552b45b6386305e57401aR86-R87).

> Also this should be called `directDependencies()` to be clear.

I'll do the renaming.

sschuberth: > > Yes, could you provide an example use case for a rule that would use this?
> 
> This commit was extracted from #4977 where also an example use is given [here](https://github.com/oss-review-toolkit/ort/pull/4977/commits/12ebf5523f6a3ab285797a0d6e512a004ac82f32#diff-826eafbbb8a3949e640571ae685f52f399c36e9b39e552b45b6386305e57401aR86-R87).

So the main point is: I don't want to check for the dependency I'm iterating on in a `DependencyRule` at which level *it* is, but get *its* direct dependencies.

sschuberth:
@fviernau or @mnonnenmacher can I please get another review for this small change?

fviernau:
Reading the above comments, in particular following the link provided [1] reveals that the ultimate goal of this PR is to provide an API which allows to write a rule which checks the OSADL matrix. So, in my view this PR should be reviewed against that, e.g.: is the API added suitable for that. Please correct me if I'm wrong with that @sschuberth.

When trying to do such review the following questions popped up

1. How is the compatibility check defined exactly, in particular
  - which licenses have to be checked for a particular dependency?
    Consider the dependency tree rooted at `a`with the following edges: `(a -> b), (b -> c1) (b -> c2)`.
    If you make the check for edge `(a -> b)` which licenses do need to be checked against the ones of `a`?
    Is it the licenses 
    - the licenses of (`b + c1 + c2`) or just the ones of `b`
  - Maybe a bit similar, is the check transitive?
  - @sschuberth, do you maybe have a reference where the algorithm for checking an edge is explained?
2. Efficiency / using `dependencyRule`.
   - If I have a dependency X which is contained in `N` scopes and appearing `M` times in each, how often
     is the `DepedencyRule` triggered for X. 
   - Put above differently, does the current implementation [1] perform the check for each distinct edge potentially more than once?
     If so, that would, apart from efficiency, also lead to redundant violations.
   - Would a dedicated rule type for checking distinct edges makes sense? (Whether this question makes sense at all depends
     on the answers for point 1. 

[1] https://github.com/oss-review-toolkit/ort/pull/4977/commits/12ebf5523f6a3ab285797a0d6e512a004ac82f32#diff-826eafbbb8a3949e640571ae685f52f399c36e9b39e552b45b6386305e57401aR86-R87

fviernau (approved):

sschuberth:
> the ultimate goal of this PR is to provide an API which allows to write a rule which checks the OSADL matrix. So, in my view this PR should be reviewed against that, e.g.: is the API added suitable for that. Please correct me if I'm wrong with that @sschuberth.

That's only partly correct: While indeed the need for this helper function first occurred when writing a rule that checks the OSADL matrix, I was actually surprised that we didn't have such a function until now, as I deem it to be valuable independently of the OSADL matrix rule.

In fact, it might be that a new iteration of the OSADL matrix rule PR will not require this function anymore, so I pulled it out to a separate PR as in my view the context of the OSADL matrix PR is not necessary here.

> *I didn't want to start a large discussion out of this small change. Even if the API would turn out to not work for the compliance checks, it could be useful for other use cases.*

Exactly. Thanks for confirming.

PR #5827, opened 2022-09-20
VcsInfo: Drop support for deserializing the legacy format

fviernau:
The last change to the format has been made in June 2021, see [1]. As support for deserializing legacy ORT formats dating back to about more than a year ago has been dropped in recent changes, remove also this deserializer to reduce technical debt.

Note: The deserializer had been introduced for assigning empty strings as default values to properties with absent values [2]. As the serializer always serializes all properties, that mechanism is not needed (anymore).

[1] https://github.com/oss-review-toolkit/ort/commit/abd41579bd1e882870350dbcc7db781e22ec5c56 
[2] https://github.com/oss-review-toolkit/ort/commit/66aa368e7cbddf94e090555ba115f92e9b8ebc45

Node: I've added the `breaking change` label as it removes deserializing legacy format. However, it is not really breaking as it's been broken already by preceeding changes.

### DIFF
--- a/model/build.gradle.kts
+++ b/model/build.gradle.kts
@@ -32,7 +32,6 @@ dependencies {
     api(libs.jacksonDataformatXml)
     api(libs.jacksonDataformatYaml)
 
-    implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation(libs.bundles.exposed)
     implementation(libs.bundles.hoplite)
     implementation(libs.hikari)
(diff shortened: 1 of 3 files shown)


sschuberth:
Please note that this conflicts with @mnonnenmacher's https://github.com/oss-review-toolkit/ort/pull/5817, where we already discussed the removal of the custom deserializer.

sschuberth:
This should now also drop `implementation("org.jetbrains.kotlin:kotlin-reflect")` from the build file.

sschuberth (approved):

PR #5830, opened 2022-09-20
Docker tool components

sschuberth:

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -112,7 +112,7 @@ jobs:
     - name: Validate Batect wrapper scripts
       uses: batect/batect-wrapper-validation-action@v0
     - name: Run functional tests
-      run: BATECT_QUIET_DOWNLOAD=true ./batect --enable-buildkit --config-var gradle_console=plain funTest -- jacocoFunTestReport -p analyzer
+      run: BATECT_QUIET_DOWNLOAD=true ./batect --enable-buildkit --config-var gradle_console=plain funTestAnalyzer
     - name: Upload code coverage data
       uses: codecov/codecov-action@v3
       with:
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@
 # Copyright (C) 2020 Bosch Software Innovations GmbH
 # Copyright (C) 2021-2022 Bosch.IO GmbH
 # Copyright (C) 2021 Alliander N.V.
+# Copyright (C) 2021-2022 BMW CarIT GmbH
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


mnonnenmacher:
@fviernau FYI, I have removed the Docker socket mount because the analyzer tests do not use testcontainers (https://www.testcontainers.org/supported_docker_environment/continuous_integration/dind_patterns/). I think it was temporarily required for the PostgreSQL tests when the `-p analyzer` argument was not correctly passed to the batect command in some previous iteration of the change and therefore the scanner tests were executed as well.

sschuberth:
I've retriggered the `funTest-analyzer-docker` which failed for reasons probably unrelated to this PR (network timeout).

docker/00-add_local_path.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
sschuberth, line 1: All shell scripts should be committed with mode 100755 (instead of 100644).

Dockerfile
```diff
@@ -3,6 +3,7 @@
 # Copyright (C) 2020 Bosch Software Innovations GmbH
 # Copyright (C) 2021-2022 Bosch.IO GmbH
 # Copyright (C) 2021 Alliander N.V.
+# Copyright (C) 2021-2022 BMW CarIT GmbH
```
sschuberth, line 6: > These "components" are installed in separate layers to benefit from Docker's caching if a
tool version changes, but others stay fixed.

I believe we should improve this bit of my own wording as say:

These "components" are installed in separate multi-images to benefit from Docker's parallel building of those, and the tool binaries are then copied to the final image. In contrast to a layered approach this avoids rebuilding if only a single tool version changes.

scripts/docker_build.sh
```diff
@@ -22,6 +22,7 @@ DOCKER_ARGS=$@
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
 GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty)
+ORT_DOCKER=${ORT_DOCKER:-ort}
```
sschuberth, line 25: This should either be dropped as we have that inline in line 30 below. Or we keep this line and just use `$ORT_DOCKER` in line 30. Such an "outlining" (opposite of "inlining") of the variable should then be done in a separate commit, though.

.github/workflows/build-and-test.yml
```diff
@@ -112,7 +112,7 @@ jobs:
     - name: Validate Batect wrapper scripts
       uses: batect/batect-wrapper-validation-action@v0
     - name: Run functional tests
-      run: BATECT_QUIET_DOWNLOAD=true ./batect --enable-buildkit --config-var gradle_console=plain funTest -- jacocoFunTestReport -p analyzer
+      run: BATECT_QUIET_DOWNLOAD=true ./batect --enable-buildkit --config-var gradle_console=plain funTestAnalyzer
```
sschuberth, line 115: Commit message nits:
- "Use a login shell as entrypoint" -> "Use a login shell as *the* entrypoint"
- "does not work anymore the pass additional arguments" -> "does not work anymore *to* the pass additional arguments"

batect.yml
```diff
@@ -85,12 +85,12 @@ tasks:
...
     description: >-
-      A task to execute ORT's functional tests in the runtime container. Can be limited to e.g. the "analyzer" project
-      like `./batect funTest -- -p analyzer`.
+      A task to execute ORT's functional tests for the analyzer in the runtime container.
     prerequisites:
       - buildFunTest
     run:
       container: run
-      command: ./gradlew funTest
+      command: "'./gradlew funTest jacocoFunTestReport -p analyzer'"
```
sschuberth, line 95: Could you amend you commit message about why this nested quoting with double *and* single-quotes is required? Have you verified that it does not work with either type of quoting only?

sschuberth: Ah, I just saw this is nicely explained at https://batect.dev/docs/reference/config/tasks/#command. Would you mind adding that link to the commit message?

sschuberth (commented):
I can't request changes on my own PR 😭

batect.yml
sschuberth: PS: I also raised the [question](https://github.com/batect/batect/discussions/1355) whether the current Batect behavior is intended. My hunch is that it's not, and Batect could be smarter about adding additional task arguments to inside of quotes in a `command`.

sschuberth (commented):
LGTM (I cannot approve a PR that I myself created). Please approve yourself @mnonnenmacher.

mnonnenmacher (approved):

sschuberth:
@nnobelis any objections to merge this?

tsteenbe (approved):

sschuberth:
> I am in favor of merging it Monday to handle the change on our side.

Ok, I'll ping you on Monday then again.

fviernau (approved):

nnobelis:
> > I am in favor of merging it Monday to handle the change on our side.
> 
> Ok, I'll ping you on Monday then again.

You can merge the PR: I am watching its state.

PR #5831, opened 2022-09-20
Miscellaneous small fixes

sschuberth:

### DIFF
--- a/model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
+++ b/model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
@@ -122,9 +122,9 @@ data class ResolvedLicenseInfo(
         licenseChoices: List<SpdxLicenseChoice>,
         licenseView: LicenseView = LicenseView.ALL
     ): ResolvedLicenseInfo {
-        val licenses = effectiveLicense(licenseView, licenseChoices)?.decompose().orEmpty()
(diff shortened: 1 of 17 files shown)


model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
```diff
@@ -122,9 +122,9 @@ data class ResolvedLicenseInfo(
         licenseChoices: List<SpdxLicenseChoice>,
         licenseView: LicenseView = LicenseView.ALL
     ): ResolvedLicenseInfo {
-        val licenses = effectiveLicense(licenseView, licenseChoices)?.decompose().orEmpty()
+        val effectiveLicenses = effectiveLicense(licenseView, licenseChoices)?.decompose().orEmpty()
```
mnonnenmacher, line 125: Unrelated renaming.

mnonnenmacher (changes requested):

model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
sschuberth: It's not unrelated: In order to be able to simplify `this.licenses`, the local `licenses` has to be renamed.

mnonnenmacher (approved):

PR #5844, opened 2022-09-21
model: Split out classes contained in `ResolvedLicenseInfo` to separate files

sschuberth:
This helps to get a better overview over the many classes.


model/src/main/kotlin/licenses/ResolvedCopyright.kt
```diff
@@ -0,0 +1,38 @@
+/*
```
fviernau, line 1: Is it worth turning these into a separate package @sschuberth @mnonnenmacher ?

fviernau (approved):

model/src/main/kotlin/licenses/ResolvedCopyright.kt
sschuberth: Let's see... maybe some more refactoring will be coming up from my side, and maybe that shows whether it's worth it.

Issue #5763, opened 2022-09-09
helper-cli: Provide a command to generate path excludes for package configurations


fviernau:
The creation of `PathExclude`s for package configurations is to some extend a repetitive task.

A `helper-cli` command should be added which automates this by generating path excludes which are **very likely
but not necessarily apply**. That command should assume that the output of the command always needs a manual review.

**Potential input**

1. file paths
  - Some directory name patterns can always be excluded
  - Some filename patterns can always be excluded
  - Some file extensions can always be excluded
  - For now the findings paths can be used, but ideally the command would use the full file (which will be added to OrtResult in the future)
2. existing package configurations for the same package

**Things to consider**

1. Granularity of the generated excludes:
   - fine grained
     - it's clear what is being excluded 
     - if too fine grained, may become hard to manage / review 
     - merging more fine grained ones into more coarse grained ones is relatively easy
   - coarse grained
     - fewer patterns are more easy to review, manage as long as it is clear what is being excluded 
     - if too coarse grained, may become unclear what is being excluded, so at least the review becomes unnecessarily hard
       rendering the command not valuable in the worst case
     - manually turning coarse grained patterns into finer grained ones isn't easy / quick
   - consequence
     - patterns should use wildcards and be as coarse grained as possible, but must certainly not be too coarse grained
     - how to archive that?
       - whether patterns can be merged / squashed can depend on the actual pattern
       - be careful with patterns which cross repository boundaries / submodules. It may make
         sense to disable merging patterns from sibling submodules
2. Only regard files with offending findings (optionally)
   - It seems to make sense to have separate toggle switches for filename patterns and directory name patterns.
3. Quality
  - The patterns need to be good from the start, because poor patterns commited to many files may create unnecessary maintenance overhead in the longer run. Typically, such command would improve incrementally by the users of it while creating path excludes. So it is difficult / impossible to avoid that problem fully, but it should be minimizied by developing it against a rather large data set of real world cases.

**Scope** 

Not all of the above has to be done in scope of this ticket, but a reasonable MVP.

PR #5845, opened 2022-09-21
PathExcludeGenerator: Enable filename specific excludes

fviernau:
See individual commits.

Part of #5763.

### DIFF
--- a/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
+++ b/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
@@ -37,10 +37,10 @@ internal object PathExcludeGenerator {
      * Return path excludes which likely but not necessarily apply to a source tree containing all given [filePaths]
      * which must be relative to the root directory of the source tree.
      */
-    fun generatePathExcludes(filePaths: Collection<String>): List<PathExclude> {
+    fun generatePathExcludes(filePaths: Collection<String>): Set<PathExclude> {
         val files = filePaths.mapTo(mutableSetOf()) { File(it) }
 
-        return generateExcludesForDirectories(files).toList()
+        return generateExcludesForDirectories(files)
     }
 
     private fun generateExcludesForDirectories(files: Set<File>): Set<PathExclude> {
--- a/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
+++ b/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
@@ -37,8 +37,13 @@ internal object PathExcludeGenerator {
      * Return path excludes which likely but not necessarily apply to a source tree containing all given [filePaths]
      * which must be relative to the root directory of the source tree.
      */
-    fun generatePathExcludes(filePaths: Collection<String>): Set<PathExclude> =
-        generateExcludesForDirectories(filePaths)
+    fun generatePathExcludes(filePaths: Collection<String>): Set<PathExclude> {
+        val directoryExcludes = generateExcludesForDirectories(filePaths)
+        val remainingFilePaths = filePaths.filterNot { filePath -> directoryExcludes.any { it.matches(filePath) } }
+        val fileExcludes = generateExcludesForFiles(remainingFilePaths)
+
+        return directoryExcludes + fileExcludes
+    }
 
     /**
      * Return path excludes matching entire directories which likely but not necessarily apply to a source tree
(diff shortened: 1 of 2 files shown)


helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -37,13 +37,15 @@ internal object PathExcludeGenerator {
...
-    }
+    fun generatePathExcludes(filePaths: Collection<String>): Set<PathExclude> =
+        generateExcludesForDirectories(filePaths)
 
-    private fun generateExcludesForDirectories(files: Set<File>): Set<PathExclude> {
+    /**
+     * Return path excludes matching entire directories which likely but not necessarily apply to a source tree
+     * containing all given [filePaths] which must be relative to the root directory of the source tree.
+     */
+    fun generateExcludesForDirectories(filePaths: Collection<String>): Set<PathExclude> {
```
sschuberth, line 52: Unrelated, but I'm mentioning it anyway: While re-reading the function name, it occurred to me that "...ForDirectories" sounds as if it expects `filePath` to point to directories. But instead what's meant is that the generated *excludes* refer to directories. So maybe `generateDirectoryExcludes` would be a better function name? That would also nicely complement `generatePathExcludes` with "Path" being the umbrella term for "Directory" and "File".

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -71,6 +76,43 @@ internal object PathExcludeGenerator {
...
+            pathExcludes += patterns.map { PathExclude(it, reason) }
+        }
+
+        val filesForPathExcludes = pathExcludes.associateWith { pathExcludeExclude ->
+            files.filter { pathExcludeExclude.matches(it.path) }.toSet()
+        }
+
+        return greedySetCover(filesForPathExcludes).toSet()
+    }
+    internal fun createExcludePatterns(filenamePattern: String, files: Set<File>): Set<String> {
```
sschuberth, line 96: Missing blank line.

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -71,6 +76,43 @@ internal object PathExcludeGenerator {
 
         return result
     }
+
+    private fun generateExcludesForFiles(filePaths: Collection<String>): Set<PathExclude> {
```
sschuberth, line 80: Similar to my previous comment, maybe name this `generateFileExcludes`.

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -71,6 +76,43 @@ internal object PathExcludeGenerator {
...
+            } else {
+                filenamePattern
+            },
+            matchSubdirectories = matchingFiles.distinctBy { it.parentFile ?: File("") }.size > 1
+        ).let { setOf(it) }
+    }
+
+    private fun createExcludePattern(directory: String, filenamePattern: String, matchSubdirectories: Boolean): String {
+        val dir = "".takeIf { directory.isEmpty() } ?: "$directory/"
+        val wildcard = "**/".takeIf { matchSubdirectories } ?: ""
```
sschuberth, line 113: Use `.orEmpty()`.

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -71,6 +76,43 @@ internal object PathExcludeGenerator {
...
+                matchingFiles.first().name
+            } else {
+                filenamePattern
+            },
+            matchSubdirectories = matchingFiles.distinctBy { it.parentFile ?: File("") }.size > 1
+        ).let { setOf(it) }
+    }
+
+    private fun createExcludePattern(directory: String, filenamePattern: String, matchSubdirectories: Boolean): String {
+        val dir = "".takeIf { directory.isEmpty() } ?: "$directory/"
```
sschuberth, line 112: directory.takeIf { it.isEmpty() } ?: "$directory/"

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -92,6 +134,20 @@ private fun File.getAncestorFiles(): List<File> {
     return result
 }
 
+private fun getClosestCommonAncestor(files: Collection<File>): File {
```
sschuberth, line 137: I admit I haven't carefully check what *exactly* this does, but it sounds like our existing `getCommonFileParent` utility function, or?

sschuberth (changes requested):

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
fviernau: I just tried out `getCommonFileParent()`. It doesn't work as-is, because it turns the paths into absolute paths.
The functions in this file only use relative paths.

sschuberth: I can look into changing `getCommonFileParent()` in that regard, then we could eventually simplify this later.

sschuberth (approved):

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
fviernau: Thanks!

PR #5846, opened 2022-09-21
Apply some minor `PathExcludeGenerator` improvements

sschuberth:

### DIFF
--- a/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
+++ b/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
@@ -68,7 +68,7 @@ internal object PathExcludeGenerator {
         dirsToExclude.forEach { (dir, reason) ->
             if (dir.getAncestorFiles().intersect(dirsToExclude.keys).isEmpty()) {
                 result += PathExclude(
-                    pattern = "${dir.path}/**",
+                    pattern = "${dir.invariantSeparatorsPath}/**",
                     reason = reason
                 )
             }
@@ -99,7 +99,7 @@ internal object PathExcludeGenerator {
             ?: return emptySet()
 
         return createExcludePattern(
-            directory = getClosestCommonAncestor(matchingFiles).path,
+            directory = getClosestCommonAncestor(matchingFiles).invariantSeparatorsPath,
             filenamePattern = if (matchingFiles.distinctBy { it.name }.size == 1) {
                 matchingFiles.first().name
             } else {
--- a/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
+++ b/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
@@ -27,6 +27,7 @@ import org.ossreviewtoolkit.model.config.PathExcludeReason.BUILD_TOOL_OF
 import org.ossreviewtoolkit.model.config.PathExcludeReason.DOCUMENTATION_OF
 import org.ossreviewtoolkit.model.config.PathExcludeReason.TEST_OF
 import org.ossreviewtoolkit.utils.common.FileMatcher
+import org.ossreviewtoolkit.utils.common.getCommonParentFile
 import org.ossreviewtoolkit.utils.common.getDuplicates
 
 /**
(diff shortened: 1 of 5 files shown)


helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -68,7 +68,7 @@ internal object PathExcludeGenerator {
         dirsToExclude.forEach { (dir, reason) ->
             if (dir.getAncestorFiles().intersect(dirsToExclude.keys).isEmpty()) {
                 result += PathExclude(
-                    pattern = "${dir.path}/**",
+                    pattern = "${dir.invariantSeparatorsPath}/**",
```
fviernau, line 72: The same bug should exist in line 102.
Could you fix this one as well?

fviernau (changes requested):

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -99,7 +100,7 @@ internal object PathExcludeGenerator {
             ?: return emptySet()
 
         return createExcludePattern(
-            directory = getClosestCommonAncestor(matchingFiles).path,
+            directory = getCommonParentFile(matchingFiles).path,
```
fviernau, line 103: This breaks one of the unit test in PathExcludeGeneratorTest (at least on Linux).

fviernau (changes requested):

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -51,14 +52,12 @@ internal object PathExcludeGenerator {
...
-        val dirs = getAllDirectories(files)
-
-        val dirsToExclude = mutableMapOf<File, PathExcludeReason>()
+        val dirs = getAllDirectories(filePaths)
+        val dirsToExclude = mutableMapOf<String, PathExcludeReason>()
 
         dirs.forEach { dir ->
             val (_, reason) = PATH_EXCLUDES_REASON_FOR_DIR_NAME.find { (pattern, _) ->
-                FileMatcher.match(pattern, dir.name, ignoreCase = true)
+                FileMatcher.match(pattern, dir.substringAfterLast('/'), ignoreCase = true)
```
fviernau, line 60: I find it a bit more convenient to have the helper functions work with `Files` and also more consistent.
Why are you changing these to use `String`? e.g. turn `dir.name` into `dir.substringAfterLast('/')` ?

Would you be ok dropping this commit from this PR and maybe do that a bit later?
I have more changes in my stash, but I'm not yet sure how code will look in the end,
e.g. if there will be further calls to File.getAncestorFiles(), but I think there is a good chance.

sschuberth: > I find it a bit more convenient to have the helper functions work with `Files` and also more consistent.

I basically agree, but actually your calling code is operating on strings and your mapping to files, so using the existing function that works with strings actually makes the code simpler.

> Why are you changing these to use `String`?

Note that I'm not changing the existing helper function to take a string, but my goal was to reuse the existing string-based function.

> e.g. turn `dir.name` into `dir.substringAfterLast('/')` ?

I agree this bit looks uglier than before, but IMO it's worth it for the sake of the other simplifications / code deduplications.

> if there will be further calls to File.getAncestorFiles(), but I think there is a good chance.

If your upcoming code would indeed look much uglier with the string-version I'd instead consider to rewrite the common `getAllAncestorDirectories()` utility function to operate on files. There is just one single other called of it in the `RootLicenseMatcher`. And we should really try to not duplicate so much code and work.

fviernau: Could we then please split out this single (in my view controversial commit) into a separate PR, so that I continue my work while we sort this one out?

sschuberth: I've dropped the last commit.

PS: These refactoring would not be necessary if you did a more thorough research of what functions we already have in the first place 😉

fviernau (approved):

PR #5851, opened 2022-09-22
SCANOSS test fixes

sschuberth:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -20,7 +20,8 @@ reporter/src/funTest/assets/*-expected-*	text eol=lf
 reporter/src/main/resources/**/*.ftl	text eol=lf
 
 # Use Unix line endings for scan results which are misused for also testing the calculation of the SPDX package verification code.
-scanner/src/test/assets/*.json	text eol=lf
+scanner/src/test/assets/*.json				text eol=lf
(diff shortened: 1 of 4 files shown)


.gitattributes
```diff
@@ -20,7 +20,8 @@ reporter/src/funTest/assets/*-expected-*	text eol=lf
 reporter/src/main/resources/**/*.ftl	text eol=lf
```
nnobelis, line 21: i don't understand how this change is relevant to the bug ?!
Why a change to a Reporter test would have an impact on a Scanner test ?

sschuberth: I'm not sure what you're referring to, I'm not changing a reporter-test related line in this file. I'm just adding a line for `scanner/src/test/assets/scanoss/filesToScan/*.kt` (and adjusting the indentation in the line above).

nnobelis: Yes, it was a mistake from me..

nnobelis (approved):

sschuberth:
Note that `funTest-analyzer-docker` has actually [passed](https://github.com/oss-review-toolkit/ort/actions/runs/3104576344/jobs/5029209495#step:6:938), just reporting that back does not seem to have worked. So merging nonetheless.

PR #5857, opened 2022-09-23
Fix a StackOverflowError in DependencyGraphBuilder

oheger-bosch:
Refer to the single commits for further information.


model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -149,7 +149,7 @@ class DependencyGraphBuilder<D>(
      * the [DependencyGraph].
      */
     fun addDependency(scopeName: String, dependency: D): DependencyGraphBuilder<D> =
-        apply { addDependencyToGraph(scopeName, dependency, transitive = false) }
+        apply { addDependencyToGraph(scopeName, dependency, transitive = false, emptySet()) }
```
sschuberth, line 152: Commit message nit: Missing dot at the end of the last sentence that starts with "Resolves".

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -238,7 +238,12 @@ class DependencyGraphBuilder<D>(
      * Update the dependency graph by adding the given [dependency], which may be [transitive], for the scope with name
      * [scopeName]. All the dependencies of this dependency are processed recursively.
      */
-    private fun addDependencyToGraph(scopeName: String, dependency: D, transitive: Boolean): DependencyReference {
+    private fun addDependencyToGraph(
+        scopeName: String,
+        dependency: D,
+        transitive: Boolean,
+        processed: Set<D>
```
sschuberth, line 245: Please mention this new parameter in the function docs.

sschuberth: Or probably it's enough to document it only for the most low-level function for which this new parameter was introduced.

analyzer/src/funTest/assets/projects/synthetic/gradle-android-cyclic/app/src/main/res/drawable-v24/ic_launcher_foreground.xml
```diff
@@ -0,0 +1,34 @@
+<vector xmlns:aapt="http://schemas.android.com/aapt"
```
sschuberth, line 1: Could we strip all resources to keep the number of files in this synthetic project small?

analyzer/src/funTest/kotlin/managers/GradleAndroidFunTest.kt
```diff
@@ -77,6 +77,20 @@ class GradleAndroidFunTest : StringSpec() {
 
             result.toYaml() shouldBe expectedResult
         }
+
+        "Cyclic dependencies can be handled" {
```
sschuberth, line 81: Could you highlight the semantic differnece to `"deal with cycles in dependencies"` in `DependencyGraphBuilderTest`?

analyzer/src/funTest/assets/projects/synthetic/gradle-android-cyclic/app/src/main/res/drawable-v24/ic_launcher_foreground.xml
oheger-bosch: Actually, I am not sure whether this test should be added. Currently it is failing (for reasons that need to be investigated); but if it runs through, it takes a rather long time (almost 20 minutes on my machine) - these Android projects are really heavy-weighted.

Unfortunately, I was unable to come up with a minimum synthetic test that reproduces the issue.

What is your opinion about the test?

sschuberth: If it's really taking 20 minutes, then maybe we could still add it but tag it with `AndroidTag` and `ExpensiveTag`, and again start not running those in CI.

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
oheger-bosch: The function docs have been extended.

analyzer/src/funTest/assets/projects/synthetic/gradle-android-cyclic/app/src/main/res/drawable-v24/ic_launcher_foreground.xml
oheger-bosch: I have added the mentioned tags. Let's see whether the test passes now, then I can promote this PR to 'ready to review'.

analyzer/src/funTest/kotlin/managers/GradleAndroidFunTest.kt
oheger-bosch: I have improved the test name. The difference is basically that the test in `DependencyGraphBuilder` uses a minimum cycle. In this real-world example, the cycle spans multiple packages; this causes a different behavior of the builder.

oheger-bosch:
There are now new (unrelated) test failures, but the newly added test is successful.

oheger-bosch:
@sschuberth, the test is now successful, but according to the logs, the execution of all analyzer fun tests now took 70 m. How do we proceed?

sschuberth:
> How do we proceed?

As mentioned [above](https://github.com/oss-review-toolkit/ort/pull/5857#discussion_r978591990), I'd exclude `ExpensiveTag` tests from running as part of Batect's `funTestAnalyzer` task via `-Dkotest.tags=!ExpensiveTag`, similar to like we did in the old Azure pipelines before 4ff68955437d95400ca7f7835c6e4708bdf6694b.

However, before that we should probably evaluate which of the current `ExpensiveTag` tests we still regard as "expensive"; they have all been executed on each PR after all for quite a while now. So I just did https://github.com/oss-review-toolkit/ort/pull/5865 to be able to see how long tests ran, and maybe everything above *X* minutes should get the `ExpensiveTag` (unless it's a crucial test, like the only one we have for a specific package manager or so).

Then we can disable `ExpensiveTag` for PRs and instead only run them once a day or so.

oheger-bosch:
> As mentioned [above](https://github.com/oss-review-toolkit/ort/pull/5857#discussion_r978591990), I'd exclude `ExpensiveTag` tests from running as part of Batect's `funTestAnalyzer` task via `-Dkotest.tags=!ExpensiveTag`, similar to like we did in the old Azure pipelines before [4ff6895](https://github.com/oss-review-toolkit/ort/commit/4ff68955437d95400ca7f7835c6e4708bdf6694b).
> 
> However, before that we should probably evaluate which of the current `ExpensiveTag` tests we still regard as "expensive"; they have all been executed on each PR after all for quite a while now. So I just did #5865 to be able to see how long tests ran, and maybe everything above _X_ minutes should get the `ExpensiveTag` (unless it's a crucial test, like the only one we have for a specific package manager or so).
> 
> Then we can disable `ExpensiveTag` for PRs and instead only run them once a day or so.

Sounds good. So with #5865 in place, can this be merged now?

sschuberth:
> Sounds good. So with #5865 in place, can this be merged now?

Not yet. #5865 only enabled the build scans for tests to *get the durations of tests*. We'd still need to agree on from when on a test counts as "expensive", retag all tests accordingly, exclude expensive tags from PR runs, and set up a new job that runs expensive tests on a nighly basis. I don't have the time to work on all of this.

To get a feeling for recent test durations, see
* [(unit) tests](https://scans.gradle.com/s/oyguk3fb3r67s/tests/slowest-tests)
* [non-analyzer functional tests](https://scans.gradle.com/s/6or2n3xwckats/tests/slowest-tests)
* and I just realized I missed to enable build scans for analyzer functional tests (will do that now)

But in order to get your fixed merged, can we either disable the test, or split it out to a separate PR? And while doing that, we could also consider to split the addition of the rather big test project from the addition of the test itself.

sschuberth (approved):

PR #5862, opened 2022-09-26
Consistently use `filePath` instead of `filepath` in Kotlin code

fviernau:


evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -88,8 +88,8 @@ open class ProjectSourceRule(
     fun projectSourceGetDetectedLicensesByFilePath(
```
nnobelis, line 88: Commit message `filepath` with a small P.

tsteenbe (approved):

PR #5863, opened 2022-09-26
PathExcludeGenerator: Misc minor improvements

fviernau:
See individual commits.

Fixes: #5763.


helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -154,6 +154,7 @@ private fun <T> Collection<T>.checkPatterns(patternSelector: (T) -> String) =
         }
     }
 
+/** Case insensitive glob patterns matched against directory names */
```
mnonnenmacher, line 158: Nit: Missing dot at the end.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5872, opened 2022-09-27
Require ORT versions to start with a digit

sschuberth:
This prevents Git from determining the ORT version relative to arbitrary tags like the recently added "legacy-dockerfile" tag.

### DIFF
--- a/scripts/docker_build.sh
+++ b/scripts/docker_build.sh
@@ -21,7 +21,7 @@
 DOCKER_ARGS=$@
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
-GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty)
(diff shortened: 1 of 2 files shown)


scripts/docker_build.sh
```diff
@@ -21,7 +21,7 @@
 DOCKER_ARGS=$@
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
-GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty)
+GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty --match=[0-9]*)
```
mnonnenmacher, line 24: Interesting, why does this match revisions that start with a letter like "b9e8b63ace"?

sschuberth: It's the *name of the tag* being matched, not the revision.

PR #5873, opened 2022-09-27
Conan: Do not log passwords

oheger-bosch:
Since Conan uses `ProcessCapture` to run `conan user` commands, the whole command line including the password is logged. This PR adds a mechanism to mask specific arguments when using `ProcessCapture`, so that sensitive information can be hidden.

### DIFF
--- a/utils/common/src/main/kotlin/MaskedString.kt
+++ b/utils/common/src/main/kotlin/MaskedString.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright (C) 2022 Bosch.IO GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.utils.common
+
+/**
+ * A class that wraps a [CharSequence] and overrides some functions to return a configurable mask string
+ * instead of the actual content. The main use case for this class is to prevent that sensitive information like
+ * passwords are visible in log output. The class can be used everywhere a [CharSequence] is accepted.
+ */
+data class MaskedString(
+    /** The wrapped string. */
+    val value: CharSequence,
+
+    /** A string that is returned by the toString method instead of the actual value. */
+    val mask: String = DEFAULT_MASK
+) : CharSequence by mask {
+    companion object {
+        const val DEFAULT_MASK = "*****"
+    }
+
+    override fun toString() = mask
+}
+
+/**
+ * Return an array with strings from the given [args]. If one of the input strings is a [MaskedString], include its
+ * actual value. So, this function can be used to access the unmasked strings.
+ */
+fun unmaskedStrings(vararg args: CharSequence): Array<out String> =
+    args.map { s ->
+        when (s) {
+            is MaskedString -> s.value.toString()
+            else -> s.toString()
+        }
+    }.toTypedArray()
+
+/**
+ * Convert this [CharSequence] to a [MaskedString] which uses the provided [mask].
+ */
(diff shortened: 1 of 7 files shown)


sschuberth:
I'm wondering whether that masking feature shouldn't be part of the logging instead, so it would also apply to logging e.g. URLs with encoded credentials. But I have to admit I don't see a simple way to implement that.

mnonnenmacher:
> I'm wondering whether that masking feature shouldn't be part of the logging instead, so it would also apply to logging e.g. URLs with encoded credentials. But I have to admit I don't see a simple way to implement that.

Hoplite uses a special [`Masked`](https://github.com/sksamuel/hoplite#masked-values) type for that, maybe we could implement a similar mechanism. It's basically just a [wrapper around `String`](https://github.com/sksamuel/hoplite/blob/4ed1f3397c13b12e31bc48ac06d0e9b40e9dcd37/hoplite-core/src/main/kotlin/com/sksamuel/hoplite/types.kt) that ensures that `toString()` does not print the actual value.

@oheger-bosch Would this be a feasible approach here? I think the difficulty is what type `vararg command` should have then, but the current mechanism to provide the argument index seems to be a bit fragile.

oheger-bosch:
> Hoplite uses a special [`Masked`](https://github.com/sksamuel/hoplite#masked-values) type for that, maybe we could implement a similar mechanism. It's basically just a [wrapper around `String`](https://github.com/sksamuel/hoplite/blob/4ed1f3397c13b12e31bc48ac06d0e9b40e9dcd37/hoplite-core/src/main/kotlin/com/sksamuel/hoplite/types.kt) that ensures that `toString()` does not print the actual value.
> 
> @oheger-bosch Would this be a feasible approach here? I think the difficulty is what type `vararg command` should have then, but the current mechanism to provide the argument index seems to be a bit fragile.

Agreed to the fragility argument. Could the wrapper type implement `CharSequence`, maybe using Kotlin's `by` delegate mechanism?

mnonnenmacher:
> Agreed to the fragility argument. Could the wrapper type implement `CharSequence`, maybe using Kotlin's `by` delegate mechanism?

This might be a good idea, something like this could work:

```kotlin
class Masked(val value: String) : CharSequence by MASK {
    companion object {
        private const val MASK = "***"
    }

    override fun toString(): String = MASK
}
```

This would ensure that by default all `toString()` calls get the mask but you can check for the type to get the actual value.

utils/common/src/main/kotlin/MaskedString.kt
```diff
@@ -0,0 +1,52 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+package org.ossreviewtoolkit.utils.common
```
mnonnenmacher, line 19: Nit: Please add an empty line between license and package.

utils/common/src/test/kotlin/MaskedStringTest.kt
```diff
@@ -0,0 +1,72 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+package org.ossreviewtoolkit.utils.common
```
mnonnenmacher, line 19: Nit: Please add an empty line between license and package.

utils/common/src/main/kotlin/MaskedString.kt
```diff
@@ -0,0 +1,52 @@
...
+    /** A string that is returned by the toString method instead of the actual value. */
+    val mask: String = DEFAULT_MASK
+) : CharSequence by mask {
+    companion object {
+        const val DEFAULT_MASK = "*****"
+    }
+
+    override fun toString(): String {
+        return mask
+    }
```
mnonnenmacher, line 39: Could be shorter: `override fun toString() = mask`

utils/common/src/main/kotlin/MaskedString.kt
```diff
@@ -25,7 +25,7 @@ package org.ossreviewtoolkit.utils.common
  */
 data class MaskedString(
     /** The wrapped string. */
-    val s: CharSequence,
+    val value: CharSequence,
```
mnonnenmacher, line 28: I think this commit is supposed to be squashed into the previous commit.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5876, opened 2022-09-27
OrtAuthenticator: Use credentials in URL when provided

nnobelis:
When a URL contains credentials (in the 'userInfo' part of the URL), the Authenticator should use them and stop searching for credentials. This commit extracts the credentials from the URL and use them for password authentication.

### DIFF
--- a/utils/ort/src/main/kotlin/OrtAuthenticator.kt
+++ b/utils/ort/src/main/kotlin/OrtAuthenticator.kt
@@ -68,9 +68,9 @@ class OrtAuthenticator(private val original: Authenticator? = null) : Authentica
         }
     }
 
-    // First look for (potentially machine-specific) credentials in a netrc-style file, then look for generic
-    // credentials passed as environment variables.
-    private val delegateAuthenticators = listOf(NetRcAuthenticator(), EnvVarAuthenticator())
+    // First look if the credentials are already present in the URL, then search for (potentially machine-specific)
+    // credentials in a netrc-style file, and finally look for generic credentials passed as environment variables.
+    private val delegateAuthenticators = listOf(UserInfoAuthenticator(), NetRcAuthenticator(), EnvVarAuthenticator())
 
     private val serverAuthentication: ConcurrentHashMap<String, PasswordAuthentication> = ConcurrentHashMap()
 
--- a/utils/ort/src/main/kotlin/UserInfoAuthenticator.kt
+++ b/utils/ort/src/main/kotlin/UserInfoAuthenticator.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2022 The ORT Project Authors (See <https://github.com/oss-review-toolkit/ort-server/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.utils.ort
+
+import java.net.Authenticator
+import java.net.PasswordAuthentication
+
+/**
+ * A trivial [Authenticator] that checks if the credentials are already present in the URL. If this is the case, the
+ * credentials will be used for password authentication.
+ */
+class UserInfoAuthenticator : Authenticator() {
+    override fun getPasswordAuthentication(): PasswordAuthentication? {
+        if (requestorType == RequestorType.SERVER) {
+            requestingURL?.userInfo?.let {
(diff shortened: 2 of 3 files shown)


utils/ort/src/test/kotlin/UtilsTest.kt
```diff
@@ -410,5 +413,28 @@ class UtilsTest : WordSpec({
...
+                port,
+                null,
+                null,
+                scheme,
+                url,
+                Authenticator.RequestorType.SERVER
+            )
+            auth shouldNotBeNull {
+                this.userName shouldBe "foo"
+                String(this.password) shouldBe "bar"
```
sschuberth, line 435: Please drop superfluous `this.` prefixes.

utils/ort/src/test/kotlin/UtilsTest.kt
```diff
@@ -410,5 +413,28 @@ class UtilsTest : WordSpec({
...
+                host,
+                null,
+                port,
+                null,
+                null,
+                scheme,
+                url,
+                Authenticator.RequestorType.SERVER
+            )
+            auth shouldNotBeNull {
```
sschuberth, line 433: Add a blank line before this one to follow the AAA pattern.

utils/ort/src/main/kotlin/OrtAuthenticator.kt
```diff
@@ -86,6 +86,12 @@ class OrtAuthenticator(private val original: Authenticator? = null) : Authentica
             }
 
             RequestorType.SERVER -> {
+                requestingURL.userInfo?.let {
```
sschuberth, line 89: I'm actually surprised that `getPasswordAuthentication()` even gets called when the credentials are in the URL... but I guess you've verified that this happens in real-life.

sschuberth: Do you think it's worth implementing this is a separate `UserInfoAuthenticator` that is added as the first entry to `delegateAuthenticators`?

sschuberth (changes requested):

utils/ort/src/main/kotlin/OrtAuthenticator.kt
nnobelis: > I'm actually surprised that `getPasswordAuthentication()` even gets called when the credentials are in the URL... but I guess you've verified that this happens in real-life.

 Yes this is pretty strange behavior.

My understanding is, when an `Authenticator` is installed, all requests goes through a check in  the JDK. The latter consists of test request to the server (HEAD ?) to see if it requires authentication. If yes, it goes through the `Authenticator`.
When a request comes in with userinfo,  the latter is stripped before doing this test request, hence the request lands in `Authenticator`.

I could not find anything online to back it up so it's just my fuzzy explanation.

> Do you think it's worth implementing this is a separate `UserInfoAuthenticator` that is added as the first entry to `delegateAuthenticators`?

Good idea I will change my PR.

utils/ort/src/main/kotlin/UserInfoAuthenticator.kt
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A trivial [Authenticator] that checks if the credentials are already present in the URL. If this is the case, the
+ * credentials will be used for password authentication.
+ */
+class UserInfoAuthenticator : Authenticator() {
+    override fun getPasswordAuthentication(): PasswordAuthentication? {
+        if (requestorType == RequestorType.SERVER) {
+            requestingURL?.userInfo?.let {
+                val credentials = it.split(":")
```
sschuberth, line 33: Nit: We could use the `limit = 2` argument to stop splitting after two parts.

Thinking about it, we probably even must do that to not omit any parts from the password (Which might contain unescaped colons?) below when using `credentials[1]`.

utils/ort/src/main/kotlin/UserInfoAuthenticator.kt
```diff
@@ -0,0 +1,41 @@
...
+                if (credentials.size > 1) {
```
sschuberth, line 34: With the limit in place, this should become `credentials.size == 2`.

utils/ort/src/main/kotlin/UserInfoAuthenticator.kt
```diff
@@ -0,0 +1,41 @@
...
+                    return PasswordAuthentication(credentials.first(), credentials[1].toCharArray())
```
sschuberth, line 35: With the limit in place, `credentials[1]` could become a more speaking `credentials.last()` (which nicely matches `credentials.first()`).

utils/ort/src/main/kotlin/UserInfoAuthenticator.kt
```diff
@@ -0,0 +1,41 @@
...
+                }
+            }
+        }
+        return super.getPasswordAuthentication()
```
sschuberth, line 39: Nit: Blank line before this one, to separate from the block above.

utils/ort/src/test/kotlin/UtilsTest.kt
```diff
@@ -410,5 +413,28 @@ class UtilsTest : WordSpec({
...
+                port,
+                null,
+                null,
+                scheme,
+                url,
+                Authenticator.RequestorType.SERVER
+            )
+
+            auth shouldNotBeNull {
+                this.userName shouldBe "foo"
```
sschuberth, line 435: `this.` should still be dropped here.

sschuberth (changes requested):

sschuberth (approved):

PR #5883, opened 2022-09-29
Add linebreak so "==Dependencies" is correctly displayed as headline

hanna-modica:
Fixes the issue seen in below screenshot:

grafik

### DIFF
--- a/reporter/src/funTest/assets/asciidoc-template-reporter-expected-result.adoc
+++ b/reporter/src/funTest/assets/asciidoc-template-reporter-expected-result.adoc
@@ -20,6 +20,7 @@ The applicable license information is listed below:
 ** +Copyright 1+
 
 <<<
+
 == Dependencies
 
 This software depends on external packages and source code.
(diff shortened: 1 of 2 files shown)


mnonnenmacher (dismissed):
Please add a `Signed-off-by:` line to the commit message for the DCO check, see the git log for examples.

hanna-modica:
Hi @mnonnenmacher,
thanks for the hint. Seems I have forgotten to configure it for this project in my IDE. Added it now.

sschuberth:
Unrelated, but the bullets for the list items seems to be broken, too. Is that something you could also fix along the way, @hanna-modica?o

MarcelBochtler (approved):

hanna-modica:
Hi @sschuberth, thanks for the hint. I think not the bullet points are broken, but the (c)-symbol. So I think the problem maybe lies not in this report template. I will ask the customer who reported this issue to verify, that the fix is working and afterwards confirm, if the wrong display of the character is still there. If yes, I will open a separate issue.

MarcelBochtler:
@hanna-modica This expected test result also needs to be adjusted:
https://github.com/oss-review-toolkit/ort/blob/main/reporter/src/funTest/assets/asciidoc-template-reporter-expected-result.adoc

MarcelBochtler (changes requested):
@hanna-modica This expected test result also needs to be adjusted:
https://github.com/oss-review-toolkit/ort/blob/main/reporter/src/funTest/assets/asciidoc-template-reporter-expected-result.adoc

MarcelBochtler (approved):

PR #5884, opened 2022-09-29
Pip: Upgrade python-inspector to 0.7.1

mnonnenmacher:
For the release notes see [1].

To fix the analysis of `setup.py` files, always enable the new option `--analyze-setup-py-insecurely`. This will be made configurable in a later commit.

[1]: https://github.com/nexB/python-inspector/blob/main/CHANGELOG.rst#v070


analyzer/src/funTest/assets/projects/external/spdx-tools-python-expected-output.yml
```diff
@@ -27,8 +27,47 @@ project:
     dependencies:
     - id: "PyPI::ply:3.11"
     - id: "PyPI::rdflib:5.0.0"
-    - id: "PyPI::six:1.16.0"
+      dependencies:
```
fviernau, line 30: Does the upgrade without adding the new option also affect the expected output? / would it make sense to have separate commits for (1) upgrade and (2) adding the new option?

fviernau (approved):

analyzer/src/funTest/assets/projects/external/spdx-tools-python-expected-output.yml
mnonnenmacher: The option is required for some test projects, I have added this to the commit message, see: https://github.com/nexB/python-inspector/issues/72

fviernau (approved):

analyzer/src/funTest/assets/projects/external/spdx-tools-python-expected-output.yml
sschuberth, line 30: One should probably note that this restores the expected test result to (almost) how it was before 16c4d66b0c6f534ced9e9220761895dce114a90b. So that's good.

sschuberth (approved):

PR #5885, opened 2022-09-30
Dockerfile: Upgrade the Cargo version available in Ubuntu Jammy

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Cargo.kt
+++ b/analyzer/src/main/kotlin/managers/Cargo.kt
@@ -108,9 +108,22 @@ class Cargo(
         }
 
         val contents = Toml().read(lockfile)
-        val metadata = contents.getTable("metadata")?.toMap().orEmpty()
-        return metadata.mapNotNull { (k, v) ->
-            (v as? String)?.let { k.unquote().removePrefix("checksum ") to v }
+        return when (contents.getLong("version")) {
+            3L -> {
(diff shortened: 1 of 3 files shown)


mnonnenmacher (approved):

analyzer/src/main/kotlin/managers/Cargo.kt
```diff
@@ -107,17 +107,8 @@ class Cargo(
...
-        val metadataMapNotNull = mutableMapOf<String, String>()
-        metadataMap.forEach { (key, value) ->
-            if (key != null && (value as? String) != null) {
-                metadataMapNotNull[key] = value
-            }
-        }
-
-        return metadataMapNotNull
+        val metadata = contents.getTable("metadata")?.toMap().orEmpty()
+        return metadata.mapNotNull { (k, v) -> (v as? String)?.let { k to v } }.toMap()
```
fviernau: Was `key != null` unnecessary before?

sschuberth: According to the Toml library function docs, yes:

```
  /**
   * @param key A table name, not including square brackets.
   * @return A new Toml instance or <code>null</code> if no value is found for key.
   */
  @SuppressWarnings("unchecked")
  public Toml getTable(String key) {
```

While `value` is documented to probably be null, ` key` is not. Allowing null keys would also make it hard to distinguish from non-existing keys.

PR #5887, opened 2022-09-30
Upgrade pipenv to 2022.9.24

mnonnenmacher:


analyzer/src/main/kotlin/managers/Pipenv.kt
```diff
@@ -72,7 +73,11 @@ class Pipenv(
 
         logger.info { "Generating '${requirementsFile.name}' file in '$workingDir' directory..." }
 
-        val requirements = ProcessCapture(workingDir, command(), "lock", "--requirements").requireSuccess().stdout
+        val requirements = if (Semver(getVersion()) >= Semver("2022.8.13")) {
```
sschuberth, line 76: Should we maybe guard against `getVersion()` not being parsable as a `Semver` here? and maybe extract `Semver("2022.8.13")` to a class variable with a speaking name?

analyzer/src/main/kotlin/managers/Pipenv.kt
```diff
@@ -72,7 +73,11 @@ class Pipenv(
...
+            ProcessCapture(workingDir, command(), "requirements").requireSuccess().stdout
+        } else {
+            ProcessCapture(workingDir, command(), "lock", "--requirements").requireSuccess().stdout
```
sschuberth, line 79: Nit: Maybe pull out `.requireSuccess().stdout` of the ` if`?

analyzer/src/main/kotlin/managers/Pipenv.kt
sschuberth: Did you intend to also address the point from my first sentence?

mnonnenmacher: I forgot to respond, no, all versions of Pipenv so far are semver compliant so I don't think this is necessary.

sschuberth (approved):

PR #5889, opened 2022-09-30
FossID: Support URL mappings

oheger-bosch:
In our setup, we have cases in which FossID requires a different URL to scan a project than the repository URL used by ORT. To support such cases, this PR adds a mechanism based on regular expressions to map repository URLs to different URLs. To enable this, the FossID-specific scanner configuration can list an arbitrary number of mapping regular expressions. These are applied to the URL of the current package before it is passed to FossID.

### DIFF
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -156,7 +156,6 @@ ort:
         namingVariableVar3: myUnit
 
         waitForResult: false
-        addAuthenticationToUrl: false
 
         keepFailedScans: false
         deltaScans: true
@@ -166,6 +165,8 @@ ort:
         detectCopyrightStatements: true
 
         timeout: 60
+
+        urlMappingExample: "https://my-repo.example.org(?<repoPath>.*) -> ssh://my-mapped-repo.example.org${repoPath}"
 
     storages:
       local:
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
@@ -210,5 +210,5 @@ internal data class FossIdConfig(
     /**
      * Create a [FossIdUrlProvider] helper object based on the configuration stored in this object.
      */
-    fun createUrlProvider() = FossIdUrlProvider.create(addAuthenticationToUrl)
+    fun createUrlProvider() = FossIdUrlProvider.create()
 }
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
@@ -33,7 +33,6 @@ import org.ossreviewtoolkit.model.config.ScannerConfiguration
  * * **"apiKey":** The API key of the user which connects to the FossID server.
  * * **"waitForResult":** When set to false, ORT does not wait for repositories to be downloaded nor scans to be
  *   completed. As a consequence, scan results won't be available in ORT result.
- * * **"addAuthenticationToUrl":** If set, ORT will add credentials from its Authenticator to the URLs sent to FossID.
  * * **"deltaScans":** If set, ORT will create delta scans. When only changes in a repository need to be scanned,
  *   delta scans reuse the identifications of the latest scan on this repository to reduce the amount of findings. If
  *   *deltaScans* is set and no scan exist yet, an initial scan called "origin" scan will be created.
@@ -50,6 +49,28 @@ import org.ossreviewtoolkit.model.config.ScannerConfiguration
  *   variables prefixed by "$" e.g. "$Var1_$Var2". Variables are also passed as options and are prefixed by
  *   [NAMING_CONVENTION_VARIABLE_PREFIX] e.g. namingVariableVar1 = "foo".
  * * **"namingScanPattern":** A pattern for scan names when scans are created on the FossID instance.
+ *
+ * URL mapping options. These options allow transforming the URLs of specific repositories before they are passed to
+ * the FossID service. This may be necessary if FossID uses a different mechanism to clone a repository, e.g. via SSH
+ * instead of HTTP. Options of this form start with the prefix [FossIdUrlProvider.PREFIX_URL_MAPPING] followed by an
+ * arbitrary name. Their values define the mapping to be applied consisting of two parts separated by the string
+ * " -> ":
+ * * A regular expression to match the repository URL.
+ * * The replacement to be used for this repository URL. It can access the capture groups defined by the regular
+ * expression, so that rather flexible transformations can be achieved. In addition, it can contain the variables
+ * "#user" and "#password" that are replaced by the credentials known for the target host.
+ *
+ * The example
+ *
+ * `mapExampleRepo = https://my-repo.example.org(?<repoPath>.*) -> ssh://my-mapped-repo.example.org${repoPath}`
+ *
+ * would change the scheme from "https" to "ssh" and the host name for all repositories hosted on
+ * "my-repo.example.org". With
+ *
+ * `mapAddCredentials =
+ *   (?<scheme>)://(?<host>)(?<port>:\\d+)?(?<repoPath>.*) -> ${scheme}://#user:#password@${host}${port}${repoPath}`
+ *
+ * every repository URL would be added credentials. Mappings are applied in the order they are defined.
  */
 internal data class FossIdConfig(
     /** The URL where the FossID service is running. */
@@ -63,9 +84,6 @@ internal data class FossIdConfig(
 
     /** Flag whether the scanner should wait for the completion of FossID scans. */
     val waitForResult: Boolean,
-
-    /** Flag whether credentials should be passed to FossID in URLs. */
-    val addAuthenticationToUrl: Boolean,
 
     /** Flag whether failed scans should be kept. */
     val keepFailedScans: Boolean,
@@ -99,9 +117,6 @@ internal data class FossIdConfig(
 
         /** Name of the configuration property for the API key. */
         private const val API_KEY_PROPERTY = "apiKey"
-
-        /** Name of the configuration property controlling that credentials are added to URLs. */
-        private const val CREDENTIALS_IN_URL_PROPERTY = "addAuthenticationToUrl"
 
         /** Name of the configuration property controlling whether ORT should wait for FossID results. */
         private const val WAIT_FOR_RESULT_PROPERTY = "waitForResult"
@@ -152,7 +167,6 @@ internal data class FossIdConfig(
                 ?: throw IllegalArgumentException("No FossID API Key configuration found.")
 
             val waitForResult = fossIdScannerOptions[WAIT_FOR_RESULT_PROPERTY]?.toBoolean() ?: true
-            val addAuthenticationToUrl = fossIdScannerOptions[CREDENTIALS_IN_URL_PROPERTY]?.toBoolean() ?: false
 
             val keepFailedScans = fossIdScannerOptions[KEEP_FAILED_SCANS_PROPERTY]?.toBoolean() ?: false
             val deltaScans = fossIdScannerOptions[DELTA_SCAN_PROPERTY]?.toBoolean() ?: false
@@ -176,7 +190,6 @@ internal data class FossIdConfig(
                 user = user,
                 apiKey = apiKey,
                 waitForResult = waitForResult,
-                addAuthenticationToUrl = addAuthenticationToUrl,
                 keepFailedScans = keepFailedScans,
                 deltaScans = deltaScans,
                 deltaScanLimit = deltaScanLimit,
@@ -210,5 +223,5 @@ internal data class FossIdConfig(
     /**
      * Create a [FossIdUrlProvider] helper object based on the configuration stored in this object.
      */
-    fun createUrlProvider() = FossIdUrlProvider.create()
+    fun createUrlProvider() = FossIdUrlProvider.create(options)
 }
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
@@ -206,4 +206,9 @@ internal data class FossIdConfig(
 
         return FossIdNamingProvider(namingProjectPattern, namingScanPattern, namingConventionVariables)
     }
+
+    /**
+     * Create a [FossIdUrlProvider] helper object based on the configuration stored in this object.
+     */
+    fun createUrlProvider() = FossIdUrlProvider.create(addAuthenticationToUrl)
 }
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2022 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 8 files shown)


scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -50,6 +50,19 @@ import org.ossreviewtoolkit.model.config.ScannerConfiguration
  *   variables prefixed by "$" e.g. "$Var1_$Var2". Variables are also passed as options and are prefixed by
  *   [NAMING_CONVENTION_VARIABLE_PREFIX] e.g. namingVariableVar1 = "foo".
  * * **"namingScanPattern":** A pattern for scan names when scans are created on the FossID instance.
+ *
+ * URL mapping options. These options allow transforming the URLs of specific repositories before they are passed to
+ * the FossID service. This may be necessary if FossID uses a different mechanism to clone a repository, e.g. via SSH
```
sschuberth, line 55: I'm not sure I can follow here: Isn't the mechanism to clone a repository implied by the URL / protocol? So if FossID is passed a HTTP URL, how can it still decide by itself to use a different mechanism like SSH?

model/src/main/resources/reference.yml
```diff
@@ -167,6 +167,9 @@ ort:
 
         timeout: 60
 
+        'map:https://my-repo.example.org(?<repoPath>.*)': "ssh://my-mapped-repo.example.org${repoPath}"
```
sschuberth, line 170: From a pure user perspective, I believe this reads quite ugly. Can't we have nicely named keys and define the mapping in the values only?

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
```diff
@@ -0,0 +1,68 @@
...
+    companion object {
+        fun create(addAuthenticationToUrl: Boolean): FossIdUrlProvider = FossIdUrlProvider(addAuthenticationToUrl)
+
+        /**
+         * This function fetches credentials for [repoUrl] and insert them between the URL scheme and the host. If no
+         * matching host is found by [Authenticator], the [repoUrl] is returned untouched.
+         */
+        private fun queryAuthenticator(repoUrl: String): String {
+            val repoUri = repoUrl.toUri().getOrElse {
+                FossId.logger.warn { "The repository URL '$repoUrl' is not valid." }
```
nnobelis, line 81: Is there a reason for not using `FossIdUrlProvider.logger` here and underneath ?

scanner/src/test/kotlin/scanners/fossid/FossIdUrlProviderTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        val url = urlProvider.getUrl(REPO_URL)
+
+        url shouldBe expectedUrl
+    }
+
+    "Missing credentials should be handled" {
+        mockAuthenticator(authentication = null)
+        val urlProvider = FossIdUrlProvider.create(addAuthenticationToUrl = true)
+
+        urlProvider.getUrl(REPO_URL) shouldBe REPO_URL
```
nnobelis, line 54: Please make this line consistent with line 45.

nnobelis (commented):
IMO this solution is a bit over engineered. 
We talked only about replacing the scheme from `https://` to `ssh://`. Apart from that, I am not sure the need to rewrite other parts to the URL will ever arise...

However I won't oppose the merge.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
oheger-bosch: This is a ,misunderstanding: FossID tries to clone exactly the URL it is passed. However, in our setup ORT and FossID run in different network zones, and in some cases the URL used by ORT to clone a repository does not work for FossID. Therefore, with this new mechanism the URL is transformed accordingly by ORT before it is propagated to FossID.

model/src/main/resources/reference.yml
oheger-bosch: Yes, this is certainly not nice. Do you have a concrete proposal?

The key could be built from a prefix plus an arbitrary logical name, e.g. `mapHostXYZ`. I am unsure about the value though. We would need a unique separator between the regex and the replacement that should not appear in one of those.

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
oheger-bosch: Copy & paste.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
sschuberth: As the idea / wish to rewrite download / clone URLs has occurred several times in ORT's history, IMO we should think about implementing this in a higher level, so that URLs are always being rewritten, independent of the scanner being used.

Would that also work for you, or would such a global rewrite break something else for your network setup?

oheger-bosch: Hm, in our case there is indeed a scanner-specific aspect. FossID uses a different URL than the ORT downloader invoked from the ScanCode abstraction.

oheger-bosch:
> IMO this solution is a bit over engineered. We talked only about replacing the scheme from `https://` to `ssh://`. Apart from that, I am not sure the need to rewrite other parts to the URL will ever arise...
> 
Replacing only the scheme is not sufficient. For our use case, the port in the URL needs to be changed as well.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
oheger-bosch: @sschuberth, how to proceed here? Could we use this implementation as a kind of starting point to be improved later? Or do you have something different in mind?

sschuberth: > Or do you have something different in mind?

What I originally had in mind was to implement this in the analyzer, and perform some VCS URL mapping in the process of creating `vcsProcessed` from `vcs`. However, as I understand your use-case that would not work, as for some scanners you need the original URL, and for others the mapped URL, correct?

So the mapping basically depends on the infrastructure the scanner runs on. Which means it probably doesn't even make sense to move the mapping logic to a more common place like `ProvenanceScannerWrapper`.

fviernau: I'm afraid this is complicating the current (and future added) logic for handling provenance / resolved provenance.
I believe a complication could be avoided if this remained completely transparent: e.g. would it make sense that the scan result returned from the scanner would still return the original repository URL (not the mapped one) so that the generic code can remain completely unaware ?

fviernau:
> Replacing only the scheme is not sufficient. For our use case, the port in the URL needs to be changed as well.

From above discussion I understood that the mapped URL could only be used by the FossId scanner,
while from this comment it seems to me that the mapped URL could in fact be used by all scanners. Can you clarify @oheger-bosch ?

model/src/main/resources/reference.yml
sschuberth: > Yes, this is certainly not nice. Do you have a concrete proposal?

I have only skimmed over the changes, and I'm not sure if I'm getting all use-cases. Is there a use-case where credentials that haven't been part of the URL need to be part of the URL after the mapping? I.e. a simple mapping is not enough?

If so, maybe we could per-define "named groups" / placeholders that don't actually belong to a regex group, but refer to the username / password the authenticator providers for the URL. So you wouldn't have to distinguish between `map` and `mapCred` anymore, and can just use `$username` / `$password` in the replacement URL, as if there were regex groups named like that.

oheger-bosch:
> > Replacing only the scheme is not sufficient. For our use case, the port in the URL needs to be changed as well.
> 
> From above discussion I understood that the mapped URL could only be used by the FossId scanner, while from this comment it seems to me that the mapped URL could in fact be used by all scanners. Can you clarify @oheger-bosch ?

No, the mapped URL can only be used by the FossId scanner. In our constellation, there are two different URLs: one for the ORT downloader / the local ScanCode scanner, and one for FossId.

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
sschuberth: > would it make sense that the scan result returned from the scanner would still return the original repository URL (not the mapped one) so that the generic code can remain completely unaware ?

I makes me feel a bit uneasy that strictly speaking we're then not returning the URL / provenance that has been scanned in this case... which is esp. bad if you have some error in your defined mapping, and scan some completely different than you intend to.

oheger-bosch: With the current implementation in this PR, the URL is only transformed before it is passed to the FossID server. The URL reported in the provenance of the scan result is still the original one. I am not sure, which one to prefer. At least in our case, the same repository should be accessed, only with a different protocol.

model/src/main/resources/reference.yml
oheger-bosch: I have reworked the URL mapping configuration, so that the regex and the replacement are now defined in the value using ' -> ' as separator. Credentials can now be injected using special variables. Adding credentials to URLs is now handled by URL mappings as well; therefore, the corresponding flag was removed from the FossId configuration. As this changes the configuration, I added the "breaking change" label.

scanner/src/test/kotlin/scanners/fossid/FossIdUrlProviderTest.kt
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (C) 2022 Bosch.IO GmbH
```
sschuberth, line 2: Please use now

    Copyright (C) 2022 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)

scanner/src/main/kotlin/scanners/fossid/FossIdConfig.kt
```diff
@@ -206,4 +206,9 @@ internal data class FossIdConfig(
 
         return FossIdNamingProvider(namingProjectPattern, namingScanPattern, namingConventionVariables)
     }
+
+    /**
+     * Create a [FossIdUrlProvider] helper object based on the configuration stored in this object.
+     */
+    fun createUrlProvider(): FossIdUrlProvider = FossIdUrlProvider.create(addAuthenticationToUrl)
```
sschuberth: Nit: Maybe drop the explicit return type to avoid the `FossIdUrlProvider = FossIdUrlProvider` duplication?

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
```diff
@@ -31,40 +32,109 @@ import org.ossreviewtoolkit.utils.ort.requestPasswordAuthentication
...
-                repoUrl.replaceCredentialsInUri("${creds.userName}:${String(creds.password)}")
-            } ?: repoUrl
+            return requestPasswordAuthentication(repoUri)
+        }
+
+        /**
+         * Handle the variables referencing credentials if they occur in the [replacement string][replaced].
+         */
+        private fun replaceVariables(replaced: String): String =
+            replaced.takeUnless { it.contains(VAR_USERNAME) || it.contains(VAR_PASSWORD) }
```
sschuberth, line 96: Prefer `in` over `contains`.

scanner/src/test/kotlin/scanners/fossid/FossIdUrlProviderTest.kt
```diff
@@ -65,10 +104,22 @@ private const val USER = "scott"
...
-private fun mockAuthenticator(url: String = REPO_URL, authentication: PasswordAuthentication? = AUTHENTICATION) {
+private fun mockAuthenticator(
+    host: String = HOST,
+    port: Int = PORT,
+    authentication: PasswordAuthentication? = AUTHENTICATION
+) {
     mockkStatic("org.ossreviewtoolkit.utils.ort.UtilsKt")
-    every { requestPasswordAuthentication(URI(url)) } returns authentication
+    every { requestPasswordAuthentication(any()) } answers {
+        val uri: URI = firstArg()
```
sschuberth, line 122: Nit: Maybe use `firstArg<URI>()` instead?

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
```diff
@@ -31,40 +32,109 @@ import org.ossreviewtoolkit.utils.ort.requestPasswordAuthentication
...
+     * The URL mapping. URLs matched by a key [Regex] are replaced by the URL in the value. The replacement string can
+     * refer to matched groups of the [Regex]. It can also contain the variables [VAR_USERNAME] and [VAR_PASSWORD] to
+     * allow the injection of credentials.
+     */
+    private val urlMapping: Map<Regex, String>
 ) {
     companion object : Logging {
-        fun create(addAuthenticationToUrl: Boolean): FossIdUrlProvider = FossIdUrlProvider(addAuthenticationToUrl)
+        /** Variable in the mapping replacement string that references the username. */
+        private const val VAR_USERNAME = "#user"
```
sschuberth, line 50: I guess you deliberately chose `#` over `$` as a marker here, but if `$` works I guess I would prefer that, so the user does not have to remember that other marker char, and because "#user" might actually be an anchor a URL (very unlikely for a Git repo URL, though).

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
```diff
@@ -43,6 +43,9 @@ internal class FossIdUrlProvider private constructor(
     private val urlMapping: Map<Regex, String>
 ) {
     companion object : Logging {
+        /** The prefix of option keys that define a URL mapping. */
+        internal const val PREFIX_URL_MAPPING = "map"
```
sschuberth, line 47: Should we be more verbose here, like "urlMapping" or even "repo(sitory)UrlMapping"?

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
oheger-bosch: Yes, I wanted to avoid conflicts with the named groups in the regular expression. And indeed: when changing the variables to use the `$` prefix, the replace operation fails with:
`java.lang.IllegalArgumentException: Illegal group reference`

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
oheger-bosch: Changed the prefix to "urlMapping".

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
sschuberth: Ok, thanks for checking; then let's stick to `#`.

scanner/src/main/kotlin/scanners/fossid/FossIdUrlProvider.kt
```diff
@@ -0,0 +1,150 @@
...
+     * allow the injection of credentials.
+     */
+    private val urlMapping: Map<Regex, String>
+) {
+    companion object : Logging {
+        /** The prefix of option keys that define a URL mapping. */
+        internal const val PREFIX_URL_MAPPING = "urlMapping"
+
+        /** Variable in the mapping replacement string that references the username. */
+        private const val VAR_USERNAME = "#user"
```
sschuberth, line 50: One last (optional) remark: You're documenting the variable as "username", you're calling the variable "USERNAME", but the marker is just "user". Should the full marker also be "#username" for consistency, and to reduce the likeliness that the shorter "#user" appears anywhere else in the URL?

oheger-bosch: Yes, makes sense. Changed to the variable name "#username".

PR #5891, opened 2022-09-30
Dockerfile: Re-Add GRADLE_USER_HOME

jens-erdmann:
Re-add setting of GRADLE_USER_HOME to enable set_gradle_proxy.sh. Without the variable set gradle does not consider the configuration set by set_gradle_proxy.sh.
GRADLE_USER_HOME got lost in 10b38cb65955e18bac471f06ec2c25a1d19e544a.

Signed-off-by: Jens Erdmann <jens.erdmann@web.de>


sschuberth:
> Re-add setting of GRADLE_USER_HOME to enable set_gradle_proxy.sh. Without the variable set gradle does not consider the configuration set by set_gradle_proxy.sh.

The `set_gradle_proxy.sh` script has `${GRADLE_USER_HOME:-$HOME/.gradle}`, so isn't "$HOME/.gradle" a valid path in the new Dockerfile?

jens-erdmann:
> > Re-add setting of GRADLE_USER_HOME to enable set_gradle_proxy.sh. Without the variable set gradle does not consider the configuration set by set_gradle_proxy.sh.
> 
> The `set_gradle_proxy.sh` script has `${GRADLE_USER_HOME:-$HOME/.gradle}`, so isn't "$HOME/.gradle" a valid path in the new Dockerfile?

Apparently not. My setup is behind a proxy and without the change the docker build fails.

jens-erdmann:
I can invest a bit of time next week to investigate further on why the default is not working properly.

jens-erdmann:
I dissected this issue a bit further. 10b38cb65955e18bac471f06ec2c25a1d19e544a introduced some changes, that makes it hard for the script to provide a proper fallback. The script itself is Ok IMHO. The problem is, that in the [Dockerfile line 68](https://github.com/oss-review-toolkit/ort/blob/d09822fa366fa6ecbecb91ca9275a1b261e17041/Dockerfile#L68) the `HOME` variable gets modified. At the time of the execution of [gradle in line 261](https://github.com/oss-review-toolkit/ort/blob/d09822fa366fa6ecbecb91ca9275a1b261e17041/Dockerfile#L261) the user is root. So gradle would expect the config file to be in `/root/` but the script writes int `/home/ort/`.  
In addition in [Dockerfile line 255](https://github.com/oss-review-toolkit/ort/blob/d09822fa366fa6ecbecb91ca9275a1b261e17041/Dockerfile#L255) a predefined gradle config gets copied into the container. I am not entirely sure if the gradle wrapper script is recognizing it or not. I modified my changes to make the script write into this file instead.

jens-erdmann:
@sschuberth sorry for pinging you. Could you have another look? That coverage ci is nothing I can do much about, I think.

Dockerfile
```diff
@@ -256,7 +256,8 @@ COPY . /usr/local/src/ort
 WORKDIR /usr/local/src/ort
 
 # Prepare Gradle
-RUN scripts/import_proxy_certs.sh \
+RUN export GRADLE_USER_HOME=/usr/local/src/ort/ \
```
sschuberth, line 259: It looks weird that this is pointing to ORT sources now. Shouldn't this rather be `/home/ort/.gradle`?

sschuberth (changes requested):

Dockerfile
jens-erdmann: > It looks weird that this is pointing to ORT sources now. Shouldn't this rather be `/home/ort/.gradle`?

This setting is only temporary. It is only in effect at build time of ORT. At build time the ort user is not used but root.
To my understanding this is the right place as the gradle wrapper uses the properties file that is copied in and this change is only making sure, that this properties file is expanded with the proxy settings. These settings will not be added to the final image as of now.

sschuberth: But to my understanding it does not matter *when* this setting will be applied; setting Gradle's home directory to the ORT source code directory should *never* be correct.

jens-erdmann: Ok, might be worth a discussion. But were does this leave us now?  
With the presented fix my problems are solved. I can build the ORT docker image behind a proxy again. I can also use the set_gradle_proxy.sh script in my workflows to append the proxy settings for gradle at runtime.

> it should be the [user-global properties file](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties) below the current user's home directory.

That is still possible but in the current setup of the dockerfile it does not at build time of the container.

sschuberth (approved):

PR #5903, opened 2022-10-04
WorkingTreeCache: Ignore the VCS path

mnonnenmacher:
Ignore the VCS path when initializing the working tree. The scanner
    implementation always scans full repositories, so sparse checkout are
    not required. This improves the performance if an analyzer repositories
    contains many projects in different directories, because previously one
    working tree for each of them would have been created during scanning.
    Now only one working tree is created for this repository.
    
    Add a check to the `PackageProvenanceResolver` to verify that the VCS
    path from the package metadata exists, because the working tree would
    now not throw an exception anymore if it does not exist.

### DIFF
--- a/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
+++ b/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
@@ -220,7 +220,19 @@ class DefaultPackageProvenanceResolver(
 
             revisionCandidates.forEachIndexed { index, revision ->
                 logger.info { "Trying revision candidate '$revision' (${index + 1} of ${revisionCandidates.size})." }
-                val result = vcs.updateWorkingTree(workingTree, revision, pkg.vcsProcessed.path, recursive = false)
+                val result = vcs.updateWorkingTree(workingTree, revision, recursive = false)
+
+                if (pkg.vcsProcessed.path.isNotBlank() &&
+                    !workingTree.workingDir.resolve(pkg.vcsProcessed.path).exists()
+                ) {
+                    logger.info {
+                        "Discarding revision '$revision' because the requested VCS path '${pkg.vcsProcessed.path}' " +
+                                "does not exist."
+                    }
+
+                    return@forEachIndexed
+                }
+
                 if (result.isSuccess) {
                     val resolvedRevision = workingTree.getRevision()
 
--- a/scanner/src/main/kotlin/provenance/NestedProvenanceResolver.kt
+++ b/scanner/src/main/kotlin/provenance/NestedProvenanceResolver.kt
@@ -83,7 +83,6 @@ class DefaultNestedProvenanceResolver(
             vcs.updateWorkingTree(
                 workingTree,
                 provenance.resolvedRevision,
-                path = provenance.vcsInfo.path,
                 recursive = true
             ).onFailure { throw it }
 
--- a/scanner/src/main/kotlin/utils/WorkingTreeCache.kt
+++ b/scanner/src/main/kotlin/utils/WorkingTreeCache.kt
@@ -38,8 +38,9 @@ interface WorkingTreeCache {
     /**
      * Apply the [block] to the working tree defined by [vcsInfo]. The [VcsInfo.revision] is ignored during
      * initialization of the working tree, it is the caller's responsibility to update the working tree to the correct
-     * revision. It is ensured that only one [block] at a time can access a single working tree, separate working trees
-     * can be accessed in parallel.
+     * revision. The [VcsInfo.path] is also ignored, so the caller needs to verify that the expected path exists. It is
+     * ensured that only one [block] at a time can access a single working tree, separate working trees can be accessed
+     * in parallel.
      *
      * Throws an [IllegalStateException] if the cache was already [shut down][shutdown].
      */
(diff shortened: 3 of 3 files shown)


scanner/src/main/kotlin/experimental/WorkingTreeCache.kt
```diff
@@ -64,13 +64,15 @@ class DefaultWorkingTreeCache : WorkingTreeCache {
...
+    private fun getMutexKey(vcsInfo: VcsInfo) = "${vcsInfo.type}|${vcsInfo.url}"
+
+    private fun getWorkingTreeKey(vcsInfo: VcsInfo) = "${vcsInfo.type}|${vcsInfo.url}|${vcsInfo.path}"
 
     private suspend fun getWorkingTreeMutex(vcsInfo: VcsInfo) =
         mutex.withLock {
             check(!terminated) { "The cache was already shut down." }
 
-            workingTreeMutexes.getOrPut(getKey(vcsInfo)) { Mutex() }
+            workingTreeMutexes.getOrPut(getMutexKey(vcsInfo)) { Mutex() }
```
fviernau, line 75: I don't understand the existing approach, in particular why multiple working trees can exist for same (type, url) but with different path (or also revision). Mind giving me a hint?

(I wonder whether switching to max 1 working tree per (type, url) would be better)

mnonnenmacher: The historic reason was that the sparse checkout configuration is written to a file, so I wanted to avoid issues with re-using the working tree with and without sparse checkout. This is currently a draft because I just wanted to quickly see if this would fix our issue. I will alter replace it with another change that does indeed remove sparse checkouts. This is something I had on my list for some time because the current implementation clones the same repository once for each project found in a different path.

mnonnenmacher: It's updated now.

fviernau (dismissed):

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
```diff
@@ -221,6 +221,18 @@ class DefaultPackageProvenanceResolver(
             revisionCandidates.forEachIndexed { index, revision ->
                 logger.info { "Trying revision candidate '$revision' (${index + 1} of ${revisionCandidates.size})." }
                 val result = vcs.updateWorkingTree(workingTree, revision, pkg.vcsProcessed.path, recursive = false)
+
```
sschuberth, line 224: Commit message:
- "if an analyzer repositories contains" -> "if an analyzed repository contains" (?)

sschuberth (changes requested):

sschuberth:
BTW, I checked the callers of `VersionControlSystem.updateWorkingTree`, and there are only two more places that actually pass a `path`. Would you mind checking whether passing those paths is actually still required? Because if not, we could remove the feature of sparse checkouts completely from our VCS implementations which would nicely simplify the code, and allow us to completely switch to JGit.

mnonnenmacher:
> BTW, I checked the callers of `VersionControlSystem.updateWorkingTree`, and there are only two more places that actually pass a `path`. Would you mind checking whether passing those paths is actually still required? Because if not, we could remove the feature of sparse checkouts completely from our VCS implementations which would nicely simplify the code, and allow us to completely switch to JGit.

Not removing the one in `PackageProvenanceResolver` in this commit was actually a mistake which I fixed now. So the only occurrence left is in `VersionControlSystem.download()` which is still used by the downloader. We still depend on that to limit analysis of repositories to a sub-directory. If you are eager to remove the sparse checkout code, one option would be to "simulate" the sparse checkout by removing files manually after the checkout, then the feature would still work but you could clean up the VCS classes already.

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
```diff
@@ -220,7 +220,19 @@ class DefaultPackageProvenanceResolver(
...
-                val result = vcs.updateWorkingTree(workingTree, revision, pkg.vcsProcessed.path, recursive = false)
+                val result = vcs.updateWorkingTree(workingTree, revision, recursive = false)
+
+                if (pkg.vcsProcessed.path.isNotBlank() &&
+                    !workingTree.workingDir.resolve(pkg.vcsProcessed.path).exists()
+                ) {
+                    logger.info {
+                        "Discarding revision '$revision' because the requested VCS path '${pkg.vcsProcessed.path}' " +
+                                "does not exist."
+                    }
```
fviernau, line 231: Sorry, for not noticing this earlier. I just though about the use case where a user
sets a wrong (non-existing) VCS path in a curation. I'm not certain IIRC, but I think previously this lead to an ORT issue indicating that the VCS path does not exist. As of this change, the problem would only be logged, while the actual failure would be: "Cannot determine a revision to checkout". 

In another (more edge case) scenario, where multiple candidates match (by tag name) I'm not sure if using the path for discarding is a good idea.

Would it make sense to not use the path for discarding and do the following instead?
1. Determine the revision to checkout agnostic of the path
2. Report an OrtIssue if the VCS path does not exist in the guessed revision

fviernau (changes requested):

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
mnonnenmacher: That's a good point, I will make a separate PR to improve the error reporting to catch also other errors which are currently swallowed.

fviernau: So, do you agree that the revision should not be discarded based on the path?

mnonnenmacher: No, I think it should be discarded, better fail early than fail in the scanner step when we already know that the path does not exist here. What I want to improve is the error message so that the user knows why no revision was found, this will end up in an `OrtIssue`.

fviernau: I'm not 100% convinced about including the path in matching candidates, but also needs not be discussed in scope of this PR.

fviernau (approved):

PR #5905, opened 2022-10-04
More preparations for running Detekt with type resolution

sschuberth:

### DIFF
--- a/advisor/src/main/kotlin/advisors/Osv.kt
+++ b/advisor/src/main/kotlin/advisors/Osv.kt
@@ -111,7 +111,7 @@ class Osv(name: String, advisorConfiguration: AdvisorConfiguration) : AdviceProv
             }
         }.onFailure {
             logger.error {
-                "Requesting vulnerabilities IDs for packages failed: ${result.exceptionOrNull()!!.collectMessages()}"
+                "Requesting vulnerabilities IDs for packages failed: ${it.collectMessages()}"
             }
         }
 
@@ -123,7 +123,7 @@ class Osv(name: String, advisorConfiguration: AdvisorConfiguration) : AdviceProv
 
         return result.getOrElse {
             logger.error {
-                "Requesting vulnerabilities IDs for packages failed: ${result.exceptionOrNull()!!.collectMessages()}"
+                "Requesting vulnerabilities IDs for packages failed: ${it.collectMessages()}"
             }
             emptyList()
         }
@@ -185,13 +185,12 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
         }
     } ?: (null to null)
 
-    if (severity == null && databaseSpecific != null) {
+    val specificSeverity = databaseSpecific?.get("severity")
+    if (severity == null && specificSeverity != null) {
         // Fallback to the 'severity' property of the unspecified 'databaseSpecific' object.
         // See also https://github.com/google/osv.dev/issues/484.
-        databaseSpecific!!["severity"]?.let {
-            if (it is JsonPrimitive) {
-                severity = it.contentOrNull
-            }
+        if (specificSeverity is JsonPrimitive) {
+            severity = specificSeverity.contentOrNull
         }
     }
 
--- a/clients/fossid-webapp/src/main/kotlin/model/identification/markedAsIdentified/MarkedAsIdentifiedFile.kt
+++ b/clients/fossid-webapp/src/main/kotlin/model/identification/markedAsIdentified/MarkedAsIdentifiedFile.kt
@@ -47,7 +47,7 @@ data class MarkedAsIdentifiedFile(
         val licenses = file.licenses?.let { licenses ->
             licenses.values.map {
                 License(
-                    identifier = it.file.licenseIdentifier!!,
+                    identifier = checkNotNull(it.file.licenseIdentifier),
                     name = it.file.licenseName,
                     origin = it.type
                 )
@@ -61,7 +61,7 @@ data class MarkedAsIdentifiedFile(
         )
     }
 
-    override fun getFileName(): String = file.path!!
+    override fun getFileName(): String = checkNotNull(file.path)
 
     override fun getCopyright(): String = identificationCopyright
 }
--- a/clients/osv/src/main/kotlin/OsvService.kt
+++ b/clients/osv/src/main/kotlin/OsvService.kt
@@ -43,10 +43,12 @@ class OsvService(serverUrl: String = OsvApiClient.SERVER_URL_PRODUCTION, httpCli
         request: VulnerabilitiesForPackageRequest
     ): Result<List<Vulnerability>> {
         val response = client.getVulnerabilitiesForPackage(request).execute()
+        val body = response.body()
 
-        return when (response.isSuccessful) {
-            true -> Result.success(response.body()!!.vulnerabilities)
-            else -> Result.failure(IOException(response.message()))
+        return if (response.isSuccessful && body != null) {
+            Result.success(body.vulnerabilities)
+        } else {
+            Result.failure(IOException(response.message()))
         }
     }
 
@@ -63,10 +65,11 @@ class OsvService(serverUrl: String = OsvApiClient.SERVER_URL_PRODUCTION, httpCli
         requests.chunked(OsvApiClient.BATCH_REQUEST_MAX_SIZE).forEach { requestsChunk ->
             val batchRequest = VulnerabilitiesForPackageBatchRequest(requestsChunk)
             val response = client.getVulnerabilityIdsForPackages(batchRequest).execute()
+            val body = response.body()
 
-            if (!response.isSuccessful) return Result.failure(IOException(response.message()))
+            if (!response.isSuccessful || body == null) return Result.failure(IOException(response.message()))
 
-            result += response.body()!!.results.map { batchResponse ->
+            result += body.results.map { batchResponse ->
                 batchResponse.vulnerabilities.mapTo(mutableListOf()) { it.id }
             }
         }
@@ -79,10 +82,12 @@ class OsvService(serverUrl: String = OsvApiClient.SERVER_URL_PRODUCTION, httpCli
      */
     fun getVulnerabilityForId(id: String): Result<Vulnerability> {
         val response = client.getVulnerabilityForId(id).execute()
+        val body = response.body()
 
-        return when (response.isSuccessful) {
-            true -> Result.success(response.body()!!)
-            else -> Result.failure(IOException(response.message()))
+        return if (response.isSuccessful && body != null) {
+            Result.success(body)
+        } else {
+            Result.failure(IOException(response.message()))
         }
     }
 
--- a/model/src/main/kotlin/config/FileStorageConfiguration.kt
+++ b/model/src/main/kotlin/config/FileStorageConfiguration.kt
@@ -43,24 +43,22 @@ data class FileStorageConfiguration(
      * Create a [FileStorage] based on this configuration.
      */
     fun createFileStorage(): FileStorage {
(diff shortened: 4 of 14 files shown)


clients/fossid-webapp/src/main/kotlin/model/identification/markedAsIdentified/MarkedAsIdentifiedFile.kt
```diff
@@ -47,7 +47,7 @@ data class MarkedAsIdentifiedFile(
         val licenses = file.licenses?.let { licenses ->
             licenses.values.map {
                 License(
-                    identifier = it.file.licenseIdentifier!!,
+                    identifier = checkNotNull(it.file.licenseIdentifier),
```
fviernau, line 50: Mind giving a hint why `checkNotNull` is better than `!!``?

model/src/main/kotlin/config/FileStorageConfiguration.kt
```diff
@@ -42,25 +42,22 @@ data class FileStorageConfiguration(
     /**
      * Create a [FileStorage] based on this configuration.
      */
-    fun createFileStorage(): FileStorage {
-        require((httpFileStorage == null) xor (localFileStorage == null)) {
-            "Exactly one implementation must be configured for a FileStorage."
-        }
+    fun createFileStorage(): FileStorage =
+        when {
```
fviernau, line 46: For me the first three lines of the old function helped quite a bit in terms of readability as it made it very obvious
that excatly one of the storages must be configured. The change IMO sacrifices that, which is why I'd prefer to fix that by just replacing `!!` with `checkNotNull`.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -185,13 +185,12 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
         }
     } ?: (null to null)
 
-    if (severity == null && databaseSpecific != null) {
+    val specificSeverity = databaseSpecific?.get("severity")
```
fviernau, line 188: Consider following alternative which avoid the variable `specificSeverity`:

```
if (severity == null) {
   databaseSpecific?.get("severity")?.let {
     if (it is JsonPrimitive) {
         severity = it.contentOrNull
     }
   }
}
```

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -185,13 +185,12 @@ private fun Vulnerability.toOrtVulnerability(): org.ossreviewtoolkit.model.Vulne
...
+    if (severity == null && specificSeverity != null) {
         // Fallback to the 'severity' property of the unspecified 'databaseSpecific' object.
         // See also https://github.com/google/osv.dev/issues/484.
-        databaseSpecific!!["severity"]?.let {
-            if (it is JsonPrimitive) {
-                severity = it.contentOrNull
-            }
+        if (specificSeverity is JsonPrimitive) {
```
fviernau, line 192: I don't see any problem with the assertion as was before, do you?
I wonder if the detect rule flags all assertions and if we should disable that rule.
Mis-use IMO can be spotted in code reviews.

What do you think?

clients/fossid-webapp/src/main/kotlin/model/identification/markedAsIdentified/MarkedAsIdentifiedFile.kt
sschuberth: Quite simply because Detekt does not complain anymore this way. Personally, I also find it more expressive than using the rather low-level `!!` "cast".

fviernau: So, the agreement would be to never use `!!`. For cases where `!!` is valid, we should instead use the semantically identical `!!` right ?

clients/osv/src/main/kotlin/OsvService.kt
```diff
@@ -43,10 +43,12 @@ class OsvService(serverUrl: String = OsvApiClient.SERVER_URL_PRODUCTION, httpCli
...
     ): Result<List<Vulnerability>> {
         val response = client.getVulnerabilitiesForPackage(request).execute()
+        val body = response.body()
 
-        return when (response.isSuccessful) {
-            true -> Result.success(response.body()!!.vulnerabilities)
-            else -> Result.failure(IOException(response.message()))
+        return if (response.isSuccessful && body != null) {
+            Result.success(body.vulnerabilities)
+        } else {
```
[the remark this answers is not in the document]
sschuberth: > if the result is success then the body must not be empty.

I believe instead of "empty" you mean "null", so I guess what you're trying to say is: Previously, a null body threw an exception due to the failing `!!` cast, and now a null-body is "silently" turned into a failure result, and the message of the wrapped `IOException` might not reveal the reason for the body to be null.

While that might indeed be, I believe the new behavior better fits the semantics of the function, because as a user I'd not expect a function that returns a `Result` to throw, but always wrap any exception into a `Result.failure`.

As a side note, it should anyway never happen that the body is null; we just need to jump through some hoops to please Kotlin's nullability checks, and [upcoming versions of OkHttp will make the body non-nullable](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500-alpha7).

model/src/main/kotlin/config/FileStorageConfiguration.kt
sschuberth: I was actually trying to "kill two birds with one stone" here and also get rid of the IMO rather unreadable `let` / `?:` construct in this case. I'll see if I can come up with something that makes the "either or" check again performed earlier.

advisor/src/main/kotlin/advisors/Osv.kt
sschuberth: I prefer to keep the current code as it makes clear already in the outer condition to only do something if we don't have a `severity` *but* we have a `specificSeverity`.

clients/osv/src/main/kotlin/OsvService.kt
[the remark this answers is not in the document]
fviernau: >  I believe the new behavior better fits the semantics of the function, because as a user I'd not expect a function that returns a Result to throw

I believe the reasoning why it's better gets invalidated by your next sentence saying that

> As a side note, it should anyway never happen

In my view this change sacrifices on readability, and as it should never happen I prefer not to sacrifice readability.

So, I do not agree with that change but won't block it.

advisor/src/main/kotlin/advisors/Osv.kt
sschuberth: > I don't see any problem with the assertion as was before, do you?

The use of `!!` per se is considered bad practice (by Detekt and [common](https://unity-technologies.github.io/kotlin-guide/#safe-variable-declarations) [Kotlin](https://medium.com/android-news/how-to-remove-all-from-your-kotlin-code-87dc2c9767fb) [guidelines](https://www.baeldung.com/kotlin/not-null-assertion#use-cases)), independently of the context, and personally agree to that. IMO we should avoid the use of `!!` where ever possible, for readability and safety. To me, `!!` is like an unchecked low-level cast in C++; you wouldn't do that either, unless absolutely necessary.

> I wonder if the detect rule flags all assertions and if we should disable that rule.

Yes, Detekt (with type resolution enabled) flags the use of `!!` in general, and I believe that rule to be good, so I wouldn't disable it.

> Mis-use IMO can be spotted in code reviews.

I believe a tiny `!!` can easily be overlooked, and differentiating a misuse from proper use can be challenging without a very deep review, which we might not always have time to do.

clients/fossid-webapp/src/main/kotlin/model/identification/markedAsIdentified/MarkedAsIdentifiedFile.kt
sschuberth: > For cases where `!!` is valid, we should instead use the semantically identical `!!` right ?

I guess by the last `!!` you mean `checkNotNull`? Anyway, a "semantically required" `!!` should not always blindly be replaced with `checkNotNull`. You could also use `requireNotNull` depending on what you want to express. (Are you testing arguments or state? Also see https://bignerdranch.com/blog/write-better-code-using-kotlins-require-check-and-assert/.)

One of my problems with `!!` is that it does not really tell you *why* you can assume a value to be not-null; `checkNotNull` / `requireNotNull` are more expressive there, IMO.

fviernau: > I guess by the last !! you mean checkNotNull?

yes.

> One of my problems with !! is that it does not really tell you why you can assume a value to be not-null; checkNotNull / requireNotNull are more expressive there, IMO.

True, if a message is provided. So, given that I'm ok with ditching `!!` everywhere makes sense, so that at least we have it consistent even if there is no message provided.

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: Ok, let's keep the code as you wrote it then.

clients/fossid-webapp/src/main/kotlin/model/identification/markedAsIdentified/MarkedAsIdentifiedFile.kt
sschuberth: > True, if a message is provided.

I believe even without a message they are more expressive, as `checkNotNull` implies that something with the object state is wrong, while `requireNotNull` implies that an argument constraint was not met.

model/src/main/kotlin/config/FileStorageConfiguration.kt
sschuberth: I found a solution that I actually believe to be quite elegant, please have a look.

fviernau (approved):

utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -277,7 +277,7 @@ fun <K, V : Collection<T>, T> Map<K, V>.zipWithCollections(other: Map<K, V>): Ma
     zip(other) { a, b ->
         when {
             // When iterating over the combined key set, not both values can be null.
-            a == null -> b!!
+            a == null -> checkNotNull(b)
```
MarcelBochtler, line 280: Java 14 improved the `NullPointerExceptions` to display the actual variable name which was null causing this exception.
`checkNotNull()` only shows `Required value was null.`.

I'm not sure if this Java 14 feature would also work with NPEs from Kotlin, and if it would be more helpful. But it might be one thing to consider.

MarcelBochtler: Aside from that, when looking at this commit, if these are the only places where it had to be changed, I would also vote for keeping this the rule to disallow `!!`.

MarcelBochtler (approved):

utils/common/src/main/kotlin/Extensions.kt
MarcelBochtler: Additionally, this rule seems not to disallow _every_ usage of `!!`.

sschuberth: > I'm not sure if this Java 14 feature would also work with NPEs from Kotlin, and if it would be more helpful.

That's a nice Java feature indeed, and I believe it would work with any JVM language; however, for the time being we're probably stuck with Java 11 anyway, also see https://github.com/oss-review-toolkit/ort/pull/4948 😞 

> Aside from that, when looking at this commit, if these are the only places where it had to be changed

To be fair, there might be might more places (in non-main source sets) that would need to be changed when keeping this rule enabled *and* using Detekt with type resolution. I'm not yet done with looking at all the code, but I'll probably stop here for now with this preparation.

Issue #5901, opened 2022-10-04
Wrong matching with [AntPathMatcher] for path exclude


nnobelis:
We have a project with the following path exclude
```
paths:
    - pattern: "{gradlew,gradlew.bat}"
      reason: "BUILD_TOOL_OF"
      comment: "Gradle build files"
```
However it seems this pattern excludes the `build.gradle` in the project.

After debugging, I came to evaluate this bit of code in ORT Reporter:
```
AntPathMatcher().apply { setCaseSensitive(true) }.match("{gradlew,gradlew.bat}", "build.gradle")
```
and to my surprise it evaluates to  `true`.

Has anyone an idea what happens here ?
It does not seem to conform to what's in Spring Javadoc [1].

As a sidenote, should ORT not use [2] instead for glob matching ?

[1]: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html
[2]: https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)

nnobelis:
Thinking about it maybe this is not a bug. Maybe this expression is just not supported.
Please advise.

sschuberth:
> As a sidenote, should ORT not use [2](https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)) instead for glob matching ?

Unfortunately, that standard implementation behaves oddly in some edge cases, i.e. it does not match files that we expect to be matched for certain patterns.

sschuberth:
> As anyone an idea what happens here ?

Probably @fviernau can advise.

fviernau:
@nnobelis @sschuberth , I wonder if the `AntPathMatcher` syntax is actually the best choice.
When I have some time I'd make some research what other implementations are available, so that we could switch
to something else. Do you agree at least rearching it makes sense?

nnobelis:
>  Do you agree at least searching it makes sense?

Definitively. Spring also as a replacement for `AntPathMatcher` if I am not mistaken:
https://www.baeldung.com/spring-5-mvc-url-matching

sschuberth:
> When I have some time I'd make some research what other implementations are available, so that we could switch
> to something else. Do you agree at least rearching it makes sense?

[...]

PR #5907, opened 2022-10-04
Re-align path exclude patterns with the changed behavior

fviernau:
The migration from `Java`'s path matcher to `AntPathMatcher` was a breaking change (unintended) for patterns which
match groups of subpatterns. Migrate existing patterns to the new notation and add a test case illustrating the current behavior.

Fixes: #5901.

### DIFF
--- a/.ort.yml
+++ b/.ort.yml
@@ -1,7 +1,7 @@
 ---
 excludes:
   paths:
-  - pattern: "**/src/{funTest,test}/**"
+  - pattern: "**/src/{:funTest|test}/**"
     reason: "TEST_OF"
     comment: >-
       Licenses contained in this directory are used for testing and do not apply to the OSS Review Toolkit.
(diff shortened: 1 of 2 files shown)


nnobelis:
Are you sure `/{:funTest|test}/` is a supported syntax of `AntPathMatcher` ?

I don't see it mentionned in the doc: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html

fviernau:
> Are you sure `/{:funTest|test}/` is a supported syntax of `AntPathMatcher` ?
> 
> I don't see it mentionned in the doc: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html

The test shows it works. In the docs it says "{spring:[a-z]+} matches the regexp [a-z]+ as a path variable named "spring".
In that case `spring` is the variable name. I just tried if one can omit the variable name, and it seems to work.

.ort.yml
```diff
@@ -1,15 +1,15 @@
 ---
```
nnobelis, line 1: Commit message:
```
Unfortunately [1] has changed the notation for matching a group of
subpatterns  (the group matches if any subpattern in the group
matches) has changed.
```
2x `has changed`

.ort.yml
```diff
@@ -1,15 +1,15 @@
...
 excludes:
   paths:
-  - pattern: "**/src/{funTest,test}/**"
+  - pattern: "**/src/{:funTest|test}/**"
```
sschuberth, line 4: Just as a side note: IMO that new syntax is less intuitive, and it's a bit sad we have to do this, but it's good to have this aligned and documented now; maybe we find a way to revert to the previous syntax by using (yet) another glob matcher implementation in the future...

sschuberth (approved):

tsteenbe:
@fviernau  https://github.com/oss-review-toolkit/ort/blob/main/docs/config-file-ort-yml.md still points to glob expression tutorial "For how to write a glob pattern, please see this [tutorial](https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob)." 

Could you fill a separate PR to update this to a AntPathMatcher one?

Issue #5899, opened 2022-10-04
Scanner: Runs into `NoSuchElementException`


fviernau:
While scanning ORT with ORT I noticed that ORT exits scanning with the following exception:

```
Exception in thread "main" java.util.NoSuchElementException: Key RepositoryProvenance(vcsInfo=VcsInfo(type=Git, url=https://github.com/square/okhttp.git, revision=parent-3.10.0, path=), resolvedRevision=c0739a419949a24d0c34cf38a25953c60871268b) is missing in the map.
	at kotlin.collections.MapsKt__MapWithDefaultKt.getOrImplicitDefaultNullable(MapWithDefault.kt:24)
	at kotlin.collections.MapsKt__MapsKt.getValue(Maps.kt:346)
	at org.ossreviewtoolkit.scanner.experimental.ScanController.buildNestedProvenanceScanResult(ScanController.kt:280)
	at org.ossreviewtoolkit.scanner.experimental.ScanController.getNestedScanResultsByPackage(ScanController.kt:170)
	at org.ossreviewtoolkit.scanner.experimental.ExperimentalScanner.scan(ExperimentalScanner.kt:158)
	at org.ossreviewtoolkit.scanner.experimental.ExperimentalScanner$scan$4.invokeSuspend(ExperimentalScanner.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:284)
	at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:85)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59)
	at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:38)
	at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source)
	at org.ossreviewtoolkit.cli.commands.ScannerCommand.runExperimental(ScannerCommand.kt:373)
	at org.ossreviewtoolkit.cli.commands.ScannerCommand.run(ScannerCommand.kt:250)
	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:198)
[...]

PR #5912, opened 2022-10-05
Gracefully handle nested provenance resolution failures

fviernau:
Add a fallback so that an assumption in the code holds in additional cases, where that assumption lead to a crash before.

See individual commits.

### DIFF
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -228,6 +228,9 @@ class Scanner(
                             )
                         )
                     }
+
+                    // If resolving the nested repositories failed, we can still scan the main repository.
+                    controller.addNestedProvenance(provenance, NestedProvenance(provenance, emptyMap()))
                 }
             }
         }
--- a/scanner/src/main/kotlin/ScanController.kt
+++ b/scanner/src/main/kotlin/ScanController.kt
@@ -21,6 +21,8 @@ package org.ossreviewtoolkit.scanner
 
 import java.time.Instant
 
+import org.apache.logging.log4j.kotlin.Logging
+
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.KnownProvenance
 import org.ossreviewtoolkit.model.OrtIssue
@@ -54,6 +56,8 @@ class ScanController(
      */
     val config: ScannerConfiguration
 ) {
+    companion object : Logging
+
     /**
      * A map of package [Identifier]s to a list of [OrtIssue]s that occurred during provenance resolution for the
      * respective package.
@@ -89,10 +93,14 @@ class ScanController(
     private val scanResults = mutableMapOf<ScannerWrapper, MutableMap<KnownProvenance, MutableList<ScanResult>>>()
 
     fun addProvenanceResolutionIssue(id: Identifier, issue: OrtIssue) {
+        logger.log(issue.severity.toLog4jLevel()) { issue.message }
+
         provenanceResolutionIssues.getOrPut(id) { mutableListOf() } += issue
     }
 
     fun addIssue(id: Identifier, issue: OrtIssue) {
+        logger.log(issue.severity.toLog4jLevel()) { issue.message }
+
         issues.getOrPut(id) { mutableListOf() } += issue
     }
 


scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -19,6 +19,7 @@
 
 package org.ossreviewtoolkit.scanner.experimental
 
+import org.apache.logging.log4j.kotlin.Logging
```
sschuberth, line 22: Wrong import ordering.

scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -52,6 +53,7 @@ class ScanController(
      */
     val config: ScannerConfiguration
 ) {
+    companion object : Logging
```
sschuberth, line 56: Add an empty line after this one.

scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -87,10 +89,14 @@ class ScanController(
     private val scanResults = mutableMapOf<ScannerWrapper, MutableMap<KnownProvenance, MutableList<ScanResult>>>()
 
     fun addProvenanceResolutionIssue(id: Identifier, issue: OrtIssue) {
+        logger.warn { issue.message }
```
sschuberth, line 92: I believe we can't just always use `warn` here but should try to match the issue's severity. So basically do something similar to what we already do in `createAndLogIssue`:

    logger.log(issue.severity.toLog4jLevel()) { issue.message }

Same below.

sschuberth (changes requested):

scanner/src/main/kotlin/experimental/ScanController.kt
```diff
@@ -87,10 +91,14 @@ class ScanController(
     private val scanResults = mutableMapOf<ScannerWrapper, MutableMap<KnownProvenance, MutableList<ScanResult>>>()
 
     fun addProvenanceResolutionIssue(id: Identifier, issue: OrtIssue) {
+        logger.log(issue.severity.toLog4jLevel()) { issue.message }
+
         provenanceResolutionIssues.getOrPut(id) { mutableListOf() } += issue
     }
 
     fun addIssue(id: Identifier, issue: OrtIssue) {
+        logger.log(issue.severity.toLog4jLevel()) { issue.message }
```
sschuberth, line 100: BTW, are these issues *additionally* logged else where in a single blow? If so, we should probably remove that other logging. Maybe @mnonnenmacher knows?

scanner/src/main/kotlin/experimental/ExperimentalScanner.kt
```diff
@@ -223,6 +223,9 @@ class ExperimentalScanner(
                             )
                         )
                     }
+
+                    // If resolving the nested repositories failed, we can still scan the main repository.
+                    controller.addNestedProvenance(provenance, NestedProvenance(provenance, emptyMap()))
```
sschuberth, line 228: I'm unsure whether this dummy entry is the right approach, or whether we rather should change the linked lookup in `ScanController` to handle non-existing entries. Maybe again @mnonnenmacher can advise?

scanner/src/main/kotlin/experimental/ScanController.kt
fviernau: The issues are created [here](https://github.com/oss-review-toolkit/ort/blob/c4a628d37017fb34e0ba7218f52eb52172cb7397/scanner/src/main/kotlin/experimental/ExperimentalScanner.kt#L183-L191) and they arise if PackageProvenanceResolver throws from [here](https://github.com/oss-review-toolkit/ort/blob/c4a628d37017fb34e0ba7218f52eb52172cb7397/scanner/src/main/kotlin/experimental/PackageProvenanceResolver.kt#L223).

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -228,6 +228,9 @@ class Scanner(
                             )
                         )
                     }
+
+                    // If resolving the nested repositories failed, we can still scan the main repository.
+                    controller.addNestedProvenance(provenance, NestedProvenance(provenance, emptyMap()))
```
mnonnenmacher, line 233: I deliberately did not catch or handle those errors so far because they indicate bugs, this helped fixing issues in the past. Can you provide a stacktrace for the exception to see where exactly it did occur? I suspect that the root cause might be in `getNestedScanResultsByPackage()` which does not filter packages with provenance resolution errors.

fviernau: For the stack trace, please see https://github.com/oss-review-toolkit/ort/issues/5899.

mnonnenmacher: Ok, then it is indeed `getNestedScanResultsByPackage()`. I think this function should be changed to not try to build a scan result for a package where provenance resolution failed by calling `hasCompleteScanResult()` on the packages ((after changing `hasCompleteScanResult()` to take an `Identifier` instead of a `Package` as argument)).

fviernau: IIUC you propose the following: In case nested provenance resolution failed AND provenance resolution of main repository succeeded, do not return any scan result even if the main repository has been scanned. Can you explain "why" you think it is better to not use the scan result of the main repository?

fviernau: > Ok, then it is indeed `getNestedScanResultsByPackage()`. I think this function should be changed to not try to build a scan result for a package where provenance resolution failed by calling `hasCompleteScanResult()` on the packages (after changing `hasCompleteScanResult()` to take an `Identifier` instead of a `Package` as argument).

I tried your approach adjusting `hasCompleteScanResult()`, but failed as that has `scanner` as second parameter which seems to me to not fit the use case. So, I've researched a bit further and found a solution which I hope aligns with your idea. 
I made it a separate PR, https://github.com/oss-review-toolkit/ort/pull/5933, as I didn't feel ready to scratch this one yet.
However, we can use that PR as replacement if we consider it to be ok.

mnonnenmacher: > IIUC you propose the following: In case nested provenance resolution failed AND provenance resolution of main repository succeeded, do not return any scan result even if the main repository has been scanned. Can you explain "why" you think it is better to not use the scan result of the main repository?

If I'm not wrong the main repository is not scanned if the nested provenance resolution failed. It could only happen as a side-effect if another package with successful nested provenance resolution uses the same repository. Not providing a result for a package where nested provenance resolution is better in my opinion because it could give the false impression that the scan was successful. This could also be an issue with package based storages, if the incomplete result would end up in the storage.

mnonnenmacher: > I tried your approach adjusting `hasCompleteScanResult()`, but failed as that has `scanner` as second parameter which seems to me to not fit the use case.

Right, a better approach would have been to filter packages in `getNestedScanResultsByPackage` using `getNestedProvenance(id)`. I still need to look at your other PR.

fviernau: > Right, a better approach would have been to filter packages in `getNestedScanResultsByPackage` using `getNestedProvenance(id)`. I still need to look at your other PR.

That's basically what my other PR does.

mnonnenmacher: Is this PR obsolete now that #5933 is approved?

fviernau: yes, let me close it.

fviernau:
Closing as this PR has been replaced by https://github.com/oss-review-toolkit/ort/pull/5933.

PR #5915, opened 2022-10-06
docker: Decouple cargo/rust from base linux

heliocastro:
Signed-off-by: Helio Chissini de Castro <heliocastro@gmail.com>


batect.yml
```diff
@@ -32,9 +32,6 @@ config_variables:
     default: simple
 
 .common-container-config: &common-container-config
-  run_as_current_user:
```
sschuberth, line 35: Why was this removed? Running as the current user (i.e. non-root) still is a good thing, or?

heliocastro: Was a test to reproduce on CI, i already check the results

sschuberth: > Was a test to reproduce on CI, i already checked the results.

Just as a hint, if you'd like to check something on CI while the PR is not yet ready, you can use a draft PR. These will also trigger CI checks.

batect.yml
```diff
@@ -21,7 +21,7 @@ config_variables:
   gradle_build_scan:
     description: Enabling of Gradle build scans
     # Supported values: "true", "false"
-    default: false
+    default: "false"
```
tsteenbe, line 24: Propose to move the change to batect.yml to a separate commit something like "batect: Quote boolean string"

tsteenbe (changes requested):

batect.yml
sschuberth: Is this change even necessary?

mnonnenmacher (approved):

PR #5920, opened 2022-10-07
Eliminate repetition in the construction of `ScannerRun`s

fviernau:
See individual commits.

In context of https://github.com/oss-review-toolkit/ort/issues/5681.


model/src/main/kotlin/ScannerRun.kt
```diff
@@ -29,14 +29,14 @@ import org.ossreviewtoolkit.utils.ort.Environment
  */
 data class ScannerRun(
     /**
-     * The [Instant] the scanner was started. The default value exists only for backward compatibility.
+     * The [Instant] the scanner was started.
```
MarcelBochtler, line 32: Commit message: The reference for `[1]` is missing.

MarcelBochtler (changes requested):

MarcelBochtler (dismissed):

MarcelBochtler (approved):

PR #5921, opened 2022-10-07
Eliminate repetition in the construction of `EvaluatorRun`s 

fviernau:
See individual commits.

In context of https://github.com/oss-review-toolkit/ort/issues/5681.


model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -307,7 +307,7 @@ class OrtResultTest : WordSpec({
                         )
                     )
                 ),
-                evaluator = EvaluatorRun(
+                evaluator = EvaluatorRun.EMPTY.copy(
```
sschuberth, line 310: Nit: The argument "for specifying only the needed arguments" is not really convincing here, as no unneeded arguments can be dropped here.

sschuberth (approved):

PR #5926, opened 2022-10-09
github-graphql: Upgrade the schema to version 12.11.0

sschuberth:
See [1]. While at it, drop the "doc" infix from the schema name to align with the upstream name.

[1]: https://github.com/octokit/graphql-schema/releases/tag/v12.11.0


fviernau (approved):

sschuberth:
@oheger-bosch just to be on the safe side: Was there a reason to call the file `schema.docs.graphql` (with the "doc" infix) initially?

oheger-bosch:
> @oheger-bosch just to be on the safe side: Was there a reason to call the file `schema.docs.graphql` (with the "docs" infix) initially?

Sorry, can't remember. But this is certainly not a naming convention invented by me. Was this the default name used by the plugin that processes this information?

sschuberth:
> Was this the default name used by the plugin that processes this information?

No idea. At least the name was explicitly configured. I'm now using the filename as-is from upstream.

PR #5929, opened 2022-10-10
DefaultPackageProvenanceResolver: Improve error reporting

mnonnenmacher:
Add the reasons why revision candidates could not be resolved to the exception that is thrown if no revision could be resolved. This ensures that these reasons do not only appear in the log output but also in the `OrtIssue` which will be added to the reports.

### DIFF
--- a/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
+++ b/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
@@ -218,7 +218,7 @@ class DefaultPackageProvenanceResolver(
 
             if (revisionCandidates.isEmpty()) {
                 addAndLogMessage(
-                    "Could not find any revision candidates for package '${pkg.id.toCoordinates()}' with VCS " +
+                    "Could not find any revision candidates for package '${pkg.id.toCoordinates()}' with " +
                             "${pkg.vcsProcessed}."
                 )
             }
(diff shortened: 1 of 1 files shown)


fviernau (approved):

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
```diff
@@ -209,13 +209,16 @@ class DefaultPackageProvenanceResolver(
...
             }.getOrDefault(emptySet())
 
-            if (revisionCandidates.isEmpty()) {
-                val message = "Could not find any revision candidates for package '${pkg.id.toCoordinates()}' with " +
-                        "VCS ${pkg.vcsProcessed}."
+            val errorMessages = mutableListOf<String>()
 
-                storage.putProvenance(pkg.id, pkg.vcsProcessed, UnresolvedPackageProvenance(message))
+            fun addAndLogError(message: String) {
+                logger.info { message }
```
sschuberth, line 215: Shouldn't errors also be logged at `error` level?

sschuberth (changes requested):

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
```diff
@@ -209,13 +209,16 @@ class DefaultPackageProvenanceResolver(
...
+                errorMessages += message
+            }
 
-                throw IOException(message)
+            if (revisionCandidates.isEmpty()) {
+                addAndLogError("Could not find any revision candidates for package '${pkg.id.toCoordinates()}' with " +
+                        "VCS ${pkg.vcsProcessed}.")
```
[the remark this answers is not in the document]
mnonnenmacher: I have added a commit to remove the redundant "VCS".

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
mnonnenmacher: I think "info" is the right log level for these messages because issues during provenance resolution are expected and often do not lead to an error but just to trying another candidate. But I have renamed the function to accommodate for that.

sschuberth (approved):

PR #5933, opened 2022-10-10
ScanController: Fix a crash when nested provenance resoltion failed

fviernau:
The first line of the function, `nestedProvenances[root].getValue()`, crashes in case there is no map entry for the given key. Such crash happens when the nested provenance resolution failed. Stop returning any scan results for packages for which the nested provenance is not available.

Note: It may make sense to return the scan results for the provenances which are actually available, in particular the scan result corresponding to the non-nested provenance / main code repository.

Relates to https://github.com/oss-review-toolkit/ort/issues/5899.


scanner/src/main/kotlin/ScanController.kt
```diff
@@ -169,7 +169,7 @@ class ScanController(
     /**
      * Get the [NestedProvenanceScanResult] for the provided [id].
      */
-    fun getNestedScanResult(id: Identifier): NestedProvenanceScanResult =
+    fun getNestedScanResult(id: Identifier): NestedProvenanceScanResult? =
```
mnonnenmacher, line 172: Commit message: You forgot to mention which function you refer to in the first sentence (`buildNestedProvenanceScanResult()`).

scanner/src/main/kotlin/ScanController.kt
```diff
@@ -178,10 +178,16 @@ class ScanController(
...
             val issues = issues[id].orEmpty()
-            packages.first { it.id == id } to buildNestedProvenanceScanResult(provenance, issues)
-        }.filterByVcsPath().filterByIgnorePatterns()
+            val scanResult = buildNestedProvenanceScanResult(provenance, issues)
+
+            if (scanResult != null) {
+                packages.first { it.id == id } to scanResult
+            } else {
+                null
+            }
```
mnonnenmacher, line 189: Shorter:
```kotlin
buildNestedProvenanceScanResult(provenance, issues)?.let { scanResult ->
  packages.first { it.id == id } to scanResult
}
```

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #5940, opened 2022-10-11
Pip: Get project metadata from python-inspector

mnonnenmacher:

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
+++ b/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
@@ -275,7 +275,7 @@ internal fun List<PythonInspector.Package>.toOrtPackages(): SortedSet<Package> =
             } ?: RemoteArtifact.EMPTY
 
         fun PythonInspector.Package.getDeclaredLicenses() =
-            listOfNotNull(declaredLicense.takeIf { it.isNotBlank() && it != "UNKNOWN" }).toSortedSet()
+            listOfNotNull(getLicenseFromLicenseField(declaredLicense)).toSortedSet()
 
         Package(
             // The package has a namespace property which is currently always empty. Deliberately set the namespace to
--- a/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
+++ b/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
@@ -76,6 +76,15 @@ internal object PythonInspector : CommandLineTool {
             }
 
             add(definitionFile.absolutePath)
+
+            if (definitionFile.name != "setup.py") {
+                // If a setup.py file exists, add it to the analysis to capture additional project metadata.
+                val setupFile = definitionFile.resolveSibling("setup.py")
+                if (setupFile.isFile) {
+                    add("--setup-py")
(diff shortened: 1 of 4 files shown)


sschuberth (approved):

analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
```diff
@@ -76,6 +76,15 @@ internal object PythonInspector : CommandLineTool {
             }
 
             add(definitionFile.absolutePath)
+
+            if (definitionFile.name != "setup.py") {
+                // If a setup.py file exists, add it to the analysis to capture additional project metadata.
+                val setupFile = definitionFile.parentFile.resolve("setup.py")
```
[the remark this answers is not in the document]
sschuberth: This could probably be worked around by using `resolveSibling()` instead.

analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
```diff
@@ -173,6 +178,85 @@ internal object PythonInspector : CommandLineTool {
...
 private const val TYPE = "PyPI"
 
+internal fun PythonInspector.Result.toOrtProject(
+    managerName: String,
+    analysisRoot: File,
+    definitionFile: File
+): Project {
+    val id = resolveIdentifier(managerName, analysisRoot, definitionFile)
+
+    val setupProject = projects.find { it.path.endsWith("/setup.py") }
```
fviernau, line 188: I believe that a `setup.py` in the analysis root won't be found.

Note: Same code exists multiple times below.

fviernau (changes requested):

analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
```diff
@@ -86,68 +86,68 @@ internal object PythonInspector : CommandLineTool {
...
+        val name: String?,
+        val email: String?,
+        val url: String?
+    )
+}
 
 private const val TYPE = "PyPI"
 
-internal fun List<PythonInspectorPackage>.toOrtPackages(): SortedSet<Package> =
+internal fun List<PythonInspector.Package>.toOrtPackages(): SortedSet<Package> =
```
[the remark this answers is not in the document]
mnonnenmacher: That`s not possible as another `Package` class is already imported.

analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
mnonnenmacher: It's not looking at the root because it uses `endsWith`. Also, the python-inspector output uses absolute paths so it would not be the analysis root but the file system root.

sschuberth (approved):

analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
fviernau: Consider dropping the assumption about absolute paths anyway,e.g.: `it.path.substringAfterLast('/') == "setup.py"`.

fviernau (approved):

PR #5942, opened 2022-10-12
GeneratePathExcludeCommand: Add a preference for filename patterns

fviernau:
Make the behavior a bit less dependent on the the order of the hard-coded patterns, and hopefully return simpler patterns.

See individual commits.

### DIFF
--- a/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
+++ b/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
@@ -97,7 +97,7 @@ internal object PathExcludeGenerator {
             files.filter { pathExcludeExclude.matches(it.path) }.toSet()
         }
 
-        return greedySetCover(filesForPathExcludes).toSet()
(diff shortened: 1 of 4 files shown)


helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -97,7 +97,9 @@ internal object PathExcludeGenerator {
             files.filter { pathExcludeExclude.matches(it.path) }.toSet()
         }
 
-        return greedySetCover(filesForPathExcludes).toSet()
+        return greedySetCover(filesForPathExcludes) { a, b ->
+            b.pattern.length - a.pattern.length
```
sschuberth, line 101: Just a thought: I wonder whether "shorter" is the best metric to identify "simpler" patterns (if getting "simpler" patters from a human perspective is the goal). Instead, maybe "contains fewer wildcards" could be an alternative metric.

sschuberth (commented):
There seem to be detekt issues.

sschuberth (approved):

fviernau:
The failing tests was unrelated, merged regardlessly.

Issue #5484, opened 2022-06-23
Automatical VCS path detection


fviernau:
Depending on the package manager it may be possible to better set the VCS path so that scan scope is narrowed down a bit.
For example:

1. GoMod: The source tree of a module is guaranteed to be root at the directory where the `go.mod` file resides in. So, one could find that directory and set the VCS path without risk
2. Maven: Often times the source tree is rooted at the directory where the `pom.xml` resides.

...TBC

tsteenbe:
@fviernau  I would word the above slightly differently

2. Maven: For a repository containing multiple Maven packages, the directory with the same name as the package containing pom.xml is often the right VCS path, see e.g. https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-accessanalyzer/pom.xml
3. npm/yarn: For a repository containing multiple Node packages, the directory with the same name as the package containing package.json is often the right VCS path, see e.g. https://github.com/heremaps/harp.gl/tree/master/%40here.

mnonnenmacher:
A good place to implement such logic could be the `PackageProvenanceResolver`:
* It has to clone the repository anyway to verify that the revision exists or to find the revision.
* The resolution result is already stored in a database and associated with the related package, therefore it's traceable.

A possible risk if this is implemented outside the analyzer is that it needs to be verified that all code relying on the VCS path of the package is made aware that the scan result might use a different path.

PR #5943, opened 2022-10-12
MavenSupport: Also take the `url` into account when fixing the SCM

MarcelBochtler:
For POMs that do not set an SCM URL itself, the fixup of the SCM does
not work if the parent SCM does not set an `connection`, but only the
`url`. This results in an unwanted original SCM URL.
For instance, spring-boot-starter-parent [1] only sets the `url` and not
the `connection`. This results in an original SCM `url` like
'https://github.com/spring-projects/spring-boot/childArtifactName'.
The issue this causes is, that the `Project`'s `VcsInfo` is merged with
this original SCM's URL as `fallbackUrl`, causing the path to be set to
`childArtifactName`, which is not available in the child repository.
Fix this issue, by also taking the `url` into account when replacing the
SCM with the parent's one.

[1]: https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-starter-parent/2.7.4/spring-boot-starter-parent-2.7.4.pom

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/maven-expected-output-app.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/maven-expected-output-app.yml
@@ -12,8 +12,8 @@ project:
     mapped:
       Apache License, Version 2.0: "Apache-2.0"
   vcs:
-    type: ""
-    url: ""
+    type: "Git"
+    url: "https://example.com/git/repository.git"
     revision: ""
     path: ""
   vcs_processed:
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -180,6 +180,12 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
                             scm.connection = parentScm.connection
                         }
                     }
+
+                    parentScm.url?.let { parentUrl ->
+                        if (parentUrl.isNotBlank() && scm.url.startsWith(parentUrl)) {
+                            scm.url = parentScm.url
+                        }
+                    }
                 }
 
                 parent = parent.parent
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -189,10 +189,10 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
         }
 
         private fun parseScm(scm: Scm?): VcsInfo {
-            val connection = scm?.connection.orEmpty()
-            val tag = scm?.tag?.takeIf { it != "HEAD" }.orEmpty()
+            val connection = scm?.connection
+            if (connection.isNullOrEmpty()) return VcsInfo.EMPTY
 
-            if (connection.isEmpty()) return VcsInfo.EMPTY
+            val tag = scm.tag?.takeIf { it != "HEAD" }.orEmpty()
 
             return SCM_REGEX.matcher(connection).let { matcher ->
                 if (matcher.matches()) {
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -159,15 +159,15 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
         fun parseVcsInfo(mavenProject: MavenProject) = parseScm(getOriginalScm(mavenProject))
 
         /**
-         * When asking Maven for the SCM URL of a POM that does not itself define an SCM URL, Maven returns the SCM
-         * URL of the first parent POM (if any) that defines one and appends the artifactIds of all child POMs to it,
-         * separated by slashes.
-         * This behavior is fundamentally broken because it invalidates the SCM URL for all VCS that cannot limit
-         * cloning to a specific path within a repository, or use a different syntax for that. Also, the assumption
-         * that the source code for a child artifact is stored in a top-level directory named like the artifactId
-         * inside the parent artifact's repository is often not correct.
-         * To address this, determine the SCM URL of the parent (if any) that is closest to the root POM and whose
-         * SCM URL still is a prefix of the child POM's SCM URL.
+         * When asking Maven for the SCM connection or the SCM URL of a POM that does not itself define these values,
+         * Maven returns the values of the first parent POM (if any) that defines one and appends the artifactIds of all
+         * child POMs to it, separated by slashes.
+         * This behavior is fundamentally broken because it invalidates the SCM connection / URL for all VCS that cannot
+         * limit cloning to a specific path within a repository, or use a different syntax for that. Also, the
+         * assumption that the source code for a child artifact is stored in a top-level directory named like the
+         * artifactId inside the parent artifact's repository is often not correct.
+         * To address this, determine the SCM connection and URL of the parent (if any) that is closest to the root POM
+         * and whose SCM connection / URL still is a prefix of the child POM's SCM values.
          */
         fun getOriginalScm(mavenProject: MavenProject): Scm? {
             val scm = mavenProject.scm
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -170,14 +170,14 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
          * SCM URL still is a prefix of the child POM's SCM URL.
          */
         fun getOriginalScm(mavenProject: MavenProject): Scm? {
-            var scm = mavenProject.scm
+            val scm = mavenProject.scm
             var parent = mavenProject.parent
 
             while (parent != null) {
                 parent.scm?.let { parentScm ->
                     parentScm.connection?.let { parentConnection ->
                         if (parentConnection.isNotBlank() && scm.connection.startsWith(parentConnection)) {
-                            scm = parentScm
+                            scm.connection = parentScm.connection
                         }
                     }
                 }
--- a/analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
+++ b/analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
@@ -30,6 +30,27 @@ import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
 
 class MavenSupportTest : WordSpec({
+    "getOriginalScm()" should {
+        "return the parent's SCM connection if the child SCM uses the parent's SCM connection implicitly" {
+            val mavenProject = MavenProject().apply {
+                scm = Scm().apply {
+                    connection = "scm:git:https://github.com/spring-projects/spring-boot.git/childArtifactName"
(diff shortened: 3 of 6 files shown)


sschuberth:
FYI, this is a bit related to https://github.com/oss-review-toolkit/ort/issues/5484 for the Maven case.

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
```diff
@@ -30,6 +30,23 @@ import org.ossreviewtoolkit.model.VcsInfo
...
+                    url = "https://github.com/spring-projects/spring-boot/childArtifactName"
+                }
+                parent = MavenProject().apply {
+                    scm = Scm().apply {
+                        url = "https://github.com/spring-projects/spring-boot"
+                    }
+                }
+            }
+
+            MavenSupport.getOriginalScm(mavenProject) shouldBe mavenProject.parent.scm
```
sschuberth, line 46: Do you really want to assert instances here? For readablity anyway it would be easier to only assert the URL as a string and spell out the expected value explicitly.

sschuberth:
> For POMs that do not set an SCM URL itself

Either singular "POM" or use "themselves" at the end.

Also, shouldn't the test case then *not* set an SCM URL for the child?

> The issue this causes is

Maybe start a new paragraph with this sentence?

sschuberth:
Could you please also provide a link to a child POM in the commit message?

MarcelBochtler:
> Also, shouldn't the test case then not set an SCM URL for the child?

The implicit adding of the SCM URL is not happening when accessing the `Scm` object.
It seems to be added when the POM is parsed:
https://github.com/oss-review-toolkit/ort/blob/c7bce54eb435b7c5f1dd793ffa0cfaca33c8fba7/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt#L444

So for keeping the test scope limited, I added the (typically implicitly added) value manually.

MarcelBochtler:
> Could you please also provide a link to a child POM in the commit message?

I don't have one at hand that I can share. But it is basically every POM that doesn't set the `SCM` and uses `spring-boot-starter-parent` as parent.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -175,8 +175,11 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
 
             while (parent != null) {
                 parent.scm?.let { parentScm ->
-                    parentScm.connection?.let { parentConnection ->
-                        if (parentConnection.isNotBlank() && scm.connection.startsWith(parentConnection)) {
+                    val parentScmUrl: String? = parentScm.connection ?: parentScm.url
+                    parentScmUrl?.let { url ->
+                        if (url.isNotBlank()
+                            && (scm.connection?.startsWith(url) == true || scm.url?.startsWith(url) == true)
```
sschuberth, line 181: It makes me feel a bit uneasy to compare the semantically different `connection` and `url` fields here. Shouldn't we instead simply duplicate the whole check for the `url` field? Like by adding:

                    parentScm.url?.let { parentUrl ->
                        if (parentUrl.isNotBlank() && scm.url.startsWith(parentUrl)) {
                            scm = parentScm
                        }
                    }

Or maybe combine the two checks like

                    val startsWithConnection = parentScm.connection.takeUnless { it.isBlank() }
                        ?.let { scm.connection.startsWith(it) } ?: false

                    val startsWithUrl = parentScm.url.takeUnless { it.isBlank() }
                        ?.let { scm.url.startsWith(it) } ?: false

                    if (startsWithConnection || startsWithUrl) scm = parentScm

sschuberth (dismissed):

MarcelBochtler:
The funTest found an issue here.
For the `Maven:org.jenkins-ci:version-number:1.4` package, the VCS is now incorrectly replaced with the parent's URL and revision.

I'll investigate how to fix this. Converting to draft for now.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -189,10 +189,8 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
         }
 
         private fun parseScm(scm: Scm?): VcsInfo {
-            val connection = scm?.connection.orEmpty()
-            val tag = scm?.tag?.takeIf { it != "HEAD" }.orEmpty()
-
-            if (connection.isEmpty()) return VcsInfo.EMPTY
+            val connection = scm?.connection ?: return VcsInfo.EMPTY
+            val tag = scm.tag?.takeIf { it != "HEAD" }.orEmpty()
```
mnonnenmacher, line 199: You change functionality here, previously `VcsInfo.EMPTY` was also returned if `scm.connection` was empty, now only if it is null. I would suggest to use `val connection = scm?.connection?.takeIf { it.isnotEmpty() } ?: return VcsInfo.EMPTY`.
What I don't know is if Maven sets `scm.connection` to null or to empty if an empty `<connection/>` tag is provided in the POM, so this might not make a difference.

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
```diff
@@ -30,6 +30,23 @@ import org.ossreviewtoolkit.model.VcsInfo
...
+                }
+                parent = MavenProject().apply {
+                    scm = Scm().apply {
+                        url = "https://github.com/spring-projects/spring-boot"
+                    }
+                }
+            }
+
+            MavenSupport.getOriginalScm(mavenProject)?.url shouldBe "https://github.com/spring-projects/spring-boot"
+        }
```
mnonnenmacher, line 47: Could you please add a similar test for the connection as well? Or update this test to test both?

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -170,14 +170,14 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
...
-            var scm = mavenProject.scm
+            val scm = mavenProject.scm
             var parent = mavenProject.parent
 
             while (parent != null) {
                 parent.scm?.let { parentScm ->
                     parentScm.connection?.let { parentConnection ->
                         if (parentConnection.isNotBlank() && scm.connection.startsWith(parentConnection)) {
-                            scm = parentScm
+                            scm.connection = parentScm.connection
```
mnonnenmacher, line 180: Not directly related to this change, but I just noticed that the function docs are confusing because they only talk about the "SCM URL" when actually so far only the connection is compared. After this PR both the URL and connection are processed, could you maybe add a commit to update the documentation?

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: Good points. More thoughts: Maybe prefer `takeUnless { it.isEmpty() }` to express that the expected default is to take the value. Or be a bit more explicit and use

            val connection = scm?.connection
            if (connection.isNullOrEmpty()) return VcsInfo.EMPTY

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: Commit message:
- "child'd" -> "child's"
- no comma in "was, that"

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
```diff
@@ -30,6 +30,27 @@ import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
 
 class MavenSupportTest : WordSpec({
+    "getOriginalScm()" should {
+        "return the parent's SCM connection if the child SCM uses the parent's SCM connection implicitly" {
+            val mavenProject = MavenProject().apply {
+                scm = Scm().apply {
+                    connection = "https://github.com/spring-projects/spring-boot/childArtifactName"
```
sschuberth, line 37: Even if it does not matter for this test, please use the correct semantics for the `connection` field and prefix it with `scm:git:` and use the real clone URL that ends in `.git`. That way it also can't be mixed up with the `url` field (like right now in line 42 and 43).

sschuberth (changes requested):

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
MarcelBochtler: When using these semantics, would it look like this?

The child gets an SCM connection from Maven like this:
`scm:git:https://github.com/spring-projects/spring-boot/childArtifactName.git`
if the parent uses:
`scm:git:https://github.com/spring-projects/spring-boot.git`

If this is how it works, the fixup does no longer work.

I guess it is time to prepare an example project and to write a funTest to find out.

MarcelBochtler: Ok, I tested it, and Maven really uses this SCM connection for the child:
```
scm:git:https://example.com/git/repository.git/maven-app
```

mnonnenmacher (dismissed):

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: There's still one more "child'd".

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
```diff
@@ -49,6 +49,25 @@ class MavenSupportTest : WordSpec({
...
+
+        "return the parent's SCM URL if the child SCM uses the parent's SCM URL implicitly" {
+            val mavenProject = MavenProject().apply {
+                scm = Scm().apply {
+                    connection = "scm:git:https://github.com/oss-review-toolkit/childConnection.git"
+                    url = "https://github.com/spring-projects/spring-boot/childArtifactName"
+                }
+                parent = MavenProject().apply {
+                    scm = Scm().apply {
+                        connection = "scm:git:https://github.com/oss-review-toolkit/parent/parentConnection.git"
```
sschuberth, line 61: Shouldn't "/parent" be dropped from here? (Otherwise there are 3 path components in here.)

analyzer/src/funTest/assets/projects/synthetic/maven-expected-output-app.yml
```diff
@@ -12,8 +12,8 @@ project:
     mapped:
       Apache License, Version 2.0: "Apache-2.0"
   vcs:
-    type: ""
-    url: ""
+    type: "Git"
```
sschuberth, line 15: FYI, this test also passes when cherry-picked to `main` without any of the other changes in this PR... so I'm not sure what it proves.

MarcelBochtler: Yes, because this test only checks if the replacement works with `connection`, which was before not tested in a `funTest`.
IMO, using the connection is the preferred way by Maven, and the test did help me to debug the issue.

To test the same functionality with the `url` would require me to create an additional test project, which I'd like to avoid, as the functionality is also covered by unit tests.

mnonnenmacher (approved):

sschuberth (approved):

PR #5951, opened 2022-10-14
analyzer: Improve Git protocol in VCS processed management

fb33:
Add management of urls like `git@github.com:someOrg/someProject.git`

Signed-off-by: Barbe François <f.barbe@lectra.com>

### DIFF
--- a/utils/ort/src/main/kotlin/Utils.kt
+++ b/utils/ort/src/main/kotlin/Utils.kt
@@ -181,6 +181,9 @@ fun normalizeVcsUrl(vcsUrl: String): String {
     if (url.startsWith("git://")) {
         url = "https://${url.removePrefix("git://")}"
     }
+    if (url.startsWith("git@") && url.count { it == ':' } == 1) {
+        url = "https://${url.replace(':', '/').removePrefix("git@")}"
+    }
 
     // URLs to Git repos may omit the scheme and use an SCP-like URL that uses ":" to separate the host from the path,
     // see https://git-scm.com/docs/git-clone#_git_urls_a_id_urls_a. Make this an explicit ssh URL, so it can be parsed


utils/ort/src/main/kotlin/Utils.kt
```diff
@@ -181,6 +181,9 @@ fun normalizeVcsUrl(vcsUrl: String): String {
     if (url.startsWith("git://")) {
         url = "https://${url.removePrefix("git://")}"
     }
+    if (url.startsWith("git@") && url.count { it == ':' } == 1) {
```
sschuberth, line 184: This turns SCP-like SSH URLs into HTTPS URLs, which IMO is not correct (we must not silently change the SSH to the HTTPS protocol). Turning these SCP-like URLs into regular SSH URLs is already handled in line 197.

Can you elaborate a bit more what problem you're trying to solve?

sschuberth (changes requested):

utils/ort/src/main/kotlin/Utils.kt
fb33: Yes, I was thinking that [4c57907](https://github.com/oss-review-toolkit/ort/commit/4c57907f5f3c95f229f824c8cd802ce0efdf660b) was to force all package git urls to be used with https.
But this part still forces the ssh
https://github.com/oss-review-toolkit/ort/blob/4c57907f5f3c95f229f824c8cd802ce0efdf660b/utils/ort/src/main/kotlin/Utils.kt#L193-L196
so I add this condition ` if (url.startsWith("git@") && url.count { it == ':' } == 1)` to be sure not to override the code L193-L196
but we could also change this code to : 
```
    // Fixup scp-like Git URLs that do not use a ':' after the server part.
    if (url.startsWith("git@") && !url.contains(":") {
        url = "ssh://$url"
    }
```

sschuberth: > Yes, I was thinking that [4c57907](https://github.com/oss-review-toolkit/ort/commit/4c57907f5f3c95f229f824c8cd802ce0efdf660b) was to force all package git urls to be used with https.

No, that's not the case. Git repos can be cloned via 3 different protocols: `git://`, `ssh://` and `https://`. Commit [4c57907](https://github.com/oss-review-toolkit/ort/commit/4c57907f5f3c95f229f824c8cd802ce0efdf660b) does not change *all* Git repos to be cloned via `https://`, but only those that used `git://` before, as `git://` is a legacy protocol that does not support authentication and was deprecated on GitHub and other VCS hosting sites.

Both `ssh://` and `https://` support authentication, and we should not change these. Finally, the SCP-like syntax `[user@]host.xz:path/to/repo.git/` (as in `git@github.com:someOrg/someProject.git`) is just an [alternative way of writing](https://git-scm.com/docs/git-clone#_git_urls) `ssh://[user@]host.xz[:port]/path/to/repo.git/`; so the former is *not* using the `git://` protocol.

TL;DR I believe we should close this PR unmerged.

PR #5952, opened 2022-10-14
Pub: Use stdout instead of stderr to get version

MarcelBochtler:
The version is printed on stdout, not on stderr. This fixes the issue
that no version is detected when running `requirements.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -601,7 +601,7 @@ class Pub(
     override fun getVersion(workingDir: File?): String {
         val result = ProcessCapture(workingDir, command(workingDir), getVersionArguments()).requireSuccess()
 
-        return transformVersion(result.stderr)
+        return transformVersion(result.stdout)
     }
 
     override fun command(workingDir: File?): String =


analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -601,7 +601,7 @@ class Pub(
     override fun getVersion(workingDir: File?): String {
         val result = ProcessCapture(workingDir, command(workingDir), getVersionArguments()).requireSuccess()
 
-        return transformVersion(result.stderr)
+        return transformVersion(result.stdout)
```
sschuberth, line 604: Commit message: Missing final backtick after "requirements".

Out of curiosity, if this was wrong, how could the version requirement ever have passed?

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/Pub.kt
MarcelBochtler: > Out of curiosity, if this was wrong, how could the version requirement ever have passed?

I have no idea, when I implemented this check I was also confused why the version was printed on stderr. My best guess is, that this changed with some Flutter / Dart version.

sschuberth (approved):

PR #5955, opened 2022-10-17
Gradle: Allow Kotlin script files in source root again

sschuberth:
Kotlin 1.7.20 changed its behavior and warns about Kotlin script files that are "used along with regular Kotlin sources", which fails the build as ORT treats warnings as errors. Restore the old behavior to allow such script files.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -254,6 +254,7 @@ subprojects {
 
     tasks.withType<KotlinCompile>().configureEach {
         val customCompilerArgs = listOf(
+            "-Xallow-any-scripts-in-source-roots",
             "-Xallow-result-return-type",
             "-opt-in=kotlin.contracts.ExperimentalContracts",
             "-opt-in=kotlin.io.path.ExperimentalPathApi",


build.gradle.kts
```diff
@@ -254,6 +254,7 @@ subprojects {
 
     tasks.withType<KotlinCompile>().configureEach {
         val customCompilerArgs = listOf(
+            "-Xallow-any-scripts-in-source-roots",
```
fviernau, line 257: commit-msg: Could you provide a rationale _why_ this change is done?

fviernau (changes requested):

build.gradle.kts
sschuberth: I'm (honestly) unsure what beyond

> Kotlin 1.7.20 changed its behavior and warns about Kotlin script files that are "used along with regular Kotlin sources", which fails the build as ORT treats warnings as errors. Restore the old behavior to allow such script files.

I could say. The rationale is to restore the pre-Kotlin-1.7.20 behavior to again be able to compile scripts that are located in source root, like resource folders, without any warnings.

fviernau: I was expecting a rationale "why" you allow scripts in the root. Anyhow, I think it's good enough.

sschuberth: Maybe the term "root" is misleading here, as what Kotlin means actually is "any directory below a source root" (in the IDE sense), which includes directories like "rules" inside "resource" directories.

For some semi-related discussion maybe also see the comments at https://youtrack.jetbrains.com/issue/KTIJ-22806.

PR #5958, opened 2022-10-17
model: Remove support for Hocon configuration files

mnonnenmacher:
Remove the support for Hocon configuration files as announced in 211930c.

**Breaking change:**
Hocon configuration files are not supported anymore, you have to migrate to YAML. To show your Hocon configuration as YAML run `ort config --show-active` with a previous ORT version.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -103,7 +103,6 @@ graphQlKtorClient = { module = "com.expediagroup:graphql-kotlin-ktor-client", ve
 greenmail = { module = "com.icegreen:greenmail", version.ref = "greenmail" }
 hikari = { module = "com.zaxxer:HikariCP", version.ref = "hikari" }
 hopliteCore = { module = "com.sksamuel.hoplite:hoplite-core", version.ref = "hoplite" }
-hopliteHocon = { module = "com.sksamuel.hoplite:hoplite-hocon", version.ref = "hoplite" }
 hopliteYaml = { module = "com.sksamuel.hoplite:hoplite-yaml", version.ref = "hoplite" }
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -163,6 +162,6 @@ xz = { module = "org.tukaani:xz", version.ref = "xz" }
 
 [bundles]
 exposed = ["exposedCore", "exposedDao", "exposedJdbc", "exposedJavaTime"]
-hoplite = ["hopliteCore", "hopliteHocon", "hopliteYaml"]
+hoplite = ["hopliteCore", "hopliteHocon"]
```
sschuberth, line 165: Should this drop `"hopliteHocon"` instead?

mnonnenmacher: Yes, of course :)

sschuberth (approved):

model/src/test/kotlin/config/OrtConfigurationTest.kt
```diff
@@ -310,20 +302,15 @@ class OrtConfigurationTest : WordSpec({
...
-                }
+                ort:
+                  scanner:
+                    storages:
+                      postgresStorage:
+                        connection:
+                          url: "postgresql://your-postgresql-server:5444/your-database"
+                          schema: public
+                          username: ${'$'}{POSTGRES_USERNAME}
+                          password: ${'$'}{POSTGRES_PASSWORD}
```
sschuberth, line 313: Just checking: Is this syntax really still correct with YAML?

mnonnenmacher: It's required for Kotlin to escape `$` in a string literal.

PR #5970, opened 2022-10-19
Add some scanning progress information

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -338,12 +338,13 @@ class Scanner(
      */
     private fun runPathScanners(controller: ScanController, context: ScanContext) {
         controller.getAllProvenances().forEach { provenance ->
-            val scannersForProvenance =
-                controller.getPathScanners().filterNot { controller.hasScanResult(it, provenance) }
+            val scannersWithoutResults = controller.getPathScanners().filterNot {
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/Scanner.kt
```diff
@@ -234,7 +234,13 @@ class Scanner(
      * Run package scanners for packages with incomplete scan results.
      */
     private fun runPackageScanners(controller: ScanController, context: ScanContext) {
-        controller.getPackagesConsolidatedByProvenance().forEach { (provenance, packages) ->
+        val packagesByProvenance = controller.getPackagesConsolidatedByProvenance()
+
+        packagesByProvenance.onEachIndexed { index, entry ->
```
sschuberth, line 239: FYI: There is no `forEachIndexed` for maps.

mnonnenmacher: You could still do `onEachIndexed { index, (provenance, packages) ->`.

mnonnenmacher (changes requested):

scanner/src/main/kotlin/Scanner.kt
sschuberth: Ah, great! I was pretty sure I checked this some time ago, and I believe back then that syntax was not supported. But good to see that it is now!

mnonnenmacher (approved):

Issue #5105, opened 2022-02-25
Add the concept of "trusted" vendors / frameworks / SDKs


sschuberth:
When scanning code for e.g. Android or iOS apps, there are some reoccurring fundamental SDK dependencies which usually contains a *lot* of source code. It makes little sense to scan these dependencies over and over again for (small) version changes.

To address the issue, I believe it's fair to assume that if a company decided to develop e.g. Android / iOS apps, the (business) decision to accept the use of the *required* SDKs has effectively already been made. With that in mind, it would be nice if ORT had a concept of "trusted vendors" where a copyright holder and license need to be specified for whole SDK / framework, and then scanning for all belonging / matching packages would be skipped.

This is conceptually similar to the existing mechanism of curating `authors` and `concludedLicense` in conjunction with enabling `skipConcluded`. A reason why this existing mechanism is not sufficient is that curations can only use ranges / wildcards as part of the `version`. However, in order to trust e.g. the Amazon AWS SDK as a whole, which includes a component like `Maven:com.amazonaws:aws-java-sdk-models:1.12.162`, we'd also need to allow wildcards for the `name` (at least), so that `authors` and `concludedLicense` could be curated for something like `Maven:com.amazonaws:aws-*:*`.

So a proposal to support the concept of "trusted" vendors / frameworks / SDKs would be to extend the curation logic to also accept wildcards / regexes for the `name`.

sschuberth:
Another proposal just came to my mind: "Trusted" in this context should mean that we trust the *declared* license, and thus also do not need to scan to get the detected licenses, even if no explicit concluded license is present.

sschuberth:
It has slipped my mind that we already [support the name of a package curation to be empty](https://github.com/oss-review-toolkit/ort/blob/034eba71f22d391efde8ac3551ffbc7a0d495801/model/src/main/kotlin/PackageCuration.kt#L53), so we can conclude the license (and author) for whole namespaces like `Maven:com.amazonaws` or `Maven:com.android.tools`.

So, it that mechanism maybe already good enough?

Note that this would still require separate curations for e.g. `Maven:com.google.android.datatransport` and `Maven:com.google.android.gms` etc., and we could not say `Maven:com.google.android.gms.*` (or similar).

sschuberth:
> So, it that mechanism maybe already good enough?

I believe yes. Also see https://github.com/oss-review-toolkit/ort/pull/5971.

sschuberth:
[...]

PR #5971, opened 2022-10-19
Test curations with empty name / version

sschuberth:

### DIFF
--- a/model/src/test/kotlin/PackageCurationTest.kt
+++ b/model/src/test/kotlin/PackageCurationTest.kt
@@ -19,6 +19,7 @@
 
 package org.ossreviewtoolkit.model
 
+import io.kotest.assertions.assertSoftly
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.collections.containExactlyInAnyOrder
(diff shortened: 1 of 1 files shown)


model/src/test/kotlin/PackageCurationTest.kt
```diff
@@ -409,6 +410,22 @@ class PackageCurationTest : WordSpec({
     }
 
     "isApplicable()" should {
+        "accept empty an name and / or version" {
```
mnonnenmacher, line 413: Typo: "an empty"

model/src/test/kotlin/PackageCurationTest.kt
```diff
@@ -409,6 +410,22 @@ class PackageCurationTest : WordSpec({
...
+                data = PackageCurationData(
+                    authors = sortedSetOf("The Android Open Source Project"),
+                    concludedLicense = "Apache-2.0".toSpdx()
+                )
+            )
+
+            assertSoftly {
+                curation.isApplicable(Identifier("Maven:com.android.tools:common:25.3.0")) shouldBe true
+                curation.isApplicable(Identifier("Maven:com.android.tools:common")) shouldBe true
+                curation.isApplicable(Identifier("Maven:com.android.tools")) shouldBe true
```
mnonnenmacher, line 425: Add `curation.isApplicable(Identifier("Maven:com.android.tools::25.2.0")) shouldBe true` to test all cases (IIRC).

mnonnenmacher (changes requested):

mnonnenmacher (approved):

model/src/test/kotlin/PackageCurationTest.kt
```diff
@@ -409,6 +410,23 @@ class PackageCurationTest : WordSpec({
     }
 
     "isApplicable()" should {
+        "accept an empty name and / or version" {
+            val curation = PackageCuration(
+                id = Identifier("Maven:com.android.tools"),
+                data = PackageCurationData(
+                    authors = sortedSetOf("The Android Open Source Project"),
```
fviernau, line 417: Setting authors and concluded license is not needed and should be dropped.

fviernau (changes requested):

model/src/test/kotlin/PackageCurationTest.kt
sschuberth: Actually, I added this deliberately even if not strictly needed to showcase the idea described in the [comment linked from the commit message](https://github.com/oss-review-toolkit/ort/issues/5105#issuecomment-1108228824). I'll also add a code comment.

fviernau: The test is about the matching, not about overriding the `authors` and `concluded license` property. 
I doubt the reader of the test will be aware of the comment link you provided above.
So, from the perspective of that reader the data would be irrelevant as well I guess.

fviernau (dismissed):

fviernau (approved):

PR #5981, opened 2022-10-21
Docker: Remove pyenv

mnonnenmacher:


Dockerfile
```diff
@@ -125,20 +125,21 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
...
 ARG SCANCODE_VERSION=30.1.0
 
-# Scancode need restrict commoncode  version
 RUN pip install -U \
     pip=="${PIPTOOL_VERSION}" \
     wheel \
     && pip install -U \
     Mercurial \
-    commoncode==30.0.0 \
+    commoncode=="${COMMONCODE_VERSION}" \
```
sschuberth, line 142: Couldn't we drop the manual install of `commoncode` completely by using the requirements file like we do in the legacy Dockerfile:

https://github.com/oss-review-toolkit/ort/blob/00fe1a2fbe0f575e993c137c8825477ad81a991d/docker/legacy/Dockerfile#L153-L157

sschuberth (approved):

heliocastro:
This is really a bad idea.

We just made a change recently to move cargo/rust away from distro specifics because issues with version dependency
Now you doing the reverse move for python, making it totally distro depended.

Is there any specific reason to do that beyond "save time" ?

heliocastro (changes requested):
Avoid create another distro direct dependency.

sschuberth:
> We just made a change recently to move cargo/rust away from distro specifics because issues with version dependency Now you doing the reverse move for python, making it totally distro depended.
> 
> Is there any specific reason to do that beyond "save time" ?

The rationale is given in the message of the first commit:

```
Use the default Python 3 of the Ubuntu distribution instead of
installing it with pyenv. With the introduction of python-inspector it
is not required anymore to have a configurable Python version.
```

So if no specific Python version is required anymore, why do we need to keep the logic to install any Python version independently of the Linux distribution?

heliocastro:
The same rationale as cargo.
If python-inspector for some reason needs a new version of python, we will be screwed as same way happened with cargo.
So, i'm still with the opinion that is a mistake to do, and if the reason is speed of the build, i have the final step to fix this cooking, just need change the priorities. @tsteenbe should i look for speed up the build first than the OSS sources ?

sschuberth:
> If python-inspector for some reason needs a new version of python, we will be screwed as same way happened with cargo.

BTW, I believe we should eventually move towards running `python-inspector` with Jython or better GraalVM anyway.

> and if the reason is speed of the build, i have the final step to fix this cooking

I believe equally important reasons are to reduce complexity by removing unneeded dependencies, but @mnonnenmacher might want to chime in here about further reasons.

mnonnenmacher:
> > If python-inspector for some reason needs a new version of python, we will be screwed as same way happened with cargo.
> 
> BTW, I believe we should eventually move towards running `python-inspector` with Jython or better GraalVM anyway.

I think it's very unlikely that `python-inspector` will ever not run with the default Python version of the current Ubunut LTS release, but I'm open to keeping it, see below.

> > and if the reason is speed of the build, i have the final step to fix this cooking
> 
> I believe equally important reasons are to reduce complexity by removing unneeded dependencies, but @mnonnenmacher might want to chime in here about further reasons.

Yes, reducing complexity is one point, but build performance is currently the far more pressing issue to me, although I did not measure if this PR does actually affect it significantly. @heliocastro I would be curious to learn which performance improvements you have in the pipeline.

Btw, I am converting this to a draft because there are failing tests and I have no time to look into it.

Issue #5973, opened 2022-10-20
Filter environment variables passed to ProcessCapture to increase security


oheger-bosch:
This is an idea how to reduce a security risk when running ORT, for which I would like to get some feedback:

Especially when running the Analyzer, external logic in build scripts is executed which could do potentially harmful thing. One attack vector could be reading sensitive information from environment variables, such as database credentials or credentials to other services provided to the ORT process.

At least when external processes are created via the `ProcessCapture` class, this risk could be mitigated by filtering the map with environment variables passed to the new process. Since the variables are specific to a concrete runtime environment, probably a configurable filter mechanism is required. For instance, users could define inclusion and/or exclusion filters in the ORT configuration file that are applied when constructing the environment. This would support use cases such as including only variables starting with a specific prefix (which are explicitly intended to be used by external tools) or removing variables with "postgres" in their name which are specific for the database.

This is of course not bullet-proof, but would be an improvement in this area.

oheger-bosch:
The topic has been discussed in the ORT Community Meeting on 2022-10-20 with the following outcome:
* There is general agreement that a solution addressing this issue would be useful.
* The most secure solution - that should be followed here - is a whitelist approach. This means:
  * In the ORT configuration, users have to explicitly list the environment variables which should be passed through to external processes. All others are removed from the environment when starting an external process.
  * ORT provides a default for this setting that includes a minimum set of variables that are known to be required by the tools used, such as `GRADLE_HOME`, `MAVEN_HOME`, `M2_REPO`, etc.
* For the beginning, a global whitelist is sufficient. Later on, tool-specific whitelists may be supported.

oheger-bosch:
The approach of listing all allowed variables used by the various tools is hardly practicable. In a first (and incomplete) attempt to compile such a list, I came up with more than 300 variables. This list would also have to be maintained and adapted to changes in the supported tools. So, while this may be the most secure approach, I doubt that it is feasible.

Therefore, I would suggest an alternative approach: ORT could define some patterns for variables it excludes by default, e.g. ".*PASS.*", ".*USER.*", ".*_KEY". The user could then configure a list of variables that match a pattern, but should be nevertheless included.

nnobelis:
This is a good idea. `.*TOKEN.*` too.

PR #5984, opened 2022-10-21
Filter environment variables when creating child processes

oheger-bosch:
This PR addresses #5973. Refer to the single commits for further details.

### DIFF
--- a/cli/src/funTest/kotlin/OrtMainFunTest.kt
+++ b/cli/src/funTest/kotlin/OrtMainFunTest.kt
@@ -37,6 +37,7 @@ import org.ossreviewtoolkit.model.config.OrtConfiguration
 import org.ossreviewtoolkit.model.config.OrtConfigurationWrapper
 import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.model.writeValue
+import org.ossreviewtoolkit.utils.common.EnvironmentVariableFilter
 import org.ossreviewtoolkit.utils.common.redirectStdout
 import org.ossreviewtoolkit.utils.ort.normalizeVcsUrl
 import org.ossreviewtoolkit.utils.test.createSpecTempFile
@@ -201,6 +202,18 @@ class OrtMainFunTest : StringSpec() {
                     "--rules-resource", "DUMMY"
                 )
             }
+        }
+
+        "EnvironmentVariableFilter is correctly initialized" {
+            val referenceConfigFile = File("../model/src/main/resources/reference.yml").absolutePath
+            runMain(
+                "-c",
+                referenceConfigFile,
+                "config"
+            )
+
+            EnvironmentVariableFilter.isAllowed("PASSPORT") shouldBe true
+            EnvironmentVariableFilter.isAllowed("DB_PASS") shouldBe false
         }
     }
 
--- a/model/src/main/kotlin/config/OrtConfiguration.kt
+++ b/model/src/main/kotlin/config/OrtConfiguration.kt
@@ -29,6 +29,7 @@ import java.io.File
 import org.apache.logging.log4j.kotlin.Logging
 
 import org.ossreviewtoolkit.model.Severity
+import org.ossreviewtoolkit.utils.common.EnvironmentVariableFilter
 
 /**
  * The configuration model for all ORT components.
@@ -67,6 +68,19 @@ data class OrtConfiguration(
      * configurations from the .ort.yml take precedence.
      */
     val enableRepositoryPackageConfigurations: Boolean = false,
+
+    /**
+     * A set with substrings to filter out environment variables before creating child processes to prevent that those
+     * processes can access sensitive information. See [EnvironmentVariableFilter] for further details.
+     */
+    val deniedProcessEnvironmentVariablesSubstrings: Set<String> = EnvironmentVariableFilter.DEFAULT_DENY_SUBSTRINGS,
+
+    /**
+     * A set with the names of environment variables that are explicitly allowed to be passed to child processes,
+     * even if they are matched by one of the [deniedProcessEnvironmentVariablesSubstrings].
+     * See [EnvironmentVariableFilter] for further details.
+     */
+    val allowedProcessEnvironmentVariableNames: Set<String> = EnvironmentVariableFilter.DEFAULT_ALLOW_NAMES,
 
     /**
      * The configuration of the analyzer.
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -30,6 +30,15 @@ ort:
   enableRepositoryPackageCurations: true
   enableRepositoryPackageConfigurations: true
 
+  deniedProcessEnvironmentVariablesSubstrings:
+    - PASS
+    - SECRET
+    - TOKEN
+    - USER
+  allowedProcessEnvironmentVariableNames:
+    - PASSPORT
+    - USER_HOME
+
   analyzer:
     allowDynamicVersions: true
 
--- a/utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
+++ b/utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2022 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.utils.common
+
+/**
+ * An object providing functionality to filter environments that are passed to newly created processes.
+ *
+ * For many tasks, ORT spawns new processes using the [ProcessCapture] class. When creating a new process, the child
+ * process by default inherits all environment variables from the parent. This could impose a security risk, for
+ * instance if logic in build scripts could theoretically access sensitive information stored in environment
+ * variables, such as database or service credentials.
+ *
+ * To reduce this risk, this object filters the environment variables passed to child processes based on the
+ * following criteria:
+ * - Substrings for variable names can be defined to determine variables with sensitive information. The object provides
+ *   some default strings to match variable names like "PASS", "USER", "TOKEN", etc.
+ * - There is an allow list to include variables even if they contain one of these substrings.
+ *
+ * So in order to determine whether a specific variable "E" can be passed to a child process, this filter applies the
+ * following steps:
+ * - If E is contained in the allow list, it is included.
+ * - Otherwise, E is included if and only if its name does not contain one of the exclusion substrings (ignoring case).
+ *
+ * TODO: Find an alternative mechanism to initialize this object from the ORT configuration (maybe using dependency
+ *       injection) which does not require this object to be public.
+ */
+object EnvironmentVariableFilter {
+    /**
+     * A set with substrings contained in variable names that are denied by default. All variables containing one of
+     * these strings (ignoring case) are not propagated to child processes.
+     */
+    val DEFAULT_DENY_SUBSTRINGS = setOf(
+        "key",
+        "pass",
+        "pwd",
+        "token",
+        "user"
+    )
+
+    /** A set of known variable names that are allowed despite being matched by deny substrings. */
+    val DEFAULT_ALLOW_NAMES = setOf(
+        "CARGO_HTTP_USER_AGENT",
+        "COMPOSER_ALLOW_SUPERUSER",
+        "CONAN_LOGIN_ENCRYPTION_KEY",
+        "CONAN_LOGIN_USERNAME",
+        "CONAN_PASSWORD",
+        "CONAN_USERNAME",
+        "CONAN_USER_HOME",
+        "CONAN_USER_HOME_SHORT",
+        "DOTNET_CLI_CONTEXT_ANSI_PASS_THRU",
+        "GIT_ASKPASS",
+        "GIT_HTTP_USER_AGENT",
+        "GRADLE_USER_HOME",
+        "HACKAGE_USERNAME",
+        "HACKAGE_PASSWORD",
+        "HACKAGE_KEY",
+        "PWD",
+        "USER",
+        "USERPROFILE"
+    )
+
+    /** Stores the current deny substrings used by this filter. */
+    private var denySubstrings = DEFAULT_DENY_SUBSTRINGS
+
+    /** Stores the current allow names used by this filter. */
+    private var allowNames = DEFAULT_ALLOW_NAMES
+
+    /**
+     * Reset this filter to use the given [denySubstrings] and [allowNames].
+     */
+    fun reset(
+        denySubstrings: Collection<String> = DEFAULT_DENY_SUBSTRINGS,
+        allowNames: Collection<String> = DEFAULT_ALLOW_NAMES
+    ) {
+        this.denySubstrings = denySubstrings.toSet()
+        this.allowNames = allowNames.toSet()
+    }
+
+    /**
+     * Test whether the variable with the given [name] can be passed to a child process according to the criteria
+     * described in the header comment.
+     */
+    fun isAllowed(name: String): Boolean {
+        return name in allowNames || denySubstrings.none { name.contains(it, ignoreCase = true) }
+    }
+
+    /**
+     * Remove all keys from [environment] that do not pass this filter.
+     */
+    fun filter(environment: MutableMap<String, String>): MutableMap<String, String> {
+        val keysToRemove = environment.keys.filterNot(EnvironmentVariableFilter::isAllowed)
+
+        @Suppress("ConvertArgumentToSet") // The list cannot contain duplicates.
+        environment -= keysToRemove
+
+        return environment
+    }
+}
--- a/utils/common/src/main/kotlin/ProcessCapture.kt
+++ b/utils/common/src/main/kotlin/ProcessCapture.kt
@@ -84,7 +84,9 @@ class ProcessCapture(
         .redirectOutput(stdoutFile)
         .redirectError(stderrFile)
         .apply {
-            environment().putAll(environment)
+            // Apply the environment filter to the map with inherited variables, but not to the variables added
+            // explicitly via the environment parameter, since these are expected to be always relevant for the command.
+            EnvironmentVariableFilter.filter(environment()).putAll(environment)
         }
 
     val commandLine = command.joinToString(" ")
--- a/utils/common/src/test/kotlin/EnvironmentVariableFilterTest.kt
+++ b/utils/common/src/test/kotlin/EnvironmentVariableFilterTest.kt
@@ -1,0 +1,83 @@
+/*
+ * Copyright (C) 2022 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.utils.common
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.inspectors.forAll
+import io.kotest.matchers.maps.containExactly
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.types.beTheSameInstanceAs
+
+class EnvironmentVariableFilterTest : StringSpec({
+    afterAny {
+        // Reset default values after each test.
+        EnvironmentVariableFilter.reset()
+    }
+
+    "Uncritical variables should be accepted" {
+        listOf("FOO", "bar", "boring", "uncritical", "PATH").forAll { variable ->
+            EnvironmentVariableFilter.isAllowed(variable) shouldBe true
+        }
+    }
+
+    "Variables should be filtered based on default deny substrings" {
+        listOf("DATABASE_PASSWORD", "repositoryToken", "service-user", "APIKEY").forAll { variable ->
+            EnvironmentVariableFilter.isAllowed(variable) shouldBe false
+        }
+    }
+
+    "Variables should be accepted if they are on the allowed list" {
+        listOf("USER", "USERPROFILE", "GRADLE_USER_HOME", "GIT_HTTP_USER_AGENT").forAll { variable ->
+            EnvironmentVariableFilter.isAllowed(variable) shouldBe true
+        }
+    }
+
+    "Denied substrings can be changed" {
+        EnvironmentVariableFilter.reset(listOf("foo"))
+
+        EnvironmentVariableFilter.isAllowed("ENV_FOO") shouldBe false
+        EnvironmentVariableFilter.isAllowed("MY_PASSWORD") shouldBe true
+    }
+
+    "Allowed names can be changed" {
+        val allowedSecret = "MAVEN_REPO_PASSWORD"
+
+        EnvironmentVariableFilter.reset(allowNames = setOf(allowedSecret))
+
+        EnvironmentVariableFilter.isAllowed(allowedSecret) shouldBe true
+    }
+
+    "The environment map can be filtered" {
+        val environment = mutableMapOf(
+            "USER" to "scott",
+            "password" to "tiger",
+            "foo" to "bar"
+        )
+        val filteredEnvironment = mapOf(
+            "USER" to "scott",
+            "foo" to "bar"
+        )
+
+        EnvironmentVariableFilter.filter(environment) should beTheSameInstanceAs(environment)
+
+        environment should containExactly(filteredEnvironment)
+    }
+})
--- a/utils/common/src/test/kotlin/ProcessCaptureTest.kt
+++ b/utils/common/src/test/kotlin/ProcessCaptureTest.kt
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.utils.common
 
 import io.kotest.core.spec.style.StringSpec
+import io.kotest.extensions.system.withEnvironment
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.string.shouldContain
 
@@ -34,6 +35,21 @@ class ProcessCaptureTest : StringSpec({
 
         proc.exitValue shouldBe 0
         proc.stdout.trimEnd() shouldBe "This is some path: /foo/bar"
+    }
+
+    "Only allowed environment variables should be passed" {
+        val env = mapOf("DB_USER" to "scott", "DB_PASSWORD" to "tiger", "DB_CONN" to "my-db.example.org")
+
+        withEnvironment(env) {
+            val proc = if (Os.isWindows) {
+                ProcessCapture("cmd.exe", "/c", "echo %DB_USER%:%DB_PASSWORD%.%DB_CONN%")
+            } else {
+                ProcessCapture("sh", "-c", "echo \$DB_USER:\$DB_PASSWORD.\$DB_CONN")
+            }
+
+            proc.exitValue shouldBe 0
+            proc.stdout.trimEnd() shouldBe ":.my-db.example.org"
+        }
     }
 
     "Masked strings should be processed correctly" {
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -45,6 +45,7 @@ import org.ossreviewtoolkit.cli.commands.*
 import org.ossreviewtoolkit.cli.utils.logger
 import org.ossreviewtoolkit.model.config.LicenseFilenamePatterns
 import org.ossreviewtoolkit.model.config.OrtConfiguration
+import org.ossreviewtoolkit.utils.common.EnvironmentVariableFilter
 import org.ossreviewtoolkit.utils.common.Os
 import org.ossreviewtoolkit.utils.common.expandTilde
 import org.ossreviewtoolkit.utils.ort.Environment
@@ -186,6 +187,11 @@ class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
         val ortConfiguration = OrtConfiguration.load(configArguments, configFile)
         currentContext.findOrSetObject { GlobalOptions(ortConfiguration, forceOverwrite) }
         LicenseFilenamePatterns.configure(ortConfiguration.licenseFilePatterns)
+
+        EnvironmentVariableFilter.reset(
+            ortConfiguration.deniedProcessEnvironmentVariablesSubstrings,
+            ortConfiguration.allowedProcessEnvironmentVariableNames
+        )
 
         if (helpAll) {
             registeredSubcommands().forEach {
--- a/model/src/test/kotlin/config/OrtConfigurationTest.kt
+++ b/model/src/test/kotlin/config/OrtConfigurationTest.kt
@@ -24,6 +24,7 @@ import io.kotest.core.TestConfiguration
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.extensions.system.withEnvironment
 import io.kotest.matchers.collections.containExactly
+import io.kotest.matchers.collections.containExactlyInAnyOrder
 import io.kotest.matchers.collections.shouldContainExactly
 import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
 import io.kotest.matchers.maps.containExactly as containExactlyEntries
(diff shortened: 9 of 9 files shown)


model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -68,6 +68,24 @@ data class OrtConfiguration(
...
+     * instead.
+     */
+    val variablesToPropagate: Set<String> = emptySet(),
+
+    /**
+     * A set with the names of additional environment variables that are passed to child processes created by ORT.
+     * Defining this property in a configuration file is the easiest way to extend the set of variables to pass
+     * through, since it is not necessary to redefine all the variables ORT takes into account by default.
+     */
+    val additionalVariablesToPropagate: Set<String> = emptySet(),
```
sschuberth, line 87: I'd prefer to not need this. If above we'd set a `setOf(...)` default variables, if should be easy enough for the user to run `ort config --show-default`, copy over the default, append additional wanted variables, and set the combined collection as the new value.

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -68,6 +68,24 @@ data class OrtConfiguration(
...
+    /**
+     * A set with the names of environment variables that are passed to child processes created by ORT, e.g. when
+     * running external package managers or scanners. All variables not listed here are not passed through. This
+     * property aims to prevent that sensitive information from environment variables like passwords or service
+     * credentials is passed to external tools. As a default value, a larger number of well-known variables evaluated
+     * by the tools invoked by ORT is set. Users can choose to override this by redefining this property in their
+     * configuration file. In order to extend the set, the [additionalVariablesToPropagate] property should be used
+     * instead.
+     */
+    val variablesToPropagate: Set<String> = emptySet(),
```
sschuberth, line 80: I'd prefer if the variable name would contain the word "environment". Maybe something like `allowedProcessEnvironmentVariables`?

model/src/main/resources/defaults.yml
```diff
@@ -0,0 +1,358 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+# License-Filename: LICENSE
+
+# A file with default configuration options used by ORT. These options can be overridden in custom configuration files.
+
+ort:
+  variablesToPropagate:
```
sschuberth, line 21: Phew. Looking at the length of this, I really wonder whether the "allow"-approach is the right one.

In either case we should probably consider to support wildcards or regexes, so one could allow / deny e.g. `CARGO_*`.

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -117,6 +117,9 @@ data class OrtConfiguration(
     val notifier: NotifierConfiguration = NotifierConfiguration()
 ) {
     companion object : Logging {
+        /** The name of the configuration file with the default values. */
+        private const val DEFAULT_CONFIG_FILE = "/defaults.yml"
```
sschuberth, line 121: IIRC at some point we decided to not externalize default values to files, but programmatically have them as default arguments. I believe @fviernau's main argument for this was support in test and refactoring.

utils/common/src/main/kotlin/ProcessCapture.kt
```diff
@@ -60,6 +65,17 @@ class ProcessCapture(
                 message
             }
         }
+
+        /**
+         * Filter out the variables that are not white-listed to be propagated from [environment].
```
sschuberth, line 79: Nit: We should probably try to avoid the white / black list terms, and rather speak of allow / deny lists.

utils/common/src/main/kotlin/ProcessCapture.kt
```diff
@@ -70,12 +79,21 @@ class ProcessCapture(
          * Filter out the variables that are not white-listed to be propagated from [environment].
          */
         private fun filterEnvironmentVariables(environment: MutableMap<String, String>): MutableMap<String, String> {
-            val keysToRemove = environment.keys.filterNot { it in variablesToPropagate }
+            val keys = environment.keys.handleCase()
```
sschuberth, line 82: Alternatively to this, apply `.toSortedMap(String.CASE_INSENSITIVE_ORDER)` to `environment` on Windows before looking up the keys.

sschuberth (changes requested):

model/src/main/resources/reference.yml
```diff
@@ -30,6 +30,13 @@ ort:
   enableRepositoryPackageCurations: true
   enableRepositoryPackageConfigurations: true
 
+  variablesToPropagate:
+    - HOME
+    - JAVA_HOME
+  additionalVariablesToPropagate:
+    - BUILD_PATH
+    - FAST_MODE
+
```
nnobelis, line 39: Should we also propagate  `PATH` ?

When I run the Analyzer locally, I often run it with a modified `PATH` for it to find the different tools.

nnobelis: We should also propagate `USER`, `PWD` and `LANG`.

model/src/main/resources/defaults.yml
```diff
@@ -0,0 +1,358 @@
...
+    - LANG
+    - LS_COLORS
+    - netrc
+    - netrcPath
+    - NO_PROXY
+    - no_proxy
+    - NODE_PATH
+    - NODE_EXTRA_CA_CERTS
+    - PREFIX
+    - TERM
```
nnobelis, line 41: We should also propagate `PATH` and `PWD`.

oheger-bosch: `PATH` really seems to be required. Lots of tests were failing because tools could not be found.

model/src/main/resources/defaults.yml
oheger-bosch: Yes, I was already sceptical when the decision was taken. It will be very difficult to come up with an exhaustive list. In addition, this list will have to be maintained when there are new versions of tools.

Using of wildcards will simplify the configuration, but again increase the risk that an unwanted variable is passed through.

model/src/main/kotlin/config/OrtConfiguration.kt
oheger-bosch: Because of the length of the list, I would not want to define it in `OrtConfiguration`. So, would it be okay to create an object `AllowedProcessEnvironmentVariables` which contains the list and also functionality to deal with regular expressions?

sschuberth: > So, would it be okay to create an object `AllowedProcessEnvironmentVariables` which contains the list and also functionality to deal with regular expressions?

Sounds like the best compromise 😒

model/src/main/resources/defaults.yml
sschuberth: I wonder whether we should come up with something completely different... a "crazy" approach could be to look at the entropy of the value of an environment variable to "guess" whether it's likely a password or so, and only then do not pass it on to the spawned process.

Or, instead of a global list, create tool-specific lists that are used in `CommandLineTool` `object`s. I.e. we should wrap each tool we call in an `object` instead of implementing the interface in the main class itself (see [`MercurialCommand`](https://github.com/oss-review-toolkit/ort/blob/4d90096fb049344092a4c067edcc74f5d1472b58/downloader/src/main/kotlin/vcs/Mercurial.kt#L39-L52) for an example), and each of these objects gets its own (hard-coded) list of allowed environment variables. That should at least ease maintainability a bit.

oheger-bosch: I fear the entropy approach is too clever to work in practice.

If the allowed variables were configured at the command line tools, this could indeed make the maintenance a bit easier. But how could these configurations then be extended? For our use case, we will need to pass additional variables with credentials to package managers.

oheger-bosch:
I have implemented an alternative approach as described in my last comment on #5973. This seems to be significantly easier by (hopefully) offering a comparable level of security.

model/src/main/kotlin/config/OrtConfiguration.kt
oheger-bosch: The mechanism has been completely reworked.

model/src/main/kotlin/config/OrtConfiguration.kt
oheger-bosch: Changed implementation.

model/src/main/kotlin/config/OrtConfiguration.kt
oheger-bosch: The default values and logic is now contained in an object named `EnvironmentVariableFilter`.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,119 @@
...
+ *   some default patterns to match variables containing strings like "PASS", "USER", "TOKEN", etc.
+ * - There is an allow list to include variables even if they are matched by a pattern.
+ *
+ * So in order to determine whether a specific variable "E" can be passed to a child process, this filter applies the
+ * following steps:
+ * - If E is contained in the allow list, it is included.
+ * - Otherwise, E is included if and only if it is not matched by one of the exclusion patterns.
+ */
+object EnvironmentVariableFilter {
+    /** A list with patterns for variable names that are excluded by default. */
```
sschuberth, line 42: Should this be documented to be be case-insensitive?

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,119 @@
...
+    val DEFAULT_EXCLUSION_PATTERNS = listOf(
```
sschuberth, line 43: Should this also be a set to not imply any meaning to the order?

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,119 @@
...
+        "key",
+        "pass",
+        "pwd",
+        "token",
+        "user"
+    )
+
+    /** A list with known variable names that are accepted despite they are matched by exclusion patterns. */
+    val DEFAULT_ALLOW_LIST = setOf(
```
sschuberth, line 52: I would align the terms used as part of the variable names here and above: When using "allow" here use "deny" above, when using "exclusion" (or IMO better "exclude") above, use "include" here.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,119 @@
...
+    private var exclusionPatterns = DEFAULT_EXCLUSION_PATTERNS.toPatterns()
+
+    /** Stores the current allow list used by this filter. */
+    private var allowList = DEFAULT_ALLOW_LIST
+
+    /**
+     * Initialize this filter from the given [exclusionPatterns] and [allowList]. The [exclusionPatterns] are
+     * converted to regular expressions, so that they can be matched anywhere in variable names, ignoring case.
+     */
+    fun init(
```
sschuberth, line 83: Please don't call this function `init`. That's way too easy to confuse with the `init {}` block. Maybe `reset`?

Maybe this `object` should become a `class` after all, with a factory method in a `companion object` that creates a default instance?

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,119 @@
...
+        this.exclusionPatterns = exclusionPatterns.toPatterns()
+        this.allowList = allowList.toSet()
+    }
+
+    /**
+     * Test whether the variable with the given [name] can be passed to a child process according to the criteria
+     * described in the header comment.
+     */
+    fun acceptVariable(name: String): Boolean {
+        return name in allowList || exclusionPatterns.all { !it.matches(name) }
```
sschuberth, line 96: Use `.none {}` instead for readability.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,119 @@
...
+    ) {
+        this.exclusionPatterns = exclusionPatterns.toPatterns()
+        this.allowList = allowList.toSet()
+    }
+
+    /**
+     * Test whether the variable with the given [name] can be passed to a child process according to the criteria
+     * described in the header comment.
+     */
+    fun acceptVariable(name: String): Boolean {
```
sschuberth, line 95: Maybe better any of these: `isAccepted`, `canBePassed`, `shouldPropagate` or similar.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,119 @@
...
+
+        environment -= keysToRemove.toSet()
+
+        return environment
+    }
+
+    /**
+     * Construct an exclusion pattern regular expression from the given [string][s].
+     */
+    private fun pattern(s: String): Regex = Regex(".*$s.*", RegexOption.IGNORE_CASE)
```
sschuberth, line 113: I believe this can be inline without sacrificing readability.

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -68,6 +69,19 @@ data class OrtConfiguration(
      */
     val enableRepositoryPackageConfigurations: Boolean = false,
 
+    /**
+     * A list with patterns to filter out environment variables before creating child processes to prevent that those
+     * processes can access sensitive information. See [EnvironmentVariableFilter] for further details.
+     */
+    val environmentVariablesExclusionPatterns: List<String> = EnvironmentVariableFilter.DEFAULT_EXCLUSION_PATTERNS,
```
sschuberth, line 76: While the "process" in `allowedProcessEnvironmentVariables` nicely mentions the contetxt in which environment variables are allowed, the `environmentVariablesExclusionPatterns` name lacks that context.

I propose `deniedProcessEnvironmentVariablePatterns` and `allowedProcessEnvironmentVariableNames` to make the variable names nicely resemble each other with "denied vs allowed" and "patterns vs names", and still naming the usage context.

utils/common/src/main/kotlin/ProcessCapture.kt
```diff
@@ -84,7 +84,7 @@ class ProcessCapture(
         .redirectOutput(stdoutFile)
         .redirectError(stderrFile)
         .apply {
-            environment().putAll(environment)
+            EnvironmentVariableFilter.filterEnvironment(environment()).putAll(environment)
```
sschuberth, line 87: Shouldn't the filter be applied on the second `environment` argument?

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,119 @@
...
+     * described in the header comment.
+     */
+    fun acceptVariable(name: String): Boolean {
+        return name in allowList || exclusionPatterns.all { !it.matches(name) }
+    }
+
+    /**
+     * Remove all keys from [environment] that do not pass this filter.
+     */
+    fun filterEnvironment(environment: MutableMap<String, String>): MutableMap<String, String> {
```
sschuberth, line 102: Just call this `filter`. The context of environment variables is clear here.

sschuberth (changes requested):

utils/common/src/main/kotlin/ProcessCapture.kt
oheger-bosch: You are probably right, but I am not 100% sure. My thinking was that environment variables that are explicitly provided when the process is created are probably mandatory for the command to function correctly and should therefore not be filtered. From what I have seen, only at few places environment variables are passed, for instance to propagate the GO path or to disable SSL verification in NPM.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: Documentation was extended to mention this.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: Constants and variables now use the terms "DENY_PATTERNS" and "ALLOW_NAMES".

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: Renamed the function to reset. As discussed during a call, before converting the `object` to a `class`, we wait for the introduction of dependency injection facilities.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: I liked `isAccepted` best.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
sschuberth: See https://github.com/oss-review-toolkit/ort/pull/6037 for some early experiments.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
sschuberth: Actually, with this [other renaming done](https://github.com/oss-review-toolkit/ort/pull/5984#discussion_r1012811315), maybe now better `isAllowed`?

utils/common/src/main/kotlin/ProcessCapture.kt
sschuberth: > environment variables that are explicitly provided when the process is created are probably mandatory for the command to function correctly and should therefore not be filtered

I agree. Could you maybe add a code comment about that? I.e. explain why `environment()` is filtered but `environment` not?

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: Changed to `isAllowed`.

utils/common/src/main/kotlin/ProcessCapture.kt
oheger-bosch: Added a comment.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: Alternatively, this PR could be merged as is, and the conversion from an `object` to a `class` is done afterwards. I guess, the best way forward depends on the time you need for your experiments. As I said, this PR is not that urgent for us, but if it remains open for a longer time, there is always the risk for conflicts.

sschuberth: > Alternatively, this PR could be merged as is, and the conversion from an `object` to a `class` is done afterwards.

Yes, I'd be fine with this, to not put me and my experiments in the position of a bottleneck. I'll review this PR once more from that perspective. But I'd actually appreciate also a review from another Bosch contributor, as the feature seems to be mostly wanted from your side.

oheger-bosch: Thanks, fair enough. And I just noticed that my latest renaming has caused a compilation failure.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,117 @@
...
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.utils.common
+
+/**
+ * An object providing functionality to filter environments that are passed to newly created processes.
+ *
+ * For many tasks, ORT spawns new processes using the [ProcessCapture] class. When creating a new process, the child
+ * process per defaults inherits all environment variables from the parent. This could impose a security risk, for
```
sschuberth, line 26: "by default"

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,117 @@
...
+ * following steps:
+ * - If E is contained in the allow list, it is included.
+ * - Otherwise, E is included if and only if it is not matched by one of the exclusion patterns.
+ */
+object EnvironmentVariableFilter {
+    /**
+     * A set with patterns for variable names that are denied by default. All variables containing one of these
+     * strings (ignoring case) are not propagated to child processes.
+     */
+    val DEFAULT_DENY_PATTERNS = setOf(
```
sschuberth, line 46: Thinking about it, the term "pattern" in the variable name (and also in the docs above) probably is a bit too auspicious as these neither are glob nor regex patterns (I was looking for docs by which principle patterns are matched). So, as this simply is a case-insensitive sub-string search, maybe just use "term"?

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,117 @@
...
+        "HACKAGE_USERNAME",
+        "HACKAGE_PASSWORD",
+        "HACKAGE_KEY",
+        "PWD",
+        "USER",
+        "USERPROFILE"
+    )
+
+    /** Stores the current deny patterns used by this filter. */
+    private var denyPatterns = DEFAULT_DENY_PATTERNS.toPatterns()
```
sschuberth, line 77: BTW, this line confirms me that "patters" is the wrong word for the variable, because if they already were patters, there would be no point in converting them `toPatterns()`.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Remove all keys from [environment] that do not pass this filter.
+     */
+    fun filter(environment: MutableMap<String, String>): MutableMap<String, String> {
+        val keysToRemove = environment.keys.filterNot(EnvironmentVariableFilter::isAllowed)
+
+        environment -= keysToRemove.toSet()
+
+        return environment
```
sschuberth, line 110: Shorter: `= environment.filter { (name, _) -> isAllowed(name) }`

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Convert the strings in this collection to deny pattern regular expressions.
+     */
+    private fun Collection<String>.toPatterns() = map { Regex(".*$it.*", RegexOption.IGNORE_CASE) }
```
sschuberth, line 116: This basically just implements a case-insensitive sub-string search. So why use a regex at after, instead of simply using `denyPatterns.none { it.contains (name, ignoreCase = true) }` above?

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
sschuberth: Or "substrings", also see my comment about not using a regexes below.

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -68,6 +69,19 @@ data class OrtConfiguration(
      */
     val enableRepositoryPackageConfigurations: Boolean = false,
 
+    /**
+     * A set with patterns to filter out environment variables before creating child processes to prevent that those
+     * processes can access sensitive information. See [EnvironmentVariableFilter] for further details.
+     */
+    val deniedProcessEnvironmentVariablesPatterns: Set<String> = EnvironmentVariableFilter.DEFAULT_DENY_PATTERNS,
```
sschuberth, line 76: Remember to eventually also adjust the variable name here.

sschuberth (changes requested):

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: You are right. This is by far easier and completely sufficient. Changed.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: I find my original construct ugly, too. But the problem is that the filtering must be done in place instead of returning a new filtered map. This is because of the API of `ProcessBuilder` which returns a mutable map with environment variables, but does not allow setting this map.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: Now using the term "substrings" (hopefully) everywhere.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
sschuberth: > But the problem is that the filtering must be done in place instead of returning a new filtered map.

Ah, right, sorry for missing that. Not sure if it's nicer, but this should work alternatively:

```
        val i = environment.iterator()
        while (i.hasNext()) {
            if (!isAllowed(i.next().key)) i.remove()
        }

        return environment
```

Actually, I guess I like your solution better as it's more compact. The only thing that bothered me is the `toSet()` conversion. I know that without it we get an inspection hint, but the hint is bogus in this case as there can be no duplicates in the list. So I'd indeed prefer your solution, but with the `toSet()` removed.

sschuberth:
BTW @oheger-bosch, did you forget to push?

oheger-bosch:
> BTW @oheger-bosch, did you forget to push?

I did not manage to finish this yesterday. Sorry for the confusion. Now all changes should be available.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,112 @@
...
+     */
+    val DEFAULT_DENY_SUBSTRINGS = setOf(
+        "key",
+        "pass",
+        "pwd",
+        "token",
+        "user"
+    )
+
+    /** A set with known variable names that are accepted despite they are matched by deny patterns. */
```
sschuberth, line 54: "A set of known variable names that are allowed despite being matched by deny substrings."

utils/common/src/test/kotlin/EnvironmentVariableFilterTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        EnvironmentVariableFilter.reset()
+    }
+
+    "Uncritical variables should be accepted" {
+        listOf("FOO", "bar", "boring", "uncritical", "PATH").forAll { variable ->
+            EnvironmentVariableFilter.isAllowed(variable) shouldBe true
+        }
+    }
+
+    "Variables should be filtered based on default patterns" {
```
sschuberth, line 41: "... based on default deny substrings"

utils/common/src/test/kotlin/EnvironmentVariableFilterTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    "Variables should be accepted if they are on the allowed list" {
+        listOf("USER", "USERPROFILE", "GRADLE_USER_HOME", "GIT_HTTP_USER_AGENT").forAll { variable ->
+            EnvironmentVariableFilter.isAllowed(variable) shouldBe true
+        }
+    }
+
+    "Exclusion patterns can be changed" {
```
sschuberth, line 53: "Denied substrings ..."

utils/common/src/test/kotlin/EnvironmentVariableFilterTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        EnvironmentVariableFilter.reset(listOf("foo"))
+
+        EnvironmentVariableFilter.isAllowed("ENV_FOO") shouldBe false
+        EnvironmentVariableFilter.isAllowed("MY_PASSWORD") shouldBe true
+    }
+
+    "The allow list can be changed" {
```
sschuberth, line 60: "Allowed names ..."

utils/common/src/test/kotlin/ProcessCaptureTest.kt
```diff
@@ -36,6 +37,21 @@ class ProcessCaptureTest : StringSpec({
         proc.stdout.trimEnd() shouldBe "This is some path: /foo/bar"
     }
 
+    "Only allowed environment variables should be passed" {
+        val env = mapOf("DB_USER" to "scott", "DB_PASSWORD" to "tiger", "DB_CONN" to "my-db.example.org")
+
+        withEnvironment(env) {
+            val proc = if (Os.isWindows) {
+                ProcessCapture("cmd.exe", "/c", "echo %DB_USER%:%DB_PASSWORD%.%DB_CONN%")
```
sschuberth, line 45: Just wondering, why are you using `.` as a separator before the `DB_CONN`? For a more realistic example, shouldn't it be `@`?

sschuberth (changes requested):
Just some (hopefully) final nits.

utils/common/src/test/kotlin/ProcessCaptureTest.kt
oheger-bosch: Yes, but the "@" character could be problematic when it is interpreted by the shell. I did not want to mess around with escaping syntax, since for the test it is only relevant that some of the values are replaced while others are not.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,112 @@
...
+ * For many tasks, ORT spawns new processes using the [ProcessCapture] class. When creating a new process, the child
+ * process by default inherits all environment variables from the parent. This could impose a security risk, for
+ * instance if logic in build scripts could theoretically access sensitive information stored in environment
+ * variables, such as database or service credentials.
+ *
+ * To reduce this risk, this object filters the environment variables passed to child processes based on the
+ * following criteria:
+ * - Substrings for variable names can be defined to determine variables with sensitive information. The object provides
+ *   some default strings to match variable names like "PASS", "USER", "TOKEN", etc.
+ * - There is an allow list to include variables even if they contain one of these substrings.
```
nnobelis, line 34: is a _whitelist_...

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * So in order to determine whether a specific variable "E" can be passed to a child process, this filter applies the
+ * following steps:
+ * - If E is contained in the allow list, it is included.
```
nnobelis, line 38: same as above

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,112 @@
...
+     * described in the header comment.
+     */
+    fun isAllowed(name: String): Boolean {
+        return name in allowNames || denySubstrings.none { name.contains(it, ignoreCase = true) }
+    }
+
+    /**
+     * Remove all keys from [environment] that do not pass this filter.
+     */
+    fun filter(environment: MutableMap<String, String>): MutableMap<String, String> {
```
nnobelis, line 104: Having a function that modifies it parameters AND returns the modified version is a bit of codesmell.

I would rather have either an extension function `MutableMap<String, String>.filter` or no return value.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
```diff
@@ -0,0 +1,112 @@
...
+ * - Substrings for variable names can be defined to determine variables with sensitive information. The object provides
+ *   some default strings to match variable names like "PASS", "USER", "TOKEN", etc.
+ * - There is an allow list to include variables even if they contain one of these substrings.
+ *
+ * So in order to determine whether a specific variable "E" can be passed to a child process, this filter applies the
+ * following steps:
+ * - If E is contained in the allow list, it is included.
+ * - Otherwise, E is included if and only if its name does not contain one of the exclusion substrings (ignoring case).
+ */
+object EnvironmentVariableFilter {
```
nnobelis, line 41: Should these objects not be `internal` ?
Theoretically, I could develop a malicious ORT plugins that modifies these global filters ?

sschuberth (approved):

cli/src/funTest/kotlin/OrtMainFunTest.kt
```diff
@@ -202,6 +203,17 @@ class OrtMainFunTest : StringSpec() {
...
+
+        "EnvironmentVariableFilter is correctly initialized" {
+            val referenceConfigFile = File("../model/src/main/resources/reference.yml").absolutePath
+            runMain(
+                "-c",
+                referenceConfigFile,
+                "config"
+            )
+
+            EnvironmentVariableFilter.isAllowed("PASSPORT") shouldBe true
```
nnobelis, line 215: Could you also add a negative test, i.e. something disallowed by default ?

nnobelis (changes requested):

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
sschuberth: We should avoid the terms white- / blacklist (for political correctness...)

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
sschuberth: I like the idea of an extension function.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: The names "allow list" and "deny list" have been carefully chosen.

nnobelis: Understood.

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: Valid point. Currently, the object needs to be public because the filters are initialized by the ORT main function from the configuration. The plan is to change this later with usage of dependency injection.

nnobelis: Ok Then maybe add a TODO ?

utils/common/src/main/kotlin/EnvironmentVariableFilter.kt
oheger-bosch: The parameter is returned to have a single expression in `ProcessCapture` where the filter is applied. I am not sure whether I understand the proposal for an extension function. How would the signature of such an extension function look like and how would it help in this context?

nnobelis: As written `MutableMap<String, String>.filter()` and applied like that:
```
with(EnvironmentVariableFilter) {
    environment().filter().putAll(environment)
}
```
but it is more verbose.

nnobelis: Or remove the return value, you don't use it in `ProcessCapture`.

nnobelis: You're right: since `environment()` is a `MutableMap` and there is no 'cleaner' way of changing it, I am find to leave the code as it is.

nnobelis (approved):

sschuberth (approved):

PR #5994, opened 2022-10-26
EvaluatorCommand: Fix-up replacing curations

fviernau:
The command allows overriding all curations by the ones provided via the option `--package-curations-file` or `--package-curations-dir`. Make that override work no matter if the curations are empty, to allow for "clearing" the curations.

### DIFF
--- a/cli/src/main/kotlin/commands/EvaluatorCommand.kt
+++ b/cli/src/main/kotlin/commands/EvaluatorCommand.kt
@@ -267,8 +267,8 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
             ortResultInput = ortResultInput.replaceConfig(config)
         }
 
-        val curations = FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations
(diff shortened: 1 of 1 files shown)


cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -267,8 +267,8 @@ class EvaluatorCommand : CliktCommand(name = "evaluate", help = "Evaluate ORT re
             ortResultInput = ortResultInput.replaceConfig(config)
         }
 
-        val curations = FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations
-        if (curations.isNotEmpty()) {
+        if (packageCurationsFile != null || packageCurationsDir != null) {
```
sschuberth, line 270: I'd propose to use `packageConfigurationOption != null` instead as it's shorter and nicely emphasizes that you want to do something as soon as the option is specified.

sschuberth (changes requested):

sschuberth (approved):

PR #5995, opened 2022-10-26
CreateCommand: Allow for specifying non-offending license IDs

fviernau:
The command already allows specifying non-offending license categories. Allow specifying license IDs in addition to support further use cases. This unlocks new use cases as the user can now specify uncategorized licenses or the SPDX constant NOASSERTION.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
@@ -99,9 +99,15 @@ internal class CreateCommand : CliktCommand(
 
     private val nonOffendingLicenseCategories by option(
         "--non-offending-license-categories",
-        help = "Configure the path exclude generation to not create excludes for files or directories which only " +
-                "contain non-offending licenses specified by this comma separated list of license category names. " +
-                "Each category name must be present in the given license classifications file."
+        help = "Specify licenses by their category which should be considered non-offending. Path excludes are not" +
+                "generated for files or directories which only contain non-offending licenses. Each category name " +
+                "must be present in the given license classifications file."
+    ).split(",").default(emptyList())
+
+    private val nonOffendingLicenseIds by option(
+        "--non-offending-license-ids",
+        help = "Specify license IDs which should be considered non-offending. Path excludes are not generated for " +
+                "files or directories which only contain non-offending licenses."
     ).split(",").default(emptyList())
 
     override fun run() {
@@ -174,8 +180,15 @@ internal class CreateCommand : CliktCommand(
         }
 
     private fun getNonOffendingLicenses(): Set<SpdxSingleLicenseExpression> {
+        val result = mutableSetOf<SpdxSingleLicenseExpression>()
+
         // Filter blanks to allow passing an empty list as argument to simplify caller code.
-        if (nonOffendingLicenseCategories.all { it.isBlank() }) return emptySet()
+        nonOffendingLicenseIds.filterNot { it.isBlank() }.mapTo(result) {
+            SpdxSingleLicenseExpression.parse(it)
+        }
+
+        // Filter blanks to allow passing an empty list as argument to simplify caller code.
+        if (nonOffendingLicenseCategories.all { it.isBlank() }) return result
 
         val licenseClassifications = licenseClassificationsFile?.readValue<LicenseClassifications>()
             ?: throw UsageError(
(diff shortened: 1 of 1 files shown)


helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
```diff
@@ -100,8 +100,17 @@ internal class CreateCommand : CliktCommand(
     private val nonOffendingLicenseCategories by option(
         "--non-offending-license-categories",
         help = "Configure the path exclude generation to not create excludes for files or directories which only " +
-                "contain non-offending licenses specified by this comma separated list of license category names. " +
-                "Each category name must be present in the given license classifications file."
+                "contain non-offending licenses specified by this comma separated list of license category names and " +
+                "by the option '--non-offending-license-ids'. Each category name must be present in the given " +
+                "license classifications file."
```
mnonnenmacher, line 105: Maybe just say "cumulative with '--non-offending-license-ids'" at the end.

helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
```diff
@@ -100,8 +100,17 @@ internal class CreateCommand : CliktCommand(
...
+    ).split(",").default(emptyList())
+
+    private val nonOffendingLicenseIds by option(
+        "--non-offending-license-ids",
+        help = "Configure the path exclude generation to not create excludes for files or directories which only " +
+                "contain non-offending licenses specified by this comma separated list of license IDs and by " +
+                "the option '--non-offending-license-categories'. Each category name must be present in the given " +
+                "license classifications file."
```
mnonnenmacher, line 113: The last sentence does not fit here, and maybe just say "cumulative with '--non-offending-license-categories'".

mnonnenmacher (changes requested):

mnonnenmacher (dismissed):

helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
```diff
@@ -174,8 +182,15 @@ internal class CreateCommand : CliktCommand(
...
+        val result = mutableSetOf<SpdxSingleLicenseExpression>()
+
         // Filter blanks to allow passing an empty list as argument to simplify caller code.
-        if (nonOffendingLicenseCategories.all { it.isBlank() }) return emptySet()
+        nonOffendingLicenseIds.filterNot { it.isBlank() }.mapTo(result) {
+            SpdxSingleLicenseExpression.parse(it)
+        }
+
+        // Filter blanks to allow passing an empty list as argument to simplify caller code.
+        if (nonOffendingLicenseCategories.all { it.isBlank() }) return result
```
sschuberth, line 193: I know the code was like that before, but strictly speaking this code does not *filter*, but return early if *all* `nonOffendingLicenseCategories` are blank. Maybe this is a good opportunity to fix this? Isn't `nonOffendingLicenseCategories.isEmpty()` sufficient to account for `--non-offending-license-categories=` being specified (which I believe to be the intention)?

sschuberth (dismissed):

helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
fviernau: It allows for specifying `--non-offending-license-categories ""`

helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
```diff
@@ -174,8 +183,15 @@ internal class CreateCommand : CliktCommand(
         }
 
     private fun getNonOffendingLicenses(): Set<SpdxSingleLicenseExpression> {
+        val result = mutableSetOf<SpdxSingleLicenseExpression>()
```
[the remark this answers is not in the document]
fviernau: To me `result` indicates that this will be the return value of the function. I believe it's quite clear and also widely used in ORT. I'd prefer to keep it.

helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
sschuberth: For reference, I just checked, and *clikt* returns a list with a single element which is an empty string in this case.

fviernau: Yes, this was the reason the code was invented.

helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
```diff
@@ -99,9 +99,15 @@ internal class CreateCommand : CliktCommand(
 
     private val nonOffendingLicenseCategories by option(
         "--non-offending-license-categories",
-        help = "Configure the path exclude generation to not create excludes for files or directories which only " +
-                "contain non-offending licenses specified by this comma separated list of license category names. " +
-                "Each category name must be present in the given license classifications file."
+        help = "Specify license IDs by their category which should be considered non-offending. Path excludes are not" +
```
sschuberth, line 102: I believe "IDs" should be dropped here as that's not what's being specified, so it should read "Specify licenses by their category ..."

sschuberth (approved):

PR #6001, opened 2022-10-26
Improvements to Copyright handling

sschuberth:

### DIFF
--- a/model/src/main/kotlin/licenses/ResolvedLicense.kt
+++ b/model/src/main/kotlin/licenses/ResolvedLicense.kt
@@ -121,19 +121,20 @@ data class ResolvedLicense(
 }
 
 private fun Collection<ResolvedCopyrightFinding>.toResolvedCopyrights(process: Boolean): List<ResolvedCopyright> {
-    val allStatements = map { it.statement }
-    val processedStatements = if (process) {
-        CopyrightStatementsProcessor().process(allStatements).toMap()
-    } else {
-        allStatements.associateBy({ it }, { setOf(it) })
+    if (!process) return map { ResolvedCopyright(it.statement, setOf(it)) }
+
+    val statementToFinding = associateBy { it.statement }
+    val result = CopyrightStatementsProcessor().process(statementToFinding.keys)
+
+    val processedCopyrights = result.processedStatements.map { (statement, originalStatements) ->
+        val findings = originalStatements.mapTo(mutableSetOf()) { statementToFinding.getValue(statement) }
+        ResolvedCopyright(statement, findings)
     }
 
-    return processedStatements.mapValues { (_, originalStatements) ->
-        filter { it.statement in originalStatements }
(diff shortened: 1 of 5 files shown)


model/src/main/kotlin/licenses/ResolvedLicense.kt
```diff
@@ -121,19 +121,20 @@ data class ResolvedLicense(
...
-    val processedStatements = if (process) {
-        CopyrightStatementsProcessor().process(allStatements).toMap()
-    } else {
-        allStatements.associateBy({ it }, { setOf(it) })
+    if (!process) return map { ResolvedCopyright(it.statement, setOf(it)) }
+
+    val statementToFinding = associateBy { it.statement }
+    val (processedStatements, unprocessedStatements) = CopyrightStatementsProcessor().process(statementToFinding.keys)
+
+    val processedCopyrights = processedStatements.mapNotNull { (statement, originals) ->
```
fviernau, line 129: Was you mention optimizations, was this function too slow?

sschuberth: > was this function too slow?

I did not measure it. It just occurred to me from the code that for some cases too much work was being done, like `in`-checks in a double-nested loop.

model/src/main/kotlin/licenses/ResolvedLicense.kt
```diff
@@ -121,19 +121,20 @@ data class ResolvedLicense(
...
 private fun Collection<ResolvedCopyrightFinding>.toResolvedCopyrights(process: Boolean): List<ResolvedCopyright> {
-    val allStatements = map { it.statement }
-    val processedStatements = if (process) {
-        CopyrightStatementsProcessor().process(allStatements).toMap()
-    } else {
-        allStatements.associateBy({ it }, { setOf(it) })
+    if (!process) return map { ResolvedCopyright(it.statement, setOf(it)) }
+
+    val statementToFinding = associateBy { it.statement }
+    val (processedStatements, unprocessedStatements) = CopyrightStatementsProcessor().process(statementToFinding.keys)
```
mnonnenmacher: I'm not a fan of deconstructing data classes like this. Mainly because it makes assumptions about the property order which should be an implementation details of the class, but also because it has the risk that chosen names are misleading (I think refactoring the names of the data class would also not update them here).

model/src/main/kotlin/licenses/ResolvedLicense.kt
```diff
@@ -121,19 +121,20 @@ data class ResolvedLicense(
...
+
+    val processedCopyrights = processedStatements.map { (statement, originals) ->
```
mnonnenmacher, line 129: Consider `originalStatements` for clarity.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

sschuberth:

PR #6004, opened 2022-10-27
Generalize a function to collapse whitespace

sschuberth:

### DIFF
--- a/utils/common/src/main/kotlin/Extensions.kt
+++ b/utils/common/src/main/kotlin/Extensions.kt
@@ -310,6 +310,13 @@ fun <K, V : Set<T>, T> Map<K, V>.zipWithCollections(other: Map<K, V>): Map<K, V>
 fun Number.bytesToMib(): Double = toDouble() / (1024 * 1024)
 
 /**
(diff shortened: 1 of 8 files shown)


utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -309,6 +309,13 @@ fun <K, V : Set<T>, T> Map<K, V>.zipWithCollections(other: Map<K, V>): Map<K, V>
  */
 fun Number.bytesToMib(): Double = toDouble() / (1024 * 1024)
 
+/**
+ * Trim leading and trailing whitespace, and collapse consecutive inner whitespace to a single space.
+ */
+fun String.collapseWhitespace() = trim().replace(CONSECUTIVE_WHITESPACE_REGEX, " ")
```
oheger-bosch, line 315: Out of curiosity: What was the reason why you switched the order of the `trim` and `replace` operations?

oheger-bosch (approved):

utils/common/src/main/kotlin/Extensions.kt
sschuberth: I was thinking that trimming first potentially reduces the number of whitespaces collapsed later.

PR #6006, opened 2022-10-27
Npm: Cache the results of `npm view`

mnonnenmacher:
Cache the results of the `npm view` call to improve performance in scans with many projects that have many overlapping dependencies.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -113,6 +113,8 @@ open class Npm(
 
     private val graphBuilder = DependencyGraphBuilder(NpmDependencyHandler(this))
 
+    private val npmViewCache = ConcurrentHashMap<String, JsonNode>()
+
     /**
      * Search depth in the `node_modules` directory for `package.json` files used for collecting all packages of the
      * projects.
@@ -360,10 +362,11 @@ open class Npm(
         return Pair(id.toCoordinates(), module)
     }
 
-    protected open fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
-        val process = run(workingDir, "view", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)
-    }
+    protected open fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode =
+        npmViewCache.getOrPut(packageName) {
+            val process = run(workingDir, "view", "--json", packageName)
+            jsonMapper.readTree(process.stdout)
+        }
 
     /** Cache for submodules identified by its moduleDir absolutePath */
     private val submodulesCache: ConcurrentHashMap<String, List<File>> = ConcurrentHashMap()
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -113,6 +113,8 @@ open class Npm(
 
     private val graphBuilder = DependencyGraphBuilder(NpmDependencyHandler(this))
 
+    private val npmViewCache = mutableMapOf<String, JsonNode>()
```
oheger-bosch, line 116: IIRC, the metadata for the packages is retrieved in multiple threads in parallel. So, the cache needs to be a concurrent map then.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -360,10 +362,11 @@ open class Npm(
         return Pair(id.toCoordinates(), module)
     }
 
-    protected open fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
-        val process = run(workingDir, "view", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)
-    }
+    protected open fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode =
```
sschuberth, line 365: This function might be called in parallel via `parsePackage`, so should `npmViewCache` become a `ConcurrentHashMap` like `submodulesCache` below?

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -360,10 +362,11 @@ open class Npm(
...
+        npmViewCache.getOrPut(packageName) {
```
oheger-bosch, line 366: I am also not sure about the semantics of `getOrPut` regarding concurrent access. `ConcurrentMap` defines special atomic update functions. Maybe the Kotlin extension is aware of this, but I do not for sure.

sschuberth: > Maybe the Kotlin extension is aware of this, but I do not for sure.

Yes it is, there's a special `fun ConcurrentMap<K, V>.getOrPut()`.

mnonnenmacher: Yes, the `getOrPut` implementation for `ConcurrentHashMap` uses `putIfAbsent`.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -360,10 +362,11 @@ open class Npm(
...
+            val process = run(workingDir, "view", "--json", packageName)
```
oheger-bosch, line 367: Here a potentially long-running process is invoked with the map partially locked. This is probably not ideal.
I will approve this PR because I think it improves the current situation in any case; but there is still room for optimization.

oheger-bosch (approved):

mnonnenmacher:
The failing ClearlyDefined test is unrelated to the change.

Issue #5483, opened 2022-06-23
SpdxDocumentFile: documentLocation is not used if the dependency is local


bennati:
If the dependency is stored in a subfolder, the analyzer infers the VCS URL from the project instead of from the downloadLocation attribute of the spdx file.

This is caused by the VCS info being determined from the working tree first, and from the download location as fallback.
https://github.com/oss-review-toolkit/ort/blob/main/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt#L282

The current implementation makes it impossible to reference a package where the source code is located in another repository than the SPDX file. A scenario where this is no issue is if you put the SPDX file in a fork of the packages source code.

tsteenbe:
Below I wrote out current and desired handling of package.spdx.yml by ORT:

**Current**

@bennati  Did a scan of  github.com/movetk/movetk/ and [master/third_party/GsTL/package.spdx.yml](https://github.com/movetk/movetk/blob/master/third_party/GsTL/package.spdx.yml) currently shows up as a project.
Which is correct as packageFileName: "./" is set indicating that the sources for SPDXRef-Package-GsTL are in the same directory as package.spdx.yml

```
SPDXID: "SPDXRef-DOCUMENT"
spdxVersion: "SPDX-2.2"
creationInfo:
  created: "2020-09-07T13:30:00Z"
  creators:
  - "Organization: HERE Europe B.V."
  - "Person: Thomas Steenbergen"
  licenseListVersion: "3.9"
name: "GsTL-1.2.1"
dataLicense: "CC0-1.0"
documentNamespace: "http://spdx.org/spdxdocs/spdx-document-GsTL"
documentDescribes:
- "SPDXRef-Package-GsTL"
packages:
- SPDXID: "SPDXRef-Package-GsTL"
  copyrightText: "NOASSERTION"
  description: "GsTL is a C++ library that provides a comprehensive set of tools and algorithms for geostatistics."
  downloadLocation: "http://prdownloads.sourceforge.net/gstl/GsTL-1.2.1.tgz?download"
  filesAnalyzed: false
  homepage: "http://gstl.sourceforge.net/"
  licenseConcluded: "NOASSERTION"
  licenseDeclared: "BSD-4-Clause"
  name: "GsTL"
  sourceInfo: ""
  versionInfo: "1.2.1"
  packageFileName: "./"
```

**Desired**

[...]

PR #6007, opened 2022-10-27
SpdxDocumentFile: Prefer VCS information from SPDX over working tree

mnonnenmacher:
Prefer the VCS information provided by the SPDX metadata of a package over the VCS information gathered from the local working tree. Otherwise it is not possible to define in SPDX that the source code of a package is located in a different VCS because the local working tree will always be preferred.

### DIFF
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -296,10 +296,10 @@ class SpdxDocumentFile(
     private fun SpdxPackage.toPackage(definitionFile: File?, doc: SpdxResolvedDocument): Package {
         val packageDescription = description.takeUnless { it.isEmpty() } ?: summary
 
-        // If the VCS information cannot be determined from the VCS working tree itself, fall back to try getting it
-        // from the download location.
+        // If the VCS information cannot be parsed from the download location, fall back to try getting it from the VCS
+        // working tree.
         val packageDir = definitionFile?.resolveSibling(packageFilename)
-        val vcs = packageDir?.let { VersionControlSystem.forDirectory(it)?.getInfo() } ?: getVcsInfo().orEmpty()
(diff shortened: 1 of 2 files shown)


mnonnenmacher:
> IIUC, this PR should mark #5483 as resolved. However, as it effectively seems to revert the changes from #4277, I'd like to kindly ask @mnonnenmacher to cross-check that this is indeed intended, as the latter actually resolved #4276.
> 
> All in all, these sources should be referred to from the commit message to document what has been considered.

Thanks for the links, I'm waiting for the results of a test scan before further looking into this. I think if `packageFilename` is set it should also be verified that the file/directory actually exists, so maybe I'll either make another commit or rework the existing one.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -296,10 +296,10 @@ class SpdxDocumentFile(
     private fun SpdxPackage.toPackage(definitionFile: File?, doc: SpdxResolvedDocument): Package {
         val packageDescription = description.takeUnless { it.isEmpty() } ?: summary
 
-        // If the VCS information cannot be determined from the VCS working tree itself, fall back to try getting it
-        // from the download location.
+        // If the VCS information cannot be parsed from the download location, fall back to try getting it from the VCS
+        // working tree.
         val packageDir = definitionFile?.resolveSibling(packageFilename)
-        val vcs = packageDir?.let { VersionControlSystem.forDirectory(it)?.getInfo() } ?: getVcsInfo().orEmpty()
+        val vcs = getVcsInfo() ?: packageDir?.let { VersionControlSystem.forDirectory(it)?.getInfo() }.orEmpty()
```
fviernau, line 302: Does this change apply only to VCS info of `Packages` (not `Projects`) ?

mnonnenmacher: Yes.

mnonnenmacher:
After reading the related documentation and issues I came to the conclusion that this needs to be fixed differently. I will make another PR tomorrow.

PR #6010, opened 2022-10-29
analyzer: Upgrade Gradle test projects to Gradle 7.5.1

mnonnenmacher:
Upgrade the Gradle test projects to Gradle 7.5.1 and update the configuration where required.


sschuberth:
Don't we want to deliberately keep them at old versions to ensure we can still analyze such projects?

mnonnenmacher:
> Don't we want to deliberately keep them at old versions to ensure we can still analyze such projects?

IMO no, we have a dedicated test for compatibility even though it is currently deactivated. And if we do compatibility tests it should happen with a strategy, not by testing arbitrary versions as part of tests which were created to test other functionality.

analyzer/src/funTest/assets/projects/synthetic/gradle-all-dependencies-expected-result-with-curations.yml
```diff
@@ -47,7 +47,7 @@ analyzer:
     variables: {}
     tool_versions: {}
   config:
-    allow_dynamic_versions: false
+    allow_dynamic_versions: true
```
sschuberth, line 50: Any idea why this is `true` now? The default is `false`, and the only `OrtMainFunTest` which uses this file does not seem to modify the default...

sschuberth: Maybe come local configuration of your is being taken into account, as the test runs the full `OrtMain`?

mnonnenmacher:
@sschuberth FYI, I have started to work on activating the Gradle compatibility tests again by updating and reducing the size of the test.

analyzer/src/funTest/assets/projects/synthetic/gradle-all-dependencies-expected-result-with-curations.yml
mnonnenmacher: Indeed, fixed. This should not be the case, I think I'll make a separate PR to decouple the test from the user configuration.

sschuberth (approved):

PR #6014, opened 2022-10-30
Continued `CopyrightStatementsProcessor` improvements

sschuberth:

### DIFF
--- a/utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
+++ b/utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
@@ -26,32 +26,6 @@ import java.util.SortedSet
 
 import org.ossreviewtoolkit.utils.common.collapseWhitespace
 
-private data class Parts(
-    val prefix: String,
-    val years: Set<Int>,
-    val owner: String,
-    val originalStatements: List<String>
-) {
-    companion object {
-        val COMPARATOR = compareBy<Parts>({ it.owner }, { prettyPrintYears(it.years) }, { it.prefix })
-    }
-
-    override fun toString() =
-        buildString {
-            append(prefix)
-
-            if (years.isNotEmpty()) {
-                append(" ")
-                append(prettyPrintYears(years))
-            }
-
-            if (owner.isNotEmpty()) {
-                append(" ")
-                append(owner)
-            }
-        }
-}
-
 private val INVALID_OWNER_START_CHARS = charArrayOf(' ', ';', '.', ',', '-', '+', '~', '&')
 private val INVALID_OWNER_KEY_CHARS = charArrayOf('<', '>', '(', ')', '[', ']') + INVALID_OWNER_START_CHARS
 
@@ -112,49 +86,6 @@ private fun prettyPrintYears(years: Collection<Int>): String {
     return getYearRanges(years).joinToString { (fromYear, toYear) ->
         if (fromYear == toYear) fromYear.toString() else "$fromYear-$toYear"
     }
-}
-
-/**
- * Split the [copyrightStatement] into its [Parts], or return null if the [Parts] could not be determined.
- */
-private fun determineParts(copyrightStatement: String): Parts? {
-    /**
-     * Strip the longest [known copyright prefix][KNOWN_PREFIX_REGEX] from [copyrightStatement] and return a pair of the
-     * copyright statement without the prefix and the prefix that was stripped from it.
-     */
-    fun stripKnownCopyrightPrefix(copyrightStatement: String): Pair<String, String> {
-        val copyrightStatementWithoutPrefix = KNOWN_PREFIX_REGEX.map { regex ->
-            copyrightStatement.replace(regex, "")
-        }.minByOrNull {
-            it.length
-        } ?: return Pair(first = copyrightStatement, second = "")
-
-        return Pair(
-            first = copyrightStatementWithoutPrefix,
-            second = copyrightStatement.removeSuffix(copyrightStatementWithoutPrefix)
-        )
-    }
-
-    /**
-     * Remove all years from the [copyrightStatement] and return the stripped string paired to the set of years.
-     */
-    fun stripYears(copyrightStatement: String): Pair<String, Set<Int>> =
-        replaceYears(copyrightStatement).let {
-            it.copy(first = it.first.replace(YEAR_PLACEHOLDER, ""))
-        }
-
-    val prefixStripResult = stripKnownCopyrightPrefix(copyrightStatement)
-    if (prefixStripResult.second.isEmpty()) return null
-
-    val yearsStripResult = stripYears(prefixStripResult.first)
-    return Parts(
-        prefix = prefixStripResult.second,
-        years = yearsStripResult.second,
-        owner = yearsStripResult.first
-            .trimStart(*INVALID_OWNER_START_CHARS)
-            .collapseWhitespace(),
-        originalStatements = listOf(copyrightStatement)
-    )
 }
 
 /**
@@ -256,6 +187,32 @@ private fun replaceYears(copyrightStatement: String): Pair<String, Set<Int>> {
  * TODO: Maybe treat URLs similar to years, e.g. entries which differ only in URLs and years can be merged.
  */
 object CopyrightStatementsProcessor {
+    data class Parts(
+        val prefix: String,
+        val years: Set<Int>,
+        val owner: String,
+        val originalStatements: List<String>
+    ) {
+        companion object {
+            val COMPARATOR = compareBy<Parts>({ it.owner }, { prettyPrintYears(it.years) }, { it.prefix })
+        }
+
+        override fun toString() =
+            buildString {
+                append(prefix)
+
+                if (years.isNotEmpty()) {
+                    append(" ")
+                    append(prettyPrintYears(years))
+                }
+
+                if (owner.isNotEmpty()) {
+                    append(" ")
+                    append(owner)
+                }
+            }
+    }
+
     data class Result(
         /**
          * The copyright statements that were processed by the [CopyrightStatementsProcessor], mapped to the original
@@ -271,6 +228,49 @@ object CopyrightStatementsProcessor {
     ) {
         @get:JsonIgnore
         val allStatements by lazy { unprocessedStatements + processedStatements.keys }
+    }
+
+    /**
+     * Split the [copyrightStatement] into its [Parts], or return null if the [Parts] could not be determined.
+     */
+    fun determineParts(copyrightStatement: String): Parts? {
+        /**
+         * Strip the longest [known copyright prefix][KNOWN_PREFIX_REGEX] from [copyrightStatement] and return a pair of
+         * the copyright statement without the prefix and the prefix that was stripped from it.
+         */
+        fun stripKnownCopyrightPrefix(copyrightStatement: String): Pair<String, String> {
+            val copyrightStatementWithoutPrefix = KNOWN_PREFIX_REGEX.map { regex ->
+                copyrightStatement.replace(regex, "")
+            }.minByOrNull {
+                it.length
+            } ?: return Pair(first = copyrightStatement, second = "")
+
+            return Pair(
+                first = copyrightStatementWithoutPrefix,
+                second = copyrightStatement.removeSuffix(copyrightStatementWithoutPrefix)
+            )
+        }
+
+        /**
+         * Remove all years from the [copyrightStatement] and return the stripped string paired to the set of years.
+         */
+        fun stripYears(copyrightStatement: String): Pair<String, Set<Int>> =
+            replaceYears(copyrightStatement).let {
+                it.copy(first = it.first.replace(YEAR_PLACEHOLDER, ""))
+            }
+
+        val prefixStripResult = stripKnownCopyrightPrefix(copyrightStatement)
+        if (prefixStripResult.second.isEmpty()) return null
+
+        val yearsStripResult = stripYears(prefixStripResult.first)
+        return Parts(
+            prefix = prefixStripResult.second,
+            years = yearsStripResult.second,
+            owner = yearsStripResult.first
+                .trimStart(*INVALID_OWNER_START_CHARS)
+                .collapseWhitespace(),
+            originalStatements = listOf(copyrightStatement)
+        )
     }
 
     /**
--- a/utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
+++ b/utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
@@ -31,7 +31,11 @@ private data class Parts(
     val years: Set<Int>,
     val owner: String,
     val originalStatements: List<String>
-)
+) {
+    companion object {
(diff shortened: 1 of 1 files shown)


utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
```diff
@@ -288,6 +261,33 @@ object CopyrightStatementsProcessor {
      * collapsed. The returned [Result] contains successfully processed as well as unprocessed statements.
      */
     fun process(copyrightStatements: Collection<String>): Result {
+        /**
```
fviernau, line 266: Personally I'm not a fan of using this nesting heavily, IMO it decreases readability as
levels of abstraction become mixed again.

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
```diff
@@ -239,6 +187,17 @@ private fun replaceYears(copyrightStatement: String): Pair<String, Set<Int>> {
  * TODO: Maybe treat URLs similar to years, e.g. entries which differ only in URLs and years can be merged.
  */
 object CopyrightStatementsProcessor {
+    data class Parts(
```
fviernau, line 190: PR: Is there any bigger umbrella this PR (series) is made for. Would you mind sharing the ultimate goal, e.g. by linking the PRs to a GitHub issue describing the goal?

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
```diff
@@ -239,6 +187,17 @@ private fun replaceYears(copyrightStatement: String): Pair<String, Set<Int>> {
...
+        val prefix: String,
+        val years: Set<Int>,
+        val owner: String,
+        val originalStatements: List<String>
+    ) {
```
fviernau, line 195: Commit: Mind explaining the use case a bit, in particular: why can't the custom report use the process function, but needs the parts class?

fviernau (changes requested):

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
sschuberth: These were just a series of ad-hoc PRs that I created in the context of getting an overview for / understanding the algorithms in `CopyrightStatementsProcessor`, and addressing miscellaneous things I came across while doing so. Eventually, the only "goal" I had was to make the `determineParts()` logic public, so this PR concludes that work. I do not plan to make further `CopyrightStatementsProcessor`-related PRs for the time being.

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
sschuberth: > why can't the custom report use the process function, but needs the parts class?

Because the `owner` information is lost / mangled into the statement again, which hinders grouping by owner easily.

fviernau: Would it make sense / be possible to adjust the data structure of the result of `processCopyrights` so that it works with your use case (instead of making `determineParts` public ?

sschuberth: I guess I'll simply make the keys in `processedStatements` the `Parts`...

fviernau: Haven't investigated it yet,...was really more like a question

fviernau: I've taken a look at it again. The parts in the results are already merged, while exposing the function as you initially did allowed you to use determining parts independently of any merging applied. So, maybe exposing determine parts actually is simpler / more flexible.

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
```diff
@@ -35,6 +35,21 @@ private data class Parts(
...
+            append(prefix)
+
+            if (years.isNotEmpty()) {
+                append(" ")
+                append(prettyPrintYears(years))
+            }
+
+            if (owner.isNotEmpty()) {
+                append(" ")
+                append(owner)
```
fviernau: Previously, copyright statements without an owner have been skipped. With this change this is no more the case: it now produces copyright statement without an owner. I prefer skipping statements with empty owners as it was before.

fviernau (changes requested):

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
```diff
@@ -256,6 +187,32 @@ private fun replaceYears(copyrightStatement: String): Pair<String, Set<Int>> {
...
  */
 object CopyrightStatementsProcessor {
+    data class Parts(
+        val prefix: String,
+        val years: Set<Int>,
+        val owner: String,
+        val originalStatements: List<String>
+    ) {
+        companion object {
+            val COMPARATOR = compareBy<Parts>({ it.owner }, { prettyPrintYears(it.years) }, { it.prefix })
```
fviernau, line 197: Should this comparator now become private, as it is an internal of `CopyrightStatementsProcessor` ?

fviernau (changes requested):

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
sschuberth: > I prefer skipping statements with empty owners as it was before.

It's still like that as the caller still has `if (it.owner.isNotEmpty()) {` like before. The `if (owner.isNotEmpty()) {` introduced here is on top of that. I'll address this temporary "duplication" in a followup PR.

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
sschuberth: If you don't mind, I'd prefer to do also that in a later PR as I'm planning to implement `Comparable` here.

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
sschuberth: My plan is actually to remove the `if (it.owner.isNotEmpty()) {` from the caller and to handle empty owners as unprocessed. I.e. there should be no statement that is neither processed nor unprocessed. And this logic change I'd prefer to keep out of the "factor-out the to-string"- topic.

sschuberth: If you're curious, [here's a draft](https://github.com/oss-review-toolkit/ort/pull/6022) that is based on this PR.

fviernau: After looking at this draft, I again ask (as more PRs are necessary unexpectedly) what the goal of this refactoring is, so that I can consider it in the review.

sschuberth: I can only repeat myself from [here](https://github.com/oss-review-toolkit/ort/pull/6014#discussion_r1009234410). There still is no ultimate goal or hidden agenda. It's more like an unplanned "code audit" that was partly triggered by your review comments actually. And while playing around with what you suggested (like changing the return value of `process` instead of making `determineParts` public), I discovered issues like statements without owners being swallowed, and the returned owner being dependent on the order of the original statements. I fixed these issues now as I discovered them, but I have no further concrete plans for this class (unless more issues pop up).

fviernau: Thanks! 

> I can only repeat myself from 

That comment says "this PR concludes that work", so the situation has changed which is why I think asking the question again is valid.

> I discovered issues like statements without owners being swallowed

To me it is not obvious whether this is an issue. Let's just do a quick poll on this from the core maintainers.

sschuberth: > That comment says "this PR concludes that work", so the situation has changed which is why I think asking the question again is valid.

Sure, asking does not hurt. But I also wrote "for the time being" there, as I of course couldn't rule out more work coming up due to the reviews.

> To me it is not obvious whether this is an issue. Let's just do a quick poll on this from the core maintainers.

Fine with me. But can we proceed with *this* PR then?

fviernau: > Fine with me. But can we proceed with _this_ PR then?

Not sure if I'm missing something but doesn't this PR assume that swallowing will be removed? e.g. the code change in this PR would be different otherwise?

sschuberth: > e.g. the code change in this PR would be different otherwise?

No, it wouldn't. I wrote this code before I had the idea to record empty owners as unprocessed statements.

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
```diff
@@ -26,32 +26,6 @@ import java.util.SortedSet
 
 import org.ossreviewtoolkit.utils.common.collapseWhitespace
```
fviernau, line 27: It would be really interesting to know this use case. Would you mind providing an example how this use case requires the copyrights to be processed / merged?

sschuberth: @porsche-rbieniek can provide you the details about the case for using `determineParts` in report generation.

But besides that, to all @oss-review-toolkit/kotlin-devs can we now *please* proceed and stop blocking this unless there are good reasons beyond questions about the scope or those asked just out of curiosity?

fviernau: I would like to understand the use case before approving it, if it is possible to share it.

rbieniek: The Porsche use case is that we need to 'group' copyright statements by the copyright holder. 

We need to understand two facts here:
- list of copyright holders per package
- list of copyright statements issued by a particular copyright holder per package

To give an example copyright statement list here:
- "(c) Zeno Rocha",
- "Copyright (C) The Bootstrap Authors",
- "Copyright (c) 2011-2021 The Bootstrap Authors",
- "Copyright (c) 2011-2021 Twitter, Inc.",
- "Copyright (c) 2021 Bryan Braun",
- "Copyright 2011 $ year pkg.author",
- "Copyright 2011-2021 The Bootstrap Authors",
- "Copyright 2011-2021 The Bootstrap Authors (https://github.com/twbs/bootstrap/graphs/contributors)",
- "Copyright 2011-2021 Twitter, Inc.",
- "copyright 2011-2021 the Bootstrap Authors (https://github.com/twbs/bootstrap/graphs/contributors) and Twitter, Inc. (https://twitter.com)"

I would need to group as this:
- "The Bootstrap Authors": { "Copyright (C) The Bootstrap Authors", "Copyright (c) 2011-2021 The Bootstrap Authors", "Copyright 2011-2021 The Bootstrap Authors (https://github.com/twbs/bootstrap/graphs/contributors)", "copyright 2011-2021 the Bootstrap Authors (https://github.com/twbs/bootstrap/graphs/contributors) and Twitter, Inc. (https://twitter.com)" }
- "Twitter, Inc.": { "Copyright (c) 2011-2021 Twitter, Inc.", "copyright 2011-2021 the Bootstrap Authors (https://github.com/twbs/bootstrap/graphs/contributors) and Twitter, Inc. (https://twitter.com)" }
- "Zeno Rocha": { "(c) Zeno Rocha" }
- "Bryan Braun": { "Copyright (c) 2021 Bryan Braun" }
- "$ year pkg.author": {"Copyright 2011 $ year pkg.author" }

fviernau: Thanks @rbieniek !

Can you explain how the logic is supposed to arrive at these groups? 
In particular, do you think you will have a heuristic which produces the output you want solely based on the given statements above, or would there be additional (manual human) input needed?

sschuberth: > do you think you will have a heuristic which produces the output you want solely based on the given statements above

I honestly don't see how the outcome of this conversation is in anyway helpful, let alone required, for being able to review the code changes in this PR. But @rbieniek / @porsche-rbieniek can you please yet respond timely to move this forward?

rbieniek: Speaking from the perspective of Porsche. it would be helpful to have some kind heuristic, either built-in or implemented seperately, for grouping the copyright statements by the copyright holder for easing the life of the (human) analysts who need to review the output of ORT anyway. And who need to take appropiate actions based on that outcome.

Speaking from my personal point of view, as an user of ORT I simply asked for some capability already built into ORT to be exposed in a way that I can make use of it in custom code I implement for supporting internal workflows. 
I do not fully understand why and how the discussion about relatively small code change could have gotten so much out of the way into a somewhat fundamental discussion.

fviernau: You ask me to review a series of refactoring changes, without providing the use case / context.
Now I've just asked for the information which I would like to consider when thinking about the proposed approach.
I have really have no clue what problem you have with that.

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
```diff
@@ -31,7 +31,11 @@ private data class Parts(
     val years: Set<Int>,
     val owner: String,
     val originalStatements: List<String>
-)
+) {
```
fviernau: In my view the comparator belongs to `CopyrightStatementsProcesser` because it defines how `CopyrightStatementsProcesser` orders its results. So, moving it to companion object to `CopyrightStatementsProcesser` clarifies that ownership IMO better and also enables sharing. E.g. one can just change the ordering `CopyrightStatementsProcesser` without worrying about (custom) calling code.

What do you think?

fviernau (changes requested):

fviernau (approved):

utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
sschuberth: > In my view the comparator belongs to `CopyrightStatementsProcesser` because it defines how `CopyrightStatementsProcesser` orders its results.

I can see the point that there is some subjectivity to *how* `Parts` should be sorted, and that the current ordering might be seen as `CopyrightStatementsProcesser`-specific. However, in general I believe a data class knows best for itself how it should be sorted. At least it should have a default, expressed as `Comparable`. Otherwise such data classes could not be put into sorted sets or be used as keys in sorted maps, which is functionality I rely on in the later changes.

> It'll become harder to figure out which callers get affected when changing the ordering.

True, but I guess that's a rather theoretical argument as the ordering has been the same for years, and I don't foresee any changes to it.

> I'm approving this anyway, as you seem to be on a rush but would prefer if it was changed.

I'm not on a "rush", but I simply believe that discussing 4 days with > 30 comments over a small change that basically just moves existing code is enough. So, thanks for approving finally.

PR #6015, opened 2022-10-31
`ServiceLoader`-related improvements for the CLI commands

sschuberth:

### DIFF
--- a/advisor/src/main/kotlin/Advisor.kt
+++ b/advisor/src/main/kotlin/Advisor.kt
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.advisor
 
 import java.time.Instant
 import java.util.ServiceLoader
+import java.util.SortedMap
 
 import kotlinx.coroutines.async
 import kotlinx.coroutines.runBlocking
@@ -47,10 +48,10 @@ class Advisor(
         private val LOADER = ServiceLoader.load(AdviceProviderFactory::class.java)
 
         /**
-         * The set of all available [advice provider factories][AdviceProviderFactory] in the classpath, sorted by name.
+         * All [advice provider factories][AdviceProviderFactory] available in the classpath, associated by their names.
          */
-        val ALL: Set<AdviceProviderFactory> by lazy {
-            LOADER.iterator().asSequence().toSortedSet(compareBy { it.providerName })
+        val ALL: SortedMap<String, AdviceProviderFactory> by lazy {
+            LOADER.iterator().asSequence().associateByTo(sortedMapOf(String.CASE_INSENSITIVE_ORDER)) { it.providerName }
         }
     }
 
--- a/advisor/src/main/kotlin/Advisor.kt
+++ b/advisor/src/main/kotlin/Advisor.kt
@@ -44,7 +44,7 @@ class Advisor(
     private val config: AdvisorConfiguration
 ) {
     companion object : Logging {
-        private val LOADER = ServiceLoader.load(AdviceProviderFactory::class.java)!!
+        private val LOADER = ServiceLoader.load(AdviceProviderFactory::class.java)
 
         /**
          * The set of all available [advice provider factories][AdviceProviderFactory] in the classpath, sorted by name.
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -65,7 +65,7 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
     @JvmOverloads
     fun findManagedFiles(
         absoluteProjectPath: File,
-        packageManagers: Set<PackageManagerFactory> = PackageManager.ALL,
+        packageManagers: Collection<PackageManagerFactory> = PackageManager.ALL.values,
         repositoryConfiguration: RepositoryConfiguration = RepositoryConfiguration()
     ): ManagedFileInfo {
         require(absoluteProjectPath.isAbsolute)
@@ -74,14 +74,16 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
             "Using the following configuration settings:\n${yamlMapper.writeValueAsString(repositoryConfiguration)}"
         }
 
+        val distinctPackageManagers = packageManagers.distinct()
+
         // Associate files by the package manager factory that manages them.
-        val factoryFiles = if (packageManagers.size == 1 && absoluteProjectPath.isFile) {
+        val factoryFiles = if (distinctPackageManagers.size == 1 && absoluteProjectPath.isFile) {
             // If only one package manager is activated and the project path is in fact a file, assume that the file is
             // a definition file for that package manager. This is useful to limit analysis to a single project e.g. for
             // debugging purposes.
-            mutableMapOf(packageManagers.first() to listOf(absoluteProjectPath))
+            mutableMapOf(distinctPackageManagers.first() to listOf(absoluteProjectPath))
         } else {
-            PackageManager.findManagedFiles(absoluteProjectPath, packageManagers).toMutableMap()
+            PackageManager.findManagedFiles(absoluteProjectPath, distinctPackageManagers).toMutableMap()
         }
 
         // Associate mapped files by the package manager that manages them.
@@ -99,7 +101,7 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
         }
 
         if (!hasOnlyManagedDirs) {
-            packageManagers.find { it is Unmanaged.Factory }
+            distinctPackageManagers.find { it is Unmanaged.Factory }
                 ?.create(absoluteProjectPath, config, repositoryConfiguration)
                 ?.run { managedFiles[this] = listOf(absoluteProjectPath) }
         }
--- a/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
+++ b/analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
@@ -361,7 +361,7 @@ class SpdxDocumentFile(
         val packageFile = definitionFile.parentFile.resolve(spdxPackage.packageFilename)
 
         if (packageFile.isFile) {
-            val managedFiles = findManagedFiles(packageFile.parentFile, ALL)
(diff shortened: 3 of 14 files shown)


cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -151,11 +151,11 @@ class AnalyzerCommand : CliktCommand(name = "analyze", help = "Determine depende
...
 
-    private val activatedPackageManagers by option(
+    private val enabledPackageManagers by option(
         "--package-managers", "-m",
-        help = "The comma-separated package managers to activate, any of ${allPackageManagersByName.keys}. Note that " +
-                "deactivation overrides activation. If set, package manager activation settings from configuration " +
-                "files are ignored."
+        help = "The comma-separated package managers to enable, any of ${allPackageManagersByName.keys}. Note that " +
+                "disabling overrides enabling. If set, package manager enabling settings from configuration files " +
+                "are ignored."
```
mnonnenmacher, line 155: While at it you could rephrase the sentence to be a more specific, for example: "If set, the 'enabledPackageManagers' property from configuration files is ignored." Same for the option below.

advisor/src/main/kotlin/Advisor.kt
```diff
@@ -47,10 +48,10 @@ class Advisor(
         private val LOADER = ServiceLoader.load(AdviceProviderFactory::class.java)
 
         /**
-         * The set of all available [advice provider factories][AdviceProviderFactory] in the classpath, sorted by name.
+         * All [advice provider factories][AdviceProviderFactory] available in the classpath, associated by their names.
          */
-        val ALL: Set<AdviceProviderFactory> by lazy {
-            LOADER.iterator().asSequence().toSortedSet(compareBy { it.providerName })
+        val ALL: SortedMap<String, AdviceProviderFactory> by lazy {
+            LOADER.iterator().asSequence().associateByTo(sortedMapOf(String.CASE_INSENSITIVE_ORDER)) { it.providerName }
```
fviernau, line 54: I believe we should fail if duplicate names are used.
E.g. add a require block.

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -74,14 +74,16 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
             "Using the following configuration settings:\n${yamlMapper.writeValueAsString(repositoryConfiguration)}"
         }
 
+        val distinctPackageManagers = packageManagers.distinct()
```
fviernau, line 77: I'd also prefer failing here, e.g. use require instead of distinct. (Same in several other places below)

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
```diff
@@ -361,7 +361,7 @@ class SpdxDocumentFile(
         val packageFile = definitionFile.parentFile.resolve(spdxPackage.packageFilename)
 
         if (packageFile.isFile) {
-            val managedFiles = findManagedFiles(packageFile.parentFile, ALL)
+            val managedFiles = findManagedFiles(packageFile.parentFile)
```
mnonnenmacher, line 364: IIRC the usage of `ALL` was deliberate here to make sure that an error is created if a package manager dependency was defined but the required package manager is disabled.

fviernau (changes requested):

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
sschuberth: I'm not sure what you mean... `ALL` is the default argument, and it always returns all package manager implementations from the classpath, no matter whether they were disabled at the CLI or not.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
sschuberth: I'm usually not a fan of hard-coding other option names from an option's help text as such references are hard to maintain. But I'll do it here nonetheless as the other options are being referred to in the deprecation messages already anyway.

advisor/src/main/kotlin/Advisor.kt
sschuberth: That, however, would be new feature; we didn't fail in the corresponding code from `AdvisorCommand` before either. But I agree that we should forbid different plugins of the same type and name on the classpath, and will consider creating another PR to address that.

analyzer/src/main/kotlin/Analyzer.kt
sschuberth: Here, I believe any duplicates do no harm, and the deduplication is basically just a performance optimization to not find definition files with the same package manager twice.

Why do you believe it should be an error if the user says "find files with Gradle, Gradle and NPM" instead of just saying "find files with Gradle and NPM"? I.e. what's the risk?

advisor/src/main/kotlin/Advisor.kt
fviernau: Well, before this change having different provider with the same name resulted in all providers being used. As of this change, providers of duplicate name get silently discarded. That's a behavioral change.

analyzer/src/main/kotlin/Analyzer.kt
fviernau: In my view tolerating duplicate values as command line option is fine. However, I believe that should be handled only in the `AnalyzerCommand`, not all over the place. 

In other places (assuming that we all agree on this change which puts the factories by name on the map) I believe the assumption should be made that pkg mgrs are always `disctinctBy` their name, to avoid the IDs in the ORT file becoming ambigeous.

fviernau: In my view tolerating duplicate values as command line option is fine. However, I believe that should be handled only in the `AnalyzerCommand`, not all over the place. 

In other places (assuming that we all agree on this change which puts the factories by name on the map) I believe the assumption should be made that pkg mgrs are always `disctinctBy` their name, to avoid the IDs in the ORT file becoming ambigeous.

fviernau (changes requested):

advisor/src/main/kotlin/Advisor.kt
sschuberth: > Well, before this change having different provider with the same name resulted in all providers being used.

How? In the former `allVulnerabilityProvidersByName` map, providers were also deduplicated by name.

analyzer/src/main/kotlin/Analyzer.kt
sschuberth: > However, I believe that should be handled only in the `AnalyzerCommand`, not all over the place.

More concretely, are you suggesting to keep the `packageManagers` parameter type at a `Set` instead of a `Collection`? I thought I'm doing you a favor here, as it's usually you who advocates for functions to take more generic collections.

advisor/src/main/kotlin/Advisor.kt
fviernau: You're right. I've missed that the de-duplicated did exist before.

analyzer/src/main/kotlin/Analyzer.kt
fviernau: > More concretely, are you suggesting to keep the `packageManagers` parameter type at a `Set` instead of a `Collection`?

No. My suggestion was to use a `require` block as I wrote above.

> I thought I'm doing you a favor here, as it's usually you who advocates for functions to take more generic collections.

If there is no need for a parameter type to be "more specific" I believe that it's better to stick with the more general type, to allow for calling the function more easily (without prior conversion). This is what I usually try to adhere to when choosing the parameter type. Not sure if this is what you're talking about.

sschuberth: > No. My suggestion was to use a `require` block as I wrote above.

That doesn't make sense to me then, sorry. I don't see a point in failing hard here if we either could

- enforce "intended" use by keeping the function parameter type at a `Set` instead of a `Collection`, or
- "recover" from "unintended" use by simply deduplicating the collection.

fviernau: > enforce "intended" use by keeping the function parameter type at a Set instead of a Collection, or

That doesn't make sense to me because making the type a `Set` makes the factories distinct by instance, while as per the introduce map they should be distinct by their name, sorry.

> "recover" from "unintended" use by simply deduplicating the collection.

I'm not against deduplicating at all. I doesn't make sense to me to do this in multiple places rather than just in the CLI, sorry again.

Anyhow, as [1] has been resolved I won't block this and resolve this discussion as well.

[1] https://github.com/oss-review-toolkit/ort/pull/6015#discussion_r1010177969

fviernau (approved):

analyzer/src/main/kotlin/Analyzer.kt
sschuberth: > I doesn't make sense to me to do this in multiple places rather than just in the CLI, sorry again.

Please keep in mind that the `findManagedFiles` functions are part of the public API which might be used programmatically e.g. in the ORT server context.

analyzer/src/main/kotlin/managers/SpdxDocumentFile.kt
mnonnenmacher: What I meant is it was deliberate to be explicit following the idea that:
* If you don't care about the value of an argument, rely on the default value.
* If you care about the value of an argument, be explicit and do not rely on the default value to never change.

mnonnenmacher: But I mainly wanted to mention it, I'll leave it up to you if you want to change it or not as it does not break anything.

sschuberth: I believe there's a third idea to default arguments in general, which is to not repeat arguments that barely need to be changed. So you *do* care about the argument, *and* you're fine with the default value. From that perspective I'd prefer to keep the change, esp. as I cannot envision us to ever change the default value for the `packageManagers` argument to anything else than all package managers.

mnonnenmacher (approved):

PR #6016, opened 2022-10-31
ResolvedLicense: Fix `toResolvedCopyrights()`

mnonnenmacher:
When searching the copyright findings for a statement the code was erroneously using the processed statement instead of the unprocessed statement. If the processed statement was different to the original statement this lead to an exception, because the map of original statements to their findings did not contain an entry with that key. Also, this error limited the list of findings in the `ResolvedCopyright` to those which exactly match the processed statement.

This is fixup for 3e9d11d.

### DIFF
--- a/model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
+++ b/model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
@@ -281,6 +281,17 @@ class LicenseInfoResolverTest : WordSpec({
                 "(c) 2011 Holder",
                 "(c) 2012 Holder"
             )
+
+            result should containCopyrightStatementsForLicenseExactly(
+                "Apache-2.0",
+                "(c) 2009-2010 Holder",
+                process = true
+            )
+            result should containCopyrightStatementsForLicenseExactly(
+                "MIT",
+                "(c) 2011-2012 Holder",
+                process = true
+            )
         }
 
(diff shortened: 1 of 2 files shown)


sschuberth (approved):

mnonnenmacher:
@sschuberth Please review again, I have added a test that would have prevented the issue.

MarcelBochtler (dismissed):

MarcelBochtler (approved):

model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
```diff
@@ -281,6 +281,17 @@ class LicenseInfoResolverTest : WordSpec({
                 "(c) 2011 Holder",
                 "(c) 2012 Holder"
             )
+
```
sschuberth, line 284: Nit: I guess the blank line is intended here to separate the `process = true` cases?

sschuberth (approved):

model/src/test/kotlin/licenses/LicenseInfoResolverTest.kt
mnonnenmacher: Exactly.

mnonnenmacher:
Merged without waiting for the analyzer tests because the change does not affect the analyzer. The failing `GitWorkingTreeFunTest` is also unrelated to the change.

PR #6022, opened 2022-11-01
Yet more `CopyrightStatementsProcessor` improvements

sschuberth:


fviernau:
Should this be moved back to "draft" state, as it introduces copyright statements without holder into e.g. the notices?

sschuberth:
I've split out some stuff to https://github.com/oss-review-toolkit/ort/pull/6061.

sschuberth:
> Should this be moved back to "draft" state, as it introduces copyright statements without holder into e.g. the notices?

@fviernau, I finally found the time to continue working on this. In the meantime, I had verified with @LeChasseur that, to be on the safe side, even Copyright "stubs" without a holder, like "Copyright (c) 2006", should be maintained to fully satisfy e.g. `BSD-3-Clause`'s "Redistributions in binary form must reproduce the above copyright notice" (whatever it reads) obligation.

And the second commit also proves to be the right thing in the context of the recent `compareTo()` vs. `equals()` alignment discussions, IMO.

fviernau:
> @fviernau, I finally found the time to continue working on this. In the meantime, I had verified with @LeChasseur that, to be on the safe side, even Copyright "stubs" without a holder, like "Copyright (c) 2006", should be maintained to fully satisfy e.g. `BSD-3-Clause`'s "Redistributions in binary form must reproduce the above copyright notice" (whatever it reads) obligation.
> 
> And the second commit also proves to be the right thing in the context of the recent `compareTo()` vs. `equals()` alignment discussions, IMO.

I fear I won't find the time to do a review prior to my X-mas holidays. Is this urgent?

sschuberth:
> I fear I won't find the time to do a review prior to my X-mas holidays. Is this urgent?

It's not super urgent. But it's also not a lot of code. If you agree with the basic idea, and trust the test coverage, then also anyone else could approve, if you're fine with that. I just wanted to give you a heads up.

mnonnenmacher:
Code looks good to me, but I wonder if the outcome of the change should be discussed in a community meeting first. I'd also like to discuss this internally at Bosch before the merge but that won't be possible before next year.

fviernau:
> Code looks good to me, but I wonder if the outcome of the change should be discussed in a community meeting first. I'd also like to discuss this internally at Bosch before the merge but that won't be possible before next year.

+1

fviernau:
Needs to be discussed in EPAM as well after the holidays.

pombredanne:
Is this something that could be fixed upstream in ScanCode instead?

sschuberth:
> Is this something that could be fixed upstream in ScanCode instead?

I don't think so. ScanCode already does report Copyright statements without holders (if that's how they're declared in the source), but our post-processing so far removed these. So it's something we need to tackle in ORT.

PR #6036, opened 2022-11-03
PythonInspector: Fix a deserialization issue

mnonnenmacher:
If a project defines no dependencies python-inspector returns an empty object instead of an empty array [1]. Create a custom deserializer as a workaround to not fail during deserialization.

[1]: https://github.com/nexB/python-inspector/issues/94

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
+++ b/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
@@ -26,7 +26,11 @@ import java.util.SortedSet
 
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.builtins.ListSerializer
 import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonArray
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
```diff
@@ -112,10 +116,19 @@ internal object PythonInspector : CommandLineTool, Logging {
     @Serializable
     internal data class Result(
         @SerialName("files") val projects: List<Project>,
-        @SerialName("resolved_dependencies_graph") val resolvedDependenciesGraph: List<ResolvedDependency>,
+        @SerialName("resolved_dependencies_graph")
+        @Serializable(with = DependencyGraphDeserializer::class)
```
sschuberth, line 120: Would it make sense to do the changes in this file in a separate commit that can be cleanly reverted once the issue is fixed, but the changes to `PipFunTest` are kept?

mnonnenmacher:
The root cause was fixed in the meantime, I will replace this with a python-inspector upgrade.

PR #6037, opened 2022-11-03
Inject dependencies on the ORT configuration via Koin

sschuberth:


sschuberth:
I do not plan to pursue this idea further currently.

PR #6045, opened 2022-11-07
Revert "analyzer: Stop using the unauthenticated Git protocol in VCS processed"

zhernovs:
This reverts commit 4c57907f5f3c95f229f824c8cd802ce0efdf660b as it prevents ORT to download repositories that doesn't support https.

Signed-off-by: Andriy Zhernovskyi <ext-andriy.zhernovskyi@here.com>

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/gradle-android-cyclic-expected-output-app.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gradle-android-cyclic-expected-output-app.yml
@@ -11479,7 +11479,7 @@ shared_packages:
     path: ""
   vcs_processed:
     type: "Git"
-    url: "https://android.googlesource.com/platform/tools/sherpa.git"
+    url: "git://android.googlesource.com/platform/tools/sherpa.git"
     revision: ""
     path: ""
 - id: "Maven:androidx.constraintlayout:constraintlayout-solver:2.0.1"
@@ -11511,7 +11511,7 @@ shared_packages:
     path: ""
   vcs_processed:
     type: "Git"
-    url: "https://android.googlesource.com/platform/tools/sherpa.git"
+    url: "git://android.googlesource.com/platform/tools/sherpa.git"
     revision: ""
     path: ""
 - id: "Maven:androidx.dynamicanimation:dynamicanimation:1.0.0"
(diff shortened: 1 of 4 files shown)


zhernovs:
There are few old gerrit hosts in our company. They are internal in our company. 
Also it seems that cloning dependencies from https://svn.apache.org/repos/asf/xml/commons is timing out too.

sschuberth:
> There are few old internal gerrit hosts in our company.

AFAIK Gerrit supports cloning via `https://` since its inception. Gerrit is using JGit (like we do) after all. The only point is that you need to configure authentication correctly, using an [HTTP password](https://gerrit-documentation.storage.googleapis.com/Documentation/2.9.3/access-control.html#capability_generateHttpPassword).

> Also it seems that cloning dependencies from https://svn.apache.org/repos/asf/xml/commons is timing out due to this change too.

How's cloning from a Subversion repo related to Git? No, https://svn.apache.org/repos/asf/xml/commons simply results in a 404 also in the browser.

All in all I'm :-1: on this change as the real issue seems to be a lack of proper Gerrit / authentication configuration, and falling back to unauthenticated cloning does not seem like a good idea to me.

What's your take on this @fviernau as it's your change being reverted here?

fviernau:
> What's your take on this @fviernau as it's your change being reverted here?

Prior to my change, the problem was two fold

1. The Clones of `git://` weren't successful
2. The unsuccessful attempts to clone took very long (to time out). Haven't measured it but I guess >> 30 seconds

So, reverting this won't just make curations necessary but will also slow down affected scans tremendously (until VCS URLs are curated). Whether I'm ok with reverting this depends on the reason. For the given reason (authentication for https not setup for gerrit, but could be done) I **prefer to not do the revert**.

sschuberth:
> 1. The Clones of `git://` weren't successful

To clarify, to my understanding this was the case when metadata still pointed to `git://`, whereas the host meanwhile disabled `git://` in favor of `https://`. If the host still offered `git://`, cloning would have been successful.

fviernau:
> To clarify, to my understanding this was the case when metadata still pointed to `git://`, whereas the host meanwhile disabled `git://` in favor of `https://`. If the host still offered `git://`, cloning would have been successful.

This exactly is the case I tried to outline.

PR #6046, opened 2022-11-07
feat(nuget)!: Split the (technology) namespace from the name

sschuberth:
Use up to two namespace nodes from the name that identify the company
and product / technology as defined at [1] as the namespace. This allows
to write curations with empty name and version that affect a whole product
/ technology, and to set e.g. a concluded license for all
"Microsoft.AspNetCore" packages.

[1]: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces

> **Note**
> This is a breaking change as IDs that refer to NuGet packages, like in package curations, probably need to be adjusted to accommodate for the proper namespace.


sschuberth:
See [this](https://github.com/oss-review-toolkit/ort/issues/5105#issuecomment-1305736004) for more context. Even if this would be a quite massive breaking change for existing NuGet curations, I believe it's the right thing to do, and afterwards writing (generic) NuGet curations would be much easier / possible. What do you thing @oss-review-toolkit/core-devs?

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -289,8 +293,12 @@ private fun parseAuthors(spec: PackageSpec?): SortedSet<String> =
 private fun resolveLocalSpec(definitionFile: File): File? =
     definitionFile.parentFile?.resolve(".nuspec")?.takeIf { it.isFile }
 
-private fun getIdentifier(name: String, version: String) =
-    Identifier(type = "NuGet", namespace = "", name = name, version = version)
+internal fun getIdentifier(name: String, version: String): Identifier {
+    val namespace = name.split('.', limit = 3).toMutableList()
```
sschuberth, line 297: Maybe the splitting here should be made more strict and also require PascalCase as described in the linked article.

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetInspector.kt
```diff
@@ -197,6 +197,13 @@ private fun List<NuGetInspector.PackageData>.toPackageReferences(): Set<PackageR
         )
     }
 
+internal fun getIdentifierWithNamespace(name: String, version: String): Identifier {
```
sschuberth, line 200: @mnonnenmacher am I correct to assume that `public` access to this would be required here in order for [`GenerateAspNetCoreCurationsTask` et al](https://github.com/oss-review-toolkit/ort-config/blob/80f72cbf3b6c0c3c9c62ef78b62fc9e2752bb427/tools/curations/buildSrc/src/main/kotlin/GenerateAspNetCoreCurationsTask.kt#L27) to make use of the new logic?

sschuberth: If so, I'd probably move this to `NuGet.kt`, as strictly speaking the logic is not *nuget-inspector* specific.

mnonnenmacher: I think it would be better to make it public just in case there are changes in the future. Otherwise the copy might get out of sync.

plugins/package-managers/python/src/main/kotlin/utils/PythonInspectorExtensions.kt
```diff
@@ -0,0 +1,182 @@
...
+import org.ossreviewtoolkit.model.PackageReference
+import org.ossreviewtoolkit.model.Project
+import org.ossreviewtoolkit.model.RemoteArtifact
+import org.ossreviewtoolkit.model.Scope
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+
+private const val TYPE = "PyPI"
+
+internal fun PythonInspector.Result.toOrtProject(
```
fviernau, line 38: I wonder if it's clear to future code readers that the file `PythonInspectorExtensions` is only about functions which map betweeen two models. Would it make sense to make the file name more specific?

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.plugins.packagemanagers.nuget.utils
+
+import org.ossreviewtoolkit.model.Identifier
+
+fun getIdentifierWithNamespace(type: String, name: String, version: String): Identifier {
```
fviernau, line 24: Hm, I believe I'm missing something, maybe you can help me finding out from where (besides the test) this (new) function is called.

plugins/package-managers/python/src/main/kotlin/utils/PythonInspectorExtensions.kt
sschuberth: > Would it make sense to make the file name more specific?

I'm open for suggestions!

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetUtils.kt
sschuberth: See a few comments [above](https://github.com/oss-review-toolkit/ort/pull/6046#discussion_r1314748990).

fviernau: Sorry, not following you.

sschuberth: Did you follow the link? The code needs to be public in order to be able to call it from the code in `ort-config` that generates the NuGet curations. (And then these curations need to be regenerated, obviously.)

plugins/package-managers/python/src/main/kotlin/utils/PythonInspectorExtensions.kt
sschuberth: Ok, so it seems to be up on me to propose names for your argument... I could call it `PythonInspectorModelUtils.kt` or `PythonInspectorModelMapper.kt` (and similar for NuGet then) if you prefer.

fviernau: > Ok, so it seems to be up on me...

No, I just wanted to point out the issue. (Because the commit message indicated to me that this separation maybe important to you) I didn't mean to ask you to change it.

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetUtils.kt
fviernau: > Did you follow the link? 

Yes, but hadn't noticed it was a different repo.

fviernau, line 24: Please add KDoc as this is now public. Maybe adding some hint why this unused function exists would  help as well?

fviernau (changes requested):

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetUtils.kt
sschuberth: > Hm, I believe I'm missing something, maybe you can help me finding out from where (besides the test) this (new) function is called.

Ah, now I realize what you were confused by: Due to a wrong conflict resolution while rebasing, indeed the actual call to `getIdentifierWithNamespace()` got lost. I've fixed that now (plus doing some more other fixes).

fviernau (approved):

PR #6050, opened 2022-11-08
Introduce a plugin manager for named plugins

sschuberth:

### DIFF
--- a/advisor/src/main/kotlin/Advisor.kt
+++ b/advisor/src/main/kotlin/Advisor.kt
@@ -20,8 +20,6 @@
 package org.ossreviewtoolkit.advisor
 
 import java.time.Instant
-import java.util.ServiceLoader
-import java.util.SortedMap
 
 import kotlinx.coroutines.async
 import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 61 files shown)


advisor/src/main/kotlin/Advisor.kt
```diff
@@ -45,14 +44,10 @@ class Advisor(
...
     companion object : Logging {
-        private val LOADER = ServiceLoader.load(AdviceProviderFactory::class.java)
-
         /**
          * All [advice provider factories][AdviceProviderFactory] available in the classpath, associated by their names.
          */
-        val ALL: SortedMap<String, AdviceProviderFactory> by lazy {
-            LOADER.iterator().asSequence().associateByTo(sortedMapOf(String.CASE_INSENSITIVE_ORDER)) { it.name }
-        }
+        val ALL = NamedPlugin.getAll<AdviceProviderFactory>()
```
fviernau, line 50: commit-msg: Please clarify the intention of removing the `lazy` execution.

fviernau (changes requested):

advisor/src/main/kotlin/Advisor.kt
sschuberth: The removal of `lazy` was actually an oversight; I've added it back.

fviernau (approved):

PR #6052, opened 2022-11-08
NuGetSupport: Simplify the code to map object values from URLs

sschuberth:
Replace the `mapFromUrl()` function with an extension function that
leverages `downloadText()` to simplify the code a bit.

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
@@ -40,7 +40,6 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.runBlocking
 
 import okhttp3.CacheControl
-import okhttp3.Request
 
 import org.apache.logging.log4j.kotlin.Logging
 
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -115,7 +113,10 @@ class NuGetSupport(serviceIndexUrls: List<String> = listOf(DEFAULT_SERVICE_INDEX
 
     private val serviceIndices = runBlocking {
         serviceIndexUrls.map {
-            async { mapFromUrl<ServiceIndex>(JSON_MAPPER, it) }
+            async {
+                val json = client.downloadText(it).getOrThrow()
+                JSON_MAPPER.readValue<ServiceIndex>(json)
```
mnonnenmacher, line 118: These two lines are repeated several times, I think refactoring `mapFromUrl` to contain those two lines and have a better name would be better, for example `ObjectMapper.downloadValue(url: String): T`.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
sschuberth: Done using a different function name.

mnonnenmacher (approved):

PR #6057, opened 2022-11-09
Client service code alignments

sschuberth:

### DIFF
--- a/advisor/src/main/kotlin/advisors/OssIndex.kt
+++ b/advisor/src/main/kotlin/advisors/OssIndex.kt
@@ -76,7 +76,10 @@ class OssIndex(name: String, serverUrl: String = OssIndexService.DEFAULT_BASE_UR
         return try {
             val componentReports = mutableMapOf<String, OssIndexService.ComponentReport>()
 
-            components.chunked(BULK_REQUEST_SIZE).forEach { chunk ->
+            val chunks = components.chunked(BULK_REQUEST_SIZE)
+            chunks.forEachIndexed { index, chunk ->
+                logger.debug { "Getting report for ${chunk.size} components (chunk $index of ${chunks.size})." }
+
                 val requestResults = getComponentReport(service, chunk).associateBy {
                     it.coordinates
                 }
@@ -132,7 +135,6 @@ class OssIndex(name: String, serverUrl: String = OssIndexService.DEFAULT_BASE_UR
         coordinates: List<String>
     ): List<OssIndexService.ComponentReport> =
         try {
-            logger.debug { "Querying component report from ${OssIndexService.DEFAULT_BASE_URL}." }
             service.getComponentReport(OssIndexService.ComponentReportRequest(coordinates))
         } catch (e: HttpException) {
             throw IOException(e)
--- a/advisor/src/main/kotlin/advisors/OssIndex.kt
+++ b/advisor/src/main/kotlin/advisors/OssIndex.kt
@@ -71,16 +71,16 @@ class OssIndex(name: String, serverUrl: String = OssIndexService.DEFAULT_BASE_UR
     override suspend fun retrievePackageFindings(packages: List<Package>): Map<Package, List<AdvisorResult>> {
         val startTime = Instant.now()
 
-        val components = packages.map { it.purl }
+        val purls = packages.map { it.purl }
 
         return try {
             val componentReports = mutableMapOf<String, OssIndexService.ComponentReport>()
 
-            val chunks = components.chunked(BULK_REQUEST_SIZE)
-            chunks.forEachIndexed { index, chunk ->
-                logger.debug { "Getting report for ${chunk.size} components (chunk $index of ${chunks.size})." }
+            val chunks = purls.chunked(BULK_REQUEST_SIZE)
+            chunks.forEachIndexed { index, chunkOfPurls ->
+                logger.debug { "Getting report for ${chunkOfPurls.size} components (chunk $index of ${chunks.size})." }
 
-                val requestResults = getComponentReport(service, chunk).associateBy {
+                val requestResults = getComponentReport(service, chunkOfPurls).associateBy {
                     it.coordinates
                 }
 
@@ -127,15 +127,15 @@ class OssIndex(name: String, serverUrl: String = OssIndexService.DEFAULT_BASE_UR
     }
 
     /**
-     * Invoke the [OSS Index service][service] to request detail information for the given [coordinates]. Catch HTTP
+     * Invoke the [OSS Index service][service] to request detail information for the given [purls]. Catch HTTP
      * exceptions thrown by the service and re-throw them as [IOException].
      */
     private suspend fun getComponentReport(
         service: OssIndexService,
-        coordinates: List<String>
+        purls: List<String>
     ): List<OssIndexService.ComponentReport> =
         try {
-            service.getComponentReport(OssIndexService.ComponentReportRequest(coordinates))
+            service.getComponentReport(OssIndexService.ComponentReportRequest(purls))
         } catch (e: HttpException) {
             throw IOException(e)
         }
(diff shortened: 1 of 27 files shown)


clients/scanoss/src/main/kotlin/ScanOssService.kt
```diff
@@ -38,6 +38,11 @@ typealias FullScanResponse = Map<String, List<ScanResponse>>
 
 interface ScanOssService {
     companion object {
+        /**
+         * The default API URL to use.
```
fviernau, line 42: Consider stripping "to use".

clients/osv/src/main/kotlin/OsvApiClient.kt
```diff
@@ -44,24 +44,41 @@ import retrofit2.http.Path
...
-                .baseUrl(serverUrl ?: SERVER_URL_PRODUCTION)
+                .baseUrl(serverUrl ?: Server.PRODUCTION.url)
                 .addConverterFactory(converterFactory)
                 .build()
                 .create(OsvApiClient::class.java)
         }
     }
 
+    enum class Server(val url: String) {
+        /**
```
fviernau, line 71: Consider dropping "OSV" from the docs as it is implied by the context.

clients/osv/src/main/kotlin/OsvApiClient.kt
```diff
@@ -44,24 +44,41 @@ import retrofit2.http.Path
  */
 interface OsvApiClient {
     companion object {
-        const val SERVER_URL_PRODUCTION = "https://api.osv.dev"
-        const val SERVER_URL_STAGING = "https://api-staging.osv.dev"
         const val BATCH_REQUEST_MAX_SIZE = 1000
 
         val JSON = Json.Default
 
+        /**
```
fviernau, line 51: commit-message-nit: "Mutual exclusive use" seems a bit confusing to me. It was obviously mutually exclusive before, because there is just on single parameter of type string which means: only one value can be passed, not multiple.

clients/osv/src/main/kotlin/OsvApiClient.kt
```diff
@@ -44,24 +44,41 @@ import retrofit2.http.Path
...
+         * Create an OsvApiClient instance for communicating with the given [server], optionally using a pre-built
+         * OkHttp [client].
+         */
+        fun create(server: Server, client: OkHttpClient? = null): OsvApiClient =
+            create(server.url, client)
```
fviernau, line 56: If we don't want to limit the use of the library to just 2 server URLs (which I'd agree we shouldn't), then IMO it adds quite a bit boiler plate compared to the value it brings - as still values other than staging and production can be passed.

advisor/src/main/kotlin/advisors/OssIndex.kt
```diff
@@ -132,7 +132,7 @@ class OssIndex(name: String, serverUrl: String = OssIndexService.DEFAULT_BASE_UR
         coordinates: List<String>
     ): List<OssIndexService.ComponentReport> =
         try {
-            logger.debug { "Querying component report from ${OssIndexService.DEFAULT_BASE_URL}." }
+            logger.debug { "Querying component report." }
```
fviernau, line 135: Out of scope: 
- maybe we should rename coordinates to e.g. purls for clarify.
- maybe move the output to the caller, which could then at least tell "Querying component reports chunk $i of $n."

fviernau (dismissed):

advisor/src/main/kotlin/advisors/OssIndex.kt
sschuberth: > maybe we should rename coordinates to e.g. purls for clarity.

Good idea, done in a new commit.

> maybe move the output to the caller, which could then at least tell "Querying component reports chunk $i of $n."

I like that also, done.

clients/scanoss/src/main/kotlin/ScanOssService.kt
sschuberth: This was moved from `ScanOssConfig.kt`, but as I'm slightly adjusting the formatting anyway, I agree I could as well drop this suffix. Will do.

fviernau: Hadn't notice it was moved...

clients/osv/src/main/kotlin/OsvApiClient.kt
sschuberth: I agree and have dropped that part from the commit message.

clients/osv/src/main/kotlin/OsvApiClient.kt
sschuberth: I agree this commit is a bit "fluffy", but I somehow liked how it's done for `ClearlyDefined`. As my goal was general alignment, the alternative would have been to drop the server enum code from `ClearlyDefined`, which I liked slightly less.

PS: I'll add a new commit that hopefully adds some more value to having a server enum at least in the `ClearlyDefined` case.

advisor/src/main/kotlin/advisors/OssIndex.kt
```diff
@@ -77,6 +77,8 @@ class OssIndex(name: String, serverUrl: String = OssIndexService.DEFAULT_BASE_UR
             val componentReports = mutableMapOf<String, OssIndexService.ComponentReport>()
 
             components.chunked(BULK_REQUEST_SIZE).forEach { chunk ->
+                logger.debug { "Querying component report (chunk $chunk of $BULK_REQUEST_SIZE)." }
```
fviernau: Was it intended to include all PURLs and the size limit of a chunk into the log output?

```
val chunks = compontents.chunked(BULK_REQUEST_SIZE)
chunks.forEachIndexed ...
    "Querying $chunk.size component reports (chunk "$i+1 of $chunks.size)"
```

Note: I believe it's a batch query so report**s** IMO should use plural form.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
```diff
@@ -76,21 +76,27 @@ interface ClearlyDefinedService {
...
-        DEVELOPMENT("https://dev-api.clearlydefined.io"),
+        DEVELOPMENT(
+            "https://dev-api.clearlydefined.io",
+            "https://dev.clearlydefined.io",
+            "https://github.com/clearlydefined/curated-data-dev"
+        ),
 
         /**
-         * The ClearlyDefined server when running locally.
+         * The server when running locally.
```
fviernau, line 99: Should LOCAL use empty strings rather than null values. E.g. I wonder if the caller of local expects a fallback to PRODUCTION.

fviernau (changes requested):

advisor/src/main/kotlin/advisors/OssIndex.kt
sschuberth: > Was it intended to include all PURLs and the size limit of a chunk into the log output?

No, that was an oversight. Thanks for finding it.

> Note: I believe it's a batch query so reports IMO should use plural form.

Within a single loop iteration just one report (for multiple pURLs) is queried, so I believe singular to be correct.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
sschuberth: I believe these environments shouldn't be mixed, and there should be no fallback from / combination of LOCAL and PRODUCTION URLs. With a non-nullable value there's a risk that it's used as-is as the semantics of an empty string to mean "undefined" are a bit unclear. IMO `null` is a better value value for a string / URL if you want to express "undefined".

fviernau: I see. I just had blindly assumed null does not mean "undefined" but it does mean "fall back to default" as it does so in other places of the client server code.

PR #6061, opened 2022-11-10
CopyrightStatementsProcessor: Make `process()` deterministic

sschuberth:

### DIFF
--- a/utils/ort/src/test/kotlin/CopyrightStatementsProcessorTest.kt
+++ b/utils/ort/src/test/kotlin/CopyrightStatementsProcessorTest.kt
@@ -29,12 +29,12 @@ import org.ossreviewtoolkit.model.yamlMapper
 class CopyrightStatementsProcessorTest : WordSpec({
     "process" should {
         "return a result with items merged by owner and prefix, sorted by owner and year" {
-            val input = File("src/test/assets/copyright-statements.txt").readLines()
-
-            val result = CopyrightStatementsProcessor.process(input).toYaml()
-
+            val statements = File("src/test/assets/copyright-statements.txt").readLines()
             val expectedResult = File("src/test/assets/copyright-statements-expected-output.yml").readText()
-            result shouldBe expectedResult
+
+            val actualResult = CopyrightStatementsProcessor.process(statements.shuffled()).toYaml()
+
+            actualResult shouldBe expectedResult
         }
     }
 })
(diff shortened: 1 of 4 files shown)


utils/ort/src/test/kotlin/CopyrightStatementsProcessorTest.kt
```diff
@@ -28,13 +28,13 @@ import org.ossreviewtoolkit.model.yamlMapper
...
 class CopyrightStatementsProcessorTest : WordSpec({
     "process" should {
-        "return a result with items merged by owner and prefix, sorted by owner and year" {
-            val input = File("src/test/assets/copyright-statements.txt").readLines()
+        val statements = File("src/test/assets/copyright-statements.txt").readLines()
+        val expectedResult = File("src/test/assets/copyright-statements-expected-output.yml").readText()
 
-            val result = CopyrightStatementsProcessor.process(input).toYaml()
+        "return a result with items merged by owner and prefix, sorted by owner and year".config(invocations = 5) {
+            val actualResult = CopyrightStatementsProcessor.process(statements.shuffled()).toYaml()
```
fviernau, line 35: For the reader it's IMO unclear why `5` invocations. A presume this was due to having some verification for it being deterministic. 

If the test should become flake, we'll notice it anyway. Making the notice 5x more early isn't necessary IMO. What do you think?

fviernau (changes requested):

utils/ort/src/test/kotlin/CopyrightStatementsProcessorTest.kt
```diff
@@ -28,13 +28,13 @@ import org.ossreviewtoolkit.model.yamlMapper
 
 class CopyrightStatementsProcessorTest : WordSpec({
     "process" should {
-        "return a result with items merged by owner and prefix, sorted by owner and year" {
-            val input = File("src/test/assets/copyright-statements.txt").readLines()
+        val statements = File("src/test/assets/copyright-statements.txt").readLines()
```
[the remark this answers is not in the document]
sschuberth: I had moved the initialization of these variables out of the inner test for performance reasons: Due to the `invocations = 5` the files would otherwise be read on each iteration, which is unnecessary.

utils/ort/src/test/kotlin/CopyrightStatementsProcessorTest.kt
sschuberth: > For the reader it's IMO unclear why `5` invocations.

The choice of 5 was indeed arbitrary; I had aligned it with other similar test code in the code base.

> If the test should become flake, we'll notice it anyway.

I agree that a single pass that calls `shuffle()` probably is enough, will change.

fviernau (approved):

PR #6068, opened 2022-11-10
Support the public VulnerableCode instance 

sschuberth:

/cc @pombredanne

### DIFF
--- a/clients/vulnerable-code/src/main/kotlin/VulnerableCodeService.kt
+++ b/clients/vulnerable-code/src/main/kotlin/VulnerableCodeService.kt
@@ -38,6 +38,8 @@ import retrofit2.http.POST
  */
 interface VulnerableCodeService {
     companion object {
+        const val PUBLIC_SERVER_URL = "https://public.vulnerablecode.io"
+
         /**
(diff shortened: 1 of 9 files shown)


fviernau (dismissed):

clients/vulnerable-code/src/main/kotlin/VulnerableCodeService.kt
```diff
@@ -46,11 +48,21 @@ interface VulnerableCodeService {
         /**
          * Create a new service instance that connects to the [url] specified and uses the optionally provided [client].
          */
-        fun create(url: String, client: OkHttpClient? = null): VulnerableCodeService {
+        fun create(url: String? = null, apiKey: String? = null, client: OkHttpClient? = null): VulnerableCodeService {
+            val vulnerableCodeClient = (client ?: OkHttpClient()).newBuilder()
```
oheger-bosch, line 52: Here an interceptor is always added, but it actually does anything only if an API key is defined. I think the logic should be changed, so that the interceptor is added only if the API key is defined - i.e. move the conditional logic from runtime to initialization time.

oheger-bosch (approved):

PR #6071, opened 2022-11-11
Support deserializing NuGet package references without a version 

sschuberth:


analyzer/src/main/kotlin/managers/utils/NuGetAllPackageData.kt
```diff
@@ -23,7 +23,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty
 
-internal class NuGetAllPackageData(
+class NuGetAllPackageData(
```
fviernau, line 28: commit msg: "public to that its nested" is unclear to me

commit: Why is it necessary to expose things beyond the module scope? E.g. why do the functions
which return `PackageSpec` need to be made public?

analyzer/src/main/kotlin/managers/utils/NuGetServiceIndex.kt
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.ossreviewtoolkit.analyzer.managers.utils
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonProperty
+
+// See https://docs.microsoft.com/en-us/nuget/api/service-index.
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class ServiceIndex(
```
fviernau, line 27: I've investigated the data classes in the original file.
To me it seems that all the classes in `NuGetAllPackageData.kt` except for the 
top-level class do represent the NuGet data model and are used for de-serialization,
while the top-level one is something different: it's a data model of Ort's Nuget implementation.

I believe that separating model classes while keeping the non-model class seems to not extract the things which belongs there the least, while it extracts two classes which actually belong to the model. Also the thing which gets extracted is similar close related as the classes which remain in the file. Does this make any sense?

How about first doing the following?

1. move the top level class `NuGetAllPackageData` into it's only user. IMO it is similar to `NuGetDependency` and belongs next to it.
2. rename the file to NuGetModel
and Note: The result would be no more nesting of classes in this file.

Given that state, maybe it suffices to keep the classes in a single file
and structure / nest them as follows? 
```
PackageData

PackageDetails
  DependencyGroup
  Dependency

PackageSpec
  Metadata
  License
  Repository

ServiceIndex
  ServiceResouce -> rename to Resource?
```

fviernau (changes requested):

analyzer/src/main/kotlin/managers/utils/NuGetAllPackageData.kt
sschuberth: > commit msg: "public to that its nested" is unclear to me

"to" was a typo and was supposed to read "so".

> Why is it necessary to expose things beyond the module scope?

I've solved this now in the alternative way and made `parseLicenses()` / `parseAuthors()` internal.

analyzer/src/main/kotlin/managers/utils/NuGetServiceIndex.kt
sschuberth: > all classses [...] represent the NuGet data model and are used for de-serialization, while the top-level one is something different: it's a data model of Ort's Nuget implementation.

That's correct.

> Also the thing which gets extracted is similar close related as the classes which remain in the file. Does this make any sense?

I view it a bit differently (and I've reworded the commit message accordingly): My goal in that commit was to separate classes that are used for deserialization, but that deserialize completely different things. One set of classes deserializes information about where to get package metadata *from*, whereas the other set of classes deserializes the actual package metadata.

I can see the point in further separating classes used directly for deserialization from classes that just bundle / make use of these classes and will take a look at that.

PS: About "separating model classes while keeping the non-model class", to me "model classes" are not only those used directly for deserialization, but any (data) classes that represent the data model, including classes that just bundle the classes used for deserialization.

fviernau: > PS: About "separating model classes while keeping the non-model class", to me "model classes" are not only those used directly for deserialization, but any (data) classes that represent the data model, including classes that just bundle the classes used for deserialization.

Well, when I used the term "model classes" it seems it wasn't sufficiently clear.

I wanted to propose to clearly separate the following:
1. Classes which represent upstream data model (these are the ones which I referred to as "model" classes)
2. Classes which represent ORT's custom data model

fviernau: Yes, that new commit is what I actually thought would be needed in order to satisfy this one, thanks!

analyzer/src/main/kotlin/managers/utils/NuGetDependency.kt
```diff
@@ -0,0 +1,125 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
```
fviernau, line 19: The extracted code to me seems a little bit like cyclic dependency. I believe it needs refactoring 
and doing so I guess it's easier if it's not separate before refactoring. 

I'm good with all other commits, would you be fine splitting this out so we can merge the rest and investigate this commit separatly?

fviernau: Ah sorry, then I overlooked that - thought it was relatively easy. Thanks anyway.

fviernau (approved):

PR #6073, opened 2022-11-11
PostgresConnection: Exclude some default value from serialization

fviernau:
Clean-up the output of `ort config --show-active` a bit.

### DIFF
--- a/model/src/main/kotlin/config/PostgresConnection.kt
+++ b/model/src/main/kotlin/config/PostgresConnection.kt
@@ -19,6 +19,8 @@
 
 package org.ossreviewtoolkit.model.config
 
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonInclude.Include
 import com.fasterxml.jackson.annotation.JsonProperty
 
 data class PostgresConnection(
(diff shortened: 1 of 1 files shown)


model/src/main/kotlin/config/PostgresConnection.kt
```diff
@@ -53,18 +54,21 @@ data class PostgresConnection(
      * The full path of the certificate file.
      * See: https://jdbc.postgresql.org/documentation/head/connect.html
      */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
```
sschuberth, line 57: Shouldn't instead the whole class be annotated with `@JsonInclude(JsonInclude.Include.NON_NULL)` or even `@JsonInclude(JsonInclude.Include.NON_DEFAULT)`? (The latter probably hides too much what a user might be interested to see in the active configuration.)

fviernau: I decided for per-property because I thought it is more obvious / harder to miss that irrelevant properties shall be excluded. 
Does this make sense to you as well?

sschuberth: Per-property is fine with me, but I still believe it should be `NON_NULL` instead of `NON_EMPTY`, or?

sschuberth (approved):

PR #6041, opened 2022-11-06
docker: Build independent stages as images

heliocastro:
New docker_build will generate images on local docker cache and then future ort build in docker will only
recreate necessary transient images and ort image itself instead of rebuild everything.

- Create multiple images as transient images for final container
- Create single place for set versions ( .versions )
- Set base as defined non-privileged user. Default "ort"
- Build all components and packages installs as non-privileged user
- Remove all auxiliary scripts as not necessary anymore

Signed-off-by: Helio Chissini de Castro <heliocastro@gmail.com>

PR #6075, opened 2022-11-11
docker: Prepare base layer to non privileged user

heliocastro:
To accomplish the last step of docker multi stage change, we need take small step to achieve the result of the draft PR https://github.com/oss-review-toolkit/ort/pull/6041, as too big to change in a single step.

This initial needed changes accomplish:

- Updated the version of dockerfile spec to accept build-contexts ( buildx )
- Add extra packages to base to eliminate the need of a builder image, reducing build time after cached and providing the minimal packages needed to language package managers  build native bindings, mostly cargo, ruby and python 
- Make base image run as non-privileged user instead of root
- Add proxy script for apt, as now image is user based and apt will run in sudo mode
- Modify permissions on android deploy dir, as now need to be run as user, and github actions use a **runner** user called runner.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:1.3
+# syntax=docker/dockerfile:1.4
 
 # Copyright (C) 2020 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
 #
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,6 +72,7 @@ ARG USER_ID=1000
 ARG USER_GID=$USER_ID
 ARG HOMEDIR=/home/ort
 ENV HOME=$HOMEDIR
+ENV USER=$USERNAME
 
 # Non privileged user
 RUN groupadd --gid $USER_GID $USERNAME \
@@ -81,6 +82,9 @@ RUN groupadd --gid $USER_GID $USERNAME \
     --shell /bin/bash \
     --home-dir $HOMEDIR \
     --create-home $USERNAME
+
+RUN chgrp $USER /opt \
+    && chmod g+wx /opt
 
 # sudo support
 RUN echo "$USERNAME ALL=(root) NOPASSWD:ALL" > /etc/sudoers.d/$USERNAME \
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,8 +83,8 @@ RUN echo "$USERNAME ALL=(root) NOPASSWD:ALL" > /etc/sudoers.d/$USERNAME \
 
 COPY docker/00-add_local_path.sh /etc/profile.d/
 
-# Copy ort scripts
-COPY scripts /etc/scripts
+# Import certificates scripts only
+COPY scripts/import_certificates.sh /etc/scripts/import_certificates.sh
 
 # Set this to a directory containing CRT-files for custom certificates that ORT and all build tools should know about.
 ARG CRT_FILES=""
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,11 +40,16 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     dirmngr \
     gcc \
     git \
+    g++ \
     gnupg2 \
     iproute2 \
     libarchive-tools \
+    libffi-dev \
+    libgmp-dev \
     libz-dev \
     locales \
+    lzma \
+    make \
     netbase \
     openssh-client \
     openssl \
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,10 +24,6 @@ ENV LANG=en_US.UTF-8
 ENV LANGUAGE=en_US:en
 ENV LC_ALL=en_US.UTF-8
 
-RUN echo $LANG > /etc/locale.gen \
-    && locale-gen en_US.UTF-8 \
-    && update-locale LANG=en_US.UTF-8
-
 # Base package set
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     --mount=type=cache,target=/var/lib/apt,sharing=locked \
@@ -57,6 +53,10 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     wget \
     xz-utils \
     && rm -rf /var/lib/apt/lists/*
+
+RUN echo $LANG > /etc/locale.gen \
+    && locale-gen $LANG \
+    && update-locale LANG=$LANG
 
 ARG USERNAME=ort
 ARG USER_ID=1000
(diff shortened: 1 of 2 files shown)


nicorikken (dismissed):
Great, now up to the rest of the improvements.

MarcelBochtler:
@heliocastro there are a lot of unrelated changes in a single commit.
For instance, renaming `ort-base-image`, printing the architecture and setting `-qq` for `apt-get` are all unrelated to the change the commit message mentions.
Can you please split the unrelated changes into separate commits?
Afterwards, please describe in the commit message why some changes are required (e.g. I don't understand why the additionally installed tools are required to run docker in non-root mode).

sschuberth:
> This changes are needed if you saw the related drafted PR mentioned.

But the question is *why* are these changes needed in the context of "Prepare base layer to non privileged user"? The explanation needs to go into the respective commit's message, not in another draft PR that you refer to.

> I honestly think break more down this commit will end up in a single line unexplained commits

Single-line commit are perfectly fine, but of course not unexplained ones. There are single-line code changes (not in this PR, but in general) that deserve a full paragraph of commit message to explain *why* this change is the right thing to do, in the context of the PR.

> So is the explanations good enough ?

*Which* explanations? There are none in the commit message body, and that's the only place that counts. PR comments do not count as they're not part of the Git history.

heliocastro:
Updated commit log

MarcelBochtler:
> Updated commit log

Unfortunately, that does not fix the issues I mentioned in my comment above.
I'll try to rephrase:
As outlined in our [CONTRIBUTING.md](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md#git-history) you should:
- Make separate commits for logically separate changes.
- Describe your changes well.

As an example from your changes, let's take setting the apt proxy which is required for a non-root image:
1. You'd add these changes:
```
# Set proxy for apt.
COPY docker/set_apt_proxy.sh /etc/scripts/set_apt_proxy.sh
RUN /etc/scripts/set_apt_proxy.sh
```
together with the new `set_apt_proxy.sh` in one commit.

2. The commit message for this single commit explains why this change is required.
In this case, it could be something like (I'm sure you can explain it better than I tried to):
```
Docker: Add proxy script for apt

As a preparation for a docker image running as non privileged user, the
proxy has to be set explicitly because apt will continue to run as root.

Signed-off-by: ....
```

heliocastro:
Multiple commits now

nicorikken (dismissed):
Ok I must admit the smaller commits do make it even easier to review. I still approve it.

Dockerfile
```diff
@@ -330,7 +330,7 @@ ENV ANDROID_HOME=/opt/android-sdk
 COPY --from=androidbuild /usr/bin/repo /usr/bin/
 COPY --from=androidbuild /etc/profile.d/android.sh /etc/profile.d/
 COPY --chown=$USERNAME:$USERNAME --from=androidbuild ${ANDROID_HOME} ${ANDROID_HOME}
-RUN chmod o+rw ${ANDROID_HOME}
+RUN chmod -R o+rw ${ANDROID_HOME}
```
MarcelBochtler, line 333: Commit message: This doesn't only affect GitHub actions, it also affects all other users of the image, right?
Maybe just remove the `in github actions` part of the commit message.

MarcelBochtler (commented):
Thanks for splitting up the PR. Quite clear now what you're doing and why you're doing the changes.

One nit though: Please check your commit messages for grammatical errors and the usage of correct punctuations (e.g. multiple sentences are missing a trailing `.`, and the first commit only contains a single list item)

Regarding the changes, I added my questions in line.

Dockerfile
```diff
@@ -72,6 +72,7 @@ ARG USER_ID=1000
 ARG USER_GID=$USER_ID
 ARG HOMEDIR=/home/ort
 ENV HOME=$HOMEDIR
+ENV USER=$USERNAME
```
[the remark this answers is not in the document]
heliocastro: Not only. So, the subsequent layers that will build using base will need to be assigned to USER ( Linux base user ), and default to it. 
So, before we're setting USERNAME and worst, we are setting username as ARG, which is not passed along to other layers, different from ENV, which then was completely useless before.
I will update commit message properly

Dockerfile
heliocastro: Ok, changed the commit message

MarcelBochtler:
> Thanks for splitting up the PR. Quite clear now what you're doing and why you're doing the changes.
> 
> One nit though: Please check your commit messages for grammatical errors and the usage of correct punctuations (e.g. multiple sentences are missing a trailing `.`, and the first commit only contains a single list item)

@heliocastro I'm waiting for a response regarding the quoted issues. Otherwise, the changes look good to me.

heliocastro:
Sorry, was on a work travel.
@MarcelBochtler I think i already solved the last quoted entries, but will double check.

MarcelBochtler (changes requested):
Nearly there.
I'm sorry for being that nitpicky. But it is important that we are able to understand (and are able to backtrack) the reasons why changes are done.

Dockerfile
```diff
@@ -330,7 +330,7 @@ ENV ANDROID_HOME=/opt/android-sdk
 COPY --from=androidbuild /usr/bin/repo /usr/bin/
```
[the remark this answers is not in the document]
heliocastro: Fixed

Dockerfile
```diff
@@ -72,6 +72,7 @@ ARG USER_ID=1000
 ARG USER_GID=$USER_ID
```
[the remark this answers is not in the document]
heliocastro: Fixed

MarcelBochtler (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:1.3
+# syntax=docker/dockerfile-upstream:1.4
```
sschuberth, line 1: I believe the "-upstream" part could (and should) be dropped here.

Dockerfile
```diff
@@ -82,6 +83,9 @@ RUN groupadd --gid $USER_GID $USERNAME \
     --home-dir $HOMEDIR \
     --create-home $USERNAME
 
+RUN chgrp $USER /opt \
```
sschuberth, line 86: Do the leading bullet item in the commit message still make sense? Their look like a left-over from rewording the commit message.

heliocastro: Yes. 
The change above this line, set the environment to user ( line 75 ) and this reflects on the layers that depends on base now

tsteenbe (approved):

MarcelBochtler (approved):

PR #6077, opened 2022-11-11
docker: Build with detected architecture platform

heliocastro:


scripts/docker_build.sh
```diff
@@ -21,12 +21,13 @@ DOCKER_ARGS=$@
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
 GIT_REVISION=$(git describe --abbrev=10 --always --tags --dirty --match=[0-9]*)
+PLATFORM="${PLATFORM:-linux/$(arch)}"
```
MarcelBochtler, line 24: On my Intel Macbook:
```
❯ arch
i386
```

Would this cause issues?

MarcelBochtler: I checked, and it causes issues:
```
 => ERROR [internal] load metadata for docker.io/library/eclipse-temurin:11-jdk-jammy                                                                       1.8s
------
 > [internal] load metadata for docker.io/library/eclipse-temurin:11-jdk-jammy:
------
ERROR: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: rpc error: code = Unknown desc = no match for platform in manifest sha256:ff15d0a337affefa41a3619cb5d8626c11e0d38c84e85ac3b0fafc5f86ff68e9: not found
```

sschuberth: Similarly, my Fedora Linux returns "x86_64" for `arch`, so that's also not the expected "amd64"...

sschuberth: PS: Looking at the normalization done [here](https://github.com/containerd/containerd/blob/v1.4.3/platforms/platforms.go#L63), I believe we can't simply use `arch`.

sschuberth: The [docs](https://docs.docker.com/engine/reference/commandline/buildx_build/#platform) say:

> The default value will be the current platform of the buildkit daemon.

So can't we simply drop the whole `--platform` option and be good?

sschuberth (approved):

MarcelBochtler:
The failing test looks unrelated to this change. Restarted it anyway to be sure.

PR #6082, opened 2022-11-14
Unify hasLabel() and labelContains()

fviernau:
See individual commits.

**Breaking change:** Policy rules which use `labelContains()` must be changed to use `hasLabel()` instead.

### DIFF
--- a/evaluator/src/main/kotlin/Rule.kt
+++ b/evaluator/src/main/kotlin/Rule.kt
@@ -107,25 +107,14 @@ abstract class Rule(
 
     /**
      * A [RuleMatcher] that checks whether a [label] exists in the [ORT result][OrtResult.labels]. If [value] is null
-     * the value of the label is ignored.
+     * the value of the label is ignored. If [splitValue] is true, the label value is interpreted as comma-separated
+     * list.
      */
-    fun hasLabel(label: String, value: String? = null) =
+    fun hasLabel(label: String, value: String? = null, splitValue: Boolean = true) =
         object : RuleMatcher {
-            override val description = "hasLabel(${listOfNotNull(label, value).joinToString()})"
+            override val description = "hasLabel(${listOfNotNull(label, value, splitValue).joinToString()})"
 
-            override fun matches() = ruleSet.ortResult.hasLabel(label, value, splitValue = false)
-        }
-
-    /**
-     * A [RuleMatcher] that checks whether a [label] exists in the [ORT result][OrtResult.labels] and contains a
-     * specific [value]. The value of the label is interpreted as a comma-separated list. The check is successful if
-     * this list contains the [value].
-     */
-    fun labelContains(label: String, value: String) =
-        object : RuleMatcher {
-            override val description = "labelContains($label, $value)"
-
-            override fun matches() = value in ruleSet.ortResult.getLabelValues(label)
+            override fun matches() = ruleSet.ortResult.hasLabel(label, value, splitValue)
         }
 
     /**
--- a/evaluator/src/main/kotlin/Rule.kt
+++ b/evaluator/src/main/kotlin/Rule.kt
@@ -113,7 +113,7 @@ abstract class Rule(
         object : RuleMatcher {
             override val description = "hasLabel(${listOfNotNull(label, value).joinToString()})"
 
-            override fun matches() = ruleSet.ortResult.hasLabel(label, value)
+            override fun matches() = ruleSet.ortResult.hasLabel(label, value, splitValue = false)
         }
 
     /**
--- a/evaluator/src/test/kotlin/RuleTest.kt
+++ b/evaluator/src/test/kotlin/RuleTest.kt
@@ -144,25 +144,29 @@ class RuleTest : WordSpec() {
 
                 matcher.matches() shouldBe false
             }
-        }
 
-        "labelContains()" should {
             "return true if the ORT result contains the label with the value" {
-                val matcher = createRule().labelContains("list", "value2")
+                val matcher = createRule().hasLabel("list", "value2")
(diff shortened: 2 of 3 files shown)


MarcelBochtler (dismissed):

evaluator/src/main/kotlin/Rule.kt
```diff
@@ -106,18 +106,18 @@ abstract class Rule(
     }
 
     /**
-     * A [RuleMatcher] that checks whether a [label] exists in the [ORT result][OrtResult.labels]. If [value] is null
-     * the value of the label is ignored.
+     * A [RuleMatcher] that checks whether a label denoted by the given [key] exists and optionally (in case the given
```
sschuberth, line 109: Maybe simply: ".. a label named [key] ..."

evaluator/src/main/kotlin/Rule.kt
```diff
@@ -106,18 +106,18 @@ abstract class Rule(
...
+     * [value] is not null) if [value] is amongst the label values of that label.
```
sschuberth, line 110: I believe that description is misleading now. It sounds like if the label's value is e.g. a comma-separated string, then that string is interpreted as a list, and it is checked whether `value` is part of that list. But that's not how it's implemented.

sschuberth (changes requested):

evaluator/src/main/kotlin/Rule.kt
sschuberth: > I wonder whether there is actually any use case for `hasLabel()`.

I believe there is. Lazy people might simply want to check for the presence of a `DO_SOMETHING` label instead of requiring the value to be "true" / "false" and parsing it as a boolean.

fviernau: Let me try to clarify what I meant. 

Can we replace labelContains() and hasLabel() (see Rule.kt) with the following one in OrtResult plus 
an identical one (which returns a matcher instead of bool) in the Rule class?
```
    fun hasLabel(key: String, value: String? = null) =
        if (value == null) {
            key in labels
        } else {
            value in getLabelValues(key)
        }
```

fviernau: @sschuberth I've extracted the IIUC uncontroversial bit to https://github.com/oss-review-toolkit/ort/pull/6084 to unblock me.
Once this is merged, I'll make a force push to this PR proposing something on top.

mnonnenmacher: > Lazy people might simply want to check for the presence of a DO_SOMETHING label instead of requiring the value to be "true" / "false" and parsing it as a boolean.

It has nothing to do with being lazy, but yes, a label does not require to have a value.

> Can we replace labelContains() and hasLabel() (see Rule.kt) with the following one...

That would make it impossible to check the exact value of a label, because it is always interpreted as a comma-separated list. Currently the user can decide how the value is interpreted so I think I would prefer to keep it as is.

evaluator/src/main/kotlin/Rule.kt
```diff
@@ -106,18 +106,18 @@ abstract class Rule(
...
      */
-    fun hasLabel(label: String, value: String? = null) =
+    fun hasLabel(key: String, value: String? = null) =
```
mnonnenmacher, line 112: I prefer the old name because the string before the `=` is the actual label.

evaluator/src/main/kotlin/Rule.kt
sschuberth: > It has nothing to do with being lazy, but yes, a label does not require to have a value.

I should have put "lazy" in quotes as it was a flippant remark to be taken with a wink.

> Currently the user can decide how the value is interpreted so I think I would prefer to keep it as is.

I tend to agree, although it's probably quite unlike that a user is using commas in the value and these commas are *not* list item delimiters.

fviernau: I've updated it and hope it now fits the preferences of both of you @mnonnenmacher @sschuberth .

evaluator/src/main/kotlin/Rule.kt
```diff
@@ -113,7 +113,7 @@ abstract class Rule(
         object : RuleMatcher {
             override val description = "hasLabel(${listOfNotNull(label, value).joinToString()})"
 
-            override fun matches() = ruleSet.ortResult.hasLabel(label, value)
+            override fun matches() = ruleSet.ortResult.hasLabel(label, value, false)
```
sschuberth: Looking at this commit only, it's unclear what the benefit is to use `hasLabel` instead of `getLabelValues` in `labelContains`. And it's probably hard to find an argument for the commit message, as the next commit even removes `labelContains`. So why bother changing it here in the first place?

sschuberth: Boolean arguments should be named for clarity.

evaluator/src/test/kotlin/RuleTest.kt
```diff
@@ -144,26 +144,30 @@ class RuleTest : WordSpec() {
...
 
             "return false if the ORT result does not contain the label with the value" {
-                val matcher = createRule().labelContains("list", "value4")
+                val matcher = createRule().hasLabel("list", "value4")
 
                 matcher.matches() shouldBe false
             }
+
+            "return true if the ORT result contains the label values which includes commas" {
+                val matcher = createRule().hasLabel("list", "value1, value2, value3", false)
```
sschuberth, line 167: The boolean argument should be named.

sschuberth (changes requested):

sschuberth (approved):

PR #6085, opened 2022-11-14
Analyzer: Fix `mustRunAfter` to eventually hang when configured globally 

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -153,7 +153,7 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
         runBlocking {
             managedFiles.entries.map { (manager, files) ->
                 val mustRunAfter = config.getPackageManagerConfiguration(manager.managerName)?.mustRunAfter?.toSet()
-                    ?: packageManagerDependencies[manager]?.mapTo(mutableSetOf()) { it.managerName }.orEmpty()
+                    ?: packageManagerDependencies[manager].orEmpty()
 
                 PackageManagerRunner(
                     manager = manager,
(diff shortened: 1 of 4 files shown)


analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -175,37 +172,42 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
...
-        val result = mutableMapOf<PackageManager, MutableSet<PackageManager>>()
+        val result = mutableMapOf<PackageManager, MutableSet<String>>()
 
         managedFiles.keys.forEach { packageManager ->
             val dependencies = packageManager.findPackageManagerDependencies(managedFiles)
-            dependencies.mustRunAfter.forEach { name ->
-                val managerForName = allPackageManagers[name]
+
+            val mustRunAfter = config.getPackageManagerConfiguration(packageManager.managerName)?.mustRunAfter?.toSet()
+                ?: dependencies.mustRunAfter
```
sschuberth, line 186: There might still be a problem with this solution as now a dynamic `mustRunBefore` (see below) may override an inverse `mustRunAfter` from the static / global configuration, which might not be what we want. I need to clarify this with @mnonnenmacher.

sschuberth: I've added a fixup commit that I'll eventually squash later that shows how this could be addressed.

sschuberth: Probably yes, but that filter would need to be done on the `config`-retrieved dependencies only, as `determinePackageManagerDependencies()` and thus the `packageManagerDependencies` map already have that filtering applied. And as `determinePackageManagerDependencies()` already does that filtering, I wanted to rather make reuse of it instead of implementing it again. Also, reusing `determinePackageManagerDependencies()` has the benefit to also get the debug log output when a package manager dependency is ignored.

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -152,14 +152,11 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
...
             managedFiles.entries.map { (manager, files) ->
-                val mustRunAfter = config.getPackageManagerConfiguration(manager.managerName)?.mustRunAfter?.toSet()
-                    ?: packageManagerDependencies[manager].orEmpty()
-
                 PackageManagerRunner(
                     manager = manager,
                     definitionFiles = files,
                     labels = labels,
-                    mustRunAfter = mustRunAfter,
+                    mustRunAfter = packageManagerDependencies[manager].orEmpty(),
```
mnonnenmacher, line 159: Commit message nit: "definit**i**on files"

mnonnenmacher (changes requested):
LGTM, just one nit.

mnonnenmacher (approved):

PR #6088, opened 2022-11-16
docker: Sanitize Rust in docker

heliocastro:
- Create scratch layer only with result dir to future small caching
- Add PATH change and ENV entries to runtime image
- Remove the auxiliary script as not needed anymore

This change is part of the process to remove all bash login requirements, simplifying pipeline usage


Dockerfile
```diff
@@ -197,14 +197,19 @@ RUN . $NVM_DIR/nvm.sh \
 
 #------------------------------------------------------------------------
 # RUST - Build as a separate component
-FROM build AS rustbuild
+FROM ort-base-image AS rustbuild
```
MarcelBochtler, line 200: Unrelated change. Please split to separate commit.

MarcelBochtler (changes requested):
Please adjust the commit message(s) to comply with our [contribution guidelines](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md):
- One change per commit.
- Grammatically correct commit message.
- Sentences should end with a `.`.

Dockerfile
heliocastro: @MarcelBochtler How this is unrelated ?

tsteenbe: @MarcelBochtler  Not following you here as well ...

Dockerfile
```diff
@@ -315,10 +315,12 @@ COPY --chown=$USERNAME:$USERNAME --from=nodebuild ${NVM_DIR} ${NVM_DIR}
...
-ARG RUST_HOME=/opt/rust
-ARG CARGO_HOME=${RUST_HOME}/cargo
-COPY --chown=$USERNAME:$USERNAME --from=rustbuild /opt/rust /opt/rust
-COPY docker/rust.sh /etc/profile.d/
+ENV RUST_HOME=/opt/rust
+ENV CARGO_HOME=${RUST_HOME}/cargo
+ENV RUSTUP_HOME=${RUST_HOME}/rustup
+ENV PATH=$PATH:${CARGO_HOME}/bin:${RUSTUP_HOME}/bin
+
+COPY --chown=$USER:$USER --from=rust /opt/rust /opt/rust
```
fviernau, line 323: How is changing `$USERNAME` to `$USER` related? Also switching from `rustbuild` to `rust`.
The latter imo should be squashed into previous commit.
IMO also deleting `rust.sh` should be done in that same commit.

fviernau (changes requested):

Dockerfile
heliocastro: @fviernau You are correct over the USER part, the change is not suppose to be applied for this new entry.
The user change will happen on last commit after all sublayers are committed.
For the remove commit, please come to an agreement with @MarcelBochtler 
He complained to me that this changes need to be in a separated commit, so i honestly would keep in this way

heliocastro: @fviernau If you noticed that there are rustbuild and rust layers, and i prefer not sqaush any commit anymore
I will not start a who-is-right fight, so @MarcelBochtler came first, i did what they asked. 
I already did all the change requests on changelog from @tsteenbe, so if you have any technical issue, i can solve it, otherwise prefer stay the way it is.

fviernau: > For the remove commit, please come to an agreement with @MarcelBochtler
> He complained to me that this changes need to be in a separated commit, so i honestly would keep in this way

 I didn't review previous discussions and don't want to introduce back and force.

fviernau (approved):

PR #6091, opened 2022-11-17
NuGetSupport: Refactoring first couple of steps

fviernau:
The relationship between the the logic in NugetSupport and the package managers using it was partially of cyclic nature.
Also the code has some tight coupling making it overall relatively hard to work with. 
This PR removes these cyclic things and removes a few further hurdles in order to enable further refactoring and clean-up.
So, the goal is: enable further refactoring (with less complexity).

Breaking change: Because the issue message has been changed slightly. Resolution might need adjustments, see change in `dotnet-expected-output-with-nuspec.yml` for how to adjust the test.

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/dotnet-direct-dependencies-only-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/dotnet-direct-dependencies-only-expected-output.yml
@@ -171,7 +171,6 @@ packages:
 issues:
 - timestamp: "1970-01-01T00:00:00Z"
   source: "NuGet"
-  message: "Failed to get package data for 'NuGet::foobar:1.2.3': IOException: Failed\
-    \ to retrieve package data for 'foobar' from any of [https://api.nuget.org/v3/registration5-gz-semver2,\
+  message: "Failed to get package data for 'NuGet::foobar:1.2.3' from any of [https://api.nuget.org/v3/registration5-gz-semver2,\
     \ https://pkgs.dev.azure.com/ms/20953fb4-d427-4ad6-b78c-13142d0c5462/_packaging/64e3bf16-3e04-41cd-b90f-60fdc9ccf494/nuget/v3/registrations2-semver2]."
   severity: "ERROR"
--- a/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
@@ -250,7 +250,7 @@ class NuGetSupport(
                 val packageReferences = sortedSetOf<PackageReference>()
 
                 buildDependencyTree(
-                    references = allDependencies.map { Identifier("NuGet::${it.name}:${it.version}") },
+                    references = allDependencies.map { it.getId() },
                     dependencies = packageReferences,
                     packages = packages,
                     issues = issues,
@@ -342,6 +342,8 @@ private fun resolveLocalSpec(definitionFile: File): File? =
 
 private fun getIdentifier(name: String, version: String) =
     Identifier(type = "NuGet", namespace = "", name = name, version = version)
+
+private fun NuGetDependency.getId() = Identifier("NuGet::$name:$version")
 
 /**
  * A class that bundles properties of a single NuGet dependency.
--- a/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
@@ -70,9 +70,9 @@ private val VERSION_RANGE_CHARS = charArrayOf('[', ']', '(', ')', ',')
 private val JSON_MAPPER = JsonMapper().registerKotlinModule()
 
 class NuGetSupport(
-    val managerName: String,
-    val analysisRoot: File,
-    val reader: XmlPackageFileReader,
+    private val managerName: String,
+    private val analysisRoot: File,
+    private val reader: XmlPackageFileReader,
     definitionFile: File
 ) {
     companion object : Logging {
@@ -184,7 +184,7 @@ class NuGetSupport(
         }
     }
 
-    fun buildDependencyTree(
+    private fun buildDependencyTree(
         references: Collection<Identifier>,
         dependencies: MutableCollection<PackageReference>,
         packages: MutableCollection<Package>,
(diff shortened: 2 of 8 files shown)


sschuberth:
> the logic in NugetSupport

Nit: Please consistently spell it like "NuGetSupport" in commit messages.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -111,18 +104,23 @@ class NuGetSupport(serviceIndexUrls: List<String> = listOf(DEFAULT_SERVICE_INDEX
...
+        }
 
-    // Note: Remove a trailing slash as one is always added later to separate from the path, and a double-slash would
-    // break the URL!
-    private val registrationsBaseUrls = serviceIndices
-        .flatMap { it.resources }
-        .filter { it.type == REGISTRATIONS_BASE_URL_TYPE }
-        .map { it.id.removeSuffix("/") }
+        // Note: Remove a trailing slash as one is always added later to separate from the path, and a double-slash
+        // would break the URL!
```
sschuberth: Nit: I know the comment has been at this place before, but now that you're moving this code anyway, how about also moving this comment just before the belonging line 122?

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -111,18 +104,23 @@ class NuGetSupport(serviceIndexUrls: List<String> = listOf(DEFAULT_SERVICE_INDEX
         }
     }
 
-    private val serviceIndices = runBlocking {
-        serviceIndexUrls.map {
-            async { JSON_MAPPER.readValueFromUrl<ServiceIndex>(it) }
-        }.awaitAll()
-    }
+    private val registrationsBaseUrls: List<String> = run {
```
sschuberth: Maybe make this `by lazy {}` now?

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -109,23 +109,7 @@ class NuGetSupport(
...
-        }
-
-        // Note: Remove a trailing slash as one is always added later to separate from the path, and a double-slash
-        // would break the URL!
-        serviceIndices
-            .flatMap { it.resources }
-            .filter { it.type == REGISTRATIONS_BASE_URL_TYPE }
-            .map { it.id.removeSuffix("/") }
-    }
+    private val registrationsBaseUrls: List<String> = getRegistrationBaseUrls(definitionFile)
```
sschuberth, line 111: IMO this would have been a fitting change already in the first commit, and could be squashed.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -343,6 +343,8 @@ private fun resolveLocalSpec(definitionFile: File): File? =
 private fun getIdentifier(name: String, version: String) =
     Identifier(type = "NuGet", namespace = "", name = name, version = version)
 
+private fun NuGetDependency.getId() = Identifier("NuGet::$name:$version")
```
sschuberth, line 352: This should probably call the above `getIdentifier()` to not hard-code "NuGet" more than needed.

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: Done in a commit on top.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: This is out of scope and won't be needed anyway, because it will become lazy even without the `lazy` operator in a following change.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: Done in a separate commit on top.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: I prefer to not squash it, as perfect order is not necessary. This is the order things popped up in the refactoring. IMO it's fine to keep that (mikado method) ordering.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
sschuberth: > This is out of scope

You're now moving more (IO-related) code here, making it potentially more expensive to run, so I don't see how in turn making it lazy would be out of scope, just because it slightly changes the behavior of the property.

In general I'm not very happy with this commit on its own as it mangles getting the base URLs with reading from them. But I get that there's not much point in debating about it, as you're changing the code later anyway.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -69,7 +69,12 @@ private val VERSION_RANGE_CHARS = charArrayOf('[', ']', '(', ')', ',')
 
 private val JSON_MAPPER = JsonMapper().registerKotlinModule()
 
-class NuGetSupport(definitionFile: File) {
+class NuGetSupport(
+    val managerName: String,
+    val analysisRoot: File,
+    val xmlPackageFileReader: XmlPackageFileReader,
```
sschuberth: Nit: Maybe just `reader` here, too?

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
sschuberth: Why only on top? It's really inconvenient to come across this again in incremental reviews. Do you want to bump your commit stats? 😉

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -343,7 +343,7 @@ private fun resolveLocalSpec(definitionFile: File): File? =
 private fun getIdentifier(name: String, version: String) =
     Identifier(type = "NuGet", namespace = "", name = name, version = version)
 
-private fun NuGetDependency.getId() = Identifier("NuGet::$name:$version")
+private fun NuGetDependency.getId() = getIdentifier(name, version)
```
sschuberth, line 351: It really makes no sense to keep this separate from the previous commit. Please squash it.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -198,17 +198,7 @@ class NuGetSupport(
...
-                            // https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution#lowest-applicable-version.
-                            val version = dependency.range.trim { it.isWhitespace() || it in VERSION_RANGE_CHARS }
-                                .split(',').first().trim()
-
-                            // TODO: Add support resolving to the highest version for floating versions, see
-                            //       https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution#floating-versions.
-
-                            getIdentifier(dependency.id, version)
-                        },
+                        referredDependencies.map { it.getId() },
```
sschuberth: Please explain in the commit message why this is done, i.e. what's the plan?

analyzer/src/funTest/assets/projects/synthetic/dotnet-direct-dependencies-only-expected-output.yml
```diff
@@ -171,7 +171,6 @@ packages:
 issues:
 - timestamp: "1970-01-01T00:00:00Z"
   source: "NuGet"
-  message: "Failed to get package data for 'NuGet::foobar:1.2.3': IOException: Failed\
-    \ to retrieve package data for 'foobar' from any of [https://api.nuget.org/v3/registration5-gz-semver2,\
+  message: "Failed to get package data for 'NuGet::foobar:1.2.3' from any of [https://api.nuget.org/v3/registration5-gz-semver2,\
```
sschuberth, line 174: Commit message: "It's become easier" -> "It has become easier"

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -113,32 +112,58 @@ class NuGetSupport(
...
         return readValue(text)
     }
 
-    private fun getAllPackageData(id: Identifier): AllPackageData {
-        // Note: The package name in the URL is case-sensitive and must be lower-case!
-        val lowerId = id.name.lowercase()
+    private fun getAllPackageData(
+        directDependencies: Collection<Identifier>
+    ): Pair<Map<Identifier, AllPackageData>, Set<OrtIssue>> {
+        val issues = mutableSetOf<OrtIssue>()
```
sschuberth, line 121: The below code seems to do significantly more than just to "Separate fetching data from mapping it". The more important change clearly is that now multiple ids are fetched at once. It's a bit inconvenient to review both changes in the same diff.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -113,32 +112,58 @@ class NuGetSupport(
...
+        val result = mutableMapOf<Identifier, AllPackageData>()
+        val queue = LinkedList(directDependencies)
+
+        while (queue.isNotEmpty()) {
```
sschuberth, line 125: @oheger-bosch can you do me a favor and review this code from a concurrency perspective?

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: > You're now moving more (IO-related) code here, making it potentially more expensive to run, 

To my understanding that's incorrect. Prior and after this change the exact same logic is executed during construction time.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: It just follows your rules to do one thing at a time. It's a very minor issue. In my view you are overreacting to this nit. I prefer to keep it still.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: It really makes no sense to discuss this any further. See may other comment. Let's just disagree and get on with more important stuff.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: Resolving this as this seems to be just a general remark.

MarcelBochtler (approved):

Issue #5782, opened 2022-09-12
Stuck at JNI boundary?


DaveJarvis:
# Background

A Java application has ties to a C/C++ program via the JNI.

# Problem

ORT did not complete processing the build.gradle file, yet gave no indication as to an error.

      ~/dev/ort/bin/ort --info analyze -f JSON -i . -o ./ort/analyzer
      12:34:06.484 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.analyzer.PackageManager - Resolving Gradle dependencies for path 'conformance/build.gradle'...

# Expected

ORT shows an error to indicate that it encountered an issue resolving licensing information or the hanging loop prints a progress report to the user to indicate what's happening.

# Actual

ORT hangs. After 30 minutes, no progress made, I stopped ORT.

# Work around

1. Comment out the references to JNI projects in `conformance/build.gradle`.
1. Re-run ORT.

This allowed ORT to proceed without hanging. (At least until the troublesome JNI project was referenced from another project, which resulted in another hang, but it appears to be the same issue.)

sschuberth:
> 1. Comment out the references to JNI projects in `conformance/build.gradle`.

Are you able to share the source code to the project?

Also, do you get more details when running with `--debug`?

DaveJarvis:
> Are you able to share the source code to the project?

Proprietary, sorry. I'll see about running with `--debug` (aside, it'd be nice to add that option to the README.md file in the debugging section).

DaveJarvis:
Ran using:

    ~/dev/ort/bin/ort --debug --info analyze -f JSON -i . -o ./ort/analyzer

No additional processing information was written to the console.

Perhaps the first task in tracking this down would be to add more logging into the system around the resolver?

    13:13:20.211 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.analyzer.PackageManager - Resolving Gradle dependencies for path 'project.name'...

sschuberth:
> No additional processing information was written to the console.

That's probably because you use both `--debug` and `--info`, and the latter overrides the former. Please use `--debug` *instead* of `--info` (or put `--debug` last).

DaveJarvis:
Hangs here:

    14:32:37.186 [DefaultDispatcher-worker-1] DEBUG org.gradle.tooling.GradleConnector - Connecting from tooling API consumer version 7.5.1

oooooo99999:
[...]

Issue #5622, opened 2022-07-28
ORT hanging on analyze for multi-project Android app


pirvudoru:
### Description

I am trying to extract the list of dependencies of a multi-project, multi-flavor Android application using ORT.
What I am experiencing is that ort, once it reaches a certain project, it hangs. 

### Setup
Built ORT docker image at https://github.com/oss-review-toolkit/ort/commit/c2b17a999520808bb9d9dde3a51f28f66507999a
Started the docker image with --cpus="3" to give it 3 cpus
export JAVA_OPTS="-Xmx12288M" to give java room to grow up to 12 gb of memory

Example in the image:
![image](https://user-images.githubusercontent.com/2707601/181501622-fba0421c-e151-4a01-b954-c14c374e0bc7.png)

Reference to sample app to reproduce the same "hanging" behavior:
https://github.com/pirvudoru/SampleArchApp

Is there anything that can be done so that we get the analyzer-result for this type of project?

sschuberth:
Might be related to https://github.com/oss-review-toolkit/ort/issues/4694.

pirvudoru:
Thanks @sschuberth. Is there any way to work around this issue for the moment?

sschuberth:
@pirvudoru there's a chance that https://github.com/oss-review-toolkit/ort/pull/6276 also helps here, would you mind giving it a try?

sschuberth:
I've verified that with current ORT the analyzer is not hanging when scanning https://github.com/pirvudoru/SampleArchApp.

PR #6092, opened 2022-11-17
init.gradle: Ignore BOM dependencies when comparing dependency trees

nnobelis:
When comparing dependency trees, the count of sub dependencies should not include BOM dependencies, since they are skipped by the `fetchDependency` function.
This solves a bug where an Android project shipping 'androidx.appcompat:appcompat:1.5.1' would hang indefinitely. This is because, for instance, 'kotlinx-coroutines-android:1.6.1' depends on 'kotlinx-coroutines-core:1.6.1, kotlin-stdlib-jdk8:1.6.0 and kotlinx-coroutines-bom:1.6.1'. Since no ORT dependency is created for 'kotlinx-coroutines-bom' because it's a BOM file, it should also not be taken in account when comparing dependency trees. Otherwise, comparison would always fail, leading to a cache miss and overgrowth of the latter.

### DIFF
--- a/analyzer/src/main/resources/scripts/init.gradle
+++ b/analyzer/src/main/resources/scripts/init.gradle
@@ -251,7 +251,7 @@ class AbstractDependencyTreePlugin<T> implements Plugin<T> {
         /**
          * Return true if the provided DependencyResult represents an imported Maven BOM.
          */
-        private boolean isBom(DependencyResult dependencyResult) {
+        private static boolean isBom(DependencyResult dependencyResult) {
             if (dependencyResult instanceof ResolvedDependencyResult &&
                     dependencyResult.selected.hasProperty('variants')) {
                 return dependencyResult.selected.variants.every {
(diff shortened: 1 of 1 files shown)


mnonnenmacher (approved):

sschuberth:
@mpbecker and @pirvudoru, could help confirming whether this fixes your issues described [here](https://github.com/oss-review-toolkit/ort/issues/5782#issuecomment-1282130499) and [here](https://github.com/oss-review-toolkit/ort/issues/5622), respectively?

analyzer/src/main/resources/scripts/init.gradle
```diff
@@ -251,7 +251,7 @@ class AbstractDependencyTreePlugin<T> implements Plugin<T> {
         /**
          * Return true if the provided DependencyResult represents an imported Maven BOM.
          */
-        private boolean isBom(DependencyResult dependencyResult) {
+        private static boolean isBom(DependencyResult dependencyResult) {
```
sschuberth, line 254: Nit: Unrelated change.

sschuberth: Ah, sorry, expanding the diffs shows that it's not unrelated as `isBom` is called from the static `dependencyTreeEquals` now.

analyzer/src/main/resources/scripts/init.gradle
```diff
@@ -394,7 +394,8 @@ class AbstractDependencyTreePlugin<T> implements Plugin<T> {
 
             def resultDependencies = dependencyResult instanceof ResolvedDependencyResult ?
                     dependencyResult.selected.dependencies : [] as Set<DependencyResult>
-            if (dependency.dependencies.size() != resultDependencies.size()) {
+            // Ignore the BOM dependencies for calculating the count since they are skipped by `fetchDependency`.
+            if (dependency.dependencies.size() != resultDependencies.findAll { !isBom(it) }.size()) {
```
sschuberth, line 398: Nit: Could we use `count { !isBom(it) }` here?

sschuberth (approved):

PR #6096, opened 2022-11-21
Upgrade Detekt to version 1.22

sschuberth:
The `ComplexMethod` rule was renamed to `CyclomaticComplexMethod`, and revisiting the `threshold` revealed that it can be lowered compared to before (but still needs to be higher than the default of 15).

New rules demand annotated declarations to be on separate lines and to be separated by blank lines, like

    @SerialName("npm")
    NPM,

    @SerialName("crate")
    CRATE,

While this notation makes sense if the declarations are documented (and thus already separated by blank lines), for undocumented declarations (esp. as part of `@Serializable` classes) a more compact notation makes sense:

    @SerialName("package_name") val packageName: String,
    @SerialName("installed_version") val installedVersion: String,

Disable rules that prevent this notation, as it is already used e.g. in the `PythonInspector` class. Also, in general disable rules that would require too many code changes at this point.

Otherwise address the code inspection hints.

For the release notes see [1].

[1]: https://github.com/detekt/detekt/releases/tag/v1.22.0

### DIFF
--- a/.detekt.yml
+++ b/.detekt.yml
@@ -3,8 +3,8 @@
 # https://github.com/arturbosch/detekt/blob/master/detekt-cli/src/main/resources/default-detekt-config.yml.
 
 complexity:
-  ComplexMethod:
-    threshold: 33
+  CyclomaticComplexMethod:
+    threshold: 25
   LongMethod:
     threshold: 158
   LongParameterList:
@@ -23,6 +23,10 @@ coroutines:
 # Formatting rules are implemented via the ktlint plugin. As ktlint does not allow exceptions, we need to disable
 # respective rules completely.
 formatting:
+  AnnotationOnSeparateLine:
+    active: false
+  ArgumentListWrapping:
+    active: false
   ChainWrapping:
     active: false
   FinalNewline:
@@ -38,6 +42,8 @@ formatting:
   NoWildcardImports:
     active: false
   ParameterListWrapping:
+    active: false
+  SpacingBetweenDeclarationsWithAnnotations:
     active: false
 
 naming:
--- a/clients/clearly-defined/src/main/kotlin/Enums.kt
+++ b/clients/clearly-defined/src/main/kotlin/Enums.kt
@@ -29,32 +29,19 @@ import kotlinx.serialization.json.jsonPrimitive
  */
 @Serializable
 enum class ComponentType {
-    @SerialName("npm")
-    NPM,
-    @SerialName("crate")
-    CRATE,
-    @SerialName("git")
-    GIT,
-    @SerialName("maven")
-    MAVEN,
-    @SerialName("composer")
-    COMPOSER,
-    @SerialName("nuget")
-    NUGET,
-    @SerialName("gem")
-    GEM,
-    @SerialName("go")
-    GO,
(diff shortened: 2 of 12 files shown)


clients/fossid-webapp/src/main/kotlin/model/status/DownloadStatus.kt
```diff
@@ -24,8 +24,10 @@ import com.fasterxml.jackson.annotation.JsonProperty
 enum class DownloadStatus {
     FAILED,
     FINISHED,
+
```
fviernau, line 27: The annotation below is also for serialization and the enum value is undocumented, but the annotation remains on a separate line. This seems to be inconsistent. Is there any hidden rule there I'm missing?

clients/clearly-defined/src/main/kotlin/Enums.kt
```diff
@@ -113,27 +87,20 @@ enum class Provider {
...
-    MISSING,
-    @SerialName("Incorrect")
-    INCORRECT,
-    @SerialName("Incomplete")
-    INCOMPLETE,
-    @SerialName("Ambiguous")
-    AMBIGUOUS,
-    @SerialName("Other")
-    OTHER
+    @SerialName("Missing") MISSING,
```
fviernau, line 99: Not sure if we agreed already on having these on a single line. If so, please just ignore this comment:

The approach seems to work as long as there are only 1 annotation per value / property. As soon as one needs to add a second annotation, it will become ugly. Would we but 2,3,4 annotations on a single line then? How would it look then?

.detekt.yml
```diff
@@ -3,8 +3,8 @@
 # https://github.com/arturbosch/detekt/blob/master/detekt-cli/src/main/resources/default-detekt-config.yml.
 
 complexity:
-  ComplexMethod:
-    threshold: 33
+  CyclomaticComplexMethod:
```
fviernau, line 6: Please explain why for undocumented declaration the single line approach makes more sense?

I don't think this is obvous, in particular in cases where a second annotation is added to a declaration or in case declaration pass constructor args where the call takes multiple lines.

sschuberth: > Please explain why for undocumented declaration the single line approach makes more sense?

Undocumented field declarations in our code base are often not separated by blank lines, i.e. they come in the form

    @SerialName("npm")
    NPM,
    @SerialName("crate")
    CRATE,
    @SerialName("git")
    GIT,
    @SerialName("maven")
    MAVEN,
    ...

Instead of blowing up the code length by adding blank lines like in

    @SerialName("npm")
    NPM,

    @SerialName("crate")
    CRATE,

    @SerialName("git")
    GIT,

    @SerialName("maven")
    MAVEN,

    ...

I decided to follow @mnonnenmacher's preferred style (introduced with `PythonInspector`) to have the annotation and declaration compact in a single line:

    @SerialName("npm") NPM,
    @SerialName("crate") CRATE,
    @SerialName("git") GIT,
    @SerialName("maven") MAVEN,
    ...

> If decision for single line approach has been made already, please feel free to ignore this comment

I don't think a decision has been made yet, but IMO the single-line style should at least be allowed.

PR #6097, opened 2022-11-22
LicenseClassification: Ensure `licensesByCategory` contains all keys

fviernau:
Make the `Map` contain all category names as keys in order to simplify
its use.

### DIFF
--- a/model/src/main/kotlin/licenses/LicenseClassifications.kt
+++ b/model/src/main/kotlin/licenses/LicenseClassifications.kt
@@ -56,6 +56,10 @@ data class LicenseClassifications(
                     getOrPut(category) { mutableSetOf() } += license.id
                 }
             }
+
+            categories.forEach { category ->
+                putIfAbsent(category.name, mutableSetOf())
+            }
         }
     }
 
--- a/model/src/test/kotlin/licenses/LicenseClassificationsTest.kt
+++ b/model/src/test/kotlin/licenses/LicenseClassificationsTest.kt
@@ -22,6 +22,7 @@ package org.ossreviewtoolkit.model.licenses
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.collections.containExactlyInAnyOrder
+import io.kotest.matchers.collections.shouldBeEmpty
 import io.kotest.matchers.nulls.beNull
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
@@ -110,6 +111,14 @@ class LicenseClassificationsTest : WordSpec({
             permissiveLicenses shouldNotBeNull {
                 permissiveLicenses should containExactlyInAnyOrder(lic1.id, lic2.id)
             }
+        }
+
+        "maintain empty categories" {
+            val licenseClassifications = LicenseClassifications(
+                categories = listOf(LicenseCategory("permissive"))
+            )
+
+            licenseClassifications.licensesByCategory["permissive"] shouldNotBeNull { shouldBeEmpty() }
         }
 
         "return null when querying the licenses for an unknown category" {


model/src/main/kotlin/licenses/LicenseClassifications.kt
```diff
@@ -56,6 +56,10 @@ data class LicenseClassifications(
                     getOrPut(category) { mutableSetOf() } += license.id
                 }
             }
+
+            categories.forEach { category ->
```
sschuberth, line 60: > Make the `Map` contain all category names as keys in order to simplify its use.

I guess the more important thing here is to be able to distinguish map look-ups with a non-existing category as the key vs. an existing category that has no licenses assigned.

Here's a more compact implementation suggestion:

    val licensesByCategory: Map<String, Set<SpdxSingleLicenseExpression>> by lazy {
        buildMap<String, MutableSet<SpdxSingleLicenseExpression>> {
            categorizations.forEach { license ->
                categoryNames.forEach { category ->
                    val licenses = getOrPut(category) { mutableSetOf() }
                    if (category in license.categories) licenses += license.id
                }
            }
        }
    }

sschuberth: Also, please add a test-case in `LicenseClassificationsTest`'s `"licensesByCategory" should` test case that highlights that for existing but empty categorizations now an empty set instead of `null` is returned (above `"return null when querying the licenses for an unknown category")`.

sschuberth (changes requested):

model/src/main/kotlin/licenses/LicenseClassifications.kt
fviernau: > I guess the more important thing here is to be able to distinguish map look-ups with a non-existing category as the key vs. an existing category that has no licenses assigned.

Is this a general remark, or are you saying the commit message is not detailed enough and must be changed?

sschuberth: It's a general remark. But personally, I would have mentioned (only) that in the commit message. IMO it's clearer than saying "to simplify its use" without an example of how the use would have looked before and after the change.

TL;DR no strict need to change the commit message, just a suggestion.

model/src/test/kotlin/licenses/LicenseClassificationsTest.kt
```diff
@@ -63,7 +64,7 @@ class LicenseClassificationsTest : WordSpec({
             exception.message shouldContain "[ASL-1]"
         }
 
-        "detect licenses referencing non-existing categories" {
+            "detect licenses referencing non-existing categories" {
```
sschuberth, line 67: Indentation was changed.

model/src/test/kotlin/licenses/LicenseClassificationsTest.kt
```diff
@@ -112,6 +113,14 @@ class LicenseClassificationsTest : WordSpec({
             }
         }
 
+        "contain empty categories" {
```
sschuberth, line 116: Better "maintain" instead of "contain", as the latter sounds like "any arbitrary empty categories".

model/src/main/kotlin/licenses/LicenseClassifications.kt
```diff
@@ -56,6 +56,10 @@ data class LicenseClassifications(
...
+                putIfAbsent(category.name, mutableSetOf())
+            }
```
sschuberth, line 62: Just double-checking that this is no oversight as you did not comment on my proposed code in any way: It's intended that you did not change this, correct? I'm really just asking, without implying that I'd prefer that you should have changed it.

sschuberth (changes requested):

model/src/main/kotlin/licenses/LicenseClassifications.kt
fviernau: Yes, I intentionally haven't changed that.

model/src/main/kotlin/licenses/LicenseClassifications.kt
fviernau: ok.

sschuberth (approved):

PR #6105, opened 2022-11-23
Upgrade ScanCode to the latest version 31.2.1

sschuberth:


scanner/src/funTest/kotlin/scanners/scancode/ScanCodeScannerFunTest.kt
```diff
@@ -36,13 +36,15 @@ class ScanCodeScannerFunTest : AbstractPathScannerWrapperFunTest(setOf(Expensive
     override val scanner = ScanCode("ScanCode", scannerConfig)
 
     override val expectedFileLicenses = listOf(
-        LicenseFinding("Apache-2.0", TextLocation("LICENSE", 1, 201), 99.94f)
+        LicenseFinding("Apache-2.0", TextLocation("LICENSE", 1, 187), 100.0f),
+        LicenseFinding("Apache-2.0", TextLocation("LICENSE", 191, 201), 100.0f)
```
sschuberth, line 40: Note, at first I though these required changes might render existing package configurations with license finding curations invalid, but luckily these only depend on the start line, not the end line. So only additional license finding curations might be needed.

Dockerfile
```diff
@@ -140,17 +140,15 @@ ARG PYTHON_INSPECTOR_VERSION=0.9.2
 ARG PYTHON_PIPENV_VERSION=2022.9.24
 ARG PYTHON_POETRY_VERSION=1.2.2
 ARG PIPTOOL_VERSION=22.2.2
-ARG SCANCODE_VERSION=30.1.0
+ARG SCANCODE_VERSION=31.2.1
```
mnonnenmacher, line 143: If you want to continue using the old version some way to force the commoncode version is required, e.g. as an additional argument here which is ignored if empty.

mnonnenmacher (changes requested):
I'm surprised, what happened to the announced breaking changes in the output format?

model/src/main/resources/reference.yml
```diff
@@ -152,7 +152,7 @@ ort:
         # Criteria for matching stored scan results. These can be configured for any scanner that uses semantic
         # versioning. Note that the 'maxVersion' is exclusive and not part of the range of accepted versions.
         minVersion: '3.2.1-rc2'
-        maxVersion: '30.2.0'
+        maxVersion: '32.0.0'
```
mnonnenmacher, line 155: Are you sure 32.0.0 will work without additional changes?

mnonnenmacher (changes requested):

model/src/main/resources/reference.yml
sschuberth: No :grin: But `maxVersion` is *exclusive* anyway.

Dockerfile
sschuberth: So should we go for the `requirements.txt` approach here, too?

sschuberth:
> I'm surprised, what happened to the announced breaking changes in the output format?

Apparently only new stuff was added, or stuff that we don't read broke... maybe @pombredanne can comment on the changes between output format version 1.0.0 and 2.0.0? Are they really as big as the version bump indicates?

Dockerfile
mnonnenmacher: Yes, or maybe install the tar.gz in case there could be version conflicts with other tools installed via Pip?

model/src/main/resources/reference.yml
mnonnenmacher: Right, as the comment above says... Still unintuitive.

Dockerfile
sschuberth: > or maybe install the tar.gz

Yes, that's documented as the [recommended method](https://github.com/nexB/scancode-toolkit#installation).

sschuberth: However, I currently don't have time to continue working on this.

sschuberth: > So should we go for the `requirements.txt` approach here, too?

In view of the [problems I'm facing with installing ScanCode from binaries](https://github.com/oss-review-toolkit/ort/pull/6106#discussion_r1035739831), I went for the proven `requirements.txt` approach (that is successfully used in the "legacy" `Dockerfile`) also in the root `Dockerfile`.

fviernau (approved):

tsteenbe (approved):

mnonnenmacher (approved):

PR #6106, opened 2022-11-23
ScannerWrapper: Inline `ScannerDetails` to deduplicate the `name`

sschuberth:
Having `ScannerDetails` as part of `ScannerWrapper` creates a
recursion pitfall for bootstrappable `CommandLinePathScannerWrapper`s as
they must not access the `ScannerDetails` to get the `name` during
bootstrapping as that would also trigger getting the `version` which is
not known before bootstrapping succeeded.

To avoid running into that pitfall, inline the `version` and
`configuration` properties, and only create `ScannerDetails` when
needed.

### DIFF
--- a/scanner/src/main/kotlin/ScanController.kt
+++ b/scanner/src/main/kotlin/ScanController.kt
@@ -29,6 +29,7 @@ import org.ossreviewtoolkit.model.Provenance
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
 import org.ossreviewtoolkit.model.ScanSummary
+import org.ossreviewtoolkit.model.ScannerDetails
 import org.ossreviewtoolkit.model.UnknownProvenance
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.scanner.provenance.NestedProvenance
@@ -249,7 +250,7 @@ class ScanController(
             scanners.map { scanner ->
                 ScanResult(
                     provenance = UnknownProvenance,
-                    scanner = scanner.details,
+                    scanner = ScannerDetails(scanner.name, scanner.version, scanner.configuration),
                     summary = ScanSummary(
                         startTime = Instant.now(),
                         endTime = Instant.now(),
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -188,14 +188,14 @@ class ScannerCommand : CliktCommand(name = "scan", help = "Run external license 
 
         if (projectScannerWrappers.isNotEmpty()) {
             println("Scanning projects with:")
-            println(projectScannerWrappers.joinToString { "\t${it.name} (version ${it.details.version})" })
+            println(projectScannerWrappers.joinToString { "\t${it.name} (version ${it.version})" })
         } else {
             println("Projects will not be scanned.")
         }
 
         if (packageScannerWrappers.isNotEmpty()) {
             println("Scanning packages with:")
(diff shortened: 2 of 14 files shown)


fviernau (approved):
LGTM. Disabled auto-merge, so that it is possible to give @mnonnenmacher a chance to have a look.

sschuberth:
> LGTM. Disabled auto-merge, so that it is possible to give @mnonnenmacher a chance to have a look.

I'm running into a recursive problem and need to reword this anyway...

sschuberth:
> I'm running into a recursive problem and need to rework this anyway...

I'm now solving the problem now in basically the opposite way: Instead of keeping only `ScannerDetails`, I'm inlining it into `ScannerWrapper`. I'm curious about @mnonnenmacher's opinion.

scanner/src/main/kotlin/ScanController.kt
```diff
@@ -249,7 +250,7 @@ class ScanController(
             scanners.map { scanner ->
                 ScanResult(
                     provenance = UnknownProvenance,
-                    scanner = scanner.details,
+                    scanner = ScannerDetails(scanner.name, scanner.version, scanner.configuration),
```
mnonnenmacher, line 253: That seems very inconvenient to repeat the creation of `ScannerDetails` in so many places. Can you add either a lazy `details` property to `ScannerWrapper` or a `getDetails()` function?

mnonnenmacher (changes requested):

scanner/src/main/kotlin/ScanController.kt
sschuberth: The point was exactly to get rid of a `ScannerDetails` property / function in any of the base classes / interfaces of `CommandLinePathScannerWrapper` to avoid the potential misuse. I don't really see a better and safe way other than repeating this. But I'm absolutely open to suggestions.

sschuberth: > it's easy to avoid calling it there.

Well, at least I fell into this pitfall about 3 times over the last years...

> It makes sense to me to have name, version and configuration as separate properties, but not having a convenient way to create the details object does not.

Or should we still try to pursue the original idea of this PR to keep `ScannerDetails` in `ScannerWrapper` (and remove `name`from there), plus find a way to forbid use of `ScannerDetails` in `CommandLinePathScannerWrapper`, like by making it private in the base class?

mnonnenmacher: But I thought the bootstrapping code needs access to the name so it cannot be part of `ScannerDetails`?

sschuberth: > But I thought the bootstrapping code needs access to the name so it cannot be part of `ScannerDetails`?

Yes, sorry, my proposal didn't make sense.

> Also, didn't we plan to remove bootstrapping anyway?

We discussed this at some point, yes, but [in the context of installing ScanCode from binaries](https://github.com/oss-review-toolkit/ort/pull/6105#discussion_r1030833446) I was thinking about using ORT's existing bootstrapping mechanism to also install scanners as part of a Docker build. That's something I'm playing with right now.

sschuberth: Ok, I'm giving up on installing ScanCode from binaries as there are issues:

* https://github.com/nexB/scancode-toolkit/issues/2836
* https://github.com/nexB/scancode-toolkit/issues/2731
* https://github.com/nexB/scancode-toolkit/issues/2730
* https://github.com/nexB/scancode-toolkit/issues/769

Looks like indeed simply dropping the scanner bootstrapping feature is the way to go.

sschuberth:
> Also, didn't we plan to remove bootstrapping anyway?

I'm now indeed dropping this PR in favor of another one that will remove bootstrapping of scanners as [agreed](https://github.com/oss-review-toolkit/ort/wiki/Community-Meeting#2022-12-01).

PR #6109, opened 2022-11-24
Accept an uppercase "(C)" as part of the Copyright prefix

sschuberth:

### DIFF
--- a/utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
+++ b/utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
@@ -34,20 +34,20 @@ private const val YEAR_PLACEHOLDER = "<ORT_YEAR_PLACEHOLDER_TRO>"
 private val COMMA_SEPARATED_YEARS_REGEX = "(?=.*)\\b(\\d{4})\\b( *, *)\\b(\\d{4})\\b".toRegex()
 
 private val KNOWN_PREFIX_REGEX = listOf(
-    "^(?:\\(c\\))",
-    "^(?:\\(c\\) [C|c]opyright)",
-    "^(?:\\(c\\) [C|c]opyrighted)",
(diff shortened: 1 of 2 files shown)


model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
```diff
@@ -0,0 +1,60 @@
...
+package org.ossreviewtoolkit.model.licenses
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.collections.shouldHaveSize
+import io.kotest.matchers.shouldBe
+
+import org.ossreviewtoolkit.model.TextLocation
+
+class ResolvedLicenseTest : StringSpec({
+    "Copyright findings should be grouped across different files and maintained in the result" {
```
fviernau, line 29: This test belongs to `CopyrightStatementsProcessor` and should be implemented in its dedicated test file.

model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
```diff
@@ -50,8 +50,7 @@ class ResolvedLicenseTest : StringSpec({
 
         resolvedCopyrights shouldHaveSize 1
         with(resolvedCopyrights.first()) {
-            // TODO: The "(C)" should come directly after "Copyright".
-            statement shouldBe "Copyright 2019, 2022 (C) The ORT Project Authors"
+            statement shouldBe "Copyright (C) 2019, 2022 The ORT Project Authors"
```
fviernau, line 53: Should this commit auto resolve the Github issue ?

sschuberth: No, the issue is actually about another problem that I *tried* to reproduce with this test, but instead discovered this problem.

fviernau (approved):

PR #6110, opened 2022-11-24
Prefer the more speaking `first()` over `[0]`

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -273,7 +273,7 @@ class GoMod(
 
         list.stdout.lines().forEach { line ->
             val columns = line.splitOnWhitespace()
-            if (columns.size in 1..2) result += columns[0]
+            if (columns.size in 1..2) result += columns.first()
         }
 
         return result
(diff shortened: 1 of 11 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -166,8 +166,8 @@ class GoMod(
             val columns = line.splitOnWhitespace()
             require(columns.size == 2) { "Expected exactly one occurrence of ' ' on any non-blank line." }
 
-            val parent = parseModuleEntry(columns[0])
-            val child = parseModuleEntry(columns[1])
+            val parent = parseModuleEntry(columns.first())
+            val child = parseModuleEntry(columns.last())
```
fviernau, line 170: I believe [1] is more speaking than `last` and more appropriate.
Due to that I believe [0] should also be kept for consistency.

fviernau (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
sschuberth: I've reverted that occurrence.

fviernau (approved):

PR #6104, opened 2022-11-23
ResolvedLicense: fix toResolvedCopyrights to identify the same copyri…

bennati:
…ght in multiple files.

Fix a regression when a copyright is present in multiple files in the same repository.
The expected behavior is that the all findings of the copyright are returned, if they have different paths.
The previous behavior was that only the last finding was kept, as 'statementToFinding' is built using 'associateBy' which "If any two elements would have the same key returned by keySelector the last one gets added to the map."

Signed-off-by: Stefano Bennati <stefano.bennati@here.com>

Please ensure that your pull request adheres to our [contribution guidelines](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md). Thank you!

PR #6113, opened 2022-11-24
ResolvedLicense: Fix a regression in `toResolvedCopyrights()`

sschuberth:
The `associateBy()` was unintentionally deduplicating original findings with the same statement. Fix that by using `groupBy()` (and `flatMapTo()`) instead. Also add a test for the wanted behavior. See [1] for the original report with another proposed fix; this fix is more minimal.

[1]: https://github.com/oss-review-toolkit/ort/pull/6104

### DIFF
--- a/model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
+++ b/model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
@@ -19,40 +19,72 @@
 
 package org.ossreviewtoolkit.model.licenses
 
-import io.kotest.core.spec.style.StringSpec
+import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.shouldBe
 
 import org.ossreviewtoolkit.model.TextLocation
 
-class ResolvedLicenseTest : StringSpec({
-    "Copyright findings with the same statement across different files should be listed in the result" {
-        val originalFindings = listOf(
-            ResolvedCopyrightFinding(
-                statement = "Copyright (C) 2022 The ORT Project Authors",
-                location = TextLocation(
-                    path = "/path/to/file/A",
-                    line = 2
+class ResolvedLicenseTest : WordSpec({
+    "Copyright findings across different files" should {
+        "be consolidated without processing if their statements are exactly the same" {
+            val originalFindings = listOf(
+                ResolvedCopyrightFinding(
+                    statement = "Copyright (C) 2022 The ORT Project Authors",
+                    location = TextLocation(
+                        path = "/path/to/file/A",
+                        line = 2
+                    ),
+                    matchingPathExcludes = emptyList()
                 ),
-                matchingPathExcludes = emptyList()
-            ),
-            ResolvedCopyrightFinding(
-                statement = "Copyright (C) 2022 The ORT Project Authors",
-                location = TextLocation(
-                    path = "/path/to/file/B",
-                    line = 2
+                ResolvedCopyrightFinding(
+                    statement = "Copyright (C) 2022 The ORT Project Authors",
+                    location = TextLocation(
+                        path = "/path/to/file/B",
+                        line = 2
+                    ),
+                    matchingPathExcludes = emptyList()
                 ),
-                matchingPathExcludes = emptyList()
-            ),
-        )
+            )
 
-        val resolvedCopyrights = originalFindings.toResolvedCopyrights(process = false)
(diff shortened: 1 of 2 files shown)


model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
```diff
@@ -0,0 +1,59 @@
...
+package org.ossreviewtoolkit.model.licenses
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.collections.shouldHaveSize
+import io.kotest.matchers.shouldBe
+
+import org.ossreviewtoolkit.model.TextLocation
+
+class ResolvedLicenseTest : StringSpec({
+    "Copyright findings with the same year across different files should be maintained in the result" {
```
fviernau, line 29: I wonder if "with same year" may be misleading. 
Would it make sense to say:  "copyright findings which get processed to identical statements should be maintained"?

model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+        val resolvedCopyrights = originalFindings.toResolvedCopyrights(process = true)
+
+        resolvedCopyrights shouldHaveSize 1
+        with(resolvedCopyrights.first()) {
+            statement shouldBe "Copyright (C) 2022 The ORT Project Authors"
+            findings shouldHaveSize 2
+            findings.first().location.path shouldBe "/path/to/file/A"
+            findings.last().location.path shouldBe "/path/to/file/B"
```
fviernau, line 56: consider:
```
findings.map { it.location.path } shouldContainExactlyInAnyOrder("/path/to/file/A", "/path/to/file/B")?
```

fviernau (approved):

model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
sschuberth: Actually, the bug of unintended deduplication happened *before* the processing already. Which made me think that the test should also pass with `process = false`, which it didn't, due to another regression 😢 I've also fixed that one in a new commit; I'm really sorry for having introduced these. On the good side, I've now added more tests to guard against such things in the future.

model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
```diff
@@ -19,41 +19,74 @@
...
+                    matchingPathExcludes = emptyList()
+                ),
+            )
+
+            val resolvedCopyrights = originalFindings.toResolvedCopyrights(process = true)
+
+            resolvedCopyrights shouldHaveSize 1
+            with(resolvedCopyrights.first()) {
+                statement shouldBe "Copyright (C) 2022 The ORT Project Authors"
+                findings shouldHaveSize 2
```
fviernau, line 87: The assertion in line 87 is redundant and should be dropped. See line 88.

Same in line 56.

model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
```diff
@@ -19,41 +19,74 @@
...
+                    ),
+                    matchingPathExcludes = emptyList()
                 ),
-                matchingPathExcludes = emptyList()
-            ),
-        )
+            )
 
-        val resolvedCopyrights = originalFindings.toResolvedCopyrights(process = false)
+            val resolvedCopyrights = originalFindings.toResolvedCopyrights(process = false)
```
fviernau, line 51: If I didn't miss anything these two test cases only differ in terms of the boolean value for `process`, e.g. `true` vs. `false`. While it make sense to test one thing at a time, I believe in this case it would make sense to merge these tests like: 
```
should "be consolidated with and without processing if their statements are exactly the same" {
...

listOf(true, false).forall { process ->
  // assert here
}
```
What do you think?

fviernau (changes requested):

model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
fviernau: (Same also in previous commit)

model/src/test/kotlin/licenses/ResolvedLicenseTest.kt
sschuberth: > If I didn't miss anything these two test cases only differ in terms of the boolean value for `process`

No, the input `statement` also differs see the `// Note the "." at the end.` comment. Thus I'd prefer to keep these test cases separate.

fviernau (approved):

PR #6124, opened 2022-11-28
Set the code point limit for SnakeYAML to 1GB

mnonnenmacher:
Since version 1.32 [1] SnakeYAML sets the limit for incoming data to 3MB. Since Jackson 2.14 [2] it is possible to override this default, so hardcode the limit to 1GB to allow parsing large ORT result files.

[1]: https://bitbucket.org/snakeyaml/snakeyaml/wiki/Changes
[2]: https://github.com/FasterXML/jackson-dataformats-text/issues/337


sschuberth:
Also, should we add a test that writes a 4 MiB dummy YAML file?

sschuberth:
Let's address my remarks manually / separately and merge this to fixup https://github.com/oss-review-toolkit/ort/pull/6123 for the changes.

sschuberth (approved):

PR #6134, opened 2022-11-28
github/workflows: Enable CodeQL analysis for Kotlin / Java

sschuberth:
See [1].

[1]: https://github.blog/changelog/2022-11-28-codeql-code-scanning-launches-kotlin-analysis-support-beta/

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,3 @@
+paths-ignore:
+  - '**/src/funTest/assets/**'
+  - '**/src/test/assets/**'
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -45,6 +45,30 @@ jobs:
       uses: gradle/gradle-build-action@v2
       with:
         arguments: --scan :reporter-web-app:yarnBuild
+  codeql-analysis:
+    needs: build
+    runs-on: ubuntu-22.04
+    permissions:
+      security-events: write
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v3
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: java
+        tools: latest
+    - name: Setup Java
+      uses: actions/setup-java@v3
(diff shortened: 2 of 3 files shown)


sschuberth:
Moving to draft state due to memory issues, also see https://github.com/github/codeql/discussions/11460.

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -133,6 +136,14 @@ fun File.unpack7Zip(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = {
 
             val target = targetDirectory.resolve(entry.name)
 
+            if (!target.canonicalFile.startsWith(targetDirectory)) {
```
Marcono1234, line 139: Would it make sense to compare this with `targetDirectory.canonicalFile` instead?
```suggestion
            if (!target.canonicalFile.startsWith(targetDirectory.canonicalFile)) {
```
(respectively evaluate `targetDirectory.canonicalFile` once outside the loop)

Otherwise I assume this check would fail if the `targetDirectory` value is not the canonical file path, even though the archive entry name is valid. Same applies to the other checks introduced by this PR as well.

(Sorry if this comment here is a bit intrusive)

sschuberth: > Would it make sense to compare this with `targetDirectory.canonicalFile` instead?

Yes, you're right. Fixed.

> Sorry if this comment here is a bit intrusive

No worries, and thanks for the review!

sschuberth:
Still blocked until [mid-January 2023](https://github.com/github/codeql/discussions/11460#discussioncomment-4419634).

sschuberth:
> Still blocked until [mid-January 2023](https://github.com/github/codeql/discussions/11460#discussioncomment-4419634).

Now blocked by a [lack of support for Kotlin 1.8](https://github.com/github/codeql/discussions/11460#discussioncomment-4734997).

sschuberth:
Ping @oss-review-toolkit/kotlin-devs 😉

mnonnenmacher (approved):

PR #6142, opened 2022-11-29
Introduce a Yarn info cache

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Yarn.kt
+++ b/analyzer/src/main/kotlin/managers/Yarn.kt
@@ -25,6 +25,8 @@ import com.vdurmont.semver4j.Requirement
 
 import java.io.File
 
+import kotlin.time.Duration.Companion.days
+
 import org.apache.logging.log4j.kotlin.Logging
 
 import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
@@ -33,7 +35,15 @@ import org.ossreviewtoolkit.analyzer.managers.utils.mapDefinitionFilesForYarn
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.jsonMapper
+import org.ossreviewtoolkit.utils.common.DiskCache
 import org.ossreviewtoolkit.utils.common.Os
+import org.ossreviewtoolkit.utils.ort.ortDataDirectory
+
+private val yarnInfoCache = DiskCache(
+    directory = ortDataDirectory.resolve("cache/analyzer/yarn/info"),
+    maxCacheSizeInBytes = 100L * 1024L * 1024L,
+    maxCacheEntryAgeInSeconds = 7.days.inWholeSeconds
+)
 
 /**
  * The [Yarn](https://classic.yarnpkg.com/) package manager for JavaScript.
@@ -72,9 +82,16 @@ class Yarn(
     override fun runInstall(workingDir: File) = run(workingDir, "install", "--ignore-scripts", "--ignore-engines")
 
     override fun getRemotePackageDetails(workingDir: File, packageName: String): JsonNode {
+        yarnInfoCache.read(packageName)?.let { return jsonMapper.readTree(it) }
+
         val process = run(workingDir, "info", "--json", packageName)
-        return jsonMapper.readTree(process.stdout)["data"] ?: jsonMapper.readTree(process.stderr)["data"].also {
-            logger.warn { "Error running '${process.commandLine}' in directory $workingDir: $it" }
+
+        return jsonMapper.readTree(process.stdout)["data"]?.also {
+            yarnInfoCache.write(packageName, it.toString())
+        } ?: run {
+            jsonMapper.readTree(process.stderr)["data"].also {
+                logger.warn { "Error running '${process.commandLine}' in directory $workingDir: $it" }
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -107,20 +109,17 @@ private val File?.safePath: String
...
 
-        private val remoteArtifactCache =
-            DiskCache(
-                ortDataDirectory.resolve("cache/remote_artifacts"),
-                MAX_DISK_CACHE_SIZE_IN_BYTES, MAX_DISK_CACHE_ENTRY_AGE_SECONDS
-            )
+        private val remoteArtifactCache = DiskCache(
+            directory = ortDataDirectory.resolve("cache/remote-artifacts"),
+            maxCacheSizeInBytes = 1024L * 1024L * 1024L,
+            maxCacheEntryAgeInSeconds = 6.hours.inWholeSeconds
```
sschuberth, line 121: BTW, I don't think that these historic number make sense still. IMO we should use a smaller cache with a larger max cache age.

Can anyone share their current cache sizes on disk?

sschuberth:
If this is fine with you guys, we could do something similar for `npm view` instead of the `ConcurrentHashMap`.

oheger-bosch:
Caching is certainly beneficial here; I have also been thinking about something similar for npm: here the `npm view` calls typically are responsible for the major part of the analysis time.

However, especially in the ORT server context, a disk cache may not be the ideal choice for caching such data; here a database would probably be preferable. ORT already has the concept of storages for scan results. I wonder whether this concept could be generalized and extended to a generic caching service, which could then be used by multiple components.

nnobelis:
Additionally, this improves only Yarn, not Yarn 2+. Therefore I would prefer something more generic as @oheger-bosch said.

sschuberth:
> I wonder whether this concept could be generalized and extended to a generic caching service, which could then be used by multiple components.

I like the idea, but that goes way beyond the scope of this PR as it involved major refactoring around a completely new component.

> Additionally, this improves only Yarn, not Yarn 2+.

We have to start somewhere, right? Regard this as a PoC implementation to remind ourselves of the general idea to use persistent caches here (and in the future in other places), not only in-memory caching via maps.

analyzer/src/main/kotlin/managers/Yarn.kt
```diff
@@ -33,7 +35,15 @@ import org.ossreviewtoolkit.analyzer.managers.utils.mapDefinitionFilesForYarn
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.jsonMapper
+import org.ossreviewtoolkit.utils.common.DiskCache
 import org.ossreviewtoolkit.utils.common.Os
+import org.ossreviewtoolkit.utils.ort.ortDataDirectory
+
+private val yarnInfoCache = DiskCache(
+    directory = ortDataDirectory.resolve("cache/yarn-info"),
```
fviernau, line 43: Should we use `cache/analyzer/yarn-info` instead?

Note: If I'd be debugging a random "analyzer reports wrong meta-data" bug report, I'd probably start with reproducing trying to delete the whole analyzer cache first. Also I'd probably ask the user if it reproduces with the cache deleted. That deletion step would become simpler then.

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Yarn.kt
sschuberth: > Should we use `cache/analyzer/yarn-info` instead?

I'd be fine with that, but I'd then like to move Maven's "cache/remote_artifacts" accordingly.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: @mnonnenmacher and @fviernau, my "cache/remote_artifacts" is just several hundred KB, what about yours?

analyzer/src/main/kotlin/managers/Yarn.kt
fviernau: Oh, wasn't aware of that one. Maybe then

`cache/analyzer/maven/remote-artifacts`
`cache/analyzer/yarn/yarn-info`

This allows (1) simple handling of all analyzer cache, e.g. delete or docker mount and (2) simple handling of all cache data for a specific package manager, e.g. debugging / deleting.

sschuberth: Also fine with me, although I probably would say just "cache/analyzer/yarn/info" then to not repeat "yarn".

sschuberth: I've made these changes, please re-review @oss-review-toolkit/kotlin-devs.

utils/common/src/main/kotlin/DiskCache.kt
```diff
@@ -45,7 +45,7 @@ class DiskCache(
     /**
      * Duration in seconds that cache entries are valid.
      */
-    private val maxCacheEntryAgeInSeconds: Int
+    private val maxCacheEntryAgeInSeconds: Long
```
fviernau, line 48: Int was 68 years was it?

fviernau (approved):

utils/common/src/main/kotlin/DiskCache.kt
sschuberth: I don't know, but the math using `maxCacheEntryAgeInSeconds` was using `Long` already (due to API requirements), so it anyway made sense to turn this into `Long`, too.

PR #6143, opened 2022-11-30
Jenkins pipeline improvements

sschuberth:


fviernau (approved):
Out of scope: As I see this screenshot, just sharing an idea:

{analyzer|scan|evaluator|advisor}-result.yml IMO should be turned into separate
{analyzer|scan|evaluator|advisor}-result.yml.xz files using `xz -d1`.

This (1) allows user to very quickly download the exact result file they need and (2) save a lot of space. 

Note: the other files wouldn't need compression as they are small anyway and can be viewed instantly.

Would it make sense to have a ticket to implement that?

fviernau:
> Actually, I like to be able to view these files quickly from within Jenkins. Compressing them would prevent that. So I'm not sure about implementing the feature in the pipeline definition. Interested users might be better off by using e.g. https://plugins.jenkins.io/compress-artifacts/.

Mind sharing why you believe that viewing the files in Jenkins UI is so important that showing things locally in your favorite editor should be slowed down?

> https://plugins.jenkins.io/compress-artifacts/.

Seems to only support zip and not sure if it supports compressing artifacts separately.

fviernau:
> What do you mean by "slowed down"? Or phrased differently, how does compressing these files speed up showing the files in your local editor? Sure, the download would be faster, but in exchange you have the additional (manual) step of uncompressing a file in order to display it in your editor.

Slower download, yes. (It can make a huge difference depending on the file size.)
The additional (manual) step needs not be manual, it can be automated easily.
I believe the ones which really do investigations of scan results want the file locally. Why
should this target folks be prioritized below the folks which only wants to quickly view files in Jenkins, but I presume not do any deeper investigations?

sschuberth:
> Why should this target folks be prioritized below the folks which only wants to quickly view files in Jenkins, but I presume not do any deeper investigations?

Because to me, the `Jenkinsfile` pipeline is a reference implementation to get an ORT pipeline setup quickly, to *learn* about ORT.

I believe it's a good starting point for a production-ready pipeline, but it anyway lacks several things (like a DB-based scan storage) in that regard. That's why for me, the ability to easily view ORT results and learn about ORT's inner workings goes above a highly tuned investigation / curation workflow here, if these two purposes conflict.

sschuberth:
PS: What would probably change my mind is, if we had a custom file extension for (compressed) ORT results so that these files could be associated with a viewer application (like the ORT workbench could become one). Then clicking on such a Jenkins artifact would directly open up the file in the dedicated local viewer application. Additionally, all ORT command should be changed to work with compressed ORT result files as input.

PR #6144, opened 2022-11-30
cli: Turn commands into plugins

sschuberth:
This allows to provide external subcommands to ORT at runtime.

### DIFF
--- a/cli/src/main/kotlin/OrtCommand.kt
+++ b/cli/src/main/kotlin/OrtCommand.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2022 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.cli
+
+import com.github.ajalt.clikt.core.CliktCommand
+
+import org.ossreviewtoolkit.utils.common.NamedPlugin
+
+/**
+ * An interface for [CliktCommand]-based ORT commands that come as named plugins.
+ */
+abstract class OrtCommand(name: String, help: String) : CliktCommand(name = name, help = help), NamedPlugin {
+    companion object {
+        /**
+         * All ORT commands available in the classpath, associated by their names.
+         */
+        val ALL by lazy { NamedPlugin.getAll<OrtCommand>() }
+    }
+
+    override val name = commandName
+}
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -151,20 +151,7 @@ class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
             helpFormatter = OrtHelpFormatter()
         }
 
-        subcommands(
-            AdvisorCommand(),
-            AnalyzerCommand(),
-            ConfigCommand(),
-            DownloaderCommand(),
-            EvaluatorCommand(),
-            NotifierCommand(),
-            ReporterCommand(),
-            RequirementsCommand(),
-            ScannerCommand(),
-            UploadCurationsCommand(),
-            UploadResultToPostgresCommand(),
-            UploadResultToSw360Command()
-        )
+        subcommands(OrtCommand.ALL.values)
 
         versionOption(
             version = env.ortVersion,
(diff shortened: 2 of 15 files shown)


cli/src/main/kotlin/OrtCommand.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    override val name: String
+        get() = commandName
+
+    /**
+     * The command name as provided by [CliktCommand].
+     */
+    val commandName: String
+}
```
fviernau, line 44: While trying to find a way to make this a bit more type-safe by inheriting CliktCommand,
I found that CliktCommand is an abstract class so IMO it would not be nice to inherit from it, but stick to an interface. Anyhow, I believe we can make clearer that this plugin is about CliktCommand by just adding a factory function to the plugin interface. This nicely separates plugin related things, avoids casts from the command, see the following:
```
interface OrtCommandPlugin : NamedPlugin {
     fun createCommand(): CliktCommand
     /* The name of the plugin, as per NamedPlugin (no override) **/
     val name: String
}
```

What do you think?

sschuberth: I dislike the overhead of factory functions, esp. as in this case they're not needed as the subcommands do not require constructor arguments and so `ServiceLoader` can instantiate them directly. But I'll probably play around how it makes the code look like.

sschuberth: Looks like I then would need to add wrappers to each command, like

```
class AdvisorCommandPlugin(override val name: String = "advise") : OrtCommandPlugin {
    override fun createCommand() = AdvisorCommand(name)
}

class AdvisorCommand(name: String) : CliktCommand(
    name = name,
    help = "Check dependencies for security vulnerabilities."
) {
```

and in OrtMain then say

```
subcommands(OrtCommandPlugin.ALL.values.map { it.createCommand() })
```

I'm not sure if I like that better. What about other @oss-review-toolkit/kotlin-devs?

fviernau: I'm not sure if the wrappers are needed like this. If you look at https://github.com/oss-review-toolkit/ort/blob/3d009a0af5a6400c91dc97c16f93ab72027d97d4/cli/src/main/kotlin/OrtMain.kt#L154-L167, you could just turn this into a list of lambdas:

```

val factories = listOf(
 { AdvisorCommand() },
 { AnalyzerCommand() },
....
).map { 
   object : OrtCommandPlugin {
       createCommand() = it()
   }
}
```

sschuberth: But hard-coding the classes like

```
val plugins = listOf(
 { AdvisorCommand() },
 { AnalyzerCommand() },
```

defeats the purpose of making these plugins, as we don't know the class names of externally added plugins, or?

oheger-bosch: To make this type-safe, how about making `OrtCommand` an abstract class that extends `CliktCommand` instead of an interface? You could then even drop the `commandName` property.

sschuberth: > To make this type-safe, how about making `OrtCommand` an abstract class that extends `CliktCommand` instead of an interface?

That's the approach I tried first, but it requires to repeat all `CliktCommand` constructor arguments in `OrtCommand`, which I dislike.

I have another idea on how to improve type safety, let me check something.

sschuberth: > I have another idea on how to improve type safety, let me check something.

My idea about using a special / generic extension-getter-function did not work out. But I've filed https://github.com/ajalt/clikt/issues/378 in between.

All in all, I still like the current implementation best (for its minimally invasive approach) when compared to the other proposals.

fviernau: > defeats the purpose of making these plugins, as we don't know the class names of externally added plugins, or?

Wasn't the purpose "enable loading external implementations", which would still work IIUC?

fviernau: > defeats the purpose of making these plugins, as we don't know the class names of externally added plugins, or?

Why does it defeat the purpose. E.g. why wouldn't the following work?

```
val plugins = listOf(
 { AdvisorCommand() },
 { AnalyzerCommand() },
....
).map { 
   object : OrtCommandPlugin {
       createCommand() = it()
   }
} + NamedPlugin.getAll<OrtCommandPlugin>()
```

sschuberth: That would probably technically work, but I believe it's not very elegant to treat built-in commands and external commands differently.

fviernau: Only instantiation would be different, afterwards they are handled via their interface. Anyhow, let's not discuss it further - it was just an idea / proposal. I don't want to force you to use it.

sschuberth: > > To make this type-safe, how about making `OrtCommand` an abstract class that extends `CliktCommand` instead of an interface?
> 
> That's the approach I tried first, but it requires to repeat all `CliktCommand` constructor arguments in `OrtCommand`, which I dislike.

@oheger-bosch, @fviernau, I've changed my mind and turned `OrtCommand` into an abstract class inheriting from `CliktCommand`. The code looks much worse than I initially though, as we only use very few of the many `CliktCommand` constructor arguments, so I also only need to forward very few of them.

oheger-bosch: Looks good to me. This has the additional benefit that the base class could be later extended by functionality common to typical commands, maybe such as reading ORT result files or stuff like that.

oheger-bosch (approved):

PR #6147, opened 2022-11-30
docker: Sanitize Golang in docker

heliocastro:
Docker go sanitize

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -229,7 +229,7 @@ COPY --from=rustbuild /opt/rust /opt/rust
 
 #------------------------------------------------------------------------
 # GOLANG - Build as a separate component
-FROM build AS gobuild
+FROM ort-base-image AS gobuild
 
 ARG GO_DEP_VERSION=0.5.4
 ARG GO_VERSION=1.18.3
--- a/Dockerfile
+++ b/Dockerfile
@@ -240,7 +240,6 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN arch=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/) \
     && curl -L https://dl.google.com/go/go${GO_VERSION}.linux-${arch}.tar.gz | tar -C /opt -xz \
     && curl -ksS https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | bash
-RUN echo "add_local_path /opt/go/bin:\$PATH" > /etc/profile.d/go.sh
 
 FROM scratch AS golang
 COPY --from=gobuild /opt/go /opt/go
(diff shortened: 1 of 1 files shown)


tsteenbe (commented):
....

tsteenbe (changes requested):
@heliocastro  I mostly see grammar issue in the commit message in this PR see my suggestions below. Please remove the merge commit _Merge branch 'main' into hcc/docker_go_sanitize_


_docker: Use ort-base-image instead of ort-base-build_ 

**current body:**
"Ort-base-build layer is not required anymore for golang layer." 
**proposed body:**
"The ort-base-build layer is no longer required by the golang layer."

_docker: Download and install golang with correct architecture_
**current body:**
Only amd64 and arm64 are supported as main ORT base Java
Temurin container supports only this two architectures.

Environment PATH is required to golang dep package manager installation.
**proposed body:**
Only amd64 and arm64 are supported as these are the architectures
supported by eclipse-temurin:11-jdk-jammy which is ORT base image.

Environment PATH is required to install the golang dep package manager.

_docker: Create golang binary install layer_
**current body**:
Improve docker cache efficiency having a single scratch layer with
only installation binaries.
**proposed body:**
Improve docker cache efficiency via a single scratch layer with
only installed binaries.

heliocastro:
> 

@tsteenbe  Done

tsteenbe (dismissed):

Dockerfile
```diff
@@ -224,11 +224,12 @@ FROM ort-base-image AS gobuild
 
 ARG GO_DEP_VERSION=0.5.4
 ARG GO_VERSION=1.18.3
-ENV GOPATH=/opt/go
-RUN curl -L https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar -C /opt -xz
-ENV PATH=/opt/go/bin:$PATH
-RUN go version
-RUN curl -ksS https://raw.githubusercontent.com/golang/dep/v$GO_DEP_VERSION/install.sh | bash
+ENV GOBIN=/opt/go/bin
```
fviernau, line 227: commit message: are supported as **these** are the architectures.

fviernau, line 227: Is it intended to remove `GOPATH=/opt/go`? (or replace it with `GOBIN=/opt/go/bin`)?
Why is that done?

fviernau (changes requested):

Dockerfile
```diff
@@ -323,8 +322,8 @@ COPY docker/rust.sh /etc/profile.d/
 RUN chmod o+rwx ${CARGO_HOME}
 
 # Golang
-COPY --chown=$USERNAME:$USERNAME --from=gobuild /opt/go /opt/go/
-COPY --from=gobuild  /etc/profile.d/go.sh /etc/profile.d/
+ENV PATH=$PATH:/opt/go/bin
+COPY --from=golang /opt/go /opt/go
```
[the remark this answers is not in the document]
heliocastro: Is the same as previous one, is stacked, so it will be used afterwards.
I prefer stays like this, and be consistent with all other commits

Dockerfile
heliocastro: Yes, intentional.
GOPATH forces go to always try to use /opt, that creates another permission issue for non root user.
Removing it, Go uses first home dir, then go dir.

heliocastro: Fixed

fviernau (approved):

tsteenbe (approved):

PR #6150, opened 2022-12-01
docker: Sanitize Dart in docker

heliocastro:


tsteenbe (changes requested):
docker: Move Dart to independent layer 

You say move here so I was expecting a delete - should this be "docker: Create Dart build layer"?

current body:
Dart was installed using their own debian based repository, which not provided
aarch64 builds.
Moving it for independent layer make it distro neutral and allow to
properly choose the architecture.
proposed body:
Dart was installed via its debian based repository, which does not provide
aarch64 builds. Switch to an independent build layer which is distro neutral and allows
setting the correct architecture.

docker: Create dart binary install layer 
proposed message: Create Dart binary install layer 
current body:
Improve docker cache efficiency as have single layer from scratch with
only installation binaries.
proposed body:
Improve docker cache efficiency via a single scratch layer with
only installation binaries.

docker: Replace Dart handling to use new dart layer 
proposed message: Replace Dart installation with new binary layer

tsteenbe (approved):

Dockerfile
```diff
@@ -261,6 +261,23 @@ RUN curl -Os https://dl.google.com/android/repository/commandlinetools-linux-${A
     && chmod -R o+rw $ANDROID_HOME
 COPY docker/android.sh /etc/profile.d
 
+#------------------------------------------------------------------------
```
fviernau, line 264: From the commit message, I was expecting that the old way of setting up Dart (via APT) was replaced with the new way: separate layer + copy command. I believe the use of terms "move"  and "switch" lead me into thinking so.

Looking at the diff, it seems that the old way still remains and the new way is added but not used.
I believe that this whole "move" step actually is atomic, and should not be split into two commits.
If you agree with that this can be easily fixed by squashing the commit "docker: Replace Dart installation with new binary layer " into this one. (The third commit IMO is nice to have separately).

fviernau (changes requested):

Dockerfile
fviernau: In other PRs it was agreed with other reviewers to have these separate commits, so keep as-is as it is analog.

fviernau (approved):

Dockerfile
```diff
@@ -278,6 +278,9 @@ RUN --mount=type=tmpfs,target=/dart \
     && curl -o /dart/dart.zip -L https://storage.googleapis.com/dart-archive/channels/stable/release/${DART_VERSION}/sdk/dartsdk-linux-${arch}-release.zip \
     && unzip /dart/dart.zip
 
+FROM scratch AS dart
```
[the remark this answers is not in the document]
heliocastro: The caching is not related to local environment, this is granted by docker. The issue lives on github itself.
Today cache is done with entire image, so is *big*, every intermediate layer includes the build layer stack, and is never discarded.
With this approach, the layers used to build will be discarded, and only this generated from scratch will be cached, using this https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-to

So, after this layers are all merged and the build layer removed, i will set another commit for the workflow, which on the first build will be slow, for docker, but the next one will use the small cached images and build only ort, which will be near 80% faster. 
The caches will be only invalidated when effectively we change one of the language versions, or the base layer ( which will invalidate all ), but this is really not so frequent

PR #6154, opened 2022-12-01
Switch to the Semver4j fork

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Bower.kt
+++ b/analyzer/src/main/kotlin/managers/Bower.kt
@@ -21,8 +21,6 @@ package org.ossreviewtoolkit.analyzer.managers
 
 import com.fasterxml.jackson.databind.JsonNode
 
-import com.vdurmont.semver4j.Requirement
-
 import java.io.File
 import java.util.SortedSet
 import java.util.Stack
@@ -49,6 +47,9 @@ import org.ossreviewtoolkit.utils.common.fieldNamesOrEmpty
 import org.ossreviewtoolkit.utils.common.fieldsOrEmpty
 import org.ossreviewtoolkit.utils.common.stashDirectories
 import org.ossreviewtoolkit.utils.common.textValueOrEmpty
+
+import org.semver4j.RangesList
+import org.semver4j.RangesListFactory
 
 /**
  * The [Bower](https://bower.io/) package manager for JavaScript.
@@ -225,7 +226,7 @@ class Bower(
 
     override fun command(workingDir: File?) = if (Os.isWindows) "bower.cmd" else "bower"
 
-    override fun getVersionRequirement(): Requirement = Requirement.buildIvy("[1.8.8,)")
+    override fun getVersionRequirement(): RangesList = RangesListFactory.create("[1.8.8,)")
 
     override fun beforeResolution(definitionFiles: List<File>) = checkVersion()
 
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -24,8 +24,6 @@ package org.ossreviewtoolkit.analyzer.managers
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.node.ObjectNode
 
-import com.vdurmont.semver4j.Requirement
-
 import java.io.File
 import java.util.concurrent.ConcurrentHashMap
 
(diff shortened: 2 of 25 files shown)


analyzer/src/main/kotlin/managers/Bower.kt
```diff
@@ -21,8 +21,6 @@ package org.ossreviewtoolkit.analyzer.managers
 
 import com.fasterxml.jackson.databind.JsonNode
 
-import com.vdurmont.semver4j.Requirement
```
MarcelBochtler, line 24: Commit message: You're never referencing `[1]`.

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -145,7 +146,7 @@ open class Npm(
 
     override fun command(workingDir: File?) = if (Os.isWindows) "npm.cmd" else "npm"
 
-    override fun getVersionRequirement(): Requirement = Requirement.buildNPM("6.* - 8.*")
+    override fun getVersionRequirement(): RangesList = RangesListFactory.create("6.* - 8.*")
```
MarcelBochtler, line 150: Note for myself: The NPM ranges are covered by the [XRangeProcessor](https://github.com/semver4j/semver4j/blob/2258ea7b3aa3a43afa8ce41203af9ac844b95d4f/src/main/java/org/semver4j/internal/range/processor/XRangeProcessor.java).

cli/src/main/kotlin/commands/RequirementsCommand.kt
```diff
@@ -120,7 +122,7 @@ class RequirementsCommand : OrtCommand(
                             val actualVersion = tool.getVersion()
                             runCatching {
                                 val isRequiredVersion = tool.getVersionRequirement().let {
-                                    it == CommandLineTool.ANY_VERSION || it.isSatisfiedBy(actualVersion)
+                                    Semver.coerce(actualVersion).satisfies(it)
```
MarcelBochtler, line 125: Why is the check for `ANY_VERSION` dropped?

downloader/src/main/kotlin/VersionControlSystem.kt
```diff
@@ -404,7 +404,7 @@ abstract class VersionControlSystem {
      * Check whether the VCS tool is at least of the specified [expectedVersion], e.g. to check for features.
      */
     fun isAtLeastVersion(expectedVersion: String): Boolean {
-        val actualVersion = Semver(getVersion(), Semver.SemverType.LOOSE)
-        return actualVersion.isGreaterThanOrEqualTo(expectedVersion)
+        val actualVersion = Semver.coerce(getVersion())
+        return actualVersion.isGreaterThanOrEqualTo(Semver.coerce(expectedVersion))
```
MarcelBochtler, line 408: Note for myself: `SemverType.LOOSE` was implicitly used before as well for `expectedVersion`.

model/src/main/kotlin/PackageCuration.kt
```diff
@@ -65,10 +64,7 @@ data class PackageCuration(
      */
     private fun isApplicableIvyVersion(pkgId: Identifier) =
         runCatching {
-            // TODO: This check does not completely comply to the Ivy version-matchers specification. E.g. the version
-            //       '1.0' does not satisfy the version range [1.0,2.0], see
-            //       https://github.com/vdurmont/semver4j/issues/67.
-            Requirement.buildIvy(id.version).isSatisfiedBy(Semver(pkgId.version, Semver.SemverType.LOOSE))
+            Semver.coerce(pkgId.version).satisfies(id.version)
```
MarcelBochtler, line 67: `satisfies(String)` never uses `coerce` internally.
Isn't this required to solve the `TODO` above?

cli/src/main/kotlin/commands/RequirementsCommand.kt
sschuberth: `ANY_VERSION` was introduced here in d8c65c3f8ac329fc89c01f2694b03f5cf6d41309 to work around a limitation of the original `isSatisfiedBy()` call to not handle non-semantic versions, but even those should be accepted here. This is not necessary anymore with the new code as it uses `coerce()` before.

model/src/main/kotlin/PackageCuration.kt
sschuberth: It shouldn't be necessary (anymore), see https://github.com/semver4j/semver4j/issues/100. And I just manually verified that `Semver.coerce("1.0").satisfies("[1.0,2.0]") == true`.

MarcelBochtler (approved):

PR #6159, opened 2022-12-02
Docker: Sanitize Ruby on docker

heliocastro:


tsteenbe (changes requested):
docker: Use ort-base-image instead of build layer

Commit message says 'python' layer instead of 'ruby layer'.


docker: Set GEM_HOME to gem iuser installs

Commit message says 'iuser' instead of 'user', may better commit message "docker: Set GEM_HOME to permissible /var/tmp"

current body:
Running as a non-provileged user requires a user writable directory, so
redirecting to permissable /var/tmp will allow gem to create the
needed userspace.
proposed body:
Use permissible /var/tmp for gem to create user space as running
Docker image as a non-privileged user requires a user writable directory.

tsteenbe (dismissed):

Dockerfile
```diff
@@ -314,9 +314,9 @@ COPY --from=pythonbuild /etc/profile.d/python.sh /etc/profile.d/
 RUN chmod o+rwx ${PYENV_ROOT}
 
 # Ruby
-ARG RBENV_ROOT=/opt/rbenv/
-COPY --chown=$USERNAME:$USERNAME --from=rubybuild ${RBENV_ROOT} ${RBENV_ROOT}
-COPY --from=rubybuild /etc/profile.d/ruby.sh /etc/profile.d/
+ENV RBENV_ROOT=/opt/rbenv/
+ENV PATH=$PATH:${RBENV_ROOT}/bin:${RBENV_ROOT}/shims:${RBENV_ROOT}/plugins/ruby-install/bin
+COPY --from=ruby ${RBENV_ROOT} ${RBENV_ROOT}
```
fviernau, line 318: Similar question as in other PR: Could the change in line 319 be done in preceeding commit instead, so that 
`ruby` is taken into use instantly? (not unused)

fviernau (approved):

tsteenbe (approved):

Dockerfile
tsteenbe: It will be used afterwards. I prefer stays like this, and be consistent with all other commits

Issue #6146, opened 2022-11-30
Git returns wrong nested repositories for uninitialized submodule


mnonnenmacher:
When running the analyzer on a project with uninitialized Git submodules, the nested repository information is wrong. For each uninitialized submodule it contains the original repository. The expected behavior would be to either fail or not return any entry for the uninitialized submodule.

This can be reproduced by running the analyzer on the https://github.com/oss-review-toolkit/ort-test-data-git-submodules repository without initializing the submodules. The analyzer result contains the following repositories:

```yaml
repository:
  vcs:
    type: "Git"
    url: "git@github.com:oss-review-toolkit/ort-test-data-git-submodules.git"
    revision: "fcea94bab5835172e826afddb9f6427274c983b9"
    path: ""
  vcs_processed:
    type: "Git"
    url: "ssh://git@github.com/oss-review-toolkit/ort-test-data-git-submodules.git"
    revision: "fcea94bab5835172e826afddb9f6427274c983b9"
    path: ""
  nested_repositories:
    commons-text:
      type: "Git"
      url: "git@github.com:oss-review-toolkit/ort-test-data-git-submodules.git"
      revision: "fcea94bab5835172e826afddb9f6427274c983b9"
      path: "commons-text"
    test-data-npm:
      type: "Git"
      url: "git@github.com:oss-review-toolkit/ort-test-data-git-submodules.git"
      revision: "fcea94bab5835172e826afddb9f6427274c983b9"
      path: "test-data-npm"
```

sschuberth:
> should the `GitWorkingTree` fail if there are no uninitialized submodules or should it ignore them?

I believe you mean "no initialized" or just "uninitalized", @mnonnenmacher?

I think it should ignore them, otherwise we'd not be able to analyze repos where "unimportant" stuff (like we do  with test projects) have been outsources to Git submodules; such things would usually not be relevant for analysis anyway.

PR #6166, opened 2022-12-05
GitWorkingTree: Clear VCS path if submodule is not initialized.

bennati:
The nested repository information and the scanner cache require that the VCS path is empty, which is typically done when resolving a submodule (from the local path to the upstream repo).
This resolution is not possbile if the submodule is not initialized.

Closes: #6146 

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -97,6 +97,8 @@ jobs:
           pip install --no-cache-dir --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION
     - name: Checkout Repository
       uses: actions/checkout@v3
+      with:
+        submodules: recursive
     - name: Setup Java
       uses: actions/setup-java@v3
       with:
--- a/downloader/src/main/kotlin/vcs/GitWorkingTree.kt
+++ b/downloader/src/main/kotlin/vcs/GitWorkingTree.kt
@@ -70,18 +70,17 @@ open class GitWorkingTree(
             SubmoduleWalk.forIndex(repo).use { walk ->
                 while (walk.next()) {
                     val path = "$prefix${walk.path}"
-                    paths += path
 
                     if (walk.repository == null) {
                         logger.info {
                             "Git submodule at '$path' not initialized. Cannot recursively list its submodules."
                         }
+                    } else {
+                        paths += path
 
-                        continue
-                    }
-
-                    walk.repository.use { submoduleRepo ->
-                        listSubmodules(path, submoduleRepo, paths)
+                        walk.repository.use { submoduleRepo ->
+                            listSubmodules(path, submoduleRepo, paths)
+                        }
                     }
                 }
             }
--- a/downloader/src/main/kotlin/vcs/GitWorkingTree.kt
+++ b/downloader/src/main/kotlin/vcs/GitWorkingTree.kt
@@ -72,7 +72,7 @@ open class GitWorkingTree(
                     val path = "$prefix${walk.path}"
 
                     if (walk.repository == null) {
-                        logger.info {
+                        logger.warn {
                             "Git submodule at '$path' not initialized. Cannot recursively list its submodules."
                         }
                     } else {


sschuberth (changes requested):

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -70,18 +70,18 @@ open class GitWorkingTree(
...
                         logger.info {
                             "Git submodule at '$path' not initialized. Cannot recursively list its submodules."
                         }
 
-                        continue
-                    }
-
-                    walk.repository.use { submoduleRepo ->
-                        listSubmodules(path, submoduleRepo, paths)
+                        paths += ""
```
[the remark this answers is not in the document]
bennati: i amended the commit

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -70,18 +70,16 @@ open class GitWorkingTree(
...
                             "Git submodule at '$path' not initialized. Cannot recursively list its submodules."
                         }
-
-                        continue
-                    }
-
-                    walk.repository.use { submoduleRepo ->
-                        listSubmodules(path, submoduleRepo, paths)
+                    } else {
+                        paths += path
```
sschuberth, line 79: The code looks ok to me now, but the commit message doesn't:

This is not about clearing a VCS path, but about only adding paths to *initialized* submodules. So I'd say:

```
GitWorkingTree: Only add paths to initialized submodules

Do not add paths to uninitialized submodules. As these are not present in
the working tree, they should not count towards the provenance information
of that is going to be scanned.
```
```

**Review by sschuberth — CHANGES_REQUESTED** [review]

File: downloader/src/main/kotlin/vcs/GitWorkingTree.kt (hunk 2 above)
Comment (sschuberth) [reply downloader/src/main/kotlin/vcs/GitWorkingTree.kt:79]: Unrelated, but could you please add an additional commit that changes the above `info`-log to `warn`? With the rationale that it should be an exceptional case that submodules are not initialized, as it might hide stuff from getting scanned.

File: downloader/src/main/kotlin/vcs/GitWorkingTree.kt (hunk 1 above)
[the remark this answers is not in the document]
Comment (sschuberth) [reply downloader/src/main/kotlin/vcs/GitWorkingTree.kt]: Oh, and if this also fixes #6070, please mention that in the commit message, too.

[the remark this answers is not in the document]
Comment (bennati) [reply downloader/src/main/kotlin/vcs/GitWorkingTree.kt]: I am not yet sure it fixes the issue, still looking into it

File: downloader/src/main/kotlin/vcs/GitWorkingTree.kt [hunk 3]
```diff
@@ -70,18 +70,16 @@ open class GitWorkingTree(
...
                         logger.info {
                             "Git submodule at '$path' not initialized. Cannot recursively list its submodules."
                         }
-
-                        continue
-                    }
-
-                    walk.repository.use { submoduleRepo ->
-                        listSubmodules(path, submoduleRepo, paths)
+                    } else {
```
fviernau, line 78: commit: I'm not sure this change is ok:

1. When the working tree is used for the sources of a project, there could be a point in saying that non-recursive cloning is fine. Or maybe the source have been clone manually, and only some submodules are initialized.
2. When the working tree is used for the sources of a dependency, I believe ORT does recursive cloning always. Also the provenance resolver recursively determines submodules. 

Is this fix targeting an issue which appears with (1) projects or with (2) dependencies?

fviernau (dismissed):

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
bennati: it targets issues with (2) dependencies: these appear both in git projects, which depend on protobuf, as well as in monorepo projects which contain the protobuf code in a subfolder.

fviernau: @bennati : Just to double check: By dependencies, we're talking about packages consumed via a package manager (not dependencies included in the source code of the project like in C++ / Git-Repo projects), right?

bennati: both: in the case of the git project, protobuf is fetched using a package manager, in the case of the monorepo project, protobuf is part of the source code. The problem in both cases comes from uninitialized submodules inside protobuf's code.

sschuberth: > If you see another behavior @bennati, then we have yet another problem.

But independently of that, I still believe that not listing something that's not there is the right thing to do in this context, no matter *why* it's not there.

bennati: ORT downloader does clone recursively, but that does not mean it always resolves all submodules correctly: for example we had a repository which contained a submodule pointing to itself, which caused the scanner to crash. We fixed it by removing the recursive submodule in that case.

sschuberth: > that does not mean it always resolves all submodules correctly

Well, the downloader *should* resolve all submodules correctly *if* they are properly configured.

> we had a repository which contained a submodule pointing to itself

Do you think it makes sense add a check fort this, and skip such "submodules", too?

fviernau: > But independently of that, I still believe that not listing something that's not there is the right thing to do in this context, no matter _why_ it's not there.

Hm, as we ditched the legacy scanner now all scanning of dependencies uses the provenance resolver.
An implementation detail of the provenance resolver is that it's using Git's recursive submodule initialization.
To my understanding this algorithm is sub-optimal and can be improved by not using Git's recursive cloning anymore. I've assumed that the current implementation just was taking a short cut, but that longer term plan was to do exactly that. This basically means one would want to list only the (direct) submodules of a repository, e.g. `getNested()` would be changed. So, the case to list submodules without initializing them becomes the desired use case. 

So, I fear we're moving into the opposite direction here. @mnonnenmacher inupt would be interresting here as he has written the provenance resolver.

But going back from future to the code-as is, wouldn't the provenance resolver need to know if any submodule could not be intialized? (So that it can add the corresponding OrtIssue to the scan result)

Given these two, I'm still not convinced yet that the changes are desired.

bennati: > Well, the downloader should resolve all submodules correctly if they are properly configured.
In case the project is not properly configured, we don't need this change at all because the project can simply be fixed. The issue arises when dependencies such as protobuf are not properly configured and the scanner crashes.

sschuberth: > The issue arises when dependencies such as protobuf are not properly configured and the scanner crashes.

We should probably align on what we call "properly configured" WRT submodules: To me, a repo with uninitialized submodules is still properly configured. But a repo with submodules that point to the parent repo is not properly configured.

devcooch:
Maybe small observation on the problem we try to fix with @bennati.
I tried both repo v2.21 (default) and v2.30 (latest).
There are two commands needed to initialize working tree locally:
`repo init --groups=all --no-repo-verify --no-clone-bundle --submodules --repo-branch=$GIT_REPO_BRANCH -u $MANIFEST_URL -b $MANIFEST_BRANCH -m $MANIFEST_FILENAME`
and then
`repo sync -c --force-sync --fetch-submodules`

However there is a behavior difference between ORT and git-repo.
Git-repo is acting on `.gitsubmodules` located in the root of repository (where `.git` folder appears), while ORT recursively scans all subfolders.
When we have nested repository in `upstream` folder with `.gitmodules` inside it - repo is silently ignoring it, while ORT is expecting it work.

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
bennati: Ok, let's use your definition

bennati:
@fviernau that is correct, this PR is not trying to fix point 5, i.e. have ORT ignore the .gitmodules file, but to have ORT ignore non-initialized submodules, which are defined in .gitmodules. 
Were git-repo initializing these submodules, we'd be happy to have ORT scan them :)

fviernau:
> @fviernau that is correct.
> This PR is not trying to fix point 5, i.e. have ORT ignore the .gitmodules file, but to have ORT ignore non-initialized submodules, which are defined in .gitmodules.
> Were git-repo initializing these submodules, we'd be happy to have ORT scan them :)

Are you talking about `.gitmodules` file in the root dir, or elsewhere (in `upstream`)?

fviernau:
> BTW, I can confirm that the fix #6168 does not solve the issue for our git-repo projects (ran it with ORT [a35f251](https://github.com/oss-review-toolkit/ort/commit/a35f251f3f2c8f332ff08d674766e37ff70ea984)).

Thanks! With that commit I didn't intent to fix the issue. I just found that when reading the code.

zhernovs:
@mnonnenmacher @sschuberth do we still need any testing before merging this?

mnonnenmacher:
> > @mnonnenmacher @sschuberth do we still need any testing before merging this?
> 
> From my side, we need to fix the currently broken downloader funTests on CI first, I'm currently looking into that. Then this PR needs to be rebased. WRT additional testing, I'm referring to @mnonnenmacher.

I'm fine with the changes, once rebased after funTests are fixed.

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -70,18 +70,16 @@ open class GitWorkingTree(
...
+                        paths += path
+                        walk.repository.use { submoduleRepo ->
```
sschuberth, line 80: Meanwhile, a tiny style nit: We prefer to have empty lines before the start of code blocks, like here.

bennati: fixed

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -77,7 +77,8 @@ open class GitWorkingTree(
                         }
                     } else {
                         paths += path
-                        walk.repository.use { submoduleRepo ->
+                        walk.repository.use {
+                            submoduleRepo ->
```
sschuberth, line 81: That's not what I mean. I meant the code should be formatted like

```
                        paths += path

                        walk.repository.use { submoduleRepo ->
                            listSubmodules(path, submoduleRepo, paths)
                        }
```

And please squash this change with the first commit instead of adding a new one.

sschuberth:
Commit message titles now need to be rewritten to follow [Conventional Commits](https://www.conventionalcommits.org/).

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
bennati: done

bennati:
@sschuberth I reworded the commits, let me know if they look good to you.

sschuberth:
> @sschuberth I reworded the commits, let me know if they look good to you.

Please have a look at the failing checks:

https://github.com/oss-review-toolkit/ort/actions/runs/3938307580/jobs/6736845458

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
```diff
@@ -72,7 +72,7 @@ open class GitWorkingTree(
                     val path = "$prefix${walk.path}"
 
                     if (walk.repository == null) {
-                        logger.info {
+                        logger.warn {
```
sschuberth, line 75: From the "type", this is rather a "chore" than a "fix".

bennati: fixed

sschuberth (dismissed):

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
bennati: is this conversation resolved?

sschuberth: From my side, yes.

sschuberth:
The remaining problem seems to be that `GitWorkingTreeFunTest` fails, probably because the GitHub action does not clone recursively, and thus submodules are now not found anymore as they're not initialized. Locally the test passes for me. So youo should probably amend your first commit to also add `submodules: recursive` after https://github.com/oss-review-toolkit/ort/blob/d91b746f5986b04e970bdd85fad114165cae3e1b/.github/workflows/build-and-test.yml#L99

.github/workflows/build-and-test.yml
```diff
@@ -97,6 +97,7 @@ jobs:
           pip install --no-cache-dir --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION
     - name: Checkout Repository
       uses: actions/checkout@v3
+      submodules: recursive
```
sschuberth, line 100: This needs to be indented and nested below `with`. See the [usage](https://github.com/actions/checkout#usage).

bennati: fixed

sschuberth (approved):

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
bennati: Can we then resolve this conversation and merge the change? @fviernau

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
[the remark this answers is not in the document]
bennati: done

sschuberth (approved):

downloader/src/main/kotlin/vcs/GitWorkingTree.kt
bennati: @fviernau we are waiting on you

sschuberth: @tsteenbe mentioned in the community meeting that @fviernau is sick. I guess it's ok to dismiss this review and eventually do a fixup.

PR #6179, opened 2022-12-06
Some smaller CLI code improvements

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -115,8 +115,8 @@ class AnalyzerCommand : OrtCommand(
 
     private val repositoryConfigurationFile by option(
         "--repository-configuration-file",
-        help = "A file containing the repository configuration. If set, the '$ORT_REPO_CONFIG_FILENAME' file from " +
-                "the repository is ignored."
+        help = "A file containing the repository configuration. If set, overrides any repository configuration " +
+                "contained in a '$ORT_REPO_CONFIG_FILENAME' file in the repository."
     ).convert { it.expandTilde() }
         .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
         .convert { it.absoluteFile.normalize() }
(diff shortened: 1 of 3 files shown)


cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -179,8 +179,8 @@ class EvaluatorCommand : OrtCommand(
 
     private val repositoryConfigurationFile by option(
         "--repository-configuration-file",
-        help = "A file containing the repository configuration. If set, the '$ORT_REPO_CONFIG_FILENAME' overrides " +
-                "the repository configuration contained in the ORT result from the input file."
+        help = "A file containing the repository configuration. If set, the '$ORT_REPO_CONFIG_FILENAME' file from " +
```
fviernau, line 182: - The new description to me sounds as if the evaluator would read the ort.yml from the repository if the option is not provided, which would be misleading. 
- The repository configuration contained in the ORT file does not necessarily original from ORT_REPO_CONFIG_FILENAME.
-  nit: Also if $ORT_REPO_CONFIG_FILENAME is provided as argument for that option, that file is not ignored.

Would it make sense to say:  "... If set, the repository configuration contained in the ORT file is overridden by the given one."

fviernau (changes requested):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -115,8 +115,8 @@ class AnalyzerCommand : OrtCommand(
 
     private val repositoryConfigurationFile by option(
         "--repository-configuration-file",
-        help = "A file containing the repository configuration. If set, the '$ORT_REPO_CONFIG_FILENAME' file from " +
-                "the repository is ignored."
+        help = "A file containing the repository configuration. If set, overrides any repository configuration " +
```
fviernau, line 118: Would it make sense to simplify like follows? (up to you)

The filepath to the repository configuration file, relative to the repository root. Defaults to '$ORT_REPO_CONFIG_FILENAME'.

fviernau (approved):

PR #6180, opened 2022-12-06
Add a `helper-cli` commands to import and merge license classifications

sschuberth:

### DIFF
--- a/helper-cli/src/main/kotlin/commands/classifications/ImportCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/classifications/ImportCommand.kt
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2022 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands.classifications
+
+import com.fasterxml.jackson.module.kotlin.readValue
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.arguments.argument
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.types.enum
+import com.github.ajalt.clikt.parameters.types.file
+
+import java.net.URL
+
+import org.apache.logging.log4j.kotlin.Logging
+
+import org.ossreviewtoolkit.model.jsonMapper
+import org.ossreviewtoolkit.model.licenses.LicenseCategorization
+import org.ossreviewtoolkit.model.licenses.LicenseCategory
+import org.ossreviewtoolkit.model.licenses.LicenseClassifications
+import org.ossreviewtoolkit.model.yamlMapper
+import org.ossreviewtoolkit.utils.common.expandTilde
+import org.ossreviewtoolkit.utils.spdx.SpdxSingleLicenseExpression
+
+internal class ImportCommand : CliktCommand(
+    help = "Import license classifications from supported providers to ORT format."
+) {
+    private val provider by argument(
+        help = "The name of the provider to import license classifications from. Must be one of " +
+                "${enumValues<LicenseClassificationProvider>().map { it.name }}."
+    ).enum<LicenseClassificationProvider>()
+
+    // TODO: Make this default to the provider name with clikt 4, see https://github.com/ajalt/clikt/issues/381.
+    private val prefix by option(
+        "--prefix", "-p",
+        help = "A prefix to use for all category names declared by the provider's classifications."
+    )
+
+    private val licenseClassificationsFile by option(
+        "--output", "-o",
+        help = "The file to write the provided license classifications to."
+    ).convert { it.expandTilde() }
+        .file(mustExist = false, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+
+    override fun run() {
+        val classifications = provider.getClassifications().run {
+            prefix?.let { prefixCategoryNames(it) } ?: this
+        }.sort()
+
+        val yaml = yamlMapper.writeValueAsString(classifications)
+
+        licenseClassificationsFile?.run {
+            writeText(yaml)
+        } ?: println(yaml)
+    }
+}
+
+private data class EclipseLicenses(
+    val meta: Map<String, String>,
+    val approved: Map<String, String>,
+    val restricted: Map<String, String>
+)
+
+private enum class LicenseClassificationProvider(val url: String) : Logging {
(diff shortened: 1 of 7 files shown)


helper-cli/src/main/kotlin/commands/ImportLicenseClassificationsCommand.kt
```diff
@@ -0,0 +1,125 @@
...
+import org.ossreviewtoolkit.model.licenses.LicenseCategorization
+import org.ossreviewtoolkit.model.licenses.LicenseCategory
+import org.ossreviewtoolkit.model.licenses.LicenseClassifications
+import org.ossreviewtoolkit.model.yamlMapper
+import org.ossreviewtoolkit.utils.common.expandTilde
+import org.ossreviewtoolkit.utils.ort.ORT_LICENSE_CLASSIFICATIONS_FILENAME
+import org.ossreviewtoolkit.utils.ort.ortConfigDirectory
+import org.ossreviewtoolkit.utils.spdx.SpdxSingleLicenseExpression
+
+internal class ImportLicenseClassificationsCommand : CliktCommand(
```
fviernau, line 47: The command is doing multiple things at once.

1. Importing / mapping the models of the providers to the Ort model
2. Renaming categories / adding prefixes
3. Merging license classification files

To allow for more use cases, I propose to change the import command to import classifications from a single provider to a new empty `license-classifications.yml` file. E.g. `orth import-license-classifications --provider ECLIPSE`.

Then also have a dedicated merge command, which allows merging multiple license classification files.

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/ImportLicenseClassificationsCommand.kt
```diff
@@ -0,0 +1,125 @@
...
+                    categorization.copy(categories = categorization.categories.mapTo(mutableSetOf()) { "$prefix:$it" })
+                }
+            )
+        }
+    };
+
+    protected val prefix = name.lowercase()
+
+    abstract fun getClassifications(): LicenseClassifications
+}
```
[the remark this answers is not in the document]
sschuberth: > I'd expect the classifications to overlap.

Speaking of overlap: There are two aspects to it. Theoretically, classification might *syntactically* overlap, i.e. there could be clashes in category names (maybe with different meanings). This is circumvented by adding prefix as "category scopes".

Then there's semantical overlap possible, meaning the same or different category names actually describe the same thing. But that's fine, IMO. Maybe I want to pull in *X* different classification providers, and only if *N* out of *X* classify a particular license as "permissive" (modulo the prefix), I trust that information in my rules and act accordingly.

helper-cli/src/main/kotlin/commands/ImportLicenseClassificationsCommand.kt
sschuberth: I would have preferred to avoid the overhead of creating multiple commands. But I can see the benefit of having a separate merge command. I'll give it a try to implement this via more options to the same command.

fviernau: Maybe even 3: import, merge, rename. 
This will keep things simple and allow for more flexibility. Integrating a flexible renaming into the merge command IMO is way more complicated that having a separate rename categories command.

sschuberth: > This will keep things simple

Actually, this will add more complexity to the users at least I have in mind, as they now need to call 2 / 3 tools with arguments instead of just one.

> allow for more flexibility.

For a balance between flexibility and ease of use, I'm now planning for 2 commands: import and merge. The import will take an optional `--prefix` for the renaming / scoping of categories.

helper-cli/src/main/kotlin/commands/classifications/ImportCommand.kt
```diff
@@ -0,0 +1,121 @@
...
+    ).convert { it.expandTilde() }
+        .file(mustExist = false, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+
+    override fun run() {
+        val classifications = provider.getClassifications().run {
+            prefix?.let { prefixCategoryNames(it) } ?: this
+        }
+
+        val yaml = yamlMapper.writeValueAsString(classifications)
```
fviernau, line 69: Please sort the categories by name and the categorizations by id.

helper-cli/src/main/kotlin/commands/classifications/ImportCommand.kt
```diff
@@ -0,0 +1,121 @@
...
+    LDBCOLLECTOR("https://github.com/maxhbr/LDBcollector/raw/generated/ort/license-classifications.yml") {
+        override fun getClassifications(): LicenseClassifications = yamlMapper.readValue(URL(url))
+    };
+
+    abstract fun getClassifications(): LicenseClassifications
+}
+
+private fun LicenseClassifications.prefixCategoryNames(prefix: String) =
+    LicenseClassifications(
+        categories = categories.map { it.copy(name = prefix + it.name) },
```
fviernau, line 117: Do we commonly use string interpolation / should this be done here as well? (Same below)

helper-cli/src/main/kotlin/commands/classifications/MergeCommand.kt
```diff
@@ -0,0 +1,69 @@
...
+        .convert { it.absoluteFile.normalize() }
+
+    override fun run() {
+        val mergedClassifications = licenseClassificationsFiles.map {
+            yamlMapper.readValue<LicenseClassifications>(it)
+        }.reduce { existing, other ->
+            existing.merge(other)
+        }
+
+        val yaml = yamlMapper.writeValueAsString(mergedClassifications)
```
fviernau, line 63: Let's sort the entries in the output by name / id.

helper-cli/src/main/kotlin/commands/classifications/MergeCommand.kt
```diff
@@ -0,0 +1,69 @@
...
+    private val licenseClassificationsFiles by argument(
+        help = "The license classifications file to merge, in order. Existing classifications will be maintained " +
+                "unless they are redefined, in which case they will be overwritten."
+    ).convert { it.expandTilde() }
+        .file(mustExist = false, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = false)
+        .convert { it.absoluteFile.normalize() }
+        .multiple()
+        .check("At least two files are required for merging.") { it.size >= 2 }
+
+    private val mergedLicenseClassificationsFile by option(
```
fviernau, line 49: Would it make sense to name it `outputFile` , to align a bit with the option name?

fviernau (changes requested):

helper-cli/src/main/kotlin/commands/classifications/MergeCommand.kt
sschuberth: I deliberately chose a more speaking variable name for the output, as something like `outputFile` does not really tell you anything about *what* the output is.

helper-cli/src/main/kotlin/commands/classifications/ImportCommand.kt
sschuberth: I believe we only use / prefer string interpolation when mixing with literal stings.

helper-cli/src/main/kotlin/commands/classifications/ImportCommand.kt
fviernau: I'd prefer to keep the discussion isolated, my original intention was that I believe 

1. It's more valuable to have the output sorted than having append or even random order

In your above comment there are the points

2. Should license classifications be changed to enforce sorted order
3. Should a wrapping / companion class be added
4. probably more

As 1. is pretty much independent of all the rest and can be implemented with typing just a couple of characters, I believe it's more appropriate to decide 1. in isolation an implement already now what we believe is best suited for the output.

sschuberth: Ok, I've implemented the sorting.

fviernau: thx!

fviernau (approved):

Issue #6183, opened 2022-12-07
NuGet: Does not handle version ranges / wildcards


fviernau:
When the "Package Reference" notation is used to specify dependencies, e.g. in a `.csproj` file, then one can specify version ranges / wildcards as outlined in [1]. When doing so, for example with a `*` as version, ORT fails with e.g. [2]. 

[1] https://learn.microsoft.com/en-us/nuget/concepts/package-versioning
[2] 
```
Failed to get package data for 'NuGet::System.IO:*': IOException: Failed to retrieve package data for 'system.io' from any of [https://api.nuget.org/v3/registration5-gz-semver2].
```

sschuberth:
This issue is somewhat related to the `TODO` about floating versions, see

https://github.com/oss-review-toolkit/ort/blob/2c8e7ad0c1afb3d7371285ebb3d9bddb59b6657d/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt#L360-L361

sschuberth:
@pombredanne was this also addressed as part of https://github.com/oss-review-toolkit/ort/pull/6209?

pombredanne:
@sschuberth correct, nuget-inspector supports star versions aka. "floating ranges" in NuGet.
As a side note, these are the weirdest things as this is the only version range syntax that I know of that also influences how the underlying dependency resolution works. Here it changes whether the resolutions picks the lowest or highest version. That's a unique whart across all package managers I know of.
Feedback and tests are welcomed.

PR #6184, opened 2022-12-07
NuGetSupport: Improve the message of an OrtIssue

fviernau:
Make it obvious that version ranges and wildcards are not supported, to potentially safe some debugging time.

See also [1], [2].

[1] https://learn.microsoft.com/en-us/nuget/concepts/package-versioning 
[2] https://github.com/oss-review-toolkit/ort/issues/6183

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
@@ -64,7 +64,7 @@ internal const val OPTION_DIRECT_DEPENDENCIES_ONLY = "directDependenciesOnly"
 // See https://docs.microsoft.com/en-us/nuget/api/overview.
 private const val DEFAULT_SERVICE_INDEX_URL = "https://api.nuget.org/v3/index.json"
 private const val REGISTRATIONS_BASE_URL_TYPE = "RegistrationsBaseUrl/3.6.0"
-private val VERSION_RANGE_CHARS = charArrayOf('[', ']', '(', ')', ',')
+private val VERSION_RANGE_CHARS = charArrayOf('[', ']', '(', ')', ',', '*', '+')
 
 private val JSON_MAPPER = JsonMapper().registerKotlinModule()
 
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -382,3 +387,6 @@ interface XmlPackageFileReader {
      */
     fun getDependencies(definitionFile: File): Set<NuGetDependency>
 }
+
+private fun String.isFixedVersion() =
+    listOf('(', ')', '[', ']', '*', '+').none { it in this } && isNotBlank()
```
sschuberth, line 392: How about reusing `VERSION_RANGE_CHARS` (after adding `*` and `+` to it)? Also, maybe make the cheap `isNotBlank()` condition go first.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -137,11 +137,16 @@ class NuGetSupport(
...
             if (packageData == null) {
+                val hint = " Support for non-fixed versions, like '${id.version}', is not implemented".takeIf {
+                    !id.version.isFixedVersion()
+                }.orEmpty()
+
                 issues += createAndLogIssue(
                     source = "NuGet",
                     message = "Failed to get package data for '${id.toCoordinates()}' from any of " +
-                            "$registrationsBaseUrls."
+                            "$registrationsBaseUrls. $hint"
```
sschuberth, line 147: The separating single space is already part of `hint` and thus should be omitted here.

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
```diff
@@ -137,11 +137,16 @@ class NuGetSupport(
             val packageData = registrationsBaseUrls.firstNotNullOfOrNull { getPackageData(id, it) }
 
             if (packageData == null) {
+                val hint = " Support for non-fixed versions, like '${id.version}', is not implemented".takeIf {
```
sschuberth, line 140: "like" sounds like an example, but in fact you're checking exactly that version, so I'd simply say " Support for non-fixed version '${id.version}' is not implemented." (also note the added trailing dot at the end of the sentence).

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/utils/NuGetSupport.kt
fviernau: Do you mean I should change the value of VERSION_RANGE_CHARS and thereby also the behavior of `Dependency.getId()` ?

sschuberth: > Do you mean I should change the value of VERSION_RANGE_CHARS and thereby also the behavior of `Dependency.getId()` ?

Yes. The behavior change should not matter, but helps you to reuse the variable. This, however, should be stated in the commit message then.

sschuberth (approved):

PR #6192, opened 2022-12-07
EvaluatorCommand: Support a rule file and a rule resource at the same time

sschuberth:

### DIFF
--- a/cli/src/funTest/kotlin/OrtMainFunTest.kt
+++ b/cli/src/funTest/kotlin/OrtMainFunTest.kt
@@ -237,9 +237,10 @@ class OrtMainFunTest : StringSpec() {
                 runMain(
                     "-c", configFile.path,
                     "evaluate",
-                    "-i", "build.gradle.kts",
-                    "--rules-file", "build.gradle.kts",
-                    "--rules-resource", "DUMMY"
+                    "-i", "src/funTest/assets/semver4j-ort-result.yml",
+                    "--rules-resource", "/rules/osadl.rules.kts",
+                    "--package-configuration-dir", "src",
+                    "--package-configuration-file", "build.gradle.kts"
                 )
             }
         }
--- a/cli/src/main/kotlin/commands/EvaluatorCommand.kt
+++ b/cli/src/main/kotlin/commands/EvaluatorCommand.kt
@@ -19,10 +19,10 @@
 
 package org.ossreviewtoolkit.cli.commands
 
+import com.github.ajalt.clikt.core.BadParameterValue
 import com.github.ajalt.clikt.core.ProgramResult
 import com.github.ajalt.clikt.core.UsageError
 import com.github.ajalt.clikt.core.requireObject
-import com.github.ajalt.clikt.parameters.groups.default
 import com.github.ajalt.clikt.parameters.groups.mutuallyExclusiveOptions
 import com.github.ajalt.clikt.parameters.groups.single
 import com.github.ajalt.clikt.parameters.options.associate
@@ -40,11 +40,8 @@ import kotlin.time.measureTimedValue
 
 import org.ossreviewtoolkit.analyzer.curation.FilePackageCurationProvider
 import org.ossreviewtoolkit.cli.GlobalOptions
-import org.ossreviewtoolkit.cli.GroupTypes.FileType
-import org.ossreviewtoolkit.cli.GroupTypes.StringType
 import org.ossreviewtoolkit.cli.OrtCommand
 import org.ossreviewtoolkit.cli.utils.OPTION_GROUP_CONFIGURATION
-import org.ossreviewtoolkit.cli.utils.OPTION_GROUP_RULE
 import org.ossreviewtoolkit.cli.utils.PackageConfigurationOption
 import org.ossreviewtoolkit.cli.utils.SeverityStats
 import org.ossreviewtoolkit.cli.utils.configurationGroup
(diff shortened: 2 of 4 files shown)


cli/src/funTest/kotlin/OrtMainFunTest.kt
```diff
@@ -237,9 +237,10 @@ class OrtMainFunTest : StringSpec() {
                 runMain(
                     "-c", configFile.path,
                     "evaluate",
-                    "-i", "build.gradle.kts",
-                    "--rules-file", "build.gradle.kts",
-                    "--rules-resource", "DUMMY"
+                    "-i", "src/funTest/assets/semver4j-ort-result.yml",
```
fviernau, line 240: Have you considered to instead provide a `osadlLicenseRule`, or however one would call it, in ORT's API so that the rules can make use of it, just as it can for all other rules in ORT's API?

I'm asking because that would keep the evaluator simpler. So, is there any particular reason why this is not possible / desired?

fviernau (dismissed):

cli/src/funTest/kotlin/OrtMainFunTest.kt
sschuberth: > Have you considered to instead provide a `osadlLicenseRule`, or however one would call it, in ORT's API

TBH I haven't considered it, I guess because currently the "highest level" of reusable rules code in the ORT API is a `Rule`, but here we'd need a whole `RuleSet`.

> So, is there any particular reason why this is not possible / desired?

I had two independent asks for the evaluator to support multiple scripts recently. It's simply more convenient for users to just refer to those scripts without any need to write own "glue code". Also, separate scripts allow users to maintain them independently, e.g. one for OSS compliance folks and once for security officers to create violations on security vulnerabilities.

sschuberth: > Having multiple rule files will makes things hard to maintain as an ORT user

I disagree. There is no breaking change in the user-facing API (i.e. the CLI) here, so everything that worked before will continue to work, and there is no *need* for the user to actually use / maintain multiple rules files.

> prefer to keep things simple

Keeping things simple for the *user* is fine. But I believe we should not recoil from adding a *bit* of complexity on our side if it actually helps our users, like in this case.

> not seeing the use case to introduce this feature.

I guess our comments overlapped. The concrete use-case / ask for this feature is described in my comment above.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -241,18 +246,10 @@ class EvaluatorCommand : OrtCommand(
...
-                javaClass.getResource(rulesResource)?.readText()
-                    ?: throw UsageError("Invalid rules resource '$rulesResource'.")
-            }
-        }
-
         if (checkSyntax) {
-            if (Evaluator().checkSyntax(script)) {
+            val evaluator = Evaluator()
+
+            if (scripts.all { evaluator.checkSyntax(it) }) {
```
mnonnenmacher, line 252: It would be nice if the "Syntax check succeeded/failed" logs would also print which resource they refer to. E.g. `script` could be an object that contains not only the text but also the path of the resource, then this information would be available here. Otherwise it could be difficult to find out which file needs to be fixed.
The output could look like:
```
Syntax check for 'path1' succeeded.
Syntax check for 'path2' failed.
Syntax check for 'path1' succeeded.
```

evaluator/src/main/kotlin/Evaluator.kt
```diff
@@ -49,11 +49,8 @@ class Evaluator(
         scriptsInstancesSharing(true)
     }
 
-    fun run(script: String): EvaluatorRun {
-        val startTime = Instant.now()
+    fun run(script: String): List<RuleViolation> {
```
mnonnenmacher, line 52: I would prefer if the return type would stay `EvaluatorRun` because it is better for reusability of the function, otherwise if the function is used programmatically (e.g. in the upcoming server) the code to generate the `EvaluatorRun` needs to be duplicated. This would become more of an issue if more properties like `options` or `environment` would be added to `EvaluatorRun` later.
Maybe instead of calling this function multiple times, change the function to take multiple scripts?

mnonnenmacher (changes requested):

cli/src/funTest/kotlin/OrtMainFunTest.kt
mnonnenmacher: After the discussion in the community meeting today I agree that there are valid use cases for multiple rule files and that this does not add much complexity as you can keep using a single file if you prefer, so I am fine with this change.

evaluator/src/main/kotlin/Evaluator.kt
sschuberth: > Maybe instead of calling this function multiple times, change the function to take multiple scripts?

Good idea, done.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -106,20 +103,19 @@ class EvaluatorCommand : OrtCommand(
...
-        option(
-            "--rules-resource",
-            help = "The name of a script resource on the classpath that contains rules. Must not be used together " +
-                    "with '--rules-file'."
-        ).convert { StringType(it) },
-        name = OPTION_GROUP_RULE
-    ).single().default(FileType(ortConfigDirectory.resolve(ORT_EVALUATOR_RULES_FILENAME)))
+    private val rulesFile by option(
+        "--rules-file", "-r",
+        help = "The name of a script file containing rules. Must not be used together with '--rules-resource'."
```
mnonnenmacher, line 108: "Must not be used together with..." is obsolete now, same below.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -238,24 +243,21 @@ class EvaluatorCommand : OrtCommand(
...
-        }
+            var allChecksSucceeded = true
 
-        if (checkSyntax) {
-            if (Evaluator().checkSyntax(script)) {
-                println("Syntax check succeeded.")
-                return
+            scriptUrls.forEach {
+                if (evaluator.checkSyntax(it.readText())) {
+                    println("Syntax check for $it succeeded.")
```
mnonnenmacher, line 253: Maybe single quotes around `$it`? (same below)

mnonnenmacher (changes requested):

cli/src/main/kotlin/commands/EvaluatorCommand.kt
sschuberth: These are `URL`s which we usually don't quote.

mnonnenmacher (approved):

PR #6204, opened 2022-12-09
For Git, eventually add "main" as a revision candidate in addition to "master"

sschuberth:


downloader/src/main/kotlin/VersionControlSystem.kt
```diff
@@ -358,6 +358,11 @@ abstract class VersionControlSystem {
 
         addMetadataRevision(pkg.vcsProcessed.revision)
 
+        if (type == VcsType.GIT && revisionCandidates.singleOrNull() == "master") {
```
mnonnenmacher, line 361: I think this should only be done if "master" is coming from `pkg.vcsProcessed.revision`, because if `listRemoteBranches()` returns only "master" it does not make sense to check "main" as well.

mnonnenmacher (changes requested):

downloader/src/main/kotlin/VersionControlSystem.kt
```diff
@@ -358,6 +358,11 @@ abstract class VersionControlSystem {
 
         addMetadataRevision(pkg.vcsProcessed.revision)
 
+        if (type == VcsType.GIT && pkg.vcsProcessed.revision == "master") {
+            // Also try with Git's upcoming default branch name in case the repository is already using it.
+            addMetadataRevision("main")
```
mnonnenmacher, line 363: Should we also verify that `listRemoteBranches()` contains "main" before adding it?

sschuberth: That check basically is implicitly done as part of `addMetadataRevision()` when determining whether `main` is a fixed revision, which involves also listing the remote branches.

However, you're right that there's a general issue with the way `isFixedRevision()` currently works: For non-existing branches, it cannot *generally* tell whether it's a fixed revision or not. For example, consider the revision string "deadbeef". That could either be a non-existing branch name, or an existing short SHA1 for a commit. Distinguishing this better has been on my mental TODO-list for a while, but I'd like to do that separately of this PR.

mnonnenmacher (approved):

PR #6218, opened 2022-12-12
docker: Add bash entrypoint for base layer

heliocastro:


tsteenbe (approved):

Dockerfile
```diff
@@ -102,6 +102,8 @@ RUN /etc/scripts/import_certificates.sh \
         /etc/scripts/import_certificates.sh /tmp/certificates/; \
     fi
 
+ENTRYPOINT [ "/bin/bash" ]
```
sschuberth, line 105: An explanation in the commit message *why* to have an entrypoint in the base layer would have been nice.

PR #6225, opened 2022-12-13
Docker: Build components and run as (ort) user instead of root

heliocastro:


Dockerfile
```diff
@@ -379,57 +379,55 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 # Python
```
nnobelis, line 383: commit message "copied as correct"

Dockerfile
```diff
@@ -345,6 +345,8 @@ WORKDIR ${HOME}/src/ort
 RUN --mount=type=cache,target=/var/tmp/gradle \
```
nnobelis, line 347: commit message ": Change"

tsteenbe (changes requested):
Grammar/Spelling mistakes in commit message

```
docker: Chnage ownership of cache and build binded directories

BuildKit mounts types cache and bind are always as superuser,
so change ownership of both directories are required to
build ort as regular user.
``

to 

```
docker: Change ownership of cache and build binded directories

BuildKit mounts types cache and bind are always as superuser,
so change of ownership for both directories is required to
build ort as a regular user.
```

**Review by tsteenbe — CHANGES_REQUESTED** [review]
`docker Prepare components to be copy as correct user` -> missing `:` in commit message after `docker`

File: Dockerfile (hunk 1 above)
Comment (heliocastro) [reply Dockerfile:384]: > I thought the current image was already running as an user `ort`. was it not the case ?
Yes, the end image was running as ort, you are correct, but all components and the build stages were all done as root.
So now, if you look under the directory /opt, all components belongs to image built user. 
Same way, now base image, not only run image is running as user ort, so any change below that layer will be bounded to user ort, if try to escape the user jail, will fail.
Same as the copies after, that always end up with root.
Unfortunately this not solve *all* permission issues, as android-sdk tries to install in sdk root, and if user is not ort or root ( like github actions user ), we will end up in the permissions issues.

File: Dockerfile (hunk 2 above)
Comment (heliocastro) [reply Dockerfile:348]: Fixed

**Comment by heliocastro** [comment]
> I thought the current image was already running as an user `ort`. was it not the case ?

Not exactly. The current image has user ort, but the rest was built with root user, including the 

> > I thought the current image was already running as an user `ort`. was it not the case ?
> 
> I'm also curious about that. Already with the legacy Docker image it was [possible to run as non-root](https://github.com/oss-review-toolkit/ort/issues/3764#issuecomment-1042306906), and [Batect runs the analyzer funTests as the current user](https://github.com/oss-review-toolkit/ort/blob/main/batect.yml#L35-L36) for a long time already, so I wonder what the change is.

Same answer as above to Nico.
Is not anymore only run last layer as user ort, is build AND having ownership as user ort ( or custom user if defined on build )

**Comment by heliocastro** [comment]
> I thought the current image was already running as an user `ort`. was it not the case ?

Yes, the end image was running as ort, you are correct, but all components and the build stages were all done as root.
So now, if you look under the directory /opt, all components belongs to image built user. 
Same way, now base image, not only run image is running as user ort, so any change below that layer will be bounded to user ort, if try to escape the user jail, will fail.
Same as the copies after, that always end up with root.
Unfortunately this not solve *all* permission issues, as android-sdk tries to install in sdk root, and if user is not ort or root ( like github actions user ), we will end up in the permissions issues.

**Review by tsteenbe — APPROVED** [review]

**Comment by sschuberth** [comment]
> So now, if you look under the directory /opt, all components belongs to image built user.

I wonder whether that's the right approach to handle any permission issues. I mean, in a regular Linux installation also not everything that goes to `/opt` belongs to a certain user, but still all current and future users being created can make use of / read the stuff in `/opt`.

It [seems](https://askubuntu.com/a/169317) that stuff to which users only need read-access can / should go to `/opt`, and `/opt`'s permissions should be `drwxr-xr-x` and belong to `root:root`.

> Unfortunately this not solve all permission issues, as android-sdk tries to install in sdk root, and if user is not ort or root ( like github actions user ), we will end up in the permissions issues.

Which permission issues exactly? Is it about reading or writing?

**Comment by heliocastro** [comment]
> > So now, if you look under the directory /opt, all components belongs to image built user.
> 
> I wonder whether that's the right approach to handle any permission issues. I mean, in a regular Linux installation also not everything that goes to `/opt` belongs to a certain user, but still all current and future users being created can make use of / read the stuff in `/opt`.
> 
> It [seems](https://askubuntu.com/a/169317) that stuff to which users only need read-access can / should go to `/opt`, and `/opt`'s permissions should be `drwxr-xr-x` and belong to `root:root`.

For this specific then we need to quote LSB and you can find a lot of interpretation, even if you go further /opt was considered a non valid location on some linux implementations. But aside this, moving the implementation to any directory would not solve the problem, been /usr /var/local, /var/lib ( like postgres like to install ). We always end up on not be able to track it down proper file needs because root can write everywhere.

As we have a contained environment, it is not a bare metal production machine., is perfectly acceptable ownership of user there. There's even cases for example that some software (let's say postgres again) installs there as default postgres user.

What you can argue, and i'm ok with this, is that we keep root as the owner of /opt directories, but change group to a more permissive way, like  **ort** group, with r/w permissions, so any new user added to this group during process would have direct access.
Would not change much from current approach, tough.

> > Unfortunately this not solve all permission issues, as android-sdk tries to install in sdk root, and if user is not ort or root ( like github actions user ), we will end up in the permissions issues.
> 
> Which permission issues exactly? Is it about reading or writing?

Writing. Android applications targte to an specific sdk version. For obvious reasons, we don't have all sdk sources in Ort container, then when the android app buildsystem is hit to request the depes to scan, it will download the target files for the required sdk version.
Android downloads the specific files under the sdk root dir ( even set user base for local user ), and then if you are a user that is now owner of the sdk dir or root, it will fails.

Similar thing to Rust Cargo, but that i need do better check

**Comment by sschuberth** [comment]
For me, building this locally by calling `scripts/docker_build.sh` fails with

```
#0 0.772 Exception in thread "main" java.io.FileNotFoundException: /var/tmp/gradle/wrapper/dists/gradle-7.6-bin/9l9tetv7ltxvx3i8an4pb86ye/gradle-7.6-bin.zip.lck (Permission denied)
```

**Comment by heliocastro** [comment]
> #0 0.772 Exception in thread "main" java.io.FileNotFoundException: /var/tmp/gradle/wrapper/dists/gradle-7.6-bin/9l9tetv7ltxvx3i8an4pb86ye/gradle-7.6-bin.zip.lck (Permission denied)

This doesn't make sense. I built in more than 4 different machines, one is even a Mac M1, is there any GRADLE special env var on your machine ? I would understand this happening in some case in github, but not in local build

**Comment by sschuberth** [comment]
> > #0 0.772 Exception in thread "main" java.io.FileNotFoundException: /var/tmp/gradle/wrapper/dists/gradle-7.6-bin/9l9tetv7ltxvx3i8an4pb86ye/gradle-7.6-bin.zip.lck (Permission denied)
> 
> This doesn't make sense. I built in more than 4 different machines, one is even a Mac M1, is there any GRADLE special env var on your machine ? I would understand this happening in some case in github, but not in local build

I tried again after first pruning Docker and then running from scratch, that made the build work for me. Not sure what went wrong before.

**Comment by sschuberth** [comment]
> As we have a contained environment, it is not a bare metal production machine., is perfectly acceptable ownership of user there.

TBH I'm feeling uneasy with this. What I generally miss is a clear separation of stuff the ORT user needs only read / execute access to, and stuff the user needs write access to. The former should be installed in some shared system directory, and the latter probably below the user home. Currently, it seems everything is just chowned to the ORT user everywhere, even if more generic / restrictive permissions would suffice.

File: Dockerfile [hunk 3]
```diff
@@ -381,57 +381,56 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 # Python
 ENV PYENV_ROOT=/opt/python
 ENV PATH=$PATH:${PYENV_ROOT}/shims:${PYENV_ROOT}/bin
-COPY --from=python ${PYENV_ROOT} ${PYENV_ROOT}
+COPY --from=python --chown=$USER:$USER ${PYENV_ROOT} ${PYENV_ROOT}
 #RUN sudo chown -R ${USER}.${USER} ${PYENV_ROOT}
```
sschuberth, line 387: Can you please drop this out-commented line along the way?

Dockerfile
```diff
@@ -346,8 +346,7 @@ WORKDIR ${HOME}/src/ort
 RUN --mount=type=cache,target=/var/tmp/gradle \
     --mount=type=bind,target=${HOME}/src/ort,rw \
     export GRADLE_USER_HOME=/var/tmp/gradle \
-    && scripts/export_proxy_certificates.sh /tmp/certificates/ \
-    && scripts/import_certificates.sh /tmp/certificates/ \
+    && sudo chown -R "$USER". ${HOME}/src/ort /var/tmp/gradle \
```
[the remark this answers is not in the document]
sschuberth: At least for `--mount=type=cache`, [`uid` and `gui` seem to be supported](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mounttypecache). Can you please give this a try?

sschuberth (changes requested):

Dockerfile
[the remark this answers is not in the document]
heliocastro: Sorry, but not works. Cache assume you come from an empty directory.

There's two solutions on this case, they way it is today or back to COPY, but then any alteration locally on source would invalidate.

The perfect solution would be is gradle allows to do out-of-source builds, but i failed to find out and i think is not possible

Dockerfile
```diff
@@ -447,6 +447,8 @@ FROM components AS run
 # ORT
 COPY --from=ort --chown=$USER:$USER /opt/ort /opt/ort
 ENV PATH=$PATH:/opt/ort/bin
+RUN sudo ln -sf /opt/ort/bin/ort /usr/bin/ort \
```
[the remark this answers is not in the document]
heliocastro: Removed commit

Dockerfile
```diff
@@ -433,7 +433,10 @@ COPY --from=sbt ${SBT_HOME} ${SBT_HOME}
 
 # PHP composer
 ARG COMPOSER_VERSION=2.2
-RUN curl -ksS https://getcomposer.org/installer | php -- --install-dir=/bin --filename=composer --$COMPOSER_VERSION
+
+ENV PATH=$PATH:/opt/php/bin
+RUN mkdir -p /opt/php/bin \
+    && curl -ksS https://getcomposer.org/installer | php -- --install-dir=/opt/php/bin --filename=composer --$COMPOSER_VERSION
```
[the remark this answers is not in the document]
heliocastro: When you reach this stage, you already are running as user, so all operations are done with the ort user ( or the defined one )

Dockerfile
heliocastro: Removed

Dockerfile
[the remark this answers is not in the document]
heliocastro: If you want to do any operation as root, sudo is necessary

fviernau (approved):

PR #6227, opened 2022-12-13
utils: Add extension properties to convert values to mebi- and gibibytes

mnonnenmacher:
Add convenience properties for converting values to mebibytes and gibibytes.

### DIFF
--- a/utils/common/src/main/kotlin/Extensions.kt
+++ b/utils/common/src/main/kotlin/Extensions.kt
@@ -227,6 +227,21 @@ fun File.toSafeUri(): URI {
 }
 
 /**
+ * Return the bytes equal to this [Int] number of kibibytes (KiB).
+ */
+inline val Int.kibibytes get(): Long = this * 1024L
+
+/**
+ * Return the bytes equal to this [Int] number of mebibytes (MiB).
+ */
+inline val Int.mebibytes get(): Long = kibibytes * 1024L
+
+/**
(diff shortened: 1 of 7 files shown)


utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -226,6 +226,16 @@ fun File.toSafeUri(): URI {
...
 
+/**
+ * Return the bytes equal to this [Int] number of mebibytes (MiB).
+ */
+inline val Int.mebibytes get(): Long = this * 1024L * 1024L
+
+/**
+ * Return the bytes equal to this [Int] number of gibibytes (GiB).
+ */
+inline val Int.gibibytes get(): Long = this * 1024L * 1024L * 1204L
```
sschuberth, line 237: Should the implementation call `mebibytes` internally?

sschuberth (commented):
BTW, should we ever require more of this, we could take a look at https://github.com/nacular/measured.

utils/common/src/main/kotlin/Extensions.kt
mnonnenmacher: I wasn't sure if that's not overdoing it, what would you prefer?

sschuberth: I think I'd have a slight tendency towards doing it, as it IMO increases readability (by reducing the operations) and it documents the relation between the units.

sschuberth (approved):

PR #6228, opened 2022-12-13
README: Document the FossId and EvaluatedModel reporters

sschuberth:

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,6 +678,7 @@ following formats are supported (reporter names are case-insensitive):
   (`-f CtrlXAutomation`)
 * [CycloneDX](https://cyclonedx.org/) BOM (`-f CycloneDx`)
 * [Excel](https://en.wikipedia.org/wiki/Microsoft_Excel) sheet (`-f Excel`)
+* [FossId](https://fossid.com/open-source-compliance/) report download (HTML, SPDX, and Excel types)
 * [GitLabLicenseModel](https://docs.gitlab.com/ee/ci/pipelines/job_artifacts.html#artifactsreportslicense_scanning-ultimate)
   (`-f GitLabLicenseModel`)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -690,6 +691,9 @@ following formats are supported (reporter names are case-insensitive):
 * [SPDX Document](https://spdx.dev/specifications/), version 2.2 (`-f SpdxDocument`)
 * Static HTML (`-f StaticHtml`)
 * Web App (`-f WebApp`)
+  * Also see the [EvaluatedModelReporter](reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelReporter.kt)
+    (`-f EvaluatedModel`) which is stable JSON / YAML format used by the Web App report that is also suitable for custom
```
MarcelBochtler, line 695: Is the `evaluated-model` really considered stable? Just because it is used by the Web App report, doesn't make it stable, as changes in the Web App report might also cause / require changes in the evaluated model.
I'd like to hear other opinions on that.

sschuberth: Just FYI, "stable" refers here to the stability of the content of the report, not to the stability of the feature per se (which is a nit-picking difference, I agree).

The reason I'm documenting this is that I see users over and over again performing post-processing on ORT result files, and then suffering from breakages. These users were not aware of the evaluated model as an alternative, because it was not documented until now. But I do think it's worth poiniting out this report format independently of its use in the web-app report, and emphasizing that it's considered less likely to change than the ORT result files.

sschuberth: @MarcelBochtler I've simply dropped the word "stable"; "suitable for post-processing" should explain enough the purpose.

sschuberth: @nnobelis, @mnonnenmacher, any preferences here?

mnonnenmacher (approved):

README.md
mnonnenmacher: I'm fine with the wording.

PR #6229, opened 2022-12-13
model: Fix the `Set` contract for data classes that implement `Comparable`

sschuberth:
Kotlin's data classes have auto-generated `equals()` functions that take all primary properties into account. When implementing the `Comparable` interface for those classes, i.e. to give them a defined order in a `SortedSet`, special attention needs to be put on `compareTo()` to return 0 if and only if `equals()` returns true. This is required because "a sorted set performs all element comparisons using its compareTo method, so two elements that are deemed equal by this method are, from the standpoint of the sorted set, equal" [1]. This means that sorted sets are considered equal if for all their respective elements `compareTo()` returns 0. So if `compareTo()` is implemented to not consider all properties for an equality check, this can lead to unexpected results, for example when using Kotest's data class matchers.

Fix this and obey the general contract of the `Set` interface by taking all properties into account if the properties to sort on are equal. As the ordering does not matter if not all properties are equal but the properties to sort on are, arbitrarily return 1 from `compareTo()` in that case.

[1]: https://docs.oracle.com/javase/8/docs/api/java/util/SortedSet.html

### DIFF
--- a/model/src/main/kotlin/CuratedPackage.kt
+++ b/model/src/main/kotlin/CuratedPackage.kt
@@ -44,7 +44,12 @@ data class CuratedPackage(
     /**
      * A comparison function to sort packages by their identifier.
      */
-    override fun compareTo(other: CuratedPackage) = metadata.id.compareTo(other.metadata.id)
+    override fun compareTo(other: CuratedPackage) =
+        when {
+            metadata.id < other.metadata.id -> -1
+            metadata.id > other.metadata.id -> 1
+            else -> if (this == other) 0 else 1 // Return arbitrary inequality.
+        }
 
     /**
      * Return a [Package] representing the same package as this one but which does not have any curations applied.
--- a/model/src/main/kotlin/PackageReference.kt
+++ b/model/src/main/kotlin/PackageReference.kt
@@ -98,7 +98,12 @@ data class PackageReference(
      * A comparison function to sort package references by their identifier. This function ignores all other properties
      * except for [id].
      */
-    override fun compareTo(other: PackageReference) = id.compareTo(other.id)
+    override fun compareTo(other: PackageReference) =
+        when {
+            id < other.id -> -1
+            id > other.id -> 1
+            else -> if (this == other) 0 else 1 // Return arbitrary inequality.
+        }
 
     /**
      * Return whether the package identified by [id] is a (transitive) dependency of this reference.
--- a/model/src/main/kotlin/Scope.kt
+++ b/model/src/main/kotlin/Scope.kt
@@ -63,7 +63,12 @@ data class Scope(
     /**
      * A comparison function to sort scopes by their name.
      */
-    override fun compareTo(other: Scope) = compareValuesBy(this, other) { it.name }
+    override fun compareTo(other: Scope) =
+        when {
+            name < other.name -> -1
+            name > other.name -> 1
+            else -> if (this == other) 0 else 1 // Return arbitrary inequality.
+        }
 
     /**
      * Return whether the package identified by [id] is contained as a (transitive) dependency in this scope.
--- a/model/src/main/kotlin/DependencyGraph.kt
+++ b/model/src/main/kotlin/DependencyGraph.kt
@@ -332,10 +332,14 @@ class DependencyReference(
      * Define an order on [DependencyReference] instances. Instances are ordered by their indices and fragment indices.
      */
     override fun compareTo(other: DependencyReference): Int =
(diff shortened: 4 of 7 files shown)


model/src/main/kotlin/PackageReference.kt
```diff
@@ -98,7 +98,12 @@ data class PackageReference(
      * A comparison function to sort package references by their identifier. This function ignores all other properties
      * except for [id].
      */
-    override fun compareTo(other: PackageReference) = id.compareTo(other.id)
+    override fun compareTo(other: PackageReference) =
+        when {
+            id < other.id -> -1
+            id > other.id -> 1
+            else -> if (this == other) 0 else 1 // Return arbitrary inequality.
```
fviernau, line 105: The changes in this commit do impact the performance negatively.

In this particular case it leads to recursion: the whole subtree rooted at `this` and `other` is compared, which I doubt is a good idea.

model/src/main/kotlin/Scope.kt
```diff
@@ -63,7 +63,12 @@ data class Scope(
     /**
      * A comparison function to sort scopes by their name.
      */
-    override fun compareTo(other: Scope) = compareValuesBy(this, other) { it.name }
+    override fun compareTo(other: Scope) =
+        when {
+            name < other.name -> -1
+            name > other.name -> 1
+            else -> if (this == other) 0 else 1 // Return arbitrary inequality.
```
fviernau, line 70: Again, this is a performance problem.

model/src/main/kotlin/CuratedPackage.kt
```diff
@@ -44,7 +44,12 @@ data class CuratedPackage(
     /**
      * A comparison function to sort packages by their identifier.
      */
-    override fun compareTo(other: CuratedPackage) = metadata.id.compareTo(other.metadata.id)
+    override fun compareTo(other: CuratedPackage) =
+        when {
+            metadata.id < other.metadata.id -> -1
+            metadata.id > other.metadata.id -> 1
+            else -> if (this == other) 0 else 1 // Return arbitrary inequality.
+        }
```
fviernau, line 52: See this commit, I would again like to bring up the question whether
using a combination of the following two for the task of sorting things in serialization is the right approach:

1. Sorted collections
2. Implementing Comparable

E.g.: If implementing Comparable was avoided (for ordering things in the serializer), would these kind of problems then be not present?

fviernau (changes requested):

model/src/main/kotlin/Scope.kt
sschuberth: In this concrete case, I don't think there's a performance problem as the expensive comparison would only happen if scope names were equal, which should never be the case as a project cannot have multiple scopes with the same name.

model/src/main/kotlin/CuratedPackage.kt
fviernau: Seems we agree that this could be good to do. I've tried to write this down in an issue, see https://github.com/oss-review-toolkit/ort/issues/6235.

fviernau: How do you prefer to move this PR forward?

sschuberth: I don't know yet. This is just a draft to play with (I wanted to see what tests it breaks, if any).

model/src/main/kotlin/PackageReference.kt
sschuberth: > the whole subtree rooted at `this` and `other` is compared, which I doubt is a good idea.

Actually, that's exactly the point in order to be able to use Kotest's data class matchers e.g. on whole OrtResults. This would allow to get rid of the fragile text line comparison which is really hard to grasp on CI.

mnonnenmacher: But the text comparison also makes it very easy to see the effect of changes during development and to update the expected results files, if the plan is to change the matching for the functional analyzer tests we need to consider this.

sschuberth: > But the text comparison also makes it very easy to see the effect of changes

True, and it makes it easy to take over new expected results from actual results. My main point is that I believe Kotest's data class matchers *should* work with out code base in case someone decides to use them in tests, but currently they don't as soon as a `SortedSet` is involved anywhere in the hierarchy.

sschuberth:
I'm closing this in favor of https://github.com/oss-review-toolkit/ort/pull/6244.

PR #6233, opened 2022-12-14
Introduce extension functions for chunked ClearlyDefined service requests

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
@@ -23,9 +23,6 @@ import com.fasterxml.jackson.databind.JsonMappingException
 
 import java.net.HttpURLConnection
 
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.runBlocking
-
 import okhttp3.OkHttpClient
 
 import org.apache.logging.log4j.kotlin.Logging
@@ -36,6 +33,7 @@ import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService.Server
 import org.ossreviewtoolkit.clients.clearlydefined.ComponentType
 import org.ossreviewtoolkit.clients.clearlydefined.Coordinates
 import org.ossreviewtoolkit.clients.clearlydefined.SourceLocation
+import org.ossreviewtoolkit.clients.clearlydefined.callBlocking
 import org.ossreviewtoolkit.model.Hash
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.PackageCuration
@@ -113,7 +111,7 @@ class ClearlyDefinedPackageCurationProvider(
         val contributedCurations = runCatching {
             buildMap {
                 coordinatesToIds.keys.chunked(BULK_REQUEST_SIZE).forEach { coordinates ->
-                    putAll(runBlocking(Dispatchers.IO) { service.getCurations(coordinates) })
+                    putAll(service.callBlocking { getCurations(coordinates) })
                 }
             }
         }.onFailure { e ->
--- a/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
@@ -33,7 +33,7 @@ import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService.Server
 import org.ossreviewtoolkit.clients.clearlydefined.ComponentType
 import org.ossreviewtoolkit.clients.clearlydefined.Coordinates
 import org.ossreviewtoolkit.clients.clearlydefined.SourceLocation
-import org.ossreviewtoolkit.clients.clearlydefined.callBlocking
+import org.ossreviewtoolkit.clients.clearlydefined.getCurationsChunked
 import org.ossreviewtoolkit.model.Hash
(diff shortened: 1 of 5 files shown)


analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
```diff
@@ -110,12 +105,8 @@ class ClearlyDefinedPackageCurationProvider(
...
         }.toMap()
 
-        val contributedCurations = runCatching {
-            buildMap {
-                coordinatesToIds.keys.chunked(BULK_REQUEST_SIZE).forEach { coordinates ->
-                    putAll(runBlocking(Dispatchers.IO) { service.getCurations(coordinates) })
-                }
-            }
+        val curations = runCatching {
+            runBlocking(Dispatchers.IO) { service.getCurationsChunked(coordinatesToIds.keys) }
```
fviernau, line 109: Can you help me understanding this please? 
I understand that the single call to `getCurationsChunked()` is scheduled / executed on another thread, and this thread blocks and waits for it to return. I wonder what's the reason to do that context switch....

sschuberth: The original `forEach` loop from this code location has been moved to `getCurationsChunked`, but *without* the `runBlocking``around the inner `getCurations` suspend function call. So as `getCurationsChunked` now calls a suspend function, it needs to be a suspend function itself, unless it uses `runBlocking` itself. The reason why I chose not to make `getCurationsChunked` use `runBlocking` itself is that it's usually a good idea to block as late as possible, i.e. the caller should decide whether it need the result now, or if it could wait a bit as it does more other things in parallel.

sschuberth: Maybe @oheger-bosch, any opinion here?

mnonnenmacher: It makes sense for `getCurationsChunked` to be a suspend function because the rest of the `ClearlyDefinedService` API is also suspend functions. So we have to use `runBlocking`here because `getCurationsFor` is not a suspend function.
Regarding `Dispatchers.IO` it is just best practice to use it for IO calls. It's of limited use here, because the current thread is blocked anyway. For example, it would make more sense to use it if this code was a coroutine running in a Dispatcher with limited threads, like the UI thread. However, according to the [docs ](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-i-o.html), the use of `Dispatchers.IO` here might not even lead to a context switch:

> This dispatcher and its views share threads with the [Default](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-default.html) dispatcher, so using withContext(Dispatchers.IO) { ... } when already running on the [Default](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-default.html) dispatcher typically does not lead to an actual switching to another thread. In such scenarios, the underlying implementation attempts to keep the execution on the same thread on a best-effort basis.

sschuberth: So, anyone dares to approve?

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
```diff
@@ -268,3 +274,13 @@ interface ClearlyDefinedService {
...
+
+suspend fun ClearlyDefinedService.getCurationsChunked(
+    coordinates: Collection<Coordinates>,
+    chunkSize: Int = ClearlyDefinedService.MAX_REQUEST_CHUNK_SIZE
+): Map<Coordinates, Curation> =
+    buildMap {
+        coordinates.chunked(chunkSize).forEach { chunk ->
+            getCurations(chunk).values.forEach { putAll(it.curations) }
+        }
+    }
```
mnonnenmacher, line 296: Would it make sense to run the requests in parallel using a `map { async { ... }}.awaitAll()` construct?

sschuberth: I was asking myself the same thing as well but was undecided, so I wanted to wait whether it would come up during the review. As it does, I'll change it 😸

sschuberth: But wait, now I remember, it basically boils down to the same question: Should the function be blocking. Because `async` needs a `CoroutineScope`, so I'd have to wrap it in `runBlocking` again.

sschuberth: I've reworked the code quite a bit, please have another look.

fviernau (approved):

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
mnonnenmacher: You don't need `runBlocking` for that, you could have wrapped the async code with `coroutineScope {}` instead and kept it a suspend function.

sschuberth: Ah, right. Well, now the extension functions are higher level convenience functions that already do the blocking, which was anyway required by all current callers. So I believe it's ok that way, but should we ever need to change it again, we should do like you propose.

PR #6240, opened 2022-12-15
Dockerfile: Ensure that the `.ort` data directory exists

sschuberth:
It is common practice to mount a host-side ORT config directory into the Docker container at `/home/ort/.ort/config`. While Docker infers the permissions for the `config` directory from the corresponding directory on the host, for the intermediate `.ort` directory (which did not exist in the container prior to this change) there is no corresponding directory on the host, so Docker uses `root:root` ownership. Fix the ownership to be `ort:ort` by ensuring that the `.ort` directory exists in the container.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -129,7 +129,11 @@ jobs:
     - name: Run functional tests
       run: |
           echo "Running as $(id)."
-          BATECT_QUIET_DOWNLOAD=true ./batect --enable-buildkit --config-var gradle_build_scan=true --config-var gradle_console=plain funTestAnalyzer
+          BATECT_QUIET_DOWNLOAD=true ./batect --enable-buildkit \
+              --config-var docker_build_user_id=1001 \
+              --config-var gradle_build_scan=true \
+              --config-var gradle_console=plain \
+              funTestAnalyzer
     - name: Upload code coverage data
       uses: codecov/codecov-action@v3
       with:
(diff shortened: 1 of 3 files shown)


MarcelBochtler (approved):

tsteenbe (dismissed):
Not sure this is a good change, at minimum it should be a `mkdir -p` which fails silently e.g. `mkdir -p $HOMEDIR/.ort/ || :`

I have been running into Docker file ownership issues with GitHub action for ORT where the docker creates files in user 'ort' instead of the host user - we would not want the .ort dir to be created in the user's homedir that is not accessible by the user afterwards.

sschuberth:
> Not sure this is a good change

Why not? What's your concern? It has been tested to fix a real permission issue.

> at minimum it should be a `mkdir -p` which fails silently

Done.

sschuberth:
Hmm, this seem to indeed [break the CI tests](https://github.com/oss-review-toolkit/ort/actions/runs/3703038266/jobs/6274008283#step:6:643):

```
java.io.FileNotFoundException: /home/ort/.ort/cache/analyzer/maven/remote-artifacts/journal.tmp (No such file or directory)
```

Very weird, Docker drives my crazy :zany_face:

sschuberth:
> Hmm, this seem to indeed [break the CI tests](https://github.com/oss-review-toolkit/ort/actions/runs/3703038266/jobs/6274008283#step:6:643):

Fixed.

sschuberth:
> > Not sure this is a good change
> 
> Why not? What's your concern? It has been tested to fix a real permission issue.
> 
> > at minimum it should be a `mkdir -p` which fails silently
> 
> Done.

Any update @tsteenbe? Tests pass now.

mnonnenmacher (approved):

PR #6244, opened 2022-12-16
analyzer: Stop using a `SortedSet` in `ProjectAnalyzerResult`

sschuberth:
The `ProjectAnalyzerResult` is only serialized as part of analyzer functional tests when comparing to expected results. So in memory, there is no need for the `packages` to be a `SortedSet`. Change the code to only sort `packages` on serialization via a Jackson converter to have a defined order when comparing to expected results.

### DIFF
--- a/analyzer/src/funTest/kotlin/managers/Extensions.kt
+++ b/analyzer/src/funTest/kotlin/managers/Extensions.kt
@@ -83,7 +83,7 @@ fun PackageManagerResult.resolveScopes(projectResult: ProjectAnalyzerResult): Pr
     // all packages; this handles corner cases with package managers producing packages not assigned to project scopes.
     val packages = projectResult.packages.takeUnless { it.isEmpty() }
         ?: if (projectResults.size > 1) resolvedProject.filterReferencedPackages(sharedPackages) else sharedPackages
-    return projectResult.copy(project = resolvedProject, packages = packages.toSortedSet())
+    return projectResult.copy(project = resolvedProject, packages = packages)
 }
 
 /**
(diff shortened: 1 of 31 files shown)


model/src/main/kotlin/utils/SortedSetConverter.kt
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model.utils
+
+import com.fasterxml.jackson.databind.util.StdConverter
+
+class SortedSetConverter<T : Comparable<T>> : StdConverter<Set<T>, Set<T>>() {
```
sschuberth, line 24: For now, this is a generic implementation that works on `Comparable<T>`, which has the downside that the classes still need to implement `Comparable`, and the current implementations are confusing as they're not in line with `equals()`.

So maybe we should prefer to have dedicated converters per class that we want to sort as these could hard-code the property to sort on then, like

class SortedSetConverter : StdConverter<Set<Package>, Set<Package>>() {
    override fun convert(value: Set<Package>): Set<Package> = value.toSortedSet(compareBy { it.id })
}

That would allow us to remove the `Comparable` implementations in exchange for several more converted classes. But as there are only ~10 implementations that we need to care about as they influence serialization, that's maybe still a good idea.

sschuberth:
FYI @mnonnenmacher, a `SortedSet` keeps only the first entry that's equal according to `compareTo`, see https://pl.kotl.in/zzR5V7Npo.

Which means that this change may uncover previously unnoticed cases where packages with the same id but different other properties are being added to the result.

fviernau (dismissed):

model/src/main/kotlin/utils/SortedSetConverter.kt
sschuberth: > So maybe we should prefer to have dedicated converters per class [...] That would allow us to remove the `Comparable` implementations in exchange for several more converted classes.

That's what I did now.

mnonnenmacher (approved):

PR #6246, opened 2022-12-17
Miscellanous console output tweaking

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
+++ b/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
@@ -131,7 +131,7 @@ class DefaultPackageProvenanceResolver(
 
             else -> {
                 logger.info {
-                    "Could not find a stored artifact resolution result for package '${pkg.id.toCoordinates()}'," +
+                    "Could not find a stored artifact resolution result for package '${pkg.id.toCoordinates()}', " +
                             "attempting resolution."
                 }
             }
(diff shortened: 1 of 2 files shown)


utils/common/src/main/kotlin/ProcessCapture.kt
```diff
@@ -129,9 +129,7 @@ class ProcessCapture(
         }
 
     init {
-        logger.info {
-            "Running '$commandLine' in '$usedWorkingDir'..."
-        }
+        logger.debug { "Running '$commandLine' in '$usedWorkingDir'..." }
```
mnonnenmacher, line 132: I find this to be one of the most useful log lines when debugging failed runs, I wouldn't want it to appear only in debug.

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
```diff
@@ -131,7 +131,7 @@ class DefaultPackageProvenanceResolver(
 
             else -> {
                 logger.info {
-                    "Could not find a stored artifact resolution result for package '${pkg.id.toCoordinates()}'," +
+                    "Could not find a stored artifact resolution result for package '${pkg.id.toCoordinates()}', " +
```
mnonnenmacher, line 134: Commit message nit: "after", not "before"

mnonnenmacher (changes requested):

utils/common/src/main/kotlin/ProcessCapture.kt
sschuberth: Ok, I've dropped it.

mnonnenmacher (approved):

sschuberth:

PR #6249, opened 2022-12-18
EvaluatorCommand: Fix the check if package curations should be replaced

mnonnenmacher:
Use the `packageCurationsDir` and `packageCurationsFile` options to check if the curations should be replaced, instead of the unrelated `packageConfigurationOption`.

This is fixup for 84ef89f.

### DIFF
--- a/cli/src/main/kotlin/commands/EvaluatorCommand.kt
+++ b/cli/src/main/kotlin/commands/EvaluatorCommand.kt
@@ -270,7 +270,7 @@ class EvaluatorCommand : OrtCommand(
             ortResultInput = ortResultInput.replaceConfig(config)
         }
 
-        if (packageConfigurationOption != null) {
+        if (packageCurationsDir != null || packageCurationsFile != null) {
             val curations = FilePackageCurationProvider.from(packageCurationsFile, packageCurationsDir).packageCurations
             ortResultInput = ortResultInput.replacePackageCurations(curations)
         }


cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -270,7 +270,7 @@ class EvaluatorCommand : OrtCommand(
             ortResultInput = ortResultInput.replaceConfig(config)
         }
 
-        if (packageConfigurationOption != null) {
+        if (packageCurationsDir != null || packageCurationsFile != null) {
```
sschuberth, line 273: Actually, that's how @fviernau had it before [my review comment](https://github.com/oss-review-toolkit/ort/pull/5994#discussion_r1005498327). Does the current code not work? Because if it does, I still believe it reads nicer to emphasize that you want to do something as soon as any of the mutually exclusive options is specified.

mnonnenmacher: The current code is wrong, `packageConfigurationOption` is for `PackageConfiguration` providers, but this code is about replacing `PackageCuration`s. For those there is no `packageCurationOption`, just the two separate options.

sschuberth: Ah, right, a *configuration* vs. *curation* mixup.

sschuberth (approved):

fviernau (approved):

PR #6254, opened 2022-12-19
Conan: Add support for Conan lockfiles

yarosevych:
Require a lockfile if a path to the lockfile is configured in `config.yml`.

Only lockfiles located in the same directory as the definition file are
supported.

The test for Conan lockfiles cannot complete successfully when ran
locally as `package id` differs depending on operating system. The
`package id` in the test is set to the one calculated on Linux.

Signed-off-by: Oleksandr Yarosevych <fixed-term.oleksandr.yarosevych@bosch.io>

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/conan-py/lockfile.lock
+++ b/analyzer/src/funTest/assets/projects/synthetic/conan-py/lockfile.lock
@@ -1,0 +1,33 @@
+{
+ "graph_lock": {
+  "nodes": {
+   "0": {
+    "ref": "Poco/1.9.3",
+    "options": "cxx_14=False\nenable_apacheconnector=False\nenable_cppparser=False\nenable_crypto=True\nenable_data=True\nenable_data_mysql=False\nenable_data_odbc=False\nenable_data_sqlite=True\nenable_json=True\nenable_mongodb=True\nenable_net=True\nenable_netssl=True\nenable_netssl_win=True\nenable_pagecompiler=False\nenable_pagecompiler_file2page=False\nenable_pdf=False\nenable_pocodoc=False\nenable_redis=True\nenable_sevenzip=False\nenable_tests=False\nenable_util=True\nenable_xml=True\nenable_zip=True\nforce_openssl=True\npoco_unbundled=False\nshared=False\nopenssl:386=False\nopenssl:capieng_dialog=False\nopenssl:enable_capieng=False\nopenssl:enable_weak_ssl_ciphers=False\nopenssl:no_aria=False\nopenssl:no_asm=False\nopenssl:no_async=False\nopenssl:no_bf=False\nopenssl:no_blake2=False\nopenssl:no_camellia=False\nopenssl:no_cast=False\nopenssl:no_chacha=False\nopenssl:no_cms=False\nopenssl:no_comp=False\nopenssl:no_ct=False\nopenssl:no_deprecated=False\nopenssl:no_des=False\nopenssl:no_dgram=False\nopenssl:no_dh=False\nopenssl:no_dsa=False\nopenssl:no_dso=False\nopenssl:no_ec=False\nopenssl:no_ecdh=False\nopenssl:no_ecdsa=False\nopenssl:no_engine=False\nopenssl:no_filenames=False\nopenssl:no_fips=False\nopenssl:no_gost=False\nopenssl:no_idea=False\nopenssl:no_legacy=False\nopenssl:no_md2=True\nopenssl:no_md4=False\nopenssl:no_mdc2=False\nopenssl:no_ocsp=False\nopenssl:no_pinshared=False\nopenssl:no_rc2=False\nopenssl:no_rc4=False\nopenssl:no_rc5=False\nopenssl:no_rfc3779=False\nopenssl:no_rmd160=False\nopenssl:no_seed=False\nopenssl:no_sm2=False\nopenssl:no_sm3=False\nopenssl:no_sm4=False\nopenssl:no_sock=False\nopenssl:no_srp=False\nopenssl:no_srtp=False\nopenssl:no_sse2=False\nopenssl:no_ssl=False\nopenssl:no_ssl3=False\nopenssl:no_stdio=False\nopenssl:no_threads=False\nopenssl:no_tls1=False\nopenssl:no_ts=False\nopenssl:no_whirlpool=False\nopenssl:no_zlib=False\nopenssl:openssldir=None\nopenssl:shared=False\nzlib:shared=False",
+    "requires": [
+     "1"
+    ],
+    "path": "conanfile.py",
+    "context": "host"
+   },
+   "1": {
+    "ref": "openssl/3.0.0",
+    "options": "386=False\ncapieng_dialog=False\nenable_capieng=False\nenable_weak_ssl_ciphers=False\nno_aria=False\nno_asm=False\nno_async=False\nno_bf=False\nno_blake2=False\nno_camellia=False\nno_cast=False\nno_chacha=False\nno_cms=False\nno_comp=False\nno_ct=False\nno_deprecated=False\nno_des=False\nno_dgram=False\nno_dh=False\nno_dsa=False\nno_dso=False\nno_ec=False\nno_ecdh=False\nno_ecdsa=False\nno_engine=False\nno_filenames=False\nno_fips=False\nno_gost=False\nno_idea=False\nno_legacy=False\nno_md2=True\nno_md4=False\nno_mdc2=False\nno_ocsp=False\nno_pinshared=False\nno_rc2=False\nno_rc4=False\nno_rc5=False\nno_rfc3779=False\nno_rmd160=False\nno_seed=False\nno_sm2=False\nno_sm3=False\nno_sm4=False\nno_sock=False\nno_srp=False\nno_srtp=False\nno_sse2=False\nno_ssl=False\nno_ssl3=False\nno_stdio=False\nno_threads=False\nno_tls1=False\nno_ts=False\nno_whirlpool=False\nno_zlib=False\nopenssldir=None\nshared=False\nzlib:shared=False",
+    "package_id": "73fc7f3ed0faf0fd80110346c52171985a5794b4",
+    "requires": [
+     "2"
+    ],
+    "context": "host"
+   },
+   "2": {
+    "ref": "zlib/1.2.12",
+    "options": "shared=False",
+    "package_id": "7bc8c2c85db7a618e5320dc997f27fc33e1df074",
+    "context": "host"
+   }
+  },
+  "revisions_enabled": false
+ },
+ "version": "0.4",
+ "profile_host": "[settings]\narch=x86_64\narch_build=x86_64\nbuild_type=Release\ncompiler=gcc\ncompiler.libcxx=libstdc++\ncompiler.version=6.3\nos=Windows\nos_build=Windows\n[options]\n[build_requires]\n[env]\n"
+}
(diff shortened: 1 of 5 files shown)


analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -144,25 +142,47 @@ class Conan(
         stashDirectories(directoryToStash).use {
             configureRemoteAuthentication(conanConfig)
 
+            // TODO: Support lockfiles which are located in a different directory than the definition file.
+            val lockfileName = analyzerConfig.packageManagers?.get(managerName)?.options?.get("lockfileName")
```
MarcelBochtler, line 146: Please document the `lockfileName` configuration option.
You can do this in the reference.yml in the package manager section:
https://github.com/oss-review-toolkit/ort/blob/509aeb2b3d21a2778e668e018401b448192c3e3f/model/src/main/resources/reference.yml#L79

MarcelBochtler (changes requested):

analyzer/src/funTest/assets/projects/synthetic/conan-py/lockfile.lock
```diff
@@ -0,0 +1,33 @@
...
+    "options": "shared=False",
+    "package_id": "7bc8c2c85db7a618e5320dc997f27fc33e1df074",
+    "context": "host"
+   }
+  },
+  "revisions_enabled": false
+ },
+ "version": "0.4",
+ "profile_host": "[settings]\narch=x86_64\narch_build=x86_64\nbuild_type=Release\ncompiler=gcc\ncompiler.libcxx=libstdc++\ncompiler.version=6.3\nos=Windows\nos_build=Windows\n[options]\n[build_requires]\n[env]\n"
+}
```
MarcelBochtler, line 33: Please add an empty line at the end of this file.

MarcelBochtler (changes requested):

MarcelBochtler (approved):
LGTM.
@sschuberth I think the commitlint check fails because @yarosevych's sign-off is too long. Do you know of a way to allow this?

sschuberth:
> LGTM. @sschuberth I think the commitlint check fails because @yarosevych's sign-off is too long. Do you know of a way to allow this?

It should [default to ` Infinity`](https://github.com/conventional-changelog/commitlint/blob/master/docs/reference-rules.md#footer-max-line-length), strange. Can you try explicitly setting it as part of this PR?

analyzer/src/main/kotlin/managers/Conan.kt
mnonnenmacher: The option also needs to be documented in the docs of the Conan class, for an example see `Pip.kt`.

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -106,9 +106,7 @@ class Conan(
 
     override fun command(workingDir: File?) = "conan"
 
-    // TODO: Add support for Conan lock files.
-
-    // protected open fun hasLockFile(projectDir: File) = null
+    private fun hasLockFile(directory: String) = File(directory).isFile
```
mnonnenmacher, line 109: Parameter should be called `file` instead of `directory`.

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -144,25 +142,47 @@ class Conan(
         stashDirectories(directoryToStash).use {
             configureRemoteAuthentication(conanConfig)
 
+            // TODO: Support lockfiles which are located in a different directory than the definition file.
+            val lockfileName = analyzerConfig.packageManagers?.get(managerName)?.options?.get("lockfileName")
+            requireLockfile(workingDir) { lockfileName?.let { hasLockFile(workingDir.resolve(it).path) } ?: false }
```
mnonnenmacher, line 147: I think we should verify that the provided lockfile is below the `analysisRoot` directory to prevent that users reference files which do not belong to the project. You can see d45a824 for an example where a similar feature was implemented for the `ArchiveUtils` class.

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -144,25 +142,47 @@ class Conan(
...
+                val lockfileNodes = jsonMapper.readTree(definitionFile.resolveSibling(lockfileName))
+                val lockfilePackages = lockfileNodes["graph_lock"]["nodes"].mapNotNull {
+                    it["ref"]?.textValue()
+                }
+
+                jsonMapper.createArrayNode().addAll(
+                    pkgInfos.filter { jsonNode ->
+                        lockfilePackages.contains(jsonNode["reference"].textValueOrEmpty())
+                                || jsonNode["reference"].textValueOrEmpty().contains(definitionFile.name)
+                    }
```
mnonnenmacher, line 165: Can you explain why this was implemented as a filter instead of using the `--lockfile` option of the `conan info` command?

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/Conan.kt
yarosevych: @mnonnenmacher you are right. I must have overlooked this because when I was testing conan with a complex example project that had a lot of interconnected dependencies, it gave me an impression that `-l` option is used to create the lockfile instead of reading it. I will make the change to this PR.

analyzer/src/main/kotlin/managers/Conan.kt
yarosevych: @mnonnenmacher Implemented, see `verifyLockfileBelongsToProject` function

analyzer/src/main/kotlin/managers/Conan.kt
yarosevych: @mnonnenmacher the change has been done

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -186,6 +201,14 @@ class Conan(
         }
     }
 
+    private fun verifyLockfileBelongsToProject(workingDir: File, lockfileName: String?) {
+        require(workingDir.resolve(lockfileName.orEmpty()).canonicalFile.startsWith(workingDir.canonicalFile)) {
+            "The provided lockfile path points to the directory outside of the analysed project: '$lockfileName' and " +
```
mnonnenmacher, line 206: Nit "analy**z**ed"

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -144,8 +147,18 @@ class Conan(
...
+
             val jsonFile = createOrtTempDir().resolve("info.json")
-            run(workingDir, "info", definitionFile.name, "--json", jsonFile.absolutePath, *DUMMY_COMPILER_SETTINGS)
+            if (!analyzerConfig.allowDynamicVersions && lockfileName != null) {
+                verifyLockfileBelongsToProject(workingDir, lockfileName)
+                run(workingDir, "info", definitionFile.name, "-l", lockfileName, "--json", jsonFile.absolutePath)
+            } else {
+                run(workingDir, "info", definitionFile.name, "--json", jsonFile.absolutePath, *DUMMY_COMPILER_SETTINGS
+                )
+            }
```
mnonnenmacher, line 161: If `allowDynamicVersion == true` and the lockfile is set it should still be used. Best reuse the `requireLockfile` function from `PackageManager`, see e.g. `Pub` for how it is used, and change the logic like that:
```kotlin
requireLockfile(...)

if (lockfileName != null) {
  run(...)
} else {
  run(...)
}
```

mnonnenmacher (changes requested):

yarosevych:
> > @sschuberth I think the commitlint check fails because @yarosevych's sign-off is too long. Do you know of a way to allow this?
> 
> It should [default to `Infinity`](https://github.com/conventional-changelog/commitlint/blob/master/docs/reference-rules.md#footer-max-line-length), strange. Can you try explicitly setting it as part of this PR?

This issue is caused by commitlint treating the "Signed-off-by" footer as a commit body line, according to the error: `body's lines must not be longer than 75 characters [body-max-line-length]`. The issue is reported here: https://github.com/conventional-changelog/commitlint/issues/3411

mnonnenmacher:
> > > @sschuberth I think the commitlint check fails because @yarosevych's sign-off is too long. Do you know of a way to allow this?
> > 
> > 
> > It should [default to `Infinity`](https://github.com/conventional-changelog/commitlint/blob/master/docs/reference-rules.md#footer-max-line-length), strange. Can you try explicitly setting it as part of this PR?
> 
> This issue is caused by commitlint treating the "Signed-off-by" footer as a commit body line, according to the error: `body's lines must not be longer than 75 characters [body-max-line-length]`. The issue is reported here: [conventional-changelog/commitlint#3411](https://github.com/conventional-changelog/commitlint/issues/3411)

@yarosevych I have added a commit to #6366 that configures commitlint to tread "Signed-off-by" lines as footers, you can rebase once the PR is merged.

analyzer/src/funTest/kotlin/managers/ConanFunTest.kt
```diff
@@ -71,12 +72,48 @@ class ConanFunTest : StringSpec() {
...
+            val result = createConanDynamicVersions().resolveSingleProject(definitionFile)
+
+            patchActualResult(result.toYaml()) shouldBe expectedResult
+        }
+
+        /**
+         * This test cannot complete successfully when you run it locally as `package id` differs depending on operating
+         * system. The `package id` is set to the one calculated on Linux.
+         */
+        "Project dependencies are detected correctly with the lockfile" {
```
mnonnenmacher, line 84: Please configure the test so that it only runs on Linux:
```
"Project dependencies are detected correctly with the lockfile".config(enabled = Os.isLinux) {
```
(using `org.ossreviewtoolkit.utils.common.Os`)

analyzer/src/main/kotlin/managers/Conan.kt
```diff
@@ -61,6 +63,9 @@ import org.semver4j.RangesListFactory
 /**
  * The [Conan](https://conan.io/) package manager for C / C++.
  *
+ * This package manager supports the following [options][PackageManagerConfiguration.options]:
+ * - *lockfileName*: the name of the lockfile, which is used for analysis if allowDynamicVersions is set to false.
```
mnonnenmacher, line 67: Nit: Please capitalize "The".

mnonnenmacher (changes requested):

analyzer/src/funTest/kotlin/managers/SpdxDocumentFileFunTest.kt
```diff
@@ -202,7 +202,7 @@ class SpdxDocumentFileFunTest : WordSpec({
                 revision = vcsRevision
             )
 
-            val analyzer = Analyzer(AnalyzerConfiguration())
+            val analyzer = Analyzer(AnalyzerConfiguration(true))
```
mnonnenmacher, line 205: Please use a named argument when passing a boolean for readability.

mnonnenmacher (changes requested):

analyzer/src/funTest/kotlin/managers/SpdxDocumentFileFunTest.kt
sschuberth: In Kotlin, [named arguments](https://kotlinlang.org/docs/functions.html#named-arguments) mean to write `allowDynamicVersions = true` directly in the list of arguments in line 205.

yarosevych: @sschuberth thanks for the tip, corrected.

mnonnenmacher (approved):

sschuberth:
Merging despite the unrelated SW360 test failures, see https://www.eclipsestatus.io/.

PR #6257, opened 2022-12-20
analyzer: Make `ClearlyDefinedPackageCurationProviderTest` a `funTest`

sschuberth:
The test talks to the real servers and thus should be a functional test. This makes room for `ClearlyDefinedPackageCurationProviderMockTest` to drop "Mock" from its name in turn.


analyzer/src/test/kotlin/curation/ClearlyDefinedPackageCurationProviderTest.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ * Copyright (C) 2021 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
```
mnonnenmacher, line 2: Maybe do this in separate commits for a cleaner diff?

mnonnenmacher (approved):

PR #6260, opened 2022-12-20
OkHttpClientHelperTest: Add a test for getting different clients

sschuberth:

### DIFF
--- a/utils/ort/src/test/kotlin/OkHttpClientHelperTest.kt
+++ b/utils/ort/src/test/kotlin/OkHttpClientHelperTest.kt
@@ -26,6 +26,7 @@ import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.types.beInstanceOf
 import io.kotest.matchers.types.shouldBeSameInstanceAs
+import io.kotest.matchers.types.shouldNotBeSameInstanceAs
 
 import io.mockk.every
 import io.mockk.mockk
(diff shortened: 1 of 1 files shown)


utils/ort/src/test/kotlin/OkHttpClientHelperTest.kt
```diff
@@ -58,6 +59,24 @@ class OkHttpClientHelperTest : WordSpec({
...
+        "return different clients when specifying different configurations" {
+            val timeoutA: BuilderConfiguration = { readTimeout(Duration.ofSeconds(100)) }
+            val timeoutB: BuilderConfiguration = { readTimeout(Duration.ofSeconds(500)) }
+            val clientA = OkHttpClientHelper.buildClient(timeoutA)
+            val clientB = OkHttpClientHelper.buildClient(timeoutB)
+
+            clientA shouldNotBeSameInstanceAs clientB
+        }
+
+        "return different clients when specifying different configurations instances with the same values" {
```
fviernau, line 72: configurations instances -> configuration instances?

utils/ort/src/test/kotlin/OkHttpClientHelperTest.kt
```diff
@@ -58,6 +59,24 @@ class OkHttpClientHelperTest : WordSpec({
 
             clientA shouldBeSameInstanceAs clientB
         }
+
+        "return different clients when specifying different configurations" {
```
fviernau, line 63: Would it make sense to say here: different configuration instances with same values?

(To emphasize that it is just about instance equality, not about the equality of the values it sets).

fviernau (changes requested):

utils/ort/src/test/kotlin/OkHttpClientHelperTest.kt
sschuberth: > different configuration instances with same values

But that's the test case below, isn't it?

utils/ort/src/test/kotlin/OkHttpClientHelperTest.kt
sschuberth: I've reworded this yet more to emphasize that differently clients are being returned despite the configuration being the same (value-wise).

utils/ort/src/test/kotlin/OkHttpClientHelperTest.kt
fviernau: The proposal was for the one in line 63 - anyhow, leaving it up to you.

fviernau (approved):

utils/ort/src/test/kotlin/OkHttpClientHelperTest.kt
sschuberth: > The proposal was for the one in line 63 - anyhow, leaving it up to you.

What I was trying to say is that IMO it's incorrect to rename the test in line 63 to "different configuration instances with same values" as that's a different case that's already tested in line 72.

PR #6263, opened 2022-12-20
common-utils: Make safeMkdirs fluent

mnonnenmacher:
Make the `safeMkdirs` function fluent because it allows nicer code in some places.


utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -161,12 +160,12 @@ fun File.safeDeleteRecursively(force: Boolean = false, baseDirectory: File? = nu
  *
  * @throws IOException if any missing directory could not be created.
  */
-fun File.safeMkdirs() {
+fun File.safeMkdirs(): File {
```
MarcelBochtler, line 163: Maybe add it to the documentation that the created `File` is returned.

fviernau: Just a hint: it may be that multiple files (dirs) get created.

sschuberth (approved):

PR #6267, opened 2022-12-21
HttpFileStorage: Further increase the connect timeout

fviernau:


utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
```diff
@@ -34,7 +34,7 @@ import org.apache.logging.log4j.kotlin.Logging
 import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.execute
 
-private const val HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = 30L
+private const val HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = 60L
```
sschuberth, line 37: This timeout is massive now. While it probably has no negative impact on fast networks, maybe we should consider making this configurable in the future.

sschuberth (approved):

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
fviernau: > This timeout is massive now. While it probably has no negative impact on fast networks, maybe we should consider making this configurable in the future.

Sorry, I clicked the "create PR" button on the wrong repo by accident. 
I didn't intend to create a PR, I pushed this only for debugging remotely, I'll make it a draft for now.

PR #6270, opened 2022-12-21
AdvisorCommand: Print a summary of how many packages are vulnerable

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/AdvisorCommand.kt
+++ b/cli/src/main/kotlin/commands/AdvisorCommand.kt
@@ -136,6 +136,16 @@ class AdvisorCommand : OrtCommand(
             throw ProgramResult(1)
         }
 
+        with(advisorResults.getVulnerabilities()) {
+            val totalPackageCount = ortResultOutput.getPackages(omitExcluded = true).size
+            val vulnerabilityCount = values.sumOf { it.size }
+
+            println(
+                "$size of $totalPackageCount package(s) (not counting excluded ones) are vulnerable, with " +
+                        "$vulnerabilityCount vulnerabilities in total."
+            )
+        }
+
         val resolutionProvider = DefaultResolutionProvider.create(ortResultOutput, resolutionsFile)
         val (resolvedIssues, unresolvedIssues) =
             advisorResults.collectIssues().flatMap { it.value }.partition { resolutionProvider.isResolved(it) }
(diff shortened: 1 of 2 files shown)


cli/src/main/kotlin/commands/AdvisorCommand.kt
```diff
@@ -136,6 +136,10 @@ class AdvisorCommand : OrtCommand(
             throw ProgramResult(1)
         }
 
+        val vulnerablePackageCount = advisorResults.getVulnerabilities().size
+        val totalPackageCount = ortResultOutput.getPackages().size
```
fviernau, line 140: The advisor may not have queried excluded packages, so the value for `totalPcakageCount` may be mis-leading the user into thinking that all these packages have been queried for vulnerabilities. So, should `skipExcluded` be passed to `getPackages()` ?

Apart from that, I wonder if total package count does matter at all or whether it should be dropped.
IMO adding the amount of distinct vulnerabilites or non-distinct vulnerabilities would be more interresting than total package count.

fviernau (changes requested):

cli/src/main/kotlin/commands/AdvisorCommand.kt
sschuberth: > So, should `skipExcluded` be passed to `getPackages()` ?

I've now simply aligned the wording with the one in the `AnalyzerCommand`: https://github.com/oss-review-toolkit/ort/blob/2e216452db7db7c984306ff1d4d4f875ca1f2067/cli/src/main/kotlin/commands/AnalyzerCommand.kt#L284

> Apart from that, I wonder if total package count does matter at all or whether it should be dropped.

I believe it's informative to know what share of your packages are vulnerable.

> IMO adding the amount of distinct vulnerabilites or non-distinct vulnerabilities would be more interresting

I'm now also mentioning the total number of vulnerabilities, so one can quickly see if there's more than one vulnerability per package.

cli/src/main/kotlin/commands/AdvisorCommand.kt
```diff
@@ -136,6 +136,16 @@ class AdvisorCommand : OrtCommand(
             throw ProgramResult(1)
         }
 
+        with(advisorResults.getVulnerabilities()) {
+            val totalPackageCount = ortResultOutput.getPackages().size
```
fviernau, line 140: this is counting excluded packages, so it contradicts the log message, see: OrtResult.getPackages(omitExcluded: Boolean = false)

fviernau (changes requested):

cli/src/main/kotlin/commands/AdvisorCommand.kt
sschuberth: Phew, this was also broken for the `AnalyzerCommand`. Thanks, both fixed.

fviernau (approved):

PR #6275, opened 2022-12-22
HttpFileStorage: Reduce keep alive duration

fviernau:
See individual commits.

### DIFF
--- a/utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
+++ b/utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
@@ -25,6 +25,7 @@ import java.time.Duration
 import java.util.concurrent.TimeUnit
 
 import okhttp3.CacheControl
+import okhttp3.ConnectionPool
 import okhttp3.Headers.Companion.toHeaders
 import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
@@ -35,6 +36,8 @@ import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.execute
 
 private const val HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = 30L
+private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 5 * 60L
+private const val HTTP_CLIENT_MAX_IDLE_CONNECTIONS = 5
 
 /**
  * A [FileStorage] that stores files on an HTTP server.
@@ -68,6 +71,13 @@ class HttpFileStorage(
 
     private val httpClient by lazy {
         OkHttpClientHelper.buildClient {
+            val connectionPool = ConnectionPool(
+                HTTP_CLIENT_MAX_IDLE_CONNECTIONS,
+                HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS,
+                TimeUnit.SECONDS
+            )
+
+            connectionPool(connectionPool)
             connectTimeout(Duration.ofSeconds(HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS))
         }
     }
--- a/utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
+++ b/utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
@@ -36,7 +36,7 @@ import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.execute
 
 private const val HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = 30L
-private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 5 * 60L
+private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 1 * 60L
 private const val HTTP_CLIENT_MAX_IDLE_CONNECTIONS = 5
 
 /**


utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
```diff
@@ -35,6 +36,8 @@ import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.execute
 
 private const val HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = 30L
+private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 5 * 60L
+private const val HTTP_CLIENT_MAX_IDLE_CONNECTIONS = 5
```
sschuberth, line 40: It took me a while to understand that you're just repeating the default values here, in order to change to a non-default keep-alive duration only in the next commit. This could either be made explicit in the commit message, or (preferably) just squash the commits and say right away that you're customizing the keep-alive duration.

Speaking of the commit message, I find it also a bit confusing to refer to [1] twice in the same sentence.

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
```diff
@@ -68,6 +71,13 @@ class HttpFileStorage(
 
     private val httpClient by lazy {
         OkHttpClientHelper.buildClient {
+            val connectionPool = ConnectionPool(
+                HTTP_CLIENT_MAX_IDLE_CONNECTIONS,
+                HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS,
+                TimeUnit.SECONDS
```
sschuberth, line 77: Sticking to the default of `TimeUnit.MINUTES` would make sense here to avoid the multiplication in `HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS` above.

sschuberth (changes requested):

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
fviernau: I've adjusted the commit message accordingly.

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
fviernau: I prefer seconds as this allows a more fine grained adjustment. (In case the error should again pop up, I'd probably further reduce the value and then likely make it configurable). Apart from that the value is consistent with `HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS ` which is defined right next to it. I don't really see an issue with a multiplication.

Would you be ok with keeping this as-is?

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
```diff
@@ -35,6 +36,8 @@ import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.execute
 
 private const val HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = 30L
+private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 5 * 60L
```
sschuberth: Commit message: The "While at it," part as if you're doing another thing at the same time, but in fact passing the default constructor arguments explicitly *is* the way to stop sharing the connection pool. So I'd simply say "Stop sharing the connection pool with HTTP clients used elsewhere by passing the default constructor parameter values explicitly". Also, `IOException` should be put in backticks as well.

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
sschuberth: > The multiplication would get removed by the compiler anyway

This is not about performance, but about human readability.

> Would you be ok with keeping this as-is?

I think it's a bit odd, esp. as you otherwise say to repeat the default arguments, which strictly speaking you don't, as the `timeUnit` changes from `TimeUnit.MINUTES` to `TimeUnit.SECONDS`. But I believe it's not worth wasting more time by debating over it, so just keep it.

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
```diff
@@ -36,7 +36,7 @@ import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.execute
 
 private const val HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = 30L
-private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 5 * 60L
+private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 1 * 60L
```
sschuberth, line 39: Can you please share some details in the commit message about *why* / *how* this is actually fixing the underlying issue? It's completely unclear to me how the keep alive duration of the connection pool can have such an impact.

sschuberth (changes requested):

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
fviernau: > The "While at it," part as if you're doing another thing at the same time

I wrote "while at it" because I'm actually doing two things at the same time:

1. Using a separate connection pool
2. passing default constructor arguments explicitly

...because 1 could also be done without 2. 
Anyhow, changing the commit message to how you want it to be.

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
fviernau: > This is not about performance, but about human readability.

Ok, so I removed the multiplication to address this.

sschuberth: The multiplication / `TimeUnit.SECONDS` is this there, but like I said, let's keep it like this then.

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
```diff
@@ -36,7 +36,7 @@ import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.execute
 
 private const val HTTP_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = 30L
-private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 5 * 60L
+private const val HTTP_CLIENT_KEEP_ALIVE_DURATION_IN_SECONDS = 60L
```
sschuberth, line 39: Ah, so you were dropping the multiplication *here*. That's not what I meant. I meant that the first commit should use `TimeUnit.MINUTES` and thus no `* 60L` in the constant.

sschuberth: Also, there are typos in "observeed" and "preceeding". Sorry for not noticing earlier.

sschuberth (changes requested):

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
fviernau: > Ah, so you were dropping the multiplication _here_. That's not what I meant. I meant that the first commit should use `TimeUnit.MINUTES` and thus no `* 60L` in the constant.

I would go your proposed route to say it's not worth discussing any further if this still is an option: https://github.com/oss-review-toolkit/ort/pull/6275#discussion_r1056101815.

sschuberth (approved):

PR #6282, opened 2022-12-27
Remove the CLI's `GlobalOptions`

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -171,13 +171,13 @@ class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
         printStackTrace = stacktrace
 
         // Make options available to subcommands and apply static configuration.
-        val ortConfiguration = OrtConfiguration.load(configArguments, configFile)
-        currentContext.findOrSetObject { GlobalOptions(ortConfiguration, forceOverwrite) }
-        LicenseFilePatterns.configure(ortConfiguration.licenseFilePatterns)
+        val ortConfig = OrtConfiguration.load(configArguments, configFile)
+        currentContext.findOrSetObject { GlobalOptions(ortConfig, forceOverwrite) }
+        LicenseFilePatterns.configure(ortConfig.licenseFilePatterns)
 
         EnvironmentVariableFilter.reset(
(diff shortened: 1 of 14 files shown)


model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -67,6 +67,11 @@ data class OrtConfiguration(
      */
     val enableRepositoryPackageCurations: Boolean = false,
 
+    /**
+     * Force overwriting if any existing output files.
```
mnonnenmacher, line 71: Typo: "if" -> "of"

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -171,8 +168,11 @@ class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
         printStackTrace = stacktrace
 
         // Make options available to subcommands and apply static configuration.
-        val ortConfig = OrtConfiguration.load(configArguments, configFile)
-        currentContext.findOrSetObject { GlobalOptions(ortConfig, forceOverwrite) }
+        val ortConfig = OrtConfiguration.load(
+            args = mapOf("ort.forceOverwrite" to "$forceOverwrite") + configArguments,
```
mnonnenmacher, line 172: How does this behave if the option is set in the configuration file? Does it always overwrite (even if the option is not set on the command line) or never?

mnonnenmacher (changes requested):

cli/src/main/kotlin/OrtMain.kt
sschuberth: Good point, yes, fixed.

mnonnenmacher: Should the order not be the other way round, as if I understand that correctly the value from the config file would now always override the command line option?

```kotlin
putAll(configArguments)
if (forceOverwrite) put("ort.forceOverwrite", "true")
```

sschuberth: No, `configArguments` only contains the `-P`-style configuration arguments given on the command line, not any configuration from `config.yml` (that's passed in the next line via `configFile`). And as `--force-overwrite` is being deprecated in favor of `-P ort.forceOverwrite=...`, the latter should override the former syntax. That's why I believe the order to be correct.

mnonnenmacher (approved):

PR #6290, opened 2022-12-29
Improved vulnerability report

sschuberth:


sschuberth:
Example for how a report generated with the code in `main` looks:

[AsciiDoc_vulnerability_report-main.pdf](https://github.com/oss-review-toolkit/ort/files/10321888/AsciiDoc_vulnerability_report-main.pdf)

And for comparison a report for the same data with the code form this PR:

[AsciiDoc_vulnerability_report-pr.pdf](https://github.com/oss-review-toolkit/ort/files/10321889/AsciiDoc_vulnerability_report-pr.pdf)

reporter/src/main/resources/templates/asciidoc/pdf-theme.yml
```diff
@@ -0,0 +1,4 @@
+extends: default
+role:
+  alert:
+    font-color: #ff0000
```
mnonnenmacher, line 18: Consider a "friendlier" color like #cc0000.

reporter/src/main/resources/templates/asciidoc/vulnerability_report.ftl
```diff
@@ -23,7 +23,11 @@
 :sectnums:
 :toc:
 
-= Vulnerability Report
+[#assign firstProjectId = projects?first.id]
+[#assign firstProjectUrl = ortResult.getProject(firstProjectId).vcsProcessed.url]
```
mnonnenmacher, line 31: Commit message: "Use the **first** project's name..." to emphasize that the choice is a bit arbitrary.
Off-topic, but I think we need a way to set a name for an ORT run, would be useful here and in many other places. Could default to the Git repository or directory name, for example.

reporter/src/main/resources/templates/asciidoc/vulnerability_report.ftl
```diff
@@ -46,9 +46,7 @@ lack relevant vulnerabilities. Further details about the issues that occurred ca
 == Packages
 [#assign advisorResults = helper.advisorResultsWithVulnerabilities()]
 [#list advisorResults as id, results]
-=== ${id.name}
-
-Package URL: _${ModelExtensions.toPurl(id)}_
+=== ${ModelExtensions.toPurl(id)}
```
mnonnenmacher, line 49: Commit message: "instead **of**"
For consistency, this should actually find the package metadata by id and use the purl from there, because it could have been set in a curation. But as it was like that before that's fine with me.

mnonnenmacher (changes requested):

reporter/src/main/resources/templates/asciidoc/pdf-theme.yml
sschuberth: I was aligning with

https://github.com/oss-review-toolkit/ort/blob/9c48a8274b4196cc74b3929c3e620e86aa9fea5d/examples/asciidoctor-pdf-theme.yml#L31-L33

but I'm planning to remove that example file later anyway, so I'll change the color here.

reporter/src/main/resources/templates/asciidoc/vulnerability_report.ftl
sschuberth: > Commit message: "Use the **first** project's name..." to emphasize that the choice is a bit arbitrary.

Done.

> Off-topic, but I think we need a way to set a name for an ORT run, would be useful here and in many other places. Could default to the Git repository or directory name, for example.

Agreed. This relates to our old discussion about what a "project" should be: A technical / build system project like ORT uses the term now, or a "product owner project", which could be compromised of multiple (e.g. frontend and backend) technical projects.

reporter/src/main/resources/templates/asciidoc/vulnerability_report.ftl
sschuberth: > Commit message: "instead **of**"

"for" should actually be correct here, as it's supposed to read like "use them instead, for listing the packages". I've anyway reworded the commit message.

> But as it was like that before that's fine with me.

Ok, I'll then leave it as-is for now.

mnonnenmacher (approved):

PR #6295, opened 2023-01-03
Do not use sorted sets / maps in `AnalyzerResult`

sschuberth:

### DIFF
--- a/advisor/src/main/kotlin/AdviceProvider.kt
+++ b/advisor/src/main/kotlin/AdviceProvider.kt
@@ -40,12 +40,10 @@ abstract class AdviceProvider(val providerName: String) {
     companion object : Logging
 
     /**
-     * For a given list of [Package]s, retrieve  findings and return a map that associates each package with a list
-     * of [AdvisorResult]s. Needs to be implemented by child classes.
+     * For a given list of [Package]s, retrieve findings and return a map that associates each package with a list of
+     * [AdvisorResult]s. Needs to be implemented by child classes.
      */
-    abstract suspend fun retrievePackageFindings(
-        packages: List<Package>
-    ): Map<Package, List<AdvisorResult>>
+    abstract suspend fun retrievePackageFindings(packages: List<Package>): Map<Package, List<AdvisorResult>>
 
     /**
      * An object with detail information about this [AdviceProvider].
@@ -53,8 +51,8 @@ abstract class AdviceProvider(val providerName: String) {
     abstract val details: AdvisorDetails
 
     /**
-     * A generic method that creates a failed [AdvisorResult] for [Package]s if there was an issue
-     * constructing the provider-specific information.
+     * A generic method that creates a failed [AdvisorResult] for [Package]s if there was an issue constructing the
+     * provider-specific information.
      */
     protected fun createFailedResults(
         startTime: Instant,
--- a/advisor/src/main/kotlin/AdviceProvider.kt
+++ b/advisor/src/main/kotlin/AdviceProvider.kt
@@ -40,10 +40,10 @@ abstract class AdviceProvider(val providerName: String) {
     companion object : Logging
 
     /**
-     * For a given list of [Package]s, retrieve findings and return a map that associates each package with a list of
+     * For a given set of [Package]s, retrieve findings and return a map that associates each package with a list of
      * [AdvisorResult]s. Needs to be implemented by child classes.
      */
-    abstract suspend fun retrievePackageFindings(packages: List<Package>): Map<Package, List<AdvisorResult>>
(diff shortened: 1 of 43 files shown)


advisor/src/main/kotlin/AdviceProvider.kt
```diff
@@ -43,7 +43,7 @@ abstract class AdviceProvider(val providerName: String) {
      * For a given list of [Package]s, retrieve findings and return a map that associates each package with a list of
      * [AdvisorResult]s. Needs to be implemented by child classes.
      */
-    abstract suspend fun retrievePackageFindings(packages: List<Package>): Map<Package, List<AdvisorResult>>
+    abstract suspend fun retrievePackageFindings(packages: Set<Package>): Map<Package, List<AdvisorResult>>
```
fviernau, line 46: Should the docs be updated? 
e.g. "For a given **set**..."

fviernau: Also, would it make sense to use `Collection` as type in function which are visible beyond class scope so that they can be called easily with any collection? (It would also simplify the caller code a bit (even as it is after applying the changes of this commit)

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
```diff
@@ -89,11 +89,14 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
         createExcludeInfo()
         createScopes()
 
-        input.ortResult.analyzer?.result?.projects?.forEach { project ->
+        val resultProjects = input.ortResult.getProjects().sortedBy { it.id }
```
fviernau, line 92: I'd prefer to not introduce these variables and do the sorting as part of the constructor call in line 124.
This would make it more obvious that packages and projects are actually sorted correctly: in the current code one needs to understand all the side effect the called functions like `addProject`, `addPackage` , `addDependencyTree` and `addShortestPaths` have. Also it's simpler to do the sorting separately, as all the other code would not need to worry about sorting at all.

Another option would be to not do the sorting at all in this data structure, but do it in the function which actually serializes the model.

What do you think?

fviernau (changes requested):

fviernau:
There are a couple of places now which ensure that the project's and package in the data class instances are actually sorted. This could be avoided by moving the sorting the function which actually serializes the model, for better separation of concerns. As result there would be only one single place which would need to care about sorting, which simplifies all the other places. What do you think?

advisor/src/main/kotlin/AdviceProvider.kt
sschuberth: One of the points of this change was to avoid all implementations to deduplicate the list before querying packages. Changing this to a generic collection would render that idea void. In general, if order and duplicates do not matter, I agree collection should be preferred over list / set. But here, I'd like to semantically express that implementation can rely on no duplicates being present.

sschuberth:
> This could be avoided by moving the sorting the function which actually serializes the model, for better separation of concerns.

You're right, for reporters that are mapper-based, I'll look into sorting only on serialization, just like for the `OrtResult` itself.

reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt
sschuberth: > I'd prefer to not introduce these variables and do the sorting as part of the constructor call in line 124.

Unfortunately, that's a lot more work as it's not enough to just sort (evaluated) projects and packages, but also licenses, license findings, dependency trees etc. would all need to be sorted, as they're sorted in lists, and the insertion order is determined by the iteration order over the original projects / packages.

> Another option would be to not do the sorting at all in this data structure, but do it in the function which actually serializes the model.

It would have been nice to be able to (easily) do that, but due to the complexity of the involved Jackson magic and the plethora of lists being used, it's hard to do. For example, `licenses` are not sorted by their own string representation, but by the id of the packages they stem from. IMO, the evaluated model is quite messed up in general, and I'm reluctant to fix it in the context of this change.

sschuberth:
> > This could be avoided by moving the sorting the function which actually serializes the model, for better separation of concerns.
> 
> You're right, for reporters that are mapper-based, I'll look into sorting only on serialization, just like for the `OrtResult` itself.

I've check, but unfortunately for none of these reporters just annotating the properties in the model to get sorted works. That's because the properties are not sorted by any of their own properties, but implicitly sorted by the iteration order to the original packages by their ids.

Of course, this can be fixed by e.g. changing the SPDX reporter to sort packages by their SPDX id (instead of the id of the original ORT package) and update the expected results accordingly, but that's another bigger change that's prefer to not do as part of this change, which already grew much larger than I anticipated.

fviernau (approved):

Issue #6064, opened 2022-11-10
evaluator: ProjectSourceRule should clone at most once


fviernau:
It currently clones once per rule.

PR #6304, opened 2023-01-05
evaluator: Clone project sources at most once

fviernau:
When executing multiple project source rules which access the project's source tree, the sources get cloned once per `ProjectSourceRule` instance contained in the `RuleSet`. These redundant clone operations slow down the execution without need.

Ensure that project sources are cloned at most once per rule set or rather evaluator run respectively.

Fixes: #6064.

### DIFF
--- a/evaluator/src/main/kotlin/ProjectSourceRule.kt
+++ b/evaluator/src/main/kotlin/ProjectSourceRule.kt
@@ -22,7 +22,6 @@ package org.ossreviewtoolkit.evaluator
 import java.io.File
 
 import org.ossreviewtoolkit.model.LicenseSource
-import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.model.utils.getRepositoryPath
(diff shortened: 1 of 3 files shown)


evaluator/src/main/kotlin/ProjectSourceRule.kt
```diff
@@ -34,7 +33,7 @@ import org.ossreviewtoolkit.utils.common.FileMatcher
 open class ProjectSourceRule(
     ruleSet: RuleSet,
     name: String,
-    projectSourceResolver: SourceTreeResolver = ruleSet.ortResult.createResolver()
+    projectSourceResolver: SourceTreeResolver = ruleSet.projectSourceResolver
```
sschuberth, line 36: Is there now a reason to keep the `projectSourceResolver` constructor parameter instead of accessing `ruleSet.projectSourceResolver` directly? If it was for testability, then `ProjectSourceRuleTest` could also either use the `RuleSet` constructor directly instead of the `ruleSet` helper function, or extend the latter to also accept a `SourceTreeResolver`.

fviernau: I've implemented your 2nd proposal.

PR #6306, opened 2023-01-06
CopyrightStatementsProcessor: Do not use sorted maps / sets in the `Result`

sschuberth:
There is no reason for the in-memory representation to be sorted, so sort
only the "visualizations" for the serialized result in
`CopyrightStatementsProcessorTest` and the console output in
`ListCopyrightsCommand`.

### DIFF
--- a/utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
+++ b/utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
@@ -20,10 +20,10 @@
 package org.ossreviewtoolkit.utils.ort
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
 
-import java.util.SortedMap
-import java.util.SortedSet
-
+import org.ossreviewtoolkit.utils.common.StringSortedSetConverter
 import org.ossreviewtoolkit.utils.common.collapseWhitespace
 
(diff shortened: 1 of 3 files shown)


utils/ort/src/main/kotlin/CopyrightStatementsProcessor.kt
```diff
@@ -20,10 +20,10 @@
 package org.ossreviewtoolkit.utils.ort
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
```
fviernau, line 23: Please also ensure the output of `ListCopyrightsCommand` remains sorted.

fviernau (changes requested):

fviernau:
Does this PR impact any ordering of the output of any of the reporter implementations like e.g. [1] or maybe more importantly the notice template reporter with the NOTICE templates? 

[1] https://github.com/oss-review-toolkit/ort/blob/94de43af0ac8dea417c05b7bcdab797624cd0523/reporter/src/main/kotlin/reporters/CycloneDxReporter.kt#L275

sschuberth:
> Does this PR impact any ordering of the output of any of the reporter implementations

According to the passing tests, no. And that's probably because insertion order in sets is maintained.

fviernau:
> According to the passing tests, no. And that's probably because insertion order in sets is maintained.

What do you think about sorting explicitly also in the reporters (for same reason as for doing so in jackson based serialization)?

fviernau (approved):

Issue #6056, opened 2022-11-09
Turn curation providers into plugins


sschuberth:
Similar to e.g. advisors and scanners, package curation providers should become ("named") plugins (discovered by `ServiceLoader`). This would enable third parties to write custom curation providers without modifying ORT, like curation providers that "calculate" curations based on package properties, or implement other feature like [wildcards that ORT core currently does not support](https://github.com/oss-review-toolkit/ort/issues/5105).

The implementation of this relates to https://github.com/oss-review-toolkit/ort/issues/6055.

Issue #6055, opened 2022-11-09
Make curation providers configurable via `config.yml`


sschuberth:
Currently, package curation providers are configured exclusively via the `AnalyzerCommand` CLI:

https://github.com/oss-review-toolkit/ort/blob/00e79bedc8a248c126465f6f1c8f746ccf96cc08/cli/src/main/kotlin/commands/AnalyzerCommand.kt#L233-L256

This makes it hard to leverage curation providers when using the analyzer programmatically, like in the ORT server. Instead, each curation provider should by itself look for its own settings in the `OrtConfiguration` / `AnalyzerConfiguration`, similar to like advisors and scanners already do.

This would also open up the way to [make curation providers plugins](https://github.com/oss-review-toolkit/ort/issues/6056).

Issue #6188, opened 2022-12-07
Allow curations to get applied without rerunning the analyzer


sschuberth:
Currently, the turn around-times to test (technical) curations (also see #6187) locally or on CI are rather high as curations are "baked" into Analyzer results. This means one needs to rerun the analyzer, even if the previous analysis was successful, just in order to get updated curation data applied to its results.

Just brainstorming some ideas how to address this (without any ordering implied):
- Do not store curations as part of Analyzer results at all, but create a new "Curator" tool that tales the Analyzer result as input and adds curations data to a separate section in the ORT result file, similar to like any other tool except for the Analyzer works.
- Create a helper command that can patch-up analyzer results with updates / different curations. Actaually, there already is the [PackageCurationsCommand](https://github.com/oss-review-toolkit/ort/blob/2c8e7ad0c1afb3d7371285ebb3d9bddb59b6657d/helper-cli/src/main/kotlin/commands/packagecuration/PackageCurationsCommand.kt)'s [SetCommand](https://github.com/oss-review-toolkit/ort/blob/2c8e7ad0c1afb3d7371285ebb3d9bddb59b6657d/helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt) which could probably be used for this.

sschuberth:
BTW, depending on the implementation, this could also solve https://github.com/oss-review-toolkit/ort/issues/5637.

mnonnenmacher:
Suggested name for the new ORT result section that could contain package curations,  package configurations and resolutions: "corrections"

mnonnenmacher:
To ensure consistency within a pipeline of the ORT tools it is important that configuration is only applied or replaced before it is used as input for a tool. For example, replacing package curations after running the scanner could lead to inconsistencies if the provenance metadata of a package is changed and as a result the scan result for that package does not match the provenance anymore.

A summary of the order in which tools should be run and configuration should be added to the result file to ensure consistency in later tools in the pipeline (package curations are split into technical "metadata corrections" and "legal curations" based on the idea from #6187):

Analyzer
-> package metadata corrections
-> Scanner, Advisor
-> package configurations, issue resolutions, vulnerability resolutions
-> Evaluator
-> rule violation resolutions
-> Reporter

mnonnenmacher:
Based on my above comment I propose that we extend the `OrtResult` model to contain the configuration used by each tool. I see two options:

[...]

PR #6308, opened 2023-01-06
Package curation provider plugins

mnonnenmacher:
Make package curation providers configurable plugins. Please see the commit messages for details.

**Breaking change:**
This PR adds a new section `packageCurationProviders` to the `config.yml` which is used to configure the package curation providers. The configuration of the `Sw360PackageCurationProvider` moved from `AnalyzerConfiguration` to this new section. The command line options of `AnalzyerCommand` to enable curation providers have been removed, instead this now needs to be configured in `config.yml`. See the changes to `reference.yml` in this PR for possible configuration options, this includes also an example for how to enable/disable a specific provider using an environment variable.


analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -30,6 +31,9 @@ import org.ossreviewtoolkit.utils.common.NamedPlugin
 interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<PackageCurationProvider> {
     companion object {
         val ALL = NamedPlugin.getAll<PackageCurationProviderFactory<*>>()
+
+        fun create(configurations: List<PackageCurationProviderConfiguration>) =
+            configurations.map { ALL.getValue(it.name).create(it.config) }.reversed()
```
sschuberth, line 36: Why is the list reversed here? Could you maybe add a code comment?

sschuberth (changes requested):

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.analyzer
 
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.PackageCuration
+import org.ossreviewtoolkit.model.config.PackageCurationProviderConfiguration
```
sschuberth, line 24: This commit should auto-close https://github.com/oss-review-toolkit/ort/issues/6055.

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -21,6 +21,29 @@ package org.ossreviewtoolkit.analyzer
 
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.PackageCuration
+import org.ossreviewtoolkit.utils.common.ConfigurablePluginFactory
```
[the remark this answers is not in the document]
sschuberth: Also, this commit should auto-close https://github.com/oss-review-toolkit/ort/issues/6056.

mnonnenmacher:
@oss-review-toolkit/core-devs I decided move the files to the new plugins folder which we discussed yesterday in a separate PR as I added some commits to clean up `reference.yml`  and `OrtConfigurationTest` and didn't want make this PR too big.

sschuberth:
FYI @mnonnenmacher, `OrtMainFunTest` still seems to fail.

mnonnenmacher:
> FYI @mnonnenmacher, `OrtMainFunTest` still seems to fail.

Should be fixed now.

analyzer/src/main/kotlin/PackageCurationProvider.kt
sschuberth, line 25: > as a preparation for making `PackageCurationProvider`s plugins.

Strictly speaking this is a bit misleading: `PackageCurationProvider`s themselves will never become plugins, but only their factories, and the factories become plugins already in this commit, so this is not a "preparation".

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -30,6 +31,11 @@ import org.ossreviewtoolkit.utils.common.NamedPlugin
 interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<PackageCurationProvider> {
     companion object {
         val ALL = NamedPlugin.getAll<PackageCurationProviderFactory<*>>()
+
+        fun create(configurations: List<PackageCurationProviderConfiguration>) =
+            // Reverse the list so that curations from providers with higher priority are applied later and can
+            // overwrite curations from providers with lower priority.
+            configurations.filter { it.enabled }.map { ALL.getValue(it.name).create(it.config) }.reversed()
```
sschuberth, line 38: Use `.asReversed()` to avoid creating a copy.

sschuberth: I'm actually unsure about this. Personally, I would expect curation providers to be queried / curations be applied in the order I define them in the configuration, and *not* for the order to express a priority.

cli/src/funTest/kotlin/OrtMainFunTest.kt
```diff
@@ -63,7 +64,18 @@ class OrtMainFunTest : StringSpec() {
...
     override suspend fun beforeSpec(spec: Spec) {
         configFile = createSpecTempFile(suffix = ".yml")
-        configFile.writeValue(OrtConfigurationWrapper(OrtConfiguration()))
+        configFile.writeValue(
+            OrtConfigurationWrapper(
+                OrtConfiguration(
+                    packageCurationProviders = listOf(
+                        PackageCurationProviderConfiguration(
+                            name = "File",
+                            config = mapOf("path" to projectDir.resolve("gradle/curations.yml").toString())
```
sschuberth, line 73: `.path` is probably more speaking than `.toString()` as it matches the key name.

cli/src/funTest/kotlin/OrtMainFunTest.kt
```diff
@@ -186,29 +198,6 @@ class OrtMainFunTest : StringSpec() {
         }
 
         "Analyzer creates correct output" {
-            val expectedResult = patchExpectedResult(
```
sschuberth, line 189: Why was this test removed?

model/src/test/kotlin/config/OrtConfigurationTest.kt
```diff
@@ -37,6 +37,7 @@ import io.kotest.matchers.types.shouldBeInstanceOf
 import java.io.File
 import java.lang.IllegalArgumentException
 
+import org.ossreviewtoolkit.model.Severity
```
sschuberth, line 40: Commit message nit: Maybe put `reference.yml` inside backticks.

sschuberth (changes requested):

analyzer/src/main/kotlin/PackageCurationProvider.kt
mnonnenmacher: I have removed that part of the sentence.

cli/src/funTest/kotlin/OrtMainFunTest.kt
mnonnenmacher: I have added an explanation to the commit message.

analyzer/src/main/kotlin/PackageCurationProvider.kt
sschuberth: Do we have other places in the config where order expresses priority that we could align with?

mnonnenmacher: I don't think so.

model/src/main/resources/reference.yml
```diff
@@ -19,26 +19,50 @@
...
     patentFilenames: [patents]
     rootLicenseFilenames: ['readme*']
 
+  packageCurationProviders:
+    - name: DefaultFile
+    - name: DefaultDir
+    - name: File
+      config:
+        path: '/some-path/curations.yml'
+    - name: File
```
zhernovs, line 45: Shouldn't this be "Dir" or "Directory"?

mnonnenmacher: There is currently only the `FilePackageCurationProvider` which handles both files and dirs. I would like to split this later on, but not in this PR.

analyzer/src/main/kotlin/PackageCurationProvider.kt
sschuberth: > Use `.asReversed()` to avoid creating a copy.

I believe this is the only open remark, otherwise I'd simply approve now.

mnonnenmacher: Fixed.

sschuberth (approved):

zhernovs:
Just curious, if `--package-curations-dir` and `--package-curations-file` options were deprecated for `AnalyzerCommand`, shouldn't it be also deprecated for `EvaluatorCommand`?

sschuberth:
> Just curious, if `--package-curations-dir` and `--package-curations-file` options were deprecated for `AnalyzerCommand`, shouldn't it be also deprecated for `EvaluatorCommand`?

In the `AnalyzerCommand` these options were not *deprecated*, but *removed*, because there's now a new way to configure them. In the `EvaluatorCommand`, however, these options are used to *override* the configured curations, so these options should be kept.

mnonnenmacher:
On the long term the `EvaluatorCommand` options should also be removed because it's inconsistent that not all curation providers are supported. Instead, a separate command to replace existing curations would be a better option, also see #6188.

PR #6311, opened 2023-01-09
Jitpack: Switch back to OpenJDK 11

mnonnenmacher:
Switch back to OpenJDK 11 for building the Jitpack artifacts. Otherwise it is not possible to use those artifacts to build plugins that still need to run on Java 11.

This partly reverts 171fa3d.


sschuberth:
> plugins that still need to run on Java 11.

Could you please give an example for *why* plugins would still need to run on Java 11?

Ah, I guess this is for the case that you have a custom ORT distribution built with Java 11 in order to analyze old Gradle projects, right? If so, could you maybe clarify this in the commit message?

sschuberth (approved):

Issue #3281, opened 2020-11-03
Migrate the Git implementation to pure JGit


sschuberth:
We currently [use](https://github.com/oss-review-toolkit/ort/blob/master/downloader/src/main/kotlin/vcs/Git.kt) a mix of Git CLI and JGit as [JGit does not support sparse checkouts](https://bugs.eclipse.org/bugs/show_bug.cgi?id=383772). However, [this post](https://www.eclipse.org/forums/index.php?t=msg&th=1094825&goto=1794275&#msg_1794275) seems to have an answer how to emulate sparse checkouts in pure JGit.

As a side note, we also have the idea to [get rid of the need for sparse checkouts altogether](https://github.com/oss-review-toolkit/ort/issues/2668) for any VCS implementation, but that requires significantly more work.

MarcelBochtler:
The proposed solution using 
```kotlin
git.checkout()
    .setStartPoint(revision)
    .addPaths(pathsToCheckout)
    .call()
```
unfortunately doesn't work.

The issue with this approach is that `workingTree.getRevision()` won't work anymore, because the partial checkout does not actually checkout the git revision.
`workingTree.getRevision()` returns an empty String when using the above method.

Another considered idea was to delete the non-required files, but this would decrease the downloader performance significantly for large repositories.

Leaving the ticket open because ORT should still migrate to a pure JGit implementation as soon as JGit supports sparse-checkouts.

sschuberth:
Closed as part of backlog grooming. Feel free to comment if you would like to contribute to this.

PR #6319, opened 2023-01-10
Git: Fall back to Git CLI if the shallow fetch is failing using JGit

MarcelBochtler:
JGit handles unshallow server responses incorrectly [1]. Add a fallback to the Git CLI if an unshallow fetch fails with JGit.

[1]: https://bugs.eclipse.org/bugs/show_bug.cgi?id=581228


MarcelBochtler:
While this is not really desired with regard to https://github.com/oss-review-toolkit/ort/issues/3281, JGit causes unrecoverable issues when [this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=581228) is triggered.
Therefore, it would be great if we could add this fallback, as ORT is currently using the Git CLI along with JGit anyway.

MarcelBochtler:
Related commit which introduced the issue: d899156a73fb8cbe82f856c6b01e12f599487f57

downloader/src/main/kotlin/vcs/Git.kt
```diff
@@ -218,13 +218,23 @@ class Git : VersionControlSystem(), CommandLineTool {
             logger.info { "Falling back to fetch everything including tags." }
 
             git.fetch().setUnshallow(true).setTagOpt(TagOpt.FETCH_TAGS).call()
+        }.recoverCatching {
+            it.showStackTrace()
+
+            logger.info { "Could not fetch everything using JGit: ${it.collectMessages()}" }
+            logger.info { "Could not fetch everything using JGit. Falling back to Git CLI." }
```
sschuberth, line 225: I believe only one of these / a combination of both would be enough.

sschuberth: And maybe explicitly say that this partly reverts d899156a73fb8cbe82f856c6b01e12f599487f57.

MarcelBochtler: > I believe only one of these / a combination of both would be enough.

I accidentally duplicated, added the failure message. Removed this. Now the logs are following the previous fallback logs.

sschuberth (approved):

PR #6326, opened 2023-01-11
FilePackageCurationProvider: Fix creating directory providers from config

sschuberth:
While the previous implementation did an `isDirectory` check, the implementation was still calling the `FilePackageCurationProvider` constructor directly which assumes all `File` instances to really be files.

Address that by instead relying on the existing logic in `from()`, and creating another convenience `from()` function that takes a `FilePackageCurationProviderConfig`.

### DIFF
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -38,7 +38,7 @@ class FilePackageCurationProviderConfig(
     /**
      * The path of the package curation file or directory.
      */
-    val path: String
+    val path: File
 )
 
 open class FilePackageCurationProviderFactory : PackageCurationProviderFactory<FilePackageCurationProviderConfig> {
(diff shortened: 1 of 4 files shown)


fviernau (approved):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -38,7 +38,7 @@ class FilePackageCurationProviderConfig(
     /**
      * The path of the package curation file or directory.
      */
-    val path: String
+    val path: File
```
mnonnenmacher, line 41: The property should also be renamed to `path`.
Commit message: "simplifies"

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -70,25 +70,18 @@ class DefaultDirPackageCurationProviderFactory : FilePackageCurationProviderFact
  * specified in [FileFormat].
  */
 class FilePackageCurationProvider(
-    curationFiles: List<File>
-) : SimplePackageCurationProvider(readCurationFiles(curationFiles)) {
-    constructor(curationFile: File) : this(listOf(curationFile))
+    vararg paths: File?
```
mnonnenmacher, line 73: I would prefer `files` here as well.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
sschuberth: > The property should also be renamed to `path`.

Uhm, the property is already named `path`...

> Commit message: "simplifies"

:+1:

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
sschuberth: I'd prefer to stick with `paths` to align with the naming in the config, and to make more clear that these can also be directories.

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
sschuberth: > Uhm, the property is already named `path`...

But if you actually meant "file", then I'd prefer to stick with "path" for the same reason as mentioned [here](https://github.com/oss-review-toolkit/ort/pull/6326#discussion_r1067125867), and to not break the config again.

mnonnenmacher: Yes, I meant `file`.

mnonnenmacher (approved):

PR #6329, opened 2023-01-11
Remove the mention of `--package-curations-*` for the analyzer

sschuberth:


docs/config-file-curations-yml.md
```diff
@@ -86,47 +86,31 @@ Where the list of available options for curations is defined in
...
-
-```bash
-cli/build/install/ort/bin/ort analyze
-  -i [source-code-of-project-dir]
-  -o [analyzer-output-dir]
-  --package-curations-file $ORT_CONFIG_DIR/curations.yml
-  --clearly-defined-curations
-```
+Alternatively to a single file, curations may also be split across multiple files below a directory, by default
+`$ORT_CONFIG_DIR/curations`. Both file and directory package curation providers may also be configured as
```
fviernau, line 99: Should we add a comma after "Both"?

docs/config-file-curations-yml.md
```diff
@@ -86,47 +86,31 @@ Where the list of available options for curations is defined in
...
+[FilePackageCurationProviders](../analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt) in
+`$ORT_CONFIG_DIR/config.yml`. Similarly, ORT can use [ClearlyDefined](https://clearlydefined.io/) and
```
fviernau, line 101: Would it make sense to somehow say what can be configured, e.g. using a different location?

...may be configured to use locations **different from the default ones**, see...

fviernau (changes requested):

docs/config-file-curations-yml.md
sschuberth: If so, I believe we'd need another comma after "providers", too. But I'm fine either way.

sschuberth: I guess I'll simply drop "Both" as it sounds a bit as if only these two can be configured in `config.yml`.

docs/config-file-curations-yml.md
sschuberth: > Would it make sense to somehow say what can be configured, e.g. using a different location?

Which type of provider are you referring to, only the `FilePackageCurationProviders`? Because for other, more can be configured (like credentials), but I didn't want to go into the details of the configuration options for each type here.

fviernau: Agree, it doesn't make sense to repeat these details. I thought that the point of the sentence, as-is, was to make clear that you can deviate from the default location. If that wasn't the case than my comment above comment does not really make sense.

fviernau (approved):

PR #6331, opened 2023-01-12
FilePackageCurationProvider: Filter out non-existing files

sschuberth:
This avoids the need for `$ORT_CONFIG_DIR/curations.yml` to be present.

### DIFF
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -38,7 +38,12 @@ class FilePackageCurationProviderConfig(
     /**
      * The path of the package curation file or directory.
      */
-    val path: File
+    val path: File,
+
+    /**
+     * A flag to denote whether the path is required to exist.
+     */
+    val mustExist: Boolean
 )
 
 open class FilePackageCurationProviderFactory : PackageCurationProviderFactory<FilePackageCurationProviderConfig> {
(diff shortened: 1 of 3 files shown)


analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -79,7 +79,11 @@ class FilePackageCurationProvider(
             val allCurations = mutableListOf<Pair<PackageCuration, File>>()
 
             val curationFiles = paths.flatMap {
-                if (it.isDirectory) FileFormat.findFilesWithKnownExtensions(it) else listOf(it)
+                when {
```
fviernau, line 82: Have you considered the case where one explicitly specified a non-default file location? e.g. in such case should it be a program error if the file does not exist?

fviernau (changes requested):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
sschuberth: Yeah, that's the problem highlighted by `FilePackageCurationProviderTest` now: In some cases (when the default location is used), we do *not* want an exception to be thrown. But if a non-default location that does not exist is specified, we (still?) want an exception to be thrown. With the unified code this becomes a bit nasty to implement. Let me think about it...

fviernau (approved):

analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
```diff
@@ -38,7 +38,12 @@ class FilePackageCurationProviderConfig(
     /**
      * The path of the package curation file or directory.
      */
-    val path: File
+    val path: File,
+
+    /**
+     * A flag to denote whether the path is required to exist.
+     */
+    val requireExistence: Boolean
```
mnonnenmacher, line 46: What about `mustExist` instead?

maxhbr (approved):
Also solves the issue for me

mnonnenmacher (approved):

sschuberth:
Merging despite `funTest-non-analyzer` being stuck.

PR #6335, opened 2023-01-12
ci(github): Start enforcing of Conventional Commits

sschuberth:
See [1]. This is a preparation for release automation including release
notes. At the same time, configure Renovate to use "deps" as the type
and no scope at all, see [2].

[1]: https://www.conventionalcommits.org/
[2]: https://docs.renovatebot.com/semantic-commits/

Co-authored-by: Martin Nonnenmacher <martin.nonnenmacher@gmail.com>

### DIFF
--- a/.commitlintrc.yml
+++ b/.commitlintrc.yml
@@ -1,0 +1,62 @@
+# Commitlint configuration.
+# See: https://github.com/conventional-changelog/commitlint/blob/master/docs/reference-rules.md
+---
+rules:
+  body-leading-blank:
+    - 2
+    - always
+  body-max-line-length:
+    - 2
+    - always
+    - 75
+  footer-leading-blank:
+    - 2
+    - always
+  header-max-length:
+    - 2
+    - always
+    - 75
+  scope-case:
+    - 2
+    - never
+    - - sentence-case
+      - start-case
+  subject-case:
+    - 2
+    - always
+    - - pascal-case
+      - sentence-case
+      - start-case
+      - upper-case
+  subject-empty:
+    - 2
+    - never
+  subject-full-stop:
+    - 2
+    - never
+    - .
+  type-case:
+    - 2
+    - always
+    - lower-case
+  type-empty:
+    - 2
+    - never
+  type-enum:
+    - 2
+    - always
+    - - build
+      - chore
+      - ci
+      - deps
+      - docs
+      - feat
+      - fix
+      - perf
+      - refactor
+      - revert
+      - style
+      - test
+  signed-off-by:
+    - 2
+    - always
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -9,6 +9,15 @@ on:
       - main
 
 jobs:
+  commit-lint:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: wagoid/commitlint-github-action@v5
+        with:
+          configFile: .commitlintrc.yml
   copyright-license:
     runs-on: ubuntu-22.04
     env:
--- a/renovate.json
+++ b/renovate.json
@@ -2,7 +2,9 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "config:base",
-    "docker:disable"
+    "docker:disable",
+    ":semanticCommitScopeDisabled",
+    ":semanticCommitTypeAll(deps)"
   ],
   "dependencyDashboard": false,
   "batect": {


tsteenbe (changes requested):
Please add an update to the Contributing.md to this PR which mentions that the ORT project requires all contributed code commits to conform conventional commits standard.

sschuberth:
> Please add an update to the Contributing.md to this PR

I can't, because the file resides at https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md now. I'll make a follow-up PR to that repo afterwards.

.commitlintrc.yml
```diff
@@ -0,0 +1,82 @@
...
+      - ci
+      - docs
+      - feat
+      - fix
+      - perf
+      - refactor
+      - revert
+      - style
+      - test
+      - ui
```
mnonnenmacher, line 79: Why do you include `ui`? It's not in the default Angular types: https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional#type-enum

.commitlintrc.yml
```diff
@@ -0,0 +1,82 @@
...
+      - evaluator
+      - examples
+      - github
+      - helper
+      - integrations
+      - model
+      - notifier
+      - reporter
+      - scanner
+      - utils
```
mnonnenmacher, line 39: Didn't we discuss not to enforce the scopes? It's maybe hard to predict all possible scopes, but we could try with that list.

.commitlintrc.yml
mnonnenmacher: And maybe we should use `deps:` instead of `chore(deps):` for dependency updates, could make it a bit easier to have a separate section in the changelog for dependency udpates. E.g. jreleaser by default only groups by type, not by scope, but that's configurable of course:
https://github.com/jreleaser/jreleaser/blob/ddc4d3545ea65a6947890aa0415770677a863c37/core/jreleaser-model-impl/src/main/resources/META-INF/jreleaser/changelog/preset-conventional-commits.yml

mnonnenmacher (changes requested):

.commitlintrc.yml
```diff
@@ -0,0 +1,82 @@
...
+    - 2
+    - never
+  subject-full-stop:
+    - 2
+    - never
+    - .
+  type-case:
+    - 2
+    - always
+    - kebab-case
```
mnonnenmacher, line 61: Why not lower-case like for `scope`?

.commitlintrc.yml
sschuberth: > Why do you include `ui`?

I was thinking of updates to assets like logos. Where else would that fit?

sschuberth: > And maybe we should use `deps:` instead of `chore(deps):` for dependency updates

Fine with me, esp. as dependency updates could be more important than "just a chore", but that would require some Renovate-config changes, right?

.commitlintrc.yml
sschuberth: Copy & paste from your config 😉

.commitlintrc.yml
sschuberth: I've decided to drop `scope-enum` as you sometimes might want to specify single file names.

.commitlintrc.yml
sschuberth: Actually, I believe `lower-case` *only* is enough, as we also configure `type-enum`.

.commitlintrc.yml
mnonnenmacher: > I was thinking of updates to assets like logos. Where else would that fit?

ORT doesn't have a UI, so maybe `chore(assets)`?

> Fine with me, esp. as dependency updates could be more important than "just a chore", but that would require some Renovate-config changes, right?

Yes, see: https://docs.renovatebot.com/semantic-commits

.commitlintrc.yml
```diff
@@ -0,0 +1,62 @@
...
+    - always
+  header-max-length:
+    - 2
+    - always
+    - 75
+  scope-case:
+    - 2
+    - never
+    - - sentence-case
+      - start-case
```
mnonnenmacher, line 23: Based on these examples:
```
  'lower-case',    // default
  'upper-case',    // UPPERCASE
  'camel-case',    // camelCase
  'kebab-case',    // kebab-case
  'pascal-case',   // PascalCase
  'sentence-case', // Sentence case
  'snake-case',    // snake_case
  'start-case'     // Start Case
```

I would suggest we allow pascal-case and lower-case instead of sentence-case and start-case, because I think there should not be spaces in the scope.
Currently we often use the class name as prefix, this could become for example `fix(MavenSupport): ...`

sschuberth: > I think there should not be spaces in the scope.

I agree, that's why I use `never` (instead of `always` previously) now.

.commitlintrc.yml
sschuberth: > ORT doesn't have a UI, so maybe `chore(assets)`?

That should work. I've removed `ui` again.

> Yes, see: https://docs.renovatebot.com/semantic-commits

:+1:

.commitlintrc.yml
```diff
@@ -0,0 +1,62 @@
...
+      - ci
+      - deps
+      - docs
+      - feat
+      - fix
+      - perf
+      - refactor
+      - revert
+      - style
+      - test
```
sschuberth, line 59: I'm note sure about the "test" type. When I correct styling in a test, would I choose "test" or "style"? Also, the type should be, well, about the type of the change, not about the area of code it touches. And "test" sounds to me more like an area than a type.

mnonnenmacher: From https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format:

```
test: Adding missing tests or correcting existing tests
```

mnonnenmacher (approved):

sschuberth:
Ok, let's start with this.

PR #6337, opened 2023-01-12
Allow to configure a minimum ClearlyDefined license score

sschuberth:

/cc @borisbaldassari

### DIFF
--- a/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
@@ -35,6 +35,7 @@ import org.ossreviewtoolkit.clients.clearlydefined.ComponentType
 import org.ossreviewtoolkit.clients.clearlydefined.Coordinates
 import org.ossreviewtoolkit.clients.clearlydefined.SourceLocation
 import org.ossreviewtoolkit.clients.clearlydefined.getCurationsChunked
+import org.ossreviewtoolkit.clients.clearlydefined.getDefinitionsChunked
 import org.ossreviewtoolkit.model.Hash
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.PackageCuration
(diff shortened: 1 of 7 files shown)


mnonnenmacher (dismissed):

analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
```diff
@@ -55,32 +56,46 @@ class ClearlyDefinedPackageCurationProviderConfig(
     /**
      * The URL of the ClearlyDefined server to use. If null, uses the [production server][Server.PRODUCTION.apiUrl].
      */
-    val serverUrl: String? = null
+    val serverUrl: String,
+
+    /**
+     * The minimum total score for a curation to be accepted.
+     */
+    val minTotalLicenseScore: Int
```
fviernau, line 64: What about naming this `minLicenseClarityScore` or even `minCoreLicenseClarityScore` to indicated we care only about the `core` facet? (See [1]].

Also, shall we add a link to [1] so that one can read up on it easily. IMO it's not obvious what the semantics of this is.

[1] https://github.com/clearlydefined/license-score/blob/b7977ca61087f0e5aaedd1f6a682a99c8a852352/ClearlyLicensedMetrics.md

fviernau (changes requested):

analyzer/src/main/kotlin/curation/ClearlyDefinedPackageCurationProvider.kt
sschuberth: IIUC, basically only the ["per-file license and copyright" score](https://github.com/clearlydefined/license-score/blob/b7977ca61087f0e5aaedd1f6a682a99c8a852352/ClearlyLicensedMetrics.md#per-file-license-and-copyright) is limited to the "core" facet. The [total score](https://github.com/clearlydefined/license-score/blob/b7977ca61087f0e5aaedd1f6a682a99c8a852352/ClearlyLicensedMetrics.md#clearlylicensed-scoring-formula) not necessarily is.

In any case, I'd prefer to stick closely to the "path" (`licensed` -> `score` -> `total`) to the value in CD's own data model, and not expose too much of the implementation detail of how the score is calculated (which might change at any time; the docs explicitly say "The focus for *now* is only on the core facet when facets are available and in *some* cases only on source code in the core facet").

But thanks for the link anyway, I've included it into the commit message.

fviernau (approved):

mnonnenmacher (approved):

sschuberth:
Merging despite tests being stuck that passed before.

PR #6338, opened 2023-01-13
Address some vulnerabilities reported for ORT using resolutions in .ort.yml

oheger-bosch:
This PR is in the context of scanning ORT with ORT. Sonatype NexusIQ which we use currently as advisor issues a number of vulnerabilities for ORT. In this PR I try to address some of them by resolutions in `.ort.yml` that (according to my analysis and understanding) should not affect ORT in the way it is using the affected components.

There are other vulnerabilities though which probably cannot be handled by such resolutions in general, since they may depend on the concrete context or environment in which ORT is used.

### DIFF
--- a/.ort.yml
+++ b/.ort.yml
@@ -57,6 +57,14 @@ resolutions:
     comment: >-
       This is a duplicate for CVE-2022-22965 reported by Sonatype NexusIQ, as Sonatype reported this issue before a
       CVE ID was officially released.
+  - id: "CVE-2016-7954"
+    reason: "INEFFECTIVE_VULNERABILITY"
+    comment: >-
+      This vulnerability is reported for the JRuby dependencies used within the Analyzer and the Reporter. The CVE is
+      actually assigned to Bundler which is shipped with JRuby. It is about a possibility to inject malicious code for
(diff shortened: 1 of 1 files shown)


sschuberth (approved):
Very nicely crafted resolution comments, @oheger-bosch!

.ort.yml
```diff
@@ -31,6 +31,10 @@ excludes:
     reason: "DEV_DEPENDENCY_OF"
     comment: >-
       Packages for development only. Not part of released artifacts.
+  - pattern: "dokka.*"
+    reason: "BUILD_DEPENDENCY_OF"
+    comment: >-
+      Packages used to build artifacts. Not part of released artifacts.
```
fviernau, line 37: Consider instead: "Packages for the Dokka documentation engine." 

See also GenerateScopeExcludesCommand.kt.

fviernau (approved):

sschuberth:
Looks like `funTest-non-analyzer` got stuck again... we probably need to have a look at what's going on there separately.

.ort.yml
sschuberth: See https://github.com/oss-review-toolkit/ort/pull/6340.

PR #6339, opened 2023-01-13
feat(DownloadCommand): Add a `--skip-excluded` option

bs-ondem:
The new option `--skip-excluded` allows the `DownloadCommand` to ignore downloading excluded packages/projects (via `RepositoryConfiguration`). By default the option is set to true and can be negate with `--include-excluded`.

### DIFF
--- a/cli/src/main/kotlin/commands/DownloaderCommand.kt
+++ b/cli/src/main/kotlin/commands/DownloaderCommand.kt
@@ -27,6 +27,7 @@ import com.github.ajalt.clikt.parameters.groups.required
 import com.github.ajalt.clikt.parameters.groups.single
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
 import com.github.ajalt.clikt.parameters.options.split
@@ -179,6 +180,11 @@ class DownloaderCommand : OrtCommand(
                 "result to limit downloads to. If not specified, all packages are downloaded."
     ).split(",")
 
+    private val skipExcluded by option(
+        "--skip-excluded",
+        help = "Do not download excluded projects or packages. Works only with the '--ort-file' parameter."
+    ).flag()
+
     private val ortConfig by requireObject<OrtConfiguration>()
 
     override fun run() {
@@ -205,9 +211,8 @@ class DownloaderCommand : OrtCommand(
         )
 
         val ortResult = readOrtResult(ortFile)
-        val analyzerResult = ortResult.analyzer?.result
 
-        if (analyzerResult == null) {
+        if (ortResult.analyzer?.result == null) {
             logger.warn {
                 "Cannot run the downloader as the provided ORT result file '${ortFile.canonicalPath}' does " +
                         "not contain an analyzer result. Nothing will be downloaded."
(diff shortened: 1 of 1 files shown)


cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -27,6 +27,7 @@ import com.github.ajalt.clikt.parameters.groups.required
 import com.github.ajalt.clikt.parameters.groups.single
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.flag
```
MarcelBochtler, line 30: Commit message nit: "Can be negate**d** with `--include-excluded`

cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -27,6 +27,7 @@ import com.github.ajalt.clikt.parameters.groups.required
 import com.github.ajalt.clikt.parameters.groups.single
```
nnobelis, line 27: Commit message: can be negate_d_

cli/src/main/kotlin/commands/DownloaderCommand.kt
```diff
@@ -179,6 +180,11 @@ class DownloaderCommand : OrtCommand(
                 "result to limit downloads to. If not specified, all packages are downloaded."
     ).split(",")
 
+    private val skipExcluded by option(
+        "--skip-excluded",
+        help = "Do not download excluded projects or packages. Works only with the '--ort-file' parameter."
+    ).flag("--include-excluded", default = true)
```
MarcelBochtler, line 186: Setting this to true would make this a breaking change.

I personally think that defaulting to omitting excluded projects is good, other tools (scanner) default to include excluded projects.
So we might want to default to `false` in this case as well, for consistency.

What do others think?

MarcelBochtler (commented):
Implementation looks good.

cli/src/main/kotlin/commands/DownloaderCommand.kt
mnonnenmacher: @MarcelBochtler I do agree, I would prefer `true` as default, but for consistency it would be better to use `false`.
@bs-ondem Please use a name for the first argument too.

sschuberth: Why do we need `--include-excluded` to invert this after all? The other commands that support `--skip-excluded` (`AdvisorCommand`, `ScannerCommand`) don't have it, either. In any case, I'd vote for consistency here. Also "include exclude" sounds a bit odd. Maybe `--contain-excluded` or `--download-excluded`?

mnonnenmacher: That's right, if the option has a default value the inverted flag is not required.

bs-ondem: > Why do we need --include-excluded to invert this after all?

If the default value for a flag is set to true, you need a `secondaryName` to set it to false. Otherwise, you can never set it to false. By setting the default value to  `false` now, the `secondaryName` is not needed anymore.

sschuberth: BTW, in the context of our general effort to reduce CLI options in favor of config options, should this really become a new CLI option? Like, what's the use-case exactly / how often would you invoke the downloader with different values for this option?

bs-ondem: The generated source code bundle (`--archive-all` option) contains sources for e.g. test related packages, which are actually excluded. With this option, we can prevent adding sources of excluded packages in it.

sschuberth: So, this means you *always* want to run the downloader CLI with `--skip-excluded`, and from that perspective it *could* become a config option. But I'm fine with not adressing this now as part of this PR, as it would require some alignment with `AdvisorCommand` and `ScannerCommand`, too.

sschuberth (dismissed):

mnonnenmacher (approved):

sschuberth (approved):

PR #6351, opened 2023-01-17
Address some more vulnerabilities findings

oheger-bosch:
This is a follow-up of #6338. It suppresses some more vulnerabilities by excluding a scope for generation of GraphQL clients and adding a constraint to enforce a higher version of a transitive gson dependency.

### DIFF
--- a/analyzer/build.gradle.kts
+++ b/analyzer/build.gradle.kts
@@ -72,7 +72,13 @@ dependencies {
     implementation(libs.kotlinxSerialization)
     implementation(libs.semver4j)
     implementation(libs.sw360Client)
+
     implementation(libs.toml4j)
+    constraints {
+        implementation("com.google.code.gson:gson:2.8.9") {
+            because("Earlier versions have vulnerabilities.")
+        }
+    }
 
     testImplementation(libs.mockk)
     testImplementation(libs.wiremock)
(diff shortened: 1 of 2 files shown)


analyzer/build.gradle.kts
```diff
@@ -73,6 +73,11 @@ dependencies {
     implementation(libs.semver4j)
     implementation(libs.sw360Client)
     implementation(libs.toml4j)
+    constraints {
```
sschuberth, line 76: Nit: Maybe also add an empty line before line 75 to more clearly group the constraint to the direct dependency that introduces this transitive dependency.

sschuberth (dismissed):

analyzer/build.gradle.kts
```diff
@@ -72,7 +72,13 @@ dependencies {
     implementation(libs.kotlinxSerialization)
     implementation(libs.semver4j)
     implementation(libs.sw360Client)
+
     implementation(libs.toml4j)
+    constraints {
+        implementation("com.google.code.gson:gson:2.8.9") {
+            because("Earlier versions have vulnerabilities.")
+        }
+    }
```
mnonnenmacher, line 81: Commit message: Type should be `build` instead of `chore(build)`.

mnonnenmacher (changes requested):

analyzer/build.gradle.kts
oheger-bosch: Changed commit message.

sschuberth (approved):

mnonnenmacher (approved):

PR #6364, opened 2023-01-19
fix(commitlint): Raise the max length of footer

heliocastro:
Commitlint extends the config set from
https://www.npmjs.com/package/@commitlint/config-conventional, which has validation entries for footer as well.

Signed-off-by: Helio Chissini de Castro <heliocastro@gmail.com>

### DIFF
--- a/.commitlintrc.yml
+++ b/.commitlintrc.yml
@@ -9,6 +9,10 @@ rules:
     - 2
     - always
     - 75
+  footer-max-line-length:
+    - 2
+    - always
+    - 120
   footer-leading-blank:
     - 2
     - always


.commitlintrc.yml
```diff
@@ -9,6 +9,10 @@ rules:
     - 2
     - always
     - 75
+  footer-max-line-length:
+    - 2
+    - always
+    - 120
```
sschuberth, line 15: As mentioned [here](https://github.com/oss-review-toolkit/ort/pull/6254#issuecomment-1385168926) the default is `Infinity`; so any idea why this should be needed?

sschuberth: Moreover, as you can see e.g. [here](https://github.com/oss-review-toolkit/ort/actions/runs/3958505589/jobs/6780504530#step:4:18), *commitlint* complains about `body-max-line-length` being exceeded for the `Signed-off-by` line. So it's not even recognizing that line as being part of the footer. Which is why I indeed believe that setting `footer-max-line-length` won't help.

mnonnenmacher: I have added a commit to #6366 that configures commitlint to treat "Signed-off-by" lines as footers.

PR #6368, opened 2023-01-20
refactor(Advisor)!: Make advise functions suspendable

oheger-bosch:
With this change, the Advisor can be better integrated with code that is already using coroutines. This also aligns the API with other ORT components like the Scanner.

BREAKING CHANGE: Clients now have to start coroutines.

### DIFF
--- a/advisor/src/main/kotlin/Advisor.kt
+++ b/advisor/src/main/kotlin/Advisor.kt
@@ -21,8 +21,9 @@ package org.ossreviewtoolkit.advisor
 
 import java.time.Instant
 
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
 
 import org.apache.logging.log4j.kotlin.Logging
 
@@ -55,8 +56,7 @@ class Advisor(
      * Query the [advice providers][providerFactories] and add the result to the provided [ortResult]. Excluded packages
      * can optionally be [skipped][skipExcluded].
      */
-    @JvmOverloads
-    fun advise(ortResult: OrtResult, skipExcluded: Boolean = false): OrtResult {
+    suspend fun advise(ortResult: OrtResult, skipExcluded: Boolean = false): OrtResult {
         if (ortResult.analyzer == null) {
             logger.warn {
                 "Cannot run the advisor as the provided ORT result does not contain an analyzer result. " +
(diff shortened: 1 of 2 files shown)


advisor/src/main/kotlin/Advisor.kt
```diff
@@ -56,7 +57,7 @@ class Advisor(
      * can optionally be [skipped][skipExcluded].
      */
     @JvmOverloads
-    fun advise(ortResult: OrtResult, skipExcluded: Boolean = false): OrtResult {
+    suspend fun advise(ortResult: OrtResult, skipExcluded: Boolean = false): OrtResult {
```
fviernau, line 60: Does this change make it harder to call the API from Java? Mind giving a hint why it becomes easier to intergate? (just asking to learn as I'm not yet deep into Kotlin's coroutines).

sschuberth: > Does this change make it harder to call the API from Java?

I was wondering about this as well, and it [seems like](https://stackoverflow.com/a/71599237/1127485) you'd need a `CompletableFuture` wrapper in that case. Put short, `suspend`functions are not directly callable from Java, so if we're fine with what, you could drop `@JvmOverloads` from here as well as a consequence.

sschuberth: More reading on that topic: https://jorgecastillo.dev/digging-into-kotlin-continuations

advisor/src/main/kotlin/Advisor.kt
```diff
@@ -74,7 +75,7 @@ class Advisor(
     /**
      * Query the [advice providers][providerFactories] for the provided [packages].
      */
-    fun advise(packages: Set<Package>): AdvisorRun {
+    suspend fun advise(packages: Set<Package>): AdvisorRun = withContext(Dispatchers.IO) {
```
sschuberth, line 78: Nit: Formatting should be like

```
suspend fun advise(packages: Set<Package>): AdvisorRun =
    withContext(Dispatchers.IO) {
```

advisor/src/main/kotlin/Advisor.kt
oheger-bosch: @fviernau, one big advantage of coroutines is that you can do things concurrently with a low number of threads. In this model, an action that would traditionally block a thread is put aside, and the thread can continue to run and process other actions available. You need to be careful, however, when calling code that is not prepared for this model and that might block in the traditional way; this would potentially impact all your coroutines. Therefore, you would typically take special care to prevent this, e.g. by having a special executor with a dedicated threadpool for running such blocking code.

By marking the functions of the Advisor with the *suspend* keyword, the component explicitly indicates that it adheres to the coroutine programming model and can be used directly without additional means.

advisor/src/main/kotlin/Advisor.kt
oheger-bosch: Changed formatting.

advisor/src/main/kotlin/Advisor.kt
sschuberth: I still believe `@JvmOverloads` should be dropped.

oheger-bosch: Done.

sschuberth (approved):

fviernau (approved):

PR #6369, opened 2023-01-20
Mercurial: Fix a problem with testing whether a VCS URL is applicable

oheger-bosch:
Refer to the single commits for further details.

### DIFF
--- a/downloader/src/main/kotlin/VersionControlSystem.kt
+++ b/downloader/src/main/kotlin/VersionControlSystem.kt
@@ -210,6 +210,12 @@ abstract class VersionControlSystem {
      */
     fun isAvailable(): Boolean = this !is CommandLineTool || isInPath()
 
+    /**
+     * Test - in a way specific to this [VersionControlSystem] - whether it can be used to download from the provided
+     * [vcsUrl]. This function is called by [isApplicableUrl] if it cannot be determined from parsing the [vcsUrl]
+     * whether it is applicable for this [VersionControlSystem] or not. A concrete implementation can do specific
+     * checks here that may also include network requests.
+     */
     protected abstract fun isApplicableUrlInternal(vcsUrl: String): Boolean
 
     /**
--- a/downloader/src/main/kotlin/VersionControlSystem.kt
+++ b/downloader/src/main/kotlin/VersionControlSystem.kt
@@ -38,7 +38,13 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
 
 import org.semver4j.Semver
 
-abstract class VersionControlSystem {
+abstract class VersionControlSystem(
+    /**
+     * The command line tool used by this implementation if any. Via this property, the instance can check whether
+     * the version control system is available.
+     */
+    private val commandLineTool: CommandLineTool? = null
+) {
     companion object : Logging {
         private val LOADER = ServiceLoader.load(VersionControlSystem::class.java)
 
@@ -208,7 +214,7 @@ abstract class VersionControlSystem {
     /**
      * Return true if this [VersionControlSystem] is available for use.
      */
-    fun isAvailable(): Boolean = this !is CommandLineTool || isInPath()
+    fun isAvailable(): Boolean = commandLineTool?.isInPath() ?: true
 
     /**
      * Test - in a way specific to this [VersionControlSystem] - whether it can be used to download from the provided
--- a/downloader/src/main/kotlin/vcs/Mercurial.kt
+++ b/downloader/src/main/kotlin/vcs/Mercurial.kt
@@ -51,7 +51,7 @@ object MercurialCommand : CommandLineTool {
         }
 }
 
-class Mercurial : VersionControlSystem() {
+class Mercurial : VersionControlSystem(MercurialCommand) {
     companion object : Logging
 
     override val type = VcsType.MERCURIAL
--- a/downloader/src/main/kotlin/vcs/Cvs.kt
+++ b/downloader/src/main/kotlin/vcs/Cvs.kt
@@ -45,7 +45,7 @@ object CvsCommand : CommandLineTool {
         }
 }
 
-class Cvs : VersionControlSystem() {
+class Cvs : VersionControlSystem(CvsCommand) {
     override val type = VcsType.CVS
     override val latestRevisionNames = emptyList<String>()
 
--- a/downloader/src/test/kotlin/VersionControlSystemTest.kt
+++ b/downloader/src/test/kotlin/VersionControlSystemTest.kt
@@ -28,12 +28,14 @@ import io.mockk.mockk
 
 import java.io.File
 import java.io.IOException
+import java.lang.UnsupportedOperationException
 
 import org.ossreviewtoolkit.downloader.vcs.Git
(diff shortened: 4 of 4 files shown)


downloader/src/main/kotlin/vcs/Mercurial.kt
```diff
@@ -65,7 +65,7 @@ class Mercurial : VersionControlSystem() {
     override fun getWorkingTree(vcsDirectory: File) = MercurialWorkingTree(vcsDirectory, type)
 
     override fun isApplicableUrlInternal(vcsUrl: String) =
-        ProcessCapture("hg", "identify", vcsUrl).isSuccess
+        MercurialCommand.isInPath() && ProcessCapture("hg", "identify", vcsUrl).isSuccess
```
fviernau, line 68: Would it make sense to call `isAvailable()` instead of `MercurialCommand.isInPath()`?

fviernau, line 68: commit: IIUC all production code callers only call visit this code path only if `VersionControlSystem.isAvailable()` returns `true`. Would you mind providing the callstack which leads to the mentioned exception to help me understanding it better?

fviernau, line 68: If we consider adding this fault tolerance the right thing to do, should the check then be moved a bit upwards the call tree so that it becomes general fall all `VersionControlSystem` implementations?

fviernau (changes requested):

downloader/src/main/kotlin/vcs/Mercurial.kt
oheger-bosch: The actual problem I try to solve here is that 'isAvailable()` for Mercurial returns *true* even if the `hg` command is not on the path. This is because the available check in `VersionControlSystem` is implemented as
`this !is CommandLineTool || isInPath()`

`Mercurial` is not a `CommandLineTool` (since the command line tool implementation is delegated to `MercurialCommand`), therefore, this check yields always *true*.

I think, this answers all your comments @fviernau, doesn't it?

fviernau: I believe the `isAvailable()` implementation worked out properly until [1] got merged.
So, [1] forgot to update `isAvailable()`. How about fixing up [1] to address the problem?

[1] https://github.com/oss-review-toolkit/ort/commit/55318202ea53c66a213f0c8fc5bd2fa7ecd440cb

oheger-bosch: I like this, but this would not be sufficient to do the check whether the command is in the path.

How about having a function `getAssociatedCommandLineTool(): CommandLineTool?` or similar in `VersionControlSystem` that returns the tool if there is one? Then all the properties of the returned tool object would be directly available.

sschuberth: Another idea: How about simply making `VersionControlSystem.isAvailable()` open and overriding it in `Mercurial` with `= MercurialCommand.isInPath()`? The same should then be done for `Cvs`, BTW.

oheger-bosch: I thought about this, too. One advantage of the solution proposed by @fviernau would be that it could be implemented in `VersionControlSystem` centrally without having to implement the logic in the concrete implementations.

oheger-bosch: I have updated the PR to implement the proposal. @sschuberth, @fviernau could you please have a look?

sschuberth (approved):

downloader/src/main/kotlin/VersionControlSystem.kt
```diff
@@ -205,10 +205,19 @@ abstract class VersionControlSystem {
...
     }
 
+    /**
+     * Return a reference to a [CommandLineTool] associated with this concrete [VersionControlSystem] if such a
+     * tool exists. This is used for instance to check whether this [VersionControlSystem] is available. This
+     * default implementation checks whether this class implements the [CommandLineTool] interface and if so, returns
+     * a reference to itself. Subclasses making use of a command line tool without implementing this interface should
+     * override this function.
+     */
+    open fun getCommandLineTool(): CommandLineTool? = this as? CommandLineTool
```
fviernau, line 215: There is not really need for exposing the command line tool again, I'd prefer to keep it hidden and thus use constructor injection instead. Would that make sense to you @sschuberth  @oheger-bosch ?

fviernau (changes requested):

downloader/src/main/kotlin/VersionControlSystem.kt
sschuberth: I don't have a strong opinion here.

oheger-bosch: @fviernau, how would you then implement the check whether the tool is on the path?

fviernau: In the constructor: `private val commandLineTool: CommandLineTool? = null`
And the check: `fun isAvailable(): Boolean = commandLineTool?.isInPath() ?: true`

oheger-bosch: Ah, I think I got it now. Changed accordingly.

downloader/src/test/kotlin/vcs/CvsTest.kt
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.downloader.vcs
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.shouldBe
+
+class CvsTest : StringSpec({
```
fviernau, line 25: For my taste, testing that the right instance is passed which comes at the cost of quite a bit boiler plate and exposing the commandLineTool property does not provide much value in return. In particular it seems a bit unbalanced, as this is the only thing the test checks. 

I'd prefer to drop this, leaving it up to you anyway.

fviernau (approved):

downloader/src/main/kotlin/VersionControlSystem.kt
fviernau: Thanks!

downloader/src/test/kotlin/vcs/CvsTest.kt
oheger-bosch: Dropped the tests and made the property private.

fviernau (approved):

sschuberth:
I'm not going to block this, but TBH I liked the previous approach better because e.g. the `Cvs` class referred to `CvsCommand` exclusively via it's static object name. Now `CvsCommand` is being referred to both directly and via the `commandLineTool` property, which IMO makes the code less readable. Also, this cannot easily be fixed by making `commandLineTool` protected and using it inside `Cvs` as it's nullable.

sschuberth:
Merging despite the unrelated SW360 test failures, see https://www.eclipsestatus.io/.

PR #6372, opened 2023-01-22
docs: Fix minor missing quotes in `reference.yml`

maxhbr:
a very small change


sschuberth (dismissed):
Some nits (for which *commitlint* partly does not support to add checks): We prefer to start capitalized after `:` in the title, and to use backticks for filenames (as in Markdown).

maxhbr:
Yay, github finally renders backticks in commit messages. Was not aware of that change.

sschuberth (approved):

PR #6379, opened 2023-01-23
ORT plugin renamings

fviernau:
Call the interface for ORT plugins `OrtPlugin` and renamed its property `name` to `type`.
Also, align another property name in `OrtConfiguration` for consistency.
Also see individual commits.

Breaking change:

1. The ORT plugin interface has been renamed, so plugins must implement `org.ossreviewtoolkit.utils.common.Plugin` instead of `org.ossreviewtoolkit.utils.common.Plugin`.
2. The property `Plugin.name` has been renamed to `Plugin.type`, so plugin implementations must align with that new property name.
3. `OrtConfiguration`s, e.g. `~/.ort/config/config.yml` which define `PackageCurationProviders` should align the respective property names from `name` to `type`.

### DIFF
--- a/model/src/main/kotlin/config/PackageCurationProviderConfiguration.kt
+++ b/model/src/main/kotlin/config/PackageCurationProviderConfiguration.kt
@@ -19,13 +19,16 @@
 
 package org.ossreviewtoolkit.model.config
 
+import com.sksamuel.hoplite.ConfigAlias
+
 import org.ossreviewtoolkit.utils.common.Plugin
 
 data class PackageCurationProviderConfiguration(
     /**
-     * The [name][Plugin.type] of the package curation provider.
+     * The [type][Plugin.type] of the package curation provider.
      */
-    val name: String,
+    @ConfigAlias("name")
+    val type: String,
 
     /**
      * Whether this curation provider is enabled.
--- a/model/src/main/kotlin/config/PackageCurationProviderConfiguration.kt
+++ b/model/src/main/kotlin/config/PackageCurationProviderConfiguration.kt
@@ -19,11 +19,11 @@
 
 package org.ossreviewtoolkit.model.config
 
-import org.ossreviewtoolkit.utils.common.NamedPlugin
+import org.ossreviewtoolkit.utils.common.Plugin
 
 data class PackageCurationProviderConfiguration(
     /**
-     * The [name][NamedPlugin.name] of the package curation provider.
+     * The [name][Plugin.name] of the package curation provider.
      */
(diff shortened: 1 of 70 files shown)


utils/common/src/main/kotlin/PluginManager.kt
```diff
@@ -27,14 +27,14 @@ import java.util.ServiceLoader
 inline fun <reified T : Any> getLoaderFor(): ServiceLoader<T> = ServiceLoader.load(T::class.java)
 
 /**
- * An interface to be implemented by plugins that have a name.
+ * An interface to be implemented by any ORT plugin.
  */
-interface NamedPlugin {
+interface OrtPlugin {
```
sschuberth, line 32: I thought we settled on just `Plugin` now, or? (In the context of renaming `OrtIssue` to `Issue`.)

utils/common/src/main/kotlin/PluginManager.kt
```diff
@@ -38,14 +38,14 @@ interface OrtPlugin {
...
-                it.name
+                it.type
             }
     }
 
     /**
      * The name of the plugin.
      */
-    val name: String
+    val type: String
```
sschuberth, line 48: The comment above should say "*type* of the plugin".

model/src/main/kotlin/config/PackageCurationProviderConfiguration.kt
```diff
@@ -19,13 +19,16 @@
...
+import com.fasterxml.jackson.annotation.JsonAlias
+
 import org.ossreviewtoolkit.utils.common.OrtPlugin
 
 data class PackageCurationProviderConfiguration(
     /**
      * The [name][OrtPlugin.type] of the package curation provider.
      */
-    val name: String,
+    @JsonAlias("name")
```
sschuberth, line 30: The docs about should also be adjusted.

sschuberth (changes requested):

model/src/main/kotlin/config/PackageCurationProviderConfiguration.kt
mnonnenmacher: Für Hoplite you need to add `@ConfigAlias`. I'm not sure if this is ever deserialized with Jackson so you might safely drop `@JsonAlias`.

mnonnenmacher (changes requested):

mnonnenmacher:
@fviernau Please add a summary of the breaking changes to the PR description.

sschuberth (approved):

PR #6381, opened 2023-01-23
Support more edge-cases of invalid Maven SCM connection URLs

sschuberth:

### DIFF
--- a/analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
+++ b/analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
@@ -140,6 +140,36 @@ class MavenSupportTest : WordSpec({
                 path = "boringssl-static"
             )
         }
+
+        "handle GitHub URLs with double 'git:' prefix" {
+            val mavenProject = MavenProject().apply {
+                scm = Scm().apply {
+                    connection = "scm:git:git:github.com/MarkusAmshove/Kluent.git"
+                }
+            }
+
+            MavenSupport.parseVcsInfo(mavenProject) shouldBe VcsInfo(
+                type = VcsType.GIT,
+                url = "github.com/MarkusAmshove/Kluent.git",
+                revision = "",
+                path = ""
+            )
+        }
+
+        "handle GitHub URLs with missing 'git:' provider" {
+            val mavenProject = MavenProject().apply {
+                scm = Scm().apply {
+                    connection = "scm:git@github.com/Yalantis/uCrop.git"
+                }
+            }
+
+            MavenSupport.parseVcsInfo(mavenProject) shouldBe VcsInfo(
+                type = VcsType.GIT,
+                url = "https://github.com/Yalantis/uCrop.git",
+                revision = "",
+                path = ""
+            )
+        }
     }
 
     "parseChecksum()" should {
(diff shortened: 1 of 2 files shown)


analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
```diff
@@ -140,6 +140,36 @@ class MavenSupportTest : WordSpec({
                 path = "boringssl-static"
             )
         }
+
+        "handle GitHub URLs with double 'git:' prefix" {
+            val mavenProject = MavenProject().apply {
+                scm = Scm().apply {
+                    connection = "scm:git:git:github.com/MarkusAmshove/Kluent.git"
```
fviernau, line 147: Do you think the metadata intended but forgot to specify the protocol, so intention was to specify the format as in line 132? If so, should we add `git` as a protocol to the result?

fviernau, line 147: pick 4aca038e8 refactor: Always get projects and packages via OrtResult getters

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
```diff
@@ -140,6 +140,36 @@ class MavenSupportTest : WordSpec({
...
+        "handle GitHub URLs with missing 'git:' provider" {
+            val mavenProject = MavenProject().apply {
+                scm = Scm().apply {
+                    connection = "scm:git@github.com/Yalantis/uCrop.git"
+                }
+            }
+
+            MavenSupport.parseVcsInfo(mavenProject) shouldBe VcsInfo(
+                type = VcsType.GIT,
+                url = "https://github.com/Yalantis/uCrop.git",
```
fviernau, line 168: If we mapped it to `git@github.com/Yalantis/uCrop.git` IIUC `normalizeVcsUrl()` would add the `ssh` protocol. Is `https` intended here?

fviernau (changes requested):

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
sschuberth: > pick 4aca038e8 refactor: Always get projects and packages via OrtResult getters

Hmm, I'm not sure what you're trying to tell me 👀

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
sschuberth: > Is `https` intended here?

Sort of: While the `git@` username more indicates SSH, the `/` (instead of `:`) separator after the host more indicates HTTP(S). So basically both could be valid choices. However, for our read-only use-case HTTP(S) is more suitable as it does not enforce authentication.

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
fviernau: LOL, this was a copy and paste mistake.

fviernau, line 147: Could it be that here, the `//` has just been forgotten, so it should be handled like `scm:git:git://` in 132 which maps it to the `git` protocol?

fviernau (changes requested):

analyzer/src/test/kotlin/managers/utils/MavenSupportTest.kt
sschuberth: Of course I don't know for sure, but even if that was the case, we're better off with HTTP(S) because, as you known, [GitHub deprecated the `git://` protocol](https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git).

fviernau: Yeah, I know. Was just wondering whether to keep on mapping that at a single place. Anyhow, let's keep this as is. It's probably better.

fviernau (approved):

Issue #5268, opened 2022-04-20
Support gathering of metadata for Pub dependencies not hosted on pub.dev


linuxdict:
The current implementation of the Pub package manager only gathers meta-data from pub.dev. For dependencies meta-data check it ONLY relies on querying pub.dev. This does not work for packages that are hosted elsewhere.

sschuberth:
Closed as part of backlog grooming. Feel free to comment if you would like to contribute to this.

PR #6382, opened 2023-01-23
Pub: Support Git dependencies

maklostermann:
Support Pub dependencies which are loaded from a Git repository.

This implements part of #5268 (only Git packages, but Pub supports additional sources).


sschuberth:
Hi @maklostermann, thank you for your contribution! We recently switched to [Conventional Commits](https://www.conventionalcommits.org/), so please rewrite your commit message accordingly. And while at it, please also drop the merge commit from your PR. If you want to catch up with upstream changes, please rebase instead.

maklostermann:
Thanks @sschuberth, hopefully I got it right with #6384.

sschuberth:
PS: @maklostermann, there's no need to close a PR to rewrite the commit messages. Just do the fixups locally and force-push your branch while keeping the original PR open.

PR #6366, opened 2023-01-19
Improve Advisor API

mnonnenmacher:
Add an `advise` function that is easier to use programmatically, see the commit messages for details.

**Breaking change:**
The function `Advisor.retrieveFindings()` was renamed to `Advisor.advise()`.

PR #6387, opened 2023-01-24
Change the curation provider API to take packages instead of ids

sschuberth:

> **Note**
> This is a **breaking change** in the sense that the `PackageCurationProvider` interface now takes a collection of `Package`s instead of `Identifiers`.

### DIFF
--- a/analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
+++ b/analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
@@ -28,6 +28,7 @@ import io.kotest.matchers.shouldNot
 
 import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService.Server
 import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 
 class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
@@ -35,9 +36,9 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
         val provider = ClearlyDefinedPackageCurationProvider()
 
         "return an existing curation for the javax.servlet-api Maven package" {
-            val identifier = Identifier("Maven:javax.servlet:javax.servlet-api:3.1.0")
+            val packages = createPackagesFromIds("Maven:javax.servlet:javax.servlet-api:3.1.0")
 
-            val curations = provider.getCurationsFor(listOf(identifier))
+            val curations = provider.getCurationsFor(packages)
 
             curations should haveSize(1)
             curations.values.flatten().first().data.concludedLicense shouldBe
@@ -45,18 +46,18 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
         }
 
         "return an existing curation for the slf4j-log4j12 Maven package" {
-            val identifier = Identifier("Maven:org.slf4j:slf4j-log4j12:1.7.30")
+            val packages = createPackagesFromIds("Maven:org.slf4j:slf4j-log4j12:1.7.30")
 
-            val curations = provider.getCurationsFor(listOf(identifier))
+            val curations = provider.getCurationsFor(packages)
 
             curations should haveSize(1)
             curations.values.flatten().first().data.vcs?.revision shouldBe "0b97c416e42a184ff9728877b461c616187c58f7"
         }
 
         "return no curation for a non-existing dummy NPM package" {
-            val identifier = Identifier("NPM:@scope:name:1.2.3")
+            val packages = createPackagesFromIds("NPM:@scope:name:1.2.3")
 
-            val curations = provider.getCurationsFor(listOf(identifier))
+            val curations = provider.getCurationsFor(packages)
 
             curations should beEmpty()
         }
@@ -66,18 +67,18 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
         val provider = ClearlyDefinedPackageCurationProvider(Server.DEVELOPMENT)
 
         "return an existing curation for the platform-express NPM package" {
-            val identifier = Identifier("NPM:@nestjs:platform-express:6.2.3")
+            val packages = createPackagesFromIds("NPM:@nestjs:platform-express:6.2.3")
 
-            val curations = provider.getCurationsFor(listOf(identifier))
+            val curations = provider.getCurationsFor(packages)
 
             curations should haveSize(1)
             curations.values.flatten().first().data.concludedLicense shouldBe "Apache-1.0".toSpdx()
         }
 
         "return no curation for a non-existing dummy Maven package" {
-            val identifier = Identifier("Maven:group:name:1.2.3")
+            val packages = createPackagesFromIds("Maven:group:name:1.2.3")
 
-            val curations = provider.getCurationsFor(listOf(identifier))
+            val curations = provider.getCurationsFor(packages)
 
             curations should beEmpty()
         }
@@ -92,20 +93,23 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
             val provider = ClearlyDefinedPackageCurationProvider(config)
 
             // Use an id which is known to have non-empty results from an earlier test.
-            val identifier = Identifier("Maven:org.slf4j:slf4j-log4j12:1.7.30")
+            val packages = createPackagesFromIds("Maven:org.slf4j:slf4j-log4j12:1.7.30")
 
-            val curations = provider.getCurationsFor(listOf(identifier))
+            val curations = provider.getCurationsFor(packages)
 
             curations should beEmpty()
         }
 
         "be retrieved for packages without a namespace" {
             val provider = ClearlyDefinedPackageCurationProvider()
-            val identifier = Identifier("NPM::acorn:0.6.0")
+            val packages = createPackagesFromIds("NPM::acorn:0.6.0")
 
-            val curations = provider.getCurationsFor(listOf(identifier))
+            val curations = provider.getCurationsFor(packages)
 
             curations shouldNot beEmpty()
         }
     }
 })
+
+private fun createPackagesFromIds(vararg ids: String) =
+    ids.map { Package.EMPTY.copy(id = Identifier(it)) }
--- a/analyzer/src/main/kotlin/PackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/PackageCurationProvider.kt
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.analyzer
 
 import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageCuration
 import org.ossreviewtoolkit.model.config.PackageCurationProviderConfiguration
 import org.ossreviewtoolkit.utils.common.ConfigurablePluginFactory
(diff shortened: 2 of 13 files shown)


analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
```diff
@@ -28,35 +28,36 @@ import io.kotest.matchers.shouldNot
...
+import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 
 class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
     "The production server" should {
         val provider = ClearlyDefinedPackageCurationProvider()
 
         "return an existing curation for the javax.servlet-api Maven package" {
-            val identifier = Identifier("Maven:javax.servlet:javax.servlet-api:3.1.0")
+            val packages = createPackagesFromIds("Maven:javax.servlet:javax.servlet-api:3.1.0")
```
fviernau, line 39: commit-message: I believe I don't fully get to which plan [1] refers to.
E.g. what are URL based providers and what would be example use cases?
Mind helping me understanding? 

[1] So, as a preparation for ORT to replace the
hard-coded providers [3] with real ones determined based on URLs

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -64,10 +65,10 @@ fun interface PackageCurationProvider {
...
-     * Return all available [PackageCuration]s for the provided [pkgIds], associated by the package's [Identifier]. Each
-     * list of curations must be non-empty; if no curation is available for a package, the returned map must not contain
-     * a key for that package's identifier at all.
+     * Return all available [PackageCuration]s for the provided [packages], associated by the package's [Identifier].
+     * Each list of curations must be non-empty; if no curation is available for a package, the returned map must not
+     * contain a key for that package's identifier at all.
      */
     // TODO: Maybe make this a suspend function, then all implementing classes could deal with coroutines more easily.
-    fun getCurationsFor(pkgIds: Collection<Identifier>): Map<Identifier, List<PackageCuration>>
+    fun getCurationsFor(packages: Collection<Package>): Map<Identifier, List<PackageCuration>>
```
mnonnenmacher, line 73: I would prefer an API that is more tailored for the use case. Packages contain a lot more data than required to lookup curations which leads to ugly constructs when using the API like `Package.EMPTY.copy(id = id)` in the tests. If the `Identifier` is not sufficient a dedicated request object would be better, for example if the `id` and `binaryArtifact` were required:

```kotlin
data class PackageCurationQuery(
  id: Identifier,
  binaryArtifact: RemoteArtifact = RemoteArtifact.EMPTY
)
```

analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
```diff
@@ -28,35 +28,36 @@ import io.kotest.matchers.shouldNot
 
 import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService.Server
 import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.Package
```
mnonnenmacher, line 31: Commit message: Typo in "chnage"

mnonnenmacher (changes requested):

analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
sschuberth: > Mind helping me understanding?

I'll try: Currently, [this code](https://github.com/oss-review-toolkit/ort/blob/33531c7/model/src/main/kotlin/utils/Extensions.kt#L38-L57) hard-codes e.g. the provider for all `ComponentType.MAVEN` packages to `Provider.MAVEN_CENTRAL`. But that's not correct as not all Maven packages are hosted on Maven Central. To fix that, the code needs to be able to look at the URL of the `RemoteArtifact`, and only if that *really* refers to Maven Central, use `MAVEN_CENTRAL` as the provider.

analyzer/src/main/kotlin/PackageCurationProvider.kt
sschuberth: At first, I also didn't like so much to pass in "heavyweight" `Package` instances. But actually, most "real" code (not counting tests) becomes simpler this way, as existing collections of packages do not need to be mapped to ids, avoiding the creation of temporary collections, or of temporary helper classes like your proposed `PackageCurationQuery`.

mnonnenmacher: This could be solved by having an additional `getCurationsFor(packages: Collection<Package>)` function in `PackageCurationProvider` that does the mapping for packages and calls the abstract `getCurationsFor` function while still ensuring that implementations of `PackageCurationProvider` only get the relevant data.

sschuberth: For providers, like ClearlyDefined, that actually require the full package, how would you then prevent callers from calling `fun getCurationsFor(pkgIds: Collection<Identifier>)` instead of `fun getCurationsFor(packages: Collection<Package>)`?

sschuberth: What could work is to turn `PackageCurationProvider` into an `abstract class` with:

```
    fun getCurationsFor(packages: Collection<Package>): Map<Identifier, List<PackageCuration>> =
        getCurationsFor(packages.map { it.id })

    protected fun getCurationsFor(ids: Collection<Identifier>): Map<Identifier, List<PackageCuration>> =
        throw NotImplementedError()

```

fviernau:
Hm, prior to this change I believe in ORT (code) we made the assumption that package IDs are unique and refer to a specific package. Most importantly, if a single ID refers to multiple different packages (e.g. forks), this **had to be solved outside of ORT**. That mentioned assumption is not just reflected in the package curation provider API, but almost all over the place, ORT's curation matching, curation data, package configurations, all kind of functions to do "things by ID".

Whether we move away from the previous concept IMO is a decision with large impact in terms of complexity and besides that it is not obvious at all what implications this in terms of alignment / implementation we have to do for implementing that consistently. One thing I don't like is that curations (by id) obtained from a provider would not be applicable *globally* anymore, but would be valid only *locally* in the given ORT result (in which we, as by this PR, still assume IDs to be unique). This alone IMO has huge implications on automation potential and complexity. 

...therefore, I'd propose we discuss the larger topic with @oss-review-toolkit/core-devs, before getting back into the details of this PR.

sschuberth:
> I'd propose we discuss the larger topic with @oss-review-toolkit/core-devs, before getting back into the details of this PR.

Sure, I'll put it onto the agenda. But I'd like to highlight that I believe you're over-interpreting the impact of this PR a bit: It's not about a general paradigm-shift that we must not identify packages solely by `Identifier` anymore. Instead, it's simply about giving curation providers access to more metadata, so that they can decide for themselves whether a curation is applicable or not. The fact whether they actually make use of more metadata than just the `Identifer` is an implementation detail of the curation provider.

sschuberth:
Meanwhile, I've [split out the first two commits](https://github.com/oss-review-toolkit/ort/pull/6398).

fviernau:
> But I'd like to highlight that I believe you're over-interpreting

I don't think so, because I believe it introduces a mis-match in semantics / design.

analyzer/src/main/kotlin/PackageCurationProvider.kt
mnonnenmacher: I didn't mean to keep the function that takes only ids, but that we should carefully extend the API to not include values that are not required. What I want to avoid is a situation like the one fixed [here](https://github.com/oss-review-toolkit/ort/pull/6366/commits/b1cec227e1264ee8abbe0f413ac9469f93c65c09) where the API required far more information than is actually required which unnecessarily complicates usage and testing.

So the proposal was to define which values are actually required to look up curation and limit the API to these, and to have a convenience function to extract that data from packages, for example:

```kotlin
data class PackageCurationQuery(
  val id: Identifier,
  val binaryArtifact: RemoteArtifact,
  val vcs: VcsInfo
)

// Curation providers only need to implement this function.
fun getCurationsFor(queries: Set<PackageCurationQuery>): Map<PackageCurationQuery, List<PackageCuration>>

fun getCurationsFor(packages: Set<Package>) = getCurationsFor(packages.map { it.toPackageCurationQuery() })

fun Package.toPackageCurationQuery() = PackageCurationQuery(id, binaryArtifact, vcsProcessed)
```

> For providers, like ClearlyDefined or also https://github.com/oss-review-toolkit/ort/issues/6389, that actually require the full package when implemented correctly,

If this is the case then the API must take full packages, but is it? Do ClearlyDefined or OSSelot require for example the `authors`, `description` or `isMetadataOnly` properties?

sschuberth: > If this is the case then the API must take full packages, but is it? Do ClearlyDefined or OSSelot require for example the `authors`, `description` or `isMetadataOnly` properties?

No, admittedly they don't. To me the question simply is where to draw the line: Starting with which percentage of the original properties required should we stop introducing a dedicated `Query` data class?

With [your change](https://github.com/oss-review-toolkit/ort/pull/6366/commits/b1cec227e1264ee8abbe0f413ac9469f93c65c09) advisors now operate on packages, [scanners can operate on (a single) package(s)](https://github.com/oss-review-toolkit/ort/blob/ed1be558ad222b67aa0ffee1a00fe1f6758c0045/scanner/src/main/kotlin/ScannerWrapper.kt#L72), so also from that perspective it makes sense to me if curation providers would operate on packages, too.

sschuberth:
> I don't think so, because I believe it introduces a mis-match in semantics / design, but sure I might miss something. Let's just discuss it.

But [advisors operate on packages instead of ids](https://github.com/oss-review-toolkit/ort/pull/6366/commits/b1cec227e1264ee8abbe0f413ac9469f93c65c09) now, too. And there you didn't complain about a "mis-match in semantics".

analyzer/src/main/kotlin/PackageCurationProvider.kt
mnonnenmacher: I think it would be difficult to draw a hard line but in this case it seemed to make sense to me to have a dedicated class because only a few properties would be needed in addition to the id and most package properties would not be required. I wouldn't compare the curation providers the advisor or scanner because these are top-level tools and the curation providers have a very specific use case.

What's actually your concern, is it that you don't want to introduce another class?

sschuberth: > is it that you don't want to introduce another class?

Exactly. I'm reluctant to introduce such small helper data classes (similar also for return values) as more classes simply make it harder to get an overview of class hierarchies and the code base in general.

Also the benefit of having such a class is unclear to me in this case: If we anyway have `fun getCurationsFor(packages: Set<Package>)`, which callers (currently) would call `fun getCurationsFor(queries: Set<PackageCurationQuery>)`? And: Is it really that bad to give advisors access to the rest of the (currently unneeded) `Package` properties, when in exchange it avoids a mapping?

mnonnenmacher: The benefit of a dedicated class to me is a bettter specification of the package curation provider API which also leads to cleaner tests. For example, with `Package.EMPTY.copy(id = id)` it's less clear which properties are actually used by the providers. Let's talk about this in the dev meeting today.

mnonnenmacher: For reference, it was agreed in the developer meeting to keep the API, for example because future provider implementations might need access to other package properties.

mnonnenmacher (dismissed):

mnonnenmacher (approved):

PR #6390, opened 2023-01-24
AnalyzerResultBuilder: Use apply() in all fluent API functions

fviernau:


analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -58,46 +58,44 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
...
-        val existingProject = projects.find { it.id == projectAnalyzerResult.project.id }
-
-        if (existingProject != null) {
-            val existingDefinitionFileUrl = existingProject.let {
-                "${it.vcsProcessed.url}/${it.definitionFilePath}"
-            }
-            val incomingDefinitionFileUrl = projectAnalyzerResult.project.let {
-                "${it.vcsProcessed.url}/${it.definitionFilePath}"
-            }
+    fun addResult(projectAnalyzerResult: ProjectAnalyzerResult) = apply {
```
sschuberth, line 61: @oss-review-toolkit/core-devs, so far we preferred to only start the implementation of a function expression after `=` if the whole implementation fits a single line. As that's not the case here, we would have formatted this like

```
fun addResult(projectAnalyzerResult: ProjectAnalyzerResult) =
    apply {
        // ...
    }
```

However, as recently @oheger-bosch and now also @fviernau used a syntax to start an opening block in the line of a function expression, I wonder whether we should officially endorse that style. The [Kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#expression-bodies) are a bit vague on that. Opinions? Personally, I think I'd still prefer to add that line break, because otherwise it's easy to overlook the `apply` (in this case) as usually that line would only end with the return value.

fviernau: I've got no strong opinion on this. My proposal would be: 

1. put on a new line per default
2. In cases where much spaced / less indent is helping much, then put on same line (locally consistent).

So, in short, decide case by case.

mnonnenmacher: Outside of ORT I started using this style a lot to save indentation, especially in files where most functions are wrapped with some other function, like a database class where all functions are wrapped with a `transaction { ... }`. So I would be fine with adopting this style in ORT. It would be nice if there was a detekt rule for this.

sschuberth: > It would be nice if there was a detekt rule for this.

I've [asked](https://kotlinlang.slack.com/archives/C88E12QH4/p1674638843243649).

> So, in short, decide case by case.

Ok, then let's at least allow the style.

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -58,46 +58,44 @@ class AnalyzerResultBuilder(private val curationProvider: PackageCurationProvide
...
+
+                val projectIssues = issues.getOrDefault(existingProject.id, emptyList())
+                issues[existingProject.id] = projectIssues + issue
+            } else {
+                projects += projectAnalyzerResult.project
+                addPackages(projectAnalyzerResult.packages)
+
+                if (projectAnalyzerResult.issues.isNotEmpty()) {
+                    issues[projectAnalyzerResult.project.id] = projectAnalyzerResult.issues
+                }
```
sschuberth, line 90: However, this function body needs to be de-indented.

sschuberth (changes requested):

sschuberth (approved):

PR #6393, opened 2023-01-25
Migrate the Opossum reporter to the dependency navigator

fviernau:
Mainly, migrate from accessing the legacy dependency representation to using the dependency navigator, to remove technical debt and to enable and make further simplifications.

See individual commits.

### DIFF
--- a/reporter/src/main/kotlin/reporters/opossum/OpossumReporter.kt
+++ b/reporter/src/main/kotlin/reporters/opossum/OpossumReporter.kt
@@ -38,14 +38,13 @@ import org.apache.logging.log4j.kotlin.Logging
 
 import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.model.CuratedPackage
+import org.ossreviewtoolkit.model.DependencyNode
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtIssue
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Package
-import org.ossreviewtoolkit.model.PackageReference
 import org.ossreviewtoolkit.model.Project
 import org.ossreviewtoolkit.model.ScanResult
-import org.ossreviewtoolkit.model.Scope
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.utils.getPurlType
 import org.ossreviewtoolkit.model.utils.toPurl
@@ -310,7 +309,7 @@ class OpossumReporter : Reporter {
         }
 
         private fun addDependency(
-            dependency: PackageReference,
+            dependency: DependencyNode,
             ortResult: OrtResult,
             relRoot: String,
             level: Int = 1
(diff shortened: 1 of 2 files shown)


reporter/src/main/kotlin/reporters/opossum/OpossumReporter.kt
```diff
@@ -567,3 +571,8 @@ class OpossumReporter : Reporter {
         return listOf(outputFile)
     }
 }
+
+private fun DependencyNode.getDependencies(): List<DependencyNode> =
+    buildList {
+        visitDependencies { this += it }
```
oheger-bosch, line 556: I am not sure whether this works in all constellations, especially for dependency graph nodes. Here `DependencyNode` is implemented by the [DependencyRefCursor](https://github.com/oss-review-toolkit/ort/blob/babdd1243a283513ccd1590794a10d2afdcdeb9b/model/src/main/kotlin/DependencyGraphNavigator.kt#L125) class. For efficiency reasons, this class iterates over dependencies by changing its own attributes and returning itself. So the list of dependency nodes will only contain a single instance. You will probably have to call `getStableReference()` to get different instances.

fviernau: Thanks, nice catch! Fix it.

reporter/src/main/kotlin/reporters/opossum/OpossumReporter.kt
```diff
@@ -380,7 +380,9 @@ class OpossumReporter : Reporter {
                 val rootForScope = resolvePath(definitionFilePath, scopeName)
                 val dependencies = ortResult.dependencyNavigator.directDependencies(project, scopeName).toList()
 
-                if (dependencies.isNotEmpty()) addAttributionBreakpoint(rootForScope)
+                if (dependencies.isNotEmpty()) {
+                    addAttributionBreakpoint(rootForScope)
+                }
```
sschuberth, line 384: Why? According to our (unwritten 😁) rules `if`-one-liners are fine, even if there was an `else`, as [Kotlin has no ternary operator](https://kotlinlang.org/docs/control-flow.html#if-expression).

reporter/src/main/kotlin/reporters/opossum/OpossumReporter.kt
```diff
@@ -483,20 +481,6 @@ class OpossumReporter : Reporter {
                 OpossumSignal(source, comment = issue.toString(), followUp = true, excludeFromNotice = true)
             addSignal(signal, paths.map { resolvePath(it) }.toSortedSet())
         }
-
```
sschuberth, line 488: Should this be marked as a breaking change in the commit message?

sschuberth (commented):
Only a few comments. I'd prefer @oheger-bosch and @maxhbr to approve.

reporter/src/main/kotlin/reporters/opossum/OpossumReporter.kt
fviernau: According to the unwritten rules I would guess both is fine.
I did this because I found it to be a tad more readable as the right hand side
is not super short. Anyhow, if you prefer the previous version I can drop this commit.

reporter/src/main/kotlin/reporters/opossum/OpossumReporter.kt
fviernau: I decided for not doing so, as users do not need to make any alignments. 
I considered it basically under the label "change slightly how the UI displays things", for which
I believe "breaking" is not needed. What do you think?

sschuberth: I'm fine either way. I just wanted to highlight it, also towards @maxhbr.

reporter/src/main/kotlin/reporters/opossum/OpossumReporter.kt
sschuberth: Then I'd indeed prefer to drop it as IMO there's no strong reason for the change.

fviernau:
@sschuberth @maxhbr - I've dropped the controversial commit (stop adding rootless packages) from this PR, so that we can proceed. The changes now should be non-function and easy to review.

sschuberth:
> The changes now should be non-function and easy to review. (Having this in removes a bit of hazzle in me coding workflow).

Ok, let's move ahead then as I don't know when @maxhbr will find the time to review.

sschuberth (dismissed):

benedikt-richter (approved):
lgtm

maxhbr (approved):
LGTM

sschuberth (approved):

PR #6402, opened 2023-01-26
fix(analyzer): Correctly determine the project name from Windows paths

### LINKED ISSUE #2900: Improve path handling on Windows
If the remote master of a GIT repository is set to "C:\RemoteMaster-Sample" (with a backslash) it will result in an analyzer error: 
```
'RemoteMaster-Sample' failed with: IllegalArgumentException: An identifier's properties must not contain ':' because that character is used as a separator in the string representation: type='Unmanaged', namespace='', name='C:\RemoteMaster-Sample', version='cXXXe'.
09:10 [Analyzer-1] INFO  org.ossreviewtoolkit.analyzer.PackageManager - Resolving Unmanaged dependencies for 'RemoteMaster-Sample' took 1 s.
```

You have to set the remote master to "C:/RemoteMaster-Sample" (with a forward slash) in the GIT settings for oss-review-toolkit to work. It should not be required to change your GIT repository names for oss-review-toolkit to work.

sschuberth:
Fixes #2900.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Unmanaged.kt
+++ b/analyzer/src/main/kotlin/managers/Unmanaged.kt
@@ -20,6 +20,7 @@
 package org.ossreviewtoolkit.analyzer.managers
 
 import java.io.File
+import java.net.URI
 
 import org.apache.logging.log4j.kotlin.Logging
 
@@ -101,10 +102,12 @@ class Unmanaged(
 
             else -> {
                 // For all non-GitRepo VCSes derive the name from the VCS URL.
+                val fileUrl = File(vcsInfo.url).takeIf { it.exists() } ?: File(URI(vcsInfo.url))
+
                 Identifier(
                     type = managerName,
                     namespace = "",
-                    name = vcsInfo.url.split('/').last().removeSuffix(".git"),
+                    name = fileUrl.name.removeSuffix(".git"),
                     version = vcsInfo.revision
                 )
             }


analyzer/src/main/kotlin/managers/Unmanaged.kt
```diff
@@ -101,10 +102,12 @@ class Unmanaged(
 
             else -> {
                 // For all non-GitRepo VCSes derive the name from the VCS URL.
+                val fileUrl = File(vcsInfo.url).takeIf { it.exists() } ?: File(URI(vcsInfo.url))
+
                 Identifier(
                     type = managerName,
                     namespace = "",
-                    name = vcsInfo.url.split('/').last().removeSuffix(".git"),
+                    name = fileUrl.name.removeSuffix(".git"),
```
fviernau, line 110: Have you considered to (also) change the code which puts the `\` into `VcsInfo.url`?
E.g. should that be changed to put "invariant separator paths" in there always?

fviernau (approved):

analyzer/src/main/kotlin/managers/Unmanaged.kt
sschuberth: I'll take a look at that idea and eventually create a follow-up PR later.

sschuberth: See https://github.com/oss-review-toolkit/ort/pull/6407.

Issue #5968, opened 2022-10-19
Allow to skip excluded paths during analysis


sschuberth:
In ORT's semantics "excluded" entities will still get analyzed / scanned by default, but the results from these show up marked as "excluded" in the reports.

This is a problem if you need to *really* skip e.g. a path from analysis, for example because that path contains an [Android](https://github.com/oss-review-toolkit/ort/issues/5622) or [NuGet](https://github.com/oss-review-toolkit/ort/issues/5038) project that cannot be properly analyzed.

The scanner (and also the advisor) already have a `--skip-excluded` CLI option. The proposal is to introduce `--skip-excluded` also to the analyzer, so that

1. excluded paths are not searched for definition files,
2. excluded scopes are either skipped during analysis or filtered out from the result afterwards.

That way, problematic / large projects that are *also* irrelevant compliance-wise in the specific project-context can be skipped and completely omitted from the results.

sschuberth:
Fun fact, we had a [`--remove-excludes-from-results` analyzer option](ce1c28bb151352e13fbe7bc9fdbe27e2c26f18ee) once, but it was [removed](a4d6c125ef1a54559cecee98eebe4cc1f1057d89) in favor of rendering excluded stuff differently in the reports.

hanna-modica:
Hi @sschuberth, that is indeed interesting. And also the comment from @mnonnenmacher states, that it will be reimplemented later. So it was always planned to have this feature back, I guess? I would really appreciate it :)

Issue #3825, opened 2021-03-29
Make more package managers use the new dependency graph format


sschuberth:
Currently, only the Gradle analyzer uses the new dependency graph format introduced in https://github.com/oss-review-toolkit/ort/pull/3502. We should take advantage of the new format for more (ultimately all) package manager implementations.

My proposal for package managers to be addressed next is:
- Maven
- NPM / Yarn
- Haskell / Stack

oheger-bosch:
To better track the progress, I propose to use this as an umbrella ticket and create separate tickets for specific package managers.

oheger-bosch:
The rework of the Maven package manager (#3895) is now done (see #3894). Unfortunately, the results from first tests are rather disappointing wrt reduction of file sizes. I don't have concrete numbers yet from a direct comparison (a test build is still running), but the result file in the new format is still quite big.

It seems that the situation for Maven projects is a bit different than for Gradle Android projects. In the latter case, a project typically has high redundancy due to the various scopes with almost identical dependency trees. A Maven module in a multi-module build in contrast has rather disjunct scopes, and packages do not appear multiple times in the dependency tree. As the dependency graph is constructed on a project-base, there is not much benefit from the new format.

There is, however, a lot of redundancy between the dependencies of the single Maven sub projects. So a promising approach could be to construct a global dependency graph for an analyzer run. This would be rather easy with the current `DependencyGraphBuilder` class; a new builder instance would not be created for each project, but only a single one, and there would need to be a convention to construct unique scope names qualified with project IDs.

A major problem would be that this is quite a big change in the data format of the analyzer and affects other model classes as well. Especially for projects it would be tricky to re-associate them with their part of the dependency graph.

sschuberth:
> There is, however, a lot of redundancy between the dependencies of the single Maven sub projects.

Oh, then I must have misunderstood something about the current new dependency graph format, as I though exactly that is already possible: The re-use of re-occurring subtrees within a single project / scope.

So for example, if we had

```
project
+- scope
   +- depA
   |  +- depC
   |     +- depD
   |     +- depE
   +- depB
      +- depC
         +- depD
         +- depE
```

then I though `depC` (incl. all its transitive dependencies, i.e. the whole subtree) would only be stored once, and *referred* to from `depA` and `depB`. Is this not the case?

oheger-bosch:
[...]

PR #6405, opened 2023-01-27
feat(Analyzer): Add a flag to skip excluded scopes in dependency graph

oheger-bosch:
This flag indicates that dependencies in excluded scopes should not be added to the dependency graph. For large projects, this could reduce the size of the dependency graph and the analysis time significantly.


model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -48,7 +48,15 @@ data class AnalyzerConfiguration(
...
-    val packageManagers: Map<String, PackageManagerConfiguration>? = null
+    val packageManagers: Map<String, PackageManagerConfiguration>? = null,
+
+    /**
+     * A flag to control whether excluded scopes should be skipped when constructing the dependency graph. If set to
+     * true, only the dependencies from non-excluded scopes are added to the dependency graph. This is useful for
+     * projects with large dependency graphs, for instance Gradle Android projects. By setting this flag to true,
+     * the size of the graph and the analysis time can be reduced significantly.
+     */
+    val skipExcludedScopesInDependencyGraph: Boolean = false
```
fviernau, line 59: A dependency can be excluded in two ways:

1. path exclude for the corresponding project
2. scope exclude (s)

So, I believe there should be a single flag to toggle inclusion of "excluded dependencies"
So, how about naming this just e.g.: 
- skipExcludedDependencies
- skipExcludedPackages

What do you think?

Apart from that, I'd prefer to drop the mention of "Dependency graph" as that's IMO not much relevant, just "dependencies" or "package" IMO should be enough.

model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -48,7 +48,15 @@ data class AnalyzerConfiguration(
      * Package manager specific configurations. The key needs to match the name of the package manager class, e.g.
      * "NuGet" for the NuGet package manager.
      */
-    val packageManagers: Map<String, PackageManagerConfiguration>? = null
+    val packageManagers: Map<String, PackageManagerConfiguration>? = null,
+
+    /**
```
fviernau, line 53: Did you forget to commit the handling of this flag?

fviernau (changes requested):

sschuberth:
I'd like to just point out https://github.com/oss-review-toolkit/ort/issues/5968. It would be great if this work could fully address that issue.

oheger-bosch:
@fviernau, @sschuberth: This is currently a bit experimental. What I am after is indeed not a full solution for #5968, but only a first step:

We have currently problems with an Android project that frequently crashes our ORT pipeline due to resource exhaustion. In the cases the build runs through, the Analyzer step takes more than 2 hours. The time seems to be consumed mainly when constructing the dependency graph.

I did some tests in which I filtered out test scopes when adding the dependencies to the graph builder - this made the build stable and reduced the time of the Analyzer step to 8 minutes! So, I think, this intermediate step to skip excluded scopes when building the dependency graph is already beneficial - and the implementation should not be too complex.

WDYT?

sschuberth:
> I did some tests in which I filtered out test scopes when adding the dependencies to the graph builder - this made the build stable and reduced the time of the Analyzer step to 8 minutes!

Feels a bit like cheating 😉 More seriously, I'd really like to understand *what exactly* is taking so long to construct the dependency graph. Also, I wouldn't expect test scopes to have much more dependencies than "regular" scopes, so why does excluding them make such a big difference?

oheger-bosch:
> Feels a bit like cheating wink More seriously, I'd really like to understand _what exactly_ is taking so long to construct the dependency graph. Also, I wouldn't expect test scopes to have much more dependencies than "regular" scopes, so why does excluding them make such a big difference?

Fair points. What we have noticed is that, starting from a certain size, adding further dependencies to the graph becomes rather slow. This is probably due to the fact that the subtrees that need to be compared become larger and larger. For Android projects with their numerous synthetic scopes this seems to be a real issue; so reducing the overall size of the graph by excluding unneeded scopes can actually have a significant effect.

Regarding "cheating": I think, handling scope excludes when constructing the dependency graph needs to be done anyway when implementing full support for exclusions as requested by #5968. So, these changes are not a side-track, but already a partial solution to this goal. I have pushed some more commits, so hopefully it becomes a bit clearer now what I try to achieve.

oheger-bosch:
I have extended the PR to cover the functionality requested by #5968. Still testing whether there is something missing.

analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
```diff
@@ -55,6 +55,7 @@ analyzer:
     tool_versions: {}
   config:
     allow_dynamic_versions: false
+    skip_excludes: false
```
sschuberth, line 58: > This flag determines whether the excludes defined in the repository configuration should be taken into account.

This sounds as if before excludes where not taken into account at all (which is not true, as later steps like the reporter already do take excludes into accounts.

The new thing here is that the *analyzer* takes excludes into account to skip paths. While that scope is slightly indicated by the commit message title, I propose to be more clear and say something like:

"This flag determines whether the excludes defined in the repository configuration should lead to paths being skipped completely during analysis."

model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -48,7 +48,12 @@ data class AnalyzerConfiguration(
...
+     * A flag to control whether excluded scopes and paths should be ignored during the analysis.
```
sschuberth, line 54: Consider using "skipped" instead of "ignored", which hopefully makes more clear what we're not analyzing and discarding the results, but not analyzing at all.

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -41,6 +41,7 @@ import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.Excludes
```
sschuberth, line 44: Commit message: Either "excludes" or put "Excludes" into back-ticks if you want to refer to the class.

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -199,13 +206,30 @@ abstract class PackageManager(
...
      */
     protected val options: Options = analyzerConfig.getPackageManagerConfiguration(managerName)?.options.orEmpty()
 
+    /**
+     * The [Excludes] to take into account during analysis. The [Excludes] from the [RepositoryConfiguration] are
+     * taken into account only if this is enabled in the [AnalyzerConfiguration].
+     */
+    val excludes by lazy {
+        analyzerConfig.excludes(repoConfig)
+    }
```
sschuberth, line 244: Nit: Could be a one-liner.

analyzer/src/funTest/assets/projects/synthetic/gradle-expected-output-scopes-excludes.yml
```diff
@@ -0,0 +1,185 @@
+---
```
sschuberth, line 1: The commit message still talks about "skipExclusions" instead of "skipExcludes".

sschuberth: Similar also for follow-up commits.

model/src/main/kotlin/utils/DependencyGraphConverter.kt
```diff
@@ -27,13 +27,15 @@ import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageLinkage
 import org.ossreviewtoolkit.model.PackageReference
 import org.ossreviewtoolkit.model.Project
+import org.ossreviewtoolkit.model.config.Excludes
```
sschuberth, line 30: Maybe link to https://github.com/oss-review-toolkit/ort/issues/3825 from the commit message for reference?

model/src/main/kotlin/utils/DependencyGraphConverter.kt
```diff
@@ -82,6 +94,19 @@ object DependencyGraphConverter {
...
+    /**
+     * Filter out all [packages] that are no longer referenced by one of the given [dependency graphs][graphs]. These
+     * packages have been subject of scope excludes.
+     */
+    private fun filterExcludedPackages(
+        graphs: Collection<DependencyGraph>,
+        packages: Collection<Package>
+    ): Set<Package> {
+        val includedPackages = graphs.flatMapTo(mutableSetOf()) { it.packages }
+
```
sschuberth, line 106: Nit: Maybe drop the blank line in this case?

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -174,7 +175,8 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
             }
         }
 
-        return state.buildResult()
+        val excludes = managedFiles.keys.firstOrNull()?.excludes ?: Excludes()
```
sschuberth, line 183: Should this re-use `PackageManager.EMPTY_EXCLUDES` (which would need to become at least `internal` then)?

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -43,15 +44,15 @@ class AnalyzerResultBuilder {
     private val issues = mutableMapOf<Identifier, List<Issue>>()
     private val dependencyGraphs = mutableMapOf<String, DependencyGraph>()
 
-    fun build(): AnalyzerResult {
+    fun build(excludes: Excludes = Excludes()): AnalyzerResult {
```
sschuberth, line 47: Should this re-use `PackageManager.EMPTY_EXCLUDES` (which would need to become at least `internal` then)?

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -99,20 +99,25 @@ abstract class PackageManager(
 
         /**
          * Recursively search the [directory] for files managed by any of the [packageManagers]. The search is performed
-         * depth-first so that root project files are found before any subproject files for a specific manager.
+         * depth-first so that root project files are found before any subproject files for a specific manager. Path
```
sschuberth, line 102: Nit: Maybe add a concrete use-case to the commit message? ("To skip problematic / large projects that are irrelevant compliance-wise.")

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -123,6 +128,14 @@ abstract class PackageManager(
                             return FileVisitResult.SKIP_SUBTREE
                         }
 
+                        if (excludes.isPathExcluded(rootPath.relativize(dir).toString())) {
```
sschuberth, line 131: Nit: Maybe use the slightly more speaking `pathString` than `toString()`?

analyzer/src/funTest/assets/projects/synthetic/gradle-expected-output-scopes-excludes.yml
oheger-bosch: I hope I found all stale references.

analyzer/src/main/kotlin/Analyzer.kt
oheger-bosch: As an alternative, would it make sense to have an `EMPTY` constant in the companion object of `Excludes`?

analyzer/src/main/kotlin/PackageManager.kt
oheger-bosch: I extended the commit message slightly.

analyzer/src/main/kotlin/Analyzer.kt
sschuberth: Yes, I think that would be more in-line with existing code.

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -90,22 +92,33 @@ abstract class PackageManager(
...
          * Recursively search the [directory] for files managed by any of the [packageManagers]. The search is performed
-         * depth-first so that root project files are found before any subproject files for a specific manager.
+         * depth-first so that root project files are found before any subproject files for a specific manager. Path
+         * excludes defined by the given [excludes] are taken into account; the corresponding directories are skipped.
          */
         fun findManagedFiles(
             directory: File,
-            packageManagers: Collection<PackageManagerFactory> = ALL.values
+            packageManagers: Collection<PackageManagerFactory> = ALL.values,
+            excludes: Excludes = EMPTY_EXCLUDES
```
fviernau, line 109: The `PackageManager` has an instance of `RepositoryConfiguration` containing all excludes.
So, in case `--skip-excluded` is passed, the excludes are passed redundantly. Would it make sense to pass a `Boolean` / flag instead?

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -116,6 +129,14 @@ abstract class PackageManager(
                             return FileVisitResult.SKIP_SUBTREE
                         }
 
+                        if (excludes.isPathExcluded(rootPath.relativize(dir).pathString)) {
+                            logger.info {
+                                "Not analyzing directory '$dir' as it is matched by a path exclude."
```
fviernau, line 134: optional: Maybe `PathExclude` is not needed implementation detail in "as it is matched by a path exclude.". Would it make sense to say "Not analyze directory '$dir' as it is excluded."?

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -116,6 +129,14 @@ abstract class PackageManager(
                             return FileVisitResult.SKIP_SUBTREE
                         }
 
+                        if (excludes.isPathExcluded(rootPath.relativize(dir).pathString)) {
+                            logger.info {
```
fviernau, line 133: I'm trying to understand why this is implemented here and not in the caller (`Analyzer`).
Am I right this is for optimizing performance (saving accessing some file system nodes for listing file) ?

I'm asking because this solution works only for excluded directory. Excluded definition files residing in non excluded directories won't get filtered out. Having this in the `Analyzer`
like `pkgManager.findManagedFile(..).filter { definitionFile -> pathExcludes.none  { it.matches(definitionFile) }` would work in any case.

fviernau (changes requested):

analyzer/src/main/kotlin/Analyzer.kt
oheger-bosch: Done.

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
oheger-bosch: Now using the new `Excludes.EMPTY` constant.

analyzer/src/main/kotlin/PackageManager.kt
oheger-bosch: `findManagedFiles()` is a function of the companion object. So the `RepositoryConfiguration` is not available here.

analyzer/src/main/kotlin/PackageManager.kt
oheger-bosch: To be honest, it did not occur to me that the filtering could be done in the `Analyzer`.

You are right, it is probably a bit more efficient to skip searching directory subtrees that are matched by an exclude. I have extended the code in `PackageManager` to handle excluded files as well.

oheger-bosch:
It is not clear to me why the commit-lint check fails. Could anybody help me understand what is wrong?

sschuberth:
> It is not clear to me why the commit-lint check fails. Could anybody help me understand what is wrong?

For some reason it seems to believe `graph format natively (see #3825).` is a foother line. No idea why. Maybe try omitting the parentheses and using a comma instead, like `, see #3825.`

oheger-bosch:
> > It is not clear to me why the commit-lint check fails. Could anybody help me understand what is wrong?
> 
> For some reason it seems to believe `graph format natively (see #3825).` is a foother line. No idea why. Maybe try omitting the parentheses and using a comma instead, like `, see #3825.`

The comma was not sufficient, obviously it has problems with the '#' character. I now used the reference style.

analyzer/src/main/kotlin/PackageManager.kt
fviernau: Doing so in the `Analyzer` IMO is way simpler than doing so in the `PackageManager`, also because in the `Analyzer` the `repositoryConfigration` containing all needed data is available. So, if the performance thing was not of interrest anyway, would it make sense to move the handling into the `Analyzer` ? (I fear I may be missing something)

oheger-bosch: Well, doing the filtering in the `PackageManager` is not that complicated either. BTW, I never said that performance was not of interest. A main use case behind this new feature is to reduce analysis time and resource usage by preventing unnecessary work.

What may be a bit related, I was wondering why the logic to find definition files is distributed between the `PackageManager` and the `Analyzer` as it currently is. IHMO, it is not in the responsibility of `PackageManager` to implement such a file search algorithm. Should the logic rather be moved completely to `Analyzer`, or maybe even better to a separate class? (But not in the scope of this PR, which is already quite huge.)

analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
sschuberth, line 58: Sorry for not noticing this earlier, but it just occurred to me that this should be called "skip_exclude*d*" (with "d" instead of "s") to align with the existing `--skip-excluded` CLI options, and to make it sound less like the excludes themselves would be skipped.

sschuberth (changes requested):

analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
oheger-bosch: Renamed.

sschuberth (approved):
LGTM, but I'd prefer to have at least one other approval from @oss-review-toolkit/kotlin-devs.

model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -48,7 +48,12 @@ data class AnalyzerConfiguration(
      * Package manager specific configurations. The key needs to match the name of the package manager class, e.g.
      * "NuGet" for the NuGet package manager.
      */
-    val packageManagers: Map<String, PackageManagerConfiguration>? = null
+    val packageManagers: Map<String, PackageManagerConfiguration>? = null,
+
+    /**
+     * A flag to control whether excluded scopes and paths should be skipped during the analysis.
+     */
+    val skipExcluded: Boolean = false
```
MarcelBochtler, line 56: Please document this new option. I guess the best place to do this would be this file:
https://github.com/oss-review-toolkit/ort/blob/main/docs/config-file-ort-yml.md#excludes

In there, you'd also need to adjust this sentence, as this changes with this PR:
> Note that the excluded parts are analyzed and scanned

MarcelBochtler (changes requested):

analyzer/src/main/kotlin/managers/Yarn2.kt
```diff
@@ -338,55 +338,55 @@ class Yarn2(
                 mapping += it.value
             }
         }
-        graphBuilder.addPackages(allPackages.values)
```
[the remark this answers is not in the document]
oheger-bosch: Good catch. This was never required, since the calls to `addDependency()` automatically add the packages to the graph builder. This would only be needed if packages were constructed in an alternative way, which is not the case here.

model/src/main/kotlin/config/AnalyzerConfiguration.kt
oheger-bosch: Sure, this is on my agenda, but I want to get this merged first to be sure that there are no more changes in naming or otherwise. I will then open a follow-up PR.

PR #6412, opened 2023-01-30
fix(Unmanaged): Fix ID generation for unmanaged projects

oheger-bosch:
This is a fix for 07c8633a in #6402 which changed the way the name for unmanaged projects was generated. This approach threw an IllegalArgumentException for all VCS URLs not of type file://. To fix this, obtain the project name from the path of the VCS URI.

I could successfully test the new way to obtain the project name in our setup; however, I am not sure how this looks like under Windows.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Unmanaged.kt
+++ b/analyzer/src/main/kotlin/managers/Unmanaged.kt
@@ -20,7 +20,6 @@
 package org.ossreviewtoolkit.analyzer.managers
 
 import java.io.File
-import java.net.URI
 
 import org.apache.logging.log4j.kotlin.Logging
 
@@ -102,12 +101,10 @@ class Unmanaged(
 
             else -> {
                 // For all non-GitRepo VCSes derive the name from the VCS URL.
-                val fileUrl = File(vcsInfo.url).takeIf { it.exists() } ?: File(URI(vcsInfo.url))
-
                 Identifier(
                     type = managerName,
                     namespace = "",
-                    name = fileUrl.name.removeSuffix(".git"),
+                    name = vcsInfo.url.split('/').last().removeSuffix(".git"),
                     version = vcsInfo.revision
                 )
             }


analyzer/src/main/kotlin/managers/Unmanaged.kt
```diff
@@ -102,7 +102,7 @@ class Unmanaged(
 
             else -> {
                 // For all non-GitRepo VCSes derive the name from the VCS URL.
-                val fileUrl = File(vcsInfo.url).takeIf { it.exists() } ?: File(URI(vcsInfo.url))
+                val fileUrl = File(URI(vcsInfo.url).path)
```
fviernau, line 105: The variable name `fileUrl` is now IMO misleading, because the value is no more a URL but a path.

Have you considered reverting the original change? (as we know previously it worked except for windows).
I'm not sure, but may  also fixed the issue for windows?

[1] https://github.com/oss-review-toolkit/ort/pull/6407

fviernau, line 105: Just for my understanding, could you tell where exactly the mentioned exception got thrown ?

oheger-bosch: It was actually not that easy to debug. The problem manifested itself at the very end of the analyzer run with a stack-trace as below. With the `--stack-trace` option enabled, it can be seen that the construction of the `File` from the `URI` after the Elvis operator threw an `IllegalArgumentException`, as the URI could not be converted to a file.

```
Exception in thread "main" java.lang.IllegalArgumentException: The VcsInfo(type=, url=, revision=, path=) of project 'Unmanaged::.:' cannot be found in Repository(vcs=VcsInfo(type=Git, url=XXXX.
	at org.ossreviewtoolkit.model.OrtResult.getFilePathRelativeToAnalyzerRoot(OrtResult.kt:245)
	at org.ossreviewtoolkit.model.OrtResult.getDefinitionFilePathRelativeToAnalyzerRoot(OrtResult.kt:231)
	at org.ossreviewtoolkit.model.config.Excludes.findPathExcludes(Excludes.kt:52)
	at org.ossreviewtoolkit.model.OrtResult$projects$2.invoke(OrtResult.kt:104)
	at org.ossreviewtoolkit.model.OrtResult$projects$2.invoke(OrtResult.kt:100)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at org.ossreviewtoolkit.model.OrtResult.getProjects(OrtResult.kt:100)
	at org.ossreviewtoolkit.model.OrtResult.getProject(OrtResult.kt:321)
	at org.ossreviewtoolkit.model.OrtResult.isProject(OrtResult.kt:472)
	at org.ossreviewtoolkit.model.OrtResult.isExcluded(OrtResult.kt:269)
	at org.ossreviewtoolkit.model.OrtResult.getProjects(OrtResult.kt:341)
	at org.ossreviewtoolkit.model.OrtResult.getProjects$default(OrtResult.kt:339)
	at org.ossreviewtoolkit.cli.commands.AnalyzerCommand.run(AnalyzerCommand.kt:242)
	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:198)
	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:211)
	at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:18)
	at com.github.ajalt.clikt.core.CliktCommand.parse(CliktCommand.kt:400)
	at com.github.ajalt.clikt.core.CliktCommand.parse$default(CliktCommand.kt:397)
	at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:415)
	at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:440)
	at org.ossreviewtoolkit.cli.OrtMainKt.main(OrtMain.kt:76)
```

oheger-bosch: I would like to hear @sschuberth's opinion to learn what was the original issue with Windows. I assume it was related to path separator characters, but are not sure.

fviernau: Given this stack trace, to my understanding, it's not possible that [1] introduced this bug (because it did not affect any VCS info). By coincidence I've been debugging recently as similar stack trace, and I believe this  stacktrace only happens if the VCS info of the project is not contained in the VCSes listed in `ortResult.repository`.

In particular, the function that crashes assumes that both, the project's VCS as well as the VCSes in `ortResult.repository` are normalized, see [2]. Setting a breakpoint in there IMO should help, assuming you have an Ort file to reproduce the issue.

[1] https://github.com/oss-review-toolkit/ort/commit/07c8633ac7a67c0ec8f3cb99da2fd387e3eee4fa 
[2] https://github.com/oss-review-toolkit/ort/blob/fb1d0a91aa89e6afa05b50c62bf311aaedb5fa03/model/src/main/kotlin/Repository.kt#L71-L81

fviernau: Given this stack trace, to my understanding, it's not possible that [1] introduced this bug (because it did not affect any VCS info). By coincidence I've been debugging recently as similar stack trace, and I believe this  stacktrace only happens if the VCS info of the project is not contained in the VCSes listed in `ortResult.repository`.

In particular, the function that crashes assumes that both, the project's VCS as well as the VCSes in `ortResult.repository` are normalized, see [2]. Setting a breakpoint in there IMO should help, assuming you have an Ort file to reproduce the issue.

[1] https://github.com/oss-review-toolkit/ort/commit/07c8633ac7a67c0ec8f3cb99da2fd387e3eee4fa 
[2] https://github.com/oss-review-toolkit/ort/blob/fb1d0a91aa89e6afa05b50c62bf311aaedb5fa03/model/src/main/kotlin/Repository.kt#L71-L81

fviernau (changes requested):

fviernau (approved):

analyzer/src/main/kotlin/managers/Unmanaged.kt
sschuberth: Yes, it's indeed fine to revert https://github.com/oss-review-toolkit/ort/pull/6402 as its follow-up https://github.com/oss-review-toolkit/ort/pull/6407 also fixes https://github.com/oss-review-toolkit/ort/issues/2900 in a more generic way.

PR #6413, opened 2023-01-30
Fix an issue with CuratedPackage.toUncuratedPackage()

fviernau:
Get rid of the "not always correctly working" [1] `toUncuratedPackage()` by utilizing the original uncurated packages from the `OrtResult`. While at it, do a couple of clean-ups for related `OrtResult` getters and their callers.

[1] https://github.com/oss-review-toolkit/ort/issues/5637.

BREAKING CHANGE: Only for programmatic use of ORT:
1. uncurated packages should be obtained now from the `OrtResult` instead of from the removed `toUncuratedPackage()`.
2. Some API functions get renamed, so the name change must be reflected.

### DIFF
--- a/cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
+++ b/cli/src/main/kotlin/commands/UploadResultToSw360Command.kt
@@ -194,7 +194,7 @@ class UploadResultToSw360Command : OrtCommand(
         ortResult.getProjects(omitExcluded = true).associateWith { project ->
             // Upload the uncurated packages because SW360 also is a package curation provider.
             ortResult.dependencyNavigator.projectDependencies(project)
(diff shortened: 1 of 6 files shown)


fviernau:
Could it be commit-lint is running into this bug? https://github.com/conventional-changelog/commitlint/issues/896

mnonnenmacher:
> Could it be commit-lint is running into this bug? [conventional-changelog/commitlint#896](https://github.com/conventional-changelog/commitlint/issues/896)

Looks like it, maybe the `#5637` in the body triggers the line to be read as footer because the commitlint output shows an empty line before the second line of the body. Maybe you can find a workaround, FYI you can commitlint locally like this `npx commitlint --from HEAD~2`.

PR #6416, opened 2023-01-31
OrtConfiguration: Introduce package curation provider IDs

fviernau:
Introduce package curation provider `IDs` for the `OrtConfiguration`, e.g. `~/ort/config/config.yml`.
This prepares for extending the ORT file format, to keep the association between curations and their origin, namely the provider `ID`. 

Related: #6188.

BREAKING CHANGE:
- For programmatic use of ORT there is a minor API change, e.g. drop of constant / change of parameter type.
- As of this change it is required to define an `id`  for any package curation provider in `~/ort/config/config.yml`, if it's `type` is not unique within that configuration. Note: The value of `type` is used as default value for `id`. The `id` must be defined if the default value would not be a unique amongst the `IDs`.


model/src/main/kotlin/config/PackageCurationProviderConfiguration.kt
```diff
@@ -30,6 +30,11 @@ data class PackageCurationProviderConfiguration(
     @ConfigAlias("name")
     val type: String,
 
+    /**
+     * A unique identifier for the package curation provider.
+     */
+    val id: String,
```
mnonnenmacher, line 36: I would prefer if this would default to `type` to not require specifying an id as long as no two providers of the same type are used.

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -33,14 +33,20 @@ interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<Pac
...
-            configurations.filter { it.enabled }.map { ALL.getValue(it.type).create(it.config) }.asReversed()
-    }
+        fun create(
+            configurations: List<PackageCurationProviderConfiguration>
+        ): List<Pair<String, PackageCurationProvider>> =
+            configurations.filter {
+                it.enabled
+            }.map {
+                it.id to ALL.getValue(it.type).create(it.config)
+            }.asReversed()
```
mnonnenmacher, line 48: This function should throw an exception if there are two providers with the same `id`. It's better to have the check here than just in the `AnalyzerCommand` for reusability.

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -32,9 +32,13 @@ interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<Pac
     companion object {
         val ALL = Plugin.getAll<PackageCurationProviderFactory<*>>()
 
+        /**
+         * Return a new provider instance for each [enabled][PackageCurationProviderConfiguration.enabled] provider
+         * configuration in [configurations]. The given [configurations] must be ordered highest-priority first while
+         * the returned providers are ordered lowest-priority first, e.g. the order in which the corresponding curations
```
mnonnenmacher: "e.g." means "for example", better say "which is" instead.

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -193,13 +194,23 @@ class AnalyzerCommand : OrtCommand(
             val repositoryPackageCurations = repositoryConfiguration.curations.packages
 
             if (ortConfig.enableRepositoryPackageCurations) {
-                add(SimplePackageCurationProvider(repositoryPackageCurations))
+                add("ort.yml" to SimplePackageCurationProvider(repositoryPackageCurations))
```
mnonnenmacher, line 197: If the check for duplicate ids is moved to `PackageCurationProvider` as suggested above here could be a check if the "ort.yml" id is already used.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -24,22 +24,45 @@ import org.ossreviewtoolkit.model.PackageCuration
...
+                }
+
+                val duplicateIds = getDuplicates { (id, _) -> id }.keys
+                require(duplicateIds.isEmpty()) {
+                    "Found multiple package curation providers for the IDs ${duplicateIds.joinToString()}, which is " +
+                            "not allowed. Please configure a unique ID for each package curation provider."
+                }
+
+                require(none { (id, _) -> id == ORT_YML_PROVIDER_ID }) {
+                    "Found package curation provider which uses '$ORT_YML_PROVIDER_ID' as id which is a reserved " +
```
mnonnenmacher, line 62: Missing "id" after "reversed", or maybe just "is reserved".

mnonnenmacher (changes requested):

mnonnenmacher (approved):

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -22,7 +22,7 @@ package org.ossreviewtoolkit.analyzer
 import java.io.File
 import java.time.Instant
 
-import kotlin.time.measureTimedValue
+import kotlin.time.measureTime
```
sschuberth, line 25: Hmm, at least to me it's not clear why the concept of introducing IDs generally conflicts with having a composite package curation provider. Esp. as the idea of the composite package curation provider is not actually dropped, but rather inlined, I'd say. I would have appreciated some more details in the commit message here.

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -24,22 +24,45 @@ import org.ossreviewtoolkit.model.PackageCuration
...
+         * Return a new (identifier, provider instance) tuple for each
+         * [enabled][PackageCurationProviderConfiguration.enabled] provider configuration in [configurations]. The given
+         * [configurations] must be ordered highest-priority first while the returned providers are ordered
+         * lowest-priority first, which is the order in which the corresponding curations must be applied.
          */
-        fun create(configurations: List<PackageCurationProviderConfiguration>): List<PackageCurationProvider> =
-            configurations.filter { it.enabled }.map { ALL.getValue(it.type).create(it.config) }.asReversed()
+        fun create(
+            configurations: List<PackageCurationProviderConfiguration>
+        ): List<Pair<String, PackageCurationProvider>> =
```
sschuberth, line 45: I'm surprised that this now is a list of pairs instead of a map [as discussed before](https://github.com/oss-review-toolkit/ort/pull/6375/files#diff-f97d611d1c2477b31e641e58a37794afd7b77e3e586b8cc2660562c56f85ebddR40). Have you considered it to be a map?

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -24,22 +24,45 @@ import org.ossreviewtoolkit.model.PackageCuration
...
 import org.ossreviewtoolkit.utils.common.Plugin
+import org.ossreviewtoolkit.utils.common.getDuplicates
 
 /**
  * The extension point for [PackageCurationProvider]s.
  */
 interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<PackageCurationProvider> {
     companion object {
         val ALL = Plugin.getAll<PackageCurationProviderFactory<*>>()
+        const val ORT_YML_PROVIDER_ID = "ort-yml"
```
sschuberth, line 35: Like for types, shouldn't we rather align on Camel-Case for IDs, too? Esp. as you default `id` to `type` below.

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -117,7 +117,7 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
     @JvmOverloads
     fun analyze(
         info: ManagedFileInfo,
-        curationProviders: List<PackageCurationProvider> = emptyList()
+        curationProviders: List<Pair<String, PackageCurationProvider>> = emptyList()
```
sschuberth, line 120: I would have appreciated some explanations in the commit message about the concept of the ID, what it actually is, i.e. a unique identifier for curation providers to distinguish different providers of the same type.

sschuberth (commented):
I have a few after-the-fact comments that we can discuss in a Kotlin developer meeting.

fviernau:
> I would have appreciated some explanations in the commit message about the concept of the ID

>  I would have appreciated some more details in the commit message here.

@sschuberth I was unable to guess your desires for the commit message, and I'm also unable to guess the intention of these comments. Could you please shed a lite on how I should interpret these?

analyzer/src/main/kotlin/PackageCurationProvider.kt
sschuberth: I was reminded that, as order of curation providers matters, a general map (with unordered keys) is not suitable.

analyzer/src/main/kotlin/PackageCurationProvider.kt
sschuberth: See @fviernau's https://github.com/oss-review-toolkit/ort/pull/6447.

PR #6418, opened 2023-01-31
fix(Npm): Correctly parse the output of the 'npm' command

oheger-bosch:
When parsing the output of the command to install NPM dependencies, do not expect that the output starts with the expected error level, but read the whole text if necessary. Otherwise, important error information can be missed if the output starts with a hint.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Npm.kt
+++ b/analyzer/src/main/kotlin/managers/Npm.kt
@@ -597,11 +597,12 @@ open class Npm(
         )
 
         fun mapLinesToOrtIssues(prefix: String, severity: Severity) {
-            val issueLines = lines.takeWhile { it.startsWith(prefix) }.mapNotNull { line ->
-                line.removePrefix(prefix).let {
-                    commonSecondaryPrefixes.fold(it) { remainder, prefix -> remainder.removePrefix(prefix) }
-                }.takeUnless { it.isBlank() }
-            }
+            val issueLines = lines.dropWhile { !it.startsWith(prefix) }
+                .takeWhile { it.startsWith(prefix) }.mapNotNull { line ->
+                    line.removePrefix(prefix).let {
+                        commonSecondaryPrefixes.fold(it) { remainder, prefix -> remainder.removePrefix(prefix) }
+                    }.takeUnless { it.isBlank() }
+                }
 
             if (issueLines.isNotEmpty()) {
                 issues += OrtIssue(
--- a/analyzer/src/test/kotlin/managers/NpmTest.kt
+++ b/analyzer/src/test/kotlin/managers/NpmTest.kt
@@ -1,0 +1,92 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.analyzer.managers
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.collections.beEmpty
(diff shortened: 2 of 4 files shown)


analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -597,11 +597,12 @@ open class Npm(
         )
 
         fun mapLinesToOrtIssues(prefix: String, severity: Severity) {
-            val issueLines = lines.takeWhile { it.startsWith(prefix) }.mapNotNull { line ->
-                line.removePrefix(prefix).let {
-                    commonSecondaryPrefixes.fold(it) { remainder, prefix -> remainder.removePrefix(prefix) }
-                }.takeUnless { it.isBlank() }
-            }
+            val issueLines = lines.dropWhile { !it.startsWith(prefix) }
+                .takeWhile { it.startsWith(prefix) }.mapNotNull { line ->
```
fviernau, line 601: Is `.takeWhile { it.startsWith(prefix) }` now redudant?

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -597,11 +597,12 @@ open class Npm(
         )
 
         fun mapLinesToOrtIssues(prefix: String, severity: Severity) {
-            val issueLines = lines.takeWhile { it.startsWith(prefix) }.mapNotNull { line ->
-                line.removePrefix(prefix).let {
-                    commonSecondaryPrefixes.fold(it) { remainder, prefix -> remainder.removePrefix(prefix) }
-                }.takeUnless { it.isBlank() }
-            }
+            val issueLines = lines.dropWhile { !it.startsWith(prefix) }
```
fviernau, line 600: Would move the filtering into the mapNotNull, e.g.:

```
lines.mapNotNull { line ->
    if (!line.startsWith(prefix)) null
    else {
          commonSecondaryPrefixes.fold(line.removePrefix(prefix)) { remainder, prefix ->    
                remainder.removePrefix(prefix) 
           }.takeUnless { it.isBlank() }
     }
}
```

analyzer/src/test/kotlin/managers/NpmTest.kt
```diff
@@ -0,0 +1,92 @@
...
+
+import org.ossreviewtoolkit.model.Severity
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.PackageManagerConfiguration
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.utils.common.ProcessCapture
+import org.ossreviewtoolkit.utils.test.createTestTempDir
+
+class NpmTest : StringSpec({
+    "The output of the npm command should be parsed correctly" {
```
fviernau, line 44: Would it make sense to first (prior to this commit) refactor `Npm` to have a function

```
fun getIssues(npmOuput: String): List<Issue>
```

And then write two test cases:
1. a unit test which checks that the return value is as expected. Works without mocking IMO would make the expected behavior more obvious
2. setup stub / answer for `getIssues()` and verify that `NPM` uses it.

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Npm.kt
oheger-bosch: I don't think so. This ensures that output with a different log level coming afterwards is not added to the issue description.

fviernau: Oh, I hadn't understood the code before correctly. Hope I now have....

So, after your change the code keeps the assumption that the output is grouped by prefix.
e.g. all warnings are next to one another and all errors are next to one another.

Should this assumption be kept? (If not, maybe then my proposal here https://github.com/oss-review-toolkit/ort/pull/6418/commits/1d592a67c96c788f92e16d5fc4a8f6cc2377da00#r1091792687 might make sense)

analyzer/src/main/kotlin/managers/Npm.kt
fviernau: Not as (performance) optimization, I meant it as readability optimization, simplification. 
But I understood the code in a wrong way originally.

fviernau (approved):

analyzer/src/main/kotlin/managers/Npm.kt
oheger-bosch: From what I have seen so far, it seems to be reasonable that the output is grouped by prefix; so I would keep it.

analyzer/src/test/kotlin/managers/NpmTest.kt
oheger-bosch: I agree that the test currently takes some effort to check whether the output is correctly parsed. OTOH, this could be used as a basis for further tests. For instance, it could be tested here whether correct command line arguments are provided based on the configuration (which is currently probably not tested).

Maybe the approach to mock the execution of external commands could be useful for a certain kind of tests in general. Then some basic functionality could later be extracted to the test-utils project to make this more convenient.

PR #6421, opened 2023-01-31
Curations with version 2.0 should not apply to pkg patch versions

MarcelBochtler:


model/src/main/kotlin/PackageCuration.kt
```diff
@@ -27,6 +27,11 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
 
 import org.semver4j.Semver
 
+/**
+ * A list of Strings, that are used to identify a version string as a version range in the [PackageCuration]'s version.
+ */
+private val versionRangeIndicators = listOf(",", "~", "*", "+", ">", "<", "=", " - ", "^", ".x", ".X")
```
MarcelBochtler, line 33: Since https://github.com/oss-review-toolkit/ort/pull/8142cdc06a871bc1e66d4b2df82ffcf0fbd20ee5 we implicitly support every version range specified as a `Processor` in semver4j/semver4j [1].
If we want to disable this, and only support Ivy ranges [2] has to be implemented.

[1]: https://github.com/semver4j/semver4j/blob/main/src/main/java/org/semver4j/internal/range/processor/Processor.java
[2]: https://github.com/semver4j/semver4j/issues/133

model/src/main/kotlin/PackageCuration.kt
```diff
@@ -27,6 +27,11 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
 
 import org.semver4j.Semver
 
+/**
```
fviernau, line 30: Looking down the call hierarchy I ended up here:

https://github.com/semver4j/semver4j/blob/3e5362de90b525876753404e61d849636a40120d/src/test/java/org/semver4j/RangesListFactoryTest.java#L38

I now wonder whether `||` and `<=` and so forth need also to be considered.

fviernau (changes requested):

model/src/main/kotlin/PackageCuration.kt
```diff
@@ -74,6 +85,8 @@ data class PackageCuration(
             it.showStackTrace()
         }.getOrDefault(false)
 
+    private fun String.isVersionRange() = versionRangeIndicators.any { contains(it) }
```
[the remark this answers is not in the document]
MarcelBochtler: I switched to `contains(CharSequence, Booklean)` so `it` is no longer treated as regex.

model/src/main/kotlin/PackageCuration.kt
MarcelBochtler: `||` should, yes. `<=` is already covered by `<` and `=`.

mnonnenmacher, line 30: Commit message: Duplicate sign-off.

model/src/main/kotlin/PackageCuration.kt
```diff
@@ -27,6 +27,11 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
 
 import org.semver4j.Semver
 
+/**
+ * A list of Strings, that are used to identify a version string as a version range in the [PackageCuration]'s version.
```
mnonnenmacher, line 31: Nit: No comma after "strings".

model/src/main/kotlin/PackageCuration.kt
```diff
@@ -74,6 +91,8 @@ data class PackageCuration(
             it.showStackTrace()
         }.getOrDefault(false)
 
+    private fun String.isVersionRange() = versionRangeIndicators.any { contains(it, ignoreCase = true) }
```
mnonnenmacher, line 94: Technically you could use `ignoreCase = false` because none of the indicators are letters.

model/src/test/kotlin/PackageCurationTest.kt
```diff
@@ -335,6 +335,52 @@ class PackageCurationTest : WordSpec({
             curation.isApplicable(pkgVersionInsideRange.id) shouldBe true
             curation.isApplicable(pkgVersionOutsideRange.id) shouldBe false
         }
+
+        "not apply if package uses different patch version" {
+            val pkg = Package.EMPTY.copy(
```
mnonnenmacher, line 340: In both tests you only use the `id` so there is no need to create a package.

model/src/main/kotlin/PackageCuration.kt
fviernau, line 94: The code in second in (line 77 onwards) internally does create a range list like this:
```
val curationVersion = "[2.0.0, 2.0.2]"
val rangesList = RangesListFactory.create(curationVersion)
println(rangesList.get().size)
```
This code prints `0` as output due to the white space after `,` in `[2.0.0, 2.0.2]` .
I believe it would be nice to somehow act if the range list size is `0` while there is 
range list chars in the string. If both conditions are met, we could:

1. make  `isVersionRange()` return false
2. probably to hard: throw an exception

What do you think?

model/src/test/kotlin/PackageCurationTest.kt
```diff
@@ -335,6 +335,52 @@ class PackageCurationTest : WordSpec({
...
+                )
+            )
+
+            val curation = PackageCuration(
+                id = Identifier(
+                    type = "Maven",
+                    namespace = "androidx.constraintlayout",
+                    name = "constraintlayout",
+                    version = "2.0"
+                ),
```
mnonnenmacher, line 355: If you apply above comment you could just do `id.copy(version = "2.0")` here.

mnonnenmacher (changes requested):

fviernau (changes requested):

model/src/main/kotlin/PackageCuration.kt
MarcelBochtler: `.x` is.

MarcelBochtler: I added a check for the case when `id.version` looks like a version range, but semver4j creates an invalid version range with the size of `0`.

mnonnenmacher: Missed that.

mnonnenmacher (dismissed):

fviernau (dismissed):

fviernau (approved):

model/src/main/kotlin/PackageCuration.kt
```diff
@@ -64,7 +70,24 @@ data class PackageCuration(
...
         runCatching {
-            Semver.coerce(pkgId.version).satisfies(id.version)
+            if (Semver.isValid(id.version)) {
+                // If the curation is for a valid semver, do not coerce the package version, as it also has to be valid.
+                return id.version == pkgId.version
+            }
+
+            if (id.version.isVersionRange()) {
+                // `Semver.satisfies(String)` requires a valid version range to work as expected, see:
+                // https://github.com/semver4j/semver4j/issues/132.
```
sschuberth, line 80: This upstream issue has been closed meanwhile. Is it still sensible to mention it here?

model/src/main/kotlin/PackageCuration.kt
```diff
@@ -64,7 +70,24 @@ data class PackageCuration(
      */
     private fun isApplicableIvyVersion(pkgId: Identifier) =
         runCatching {
-            Semver.coerce(pkgId.version).satisfies(id.version)
+            if (Semver.isValid(id.version)) {
+                // If the curation is for a valid semver, do not coerce the package version, as it also has to be valid.
```
sschuberth, line 74: Is this related to https://github.com/semver4j/semver4j/issues/130?

model/src/main/kotlin/PackageCuration.kt
MarcelBochtler: I think this can be removed now.
However, we still need to handle this case in our code.

model/src/main/kotlin/PackageCuration.kt
MarcelBochtler: It is related, but I actually added this check due to issue that 2.0.0 matched 2.0.0.3.
But investigating this more revealed even more issues.
We might need to remove the `coerce()` call for the version altogether: https://github.com/oss-review-toolkit/ort/pull/6430

sschuberth: The more I think about all the issues we face, the more I come to believe we should create our own Kotlin multi-platform version comparison library...

PR #6422, opened 2023-01-31
Ignore empty curation files

mnonnenmacher:

### DIFF
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -105,7 +105,7 @@ class FilePackageCurationProvider(
                 }
 
                 if (it.isDirectory) FileFormat.findFilesWithKnownExtensions(it) else listOf(it)
-            }
+            }.filterNot { it.length() == 0L }
 
             curationFiles.map { curationFile ->
                 val curations = runCatching {
--- a/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/FilePackageCurationProvider.kt
@@ -111,7 +111,7 @@ class FilePackageCurationProvider(
                 val curations = runCatching {
                     curationFile.readValue<List<PackageCuration>>()
                 }.getOrElse {
-                    throw IOException("Failed parsing package curation from '${curationFile.absolutePath}'.", it)
+                    throw IOException("Failed parsing package curation(s) from '${curationFile.absolutePath}'.", it)
                 }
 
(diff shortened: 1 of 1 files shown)


oheger-bosch (approved):

sschuberth:
I remember to have had this discussion several times, and there was already a time when we simply ignored empty files via one of our special deserialization helper functions, but IIRC so far @fviernau was always against it, as he regarded empty files to be an error, and wanted to fail hard on errors.

fviernau:
> I remember to have had this discussion several times, and there was already a time when we simply ignored empty files via one of our special deserialization helper functions, but IIRC so far @fviernau was always against it, as he regarded empty files to be an error, and wanted to fail hard on errors.

I still believe in general we should keep on failing hard.
My personal preference is still to do so in all cases in which the given file is not in expected format, because I believe
it is just fine to ask the user to have the provided file in expected format. E.g. a valid empty file would have the content `[]` IIRC. Anyhow, if it's just about 0 byte content files, I wouldn't discuss any further and go with your guys decision.

fviernau (approved):

sschuberth:
> via one of our special deserialization helper functions

BTW, this is the function I was referring to and that I would have proposed to use:

https://github.com/oss-review-toolkit/ort/blob/77555b8772046360f4cfcb03932cf1a228ded15d/model/src/main/kotlin/FileFormat.kt#L101-L108

mnonnenmacher:
> > via one of our special deserialization helper functions
> 
> BTW, this is the function I was referring to and that I would have proposed to use:
> 
> https://github.com/oss-review-toolkit/ort/blob/77555b8772046360f4cfcb03932cf1a228ded15d/model/src/main/kotlin/FileFormat.kt#L101-L108

@sschuberth "two-step process" makes it sound like it has a performance impact probably larger than `file.length() == 0`, is this correct?

sschuberth:
> @sschuberth "two-step process" makes it sound like it has a performance impact probably larger than `file.length() == 0`, is this correct?

I don't think to. Internally, Jackson seems to use two different code paths calling `_readValue` and `_convert`, respectively. And the latter is commented with:

> Actual conversion implementation: instead of using existing read and write methods, much of code is inlined. Reason for this is that we must avoid root value wrapping/unwrapping both for efficiency and for correctness. If root value wrapping/unwrapping is actually desired, caller must use explicit writeValue and readValue methods.

So maybe the "two-step" approach is even faster 😉

sschuberth:
One more note: I introduced `readValueOrNull()` because when first coming across the issue of "empty" files, it actually contained (just) a newline. Such a file would not be filtered correctly with the current code.

PR #6430, opened 2023-02-02
fix(PackageCuration): Don't apply curations with different version

MarcelBochtler:
Neither the package's nor the curation's version are guaranteed to use
valid semvers. Semver4j's `coerce()` function creates a valid semver in
cases when this is not wanted, e.g. 1.2.3.4 is parsed as 1.2.3 and
therefore a package curation for 1.2.3.9 would be seen as applicable.

Remove the coercing to valid semvers to avoid applying incorrect package
curations.

### DIFF
--- a/model/src/main/kotlin/PackageCuration.kt
+++ b/model/src/main/kotlin/PackageCuration.kt
@@ -70,10 +70,7 @@ data class PackageCuration(
      */
     private fun isApplicableIvyVersion(pkgId: Identifier) =
         runCatching {
-            if (Semver.isValid(id.version)) {
-                // If the curation is for a valid semver, do not coerce the package version, as it also has to be valid.
-                return id.version == pkgId.version
-            }
+            if (id.version == pkgId.version) return true
 
             if (id.version.isVersionRange()) {
                 // `Semver.satisfies(String)` requires a valid version range to work as expected, see:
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/PackageCuration.kt
```diff
@@ -86,8 +83,7 @@ data class PackageCuration(
                 return Semver.coerce(pkgId.version).satisfies(range)
             }
 
-            // `id.version` is invalid, but also not a version range. Therefore, coerce valid versions.
-            Semver.coerce(pkgId.version).satisfies(Semver.coerce(id.version).version)
+            return false
```
fviernau, line 86: I don't understand why we can keep on doing `coerce(pkgId.version)` in line 83, while `coerce(id.version)` is not okay anymore. Can you help?

fviernau (changes requested):

model/src/main/kotlin/PackageCuration.kt
fviernau: Thanks! I tried to find a case where the remaining `coerce()` can be problematic.
If we take above range `[1.0.0, 2.0.0]` and `pkg.id = "2.0.0.127"`, I believe `coerce("2.0.0.127")` yields `2.0.0`, so the range would be matched.

I wonder if it would work to solve the above mentioned use cases by doing the matching on a kind of "normalized" version string, whereas "normalize" would have effect only on version strings with less than 3 digits, like:
- `normalize("1") -> 1.0.0`
- `normalize("1.1) -> 1.1.0`

What do you think about that?

fviernau: Hm, even if this would make sense I believe it can be considered out of scope of this change / separate iteration. So, I'll just approve this one as-is.

fviernau (approved):

Issue #6419, opened 2023-01-31
ORT Scan started to fail when use a Gradle version lower than 7.2.0


abanias:
The ort scan job started to fail with the following error: 
```
Resolving Gradle dependencies for path 'build.gradle' failed with: BuildException: Could not fetch model of type 'DependencyTreeModel' using connection to Gradle distribution 'https://services.gradle.org/distributions/gradle-6.9-bin.zip'. Caused by: LocationAwareException: Initialization script '/tmp/ort-Gradle4749349623269099033/init15891916940425317509.gradle' line: 179 Could not compile initialization script '/tmp/ort-Gradle4749349623269099033/init15891916940425317509.gradle'. Caused by: ContextualPlaceholderException: Could not compile initialization script '/tmp/ort-Gradle4749349623269099033/init15891916940425317509.gradle'. Caused by: PlaceholderException: startup failed: initialization script '/tmp/ort-Gradle4749349623269099033/init15891916940425317509.gradle': 179: unable to resolve class org.gradle.internal.composite.IncludedBuildInternal @ line 179, column 29. def state = (it as org.gradle.internal.composite.IncludedBuildInternal).getTarget() ^ 1 error
```
The problem arose because of this change in the ort project: https://github.com/oss-review-toolkit/ort/commit/6cce314c695e9e4071ef3f8a7956b2683cf2b9d5#diff-e0c99299dd78a854b2dc32308b0222986441ce3d73fea8e2efc1cabb6e39a179Ra177

Our project uses Gradle 6.9, but the getTarget() method, used in the changes above, doesn't exist in that Gradle version.
This method was added in Gradle 7.2.0.

nnobelis:
Thanks for analysis !

Out of curiosity, does your project use the composite build feature ?
Because I don't think the error occurs if the build does not have any included builds.

@mnonnenmacher @sschuberth I guess with will have to condition the composite build logic to the Gradle version ?
Or do you see another workaround ?

abanias:
Our project doesn't use the composite build feature.
It's not even a runtime error. It's a compilation error because Gradle 6.9 can not find the getTarget() method in the org.gradle.internal.composite.IncludedBuildInternal class.

sschuberth:
> I guess with will have to condition the composite build logic to the Gradle version ?

[...]

PR #6431, opened 2023-02-02
fix(Gradle): Fix regression when project uses Gradle < 7.2.0

nnobelis:
In a previous commit ([1]), support for Composite build was added using Gradle Internal API. Unfortunately, this introduced a regression for all projects using Gradle < 7.2.0 which could not be analyzed anymore. This commit fixes the regression by using reflection to call the internal API.
For projects running Gradle < 7.2.0 and using the composite build feature, the behavior is restored to the previous one i.e. it will throw a `NullPointerException` when accessing the group ID of a subproject from an included build.

Solves https://github.com/oss-review-toolkit/ort/issues/6419.

[1]: https://github.com/oss-review-toolkit/ort/commit/6cce314c695e9e4071ef3f8a7956b2683cf2b9d5

### DIFF
--- a/analyzer/src/main/resources/scripts/init.gradle
+++ b/analyzer/src/main/resources/scripts/init.gradle
@@ -175,9 +175,10 @@ class AbstractDependencyTreePlugin<T> implements Plugin<T> {
             project.gradle.getIncludedBuilds().each {
                 project.logger.quiet("Processing included build '$it.name, $it.projectDir'.")
                 // Since the projects of an included build are not present in the public API, the internal API is used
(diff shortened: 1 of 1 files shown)


mnonnenmacher (approved):

nnobelis:
@mnonnenmacher  can we please merge this ?

The CI test failure is unrelated.

mnonnenmacher:
@nnobelis Just noticed that you should have used "Fixes" instead of "Solves" to auto-close the issue. Anyway, I'll close it manually.

fviernau:
> The CI test failure is unrelated.

Just fixed it here https://github.com/oss-review-toolkit/ort/pull/6434.

PR #6435, opened 2023-02-03
AnalyzerCommand: Fix a file path in the log output

fviernau:
Log the file path of the actually use repository configuration instead of the default file location.


mnonnenmacher (dismissed):

cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -197,8 +197,8 @@ class AnalyzerCommand : OrtCommand(
                 add(SimplePackageCurationProvider(repositoryPackageCurations))
             } else if (repositoryPackageCurations.isNotEmpty()) {
                 logger.warn {
-                    "Existing package curations from '$ORT_REPO_CONFIG_FILENAME' are not applied because the feature " +
-                            "is disabled."
+                    "Existing package curations from '${repositoryConfigurationFile.absolutePath}' are not applied " +
```
mnonnenmacher, line 200: Commit message nit: "use**d**"

mnonnenmacher (changes requested):

sschuberth (approved):

PR #6436, opened 2023-02-03
Fix-up `PackageRule.uncuratedPkg` to be not null again

fviernau:
See individual commits.

### DIFF
--- a/evaluator/src/main/kotlin/PackageRule.kt
+++ b/evaluator/src/main/kotlin/PackageRule.kt
@@ -52,7 +52,7 @@ open class PackageRule(
     private val licenseRules = mutableListOf<LicenseRule>()
 
     @Suppress("UNUSED") // This is intended to be used by rule implementations.
-    val uncuratedPkg by lazy { ruleSet.ortResult.getUncuratedPackageOrProject(pkg.metadata.id) }
+    val uncuratedPkg: Package? by lazy { ruleSet.ortResult.getUncuratedPackageOrProject(pkg.metadata.id) }
 
     override val description = "Evaluating rule '$name' for package '${pkg.metadata.id.toCoordinates()}'."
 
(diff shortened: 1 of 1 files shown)


evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -52,7 +52,7 @@ open class PackageRule(
     private val licenseRules = mutableListOf<LicenseRule>()
 
     @Suppress("UNUSED") // This is intended to be used by rule implementations.
-    val uncuratedPkg by lazy { ruleSet.ortResult.getUncuratedPackageOrProject(pkg.metadata.id) }
+    val uncuratedPkg: Package? = by lazy { ruleSet.ortResult.getUncuratedPackageOrProject(pkg.metadata.id) }
```
mnonnenmacher: Remove `=`.

mnonnenmacher (changes requested):

mnonnenmacher:
For the commitlint issue you would have to adapt the `noteKeywords` to allow `^`. However, you also just inline the short SHA1 in the commit message like we usually do.

MarcelBochtler (approved):

mnonnenmacher (approved):

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -52,7 +52,7 @@ open class PackageRule(
     private val licenseRules = mutableListOf<LicenseRule>()
 
     @Suppress("UNUSED") // This is intended to be used by rule implementations.
-    val uncuratedPkg: Package? by lazy { ruleSet.ortResult.getUncuratedPackageOrProject(pkg.metadata.id) }
+    val uncuratedPkg: Package by lazy { ruleSet.ortResult.getUncuratedPackageOrProject(pkg.metadata.id)!! }
```
sschuberth, line 55: A hint in the commit message why we can assume the return value of `getUncuratedPackageOrProject()` to never be `null` here would have been nice.

PR #6447, opened 2023-02-06
refactor(PackageCurationProvider): Align on camel case IDs

fviernau:
The values of the IDs default to the values of `type` which are all in camel case. So, consistently use camel case for all IDs.

### DIFF
--- a/analyzer/src/main/kotlin/PackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/PackageCurationProvider.kt
@@ -59,8 +59,8 @@ interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<Pac
                 }
 
                 require(none { (id, _) -> id == REPOSITORY_CONFIGURATION_PROVIDER_ID }) {
-                    "Found package curation provider which uses '$REPOSITORY_CONFIGURATION_PROVIDER_ID' as id which " +
-                             "is reserved and not allowed."
+                    "Found a package curation provider which uses '$REPOSITORY_CONFIGURATION_PROVIDER_ID' as its id " +
+                            "which is reserved and not allowed."
                 }
             }
     }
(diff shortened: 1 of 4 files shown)


analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -32,7 +32,7 @@ import org.ossreviewtoolkit.utils.common.getDuplicates
 interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<PackageCurationProvider> {
     companion object {
         val ALL = Plugin.getAll<PackageCurationProviderFactory<*>>()
-        const val ORT_YML_PROVIDER_ID = "ort-yml"
+        const val ORT_YML_PROVIDER_ID = "OrtYml"
```
sschuberth, line 35: As discussed, probably also rename this to `RepositoryConfiguration` to make more clear where the `.ort.yml` comes from (and that it actually also could be a JSON).

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -59,8 +59,8 @@ interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<Pac
                 }
 
                 require(none { (id, _) -> id == REPOSITORY_CONFIGURATION_PROVIDER_ID }) {
-                    "Found package curation provider which uses '$REPOSITORY_CONFIGURATION_PROVIDER_ID' as id which " +
-                             "is reserved and not allowed."
+                    "Found a package curation provider which uses '$REPOSITORY_CONFIGURATION_PROVIDER_ID' as id " +
```
sschuberth, line 62: Commit message: s/erro/error

While at it, maybe also say "as *the* id" or "as *its* id"?

sschuberth (changes requested):

sschuberth (approved):

sschuberth (approved):

PR #6449, opened 2023-02-06
feature(Analyzer): resolveConfiguration() improvements

fviernau:
This allows identifying slow providers. While at it, promote the log level to info, as it seems a better fit.

### DIFF
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -22,7 +22,7 @@ package org.ossreviewtoolkit.analyzer
 import java.io.File
 import java.time.Instant
 
-import kotlin.time.measureTime
+import kotlin.time.measureTimedValue
 
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -344,13 +344,15 @@ private fun resolveConfiguration(
     val packageIds = analyzerResult.packages.mapTo(mutableSetOf()) { it.id }
     val packageCurations = mutableListOf<PackageCuration>()
 
-    val duration = measureTime {
-        curationProviders.forEach { (_, curationProvider) ->
-            packageCurations += curationProvider.getCurationsFor(packageIds).values.flatten()
+    curationProviders.forEach { (id, curationProvider) ->
+        val (curations, duration) = measureTimedValue {
+            curationProvider.getCurationsFor(packageIds).values.flatten()
         }
-    }
 
-    Analyzer.logger().debug { "Getting package curations took $duration." }
+        packageCurations += curations
+
+        Analyzer.logger().info { "Getting ${curations.size} package curation(s) from provider '$id' took $duration." }
+    }
 
     return ResolvedConfiguration(packageCurations)
 }
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -346,7 +346,7 @@ private fun resolveConfiguration(
 
     curationProviders.forEach { (id, curationProvider) ->
         val (curations, duration) = measureTimedValue {
-            curationProvider.getCurationsFor(packageIds).values.flatten()
+            curationProvider.getCurationsFor(packageIds).values.flatten().distinct()
         }
 
         packageCurations += curations


analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -344,13 +344,13 @@ private fun resolveConfiguration(
...
-    val duration = measureTime {
-        curationProviders.forEach { (_, curationProvider) ->
+    curationProviders.forEach { (id, curationProvider) ->
+        val duration = measureTime {
             packageCurations += curationProvider.getCurationsFor(packageIds).values.flatten()
         }
-    }
 
-    Analyzer.logger().debug { "Getting package curations took $duration." }
+        Analyzer.logger().info { "Getting package curations from provider '$id' took $duration." }
```
sschuberth, line 352: While at it, would also make sense to log the number of curations retrieved from each provider?

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -344,13 +344,15 @@ private fun resolveConfiguration(
...
+    curationProviders.forEach { (id, curationProvider) ->
+        val (curations, duration) = measureTimedValue {
+            curationProvider.getCurationsFor(packageIds).values.flatten()
         }
-    }
 
-    Analyzer.logger().debug { "Getting package curations took $duration." }
+        packageCurations += curations
+
+        Analyzer.logger().info { "Getting ${curations.size} package curations from provider '$id' took $duration." }
```
sschuberth, line 354: Nit: The message should now probably say "package curation(s)".

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -344,13 +344,15 @@ private fun resolveConfiguration(
...
+    }
 
-    return ResolvedConfiguration(packageCurations)
+    return ResolvedConfiguration(packageCurations.distinct())
```
sschuberth, line 357: Hmm... besides this being a functional change compared to before, I wondering if we can safely do that. Assume curations A and B, where B undoes A. Then having A, B, A *without* `distinct()` would result in A, whereas *with* `distinct()` it would result A, B (which would be a no-op).

fviernau: Hm, I've read your message again. So, if we agree on "no order defined within single provider" then we should have
one distinct call per provider.

sschuberth: > So, if we agree on "no order defined within single provider" then we should have one distinct call per provider. Agree?

I'm not sure yet. While thinking about it, I'd to understand a bit more about the actual problem you're trying to solve: Did you really run into the case of unwanted duplicate curations, or is this just a theoretical thing for now?

sschuberth: BTW, I just came across 8aeb20566e1404979d7a3667d3de2e4f48e319d3, so I guess it's also ok to do that for all providers (and in exchange remove that code from `ClearlyDefinedPackageCurationProvider`).

sschuberth: BTW, I just came across 8aeb20566e1404979d7a3667d3de2e4f48e319d3, so I guess it's also ok to do that for all providers (and in exchange remove that code from `ClearlyDefinedPackageCurationProvider`).

fviernau: > I'm not sure yet. While thinking about it, I'd to understand a bit more about the actual problem you're trying to solve: Did you really run into the case of unwanted duplicate curations, or is this just a theoretical thing for now?

When implementing the separation of curations and packages in analyzer result, one of the reasons / benefits was to me always that it would drop redundant curations, because a curation may omit the `name` part of the `identifier`. I just forgot about adding this distinct call. So, given that a curation can match multiple package I believe it's not theoretical. Practical use case I remember: Add metadata to all curations under the android standard packages namespace.

fviernau: > BTW, I just came across [8aeb205](https://github.com/oss-review-toolkit/ort/commit/8aeb20566e1404979d7a3667d3de2e4f48e319d3), so I guess it's also ok to do that for all providers (and in exchange remove that code from `ClearlyDefinedPackageCurationProvider`).

Ok, so I would keep the code as-is right now (after my recent force push), and add a commit on top which drops the logic from clearly defined. Would that be the correct understanding of your proposal?

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -346,7 +346,7 @@ private fun resolveConfiguration(
 
     curationProviders.forEach { (id, curationProvider) ->
         val (curations, duration) = measureTimedValue {
-            curationProvider.getCurationsFor(packageIds).values.flatten()
+            curationProvider.getCurationsFor(packageIds).values.flatten().distinct()
```
sschuberth, line 349: Commit message: I do not understand how the fact that "a single curation may apply to multiple packages" is a precondition for having duplicates (as that's how the sentence sounds to me). Even without applying to multiple packages, i.e with hard-coded ids without ranges, two different files in a curations directory could be copies of each other, resulting in a duplicate.

So maybe just say: For curations for a given id from a single provider no ordering is assumed. As such, having duplicates (like by accidental copies of files in a `FilePackageCurationProvider`) does not make any sense. So remove any duplicates to be on the safe side.

sschuberth (changes requested):

analyzer/src/main/kotlin/Analyzer.kt
fviernau: Hm, from the perspective of the `ClearlyDefinedProvider` (without its usages from within ORT code) I believe it makes sense to keep the duplicate filtering there, in particular for programmatic use of the curation provider.
If you are fine let's keep it there just for now:

Note: I'll propose (PR) a simplification of the provider API (on top of your API change) which might bring a bit more clarity if it gets accepted.

sschuberth (approved):

fviernau:
Merge despite unrelated failures.

PR #6451, opened 2023-02-07
refactor(Analyzer)!: Improve package curation ordering

fviernau:
Conistently order the package curations or the providers repsectively by higherst-priority first. This avoids confusion compared to dealing with inconsistent ordering. This way only the logic which actually applies the curations needs to reverse the curations.

Add or update the KDoc of all releated functions and properties so that the order is always specified.

BREAKING CHANGE: There is no backwards compatibility for de-serializing ORT files prior to this change. Deserialization is possible, but leads to reversing the curation order. Note that recently deserialization has been broken anyway and very soon there will be another breaking change. So, it does not really matter whether this change is breaking as we decided for breaking it anyway.

### DIFF
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -120,6 +120,10 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
         return ManagedFileInfo(absoluteProjectPath, managedFiles, repositoryConfiguration)
     }
 
+    /**
+     * Return the result of analyzing the given [managed file][info]. The given [curationProviders] must be ordered
+     * highest-priority-first.
+     */
     @JvmOverloads
     fun analyze(
         info: ManagedFileInfo,
@@ -344,6 +348,10 @@ private class PackageManagerRunner(
     }
 }
 
+/**
+ * Return the resolved configuration for the given [analyzerResult]. The [curationProviders] must be ordered
+ * highest-priority-first.
+ */
 private fun resolveConfiguration(
     analyzerResult: AnalyzerResult,
     curationProviders: List<Pair<String, PackageCurationProvider>>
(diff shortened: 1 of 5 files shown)


analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -114,6 +114,10 @@ class Analyzer(private val config: AnalyzerConfiguration, private val labels: Ma
         return ManagedFileInfo(absoluteProjectPath, managedFiles, repositoryConfiguration)
     }
 
+    /**
```
mnonnenmacher, line 117: Commit message nit: Typos in "respectively", "highest"

mnonnenmacher (changes requested):

model/src/main/kotlin/OrtResult.kt
```diff
@@ -529,6 +529,11 @@ data class OrtResult(
         }
 }
 
+/**
+ * Returns a set containing exactly one [CuratedPackage] for each given [Package], derived from applying all
```
sschuberth, line 533: "Return" (without "s").

sschuberth:
> There is no backwards compatibility for de-serializing ORT files prior to this change. Deserialization is possible, but leads to reversing the curation order.

So, doesn't this mean that the comment / order in `reference.yml` also needs to be adjust as part of this PR?

https://github.com/oss-review-toolkit/ort/blob/0a27a3813c88014572190808c341a23424fd566a/model/src/main/resources/reference.yml#L39-L68

mnonnenmacher (approved):

fviernau:
> So, doesn't this mean that the comment / order in `reference.yml` also needs to be adjust as part of this PR?

No, this change affects only the order of the list of String inside `OrtResult` (besides Kotlin code).

sschuberth:
> entries are ordered in highest-priority-first, which hasn't been changed. There is also no plan to change it, as the idea of this PR was to use this order everywhere.

Ok, great, I just wanted to double-check; I probably was mislead and somehow though it was exactly the opposite.

PR #6463, opened 2023-02-09
Simplify and clarify the API of package curation providers

fviernau:
See individual commits.

### DIFF
--- a/analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
+++ b/analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
@@ -20,10 +20,9 @@
 package org.ossreviewtoolkit.analyzer.curation
 
 import io.kotest.core.spec.style.WordSpec
-import io.kotest.matchers.maps.beEmpty
-import io.kotest.matchers.maps.haveSize
+import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.collections.shouldHaveSingleElement
 import io.kotest.matchers.should
-import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNot
 
 import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService.Server
@@ -40,8 +39,7 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
 
             val curations = provider.getCurationsFor(packages)
 
-            curations should haveSize(1)
-            curations.values.flatten().first().data.concludedLicense shouldBe
+            curations.map { it.data.concludedLicense } shouldHaveSingleElement
                     "CDDL-1.0 OR GPL-2.0-only WITH Classpath-exception-2.0".toSpdx()
         }
 
@@ -50,8 +48,7 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
 
             val curations = provider.getCurationsFor(packages)
 
-            curations should haveSize(1)
-            curations.values.flatten().first().data.vcs?.revision shouldBe "0b97c416e42a184ff9728877b461c616187c58f7"
+            curations.map { it.data.vcs?.revision } shouldHaveSingleElement "0b97c416e42a184ff9728877b461c616187c58f7"
         }
 
         "return no curation for a non-existing dummy NPM package" {
@@ -71,8 +68,7 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
 
             val curations = provider.getCurationsFor(packages)
 
-            curations should haveSize(1)
-            curations.values.flatten().first().data.concludedLicense shouldBe "Apache-1.0".toSpdx()
+            curations.map { it.data.concludedLicense } shouldHaveSingleElement "Apache-1.0".toSpdx()
         }
 
         "return no curation for a non-existing dummy Maven package" {
--- a/analyzer/src/main/kotlin/Analyzer.kt
+++ b/analyzer/src/main/kotlin/Analyzer.kt
@@ -360,7 +360,7 @@ private fun resolveConfiguration(
 
     curationProviders.forEach { (id, curationProvider) ->
         val (curations, duration) = measureTimedValue {
-            curationProvider.getCurationsFor(analyzerResult.packages).values.flatten().distinct()
+            curationProvider.getCurationsFor(analyzerResult.packages).distinct()
         }
 
         packageCurations += curations
--- a/analyzer/src/main/kotlin/PackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/PackageCurationProvider.kt
@@ -19,7 +19,6 @@
 
 package org.ossreviewtoolkit.analyzer
 
-import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageCuration
 import org.ossreviewtoolkit.model.config.PackageCurationProviderConfiguration
@@ -83,10 +82,8 @@ interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<Pac
  */
 fun interface PackageCurationProvider {
     /**
-     * Return all available [PackageCuration]s for the provided [packages], associated by the package's [Identifier].
-     * Each list of curations must be non-empty; if no curation is available for a package, the returned map must not
-     * contain a key for that package's identifier at all.
+     * Return all available [PackageCuration]s which are applicable to any of the given [packages].
      */
     // TODO: Maybe make this a suspend function, then all implementing classes could deal with coroutines more easily.
-    fun getCurationsFor(packages: Collection<Package>): Map<Identifier, List<PackageCuration>>
+    fun getCurationsFor(packages: Collection<Package>): List<PackageCuration>
 }
--- a/analyzer/src/main/kotlin/PackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/PackageCurationProvider.kt
@@ -85,5 +85,5 @@ fun interface PackageCurationProvider {
      * Return all available [PackageCuration]s which are applicable to any of the given [packages].
      */
     // TODO: Maybe make this a suspend function, then all implementing classes could deal with coroutines more easily.
-    fun getCurationsFor(packages: Collection<Package>): List<PackageCuration>
+    fun getCurationsFor(packages: Collection<Package>): Set<PackageCuration>
 }
--- a/analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
+++ b/analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
@@ -61,7 +61,7 @@ open class OrtConfigPackageCurationProvider : PackageCurationProvider {
         }
     }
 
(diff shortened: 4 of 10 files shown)


analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
```diff
@@ -61,10 +61,10 @@ open class OrtConfigPackageCurationProvider : PackageCurationProvider {
...
     }
 
-    override fun getCurationsFor(packages: Collection<Package>): Map<Identifier, List<PackageCuration>> =
-        packages.mapNotNull { pkg ->
-            getCurationsFor(pkg.id).takeUnless { it.isEmpty() }?.let { pkg.id to it }
-        }.toMap()
+    override fun getCurationsFor(packages: Collection<Package>): List<PackageCuration> =
+        packages.flatMap { pkg ->
+            getCurationsFor(pkg.id)
+        }.distinct()
```
sschuberth, line 67: Nit: This could be come a one-liner now to align with `Sw360PackageCurationProvider`.

analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
```diff
@@ -40,18 +39,17 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
 
             val curations = provider.getCurationsFor(packages)
 
-            curations should haveSize(1)
-            curations.values.flatten().first().data.concludedLicense shouldBe
+            curations.map { it.data.concludedLicense }.shouldContainExactly(
```
sschuberth, line 42: Prefer `shouldHaveSingleElement` to benefit from infix notation here and below.

sschuberth (changes requested):

analyzer/src/test/kotlin/curation/FilePackageCurationProviderTest.kt
```diff
@@ -46,38 +45,23 @@ class FilePackageCurationProviderTest : StringSpec() {
...
+                "Maven:org.ossreviewtoolkit::"
+            ).map { Identifier(it) }
 
-            val curations = provider.getCurationsFor(packages)
-
-            curations.keys shouldContainExactlyInAnyOrder idsWithExistingCurations
-            curations.values.forAll {
-                it should haveSize(1)
-            }
+            provider.packageCurations.map { it.id }.shouldContainAll(expectedCurationIds)
```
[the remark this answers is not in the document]
fviernau: > And here, the matcher could actually stay to be shouldContainExactlyInAnyOrder.

Just FYI: The reason I went for changing was to make the assertion stronger than necessary, which makes tests a bit more resilient wrt changes. 

Anyhow, I've changed it.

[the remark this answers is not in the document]
sschuberth: > The reason I went for changing was to make the assertion stronger than necessary

Hmm, but isn't `shouldContainAll` a *weaker* assertion than `shouldContainExactlyInAnyOrder`, as it only checks that [the expected is a subset of the actual collection](https://kotest.io/docs/assertions/collection-matchers.html)?

analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
```diff
@@ -20,10 +20,9 @@
 package org.ossreviewtoolkit.analyzer.curation
 
 import io.kotest.core.spec.style.WordSpec
-import io.kotest.matchers.maps.beEmpty
-import io.kotest.matchers.maps.haveSize
+import io.kotest.matchers.collections.beEmpty
```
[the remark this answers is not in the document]
fviernau: I've updated the commit message to incorporate your proposals.

analyzer/src/main/kotlin/PackageCurationProvider.kt
```diff
@@ -83,10 +82,8 @@ interface PackageCurationProviderFactory<CONFIG> : ConfigurablePluginFactory<Pac
  */
 fun interface PackageCurationProvider {
     /**
-     * Return all available [PackageCuration]s for the provided [packages], associated by the package's [Identifier].
-     * Each list of curations must be non-empty; if no curation is available for a package, the returned map must not
-     * contain a key for that package's identifier at all.
+     * Return all available [PackageCuration]s which are applicable to any of the given [packages].
```
sschuberth, line 85: In the context of you commit message, we should probably make more clear here that "applicable" does *not* refer to `PackageCuration.isApplicable()`. Yes, the providers for ORT's own curation format make that call, but we don't enforce other implementers to make that call: They could return any curations that would not pass `PackageCuration.isApplicable()`. Which actually is a short-coming of our API that we should think about, IMO.

sschuberth: So actually, I'm currently not seeing how this is true:

> In particular, make it impossible that providers return a matching inconsistent with `PackageCuration.isApplicable()`.

fviernau: > So actually, I'm currently not seeing how this is true:

I meant that things can only be inconsistent if they amount to at least 2 things. 
One thing cannot be inconsistent as there is nothing it can be compared with.

Before this change: A curation matches if either

1. The curation is associated with the package id in the returned `Map`
2. or the `isApplicable()` returns true for a particular package

So, 1 and 2 could be inconsistent.
I removed 1, so we have a single source of truth: `isApplicable()`.

fviernau: > Yes, the providers for ORT's own curation format make that call, but we don't enforce other implementers to make that call

I don't see how this can be true. Because with the decision we made together to separate `PackageCuration`s from `Package`s in `OrtResult` implies that matching must rely on `isApplicable`, because otherwise the association was lost.

sschuberth: I'm talking about the curations returned directly from `getCurationsFor()`. There is no guarantee that these pass `isApplicable()` for all provider implementations. But ideally, that's something we should do, as it would safe us the need to check for `isApplicable()` when writing curations to the ORT result.

analyzer/src/test/kotlin/curation/FilePackageCurationProviderTest.kt
```diff
@@ -46,38 +46,23 @@ class FilePackageCurationProviderTest : StringSpec() {
...
-            curations.values.forAll {
-                it should haveSize(1)
-            }
+            provider.packageCurations.map { it.id } shouldContainExactlyInAnyOrder expectedCurationIds
         }
 
         "Provider can read from a single file and directories" {
             val provider = FilePackageCurationProvider(curationsFile, curationsDir)
+            val expectedCurationIds = listOf(
+                "NPM::ramda:[0.21.0,0.25.0]",
```
fviernau, line 60: Add comments back.

analyzer/src/funTest/kotlin/curation/ClearlyDefinedPackageCurationProviderFunTest.kt
sschuberth, line 23: Commit message proposal as discussed:

```
refactor(PackageCurationProvider)!: Make `getCurationsFor()` return a list

Previously, `getCurationsFor()` return a map keyed by the `Identifier` of
the `Package` that was used to request the curation. For curations that
match multiple ids, e.g. because the use of version ranges, this resulted
in the same curation being returned multiple times for different ids.

To avoid these duplicated, and because all callers of `getCurationsFor()`
anyway call `flatten()` on the maps's values only, drop the keys by simply
returning a list instead.

Note that the `PackageCuration.isApplicable()` needs to happen
independently of that before writing curation to the ORT result, as there
is no guarantee that all curation provider implementations only return
applicable curations in the ORT sense.
```

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -360,7 +360,7 @@ private fun resolveConfiguration(
 
     curationProviders.forEach { (id, curationProvider) ->
         val (curations, duration) = measureTimedValue {
-            curationProvider.getCurationsFor(analyzerResult.packages).values.flatten().distinct()
+            curationProvider.getCurationsFor(analyzerResult.packages)
```
sschuberth, line 363: Shouldn't the `distinct()` here be only dropped in the next commit?

analyzer/src/main/kotlin/curation/OrtConfigPackageCurationProvider.kt
```diff
@@ -61,10 +61,8 @@ open class OrtConfigPackageCurationProvider : PackageCurationProvider {
         }
     }
 
-    override fun getCurationsFor(packages: Collection<Package>): Map<Identifier, List<PackageCuration>> =
-        packages.mapNotNull { pkg ->
-            getCurationsFor(pkg.id).takeUnless { it.isEmpty() }?.let { pkg.id to it }
-        }.toMap()
+    override fun getCurationsFor(packages: Collection<Package>): List<PackageCuration> =
+        packages.flatMap { pkg -> getCurationsFor(pkg.id) }.distinct()
```
sschuberth: I believe `distinct()` should not be added here to more closely resemble the previous code, and because it would be dropped in the next commit anyway.

analyzer/src/main/kotlin/curation/SimplePackageCurationProvider.kt
```diff
@@ -20,16 +20,15 @@
...
  */
 open class SimplePackageCurationProvider(val packageCurations: List<PackageCuration>) : PackageCurationProvider {
-    override fun getCurationsFor(packages: Collection<Package>): Map<Identifier, List<PackageCuration>> =
-        packages.mapNotNull { pkg ->
-            packageCurations.filter { it.isApplicable(pkg.id) }.takeUnless { it.isEmpty() }?.let { pkg.id to it }
-        }.toMap()
+    override fun getCurationsFor(packages: Collection<Package>): List<PackageCuration> =
+        packageCurations.filter { curation ->
+            packages.any { pkg -> curation.isApplicable(pkg.id) }
+        }.distinct()
```
sschuberth: I believe `distinct()` should not be added here to more closely resemble the previous code, and because it would be dropped in the next commit anyway.

analyzer/src/main/kotlin/curation/Sw360PackageCurationProvider.kt
```diff
@@ -98,10 +98,8 @@ class Sw360PackageCurationProvider(config: Sw360StorageConfiguration) : PackageC
     private val connectionFactory = createConnection(config)
     private val releaseClient = connectionFactory.releaseAdapter
 
-    override fun getCurationsFor(packages: Collection<Package>): Map<Identifier, List<PackageCuration>> =
-        packages.mapNotNull { pkg ->
-            getCurationsFor(pkg.id).takeUnless { it.isEmpty() }?.let { pkg.id to it }
-        }.toMap()
+    override fun getCurationsFor(packages: Collection<Package>): List<PackageCuration> =
+        packages.flatMap { pkg -> getCurationsFor(pkg.id) }.distinct()
```
sschuberth: I believe `distinct()` should not be added here to more closely resemble the previous code, and because it would be dropped in the next commit anyway.

sschuberth (approved):

PR #6468, opened 2023-02-09
ClearlyDefined type mapping fixes

sschuberth:

### DIFF
--- a/model/src/main/kotlin/utils/Extensions.kt
+++ b/model/src/main/kotlin/utils/Extensions.kt
@@ -51,8 +51,8 @@ fun Package.toClearlyDefinedTypeAndProvider(): Pair<ComponentType, Provider>? =
         "Maven" -> ComponentType.MAVEN to Provider.MAVEN_CENTRAL
         "NPM" -> ComponentType.NPM to Provider.NPM_JS
         "NuGet" -> ComponentType.NUGET to Provider.NUGET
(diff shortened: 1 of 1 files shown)


model/src/main/kotlin/utils/Extensions.kt
```diff
@@ -48,7 +48,7 @@ fun Package.toClearlyDefinedTypeAndProvider(): Pair<ComponentType, Provider>? =
         "Crate" -> ComponentType.CRATE to Provider.CRATES_IO
         "DotNet", "NuGet" -> ComponentType.NUGET to Provider.NUGET
         "Gem" -> ComponentType.GEM to Provider.RUBYGEMS
-        "GoDep", "GoMod" -> ComponentType.GO to Provider.GOLANG
+        "Go" -> ComponentType.GO to Provider.GOLANG
```
fviernau, line 50: I believe e30fab04d353d7fd8163aded60c8d1dad00dbdbc is not the commit you intended to reference, but another one which is IIRC very close to it timely.

fviernau (changes requested):

model/src/main/kotlin/utils/Extensions.kt
sschuberth: Thanks, fixed by referencing the right commits (it's actually two).

fviernau (approved):

sschuberth:

Issue #155, opened 2017-12-15
Add `provider` to Package


jeffmcaffer:
Packages have a `packageManager`  (e.g, npm, maven, ...). Since a given type of package could come from many different places, the `Package` should also talk about a `provider`. The `provider` should not be the URL of the repository, rather the notional name of the repository (e.g., npmjs.org, github.com). This allows the repos to move and change their URL structure without affecting the identity of the data stored in ORT.

sschuberth:
@mnonnenmacher I believe what's called "provider" here is what we refer to as "provenance" in our idea.

jeffmcaffer:
`provenance` is a pretty loaded term that carries with it very deep meaning for some. Seems like here we need the very simple implication that "this is which package foo we are talking about"

mnonnenmacher:
I think this information should be added to the RemoteArtifact and VcsInfo models, because these are the two places where we reference URLs (apart from the homepage URL).
For RemoteArtifact this could be something like "npmjs.org", "JCenter", or "Maven Central". For VcsInfo "github.com", "bitbucket.com", and so on.
My problem is how we should auto-detect the values for those fields, e.g. if we take part of the URL like "github.com" this contradicts the idea of having something URL independent. Maybe we would have to maintain a mapping from URL to provider name?

jeffmcaffer:
@mnonnenmacher agreed, having a table that bi-directionally maps provider names to host names makes sense. It likely also makes sense to keep the provider names as generic as possible. For example, we recently ran into some identity problems because some folks were using "npmjs.org" vs "npmjs.com" as the provider. it turns out they are the same and going to npmjs.org forwards to npmjs.com.

To isolate the data from these sorts of variations and changes, using just "npmjs" would be more resilient.  That's also inline with your other examples like "maven central" etc.  For simplicity perhaps we say that provider names need to be valid url segments that do not require any quoting. (e.g., no spaces, no funny chars, ...) and have them be case insensitive and NOT case preserving. (or just spec lowercase).

sschuberth:
This somewhat relates to https://github.com/heremaps/oss-review-toolkit/issues/20.

PR #6472, opened 2023-02-10
Stop hard-coding ClearlyDefined providers

sschuberth:

### DIFF
--- a/model/src/main/kotlin/PackageProvider.kt
+++ b/model/src/main/kotlin/PackageProvider.kt
@@ -1,0 +1,132 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model
+
+/**
+ * The provider / hosting platform of a remote artifact or Version Control System. These are basically human-readable
+ * aliases for different URLs that may point to the same provider. Note that different providers might host different
+ * artifacts under the same [Identifier]. So if in question, the provider must be given in addition to the [Identifier].
+ */
+enum class PackageProvider(
+    /**
+     * The regular expression patterns that match all URLs for a specific provider.
+     */
+    vararg urlPatterns: String,
+) {
+    COCOAPODS(
+        "^https?://cocoapods\\.org/pods/"
+    ),
+    CRATES_IO(
+        "^https?://crates\\.io/api/"
+    ),
+    DEBIAN(
+        "^https?://(\\w+\\.)+debian.org/debian/"
+    ),
+    GITHUB(
+        "^(https?|ssh)://(git@)?(\\w+\\.)?github\\.com/"
+    ),
+    GITLAB(
+        "^(https?|ssh)://(git@)?(\\w+\\.)?gitlab\\.com/"
+    ),
+    GOLANG(
+        "^https?://go\\.googlesource\\.com/",
+        "^https?://proxy\\.golang\\.org/"
+    ),
+    GRADLE_PLUGIN(
+        "^https?://plugins\\.gradle\\.org/m2/"
+    ),
+    MAVEN_CENTRAL(
+        "^https?://repo\\.maven\\.apache\\.org/maven2/",
+        "^https?://repo1\\.maven\\.org/maven2/"
+    ),
+    MAVEN_GOOGLE(
+        "^https?://maven\\.google\\.com/"
+    ),
+    NPM_JS(
+        "^https?://registry\\.npmjs\\.(com|org)/"
+    ),
+    NUGET(
+        "^https?://api\\.nuget\\.org/.+\\.nupkg"
+    ),
+    PACKAGIST(
+        "^https?://repo\\.packagist\\.org/"
+    ),
+    PYPI(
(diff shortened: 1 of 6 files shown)


model/src/main/kotlin/PackageProvider.kt
```diff
@@ -0,0 +1,147 @@
+/*
```
fviernau, line 1: commit-msg: "This is will" -> "This will"

model/src/main/kotlin/PackageProvider.kt
```diff
@@ -0,0 +1,147 @@
...
+ * artifacts under the same [Identifier]. So if in question, the provider must be given in addition to the [Identifier].
+ */
+enum class PackageProvider(
+    /**
+     * The regular expression patterns that match all URLs for a specific provider.
+     */
+    vararg urlPatterns: String,
+
+    /**
+     * The name ClearlyDefined uses for this provider. If a provider if not available in ClearlyDefined (yet), use an
```
fviernau, line 34: "If a provider if" -> "If a provider i**s**"

model/src/main/kotlin/PackageProvider.kt
```diff
@@ -0,0 +1,147 @@
...
+        clearlyDefinedName = "rubygems"
+    ),
+
+    // The following have no matching ClearlyDefined provider.
+    ANDROID(
+        "^https?://android\\.googlesource\\.com/"
+    ),
+    APACHE_GIT(
+        "^https?://gitbox\\.apache\\.org/repos/.+\\.git",
+        "^https?://git-wip-us\\.apache\\.org/repos/.+\\.git"
```
fviernau, line 104: I fail to understand the point of `.+`. Could you help?

model/src/main/kotlin/PackageProvider.kt
```diff
@@ -0,0 +1,147 @@
...
+    UNKNOWN(
+        ".*"
+    );
+
+    private val urlRegexes by lazy { urlPatterns.map { it.toRegex() } }
+
+    companion object {
+        fun get(url: String) = enumValues<PackageProvider>().find { provider ->
+            provider.urlRegexes.any { it.containsMatchIn(url) }
+        } ?: UNKNOWN
```
fviernau, line 141: UNKOWN is guaranteed to match, so `?: UNKOWN` should never happen. Would it make sense to express this with `!!` or rather `checkNotNull()` ?

model/src/main/kotlin/PackageProvider.kt
```diff
@@ -0,0 +1,147 @@
...
+
+    // The following matches any (unknown) provider.
+    UNKNOWN(
+        ".*"
+    );
+
+    private val urlRegexes by lazy { urlPatterns.map { it.toRegex() } }
+
+    companion object {
+        fun get(url: String) = enumValues<PackageProvider>().find { provider ->
```
fviernau, line 139: I'm not 100% certain if the patterns match. (E.g. wondering if patterns should end with `.*`. Could you add a unit test to cover at least one or two cases, maybe picking one with one of the more complicated patterns?

model/src/main/kotlin/PackageProvider.kt
```diff
@@ -0,0 +1,147 @@
...
+    ),
+
+    // The following matches any (unknown) provider.
+    UNKNOWN(
+        ".*"
+    );
+
+    private val urlRegexes by lazy { urlPatterns.map { it.toRegex() } }
+
+    companion object {
```
fviernau, line 138: optional: Imo explicit return types would be nice to have nullability obvious.

model/src/main/kotlin/utils/Extensions.kt
```diff
@@ -27,48 +27,67 @@ import org.ossreviewtoolkit.clients.clearlydefined.Provider
...
 import org.ossreviewtoolkit.utils.common.percentEncode
 
 internal fun TextLocation.prependPath(prefix: String): String =
     if (prefix.isBlank()) path else "${prefix.removeSuffix("/")}/$path"
 
 /**
- * Map an [Package] to a ClearlyDefined [ComponentType] and [Provider]. Note that an
- * [identifier's type][Identifier.type] in ORT currently implies a default provider. Return null if a mapping is not
- * possible.
+ * Map a [Package] [identifier's type][Identifier.type] to a ClearlyDefined [ComponentType]. Return null if a mapping is
```
fviernau, line 41: maybe instead: "Map [type][Identifier.type] of a [package identifier] [Package.id] to [..]"

model/src/main/kotlin/PackageProvider.kt
```diff
@@ -0,0 +1,147 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model
```
fviernau, line 20: Is this supposed to be used only from `ClearlyDefinedPackageCurationProvider`? If so, shouldn't it be put next to it (in the analyzer) ?

fviernau (changes requested):

model/src/main/kotlin/PackageProvider.kt
sschuberth: I want to ensure that the URLs end with ".git", and that something comes between that and the base URL / base path.

model/src/main/kotlin/PackageProvider.kt
sschuberth: FYI, I created a temporary unit test when crafting these patterns, and that test went over all remote artifact URLs or VCS URLs in all of our committed ORT results files. So I'm pretty sure the patterns cover all relevant cases. But the test simply was not in shape to be committed.

model/src/main/kotlin/PackageProvider.kt
sschuberth: Strictly speaking #155 is about adding the "(package) provider" concept to ORT independently of ClearlyDefined. So while I agree the primary use-case to be in context with ClearlyDefined, I do not want to rule out other use-cases at this point by moving the class.

sschuberth: But you're right that I should separate the ClearlyDefined-specific code more from the enum. Will do.

fviernau (dismissed):

sschuberth:
I had to rebase to resolve a trivial merge conflict in imports.

fviernau (approved):

sschuberth:

PR #6474, opened 2023-02-10
analyzer: Move `CurationProvider` to the `model` module

fviernau:
Simplify adding provider IDs to `OrtResult` in a following PR. 
Note that the provider now lives next to `PackageConfigurationProvider`.

See individual commits.


sschuberth (commented):
LGTM, but I'd like @mnonnenmacher to confirm that this does not conflict with his work to move plugins to their own Gradle module.

mnonnenmacher (approved):

PR #6483, opened 2023-02-14
Fix the `OsvServiceFunTest`

sschuberth:

### DIFF
--- a/clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
+++ b/clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
@@ -59,7 +59,7 @@ class OsvServiceFunTest : StringSpec({
 
         result.shouldBeSuccess { actualData ->
             val expectedData = OsvApiClient.JSON.decodeFromString<List<Vulnerability>>(expectedResult)
-            actualData.patchModifiedTimes() shouldBe expectedData.patchModifiedTimes()
+            actualData.patchModifiedTimes() shouldContainExactlyInAnyOrder expectedData.patchModifiedTimes()
         }
     }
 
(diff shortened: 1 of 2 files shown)


clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
```diff
@@ -90,6 +90,7 @@ class OsvServiceFunTest : StringSpec({
                 listOf(
                     "GHSA-462w-v97r-4m45",
                     "GHSA-8r7q-cvjq-x353",
+                    "GHSA-fqh9-2qgg-h84h",
```
fviernau, line 93: To my understanding, the minimal diff to fix the test would only update the lines 318-417, and line 93. All other updates IIUC are not necessary.

E.g. the modified date is not compared, and the test should be agnostic of ordering of properties.

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
```diff
@@ -59,7 +59,7 @@ class OsvServiceFunTest : StringSpec({
 
         result.shouldBeSuccess { actualData ->
             val expectedData = OsvApiClient.JSON.decodeFromString<List<Vulnerability>>(expectedResult)
-            actualData.patchModifiedTimes() shouldBe expectedData.patchModifiedTimes()
+            actualData.patchModifiedTimes() shouldContainExactlyInAnyOrder expectedData.patchModifiedTimes()
```
fviernau, line 62: Just a question for my understanding: This is not related to current test failures, is it?

fviernau (approved):

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
sschuberth: Not this time (as I'm also changing orders in the new expected results), but I've seen failure before where just having this would have fixed it.

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
sschuberth: > All other updates IIUC are not necessary.

Might be, but as it's actually quite inconvenient to find out what exactly fails for this test, I've simply serialized the current actual result, pretty-printed it, used it as the new expected result, and then inspected the Git diff afterwards for plausibility.

Issue #5560, opened 2022-07-14
GoMod: go list -deps error: no matching files found


mawl:
Hey,

We have an issue with the GoMod Package Manager

```
10:35:45.499 [DefaultDispatcher-worker-1] ERROR org.ossreviewtoolkit.analyzer.managers.GoMod - Resolving GoMod dependencies for path 'go.mod' failed with: 
IOException: Running 'go list -deps -f {{with .Module}}{{.Path}} {{.Version}}{{end}} ./...' in '/builds/some-go-app' failed with exit code 1:
XXXXXX.go:10:12: pattern some.file: no matching files found
```

Content of XXXXXX.go
```go
//go:generate cp ../../../some.file .
//go:embed some.file
```
some.file ist generated during build of another component in the project.

We don't want to run "go generate/go build" as a prequisite for ort. So, are there other ways to get this working?

Cheers,
Marco

fviernau:
@mawl would you be able to share a minimal example to reproduce this issue?

fviernau:
Slightly related: https://github.com/golang/go/issues/49534

mawl:
As a workaround I use following script before running oss review toolkit, which removes the go:embed directive from relevant files:

`grep -rl --include \*.go --exclude-dir=vendor "go:embed" . | xargs sed -i 's/go\:embed//g'`

sschuberth:
@mawl, can you still reproduce the issue with the latest version of ORT?

fviernau:
> @mawl, can you still reproduce the issue with the latest version of ORT?

This is not supposed to be fixed yet, but it's in the makings. In theory it should be fixed once the next Go version is released and we upgrade to it. See

- https://github.com/golang/tools/pull/428
- https://github.com/golang/go/pull/58522

sschuberth:
Note to myself: The upstream Golang issues have been closed as Gerrit changes with fixes were submitted. These changes did not make it into the 1.20.0 Golang release, but it's a bit hard to figure out whether they have been [cherry-picked to any 1.20.x minor release](https://go.dev/doc/devel/release#go1.20.minor). We should probably just try with a Golang upgrade.

fviernau:
[...]

PR #6485, opened 2023-02-14
GoMod: Upgrade and migrate from `go list -f` to `go list -json`

fviernau:
The changes prepare for fixing #5560, see individual commits.


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -288,13 +288,26 @@ class GoMod(
...
+            JsonFactory().createParser(inputStream).apply {
+                codec = ObjectMapper()
+                nextToken()
+
+                while (hasCurrentToken()) {
+                    val moduleInfo = jsonMapper.readValue(this, DepInfo::class.java).module
+                    if (moduleInfo != null) result += moduleInfo.path
+                    nextToken()
+                }
+            }
```
sschuberth, line 310: This can be simplified to

    jsonMapper.createParser(list.stdout).use { parser ->
        val moduleIterator = jsonMapper.readValues<DepInfo>(parser)
        moduleIterator.asSequence().mapNotNullTo(result) {
            it.module?.path
        }
    }

BTW, there's similar existing code starting at line 232 that could be simplified in a similar way. If you'd don't feel like simplifying that code in a new commit in this PR, too, I can do it in a follow-up PR.

sschuberth (changes requested):

sschuberth:
FYI, looks like commit body line

    see [^1],[^2] and [^3]. Such a mechanism is not in place for the `-f` option.

is slightly too long.

sschuberth (changes requested):

sschuberth (approved):

fviernau:
Merged regardless of the unrelated failing PipFunTest.

PR #6489, opened 2023-02-15
Fix issues with querying Maven artifacts from different repositories

sschuberth:


analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -644,8 +652,11 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
      * below a "target" subdirectory of the actual project.
      */
     fun parsePackage(
-        artifact: Artifact, repositories: List<RemoteRepository>,
-        localProjects: Map<String, MavenProject> = emptyMap(), sbtMode: Boolean = false
+        artifact: Artifact,
+        repositories: List<RemoteRepository>,
+        useArtifactRepositories: Boolean,
```
sschuberth, line 657: Should we set this to `useArtifactRepositories = true` be default to that together with the default of `sbtMode = false` this describes the default Maven behavior?

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -489,7 +489,11 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
         }
     }
 
-    private fun requestRemoteArtifact(artifact: Artifact, repositories: List<RemoteRepository>): RemoteArtifact {
+    private fun requestRemoteArtifact(
+        artifact: Artifact,
+        repositories: List<RemoteRepository>,
+        useArtifactRepositories: Boolean
```
sschuberth, line 495: I'm open to better names here. Something like `useReposFromDependencyPoms` sounds a bit clumsy, though.

sschuberth: @mnonnenmacher any suggestion here?

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: @oss-review-toolkit/kotlin-devs any opinion here?

mnonnenmacher: I have no strong opinion here, I'm fine with and without default value. But it would be very good to explain the argument in the function docs.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
mnonnenmacher: I actually like the suggestion, maybe a bit clumsy but also precise. Alternatives: `useReposFromDependency`, `useReposDeclaredByDependency`.
I assume you use plural "poms" because repositories from all POMs on the path to the dependency are used?

sschuberth: > I assume you use plural "poms" because repositories from all POMs on the path to the dependency are used?

Exactly. So I'll probably go for `useReposFromDependencies` (plural).

mnonnenmacher (dismissed):
Approved, but it looks like there are more expected test results you need to update.

sschuberth:
> Approved, but it looks like there are more expected test results you need to update.

The failing test pass for me locally (also see [my message on Slack](https://ort-talk.slack.com/archives/CD1M764AJ/p1676551971669669)). Would you mind checking whether `GradleKotlinScriptFunTest` and `SbtFunTest` pass locally for you, too?

sschuberth:
On CI, I for example see

    https://repo.maven.apache.org/maven2/org/jetbrains/annotations/13.0/annotations-13.0.jar

being used instead of

    https://jcenter.bintray.com/org/jetbrains/annotations/13.0/annotations-13.0.jar

mnonnenmacher:
> The failing test pass for me locally (also see [my message on Slack](https://ort-talk.slack.com/archives/CD1M764AJ/p1676551971669669)). Would you mind checking whether `GradleKotlinScriptFunTest` and `SbtFunTest` pass locally for you, too?

Yes, tests pass locally.

sschuberth:
> Yes, tests pass locally.

As I'm totally clueless by now, I've filed https://github.com/actions/setup-java/issues/454.

mnonnenmacher:
> As I'm totally clueless by now, I've filed [actions/setup-java#454](https://github.com/actions/setup-java/issues/454).

Did you already try to clear the settings.xml after the setup-java task?

sschuberth:
> Did you already try to clear the settings.xml after the setup-java task?

Not *after* `setup-java`, but [I instructed `setup-java` to not overwrite it](https://github.com/oss-review-toolkit/ort/pull/6489/commits/30a6dd0d24a060c61bf970eb696e1b7830037c7d) (after I created an empty one).

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -529,7 +529,7 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
             val localRepositories = allRepositories - remoteRepositories
             if (localRepositories.isNotEmpty()) {
                 // No need to use curly-braces-syntax for logging here as the log level check is already done above.
-                logger.debug { "Ignoring local repositories $localRepositories." }
+                logger.debug("Ignoring local repositories $localRepositories.")
```
fviernau, line 532: What's the advantage of not using curly braces?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: Avoiding a superfluous check if the log level is set to debug.

fviernau: I believe such micro optimizations should not be valued that high, and consistency preferred in this case. Which would also allow removing that comment in line 531.

sschuberth: > Which check actually is superflous?

The `logger.delegate.isDebugEnabled` condition duplicates the check done in the implementation if `debug {}`.

> I believe such micro optimizations should not be valued that high

I'm not valuing it high, I'm just changing the code to match the documentation.

> consistency preferred in this case.

There are also other code locations where is "micro-optimization" is performed, and I prefer to stay consistent to these.

fviernau: > I'm not valuing it high

I believe my point was clear, e.g. I believe it's inappropriate to do such kind of optimizations in the given context.

> There are also other code locations where is "micro-optimization" is performed, and I prefer to stay consistent to these.

Would you mind sharing why?

I'll share my reason as well: keep it simple.

analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
```diff
@@ -94,17 +94,21 @@ class MavenDependencyHandler(
...
+
+            // There is the corner case that a dependency references a project, but in a different version than
+            // the one used by the local build. Then, this dependency is actually a package, but Maven's
+            // resolution mechanism might prefer using the project. Therefore, the check whether the dependency
+            // is a project must be done after the package resolution again.
+            if (isLocalProject(pkg.id)) {
+                localProjectIds += dependency.identifier()
+                null
+            } else {
+                pkg
```
mnonnenmacher, line 111: Unrelated refactoring.

analyzer/src/test/kotlin/managers/utils/MavenDependencyHandlerTest.kt
```diff
@@ -158,7 +158,9 @@ class MavenDependencyHandlerTest : WordSpec({
             val handler = createHandler(sbtMode = true)
 
             every { dependency.repositories } returns repos
-            every { handler.support.parsePackage(artifact, repos, sbtMode = true) } returns pkg
+            every {
+                handler.support.parsePackage(artifact, repos, sbtMode = true)
+            } returns pkg
```
mnonnenmacher, line 163: Unrelated refactoring (I'd actually keep the one-liner).

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -513,6 +508,13 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
             repositories
         }
 
+        val cacheKey = "$artifact@$allRepositories"
```
mnonnenmacher, line 511: Commit message: "This, " -> "Therefore, "?

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
mnonnenmacher, line 532: I'd remove the code comment instead because the performance impact is so insignificant. But I'm fine with doing this as well.

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
mnonnenmacher: Since we know that `remoteRepositories` is a subset of `allRepositories` we could do this instead:
```kotlin
if (allRepositories.size > remoteRepositories.size) {
  logger.debug {
    val localRepositories = allRepositories - remoteRepositories
    "Ignoring local repositories $localRepositories."
  }
}
```

mnonnenmacher (changes requested):

analyzer/src/main/kotlin/managers/utils/MavenDependencyHandler.kt
sschuberth: Moved to a preceding commit (there also was a left-over unneeded wrapping involved).

analyzer/src/test/kotlin/managers/utils/MavenDependencyHandlerTest.kt
sschuberth: Undone.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: It should have been "Thus", fixed.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: > I'd remove the code comment instead because the performance impact is so insignificant.

Looks like I'm outvoted here, so I'll change as proposed.

mnonnenmacher (approved):

PR #6490, opened 2023-02-15
Make use of `PackageProvider`  for Maven

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -312,8 +312,8 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
             val mavenCentralUrl = with(artifact) {
                 val group = groupId.replace('.', '/')
                 val name = remoteArtifact.url.substringAfterLast('/')
-                val hash = remoteArtifact.hash.algorithm.name.lowercase()
-                "https://repo.maven.apache.org/maven2/$group/$artifactId/$version/$name.$hash"
+                val algorithm = remoteArtifact.hash.algorithm.name.lowercase()
+                "https://repo.maven.apache.org/maven2/$group/$artifactId/$version/$name.$algorithm"
             }
 
             val checksum = OkHttpClientHelper.downloadText(mavenCentralUrl).getOrElse { return false }
--- a/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
+++ b/analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
@@ -306,19 +306,21 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
          * Return true if an artifact that has not been requested from Maven Central is also available on Maven Central
          * but with a different hash, otherwise return false.
          */
-        private fun isArtifactModified(artifact: Artifact, remoteArtifact: RemoteArtifact): Boolean {
-            if (remoteArtifact.url.isBlank() || isMavenCentralUrl(remoteArtifact.url)) return false
+        private fun isArtifactModified(artifact: Artifact, remoteArtifact: RemoteArtifact): Boolean =
+            with(remoteArtifact) {
+                if (url.isBlank() || isMavenCentralUrl(url)) return false
 
-            val mavenCentralUrl = with(artifact) {
-                val group = groupId.replace('.', '/')
-                val name = remoteArtifact.url.substringAfterLast('/')
-                val algorithm = remoteArtifact.hash.algorithm.name.lowercase()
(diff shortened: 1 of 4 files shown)


analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -306,19 +306,21 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
...
+        private fun isArtifactModified(artifact: Artifact, remoteArtifact: RemoteArtifact): Boolean =
+            with(remoteArtifact) {
+                if (url.isBlank() || isMavenCentralUrl(url)) return false
 
-            val checksum = OkHttpClientHelper.downloadText(mavenCentralUrl).getOrElse { return false }
-            return !remoteArtifact.hash.verify(parseChecksum(checksum, remoteArtifact.hash.algorithm.name))
-        }
+                val name = url.substringAfterLast('/')
+                val algorithm = hash.algorithm.name.lowercase()
+
```
fviernau, line 315: Having two `with` nested it gets a bit harder to see from which object the attributes come from.

Would it make sense to solve this by factoring out `fun Artifact.getMavenCentralUrl()?`

fviernau (changes requested):

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -306,19 +306,21 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
...
+                val mavenCentralUrl = with(artifact) {
+                    val group = groupId.replace('.', '/')
+                    "https://repo.maven.apache.org/maven2/$group/$artifactId/$version/$name.$algorithm"
```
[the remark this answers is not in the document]
sschuberth: Yes, the variable was renamed to avoid a name clash due to the nested use of `with`.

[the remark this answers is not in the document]
fviernau: I see, then my other comment does not make sense, because the URL value depends on the attributes of both, `artifact` and `remoteArtifact`.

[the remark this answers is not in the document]
sschuberth: But as the renaming was sensible anyway, I've extracted it now to a preceding commit.

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
sschuberth: > Would it make sense to solve this by factoring out `fun Artifact.getMavenCentralUrl()`?

That's not nicely possible, as the function would also need to have access to the `RemoteArtifact`, and I wouldn't want to pass that as an argument.

Given that the inner `with` is only two lines, and with the IDE you get proper indication about the scope of the properties, can we live with the nested `with` in this case?

utils/common/src/test/kotlin/UtilsTest.kt
```diff
@@ -65,28 +64,4 @@ class UtilsTest : WordSpec({
             )
         }
     }
-
-    "isMavenCentralUrl" should {
```
fviernau, line 69: Would be nice to move these into `PackageProviderTest` class.

fviernau (changes requested):

analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
fviernau: Yes, my comment didn't make sense as I tried to explain below.

utils/common/src/test/kotlin/UtilsTest.kt
sschuberth: > Would be nice to move these into `PackageProviderTest` class.

There's no such test class yet (remember our [discussion of the temporary test that I added during development](https://github.com/oss-review-toolkit/ort/pull/6472#discussion_r1103580937)), and as such a test for `PackageProvider` would need to cover way more URLs, I'd prefer to address that in a separate PR.

fviernau: I do remember that one, I brought it up again because now it's a different context, which is dropping test coverage.

I don't see what the problem would be with creating a new class and moving the function there, even if the class does not cover all cases but only the ones from the other test. I'll leave it optional anyway.

fviernau (approved):

utils/common/src/test/kotlin/UtilsTest.kt
sschuberth: > I don't see what the problem would be with creating a new class

Just time constraints. But I've now added a start at https://github.com/oss-review-toolkit/ort/pull/6499.

PR #6497, opened 2023-02-16
Make more use of file name constants

sschuberth:

### DIFF
--- a/cli/src/funTest/kotlin/ExamplesFunTest.kt
+++ b/cli/src/funTest/kotlin/ExamplesFunTest.kt
@@ -60,6 +60,7 @@ import org.ossreviewtoolkit.reporter.ReporterInput
 import org.ossreviewtoolkit.reporter.reporters.freemarker.asciidoc.PdfTemplateReporter
 import org.ossreviewtoolkit.utils.ort.ORT_PACKAGE_CURATIONS_FILENAME
 import org.ossreviewtoolkit.utils.ort.ORT_REPO_CONFIG_FILENAME
+import org.ossreviewtoolkit.utils.ort.ORT_RESOLUTIONS_FILENAME
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 import org.ossreviewtoolkit.utils.test.createSpecTempDir
 import org.ossreviewtoolkit.utils.test.shouldNotBeNull
@@ -118,7 +119,7 @@ class ExamplesFunTest : StringSpec() {
 
         "The resolutions file can be deserialized" {
             shouldNotThrow<IOException> {
-                takeExampleFile("resolutions.yml").readValue<Resolutions>()
+                takeExampleFile(ORT_RESOLUTIONS_FILENAME).readValue<Resolutions>()
             }
         }
 
--- a/cli/src/funTest/kotlin/ExamplesFunTest.kt
+++ b/cli/src/funTest/kotlin/ExamplesFunTest.kt
@@ -58,6 +58,7 @@ import org.ossreviewtoolkit.notifier.Notifier
 import org.ossreviewtoolkit.reporter.HowToFixTextProvider
 import org.ossreviewtoolkit.reporter.ReporterInput
 import org.ossreviewtoolkit.reporter.reporters.freemarker.asciidoc.PdfTemplateReporter
+import org.ossreviewtoolkit.utils.ort.ORT_PACKAGE_CURATIONS_FILENAME
 import org.ossreviewtoolkit.utils.ort.ORT_REPO_CONFIG_FILENAME
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 import org.ossreviewtoolkit.utils.test.createSpecTempDir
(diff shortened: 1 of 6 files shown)


model/src/test/kotlin/JsonSchemaTest.kt
```diff
@@ -79,9 +81,9 @@ class JsonSchemaTest : StringSpec({
         errors should beEmpty()
     }
 
-    "reference.yml validates successfully" {
+    "$REFERENCE_CONFIG_FILENAME validates successfully" {
```
fviernau, line 88: Whether to use these constants in test names is probably quite a bit subjective.
I'd tend to not do that, because we want consistency which in turn means that also constants with longer names need to be used. So, the amount of characters available for providing the meat of the test case name gets reduced, which makes it harder to describe the test properly.

My proposal would be to even abstract from the actual file name in the test case name, like:
"the reference config validates successfully"

As this is opinionated I'd like to see what others think, e.g.  @oss-review-toolkit/core-devs ?

fviernau (changes requested):

model/src/test/kotlin/JsonSchemaTest.kt
```diff
@@ -30,6 +30,7 @@ import io.kotest.matchers.should
 import java.io.File
 
 import org.ossreviewtoolkit.model.config.REFERENCE_CONFIG_FILENAME
+import org.ossreviewtoolkit.utils.ort.ORT_PACKAGE_CONFIGURATION_FILENAME
```
fviernau, line 33: Should this constant use plural form, like the cli option `--package-curations-file` ?

fviernau (changes requested):

model/src/test/kotlin/JsonSchemaTest.kt
sschuberth: > My proposal would be to even abstract from the actual file name in the test case name, like:
> "the reference config validates successfully"

I like this, will do.

model/src/test/kotlin/JsonSchemaTest.kt
fviernau: You're right, my bad.

cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -89,19 +89,19 @@ class ExamplesFunTest : StringSpec() {
             }
         }
 
-        "copyright-garbage.yml can be deserialized" {
+        "The Copyright garbage configuration file can be deserialized" {
```
mnonnenmacher, line 94: You could scratch "configuration" from several test names, it's not used consistently anyway and it doesn't add value. So just "The copyright garbage file..." is fine IMO.

fviernau (approved):

model/src/test/kotlin/JsonSchemaTest.kt
mnonnenmacher: While the example file might only contain a single package configuration the `EvaluatorCommand` option `--package-configuration-file` actually reads a list of package configurations from this file (and should therefore be renamed to `--package-configurations-file`). To me it doesn't make sense to introduce a global constant that relates only to this example file. If we introduce the constant it should rather reflect the CLI use case and use plural for the constant name and file name.
Also, note that the other constant `ORT_PACKAGE_CONFIGURATIONS_DIRNAME` also uses plural. And for consistency with package curations `--package-configuration-dir` should also be renamed to `--package-configurations-dir`.

mnonnenmacher (changes requested):

sschuberth:
>  While the example file might only contain a single package configuration the `EvaluatorCommand` option `--package-configuration-file` actually reads a list of package configurations from this file (and should therefore be renamed to `--package-configurations-file`). To me it doesn't make sense to introduce a global constant that relates only to this example file. If we introduce the constant it should rather reflect the CLI use case and use plural for the constant name and file name.
Also, note that the other constant `ORT_PACKAGE_CONFIGURATIONS_DIRNAME` also uses plural. And for consistency with package curations `--package-configuration-dir` should also be renamed to `--package-configurations-dir`.

Phew, looks like we have a tons of inconsistencies there. Also, I just realized we never look by default for a `~/.ort/config/package-configuration(s).yml` *file* like we do look for `~/.ort/config/curations.yml`.

Anyway, that's going beyond the scope of the PR which had just the goal to not hard-code the configuration file names. And I'd also prefer to keep the introduced `ORT_PACKAGE_CONFIGURATION_FILENAME` constant even if we'd rename it to `ORT_PACKAGE_CONFIGURATIONS_FILENAME` and make it point to "package-configuration*s*.yml" later, as it's IMO a step into the right direction to *have* this constant for consistency.

mnonnenmacher:
> Phew, looks like we have a tons of inconsistencies there. Also, I just realized we never look by default for a `~/.ort/config/package-configuration(s).yml` _file_ like we do look for `~/.ort/config/curations.yml`.
> 
> Anyway, that's going beyond the scope of the PR which had just the goal to not hard-code the configuration file names. And I'd also prefer to keep the introduced `ORT_PACKAGE_CONFIGURATION_FILENAME` constant even if we'd rename it to `ORT_PACKAGE_CONFIGURATIONS_FILENAME` and make it point to "package-configuration_s_.yml" later, as it's IMO a step into the right direction to _have_ this constant for consistency.

I have created #6501 to track this.

mnonnenmacher (approved):

PR #6511, opened 2023-02-17
Gracefully handle package curations when replacing repository configurations or curations

fviernau:
This PR finally implements support for package curations in repository configurations, for the use case of re-running the evaluator after making local changes to both, local and global package curations. 

See individual commits.

### DIFF
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -355,13 +355,17 @@ data class OrtResult(
      * associated with the given [providerId].
      */
     fun replacePackageCurations(provider: PackageCurationProvider, providerId: String): OrtResult {
-        val packageCurations = ConfigurationResolver.resolvePackageCurations(
+        require(providerId != REPOSITORY_CONFIGURATION_PROVIDER_ID) {
+            "Cannot replace curations for id '$REPOSITORY_CONFIGURATION_PROVIDER_ID' which is reserved and not allowed."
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/OrtResult.kt
```diff
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.annotation.JsonInclude
 
 import org.apache.logging.log4j.kotlin.Logging
 
+import org.ossreviewtoolkit.model.ResolvedPackageCurations.Companion.REPOSITORY_CONFIGURATION_PROVIDER_ID
```
sschuberth, line 27: Commit message: s/flat/flag/

model/src/main/kotlin/OrtResult.kt
```diff
@@ -331,7 +332,21 @@ data class OrtResult(
     /**
      * Return a copy of this [OrtResult] with the [Repository.config] replaced by [config].
      */
-    fun replaceConfig(config: RepositoryConfiguration): OrtResult = copy(repository = repository.copy(config = config))
+    fun replaceConfig(config: RepositoryConfiguration): OrtResult =
```
mnonnenmacher, line 335: Please also update the documentation of the function to explain how curations contained in `config` are handled.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #6520, opened 2023-02-20
Stash to a temporary directory instead of a parent directory by default

sschuberth:

### DIFF
--- a/utils/common/src/main/kotlin/DirectoryStash.kt
+++ b/utils/common/src/main/kotlin/DirectoryStash.kt
@@ -48,7 +48,8 @@ private class DirectoryStash(directories: Set<File>) : Closeable {
         // We need to check this on each iteration instead of filtering beforehand to properly handle parent / child
         // directories.
         if (originalDir.isDirectory) {
-            // Create a temporary directory to move directories as-is into.
+            // Create a temporary directory to move the original directory into as a sibling of the original directory
+            // to ensure it resides on the same file system for being able to perform an atomic move.
             val tempDir = createTempDirectory(originalDir.parentFile.toPath(), "stash").toFile()
 
             // Use a non-existing directory as the target to ensure the directory can be moved atomically.
--- a/utils/common/src/main/kotlin/DirectoryStash.kt
+++ b/utils/common/src/main/kotlin/DirectoryStash.kt
@@ -37,7 +37,7 @@ import org.apache.logging.log4j.kotlin.Logging
 fun stashDirectories(vararg directories: File): Closeable = DirectoryStash(setOf(*directories))
 
 /**
- * A Closable class which temporarily moves away directories and moves them back on close. Any conflicting directory
+ * A [Closeable] class which temporarily moves away directories and moves them back on close. Any conflicting directory
  * created at the location of an original directory is deleted before the original state is restored. If a specified
  * directory did not exist on initialization, it will also not exist on close.
  */
--- a/utils/common/src/main/kotlin/DirectoryStash.kt
+++ b/utils/common/src/main/kotlin/DirectoryStash.kt
@@ -49,18 +49,18 @@ private class DirectoryStash(directories: Set<File>) : Closeable {
         // directories.
         if (originalDir.isDirectory) {
             // Create a temporary directory to move directories as-is into.
-            val stashDir = createTempDirectory(originalDir.parentFile.toPath(), "stash").toFile()
+            val tempDir = createTempDirectory(originalDir.parentFile.toPath(), "stash").toFile()
 
             // Use a non-existing directory as the target to ensure the directory can be moved atomically.
-            val tempDir = stashDir.resolve(originalDir.name)
+            val stashDir = tempDir.resolve(originalDir.name)
 
             logger.info {
-                "Temporarily moving directory from '${originalDir.absolutePath}' to '${tempDir.absolutePath}'."
+                "Temporarily moving directory from '${originalDir.absolutePath}' to '${stashDir.absolutePath}'."
             }
 
-            originalDir.toPath().moveTo(tempDir.toPath(), StandardCopyOption.ATOMIC_MOVE)
+            originalDir.toPath().moveTo(stashDir.toPath(), StandardCopyOption.ATOMIC_MOVE)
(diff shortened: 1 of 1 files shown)


utils/common/src/main/kotlin/DirectoryStash.kt
```diff
@@ -32,24 +32,27 @@ import kotlin.io.path.moveTo
 import org.apache.logging.log4j.kotlin.Logging
 
 /**
- * A convenience function that stashes directories using a [DirectoryStash] instance.
+ * A convenience function that stashes [directories] to the [stashParentDir] using a [DirectoryStash] instance.
```
fviernau: commit-msg: Please explain shortly the problem this solves and why it solved it.

(Would it make sense to split-up this commit into two?)

utils/common/src/main/kotlin/DirectoryStash.kt
```diff
@@ -32,24 +32,27 @@ import kotlin.io.path.moveTo
...
  */
-fun stashDirectories(vararg directories: File): Closeable = DirectoryStash(setOf(*directories))
+fun stashDirectories(
+    vararg directories: File,
+    stashParentDir: File = createTempDirectory("stashParent").toFile()
```
fviernau: Is the new stash location of `stashParentDir` guaranteed to be on same file system?

utils/common/src/main/kotlin/DirectoryStash.kt
```diff
@@ -32,24 +32,27 @@ import kotlin.io.path.moveTo
...
+): Closeable = DirectoryStash(setOf(*directories), stashParentDir)
 
 /**
- * A Closable class which temporarily moves away directories and moves them back on close. Any conflicting directory
- * created at the location of an original directory is deleted before the original state is restored. If a specified
- * directory did not exist on initialization, it will also not exist on close.
+ * A [Closeable] class which temporarily moves away [directories] to inside a [stashParentDir] and moves them back on
```
fviernau: I believe the term `parent` is an implementation detail and should not be exposed, as this IMO leads to confusion. Would it make sense to rename this to `targetDir` or `stashDir` or `tempDir` ?

utils/common/src/main/kotlin/DirectoryStash.kt
```diff
@@ -32,24 +32,34 @@ import kotlin.io.path.moveTo
...
+) : Closeable {
     private companion object : Logging
 
     private val stashedDirectories: Map<File, File?> = directories.associateWith { originalDir ->
         // We need to check this on each iteration instead of filtering beforehand to properly handle parent / child
         // directories.
         if (originalDir.isDirectory) {
             // Create a temporary directory to move directories as-is into.
-            val tempDir = createTempDirectory(originalDir.parentFile.toPath(), "stash").toFile()
+            val tempDir = createTempDirectory(stashParentDir.toPath(), "stash").toFile()
```
fviernau, line 62: Do we have the requirement now that `stashParentDir` exists?

fviernau (changes requested):

utils/common/src/main/kotlin/DirectoryStash.kt
```diff
@@ -32,24 +32,34 @@ import kotlin.io.path.moveTo
...
 
 /**
- * A convenience function that stashes directories using a [DirectoryStash] instance.
+ * A convenience function that stashes [directories] to the [stashParentDir] using a [DirectoryStash] instance. If
+ * [deleteParentOnClose] is true, the [stashParentDir] is expected to be empty on close and will be deleted.
  */
-fun stashDirectories(vararg directories: File): Closeable = DirectoryStash(setOf(*directories))
+fun stashDirectories(
+    vararg directories: File,
+    stashParentDir: File = createTempDirectory("stashParent").toFile(),
```
fviernau, line 40: commit-message: Please provide a rationale _why_ the parameters are introduced.

fviernau (changes requested):

utils/common/src/main/kotlin/DirectoryStash.kt
sschuberth: No. I've put some effort into finding out whether there's a reliable way to do that, but I haven't found any. So the best we can probably do is to document that requirement.

utils/common/src/main/kotlin/DirectoryStash.kt
sschuberth: > I believe the term parent is an implementation detail

In what way is it an implementation detail? It's the parent directory that holds all stashed directories, which IMO is something that the user should know in order to chose a proper location for the directory.

I'd rather say that whether `stashParentDir` is a *temporary* directory or not is something that's up to the user (now), so `tempDir` as a name would be misleading.

fviernau: I believe the reason the name contains `parent` is because the implementation creates another directory inside that to ensure it does not exist. That's what I meant with implementation detail.

utils/common/src/main/kotlin/DirectoryStash.kt
sschuberth: I'm now using a completely different approach, and the last commit should explain the details better.

utils/common/src/main/kotlin/DirectoryStash.kt
sschuberth: The new approach now again ensures temporary directories to be created as siblings on the same filesystem.

utils/common/src/main/kotlin/DirectoryStash.kt
sschuberth: This became irrelevant due to the new approach.

utils/common/src/main/kotlin/DirectoryStash.kt
sschuberth: This became irrelevant due to the new approach.

utils/common/src/main/kotlin/DirectoryStash.kt
sschuberth: This became irrelevant due to the new approach.

fviernau (approved):

PR #6525, opened 2023-02-21
fix(PythonInspector): Improve the way project names are derived

### LINKED ISSUE #6524: Unable to determine a project name for Python project
I have installed the `python-inspector` package and added it to the PATH, it resolved most errors, however there is still one error that popped up

```
11:38:27.784 [DefaultDispatcher-worker-2] ERROR org.ossreviewtoolkit.analyzer.PackageManager - Resolving PIP dependencies for path 'setup.py' failed with: IllegalArgumentException: Unable to determine a project name for '/home/complab304pc30/Documents/django/setup.py'.
```

_Originally posted by @HardikRa in https://github.com/oss-review-toolkit/ort/issues/5159#issuecomment-1437912865_

sschuberth:
Actually use `requirementsName` if there is no `setupName`, improve the project name if there are both, and always use the relative path to the analyzer as a fallback.

Fixes #6524.

> **Note**
> This is a **breaking change** under some circumstances Python projects my get assigned different ids than before, which might affect existing curations.

### DIFF
--- a/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
+++ b/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
@@ -247,7 +247,7 @@ private fun PythonInspector.Result.resolveIdentifier(
         // root as a unique project name.
         !hasSetupName && hasRequirementsName -> PackageManager.getFallbackProjectName(analysisRoot, definitionFile)
         hasSetupName && hasRequirementsName -> "$setupName-requirements$requirementsSuffix"
-        else -> throw IllegalArgumentException("Unable to determine a project name for '$definitionFile'.")
+        else -> PackageManager.getFallbackProjectName(analysisRoot, definitionFile)
     }
 
     val projectVersion = setupVersion.takeIf { it.isNotEmpty() } ?: requirementsVersion
--- a/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
+++ b/analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
@@ -226,27 +226,29 @@ private fun PythonInspector.Result.resolveIdentifier(
             )
         } ?: listOf("", "")
 
-    val (requirementsName, requirementsVersion, requirementsSuffix) = projects.find { !it.path.endsWith("/setup.py") }
+    val (requirementsName, requirementsVersion) = projects.find { !it.path.endsWith("/setup.py") }
         ?.let {
             // In case of "requirements*.txt" there is no metadata at all available, so use the parent directory name
             // plus what "*" expands to as the project name and the VCS revision, if any, as the project version.
             val suffix = definitionFile.name.removePrefix("requirements").removeSuffix(".txt")
-            val name = definitionFile.parentFile.name + suffix
+            val name = if (definitionFile.parentFile != analysisRoot) {
+                definitionFile.parentFile.name + suffix
+            } else {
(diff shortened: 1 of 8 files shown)


analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
```diff
@@ -235,19 +235,17 @@ private fun PythonInspector.Result.resolveIdentifier(
...
     val hasRequirementsName = requirementsName.isNotEmpty()
 
     val projectName = when {
         hasSetupName && !hasRequirementsName -> setupName
-        // In case of only a requirements file without further metadata, use the relative path to the analyzer
-        // root as a unique project name.
-        !hasSetupName && hasRequirementsName -> PackageManager.getFallbackProjectName(analysisRoot, definitionFile)
-        hasSetupName && hasRequirementsName -> "$setupName-requirements$requirementsSuffix"
-        else -> throw IllegalArgumentException("Unable to determine a project name for '$definitionFile'.")
+        !hasSetupName && hasRequirementsName -> requirementsName
```
MarcelBochtler, line 246: As it was previously an explicit decision to _not_ use the `requirementsName` can you explain why that has to be changed?

sschuberth: Was it? I do not recall that decision, nor did I read anything in the code about it. Do you remember any details? If it's about a potential project name / id clash, why should a `requirementsName` clash be more likely than a `setupName`?

MarcelBochtler: > Do you remember any details?

No. But the code looks like it was an explicit choice by the author.

But I'm ok with the change as-is.

MarcelBochtler (dismissed):

sschuberth:
> But I'm ok with the change as-is.

Thanks. To be on the safe side I'll wait for another approval from @mnonnenmacher or @fviernau.

analyzer/src/main/kotlin/managers/utils/PythonInspector.kt
mnonnenmacher: > If it's about a potential project name / id clash, why should a requirementsName clash be more likely than a setupName?

Because `setupName` comes from the metadata in `setup.py` and can be set to a meaningful value, but `requirementsName` ist just the name of the file. If a repository contains two directories with the same name and a requirements.txt, would this now lead to a name clash? For example, if there are multiple `test/requirements.txt`?

sschuberth: Thanks for pointing me at that commit. Its message reads as if the problem was mostly with `requirements.txt` files in the project root, and the project root directly might not be very unique (e.g. in CI pipelines that run ORT on a temporary workspace).

So maybe just special-case `definitionFile.parentFile != analysisRoot`?

mnonnenmacher: The mentioning of the "project's base directory" in the commit message seems to be misleading, if it was only about a file in the root dir it would not cause a name conflict within the same analyzer run. Likely the issue was about requirements.txt files in multiple directories with the same name.
I think it would be a good idea to split the changes of the `else` and the `!hasSetupName && hasRequirementsName` branches into two separate commits, then we could easily revert one of them in case it leads to issues with many duplicate identifiers.

mnonnenmacher (approved):

Issue #5555, opened 2022-07-13
GoMod: Improve detecting the VCS info of packages


fviernau:
BLOCKED BY BELOW TICKETS

ORT's logic for determining VCS info currently works only for GitHub. It does not re-use the (non-trivial) logic from the tooling. 
As soon as the following tickets are completed / released, we can use the Go tooling to get the VCS info.
That would make VCS detection finally work and setting the source artifact URL can be dropped.

1. https://github.com/golang/go/issues/18387
2. https://github.com/golang/go/issues/44742

sschuberth:
> BLOCKED BY BELOW TICKETS

This does seem to be true anymore as your https://github.com/oss-review-toolkit/ort/pull/6529 proposes a fix while https://github.com/golang/go/issues/18387 is still unresolved, right?

PR #6529, opened 2023-02-21
GoMod: Properly resolve VCS infos of dependencies

fviernau:
This PR eliminates ORT's own logic for VCS info resolution in favor of using the Go command line tooling for that purpose.

See individual commits.

Fixes: #5555.


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -54,6 +54,7 @@ import org.ossreviewtoolkit.utils.common.Os
 import org.ossreviewtoolkit.utils.common.splitOnWhitespace
 import org.ossreviewtoolkit.utils.common.stashDirectories
 import org.ossreviewtoolkit.utils.common.withoutSuffix
+import org.ossreviewtoolkit.utils.ort.createOrtTempDir
```
sschuberth, line 56: Commit message: s/the users environment/the user's environment/

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -81,6 +82,15 @@ class GoMod(
         ) = GoMod(type, analysisRoot, analyzerConfig, repoConfig)
     }
 
+    private val goDir by lazy { createOrtTempDir() }
+
+    private val environment by lazy {
+        mapOf(
+            "GOPROXY" to "direct",
+            "GOPATH" to goDir.absolutePath
```
sschuberth, line 92: How about renaming `goDir` to `goPath` to resemble the `GOPATH` environment variable name?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -221,7 +221,10 @@ class GoMod(
         val indirect: Boolean = false,
 
         @JsonProperty("Main")
-        val main: Boolean = false
+        val main: Boolean = false,
+
+        @JsonProperty("GoMod")
+        val goMod: String
```
sschuberth, line 231: It seems a bit overly fine-grained to me to split this from the next commit.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -501,25 +486,39 @@ private const val PACKAGE_SEPARATOR = "# "
...
-        it.groupValues[1]
-    } ?: path
+private fun GoMod.ModuleInfo.toVcsInfo(): VcsInfo {
+    val info = jsonMapper.readValue<ModuleInfoFile>(File(goMod).resolveSibling("$version.info"))
+    if (info.origin.vcs != "git") return VcsInfo.EMPTY
 
-    val vcsInfo = VcsHost.parseUrl("https://$hostname")
-    return vcsInfo.copy(revision = getRevision(version))
+    return VcsInfo(
+        type = VcsType.GIT,
```
sschuberth, line 522: Instead of hard-coding Git here and inline 514, consider something like

```
    ...
    val type = info.origin.vcs?.let { VcsType.forName(it) }.takeIf { it == VcsType.GIT } ?: return VcsInfo.EMPTY

    return VcsInfo(
        type = type,
        ...
```

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -501,25 +486,39 @@ private const val PACKAGE_SEPARATOR = "# "
...
+        url = info.origin.url!!,
+        revision = info.origin.hash!!,
```
sschuberth, line 524: Consider using `checkNotNull()` instead of `!!`.

analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
```diff
@@ -40,6 +40,7 @@ class GoModFunTest : StringSpec({
 
         val result = createGoMod().resolveSingleProject(definitionFile)
 
+        testDir.resolve("gomod-expected-output.yml").writeText(result.toYaml())
```
sschuberth, line 43: Left-over?

analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
```diff
@@ -49,6 +50,7 @@ class GoModFunTest : StringSpec({
 
         val result = createGoMod().resolveSingleProject(definitionFile)
 
+        testDir.resolve("gomod-subpkg-expected-output.yml").writeText(result.toYaml())
```
sschuberth, line 53: Left-over?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -96,6 +99,8 @@ class GoMod(
 
     override fun transformVersion(output: String) = output.removePrefix("go version go").substringBefore(' ')
 
+    override fun getVersionRequirement(): RangesList = RangesListFactory.create("[1.19.0,)")
```
sschuberth, line 102: If this is a requirement to make the code in the previous commit work, this code should be squashed into the previous commit (or be moved to a preceding commit to guarantee the requirements for the upcoming change).

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -81,6 +82,15 @@ class GoMod(
         ) = GoMod(type, analysisRoot, analyzerConfig, repoConfig)
     }
 
+    private val goPath by lazy { createOrtTempDir() }
+
+    private val environment by lazy {
+        mapOf(
+            "GOPROXY" to "direct",
+            "GOPATH" to goPath.absolutePath
```
sschuberth, line 92: It just occurred to me that there are no changes to expected results in this commit. Do test really still pass as-is with these changes?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -226,7 +225,10 @@ class GoMod(
         val indirect: Boolean = false,
 
         @JsonProperty("Main")
-        val main: Boolean = false
+        val main: Boolean = false,
+
+        @JsonProperty("GoMod")
+        val goMod: String
```
sschuberth: Could this be of type `File` right away?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -514,9 +514,9 @@ private data class ModuleInfoFile(
     )
 }
 
-private fun GoMod.ModuleInfo.toVcsInfo(): VcsInfo {
+private fun GoMod.ModuleInfo.toVcsInfo(): VcsInfo? {
     val info = jsonMapper.readValue<ModuleInfoFile>(File(goMod).resolveSibling("$version.info"))
-    val type = info.origin.vcs?.let { VcsType.forName(it) }.takeIf { it == VcsType.GIT } ?: return VcsInfo.EMPTY
+    val type = info.origin.vcs?.let { VcsType.forName(it) }.takeIf { it == VcsType.GIT } ?: return null
```
sschuberth, line 519: This reminds me that we should probably rather say `.takeUnless { it.type == VcsType.UNKNOWN }` here, either as part of this commit, or even already as part of the commit that introduced `.takeIf { it == VcsType.GIT }`.

sschuberth: Also, that would allow you to keep `?: return VcsInfo.EMPTY` here, but in the caller just use `toVcsInfo()` (without `orEmpty()`).

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Yes, they still pass. It'd be quite bad if dependency trees differed depending on whether a proxy was used, wouldn't it?

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: By using `.takeUnless { it.type == VcsType.UNKNOWN }` this would extend the code beyond just `git` as VCS.
This would be completely untested. I neither would like to extend the scope of this commit to test this nor I'd like to do this fully untested.

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Probably it could, I prefer to keep it a String.

analyzer/src/main/kotlin/managers/GoMod.kt
sschuberth: I'm still wondering why you're not leaving `toVcsInfo()` untouched and instead just change line 312 to `toVcsInfo()`, because `toVcsInfo()` can never return `VcsType.UNKNOWN`.

fviernau: I find my current version cleaner, `toVcsInfo()` return non-null if and only if it was able to determine VCS info.

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: > Yes. But it's Go after all, and as you [explained](https://github.com/oss-review-toolkit/ort/pull/6529/commits/9a09a7af8992d90fa57154048a27033e0efdc8ba) it depends on the proxy instance whether VCS information is available or not (which is quite sick, IMO), so I wouldn't have been surprised if _not_ using a proxy yields different results than before.

My explanation was talking about the presence of VCS info in the `.info` files. 
In this commit, that VCS info from those info files is not yet consumed at all.
This commit only ensures that the info is in these files.

sschuberth (approved):

PR #6533, opened 2023-02-22
Allow project / package "duplicates" with the same VCS location

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
+++ b/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
@@ -77,7 +77,7 @@ class AnalyzerResultBuilder {
                 val projectIssues = issues.getOrDefault(existingProject.id, emptyList())
                 issues[existingProject.id] = projectIssues + issue
             } else {
-                projects += projectAnalyzerResult.project
+                addProject(projectAnalyzerResult.project)
                 addPackages(projectAnalyzerResult.packages)
 
                 if (projectAnalyzerResult.issues.isNotEmpty()) {
@@ -85,6 +85,12 @@ class AnalyzerResultBuilder {
                 }
             }
         }
+
+    /**
+     * Add the given [project] to this builder. This function can be used for projects that have been obtained
+     * independently of a [ProjectAnalyzerResult].
+     */
+    fun addProject(project: Project) = apply { projects += project }
 
     /**
      * Add the given [packageSet] to this builder. This function can be used for packages that have been obtained
--- a/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
+++ b/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
@@ -41,9 +41,19 @@ class AnalyzerResultBuilder {
 
     fun build(excludes: Excludes = Excludes.EMPTY): AnalyzerResult {
         val duplicates = (projects.map { it.toPackage() } + packages).getDuplicates { it.id }
-        require(duplicates.isEmpty()) {
+
+        // Some source code repositories contain projects whose build artifacts are used as packages by other projects
+        // in the same repository. So the same code is discovered once as a project, and once as a package. Allow such
+        // "duplicates" as the ids actually refer to the same code by requiring duplicates to refer to different VCS
+        // locations.
+        val realDuplicates = duplicates.filter { (_, duplicates) ->
+            val vcsInfo = duplicates.first().vcsProcessed
+            vcsInfo.path.isEmpty() || !duplicates.all { it.vcsProcessed == vcsInfo }
+        }
+
+        require(realDuplicates.isEmpty()) {
             "Unable to create the AnalyzerResult as it contains packages and projects with the same ids: " +
-                duplicates.values
+                realDuplicates.values
         }
 
         return AnalyzerResult(projects, packages, issues, dependencyGraphs)
--- a/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
+++ b/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
@@ -19,6 +19,7 @@
 
 package org.ossreviewtoolkit.analyzer
 
+import io.kotest.assertions.throwables.shouldNotThrow
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.inspectors.forAll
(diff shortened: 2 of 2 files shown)


analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -46,9 +46,18 @@ class AnalyzerResultBuilder {
 
     fun build(excludes: Excludes = Excludes.EMPTY): AnalyzerResult {
         val duplicates = (projects.map { it.toPackage() } + packages).getDuplicates { it.id }
-        require(duplicates.isEmpty()) {
+
+        // Some source code repositories contain projects that are used as packages by other contained projects. So the
+        // same code is once seen as a project, and once as a package by ORT. Allow such "duplicates" as the ids
+        // actually refer to the same thing by requiring duplicates to refer to different VCS locations.
+        val realDuplicates = duplicates.filter { (_, duplicates) ->
```
sschuberth, line 53: @oss-review-toolkit/core-devs, what do you thing about this? Should we rather only keep one of these (either the project or the package) instead of keeping both of them?

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -46,9 +46,18 @@ class AnalyzerResultBuilder {
 
     fun build(excludes: Excludes = Excludes.EMPTY): AnalyzerResult {
         val duplicates = (projects.map { it.toPackage() } + packages).getDuplicates { it.id }
-        require(duplicates.isEmpty()) {
+
```
fviernau, line 49: I've got some story maybe worth sharing here: I recently scanned ORT with ORT locally, 
it exited complaining in about duplicates in this function. I debugged it only quick and had to stop at some point due to other tasks. If I didn't make a mistake the finding was that there were multiple packages (not projects) with the same `id` which were completely identical except for on attribute. IIRC it was the homepage URL.

That's just FYI, didn't look into this PR deeply yet .

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
sschuberth: I'm specifically asking as I currently don't oversee all side-effects that might occur if we keep a project and a package with the same id in the result.

fviernau: I think it's likely that assumption that Id's are unique are spread all over the place.  To list a few:

- https://github.com/oss-review-toolkit/ort/blob/19c89ff9a0a7aa2a52d85c82fd477531da1ecf3d/model/src/main/kotlin/OrtResult.kt#L451
- https://github.com/oss-review-toolkit/ort/blob/19c89ff9a0a7aa2a52d85c82fd477531da1ecf3d/model/src/main/kotlin/OrtResult.kt#L532
- https://github.com/oss-review-toolkit/ort/blob/19c89ff9a0a7aa2a52d85c82fd477531da1ecf3d/helper-cli/src/main/kotlin/commands/ListLicensesCommand.kt#L66
- https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/kotlin/OrtResult.kt#L132

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
oheger-bosch: I also had the case of multiple packages with the same identifier. This happened when a package was referenced from multiple package managers using the same package type (for instance _GoMod_ and _GoDep_). If the package metadata is not exactly the same, multiple `Package` instances are created.

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
fviernau: > Right. But some of these only assume ids to be unique within projects _or_ packages, but not within the union of both.

Yeah, that's for sure.

fviernau: Have you considered as alternative solution: Making sure that the `Identifier.type` for projects differs from `Identifier.type` for packages?

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
fviernau: In my above case the full identifiers were identical, including the `type`.

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
oheger-bosch: ISTR that there was also a problem with the WebApp report when duplicate IDs were present: The JavaScript part uses an index-based mapping to reference objects, and if there were projects and packages with the same ID, this mapping got broken.

sschuberth: > Have you considered as alternative solution: Making sure that the `Identifier.type` for projects differs from `Identifier.type` for packages?

Actually, I was playing around with similar ideas 3 years (!) ago:
- In the [identifier-types](https://github.com/oss-review-toolkit/ort/compare/main...identifier-types) branch I tried to turn an `Identifier`'s `type` into a `sealed class` with `ProjectIdentifier` and `PackageIdentifier` sub-classes.
- In the [multiple-identifier-classes](https://github.com/oss-review-toolkit/ort/compare/main...multiple-identifier-classes) branch the `Identifier` class itself is tried to be turned in a `sealed class`.

Both approaches had the goal to distinguish project and package ids, and to iterate over all known id types (e.g. for `when` statements).

mnonnenmacher: I agree with @fviernau that there are probably lots of places in the code which assume that identifiers are unique and it could be very hard to find all of them, and with @oheger-bosch that likely the WebApp report and probably other report formats would break.
There are also situations where the same package could appear twice with different metadata (e.g. if different repositories are used), I'm not sure if we currently fail in this case or just ignore it.
I think properly addressing this issue would be a big effort, another approach than the sealed class you suggested could also be to extend the identifier with properties like `isProject` and provider details.

sschuberth: > I think properly addressing this issue would be a big effort

Indeed. That's why I never completed my branches mentioned above...

sschuberth:
> The change have been tested and is fine from our side

Thanks for the feedback. However, this PT will more likely evolve into the direction of dropping the duplicate package in favor of only keeping the project.

PR #6541, opened 2023-02-23
style(CocoaPods): Remove an unnecessary "at all"

mnonnenmacher:


sschuberth (approved):

sschuberth:
Should we mark this as a breaking change as theoretically existing error resolution could be affected?

mnonnenmacher:
> Should we mark this as a breaking change as theoretically existing error resolution could be affected?

I'm not sure, that would mean that all changes to issue messages would be breaking changes, that seems excessive.

sschuberth (approved):

PR #6544, opened 2023-02-23
Fallback project names

mnonnenmacher:
Align the fallback project names that are used when the name cannot be determined from the project's metadata. This also reduces the risk of duplicate identifiers for some package managers. See the commit messages for details.

**Breaking change:**
Project names could change for the Bundler, Carthage, and GoDep package managers.

### DIFF
--- a/analyzer/src/main/kotlin/PackageManager.kt
+++ b/analyzer/src/main/kotlin/PackageManager.kt
@@ -232,6 +232,13 @@ abstract class PackageManager(
          */
         private fun Excludes.isPathExcluded(root: Path, path: Path): Boolean =
             isPathExcluded(root.relativize(path).pathString)
+
+        /**
+         * Get a fallback project name from the [definitionFile] path relative to the [analysisRoot]. This function
+         * should be used if the project name cannot be determined from the project's metadata.
+         */
+        fun getFallbackProjectName(analysisRoot: File, definitionFile: File) =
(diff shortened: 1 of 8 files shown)


fviernau (dismissed):

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -300,7 +300,7 @@ abstract class PackageManager(
         definitionFiles.forEach { definitionFile ->
             val relativePath = definitionFile.relativeTo(analysisRoot).invariantSeparatorsPath.ifEmpty { "." }
 
-            logger.info { "Resolving $managerName dependencies for path '$relativePath'..." }
+            logger.info { "Resolving $managerName dependencies for definition file '$relativePath'..." }
```
sschuberth, line 310: This change revealed to me that the `.ifEmpty { "." }` two lines doesn't really make sense. First of all, it should not happen, and secondly even if it was triggered "." would be no "definition file". Mind adding a commit that simply removes `.ifEmpty { "." }`?

sschuberth: Ah, exactly... that's also the reason why this was more generally called "path" instead of "definition file". So should we keep it at "path"?

mnonnenmacher: Yes, I have dropped the commit.

sschuberth (approved):

PR #6548, opened 2023-02-24
chore(Osv): Improve the wording of a `TODO` statement 

### LINKED ISSUE #6536: OSV advisor TODO in C++
I found this TODO in the OSV advisor: https://github.com/oss-review-toolkit/ort/blob/af0a83a2f83b2fe80ee7e0559f135200a65478fc/advisor/src/main/kotlin/advisors/Osv.kt#L165

Also, I noticed that Conan is not there.

Could someone please detail what is missing/needed, so we can determine whether we need to contribute to this? We are C++ users and would like to ensure we are not missing any advisory.

Thanks.

sschuberth:
Resolves #6536.

### DIFF
--- a/advisor/src/main/kotlin/advisors/Osv.kt
+++ b/advisor/src/main/kotlin/advisors/Osv.kt
@@ -163,7 +163,9 @@ private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
         )
     }
 
-    // TODO: Handle C++ projects and / or requesting vulnerabilities by commit-ish.
+    // TODO: Support querying vulnerabilities by Git commit hash as described at https://osv.dev/docs/#section/OSV-API.
+    //       That would allow to generally support e.g. C / C++ projects that do not use a dedicated package manager
+    //       like Conan.
 
     return null
 }


advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -163,8 +163,6 @@ private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
         )
     }
 
-    // TODO: Handle C++ projects and / or requesting vulnerabilities by commit-ish.
```
fviernau, line 166: commit-message:

> here is no ecosystem documented at [1] that would support "requesting
vulnerabilities by commit-ish"

commit-ish is ecosystem agnostic

> , nor is it supported to generally handle
C++ projects independently of a package / dependency manager.

The following does not count?

```
curl -X POST -d '{"commit": "6879efc2c1596d11a6a6ad296f80063b558d5e0f"}' "https://api.osv.dev/v1/query"
```

> On top of
that, support for Conan (which is a package manager for C / C++) has been
added 

This reads as if all C++ projects would use Conan...

fviernau (changes requested):

advisor/src/main/kotlin/advisors/Osv.kt
sschuberth: As you didn't comment on #6536 I simply have it a shot 😉 Thanks for the clarification, I've instead reworded the `TODO` statement now.

fviernau (approved):

PR #6552, opened 2023-02-24
deps(reporter-web-app): Update antd package

mennaElnemr9:
Update antd package from version 4.16.13 to version 5.2.2.

This update is a backward-incompatible and requires some code fixes.

File antd.css that is imported in App.js and App.css no longer exists in the new version so this import line is removed.

Since version 5.2.2 antd is following CSS-in-JS approach for better support of dynamic themes all CSS files no longer exist.

Except one CSS file which is reset.css for resetting some basic styles.

The file size changes because of the update so the file range in WebAppReporterFunTest.kt is changed to 1830000L..1860000L.

### DIFF
--- a/plugins/reporters/web-app-template/src/App.css
+++ b/plugins/reporters/web-app-template/src/App.css
@@ -1,5 +1,3 @@
-@import '~antd/dist/antd.css';
-
 body {
     background: #fafafa;
     font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
@@ -15,6 +13,11 @@ th {
     padding-right: 10px;
 }
 
+b,
+strong {
+    font-weight: bolder;
+}
+
 /* === Ant Design customizations === */
 
 /*
@@ -26,39 +29,41 @@ https://ant.design/docs/react/customize-theme
 .ant-tabs,
 .ant-tag,
 .ant-table,
-.ant-tree {
-    font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+.ant-tree,
+.ant-timeline,
+.ant-timeline-item-content {
+    font-family: "Helvetica Neue", Helvetica, Arial, sans-serif !important;
 }
 
 /* Reduce margin around <Ant Col> to allow content to be displayed */
 .ant-col-offset-1 {
-    margin-left: 2.66666667%;
+    margin-left: 2.66666667% !important;
 }
 
 .ant-col-22 {
-    width: 94.16666667%;
+    width: 94.16666667% !important;
 }
 
 .ant-table-small > .ant-table-content > .ant-table-body {
-    margin: 0px;
+    margin: 0px !important;
 }
 
 tr.ant-table-expanded-row, tr.ant-table-expanded-row,
 tr.ant-table-expanded-row, tr.ant-table-expanded-row:hover {
-    background: #fff;
+    background: #fff !important;
 }
 
 .ant-steps-dot .ant-steps-item-content {
-    width: auto;
+    width: auto !important;
 }
 
 /* Styling specific for last item in SummaryView Timeline */
 .ant-timeline-item-last {
-    padding-bottom: 1em;
+    padding-bottom: 1em !important;
 }
 
 .ant-timeline-item-last > .ant-timeline-item-content {
-    min-height: 1em;
+    min-height: 1em !important;
 }
 
 /* === Styling for shared ORT elements === */
--- a/plugins/reporters/web-app-template/yarn.lock
+++ b/plugins/reporters/web-app-template/yarn.lock
@@ -3,24 +3,17 @@
 
 
 "@adobe/css-tools@^4.0.1":
-  version "4.1.0"
-  resolved "https://registry.yarnpkg.com/@adobe/css-tools/-/css-tools-4.1.0.tgz#417fef4a143f4396ad0b3b4351fee21323f15aa8"
-  integrity sha512-mMVJ/j/GbZ/De4ZHWbQAQO1J6iVnjtZLc9WEdkUQb8S/Bu2cAF2bETXUgMAdvMG3/ngtKmcNBe+Zms9bg6jnQQ==
+  version "4.2.0"
+  resolved "https://registry.yarnpkg.com/@adobe/css-tools/-/css-tools-4.2.0.tgz#e1a84fca468f4b337816fcb7f0964beb620ba855"
+  integrity sha512-E09FiIft46CmH5Qnjb0wsW54/YQd69LsxeKUOWawmws1XWvyFGURnAChH0mlr7YPFR1ofwvUQfcL0J3lMxXqPA==
 
 "@ampproject/remapping@^2.2.0":
-  version "2.2.0"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
-  integrity sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==
+  version "2.2.1"
+  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.1.tgz#99e8e11851128b8702cd57c33684f1d0f260b630"
+  integrity sha512-lFMjJTrFL3j7L9yBxwYfCq2k6qqwHyzuUl/XBnif78PWTJYyL/dfowQHWE3sp6U6ZzqWiiIZnpTMO96zhkjwtg==
   dependencies:
-    "@jridgewell/gen-mapping" "^0.1.0"
+    "@jridgewell/gen-mapping" "^0.3.0"
     "@jridgewell/trace-mapping" "^0.3.9"
-
-"@ant-design/colors@^6.0.0":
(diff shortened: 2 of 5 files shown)


mennaElnemr9:
> Update antd package from version 4.16.13 to version 5.2.2.
> 
> This update is a backward-incompatible and requires some code fixes.
> 
> File antd.css that is imported in App.js and App.css no longer exists in the new version so this import line is removed.
> 
> Since version 5.2.2 antd is following CSS-in-JS approach for better support of dynamic themes all CSS files no longer exist.
> 
> Except one CSS file which is reset.css for resetting some basic styles.
> 
> The file size changes because of the update so the file range in WebAppReporterFunTest.kt is changed to 1830000L..1860000L.
> 
> Please ensure that your pull request adheres to our [contribution guidelines](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md). Thank you!

The font of the web-app-reporter has some changes because of the antd updates, that's because antd package remove antd.css that had some styles that can be overridden by App.css but with the new update the styles cannot be overridden anymore unless we add the CSS keyword `!important` besides the styles in App.css that needed to have the highest priority and to override antd styles.

sschuberth:
Thanks a lot @mennaElnemr9 for your work in this area! With all the updates, do you also manually ensure that the web-app report still works and looks as expected?

jens-erdmann:
> Thanks a lot @mennaElnemr9 for your work in this area! With all the updates, do you also manually ensure that the web-app report still works and looks as expected?

That is what the previous comment is about. There is a visual difference due to the fact that certain styles are no longer defined by antd. This is something for the project to discuss IMHO.
Are these styles essential or not.

mennaElnemr9:
> Thanks a lot @mennaElnemr9 for your work in this area! With all the updates, do you also manually ensure that the web-app report still works and looks as expected?

After every update I run the project manually to see the effects of every update .
there is also another change besides the one I already mentioned
before updating antd:
image

After updating antd:

image

You can see that "clear filters" tab is moved to the left instead of being on the right.

sschuberth:
> You can see that "clear filters" tab is moved to the left instead of being on the right.

And the font in the table is now grayed out it seems!?

mennaElnemr9:
So one of the solutions to this problem is using `!important` which is a CSS keyword to override the antd styles with the app styles like here
image
 
 in the App.css there are some antd components with some additive features so the app looks how it should look , after the antd update the styles of the antd components styles cannot be overridden by just mentioning them in the CSS file and add the properties that we want to change as the antd styles has the priority , So when using the `!important` it gives the priority to the App.css styles
 we have to use it with all the styles that are needed to be applied over antd components.

mennaElnemr9:
> @tsteenbe It would be good to get your feedback here, especially on the CSS changes.
> 
> > So one of the solutions to this problem is using `!important` which is a CSS keyword to override the antd styles with the app styles like here image
> > in the App.css there are some antd components with some additive features so the app looks how it should look , after the antd update the styles of the antd components styles cannot be overridden by just mentioning them in the CSS file and add the properties that we want to change as the antd styles has the priority , So when using the `!important` it gives the priority to the App.css styles we have to use it with all the styles that are needed to be applied over antd components.
> 
> @mennaElnemr9 Based on the code comment below it seems we are not using the "official" way to customize the Ant look anyway, so I think any quick fix is ok as the ultimate goal should be to switch to the correct way of theme customization anyway.
> 
> https://github.com/oss-review-toolkit/ort/blob/be4bb69349126b7fb595411efe1fb770e9c16132/reporter-web-app/src/App.css#L20-L23

So shall I go with !important solution and adjust the code?

mnonnenmacher:
> So shall I go with !important solution and adjust the code?

I would be ok with that.

tsteenbe (changes requested):
Don't merge this PR. Upgrade Ant Design results in multiple deprecation warning/errors. 

@mennaElnemr9  I will merge your commit into a large upcoming WebApp PR that I have been working on in which I already rewrote the WebApp code to use Ant Design 5 components API.

jens-erdmann:
> Don't merge this PR. Upgrade Ant Design results in multiple deprecation warning/errors.
> 
> @mennaElnemr9 I will merge your commit into a large upcoming WebApp PR that I have been working on in which I already rewrote the WebApp code to use Ant Design 5 components API.

Hi @tsteenbe, can you elaborate on these errors? Don't take this personal but I rather have them fixed and this PR finalized than waiting for your PR to emerge at some point in time.

mennaElnemr9:
I just added some lines to App.css ,

So the final result looks like before updating the antd package

sschuberth:
> So the final result looks like before updating the antd package

Sounds great! Could you please rebase your PR onto latest `main`? The reporter files have moved, but a locally running `git` should be able to figure it out automatically without running into any conflicts.

sschuberth:
Please remove the `reporter-web-app/node_modules` directory from your commit.

sschuberth:
Please also drop the unrelated changes to the `.gitmodules` files. In general, please always double-check that https://github.com/oss-review-toolkit/ort/pull/6552/files contains only the expected diff.

plugins/reporters/web-app-template/src/App.css
```diff
@@ -249,4 +264,4 @@ tr.ant-table-expanded-row, tr.ant-table-expanded-row:hover {
...
     background-repeat: no-repeat;
-}
+}
```
sschuberth, line 267: Please re-introduce this final newline.

plugins/reporters/web-app-template/yarn.lock
```diff
@@ -1,13300 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
sschuberth, line 1: Why has this file been deleted? That seem wrong.

plugins/reporters/web-app-template/src/App.css
mennaElnemr9: what do you mean by "re-introduce" it ?

sschuberth: See the diff:

image

It was there before, but you've removed it without reason, so you should re-introduce it.

mennaElnemr9: I just added new lines to the file so the final line moved to line 257

sschuberth: Having new lines in the file is fine, that's not what I meant. The problem is this red icon on the right-hand side in the green area of the diff. It indicates that you removed a final newline character from the end of the file. That single character is what you need to restore.

mennaElnemr9: Done

plugins/reporters/web-app/src/funTest/kotlin/WebAppReporterFunTest.kt
```diff
@@ -36,7 +36,7 @@ class WebAppReporterFunTest : WordSpec({
 
             val report = WebAppReporter().generateReport(ReporterInput(ortResult), outputDir).single()
 
-            report.length() should beInRange(2140000L..2160000L)
+            report.length() should beInRange(1830000L..1860000L)
```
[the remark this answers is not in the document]
mennaElnemr9: done

plugins/reporters/web-app/src/funTest/kotlin/WebAppReporterFunTest.kt
```diff
@@ -36,7 +36,7 @@ class WebAppReporterFunTest : WordSpec({
 
             val report = WebAppReporter().generateReport(ReporterInput(ortResult), outputDir).single()
 
-            report.length() should beInRange(2140000L..2160000L)
+            report.length() should beInRange(1870000L..1880000L)
```
sschuberth, line 39: Any idea why the report became so much smaller? This looks a bit suspicious, to be honest.

sschuberth:
> So the final result looks like before updating the antd package

As discussed in the ORT Community meeting, @jens-erdmann and @tsteenbe please align and double-check on this.

tsteenbe:
WebApp breaks after Ant Design upgrade to v5 - have been working a proper upgrade in https://github.com/oss-review-toolkit/ort/tree/webapp-reporter-v0.3/

mennaElnemr9:
> WebApp breaks after Ant Design upgrade to v5 see for hints the developer console which turns red with errors/warnings - have been working a proper upgrade in https://github.com/oss-review-toolkit/ort/tree/webapp-reporter-v0.3/ - have already fixed the UI issues.

Can you tell me when does it exactly break ?as I tried to test it thoroughly and it works as expected, what functionality exactly that leads the app to break so I can test it
 ?

sschuberth:
@mennaElnemr9 and @tsteenbe could you try to sync up to move this forward? E.g. by creating a draft PR based on @tsteenbe's https://github.com/oss-review-toolkit/ort/tree/webapp-reporter-v0.3/ branch?

tsteenbe:
Marking this pull request as on-hold as it soon be superseded my code changes on the [webapp-reporter-v0.3](https://github.com/oss-review-toolkit/ort/tree/webapp-reporter-v0.3) branch which contains a larger update of the WebApp reporter

tsteenbe:
@mennaElnemr9 Thank you for your contribution but I am closing your pull request in favor of https://github.com/oss-review-toolkit/ort/pull/7952 which implements same update but includes WebApp code changes to remediate deprecation of various APIs in Ant Design v5.

PR #6209, opened 2022-12-11
Dotnet: Add support for nuget-inspector

pombredanne:
This PR implements a new support for .NET project and NuGet dependencies and replaces the previous "DotNet" and "NuGet" implementations. This is using nuget-inspector a new .NET-native C# utility to resolve NuGet dependencies.

The approach is to add a new plugin to analyze .Net and NuGet-based projects using nuget-inspector to replace the previous "DotNet" and "NuGet" implementations that were not able to handle several common .Net project setups including variables, conditional inclusion of package references, resolution for a specific target framework and handling the many corner
cases that the NuGet client can handle. nuget-inspector uses the .Net, MSBuild and NuGet libraries in the same way as the NuGet client to resolve NuGet dependencies in the same manner.

Reference: https://github.com/nexB/nuget-inspector
Reference: https://github.com/oss-review-toolkit/ort/issues/5038
Signed-off-by: Tushar Goel <tushar.goel.dav@gmail.com>
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

PR #6554, opened 2023-02-24
Advisors: Ensure only non-empty PURLs are used in requests

MarcelBochtler:
It is possible that a package declares an empty string as PURL, e.g. for non-published packages. This results in a `Bad Request`. Avoid this by filtering the packages, before requesting the package details.

### DIFF
--- a/advisor/src/main/kotlin/advisors/NexusIq.kt
+++ b/advisor/src/main/kotlin/advisors/NexusIq.kt
@@ -83,7 +83,7 @@ class NexusIq(name: String, private val nexusIqConfig: NexusIqConfiguration) : A
     override suspend fun retrievePackageFindings(packages: Set<Package>): Map<Package, List<AdvisorResult>> {
         val startTime = Instant.now()
 
-        val components = packages.map { pkg ->
+        val components = packages.filter { it.purl.isNotEmpty() }.map { pkg ->
             val packageUrl = buildString {
                 append(pkg.purl)
 
--- a/advisor/src/main/kotlin/advisors/OssIndex.kt
+++ b/advisor/src/main/kotlin/advisors/OssIndex.kt
@@ -71,7 +71,7 @@ class OssIndex(name: String, serverUrl: String? = null) : AdviceProvider(name) {
     override suspend fun retrievePackageFindings(packages: Set<Package>): Map<Package, List<AdvisorResult>> {
         val startTime = Instant.now()
 
-        val purls = packages.map { it.purl }
+        val purls = packages.mapNotNull { pkg -> pkg.purl.takeUnless { it.isEmpty() } }
 
         return try {
             val componentReports = mutableMapOf<String, OssIndexService.ComponentReport>()
--- a/advisor/src/main/kotlin/advisors/VulnerableCode.kt
+++ b/advisor/src/main/kotlin/advisors/VulnerableCode.kt
@@ -87,7 +87,7 @@ class VulnerableCode(name: String, vulnerableCodeConfiguration: VulnerableCodeCo
         packages: List<Package>,
         startTime: Instant
     ): Map<Package, List<AdvisorResult>> {
-        val packageMap = packages.associateBy { it.purl }
+        val packageMap = packages.filter { it.purl.isNotEmpty() }.associateBy { it.purl }
         val packageVulnerabilities = service.getPackageVulnerabilities(PackagesWrapper(packageMap.keys))
 
         return packageVulnerabilities.filter { it.unresolvedVulnerabilities.isNotEmpty() }.mapNotNull { pv ->


sschuberth:
> It is possible that a package declares an empty string as PURL, e.g. for non-published packages.

I believe that comment is a bit misleading, even if I known about the [background discussion](https://github.com/oss-review-toolkit/ort/pull/6209#discussion_r1117062705). Also for non-published packages you can construct a PURL how it would be valid once the package was published.

Anyway, for advisors that directly take a PURL as part of their API it surely makes sense to verify that a PURL is valid / present.

advisor/src/main/kotlin/advisors/NexusIq.kt
```diff
@@ -83,7 +83,7 @@ class NexusIq(name: String, private val nexusIqConfig: NexusIqConfiguration) : A
     override suspend fun retrievePackageFindings(packages: Set<Package>): Map<Package, List<AdvisorResult>> {
         val startTime = Instant.now()
 
-        val components = packages.map { pkg ->
+        val components = packages.filter { it.purl.isNotBlank() }.map { pkg ->
```
sschuberth, line 86: Would `isNotEmpty()` be sufficient? Because otherwise I would wonder where a blank (but non-empty) PURL would come from... and consider it to be a bug in the origin of such a blank PURL.

oheger-bosch:
An analogous check needs to be added to the `VulnerableCode` advisor, too. It also does not like empty PURLs.

MarcelBochtler:
> Also for non-published packages you can construct a PURL how it would be valid once the package was published.

Yes, this is correct. But does this mean, that the Analyzer should never return an empty PURL, but construct one based on the package ID (option 1)?
Alternatively we could do this in the Advisors, which have issues with empty requests (option 2).
Or filter the empty PURLs as it is currently done in this PR (option 3).

Which option do you prefer? All 3 should fix the issue at hand.

sschuberth:
> Or filter the empty PURLs as it is currently done in this PR (option 3).

IMO we should do option 3 for all advisors that take PURLs, to fix the issue also for people who use the advisors programmatically.

sschuberth:
And we'd also need the same fix for OssIndex, too, which might resolve #6564.

advisor/src/main/kotlin/advisors/VulnerableCode.kt
```diff
@@ -90,13 +90,14 @@ class VulnerableCode(name: String, vulnerableCodeConfiguration: VulnerableCodeCo
...
-            }
-        }.toMap()
+        return packageVulnerabilities.filter { it.unresolvedVulnerabilities.isNotEmpty() && it.purl.isNotEmpty() }
+            .mapNotNull { pv ->
+                packageMap[pv.purl]?.let { pkg ->
+                    val vulnerabilities = pv.unresolvedVulnerabilities.map { it.toModel() }
+                    val summary = AdvisorSummary(startTime, Instant.now())
+                    pkg to listOf(AdvisorResult(details, summary, vulnerabilities = vulnerabilities))
+                }
+            }.toMap()
```
sschuberth, line 100: Nit: I guess our preferred reformatting would be like

        return packageVulnerabilities.filter {
            it.unresolvedVulnerabilities.isNotEmpty() && it.purl.isNotEmpty()
        }.mapNotNull { pv ->
            packageMap[pv.purl]?.let { pkg ->
                val vulnerabilities = pv.unresolvedVulnerabilities.map { it.toModel() }
                val summary = AdvisorSummary(startTime, Instant.now())
                pkg to listOf(AdvisorResult(details, summary, vulnerabilities = vulnerabilities))
            }
        }.toMap()

as it reduces indentation changes and thus keeps the diff smaller.

advisor/src/main/kotlin/advisors/OssIndex.kt
```diff
@@ -71,7 +71,7 @@ class OssIndex(name: String, serverUrl: String? = null) : AdviceProvider(name) {
     override suspend fun retrievePackageFindings(packages: Set<Package>): Map<Package, List<AdvisorResult>> {
         val startTime = Instant.now()
 
-        val purls = packages.map { it.purl }
+        val purls = packages.mapNotNull { it.purl.ifEmpty { null } }
```
sschuberth, line 74: Alternatively (longer but more speaking): `it.purl.takeUnless { it.isEmpty() }`

advisor/src/main/kotlin/advisors/VulnerableCode.kt
MarcelBochtler: The filtering has to be done in a different place, therefore this change is no longer in place.

sschuberth (approved):

PR #6559, opened 2023-02-26
build: Fix executing tests

mnonnenmacher:
Since 33fdca9 test from the `test` source sets were not executed
anymore, because no test framework was configured. Fix this by
configuring JUnit Jupiter as test framework.

Note that it is not necessary to declare the test framework for the
`funTest` suite because additional test suites use JUnit Jupiter by
default [1].

[1]: https://docs.gradle.org/8.0/userguide/jvm_test_suite_plugin.html#sec:declare_an_additional_test_suite

### DIFF
--- a/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
+++ b/helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
@@ -187,7 +187,6 @@ private val PATH_EXCLUDES_REASON_FOR_DIR_NAME = listOf(
 
 /** Case-sensitive glob patterns matched against filenames. */
 private val PATH_EXCLUDE_REASON_FOR_FILENAME = listOf(
-    ORT_REPO_CONFIG_FILENAME to BUILD_TOOL_OF,
     "*.bazel" to BUILD_TOOL_OF,
     "*.cmake" to BUILD_TOOL_OF,
     "*.cmakein" to BUILD_TOOL_OF,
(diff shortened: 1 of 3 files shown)


mnonnenmacher:
Unfortunately this change breaks the fun tests, we probably need to migrate the test source sets to test suites as well. I will have a look.

mnonnenmacher:
> Unfortunately this change breaks the fun tests, we probably need to migrate the test source sets to test suites as well. I will have a look.

Found the proper way to configure the test framework which should work now.

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -204,6 +203,7 @@ private val PATH_EXCLUDE_REASON_FOR_FILENAME = listOf(
     "*_test.go" to TEST_OF,
     "*coverage*.sh" to BUILD_TOOL_OF,
     ".editorconfig" to PathExcludeReason.OTHER,
+    ORT_REPO_CONFIG_FILENAME to BUILD_TOOL_OF,
```
sschuberth, line 206: FWIW, I deliberately did not use alphabetic sorting in 929b73481a4ff1a953ac5be4fa321acebb86f51e for this special entry to make it stand out.

sschuberth (approved):

sschuberth:
> Unfortunately this change breaks the fun tests

Yes, that's what I tried to express in the commit message of 33fdca96502df4f1d4b2be9159dc432a28a0335b by linking to https://github.com/gradle/gradle/issues/23544. Quite inconvenient of Gradle to require `useJUnitPlatform()` to still be added for the built-in `test` suite. Thanks for finding this.

PR #6560, opened 2023-02-26
Move package curation providers to plugin modules

mnonnenmacher:
Move the package curation providers into a new plugin module hierarchy. See the commit messages for details.

**Breaking API change:**
* The packages of package curation providers and some related classes are changed.
* The coordinates of published Maven packages are changed.


detekt-rules/src/test/kotlin/OrtPackageNamingTest.kt
```diff
@@ -0,0 +1,157 @@
...
+    // The file needs to be created in the below directory because detekt uses it to determine relative paths in tests.
+    val parent = File("build/classes/kotlin/test/$dir").safeMkdirs()
+
+    afterTest {
+        parent.safeDeleteRecursively(force = true)
+    }
+
+    val file = File(parent, "Test.kt")
+    file.writeText(content)
+    return file
```
sschuberth, line 182: Nit: Could be `return File(parent, "Test.kt").apply { writeText(content) }`

detekt-rules/src/main/kotlin/OrtPackageNaming.kt
```diff
@@ -58,7 +58,11 @@ class OrtPackageNaming(config: Config) : Rule(config) {
 
         val (pathPrefix, pathSuffix) = path.split(pathPattern, 2).map { File(it) }
         val projectDir = pathPrefix.name
-        val projectGroup = pathPrefix.parent?.let { ".$it" }.orEmpty()
+        val projectGroup = pathPrefix.parent
+            ?.let { ".$it" }
+            ?.replace("/", ".")
+            ?.replace("-", "")
```
sschuberth, line 64: Nit: Could use the `Char`-version of `replace()` by using single quotes.

detekt-rules/src/test/kotlin/OrtPackageNamingTest.kt
```diff
@@ -0,0 +1,157 @@
...
+class OrtPackageNamingTest : WordSpec({
+    val rule = OrtPackageNaming(Config.empty)
+
+    "OrtPackageNaming rule" should {
+        "succeed for package in top-level module" {
+            val file = createFile(
+                dir = "model/src/main/kotlin",
+                content = """
+                    package org.ossreviewtoolkit.model
+                """.trimIndent()
```
sschuberth, line 45: `content` is never more than a single line in any of the tests. Would it thus make sense to not use a multi-line string for it for more compact code?

plugins/package-curation-providers/clearly-defined/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+ * License-Filename: LICENSE
+ */
+
+plugins {
+    // Apply core plugins.
+    `java-library`
+}
+
+dependencies {
+    api(project(":plugins:package-curation-providers:api"))
```
sschuberth, line 26: Isn't the API that this plugin implements just an `implementation` dependency? Similar for the other plugins.

analyzer/build.gradle.kts
```diff
@@ -53,7 +53,6 @@ repositories {
 dependencies {
     api(project(":model"))
 
-    implementation(project(":clients:clearly-defined"))
```
sschuberth, line 56: I believe the dependency on the SW360 client (and the corresponding repository declaration) should be removed, too. And maybe the commit message could briefly state that these dependencies that been moved to dedicated projects in the commit before. Or even squash this commit into the previous one.

sschuberth (changes requested):

detekt-rules/src/test/kotlin/OrtPackageNamingTest.kt
mnonnenmacher: Done in #6573.

detekt-rules/src/test/kotlin/OrtPackageNamingTest.kt
mnonnenmacher: Done in #6573.

detekt-rules/src/main/kotlin/OrtPackageNaming.kt
mnonnenmacher: See comment in #6573.

plugins/package-curation-providers/clearly-defined/build.gradle.kts
mnonnenmacher: The factory implements inherit from the `PackageCurationProviderFactory` from the `api` module, shouldn't it be an api dependency then?

detekt-rules/src/main/kotlin/OrtPackageNaming.kt
sschuberth: I don't find a comment there, but by now I guess you kept it at strings as you need to replace with an empty string which is not possible with chars.

plugins/package-curation-providers/clearly-defined/build.gradle.kts
sschuberth: Looks like you are right, as the [docs](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies) mention "types used in super classes or interfaces".

analyzer/build.gradle.kts
mnonnenmacher: I have squashed it.

plugins/package-curation-providers/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+    api(project(":plugins:package-curation-providers:clearly-defined"))
+    api(project(":plugins:package-curation-providers:file"))
+    api(project(":plugins:package-curation-providers:ort-config"))
+    api(project(":plugins:package-curation-providers:sw360"))
+}
+
+configure<PublishingExtension> {
+    publications {
+        create<MavenPublication>(name) {
+            groupId = "org.ossreviewtoolkit"
```
sschuberth, line 40: Do we need the same `getGroupId()` logic here? Or should we use it anyway for consistency?

mnonnenmacher: I have hard-coded the correct value for now.

mnonnenmacher:
I have added commit to address the problem of conflicting JAR names in the distribution. Even though not required anymore, I have kept the commit to change the group ids because I like the grouping it provides.

sschuberth (approved):

PR #6563, opened 2023-02-27
Make more use of getAssetFile() and getAssetAsString() in functional tests

fviernau:
See individual commits.

### DIFF
--- a/advisor/src/funTest/kotlin/OsvFunTest.kt
+++ b/advisor/src/funTest/kotlin/OsvFunTest.kt
@@ -24,7 +24,6 @@ import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNot
 
-import java.io.File
 import java.time.Instant
 
 import org.ossreviewtoolkit.advisor.advisors.Osv
@@ -33,6 +32,7 @@ import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.config.AdvisorConfiguration
 import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.utils.test.getAssetFile
 
 class OsvFunTest : StringSpec({
     "retrievePackageFindings() returns vulnerabilities for the supported ecosystems" {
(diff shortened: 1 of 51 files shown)


sschuberth (changes requested):

advisor/src/funTest/kotlin/OsvFunTest.kt
```diff
@@ -33,6 +32,7 @@ import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.config.AdvisorConfiguration
 import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.utils.test.getAssetFile
```
sschuberth, line 35: Sorry, I missed the function names in this commit message title, which would be put into backticks as well. Will approve anyway.

sschuberth (approved):

analyzer/src/funTest/kotlin/managers/BabelFunTest.kt
```diff
@@ -22,18 +22,17 @@ package org.ossreviewtoolkit.analyzer.managers
...
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.utils.ort.normalizeVcsUrl
 import org.ossreviewtoolkit.utils.test.USER_DIR
+import org.ossreviewtoolkit.utils.test.getAssetFile
 import org.ossreviewtoolkit.utils.test.patchActualResult
 import org.ossreviewtoolkit.utils.test.patchExpectedResult
 
 class BabelFunTest : WordSpec({
-    val projectDir = File("src/funTest/assets/projects/synthetic/npm-babel").absoluteFile
+    val projectDir = getAssetFile("projects/synthetic/npm-babel").absoluteFile
```
sschuberth, line 35: Do you think it makes sense to change the implementation of `getAssetFile()` to call `absoluteFile` so we could drop these here?

sschuberth (approved):

analyzer/src/funTest/kotlin/managers/BabelFunTest.kt
fviernau: I also stumbled across this `absoluteFile` thing. I found that `projectDir` does not necessarily need to be absolute.
The need for it only comes from `definitionFile`. I started fixing that, but then decided to first eliminate code redundancy, e.g. this bit of code is in almost any funTest. I have this on my TODO list. Once the redundancy is removed, there would be no need to change the implementation for `getAssetFile()` anymore. I'd propose to go this path, if you agree it makes sense...?!

Issue #5991, opened 2022-10-24
Create an OrtResult abstraction


mnonnenmacher:
Create an abstraction for the `OrtResult` to decouple the ORT tools from the way that ORT data is stored.

Currently, we access the data stored in `OrtResult` directly in many places. This has the downside that the tools are tightly coupled to the `OrtResult` class, while the `OrtResult` class has to be optimized for serialization at the same time, because it is used to write the result files. Creating a facade (e.g. an interface of a collection of interfaces) would have several benefits:

* We could provide different means to store ORT result data, for example a local database, and it would be easier to experiment with alternative storage formats like Protobuf.
* Breaking changes in the ORT result file format would not necessarily require changes in the interface.
* Tests often require to build large parts of an `OrtResult` or even a complete `OrtResult`, when they actually only use a much smaller part of it. In those cases mocks could be used to reduce the size of the tests.

oheger-bosch:
This does not only affect the reading of result data but also the creation. One possible approach could be based on the _builder_ pattern: Instead of creating result objects directly, ORT components are passed a builder they can use to propagate their results.

sschuberth:
FYI, [there's a similar topic](https://github.com/oss-review-toolkit/ort/issues/6191) for `LicenseClassifications`. Maybe it makes sense to try out a new API approach first in the context of such a smaller / isolated class.

sschuberth:
[Here's some good read](https://www.heise.de/en/blog/Software-Development-Abstraction-is-Overrated-11198327.html) about the motivation to separate the ORT data model (as used for serialization) from the API and other usages, even if the field names / properties may be partly the same.

PR #6570, opened 2023-02-28
feat(FreeMarkerTemplateReporter): Pass `statistics` to the templates

fviernau:
This can be useful for outputting a short summary, e.g. for an ORT run on CI.
See individual commits.

### DIFF
--- a/reporter/src/main/kotlin/ReporterInput.kt
+++ b/reporter/src/main/kotlin/ReporterInput.kt
@@ -32,6 +32,7 @@ import org.ossreviewtoolkit.model.licenses.LicenseInfoResolver
 import org.ossreviewtoolkit.model.utils.DefaultResolutionProvider
 import org.ossreviewtoolkit.model.utils.PackageConfigurationProvider
 import org.ossreviewtoolkit.model.utils.ResolutionProvider
+import org.ossreviewtoolkit.reporter.StatisticsCalculator.getStatistics
 
 /**
  * A bundle of input to be used by [Reporter] implementations.
@@ -88,4 +89,10 @@ data class ReporterInput(
      * A [HowToFixTextProvider], can be used to integrate how to fix texts for [Issue]s into reports.
      */
     val howToFixTextProvider: HowToFixTextProvider = HowToFixTextProvider.NONE
-)
+) {
+    /**
+     * Statistics for [ortResult].
+     */
+    @Suppress("UNUSED") // This can be used from templates.
+    val statistics: Statistics by lazy { getStatistics(ortResult, resolutionProvider, licenseInfoResolver) }
+}
--- a/reporter/src/main/kotlin/Statistics.kt
+++ b/reporter/src/main/kotlin/Statistics.kt
@@ -17,7 +17,7 @@
  * License-Filename: LICENSE
  */
 
-package org.ossreviewtoolkit.reporter.reporters.evaluatedmodel
+package org.ossreviewtoolkit.reporter
 
 import java.util.SortedMap
 import java.util.SortedSet
--- a/reporter/src/main/kotlin/StatisticsCalculator.kt
+++ b/reporter/src/main/kotlin/StatisticsCalculator.kt
@@ -17,7 +17,7 @@
  * License-Filename: LICENSE
  */
 
-package org.ossreviewtoolkit.reporter.reporters.evaluatedmodel
+package org.ossreviewtoolkit.reporter
 
 import java.util.SortedMap
 
--- a/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
+++ b/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModel.kt
@@ -46,6 +46,7 @@ import org.ossreviewtoolkit.model.config.VulnerabilityResolution
 import org.ossreviewtoolkit.model.mapperConfig
 import org.ossreviewtoolkit.reporter.Reporter
 import org.ossreviewtoolkit.reporter.ReporterInput
+import org.ossreviewtoolkit.reporter.Statistics
 import org.ossreviewtoolkit.reporter.reporters.WebAppReporter
 
 /**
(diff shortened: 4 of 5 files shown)


reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
```diff
@@ -120,6 +121,8 @@ class FreemarkerTemplateProcessor(
             PackageModel(pkg.metadata.id, input)
         }
 
+        val statistics = with(input) { getStatistics(ortResult, resolutionProvider, licenseInfoResolver) }
```
sschuberth, line 124: I believe we must not create a dependency on the evaluated model / reporter here, esp. with having in mind that we're going to split reporters into different Gradle modules similar to https://github.com/oss-review-toolkit/ort/pull/6560.

sschuberth (changes requested):

reporter/src/main/kotlin/reporters/freemarker/FreemarkerTemplateProcessor.kt
fviernau: What do you propose as alternative?

sschuberth: > The evaluated model is currently being used by 2 reporters already

Well, counting the evaluated model reporter as an independent reporter is a bit like cheating as it originally existed only to provide the input for the web-app reporter. And I still believe we need to rewrite the web-app report with different technologies in the mid-term (to enable more people from the core team working on it, and to avoid the model duplication that we currently have). So I see the web-app (as it currently is), the evaluated model reporter, and the evaluated model to disappear at some point in time. And I'd prefer to not make that refactoring even harder by introducing more dependencies on those pieces of code now.

> What do you propose as alternative?

We already have the `AccessStatistics` class. I think it could make sense to extend that to your needs instead.

fviernau: > Well, counting the evaluated model reporter as an independent reporter is a bit like cheating as it originally existed only to provide the input for the web-app reporter. And I still believe we need to rewrite the web-app report with different technologies in the mid-term (to enable more people from the core team working on it, and to avoid the model duplication that we currently have). So I see the web-app (as it currently is), the evaluated model reporter, and the evaluated model to disappear at some point in time. And I'd prefer to not make that refactoring even harder by introducing more dependencies on those pieces of code now.

ok, just read it without making any statement about agreeing or disagreeing. I'm just a bit surprised as IIRC people were told to use the evaluated model report (instead of ORT files), as the stable outside facing interface.

> We already have the AccessStatistics class. I think it could make sense to extend that to your needs instead.

That class is used for storage stats inside the scanner. Adding in stats about summarizing ORT run like stats about the whole ORT result does not fit in there IMO. If you agree, do you have an alternative proposal?

sschuberth: > I'm just a bit surprised as IIRC people were told to use the evaluated model report (instead of ORT files), as the stable outside facing interface.

Yes, the evaluated model report is the most stable output format that we *currently* have. So *if* people desperately want to parse some sort of result file *right now*, it should be that one instead of the ORT result files. However, that does not mean (to me) that the evaluated model report is never going to disappear. IMO it's a badly designed format crafted solely after the needs of the web-app reporter. We should make it easier for people to write their own reporters instead.

> Adding in stats about summarizing ORT run like stats about the whole ORT result does not fit in there IMO.

Right. Just don't take my words too literally. What I meant was to extend the *idea* (not the class) of having statistics as a "first-class citizen" in the ORT result, and maybe add more such classes to fulfill your needs. But this certainly is something we should discuss with @oss-review-toolkit/kotlin-devs.

fviernau: > Right. Just don't take my words too literally.

I believe it's a good thing to not interpret too much but stay closely at what was written for multiple reasons.

>  Right. Just don't take my words too literally. What I meant was to extend the idea (not the class) of having statistics as a "first-class citizen" in the ORT result, and maybe add more such classes to fulfill your needs.

The problem with adding such statistics to the `OrtResult` is redundancy. I prefer not to add redundancy that file format, as that's the exact opposite direction we went for in recent years which was stripping out redundancy in many aspects as it created problems. (I'm aware it's not yet 100% right that it's redundant because the global resolutions are not yet in the `OrtResult`, but they will be as we agreed on adding them to `ResolvedConfiguration`).

Regarding the removal of the evaluated model: I believe there have been use cases where CI scripts needed the statistics and I believe these will continue to exist. So, we could remove the evaluated model but keep only the statistics.

Overall, I'm not convinced by the counter arguments provided against using the statistics in the reporter.
Is the above argumentation convincing you in any way?

sschuberth: > Is my above argumentation convincing you in any way?

It's not convincing me that depending on the evaluated model here is the right thing, no.

But here's yet another idea: In the context of https://github.com/oss-review-toolkit/ort/issues/5991 we could create an API that makes it easy for anyone (i.e. reporter implementations) to calculate the desired statistics when needed.

fviernau: > It's not convincing me that depending on the evaluated model here is the right thing, no.

As you speak of it, I believe now you're trying to phrase it in a way that it is easy to turn down.
I would propose to rephrase it as: "let's move the stats from evaluated model to the reporter because it is needed there". Moving can take place only as soon as the evaluated model is removed, because until then the usage is shared.

Can you explain why you think? (I believe that argumentation should be limited to the data it holds, I further believe
dependency on evaluated model is a non-argument as moving is considered)

1. Statistics do not fit into data model of the template reporter, or
2. Statistics do not fit into `ReporterInput` class

> But here's yet another idea: In the context of https://github.com/oss-review-toolkit/ort/issues/5991 we could create an API that makes it easy for anyone (i.e. reporter implementations) to calculate the desired statistics when needed.

That's a long term thing. I'm fine if we find a long term solution in this discussion, but it is not sufficient for me, as I need a solution implemented soon.

sschuberth (approved):

Issue #6569, opened 2023-02-28
Generalize `NoticeTemplateReporter` as `PlainTextTemplateReporter`


fviernau:
The reporter can produce any kind of plaintext file, not just notices. So, it make sense
to rename it to `PlainTextTemplateReporter`.

It is not just renaming but requires a bit of refactoring, because the file `prefix` and `extension` reporter options would not work anymore, if one still wants to produce multiple plaintext files with different prefixes or extensions in one go, like a NOTICE file and a Markdown file. 

To solve that I propose to eliminate the logic for the filename prefix and take the name fully from the template, so the template file would be named exactly after the file it produces, plus the `.ftl` extension:
- `NOTICE.ftl` produces `NOTICE`
- `MY-SUMMARY.MD.ftl` produces `MY-SUMMARY.MD`

PR #6584, opened 2023-03-01
Generalize `NoticeTemplateReporter` to `PlainTextTemplateReporter`

fviernau:
Remove any notice specifics from the reporter, and then make it clear that arbitrary plain text files can be produced via this reporter by using a more general name.

Fixes: #6569.

BREAKING CHANGE:

1. The output filenames for built-in templates changed (from -> to):
    - `NOTICE_default` -> `NOTICE_DEFAULT`
    - `NOTICE_summary` -> `NOTICE_SUMMARY`
2. The output filenames for external templates changed (from -> to), given `my-summary.txt.ftl` 
    - `NOTICE_my-summary.txt` -> `my-summary.txt`
3. The `reporter` command CLI interface changed, e.g.:
    - old: `ort report  -f NoticeTemplate -O NoticeTemplate=template.id=NOTICE_SUMMARY`
    - new: `ort report  -f PlainTextTemplate -O PlainTextTemplate=template.id=summary`


integrations/completions/ort-completion.fish
```diff
@@ -116,7 +116,7 @@ complete -c ort -f -n __fish_use_subcommand -a report -d 'Present Analyzer, Scan
 ## Options for report
 complete -c ort -n "__fish_seen_subcommand_from report" -l ort-file -s i -r -F -d 'The ORT result file to use.'
 complete -c ort -n "__fish_seen_subcommand_from report" -l output-dir -s o -r -F -d 'The output directory to store the generated reports in.'
-complete -c ort -n "__fish_seen_subcommand_from report" -l report-formats -s f -r -d 'The comma-separated reports to generate, any of [AdocTemplate, CtrlXAutomation, CycloneDx, DocBookTemplate, EvaluatedModel, Excel, FossId, GitLabLicenseModel, HtmlTemplate, ManPageTemplate, NoticeTemplate, Opossum, PdfTemplate, SpdxDocument, StaticHtml, WebApp, XHtmlTemplate].'
+complete -c ort -n "__fish_seen_subcommand_from report" -l report-formats -s f -r -d 'The comma-separated reports to generate, any of [AdocTemplate, CtrlXAutomation, CycloneDx, DocBookTemplate, EvaluatedModel, Excel, FossId, GitLabLicenseModel, HtmlTemplate, ManPageTemplate, PlainTextTemplate, Opossum, PdfTemplate, SpdxDocument, StaticHtml, WebApp, XHtmlTemplate].'
```
sschuberth, line 119: I'm surprised that the other shell completion scripts are not affected. I guess these changes were just done via search & replace? If so, I'd propose to instead re-generate the shell completion scripts in a succeeding commit.

docs/reporters/notice-templates.md
```diff
@@ -6,9 +6,9 @@ customization of the generated open source notices with [Apache Freemarker](http
 
 ORT provides two templates that can be used as a base for creating your custom open source notices:
 
-* [default](../../reporter/src/main/resources/templates/notice/NOTICE_DEFAULT.ftl): Prints a summary of all licenses 
+* [default](../../reporter/src/main/resources/templates/plain-text/NOTICE_DEFAULT.ftl): Prints a summary of all licenses
```
sschuberth, line 10: Maybe move the changes in this file to the next commit, so all docs changes are in one commit? Together with my comment about the shell scripts there would be three commits then: Code changes, completion shell regeneration, doc changes.

sschuberth (changes requested):

sschuberth:
PS: `AdocTemplateReporter` actually is a plain text reporter, too. Maybe we could get rid of it completely afterwards, @MarcelBochtler?

README.md
```diff
@@ -684,7 +684,7 @@ following formats are supported (reporter names are case-insensitive):
   * There is a [tutorial video](https://youtu.be/dNmH_kYJ34g) by @xlgmokha
 * [NOTICE](https://infra.apache.org/licensing-howto.html) file in two variants
   * List license texts and copyrights by package (`-f NoticeTemplate`)
-  * Summarize all license texts and copyrights (`-f NoticeTemplate -O NoticeTemplate=template.id=summary`)
```
MarcelBochtler, line 687: Commit message: Typo in `NoticeTemplateRepoerter`

MarcelBochtler (changes requested):

MarcelBochtler:
> PS: `AdocTemplateReporter` actually is a plain text reporter, too. Maybe we could get rid of it completely afterwards, @MarcelBochtler?

Yes, I think that make senses to so this (in a separate PR / commit).

docs/reporters/notice-templates.md
```diff
@@ -1,37 +1,38 @@
-# NOTICE Templates
+# Plain text Templates
```
sschuberth, line 1: Should this file also be renamed as part of this commit? Maybe `docs/reporters/notice-templates.md` -> `docs/reporters/plain-text-templates.md`? However, that would require several updates to places that link to `notice-templates.md`, too.

fviernau: The reason I kept it was because 99% of it talks about notice templates. So, I regarded the name still as a good fit.
If you want it to be renamed, no problem. Just let me know.

If it needs renaming, I could also do a follow up PR to get this one merged.

sschuberth: I'm also fine with keeping the file name, but I think the heading in the file and the file name should stay aligned in either case. And as you did change the heading, I was assuming you forgot the rename.

fviernau: Ok, I've reverted the heading.

sschuberth (approved):

MarcelBochtler (approved):

PR #6586, opened 2023-03-01
feat(reporter): Add "severe" issue and violation counts to `Statistics`

fviernau:
This allows accessing that value from the reporters, templates as well as from within the WebApp.

### DIFF
--- a/reporter/src/main/kotlin/Statistics.kt
+++ b/reporter/src/main/kotlin/Statistics.kt
@@ -90,7 +90,12 @@ data class IssueStatistics(
     /**
      * The number of issues with [Severity.HINT] as severity.
      */
-    val hints: Int
+    val hints: Int,
+
+    /**
+     * The number of severe issues with regard to the configured threshold.
+     */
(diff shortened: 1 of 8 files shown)


reporter/src/main/kotlin/Statistics.kt
```diff
@@ -90,7 +90,12 @@ data class IssueStatistics(
     /**
      * The number of issues with [Severity.HINT] as severity.
      */
-    val hints: Int
+    val hints: Int,
+
+    /**
+     * The number of severe issues with regard to the configured threshold.
+     */
+    val severe: Int,
```
sschuberth, line 98: Nit: Please remove the trailing comma.

sschuberth (changes requested):
PS: Looks like we're getting more an more overlap with the `SeverityStats` class from the `cli` module. Maybe we should combine both in new `model` classes? (Outside of this PR.)

sschuberth (approved):

PR #6590, opened 2023-03-01
AnalyzerConfiguration: Make properties nullable

mnonnenmacher:
Make some propertties nullable to fix merging global and local configuration. See the commit messages for details:

**Breaking API change:**
* The `allowDynamicVersions` and `skipExcluded` properties of the `AnalyzerConfiguration` are now nullable and the default value was changed from `false` to `null`.

### DIFF
--- a/model/src/main/kotlin/config/AnalyzerConfiguration.kt
+++ b/model/src/main/kotlin/config/AnalyzerConfiguration.kt
@@ -27,10 +27,10 @@ data class AnalyzerConfiguration(
      * Enable the analysis of projects that use version ranges to declare their dependencies. If set to true,
      * dependencies of exactly the same project might change with another scan done at a later time if any of the
      * (transitive) dependencies are declared using version ranges and a new version of such a dependency was
-     * published in the meantime. If set to false, analysis of projects that use version ranges will fail. Defaults to
-     * false.
+     * published in the meantime. If set to false or null, analysis of projects that use version ranges will fail.
+     * Defaults to null.
      */
-    val allowDynamicVersions: Boolean = false,
+    val allowDynamicVersions: Boolean? = null,
 
     /**
      * A list of the case-insensitive names of package managers that are enabled. Disabling a package manager in
@@ -107,7 +107,7 @@ data class AnalyzerConfiguration(
         }
 
         return AnalyzerConfiguration(
-            allowDynamicVersions = other.allowDynamicVersions,
+            allowDynamicVersions = other.allowDynamicVersions ?: allowDynamicVersions,
             enabledPackageManagers = other.enabledPackageManagers ?: enabledPackageManagers,
             disabledPackageManagers = other.disabledPackageManagers ?: disabledPackageManagers,
             packageManagers = mergedPackageManagers,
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -27,10 +27,10 @@ data class AnalyzerConfiguration(
      * Enable the analysis of projects that use version ranges to declare their dependencies. If set to true,
      * dependencies of exactly the same project might change with another scan done at a later time if any of the
      * (transitive) dependencies are declared using version ranges and a new version of such a dependency was
-     * published in the meantime. If set to false, analysis of projects that use version ranges will fail. Defaults to
-     * false.
+     * published in the meantime. If set to false or null, analysis of projects that use version ranges will fail.
+     * Defaults to null.
      */
-    val allowDynamicVersions: Boolean = false,
+    val allowDynamicVersions: Boolean? = null,
```
sschuberth, line 33: Hmm. I remember having a discussion with @fviernau a while back about how to encode default values for the configuration in general, and esp. whether we should have default values (where a default is meaningful) in the classes, or in a default configuration file that we use by default. IIRC back then @fviernau clearly voted for having the default values in classes, to ease testing and refactoring. I just wanted to bring this up.

fviernau: Interesting topic. Do I understand the problem correctly that:

"If one wants to merge two configuration, e.g.  global + local, one needs to be able to distinguish which between whether a value was explicitly set by the user or whether it came from a default value?... which is not possible so far."

If this is correct, wouldn't we then be required to find a solution for all non-mandatory config parameters?

And also, for the proposed solution one cannot override a value set by the merge partner configuration with `ǹull` which theoretically may be needed for non-primitive types.

Note: I've recently also noticed that it's not possible to de-serialize and serialize a configuration and get `input == output`. Maybe it doesn't need to be solved.

model/src/main/kotlin/config/AnalyzerConfiguration.kt
```diff
@@ -27,10 +27,10 @@ data class AnalyzerConfiguration(
      * Enable the analysis of projects that use version ranges to declare their dependencies. If set to true,
      * dependencies of exactly the same project might change with another scan done at a later time if any of the
      * (transitive) dependencies are declared using version ranges and a new version of such a dependency was
```
fviernau, line 29: Would there be a similar problem with `skipExcluded` ?

model/src/main/kotlin/config/AnalyzerConfiguration.kt
mnonnenmacher: Note that `null` is equivalent to `false` for those properties now, so the default value is still in the class, just less clear.
I'm not super happy with that change, but this works as a quick fix. Another option would be to use a separate analyzer configuration class that is used in the repository configuration and make the properties nullable only there, but that would be a bigger change.

fviernau: I'm fine with this approach as short term solution. I would propose to consider
- align also `skipExcluded`
- maybe: write down the problem in a GitHub issue, with the goal to find a general solution so this case
  is not forgotten.

model/src/main/kotlin/config/AnalyzerConfiguration.kt
sschuberth: Isn't that already handled with the last commit?

fviernau: Right, I've overlooked that one.

mnonnenmacher:
Replaced by #6608.

PR #6592, opened 2023-03-01
Allow to match Copyright garbage by `Regex` patterns

sschuberth:

### DIFF
--- a/model/src/main/kotlin/config/CopyrightGarbage.kt
+++ b/model/src/main/kotlin/config/CopyrightGarbage.kt
@@ -21,6 +21,14 @@ package org.ossreviewtoolkit.model.config
 
 import java.util.SortedSet
 
-data class CopyrightGarbage(val items: SortedSet<String> = sortedSetOf())
+/**
+ * A class that defines which Copyright statements are to be considered as garbage instead of real findings.
+ */
+data class CopyrightGarbage(
+    /**
+     * A set of literal strings that identify garbage Copyright findings.
+     */
+    val items: SortedSet<String> = sortedSetOf()
+)
 
 fun CopyrightGarbage?.orEmpty(): CopyrightGarbage = this ?: CopyrightGarbage()
--- a/model/src/main/kotlin/config/CopyrightGarbage.kt
+++ b/model/src/main/kotlin/config/CopyrightGarbage.kt
@@ -21,8 +21,6 @@ package org.ossreviewtoolkit.model.config
 
 import java.util.SortedSet
 
-data class CopyrightGarbage(val items: SortedSet<String> = sortedSetOf()) {
-    constructor(vararg items: String) : this(items.toSortedSet())
-}
+data class CopyrightGarbage(val items: SortedSet<String> = sortedSetOf())
 
 fun CopyrightGarbage?.orEmpty(): CopyrightGarbage = this ?: CopyrightGarbage()
--- a/model/src/main/kotlin/config/CopyrightGarbage.kt
+++ b/model/src/main/kotlin/config/CopyrightGarbage.kt
@@ -27,6 +27,11 @@ data class CopyrightGarbage(
      * A set of literal strings that identify garbage Copyright findings.
      */
     val items: Set<String> = emptySet()
-)
+) {
+    /**
+     * Return whether the [statement] is garbage.
+     */
+    operator fun contains(statement: String): Boolean = statement in items
+}
 
 fun CopyrightGarbage?.orEmpty(): CopyrightGarbage = this ?: CopyrightGarbage()
--- a/model/src/main/kotlin/config/CopyrightGarbage.kt
+++ b/model/src/main/kotlin/config/CopyrightGarbage.kt
@@ -19,8 +19,6 @@
 
 package org.ossreviewtoolkit.model.config
 
-import java.util.SortedSet
-
 /**
  * A class that defines which Copyright statements are to be considered as garbage instead of real findings.
  */
@@ -28,7 +26,7 @@ data class CopyrightGarbage(
     /**
      * A set of literal strings that identify garbage Copyright findings.
      */
-    val items: SortedSet<String> = sortedSetOf()
+    val items: Set<String> = emptySet()
 )
 
 fun CopyrightGarbage?.orEmpty(): CopyrightGarbage = this ?: CopyrightGarbage()
--- a/model/src/main/kotlin/config/CopyrightGarbage.kt
+++ b/model/src/main/kotlin/config/CopyrightGarbage.kt
@@ -26,12 +26,20 @@ data class CopyrightGarbage(
     /**
      * A set of literal strings that identify garbage Copyright findings.
      */
-    val items: Set<String> = emptySet()
+    val items: Set<String> = emptySet(),
+
+    /**
+     * A set of [Regex] patterns that identify garbage Copyright findings.
+     */
+    val patterns: Set<String> = emptySet()
 ) {
+    private val regexes by lazy { patterns.map { it.toRegex() } }
+
     /**
      * Return whether the [statement] is garbage.
      */
-    operator fun contains(statement: String): Boolean = statement in items
+    operator fun contains(statement: String): Boolean =
+        statement in items || regexes.any { it.matches(statement) }
 }
 
 fun CopyrightGarbage?.orEmpty(): CopyrightGarbage = this ?: CopyrightGarbage()
--- a/model/src/main/kotlin/licenses/LicenseInfoResolver.kt
+++ b/model/src/main/kotlin/licenses/LicenseInfoResolver.kt
@@ -167,12 +167,12 @@ class LicenseInfoResolver(
     private fun DetectedLicenseInfo.filterCopyrightGarbage(
         copyrightGarbageFindings: MutableMap<Provenance, Set<CopyrightFinding>>
     ): DetectedLicenseInfo {
-        val filteredFindings = findings.map {
-            val (copyrightGarbage, copyrightFindings) = it.copyrights.partition { copyrightFinding ->
+        val filteredFindings = findings.map { finding ->
+            val (copyrightGarbage, copyrightFindings) = finding.copyrights.partition { copyrightFinding ->
                 copyrightFinding.statement in copyrightGarbage.items
             }
-            copyrightGarbageFindings[it.provenance] = copyrightGarbage.toSet()
-            it.copy(copyrights = copyrightFindings.toSet())
+            copyrightGarbageFindings[finding.provenance] = copyrightGarbage.toSet()
+            finding.copy(copyrights = copyrightFindings.toSet())
         }
         return DetectedLicenseInfo(filteredFindings)
     }
--- a/model/src/main/kotlin/licenses/LicenseInfoResolver.kt
+++ b/model/src/main/kotlin/licenses/LicenseInfoResolver.kt
@@ -169,7 +169,7 @@ class LicenseInfoResolver(
     ): DetectedLicenseInfo {
         val filteredFindings = findings.map { finding ->
             val (copyrightGarbage, copyrightFindings) = finding.copyrights.partition { copyrightFinding ->
-                copyrightFinding.statement in copyrightGarbage.items
+                copyrightFinding.statement in copyrightGarbage
             }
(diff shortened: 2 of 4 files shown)


model/src/main/kotlin/licenses/LicenseInfoResolver.kt
```diff
@@ -169,7 +169,7 @@ class LicenseInfoResolver(
     ): DetectedLicenseInfo {
         val filteredFindings = findings.map { finding ->
             val (copyrightGarbage, copyrightFindings) = finding.copyrights.partition { copyrightFinding ->
-                copyrightFinding.statement in copyrightGarbage.items
+                copyrightGarbage.items.any { copyrightFinding.statement.startsWith(it) }
```
fviernau, line 172: This change does not only allow copyright garbage entries to be prefixes, but treats all copyright garbage entries as prefixes. I'm a bit uncertain whether this is too risky, but don't have a good alternative yet...

fviernau (changes requested):

model/src/main/kotlin/licenses/LicenseInfoResolver.kt
sschuberth: I remember the discussion we had a long time ago about deliberately not supporting wildcards in garbage, and I'm still fine with not supporting them (or regexes). Interpreting all garbage as prefixes was the least intrusive thing I could think of to solve the case described in https://github.com/oss-review-toolkit/ort/discussions/6580, so I thought I'd give it a try. After all, this is anyway just a temporary solution to me until we hopefully at some point can properly [curate copyrights](https://github.com/oss-review-toolkit/ort/issues/4519). And I believe it's quite unlikely to break anything.

fviernau: > After all, this is anyway just a temporary solution 

What are you trying to say with this?

> it's quite unlikely to break anything

I think it adds risk to match entries which have not been foreseen.
My gut feeling is that this risk some users wouldn't want to take without need.
Maybe this a good one for a group discussion?

sschuberth: > What are you trying to say with this?

It's written in the same sentence: In the mid-term I'm expecting the current Copyright garbage approach to be replaced with a Copyright finding curation mechanism that allow to map findings to `NONE` to make them disappear, similar to like with licenses.

> Maybe this a good one for a group discussion?

You mean in person? For a start, feel free to join the discussion at https://github.com/oss-review-toolkit/ort/discussions/6580.

PS: I'm personally completely fine with not merging this. I was just trying to help one of our users with a quick and simple fix.

fviernau: > You mean in person? For a start, feel free to join the discussion at #6580.

I meant, maybe in Ort call thursday or monday. Anyhow, I'll join discussion, seems not too urgent so give me a night to sleep about it.

sschuberth: > give me a night to sleep about it.

I've meanwhile taken the `Regex` approach that we discussed.

dgutson: > > You mean in person? For a start, feel free to join the discussion at #6580.
> 
> I meant, maybe in Ort call thursday or monday. Anyhow, I'll join discussion, seems not too urgent so give me a night to sleep about it.

Could you please brief here the outcome of the discussion?

sschuberth: @dgutson, see my reply just one comment above yours:

> I've meanwhile taken the `Regex` approach that we discussed.

dgutson: @sschuberth because I saw it, I asked :) I recall you saying that not allowing regexes was a deliberate decision, so I wanted to know the rationale behind this change.

sschuberth: > I recall you saying that not allowing regexes was a deliberate decision, so I wanted to know the rationale behind this change.

Yes, we deliberately decided to not use regexes as the primary / only way to define garbage statements. However, I'm now proposing a new section in the garbage file for regexes. That's a compromise @fviernau indicated to be fine with.

model/src/main/kotlin/config/CopyrightGarbage.kt
```diff
@@ -28,7 +30,8 @@ data class CopyrightGarbage(
     /**
      * A set of literal strings that identify garbage Copyright findings.
      */
-    val items: SortedSet<String> = sortedSetOf()
+    @JsonSerialize(converter = StringSortedSetConverter::class)
```
fviernau, line 34: I'd prefer to use this sort order: https://github.com/oss-review-toolkit/ort/blob/6e4a3f2bb7379eaadf3a9bb01ba321dba53d8dcf/helper-cli/src/main/kotlin/commands/ImportCopyrightGarbageCommand.kt#L68

This is the order you get if you manage the garbage file via helper-cli. I do recall it was deliberately choosen, IIRC it was for consistency with the Unix sort command. Maybe the git history tells it.

model/src/main/kotlin/config/CopyrightGarbage.kt
```diff
@@ -32,6 +33,11 @@ data class CopyrightGarbage(
      */
     @JsonSerialize(converter = StringSortedSetConverter::class)
     val items: Set<String> = emptySet()
-)
+) {
+    /**
+     * Return whether the statment of the given [finding] is a garbage.
+     */
+    operator fun contains(finding: CopyrightFinding): Boolean = finding.statement in items
```
fviernau: There is no need to make this depend on the `CopyrightFinding` class.
Could you please change the parameter to `String` type, for ease of use from random contexts.

model/src/main/kotlin/config/CopyrightGarbage.kt
```diff
@@ -32,12 +32,21 @@ data class CopyrightGarbage(
...
-    val items: Set<String> = emptySet()
+    val items: Set<String> = emptySet(),
+
+    /**
+     * A set of [Regex] patterns that identify garbage Copyright findings.
+     */
+    @JsonSerialize(converter = StringSortedSetConverter::class)
+    val patterns: Set<String> = emptySet()
 ) {
+    private val regex by lazy { patterns.joinToString("|", "(", ")").toRegex() }
```
fviernau, line 43: I'm not 100% sure if the string joining always yields the correct regex. It'd be great if this was reviewed by another person.

fviernau (changes requested):

model/src/main/kotlin/config/CopyrightGarbage.kt
sschuberth: So, as this seems to be the only place where the `CopyrightGarbage` class is serialized, shouldn't it be safe to simply drop the `@JsonSerialize(converter = StringSortedSetConverter::class)` annotations and keep the `items` data type at a plain `Set`?

fviernau: Hm, as soon as another place is introduced this can easily be missed. 
What do you think about using a converter which converts a `Set` to a `Set` (not to `SortedSet`) which ensures the converted set is sorted with the above `Collator`. So to say, move the sorting from that `Import...Command` to here?

sschuberth: To be honest, that case seems to be a bit too special to me to justify the introduction of such a Collator-Set-Converter. I'm fine with having this special sorting in the special `ImportCopyrightGarbageCommand`, but I wouldn't go as far as assuming that every user of `CopyrightGarbage` would want exactly that sorting.

sschuberth: > Why isn't that too special then too?

At least to me, it's much more natural to say that strings are sorted alphabetically, or packages are sorted by id, than saying that Copyright garbage items (which also are just strings) are sorted by a collator for the English locale with UTF-8 encoding using the POSIX variant...

fviernau: > I'm fine with having this special sorting in the special ImportCopyrightGarbageCommand, but I wouldn't go as far as assuming that every user of CopyrightGarbage would want exactly that sorting.

Ok, so shall we find a way to not break it?

sschuberth: > Ok, so shall we find a way to not change sorting in ImportCopyrightGarbageCommand then?

You mean to find a way for `ImportCopyrightGarbageCommand` to not use the custom `Collator`? Well, I don't care 😸 Like I said, I'm fine with any user of `CopyrightGarbage` that requires custom sorting to apply that sorting outside of `CopyrightGarbage`, like `ImportCopyrightGarbageCommand` does. But IMO the implementation of `CopyrightGarbage` should not _enforce_ a "weird" / custom sorting for all its users. So in that regard, I consider the question to be out of scope of this PR.

fviernau: I hadn't notice that the converter actually got dropped by the force push: https://github.com/oss-review-toolkit/ort/compare/8d75b58e60d4e339ec1950026b435a9355c896d1..74804278bcd5b569f0823e96a05b2a0ae5b39471.

Agree that this is out of scope then. I thought the converter was still there which would have changed the sorting of `ImportCopyrightGarbageCommand`. So, then agree it's out of scope. Sorry.

fviernau (approved):

pombredanne:
It would super useful and likely more efficient if these garbage copyright could be reported upstream as bugs to ScanCode Toolkit where these will be fixed forever and everyone ;)

fviernau:
> We had this discussion internally already @pombredanne, and I actually believe the copyright garbage feature in ORT would not really be required all that much anymore with recent version of ScanCode. If we'd see the need to update / create a garbage file, we'd report that findings against ScanCode, rest assured wink

I agree it is a good thing to fix things upstream. But not on the critical path of an compliance scan. This implies to me that ORT must provide means to make short term fixes on that critical path, so that issues can be fixed timely. While contributing back can be done when time permits. So, I disagree that this feature can be dropped as it provide means to zap out copyright statements, which is needed even though it's use becomes less and less frequent.

Regarding contributing back, @bennati would you consider contributing your ORG's garbage entries?

bennati:
Our garbage file was filled with findings by older scancode versions, would these help when used with newer scancodes?

sschuberth:
> would these help when used with newer scancodes?

Very likely not. Only those garbage findings that are still valid for the current ScanCode version would be helpful. Actually, to remove unneeded garbage entries, I personally think it makes sense to start from scratch with an empty file and add entries anew as needed.

fviernau:
> Our garbage file was filled with findings by older scancode versions, would these help when used with newer scancodes?

@pombredanne I believe this is for you. I presumed you could just put all garbage entries into a plain text file, run latest ScanCode against it and see which are considered as copyrights statements. Which would be your list of bugs then. Does this make sense?

fviernau:
> Very likely not. Only those garbage findings that are still valid for the current ScanCode version would be helpful. Actually, to remove unneeded garbage entries, I personally think it makes sense to start from scratch with an empty file and add entries anew as needed.

Maybe we shouldn't assume that all users are on latest ScanCode version. As a compromise how about limiting it to 
a min ScanCode version. E.g. add all garbage entries for ScanCode 3.x + but not for 2.x ?

sschuberth:
> Maybe we shouldn't assume that all users are on latest ScanCode version.

That's not what I'm assuming, but I *am* assuming that @pombredanne is only interested in issue reports against the latest ScanCode version; otherwise he'd first need to verify for each report if there's still something to fix with the latest ScanCode version.

fviernau:
> That's not what I'm assuming, but I am assuming that @pombredanne is only interested in issue reports against the latest ScanCode version; 

I was proposing to contribute the copyright garbage to `ort-config` so these can be used to filter out copyright garbage by any ORT user. Apart from that, as I outlined here above [1] it may be possible to filter the entries for the latest scancode version, so that the whole list can be indeed useful for @pombredanne  as input.

[1] https://github.com/oss-review-toolkit/ort/pull/6592#issuecomment-1473596956

sschuberth:
Ok guys, this discussion is getting too off-topic now from the topic of the original PR. Please consider continuing in our [discussions forum](https://github.com/oss-review-toolkit/ort/discussions).

PR #6597, opened 2023-03-02
reporter: Add execution duration to statistics

fviernau:
See individual commits.

### DIFF
--- a/reporter/src/main/kotlin/StatisticsCalculator.kt
+++ b/reporter/src/main/kotlin/StatisticsCalculator.kt
@@ -19,7 +19,11 @@
 
 package org.ossreviewtoolkit.reporter
 
+import java.time.Duration
+import java.time.Instant
 import java.util.SortedMap
+
+import kotlin.time.toKotlinDuration
 
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Severity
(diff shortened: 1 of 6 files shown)


reporter/src/main/kotlin/StatisticsCalculator.kt
```diff
@@ -158,3 +172,6 @@ internal object StatisticsCalculator {
         )
     }
 }
+
+private fun secondsBetween(startInclusive: Instant, endInclusive: Instant) =
+    java.time.Duration.between(startInclusive, endInclusive).toKotlinDuration().inWholeSeconds
```
MarcelBochtler, line 177: What is the reason to use the fully qualified name instead of an import for `java.time.Duration`?

fviernau: I'm using `Kotlin.time.Duration` as well, so the imports would clash.

MarcelBochtler: You're never directly using the type `kotlin.time.Duration`. Only indirectly through `.toKotlinDuration()`.
There is no import for `kotlin.time.Duration`.

fviernau: Oh, right. Fixed it.

MarcelBochtler (approved):

PR #6600, opened 2023-03-03
fix(OsvServiceFunTest): Update expected results

sschuberth:


fviernau (dismissed):

fviernau:
What I perceive is that most of the flakyness comes from changes under the `database_specific` and the other "custom data" property. I believe we could make the tests resilient against changes there. Do you agree it makes sense ?

sschuberth:
> I believe we could make the tests resilient against changes there. Do you agree it makes sense?

I'm fine with giving it a try to simply ignore the `database_specific` field.

fviernau (approved):

Issue #6617, opened 2023-03-06
Ensure ORT to handle the output format 3 of the ScanCode 32 series


sschuberth:
> Ok never mind, the issue is caused by scancode v32.x.x having a different output format, which ORT does not support yet.

Ping @maxhbr who I believe was experiencing something similar with the [latest ScanCode pre-release](https://github.com/nexB/scancode-toolkit/releases/tag/v32.0.0rc2). I'll create a new issue.

_Originally posted by @sschuberth in https://github.com/oss-review-toolkit/ort/issues/6585#issuecomment-1455643624_

maxhbr:
My experienced behavior:

> using a ~2 week old ORT and https://github.com/nexB/scancode-toolkit/releases/tag/v32.0.0rc2 on windows, the scan result contained copyrights, but the licenses list was always `[]`. 
> But there were no issues or problems reported and the whole pipeline just ran through silently swallowing all scanner license information.

bennati:
The issue is caused by this line https://github.com/oss-review-toolkit/ort/blob/56400c1f467c531e4977056e3db7a43a40d0d9f5/scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt#L181 which creates an empty sequence.
The reason is that the Scancode output format has changed from
```
{
      "path": "LICENSE",
      "type": "file",
      "licenses": [
        {
          "key": "here-proprietary",
          "score": 100.0,
          "name": "HERE Proprietary License",
          "short_name": "HERE Proprietary",
          "category": "Commercial",
 ```
 
 To:
```
{
      "path": "LICENSE",
      "type": "file",
      ...
      "detected_license_expression": "here-proprietary",
      "detected_license_expression_spdx": "LicenseRef-scancode-here-proprietary",
      "license_detections": [
        {
          "license_expression": "here-proprietary",
          "matches": [
            {
              "score": 100.0,
              "start_line": 4,
              "end_line": 10,
              ...
              "license_expression": "here-proprietary",
              "rule_identifier": "here-proprietary.LICENSE",
              "rule_url": "https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data/licenses/here-proprietary.LICENSE"
            }
          ]
        }
      ],
```

sschuberth:
Work for this has been started in the [scancode-format-3](https://github.com/oss-review-toolkit/ort/tree/scancode-format-3) branch.

sschuberth:
[...]

Issue #4393, opened 2021-08-25
Add more structure to the source of ORT issues


oheger-bosch:
For some use cases it is necessary to match specific issues reported during an ORT execution. For instance, with resolutions some issues should be waived, or the new notifier module should send specific notifications when certain issues are found. 

Such matching is currently difficult, however, because the `OrtIssue` class has not sufficient information to uniquely filter out specific issues. There is a single string property for the source of the issue, which - by convention - is set to the component (like analyzer, advisor, ...) that created the issue. Other than that, one typically has to match with regular expressions against the issue's message, which is fragile when there is a change in wording. Often, the message contains the coordinates of a package the issue is about.

To simplify identifying specific issues, some additional (optional) properties could be added to `OrtIssue`. The exact amount of properties is up to discussion; I would consider the following properties helpful:
* An optional `Identifier` (or a string with package coordinates), which is set if the issue is related to a specific package.
* An ORT-specific error code: This is not trivial, since it requires some sort of terminology of possible ORT errors, and means would be needed to manage this terminology and keep it consistent; but it would be very useful to make issues machine-readable and to understand what exactly went wrong.

sschuberth:
> * An optional `Identifier` (or a string with package coordinates), which is set if the issue is related to a specific package.

This could be done similar to 521640b4f55edfbfb357f65a12db480eb9521474.

PR #6619, opened 2023-03-06
feat(ScanCodeResultParser): Warn about unsupported versions

sschuberth:
While parsing newer output format versions might not throw exceptions, results may be incomplete (see #6617 for the context). Avoid confusion by warning in that case.

### DIFF
--- a/scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
+++ b/scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
@@ -33,7 +33,9 @@ import org.ossreviewtoolkit.model.Issue
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.ScanSummary
 import org.ossreviewtoolkit.model.ScannerDetails
+import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.model.TextLocation
+import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.utils.associateLicensesWithExceptions
 import org.ossreviewtoolkit.utils.common.textValueOrEmpty
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants.LICENSE_REF_PREFIX
@@ -41,6 +43,8 @@ import org.ossreviewtoolkit.utils.spdx.calculatePackageVerificationCode
 import org.ossreviewtoolkit.utils.spdx.toSpdxId
 
 import org.semver4j.Semver
+
+const val MAX_SUPPORTED_OUTPUT_FORMAT_MAJOR_VERSION = 2
 
 internal val SCANCODE_TIMESTAMP_FORMATTER: DateTimeFormatter =
     DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HHmmss.n").withZone(ZoneId.of("UTC"))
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -101,6 +105,17 @@ internal fun generateSummary(
 ): ScanSummary {
     val header = result["headers"].single()
 
+    val issues = mutableListOf<Issue>()
+    val outputFormatVersion = header["output_format_version"]?.textValue()?.let { Semver(it) }
+    if (outputFormatVersion != null && outputFormatVersion > Semver(MAX_SUPPORTED_OUTPUT_FORMAT_VERSION)) {
+        issues += ScanCode.createAndLogIssue(
+            source = ScanCode.SCANNER_NAME,
+            message = "The output format version $outputFormatVersion exceeds the supported version " +
```
fviernau, line 113: Should we add ScanCode into the message, e.g. "The **ScanCode** output format version..."?

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
```diff
@@ -42,6 +44,8 @@ import org.ossreviewtoolkit.utils.spdx.toSpdxId
 
 import org.semver4j.Semver
 
+const val MAX_SUPPORTED_OUTPUT_FORMAT_VERSION = "2.0.0"
```
fviernau, line 47: Should this remain internal?

scanner/src/test/kotlin/scanners/scancode/ScanCodeResultParserTest.kt
```diff
@@ -211,6 +216,48 @@ class ScanCodeResultParserTest : FreeSpec({
                 )
             }
         }
+
+        "for output format 3.0.0 should" - {
```
fviernau, line 220: Consider below alternative which (1) tests both boundary cases, uses (2) `MAX_SUPPORTED_OUTPUT_FORMAT_VERSION` so the test needs not be updated when 
the supported output version is changed and (3) weakens the assertion so that changes to the error message do not require updating the test.

```
"generateSummary()" should {
     "not create a warning for highest supported output version" {
     	val summary = createOutputJson(SpdxConstants.NONE,MAX_SUPPORTED_OUTPUT_FORMAT_VERSION))
        summary.issues.count { it.severity == Severity.WARNING } shouldBe 0
     }
     
     "create a warning for lowest unsupported output version" {
        val version = Semver(MAX_SUPPORTED_OUTPUT_FORMAT_VERSION).withIncPatch()
     	val summary = createOutputJson(SpdxConstants.NONE, version.toString())
     
        summary.issues.count { it.severity == Severity.WARNING } shouldBe 1
     }
}

fun createOutputJson(outputFormatVersion: String): JsonNode {
    ...
}
```

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: I believe the `source` is sufficient here.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: I was considering that, but then though programmatic users of the scanner library might want to access this.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
fviernau: The source cannot be matched  by resolutions...So, just from the message the scanner name cannot be matched. Anyhow, we also do not have this error for other scanners right now.

scanner/src/test/kotlin/scanners/scancode/ScanCodeResultParserTest.kt
sschuberth: > "not create a warning for highest supported output version"

I believe that case is sufficiently covered by other tests.

> "create a warning for lowest unsupported output version"

This actually pointed me at the fact that we should consider less-than-major changes in the output format version compatible. I'll change the implementation accordingly.

scanner/src/main/kotlin/scanners/scancode/ScanCodeResultParser.kt
sschuberth: BTW, this generally relates to https://github.com/oss-review-toolkit/ort/issues/4393.

fviernau (approved):

sschuberth:
Merging despite the unrelated `BabelFunTest` failure.

PR #6613, opened 2023-03-04
Configurable advisor plugins

sschuberth:
Please ensure that your pull request adheres to our [contribution guidelines](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md). Thank you!

PR #6622, opened 2023-03-07
fix(Osv): Make the configuration optional again

fviernau:
See individual commits.

### DIFF
--- a/advisor/src/main/kotlin/advisors/Osv.kt
+++ b/advisor/src/main/kotlin/advisors/Osv.kt
@@ -58,7 +58,7 @@ class Osv(name: String, config: OsvConfiguration) : AdviceProvider(name) {
     class Factory : AbstractAdviceProviderFactory<Osv>("OSV") {
         override fun create(config: AdvisorConfiguration) =
             // OSV does not require any dedicated configuration to be present.
-            Osv(type, config.forProvider { osv })
+            Osv(type, config.forProvider { osv ?: OsvConfiguration() })
     }
 
     override val details: AdvisorDetails = AdvisorDetails(providerName, enumSetOf(AdvisorCapability.VULNERABILITIES))
--- a/model/src/main/kotlin/config/AdvisorConfiguration.kt
+++ b/model/src/main/kotlin/config/AdvisorConfiguration.kt
@@ -50,7 +50,7 @@ data class GitHubDefectsConfiguration(
      * The access token to authenticate against the GitHub GraphQL endpoint.
      */
     @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
-    val token: String?,
+    val token: String? = null,
 
     /**
      * The URL of the GraphQL endpoint to be accessed by the service. If undefined, default is the endpoint of the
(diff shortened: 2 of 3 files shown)


model/src/main/kotlin/config/AdvisorConfiguration.kt
```diff
@@ -50,7 +50,7 @@ data class GitHubDefectsConfiguration(
      * The access token to authenticate against the GitHub GraphQL endpoint.
      */
     @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
-    val token: String?,
+    val token: String? = null,
```
sschuberth, line 53: I believe this property is not optional, and should not have been nullable to start with. Can you confirm @oheger-bosch?

model/src/main/kotlin/config/AdvisorConfiguration.kt
```diff
@@ -155,11 +155,11 @@ data class VulnerableCodeConfiguration(
...
      */
-    val serverUrl: String?,
+    val serverUrl: String? = null,
 
     /**
      * The optional API key to use.
      */
     @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
-    val apiKey: String?
+    val apiKey: String? = null
```
sschuberth, line 164: I believe the comment is outdated and this property actually is not optional anymore. @oheger-bosch, do you know if also self-hosted VulnerableCode instances require an API key by now?

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -58,7 +58,7 @@ class Osv(name: String, config: OsvConfiguration) : AdviceProvider(name) {
     class Factory : AbstractAdviceProviderFactory<Osv>("OSV") {
         override fun create(config: AdvisorConfiguration) =
             // OSV does not require any dedicated configuration to be present.
-            Osv(type, config.forProvider { osv })
+            Osv(type, config.forProvider { osv ?: OsvConfiguration() })
```
sschuberth, line 61: Sorry for breaking this. Looks like this was broken also for `OssIndex` (from where I took the code) before. Mind finding it for `OssIndex`, too?

sschuberth (changes requested):

sschuberth:
As a side-note, I believe the `AdvisorConfiguration` to be quite messy, as they have a wild mix of `null` default values (and some other place applies the actual default value) vs. having the real default value in the configuration class itself. I believe the latter should be the approach to align on, as otherwise the used config as written to the ORT result file as part of an `AdvisorRun` doe snot contain the actually used values. This is probably something to keep in mind for https://github.com/oss-review-toolkit/ort/pull/6613.

model/src/main/kotlin/config/AdvisorConfiguration.kt
fviernau: Can this be considered out of scope of this commit / PR?

advisor/src/main/kotlin/advisors/Osv.kt
fviernau: Done.

sschuberth (approved):

PR #6624, opened 2023-03-07
feat(Npm): Group related lines when parsing issues

sschuberth:
This improves the logic added in 5311dcd by grouping on arbitrary secondary prefix instead of hard-coding them. This also creates a separate issue for each deprecated packages for a better overview.

### DIFF
--- a/analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
@@ -4095,18 +4095,26 @@ analyzer:
       NPM::isarray:2.0.5:
       - timestamp: "1970-01-01T00:00:00Z"
         source: "NPM"
-        message: "The package-lock.json file was created with an old version of npm,\n\
-          so supplemental metadata must be fetched from the registry.\nThis is a one-time\
-          \ fix-up, please be patient...\nminimatch@0.3.0: Please update to minimatch\
-          \ 3.0.2 or higher to avoid a RegExp DoS issue"
+        message: "The package-lock.json file was created with an old version of npm,\
+          \ so supplemental metadata must be fetched from the registry. This is a\
+          \ one-time fix-up, please be patient..."
+        severity: "WARNING"
+      - timestamp: "1970-01-01T00:00:00Z"
+        source: "NPM"
+        message: "deprecated minimatch@0.3.0: Please update to minimatch 3.0.2 or\
+          \ higher to avoid a RegExp DoS issue"
         severity: "WARNING"
       NPM::npm-test-project:1.0.0:
       - timestamp: "1970-01-01T00:00:00Z"
         source: "NPM"
-        message: "The package-lock.json file was created with an old version of npm,\n\
-          so supplemental metadata must be fetched from the registry.\nThis is a one-time\
-          \ fix-up, please be patient...\ncoffee-script@1.12.7: CoffeeScript on NPM\
-          \ has moved to \"coffeescript\" (no hyphen)"
+        message: "The package-lock.json file was created with an old version of npm,\
+          \ so supplemental metadata must be fetched from the registry. This is a\
+          \ one-time fix-up, please be patient..."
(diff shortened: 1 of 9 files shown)


sschuberth:
PS: This PR also fixes the flakyness of the `BabelFunTest` that we current see in `main`.

analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
```diff
@@ -4095,18 +4095,26 @@ analyzer:
       NPM::isarray:2.0.5:
       - timestamp: "1970-01-01T00:00:00Z"
         source: "NPM"
-        message: "The package-lock.json file was created with an old version of npm,\n\
-          so supplemental metadata must be fetched from the registry.\nThis is a one-time\
-          \ fix-up, please be patient...\nminimatch@0.3.0: Please update to minimatch\
-          \ 3.0.2 or higher to avoid a RegExp DoS issue"
+        message: "The package-lock.json file was created with an old version of npm,\
```
fviernau, line 4098: commit-msg: Maybe the last sentence could explicitly state that this fixes flakyness?!

analyzer/src/main/kotlin/managers/Npm.kt
```diff
@@ -591,27 +592,61 @@ open class Npm(
...
+            }
+
+            collapsedLines.forEach { line ->
+                // Skip any footer as a whole.
+                if (line == "A complete log of this run can be found in:") return
+
                 issues += Issue(
                     source = managerName,
-                    message = issueLines.joinToString("\n"),
+                    message = line,
```
fviernau, line 649: Is it necessary for fixing this flakyness to split up the issue into separate ones per line?

Apart from that, would it make sense to do this split-up in a separate commit anyway?

sschuberth: > Is it necessary for fixing this flakyness to split up the issue into separate ones per line?

More or less, yes. As the order of NPM `stderr` lines might be different on each run, we can either put a defined order on all `stderr` lines (independently of whether they belong to the same error / warning or not; this is the approach that was previously taken e.g. [here](https://github.com/oss-review-toolkit/ort/pull/6624/commits/ee1cff64ad6dfd134151c40664a9d4c49dd12b00#diff-8014b93031678f0763db8fee8d0f118858b435b467390a80813754ef81c83b80L68-L76)), or we can first associate the lines to their errors / warnings, and then disregard the order of the created issues (which is the new approach in this PR).

> Apart from that, would it make sense to do this split-up in a separate commit anyway?

That's a bit of introducing a change that will be changed in the next commit anyway, but I can do that for clarity.

analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
sschuberth: I've split out the fix for the flakyness of `BabelFunTest` to a separate commit.

fviernau, line 4098: commit-msg-title:  maybe say "improve mapping of lines to issues" ,
because "Group related lines when parsing issues" to me just indicates that grouping is changed while I believe the "split-up" of issues is of similar importance.

Is "Tests need to be adjusted to be invariant against the order of issues as
different NPM versions / on different platforms the issue order may vary." now obsolete? (e.g. belongs to the split-out code no more part of this commit?)

fviernau (changes requested):

analyzer/src/funTest/assets/projects/external/git-repo-expected-output.yml
sschuberth: > to me just indicates that grouping is changed while I believe the "split-up" of issues is of similar importance.

To me, splitting into issues also is grouping, but I've changed the text anyway.

> e.g. belongs to the split-out code no more part of this commit?

Yes, thank, I've removed that paragraph.

PR #6630, opened 2023-03-08
Decouple Pub from Gradle

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/managers/Gradle.kt
+++ b/analyzer/src/main/kotlin/managers/Gradle.kt
@@ -52,6 +52,7 @@ import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.PackageManagerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.utils.DependencyGraphBuilder
@@ -70,13 +71,16 @@ private const val JAVA_MAX_HEAP_SIZE_VALUE = "8g"
 
 /**
  * The [Gradle](https://gradle.org/) package manager for Java.
+ *
+ * This package manager supports the following [options][PackageManagerConfiguration.options]:
+ * - *gradleVersion*: The version of Gradle to use when analyzing projects. Defaults to the version defined in the
+ *   Gradle wrapper properties.
  */
 class Gradle(
     name: String,
     analysisRoot: File,
     analyzerConfig: AnalyzerConfiguration,
-    repoConfig: RepositoryConfiguration,
-    private val gradleVersion: String? = null
+    repoConfig: RepositoryConfiguration
 ) : PackageManager(name, analysisRoot, analyzerConfig, repoConfig) {
     companion object : Logging
 
@@ -169,6 +173,7 @@ class Gradle(
 
         val gradleConnector = GradleConnector.newConnector()
 
+        val gradleVersion = options["gradleVersion"]
         if (gradleVersion != null) {
             gradleConnector.useGradleVersion(gradleVersion)
         }
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -394,7 +394,13 @@ class Pub(
                 "Analyzing Android dependencies for package '$packageName' using Gradle version $GRADLE_VERSION."
             }
 
-            Gradle("Gradle", androidDir, analyzerConfig, repoConfig, GRADLE_VERSION)
+            val gradleAnalyzerConfig = analyzerConfig.withPackageManagerOption(
+                "Gradle",
+                "gradleVersion",
+                GRADLE_VERSION
+            )
+
+            Gradle("Gradle", androidDir, gradleAnalyzerConfig, repoConfig)
                 .resolveDependencies(listOf(definitionFile), labels).run {
                     projectResults.getValue(definitionFile).map { result ->
                         val project = result.project.withResolvedScopes(dependencyGraph)
--- a/analyzer/src/main/kotlin/managers/Pub.kt
+++ b/analyzer/src/main/kotlin/managers/Pub.kt
@@ -118,6 +118,8 @@ class Pub(
         ) = Pub(type, analysisRoot, analyzerConfig, repoConfig)
     }
 
+    private val gradleFactory = ALL["Gradle"]
+
     private data class ParsePackagesResult(
         val packages: Map<Identifier, Package>,
         val issues: List<Issue>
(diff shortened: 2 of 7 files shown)


analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -390,11 +391,19 @@ class Pub(
         if (!definitionFile.isFile) return emptyList()
 
         return analyzerResultCacheAndroid.getOrPut(packageName) {
+            val packageManagers = analyzerConfig.packageManagers?.toMutableMap() ?: mutableMapOf()
+            val gradleConfiguration = packageManagers["Gradle"] ?: PackageManagerConfiguration()
+            val gradleOptions = gradleConfiguration.options?.toMutableMap() ?: mutableMapOf()
+            val gradleVersion = gradleOptions.getOrPut("gradleVersion") { GRADLE_VERSION }
```
fviernau: `getOrPut` is not really necessary. Consider using `gradleOptions["gradleVersion"]`.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -390,11 +391,19 @@ class Pub(
         if (!definitionFile.isFile) return emptyList()
 
         return analyzerResultCacheAndroid.getOrPut(packageName) {
+            val packageManagers = analyzerConfig.packageManagers?.toMutableMap() ?: mutableMapOf()
+            val gradleConfiguration = packageManagers["Gradle"] ?: PackageManagerConfiguration()
+            val gradleOptions = gradleConfiguration.options?.toMutableMap() ?: mutableMapOf()
```
fviernau, line 400: maybe: `gradleConfiguration.options.orEmpty().toMutableMap()`

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -390,11 +391,19 @@ class Pub(
...
+            val gradleVersion = gradleOptions.getOrPut("gradleVersion") { GRADLE_VERSION }
+
             logger.info {
-                "Analyzing Android dependencies for package '$packageName' using Gradle version $GRADLE_VERSION."
+                "Analyzing Android dependencies for package '$packageName' using Gradle version $gradleVersion."
             }
 
-            Gradle("Gradle", androidDir, analyzerConfig, repoConfig, GRADLE_VERSION)
+            packageManagers["Gradle"] = gradleConfiguration.copy(options = gradleOptions)
+            val gradleAnalyzerConfig = analyzerConfig.copy(packageManagers = packageManagers)
```
fviernau, line 408: Please extract the new logic to a function `fun AnalyzerConfiguration.setGradleVersion(String): AnalyzerConfiguration` for readability of this large `return` block.

analyzer/src/main/kotlin/managers/Gradle.kt
```diff
@@ -75,8 +75,7 @@ class Gradle(
     name: String,
     analysisRoot: File,
     analyzerConfig: AnalyzerConfiguration,
-    repoConfig: RepositoryConfiguration,
-    private val gradleVersion: String? = null
+    repoConfig: RepositoryConfiguration
```
fviernau, line 78: commit message: "This allows the Gradle version to also be
configured when creating the Gradle instance via the factory class."

It'd be nice to have a hint why this can be useful.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -115,6 +115,9 @@ class Pub(
         ) = Pub(type, analysisRoot, analyzerConfig, repoConfig)
     }
 
+    // TODO: Maybe enable Gradle functionality only if the plugin is present, and warn if it is not.
+    private val gradleFactory = ALL.getValue("Gradle")
```
fviernau, line 119: Line 119 now throws if `Gradle` is not present. Should handling for this be added already now?

fviernau (changes requested):

analyzer/src/main/kotlin/managers/Pub.kt
sschuberth: Why is it not necessary? If no `"gradleVersion"` is present yet (that should take precedence over `GRADLE_VERSION`), we need to somehow put `GRADLE_VERSION` into the map.

analyzer/src/main/kotlin/managers/Pub.kt
```diff
@@ -114,6 +115,9 @@ class Pub(
         ) = Pub(type, analysisRoot, analyzerConfig, repoConfig)
     }
 
+    // TODO: Maybe enable Gradle functionality only if the plugin is present, and warn if it is not.
```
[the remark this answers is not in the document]
sschuberth: > What's the reason to not just let the Pub plugin depend on the Gradle plugin?

It just seemed unnecessary to do to me.

> do you think there is a valid third-party use case where people want to use the Pub plugin but don't want to have the Gradle plugin on the classpath?

I don't know. I agree it's probably unlikely, but architecture-wise it seems easy to not depend on the Gradle-class hard-coded, so it felt right to not do it.

analyzer/src/main/kotlin/managers/Pub.kt
sschuberth: I chose the more generic name `withPackageManagerOption`.

analyzer/src/main/kotlin/managers/Pub.kt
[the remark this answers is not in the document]
MarcelBochtler: There is no hard requirement for Pub to have Gradle available.
This is only the case for (most?) Flutter projects, other than that, Pub is a package manager that can and is used independently.

I think it makes sense to not have a hard Gradle dependency.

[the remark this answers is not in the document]
sschuberth: > I think it makes sense to not have a hard Gradle dependency.

I've already implemented that in the most recent PR iteration.

analyzer/src/main/kotlin/managers/Pub.kt
sschuberth: Done.

fviernau (approved):

fviernau:
LGTM but I haven't wrapped my head fully around this Pub / Gradle thing, prefer you rely on other approvals.

analyzer/src/main/kotlin/managers/Pub.kt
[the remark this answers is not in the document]
mnonnenmacher: > Also, how would you actually implememnt that? Even if the Pub-PM Gradle project would depend on the Gradle-PM Gradle project, no one forces users of ORT to actually put the Gradle-PM JAR in the classpath when running ORT. So I believe we should have that case gracefully.

I'm not sure I can follow. If it's a dependency, build tools like Gradle or Maven would add that dependency to the classpath automatically so users would have to actively exclude the dependency to not have it on the classpath. But if you exclude a transitive dependency you should not be surprised if it causes issues.

> I don't know. I agree it's probably unlikely, but architecture-wise it seems easy to not depend on the Gradle-class hard-coded, so it felt right to not do it.

I don't disagree with that, but it seemed like unnecessary work to me to support an esoteric use case. However, it's done now already. But this also means that the upcoming Pub plugin should now test both scenarios, one with Gradle on the classpath and one without, because otherwise we can't claim that both are supported.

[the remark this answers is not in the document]
sschuberth: > I'm not sure I can follow.

Yeah, my thoughts weren't were clear, as a few open questions regarding plugins were still lingering through my head. For example: You're right that when currently running the `installDist` task, core plugin JARs end up in the `lib` directory. This is in contrast to third-party plugins which by convention should be put into the `plugin` directory before running ORT. I was assuming that going forward it probably makes sense to move core plugins to the `plugin` directory, too, and was thinking about the case when something does wrong there, and not all plugin JARs are available. But that's probably something to be discussed once we've finalized the move of all plugins to the plugins Gradle project.

PR #6638, opened 2023-03-09
fix(GoMod): Make ModuleInfo.goMod nullable

oheger-bosch:
According to the documentation [1] this property is optional; so ORT must not expect that it is always present.

[1] https://go.dev/ref/mod#go-list-m


sschuberth:
Maybe this quote from [here](https://go.dev/ref/mod#go-mod-file) helps:

> A go.mod file is required for the [main module](https://go.dev/ref/mod#glos-main-module), and for any [replacement module](https://go.dev/ref/mod#go-mod-file-replace) specified with a local file path. However, a module that lacks an explicit go.mod file may still be [required](https://go.dev/ref/mod#go-mod-file-require) as a dependency, or used as a replacement specified with a module path and version; see [Compatibility with non-module repositories](https://go.dev/ref/mod#non-module-compat).

oheger-bosch:
This is covered by #6643.

Issue #6615, opened 2023-03-05
[BUG] Analyzer fails with go.mod


qequ:
Trying to analyze a Go project like this https://github.com/parvez3019/go-swagger3 ort crashes with this message

```
[DefaultDispatcher-worker-1] ERROR org.ossreviewtoolkit.analyzer.PackageManager - Resolving GoMod dependencies for path 'go.mod' failed with: MissingKotlinParameterException: Instantiation of [simple type, class org.ossreviewtoolkit.analyzer.managers.ModuleInfoFile] value failed for JSON property Origin due to missing (therefore NULL) value for creator parameter origin which is a non-nullable type
 at [Source: (File); line: 1, column: 50] (through reference chain: org.ossreviewtoolkit.analyzer.managers.ModuleInfoFile["Origin"])
``` 

The Analyzer used by ort was GoMod and the command used for testing was

```
./cli/build/install/ort/bin/ort analyze -i ~/go-swagger3/ -o ~/ort_results/
```
(built from sources)

qequ:
Checking on `getModuleInfos` the line that parses to objects `ModuleInfo` found out that some modules info lack GoMod data, like for example
```
{
	"Path": "golang.org/x/term",
	"Version": "v0.0.0-20210927222741-03fcf44c2211",
	"Time": "2021-09-27T22:27:41Z",
	"Indirect": true
}
{
	"Path": "golang.org/x/text",
	"Version": "v0.3.7",
	"Time": "2021-08-10T18:28:16Z",
	"Indirect": true
}

```

but in the ModuleInfo class the GoMod  property is mandatory
```
    @JsonProperty("GoMod")
    val goMod: String
```
that's the cause of the crashing. If setting a default paramater like ".../dummy.mod"  after that it's another crash when parsing data to create the ModuleInfoFile object 
```
@JsonIgnoreProperties(ignoreUnknown = true)
private data class ModuleInfoFile(
    @JsonProperty("Origin")
    val origin: Origin
) {
    @JsonIgnoreProperties(ignoreUnknown = true)
    data class Origin(
        @JsonProperty("VCS")
        val vcs: String? = null,
        @JsonProperty("URL")
        val url: String? = null,
        @JsonProperty("Ref")
        val ref: String? = null,
        @JsonProperty("Hash")
        val hash: String? = null,
        @JsonProperty("Subdir")
        val subdir: String? = null
    )
}
```
due that it lacks an Origin

@fviernau  @sschuberth

[...]

PR #6642, opened 2023-03-09
GoMod: Fix analyzing projects without deps

fviernau:
See individual commits.

This is part of #6615. For that issue to be fixed GoMod needs to work in case of:
- unused dependencies + used dependencies
- only unused dependencies
- no dependencies at all

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-no-deps-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-no-deps-expected-output.yml
@@ -1,0 +1,23 @@
+---
+project:
+  id: "GoMod::gomod_no_deps:<REPLACE_REVISION>"
+  definition_file_path: "<REPLACE_DEFINITION_FILE_PATH>"
+  declared_licenses: []
+  declared_licenses_processed: {}
+  vcs:
+    type: "Git"
+    url: "<REPLACE_URL>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  vcs_processed:
+    type: "Git"
+    url: "<REPLACE_URL>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  homepage_url: ""
+  scopes:
+  - name: "main"
+    dependencies: []
+  - name: "vendor"
+    dependencies: []
+packages: []
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -212,7 +212,7 @@ class GoMod(
 
     private fun ModuleInfo.toId(): Identifier =
         Identifier(
-            type = managerName.takeIf { version.isBlank() } ?: "Go",
+            type = managerName.takeIf { main } ?: "Go",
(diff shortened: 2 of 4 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -212,7 +212,7 @@ class GoMod(
 
     private fun ModuleInfo.toId(): Identifier =
         Identifier(
-            type = managerName.takeIf { version.isBlank() } ?: "Go",
+            type = managerName.takeIf { main } ?: "Go",
```
sschuberth, line 216: Commit message nits:
- "determining *a* value"
- "without *a* version" (twice)
- Probably a comma before "this change is of non-functional nature"

sschuberth: Not strictly related, but how do you think about renaming the `main` and `indirect` boolean properties to `isMain` and `isIndirect` in another commit for clarity?

analyzer/src/funTest/assets/projects/synthetic/gomod-no-deps-expected-output.yml
```diff
@@ -0,0 +1,23 @@
+---
```
sschuberth, line 1: Commit message nits:
- "is zero edges" -> "*are* zero edges"

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -170,7 +170,8 @@ class GoMod(
                 moduleInfo(moduleName).toId()
             }
 
-        var graph = Graph()
+        val mainModuleId = moduleInfoForModuleName.values.single { it.main }.toId()
+        var graph = Graph(mutableMapOf(mainModuleId to setOf()))
```
sschuberth, line 174: Use `emptySet()`.

sschuberth (dismissed):

sschuberth:
How does this relate to https://github.com/oss-review-toolkit/ort/pull/6638? Is it an alternative / replacement, and does it fix the same issue? (No issue is being reference in the commit messages.)

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: The names are exactly the same as in `golang` upstream data structures. I find it a tad simpler if these are consistent, which is why I'd prefer to keep these names.

sschuberth: > The names are exactly the same as in `golang` upstream data structures.

That's not true, as they already differ in capitalization.

fviernau:
> How does this relate to #6638? Is it an alternative / replacement, and does it fix the same issue? (No issue is being reference in the commit messages.)

PR #6638 is in context of #6615 and this PR is also in context of #6615.
I'm planning to fix #6615, as I already put quite a bit effort into it. So, I'll file another PR
which might replace #6638.

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: I think you got my point.

oheger-bosch (approved):

PR #6643, opened 2023-03-09
GoMod: Fix analyzing projects with unused dependencies

fviernau:
See individual commits.

**Note: Must be merged after** #6642.

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-unused-deps-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-unused-deps-expected-output.yml
@@ -1,0 +1,23 @@
+---
+project:
+  id: "GoMod::gomod_unused_deps:<REPLACE_REVISION>"
+  definition_file_path: "<REPLACE_DEFINITION_FILE_PATH>"
+  declared_licenses: []
+  declared_licenses_processed: {}
+  vcs:
+    type: "Git"
+    url: "<REPLACE_URL>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  vcs_processed:
+    type: "Git"
+    url: "<REPLACE_URL>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  homepage_url: ""
+  scopes:
+  - name: "main"
+    dependencies: []
+  - name: "vendor"
+    dependencies: []
+packages: []
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -516,7 +516,8 @@ private data class ModuleInfoFile(
 }
 
 private fun ModuleInfo.toVcsInfo(): VcsInfo? {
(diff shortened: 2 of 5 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -516,7 +516,8 @@ private data class ModuleInfoFile(
 }
 
 private fun ModuleInfo.toVcsInfo(): VcsInfo? {
-    val info = jsonMapper.readValue<ModuleInfoFile>(File(goMod).resolveSibling("$version.info"))
+    val infoFile = File(goMod).resolveSibling("$version.info")
```
sschuberth: Conventional commit nit: I guess the more correct type would be "refactor".

sschuberth (commented):
I'll leave the approval to @oheger-bosch who proposed the alternative PR #6638.

analyzer/src/funTest/assets/projects/synthetic/gomod-unused-deps-expected-output.yml
```diff
@@ -0,0 +1,23 @@
+---
```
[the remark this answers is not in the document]
fviernau: Making this unpersonal would IMO change the meaning. What do you suggest exactly?

oheger-bosch (dismissed):
Changes look good to me. @MarcelBochtler, is there any chance you can test this against our problematic project? Otherwise, I am fine with merging this.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -484,7 +484,7 @@ private data class ModuleInfo(
     val main: Boolean = false,
 
     @JsonProperty("GoMod")
-    val goMod: String
+    val goMod: String?
```
sschuberth, line 487: Just checking: In @oheger-bosch's [proposal](https://github.com/oss-review-toolkit/ort/pull/6638/files#diff-d0e6f488a95a59913828fe32b76dae1b1ad197ddac08af8485ffa87578a0ae79R489) this also got a default value of `null`. Is this intentionally *not* done here so somehow express that `goMod` being `null` is rather an exceptional case than the default?

fviernau: This class is only construction in context of deserialization. There is no constructor calls in the code. So, a default null value is not needed.

sschuberth: Now I'm confused: Isn't the point that the `goMod` property might be missing from the JSON to deserialize, hence we'd need a default value here?

sschuberth: Ah, seems like so. I've never made that really clear to me, it seems.

Given that in all other places where data class properties are nullable a default value of `null` *is* assigned, how about also aligning this as part of this PR? (In either way.)

sschuberth (approved):

MarcelBochtler (approved):

fviernau:
Merged as the only failing test (ClearlyDefinedStorageFunTest) was unrelated.

PR #6645, opened 2023-03-09
Start to move package managers to the `plugins` project

sschuberth:


gradle.properties
```diff
@@ -16,7 +16,7 @@
 # License-Filename: LICENSE
 
 org.gradle.caching = true
-org.gradle.jvmargs = -Xmx512m -XX:MaxMetaspaceSize=256m -Dfile.encoding=UTF-8
+org.gradle.jvmargs = -Xmx1024m -XX:MaxMetaspaceSize=512m -Dfile.encoding=UTF-8
```
mnonnenmacher, line 19: Commit message nit: "restructurings"
Just FYI, I had metaspace issues several times last week, but didn't change the settings yet because I couldn't reproduce it anymore when trying to investigate. So it's not entirely caused by adding more modules.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #6646, opened 2023-03-09
refactor(GoMod): Extract the `Graph` class and improve it a bit

fviernau:
The `GoMod` class was quite large and extracting the `Graph` class to a separate file removes a bit of distraction for the reader.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -168,7 +168,7 @@ class GoMod(
             }
 
         val mainModuleId = moduleInfoForModuleName.values.single { it.main }.toId()
-        var graph = Graph(mutableMapOf(mainModuleId to emptySet()))
+        var graph = Graph().apply { addNode(mainModuleId) }
 
         val edges = runGo("mod", "graph", workingDir = projectDir)
 
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -197,9 +197,9 @@ class GoMod(
         }.toMap()
 
         val vendorModules = getVendorModules(graph, projectDir, replacedModules)
-        if (vendorModules.size < graph.size()) {
+        if (vendorModules.size < graph.size) {
             logger.debug {
-                "Removing ${graph.size() - vendorModules.size} non-vendor modules from the dependency graph."
+                "Removing ${graph.size - vendorModules.size} non-vendor modules from the dependency graph."
             }
 
             graph = graph.subgraph(vendorModules)
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -372,19 +372,6 @@ private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>>
             }
         )
 
-    /**
-     * Search for the single package that represents the main project. This is the only package without a version. Fail
-     * if no single package with this criterion can be found.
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -338,9 +338,12 @@ class GoMod(
  * A class to represent a graph with dependencies. This representation is basically an adjacency list implemented by a
  * map whose keys are package identifiers and whose values are the identifiers of packages these packages depend on.
  */
-private class Graph(private val nodeMap: MutableMap<Identifier, Set<Identifier>> = mutableMapOf()) {
+private class Graph(nodeMap: MutableMap<Identifier, Set<Identifier>> = mutableMapOf()) {
     companion object : Logging
 
+    private val nodeMap: MutableMap<Identifier, MutableSet<Identifier>> =
+        nodeMap.mapValuesTo(mutableMapOf()) { it.value.toMutableSet() }
```
sschuberth, line 345: How about using `MutableSet<Identifier>` in the constructor instead? All but one caller already use a `MutableSet` anyway, so you only need to adjust the call in `breakCycles()` from

    return Graph(outgoingEdgesForNodes.mapValuesTo(mutableMapOf()) { it.value.toSet() })

to

    return Graph(outgoingEdgesForNodes.mapValuesTo(mutableMapOf()) { it.value.toMutableSet() })

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -515,6 +502,19 @@ private data class ModuleInfoFile(
     )
 }
 
+/**
```
sschuberth, line 393: > Make the `Graph` agnostic of `Go`

Can you explain this a bit further in the commit message? At a first glance, there seems to be nothing Go-specific in that function, and the word "Go" isn't mentioned. My guess is that it's the *assumption* being made in the function that's Go-specific, i.e. that there's only one package without a version, and that that's the main package.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -168,7 +168,7 @@ class GoMod(
             }
 
         val mainModuleId = moduleInfoForModuleName.values.single { it.main }.toId()
-        var graph = Graph(mutableMapOf(mainModuleId to emptySet()))
+        var graph = Graph().apply { addNode(mainModuleId) }
```
sschuberth, line 171: Commit message: "an" -> "and"

sschuberth (changes requested):

analyzer/src/main/kotlin/managers/GoMod.kt
sschuberth: And even if the code is Go-specific, as the whole Graph still is Go-specific (Only used in Go-context, and that will probably not change, or?), why not just leave it there?

fviernau: > why not just leave it there?

The file is very large. I find it easier to work with the code if this is in a separate file.

sschuberth: > If you mean the function: To me the logic belong more closely to `GoMod` than to that `Graph`. IMO a `Graph` is something everybody knows what it is about, if there are no built-in surprises like this.

Yes, I meant the function. As long as `Graph` is not useful / being used in another context than Go, I don't see a reason to extract it. But now that it already is, let's leave it like that.

sschuberth (approved):

fviernau:
Merge regardless of the failing ClearlyDefined flaky test.

PR #6648, opened 2023-03-10
FossID: Change the delta scan matching logic

nnobelis:
Please refer to individual commits messages.

### DIFF
--- a/cli/src/main/kotlin/commands/ScannerCommand.kt
+++ b/cli/src/main/kotlin/commands/ScannerCommand.kt
@@ -230,7 +230,7 @@ class ScannerCommand : OrtCommand(
 
             val ortResult = readOrtResult(input)
             return runBlocking {
-                scanner.scan(ortResult, skipExcluded)
+                scanner.scan(ortResult, skipExcluded, labels)
             }
         } finally {
             runBlocking { workingTreeCache.shutdown() }
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -112,6 +112,9 @@ class FossId internal constructor(
         @JvmStatic
         internal val SERVER_URL_KEY = "serverurl"
 
+        @JvmStatic
+        internal val PROJECT_REVISION_LABEL = "projectVcsRevision"
+
         /**
          * The scan states for which a scan can be triggered.
          */
@@ -265,7 +268,7 @@ class FossId internal constructor(
                     checkNotNull(scans)
 
                     val (scanCode, scanId) = if (config.deltaScans) {
-                        checkAndCreateDeltaScan(scans, url, revision, projectCode, projectName)
+                        checkAndCreateDeltaScan(scans, url, revision, projectCode, projectName, context)
                     } else {
                         checkAndCreateScan(scans, url, revision, projectCode, projectName)
                     }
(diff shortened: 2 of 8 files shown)


cli/src/main/kotlin/commands/ScannerCommand.kt
```diff
@@ -230,7 +230,7 @@ class ScannerCommand : OrtCommand(
 
             val ortResult = readOrtResult(input)
             return runBlocking {
-                scanner.scan(ortResult, skipExcluded)
+                scanner.scan(ortResult, skipExcluded, labels)
```
sschuberth, line 233: Commit message nits:
- Make "after the scanner has run" the last part of the sentence.
- "by CLI parameter" -> "on the command line"

scanner/build.gradle.kts
```diff
@@ -53,6 +53,7 @@ dependencies {
     implementation(libs.bundles.exposed)
     implementation(libs.hikari)
     implementation(libs.jacksonModuleKotlin)
+    implementation(libs.jgit)
```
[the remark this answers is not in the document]
sschuberth: JGit does not seem to be directly used in this commit. Is this a left-over?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -353,19 +353,38 @@ class FossId internal constructor(
...
+        projectRevision: String? = null,
+        defaultBranch: String? = null,
+    ): List<Scan> {
+        val scans = filter {
+            val isArchived = it.isArchived ?: false
+            // The scans in the server contain the url with the credentials, so we have to remove it for the
+            // comparison. If we don't, the scans won't be matched if the password changes!
+            val urlWithoutCredentials = it.gitRepoUrl?.replaceCredentialsInUri()
+            !isArchived && urlWithoutCredentials == url
+        }.sortedByDescending { scan -> scan.id }
```
sschuberth, line 373: Could use just `it`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -353,19 +353,38 @@ class FossId internal constructor(
...
+        return scans.filter { scan -> projectRevision == scan.comment }.ifEmpty {
```
sschuberth, line 374: Please add an empty line before this one.

sschuberth (dismissed):

scanner/build.gradle.kts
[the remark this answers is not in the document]
nnobelis: It was a leftover.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -418,6 +419,17 @@ class FossId internal constructor(
     ): Pair<String, String> {
         val projectRevision = context.labels[PROJECT_REVISION_LABEL]
 
+        val vcs = requireNotNull(VersionControlSystem.forUrl(url))
```
oheger-bosch, line 442: IIRC, there is some URL-mapping support for the repository URLs. Does this need to be taken into account here?

nnobelis: Also, the URL provider kicks in just before creating the scan.

It is mainly used for two things: add credentials, rewrite some part of the URL (e.g. changing the scheme).

For the credentials, this does not need to be done for the `VersionControlSystem` as it comes to it's own delegation to the Authenticator:
see https://github.com/oss-review-toolkit/ort/blob/c4a7a749ad214137bedfd7410c873596abb614f6/downloader/src/main/kotlin/vcs/Git.kt#L70.

However for the rewriting of the URL, I am not sure anymore. If you change the scheme, it could be because the given scheme is unsupported.

But how would you solve this issue ? We cannot call the URL provider because we don't want the credentials being added to the URL (see above).

What is your opinion ?

nnobelis: I retested the changes with a pipeline using the URL scheme rewriting and it ran without error.

oheger-bosch (approved):

PR #6650, opened 2023-03-10
Remove the `AdocTemplateReporter`

### LINKED ISSUE #6595: Remove the `AdocTemplateReporter`
The `PlainTextTemplateReporter` can be used as replacement.

1. The templates should be kept
2. `docs/reporters/notice-templates.md` should be renamed to `docs/reporters/plain-text-templates.md`
3. a.m. file should be extended to explain the `.adoc` use case, along with an example CLI call

fviernau:
The `AdocTemplateReporter` basically generated `.adoc` files using plain text templates.
It seems better to use  the `PlainTextTemplateReporter` as replacement, in particular as the former was a bit
confusing in terms of the implementation details.

Fixes #6595.

BREAKING CHANGE:
- Instead of using `AdocTemplateReporter` use `PlainTextTemplateReporter` as described in doc updates of this PR.


docs/reporters/plain-text-templates.md
```diff
@@ -2,7 +2,7 @@
 
 The [`PlainTextTemplateReporter`](../../reporter/src/main/kotlin/reporters/freemarker/PlainTextTemplateReporter.kt)
 enables customization of the generated open source notices with [Apache Freemarker](https://freemarker.apache.org/)
-templates and producing any other arbitrary plain text files.
+templates and producing any other arbitrary plain text files, such as `.adoc` files.
```
sschuberth, line 5: I was missing these changes when reviewing the first commit. Please squash to that commit. That also avoids the `fix`-type in the commit title, which is a bit awkward, as you're "fixing" something that you "broke" in the same PR.

sschuberth: And in case you're going to ask whether the completion-script re-generation should then be squashed, too: No, I regard that as a different thing, as it's committing generated files vs. hand-made changes.

sschuberth (changes requested):

sschuberth (approved):

sschuberth (changes requested):
Ah, sorry, I just saw that `docs/reporters/asciidoc-templates.md` also needs to be updated.

sschuberth (approved):

PR #6655, opened 2023-03-10
Fix several proxy configuration errors in Docker

Iggle:
- Rename proxy $USER variable to prevent collisions with system environment
- Remove trailing slashes from proxy port variables

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -290,6 +290,7 @@ RUN --mount=type=tmpfs,target=/android \
     && curl -Os https://dl.google.com/android/repository/commandlinetools-linux-${ANDROID_CMD_VERSION}_latest.zip \
     && unzip -q commandlinetools-linux-${ANDROID_CMD_VERSION}_latest.zip -d $ANDROID_HOME \
     && PROXY_HOST_AND_PORT=${https_proxy#*://} \
+    && PROXY_HOST_AND_PORT=${PROXY_HOST_AND_PORT%/} \
     && if [ -n "$PROXY_HOST_AND_PORT" ]; then \
         # While sdkmanager uses HTTPS by default, the proxy type is still called "http".
         SDK_MANAGER_PROXY_OPTIONS="--proxy=http --proxy_host=${PROXY_HOST_AND_PORT%:*} --proxy_port=${PROXY_HOST_AND_PORT##*:}"; \
--- a/scripts/set_gradle_proxy.sh
+++ b/scripts/set_gradle_proxy.sh
@@ -38,6 +38,7 @@ writeProxyStringToGradleProps () {
     fi
 
     local PORT=${PROXY##*:}
+    PORT=${PORT/%\//}
     [ "$PORT" -ge 0 ] 2>/dev/null || PORT=80
 
     grep -qF "systemProp.$PROTOCOL.proxy" $FILE 2>/dev/null && return 1
--- a/scripts/set_gradle_proxy.sh
+++ b/scripts/set_gradle_proxy.sh
@@ -32,9 +32,9 @@ writeProxyStringToGradleProps () {
         # Strip authentication info.
         HOST=${HOST#$AUTH@}
         # Extract the user.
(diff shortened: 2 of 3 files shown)


scripts/set_gradle_proxy.sh
```diff
@@ -32,12 +32,14 @@ writeProxyStringToGradleProps () {
         # Strip authentication info.
         HOST=${HOST#$AUTH@}
         # Extract the user.
-        local USER=${AUTH%%:*}
+        local PROXY_USER=${AUTH%%:*}
         # Extract the password.
-        local PASSWORD=${AUTH#*:}
+        local PROXY_PASSWORD=${AUTH#*:}
```
sschuberth, line 37: I'm curious: As these are local variables, I though there could be no conflict with a global `USER` environment variable. In the message for this commit, could you please go a bit into the details of the problem you were facing with the original code?

Dockerfile
```diff
@@ -289,7 +289,7 @@ RUN --mount=type=tmpfs,target=/android \
     cd /android \
     && curl -Os https://dl.google.com/android/repository/commandlinetools-linux-${ANDROID_CMD_VERSION}_latest.zip \
     && unzip -q commandlinetools-linux-${ANDROID_CMD_VERSION}_latest.zip -d $ANDROID_HOME \
-    && PROXY_HOST_AND_PORT=${https_proxy#*://} \
+    && PROXY_HOST_AND_PORT=${https_proxy#*://} PROXY_HOST_AND_PORT=${PROXY_HOST_AND_PORT%/}\
```
sschuberth, line 292: Can we please put this into separate lines for readability, and apply the same change to `docker/legacy/Dockerfile`, too?

Iggle: Gladly

scripts/set_gradle_proxy.sh
```diff
@@ -38,6 +38,8 @@ writeProxyStringToGradleProps () {
     fi
 
     local PORT=${PROXY##*:}
+    # Remove trailing slashes
+    PORT=${PORT/%\//}
```
sschuberth, line 42: One more style nit (Did I say yet we're picky?): We like to group comments with belonging code by adding empty lines before the comment and after the belonging code. Here, however, that would add a lot of empty lines with relation to the length of the code block. So personally, I'd either put the comment after the line like `PORT=${PORT/%\//} # Remove trailing slashes.` or drop it altogether as we also don't have comments in similar code above, and you sufficiently explained it in the commit message. But *if* you keep the comment, please end it with a dot, like a proper sentence written in a book.

sschuberth: Also, strictly speaking this would only remove a *single* trailing slash, not multiple ones, so the comment should say `# Remove any single trailing slash.`.

scripts/set_gradle_proxy.sh
sschuberth: Commit message nits:
- In the body, just start the second sentence in the first line.
- http -> HTTPS
- gradle -> Gradle

sschuberth (changes requested):

scripts/set_gradle_proxy.sh
Iggle: You're not wrong, I've removed the comment

scripts/set_gradle_proxy.sh
```diff
@@ -38,6 +38,7 @@ writeProxyStringToGradleProps () {
     fi
 
     local PORT=${PROXY##*:}
+    PORT=${PORT/%\//}
```
sschuberth, line 41: Commit message nit: It still talks about plural "slashes" there.

sschuberth (approved):
Approving despite my one remaining nit.

scripts/set_gradle_proxy.sh
Iggle: Oups

sschuberth (approved):

PR #6657, opened 2023-03-10
fix(SpdxExpression): Fix valid license choice for mixed expressions

MarcelBochtler:


utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
```diff
@@ -472,6 +472,13 @@ class SpdxExpressionTest : WordSpec() {
                 spdxExpression.isValidChoice("b AND c AND e".toSpdx()) shouldBe true
             }
 
+            "return true if a choice is valid in a mixed expression" {
+                val spdxExpression = "a AND (a OR b)".toSpdx()
+
+                spdxExpression.isValidChoice("a".toSpdx()) shouldBe true
+                spdxExpression.isValidChoice("a AND b".toSpdx()) shouldBe true
```
sschuberth, line 479: Should this better be made a `"validChoices()"` test (using `containExactlyInAnyOrder`) to also verify these are the *only* valid choices?

sschuberth: You also need to move the test to the `"validChoices()" should` section.

sschuberth (approved):

PR #6658, opened 2023-03-10
refactor(analyzer): Make all constants for option key names public

sschuberth:
These are useful for programmatic use of the package managers.

Where it is offered, also slightly rearrange related code blocks of constants to they are at the top of the file, with public ones going first.

### DIFF
--- a/analyzer/src/main/kotlin/managers/Bundler.kt
+++ b/analyzer/src/main/kotlin/managers/Bundler.kt
@@ -63,6 +63,11 @@ import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.showStackTrace
 
 /**
+ * The name of the option to specify the Bundler version.
+ */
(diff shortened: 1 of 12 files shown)


analyzer/src/main/kotlin/managers/Bundler.kt
```diff
@@ -62,6 +62,11 @@ import org.ossreviewtoolkit.utils.ort.HttpDownloadError
 import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.showStackTrace
 
+/**
```
fviernau, line 65: Have you considered putting them into the companion object?

fviernau (dismissed):

analyzer/src/main/kotlin/managers/Bundler.kt
sschuberth: Ah, right, that should be done to avoid name clashes. I'll add another commit.

fviernau (approved):

PR #6662, opened 2023-03-11
Gradle test configuration improvements

sschuberth:

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -191,9 +191,15 @@ subprojects {
 
     testing {
         suites {
-            @Suppress("UnusedPrivateMember")
-            val test by getting(JvmTestSuite::class) {
+            withType<JvmTestSuite>().configureEach {
                 useJUnitJupiter()
+
+                dependencies {
+                    implementation(project(":utils:test-utils"))
+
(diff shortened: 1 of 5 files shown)


build.gradle.kts
```diff
@@ -197,6 +197,8 @@ subprojects {
             }
 
             register<JvmTestSuite>("funTest") {
+                configurations["funTestImplementation"].extendsFrom(configurations["testImplementation"])
```
fviernau, line 207: Is this inheritance in terms of dependencies, e.g. funTestImpl inherits deps of testImpl?

sschuberth: [Yes](https://docs.gradle.org/current/userguide/declaring_dependencies.html#sub:config-inheritance-composition).

fviernau: Hm, I was unaware of that inheritance. I believe that both, `´test` and `funTest` do have dependencies which should not be shared. 

Would it make sense to share code by having 3 projects like:

1. common-test-utils
2. test-utils
3. fun-test-utils

sschuberth: I've found a yet better way to share (only the basic) dependencies across test suites, please have a look.

PR #6665, opened 2023-03-13
ScanCode: Set a lower bound for the required version

fviernau:
Support for ScanCode version 2 has been dropped a while ago [1]. So, set a lower bound for clarity.

[1] https://github.com/oss-review-toolkit/ort/commit/4663b01813494a716a767f96af89996f76b4be24

### DIFF
--- a/scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
+++ b/scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
@@ -43,6 +43,9 @@ import org.ossreviewtoolkit.utils.common.splitOnWhitespace
 import org.ossreviewtoolkit.utils.common.withoutPrefix
 import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 
+import org.semver4j.RangesList
+import org.semver4j.RangesListFactory
+
 /**
  * A wrapper for [ScanCode](https://github.com/nexB/scancode-toolkit).
  *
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
```diff
@@ -124,6 +127,8 @@ class ScanCode internal constructor(
     override fun command(workingDir: File?) =
         listOfNotNull(workingDir, if (Os.isWindows) "scancode.bat" else "scancode").joinToString(File.separator)
 
+    override fun getVersionRequirement(): RangesList = RangesListFactory.create("[3.0.0,)")
```
sschuberth, line 130: Nit: We could simplify this to `">=3.0.0"`, but that's true for several existing ranges as well.

sschuberth (approved):

scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
fviernau: I'll make a follow-up PR.

sschuberth (approved):

scanner/src/main/kotlin/scanners/scancode/ScanCode.kt
fviernau: @sschuberth see https://github.com/oss-review-toolkit/ort/pull/6673

PR #6685, opened 2023-03-16
Add support  for some new calls to the FossID API client

nnobelis:

Theses calls will be used in the future to add snippets information to ORT data model.

See also https://github.com/oss-review-toolkit/ort/issues/3265.


clients/fossid-webapp/src/main/kotlin/Extensions.kt
sschuberth, line 1: Commit message nits:
- "Fossid-webapp" -> "fossid-webapp" (as it refers to the lower-case directory name)

clients/fossid-webapp/src/main/kotlin/Extensions.kt
```diff
@@ -22,18 +22,23 @@
 package org.ossreviewtoolkit.clients.fossid
 
 import java.io.File
+import java.util.Base64
```
sschuberth, line 25: Note to myself: With Kotlin 1.8.20 we could use the [experimental Kotlin-native Base64 implementation](https://kotlinlang.org/docs/whatsnew-eap.html#experimental-support-for-base64-encoding-in-standard-library).

clients/fossid-webapp/src/main/kotlin/model/result/Snippet.kt
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (See <https://github.com/oss-review-toolkit/ort-server/blob/main/NOTICE>)
```
sschuberth, line 2: "See" -> "see"

sschuberth: And the URL needs to be https://github.com/oss-review-toolkit/ort/blob/main/NOTICE.

clients/fossid-webapp/src/main/kotlin/model/result/Snippet.kt
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.clients.fossid.model.result
+
+data class Snippet(
```
sschuberth, line 22: Are there some docs available for this class and its properties? Is really almost everything in this class optional, i.e. you cannot rely on the presence of properties?

clients/fossid-webapp/src/test/kotlin/FossIdClientReturnTypeTest.kt
```diff
@@ -122,6 +124,23 @@ class FossIdClientReturnTypeTest : StringSpec({
         }
     }
 
+    "Snippets can be listed when there is some" {
```
sschuberth, line 137: Maybe shorter: "Existing snippets can be listed".

sschuberth: But I see other tests should then be reworded accordingly as well.

clients/fossid-webapp/src/main/kotlin/model/result/MatchedLines.kt
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
```
sschuberth, line 20: This needs to be moved above the `package`.

sschuberth (changes requested):

clients/fossid-webapp/src/test/kotlin/FossIdClientReturnTypeTest.kt
nnobelis: Can we leave them as if ? I don't intend to reword all tests now.

clients/fossid-webapp/src/main/kotlin/model/result/Snippet.kt
nnobelis: No unfortunately as always with FossID, there is no documentation of the model available.

Maybe we can make the assumption `id`, `matchType`, `fieIdId` and `scanId` are not null, for the rest I don't know.

Please view https://github.com/oss-review-toolkit/ort/blob/34ca6c65da51415bfdf09c06223867fb140d2d32/clients/fossid-webapp/src/test/assets/return-type/mappings/apiphp-list_scan_snippets.json for example of the response.

sschuberth: > Maybe we can make the assumption `id`, `matchType`, `fieIdId` and `scanId` are not null

Yes, please make as many properties non-nullable as you confidently can, to simplify any upcoming code that makes use of these properties.

nnobelis: Ok I queried the snippets of several files of the Semver4j project and I make some changes to the class.

sschuberth:
Can you please rebase this onto latest `main` once?

sschuberth (approved):

PR #6692, opened 2023-03-16
FossID: increase the `deleteScan` call read timeout

nnobelis:


clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
sschuberth, line 1: > However, the timeout needs to be made adjustable for some specific
requests.

Could you explain *why* that is the case? I.e. what would be the hard in setting the max. needed timeout for all requests?

nnobelis: I am surprised by your question. Putting it to the absurd, why are we not setting a 5 minutes timeout for all requests sent by ORT ? Why don't all our softwares in everyday life have a timeout of 5 minutes ?

The answer would be responsiveness: if there is a network sluggishness (not outage because we talk about the _read_ timeout), you don't want your software crawling as much as it can, you can to be notified ASAP.
IMO, it just seem to be good practice to increase the timeout only when it need to be.

Please note that we encounter these long delays with the delete scan operations with a 1.5 GB monorepo with around 50000 pending files. Nothing say FossID performances won't degrade exponentially with larger repositories and we have to increase the timeout further.

sschuberth: > Putting it to the absurd, why are we not setting a 5 minutes timeout for all requests sent by ORT ?

Because that would affect all of ORT and not just the FossID integration which, as mentioned, already has a minimum timeout of 1 min. So it's probably *already* sluggish, and making it more sluggish probably does not weight much.

> IMO, it just seem to be good practice to increase the timeout only when it need to be.

I agree here, and that's what I was after: With the feature you're introducing it seems to make sense to remove the original `val minReadTimeout = Duration.ofSeconds(60)` code completely, and exclusively use the header-based approach per endpoint. Would that make sense to you?

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -58,8 +60,13 @@ import retrofit2.Retrofit
 import retrofit2.converter.jackson.JacksonConverterFactory
 import retrofit2.http.Body
 import retrofit2.http.GET
+import retrofit2.http.Headers
 import retrofit2.http.POST
 
+private const val READ_TIMEOUT_HEADER = "READ_TIMEOUT"
+private const val CONNECT_TIMEOUT_HEADER = "CONNECT_TIMEOUT"
+private const val WRITE_TIMEOUT_HEADER = "WRITE_TIMEOUT"
+
```
fviernau, line 69: IIUC you have the intention to only change the READ_TIMEOUT with this PR.
Put the implementation has logic also for changing CONNECT_TIMEOUT, which seems unused even after the following commit. Would it make sense to remove that?

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -227,6 +227,7 @@ interface FossIdRestService {
     suspend fun runScan(@Body body: PostRequestBody): EntityResponseBody<Nothing>
 
     @POST("api.php")
+    @Headers("$READ_TIMEOUT_HEADER:300000")
```
fviernau, line 230: maybe: "$READ_TIMEOUT_HEADER:${5 * 60 * 1000}"

fviernau (changes requested):

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
nnobelis: >  it seems to make sense to remove the original val minReadTimeout = Duration.ofSeconds(60) code completely, and exclusively use the header-based approach per endpoint. Would that make sense to you?

But if we have no default Fallback we need to put the annotation on every function ?
I don't remember which call required the 1 minute timeout....

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
nnobelis: Yes currently I only use the `READ_TIMEOUT `. But since one can configure all the three with the header, I thought it would make sense to have them all configurable.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
sschuberth: > But if we have no default Fallback we need to put the annotation on every function ?

I don't think so, as OkHttp has a built-in [default read timeout of 10 seconds](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/read-timeout/).

> I don't remember which call required the 1 minute timeout....

The Git history is your friend... 034eba71f22d391efde8ac3551ffbc7a0d495801 indeed say the 60s timeout should apply to all requests.

fviernau:
Shall we report this timeout issue to FossId? I believe the `deleteScan` API called should be turned into an async one, if it takes that long.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
sschuberth: > But since one can configure all the three with the header, I thought it would make sense to have them all configurable.

I'd prefer to keep things simple here and not make these configurable unless needed, esp. as the code otherwise contains some not-so-nice duplications.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -160,18 +167,14 @@ interface FossIdRestService {
...
-                .addConverterFactory(JacksonConverterFactory.create(JSON_MAPPER))
-                .build()
+            val retrofit = Retrofit.Builder().apply {
+                client(
+                    client?.run {
+                        val timeout = maxOf(minReadTimeout, Duration.ofMillis(readTimeoutMillis.toLong()))
+                        newBuilder().readTimeout(timeout).addInterceptor(TimeoutInterceptor).build()
+                    } ?: OkHttpClient.Builder().readTimeout(minReadTimeout).addInterceptor(TimeoutInterceptor).build()
+                )
+            }.baseUrl(url).addConverterFactory(JacksonConverterFactory.create(JSON_MAPPER)).build()
```
sschuberth, line 177: Can we keep the formatting closer to the original one here to reduce the diff?

nnobelis:
> Shall we report this timeout issue to FossId? I believe the `deleteScan` API called should be turned into an async one, if it takes that long.

We already reported it: they know about it and it's by design. The full response:

> API calls like deleting scans are synchronous, that is, they don’t return until the operation (deletion of a scan) has been completed. This guarantees that database always is in well defined and consistent state when the call returns, and in practice simplifies scripting.

> The flip side of this is of course that the call may take some time, and that the time might differ depending on how large/complex the operation is. Consequently, deleting a scan with a lot of information (a lot of files) may take longer time than deleting a scan with just a few files.

> The fact that such operations may take longer time to respond for large scans is consequently a result of this design, and not a bug per se.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
sschuberth: But actually, I really don't like the complexity / logic duplication in here now. Coming back to [this discussion](https://github.com/oss-review-toolkit/ort/pull/6692/commits/c7fd0332c07181c98b800e038aa77121b7f141ce#r1139953679), shouldn't we really instead unconditionally do `addInterceptor(TimeoutInterceptor)` once for whatever client, and indeed be explicit about all non-default read timeouts by annotating all API endpoints?

I'm also curious about @fviernau's opinion here.

fviernau: > But actually, I really don't like the complexity / logic duplication in here now. Coming back to [this discussion](https://github.com/oss-review-toolkit/ort/pull/6692/commits/c7fd0332c07181c98b800e038aa77121b7f141ce#r1139953679), shouldn't we really instead unconditionally do `addInterceptor(TimeoutInterceptor)` once for whatever client, and indeed be explicit about all non-default read timeouts by annotating all API endpoints?
> 
> I'm also curious about @fviernau's opinion here.

I agree with doing this unconditionally on the client and then annotating the problematic API calls which IIUC are just `readScan` and `deleteScan`. 

I believe moving this to all http client's used in ORT can and should be considered out of scope of this PR, e.g. I would investigate first if there could be nicer annotations 
- like the "@Timeout" one
- how it could work together with setting a value from OrtConfiguration
- maybe a bit more intense testing
- if it applies to ORT at all, because IIUC it was said that these project under `clients` are not really part of ORT but could be considered something separate. This code also dosn't use ORT's `OkHttpClientHelper` but `OkHttp` directly.

sschuberth: > I agree with doing this unconditionally on the client and then annotating the problematic API calls which IIUC are just `readScan` and `deleteScan`.

Not that (as [quoted earlier](https://github.com/oss-review-toolkit/ort/pull/6692/commits/c7fd0332c07181c98b800e038aa77121b7f141ce#r1139976953)) at least https://github.com/oss-review-toolkit/ort/commit/034eba71f22d391efde8ac3551ffbc7a0d495801 claims that the high default read timeout value is need for all API calls. Maybe @MarcelBochtler remembers the details.

> like the "@timeout" one

I don't think that's feasible as that's not a built-in annotation that Retrofit provides. Adding a custom annotation instead of this "header trick" would be much more work.

> how it could work together with setting a value from OrtConfiguration

Values for annotations have to be compile-time constants, so that's going to be hard.

fviernau: > default read timeout value is need for all API calls

thanks for pointing this out.

> I don't think that's feasible as that's not a built-in annotation that Retrofit provides. Adding a custom annotation instead of this "header trick" would be much more work.

If the full solution would be like the following one, then it would not be that much work:
https://github.com/square/retrofit/issues/2982#issuecomment-529268601

anyhow, still believe it can be out of scope.

sschuberth: > If the full solution would be like the following one

That's indeed nice.

nnobelis: > > default read timeout value is need for all API calls
> 
> thanks for pointing this out.
> 
> > I don't think that's feasible as that's not a built-in annotation that Retrofit provides. Adding a custom annotation instead of this "header trick" would be much more work.
> 
> If the full solution would be like the following one, then it would not be that much work: [square/retrofit#2982 (comment)](https://github.com/square/retrofit/issues/2982#issuecomment-529268601)
> 
> anyhow, still believe it can be out of scope.

Yes but this solution relies on reflection. The solution in current PR is based on headers.

nnobelis: Nevertheless I will change the logic to do it unconditionally .

Making it global for ORT is, as Frank said, out of scope.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -159,19 +163,13 @@ interface FossIdRestService {
...
-                        } ?: OkHttpClient.Builder().readTimeout(minReadTimeout).build()
-                    )
-                }
-                .baseUrl(url)
-                .addConverterFactory(JacksonConverterFactory.create(JSON_MAPPER))
-                .build()
+            val retrofit = Retrofit.Builder().apply {
+                client(
+                    client?.run {
+                        newBuilder().addInterceptor(TimeoutInterceptor).build()
```
fviernau, line 169: Does it make sense to de-duplicate the `Should the `.addInterceptor(TimeoutInterceptor)` call?

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -185,6 +183,21 @@ interface FossIdRestService {
...
     }
 
+    /**
+     * An interceptor to set the timeout of the requests based on the call headers. If no timeout header is present,
+     * default timeout of the client is used.
+     */
+    object TimeoutInterceptor : Interceptor {
+        override fun intercept(chain: Interceptor.Chain): okhttp3.Response {
+            val request = chain.request()
+            val newReadTimeout = chain.request().header(READ_TIMEOUT_HEADER)
```
fviernau, line 193: `chain.request()` should be replace with `request`.

fviernau (changes requested):

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -159,19 +163,13 @@ interface FossIdRestService {
...
-                .baseUrl(url)
-                .addConverterFactory(JacksonConverterFactory.create(JSON_MAPPER))
-                .build()
+            val retrofit = Retrofit.Builder().apply {
+                client(
+                    client?.run {
+                        newBuilder().addInterceptor(TimeoutInterceptor).build()
+                    } ?: OkHttpClient.Builder().addInterceptor(TimeoutInterceptor).build()
+                )
+            }.baseUrl(url).addConverterFactory(JacksonConverterFactory.create(JSON_MAPPER)).build()
```
fviernau, line 172: line 172 has unrelated re-formatting.

fviernau (approved):

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -159,17 +163,11 @@ interface FossIdRestService {
...
-                        client?.run {
-                            takeUnless { readTimeoutMillis < minReadTimeout.toMillis() }
-                                ?: newBuilder().readTimeout(minReadTimeout).build()
-                        } ?: OkHttpClient.Builder().readTimeout(minReadTimeout).build()
-                    )
-                }
-                .baseUrl(url)
+            val retrofit = Retrofit.Builder().apply {
+                client(
+                    (client?.newBuilder() ?: OkHttpClient.Builder()).addInterceptor(TimeoutInterceptor).build()
```
sschuberth, line 168: Nit: Could be shortened to `(client ?: OkHttpClient()).newBuilder()`.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -185,6 +183,21 @@ interface FossIdRestService {
...
+    /**
+     * An interceptor to set the timeout of the requests based on the call headers. If no timeout header is present,
+     * default timeout of the client is used.
+     */
+    object TimeoutInterceptor : Interceptor {
+        override fun intercept(chain: Interceptor.Chain): okhttp3.Response {
+            val request = chain.request()
+            val newReadTimeout = request.header(READ_TIMEOUT_HEADER)
+
+            val readTimeout = newReadTimeout?.takeIf { it.isNotEmpty() }?.toInt() ?: chain.readTimeoutMillis()
```
sschuberth, line 195: Nit: Could be shortened to `newReadTimeout?.toIntOrNull() ?: chain.readTimeoutMillis()`.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
sschuberth, line 1: Commit message nits:
- "is set globally, at the OkHttpClient level." -> Maybe instead say that it effected all API endpoints.
- "Global timeout has been removed" -> Please justify why this can be done. I guess the honest answer is that in contrast to what was written in 034eba71f22d391efde8ac3551ffbc7a0d495801, it was never really required to do for all API calls, but only for `listIdentifiedFiles`.

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
nnobelis: If you look at the implementations of `newBuilder()` and `OkHttpClient.Builder()`, the latter does less and therefore saves some CPU cycles :grin:

sschuberth: I know, but it's shorted and aligns with existing code. But I'm ok either way.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
sschuberth, line 1: Commit message:
- "Currently" -> "Previously" (as you're using past tense "affected" later)
- "Global timeout" -> "The timeout of all endpoints" (as it's unclear what "global" refers to; it could be understood as all of ORT)

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
nnobelis: The timeout of all FossID endpoints then ?

sschuberth (dismissed):

sschuberth:
Looks like this has conflicts now due to your other PR merged meanwhile.

MarcelBochtler (approved):

PR #6721, opened 2023-03-20
fix(SpdxExpression): Fix the DNF algorithm

MarcelBochtler:
Whenever a `SpdxExpression` contains an `AND` operator, the DNF needs to be calculated again.
For instance the correct DNF for `a AND (b OR c OR d)` is `(a AND b) OR (a AND c) OR (a AND d)`. Before, the calculated DNF for this expression was `(a AND (b OR c)) OR (a AND d)` which is not a valid DNF.

### DIFF
--- a/utils/spdx/src/main/kotlin/SpdxExpression.kt
+++ b/utils/spdx/src/main/kotlin/SpdxExpression.kt
@@ -253,22 +253,25 @@ class SpdxCompoundExpression(
         val rightDnf = right.disjunctiveNormalForm()
 
         return when (operator) {
-            SpdxOperator.OR -> SpdxCompoundExpression(leftDnf, SpdxOperator.OR, rightDnf)
+            SpdxOperator.OR -> leftDnf or rightDnf
+            SpdxOperator.AND -> distributeAndOverOr(leftDnf, rightDnf)
+        }
+    }
 
-            SpdxOperator.AND -> when {
-                leftDnf is SpdxCompoundExpression && leftDnf.operator == SpdxOperator.OR &&
-                        rightDnf is SpdxCompoundExpression && rightDnf.operator == SpdxOperator.OR ->
-                    ((leftDnf.left and rightDnf.left) or (leftDnf.left and rightDnf.right)) or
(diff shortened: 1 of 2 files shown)


MarcelBochtler:
I'm using this scanner result to check the performance impact of this change:
https://github.com/fviernau/orthw-demo/blob/main/scan-results/scan-result.json.xz

Results on my machine
Current main: 13 seconds
With fix: 1 min 14 seconds.

Therefore, the changes are currently not feasible :/

MarcelBochtler:
Extracting the `distributeAndOverOr` function and only calling this recursively results in ~45 evaluation time on my machine.
I doubt that this problem can be solved in a non-exponential time complexity.

sschuberth:
I wonder whether `disjunctiveNormalForm()` is still needed after all... in https://github.com/oss-review-toolkit/ort/pull/8106 it seems we could simplify remove it.

sschuberth:
> I'm using this scanner result to check the performance impact of this change:
> https://github.com/fviernau/orthw-demo/blob/main/scan-results/scan-result.json.xz

@MarcelBochtler what operation exactly did you perform on the result for benchmarking?

MarcelBochtler:
> > I'm using this scanner result to check the performance impact of this change:
> > https://github.com/fviernau/orthw-demo/blob/main/scan-results/scan-result.json.xz
> 
> @MarcelBochtler what operation exactly did you perform on the result for benchmarking?

I simply ran the evaluator using the linked `scan-result.json`:
```sh
ort evaluate -i scan-result.json -o ort/ --license-classifications-file license-classifications.yml --rules-file evaluator.rules.kts
```

I cannot share the license-classifications and the evaluator rules.
In the `evaluator.rules.kts` we do use `expression.validChoices()` on every expression that offers a choice. Other than that, I don't see anything that could be related to this issue.


> I wonder whether `disjunctiveNormalForm()` is still needed after all... in #8106 it seems we could simplify remove it.

Yes looks like it. Should I close this PR?

sschuberth:
> Should I close this PR?

For the sake of cleaning things up a little, I'd say yes. I believe in the context of the current code base there now nothing anymore in this PR that we could take / reuse.

PR #6728, opened 2023-03-21
Add resolutions to `ResolvedConfiguration`

mnonnenmacher:

### DIFF
--- a/cli/src/main/kotlin/commands/ReporterCommand.kt
+++ b/cli/src/main/kotlin/commands/ReporterCommand.kt
@@ -207,7 +207,9 @@ class ReporterCommand : OrtCommand(
             ortResult = ortResult.replaceConfig(config)
         }
 
-        val resolutionProvider = DefaultResolutionProvider.create(ortResult, resolutionsFile)
+        val resolutionProvider = ortResult.resolvedConfiguration.resolutions?.let { resolutions ->
+            DefaultResolutionProvider().add(resolutions)
+        } ?: DefaultResolutionProvider.create(ortResult, resolutionsFile)
 
         val licenseTextDirectories = listOfNotNull(customLicenseTextsDir.takeIf { it.isDirectory })
 
--- a/cli/src/main/kotlin/commands/ReporterCommand.kt
+++ b/cli/src/main/kotlin/commands/ReporterCommand.kt
@@ -26,6 +26,7 @@ import com.github.ajalt.clikt.parameters.groups.mutuallyExclusiveOptions
 import com.github.ajalt.clikt.parameters.groups.single
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.multiple
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.options.required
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/ResolvedConfiguration.kt
```diff
@@ -41,6 +42,12 @@ data class ResolvedConfiguration(
      */
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     val packageCurations: List<ResolvedPackageCurations> = emptyList(),
+
+    /**
+     * All resolutions that apply to this [OrtResult].
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
sschuberth, line 49: Not strictly related to this commit, but if we continue to add default values to the properties, we could swap the `@JsonInclude(JsonInclude.Include.*)` annotations on the properties with a `@JsonInclude(JsonInclude.Include.NON_DEFAULT)` annotation on the class.

cli/src/main/kotlin/commands/ReporterCommand.kt
```diff
@@ -207,9 +215,10 @@ class ReporterCommand : OrtCommand(
             ortResult = ortResult.replaceConfig(config)
         }
 
-        val resolutionProvider = ortResult.resolvedConfiguration.resolutions?.let { resolutions ->
-            DefaultResolutionProvider().add(resolutions)
-        } ?: DefaultResolutionProvider.create(ortResult, resolutionsFile)
+        val resolutionProvider =
+            ortResult.resolvedConfiguration.resolutions.takeIf { !refreshResolutions }?.let { resolutions ->
```
sschuberth, line 219: Maybe use `takeUnless` to avoid the negation?

cli/src/main/kotlin/commands/ReporterCommand.kt
```diff
@@ -207,7 +207,9 @@ class ReporterCommand : OrtCommand(
             ortResult = ortResult.replaceConfig(config)
         }
 
-        val resolutionProvider = DefaultResolutionProvider.create(ortResult, resolutionsFile)
+        val resolutionProvider = ortResult.resolvedConfiguration.resolutions?.let { resolutions ->
+            DefaultResolutionProvider().add(resolutions)
```
sschuberth: Unrelated, but this made me wonder why we don't have a `DefaultResolutionProvider(resolutions)` constructor. Also the docs for `DefaultResolutionProvider` are outdated as they don't mention vulnerability resolutions. We probably shouldn't mention all types of resolutions there anyway.

mnonnenmacher: I'll make a refactoring in a separate PR.

model/src/main/kotlin/ResolvedConfiguration.kt
sschuberth: > Jackson needs to create an instance of the class using the no-argument constructor

That's indeed crazy. Do you have a reference for that?

sschuberth (dismissed):

model/src/main/kotlin/ResolvedConfiguration.kt
mnonnenmacher: It's in the [docs](https://fasterxml.github.io/jackson-annotations/javadoc/2.7/com/fasterxml/jackson/annotation/JsonInclude.Include.html#NON_DEFAULT):
> When used for a POJO, definition is that only values that differ from the default values of POJO properties are included. This is done by creating an instance of POJO using zero-argument constructor, and accessing property values: value is used as the default value by using equals() method, except for the case where property has `null` value in which straight null check is used.

PR #6732, opened 2023-03-22
build(Gradle): Decouple `maven-core` from `maven-compat`

sschuberth:
Perspectively, the Maven-2-compatibility-layer should not be used with a
Maven 3 core, see [1], so prepare for that split for more fine-grained
dependencies.
    
[1]: https://issues.apache.org/jira/browse/MNG-6561

### DIFF
--- a/analyzer/build.gradle.kts
+++ b/analyzer/build.gradle.kts
@@ -30,7 +30,10 @@ dependencies {
     implementation(project(":utils:ort-utils"))
     implementation(project(":utils:spdx-utils"))
 
-    implementation(libs.bundles.maven)
+    implementation(libs.mavenCore)
+
+    // TODO: Remove this once https://issues.apache.org/jira/browse/MNG-6561 is resolved.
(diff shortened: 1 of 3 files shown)


mnonnenmacher (approved):

sschuberth:
> One test is failing with a Plexus related error, maybe it is important that maven-compat is on the classpath even if we don't use it directly?

Ah, crap, I keep forgetting about [this](https://github.com/oss-review-toolkit/ort/actions/runs/4490185578/jobs/7896933323?pr=6732#step:6:1180). Maybe this is because we use some deprecated classes in `MavenSupport` and thus still run into https://issues.apache.org/jira/browse/MNG-5995 😞

sschuberth:
> thus we still run into https://issues.apache.org/jira/browse/MNG-5995 :disappointed:

Or probably it's rather https://issues.apache.org/jira/browse/MNG-6561.

mnonnenmacher (approved):

PR #6739, opened 2023-03-23
Add package configurations to the resolved configuration

mnonnenmacher:

### DIFF
--- a/model/src/main/kotlin/ResolvedConfiguration.kt
+++ b/model/src/main/kotlin/ResolvedConfiguration.kt
@@ -22,6 +22,7 @@ package org.ossreviewtoolkit.model
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonInclude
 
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/ResolvedConfiguration.kt
```diff
@@ -36,6 +37,12 @@ import org.ossreviewtoolkit.utils.common.getDuplicates
  * TODO: Add further data.
  */
 data class ResolvedConfiguration(
+    /**
+     * All [PackageConfiguration]s that match the scan results in this [OrtResult].
```
sschuberth, line 41: Nit: Why does this explicitly mention "scan result"? If a `PackageConfiguration` contains no `licenseFindingCurations` but `pathExcludes`, there's nothing scanner-specific about it.

mnonnenmacher: Package configurations are always matched against the provenance of the scan result they are applied to, e.g. path exclude paths could be different for source artifacts and repositories.

sschuberth (approved):

PR #6741, opened 2023-03-24
Make all properties of resolved configuration not nullable

fviernau:
Make the null-ability consistent and a tad simpler.

### DIFF
--- a/model/src/main/kotlin/ResolvedConfiguration.kt
+++ b/model/src/main/kotlin/ResolvedConfiguration.kt
@@ -41,8 +41,8 @@ data class ResolvedConfiguration(
     /**
      * All [PackageConfiguration]s that correspond to the scan results in this [OrtResult].
      */
-    @JsonInclude(JsonInclude.Include.NON_NULL)
-    val packageConfigurations: List<PackageConfiguration>? = null,
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    val packageConfigurations: List<PackageConfiguration> = emptyList(),
 
     /**
      * The curations for all enabled providers ordered highest-priority-first. The list contains exactly one entry for
--- a/model/src/main/kotlin/ResolvedConfiguration.kt
+++ b/model/src/main/kotlin/ResolvedConfiguration.kt
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.annotation.JsonInclude
 
 import org.ossreviewtoolkit.model.config.PackageConfiguration
 import org.ossreviewtoolkit.model.config.Resolutions
+import org.ossreviewtoolkit.model.utils.ResolutionsFilter
 import org.ossreviewtoolkit.utils.common.getDuplicates
 
 /**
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/utils/JsonIncludeFilters.kt
sschuberth, line 1: Nit: Maybe call the file "JsonInclude*Value*Filters.kt" as the annotation's parameter name is "valueFilter"?

model/src/main/kotlin/ResolvedConfiguration.kt
```diff
@@ -53,8 +54,8 @@ data class ResolvedConfiguration(
     /**
      * All resolutions that apply to this [OrtResult].
      */
-    @JsonInclude(JsonInclude.Include.NON_NULL)
-    val resolutions: Resolutions? = null
+    @JsonInclude(JsonInclude.Include.CUSTOM, valueFilter = ResolutionsFilter::class)
+    val resolutions: Resolutions = Resolutions()
```
sschuberth, line 58: I don't remember anymore why we're using a custom `valueFilter` in these cases instead of just `JsonInclude.Include.NON_DEFAULT`. Can you remind me? 1fec27e3dd6789c21f11b3d273f08a890a6315f5 does not really give a clue.

mnonnenmacher: The property is nullable on purpose. If it is `null` it means no resolutions were resolved so far. An empty `Resolutions` object means that resolutions were resolved but none were found. I should have added this to the property docs.
Removing nullability breaks the handling in the  `ReporterCommand`.

model/src/main/kotlin/ResolvedConfiguration.kt
```diff
@@ -41,8 +41,8 @@ data class ResolvedConfiguration(
     /**
      * All [PackageConfiguration]s that match the scan results in this [OrtResult].
      */
-    @JsonInclude(JsonInclude.Include.NON_NULL)
-    val packageConfigurations: List<PackageConfiguration>? = null,
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    val packageConfigurations: List<PackageConfiguration> = emptyList(),
```
mnonnenmacher, line 45: Nullability is intentional here for the same reason as for `resolutions`.

model/src/main/kotlin/utils/JsonIncludeValueFilters.kt
mnonnenmacher, line 1: Commit message: Typo in "deduplicate"

model/src/main/kotlin/ResolvedConfiguration.kt
fviernau: Hm, I've adjusted the handling in the `ReporterCommand`. It seemed to me it does not need to be able to distinguish between these two cases. Have you noticed the changes there?

fviernau: I've split out some hopefully uncontroversial bits to https://github.com/oss-review-toolkit/ort/pull/6745

fviernau: sure.

fviernau:
pretty much outdated, not worth keeping around.

PR #6745, opened 2023-03-24
Json value filter improvements

fviernau:
See individual commits.

### DIFF
--- a/model/src/main/kotlin/utils/JsonIncludeValueFilters.kt
+++ b/model/src/main/kotlin/utils/JsonIncludeValueFilters.kt
@@ -43,8 +43,10 @@ internal class LicenseChoiceFilter {
         other is LicenseChoices && other.isEmpty()
 }
 
-// A custom value filter for [PackageLinkage] to work around
-// https://github.com/FasterXML/jackson-module-kotlin/issues/193.
+/**
+ * A custom value filter for [PackageLinkage] to work around
+ * https://github.com/FasterXML/jackson-module-kotlin/issues/193.
+ */
 @Suppress("EqualsOrHashCode", "EqualsWithHashCodeExist")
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/utils/JsonIncludeValueFilters.kt
```diff
@@ -19,44 +19,45 @@
 
 package org.ossreviewtoolkit.model.utils
 
+import kotlin.reflect.KClass
+import kotlin.reflect.safeCast
```
sschuberth, line 23: I'd prefer the model to stay free of reflection, which helps with GraalVM native-images and compiling to non-JVM platforms. Can we simply drop this commit?

sschuberth (changes requested):

model/src/main/kotlin/utils/JsonIncludeValueFilters.kt
fviernau: @mnonnenmacher @MarcelBochtler  does this align with your view as well?

mnonnenmacher: Looks like you removed the commit already so I can't give a specific answer, but IMO we should generally try to keep our Jackson usage simple to not hinder the planned migration to Kotlin serialization.

sschuberth: > Looks like you removed the commit already so I can't give a specific answer

You can still [see it in the diff of the force-push](https://github.com/oss-review-toolkit/ort/compare/38e5c14ca64bb944f509cfd3e70c893f96e8b524..1c442a49c545c675b96045326a02ee405f6e06b0).

sschuberth (approved):

PR #6760, opened 2023-03-28
fix(SpdxDocumentModelMapper): Only reset concluded licenses for sources

sschuberth:
As a single SPDX package cannot capture the same amount of metadata as an
ORT package can, the `SpdxDocumentModelMapper` creates up to three SPDX
packages for a single ORT package: A binary package, a VCS source package,
and a source artifact package. The latter two copy all properties from the
binary package while adjusting esp. the `downloadLocation`. As part of
these adjustments, the licenses were reset to `NOASSERTION` until now.

However, that is inconsistent and misleading, as there has been an attempt
made to determine the declared license from metadata, and metadata is
usually stored in definition files along with the source code. Thus,
source packages should contain `licenseDeclared` from the binary package
at least as a subset.

Note that detected licenses, if any, would go into the separate
`licenseInfoFromFiles` property, also see the currently pending #6757.

### DIFF
--- a/reporter/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
+++ b/reporter/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
@@ -59,7 +59,7 @@
     "filesAnalyzed" : true,
     "homepage" : "first package's homepage URL",
     "licenseConcluded" : "NOASSERTION",
-    "licenseDeclared" : "NOASSERTION",
+    "licenseDeclared" : "BSD-3-Clause AND (MIT OR GPL-2.0-only)",
     "licenseInfoFromFiles" : [ "Apache-2.0", "BSD-2-Clause" ],
     "name" : "first-package",
     "packageVerificationCode" : {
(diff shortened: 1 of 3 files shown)


mnonnenmacher (approved):

sschuberth:
At the ask of @fviernau I'm waiting with the merge until @tsteenbe had a chance to look at this.

mnonnenmacher:
> At the ask of @fviernau I'm waiting with the merge until @tsteenbe had a chance to look at this.

Yes, I was just dismissing my approval because I'm not sure about the concluded license field. The declared license should be no issue, but the concluded license usually applies to the binary package, but the source packages (esp. the vcs) could contain other licenses, e.g. in test code, so that the concluded license becomes invalid.

tsteenbe (changes requested):
The license concluded by ORT users commonly applies to the binary or the part of the source they use so apply the concluded license to source code repository or source artifact is imo incorrect. Source code repository or source artifact often contain file not relevant to binary such as optional code, build scripts, testing functionality or documentation - this is why we have path excludes.

Do not see any issues applying the declared license to source code repository but not convinced yet that this is a good idea for the source artifacts. Take for example Maven the sources JARs do not contain pom.xml file which holds package metadata although sometimes does include the LICENSE in META-INF folder. For NPM it does make sense since package.json is always include in the tar.gz file.

sschuberth:
> Take for example Maven the sources JARs do not contain pom.xml

They often actually do @tsteenbe, as part of the `META-INF` directory structure, see e.g. https://repo.maven.apache.org/maven2/org/apache/commons/commons-text/1.1/commons-text-1.1-sources.jar.

mnonnenmacher (approved):

sschuberth:
> @tsteenbe can we dismiss your review?

@tsteenbe if I'm not getting a veto from your side I plan to merge this by the end of this week.

tsteenbe (approved):

tsteenbe:
@sschuberth  Was on holiday, Ok with this PR now.

PR #6766, opened 2023-03-29
Migrate remaining test code to `patchExpectedResult2()`

fviernau:
See individual commits.

`GoDepFunTest` is still left as todo for a future PR.

### DIFF
--- a/analyzer/src/funTest/kotlin/managers/NpmVersionUrlFunTest.kt
+++ b/analyzer/src/funTest/kotlin/managers/NpmVersionUrlFunTest.kt
@@ -24,8 +24,6 @@ import com.fasterxml.jackson.module.kotlin.readValue
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.shouldBe
 
-import java.time.Instant
-
 import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
@@ -51,7 +49,3 @@ class NpmVersionUrlFunTest : WordSpec({
 
 private fun createNpm() =
     Npm("NPM", USER_DIR, AnalyzerConfiguration(allowDynamicVersions = true), RepositoryConfiguration())
-
-private fun ProjectAnalyzerResult.withInvariantIssues() =
-    // Account for different NPM versions to return issues in different order.
-    copy(issues = issues.sortedBy { it.message }.map { it.copy(timestamp = Instant.EPOCH) })
--- a/analyzer/src/funTest/kotlin/managers/NpmVersionUrlFunTest.kt
+++ b/analyzer/src/funTest/kotlin/managers/NpmVersionUrlFunTest.kt
@@ -45,7 +45,6 @@ class NpmVersionUrlFunTest : WordSpec({
     "NPM" should {
         "resolve dependencies with URLs as versions correctly" {
             val definitionFile = projectDir.resolve("package.json")
-            val config = AnalyzerConfiguration(allowDynamicVersions = true)
             val vcsPath = vcsDir.getPathToRoot(projectDir)
             val expectedResultYaml = patchExpectedResult(
                 projectDir.resolveSibling("npm-version-urls-expected-output.yml"),
@@ -56,14 +55,15 @@ class NpmVersionUrlFunTest : WordSpec({
             )
             val expectedResult = yamlMapper.readValue<ProjectAnalyzerResult>(expectedResultYaml)
 
-            val actualResult = createNpm(config).resolveSingleProject(definitionFile, resolveScopes = true)
+            val actualResult = createNpm().resolveSingleProject(definitionFile, resolveScopes = true)
 
             actualResult.withInvariantIssues() shouldBe expectedResult.withInvariantIssues()
         }
     }
 })
 
-private fun createNpm(config: AnalyzerConfiguration) = Npm("NPM", USER_DIR, config, RepositoryConfiguration())
+private fun createNpm() =
+    Npm("NPM", USER_DIR, AnalyzerConfiguration(allowDynamicVersions = true), RepositoryConfiguration())
 
 private fun ProjectAnalyzerResult.withInvariantIssues() =
     // Account for different NPM versions to return issues in different order.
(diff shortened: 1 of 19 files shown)


analyzer/src/funTest/kotlin/managers/NpmVersionUrlFunTest.kt
sschuberth, line 1: Commit message nit: "Drop" sounds a bit misleading as it's rather inlined than omitted completely.

utils/test/src/main/kotlin/Utils.kt
```diff
@@ -146,3 +147,7 @@ fun <T, U> transformingCollectionMatcher(
 fun <T, U> transformingCollectionEmptyMatcher(
     transform: (T) -> Collection<U>
 ): Matcher<T?> = neverNullMatcher { value -> beEmpty<U>().test(transform(value)) }
+
+fun ProjectAnalyzerResult.withInvariantIssues() =
+    // Account for different NPM versions to return issues in different order.
+    copy(issues = issues.sortedBy { it.message }.map { it.copy(timestamp = Instant.EPOCH) })
```
sschuberth, line 165: As this is an extension function on an Analyzer-specific class, I'd propose to move it to `analyzer/src/testFixtures/kotlin/Extensions.kt` instead.

cli/src/funTest/kotlin/OrtMainFunTest.kt
sschuberth, line 1: Commit message: I don't believe it's faster to create a temporary directory than to use an existing one. I'd simple drop that bit.

sschuberth (changes requested):
Here's my first round of reviews; I still want to think about the `ReplacementsBuilder` a bit.

cli/src/funTest/kotlin/OrtMainFunTest.kt
fviernau: An empty directory is faster to analyze, of course not to create, but isn't that obvious?

sschuberth: At least to me it wasn't obvious what exactly "make the test faster" refers to.

utils/test/src/main/kotlin/Utils.kt
```diff
@@ -80,16 +80,25 @@ fun patchExpectedResult(
...
+    infix fun String.replaces(value: String) {
+        replacements[value] = this
+    }
+}
+
 fun patchExpectedResult2(
     expectedResultFile: File,
     definitionFile: File,
-    custom: Map<String, String> = emptyMap()
+    block: ReplacementsBuilder.() -> Unit = {}
```
sschuberth, line 96: TBH, I liked the original *from*-*to*-syntax in the callers better, and I'd prefer to not introduce a helper class like `ReplacementsBuilder` for just this purpose. So my proposal would be to change this to

    block: MutableMap<String, String>.() -> Unit = {}

and also ... (see below)

utils/test/src/main/kotlin/Utils.kt
```diff
@@ -80,16 +80,25 @@ fun patchExpectedResult(
...
 ): String {
     val projectDir = definitionFile.parentFile
     val vcsDir = VersionControlSystem.forDirectory(projectDir)!!
     val vcsUrl = vcsDir.getRemoteUrl()
     val vcsRevision = vcsDir.getRevision()
     val vcsPath = vcsDir.getPathToRoot(projectDir)
+    val custom = ReplacementsBuilder().apply(block).replacements
```
sschuberth: ... change this to `val custom = buildMap(block)` (or even inline it below).

utils/test/src/main/kotlin/Utils.kt
```diff
@@ -80,7 +80,9 @@ fun patchExpectedResult(
         .replaceIfNotNull("<REPLACE_URL_PROCESSED>", urlProcessed)
 }
 
-class ReplacementsBuilder {
+class ReplacementsBuilder(
+    val urlProcessed: String
```
sschuberth, line 84: It seems a bit arbitrary to make `urlProcessed` available here just because `CarthageFunTest` happens to require it. Can't we simply remove `<REPLACE_GITHUB_PROJECT>` and hard-code it to "oss-review-toolkit" in the YAML files, similar to like you did with some other patterns? This would then also work with the `buildMap`-based approach I'm proposing,

sschuberth (changes requested):

utils/test/src/main/kotlin/Utils.kt
sschuberth: But actually, when getting rid of `<REPLACE_GITHUB_PROJECT>`, there's maybe no reason anymore to turn custom replacements into a lambda at all.

utils/test/src/main/kotlin/Utils.kt
fviernau: This is not just `CarthageFunTest` but also `GoDepFunTest` which I haven't migrated yet.

Anyhow, I'll just turn it into your build map approach and drop the `<REPLACE_GITHUB_PROJECT>` patterns.

utils/test/src/main/kotlin/Utils.kt
fviernau: > But actually, when getting rid of `<REPLACE_GITHUB_PROJECT>`, there's maybe no reason anymore to turn custom replacements into a lambda at all.

You like readability without lambda more?

sschuberth: > You like readability without lambda more?

It does not make much of a difference to me. Feel free to keep it at a lamda, if you like. Maybe, yes, it's a tad nicer after all.

plugins/package-managers/carthage/src/funTest/assets/projects/synthetic/carthage-expected-output.yml
```diff
@@ -1,6 +1,6 @@
 ---
 project:
-  id: "Carthage:oss-review-toolkit:ort:281d7b217614a57c871b23cf83d1ffa12e3912f5"
+  id: "Carthage:oss-review-toolkit:ort:<REPLACE_REVISION>"
```
sschuberth, line 3: Unintended change?

fviernau:
> Are the usual test de-indentation changes coming in another follow-up PR?

yes.

sschuberth (approved):

sschuberth (approved):

PR #6770, opened 2023-03-29
Resolved configuration refactorings

mnonnenmacher:
This is an intermediary refactoring step to consolidate some code, it will be continued in another PR. Please see the commit messages for details.


model/src/main/kotlin/utils/OrtResultExtensions.kt
```diff
@@ -42,6 +43,18 @@ fun OrtResult.addPackageConfigurations(packageConfigurationProvider: PackageConf
     return copy(resolvedConfiguration = resolvedConfiguration.copy(packageConfigurations = packageConfigurations))
 }
 
+/**
+ * Add all package curations from [packageCurationProviders] that match any packages in this [OrtResult] to
+ * [OrtResult.resolvedConfiguration], overwriting any previously contained package curations. The
+ * [packageCurationProviders] must be ordered highest-priority-first.
+ */
+fun OrtResult.addPackageCurations(packageCurationProviders: List<Pair<String, PackageCurationProvider>>): OrtResult {
```
fviernau, line 48: Would it make sense to name this `setPackageCurations()`, because "adding" might hide the fact that existing curations get dropped.

model/src/main/kotlin/utils/OrtResultExtensions.kt
```diff
@@ -42,6 +43,18 @@ fun OrtResult.addPackageConfigurations(packageConfigurationProvider: PackageConf
     return copy(resolvedConfiguration = resolvedConfiguration.copy(packageConfigurations = packageConfigurations))
 }
 
+/**
```
fviernau, line 43: commit: One of the (my) goals when introducing IDs for providers was to allow for "updating" the curations for a specific provider ID. Looking at is function and the way it's being used in the following changes seems to move further away from that, as all existing curations get overwritten.

Have you considered this use case? / what's the plan for it?

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -52,6 +52,7 @@ import org.ossreviewtoolkit.cli.utils.readOrtResult
 import org.ossreviewtoolkit.cli.utils.writeOrtResult
 import org.ossreviewtoolkit.evaluator.Evaluator
 import org.ossreviewtoolkit.model.FileFormat
+import org.ossreviewtoolkit.model.ResolvedPackageCurations.Companion.REPOSITORY_CONFIGURATION_PROVIDER_ID
```
fviernau, line 55: commit-message: Please describe the change in behavior and the reason why this change is desired.

fviernau (changes requested):

cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -275,8 +277,17 @@ class EvaluatorCommand : OrtCommand(
         }
 
         if (packageCurationsDir != null || packageCurationsFile != null) {
-            val provider = FilePackageCurationProvider(packageCurationsFile, packageCurationsDir)
-            ortResultInput = ortResultInput.replacePackageCurations(provider, providerId = "EvaluatorCommandOption")
+            val packageCurationProviders = buildList {
+                if (ortConfig.enableRepositoryPackageCurations) {
```
fviernau, line 283: This should IMO use logic analog to the SetCommand touched in this PR as well, rather than relying on `enableRepositoryPackageCurations`.

fviernau (changes requested):

model/src/main/kotlin/utils/OrtResultExtensions.kt
mnonnenmacher: > Have you considered this use case? / what's the plan for it?

Yes, one of the goals of the refactoring is to implement this in a generic way. I'll contact you separately to make sure that it works for the orthw use case.

model/src/main/kotlin/utils/OrtResultExtensions.kt
mnonnenmacher: Yes, I will rename them in a follow-up for that reason.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
mnonnenmacher: To me the previous behavior makes sense for a helper command like `SetCommand`, but for the `EvaluatorCommand` it makes more sense to behave according to the global configuration.

model/src/main/kotlin/utils/ConfigurationResolver.kt
```diff
@@ -39,18 +36,6 @@ import org.ossreviewtoolkit.model.config.PackageConfiguration
...
-    fun resolveConfiguration(
-        analyzerResult: AnalyzerResult,
-        curationProviders: List<Pair<String, PackageCurationProvider>>
-    ): ResolvedConfiguration =
-        ResolvedConfiguration(
-            packageCurations = resolvePackageCurations(analyzerResult.packages, curationProviders)
-        )
-
     /**
      * Resolved the [PackageConfiguration]s that match the [scan results][scanResultProvider] for the provided
```
fviernau, line 40: What's the rationale for dropping `resolveConfiguration()` from here? To me it's not obvious why it doesn't belong to this class.

helper-cli/src/main/kotlin/commands/packagecuration/SetCommand.kt
```diff
@@ -58,10 +60,23 @@ class SetCommand : CliktCommand(
         .convert { it.absoluteFile.normalize() }
 
     override fun run() {
-        val provider = FilePackageCurationProvider(packageCurationsFile, packageCurationsDir)
+        val ortResultInput = readOrtResult(ortFile)
```
fviernau, line 63: Adding a new function and the dropping the other separately hides a bit the behavioral changes. 
Is there any particular reason for not just changing / refactoring the existing `replace...`  function to the `add...` function?

cli/src/main/kotlin/commands/EvaluatorCommand.kt
fviernau: I preferred not to consume that config property in the evaluator, to not start worrying about config changes inbetween the stages. 

Anyhow, I believe we should talk about the overall picture how resolving configuration is supposed to work, then such details will be clear anyway. I'm missing this clarity / alignment a bit which makes it hard for me to review this.

mnonnenmacher: > I preferred not to consume that config property in the evaluator, to not start worrying about config changes inbetween the stages.

The command already reacts to global configuration changes in the rules, license classifications, rule violation threshold, and so on. It would be rather inconsistent if this was the one config property it ignores.

> Anyhow, I believe we should talk about the overall picture how resolving configuration is supposed to work, then such details will be clear anyway. I'm missing this clarity / alignment a bit which makes it hard for me to review this.

The next PR will make the idea a lot more clear, I would suggest we have the discussion there as this PR only has a minor functional change in the `EvaluatorCommand`.

model/src/main/kotlin/utils/ConfigurationResolver.kt
fviernau: > It's removed because it's unused

It wasn't unused before this PR

mnonnenmacher: But now it is, what's the point? The function was basically just a wrapper for `resolvePackageCurations` but with the generic name it did not fit well the situation that different tools need to resolve different configurations based on several conditions. So I first want to complete the picture of which tools need to resolve which configuration under which conditions and then come up with a nicer API.

cli/src/main/kotlin/commands/EvaluatorCommand.kt
fviernau: > Not sure what your points is, by default all configuration files are located in ~/.ort/config/ 

My point is to distinguish between:
 
1. the tooling config
2. entries in some upstream database like clearly defined, OSV or `ort-config`

> That's why I'm currently working on increasing the consistency while at the same time not breaking the orthw workflow.

Maybe again: propose to have a chat about how the config resolution shall work in the end and how the mentioned `orthw` workflow may actually look like in future.

model/src/main/kotlin/utils/ConfigurationResolver.kt
fviernau: > But now it is, what's the point?

I just wanted to understand the rationale for dropping it. 

> it's removed because it's unused

did not help me with that. But now it's clearer.

fviernau (approved):

PR #6772, opened 2023-03-30
test(GoDepFunTest): Make use of `patchExpectedResult2()`

fviernau:
...plus a couple of clean-ups. See individual commits.

### DIFF
--- a/analyzer/src/funTest/kotlin/managers/NpmFunTest.kt
+++ b/analyzer/src/funTest/kotlin/managers/NpmFunTest.kt
@@ -71,11 +71,7 @@ class NpmFunTest : WordSpec({
                 resolveScopes = true
             )
 
-            patchActualResult(result.toYaml()) shouldBe patchExpectedResult2(
-                expectedResultFile,
-                definitionFile,
-                custom = mapOf("<REPLACE_LOCKFILE_NAME>" to "npm-shrinkwrap.json")
-            )
+            patchActualResult(result.toYaml()) shouldBe patchExpectedResult2(expectedResultFile, definitionFile)
         }
 
         "resolve package-lock dependencies correctly" {
--- a/analyzer/src/funTest/kotlin/managers/NpmFunTest.kt
+++ b/analyzer/src/funTest/kotlin/managers/NpmFunTest.kt
@@ -26,6 +26,8 @@ import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.string.shouldContain
 
+import java.io.File
+
 import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
@@ -47,7 +49,7 @@ class NpmFunTest : WordSpec({
             val definitionFile = getAssetFile("projects/synthetic/npm/shrinkwrap/package.json")
             val expectedResultFile = getAssetFile("projects/synthetic/npm-expected-output.yml")
 
-            val result = createNpm().resolveSingleProject(definitionFile, resolveScopes = true)
+            val result = resolveSingleProject(definitionFile, resolveScopes = true)
 
             patchActualResult(result.toYaml()) shouldBe patchExpectedResult2(
                 expectedResultFile,
@@ -63,13 +65,11 @@ class NpmFunTest : WordSpec({
             val definitionFile = getAssetFile("projects/synthetic/npm/shrinkwrap/package.json")
             val expectedResultFile = getAssetFile("projects/synthetic/npm-expected-output-scope-excludes.yml")
 
-            val analyzerConfig = AnalyzerConfiguration(skipExcluded = true)
-            val scopeExclude = ScopeExclude("devDependencies", ScopeExcludeReason.TEST_DEPENDENCY_OF)
-            val excludes = Excludes(scopes = listOf(scopeExclude))
-            val repoConfig = RepositoryConfiguration(excludes = excludes)
-
-            val result = createNpm(analyzerConfig = analyzerConfig, repoConfig = repoConfig)
-                .resolveSingleProject(definitionFile, resolveScopes = true)
+            val result = resolveSingleProject(
+                definitionFile,
+                excludedScopes = setOf("devDependencies"),
+                resolveScopes = true
+            )
 
             patchActualResult(result.toYaml()) shouldBe patchExpectedResult2(
                 expectedResultFile,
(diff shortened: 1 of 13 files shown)


analyzer/src/funTest/kotlin/managers/SbtFunTest.kt
```diff
@@ -38,11 +36,9 @@ class SbtFunTest : StringSpec({
         // Clean any previously generated POM files / target directories.
         Git().run(definitionFile.parentFile, "clean", "-fd")
 
-        val ortResult = Analyzer(AnalyzerConfiguration()).run {
-            analyze(findManagedFiles(definitionFile.parentFile, setOf(Sbt.Factory())))
-        }
+        val ortResult = analyze(definitionFile.parentFile, packageManagers = setOf(Sbt.Factory()))
 
-        patchActualResult(ortResult.withResolvedScopes(), patchStartAndEndTime = true) shouldBe expectedResult
+        patchActualResult(ortResult, patchStartAndEndTime = true) shouldBe expectedResult
```
sschuberth, line 41: Shouldn't `.withResolvedScopes()` be kept here?

analyzer/src/testFixtures/kotlin/Extensions.kt
```diff
@@ -97,3 +102,18 @@ private fun Project.filterReferencedPackages(allPackages: Set<Package>): Set<Pac
...
     copy(issues = issues.sortedBy { it.message }.map { it.copy(timestamp = Instant.EPOCH) })
+
+fun Spec.analyze(
+    projectDir: File,
+    allowDynamicVersions: Boolean = false,
+    packageManagers: Collection<PackageManagerFactory> = PackageManager.ALL.values
+): OrtResult {
+    val config = AnalyzerConfiguration(
+        allowDynamicVersions = allowDynamicVersions
+    )
```
sschuberth, line 113: Nit: It probably makes to simply write this as `val config = AnalyzerConfiguration(allowDynamicVersions)`.

plugins/package-managers/pub/src/funTest/kotlin/PubFunTest.kt
```diff
@@ -70,7 +68,7 @@ class PubFunTest : WordSpec({
             val definitionFile = getAssetFile("projects/synthetic/multi-module/pubspec.yaml")
             val expectedResultFile = getAssetFile("projects/synthetic/pub-expected-output-multi-module.yml")
 
-            val analyzerResult = analyze(definitionFile.parentFile).patchPackages()
+            val analyzerResult = analyze(definitionFile.parentFile).analyzer!!.result.patchPackages()
```
sschuberth, line 71: Previously this code path called `.withResolvedScopes()`, now it doesn't anymore.

analyzer/src/funTest/kotlin/managers/SbtFunTest.kt
fviernau: I forgot to put it into the `analyze()` function. Now it's in.

analyzer/src/funTest/kotlin/managers/SbtFunTest.kt
```diff
@@ -53,9 +49,7 @@ class SbtFunTest : StringSpec({
         // Clean any previously generated POM files / target directories.
         Git().run(definitionFile.parentFile, "clean", "-fd")
 
-        val ortResult = Analyzer(AnalyzerConfiguration()).run {
-            analyze(findManagedFiles(definitionFile.parentFile, setOf(Sbt.Factory())))
-        }
+        val ortResult = analyze(definitionFile.parentFile, packageManagers = setOf(Sbt.Factory()))
 
         patchActualResult(ortResult.withResolvedScopes(), patchStartAndEndTime = true) shouldBe expectedResult
```
sschuberth, line 54: `.withResolvedScopes()` should be dropped here now.

cli/src/funTest/kotlin/AnalyzerFunTest.kt
```diff
@@ -55,9 +55,7 @@ class AnalyzerFunTest : WordSpec({
             )
             val outputDir = createTestTempDir().also { GitRepo().download(pkg, it) }
 
-            val result = Analyzer(AnalyzerConfiguration()).run {
-                analyze(findManagedFiles(outputDir))
-            }.withResolvedScopes().toYaml()
+            val result = analyze(outputDir)
```
sschuberth, line 58: Isn't this missing a `.toYaml()` now?

analyzer/src/funTest/kotlin/managers/NpmFunTest.kt
```diff
@@ -141,15 +140,30 @@ class NpmFunTest : WordSpec({
...
         }
     }
 })
 
-private fun createNpm(
-    analyzerConfig: AnalyzerConfiguration = AnalyzerConfiguration(),
-    repoConfig: RepositoryConfiguration = RepositoryConfiguration()
-) =
-    Npm("NPM", USER_DIR, analyzerConfig, repoConfig)
+private fun resolveSingleProject(
```
sschuberth, line 146: Maybe call this `resolveNpmSingleProject` to distinguish this more from `npm.resolveSingleProject`.

sschuberth (changes requested):

analyzer/src/testFixtures/kotlin/Extensions.kt
```diff
@@ -97,3 +102,15 @@ private fun Project.filterReferencedPackages(allPackages: Set<Package>): Set<Pac
 fun ProjectAnalyzerResult.withInvariantIssues() =
     // Account for different NPM versions to return issues in different order.
     copy(issues = issues.sortedBy { it.message }.map { it.copy(timestamp = Instant.EPOCH) })
+
+fun Spec.analyze(
```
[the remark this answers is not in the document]
fviernau: > Why is this an extension function on Spec as that class context doe snot seem to be required?

To not pollute other scopes with this function where it's not needed.

> And maybe for clarity name this analyzeWithResolvedScopes()?

I prefer to keep the shorter name. As soon as a caller is introduces which doesn't need resolved scopes I'd foresee anyway to add a parameter like e.g.: `analyze(resolveScopes = false)`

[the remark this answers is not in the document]
sschuberth: > As soon as a caller is introduced which doesn't need resolved scopes I'd foresee anyway to add a parameter like e.g.: `analyze(resolveScopes = false)`

I was thinking about that, too, and actually was wondering whether we should move `.withResolvedScope()` into that function in the first place, or leave it to the caller. Amount-of-code-wise that would make no difference to using a `resolveScopes = false` argument later. What do you think?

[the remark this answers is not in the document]
fviernau: In the more frequent case where `resolveScopes = true`, the parameter can be omit, because `true` would be the default value. I still prefer to move it into the function because **all** callers want the scopes to be resolved.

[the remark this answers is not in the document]
mnonnenmacher: > To not pollute other scopes with this function where it's not needed. (And to get the permission to put it into this file)

The function is only available to test code anyway, so I'm not sure what kind of "leak" you are trying to prevent here.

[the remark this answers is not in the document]
sschuberth: We could put it into a new `Utils.kt` inside `testFixtures` or so. But if the name is supposed to stay at the generic `analyze`, then I'd indeed prefer to "hide" it inside the `Spec` scope, even if that feel like cheating just to put it into `Extensions.kt`. If needed, we can anyway move this later.

PR #6777, opened 2023-03-30
Move the SPDX package manager to its own plugin project

sschuberth:


analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -58,7 +58,7 @@ import org.ossreviewtoolkit.utils.ort.Environment
  * The class to run the analysis. The signatures of public functions in this class define the library API.
  */
 class Analyzer(private val config: AnalyzerConfiguration, private val labels: Map<String, String> = emptyMap()) {
-    companion object : Logging
+    internal companion object : Logging
```
fviernau, line 61: Would it be reasonable to use `internal` is minimum scope, to avoid switching it back an forth?

fviernau (approved):

analyzer/src/main/kotlin/Analyzer.kt
sschuberth: What do you mean by "switching it back an forth"? Do you mean in case the `private` entries would need to become `internal` for some reason in the future?

sschuberth: As we already have several `private companion object : Logging` we could do a bulk change to `internal companion object : Logging` any time later, if needed.

PR #6778, opened 2023-03-30
refactor(cli): Remove deprecated `forceOverwrite` option

mnonnenmacher:
The option is deprecated since 14234ba as `-P ort.forceOverwrite=true` can be used instead.


sschuberth (approved):
Semi-off-topic: Should `helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt` be adjusted to look at `ortConfig.forceOverwrite`, too?

PR #6781, opened 2023-03-31
tests: Use `toYaml()` instead of `yamlMapper` directly

fviernau:


scanner/src/test/kotlin/ScannerTest.kt
sschuberth, line 1: Nit: I guess the type should be "refactor" (as the test relationship also becomes clear from the scope).

utils/spdx/src/test/kotlin/SpdxExpressionTest.kt
```diff
@@ -40,660 +40,658 @@ import org.ossreviewtoolkit.utils.spdx.SpdxLicenseException.*
...
 
-class SpdxExpressionTest : WordSpec() {
+class SpdxExpressionTest : WordSpec({
     private val yamlMapper = YAMLMapper()
 
-    init {
-        "toString()" should {
-            "return the textual SPDX expression" {
-                val expression = "a+ AND (b WITH exception1 OR c+) AND d WITH exception2"
+    "toString()" should {
```
sschuberth, line 46: I guess the `private` needs to be dropped from the variable above now...

scanner/src/test/kotlin/ScannerTest.kt
fviernau: I choose test: because then there is an automated way to determine commits which affect production code.

Would it make sense to you?

sschuberth: It's just a nit, feel free to keep it as-is.

sschuberth (approved):

sschuberth (dismissed):

sschuberth (approved):

fviernau:
merge: failing tests unrelated.

Issue #4720, opened 2021-11-19
Maven dependencies might be de-duplicated


sschuberth:
I just verified that ORT's analyzer suffers from the [same problem as CycloneDX's Maven plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin/issues/116): The Maven resolver library might de-duplicated dependencies that have already been reported, resulting in an incomplete tree for some  (direct) dependencies.

While at first the discussion in that issue makes it sounds like a pure [upstream maven issue](https://issues.apache.org/jira/browse/MDEP-644), the question remains why apparently the [depgraph-maven-plugin](https://github.com/ferstl/depgraph-maven-plugin) gets it right nonetheless.

sschuberth:
On a related note, there's [a tree-based API for resolution of Maven project dependencies](http://maven.apache.org/shared/maven-dependency-tree/) which might solve the problem, and we should consider using it instead of our low-level approach. That library apparently also does [only download the POMs](https://issues.apache.org/jira/browse/MSHARED-519) which would allow us to get rid of [our work-around](https://github.com/oss-review-toolkit/ort/pull/3445).

sschuberth:
> On a related note, there's [a tree-based API for resolution of Maven project dependencies](http://maven.apache.org/shared/maven-dependency-tree/) which might solve the problem

Looks like we'd need to use its [DefaultDependencyCollectorBuilder](https://maven.apache.org/shared/maven-dependency-tree/apidocs/org/apache/maven/shared/dependency/graph/internal/DefaultDependencyCollectorBuilder.html) instead of the [DefaultDependencyGraphBuilder](https://maven.apache.org/shared/maven-dependency-tree/apidocs/org/apache/maven/shared/dependency/graph/internal/DefaultDependencyGraphBuilder.html), as that's what the `maven-dependency-plugin` uses in "verbose" mode, see https://github.com/apache/maven-dependency-plugin/commit/d9d34c6674db6c9a5c0786c95d0caf96bc9636fb.

sschuberth:
Another interesting Maven artifact resolver implementation to look at is https://github.com/square/maven-archeologist. It also points out https://github.com/google/guava/wiki/GraphsExplained which might be an alternative for our own dependency graph implementation.

PR #6782, opened 2023-03-31
Introduce a Gradle inspector

sschuberth:


sschuberth:
PS: There are no tests in here, as my original goal was to reuse the existing Gradle tests, but that turned out to be more complicated than I thought. Partly, because the GradleInspector yields more complete results.

plugins/package-managers/gradle-inspector/README.md
```diff
@@ -0,0 +1,45 @@
+# GradleInspector
+
+The [GradleInspector] is an alternative analyzer for projects that use the Gradle package manager. It is supposed to
+address [several][] [shortcomings][] of the "legacy" [Gradle] analyzer, but in order to not interfere with it, the
```
mnonnenmacher, line 4: Is the `[]` required here? You don't use it for the other links.

plugins/package-managers/gradle-inspector/README.md
```diff
@@ -0,0 +1,45 @@
...
+[GradleInspector] is disabled by default.
+
+## Usage
+
+As the [GradleInspector] is disabled by default, it needs to be enabled explicitly. It is recommended to disable the
+"legacy" [Gradle] analyzer at the same time:
+
+    ort analyze -P ort.analyzer.enabledPackageManagers=GradleInspector -P ort.analyzer.disabledPackageManagers=Gradle
```
mnonnenmacher, line 12: As there is currently no way to add a package manager to the list of enabled package manager, mabye give a hint that the full list needs to be provided, for example with `=GradleInspector[,...]`.

plugins/package-managers/gradle-inspector/README.md
```diff
@@ -0,0 +1,45 @@
...
+Due to some optimizations IntelliJ IDEA performs when building Gradle projects, it might be that bundling the fat-JAR
+for the [Gradle plugin] as a resource into the [GradleInspector] does not always work reliably. In that case ensure that
+IntelliJ IDEA has "Gradle" configured for the "Build and run using" and "Run tests using" settings, and / or try to
+run the `:plugins:package-managers:gradle-plugin:fatJar` and `:plugins:package-managers:gradle-inspector:jar` tasks
+manually once.
+
+## Debugging
+
+Due to this setup of the [GradleInspector], the [OrtModelBuilder] can actually be debugged. To do so, create a run
+configuration in in IntelliJ IDEA that runs `ort analyze` and sets the *VM options* to `-Dorg.gradle.debug=true`. Also,
```
mnonnenmacher, line 32: Double "in".

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -0,0 +1,250 @@
...
+            analyzerConfig: AnalyzerConfiguration,
+            repoConfig: RepositoryConfiguration
+        ) = GradleInspector(type, analysisRoot, analyzerConfig, repoConfig)
+    }
+
+    private fun extractInitScript(): File {
+        fun extractResource(name: String, target: File): File {
+            target.outputStream().use { outputStream ->
+                val resource = checkNotNull(javaClass.getResource(name)) {
+                    "Resource '$name' not found."
```
mnonnenmacher, line 112: Is this related to the IntelliJ issue you describe in the "Building" section of the README? If so, please add a hint here that this can be fixed by running those Gradle tasks.

plugins/package-managers/gradle-plugin/src/main/kotlin/MavenModelExtensions.kt
```diff
@@ -0,0 +1,78 @@
...
+package org.ossreviewtoolkit.plugins.packagemanagers.gradleplugin
+
+import OrtVcsModel
+
+import org.apache.maven.model.Model
+import org.apache.maven.model.Scm
+import org.apache.maven.model.building.ModelBuildingResult
+
+fun Model.collectAuthors(): Set<String> =
+    mutableSetOf<String>().apply {
```
mnonnenmacher, line 29: Use `buildSet` instead.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
mnonnenmacher, line 1: > Note that this approach will fail for private Maven repository even if Gradle is configured with the credentials, as this download happens directly via OkHttp.

Is this something that can be fixed later on? Currently we use the workaround to configure the credentials for Maven as well because the `MavenSupport`.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -286,6 +289,38 @@ private fun Collection<OrtDependency>.toPackageRefs(
...
+
+/**
+ * Split the provided [checksum] by whitespace and return a [Hash] for the first element that matches the provided
+ * algorithm. If no element matches, return [Hash.NONE]. This works around the issue that Maven checksum files sometimes
+ * contain arbitrary strings before or after the actual checksum.
+ */
+private fun parseChecksum(checksum: String, algorithm: String) =
+    checksum.splitOnWhitespace().firstNotNullOfOrNull {
+        runCatching { Hash.create(it, algorithm) }.getOrNull()
+    } ?: Hash.NONE
```
mnonnenmacher, line 322: For workarounds like this it would be nice if we could share the code with the Maven implementation somehow later on, because then any improvements will automatically be available for both package managers.

mnonnenmacher (changes requested):

plugins/package-managers/gradle-inspector/README.md
sschuberth: Yes, they're required here because the other links are not HTTP / external links.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
sschuberth: This was rather something I added during development, and to deal with a non-nullable resource value. In any case, I think it's probably not a good idea to put developer-hints into this user-facing message.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
sschuberth: > Is this something that can be fixed later on?

I hope so, but I currently don't have a good idea how.

> Currently we use the workaround to configure the credentials for Maven as well because the `MavenSupport`.

Correct. In the new approach, credentials would now need to be configured via `.netrc`. It did not make much of a difference to me whether credentials additionally need to be configured via Maven or `.netrc`. In general, I believe it's a good idea to decouple Gradle from Maven as much as we can, to avoid and confusion about the two.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
sschuberth: I agree. However, I was unsure yet where that code should go, as I actually also have plans to completely rewrite the Maven implementation (see #4720).

plugins/package-managers/gradle-inspector/README.md
sschuberth: Ah, true, I've reworded the text.

mnonnenmacher (approved):

plugins/package-managers/gradle-inspector/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+In contrast to the "legacy" [Gradle] analyzer which is fully implemented as an [initialization script], the
+[GradleInspector] only uses a minimal [init.gradle] to apply a [Gradle plugin], which in turn does nothing else than
+registering the [OrtModelBuilder] for the ORT-specific [data model for Gradle projects]. The [GradleInspector] then
+injects both the [init.gradle] and a fat-JAR for the [Gradle plugin] into the project to analyze.
+
+## Limitations
+
+The retrieval of the checksum values for remote artifacts is currently done via plain OkHttp calls, which means it will
+not work out of the box for private repositories. To worka round this, crednetials need to be configured in `.netrc`
```
fviernau, line 27: To worka round -> To work around

plugins/package-managers/gradle-inspector/README.md
```diff
@@ -0,0 +1,57 @@
...
+additionally to in Gradle. This is similar to how the "legacy" [Gradle] analyzer required to additionally configure
```
fviernau, line 28: crednetials -> credentials

plugins/package-managers/gradle-inspector/README.md
```diff
@@ -0,0 +1,57 @@
+# GradleInspector
+
+The [GradleInspector] is an alternative analyzer for projects that use the Gradle package manager. It is supposed to
```
fviernau, line 3: Consider dropping "The"

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -0,0 +1,250 @@
...
+    analyzerConfig: AnalyzerConfiguration,
+    repoConfig: RepositoryConfiguration
+) : PackageManager(name, analysisRoot, analyzerConfig, repoConfig) {
+    private companion object : Logging
+
+    class Factory : AbstractPackageManagerFactory<GradleInspector>("GradleInspector", isEnabledByDefault = false) {
+        // Gradle prefers Groovy ".gradle" files over Kotlin ".gradle.kts" files, but "build" files have to come before
+        // "settings" files as we should consider "settings" files only if the same directory does not also contain a
+        // "build" file.
+        override val globsForDefinitionFiles = GRADLE_BUILD_FILES + GRADLE_SETTINGS_FILES
```
fviernau, line 99: On the first glance I didn't get what the comment for `globsForDefinitionFiles` is trying to say. I believe it could become more obvious by droping `GRADLE_BUILD_FILES` and `GRADLE_SETTINGS_FILES` and have the comment next to one list of all globs. Either
`val GLOBS_FOR_DEFINITON_FILES` as a file level constant, or directly inline. What do you think?

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -0,0 +1,250 @@
...
+
+        override fun create(
+            analysisRoot: File,
+            analyzerConfig: AnalyzerConfiguration,
+            repoConfig: RepositoryConfiguration
+        ) = GradleInspector(type, analysisRoot, analyzerConfig, repoConfig)
+    }
+
+    private fun extractInitScript(): File {
+        fun extractResource(name: String, target: File): File {
```
fviernau, line 109: This could be turned into an expression function
```
fun extractResource(name: String, target: File): File =
   target.apply {
       ...
   }
```

fviernau: For readability of `extractInitScript()` I believe it would be good to move `extractResource()` outside of the class, as a file level private function.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -0,0 +1,250 @@
...
+
+/**
+ * The names of Gradle (Groovy, Kotlin script) settings files for a Gradle build.
+ */
+private val GRADLE_SETTINGS_FILES = listOf("settings.gradle", "settings.gradle.kts")
+
+/**
+ * The name of the option to specify the Gradle version.
+ */
+const val OPTION_GRADLE_VERSION = "gradleVersion"
```
fviernau, line 78: Would it make sense to mention the lowest Gradle version this is known to work with?

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.gradle.api.artifacts.result.UnresolvedDependencyResult
+import org.gradle.api.attributes.Attribute
+import org.gradle.api.internal.artifacts.result.ResolvedComponentResultInternal
+import org.gradle.api.logging.Logging
+import org.gradle.internal.deprecation.DeprecatableConfiguration
+import org.gradle.internal.resolve.ModuleVersionResolveException
+import org.gradle.tooling.provider.model.ToolingModelBuilder
+import org.gradle.util.GradleVersion
+
+class OrtModelBuilder : ToolingModelBuilder {
```
fviernau, line 52: Would it make sense to make this `internal` visible?

(Same for a couple of other classes which could be? implementation details)

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -0,0 +1,227 @@
...
+        )
+    }
+
+    private fun Configuration.isResolvable(): Boolean {
+        val canBeResolved = GradleVersion.current() < GradleVersion.version("3.3") || isCanBeResolved
+
+        val isDeprecatedConfiguration = GradleVersion.current() >= GradleVersion.version("6.0")
+                && this is DeprecatableConfiguration && resolutionAlternatives != null
+
+        return canBeResolved && !isDeprecatedConfiguration
```
fviernau, line 123: Am I right with assuming that a deprecated configuration is not resolvable?

fviernau (changes requested):

plugins/package-managers/gradle-inspector/README.md
sschuberth: I'd prefer to keep it.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
sschuberth: Note that this comment was copied as-is from `Gradle.kt`. And at least over there, `GRADLE_SETTINGS_FILES` is also used separately. Thus I'd prefer to keep the constants separate and aligned with `Gradle.kt`.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
sschuberth: I believe `extractInitScript()` is simple enough to bear with the added complexity of the nested `extractResource()`, and I also prefer the latter to stay nested as it's really only of use in this strict context.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
sschuberth: Probably, but it's currently unknown.

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
sschuberth: No, deprecated configurations *are* resolvable. That's why they need to be filtered out explicitly.

fviernau: I got this hint from the function name, do you agree function name is mis-leading?

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
fviernau: > Thus I'd prefer to keep the constants separate and aligned with Gradle.kt.

Ok, I was reviewing it under the assumption that `Gradle.kt` is gonna be deleted anyway.
If it makes sense for the time being I'm fine with that.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
fviernau: It's general code style. As I really find the nesting often times painful to read, I added a point to the dev meeting agenda to discuss this in the larger round.

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
sschuberth: Yes, I can see how to miss the notion between "canBeResolved" and "isResolvable". I'm open for proposals for better names for the extension function. Maybe "shouldResolve", "isAccepted" or "isEligible"?

sschuberth: I think I'll go for "isRelevent" for now, also in context of a minor refactoring.

fviernau: Hm, my only idea is `shouldBeIgnored()` which I'm not sure if I like.  
Maybe separate functions could do. `project.configurations.filter { it.isResolvable() && !it.isDeprecated() }`

I believe `shouldResolve()` would make it clearer to me compared to the current code.

fviernau: Maybe while at it, a comment could be added explaining _why_ deprecated configurations are ignored.

fviernau: > Isn't that obvious? Because the are deprecated (in favor of another configuration). 

Does the deprecation imply that the build script is not using that configuration to define it's dependencies?
Or could the build script still use deprecated configs?

PR #6791, opened 2023-04-03
Extend ScanSummary to be able to return findings from snippet scanners

nnobelis:
This pull request extends ORT Data Model to be able to capture snippet from snippets scanner such as FossID or ScanOSS.

Fixes: #3265.


model/src/main/kotlin/utils/SnippetFinding.kt
```diff
@@ -0,0 +1,63 @@
...
+ */
+data class SnippetFinding(
+    /**
+     * The author/vendor of the component the code snippet is commit from.
+     */
+    val author: String,
+    /**
+     * The artifact name of the component the code snippet is commit from.
+     */
+    val artifact: String,
```
tsteenbe, line 37: Instead of using author, artifact and version any found snippet should imo point to a [ORT Package])https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/kotlin/Package.kt) within ORT model this is how SBOM handle snippets see for example https://spdx.github.io/spdx-spec/v2.3/snippet-information/.

Also why does SnippetFinding Class not have location: TextLocation property to point to file path, start, and end line of the snippet?

tsteenbe (changes requested):

model/src/main/kotlin/utils/SnippetFinding.kt
nnobelis: Don't forget that `SnippetFinding` is an abstraction of snippets and it should be mappeable from the results of all snippet scanners supported by ORT. 
To build your ORT Package, you need information such as the identifier. ScanOSS gives your the purl so it could be doable, but FossID does no give you the type of the identifier.
You also need the VCS of the file the snippet is coming from but for ScanOSS, it is under the form  "https://osskb.org/api/file_contents/6ff2427335b985212c9b79dfa795799f" which is a link to their KB, not the original file.

> Also why does SnippetFinding Class not have location: TextLocation property to point to file path, start, and end line of the snippet?

As mentioned in the commit message, it will be added in the future. It's not trivial because, while ScanOSS provides this information, FossId returns it under the following form:
 ```
"highlighting": "{\"id\":\"410668b1f35f8b27ff9ce345998448b6\",\"local_coverage\":0.9754,\"local_highlight\":{\"blocks\":[{\"byte_range\":{\"begin\":0,\"end\":712},\"char_range\":{\"begin\":0,\"end\":712},\"id\":\"abdc2b929a1b84f24155c27b752944ab\"},{\"byte_range\":{\"begin\":1395,\"end\":28013},\"char_range\":{\"begin\":1395,\"end\":28013},\"id\":\"a11cb7ad7af8d915193131a92d514ed7\"}],\"encoding\":\"UTF-8\",\"id\":\"3673e848c2d349e2f054691c952b3f2f\",\"pfm_format\":2},\"local_size\":475,\"remote_coverage\":1,\"remote_highlight\":{\"blocks\":[{\"byte_range\":{\"begin\":0,\"end\":27333},\"char_range\":{\"begin\":0,\"end\":27333},\"id\":\"871d76314c0e746c1b33d63e6c05a909\"}],\"encoding\":\"UTF-8\",\"id\":\"410668b1f35f8b27ff9ce345998448b6\",\"pfm_format\":2},\"remote_size\":475}"
		}
```

They have an API to get the matching lines but it is sub-performing. We opened a ticket in the (private) tracker to have it addressed.

Comment (fviernau) [reply model/src/main/kotlin/utils/SnippetFinding.kt:37]: I believe that having the file location (scanned provenance) of the snippet finding is the most essential bit, as it is needed to make sense out of the finding. There it should be included in the initial MVP.

Regarding the way how the upstream snippet is specified `author, artifact, version, licenses, score`. How did you arrive at this one? Which snippet scanner did you consider when designing this model for?

Comment (nnobelis) [reply model/src/main/kotlin/utils/SnippetFinding.kt:37]: > @nnobelis Thank you for your clarifications, I think we should not base ORT model on what a specific tool currently provides but what we see as model that works with the rest of ORT. IMO Snippetfinding is should be start and end line and reference to a package.

As answered to @fviernau, in fact are two line ranges are needed: `sourceLocation` and `snippetLocation`.

> I am saying this as I would like to be able to curate snippet findings to resolve to package and be able to generate SBOM with snippets info.

I understand your need but in our current project, we do not need snippet results in a SBOM: we have a special reporter plugin for that.
Regarding the  possibility to curate snippet findings with `.ort.yml` (what we call "assessment" at Bosch) we are going to propose in some future PRs an `.ort.yml` model extension and an new ORT Evaluator. I think your concern can be addressed then ?

> 
> What do you thinking if we say introduce a in ORT Model a packageReference to hold some of the info you get from the snippet tools which can then be curated into a regular ORT Package?

For this I have unfortunately not enough experience with ORT data model. I would like @sschuberth, @mnonnenmacher, @fviernau opinion on this.

Comment (fviernau) [reply model/src/main/kotlin/utils/SnippetFinding.kt:37]: > I understand your need but in our current project, we do not need snippet results in a SBOM: we have a special reporter plugin for that.

@nnobelis to my understanding any `ScanSummary` is and (after the refactoring to store scan results by provenance also in ORT result) will be implicitly linked to at least one or multiple `Package`s. I guess this fullfills @tsteenbe request to have a link to a package.

Comment (nnobelis) [reply model/src/main/kotlin/utils/SnippetFinding.kt:37]: But I thought @tsteenbe wanted a package per snippet ?!

@tsteenbe : did you really meant a link to the package for which the snippets have been listed ?

**Comment by nnobelis** [comment]
@fviernau I have added the file location and the snippet location, as required. Please review.

File: model/src/main/kotlin/utils/SnippetFinding.kt (hunk 1 above)
Comment (tsteenbe) [reply model/src/main/kotlin/utils/SnippetFinding.kt:38]: > I understand your need but in our current project, we do not need snippet results in a SBOM: we have a special reporter plugin for that.

@nnobelis I prefer not to implement snippets in ORT that's just fulfill Bosch current project scope/needs but build something where we can see a path forwards that enable curating snippets and generate SBOM with snippets. As said for a snippet is a A) path to source (`sourceLocation`?), start and end line (snippetLocation/TextLocation?) and any other provenance info related to the package is should referenced - all of these snippet properties must be curatable at some point as no tool is perfect.

**Comment by sschuberth** [comment]
@oss-review-toolkit/core-devs I currently don't have time to review this. I'd prefer to not merge this until we had a chance to discuss this at least in a Kotlin developer meeting after the Easter vacation.

File: model/src/main/kotlin/utils/SnippetFinding.kt (hunk 1 above)
Comment (tsteenbe) [reply model/src/main/kotlin/utils/SnippetFinding.kt:38]: @fviernau and I had talked about this PR a bit as part of another meeting to align SnippetFinding with existing ORT model components such as LicenseFinding and CopyrightFinding we came up with the following for SnippetFinding. Further discussion is needed such as how to curate snippetFinding but hope the below give you a better understanding of thoughts/ideas.

```
SnippetFinding
- location: textLocation e.g. file path, start and end lines
- license: SpdxExpression e.g. license applicable to the license
- score: String optional/nullable score of a snippet finding
- UpstreamSnippet
  - location: textLocation e.g. file path, start and end lines
  - Provenance
    - vcsInfo
    - sourceArtifact
  - purls e.g. array of purls optional/nullable as not every code fragment is part of a release.
```

Please describe the business case for including author. Why do you need the author or are after copyright holder?

Comment (nnobelis) [reply model/src/main/kotlin/utils/SnippetFinding.kt:38]: > license: SpdxExpression e.g. license applicable to the license

This is strange that `license` does not belong to `UpstreamSnippet` as it is a part of it.
Additional, could their be multiple licenses per snippet ? At least ScanOSS seems to believe so:

```
"licenses": [
      {
        "name": "MIT",
        "patent_hints": "no",
        "copyleft": "no",
        "checklist_url": "https://www.osadl.org/fileadmin/checklists/unreflicenses/MIT.txt",
        "osadl_updated": "2023-03-26T02:11:00+00:00",
        "source": "component_declared",
        "url": "https://spdx.org/licenses/MIT.html"
      }
    ],
```

Regarding the provenance, I think you will run in a wall: ScanOSS for instance does not give any information about the source repository:
```
"source_hash": "bd4bff27f540f4f2c9de012acc4b48a3",
"file_url": "https://osskb.org/api/file_contents/6ff2427335b985212c9b79dfa795799f",
```

As for the purl, as said above FossID gives you the `author`, `artifact`, `version` but not package type, so you would have to build some heuristics to find it.

For the author, we didn't need it explicitly, it was just a way of identifying the snippet. But in the end, this author is in your purl.

Comment (tsteenbe) [reply model/src/main/kotlin/utils/SnippetFinding.kt:38]: > This is strange that license does not belong to UpstreamSnippet as it is a part of it.

Yes, on 2nd thought that makes more sense - improved version below:

```
SnippetFinding
- location: textLocation - file path, start and end lines
- score: String - score of a snippet finding, optional/nullable
- detectedSnippet
  - location: textLocation - file path, start and end lines
  - vcs: VcsInfoCurationData - nullable as not every snippet scanner contains this info
  - sourceArtifact: RemoteArtifact -  nullable as not every snippet scanner contains this info
  - purls: [string] e.g.  array of PURLs pointing to upstream package that include snippet, optional/nullable as not every code fragment is part of a release.
  - license: SPDXExpression e.g. license the snippetScanner thinks is applicable to the snippet
  - kb-url:
```

Also thought about curating snippet a bit more, below proposed SnippetFindingCuration is based on [LicenseFindingCuration.kt](https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/kotlin/config/LicenseFindingCuration.kt)
```
SnippetFindingCuration
- path: string - glob to match the file path of a snippet finding.
- startLines: List<Int> - A matcher for the start line of a snippet finding.
- lineCount: Int - A matcher for the line count of a snippet finding
- concludedLicense: SpdxExpression - the concluded license as SPDX expression or [org.ossreviewtoolkit.utils.spdx.SpdxConstants.NONE] for no license,
- detectedLicense: SpdxExpression - a matcher for the detected license of a snippet finding
- reason: SnippetFindingCurationReason - the reason why the curation was made, out of a predefined choice.
- comment: String - A comment explaining this [SnippetFindingCuration].
```
> Additional, could their be multiple licenses per snippet ?

license: SpdxExpression so you can take the license array provided by ScanOSS and turn it into expression LICENSE AND LICENSE AND ..

> Regarding the provenance, I think you will run in a wall: ScanOSS for instance does not give any information about the source repository

Actually source code repository is included in the ScanOSS output see https://github.com/oss-review-toolkit/ort/blob/881adaea1d055d5ab63f67abe65b887e2b85d48f/scanner/src/test/assets/scanoss-semver4j-3.1.0-with-snippet.json#L91, was able to translate the ScanOSS Response from [scanoss-semver4j-3.1.0-with-snippet.json](https://github.com/oss-review-toolkit/ort/blob/881adaea1d055d5ab63f67abe65b887e2b85d48f/scanner/src/test/assets/scanoss-semver4j-3.1.0-with-snippet.json) into proposed SnippetFinding

```
{
  location: {
    path: com/vdurmont/semver4j/Range.java,
    startLine: 1,
    endLine: 710
  },
  score: 100,
  detectedSnippet: {
    location: {
      path: src/main/java/com/vdurmont/semver4j/Requirement.java,
      startLine: 1,
      endLine: 710
    },
    purls: [
        "pkg:maven/org.graylog.repackaged/semver4j",
        "pkg:maven/org.graylog2/graylog2-server",
        "pkg:github/graylog2/graylog2-server",
        "pkg:maven/org.graylog/graylog-parent",
        "pkg:maven/org.graylog/graylog-project-parent",
        "pkg:maven/org.graylog.repackaged/os-platform-finder",
        "pkg:maven/org.graylog.telemetry/graylog-telemetry-plugin",
        "pkg:maven/org.graylog.plugins/graylog-plugin-parent",
        "pkg:maven/org.graylog.shaded/kafka09",
        "pkg:maven/org.graylog.shaded/elasticsearch5"
    ],
    vcs: {
       url: "https://github.com/vdurmont/semver4j"
       revision: "b92cd7e4d588747588d1b5fe1f8c664c"
    },
    sourceArtifact: null,
    license: Apache-2.0 AND BSD-2-Clause AND EPL-2.0 AND SSPL
  }
} 
```
Manual curation or generated curation provider for ScanOSS could then resolve purl pkg:maven/org.graylog.repackaged/semver4j into vcs or sourceArtifact info.

Comment (tsteenbe) [reply model/src/main/kotlin/utils/SnippetFinding.kt:38]: @nnobelis  Let do a call next week, ping me on Slack on April 12th or later to pick a day & time.

Comment (sschuberth) [reply model/src/main/kotlin/utils/SnippetFinding.kt:38]: > Let do a call next week, ping me on Slack on April 12th or later to pick a day & time.

@nnobelis already scheduled a meeting on April 19 on this topics. Let's just have a *single* meeting *together*.

**Comment by fviernau** [comment]
@nnobelis is this up for review? / should it be turned into draft until ready?

**Comment by nnobelis** [comment]
@fviernau It is ready for review

File: scanner/src/main/kotlin/scanners/fossid/FossId.kt [hunk 2]
```diff
@@ -27,6 +27,9 @@ import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 import kotlin.time.measureTimedValue
 
+import kotlinx.coroutines.Dispatchers
```
sschuberth, line 30: Commit message: Make the last sentence a separate paragraph, and start capitalized after a colon ("issues: *T*his").

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -746,7 +762,23 @@ class FossId internal constructor(
             "${pendingFiles.size} pending files have been returned for scan '$scanCode'."
         }
 
-        return RawResults(identifiedFiles, markedAsIdentifiedFiles, listIgnoredFiles, pendingFiles)
+        val snippets = runBlocking(Dispatchers.IO) {
+            pendingFiles.map {
+                logger.info { "Listing snippet for $it..." }
```
sschuberth, line 767: This will log immediately instead of when the coroutine starts. If this is not intended, move this to inside `async`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -786,4 +868,33 @@ class FossId internal constructor(
             mapOf(SCAN_CODE_KEY to scanCode, SCAN_ID_KEY to scanId, SERVER_URL_KEY to config.serverUrl)
         )
     }
+
+    /**
+     * Return the [PurlType] as determined from the given [url], or null if there is no match. An issue will be added to
+     * [issues] in this case.
+     */
+    private fun urlToPackageType(url: String, issues: MutableList<Issue>): PurlType? = PackageProvider.get(url)?.let {
+        when (it) {
```
sschuberth, line 877: Instead use:

    private fun urlToPackageType(url: String, issues: MutableList<Issue>): PurlType? =
        when (val provider = PackageProvider.get(url)) {

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -786,4 +868,33 @@ class FossId internal constructor(
...
+            PackageProvider.NPM_JS -> PurlType.NPM
+            PackageProvider.NUGET -> PurlType.NUGET
+            PackageProvider.PACKAGIST -> PurlType.COMPOSER
+            PackageProvider.PYPI -> PurlType.PYPI
+            PackageProvider.RUBYGEMS -> PurlType.GEM
+
+            else -> {
+                issues += FossId.createAndLogIssue(
+                    source = "FossId",
+                    message = "Cannot determine PURL type for url '$url' and provider '$it'."
```
sschuberth, line 894: Instead use:

                    message = "Cannot determine PURL type for url '$url' and provider '$provider'."

scanner/src/main/kotlin/scanners/scanoss/ScanOssResultParser.kt
```diff
@@ -23,11 +23,17 @@ import java.io.File
 import java.time.Instant
 
 import org.ossreviewtoolkit.clients.scanoss.FullScanResponse
+import org.ossreviewtoolkit.clients.scanoss.model.IdentificationType
```
sschuberth, line 26: Commit message: Start with "Map the snippets ..."

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -214,7 +214,7 @@ class FossId internal constructor(
         endTime: Instant = Instant.now(),
         issue: Issue
     ) = ScanSummary(
-        startTime, endTime, "", sortedSetOf(), sortedSetOf(), listOf(issue)
+        startTime, endTime, "", sortedSetOf(), sortedSetOf(), issues = listOf(issue)
```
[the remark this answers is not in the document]
nnobelis: Using named parameter is required because the signature of  `ScanSummary` has changed.
Otherwise the compiler throws:
`Type mismatch: inferred type is List<Issue> but SortedSet<SnippetFinding> was expected`

model/src/main/kotlin/utils/Snippet.kt
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * The license of the component the code snippet is commit from.
+     */
+    val licenses: SpdxExpression,
+) : Comparable<Snippet> {
+    companion object {
+        private val COMPARATOR = compareBy<Snippet> { it.purl }
+    }
+
+    override fun compareTo(other: Snippet) = COMPARATOR.compare(this, other)
```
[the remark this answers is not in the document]
nnobelis: You wrote `in the containing class` but it seems `SortedSetConverters.kt` is the place of these converters ?

model/src/main/kotlin/ScanSummary.kt
```diff
@@ -29,6 +29,7 @@ import java.util.SortedSet
 
 import org.ossreviewtoolkit.model.config.LicenseFilePatterns
 import org.ossreviewtoolkit.model.utils.RootLicenseMatcher
+import org.ossreviewtoolkit.model.utils.SnippetFinding
```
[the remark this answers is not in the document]
nnobelis: > I believe this sentence is misleading. It sounds as if you're trying to argue that snippet findings are different from license / copyright findings as the former require human interaction. But that's true for the latter as well; detected licenses should be double-check for false-positives, too.

Yes, true.

> 
> What are you trying to say? Are you trying to find arguments why snippets cannot be modeled via the existing license / copyright findings classes?

This is what I was trying to say.  I tried another sentence. Please review and suggest if not satisfying.

model/src/main/kotlin/utils/Extensions.kt
```diff
@@ -143,6 +143,8 @@ enum class PurlType(private val value: String) {
     DEBIAN("deb"),
     DRUPAL("drupal"),
     GEM("gem"),
+    GITHUB("github"),
```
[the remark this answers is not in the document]
sschuberth: > This allows mapping those Purl types from the corresponding `PackageProvider`s.

This still does not explain why it's exactly *these two*. A proper explanation would be that of of the snippet scanner supports these (Which scanner?), so we need to be prepared to handle them.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -746,7 +762,23 @@ class FossId internal constructor(
...
         }
 
-        return RawResults(identifiedFiles, markedAsIdentifiedFiles, listIgnoredFiles, pendingFiles)
+        val snippets = runBlocking(Dispatchers.IO) {
+            pendingFiles.map {
+                async {
+                    logger.info { "Listing snippet for $it..." }
+                    val snippetResponse = service.listSnippets(config.user, config.apiKey, scanCode, it)
+                        .checkResponse("list snippets")
+                    val snippets = requireNotNull(snippetResponse.data) {
```
sschuberth, line 771: `require(NotNull)` should be used to verify arguments / user input. Use `check(NotNull)` to verify state.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -760,10 +792,61 @@ class FossId internal constructor(
...
+                        issues += FossId.createAndLogIssue(
+                            source = "FossId",
+                            message = "Failed to parse license '$it' as an SPDX expression:" +
+                                    " ${spdxException.collectMessages()}"
+                        )
+                    }.getOrNull()
+                } ?: SpdxConstants.NOASSERTION.toSpdx()
+
+                // FossID does not return the hash of the remote artifact: it returns instead, in the property
+                // "match_file_id", the MD5 hash of the matched file IN the remote artifact.
```
sschuberth, line 822: Use:

> FossID does not return the hash of the remote artifact. Instead, it returns the MD5 hash of the matched file in the remote artifact as part of the "match_file_id" property.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -760,10 +792,61 @@ class FossId internal constructor(
...
+                val snippetProvenance = it.url?.let { url ->
+                    ArtifactProvenance(RemoteArtifact(url, Hash.NONE))
+                } ?: UnknownProvenance
```
sschuberth, line 825: Related to my earlier question: How can matched snippet ever *not* have a URL?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -760,10 +792,61 @@ class FossId internal constructor(
...
+                val purlType = it.url?.let { url -> urlToPackageType(url, issues)?.toString() } ?: "generic"
+
+                // TODO: FossId doesn't return the line numbers of the match, only the character range. One must use
+                //  another call "getMatchedLine" to retrieve the matched line numbers. Unfortunately, this is a call
+                //  per snippet which is too expensive. When it is available for a batch of snippets, it can be used
+                //  here.
```
sschuberth, line 831: Wrong indentation: Align to the word "FossID" in the first line (and consistently spell it like "FossID", not "FossId", in free text).

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -786,4 +870,33 @@ class FossId internal constructor(
...
+            PackageProvider.GITHUB -> PurlType.GITHUB
+            PackageProvider.GITLAB -> PurlType.GITLAB
+            PackageProvider.GOLANG -> PurlType.GOLANG
+            PackageProvider.MAVEN_CENTRAL, PackageProvider.MAVEN_GOOGLE -> PurlType.MAVEN
+            PackageProvider.NPM_JS -> PurlType.NPM
+            PackageProvider.NUGET -> PurlType.NUGET
+            PackageProvider.PACKAGIST -> PurlType.COMPOSER
+            PackageProvider.PYPI -> PurlType.PYPI
+            PackageProvider.RUBYGEMS -> PurlType.GEM
+            null -> null
```
sschuberth, line 892: Can't we just drop this and rely on the `else` below?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -786,4 +870,33 @@ class FossId internal constructor(
             mapOf(SCAN_CODE_KEY to scanCode, SCAN_ID_KEY to scanId, SERVER_URL_KEY to config.serverUrl)
         )
     }
+
+    /**
+     * Return the [PurlType] as determined from the given [url], or null if there is no match. An issue will be added to
+     * [issues] in this case.
```
sschuberth, line 876: Do not make the last sentence separate. Otherwise it sounds as if an issue is added in any case. Just continue the sentence instead:

> ..., in which case an issue will be added...

sschuberth (changes requested):
@mnonnenmacher and @fviernau, IMO this is close to ready for being merged. Please also have a look.

model/src/main/kotlin/utils/Snippet.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * The text location in the snippet that has matched.
+     */
+    val location: TextLocation,
+
+    /**
+     * The provenance of the snippet, either an artifact or a repository.
+     */
+    val provenance: Provenance,
```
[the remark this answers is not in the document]
nnobelis: `UnknownProvenance` can happen when FossID `url` property is `null`: see https://github.com/bosch-io/oss-review-toolkit/blob/cbc984df1bf131a80c32cea8a224cbec9d22f86c/scanner/src/main/kotlin/scanners/fossid/FossId.kt#L825.

Once again we are paying the price of the lack of data model for FossID....

However, in the various responses I get from the FossID API service, it *seems* the url is never `null` so maybe we were too caution making the property nullable in the first place. 
What do you think ? Should we make the property non nullable (in a separte commit) and then replace the  `UnknownProvenance` by a `KnownProvenance` ?

model/src/main/kotlin/ScanSummary.kt
```diff
@@ -66,6 +69,13 @@ data class ScanSummary(
     @JsonProperty("copyrights")
     val copyrightFindings: SortedSet<CopyrightFinding>,
 
+    /**
+     * The detected snippet findings.
+     */
+    @JsonProperty("snippets")
+    @JsonSerialize(converter = SnippetFindingSortedSetConverter::class)
+    val snippetFindings: Set<SnippetFinding> = setOf(),
```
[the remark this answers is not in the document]
nnobelis: I think the reasoning is that `licenseFindings ` and `copyrightFindings` are filled by all scanners whereas `snippetFindings` is only for the snippet scanners. If we don't put a default value, then we need to pass an empty set in several locations (~9 matches).

Therefore I would be in favor of keeping the default value (with `emptySet()`).

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: See my comment above.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: I added that because the tests were broken. As they use pseudo URLs, `PackageProvider.get` returned `null` and an issue was filled.
However I think filling an issue when the URL cannot be mapped is good practice so I will fix the tests instead.

model/src/main/kotlin/utils/Snippet.kt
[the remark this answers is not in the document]
sschuberth: What I meant was: If you drop comparators in favor of converters, then the converters should not go to the class that contained the comparator, but to the class that uses the class that used to contain comparators.

sschuberth (dismissed):

scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
```diff
@@ -130,6 +133,27 @@ data class NestedProvenanceScanResult(
...
+    private fun Map<KnownProvenance, List<ScanResult>>.mergeSnippetFindings(): Set<SnippetFinding> {
+        val findingsByPath = mapKeys { getPath(it.key) }.mapValues { (_, scanResults) ->
+            mutableSetOf<SnippetFinding>().also { acc ->
+                scanResults.forEach { acc += it.summary.snippetFindings }
+            }
+        }
+
+        val allFindings = mutableSetOf<SnippetFinding>()
+
+        findingsByPath.map { (path, findings) ->
```
mnonnenmacher, line 145: Use `forEach` instead as you don't use the result. You could also use `buildSet`:
```kotlin
return buildSet {
  findingsByPath.forEach { (path, findings) ->
    ...
    addAll(findings.map { ... })
}
```

scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
```diff
@@ -130,6 +133,27 @@ data class NestedProvenanceScanResult(
...
+            val prefix = if (path.isEmpty()) path else "$path/"
+
+            allFindings += findings.map { finding ->
+                val newPath = "$prefix${finding.sourceLocation.path}"
+                finding.copy(sourceLocation = finding.sourceLocation.copy(path = newPath))
+            }.toSet()
```
mnonnenmacher, line 151: Remove `toSet()` because there is no need to convert to set before adding the collection to a set.

model/src/main/kotlin/utils/Extensions.kt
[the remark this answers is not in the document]
mnonnenmacher: > Spell it like "GitHub" and "GitLab".

You still use "github".

mnonnenmacher (changes requested):

model/src/main/kotlin/utils/Snippet.kt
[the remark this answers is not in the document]
nnobelis: Yes, I opened a FossID support ticket 👍

mnonnenmacher (approved):

PR #6793, opened 2023-04-03
refactor(PdfTemplateReporter): Make setting attribute values more compact

sschuberth:
Also shorten some variable names as these are not full attributes but only their values, and use `toString()` instead of `path` to better express that a string is needed here (although both calls internally do the same thing).

### DIFF
--- a/reporter/src/main/kotlin/reporters/freemarker/asciidoc/PdfTemplateReporter.kt
+++ b/reporter/src/main/kotlin/reporters/freemarker/asciidoc/PdfTemplateReporter.kt
@@ -62,12 +62,10 @@ class PdfTemplateReporter : AsciiDocTemplateReporter("pdf", "PdfTemplate") {
 
     override fun processTemplateOptions(outputDir: File, options: MutableMap<String, String>): Attributes =
         Attributes.builder().apply {
-            val pdfThemeAttribute = options.remove(OPTION_PDF_THEME_FILE)?.let {
-                val pdfThemeFile = File(it).absoluteFile
-
-                require(pdfThemeFile.isFile) { "Could not find PDF theme file at '$pdfThemeFile'." }
-
-                pdfThemeFile.path
+            val pdfTheme = options.remove(OPTION_PDF_THEME_FILE)?.let { option ->
+                File(option).absoluteFile.also {
+                    require(it.isFile) { "Could not find PDF theme file at '$it'." }
+                }.path
             } ?: run {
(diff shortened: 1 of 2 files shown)


reporter/src/main/kotlin/reporters/freemarker/asciidoc/PdfTemplateReporter.kt
```diff
@@ -62,12 +62,10 @@ class PdfTemplateReporter : AsciiDocTemplateReporter("pdf", "PdfTemplate") {
...
-            val pdfThemeAttribute = options.remove(OPTION_PDF_THEME_FILE)?.let {
-                val pdfThemeFile = File(it).absoluteFile
-
-                require(pdfThemeFile.isFile) { "Could not find PDF theme file at '$pdfThemeFile'." }
-
-                pdfThemeFile.path
+            val pdfTheme = options.remove(OPTION_PDF_THEME_FILE)?.let { option ->
+                File(option).absoluteFile.also {
+                    require(it.isFile) { "Could not find PDF theme file at '$it'." }
+                }.toString()
```
mnonnenmacher, line 68: Why do you switch from `path` to `toString()`?

sschuberth: It does internally the same thing, but `toString()` expressed better that we *need* a string representation here, as the other branch also returns a string.

mnonnenmacher: Hm, to me using `path` makes more sense because it is defined as the string representation of the path, while `toString()` is a string representation which just coincidentally happens to be the value we are looking for.

sschuberth: Another reason why I preferred `toString()` is that using `path` somewhat raises the question why not `absolutePath` instead. The answer is: Because the `File` was already made absolute above, so `path` is sufficient here. I wanted to avoid that deliberation by using the "neutral" `toString()` instead. But if you have a clear preference for `path`still, I can change it back.

sschuberth: > I can change it back.

Done.

sschuberth:
Note that I've added a few more commits @mnonnenmacher.

mnonnenmacher (approved):

PR #6795, opened 2023-04-03
Move Docker related files

mnonnenmacher:


Dockerfile
```diff
@@ -28,7 +28,7 @@ ENV LANGUAGE=en_US:en
 ENV LC_ALL=en_US.UTF-8
 
 # Check and set apt proxy
-COPY docker/set_apt_proxy.sh /etc/scripts/set_apt_proxy.sh
+COPY scripts/set_apt_proxy.sh /etc/scripts/set_apt_proxy.sh
```
sschuberth, line 31: Commit message nit: I'd probably drop the "used by `Dockerfile`" part, as the script should be collected there independent of their use.

Dockerfile-legacy
sschuberth, line 1: `.dockerignore` needs to be updated as well.

sschuberth (changes requested):

sschuberth (approved):
Note that this is a breaking change for users that rely on the location / name of the legacy `Dockerfile` in the cloned repository.

PR #6804, opened 2023-04-04
helper-cli: Optionally skip writing empty package configurations

fviernau:
When doing bulk creation of package configurations like with ´orthw`, it is sometimes desired to only generate non-empty files.


helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
sschuberth, line 1: Commit message: The character in front of "orthw" is not a backtick.

helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
```diff
@@ -110,6 +110,11 @@ internal class CreateCommand : CliktCommand(
                 "files or directories which only contain non-offending licenses."
     ).split(",").default(emptyList())
 
+    private val skipEmpty by option(
+        "--skip-empty",
```
sschuberth, line 114: I'm wondering whether it's really worth introducing an option for this, or whether empty files should never be written out. Could you elaborate on the use-case when you actually *want* empty files?

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt
fviernau: With "empty" I was referring to a package configuration without any configuration entries, only the matcher bit is in place. It creates the directory structure plus the file with the header already setup. This is helpful from my experience, see below example:
```
orthw pc-create Maven:com.squareup.retrofit2:converter-gson:2.9.0 
Wrote a package configuration to '/home/frank/devel/ort/ort-config/package-configurations/Maven/com.squareup.retrofit2/converter-gson/2.9.0/vcs.yml'.
# ~/scans/dia/android$ cat /home/frank/devel/ort/ort-config/package-configurations/Maven/com.squareup.retrofit2/converter-gson/2.9.0/vcs.yml
---
id: "Maven:com.squareup.retrofit2:converter-gson:2.9.0"
vcs:
  type: "Git"
  url: "https://github.com/square/retrofit.git"
  revision: "bd33a5da186aa6f5365e78e27eb0292b1b8b1bff"
```

sschuberth: Would it make sense to give the feature a more fitting name then, and maybe inverse the logic by calling it "--write-templates", "--write-skeletons", "--always-create-structure" or something similar?

sschuberth:
PS: The *commit-lint* check failed.

PR #6805, opened 2023-04-04
feat(model): Re-filter scan summary when VCS path was narrowed down

fviernau:
After adding a package curation which narrows down the VCS path of a package, one has to re-run the scanner to see the impact of that curation on the detected licenses. If the new VCS path is a subdirectory of the old VCS path a re-scan is not necessary, because the scan result for the corresponding package does cover the curated VCS path.

Filter the scan result (again) in the `DefaultLicenseInfoProvider` so that the effect of the changes can be seen by only re-running the evaluator. This reduces the turn around time for seeing the effect of changes when using workflows similar to `orthw`.

### DIFF
--- a/model/src/main/kotlin/licenses/DefaultLicenseInfoProvider.kt
+++ b/model/src/main/kotlin/licenses/DefaultLicenseInfoProvider.kt
@@ -19,12 +19,15 @@
 
 package org.ossreviewtoolkit.model.licenses
 
+import java.io.File
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
 
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.RepositoryProvenance
+import org.ossreviewtoolkit.model.ScanResult
 import org.ossreviewtoolkit.model.config.LicenseFindingCuration
 import org.ossreviewtoolkit.model.config.PathExclude
 import org.ossreviewtoolkit.model.utils.PackageConfigurationProvider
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/licenses/DefaultLicenseInfoProvider.kt
```diff
@@ -82,7 +83,19 @@ class DefaultLicenseInfoProvider(
...
 
-        ortResult.getScanResultsForId(id).forEach { (provenance, _, summary, _) ->
+        ortResult.getScanResultsForId(id).forEach { scanResult ->
+            var summary = scanResult.summary
+            var provenance = scanResult.provenance
+
+            if (provenance is RepositoryProvenance) {
+                val vcsPath = ortResult.getPackage(id)?.metadata?.vcsProcessed?.path?.orEmpty()!!
+
+                if ("$vcsPath/".startsWith("${provenance.vcsInfo.path}/")) {
```
mnonnenmacher, line 93: Add `vcsPath != provenance.vcsInfo.path && ...` to avoid unnecessary filtering if the path did not change.

model/src/main/kotlin/licenses/DefaultLicenseInfoProvider.kt
```diff
@@ -82,7 +83,19 @@ class DefaultLicenseInfoProvider(
     private fun createDetectedLicenseInfo(id: Identifier): DetectedLicenseInfo {
         val findings = mutableListOf<Findings>()
 
-        ortResult.getScanResultsForId(id).forEach { (provenance, _, summary, _) ->
+        ortResult.getScanResultsForId(id).forEach { scanResult ->
+            var summary = scanResult.summary
+            var provenance = scanResult.provenance
+
+            if (provenance is RepositoryProvenance) {
```
mnonnenmacher, line 90: Please add a code comment explaining under which circumstances the re-filtering is meaningful.

mnonnenmacher (changes requested):

model/src/main/kotlin/licenses/DefaultLicenseInfoProvider.kt
mnonnenmacher: Also, shouldn't this append "/" only if the paths do not already end with a "/"?

model/src/main/kotlin/licenses/DefaultLicenseInfoProvider.kt
mnonnenmacher, line 1: A test case would be nice.

mnonnenmacher (approved):

PR #6807, opened 2023-04-05
fix(VulnerableCode): Handle invalid URLs in references

oheger-bosch:
There are some packages for which VulnerableCode returns vulnerabilities with URL references that cannot be converted to URIs, since they contain invalid characters. In such cases, no meaningful advisor result was generated. Prevent this by handling exceptions when converting to URIs.

### DIFF
--- a/advisor/src/main/kotlin/advisors/VulnerableCode.kt
+++ b/advisor/src/main/kotlin/advisors/VulnerableCode.kt
@@ -22,6 +22,8 @@ package org.ossreviewtoolkit.advisor.advisors
 import java.net.URI
 import java.time.Instant
 
+import org.apache.logging.log4j.kotlin.Logging
+
 import org.ossreviewtoolkit.advisor.AbstractAdviceProviderFactory
 import org.ossreviewtoolkit.advisor.AdviceProvider
 import org.ossreviewtoolkit.clients.vulnerablecode.VulnerableCodeService
@@ -30,11 +32,14 @@ import org.ossreviewtoolkit.model.AdvisorCapability
 import org.ossreviewtoolkit.model.AdvisorDetails
 import org.ossreviewtoolkit.model.AdvisorResult
 import org.ossreviewtoolkit.model.AdvisorSummary
+import org.ossreviewtoolkit.model.Issue
 import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.model.Vulnerability
 import org.ossreviewtoolkit.model.VulnerabilityReference
 import org.ossreviewtoolkit.model.config.AdvisorConfiguration
 import org.ossreviewtoolkit.model.config.VulnerableCodeConfiguration
+import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.utils.common.enumSetOf
 import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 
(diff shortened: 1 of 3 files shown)


sschuberth:
> There are some packages for which VulnerableCode returns vulnerabilities with URL references that cannot be converted to URIs, since they contain invalid characters

Please also report this to https://github.com/nexB/vulnerablecode/issues if not already done so.

advisor/src/main/kotlin/advisors/VulnerableCode.kt
```diff
@@ -112,11 +115,19 @@ class VulnerableCode(name: String, config: VulnerableCodeConfiguration) : Advice
...
         return scores.map { VulnerabilityReference(sourceUri, it.scoringSystem, it.value) }
     }
 
+    /**
+     * Return the URL for this reference as a URI dealing gracefully with invalid URIs. It occurred that some of the
+     * URLs returned by VulnerableCode were rejected by Java's URI parser. To prevent that in such cases the whole
+     * advisor fails, handle those URLs by returning them with a special scheme and URL-encoded.
+     */
+    private fun VulnerableCodeService.VulnerabilityReference.safeUri(): URI =
+        url.toUri().getOrDefault(URI("unparsed:${URLEncoder.encode(url, StandardCharsets.UTF_8)}"))
```
sschuberth, line 129: I wonder whether having such an "unparsed" URI is of any value, or whether we rather should create a warning issue / log for such things. I currently have a tendency for the latter. What do you think @oheger-bosch?

oheger-bosch:
> 
> Please also report this to https://github.com/nexB/vulnerablecode/issues if not already done so.

Created this issue: https://github.com/nexB/vulnerablecode/issues/1173

advisor/src/main/kotlin/advisors/VulnerableCode.kt
oheger-bosch: The problem I see here is that the vulnerability reference could also be associated with scores and severities. Consider the case that there is a reference indicating a high severity, but with an invalid URL. Wouldn't it be better than if the reference is still visible in the reports where it is expected? But since the URL is not nullable, it must contain something. The hope is that a human reader can at least get an impression from where this reference comes from.

sschuberth: > Wouldn't it be better than if the reference is still visible in the reports where it is expected?

To be honest, I don't necessarily think so. First of all, personally I do not care much about the references at all; all I want to know is if there *is* a vulnerability, and I do regard all sources that ORT's advisors use as trustful anyway.

> The hope is that a human reader can at least get an impression from where this reference comes from.

With the escaping happening, a human reading is going to have a hard time making sense of the URI anyway. So I'd still prefer the issue-approach.

sschuberth: Another idea could be to fixup the URI before parsing so it becomes valid. Is there a pattern behind what makes these URIs invalid?

oheger-bosch: The solution with an issue is also fine with me, I can try this out.
I am not sure whether we can identify a general pattern. In theory, the URI conversion could always fail.

oheger-bosch: Changed the implementation to create an issue.

advisor/src/main/kotlin/advisors/VulnerableCode.kt
```diff
@@ -49,6 +54,8 @@ private const val BULK_REQUEST_SIZE = 100
  * [VulnerableCode][https://github.com/nexB/vulnerablecode] instance.
  */
 class VulnerableCode(name: String, config: VulnerableCodeConfiguration) : AdviceProvider(name) {
+    companion object : Logging
```
sschuberth, line 57: Please make this `private` to not leak `Logging` into the API.

sschuberth (changes requested):

sschuberth (approved):

advisor/src/main/kotlin/advisors/VulnerableCode.kt
oheger-bosch: BTW, it took me a while to figure out that this declaration is needed. If it is missing, you get a strange internal Kotlin compiler error.

PR #6808, opened 2023-04-05
Ignore all PHP composer platform dependencies

sschuberth:

### DIFF
--- a/plugins/package-managers/composer/src/main/kotlin/Composer.kt
+++ b/plugins/package-managers/composer/src/main/kotlin/Composer.kt
@@ -62,12 +62,6 @@ import org.semver4j.Semver
 const val COMPOSER_PHAR_BINARY = "composer.phar"
 const val COMPOSER_LOCK_FILE = "composer.lock"
 
-private val EXCLUDED_PACKAGES = setOf(
-    "php", // Exclude the PHP runtime itself.
-    "composer-plugin-api", // Exclude the package to specify the supported composer plugin versions.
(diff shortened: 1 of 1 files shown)


plugins/package-managers/composer/src/main/kotlin/Composer.kt
```diff
@@ -302,6 +294,16 @@ class Composer(
...
 
+/**
+ * Return whether this String denotes a type of platform dependency, see
+ * https://getcomposer.org/doc/articles/composer-platform-dependencies.md#different-types-of-platform-packages.
+ */
+private fun String.isPlatformDependency(): Boolean =
+    this in (COMPOSER_PLATFORM_TYPES + PHP_PLATFORM_TYPES) || startsWith("ext-") || startsWith("lib-")
+
+private val COMPOSER_PLATFORM_TYPES = listOf("composer", "composer-plugin-api", "composer-runtime-api")
+private val PHP_PLATFORM_TYPES = listOf("php", "php-64bit", "php-ipv6", "php-zts", "php-debug")
```
fviernau, line 305: Should the lists be ordered alphabetically?

fviernau (changes requested):

plugins/package-managers/composer/src/main/kotlin/Composer.kt
sschuberth: I was copying these as-is from the linked page. When comparing with that, it probably makes sense to maintain the order. However, I've made these sets now to indicate that order does not matter.

fviernau (approved):

PR #6811, opened 2023-04-06
Gradle plugin refactorings

sschuberth:

### DIFF
--- a/plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
+++ b/plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
@@ -42,7 +42,6 @@ import org.gradle.api.internal.artifacts.DefaultModuleIdentifier
 import org.gradle.api.internal.artifacts.result.ResolvedComponentResultInternal
 import org.gradle.api.logging.Logging
 import org.gradle.internal.component.external.model.DefaultModuleComponentIdentifier
-import org.gradle.internal.deprecation.DeprecatableConfiguration
 import org.gradle.internal.resolve.ModuleVersionResolveException
 import org.gradle.maven.MavenModule
 import org.gradle.maven.MavenPomArtifact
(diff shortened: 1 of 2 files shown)


plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -173,6 +170,9 @@ internal class OrtModelBuilder : ToolingModelBuilder {
         }.filterIsInstance<ResolvedArtifactResult>()
     }
 
+    private fun AttributeContainer.getValueByName(name: String): Any? =
```
fviernau, line 173: Could be move out of class.

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -173,6 +170,9 @@ internal class OrtModelBuilder : ToolingModelBuilder {
...
+        attributes.keySet().find { it.name == name }?.let { getAttribute(it) }
+
```
fviernau, line 175: Is it not possible to do `getAttribute<Any>("key")`?

sschuberth: No, as the key into the container is the attribute (which consist of a name and a class type), but just the name. Which means you usually need to know the type of the attribute before you can look up its value. The whole point of the function is to be able to look up an attribute's value irrespective of the type, just by its name, as there sometimes are inconsistencies in the type an attribute uses (see the deleted comment, which is now not relevant anymore).

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
sschuberth: I've added a commit that introduces a new file to bundle these extension functions.

fviernau (approved):

PR #6814, opened 2023-04-06
Let `GradleInspector` filter all dependencies metadata configurations

sschuberth:

### DIFF
--- a/plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
+++ b/plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
@@ -44,7 +44,10 @@ internal fun Configuration.isRelevant(): Boolean {
     val isDeprecatedConfiguration = GradleVersion.current() >= GradleVersion.version("6.0")
             && this is DeprecatableConfiguration && resolutionAlternatives != null
 
-    return canBeResolved && !isDeprecatedConfiguration
+    // Do not try to resolve dependencies metadata configurations as there often cause failures with Gradle itself.
+    val isDependenciesMetadata = name.endsWith("DependenciesMetadata")
+
+    return canBeResolved && !isDeprecatedConfiguration && !isDependenciesMetadata
 }
 
 /**
--- a/plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
+++ b/plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
@@ -36,8 +36,11 @@ internal fun AttributeContainer.getValueByName(name: String): Any? =
  * Return whether this Gradle configuration is relevant for ORT's dependency resolution.
  */
 internal fun Configuration.isRelevant(): Boolean {
+    // Configurations can be resolved and / or consumed, but not neither, see the table at
+    // https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:resolvable-consumable-configs
     val canBeResolved = GradleVersion.current() < GradleVersion.version("3.3") || isCanBeResolved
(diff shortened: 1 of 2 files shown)


plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
```diff
@@ -44,7 +44,10 @@ internal fun Configuration.isRelevant(): Boolean {
     val isDeprecatedConfiguration = GradleVersion.current() >= GradleVersion.version("6.0")
             && this is DeprecatableConfiguration && resolutionAlternatives != null
 
-    return canBeResolved && !isDeprecatedConfiguration
+    // Do not try to resolve dependencies metadata configurations as there often cause failures with Gradle itself.
+    val isDependenciesMetadata = name.endsWith("DependenciesMetadata")
```
fviernau, line 48: Does this string always have a prefix or could the full config name also be "DependenciesMetadata" ? (lower first char)

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -65,18 +65,7 @@ internal class OrtModelBuilder : ToolingModelBuilder {
...
-                // higher, see [1]. Also see [2] for valuable information in this context.
-                //
-                // [1]: https://developer.android.com/studio/past-releases/past-agp-releases/agp-4-0-0-release-notes#dependency-metadata
-                // [2]: https://gist.github.com/h0tk3y/41c73d1f822378f52f1e6cce8dcf56aa#orgjetbrainskotlinplatformtype
-                filterNot { it.name.endsWith("DependenciesMetadata") }
-            } else {
-                this
-            }
-        }
+        val resolvableConfigurations = project.configurations.filter { it.isRelevant() }
```
fviernau, line 68: out of scope: Maybe reame resolvableConfigurations to relevantConfigurations

plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
```diff
@@ -44,7 +44,10 @@ internal fun Configuration.isRelevant(): Boolean {
...
+
```
fviernau, line 49: could we include a reference in the code comment to what this "*DependenciesMetadata" is about, and maybe also why it is ok to filter it out?

plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
sschuberth: There is always a prefix.

fviernau (approved):

plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
sschuberth: It's actually quite hard to find information about this in other places than those already given in the commit message (of the original commit that introduced the check for Android). But I've added those links here, too, for completeness.

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
sschuberth: Done in a new commit.

plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
fviernau: Thanks!

PR #6822, opened 2023-04-11
Improvements to patching results

sschuberth:

### DIFF
--- a/utils/test/src/main/kotlin/Utils.kt
+++ b/utils/test/src/main/kotlin/Utils.kt
@@ -80,8 +80,8 @@ fun patchExpectedResult(
         putAll(custom)
     }
 
-    return replacements.entries.fold(expectedResultFile.readText()) { text, entry ->
-        text.replace(entry.key, entry.value)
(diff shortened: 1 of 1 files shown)


utils/test/src/main/kotlin/Utils.kt
```diff
@@ -80,8 +80,8 @@ fun patchExpectedResult(
         putAll(custom)
     }
 
-    return replacements.entries.fold(expectedResultFile.readText()) { text, entry ->
-        text.replace(entry.key, entry.value)
+    return replacements.entries.fold(expectedResultFile.readText()) { text, (oldValue, newValue) ->
```
fviernau, line 88: Would it improve to name these `regex` and `replacement`, like the parameter of `String.replace()` ?

fviernau (approved):

utils/test/src/main/kotlin/Utils.kt
sschuberth: This here is the `String`-version of `replace()` whose parameters are exactly called `oldValue` and `newValue`.

PR #6825, opened 2023-04-11
feat(helper-cli): Add a filename to path exclude generator

fviernau:


sschuberth (dismissed):
Alternative commit message:

    feat(helper-cli): Add "libtool-ldflags" as a file to exclude

sschuberth (approved):

PR #6828, opened 2023-04-11
CycloneDX cleanups

sschuberth:

### DIFF
--- a/reporter/src/funTest/kotlin/reporters/CycloneDxReporterFunTest.kt
+++ b/reporter/src/funTest/kotlin/reporters/CycloneDxReporterFunTest.kt
@@ -41,11 +41,11 @@ import org.ossreviewtoolkit.utils.test.getAssetAsString
 
 class CycloneDxReporterFunTest : WordSpec({
     val defaultSchemaVersion = CycloneDxReporter.DEFAULT_SCHEMA_VERSION.versionString
-    val optionSingle = mapOf("single.bom" to "true")
-    val optionMulti = mapOf("single.bom" to "false")
     val outputDir = createSpecTempDir()
 
     "BOM generation with single option" should {
+        val optionSingle = mapOf("single.bom" to "true")
+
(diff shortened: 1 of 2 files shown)


reporter/src/funTest/kotlin/reporters/CycloneDxReporterFunTest.kt
```diff
@@ -43,6 +42,12 @@ class CycloneDxReporterFunTest : WordSpec({
     val defaultSchemaVersion = CycloneDxReporter.DEFAULT_SCHEMA_VERSION.versionString
     val outputDir = createSpecTempDir()
 
+    fun String.patchCycloneDxResult() =
```
fviernau, line 45: The function does not use class context. 
I believe it's nicer to have such functions outside of the class, to not distract the reader with irrelevant implementation details.

fviernau (changes requested):

reporter/src/funTest/kotlin/reporters/CycloneDxReporterFunTest.kt
sschuberth: I'd usually agree, but here I chose to put the function next to the properties as it seemed nicer to me to have all "other stuff" (than tests) in a single place than spread over the file.

fviernau: > but here I chose to put the function next to the properties as it seemed nicer to me to have all "other stuff" (than tests) in a single place 

Hm, the first commit in this PR seems to move away from that "single place" closer to the place of use. So, looking at both commits together this argumentation seems to be a bit inconsistent.

sschuberth: Yeah, I can see what you mean. Ok, I've moved the function outside of the class.

fviernau (approved):

PR #6829, opened 2023-04-11
Minor fixes for `generateReport()`

fviernau:
See individual commits.

### DIFF
--- a/reporter/src/funTest/kotlin/reporters/freemarker/PlainTextTemplateReporterFunTest.kt
+++ b/reporter/src/funTest/kotlin/reporters/freemarker/PlainTextTemplateReporterFunTest.kt
@@ -24,7 +24,6 @@ import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.shouldBe
 
 import org.ossreviewtoolkit.model.OrtResult
-import org.ossreviewtoolkit.model.config.CopyrightGarbage
 import org.ossreviewtoolkit.model.config.FileArchiverConfiguration
 import org.ossreviewtoolkit.model.config.FileStorageConfiguration
 import org.ossreviewtoolkit.model.config.LocalFileStorageConfiguration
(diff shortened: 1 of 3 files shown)


reporter/src/funTest/kotlin/reporters/opossum/OpossumReporterFunTest.kt
```diff
@@ -61,8 +61,7 @@ class OpossumReporterFunTest : WordSpec({
 private fun TestConfiguration.generateReport(ortResult: OrtResult): String {
     val input = ReporterInput(
         ortResult = ortResult,
-        resolutionProvider = DefaultResolutionProvider(ortResult.getResolutions()),
-        howToFixTextProvider = { "Some how to fix text." }
+        resolutionProvider = DefaultResolutionProvider(ortResult.getResolutions())
```
sschuberth, line 64: Commit message nit: Maybe explicitly say that this parameter is irrelevant for the test and could be anything, but for simplicity just stick to the default value.

reporter/src/funTest/kotlin/reporters/freemarker/PlainTextTemplateReporterFunTest.kt
```diff
@@ -86,13 +85,11 @@ class PlainTextTemplateReporterFunTest : WordSpec({
 private fun TestConfiguration.generateReport(
     ortResult: OrtResult,
     config: OrtConfiguration = OrtConfiguration(),
-    copyrightGarbage: CopyrightGarbage = CopyrightGarbage(),
     options: Map<String, String> = emptyMap()
 ): String {
     val input = ReporterInput(
         ortResult,
         config,
-        copyrightGarbage = copyrightGarbage,
```
sschuberth, line 95: Commit message: This is not "inlining". It would be inlining if this line was written as `copyrightGarbage = CopyrightGarbage()`. What you're doing is to remove a parameter that's always set to the default value.

reporter/src/funTest/kotlin/reporters/spdx/SpdxDocumentReporterFunTest.kt
```diff
@@ -109,10 +108,7 @@ class SpdxDocumentReporterFunTest : WordSpec({
 })
 
 private fun TestConfiguration.generateReport(ortResult: OrtResult, format: FileFormat): String {
-    val input = ReporterInput(
-        ortResult = ortResult,
-        licenseTextProvider = DefaultLicenseTextProvider()
```
sschuberth, line 114: Again, this is not "inlining", but just removing.

sschuberth (changes requested):

sschuberth (approved):

PR #6831, opened 2023-04-11
refactor(helper-cli): Drop the `--package-configuration-file` option

fviernau:
The set of package configurations as either of the following:

1. `--package-configuration-file` specifies a single file containing all package configurations.
2. `--package-configuration-dir` specifies a directory containing a hierarchical file structure with each file containing a single package configuration.

Using `--package-configuration-file` has the drawbacks that it normally leads to very large files which are unnecessarily hard to get an overview of and to maintain. At the same time it seems not to provide any benefits. So, drop the `--package-configuration-file` option as it does not add value in order to keep things simple.

Note that there are further helper-cli commands with a `--package-configuration-file` option left by intention. However, these have different semantics.

BREAKING CHANGE: `--package-configuration-file` can not be used to specify a single package curation file containing a list             of package configurations. Package configuration should be organized in a hierarchical structure instead and passed via `--package-configuration-dir`.


sschuberth:
> Note that there are further helper-cli commands with a `--package-configuration-file` option left by intention. However, these have different semantics.

Could you please go into more detail here in the commit message? What are these different semantics? Why can't files grow large in those cases?

sschuberth:
Also, what about `--package-configuration-file` in non-helper CLI commands? ARe you planning to remove these in a follow-up?

model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt
```diff
@@ -36,6 +36,10 @@ import org.ossreviewtoolkit.model.readValue
 open class SimplePackageConfigurationProvider(
     configurations: Collection<PackageConfiguration>
 ) : PackageConfigurationProvider {
+    companion object {
+        fun forDirectory(directory: File?): PackageConfigurationProvider =
```
sschuberth, line 40: IIRC it was a deliberate design choice by @mnonnenmacher to remove this function in d10bd49466ba04afbf748397e2413e6872f38f2a.

model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt
```diff
@@ -36,6 +36,10 @@ import org.ossreviewtoolkit.model.readValue
...
+            directory?.let { DirectoryPackageConfigurationProvider(directory) } ?: PackageConfigurationProvider.EMPTY
+    }
```
sschuberth, line 42: Missing empty line after this one.

model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt
fviernau: Thanks, I was aware of this choice. My commit here is removing the `createProvider()` function from `utils`.

I was in need of a replacement for this function and this was my best guess. It can also be used from `cli` module, as soon as I also remove the `--package-configuration-file` option there.

fviernau:
> Also, what about `--package-configuration-file` in non-helper CLI commands? Are you planning to remove these in a follow-up?

Yes.

model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt
sschuberth: It seems wrong to me to let `SimplePackageConfigurationProvider` create a `DirectoryPackageConfigurationProvider` instead of letting the latter do that itself. I guess I'd simply inline this implementation to the caller, also see my other comment.

helper-cli/src/main/kotlin/commands/GetPackageLicensesCommand.kt
```diff
@@ -68,25 +65,15 @@ class GetPackageLicensesCommand : CliktCommand(
...
+    private val packageConfigurationDir by option(
+        "--package-configuration-dir",
+        help = "The directory containing the package configuration files to read as input. It is searched recursively."
+    ).file(mustExist = true, canBeFile = false, canBeDir = true, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.expandTilde() }
 
     override fun run() {
         val scanResults = getStoredScanResults(packageId)
-        val packageConfigurationProvider = packageConfigurationOption.createProvider()
+        val packageConfigurationProvider = SimplePackageConfigurationProvider.forDirectory(packageConfigurationDir)
```
sschuberth, line 76: My proposal is to do

    val packageConfigurationProvider = packageConfigurationDir?.let { DirectoryPackageConfigurationProvider(it) } ?: PackageConfigurationProvider.EMPTY

here and in the other places.

model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt
sschuberth: Another alternative would be to make `DirectoryPackageConfigurationProvider`'s `directory` parameter nullable and handle this as part of its constructor.

sschuberth (approved):

fviernau:
Failing test is unrelated.

PR #6832, opened 2023-04-12
AsciiDoc reporter clean-ups

sschuberth:

> **Note**
> This is a **breaking change** as the `XHtmlTemplateReporter` reporter is removed in favor of the almost identical `HtmlTemplateReporter` reporter.


fviernau (approved):

README.md
nnobelis, line 1: I think the argument of XHML "being more modern" is fallacious. Could you justify it ?
My understanding is that, _currently_, XHTML is 'just' HTML with XML syntax.
But the current state is pretty complicated, e.g. https://softwareengineering.stackexchange.com/questions/149839/is-xhtml5-dead-or-is-it-just-an-synonym-of-html5.

sschuberth: Note that when saying "HTML" here, I don't mean "HTML5". I'm referring to the "original" HTML (version <= 4.01) to which XHTML IMO is an improvement as it's more strict. Actually, HTML5 is probably more close to XHTML than to HTML <= 4.01.

Anyway, while I personally linked the syntax in `xhtml-template-reporter-expected-result.xhtml` a little bit better than the one in `html-template-reporter-expected-result.html`, I don't really care that much. I can also remove `XHtmlTemplateReporter` instead, if that's the consensus.

nnobelis: Both are fine for me. My comment what just about the justification of the changes.

sschuberth: > I can also remove `XHtmlTemplateReporter` instead, if that's the consensus.

After reading through https://docs.asciidoctor.org/asciidoctor/latest/html-backend/, I guess that's what I'll do. We could later make it a `HtmlTemplateReporter`-specific option to output XHTML instead, if required, via the `htmlsyntax=xml` attribute.

nnobelis: Maybe adjust the commit message with "XHTML is stricter" or "XHTML has a better syntax" ?
In the second case you make is clear it's a personal preference.

nnobelis:
@sschuberth This is not a breaking change anymore, right ?

sschuberth:
> @sschuberth This is not a breaking change anymore, right ?

It is, for any previous user of the `XHtmlTemplateReporter`.

fviernau (approved):

PR #6854, opened 2023-04-16
`OrtResult` function renamings

mnonnenmacher:
Rename several `collect...` functions to `get...` and do some minor refactorings. See the commit messages for details.

**Breaking API change**:
Several `OrtResult` functions have changed their name.

### DIFF
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -164,6 +164,30 @@ data class OrtResult(
     private val scanResultsById: Map<Identifier, List<ScanResult>> by lazy { scanner?.scanResults.orEmpty() }
 
     /**
+     * Return all [AdvisorResult]s contained in this [OrtResult] or only the non-excluded ones if [omitExcluded] is
+     * true.
+     */
+    @JsonIgnore
+    fun getAdvisorResults(omitExcluded: Boolean = false): Map<Identifier, List<AdvisorResult>> =
+        advisorResultsById.filter { (id, _) ->
+            !omitExcluded || !isExcluded(id)
+        }
+
+    /**
+     * Return the list of [AdvisorResult]s for the given [id].
+     */
+    @Suppress("UNUSED")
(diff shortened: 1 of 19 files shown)


model/src/main/kotlin/OrtResult.kt
```diff
@@ -96,35 +96,22 @@ data class OrtResult(
         )
     }
 
-    /** An object that can be used to navigate the dependency information contained in this result. */
```
sschuberth, line 99: Commit message nit: Is sorting really refactoring? Maybe rather "chore" / "style"?

model/src/main/kotlin/OrtResult.kt
```diff
@@ -163,6 +163,30 @@ data class OrtResult(
 
     private val scanResultsById: Map<Identifier, List<ScanResult>> by lazy { scanner?.scanResults.orEmpty() }
 
+    /**
```
sschuberth, line 166: Same commit message nit here about the "refactor" type.

sschuberth (commented):
Do you also want to address `collectedIssues` in `DependencyGraph` / `DependencyNavigator` for consistency?

mnonnenmacher:
> Do you also want to address `collectedIssues` in `DependencyGraph` / `DependencyNavigator` for consistency?

Yes, but not in this PR, I wanted to focus on `OrtResult` first.

model/src/main/kotlin/OrtResult.kt
mnonnenmacher: I think "chore" fits best.

sschuberth (approved):
I'm fine with this, the `collect*` function names have bothered me before, too. I'm just a bit concerned about how many rule scripts this might break.

mnonnenmacher:
> I'm fine with this, the `collect*` function names have bothered me before, too. I'm just a bit concerned about how many rule scripts this might break.

Rule scripts should not be affected as long as the DSL is used, it's more likely notifier scripts like the example script, but I think the notifier is not being used that much.

sschuberth (approved):

Issue #6602, opened 2023-03-03
Move remaining plugins under the new `plugins` directory


sschuberth:
Ticked ones have already been moved:

- [x] [org.ossreviewtoolkit.advisor.AdviceProviderFactory](https://github.com/oss-review-toolkit/ort/blob/9e1d31e2247553d116c3ab0bde0c213bd2810fb8/advisor/src/main/kotlin/AdviceProviderFactory.kt#L34)
- [x] [org.ossreviewtoolkit.analyzer.PackageCurationProviderFactory](https://github.com/oss-review-toolkit/ort/blob/9e1d31e2247553d116c3ab0bde0c213bd2810fb8/analyzer/src/main/kotlin/PackageCurationProviderFactory.kt#L31)
- [x] [org.ossreviewtoolkit.analyzer.PackageManagerFactory](https://github.com/oss-review-toolkit/ort/blob/9e1d31e2247553d116c3ab0bde0c213bd2810fb8/analyzer/src/main/kotlin/PackageManagerFactory.kt#L34)
  - [x] Bower
  - [x] Bundler
  - [x] Cargo
  - [x] Carthage
  - [x] CocoaPods
  - [x] Composer
  - [x] Conan
  - [x] ~DotNet~ (was removed instead)
    - Probably wait for #6209
  - [x] GoDep
    - Should go to a combined "go" (or "golang") project
  - [x] GoMod
    - Should go to a combined "go" (or "golang") project
  - [x] Gradle
  - [x] Maven
  - [x] Npm
    - Should go to a combined "node" project
  - [x] NuGet
    - Probably wait for https://github.com/oss-review-toolkit/ort/pull/6209
  - [x] Pip
  - [x] Pipenv
  - [x] Pnpm
    - Should go to a combined "node" project
  - [x] Poetry
  - [x] Pub
  - [x] Sbt
    - Maybe do #633 first to reduce coupling to Maven
  - [x] SpdxDocumentFile
  - [x] Stack
  - [x] Unmanaged
  - [x] Yarn
    - Should go to a combined "node" project
  - [x] Yarn2
    - Should go to a combined "node" project
- [x] [org.ossreviewtoolkit.cli.OrtCommand](https://github.com/oss-review-toolkit/ort/blob/9e1d31e2247553d116c3ab0bde0c213bd2810fb8/cli/src/main/kotlin/OrtCommand.kt#L29)
  - [x] AdvisorCommand
  - [x] AnalyzerCommand
[...]

PR #6861, opened 2023-04-17
Move the Opossum reporter to its own project

sschuberth:
FYI @maxhbr & @lennartclaas, this is part of #6602.

### DIFF
--- a/plugins/reporters/opossum/src/main/kotlin/OpossumReporter.kt
+++ b/plugins/reporters/opossum/src/main/kotlin/OpossumReporter.kt
@@ -84,7 +84,7 @@ class OpossumReporter : Reporter {
         const val OPTION_SCANNER_MAX_DEPTH = "scanner.maxDepth"
     }
 
-    data class OpossumSignal(
+    internal data class OpossumSignal(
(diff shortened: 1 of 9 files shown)


maxhbr (approved):
LGTM

mnonnenmacher (approved):

plugins/reporters/opossum/src/main/kotlin/OpossumReporter.kt
```diff
@@ -17,7 +17,7 @@
  * License-Filename: LICENSE
  */
 
-package org.ossreviewtoolkit.reporter.reporters.opossum
```
fviernau, line 20: Missing the change in `META-INF`?

fviernau (changes requested):

plugins/reporters/opossum/src/main/kotlin/OpossumReporter.kt
sschuberth: Right, thanks! Fixed.

fviernau (approved):

PR #6869, opened 2023-04-18
chore(clearly-defined): Remove ORT's comment from curations

sschuberth:
The custom comment was used as a work-around to document the provider the curation is coming from. But now that the resolved configuration contains the provider explicitly, this is not needed anymore.


fviernau (approved):

sschuberth:

PR #6870, opened 2023-04-18
GradleInspector fixes in preparation for adding functional tests

sschuberth:

### DIFF
--- a/plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
+++ b/plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
@@ -162,10 +162,6 @@ internal class OrtModelBuilder : ToolingModelBuilder {
 
                     when (val id = selectedComponent.id) {
                         is ModuleComponentIdentifier -> {
-                            // Cut the graph on cyclic dependencies.
-                            if (id in visitedDependencies) return@mapNotNull null
-                            visitedDependencies += id
-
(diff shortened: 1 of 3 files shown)


plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -186,14 +182,21 @@ internal class OrtModelBuilder : ToolingModelBuilder {
                             }
 
                             val modelBuildingResult = poms.getValue(id.toString())
+                            val dependencies = if (id in visitedDependencies) {
```
mnonnenmacher, line 185: Can you provide more details on the issue, I'm not sure which of the two options below apply for the example of a "a -> b -> a" circle:

Before: "a"
After: "a -> b"

or

Before: "a -> b"
After: "a -> b -> a"

mnonnenmacher (changes requested):

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
sschuberth: With this change it's like in your second example now.

mnonnenmacher: I'm not sure if other package managers cut the circle like in the first example, but we discuss this separately.

mnonnenmacher (approved):

PR #6872, opened 2023-04-18
fix(gradle): Require the binary artifact to be empty for pure metadata

sschuberth:
This fixes the case when a dependency with a custom classifier and / or extension (see `org.apache.struts:struts2-assembly:2.5.14.1:min-lib@zip`) was treated as metadata-only.


analyzer/src/main/kotlin/managers/utils/MavenSupport.kt
```diff
@@ -790,7 +790,8 @@ class MavenSupport(private val workspaceReader: WorkspaceReader) {
             sourceArtifact = sourceRemoteArtifact,
             vcs = vcsFromPackage,
             vcsProcessed = vcsProcessed,
-            isMetadataOnly = mavenProject.packaging == "pom" || isSpringMetadataProject,
+            isMetadataOnly = (mavenProject.packaging == "pom" && binaryRemoteArtifact.url.endsWith(".pom"))
```
mnonnenmacher, line 793: Is the pom URL always used as the binary artifact in these cases or do we also have to check if `binaryRemoteArtifact.url.isEmpty()`?

sschuberth: There is indeed *one* case in `pub-expected-output-with-flutter-android-and-cocoapods.yml` where `is_metadata_only: true` and the binary artifact URL is empty. However, the URL shouldn't be empty (that's a bug in `Gradle` that `GradleInspector` fixes). So, phrased differently, in all cases (of our expected results) where *rightfully* `is_metadata_only: true`, the URL is non-empty and ends in `.pom`.

mnonnenmacher (approved):

PR #6882, opened 2023-04-19
Add functional tests for the GradleInspector

sschuberth:

### DIFF
--- a/plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-android-expected-output-app.yml
+++ b/plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-android-expected-output-app.yml
@@ -1,0 +1,693 @@
+---
+project:
+  id: "Gradle:org.ossreviewtoolkit.gradle.example:app:1.0.0"
+  definition_file_path: "plugins/package-managers/gradle/src/funTest/assets/projects/synthetic/gradle-android/app/build.gradle"
+  declared_licenses: []
+  declared_licenses_processed: {}
+  vcs:
+    type: ""
+    url: ""
+    revision: ""
+    path: ""
+  vcs_processed:
+    type: "Git"
+    url: "<REPLACE_URL_PROCESSED>"
+    revision: "<REPLACE_REVISION>"
+    path: "plugins/package-managers/gradle/src/funTest/assets/projects/synthetic/gradle-android/app"
+  homepage_url: ""
+  scopes:
+  - name: "amazonDemoDebugAndroidTestCompileClasspath"
+    dependencies:
+    - id: "Gradle:org.ossreviewtoolkit.gradle.example:lib:1.0.0"
+      linkage: "PROJECT_DYNAMIC"
+    - id: "Maven:com.squareup.okhttp3:okhttp:3.10.0"
+      dependencies:
+      - id: "Maven:com.squareup.okio:okio:1.14.0"
+  - name: "amazonDemoDebugAndroidTestRuntimeClasspath"
+    dependencies:
+    - id: "Gradle:org.ossreviewtoolkit.gradle.example:lib:1.0.0"
+      linkage: "PROJECT_DYNAMIC"
+      dependencies:
+      - id: "Maven:org.apache.commons:commons-compress:1.17"
+      - id: "Maven:org.apache.commons:commons-text:1.2"
+        dependencies:
+        - id: "Maven:org.apache.commons:commons-lang3:3.7"
+    - id: "Maven:com.squareup.okhttp3:okhttp:3.10.0"
+      dependencies:
+      - id: "Maven:com.squareup.okio:okio:1.14.0"
+  - name: "amazonDemoDebugCompileClasspath"
+    dependencies:
+    - id: "Gradle:org.ossreviewtoolkit.gradle.example:lib:1.0.0"
+      linkage: "PROJECT_DYNAMIC"
+    - id: "Maven:com.squareup.okhttp3:okhttp:3.10.0"
+      dependencies:
+      - id: "Maven:com.squareup.okio:okio:1.14.0"
+  - name: "amazonDemoDebugRuntimeClasspath"
+    dependencies:
+    - id: "Gradle:org.ossreviewtoolkit.gradle.example:lib:1.0.0"
+      linkage: "PROJECT_DYNAMIC"
+      dependencies:
+      - id: "Maven:org.apache.commons:commons-compress:1.17"
+      - id: "Maven:org.apache.commons:commons-text:1.2"
+        dependencies:
+        - id: "Maven:org.apache.commons:commons-lang3:3.7"
+    - id: "Maven:com.squareup.okhttp3:okhttp:3.10.0"
+      dependencies:
+      - id: "Maven:com.squareup.okio:okio:1.14.0"
+  - name: "amazonDemoDebugUnitTestCompileClasspath"
+    dependencies:
+    - id: "Gradle:org.ossreviewtoolkit.gradle.example:lib:1.0.0"
+      linkage: "PROJECT_DYNAMIC"
+    - id: "Maven:com.squareup.okhttp3:okhttp:3.10.0"
+      dependencies:
+      - id: "Maven:com.squareup.okio:okio:1.14.0"
+    - id: "Maven:junit:junit:4.12"
+      dependencies:
+      - id: "Maven:org.hamcrest:hamcrest-core:1.3"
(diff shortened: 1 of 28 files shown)


plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-android-expected-output-app.yml
```diff
@@ -396,10 +396,10 @@ packages:
   description: "An HTTP+HTTP/2 client for Android and Java applications"
   homepage_url: "https://github.com/square/okhttp/okhttp"
   binary_artifact:
-    url: ""
+    url: "https://repo.maven.apache.org/maven2/com/squareup/okhttp3/okhttp/3.10.0/okhttp-3.10.0.jar"
```
sschuberth, line 399: The changes in this file are obviously to the better.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-expected-output-lib-without-repo.yml
```diff
@@ -17,57 +17,22 @@ project:
...
-      issues:
-      - timestamp: "1970-01-01T00:00:00Z"
-        source: "Gradle"
-        message: "Unresolved: ModuleVersionNotFoundException: Cannot resolve external\
-          \ dependency org.apache.commons:commons-text:1.1 because no repositories\
-          \ are defined."
-        severity: "ERROR"
+    dependencies: []
 packages: []
+issues:
```
sschuberth, line 28: The GI reports these on the project level instead of per-scope, which probably makes sense as repositories are also defined on the project level. This implies some nice de-duplication.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-expected-output-lib-without-repo.yml
```diff
@@ -17,57 +17,22 @@ project:
...
+- timestamp: "1970-01-01T00:00:00Z"
+  source: "GradleInspector"
+  message: "Cannot resolve external dependency org.apache.commons:commons-text:1.1\
+    \ because no repositories are defined from project :lib-without-repo."
+  severity: "WARNING"
```
sschuberth, line 33: Should these be `ERROR` as before? It's not hindering GI to do its work (as good as Gradle CLI itself would be able to so), so I was unsure and went with `WARNING` for now.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/multi-kotlin-project-expected-output-cli.yml
```diff
@@ -24,14 +24,6 @@ project:
       dependencies:
       - id: "Maven:org.jetbrains:annotations:13.0"
       - id: "Maven:org.jetbrains.kotlin:kotlin-stdlib-common:1.7.20"
-  - name: "implementationDependenciesMetadata"
```
sschuberth: GI generally skips "DependenciesMetadata" configurations / scopes.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/multi-kotlin-project-expected-output-cli.yml
```diff
@@ -161,14 +113,6 @@ project:
       dependencies:
       - id: "Maven:org.jetbrains:annotations:13.0"
       - id: "Maven:org.jetbrains.kotlin:kotlin-stdlib-common:1.7.20"
-  - name: "testImplementationDependenciesMetadata"
```
sschuberth: GI generally skips "DependenciesMetadata" configurations / scopes.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/multi-kotlin-project-expected-output-core.yml
```diff
@@ -148,12 +105,6 @@ project:
       dependencies:
       - id: "Maven:org.jetbrains:annotations:13.0"
       - id: "Maven:org.jetbrains.kotlin:kotlin-stdlib-common:1.7.20"
-  - name: "testImplementationDependenciesMetadata"
```
sschuberth: GI generally skips "DependenciesMetadata" configurations / scopes.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/multi-kotlin-project-expected-output-cli.yml
```diff
@@ -42,9 +34,6 @@ project:
       - id: "Maven:org.jetbrains.kotlin:kotlin-reflect:1.7.20"
         dependencies:
         - id: "Maven:org.jetbrains.kotlin:kotlin-stdlib:1.7.20"
-          dependencies:
```
sschuberth: `kotlin-stdlib`'s dependencies were already seen on a level closer to the root (see line 49 and below), so they are skipped here. I'm not sure why the `Gradle` package manager implementation does not skip it, too.

sschuberth: > I'm not sure why the `Gradle` package manager implementation does not skip it, too.

That's because the "legacy" Gradle analyzer natively creates a *graph* which is then resolved back to a *tree* for test result comparison, and this expands the cycles again.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-expected-output-lib-without-repo.yml
mnonnenmacher: IMO if dependencies cannot be resolved it should be an `ERROR`.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-expected-output-lib-without-repo.yml
mnonnenmacher: I would say dependencies are defined on the scope level. What if the errors occur further down in the tree, would they also be reporter on the project level instead of inside the tree?

sschuberth: > I would say dependencies are defined on the scope level.

True. What I meant was: *Repositories* are defined on the project level, so errors caused by no repository being present could be regarded as project configuration errors, and in that sense it makes sense (heh) to report the errors on the project level. However, missing repos are probably not the only cause for a dependency to be unresolved.

Probably the question boils down to: Is the position in the dependency tree important, i.e. may it have any influence on the error. I guess not.

But I'm undecided. I'm fine either way, to record unresolved dependencies (deduplicated) on the project level, or inside the dependency tree (with possible duplicates).

sschuberth:
@mnonnenmacher @fviernau this is up for review again as I've found the cause for the failing `GradleAndroidFunTest`: The test must not be run in parallel for the `gradle-package-manager` and `gradle-inspector` projects as the Android component installation triggered by the Android Gradle plugin might fail then.

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-expected-output-lib-without-repo.yml
mnonnenmacher: > Probably the question boils down to: Is the position in the dependency tree important, i.e. may it have any influence on the error. I guess not.
> But I'm undecided. I'm fine either way, to record unresolved dependencies (deduplicated) on the project level, or inside the dependency tree (with possible duplicates).

For direct dependencies this might not be so important, but for transitive dependencies it can be very helpful to see where in the tree the issue occurs. I would appreciate if we would report issues inside the tree where possible.

mnonnenmacher (approved):

plugins/package-managers/gradle-inspector/src/funTest/assets/projects/synthetic/gradle-expected-output-lib-without-repo.yml
sschuberth: Looks like you're still fine for now with the current state as you've approved, so I'll eventually fix this up later to get the tests in now.

PR #6893, opened 2023-04-20
NuGet fixups

sschuberth:


plugins/package-managers/nuget/src/main/kotlin/utils/NuGetInspector.kt
```diff
@@ -224,7 +224,7 @@ private fun NuGetInspector.PackageData.toPackageReference() =
         dependencies = dependencies.toPackageReferences()
     )
 
-internal fun Set<NuGetInspector.PackageData>.toOrtPackages(): Set<Package> =
+internal fun Collection<NuGetInspector.PackageData>.toOrtPackages(): Set<Package> =
```
fviernau, line 226: out of scope: I believe it's nicer to have this function in "singular" form.

```
result.dependencies.map { it.toOrtPackage() }.disctinctBy {it.id}
```

Note: Currently it's a bit unclear to me why there can be duplicate "${it.name}:${it.version}", and why the "first" 
one is picked and the other ignored.

fviernau (changes requested):

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetInspector.kt
```diff
@@ -86,7 +86,7 @@ internal object NuGetInspector : CommandLineTool {
         analysisRoot: File,
         nugetConfig: String?
     ): List<ProjectAnalyzerResult> {
-        val result = runNuGetInspector(definitionFile, nugetConfig)
+        val result = run(definitionFile, nugetConfig)
```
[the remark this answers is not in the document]
sschuberth: I actually like that name. Will do for the *python-inspector*, too.

fviernau (approved):

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetInspector.kt
sschuberth: > I believe it's nicer to have this function in "singular" form.

I actually decided to go the opposite direction and only keep the "plural" form, see https://github.com/oss-review-toolkit/ort/pull/6899/commits/288f2f6e42a376f1cc7c5a18839b594a63261322.

PR #6898, opened 2023-04-21
refactor(model)!: Unqualify scope names by default

sschuberth:
Align the default values of the `createScopes()` functions. In the dependency tree representation usually scopes names are unqualified.


oheger-bosch (approved):
I think, this function is only used by test code. So, if the tests continue to work, this is fine with me.

PR #6899, opened 2023-04-21
Improve NuGet error / warning mapping

sschuberth:


MarcelBochtler (approved):

analyzer/src/testFixtures/kotlin/Extensions.kt
```diff
@@ -27,8 +27,9 @@ import io.kotest.matchers.should
 
 import java.io.File
```
MarcelBochtler, line 28: Commit message: I think you intended to say `Remove` instead of `Rename`.

MarcelBochtler (changes requested):

MarcelBochtler (approved):

PR #6905, opened 2023-04-23
Move FreeMarker and AsciiDoc reporters to own projects

sschuberth:


cli/src/funTest/kotlin/ExamplesFunTest.kt
```diff
@@ -56,8 +56,8 @@ import org.ossreviewtoolkit.model.licenses.LicenseClassifications
 import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.notifier.Notifier
 import org.ossreviewtoolkit.reporter.HowToFixTextProvider
+import org.ossreviewtoolkit.reporter.Reporter
```
fviernau, line 59: commit-msg:  

I was a bit confused by the term split in "but it makes sense to split them".
Would it make sense to say "but it makes sense to move them into separate plugins"?

"Due to this dependency it is easier" it's not obvious to me why this dependency makes it harder, can you give a hint?

fviernau (changes requested):

cli/src/funTest/kotlin/ExamplesFunTest.kt
sschuberth: I've clarified both things in a reworked commit message.

PR #6909, opened 2023-04-24
Improve `VulnerabilityReference.getSeverityString()`

mnonnenmacher:

### DIFF
--- a/model/src/main/kotlin/VulnerabilityReference.kt
+++ b/model/src/main/kotlin/VulnerabilityReference.kt
@@ -51,15 +51,17 @@ data class VulnerabilityReference(
     val severity: String?
 ) {
     companion object {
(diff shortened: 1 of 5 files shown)


model/src/main/kotlin/VulnerabilityReference.kt
```diff
@@ -51,14 +51,16 @@ data class VulnerabilityReference(
     val severity: String?
 ) {
     companion object {
+        private val SEVERITIES = listOf("NONE", "LOW", "MEDIUM", "HIGH", "CRITICAL")
```
sschuberth, line 54: Would it make sense to use `enumValues<Cvss3Rating>().map { it.name }` instead? (Esp. if we'd know that VC is actually using CVSS3 names here.)

sschuberth (approved):

PR #6917, opened 2023-04-24
fix(bundler): Make GemSpec parsing more robust

sschuberth:
Disregard empty URIs when creating VCS info. This deals with a change in SnakeYaml 2.0 (used by Jackson 2.15) that deserializes YAML maps with missing values as empty values instead of previously null values.

### DIFF
--- a/plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
+++ b/plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
@@ -54,6 +54,7 @@ import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.PackageManagerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
+import org.ossreviewtoolkit.model.orEmpty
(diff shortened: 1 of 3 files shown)


plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
```diff
@@ -425,11 +425,11 @@ internal data class GemSpec(
...
-            val vcs = when {
-                node.hasNonNull("source_code_uri") -> VcsHost.parseUrl(node["source_code_uri"].textValue())
-                node.hasNonNull("homepage_uri") -> VcsHost.parseUrl(node["homepage_uri"].textValue())
-                else -> VcsInfo.EMPTY
-            }
+            val vcs = sequenceOf(node["source_code_uri"], node["homepage_uri"]).mapNotNull { uri ->
+                uri?.textValue()?.takeIf { it.isNotEmpty() }
+            }.firstOrNull()?.let {
+                VcsHost.parseUrl(it)
+            } ?: VcsInfo.EMPTY
```
fviernau, line 432: This can use `VcsInfo.orEmpty()`.

fviernau (changes requested):

fviernau (approved):

PR #6931, opened 2023-04-26
refactor(scanner): Simplify the message for provenance issues

sschuberth:
The underlying `PackageProvenanceResolver` already uses the same prefix for the message, leading to duplications like

    Could not resolve provenance for package '...': IOException: Could not
    resolve provenance for '...' for source code origins [VCS, ARTIFACT].

Simplify that by omitting the first sentence.

### DIFF
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -192,12 +192,7 @@ class Scanner(
                 }.onFailure {
                     controller.addProvenanceResolutionIssue(
                         pkg.id,
-                        Issue(
-                            source = TOOL_NAME,
-                            severity = Severity.ERROR,
-                            message = "Could not resolve provenance for package '${pkg.id.toCoordinates()}': " +
-                                    it.collectMessages()
-                        )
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/Scanner.kt
```diff
@@ -192,12 +192,7 @@ class Scanner(
                 }.onFailure {
                     controller.addProvenanceResolutionIssue(
                         pkg.id,
-                        Issue(
-                            source = TOOL_NAME,
-                            severity = Severity.ERROR,
-                            message = "Could not resolve provenance for package '${pkg.id.toCoordinates()}': " +
-                                    it.collectMessages()
-                        )
+                        Issue(source = TOOL_NAME, severity = Severity.ERROR, message = it.collectMessages())
```
fviernau, line 195: Is `it` always thrown from line 101, or could the exception be coming from elsewhere?

sschuberth: That's the only place (other exceptions thrown in this file are collected into this one).

fviernau (approved):

sschuberth:

PR #6932, opened 2023-04-27
refactor(downloader)!: Remove CVS support

sschuberth:
According to samples in ORT provenance databases, CVS usage is down to 1 of 350000. CVS binaries have not seen any updates since 2008, and its general use is rapidly declining. So remove CVS support from ORT to reduce maintenance effort and make the code base and dependencies to external tools a bit leaner.

> **Note**
> This is a breaking change for users of CVS, or dependencies whose source code is hosted in CVS. If required, the advice is to either scan source artifacts, create a curation for a Git mirror, or curate author / copyright plus license information and configure the scanner with `skipConcluded = true`.


utils/common/src/main/kotlin/Utils.kt
```diff
@@ -31,9 +31,7 @@ val VCS_DIRECTORIES = listOf(
     ".git",
     ".hg",
     ".repo",
-    ".svn",
-    "CVS",
-    "CVSROOT"
+    ".svn"
```
fviernau, line 34: Would it make sense to keep the entry here, as it is used as "exclude" patterns in several places, e.g. scanner? To me the entry seems to make sense no matter if the corresponding VCSes are supported.

fviernau (changes requested):

utils/common/src/main/kotlin/Utils.kt
sschuberth: I was thinking about that, too, but when concluded that we should actually never come across these patterns if we're not able anymore to create CVS working trees.

fviernau: Have you considered `ort scan --input-path` ?

sschuberth: Actually, @mnonnenmacher recently proposed to drop that option, which is sometime I was planning to do soon. Are you actually using that option?

fviernau: I've been using the option very rarely. I could as well use ScanCode directly instead in my use cases.

sschuberth: Ok, so given that option would be dropped, are we good for this PR?

fviernau: I've looked again a bit closer at all the usages of `VCS_DIRECTORIES`.
Namely,

1. `packZip()`
2. `calculatePackageVerificationCode()`
3. `ScanOss.scanPath()`
4. `Analyzer.findManagedFiles()`
5. `PackageManager.IGNORED_DIRECTORY_MATCHERS`

I still prefer these to work independently of whether a VCS is supported by ORT or not. 
So, let's just get another reviewers opinion to resolved this disagreement.

sschuberth: > So, let's just get another reviewers opinion to resolved this disagreement.

There's not necessarily a disagreement. I probably just wasn't looking so deeply at the usages of `VCS_DIRECTORIES` as you did. Giving your findings, I'm actually fine with keeping CVS directories in there, even if it's still quite unlikely that we come across them.

fviernau (approved):

PR #6955, opened 2023-05-02
Use package configurations from the config directory by default

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/EvaluatorCommand.kt
+++ b/cli/src/main/kotlin/commands/EvaluatorCommand.kt
@@ -143,6 +143,7 @@ class EvaluatorCommand : OrtCommand(
                 "contain a single package configuration."
     ).convert { it.expandTilde() }
         .file(mustExist = true, canBeFile = false, canBeDir = true, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
         .default(ortConfigDirectory.resolve(ORT_PACKAGE_CONFIGURATIONS_DIRNAME))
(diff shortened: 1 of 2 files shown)


cli/src/main/kotlin/commands/EvaluatorCommand.kt
```diff
@@ -142,6 +143,7 @@ class EvaluatorCommand : OrtCommand(
                 "contain a single package configuration."
     ).convert { it.expandTilde() }
         .file(mustExist = true, canBeFile = false, canBeDir = true, mustBeWritable = false, mustBeReadable = true)
+        .default(ortConfigDirectory.resolve(ORT_PACKAGE_CONFIGURATIONS_DIRNAME))
```
fviernau, line 147: commit-msg: "This aligns with the behavior for package curations." could be mis-leading, because in evaluator command the `--package-curations-dir` option does not default to any dir.

Would it make sense to just drop that sentence?

fviernau (changes requested):

cli/src/main/kotlin/commands/EvaluatorCommand.kt
sschuberth: > commit-msg: "This aligns with the behavior for package curations." could be mis-leading, because in evaluator command the `--package-curations-dir` option does not default to any dir.

Yeah, that's the special case of overriding.

> Would it make sense to just drop that sentence?

I'd instead prefer to write "... for package curations in the analyzer."

Agreed?

fviernau (approved):

PR #6958, opened 2023-05-02
ReportTableModelMapper: First couple of style improvements

fviernau:
First couple of improvements to prepare for further ones.

### DIFF
--- a/plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
+++ b/plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
@@ -87,14 +87,14 @@ class StaticHtmlReporter : Reporter {
         outputDir: File,
         options: Map<String, String>
     ): List<File> {
-        val tabularScanRecord = ReportTableModelMapper.map(
+        val reportTableModel = ReportTableModelMapper.map(
             input.ortResult,
             input.licenseInfoResolver,
             input.resolutionProvider,
             input.howToFixTextProvider
         )
 
-        val html = renderHtml(tabularScanRecord)
+        val html = renderHtml(reportTableModel)
         val outputFile = outputDir.resolve(reportFilename)
 
         outputFile.bufferedWriter().use {
--- a/plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
+++ b/plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
@@ -87,13 +87,12 @@ class StaticHtmlReporter : Reporter {
         outputDir: File,
         options: Map<String, String>
     ): List<File> {
-        val tabularScanRecord = ReportTableModelMapper()
-            .map(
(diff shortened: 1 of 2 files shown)


plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
```diff
@@ -87,14 +87,14 @@ class StaticHtmlReporter : Reporter {
         outputDir: File,
         options: Map<String, String>
     ): List<File> {
-        val tabularScanRecord = ReportTableModelMapper.map(
+        val reportTableModel = ReportTableModelMapper.map(
```
sschuberth, line 90: Commit message: "with *the* parameter name"

reporter/src/main/kotlin/ReportTableModelMapper.kt
```diff
@@ -252,3 +212,43 @@ private fun Project.getScopesForDependencies(
...
 }
+
+private fun Issue.toResolvableIssue(
+    resolutionProvider: ResolutionProvider,
+    howToFixTextProvider: HowToFixTextProvider
+): ResolvableIssue {
+    val resolutions = resolutionProvider.getIssueResolutionsFor(this)
+    return ResolvableIssue(
+        source = this@toResolvableIssue.source,
+        description = this@toResolvableIssue.toString(),
```
sschuberth, line 223: The `this@toResolvableIssue.` prefixes can be dropped now as there is no other `this` context than that of `Issue`.

reporter/src/main/kotlin/ReportTableModelMapper.kt
```diff
@@ -45,46 +45,6 @@ import org.ossreviewtoolkit.reporter.ReportTableModel.SummaryTable
  * A mapper which converts an [OrtResult] to a [ReportTableModel].
  */
 object ReportTableModelMapper {
-    private fun Issue.toResolvableIssue(
```
sschuberth: The `object` now contains only the single `map` function. Instead of using an object to put that single function into a namespace, I wonder whether it should also become a top-level function, and maybe name it back (or not rename it in the first place) to the original `mapToReportTableModel` name, and maybe also make it an extension function on `OrtResult`.

sschuberth (changes requested):

reporter/src/main/kotlin/ReportTableModelMapper.kt
fviernau: I prefer keeping it an object.

sschuberth: > It's also consistent with other mappers IIRC.

I'd say we should rather fix then the others, too. But that can happen outside of this PR.

fviernau: Sure out scope. Let's just decide on such style thing together in the Kotlin core devs group in the mondays meeting.

PR #6961, opened 2023-05-03
refactor(examples): Liberate notifications and rules scripts

sschuberth:
It is actually not required to wrap these scripts in Gradle projects to get auto-completion to work. On the contrary, putting a script below a source root or the resources directory in particular might lead to an internal error in IntelliJ IDEA, see [1]. So do as proposed in the comments of the linked issue and move the scripts out of any source roots.

[1]: https://youtrack.jetbrains.com/issue/KTIJ-22806


fviernau (dismissed):

fviernau:
Can you help me understanding if there is any value left in keeping this gradle projects around these `.kts` files in `ort-config` repo, or would it make sense to do the same there as well?

fviernau (approved):

sschuberth:
> Can you help me understanding if there is any value left in keeping this gradle projects around these `.kts` files in `ort-config` repo, or would it make sense to do the same there as well?

My plan is to do it there as well, but I'd first like to continue with some cleanups of the examples here. For example, it occurred to me in the context of moving the CLI commands to separate plugin projects that the `ExamplesFunTest` should actually be part of the examples directory.

Issue #6943, opened 2023-04-28
scanner: Store file listings in a dedicated storage


fviernau:

fviernau:
We currently have the interface `FileArchiverStorage` with `FileArchiverFileStorage` and `PostgresFileArchiverStorage` as implementations. These can be re-used after generalizing them, e.g. rename as follows

1. `FileArchiverStorage` -> `ProvenanceFileStorage`
2. `FileArchiverFileStorage` -> `FileProvenanceFileStorage`
3. `PostgresFileArchiverStorage` -> `PosgresProvenanceFileStorage`

and then make the interface / implementation agnostic of any file extension and file name.

sschuberth:
> These can be re-used after generalizing them, e.g. rename as follows

But aren't your proposed names actually more specific than more generic? Because previously, e.g. `FileArchiverFileStorage` could store any file archives, even those not containing any provenance-related information.

PR #6966, opened 2023-05-05
utils: Improve the listing of VCS directories

fviernau:
See individual commits.

Part of #6943, e.g. files under `VCS_DIRECTORIES` will be excluded from the file listings.

### DIFF
--- a/utils/common/src/main/kotlin/Utils.kt
+++ b/utils/common/src/main/kotlin/Utils.kt
@@ -25,14 +25,14 @@ import java.security.MessageDigest
 import java.util.EnumSet
 
 /**
- * A list of directories used by version control systems to store metadata.
+ * A list of directories used by version control systems to store metadata. The list covers also version control systems
+ * not supported by ORT, because such directories should never be considered.
  */
 val VCS_DIRECTORIES = listOf(
     ".git",
     ".hg",
     ".repo",
     ".svn",
-    // Keep CVS directories in here even if ORT itself does not support CVS anymore.
     "CVS",
     "CVSROOT"
 )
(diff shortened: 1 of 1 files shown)


utils/common/src/main/kotlin/Utils.kt
```diff
@@ -26,13 +26,14 @@ import java.util.EnumSet
 
 /**
  * A list of directories used by version control systems to store metadata.
+ * The list covers also version control systems not supported by ORT, because such directories
```
sschuberth, line 29: Commit title: I guess a more fitting type would be "docs".

sschuberth: Does this deliberately start in a new line instead of continuing the first line of the comment? If so, why?

utils/common/src/main/kotlin/Utils.kt
```diff
@@ -26,13 +26,14 @@ import java.util.EnumSet
...
+ * should never be scanned.
```
sschuberth, line 30: `VCS_DIRECTORIES` are not only ignored by the scanner, but also skipped by the analyzer, so we should generalize this e.g. by saying "... never be *considered*."

utils/common/src/main/kotlin/Utils.kt
```diff
@@ -30,6 +30,7 @@ import java.util.EnumSet
  * should never be scanned.
  */
 val VCS_DIRECTORIES = listOf(
+    ".bzr",
```
sschuberth, line 33: Commit title: Consider using "feat", "fix", or "chore" as the type.

sschuberth (changes requested):

sschuberth (approved):

Issue #6945, opened 2023-04-28
Report file findings in SpdxDocuments / support complete file listings in SBOMs 


fviernau:
MVP: The `SpdxDocumentReporter` can (optionally) output license and copyright findings including
           file locations and the (mandatory) file hash as `SHA1`.

To implement this it makes sense to integrate a file listing for all scanned provenances into `OrtResult`. Properly doing
that (without adding too much redundancy) in turn requires to refactor the scan result representation in `OrtResult` to:

1. Store scan results by their provenance (not anymore by `id`).
2. Store file listings (path + `SHA1`) only once, also by their provenance.

The plan for the implementation is as follows:

- [x] https://github.com/oss-review-toolkit/ort/issues/6943
- [x] https://github.com/oss-review-toolkit/ort/issues/5950
- [x] https://github.com/oss-review-toolkit/ort/issues/7192
- [x] https://github.com/oss-review-toolkit/ort/issues/7206
- [x] https://github.com/oss-review-toolkit/ort/issues/6944
- [x] https://github.com/oss-review-toolkit/ort/issues/6947
- [x] https://github.com/oss-review-toolkit/ort/issues/6948

bugs:
- [x] https://github.com/oss-review-toolkit/ort/issues/7225
- [x] https://github.com/oss-review-toolkit/ort/issues/7231
- [x] https://github.com/oss-review-toolkit/ort/issues/7263

follow-ups:
- [ ] https://github.com/oss-review-toolkit/ort/issues/7155

Some potential future scanner improvements I've found whilst working on this epic:

1. If a Git-Repo manifest maps a Git repository to the root dir, then ORT returns VcsType.Git while it should be
    VcsType.GitRepo for the root provenance.
2. Cached Provenance resolution results may be outdated, so it may happen that provenance
    Resolution results are successful while the scanner can not download. Causes can be,
    - server no more available
    - revision no more available (force-push)
    - network hick-ups
    In that case we should at least have an issue reporter. It probably makes sense to re-resolve provenance in that
   case.
3. Provenance resolution could stop rely on recursive cloning, as performance improvement.
4. In case of GitRepo, the data model for the subrepositories is not always sufficient to tell which
    repository is a sub-repository. In particular, if the repo manifest maps one git repo under a sub directory of another.
5. Not verified: On the first glance it seems that for git repo, everything is scanned twice. One scan for the GitRepo
    provenance covering all git repositories and then seperate scans for all git repositories.

[...]

PR #6970, opened 2023-05-05
scanner: Create and store file listings for each resolved provenance

fviernau:
Create and store the file listings `(path, sha1)` for each resolved provenance. 
These can be consumed in future iterations (see #6945) for improving the generated SBOMs.

Resolves: #6943.

I've scanned the HEAD of the main branch of npm mime-types which resolved to 335 provenances. The total size of all file listings in the Postgres database is 5605372 bytes. So, about 16.4kb per file listing in average.

BREAKING CHANGE: One does not have to, but should configure a storage backend in `~/.ort/config/config.yml`.
                                      If one doesn't do so, ORT falls back to storing the file listings under `~/.ort` which may not be
                                      desired. See the modified `reference.yml` for how the new configuration feature looks like.

### DIFF
--- a/model/src/main/kotlin/config/FileListingStorageConfiguration.kt
+++ b/model/src/main/kotlin/config/FileListingStorageConfiguration.kt
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model.config
+
+import org.apache.logging.log4j.kotlin.Logging
+
+import org.ossreviewtoolkit.model.utils.DatabaseUtils
+import org.ossreviewtoolkit.model.utils.FileProvenanceFileStorage
+import org.ossreviewtoolkit.model.utils.PostgresProvenanceFileStorage
+import org.ossreviewtoolkit.model.utils.ProvenanceFileStorage
+import org.ossreviewtoolkit.utils.ort.ortDataDirectory
+import org.ossreviewtoolkit.utils.ort.storage.FileStorage
+import org.ossreviewtoolkit.utils.ort.storage.LocalFileStorage
+
+private const val TABLE_NAME = "provenance_file_listings"
+private const val FILENAME = "file_listings.xz"
+
+data class FileListingStorageConfiguration(
+    /**
+     * Configuration of the [FileStorage] used for storing the file listings.
+     */
+    val fileStorage: FileStorageConfiguration? = null,
+
+    /**
+     * Configuration of the [PostgresProvenanceFileStorage] used for storing the file listings.
+     */
+    val postgresStorage: PostgresStorageConfiguration? = null
+) {
+    private companion object : Logging
+
+    init {
+        if (fileStorage != null && postgresStorage != null) {
+            FileListingStorageConfiguration.logger.warn {
+                "'fileStorage' and 'postgresStorage' are both configured but only one storage can be used. " +
+                        "Using 'fileStorage'."
+            }
+        }
+    }
+}
+
+fun FileListingStorageConfiguration?.createStorage(): ProvenanceFileStorage =
+    when {
+        this?.fileStorage != null -> FileProvenanceFileStorage(
+            storage = fileStorage.createFileStorage(),
+            filename = FILENAME
+        )
+        this?.postgresStorage != null -> PostgresProvenanceFileStorage(
+            dataSource = DatabaseUtils.createHikariDataSource(
+                config = postgresStorage.connection,
+                applicationNameSuffix = "file-listings"
+            ),
+            tableName = TABLE_NAME
+        )
+        else -> FileProvenanceFileStorage(
+            storage = LocalFileStorage(ortDataDirectory.resolve("scanner/file-listings")),
+            filename = FILENAME
+        )
+    }
--- a/scanner/src/main/kotlin/FileListing.kt
+++ b/scanner/src/main/kotlin/FileListing.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.scanner
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
+import com.fasterxml.jackson.databind.util.StdConverter
+
+import java.util.SortedSet
+
+import org.ossreviewtoolkit.scanner.FileListing.FileEntry
+import org.ossreviewtoolkit.utils.common.StringSortedSetConverter
+import org.ossreviewtoolkit.utils.common.getDuplicates
+
+/**
+ * The model to store a file listing for a resolved provenance.
+ */
+internal data class FileListing(
+    /**
+     * The set of glob expressions which have been used to match directories to be excluded from the file listing.
+     */
+    @JsonSerialize(converter = StringSortedSetConverter::class)
+    val ignorePatterns: Set<String>,
+
+    /**
+     * The set of files contained in the resolved provenance, excluding files which are within a directory ignored by
+     * [ignorePatterns].
+     */
+    @JsonSerialize(converter = FileEntrySortedSetConverter::class)
+    val files: Set<FileEntry>
+) {
+    data class FileEntry constructor(
+        val path: String,
+        val sha1: String
+    )
+
+    init {
+        val duplicates = files.getDuplicates { it.path }.keys
+
+        require(duplicates.isEmpty()) {
+            "The file listing contains duplicate paths which is not allowed: ${duplicates.joinToString()}."
+        }
+    }
+}
+
+private class FileEntrySortedSetConverter : StdConverter<Set<FileEntry>, SortedSet<FileEntry>>() {
+    override fun convert(value: Set<FileEntry>) = value.toSortedSet(compareBy { it.path })
+}
--- a/scanner/src/main/kotlin/ScanController.kt
+++ b/scanner/src/main/kotlin/ScanController.kt
@@ -143,6 +143,18 @@ class ScanController(
         nestedProvenances.values.flatMapTo(mutableSetOf()) { it.getProvenances() }
 
     /**
+     * Return all provenances including sub-repositories associated with the identifiers of the packages they belong to.
+     */
+    fun getIdsByProvenance(): Map<KnownProvenance, Set<Identifier>> =
+        buildMap<_, MutableSet<Identifier>> {
+            getNestedProvenancesByPackage().forEach { (pkg, nestedProvenance) ->
+                nestedProvenance.getProvenances().forEach { provenance ->
+                    getOrPut(provenance) { mutableSetOf() } += pkg.id
+                }
+            }
+        }
+
+    /**
      * Get all provenances for which no scan result for the provided [scanner] is available.
      */
     private fun getMissingProvenanceScans(scanner: ScannerWrapper, nestedProvenance: NestedProvenance) =
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -45,7 +45,6 @@ import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
 import org.ossreviewtoolkit.model.ScanSummary
 import org.ossreviewtoolkit.model.ScannerRun
-import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.Options
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
@@ -192,7 +191,7 @@ class Scanner(
                 }.onFailure {
                     controller.addProvenanceResolutionIssue(
                         pkg.id,
-                        Issue(source = TOOL_NAME, severity = Severity.ERROR, message = it.collectMessages())
+                        Issue(source = TOOL_NAME, message = it.collectMessages())
                     )
                 }
             }
@@ -222,7 +221,6 @@ class Scanner(
                             id,
                             Issue(
                                 source = TOOL_NAME,
-                                severity = Severity.ERROR,
                                 message = "Could not resolve nested provenance for package " +
                                         "'${id.toCoordinates()}': ${it.collectMessages()}"
                             )
@@ -610,8 +608,7 @@ class Scanner(
                         Issue(
                             source = "Downloader",
                             message = "Could not create file archive for " +
-                                    "'${pkg.id.toCoordinates()}': ${it.collectMessages()}",
-                            severity = Severity.ERROR
+                                    "'${pkg.id.toCoordinates()}': ${it.collectMessages()}"
                         )
                     )
                 }
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -49,12 +49,14 @@ import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.Options
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.config.createFileArchiver
+import org.ossreviewtoolkit.model.config.createStorage
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.scanner.provenance.NestedProvenance
 import org.ossreviewtoolkit.scanner.provenance.NestedProvenanceResolver
 import org.ossreviewtoolkit.scanner.provenance.NestedProvenanceScanResult
 import org.ossreviewtoolkit.scanner.provenance.PackageProvenanceResolver
 import org.ossreviewtoolkit.scanner.provenance.ProvenanceDownloader
+import org.ossreviewtoolkit.scanner.utils.FileListingResolver
 import org.ossreviewtoolkit.utils.common.collectMessages
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.ort.Environment
@@ -62,6 +64,7 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
 
 const val TOOL_NAME = "scanner"
 
+@Suppress("TooManyFunctions")
 class Scanner(
     val scannerConfig: ScannerConfiguration,
     val downloaderConfig: DownloaderConfiguration,
@@ -90,6 +93,11 @@ class Scanner(
     }
 
     private val archiver = scannerConfig.archive.createFileArchiver()
+
+    private val fileListingResolver = FileListingResolver(
+        storage = scannerConfig.fileListingStorage.createStorage(),
+        provenanceDownloader = provenanceDownloader
+    )
 
     suspend fun scan(ortResult: OrtResult, skipExcluded: Boolean, labels: Map<String, String>): OrtResult {
         val startTime = Instant.now()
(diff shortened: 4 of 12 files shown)


scanner/src/main/kotlin/utils/FileListingResolver.kt
```diff
@@ -0,0 +1,106 @@
...
+                )
+                child.isDirectory -> if (!IGNORED_DIRECTORY_MATCHER.matches(path)) {
+                    queue += child
+                } else {
+                    println("ignoring: $path")
+                }
+                else -> return@forEach
+            }
+        }
+    }
```
sschuberth, line 102: Why not simply use `File.walk()`instead of a custom queue implementation?

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
```diff
@@ -0,0 +1,78 @@
...
+     * Configuration of the [PostgresProvenanceFileStorage] used for storing the file listings.
+     */
+    val postgresStorage: PostgresStorageConfiguration? = null
+) {
+    private companion object : Logging
+
+    fun createStorage(): ProvenanceFileStorage =
+        when {
+            fileStorage != null -> {
+                if (postgresStorage != null) {
```
sschuberth, line 51: Personally, I'd prefer a more "direct" check for this in an `init`-block like done in `FileArchiverConfiguration`.

sschuberth: In general, I wonder whether this should resemble `FileArchiverConfiguration` more, also regarding the extension function for the creation, and the general `enable` flag. I'm curious about @mnonnenmacher's opinion here.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -577,6 +587,41 @@ class Scanner(
...
+
+        val duration = measureTime {
+            val previousIndex = AtomicInteger(0)
+
+            provenancesMissingFileListings.parallelStream().forEach { provenance ->
+                val index = previousIndex.incrementAndGet()
+
+                logger.info { "Creating file listing for provenance $index of ${provenancesMissingFileListings.size}." }
+
+                kotlin.runCatching {
```
sschuberth, line 604: This should just be `runCatching`.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -577,6 +587,41 @@ class Scanner(
...
+                    fileListingResolver.resolve(provenance)
+                }.onFailure {
+                    idsByProvenance.getValue(provenance).forEach { id ->
+                        controller.addIssue(
+                            id,
+                            Issue(
+                                source = "Downloader",
+                                message = "Could not create file listing for " +
+                                        "'${id.toCoordinates()}': ${it.collectMessages()}",
+                                severity = Severity.ERROR
```
sschuberth, line 614: This is the default and can be omitted.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -577,6 +587,41 @@ class Scanner(
...
+                            )
+                        )
+                    }
+                }
+            }
+        }
+
+        logger.info("Created file listing for ${provenancesMissingFileListings.size} provenances in $duration.")
```
sschuberth, line 622: Use `{}` for lazy evaluation.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -577,6 +587,41 @@ class Scanner(
...
+    private fun createMissingFileListings(controller: ScanController) {
+        val idsByProvenance = controller.getIdsByProvenance()
+        val provenancesMissingFileListings = idsByProvenance.keys.filterNot { fileListingResolver.has(it) }
+
+        logger.info { "Creating file listings for ${provenancesMissingFileListings.size} provenances." }
+
+        val duration = measureTime {
+            val previousIndex = AtomicInteger(0)
+
+            provenancesMissingFileListings.parallelStream().forEach { provenance ->
```
sschuberth, line 599: As `parallelStream()` is limited to JVM targets, I'd prefer a more Kotlin-style approach with coroutines, similar to like recently done for the downloader CLI:

https://github.com/oss-review-toolkit/ort/blob/eb88bd776c0accb75f3741aaf398d858c4428739/cli/src/main/kotlin/commands/DownloaderCommand.kt#L312-L321

That'd also allow you to get rid of the manual counting with `AtomicInteger`.

sschuberth (changes requested):

scanner/src/main/kotlin/FileListing.kt
```diff
@@ -0,0 +1,60 @@
...
+        val sha1: String
+    )
+
+    init {
+        val duplicates = files.getDuplicates { it.path }.keys
+
+        require(duplicates.isEmpty()) {
+            "The file listing contains duplicate paths which is not allowed: ${duplicates.joinToString()}."
+        }
+    }
```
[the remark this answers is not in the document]
fviernau: > Have you considered using a map with `path` as the key and hash(es) (or other file properties) as values to avoid the need to this check?

Yes I considered that. I didn't / don't find this check problematic - I thought the data structure as-is would be a tad more speaking.

scanner/src/main/kotlin/utils/FileListingResolver.kt
fviernau: I changed it to using `walk()`.

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
fviernau: > Personally, I'd prefer a more "direct" check for this in an `init`-block like done in `FileArchiverConfiguration`.

That mentioned init block is executed when ort config is loaded. I thought it would be better to only log it exactly when the storage gets created. I still think it's no a bad idea, but I do not have such a strong opinion on that, so I could move it to `init` if you want.

fviernau: > In general, I wonder whether this should resemble `FileArchiverConfiguration` more, also regarding the extension function for the creation, and the general `enable` flag. I'm curious about @mnonnenmacher's opinion here.

It would have been nice to hear some rationale for this. Anyhow, I deliberately have made it mandatory so that we can count on the file listings to be present for simplicity. Making the file listing optional I prefer to not be part of the (initial) implementation.

mnonnenmacher: > In general, I wonder whether this should resemble FileArchiverConfiguration more, also regarding the extension function for the creation, and the general enable flag. I'm curious about @mnonnenmacher's opinion here.

I need to have a look at the code tomorrow before I can answer that.

> Anyhow, I deliberately have made it mandatory so that we can count on the file listings to be present for simplicity. Making the file listing optional I prefer to not be part of the (initial) implementation.

IMO the file lists should be mandatory, this makes it easier to build the features that depend on it.

> One does not have to, but should configure a storage backend in ~/.ort/config/config.yml. If one doesn't do so, ORT falls back to storing the file listings under ~/.ort which may not be desired.

Off topic, but this reminds me of #5516 which would prevent that we have yet another place to configure the same database connection. I wonder if we really need the flexibility to configure all storage types for the scanner separately, or if one central configuration to configure either a file storage or a Postgres storage would be good enough.

fviernau: @mnonnenmacher I could add #5516 to my epic, e.g. as last clean-up ticket. Ok?

scanner/src/main/kotlin/FileListing.kt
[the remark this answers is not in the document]
sschuberth: > Do you think a map would be better?

I'm unsure myself, esp. when it comes to how the serialization looks, but that's probably anyway unimportant here. I was more considering the case when more properties, like a Git-style SHA1, would be added. Maybe if would be nicer to have paths to files and files properties more clearly "separated" via key and values.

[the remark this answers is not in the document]
sschuberth: Speaking of more properties to add, have you also considered to use (a set of) `Hash` classes instead of `String`s (and ensure that at least an entry with `HashAlgorithm.SHA1` is present)?

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
sschuberth: > > In general, I wonder whether this should resemble `FileArchiverConfiguration` more, also regarding the extension function for the creation, and the general `enable` flag. I'm curious about @mnonnenmacher's opinion here.
> 
> It would have been nice to hear some rationale for this.

IMO it's wrong that `FileListingStorageConfiguration` knows how to create `ProvenanceFileStorage`. It should be the opposite: `ProvenanceFileStorage` knowns how create itself given a `FileListingStorageConfiguration`. However, the latter is hard to implement cleanly as we can have either a `FileProvenanceFileStorage` or a `PostgresProvenanceFileStorage`. So we need to move creation to a "higher level" / common place, which leads us back to `FileListingStorageConfiguration`. But *if* we do it in `FileListingStorageConfiguration`, I find it to be more clean if it's separated from the data class via an extension function like done in `FileArchiverConfiguration`.

sschuberth: > > Anyhow, I deliberately have made it mandatory so that we can count on the file listings to be present for simplicity. Making the file listing optional I prefer to not be part of the (initial) implementation.
> 
> IMO the file lists should be mandatory, this makes it easier to build the features that depend on it.

Why is it okay to have an optional file storage for license files, but it's not okay to make the storage for listings optional?

If the argument is simplicity, we should also make the license file storage mandatory, as anyway pretty much everybody seems to agree that the original wording of the license files needs to be retained.

scanner/src/main/kotlin/FileListing.kt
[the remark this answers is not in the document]
fviernau: > I'm unsure myself, esp. when it comes to how the serialization looks, but that's probably anyway unimportant here. I was more considering the case when more properties, like a Git-style SHA1, would be added. Maybe if would be nicer to have paths to files and files properties more clearly "separated" via key and values.

I believe a dedicated property is a "clear" separation already.
Adding further properties is out of scope of the this PR and the whole epic.
(This can be refactored quite easily if needed in the future)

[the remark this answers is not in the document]
fviernau: > Speaking of more properties to add, have you also considered to use (a set of) `Hash` classes instead of `String`s (and ensure that at least an entry with `HashAlgorithm.SHA1` is present)?

No, this is not needed for an MVP.

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
fviernau: > But if we do it in FileListingStorageConfiguration, I find it to be more clean if it's separated from the data class

I'm not following the arguments "separation" and "more clean, to me it's not obvious  

1. why an extension function is more clean
2. in what way the degree of separation is increased.

I'll change it anyway

scanner/src/main/kotlin/FileListing.kt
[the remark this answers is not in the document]
sschuberth: > I believe a dedicated property is a "clear" separation already.

That depends on whether you regard the path to be a property of the file, like the SHA1. Or if you believe the path to be an identifier for the file, and the file (contents) have properties separate from that, like hashes, modification time, etc.

> Adding further properties is out of scope of the this PR and the whole epic.
> (This can be refactored quite easily if needed in the future)

It's certainly needed, and it also corresponds to what we discussed some time ago: Any implementation of this epic must be prepared for adding additional hashes (like the mentioned Git-style SHA1, for Software Heritage interoperability).

> No, this is not needed for an MVP IMO.

Obviously it's not technically needed. The question is whether it makes sense / is prescient to already implement it like that. Or phrased differently, why does it make sense *not* to use the `Hash` class here, since we already use it elsewhere for serialization.

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
sschuberth: > > we should also make the license file storage mandatory, as anyway pretty much everybody seems to agree that the original wording of the license files needs to be retained.
> 
> Maybe, but can we leave this discussion out of scope here?

Sorry, no. I'm not saying we must already implement this now, but we should be allowed to discuss it, as it obviously has an impact on this PR.

I have a problem with bending arguments and changing minds every time it suites somebody's needs, just because there is pressure to implement a feature. Out of a sudden "simplicity" is an argument here, whereas in the past we were forced to add complexity by hiding several (IMO uncritical) features behind feature flags. I just miss the consistency and equality in the line of though here.

sschuberth: > I'm not following the arguments "separation" and "more clean, to me it's not obvious
> 
>     1. why an extension function is more clean
>     2. in what way the degree of separation is increased.

My answer to both is: Because the implementation is not part of the class body anymore. (I.e. it's more of a "visual" separation that expresses the intent better, as that's probably the best we can do here.)

scanner/src/main/kotlin/FileListing.kt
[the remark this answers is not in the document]
fviernau: > It's certainly needed, and it also corresponds to what we discussed some time ago: Any implementation of this epic must be prepared for adding additional hashes (like the mentioned Git-style SHA1, for Software Heritage interoperability).

As said, it's easy as-is already to add further properties. So, yes it is prepared well enough IMO.

> Obviously it's not technically needed. The question is whether it makes sense / is prescient to already implement it like that. Or phrased differently, why does it make sense not to use the Hash class here, since we already use it elsewhere for serialization.

It does make sense to not do it because it is not needed in scope of my epic.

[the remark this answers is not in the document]
sschuberth: > It does make sense to not do it because it is not needed in scope of my epic.

So is our policy now that everybody just strictly scratches their own itch, disregarding what other related use-cases others may have in the future, and leaving any additional work to them, even if it would not hurt to do it now? That's not the way at least I want to work on this project.

fviernau:
Looking at above discussion I have a hunch it may be more efficient to come to a conclusion in a short call. @sschuberth @mnonnenmacher  would you be in for that?

scanner/src/main/kotlin/FileListing.kt
```diff
@@ -0,0 +1,60 @@
+/*
```
fviernau, line 1: commit-msg: Don't mention "scanned" but say "each resolved provenance". Same for PR description.

scanner/src/main/kotlin/FileListing.kt
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.SortedSet
+
+import org.ossreviewtoolkit.scanner.FileListing.FileEntry
+import org.ossreviewtoolkit.utils.common.StringSortedSetConverter
+import org.ossreviewtoolkit.utils.common.getDuplicates
+
+/**
+ * The model to store file listings for provenances in a storage.
+ */
```
fviernau, line 33: The model to store a file listing for a resolved provenance.

scanner/src/test/assets/expected-file-listing.yml
```diff
@@ -0,0 +1,14 @@
+---
+ignored_directories:
```
fviernau, line 2: rename to : ignore_patterns

scanner/src/main/kotlin/FileListing.kt
```diff
@@ -0,0 +1,60 @@
...
+internal data class FileListing(
+    /**
+     * A set of glob expressions that match directory paths which have been excluded from the file listing.
+     */
+    @JsonSerialize(converter = StringSortedSetConverter::class)
+    val ignoredDirectories: Set<String>,
+
+    @JsonSerialize(converter = FileEntrySortedSetConverter::class)
```
fviernau, line 41: Add KDoc for `files`.

scanner/src/main/kotlin/provenance/ProvenanceDownloader.kt
```diff
@@ -39,7 +39,7 @@ import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 /**
  * An interface that provides functionality to download source code.
  */
-interface ProvenanceDownloader {
+fun interface ProvenanceDownloader {
```
fviernau, line 42: Try if FakeProvenanceDownloader can use it. If yes, make separate commit.

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
```diff
@@ -0,0 +1,75 @@
+/*
```
fviernau, line 1: commit-msg: scanned provenance -> resolved provenance

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
```diff
@@ -0,0 +1,75 @@
...
+        )
+        this?.postgresStorage != null -> PostgresProvenanceFileStorage(
+            dataSource = DatabaseUtils.createHikariDataSource(
+                config = postgresStorage.connection,
+                applicationNameSuffix = "file-listings"
+            ),
+            tableName = TABLE_NAME
+        )
+        else -> FileProvenanceFileStorage(
+            storage = LocalFileStorage(ortDataDirectory.resolve("scanner/provenance-file-listings")),
```
fviernau, line 72: just: "scanner/file-listings"

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -577,6 +586,43 @@ class Scanner(
...
+                            fileListingResolver.resolve(provenance)
+                        }.onFailure {
+                            idsByProvenance.getValue(provenance).forEach { id ->
+                                controller.addIssue(
+                                    id,
+                                    Issue(
+                                        source = "Downloader",
+                                        message = "Could not create file listing for " +
+                                                "'${id.toCoordinates()}': ${it.collectMessages()}",
+                                        severity = Severity.ERROR
```
fviernau, line 613: drop ` severity = Severity.ERROR`.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -192,7 +201,7 @@ class Scanner(
                 }.onFailure {
                     controller.addProvenanceResolutionIssue(
                         pkg.id,
-                        Issue(source = TOOL_NAME, severity = Severity.ERROR, message = it.collectMessages())
+                        Issue(source = TOOL_NAME, message = it.collectMessages())
```
fviernau, line 204: this is done in wrong commit.

scanner/src/main/kotlin/provenance/ProvenanceDownloader.kt
fviernau: The other implementation (FakeProvenanceDownloader) could be turned into a function which return an anonymous class / instance. However, this does not work with the spies used with that instance.

fviernau:
@mnonnenmacher @sschuberth I've incorporated the changes we've agreed upon in our call yesterday.

scanner/src/test/kotlin/utils/FileListingResolverTest.kt
```diff
@@ -0,0 +1,65 @@
...
+                    ".git/index",
+                    "LICENSE",
+                    "src/cli/main.cpp"
+                )
+            },
+        )
+
+        val fileListing = resolver.resolve(ArtifactProvenance(sourceArtifact = RemoteArtifact.EMPTY))
+
+        fileListing.toYaml() shouldBe File("src/test/assets/expected-file-listing.yml").readText()
```
mnonnenmacher, line 53: It would be good to have some more fine grained test, e.g. to verify that the exclude patterns work as expected. I'm fine with adding them later.
If more tests are added, it might be better for the overview to inline the expected results in the tests instead of using separate files.

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
```diff
@@ -0,0 +1,75 @@
...
+    val postgresStorage: PostgresStorageConfiguration? = null
+) {
+    private companion object : Logging
+
+    init {
+        if (fileStorage != null && postgresStorage != null) {
+            FileListingStorageConfiguration.logger.warn {
+                "'fileStorage' and 'postgresStorage' are both configured but only one storage can be used. " +
+                        "Using 'fileStorage'."
+            }
```
mnonnenmacher, line 53: Such warnings are easily overlooked, why not throw an exception?

scanner/src/main/kotlin/Scanner.kt
mnonnenmacher: What is the rationale for this? The default might change, but this would still be an error.

scanner/src/main/kotlin/ScanController.kt
```diff
@@ -142,6 +142,18 @@ class ScanController(
...
+     * Return all provenances including sub-repositories associated with the identifiers of the packages they belong to.
+     */
+    fun getIdsByProvenance(): Map<KnownProvenance, Set<Identifier>> =
+        buildMap<_, MutableSet<Identifier>> {
+            getNestedProvenancesByPackage().forEach { (pkg, nestedProvenance) ->
+                nestedProvenance.getProvenances().forEach { provenance ->
+                    getOrPut(provenance) { mutableSetOf() } += pkg.id
+                }
+            }
+        }
```
mnonnenmacher, line 155: Would `getPackagesConsolidatedByProvenance().mapValues { it.value.id }` create the same results with less code?

mnonnenmacher (changes requested):

mnonnenmacher:
> > What is the rationale for this? The default might change, but this would still be an error.
> 
> @mnonnenmacher the removal of the explicit `Severity.ERROR` happened on my request to align with the code base. If we cannot rely on the value of default arguments, it makes little sense to use default values at all...

Then I would argue for removing the default value, the issue severity is something I would like to see on the call side.

scanner/src/main/kotlin/Scanner.kt
fviernau: @sschuberth was asking in this review to not pass redundant default values. IIRC rationale was consistency.

scanner/src/test/kotlin/utils/FileListingResolverTest.kt
fviernau: I'll do it in a follow up PR then.

model/src/main/kotlin/config/FileListingStorageConfiguration.kt
fviernau: This is consistent with the FileArchiverConfiguration. There have been long discussions about this class in this PR, can we leave as-is now. This as-is IIUC what was agreed with @sschuberth .

fviernau:
> Then I would argue for removing the default value, the issue severity is something I would like to see on the call side.

Are you fine with de-scoping this from this PR then. I don't like changing this back and forth. In my view this is now a different topic. Agreed?

PR #6971, opened 2023-05-05
Unified test diff

sschuberth:


sschuberth:
@oss-review-toolkit/kotlin-devs here's something I wanted to do for a very long time: Make the diff to expected analyzer results more comprehensible and easier to apply without running the test locally. The `PipFunTest` deliberately breaks, but the diff to the expected result is now in a unified diff format that's compatible with `git apply`, which means you can simply copy the diff from the log to the clipboard, and then locally run `wl-paste | git apply` (on Wayland) to apply (but not commit) the patch required to update to new expected results.

sschuberth:
You can see some example output at https://github.com/oss-review-toolkit/ort/actions/runs/4892317433/jobs/8733935809?pr=6971#step:6:1882.

sschuberth:
PS: The last commit in this PR was already created via the unified diff / `git apply`.

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ clikt = "3.5.2"
 commonsCompress = "1.23.0"
 cvssCalculator = "1.4.1"
 cyclonedx = "7.3.2"
+diffUtils = "4.12"
```
fviernau, line 20: Does the "click to see diff" feature in IntelliJ still work with this new matcher?

plugins/package-managers/nuget/src/funTest/assets/dotnet-with-csproj-and-nuget-config-output.yml
```diff
@@ -23,17 +23,19 @@ packages:
 - id: "NuGet::xunit.analyzers:1.2.0-pre.7"
   purl: "pkg:nuget/xunit.analyzers@1.2.0-pre.7"
   authors:
-  - "jnewkirk, bradwilson, marcind"
-  declared_licenses: []
-  declared_licenses_processed: {}
+  - "jnewkirk,bradwilson,marcind"
```
fviernau, line 26: Is this commit related to other commits of this PR?

sschuberth: Not strictly, see the [note above](https://github.com/oss-review-toolkit/ort/pull/6971#issuecomment-1539504175). The commit was created via the mechanism introduced in the PR, as a coincident showcase basically.

gradle/libs.versions.toml
```diff
@@ -17,6 +17,7 @@ clikt = "3.5.2"
...
 diskLruCache = "2.0.2"
 exposed = "0.41.1"
 flexmark = "0.64.4"
```
fviernau, line 23: commit-msg: Please mention that "click to see diff" does not work as of this PR by default anymore.
And that one has to change the value manually "simple".

Also, it'd be nice to have a rationale why the default behavior is changed.

fviernau (changes requested):

gradle/libs.versions.toml
sschuberth: > Please mention that "click to see diff" does not work as of this PR by default anymore.

But that's not the case. Kotest's own default is to use [non-simple ("condensed") output for large strings](https://github.com/kotest/kotest/blob/4d11dbf9cc80c3226c6add9eda59f51fa4682db5/kotest-assertions/kotest-assertions-shared/src/commonMain/kotlin/io/kotest/assertions/eq/StringEq.kt#L57), but we did and still do override that with ["simple" when running from IntelliJ](https://github.com/oss-review-toolkit/ort/blob/6c479e2712945b3f9e7a20587a5a741668528c31/build.gradle.kts#L123) (as Kotest's own IntelliJ detection stopped working for newer IntelliJ versions).

In the latest iteration of this PR, the new unified diff is only used of the `kotest.assertions.multi-line-diff` system property is set to "unified" (which is a value that currently only we understand). And that value is only set when running functional tests via Batect.

TL;DR the default behavior did not change for local development with IntelliJ.

mnonnenmacher:
> @oss-review-toolkit/kotlin-devs here's something I wanted to do for a very long time: Make the diff to expected analyzer results more comprehensible and easier to apply without running the test locally. The `PipFunTest` deliberately breaks, but the diff to the expected result is now in a unified diff format that's compatible with `git apply`, which means you can simply copy the diff from the log to the clipboard, and then locally run `wl-paste | col -bx | git apply` (on Linux with Wayland) to apply (but not commit) the patch required to update to new expected results.

This is great! Could you document this somewhere, for example in https://github.com/oss-review-toolkit/ort#testing?

sschuberth:
> Could you document this somewhere, for example in https://github.com/oss-review-toolkit/ort#testing?

Here you go: https://github.com/oss-review-toolkit/ort/pull/6981

PR #6985, opened 2023-05-10
fix(SpdxFile): Default `licenseInfoInFiles` to `NOASSERTION`

MarcelBochtler:
According to the SPDX 2.3 specification [1] the field `licenseInfoInFiles` is not mandatory, but should default to `NOASSERTION`. While the SPDX 2.2 specification does not mention the default value, the field is also optional in this version [2].

[1]: https://github.com/spdx/spdx-spec/blob/5c893ae8010d5d972ecdf1af953a95d9f53de415/schemas/spdx-schema.json#L561-L574
[2]: https://github.com/spdx/spdx-spec/blob/2dffb51097904e03d5e9b19782e9d1b5cd7c9682/schemas/spdx-schema.json#L453-L459

### DIFF
--- a/utils/spdx/src/main/kotlin/model/SpdxFile.kt
+++ b/utils/spdx/src/main/kotlin/model/SpdxFile.kt
@@ -102,7 +102,7 @@ data class SpdxFile(
      * The license information found in this file. To represent a not present value [SpdxConstants.NONE] or
      * [SpdxConstants.NOASSERTION] must be used.
      */
-    val licenseInfoInFiles: List<String>,
+    val licenseInfoInFiles: List<String> = listOf(SpdxConstants.NOASSERTION),
 
     /**
      * License notices or other such related notices found in the file. This may include copyright statements.


utils/spdx/src/main/kotlin/model/SpdxFile.kt
```diff
@@ -99,10 +99,10 @@ data class SpdxFile(
     val licenseConcluded: String,
 
     /**
-     * The license information found in this file. To represent a not present value [SpdxConstants.NONE] or
-     * [SpdxConstants.NOASSERTION] must be used.
+     * The license information found in this file. According to the SPDX 2.3.1 spec, a not present value represents
```
sschuberth, line 102: Note that [we officially only support SPDX 2.2](https://github.com/oss-review-toolkit/ort/blob/355f239c0200dfb11ff2b3f306257c8384ca505c/utils/spdx/src/main/kotlin/model/SpdxDocument.kt#L29), so quoting 2.3 specs here seems to be a bit misplaced.

sschuberth: ... and [the 2.2 specs](https://github.com/spdx/spdx-spec/blob/a05c12a2dd4652b1396fd2659f2cd3ea1f37faba/schemas/spdx-schema.json#L535) don't mention `NOASSERTION` explicitly.

That said, I'm personally with with making this default to `NOASSERTION`, but in the same run we should probably disallow the property to be empty, to always explicitly express the intent.

tsteenbe: SPDX 2.2.1 defines `NOASSERTION` for `licenseInfoInFiles` see https://github.com/spdx/spdx-spec/blob/development/v2.2.1/chapters/package-information.md#714-all-licenses-information-from-files-field-

SPDX 2.2.1 is the ISO version of the specification so I recommend we simply upgrade our SPDX generator from 2.2 to 2.2.1.

sschuberth: > `licenseInfoInFiles` see https://github.com/spdx/spdx-spec/blob/development/v2.2.1/chapters/package-information.md#714-all-licenses-information-from-files-field-

@tsteenbe you're quoting "licenseInfo*In*Files" but linking to "licenseInfo*From*Files" docs, but these are two different fields, right?

sschuberth:
@MarcelBochtler I propose this commit message instead:

```
According to the SPDX 2.3 specification [1] the `licenseInfoInFiles` field
is optional and its absence implies `NOASSERTION`. While the SPDX 2.2
specification [2] is not so clear about this, assume this did not change
but was just clarified in version 2.3, and make the meaning of an absent
value explicit.

[1]: https://github.com/spdx/spdx-spec/blob/v2.3/schemas/spdx-schema.json#L561-L568
[2]: https://github.com/spdx/spdx-spec/blob/v2.2/schemas/spdx-schema.json#L453-L461
```

And I'd also leave the comment for the property untouched.

PR #6994, opened 2023-05-12
Use the term "file list" instead of "file listing"

fviernau:
The term has been introduced in [1]. It was agreed in the code review to follow up on it with this renaming.

[1] https://github.com/oss-review-toolkit/ort/pull/6970

Context: #6943.

BREAKING-CHANGE: The local file storage location, postgres table name for storing these lists and also the 
                                      configuration property in name in `config.yml` have been 
                                      renamed. The previous name existed only for a very short amount of time, so it'S
                                      highly unlikedly someone needs to migrate this.


model/src/main/kotlin/config/FileListStorageConfiguration.kt
```diff
@@ -29,33 +29,33 @@ import org.ossreviewtoolkit.utils.ort.ortDataDirectory
 import org.ossreviewtoolkit.utils.ort.storage.FileStorage
 import org.ossreviewtoolkit.utils.ort.storage.LocalFileStorage
 
-private const val TABLE_NAME = "provenance_file_listings"
-private const val FILENAME = "file_listings.xz"
+private const val TABLE_NAME = "provenance_file_lists"
+private const val FILENAME = "file_lists.xz"
```
mnonnenmacher, line 33: This should also be changed to singular because it contains only one file list, right?

mnonnenmacher (changes requested):

mnonnenmacher (approved):

model/src/main/kotlin/config/FileListStorageConfiguration.kt
```diff
@@ -30,32 +30,32 @@ import org.ossreviewtoolkit.utils.ort.storage.FileStorage
 import org.ossreviewtoolkit.utils.ort.storage.LocalFileStorage
 
 private const val TABLE_NAME = "provenance_file_listings"
```
MarcelBochtler, line 32: @fviernau, was it intentional that the table name continued to be named `provenance_file_listings` instead of `provenance_file_lists`?

Can we expect that this table name stays as-is now?

fviernau: In my merged renaming PR I just missed aligning the table name. Shall I file a PR quickly to fix it up ?

PR #6999, opened 2023-05-15
ScanSummary: Remove default value for `snippetFindings`

fviernau:
See individual commits.


model/src/main/kotlin/ScanSummary.kt
```diff
@@ -60,21 +60,24 @@ data class ScanSummary(
     /**
```
nnobelis, line 60: While I am ok with your change, please note that @sschuberth raised the same concerns for consistency. However I though it could be skipped here because few scanners need snippet findings:
https://github.com/oss-review-toolkit/ort/pull/6791#discussion_r1191042314

nnobelis (dismissed):

sschuberth (dismissed):

sschuberth (dismissed):

Issue #6235, opened 2022-12-14
model: Eliminate sorted collections


fviernau:
The use of sorted collections has been introduced with the goal of having entries serialized in some sorted order.
This approach has the pitfall that it requires

1. Model classes to implement the Comparable interface which
    - brings in contractual requirements with equals, which in turn adds complexity to the actual compareTo implementation
    - Tightly couples the sorting of the serializer with the model class
2. Use mutable as opposed to immutable collections
3. The client code to deal with sorted collections

It may be possible to arrive at the original goal without implementing Comparable interface / using sorted collection.
A concrete solution is not yet known but options need to be investigated / evaluated first.

fviernau:
copied hint from @sschuberth :

> There's actually an existing issue in Jackson for making this easily configurable, see [FasterXML/jackson-databind#3166](https://github.com/FasterXML/jackson-databind/issues/3166).

sschuberth:
FYI, I'm working on a new proposal that uses Jackson converters to turn sets into sorted sets on serialization only. There are a few option questions about some technical details that could be discussed in an upcoming draft PR.

sschuberth:
> FYI, I'm working on a new proposal that uses Jackson converters to turn sets into sorted sets on serialization only.

See https://github.com/oss-review-toolkit/ort/pull/6244.

> There are a few option questions about some technical details that could be discussed in an upcoming draft PR.

Or maybe we can discuss the options in our Monday meeting.

PR #7000, opened 2023-05-15
ScanSummary: Do not use sorted set for `copyrightFingings`

fviernau:
See individual commits.

Part of #6235, also needed by #6945.


model/src/main/kotlin/utils/SortedSetConverters.kt
```diff
@@ -28,6 +28,10 @@ import java.util.SortedSet
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.Project
 
+class ComparableSortedSetConverter<T : Comparable<T>> : StdConverter<Set<T>, SortedSet<T>>() {
```
sschuberth, line 31: IMO a goal of https://github.com/oss-review-toolkit/ort/issues/6235 also is to stop `data class`es to implement  `Comparable`. With than in mind, I believe we should have a dedicated converter here that is specialized for `CopyrightFinding` and, similar to `PackageSortedSetConverter` below, hard-codes the comparator to `compareBy<CopyrightFinding>({ it.statement }, { it.location })`.

sschuberth (changes requested):

model/src/main/kotlin/utils/SortedSetConverters.kt
sschuberth: Our [comments overlapped](https://github.com/oss-review-toolkit/ort/pull/7000#discussion_r1194808739). If possible, I'd prefer to stop implementing `Comparable` already in this commit.

model/src/main/kotlin/utils/SortedSetConverters.kt
```diff
@@ -25,9 +25,14 @@ import com.fasterxml.jackson.databind.util.StdConverter
 
 import java.util.SortedSet
 
+import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.Project
 
+class CopyrightFindingSortedSetConverter : StdConverter<Set<CopyrightFinding>, SortedSet<CopyrightFinding>>() {
+    override fun convert(value: Set<CopyrightFinding>) = value.toSortedSet()
```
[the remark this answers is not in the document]
fviernau: I've stopped implementing comparable, but didn't inline the `COMPARATOR`. Please see if you like rationale in the updated commit. If you still prefer to also inline `COMPARATOR` I can do that, NP.

model/src/main/kotlin/utils/SortedSetConverters.kt
fviernau: Ok, let's continue discussion below then.

model/src/main/kotlin/CopyrightFinding.kt
```diff
@@ -32,10 +32,8 @@ data class CopyrightFinding(
      * The text location where the copyright statement was found.
      */
     val location: TextLocation
-) : Comparable<CopyrightFinding> {
+) {
     companion object {
-        private val COMPARATOR = compareBy<CopyrightFinding>({ it.statement }, { it.location })
+        val COMPARATOR = compareBy<CopyrightFinding>({ it.statement }, { it.location })
```
sschuberth, line 37: I'm ok with the code, but I believe we need to tweak the rationale in the commit message a bit as there are IMO some unclarities:

> For example, the comparators of `CopyrightFinding` and `LicenseFinding` both use the comparator of `TextLocation`.

Strictly speaking, not the *comparator* of `TextLocation` is used, but they rely on `TextLocation` to *implement* `Comparable`. That's why this is not a good argument to keep the constant vs the implementation.

> So, inlining the comarator of `TextLocation` would lead to redundancy.

My proposal was to "inline" the `COMPARATOR` constant of `CopyrightFinding`, not the one of `TextLocation`.

Anyway, here's the body that I'd propose:

    Only sort on serialization for human readability and reproducibility.
    While that allows to remove the implementation of the comparable interface,
    still keep the comparator constant for use in the converter and any place
    where `CopyrightFinding`s shall be shown to users.

sschuberth: PS: Do you plan to do the same thing for `LicenseFinding`, too?

sschuberth (changes requested):

model/src/main/kotlin/CopyrightFinding.kt
fviernau: > PS: Do you plan to do the same thing for `LicenseFinding`, too?

Sure. Will be the next PR.

fviernau: See https://github.com/oss-review-toolkit/ort/pull/7004.

PR #7002, opened 2023-05-16
Statistics / ScannerRun: Do not use sorted data structures

fviernau:
See individual commits.

Part of #6235.

### DIFF
--- a/reporter/src/main/kotlin/Statistics.kt
+++ b/reporter/src/main/kotlin/Statistics.kt
@@ -19,8 +19,8 @@
 
 package org.ossreviewtoolkit.reporter
 
-import java.util.SortedMap
-import java.util.SortedSet
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
 
 import org.ossreviewtoolkit.model.Issue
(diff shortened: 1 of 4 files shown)


reporter/src/main/kotlin/Statistics.kt
```diff
@@ -155,12 +158,14 @@ data class LicenseStatistics(
     /**
      * All declared licenses, mapped to the number of [Project]s and [Package]s they are declared in.
      */
-    val declared: SortedMap<String, Int>,
+    @JsonSerialize(converter = ComparableSortedMapConverter::class)
```
sschuberth, line 161: You should be able to just use `@JsonPropertyOrder(alphabetic = true)` instead (also see the `CopyrightStatementsProcessor`). Similar below.

sschuberth (changes requested):

reporter/src/main/kotlin/Statistics.kt
fviernau: Nice, that's a great hint! FYI: Above approach didn't work because @JsonValue in `Identifier` became ignored, maybe due to type erasure.

sschuberth (approved):

sschuberth:
@fviernau, the second commit needs a conventional type in the title.

PR #7004, opened 2023-05-16
ScanSummary: Do not use sorted set for `licenseFindings`

fviernau:
See individual commits.

Part of https://github.com/oss-review-toolkit/ort/issues/6235, also needed by https://github.com/oss-review-toolkit/ort/issues/6945.


evaluator/src/test/kotlin/TestData.kt
```diff
@@ -259,7 +259,7 @@ val ortResult = OrtResult(
                     provenance = UnknownProvenance,
                     scanner = ScannerDetails.EMPTY,
                     summary = ScanSummary.EMPTY.copy(
-                        licenseFindings = sortedSetOf(
+                        licenseFindings = mutableSetOf(
```
sschuberth, line 262: Should just be `setOf()`.

sschuberth (changes requested):

sschuberth (approved):

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
```diff
@@ -635,8 +636,11 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
     ) {
         val pathExcludes = getPathExcludes(id, scanResult.provenance)
         val licenseFindingCurations = getLicenseFindingCurations(id, scanResult.provenance)
+        // TODO: Decide whether the ordering in the output should be independent of the order of the input and
+        //       either remove below sortedWith() call, or replace it with a proper ordering mechanism for the whole
+        //       model.
         val curatedFindings = curationsMatcher.applyAll(scanResult.summary.licenseFindings, licenseFindingCurations)
-            .mapNotNullTo(mutableSetOf()) { it.curatedFinding }
+            .mapNotNullTo(mutableSetOf()) { it.curatedFinding }.sortedWith(LicenseFinding.COMPARATOR).toSet()
```
sschuberth, line 643: I propose to instead use:

    .mapNotNull { it.curatedFinding }.toSortedSet(LicenseFinding.COMPARATOR)

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
```diff
@@ -635,8 +636,11 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
     ) {
         val pathExcludes = getPathExcludes(id, scanResult.provenance)
         val licenseFindingCurations = getLicenseFindingCurations(id, scanResult.provenance)
+        // TODO: Decide whether the ordering in the output should be independent of the order of the input and
+        //       either remove below sortedWith() call, or replace it with a proper ordering mechanism for the whole
+        //       model.
```
sschuberth, line 641: Given that the evaluated model report is a bit special in the sense that it tries to "pre-compute" as much as possible for quick display in the web-app report, I think it's fair to stick sorting here (if that's the way things should be show in the web-app report).

sschuberth: So maybe replace the `TODO` here with a comment that explains that sorting is done here to avoid the need to sort in the web-app each time the report is loaded.

sschuberth (changes requested):

sschuberth (approved):

PR #7007, opened 2023-05-17
fix(fossid-webapp): Don't create issue when the URL cannot be mapped

nnobelis:

### DIFF
--- a/model/src/main/kotlin/utils/Extensions.kt
+++ b/model/src/main/kotlin/utils/Extensions.kt
@@ -143,6 +143,7 @@ enum class PurlType(private val value: String) {
     DEBIAN("deb"),
     DRUPAL("drupal"),
     GEM("gem"),
+    GENERIC("generic"),
     GITHUB("github"),
     GITLAB("gitlab"),
     GOLANG("golang"),
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -825,7 +825,7 @@ class FossId internal constructor(
                 val snippetProvenance = it.url?.let { url ->
                     ArtifactProvenance(RemoteArtifact(url, Hash.NONE))
                 } ?: UnknownProvenance
-                val purlType = it.url?.let { url -> urlToPackageType(url, issues)?.toString() } ?: "generic"
+                val purlType = it.url?.let { url -> urlToPackageType(url).toString() }
 
                 // TODO: FossID doesn't return the line numbers of the match, only the character range. One must use
                 //       another call "getMatchedLine" to retrieve the matched line numbers. Unfortunately, this is a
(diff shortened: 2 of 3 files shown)


model/src/main/kotlin/utils/Extensions.kt
```diff
@@ -143,6 +143,7 @@ enum class PurlType(private val value: String) {
     DEBIAN("deb"),
     DRUPAL("drupal"),
     GEM("gem"),
+    GENERIC("generic"),
```
sschuberth, line 146: Commit message: Please link directly to https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic

sschuberth: Also note that we've historically been using `a-name` (see further up in the `enum`) for more or less the same purpose. However, `a-name` is a [NexusIq / Sonatype specific thing](https://help.sonatype.com/iqserver/product-information/referencing-package-url-%28purl%29-and-component-identifiers#ReferencingPackageURL(purl)andComponentIdentifiers-a-name). I'll probably do a follow-up PR with docs to explain this.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -825,7 +825,7 @@ class FossId internal constructor(
                 val snippetProvenance = it.url?.let { url ->
                     ArtifactProvenance(RemoteArtifact(url, Hash.NONE))
                 } ?: UnknownProvenance
-                val purlType = it.url?.let { url -> urlToPackageType(url, issues)?.toString() } ?: "generic"
+                val purlType = it.url?.let { url -> urlToPackageType(url).toString() }
```
sschuberth, line 828: Commit message title: "Do not create *an* issue *if* ..."

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -893,11 +892,10 @@ class FossId internal constructor(
...
             else -> {
-                issues += FossId.createAndLogIssue(
-                    source = "FossId",
-                    message = "Cannot determine PURL type for url '$url' and provider '$provider'."
-                )
-                null
+                logger.warn {
+                    "Cannot determine PURL type for url '$url' and provider '$provider'. Falling back to 'generic'."
+                }
+                PurlType.GENERIC
```
sschuberth, line 898: I propose

                PurlType.GENERIC.also {
                    logger.warn {
                        "Cannot determine PURL type for url '$url' and provider '$provider'. Falling back to '$it'."
                    }
                }


to not hard-code the string representation of `PurlType.GENERIC`.

sschuberth (changes requested):

sschuberth:
@nnobelis commit-lint fails. You need to use reference-syntax for URLs to work around the line length limitation, like

```
[1]: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic
```

and of course refer to `[1]` from somewhere in the text.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
nnobelis: Sorry there has been a mix up and I fixed it in the wrong branch.

It has been pushed now:
https://github.com/oss-review-toolkit/ort/compare/5416fa09f20251e030b0cda9958537a90426d79b..0f39a48bd367abfc01442350cee78eae4088334d

nnobelis: Wait, it's not there :(

sschuberth (approved):

PR #7008, opened 2023-05-17
Support authentication for the OssIndex advisor

sschuberth:

### DIFF
--- a/advisor/src/main/kotlin/advisors/OssIndex.kt
+++ b/advisor/src/main/kotlin/advisors/OssIndex.kt
@@ -64,9 +64,16 @@ class OssIndex(name: String, config: OssIndexConfiguration) : AdviceProvider(nam
 
     private val service by lazy {
         OssIndexService.create(
-            url = config.serverUrl,
-            client = OkHttpClientHelper.buildClient()
+            config.serverUrl,
+            config.username,
+            config.password,
+            OkHttpClientHelper.buildClient()
         )
+    }
+
+    private val getComponentReport by lazy {
+        val hasCredentials = config.username != null && config.password != null
+        if (hasCredentials) service::getAuthorizedComponentReport else service::getComponentReport
     }
 
     override suspend fun retrievePackageFindings(packages: Set<Package>): Map<Package, AdvisorResult> {
@@ -82,7 +89,7 @@ class OssIndex(name: String, config: OssIndexConfiguration) : AdviceProvider(nam
             logger.debug { "Getting report for ${chunk.size} components (chunk ${index + 1} of ${chunks.size})." }
 
             runCatching {
-                val results = service.getComponentReport(ComponentReportRequest(chunk)).associateBy {
+                val results = getComponentReport(ComponentReportRequest(chunk)).associateBy {
                     it.coordinates
                 }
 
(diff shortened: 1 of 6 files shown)


tsteenbe (changes requested):
@sschuberth Could you include in this PR an update of [Advisor in reference.yml]( https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/resources/reference.yml#L104) so ORT users can easily find out authentication for the OssIndex is supported?

advisor/src/main/kotlin/advisors/OssIndex.kt
```diff
@@ -64,11 +64,18 @@ class OssIndex(name: String, config: OssIndexConfiguration) : AdviceProvider(nam
...
-            url = config.serverUrl,
-            client = OkHttpClientHelper.buildClient()
+            config.serverUrl,
+            config.username,
+            config.password,
+            OkHttpClientHelper.buildClient()
         )
     }
 
+    private val getComponentReport by lazy {
```
fviernau, line 74: What do you think about explicitly specifying the type of `getComponentReport()` for readability? 

Alternatively, this performance optimization can be dropped in favor of a (IMO more readable) function. To me it sacrifices readability without an obvious need.

What do you think?

fviernau (changes requested):

advisor/src/main/kotlin/advisors/OssIndex.kt
sschuberth: > What do you think?

I think it's not less readable than a wrapping function, and (letting the performance aspect aside) on the contrary more elegant.

Adding an explicit type would actually *reduce* readability, as this is a function reference to a `suspend` function, so the type would be something like `KSuspendFunction1<ComponentReportRequest, List<ComponentReport>>`.

Finally, using a function would require to pass `config` as an argument, as functions do not have access to constructor arguments like property initialization has.

So all in all, I'd prefer to keep it as-is.

*Edit:* Here the direct comparison

```
private val getComponentReport by lazy {
    val hasCredentials = config.username != null && config.password != null
    if (hasCredentials) service::getAuthorizedComponentReport else service::getComponentReport
}
```

vs

```
private suspend fun getComponentReport(config: OssIndexConfiguration, components: ComponentReportRequest): List<ComponentReport> {
    val hasCredentials = config.username != null && config.password != null
    return if (hasCredentials) service.getAuthorizedComponentReport(components) else service.getComponentReport(components)
}
```

to me the former is less repetitive, more elegant, and thus more readable and better.

fviernau: I had something like below in mind, however - let's just leave it as is.

```
private suspend fun ComponentReportRequest.execute(): List<ComponentReport> =
   when {
        config.hasCredentials() -> service.getAuthorizedComponentReport(this)
        else -> service.getComponentReport(this)
   }

private fun OssIndexConfiguration.hasCredentials() = config.username != null && config.password != null
```

fviernau (approved):

advisor/src/main/kotlin/advisors/OssIndex.kt
sschuberth: I like the extension function, though. Let me add that in a follow-up PR.

PR #6993, opened 2023-05-11
build(Gradle): Switch from the "dokka" to the "dokkatoo" plugin

sschuberth:
The official "dokka" plugin keeps having Gradle related issues, in
particular out of memory issues. Switch to the unofficial "dokkatoo"
successor [1] to resolve these issues. For simplicity, drop the
dokka-related external link configuration; it needs to be revisited if
and how this needs to be configured with dokkatoo, too. Also generalize
the names of the wrapper tasks to not explicitly mention "dokka"
anymore.

With dokkatoo, run e.g.

    ./gradlew dokkatooGeneratePublicationHtml

to generate HTML documentation files, or run

    ./gradlew docsHtmlJar

to bundle all HTML documentation into a JAR.

[1]: https://github.com/adamko-dev/dokkatoo

PR #7009, opened 2023-05-17
build(jitpack): Change the excluded Dokka task name

nnobelis:
In https://github.com/oss-review-toolkit/ort/pull/6993, the Dokka task has been replaced by `dokkatoo`. Hence the Jitpack configuration must be adapted.

Example of failing Jitpack build:
https://jitpack.io/com/github/oss-review-toolkit/ort/48888c6b17/build.log

### DIFF
--- a/.jitpack.yml
+++ b/.jitpack.yml
@@ -2,5 +2,5 @@ jdk:
   - openjdk11
 
 install:
-  # Exclude dokkaJavadoc tasks to save build time.
-  - ./gradlew --no-daemon --stacktrace -x dokkaJavadoc publishToMavenLocal
+  # Exclude generation of docs artifacts to save build time.
+  - ./gradlew --no-daemon --stacktrace -x docsJavadocJar publishToMavenLocal


.jitpack.yml
```diff
@@ -3,4 +3,4 @@ jdk:
 
 install:
   # Exclude dokkaJavadoc tasks to save build time.
-  - ./gradlew --no-daemon --stacktrace -x dokkaJavadoc publishToMavenLocal
+  - ./gradlew --no-daemon --stacktrace -x dokkatooGenerate publishToMavenLocal
```
sschuberth, line 6: I believe that should be `docsJavadocJar` instead. `dokkatooGenerate` is just an aggregator task that `publishToMavenLocal` does not even depend on.

sschuberth: The comment above also needs to be updated. While at it, please generalize its wording an say

> Exclude generation of docs artifacts to save build time.

sschuberth (changes requested):

sschuberth (approved):

PR #7015, opened 2023-05-17
Small `PackageManagerDependency` improvements

sschuberth:


analyzer/src/main/kotlin/managers/utils/PackageManagerDependencyHandler.kt
```diff
@@ -58,23 +55,20 @@ class PackageManagerDependencyHandler(
...
-                    name = jsonMapper.writeValueAsString(
-                        PackageManagerDependency(
-                            packageManager,
-                            definitionFile,
-                            scope,
-                            linkage
-                        )
-                    ).encode(),
-                    version = ""
+                    namespace = packageManager,
```
mnonnenmacher, line 58: Commit message: Can you explain the motivation behind this change, in particular why is the new approach better/simpler than the old one?

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #7018, opened 2023-05-19
Change the Snippet Provenance to KnownProvenance

nnobelis:

### DIFF
--- a/clients/fossid-webapp/src/main/kotlin/model/result/Snippet.kt
+++ b/clients/fossid-webapp/src/main/kotlin/model/result/Snippet.kt
@@ -38,7 +38,7 @@ data class Snippet(
 
     val file: String,
     val fileLicense: String?,
-    val url: String?,
+    val url: String,
     val hits: String,
     val size: Int?,
 
--- a/model/src/main/kotlin/utils/Snippet.kt
+++ b/model/src/main/kotlin/utils/Snippet.kt
@@ -19,7 +19,7 @@
 
 package org.ossreviewtoolkit.model.utils
 
-import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.KnownProvenance
 import org.ossreviewtoolkit.model.TextLocation
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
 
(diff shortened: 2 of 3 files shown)


clients/fossid-webapp/src/main/kotlin/model/result/Snippet.kt
```diff
@@ -38,7 +38,7 @@ data class Snippet(
 
     val file: String,
     val fileLicense: String?,
-    val url: String?,
+    val url: String,
```
sschuberth, line 41: Commit message:
- Maybe better "conservatively" instead of "defensively"?
- "guarantee*s*"

Generally I'm a bit confused by this sentence:

> While the database schema indeed has this property as nullable, the CLI tool used by the scanner guarantee it is never null.

Which datebase: FossID's internal one, or our scan storage? Which CLI tool: A FossID tool, or our own scan CLI subcommand? In any case, how can a *CLI tool* guarantee non-nullabilty, as this is purely a data model thing?

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/model/result/Snippet.kt
nnobelis: FossID internal database. FossID CLI tool.

This is the message from FossID support:

> Whilst using the CLI the url property will always be populated. That said, the SQL Schema for the workbench database allows a null value to be added. So during import for example, it could return a missing value.
> It could be missing if you're using the API to import something that has the url missing. However, performing a scan would call the CLI in the background, which should return you with the URL for each of the components found.

This answer is pretty confusing but here is my interpretation: If you perform a scan (through the API), it uses the CLI in background which always populate the url property.

nnobelis: Your opinion on this ?

sschuberth: > This answer is pretty confusing but here is my interpretation: If you perform a scan (through the API), it uses the CLI in background which always populate the url property.

Indeed confusing. Performing a scan (via the API) triggers the CLI in the background which "shadows" the fact that the URL could be null in the database? That sounds like a very weird design. The CLI should simply be a user of the API, not backwards.

Anyway, I'd not repeat any of that confusing implementation details (that could change any time) as part of the commit message. Just say that the FossID support confirmed that a null URL will never be exposed via the API response.

sschuberth (approved):

PR #7022, opened 2023-05-22
feat(model): Add Snippet additionalData

nnobelis:


model/src/main/kotlin/utils/Snippet.kt
```diff
@@ -49,5 +51,12 @@ data class Snippet(
...
      */
-    val licenses: SpdxExpression
+    val licenses: SpdxExpression,
+
+    /**
+     * A map for scanner specific snippet data that cannot be mapped into any generalized property, but still needs to
+     * be made available in the scan Summary.
+     */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    val additionalData: Map<String, String> = emptyMap()
```
fviernau, line 61: Can you explain the use case which uses the `id` please? (I've got slight doubts that such `custom data` field is really needed.

Note: I was wondering already before whether there is some kind of unique identifier or fingerprint, which could be used within ORT to use to match snippet curations against. Is the `ID` mentioned in the commit message, a stable unique identifier?

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -132,6 +132,18 @@ class FossId internal constructor(
         @JvmStatic
         internal val PROJECT_REVISION_LABEL = "projectVcsRevision"
```
fviernau, line 133: commit-msg: Please explain `why` the added data is needed. E.g. what are the use cases?

fviernau (changes requested):

model/src/main/kotlin/utils/Snippet.kt
```diff
@@ -19,6 +19,8 @@
 
 package org.ossreviewtoolkit.model.utils
 
+import com.fasterxml.jackson.annotation.JsonInclude
```
sschuberth, line 22: Commit message:

> information are

information *is*

model/src/main/kotlin/utils/Snippet.kt
sschuberth: > Is the `ID` mentioned in the commit message, a stable unique identifier?

Additionally, it would be interesting to know if SCANOSS has something similar, and if so, such an ID should probably become a first-class property.

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -1020,7 +1020,7 @@ private const val PROJECT = "fossId-test-project"
 /** A test revision. */
 private const val REVISION = "test-revision"
 
-/** The version to be reported by the FossID server. */
+/** The version to be reported bysnippe the FossID server. */
```
sschuberth, line 1023: Unintended change?

sschuberth (changes requested):

model/src/main/kotlin/utils/Snippet.kt
nnobelis: > Is the ID mentioned in the commit message, a stable unique identifier?

I cannot answer this question, as there is no documentation for FossID's data model.
However some thoughts ; Yes,`id` allows to uniquely identify a snippet, but the scope of this `id` is unclear: when two source files are matching the same snippet, do the latter have the same `id`? what if the matched source file lines are different ? Do the `id`s of the snippets change after a new import of the knowledge base ? All these questions have no answer, and this is the reason we went in ORT for the Purl 'external id'.

> Can you explain the use case which uses the id please? (I've got slight doubts that such custom data field is really needed.

The id of the snippet is required to query the FossID instance for the matched lines. Please have a look at:
https://github.com/oss-review-toolkit/ort/blob/a05529d3e4deae6d214aa9fb0003c898a796f8da/clients/fossid-webapp/src/main/kotlin/Extensions.kt#L279

This call will be used in a future PR to create a detailed snippet-specific report.

Please note that, _even_ if the id was not needed, the `Match Type` is also an important information that should be accessible. The match type is a FossID specific concept that is structurally different in ScanOSS and thus cannot be abstracted in the ORT snippet model IMO. 

Does it answer your doubts  ?

nnobelis:
@sschuberth 

> Additionally, it would be interesting to know if SCANOSS has something similar, and if so, such an ID should probably become a first-class property.


Unfortunately id does not:
```
{
  "src/main/java/com/vdurmont/semver4j/RequirementForScanOSS.java":[
  {
    "id": "snippet",
    "lines": "1-710",
    "oss_lines": "1-710",
    "matched": "98%",
    "file_hash": "6ff2427335b985212c9b79dfa795799f",
    "source_hash": "bd4bff27f540f4f2c9de012acc4b48a3",
    "file_url": "https://osskb.org/api/file_contents/6ff2427335b985212c9b79dfa795799f",
    "purl": [
      "pkg:github/vdurmont/semver4j"
    ],
    "vendor": "vdurmont",
    "component": "semver4j",
    "version": "3.1.0",
    "latest": "3.1.0",
    "url": "https://github.com/vdurmont/semver4j",
    "status": "pending",
    "release_date": "2019-09-13",
    "file": "src/main/java/com/vdurmont/semver4j/Requirement.java",
    "url_hash": "b92cd7e4d588747588d1b5fe1f8c664c",
    "licenses": [
      {
        "name": "MIT",
        "patent_hints": "no",
        "copyleft": "no",
        "checklist_url": "https://www.osadl.org/fileadmin/checklists/unreflicenses/MIT.txt",
        "osadl_updated": "2023-03-26T02:11:00+00:00",
        "source": "component_declared",
        "url": "https://spdx.org/licenses/MIT.html"
      }
    ],
    "server": {
      "version": "5.2.5",
      "kb_version": {
        "monthly": "23.03",
        "daily": "23.03.19"
      }
    }
  }
]
}
```

`id` behaves _a bit_ like FossID `MatchType`: it is either `file` or `snippet`. `file` would be FossID `MatchType.FULL` and `snippet` would be `MatchType.PARTIAL`. But then you see the difference mentioned above that ScanOSS does not consider a full match a snippet, while FossID does,
And no, there is no numeric identified in ScanOSS.

model/src/main/kotlin/utils/Snippet.kt
fviernau: > I cannot answer this question, as there is no documentation for FossID's data model.
However some thoughts ; Yes,id allows to uniquely identify a snippet, but the scope of this id is unclear: when two source files are matching the same snippet, do the latter have the same id? what if the matched source file lines are different ? Do the ids of the snippets change after a new import of the knowledge base ? All these questions have no answer, and this is the reason we went in ORT for the Purl 'external id'.

Would it make sense to create a separate resarch task, to 1. figure out what properties of the identifier would be ideal for a curation mechanism and 2. to talk to the owners of the snippet scanner whether such thing can be exposed, hinting that is a necessary precondition for enabling an efficient workflow with their scanner. ?

> The id of the snippet is required to query the FossID instance for the matched lines. Please have a look at:
[...]
> This call will be used in a future PR to create a detailed snippet-specific report.

Looking at 'model/src/main/kotlin/utils/Snippet.kt', there is the `location: TextLocation`  property, which holds the line range matched. So, making this query would only yield data one already has. Am I missing something here?

> Please note that, even if the id was not needed, the Match Type is also an important information that should be accessible.

Could you tell why exactly it is "important information that should be accessible" ?

nnobelis: > Would it make sense to create a separate resarch task, to 1. figure out what properties of the identifier would be ideal for a curation mechanism and 2. to talk to the owners of the snippet scanner whether such thing can be exposed, hinting that is a necessary precondition for enabling an efficient workflow with their scanner. ?

By owner of the scanner you mean code owner in ORT ? Well correct me if I am wrong but I am both code owner of FossID and ScanOSS scanners 😄 (shared with @sschuberth).
We could do some research task but as I wrote you, my current concern is not the snippet curation but the full snippet reporting.

> Looking at 'model/src/main/kotlin/utils/Snippet.kt', there is the location: TextLocation property, which holds the line range matched. So, making this query would only yield data one already has. Am I missing something here?

Yes this is the location of the matched line. Unfortunately for FossID it is always -1:
See https://github.com/oss-review-toolkit/ort/blob/14ccf85f36e446740aa8154da5c52696a7cad08b/scanner/src/main/kotlin/scanners/fossid/FossId.kt#L853 and underneath.
The `FossIdRestService.listMatchedLines` is insanely expensive (we opened a FossID support ticket for it) so it is currently no called in the scanner. This is the reason we plan on calling it called later, in reporting.
But this is only one of the additional data: we need to collect other information not present in the current ORT model, i.e. the file matched in the snippet artifact, the release date of the snippet artifact, the match type and presents all of them in the report. 

> Could you tell why exactly it is "important information that should be accessible"

The match type describe the nature of the snippet match `PARTIAL` or `FULL`.  If `PARTIAL`, it means the matched lines needs to be fetched with `FossIdRestService.listMatchedLines`.
If the match type is not exposed, how would you represent in the current data model that a snippet match is a full match ?

fviernau: > By owner of the scanner you mean code owner in ORT ? 

I meant FOSSID team.

> The FossIdRestService.listMatchedLines is insanely expensive (we opened a FossID support ticket for it) so it is currently no called in the scanner.

Would it make sense to open up a ticket for this in ORT, and there link the FossID support ticket to have the information available for everybody in ORT?

> we need to collect other information not present in the current ORT model, i.e. the file matched in the snippet artifact, the release date of the snippet artifact, the match type and presents all of them in the report.

Does "we" refer to Bosch, or to ORT. E.g. will this information be used by a reporter implemented in ORT?

If it is for ORT, shall we create a ticket which describes the capabilities which are planned to be implemented in such reporter, so that it can be reviewed against it? 

> If the match type is not exposed, how would you represent in the current data model that a snippet match is a full match ?

Depends on the use case which I believe is not outlined yet, is it?

nnobelis: > Would it make sense to open up a ticket for this in ORT, and link the FossID support ticket to it to have the information available for everybody in ORT?

Unfortunately, FossID support is private and even I don't have access to the ticket: The Bosch team in charge of our instances opened the ticket for me.
However I can nevertheless duplicate the ticket I opened at Bosch in ORT if needed. But it won't be linked. What do you think ?

> Does "we" refer to Bosch, or to ORT. E.g. will this information be used by a reporter implemented in ORT?
If it is for ORT, shall we create a ticket which describes the capabilities which are planned to be implemented in such reporter, so that it can be reviewed against it?

Bosch needs this report and wants it implemented as an ORT reporter in ORT upstream. I can open an issue in ORT for it if you feels it would helps this PR.
I have already a prototype of such report result I could attach.

fviernau: > However I can nevertheless duplicate the ticket I opened at Bosch in ORT if needed. But it won't be linked. What do you think ?

I believe it'd make sense to have a ticket which is about "working with FossID to speed-up querying file locations", so that we can include the file locations in the `ScanSummary` already (instead of making the reporter query these slowly). If this was assigned to you or someone who is working on Bosch side, then at least the point of contanct would be clarified.

> Bosch needs this report and wants it implemented as an ORT reporter in ORT upstream.
I can open an issue in ORT for it if you feels it would helps this PR.

I believe adding a new report for upstream ORT it would be value-able to outline the use cases / requirements and the plan for implementing it, in order to enable the ORT community to provide some feedback. If this is not shared, it's hard to make consideration regarding it in the review. I believe overall it could be a time saver also for you, and increase utilization of community input.

nnobelis: Thanks for your feedback.
As suggested, I created https://github.com/oss-review-toolkit/ort/issues/7028 and https://github.com/oss-review-toolkit/ort/issues/7030.

fviernau: Thanks @nnobelis !

fviernau: My potential concerns I tried research above all are about future use of the custom fields from within ORT.
This PR does not introduce any use from ORT code, so I do not have any concerns about adding these custom fields.
So, given that the PR LGTM.

sschuberth, line 61: It just occurred to me that we already have `additionalData` in `ScanResult`. So do we really need this at this place?

sschuberth (changes requested):

model/src/main/kotlin/utils/Snippet.kt
```diff
@@ -49,5 +51,12 @@ data class Snippet(
     /**
      * The license of the component the code snippet is commit from.
      */
-    val licenses: SpdxExpression
+    val licenses: SpdxExpression,
+
+    /**
+     * A map for scanner specific snippet data that cannot be mapped into any generalized property, but still needs to
+     * be made available in the scan Summary.
```
sschuberth, line 58: This should say "scan summary" or "[ScanSummary]".

sschuberth (changes requested):
Not strictly related to this PR, but it also just occurred to me that the `SnippetFinding` class is locate in the `utils` package. Unless there's a good reason for that, I'd prefer it to go along the `LicenseFinding` and `CopyrightFinding` classes in the root model package. Could you mind adding a *preceding* commit to this PR that addressed that?

model/src/main/kotlin/SnippetFinding.kt
```diff
@@ -17,9 +17,9 @@
  * License-Filename: LICENSE
  */
 
-package org.ossreviewtoolkit.model.utils
+package org.ossreviewtoolkit.model
 
-import org.ossreviewtoolkit.model.TextLocation
+import org.ossreviewtoolkit.model.utils.Snippet
```
sschuberth, line 22: Argh, sorry, IMO `Snippet` should be moved as well, as it's not a utility to work on model classes, but a model class itself.

sschuberth (dismissed):

model/src/main/kotlin/Snippet.kt
```diff
@@ -17,10 +17,8 @@
  * License-Filename: LICENSE
  */
 
-package org.ossreviewtoolkit.model.utils
+package org.ossreviewtoolkit.model
```
mnonnenmacher, line 20: Commit message nit: The breaking api change should be indicated by an exclamation mark: "fix(model)!: ..."

sschuberth (approved):

fviernau (approved):

PR #7031, opened 2023-05-22
Stop `Scope` to implement `Comparable`

sschuberth:

### DIFF
--- a/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
+++ b/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
@@ -24,7 +24,6 @@ import io.kotest.core.spec.style.WordSpec
 import io.kotest.inspectors.forAll
 import io.kotest.matchers.collections.beEmpty
 import io.kotest.matchers.collections.containExactlyInAnyOrder
-import io.kotest.matchers.collections.shouldContainExactly
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.collections.shouldNotContain
 import io.kotest.matchers.maps.containExactly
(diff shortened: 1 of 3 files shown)


analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
```diff
@@ -367,7 +366,7 @@ class AnalyzerResultBuilderTest : WordSpec() {
 
                 analyzerResult.withResolvedScopes().apply {
                     projects.find { it.id == project.id } shouldNotBeNull {
-                        scopes shouldContainExactly sortedSetOf(
+                        scopes shouldBe setOf(
```
mnonnenmacher, line 369: Is there a reason you don't use `should containExactlyInAnyOrder`?

mnonnenmacher (changes requested):

analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
sschuberth: > Is there a reason you don't use `should containExactlyInAnyOrder`?

I felt that the "InAnyOrder" part is superfluous for sets, but actually does doesn't even work! That's probably because *another* bug has been shadowed so far: `PackageReference` still implements `Comparable`, which means `Set` equality (as used by `shouldBe`) relies on `compareTo() == 0`, but that only compares a `PackageReference`'s `id`!

Chaning the code to `should containExactlyInAnyOrder` fail with a mismatch in the `linkage` (expected is `DYNAMIC` but actual is `PROJECT_DYNAMIC`, and also `issues` are non-empty), so I guess the test needs further fixups.

mnonnenmacher: Sounds like the expected value is just wrong. Do you plan to add a fix to this PR or do you want to merge independent of the issue?

sschuberth: > Sounds like the expected value is just wrong.

So you mean the expected value should just be `PROJECT_DYNAMIC` for the indirect dependencies?

> Do you plan to add a fix to this PR or do you want to merge independent of the issue?

As this PR trivially conflicts with @fviernau's anyway, I can rework this PR after his one is merged.

fviernau (approved):

mnonnenmacher (approved):

PR #7033, opened 2023-05-23
feat(gradle-inspector): Replace deprecated JVM arguments for projects

sschuberth:
Some projects like [1] use JVM arguments that are deprecated. For example, `MaxPermSize` has been deprecated since Java 1.8, but the project still uses it despite targeting Java 11 [2]. As the Gradle Tooling API would fail to launch in such cases, replace `MaxPermSize` with its successor `MaxMetaspaceSize`.

Note that while by default the Gradle Tooling API "will use the JVM and JVM arguments that the target build is configured to use" [3], both `setJvmArguments()` and `addJvmArguments()` override that default [4]. So in order to modify the JVM arguments configured for a project, they need to be read and changed manually.

[1]: https://github.com/vector-im/element-android/blob/591b08f/gradle.properties#L11
[2]: https://github.com/vector-im/element-android/blob/591b08f/dependencies.gradle#L5-L6
[3]: https://docs.gradle.org/current/javadoc/org/gradle/tooling/GradleConnector.html
[4]: https://github.com/gradle/gradle/issues/25155

### DIFF
--- a/plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
+++ b/plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
@@ -24,6 +24,7 @@ import OrtDependencyTreeModel
 
 import java.io.ByteArrayOutputStream
 import java.io.File
+import java.util.Properties
 import java.util.concurrent.TimeUnit
 
 import org.apache.logging.log4j.kotlin.Logging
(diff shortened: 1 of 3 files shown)


plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -24,6 +24,7 @@ import OrtDependencyTreeModel
 
 import java.io.ByteArrayOutputStream
 import java.io.File
+import java.util.Properties
```
mnonnenmacher, line 27: Commit message nits:
* "**h**as been deprecated"
* "ch**an**ged"

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #7035, opened 2023-05-23
Small curation provider related improvements

sschuberth:

### DIFF
--- a/cli/src/main/kotlin/commands/AnalyzerCommand.kt
+++ b/cli/src/main/kotlin/commands/AnalyzerCommand.kt
@@ -222,7 +222,10 @@ class AnalyzerCommand : OrtCommand(
         )
 
         val curationCount = packages.sumOf { it.curations.size }
-        println("Applied $curationCount curation(s) from ${enabledCurationProviders.size} provider(s).")
+        val providerCount = ortResult.resolvedConfiguration.packageCurations.count { it.curations.isNotEmpty() }
+        println(
+            "Applied $curationCount curation(s) from $providerCount of ${enabledCurationProviders.size} provider(s)."
+        )
 
         val resolutionProvider = DefaultResolutionProvider.create(ortResult, resolutionsFile)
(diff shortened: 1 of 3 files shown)


cli/src/main/kotlin/commands/AnalyzerCommand.kt
```diff
@@ -222,7 +222,8 @@ class AnalyzerCommand : OrtCommand(
         )
 
         val curationCount = packages.sumOf { it.curations.size }
-        println("Applied $curationCount curation(s) from ${enabledCurationProviders.size} provider(s).")
+        val providerCount = ortResult.resolvedConfiguration.packageCurations.count { it.curations.isNotEmpty() }
+        println("Applied $curationCount curation(s) from $providerCount provider(s).")
```
mnonnenmacher, line 226: I'm not sure about this change, I think it could be confusing if you configure e.g. 3 providers and the log says "Applied x curation(s) from 1 provider(s)." Or does it print the used providers somewhere next to it?

sschuberth: > Or does it print the used providers somewhere next to it?

It does print them above in line 171. Having the difference about the number of enabled providers vs. how many of them actually contributed curations is exactly the stats I'm interested in.

mnonnenmacher: Then how about printing the number of curations for each provider here, wouldn't that be even better?

sschuberth: It would add more detail, yes, but I didn't want to print full-blown stats, but only a short summary that always fits into a single line.

mnonnenmacher: Ok, could you change it to print "from x of y (or x/y) provider(s)" then? This would solve the issue for me.

mnonnenmacher (approved):

PR #7039, opened 2023-05-24
Some Jackson vs kotlinx-serialization stuff

sschuberth:

### DIFF
--- a/plugins/package-curation-providers/clearly-defined/src/main/kotlin/ClearlyDefinedPackageCurationProvider.kt
+++ b/plugins/package-curation-providers/clearly-defined/src/main/kotlin/ClearlyDefinedPackageCurationProvider.kt
@@ -19,8 +19,6 @@
 
 package org.ossreviewtoolkit.plugins.packagecurationproviders.clearlydefined
 
-import com.fasterxml.jackson.databind.JsonMappingException
-
 import java.net.HttpURLConnection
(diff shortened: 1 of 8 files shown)


plugins/package-curation-providers/clearly-defined/src/main/kotlin/ClearlyDefinedPackageCurationProvider.kt
```diff
@@ -127,11 +125,6 @@ class ClearlyDefinedPackageCurationProvider(
                     }
                 }
 
-                is JsonMappingException -> {
```
mnonnenmacher, line 130: Should we catch https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serialization-exception/ instead?

sschuberth: I didn't see a reason to do that as this branch basically does the same as the `else` branch below.

mnonnenmacher: The error message was a bit more specific, but ok.

mnonnenmacher (approved):

Issue #5950, opened 2022-10-14
OrtResult: Store scan results "by provenance"


fviernau:
The ORT file currently holds the scan results per `Identifier`.
This exhibits redundancy for e.g. the `LicenseFindings` and `CopyrightFindings` in the following cases:

1. Multiple packages originating from the same code repository. 
2. The nested provenance of several packages overlap, like e.g. the code repository has `google-test` as Git submodule.

The changes to the data structure should have in mind the planned integration of a full file listing.
So, adhere to the initial proposal outline here https://github.com/oss-review-toolkit/ort/issues/5681#issuecomment-1266644671.

hanna-modica:
Hi @fviernau,

I kindly want to ask, if there is a timeline for solving this issue, since this is causing a bug for us, where all findings in FossID are multiplied with a amount of detected projects leading to a wrong total number of pending findings.

Best regards,
Hanna

dobringochev:
Hi @fviernau ,
Just a gentle reminder that we really need to have your feedback when this issue is planned to be resolved?
If you want to make ORT more accurate and popular we definitely need  so any timeline for solving this issue will be highly appreciated from us.

Best Regards,
Dobrin

fviernau:
fyi: @dobringochev I'm starting to work on the epic #6945 which covers this ticket as well.

PR #7049, opened 2023-05-26
OrtResult: Remove several serialized `has_issues` properties

fviernau:
See individual commits.

Besides removing redundancy, this prepares for introducing logic inside `OrtResult` on which `hasIssues()` would need to depend on, which is not possible, because the `*Run` classes do not have access to the enclosing `OrtResult` instance.

Context: #5950.

### DIFF
--- a/model/src/main/kotlin/AnalyzerResult.kt
+++ b/model/src/main/kotlin/AnalyzerResult.kt
@@ -102,11 +102,6 @@ data class AnalyzerResult(
         }
 
     /**
-     * True if there were any issues during the analysis, false otherwise.
-     */
-    val hasIssues by lazy { getAllIssues().isNotEmpty() }
-
-    /**
      * Return a result, in which all contained [Project]s have their scope information resolved. If this result
      * has shared dependency graphs, the projects referring to one of these graphs are replaced by corresponding
      * instances that store their dependencies in the classic [Scope]-based format. Otherwise, this instance is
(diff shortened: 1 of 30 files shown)


sschuberth:
Funny, the first commit is also part of my https://github.com/oss-review-toolkit/ort/pull/7047. Mind dropping it?

model/src/main/kotlin/AnalyzerResult.kt
```diff
@@ -101,11 +101,6 @@ data class AnalyzerResult(
             }
         }
 
-    /**
-     * True if there were any issues during the analysis, false otherwise.
-     */
-    val hasIssues by lazy { getAllIssues().isNotEmpty() }
```
sschuberth, line 107: Commit message: Could you refer to @mnonnenmacher's hint that this was originally introduced / serialized for more convenient parsing of the ORT result file in a Jenkins job, but that this is not needed anymore due to the proper exit code handling that we have by now?

Also from the wording, I probably wouldn't say that the property "holds redundant information", but that the convenience it offered is not needed anymore.

sschuberth: The same can go via copy & paste to the other commit messages then.

sschuberth (changes requested):

model/src/main/kotlin/AnalyzerResult.kt
fviernau: I found it with `grep` because it was the last file left with `has_issues` in it.

sschuberth (dismissed):

sschuberth (approved):

sschuberth (approved):

PR #7052, opened 2023-05-26
fix(EvaluatedModelMapper): Fix an issue related to object pooling

### LINKED ISSUE #7042: NoSuchElementException during EvaluatedModelReporter
During the creation of the EvaluatedModelReporter / WebAppReporter, the ORT reporter crashes with this exception:

```
java.util.NoSuchElementException: Key Identifier(type=Unknown, namespace=<unknown>, name=org.springframework.cloud_spring-cloud-sleuth-zipkin, version=) is missing in the map.
	at kotlin.collections.MapsKt__MapsKt.getValue(Maps.kt:349)
	at org.ossreviewtoolkit.plugins.reporters.evaluatedmodel.EvaluatedModelMapper.addShortestPaths(EvaluatedModelMapper.kt:694)
	at org.ossreviewtoolkit.plugins.reporters.evaluatedmodel.EvaluatedModelMapper.build(EvaluatedModelMapper.kt:126)
	at org.ossreviewtoolkit.plugins.reporters.evaluatedmodel.EvaluatedModel$Companion.create(EvaluatedModel.kt:151)
	at org.ossreviewtoolkit.plugins.reporters.evaluatedmodel.EvaluatedModelReporter.generateReport(EvaluatedModelReporter.kt:50)
	at org.ossreviewtoolkit.cli.commands.ReporterCommand$run$reportDurationMap$1$1$1$1.invokeSuspend(ReporterCommand.kt:288)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:584)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:793)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:697)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:684)
```

Unfortunately, I cannot provide an example project to reproduce this. But with an evaluator-result.yml from a customer, I can reproduce this behavior.

Git bisect shows this commit as the cause:
https://github.com/oss-review-toolkit/ort/commit/8e1a9cee3a97c38be83303b02fa2457a240038fc

Reverting this commit fixes the issue.

fviernau:
The dependency graph API offers functions for iterating over the dependencies of nodes or project scopes. They return sequences of `DependencyNode`s with the caveat that the consumption of a `DependencyNode` of the sequence has finished, it's data is no more guaranteed to be valid. In fact, the instance is re-used to represent the next item of the sequence.

So, the `sort()` calls introduced by [1] made the iteration iterate N times over the "same" node which broke the dependency tree. This broke the tree view in the WebApp, see the "Tree" tab at the very top.

Fix the problem by just calling `getStableReference()`. This basically disables the object pooling and leaves a bit more work to the garbage collector.

This is a fix-up for [1] and fixes one aspect of #7042.

[1] 8e1a9cee3a97c38be83303b02fa2457a240038fc

Fixes #7042.

### DIFF
--- a/plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
+++ b/plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
@@ -492,15 +492,18 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
             visitedNodes += getInternalId() to createDependencyNode(dependency, linkage, issues)
 
             val children = visitDependencies { dependencies ->
-                dependencies.sortedBy { it.id }.map { node ->
-                    val nodeId = node.getInternalId()
+                dependencies
+                    .map { it.getStableReference() } // Obtain stable reference to not lose node data when sorting.
+                    .sortedBy { it.id }
+                    .map { node ->
+                        val nodeId = node.getInternalId()
 
-                    if (deduplicateDependencyTree && nodeId in visitedNodes) {
-                        // Cut the duplicate subtree here and only return the node without its children.
-                        visitedNodes.getValue(nodeId).copy(children = emptyList())
-                    } else {
-                        node.toEvaluatedTreeNode(scope, path + dependency)
-                    }
+                        if (deduplicateDependencyTree && nodeId in visitedNodes) {
+                            // Cut the duplicate subtree here and only return the node without its children.
+                            visitedNodes.getValue(nodeId).copy(children = emptyList())
+                        } else {
+                            node.toEvaluatedTreeNode(scope, path + dependency)
+                        }
                 }.toList()
             }
 
@@ -511,7 +514,9 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
             // Deduplication should not happen across scopes.
             visitedNodes.clear()
 
-            val subTrees = input.ortResult.dependencyNavigator.directDependencies(project, scope).sortedBy { it.id }
+            val subTrees = input.ortResult.dependencyNavigator.directDependencies(project, scope)
+                .map { it.getStableReference() } // Obtain stable reference to not lose node data when sorting.
+                .sortedBy { it.id }
                 .mapTo(mutableListOf()) { it.toEvaluatedTreeNode(scopes.getValue(scope), mutableListOf()) }
 
             val applicableScopeExcludes = input.ortResult.getExcludes().findScopeExcludes(scope)


fviernau:
@MarcelBochtler  This PR fixes the broken dependency tree view in the WebApp. As I do not have the ORT file to reproduce the `NoSuchElementException` outlined in #7042, could you please test whether this fixes also fixes that exception?

nnobelis:
> @MarcelBochtler This PR fixes the broken dependency tree view in the WebApp. As I do not have the ORT file to reproduce the `NoSuchElementException` outlined in #7042, could you please test whether this fixes also fixes that exception?

Marcel is not working today. But I will test the fix with our pipeline.

sschuberth:
> sequences of `DependencyNode`s whith

Typo in "whith".

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
```diff
@@ -492,15 +492,18 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
...
-                dependencies.sortedBy { it.id }.map { node ->
-                    val nodeId = node.getInternalId()
-
-                    if (deduplicateDependencyTree && nodeId in visitedNodes) {
-                        // Cut the duplicate subtree here and only return the node without its children.
-                        visitedNodes.getValue(nodeId).copy(children = emptyList())
-                    } else {
-                        node.toEvaluatedTreeNode(scope, path + dependency)
-                    }
+                dependencies
```
sschuberth, line 495: Just wondering, is the problem maybe (also) related to the `dependencies` sequence being sorted and mapped lazily? In other words, would moving the final `.toList()` to the beginning also fix the problem:

                dependencies.toList().sortedBy { it.id }.map { node ->
                    val nodeId = node.getInternalId()

                    if (deduplicateDependencyTree && nodeId in visitedNodes) {
                        // Cut the duplicate subtree here and only return the node without its children.
                        visitedNodes.getValue(nodeId).copy(children = emptyList())
                    } else {
                        node.toEvaluatedTreeNode(scope, path + dependency)
                    }
                }

fviernau: Turning this into a list first still exhibits the exact same problem, that only a single instance is used for all list elements. That instance changes it's data which in this case happens already during the `toList()` call.

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
```diff
@@ -492,15 +492,18 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
...
+                    .map { it.getStableReference() } // Obtain stable reference to not lose node data in following sort.
```
sschuberth, line 496: Nit: Maybe say "... when sorting." (If so, also below.)

sschuberth (dismissed):

sschuberth (approved):

fviernau:
> EDIT @fviernau: your fix seems to have solved the issue +1

Thanks @nnobelis for testing it.

PR #7055, opened 2023-05-26
Migrate the GitLab reporter from Jackson to kotlinx-serialization 

sschuberth:


plugins/reporters/gitlab/src/funTest/kotlin/GitLabLicenseModelReporterFunTest.kt
```diff
@@ -49,15 +49,15 @@ class GitLabLicenseModelReporterFunTest : WordSpec({
         "create the expected JSON license model containing only non-excluded packages" {
             val ortResult = createOrtResult()
 
-            val jsonLicenseModel = generateReport(ortResult = ortResult, skipExcluded = true) + "\n"
+            val jsonLicenseModel = generateReport(ortResult = ortResult, skipExcluded = true)
```
mnonnenmacher, line 52: Shouldn't we rather fix the issue that the generated reports do not end with a final newline? I wonder why that is, it does not seem to be an issue with other reporters like the evaluated model reporter.

sschuberth: Well, we're just using the JSON serializer, and that (both Jackson and KXS, apparently) writes out JSON without the final newline. IIRC we already have several expected JSON files committed without a final newlines. So I believe there is nothing to fix, actually.

mnonnenmacher: I was just looking at the YML file of the evaluated model reporter, the JSON file does indeed behave the same.

mnonnenmacher (approved):

PR #7061, opened 2023-05-30
Make the logic within `NestedProvenanceScanResult.merge()` re-usable

fviernau:
The logic contained within the `merge()` function will be needed when `OrtResult`s internal representation is changed to store the results by provenance. It is supposed to be used to `map` the internal representation back to the current one.

See individual commits.

Part of: #5950.

### DIFF
--- a/scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
+++ b/scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
@@ -76,20 +76,20 @@ data class NestedProvenanceScanResult(
         val allScanners = scanResults.values.flatMapTo(mutableSetOf()) { results -> results.map { it.scanner } }
 
         return allScanners.map { scanner ->
-            val scanResultsForScannerByProvenance = scanResults.mapValues { (_, results) ->
-                results.filter { it.scanner == scanner }
+            val scanResultsForScannerByPath = scanResults.entries.associate { (provenance, result) ->
+                getPath(provenance) to result.filter { it.scanner == scanner }
             }
 
-            val scanResultsForScanner = scanResultsForScannerByProvenance.values.flatten()
+            val scanResultsForScanner = scanResultsForScannerByPath.values.flatten()
 
             val startTime = scanResultsForScanner.minByOrNull { it.summary.startTime }?.summary?.startTime
                 ?: Instant.now()
             val endTime = scanResultsForScanner.maxByOrNull { it.summary.endTime }?.summary?.endTime ?: startTime
             val issues = scanResultsForScanner.flatMap { it.summary.issues }.distinct()
 
-            val licenseFindings = scanResultsForScannerByProvenance.mergeLicenseFindings()
-            val copyrightFindings = scanResultsForScannerByProvenance.mergeCopyrightFindings()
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
```diff
@@ -76,20 +76,20 @@ data class NestedProvenanceScanResult(
         val allScanners = scanResults.values.flatMapTo(mutableSetOf()) { results -> results.map { it.scanner } }
 
         return allScanners.map { scanner ->
-            val scanResultsForScannerByProvenance = scanResults.mapValues { (_, results) ->
-                results.filter { it.scanner == scanner }
-            }
+            val scanResultsForScannerByPath = scanResults.map { (provenance, result) ->
+                getPath(provenance) to result.filter { it.scanner == scanner }
+            }.toMap()
```
mnonnenmacher: Could use `scanResults.entries.associate` to skip the intermediate list.

scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
```diff
@@ -198,6 +168,53 @@ data class NestedProvenanceScanResult(
...
 }
 
+/**
+ * Merge the nested [ScanResult]s into one [ScanResult] per used scanner, using the provenance from the [ScanResult]
+ * associated with the empty path as provenance. This maps the given [scanResultsByPath] to the format currently used
+ * by [OrtResult]. When merging multiple [ScanSummary]s the earliest start time will be used as the new start time, and
+ * the latest end time will be used as the end time. Because the [ScanSummary] does not contain the checksums of the
+ * individual files, no package verification code can be calculated.
+ */
+private fun mergeScanResults(scanResultsByPath: Map<String, List<ScanResult>>): List<ScanResult> {
```
mnonnenmacher, line 178: I would prefer `mergeScanResultsByScanner` as name as it would explain why the function returns a list of scan results.

scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
```diff
@@ -198,6 +168,53 @@ data class NestedProvenanceScanResult(
     }
 }
 
+/**
+ * Merge the nested [ScanResult]s into one [ScanResult] per used scanner, using the provenance from the [ScanResult]
+ * associated with the empty path as provenance. This maps the given [scanResultsByPath] to the format currently used
+ * by [OrtResult]. When merging multiple [ScanSummary]s the earliest start time will be used as the new start time, and
+ * the latest end time will be used as the end time. Because the [ScanSummary] does not contain the checksums of the
```
mnonnenmacher, line 175: Based on your previous changes the comment should be adapter to clarify that "the earliest start time **for each scanner** will be used as the new start time" and similar for the end time. Same for the merge function in `NestedProvenanceScanResult`.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

fviernau:
There were two failing tests which were both unrelated (GoMod + ClearlyDefined). So, merge regardlessly.

PR #7062, opened 2023-05-30
Small build system improvements

sschuberth:


gradle/libs.versions.toml
```diff
@@ -41,7 +40,6 @@ log4jApi = "2.20.0"
 log4jApiKotlin = "1.2.0"
 logback = "1.4.7"
 maven = "3.9.2"
-# @pin this version as it needs to be aligned with "maven".
 mavenResolver = "1.9.10"
```
fviernau, line 43: Is line 44 removed intentionally? Also 126.

sschuberth: Yes, those are the annotations I've mentioned in the commit message. These are only being interpreted by the plugin being removed.

PR #7066, opened 2023-05-30
Write out checksums for SPDX binary / source packages

sschuberth:
See [1] for the background discussion.

[1]: https://github.com/oss-review-toolkit/ort/discussions/7064

### DIFF
--- a/plugins/reporters/spdx/src/main/kotlin/Extensions.kt
+++ b/plugins/reporters/spdx/src/main/kotlin/Extensions.kt
@@ -41,16 +41,16 @@ import org.ossreviewtoolkit.utils.spdx.model.SpdxPackageVerificationCode
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 import org.ossreviewtoolkit.utils.spdx.toSpdxId
 
+/**
+ * Convert an [Identifier]'s coordinates to an SPDX reference ID with the specified [infix].
+ */
+internal fun Identifier.toSpdxId(infix: String) = "${SpdxConstants.REF_PREFIX}$infix-${toCoordinates()}".toSpdxId()
+
 private fun LicenseInfoResolver.getSpdxCopyrightText(id: Identifier): String {
     val copyrightStatements = resolveLicenseInfo(id).flatMapTo(sortedSetOf()) { it.getCopyrights() }
     if (copyrightStatements.isEmpty()) return SpdxConstants.NONE
     return copyrightStatements.joinToString("\n")
 }
-
-/**
- * Convert an [Identifier]'s coordinates to an SPDX reference ID with the specified [infix].
- */
-internal fun Identifier.toSpdxId(infix: String) = "${SpdxConstants.REF_PREFIX}$infix-${toCoordinates()}".toSpdxId()
 
 private fun Package.toSpdxExternalReferences(): List<SpdxExternalReference> {
     val externalRefs = mutableListOf<SpdxExternalReference>()
@@ -117,8 +117,6 @@ private fun ProcessedDeclaredLicense.toSpdxDeclaredLicense(): String =
         else -> spdxExpression.nullOrBlankToSpdxNoassertionOrNone()
     }
 
-internal fun String?.nullOrBlankToSpdxNone(): String = if (isNullOrBlank()) SpdxConstants.NONE else this
-
 internal fun ScanResult.toSpdxPackageVerificationCode(): SpdxPackageVerificationCode =
     SpdxPackageVerificationCode(
         packageVerificationCodeExcludedFiles = emptyList(),
@@ -151,6 +149,8 @@ private fun SpdxExpression?.nullOrBlankToSpdxNoassertionOrNone(): String =
         toString().isBlank() -> SpdxConstants.NONE
         else -> toString()
     }
+
+internal fun String?.nullOrBlankToSpdxNone(): String = if (isNullOrBlank()) SpdxConstants.NONE else this
 
 internal fun VcsInfo.toSpdxDownloadLocation(resolvedRevision: String?): String {
     val vcsTool = when (type) {
--- a/plugins/reporters/spdx/src/main/kotlin/Extensions.kt
+++ b/plugins/reporters/spdx/src/main/kotlin/Extensions.kt
@@ -42,9 +42,15 @@ import org.ossreviewtoolkit.utils.spdx.toSpdx
 import org.ossreviewtoolkit.utils.spdx.toSpdxId
 
 /**
- * Convert an [Identifier]'s coordinates to an SPDX reference ID with the specified [infix].
+ * Convert an [Identifier]'s coordinates to an SPDX reference ID with the specified [infix] and [suffix].
  */
-internal fun Identifier.toSpdxId(infix: String) = "${SpdxConstants.REF_PREFIX}$infix-${toCoordinates()}".toSpdxId()
+internal fun Identifier.toSpdxId(infix: String = "Package", suffix: String = "") =
+    buildString {
+        append(SpdxConstants.REF_PREFIX)
+        if (infix.isNotEmpty()) append("$infix-")
+        append(toCoordinates())
+        if (suffix.isNotEmpty()) append("-$suffix")
+    }.toSpdxId()
 
 /**
  * Get the text with all Copyright statements associated with the package of the given [id], or return `NONE` if there
--- a/plugins/reporters/spdx/src/main/kotlin/Extensions.kt
+++ b/plugins/reporters/spdx/src/main/kotlin/Extensions.kt
@@ -17,8 +17,11 @@
  * License-Filename: LICENSE
  */
 
+@file:Suppress("TooManyFunctions")
+
 package org.ossreviewtoolkit.plugins.reporters.spdx
 
+import org.ossreviewtoolkit.model.Hash
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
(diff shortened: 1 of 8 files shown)


utils/spdx/src/main/kotlin/model/SpdxPackage.kt
```diff
@@ -192,6 +192,10 @@ data class SpdxPackage(
             "The SPDX ID '$spdxId' has to start with '${SpdxConstants.REF_PREFIX}'."
         }
 
+        require(spdxId.all { it.isLetterOrDigit() || it == '.' || it == '-' }) {
```
fviernau, line 195: I do recall a discussion from long time ago. It was about what kind of checks to do in which of the following ways:

1. Check failing deserialization
2. A separate verifier / linter function which highlights issues found the the document.

IIRC there was an agreement to not implement all strictness as type 1 check, but do some checks as type 2 check. I believe this check should fall under type 2. What do you think?

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -42,6 +46,17 @@ import org.ossreviewtoolkit.utils.spdx.model.SpdxPackageVerificationCode
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 import org.ossreviewtoolkit.utils.spdx.toSpdxId
 
+/**
+ * Convert an ORT [Hash] to an [SpdxChecksum], or return null if a conversion is not possible.
+ */
+private fun Hash.toSpdxChecksum() =
```
fviernau, line 52: Please specify return type explicitly, so it's obvious that it is nullable.

fviernau (changes requested):

utils/spdx/src/main/kotlin/model/SpdxPackage.kt
sschuberth: While I agree that it makes sense to have such a separation, I'd say that all syntax checks (like invalid chars) should be done as 1, and all semantic checks (like a relation to a package that does not exist) should be done as 2. As such, I'd prefer to *have* this check, esp. as we already do other syntax checks on the SPDX ID, like requiring it to start with `SPDXRef-`.

fviernau: Have you considered that each time one chooses 1, it leads to failed de-serialization, which in turn leads to an analyzer error in case of the SPDX analyzer - While it would be possible to tolerate this kind of problem and analyze without an `issue`?

sschuberth: Yes, I'm aware of that. But the same holds true for the "SPDXRef-" prefix: Theoretically we should live without it, but we still require it as the SPDX specs also do. So where do we draw the line?

IMO it's not ORT's responsibility to be lenient WRT such syntax errors. On the contrary, if ORT was lenient, probably more related tools will be lenient, too, forcing more and more tools to also become lax in this regard, ultimately resulting in many invalid SPDX files out there in the wild.

So, in a similar way like the analyzer in general relies on and implicitly enforces to follow best engineering practices by properly configuring your build system for example, we should also help the SPDX community by enforcing correct SPDX syntax here.

fviernau: > forcing more and more tools to also become lax

I believe you misunderstood what I was trying to say. My point was solely about tolerating such issues when de-serializing stuff. Not writing out SPDX file with such problems. Being able to de-serialize files with such issues does not force any tool to become "lax"

> we should also help the SPDX community by enforcing correct SPDX syntax here

That's possible in both ways, 1 and 2. While 1 is a tad less flexible.

> So, in a similar way like the analyzer in general relies on and implicitly enforces to follow best engineering practices by properly configuring your build system for example, we should also help the SPDX community by enforcing correct SPDX syntax here.

Well, when we're talking about just this ID case I'm ok. But I really fear the checks we built in which fail deserialization will become more and more strict, which in turn can waste time of compliance engineers using ORT working on issues which are not even necessary to fix and could be tolerated. Personally, I find referential integrity more important than the syntax within an ID. Anyhow, let's just stick to as it is now, but let's not take it as agreement to put in place any kind of more strict checks in the future to fail de-serialization.

sschuberth: > Not writing out SPDX file with such problems. Being able to de-serialize files with such issues does not force any tool to become "lax"

Correct, that was an error in my thought.

> But I really fear the checks we built in which fail deserialization will become more and more strict

We absolutely should have as strict checks as possible, IMO, to discover errors early in a processing pipeline.

> working on issues which are not even necessary to fix

The question remains: Where do we draw the line then? Shouldn't we then also allow SPDX IDs *not* starting with "SPDXRef-"? And simply because there IMO is no clear answer to this question, we should stick to the safe side and adhere to the specs.

> Anyhow, let's just stick to as it is now, but let's not take it as agreement to put in place any kind of more strict checks in the future to fail de-serialization.

So, does that mean you approve this PR or not?

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -0,0 +1,171 @@
+/*
```
fviernau, line 1: Now, as we have move to such a fine grained module structure - to me it makes even less sense to determine a location for a function based on whether it's an extension function or not and then as second priority the alphabetical order. 

IMO cohesion is more important: e.g. if a non-extension utility function is closely related to an extension function, then IMO it should be located nearby.

What do you think? In particular about the alternatives to

1. Keep the functions where they currently are
2. Have a utils class where when can also put non-extension functions.

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -91,18 +92,47 @@ private fun Package.toSpdxExternalReferences(): List<SpdxExternalReference> {
...
+    PROJECT("Project"),
+    BINARY_PACKAGE("Package"),
+    SOURCE_PACKAGE("Package", "source-artifact"),
+    VCS_PACKAGE("Package", "vcs")
+}
+
+/**
+ * Convert this ORT package to an SPDX package. As an ORT package can hold more metadata about its associated artifacts
+ * and origin than an SPDX package, the [type] is used to specify which kind of SPDX package should be created from
+ * respective ORT package metadata. [licenseInfoResolver] is used for license and copyright information, and the
```
fviernau, line 122: Is "the" missing before respective?

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -91,18 +92,47 @@ private fun Package.toSpdxExternalReferences(): List<SpdxExternalReference> {
...
+ * optional [scanResult] and [provenance] are used for [type]-specific information.
```
fviernau, line 123: consider stripping the mention of "the optional" as this is obvious from signature and / or doesn't matter for the behavior.

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -42,6 +46,17 @@ import org.ossreviewtoolkit.utils.spdx.model.SpdxPackageVerificationCode
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 import org.ossreviewtoolkit.utils.spdx.toSpdxId
 
+/**
+ * Convert an ORT [Hash] to an [SpdxChecksum], or return null if a conversion is not possible.
+ */
+private fun Hash.toSpdxChecksum(): SpdxChecksum? =
+    SpdxChecksum.Algorithm.values().find { it.name in algorithm.aliases }?.let {
```
fviernau, line 53: Would it make sense to add a code comment like: `// Assume that algorithm aliases in ORT and SPDX match.`?

fviernau (approved):

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
sschuberth: > to me it makes even less sense to determine a location for a function based on whether it's an extension function or not [...]  if a non-extension utility function is closely related to an extension function, then IMO it should be located nearby

I'm open to the idea to group functions, extensions or not, by purpose instead of syntax, although having an `Extensions.kt` seems to be pretty common in Kotlin projects. Anyway, I guess we should do this consistently throughout the code base in a separate PR then.

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
sschuberth: Well, the purpose of function docs is to document the function (and its signature), so this "repetition" to "document the obvious" is by design.

fviernau: > so this "repetition" to "document the obvious" is by design.

Sure, by design, not questioning that.

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
sschuberth: I'll do it in a follow-up PR.

sschuberth:
Merging despite the unrelated `GoModFunTest` failure.

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
sschuberth: See https://github.com/oss-review-toolkit/ort/pull/7085.

PR #7068, opened 2023-05-31
feat(helper-cli): Add a command to re-serialize the test assets

fviernau:
This can be helpful for development.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
@@ -1,0 +1,207 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.helper.commands.dev
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.node.ObjectNode
+
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.options.convert
+import com.github.ajalt.clikt.parameters.options.option
+import com.github.ajalt.clikt.parameters.options.required
+import com.github.ajalt.clikt.parameters.types.file
+
+import java.io.File
+
+import kotlin.reflect.KClass
+
+import org.ossreviewtoolkit.model.AnalyzerResult
+import org.ossreviewtoolkit.model.AnalyzerRun
+import org.ossreviewtoolkit.model.EvaluatorRun
+import org.ossreviewtoolkit.model.FileFormat
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.ProjectAnalyzerResult
+import org.ossreviewtoolkit.model.ScanResult
+import org.ossreviewtoolkit.model.ScannerRun
+import org.ossreviewtoolkit.utils.common.expandTilde
+
+internal class RewriteTestAssetsCommand : CliktCommand(
+    help = "Searches all test assets directories in the given ORT sources directory for recognized serialized files " +
+            "and tries to de-serialize and serialize the file. The command can be used to update the test assets " +
+            "after making changes to the corresponding model classes or serializer configuration, e.g. after " +
+            "annotating a property to not be serialized if empty."
+) {
+    private val ortSourcesDir by option(
+        "--ort-sources-dir", "-i",
+        help = "The directory containing the source code of ORT."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = false, canBeDir = true, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+        .required()
+
+    override fun run() {
+        val candidateFiles = findCandidateFiles(ortSourcesDir)
+        println("Found ${candidateFiles.size} candidate file(s).\n")
+
+        val candidateFilesByType = patchAll(candidateFiles)
+
+        candidateFilesByType.forEach { (type, files) ->
+            files.forEach { file ->
+                println("[$type] ${file.absolutePath}")
+            }
+        }
+
+        println("\nSummary\n")
+
+        candidateFilesByType.forEach { (type, files) ->
+            println("$type: ${files.size}")
+        }
+    }
+}
+
+// Fake values to replace the replace patterns with in order to make de-serialization work. The fake values must be
+// unique and no fake value must be a substring of another.
+private val REPLACE_PATTERN_REPLACEMENTS = listOf(
+    "<REPLACE_JAVA>" to "33445501",
+    "<REPLACE_OS>" to "33445502",
+    "\"<REPLACE_PROCESSORS>\"" to "33445503",
+    "\"<REPLACE_MAX_MEMORY>\"" to "33445504",
+    "<REPLACE_DEFINITION_FILE_PATH>" to "33445505",
+    "<REPLACE_ABSOLUTE_DEFINITION_FILE_PATH>" to "33445506",
+    "<REPLACE_URL>" to "http://replace/url/non/processed",
+    "<REPLACE_REVISION>" to "33445507",
+    "<REPLACE_PATH>" to "replace/path",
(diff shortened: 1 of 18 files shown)


helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
```diff
@@ -0,0 +1,209 @@
...
+        println("\nSummary\n")
+
+        candidateFilesByType.forEach { (type, files) ->
+            println("$type: ${files.size}")
+        }
+    }
+}
+
+private val REPLACE_PATTERN_REPLACEMENTS = listOf(
+    "<REPLACE_JAVA>" to "33445501",
```
sschuberth, line 79: I guess these need to be replaced to make deserialization work. But would you maybe use fake data that is more obvio9usly fake and less arbitrary? And also add a code comment that the replacement values need to be unique?

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
```diff
@@ -0,0 +1,209 @@
...
+    "<REPLACE_OS>" to "33445502",
```
sschuberth, line 80: Nit: Should this string type also use a non-numeric fake value?

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
```diff
@@ -0,0 +1,209 @@
...
+    candidateFiles.groupByTo(sortedMapOf()) { it.patch() }
+
+private fun File.patch(): String = TARGET_CLASSES.firstOrNull { patch(it) }?.simpleName ?: "Unknown"
+
+@Suppress("SwallowedException")
+private fun File.patch(kClass: KClass<out Any>): Boolean =
+    try {
+        val content = readText().removeReplacePatterns()
+        val mapper = FileFormat.forFile(this).mapper
+        val value = mapper.readValue(content, kClass.java)
```
sschuberth, line 144: You could probably just use `File.readValue<T>()` after making this function take a generic `T` (inline / reified would also be required).

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
fviernau: That's not possible, because the content of the file first needs to be patched (replace patterns) before passing it to the object mapper.

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
fviernau: I've added documentation making clear that these are fake values, and what properties the fake values need to exhibit. 

I believe the fake values are good enough to do the job, I don't see how I can make them more obvious fake and I wouldn't see the added value this brings. Can we keep the values as is?

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
fviernau: See above.

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+private val REPLACE_PATTERN_REPLACEMENTS = listOf(
+    "<REPLACE_JAVA>" to "33445501",
+    "<REPLACE_OS>" to "33445502",
+    "\"<REPLACE_PROCESSORS>\"" to "33445503",
+    "\"<REPLACE_MAX_MEMORY>\"" to "33445504",
+    "<REPLACE_DEFINITION_FILE_PATH>" to "33445505",
+    "<REPLACE_ABSOLUTE_DEFINITION_FILE_PATH>" to "33445506",
```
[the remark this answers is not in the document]
fviernau: See above.

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
sschuberth: > I don't see how I can make them more obvious fake

For the record, to me something like e.g. "00000001", "00000002" would be obviously more fake.

> and I wouldn't see the added value this brings.

The added value is that a reviewer does not have to wonder about the meaning of those magic numbers.

> Can we keep the values as is?

Yes.

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
```diff
@@ -0,0 +1,204 @@
...
+    AnalyzerResult::class,
+    AnalyzerRun::class,
+    EvaluatorRun::class,
+    OrtResult::class,
+    ScannerRun::class,
+    ScanResult::class,
+    ProjectAnalyzerResult::class
+)
+
+// Paths to nodes in the tree of JsonNodes, whose subtree shall not be changed:
```
sschuberth, line 102: Similar to your comment above, where you explain *why* values must be replaced (in order to make deserialization work), can you please also explain *why* these subtree must not be changed?

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+private fun JsonNode.getNodeAtPath(path: String): JsonNode? {
+    val keys = path.split("/").toMutableList()
+    var result: JsonNode? = this
+
+    while (keys.isNotEmpty() && result != null) {
+        val key = keys.removeFirst()
+        result = result[key] ?: null
```
sschuberth, line 180: Can drop the `?: null` part.

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
sschuberth: I mean, I really have not clue what this (and the related code) is needed for.

sschuberth (changes requested):

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
fviernau: > For the record, to me something like e.g. "00000001", "00000002" would be obviously more fake.

The above values might be a tad more likely to be contained in a test asset than then ones I provided.
Which is why I choose a prefix different from `0000000` .

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
```diff
@@ -0,0 +1,205 @@
...
+    AnalyzerRun::class,
+    EvaluatorRun::class,
+    OrtResult::class,
+    ScannerRun::class,
+    ScanResult::class,
+    ProjectAnalyzerResult::class
+)
+
+// Paths to nodes in the tree of JsonNodes, whose subtree shall not be changed. Explicitly ignoring these subtrees is
+// needed in order to avoid brining in default values property values which are not present in the original file.
```
sschuberth, line 103: Typo in "brining".

helper-cli/src/main/kotlin/commands/dev/RewriteTestAssetsCommand.kt
```diff
@@ -0,0 +1,207 @@
...
+    AnalyzerRun::class,
+    EvaluatorRun::class,
+    OrtResult::class,
+    ScannerRun::class,
+    ScanResult::class,
+    ProjectAnalyzerResult::class
+)
+
+// Paths to nodes in the tree of JsonNodes, whose subtree shall not be changed. Explicitly ignoring these subtrees is
+// needed in order to avoid bringing in default values property values which are not present in the original file.
```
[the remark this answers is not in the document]
sschuberth: > what I believe you mean is the *machine-specific values* of the config and environment classes

Yet more detailed, if machine-specific values would be used, the replacing back of dummy values to placeholders would not work, right?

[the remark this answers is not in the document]
fviernau: > Yet more detailed, if machine-specific values would be used, the replacing back of dummy values to placeholders would not work, right?

I don't understand this question. Can you elaborate a bit?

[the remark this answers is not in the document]
sschuberth: My thinking was that you first replace e.g.

    `"\"<REPLACE_MAX_MEMORY>\""` with `"33445504"`

and later again replace

    `"33445504"` with `"\"<REPLACE_MAX_MEMORY>\""`

and in between those two replacements `"33445504"` would get replaced with the real amount of memory of the machine running the command, the back-replacing would not work anymore. But giving it a second thought, I believe this cannot happen, exactly because these values are being patched, i.e. they are present in the file, and thus no default value is used for them.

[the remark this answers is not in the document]
sschuberth: > I'm not sure what you exactly refer to with "machine-specific"

I'm referring to default values, esp. in `Environment`, that vary from machine to machine, like the `maxMemory` property. I had a feeling that the real reason for the code is that you only want to not take those properties. But if you really also do not want to take properties like `enableRepositoryPackageConfigurations`, then the term "machine-specific" obviously is not applicable.

sschuberth (approved):

PR #7069, opened 2023-05-31
feat(fossid-webapp): Add the matched lines to the snippets findings 

nnobelis:

### DIFF
--- a/clients/fossid-webapp/src/main/kotlin/model/result/MatchedLines.kt
+++ b/clients/fossid-webapp/src/main/kotlin/model/result/MatchedLines.kt
@@ -19,7 +19,9 @@
 
 package org.ossreviewtoolkit.clients.fossid.model.result
 
+import org.ossreviewtoolkit.clients.fossid.PolymorphicList
+
 data class MatchedLines(
-    val localFile: Map<String, String>,
-    val mirrorFile: Map<String, String>
+    val localFile: PolymorphicList<Int>,
+    val mirrorFile: PolymorphicList<Int>,
 )
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
@@ -24,12 +24,22 @@ import org.ossreviewtoolkit.clients.fossid.model.identification.ignored.IgnoredF
 import org.ossreviewtoolkit.clients.fossid.model.identification.markedAsIdentified.MarkedAsIdentifiedFile
 import org.ossreviewtoolkit.clients.fossid.model.result.Snippet
 import org.ossreviewtoolkit.clients.fossid.model.summary.Summarizable
+import org.ossreviewtoolkit.model.ArtifactProvenance
 import org.ossreviewtoolkit.model.CopyrightFinding
+import org.ossreviewtoolkit.model.Hash
 import org.ossreviewtoolkit.model.Issue
 import org.ossreviewtoolkit.model.LicenseFinding
+import org.ossreviewtoolkit.model.PackageProvider
+import org.ossreviewtoolkit.model.RemoteArtifact
+import org.ossreviewtoolkit.model.Snippet as OrtSnippet
+import org.ossreviewtoolkit.model.SnippetFinding
 import org.ossreviewtoolkit.model.TextLocation
+import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
+import org.ossreviewtoolkit.model.utils.PurlType
 import org.ossreviewtoolkit.utils.common.collectMessages
+import org.ossreviewtoolkit.utils.spdx.SpdxConstants
+import org.ossreviewtoolkit.utils.spdx.toSpdx
 
 /**
  * A data class to hold FossID raw results.
@@ -92,3 +102,97 @@ internal fun <T : Summarizable> List<T>.mapSummary(
         copyrightFindings = copyrightFindings
     )
 }
+
+/**
+ * Map the raw snippets to ORT [SnippetFinding]s to be included in the [ScanSummary]. If a snippet license cannot be
+ * parsed, an issues is added to [issues].
+ */
+internal fun mapSnippetFindings(
+    rawResults: RawResults,
+    scannerConfig: ScannerConfiguration,
+    issues: MutableList<Issue>
+): Set<SnippetFinding> {
+    val fakeLocation = TextLocation(".", TextLocation.UNKNOWN_LINE)
+
+    return rawResults.listSnippets.flatMap { (file, rawSnippets) ->
+        val snippets = rawSnippets.map {
+            val license = it.artifactLicense?.let {
+                runCatching {
+                    LicenseFinding.createAndMap(
+                        it,
+                        fakeLocation,
+                        detectedLicenseMapping = scannerConfig.detectedLicenseMapping
+                    ).license
+                }.onFailure { spdxException ->
+                    issues += FossId.createAndLogIssue(
+                        source = "FossId",
+                        message = "Failed to parse license '$it' as an SPDX expression:" +
+                                " ${spdxException.collectMessages()}"
+                    )
+                }.getOrNull()
+            } ?: SpdxConstants.NOASSERTION.toSpdx()
+
+            // FossID does not return the hash of the remote artifact. Instead, it returns the MD5 hash of the
+            // matched file in the remote artifact as part of the "match_file_id" property.
+            val url = checkNotNull(it.url) {
+                "The URL of snippet ${it.id} must not be null."
+            }
+            val snippetProvenance = ArtifactProvenance(RemoteArtifact(url, Hash.NONE))
+            val purl = it.purl ?: "pkg:${urlToPackageType(url)}/${it.author}/${it.artifact}@${it.version}"
+
+            val additionalSnippetData = mapOf(
+                FossId.SNIPPET_DATA_ID to it.id.toString(),
+                FossId.SNIPPET_DATA_MATCH_TYPE to it.matchType.toString(),
+                FossId.SNIPPET_DATA_RELEASE_DATE to it.releaseDate.orEmpty()
+            )
+
+            // TODO: FossID doesn't return the line numbers of the match, only the character range. One must use
+            //       another call "getMatchedLine" to retrieve the matched line numbers. Unfortunately, this is a
+            //       call per snippet which is too expensive. When it is available for a batch of snippets, it can
+            //       be used here.
+            OrtSnippet(
+                it.score.toFloat(),
+                TextLocation(it.file, TextLocation.UNKNOWN_LINE),
+                snippetProvenance,
+                purl,
+                license,
+                additionalSnippetData
+            )
+        }
+
+        val sourceLocation = TextLocation(file, TextLocation.UNKNOWN_LINE)
+        snippets.map {
+            SnippetFinding(
+                sourceLocation,
+                it
+            )
+        }
+    }.toSet()
+}
+
+/**
+ * Return the [PurlType] as determined from the given [url], or [PurlType.GENERIC] if there is no match.
+ */
+private fun urlToPackageType(url: String): PurlType =
+    when (val provider = PackageProvider.get(url)) {
+        PackageProvider.COCOAPODS -> PurlType.COCOAPODS
+        PackageProvider.CRATES_IO -> PurlType.CARGO
+        PackageProvider.DEBIAN -> PurlType.DEBIAN
+        PackageProvider.GITHUB -> PurlType.GITHUB
+        PackageProvider.GITLAB -> PurlType.GITLAB
+        PackageProvider.GOLANG -> PurlType.GOLANG
+        PackageProvider.MAVEN_CENTRAL, PackageProvider.MAVEN_GOOGLE -> PurlType.MAVEN
+        PackageProvider.NPM_JS -> PurlType.NPM
+        PackageProvider.NUGET -> PurlType.NUGET
+        PackageProvider.PACKAGIST -> PurlType.COMPOSER
+        PackageProvider.PYPI -> PurlType.PYPI
+        PackageProvider.RUBYGEMS -> PurlType.GEM
+
+        else -> {
+            PurlType.GENERIC.also {
+                FossId.logger.warn {
+                    "Cannot determine PURL type for url '$url' and provider '$provider'. Falling back to '$it'."
+                }
+            }
+        }
+    }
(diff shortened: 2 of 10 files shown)


scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -196,3 +204,37 @@ private fun urlToPackageType(url: String): PurlType =
...
     }
+
+/**
+ * FossID returns the matched lines as such:
+ *          255,256,258
+ * This function replaces the map content with:
+ *          255-256, 258
+ * and return it as [TextLocation]s.
+ */
+internal fun List<Int>.compressMatchedLines(): List<TextLocation> {
```
sschuberth, line 225: I'd prefer if we could instead slightly generalize the `CopyrightStatementsProcessor`'s `getYearRanges()` function by renaming it (to e.g. "collapseRanges"), moving it to `Utils.kt`, and them making use of it here.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -159,6 +159,16 @@ internal fun mapSnippetFindings(
...
+
+                fun List<TextLocation>?.toString(): String = this?.joinToString(", ") { location ->
+                    if (location.startLine == location.endLine) {
+                        "${location.startLine}"
+                    } else {
+                        "${location.startLine}-${location.endLine}"
+                    }
+                }.orEmpty()
+                additionalSnippetData[FossId.SNIPPET_DATA_MATCHED_LINE_SOURCE] = rawMatchedLinesSourceFile.toString()
+                additionalSnippetData[FossId.SNIPPET_DATA_MATCHED_LINE_SNIPPET] = rawMatchedLinesSnippetFile.toString()
```
sschuberth, line 171: Is it intended that these are always written even if `rawMatchedLinesSourceFile` / `rawMatchedLinesSnippetFile` is empty?

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
nnobelis: Fixed.

clients/fossid-webapp/src/main/kotlin/model/result/MatchedLines.kt
```diff
@@ -19,7 +19,9 @@
 
 package org.ossreviewtoolkit.clients.fossid.model.result
 
+import org.ossreviewtoolkit.clients.fossid.PolymorphicList
+
 data class MatchedLines(
-    val localFile: Map<String, String>,
-    val mirrorFile: Map<String, String>
+    val localFile: PolymorphicList<Int>,
```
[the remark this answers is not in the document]
nnobelis: I didn't ask them (and I would prefer not too since I don't want to harass my contact 😄) however I really believe this is the case and it's a bug / bad design / technical quirk on their side.

As seen here (https://github.com/oss-review-toolkit/ort/blob/152348f7d2d87e4aeac322b69e01906e1cb9bbd5/clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt#L114),  FossID sometimes returns some strange answers.

For instance **files_and_folders/get_fossid_results** returns:
```
"data": {
		"241": {
			"id": "241",
			"created": "2023-05-25 17:22:11",
			"scan_id": "0",
			"scan_file_id": "25436135",
```
You can notice that there is not point to return the results as a map with the entity id as a key, since the id is also present in the object: an array would be enough.
In a way, the matched line response is the same, the object returned is aggregated by its line number, but this is even weirder because the object _IS_ just a line number.

Another example to showcase such quirk. The response of **projects/get_all_scans**:

```
{
	"operation": "projects_get_all_scans",
	"status": "1",
	"data": [
		{
			"id": "8911",
			"0": "8911",
			"created": "2023-03-01 12:55:53",
			"1": "2023-03-01 12:55:53",
			"updated": "2023-03-01 12:56:04",
			"2": "2023-03-01 12:56:04",
```
Here it seems to be a badly configured serializer.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
nnobelis: Done. Please have a look how I adapted the wording of the documentation.

utils/common/src/main/kotlin/Utils.kt
```diff
@@ -128,3 +128,35 @@ fun <R> temporaryProperties(vararg properties: Pair<String, String?>, block: ()
         }
     }
 }
+
```
sschuberth, line 131: Commit message: Please use

```
refactor(utils): Generalize `prettyPrintYears()`

Extract `prettyPrintYears()` and its contained `getYearRanges()` to more
general utility functions that reflect to handle any (pairs of) integers
and not just years. These functions will be used in an upcoming change.
```

utils/common/src/main/kotlin/Utils.kt
```diff
@@ -128,3 +128,35 @@ fun <R> temporaryProperties(vararg properties: Pair<String, String?>, block: ()
...
+/**
+ * Return a comma-separated string of values or ranges that represent the [current object].
```
sschuberth, line 133: "Return a string of common-separated ranges as denoted by the list of pairs."

utils/common/src/main/kotlin/Utils.kt
```diff
@@ -128,3 +128,35 @@ fun <R> temporaryProperties(vararg properties: Pair<String, String?>, block: ()
...
+ */
+fun List<Pair<Int, Int>>.prettyPrintValues(): String {
```
sschuberth, line 135: Please call this "prettyPrintRanges".

utils/common/src/main/kotlin/Utils.kt
```diff
@@ -128,3 +128,35 @@ fun <R> temporaryProperties(vararg properties: Pair<String, String?>, block: ()
...
+    return joinToString { (startValue, endValue) ->
+        if (startValue == endValue) startValue.toString() else "$startValue-$endValue"
+    }
+}
+
+/**
+ * Collapse consecutive [values] to a list of pairs that each denote a range. A single value is represented as a
+ * range whose first and last elements are equal.
+ */
+fun collapseValues(values: Collection<Int>): List<Pair<Int, Int>> {
```
sschuberth, line 145: I'd prefer a function name that better explains *what to* values are collapsed. Maybe "createRanges", "collapseToRanges", "extractRanges", or something like that.

utils/common/src/main/kotlin/Utils.kt
sschuberth: Also, if this is an extension function now, it should actually go to `Extensions.kt`. Alternatively, make `collapseValues` also an extension function. (Which probably is the better idea.)

utils/common/src/test/kotlin/UtilsTest.kt
```diff
@@ -64,4 +66,44 @@ class UtilsTest : WordSpec({
             )
         }
     }
+
+    "collapseValues" should {
+        "not modify single value" {
```
sschuberth, line 71: "*a* single value"

utils/common/src/test/kotlin/UtilsTest.kt
```diff
@@ -64,4 +66,44 @@ class UtilsTest : WordSpec({
...
+            val lines = listOf(255)
+            collapseValues(lines).shouldContainExactly(255 to 255)
+        }
+
+        "collapse two elements in a single range" {
+            val lines = listOf(255, 256)
+            collapseValues(lines).shouldContainExactlyInAnyOrder(255 to 256)
+        }
+
+        "compress three elements in a single range" {
```
sschuberth, line 81: Stick to a single verb, i.e. use "collapse" instead of "compress".

utils/common/src/test/kotlin/UtilsTest.kt
```diff
@@ -64,4 +66,44 @@ class UtilsTest : WordSpec({
...
+            val lines = listOf(255, 256, 257)
+            collapseValues(lines).shouldContainExactlyInAnyOrder(255 to 257)
+        }
+
+        "don't include single element in a range" {
```
sschuberth, line 86: Avoid informal language: "do not a single..."

utils/common/src/test/kotlin/UtilsTest.kt
```diff
@@ -64,4 +66,44 @@ class UtilsTest : WordSpec({
...
+            val lines = listOf(255, 257, 258)
+            collapseValues(lines)
+                .shouldContainExactlyInAnyOrder(255 to 255, 257 to 258)
+        }
+
+        "compress multiple ranges" {
```
sschuberth, line 92: Stick to a single verb, i.e. use "collapse" instead of "compress".

utils/common/src/test/kotlin/UtilsTest.kt
```diff
@@ -64,4 +66,44 @@ class UtilsTest : WordSpec({
...
+            val lines = listOf(255, 256, 258, 259)
+            collapseValues(lines)
+                .shouldContainExactlyInAnyOrder(255 to 256, 258 to 259)
+        }
+
+        "compress a mix of ranges and single values" {
```
sschuberth, line 98: Stick to a single verb, i.e. use "collapse" instead of "compress".

utils/common/src/test/kotlin/UtilsTest.kt
```diff
@@ -64,4 +66,44 @@ class UtilsTest : WordSpec({
             )
         }
     }
+
+    "collapseValues" should {
+        "not modify single value" {
+            val lines = listOf(255)
+            collapseValues(lines).shouldContainExactly(255 to 255)
```
sschuberth, line 73: In general, please use the infix matchers: `should containExactly(...`

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -146,25 +150,32 @@ internal fun mapSnippetFindings(
...
+            var sourceLocation: TextLocation? = null
+            var snippetLocation: TextLocation? = null
+
+            if (it.matchType == MatchType.PARTIAL) {
+                val rawMatchedLines = rawResults.snippetMatchedLines[it.id]
+                val rawMatchedLinesSourceFile = collapseValues(rawMatchedLines?.localFile.orEmpty())
+                val rawMatchedLinesSnippetFile = collapseValues(rawMatchedLines?.mirrorFile.orEmpty())
+
+                sourceLocation = rawMatchedLinesSourceFile.firstOrNull()
+                    ?.let { pair -> TextLocation(it.file, pair.first, pair.second) }
```
sschuberth, line 163: Use a destructing declaration to give the pair members more meaningful names: `(startLine, endLine) ->`. Same below.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -162,6 +163,15 @@ internal fun mapSnippetFindings(
                     ?.let { pair -> TextLocation(it.file, pair.first, pair.second) }
                 snippetLocation = rawMatchedLinesSnippetFile.firstOrNull()
                     ?.let { pair -> TextLocation(file, pair.first, pair.second) }
+
+                if (rawMatchedLinesSourceFile.isNotEmpty()) {
+                    additionalSnippetData[FossId.SNIPPET_DATA_MATCHED_LINE_SOURCE] =
+                        rawMatchedLinesSourceFile.prettyPrintValues()
+                }
```
sschuberth, line 170: Please add an empty line after the closing block in this line.

sschuberth (changes requested):

utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -438,3 +438,35 @@ fun URI.getQueryParameters(): Map<String, List<String>> {
         .groupBy({ it.substringBefore('=') }, { it.substringAfter('=').split(',') })
         .mapValues { (_, v) -> v.flatten() }
 }
+
+/**
+ * Return a string of common-separated ranges as denoted by the list of pairs.
+ */
+fun List<Pair<Int, Int>>.prettyPrintRanges(): String {
```
sschuberth, line 445: The functions in this file should be sorted alphabetically, i.e. by the name of the class they extend. To avoid these functions being separated due to sorting, I'd propose to make this extend a `Collection` instead of a `List` as well (even if not meaningful for our current use-case).

utils/common/src/test/kotlin/ExtensionsTest.kt
```diff
@@ -61,7 +61,7 @@ class ExtensionsTest : WordSpec({
         "return all duplicates" {
             val strings = listOf("foo", "bar", "baz", "foo", "bar", "bar")
 
-            strings.getDuplicates().shouldContainExactlyInAnyOrder("foo", "bar")
+            strings.getDuplicates()should containExactlyCollection("foo", "bar")
```
sschuberth, line 64: Unintended change? In any case, there's a space missing before `should`.

sschuberth (changes requested):

scanner/src/test/kotlin/scanners/fossid/FossIdTest.kt
```diff
@@ -1322,7 +1322,6 @@ private fun createSnippetFindings(index: Int): Set<SnippetFinding> = (1..5).map
             SpdxExpression.Companion.parse("MIT"),
             mapOf(
                 FossId.SNIPPET_DATA_ID to "$snippetIndex",
-                FossId.SNIPPET_DATA_FILE to "file$snippetIndex",
```
sschuberth, line 1325: This should be squashed to the previous commit (no idea why it got separated).

sschuberth (changes requested):

sschuberth (dismissed):

nnobelis:
The GoMod test error is IMO unrelated.

utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -49,6 +49,38 @@ fun <T, K> Collection<T>.getDuplicates(keySelector: (T) -> K): Map<K, List<T>> =
...
+
+    ranges += current to sortedValues.last()
+
+    return ranges
+}
+
+/**
+ * Return a string of common-separated ranges as denoted by the list of pairs.
+ */
+fun Collection<Pair<Int, Int>>.prettyPrintRanges(): String {
```
fviernau, line 78: This could be an expression function now, as the inner function has been extracted.

fviernau (dismissed):

sschuberth (approved):

sschuberth:
> The GoMod test error is IMO unrelated.

Agreed, will merge despite that failure.

PR #7077, opened 2023-06-01
ScannerRun: Store scan results by provenance

fviernau:
See individual commits.

**This is still work in progress.**


fviernau:
Replaced by the non-recursive / simpler re-implementation: https://github.com/oss-review-toolkit/ort/pull/7099

PR #7078, opened 2023-06-01
ScanController: Stop mixing provenance resolution issues of different type

fviernau:
See individual commits.

Part of #5950.

### DIFF
--- a/scanner/src/main/kotlin/ScanController.kt
+++ b/scanner/src/main/kotlin/ScanController.kt
@@ -58,7 +58,13 @@ class ScanController(
      * A map of package [Identifier]s to a list of [Issue]s that occurred during provenance resolution for the
      * respective package.
      */
-    private val provenanceResolutionIssues = mutableMapOf<Identifier, MutableList<Issue>>()
+    private val packageProvenanceResolutionIssues = mutableMapOf<Identifier, Issue>()
+
+    /**
+     * A map of package [KnownProvenance]s to a list of [Issue]s that occurred during provenance resolution for the
+     * respective package.
+     */
+    private val nestedProvenanceResolutionIssues = mutableMapOf<KnownProvenance, Issue>()
 
     /**
      * A map of [Identifier]s associated with a list of [Issue]s that occurred during a scan besides the issues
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/ScanController.kt
```diff
@@ -88,8 +95,20 @@ class ScanController(
...
+     */
+    fun putPackageProvenanceResolutionIssue(id: Identifier, issue: Issue) {
+        packageProvenanceResolutionIssues[id] = issue
+    }
+
+    /**
+     * Set the [issue] which failed nested provenance resolution for the given [provenance].
+     */
+    fun putNestedProvenanceResolutionIssue(provenance: KnownProvenance, issue: Issue) {
+        provenance.requireEmptyVcsPath()
```
mnonnenmacher, line 108: Is this check really required here? It seems a bit over the top to me, otherwise several other function should have to do similar checks. For example, should `putPackageProvenanceResolutionIssue` and `addIssue` should verify that the `id` does exist?
But if you think it is required, it would probably make more sense to verify that `provenance` is in the values of `packageProvenancesWithoutVcsPath`.

scanner/src/main/kotlin/ScanController.kt
```diff
@@ -179,6 +178,15 @@ class ScanController(
     private fun getMissingProvenanceScans(scanner: ScannerWrapper, nestedProvenance: NestedProvenance) =
         nestedProvenance.getProvenances().filter { hasScanResult(scanner, it) }
 
+    /**
+     * Return the nested provenance resolution issues associated with the given [provenance].
+     */
+    fun getNestedProvenanceResolutionIssue(provenance: KnownProvenance): Issue? {
+        provenance.requireEmptyVcsPath()
```
mnonnenmacher, line 185: Similar to the comment in the previous commit, is this check really required?

scanner/src/main/kotlin/ScanController.kt
```diff
@@ -58,7 +58,13 @@ class ScanController(
      * A map of package [Identifier]s to a list of [Issue]s that occurred during provenance resolution for the
      * respective package.
      */
-    private val provenanceResolutionIssues = mutableMapOf<Identifier, MutableList<Issue>>()
+    private val packageProvenanceResolutionIssues = mutableMapOf<Identifier, Issue>()
```
sschuberth, line 61: Commit message: s/benifits/benefits/

mnonnenmacher (approved):

PR #7079, opened 2023-06-01
deps(Pub): Upgrade Flutter to version 3.10.5

MarcelBochtler:
Update the minimum SDK version in the synthetic tests to 2.12.0, as it
is now required for null safety compatibility [1].
Additionally, adjust the upper boundary in the synthetic tests to align
with Dart 3.0.3, which is used in Flutter 3.10.x [2].

For more information, refer to the full changelog [3].

[1]: https://dart.dev/null-safety#dart-3-and-null-safety
[2]: https://docs.flutter.dev/release/archive
[3]: https://docs.flutter.dev/release/release-notes/release-notes-3.10.0


MarcelBochtler:
This unfortunately got bigger than I expected, because with Flutter 3.10.x the Dart SDK was updated to 3.x.x.
This requires to update multiple `pubspec.yaml` files to update the minimal and maximum required SDK version. Consequently, this requires to update more expected results than usual.

Additionally, the external project `dart-http` had to be updated to also use the updated SDK version ranges.

PR #7083, opened 2023-06-02
fix(fossid-webapp): Fix FossID license mapping 

nnobelis:
Currently, the mapping of licenses from FossID identified / "marked as
identified" files and from FossID's snippets fails if the license cannot
be mapped to SPDX, for instance 'Apache 2.0'.
In this case, the FossID scanner creates issues in the scan results.

This commit changes the logic to use the `DeclaredLicenseProcessor` to
map these licenses to syntactically correct SPDX expressions. It also
adds a unit test to cover this part of the code that was previously
uncovered.

Please note that mapping of non-mappable license from identified /
"marked as identified" files still leads to the creation of an issue.
However, fewer issues should be now created with the new logic.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
@@ -36,7 +36,6 @@ import org.ossreviewtoolkit.model.RemoteArtifact
 import org.ossreviewtoolkit.model.Snippet as OrtSnippet
 import org.ossreviewtoolkit.model.SnippetFinding
 import org.ossreviewtoolkit.model.TextLocation
-import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.utils.PurlType
 import org.ossreviewtoolkit.utils.common.collapseToRanges
@@ -114,20 +113,13 @@ internal fun <T : Summarizable> List<T>.mapSummary(
  */
 internal fun mapSnippetFindings(
     rawResults: RawResults,
-    scannerConfig: ScannerConfiguration,
     issues: MutableList<Issue>
 ): Set<SnippetFinding> {
-    val fakeLocation = TextLocation(".", TextLocation.UNKNOWN_LINE)
-
     return rawResults.listSnippets.flatMap { (file, rawSnippets) ->
         rawSnippets.map { snippet ->
             val license = snippet.artifactLicense?.let {
                 runCatching {
-                    LicenseFinding.createAndMap(
-                        it,
-                        fakeLocation,
-                        detectedLicenseMapping = scannerConfig.detectedLicenseMapping
-                    ).license
+                    it.toSpdx()
                 }.onFailure { spdxException ->
                     issues += FossId.createAndLogIssue(
                         source = "FossId",
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
@@ -41,6 +41,7 @@ import org.ossreviewtoolkit.model.utils.PurlType
 import org.ossreviewtoolkit.utils.common.collapseToRanges
 import org.ossreviewtoolkit.utils.common.collectMessages
 import org.ossreviewtoolkit.utils.common.prettyPrintRanges
+import org.ossreviewtoolkit.utils.ort.DeclaredLicenseProcessor
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 
@@ -118,15 +119,14 @@ internal fun mapSnippetFindings(
     return rawResults.listSnippets.flatMap { (file, rawSnippets) ->
         rawSnippets.map { snippet ->
             val license = snippet.artifactLicense?.let {
-                runCatching {
-                    it.toSpdx()
-                }.onFailure { spdxException ->
-                    issues += FossId.createAndLogIssue(
-                        source = "FossId",
-                        message = "Failed to parse license '$it' as an SPDX expression:" +
-                                " ${spdxException.collectMessages()}"
-                    )
-                }.getOrNull()
+                DeclaredLicenseProcessor.process(it).also { expression ->
+                    if (expression == null) {
+                        issues += FossId.createAndLogIssue(
+                            source = "FossId",
+                            message = "Failed to map license '$it' as an SPDX expression."
+                        )
+                    }
+                }
             } ?: SpdxConstants.NOASSERTION.toSpdx()
 
             // FossID does not return the hash of the remote artifact. Instead, it returns the MD5 hash of the
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -828,7 +828,7 @@ class FossId internal constructor(
             Issue(source = name, message = "Pending identification for '$it'.", severity = Severity.HINT)
         }
 
-        val snippetFindings = mapSnippetFindings(rawResults, scannerConfig, issues)
+        val snippetFindings = mapSnippetFindings(rawResults, issues)
 
         val ignoredFiles = rawResults.listIgnoredFiles.associateBy { it.path }
 
--- a/scanner/src/test/kotlin/scanners/fossid/FossIdLicenseMappingTest.kt
+++ b/scanner/src/test/kotlin/scanners/fossid/FossIdLicenseMappingTest.kt
@@ -1,0 +1,174 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.scanner.scanners.fossid
+
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.collections.beEmpty
+import io.kotest.matchers.collections.haveSize
+import io.kotest.matchers.should
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldStartWith
+
+import org.ossreviewtoolkit.clients.fossid.model.identification.common.LicenseMatchType
+import org.ossreviewtoolkit.clients.fossid.model.identification.markedAsIdentified.File
+import org.ossreviewtoolkit.clients.fossid.model.identification.markedAsIdentified.License
+import org.ossreviewtoolkit.clients.fossid.model.identification.markedAsIdentified.LicenseFile
+import org.ossreviewtoolkit.clients.fossid.model.identification.markedAsIdentified.MarkedAsIdentifiedFile
+import org.ossreviewtoolkit.clients.fossid.model.result.LicenseCategory
+import org.ossreviewtoolkit.clients.fossid.model.result.MatchType
+import org.ossreviewtoolkit.clients.fossid.model.result.Snippet
+import org.ossreviewtoolkit.model.Issue
+import org.ossreviewtoolkit.utils.spdx.SpdxConstants
+import org.ossreviewtoolkit.utils.spdx.toSpdx
+import org.ossreviewtoolkit.utils.test.shouldNotBeNull
+
+private const val FILE_PATH = "filePath"
+private const val FILE_PATH_SNIPPET = "filePathSnippet"
+
+class FossIdLicenseMappingTest : WordSpec({
+    "FossIdScanResults" should {
+        "create an issue when a license in an identified file cannot be mapped" {
+            val sampleFile = createMarkAsIdentifiedFile("invalid license")
+            val issues = mutableListOf<Issue>()
+
+            val findings = listOf(sampleFile).mapSummary(emptyMap(), issues, emptyMap())
+
+            issues should haveSize(1)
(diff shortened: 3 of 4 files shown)


sschuberth:
> This could cause issues in the Webapp report.

Instead of "issues", do you mean "rule violations for custom rules"? Because actually the web-app report should be able to display unmapped / unmappable licenses.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -83,7 +84,14 @@ internal fun <T : Summarizable> List<T>.mapSummary(
 
         summary.licences.forEach {
             runCatching {
-                LicenseFinding.createAndMap(it.identifier, location, detectedLicenseMapping = detectedLicenseMapping)
+                val declaredLicense = DeclaredLicenseProcessor.process(setOf(it.identifier))
+                declaredLicense.spdxExpression?.let { spdx ->
```
sschuberth, line 88: This ignores unmapped licenses. As these could syntax-wise still be valid license IDs (i.e. they do not contain spaces), IMO they should be considered here, too.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -123,7 +131,8 @@ internal fun mapSnippetFindings(
         rawSnippets.map {
             val license = it.artifactLicense?.let {
                 runCatching {
-                    LicenseFinding.createAndMap(
+                    val declaredLicense = DeclaredLicenseProcessor.process(setOf(it))
+                    declaredLicense.spdxExpression ?: LicenseFinding.createAndMap(
```
sschuberth, line 135: See above.

sschuberth (changes requested):

nnobelis:
> > This could cause issues in the Webapp report.
> 
> Instead of "issues", do you mean "rule violations for custom rules"? Because actually the web-app report should be able to display unmapped / unmappable licenses.

No I meant ORT issues in the scan results. I agree this is confusing therefore I would like to change this sentence with

> This could cause issues in the scan results.

What do you think ?

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
nnobelis: I don't understand your comment.
When there is an unmapped license, the `spdxExpression` is `null` therefore the `?:` branch will be called: there, the old behavior of calling `LicenseFinding.createAndMap` and potentially raising an `Issue` on failure would still apply.
Am I missing something ?

sschuberth:
> > This could cause issues in the scan results.
> 
> What do you think ?

That still does not explain what *kind* of issue you're talking about. What's the message of the issue you're trying to get rid of?

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
sschuberth: > When there is an unmapped license, the `spdxExpression` is `null`

No, that's not correct, see e.g.

https://github.com/oss-review-toolkit/ort/blob/ab1779038b17773544d6f0ecf436d1f4e9f059d1/utils/ort/src/test/kotlin/DeclaredLicenseProcessorTest.kt#L107-L115

So *both* `spdxExpression` *and* `unmapped` may contain (disjunctive) values.

Not imagine the test would not use the string "invalid" but "ALicenseNameOnlyUsedbyFossId". The current implementation would swallow "ALicenseNameOnlyUsedbyFossId", but my guess is that you'd want to keep it, although it's neither an SPDX (core) license ID nor a `LicenseRef-`.

nnobelis:
> That still does not explain what _kind_ of issue you're talking about. What's the message of the issue you're trying to get rid of?

`Failed to parse license '$it' as an SPDX expression:`

They are thrown here
https://github.com/oss-review-toolkit/ort/blob/4f36a432b44b27b1a83c031f4b8ea72f181d9107/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt#L97-L102
and here:
https://github.com/oss-review-toolkit/ort/blob/4f36a432b44b27b1a83c031f4b8ea72f181d9107/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt#L140-L145

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
nnobelis: First, please not that the `DeclaredLicenseProcessor.process` is called with only one license in both cases. 
Therefore, if `ALicenseNameOnlyUsedbyFossId` is given to `process`. It is EITHER in `spdxExpression` or in `unmapped`.
I cannot land in both.

If this is in `unmapped`, the `spdxExpression` will be `null`. Hence the `?:` branch will be called, this will then call `spdxExpression`. The latter does a `toSpdx`.
Therefore finally, the `onFailure` will be called which will create an issue.

I don't see what else can be done ?!

sschuberth:
> They are thrown here:

Ah, ok, the issues are created in the FossID scanner itself! As you were initially saying

> This could cause issues in the *Webapp* report.

(emphasis mine) I thought it's the web-app report who's running into issues due to the way the FossID scanner provides the data. But in fact these generic issues would show up in *any* report.

nnobelis:
> 
> (emphasis mine) I thought it's the web-app report who's running into issues due to the way the FossID scanner provides the data. But in fact these generic issues would show up in _any_ report.

Yes I realize the wording was pretty confusing. I changed it please review.

nnobelis:
@sschuberth  I added a second commit to fix a severe bug I just found.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -169,9 +169,9 @@ internal fun mapSnippetFindings(
                 val rawMatchedLinesSnippetFile = rawMatchedLines?.mirrorFile.orEmpty().collapseToRanges()
 
                 sourceLocation = rawMatchedLinesSourceFile.firstOrNull()
-                    ?.let { (startLine, endLine) -> TextLocation(it.file, startLine, endLine) }
-                snippetLocation = rawMatchedLinesSnippetFile.firstOrNull()
                     ?.let { (startLine, endLine) -> TextLocation(file, startLine, endLine) }
+                snippetLocation = rawMatchedLinesSnippetFile.firstOrNull()
```
sschuberth, line 173: How about giving `it` the explicit `snippet` name in line 131 for clarity in a preceding commit? That would make this fix then more obvious.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -148,28 +148,29 @@ internal fun mapSnippetFindings(
...
             // matched file in the remote artifact as part of the "match_file_id" property.
-            val url = checkNotNull(it.url) {
-                "The URL of snippet ${it.id} must not be null."
+            val url = checkNotNull(snippet.url) {
+                "The URL of snippet ${snippet.id} must not be null."
             }
             val snippetProvenance = ArtifactProvenance(RemoteArtifact(url, Hash.NONE))
-            val purl = it.purl ?: "pkg:${urlToPackageType(url)}/${it.author}/${it.artifact}@${it.version}"
+            val purl =
+                snippet.purl ?: "pkg:${urlToPackageType(url)}/${snippet.author}/${snippet.artifact}@${snippet.version}"
```
sschuberth, line 156: Nit: For next time, the wrapping should rather be like

    val purl = snippet.purl
        ?: "pkg:${urlToPackageType(url)}/${snippet.author}/${snippet.artifact}@${snippet.version}"

as that also more nicely visualizes default vs. fallback value.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
sschuberth: > I don't see what else can be done ?!

My proposal is to make the `DeclaredLicenseProcessor.process()` overload that only takes a single string public, and then simply use

    val license = snippet.artifactLicense?.let { DeclaredLicenseProcessor.process(it) }
        ?: SpdxConstants.NOASSERTION.toSpdx()

This code could also be moved to before the line `val ortSnippet = OrtSnippet(` to be closer to its use.

That addressed several issues with the proposed code:
- It removes complexity only needed for compound expressions, which cannot occur here.
- `DeclaredLicenseProcessor.process()` doe snot throw, so the `onFailure` handling is unneeded
- For the use in line 135, it avoids creating a `LicenseFinding` from which only `license` is being used.

sschuberth (changes requested):

sschuberth:
@nnobelis mind splitting out the last two less "controversal" commits to a separate PR?

nnobelis:
> - It removes complexity only needed for compound expressions, which cannot occur here.
> - DeclaredLicenseProcessor.process() doe snot throw, so the onFailure handling is unneeded
> - For the use in line 135, it avoids creating a LicenseFinding from which only license is being used.

While I agree on point 1,  `runCatching` is still needed because of the `LicenseFinding.createAndMap`. The latter is just needed because of the `detectedLicenseMapping`. It would be nice to pass it as last parameter to `process` -> `customLicenseMapping`.

Unfortunately, `customLicenseMapping` is `Map<String, SpdxExpression>` and `detectedLicenseMapping` is `Map<String, String>`. How would do it elegantly ?


> @nnobelis mind splitting out the last two less "controversal" commits to a separate PR?

https://github.com/oss-review-toolkit/ort/pull/7094

sschuberth:
> The latter is just needed because of the `detectedLicenseMapping`.

I was wondering whether it's even correct to take those mappings into account, as these aren't *detected* licenses IMO, but the licenses *declared* for the packages the snippets belong to.

So, is it really correct to call `LicenseFinding.createAndMap()` from *both* `mapSummary()` *and* `mapSnippetFindings()`, or should it maybe only be called from the former, or even from neither? Just double-checking, I'm not trying to imply an answer.

nnobelis:
> So, is it really correct to call `LicenseFinding.createAndMap()` from _both_ `mapSummary()` _and_ `mapSnippetFindings()`, or should it maybe only be called from the former, or even from neither? Just double-checking, I'm not trying to imply an answer.

These mapping were introduced by the following commit by @MarcelBochtler:  https://github.com/oss-review-toolkit/ort/commit/8a8626e1eaef45ee9f86f390b760592b09d6b0a1.
My guess if they were introduced globally and therefore in FossID `mapSummary`.

They are also present in the snippet mapping because I took inspiration of `mapSummary` to write `mapSnippetFindings`.
It was done conservatively, not because it was needed.

After having another look they are definitively not needed for snippets.
Therefore, IMO we could remove this mapping for `mapSnippetFindings`.

Maybe even in both location depending on @MarcelBochtler's opinion.

sschuberth:
> After having another look they are definitively not needed for snippets.

So, does that mean we can simplify line 124 to just

    val license = it.artifactLicense ?: SpdxConstants.NOASSERTION

and use `license.toSpdx()` when creating the `OrtSnippet`?

nnobelis:
> So, does that mean we can simplify line 124 to just
> 
> ```
> val license = it.artifactLicense ?: SpdxConstants.NOASSERTION
> ```
> 
> and use `license.toSpdx()` when creating the `OrtSnippet`?

Wenn no because you still need the `DeclaredLicenseProcessor.process` to correct wrong license from FossID.
I just pushed a suggestion, please have a look.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -114,27 +121,20 @@ internal fun <T : Summarizable> List<T>.mapSummary(
...
         rawSnippets.map {
-            val license = it.artifactLicense?.let {
-                runCatching {
-                    LicenseFinding.createAndMap(
-                        it,
-                        fakeLocation,
-                        detectedLicenseMapping = scannerConfig.detectedLicenseMapping
-                    ).license
-                }.onFailure { spdxException ->
+            val license = it.artifactLicense?.let { artifactLicense ->
```
sschuberth, line 127: After rebasing (which introduces the `snippet` lambda variable) I'd propose the slightly more compact:

            val license = snippet.artifactLicense?.let {
                DeclaredLicenseProcessor.process(it).also { expression ->
                    if (expression == null) {
                        issues += FossId.createAndLogIssue(
                            source = "FossId",
                            message = "Failed to map license '$it' as an SPDX expression."
                        )
                    }
                }
            } ?: SpdxConstants.NOASSERTION.toSpdx()

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -83,7 +83,13 @@ internal fun <T : Summarizable> List<T>.mapSummary(
 
         summary.licences.forEach {
             runCatching {
-                LicenseFinding.createAndMap(it.identifier, location, detectedLicenseMapping = detectedLicenseMapping)
+                DeclaredLicenseProcessor.process(it.identifier)?.let { spdx ->
```
sschuberth, line 86: Given our latest thoughts, it really correct to use a mapping of _declared_ licenses here, *and* in the fallback a mapping of _detected_ licenses?

If the line of thinking is that we only get detected licenses here, then we should only apply `detectedLicenseMapping` here.

Or phrased differently, the issue you're mentioning in the commit message would already be solved with the changes in `mapSnippetFindings()` alone, correct?

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
nnobelis: Yes you're right. The issue I mentioned only needs the fix in `mapSnippetFindings()`.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -36,12 +36,12 @@ import org.ossreviewtoolkit.model.RemoteArtifact
 import org.ossreviewtoolkit.model.Snippet as OrtSnippet
 import org.ossreviewtoolkit.model.SnippetFinding
 import org.ossreviewtoolkit.model.TextLocation
-import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.utils.PurlType
 import org.ossreviewtoolkit.utils.common.collapseToRanges
 import org.ossreviewtoolkit.utils.common.collectMessages
 import org.ossreviewtoolkit.utils.common.prettyPrintRanges
+import org.ossreviewtoolkit.utils.ort.DeclaredLicenseProcessor
```
sschuberth, line 44: This import should only get added in the next commit.

sschuberth (changes requested):
Only one more change request 😉

sschuberth (approved):

PR #7088, opened 2023-06-05
fix(Gradle): Avoid a JVM >= 17 warning when running tests

sschuberth:
This is a fixup for c903fbf to allow more packages to be accessible at runtime via reflection. While at it, slightly align the argument syntax to other similar code.


mnonnenmacher (approved):

sschuberth:
Merging despite the unrelated failure.

PR #7092, opened 2023-06-05
fix(scanner): Sort packages with the same provenance

nnobelis:

@mnonnenmacher FYI

### DIFF
--- a/scanner/src/main/kotlin/ScanController.kt
+++ b/scanner/src/main/kotlin/ScanController.kt
@@ -33,6 +33,7 @@ import org.ossreviewtoolkit.model.UnknownProvenance
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.scanner.provenance.NestedProvenance
 import org.ossreviewtoolkit.scanner.provenance.NestedProvenanceScanResult
+import org.ossreviewtoolkit.utils.common.PATH_STRING_COMPARATOR
 
 /**
  * A controller for the data related to a run of the [Scanner].
@@ -210,9 +211,14 @@ class ScanController(
     /**
      * Return a map of [KnownProvenance]s associated with all [packages] with the same provenance, ignoring any VCS
      * path. Packages without a resolved provenance are not included in the result.
+     * For a given provenance, the packages are sorted by the ascending amount of path separators '/' in their names,
+     * then by their identifiers. For instance, a package with name 'conanfile.txt' will come before one with name
+     * 'components/conanfile.txt'. This is because scanner interfaces receive packages as input, and this aims at
+     * providing a deterministic ordering when choosing a reference package for packages with the same provenance.
      */
     fun getPackagesConsolidatedByProvenance(): Map<KnownProvenance, List<Package>> {
-        val packagesByProvenance = mutableMapOf<KnownProvenance, MutableList<Package>>()
+        val packagesByProvenance = mutableMapOf<KnownProvenance, MutableSet<Package>>()
+        val comparator = compareBy<Package, String>(PATH_STRING_COMPARATOR) { it.id.name }.thenBy { it.id }
 
         packages.forEach { pkg ->
             val consolidatedProvenance = when (val provenance = packageProvenances[pkg.id]) {
@@ -221,10 +227,10 @@ class ScanController(
                 else -> provenance
             }
 
-            packagesByProvenance.getOrPut(consolidatedProvenance) { mutableListOf() } += pkg
+            packagesByProvenance.getOrPut(consolidatedProvenance) { sortedSetOf(comparator) } += pkg
         }
 
-        return packagesByProvenance
+        return packagesByProvenance.mapValues { it.value.toList() }
     }
 
     /**
--- a/utils/spdx/src/main/kotlin/Utils.kt
+++ b/utils/spdx/src/main/kotlin/Utils.kt
@@ -27,17 +27,13 @@ import java.net.URL
 import java.security.MessageDigest
 
 import org.ossreviewtoolkit.utils.common.Os
+import org.ossreviewtoolkit.utils.common.PATH_STRING_COMPARATOR
 import org.ossreviewtoolkit.utils.common.VCS_DIRECTORIES
 import org.ossreviewtoolkit.utils.common.calculateHash
 import org.ossreviewtoolkit.utils.common.encodeHex
 import org.ossreviewtoolkit.utils.common.isSymbolicLink
 import org.ossreviewtoolkit.utils.common.realFile
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants.LICENSE_REF_PREFIX
-
-/**
- * A comparator that sorts parent paths before child paths.
- */
-internal val PATH_STRING_COMPARATOR = compareBy<String>({ path -> path.count { it == '/' } }, { it })
 
 /**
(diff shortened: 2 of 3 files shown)


scanner/src/main/kotlin/ScanController.kt
```diff
@@ -212,7 +213,10 @@ class ScanController(
      * path. Packages without a resolved provenance are not included in the result.
      */
     fun getPackagesConsolidatedByProvenance(): Map<KnownProvenance, List<Package>> {
-        val packagesByProvenance = mutableMapOf<KnownProvenance, MutableList<Package>>()
+        val packagesByProvenance = mutableMapOf<KnownProvenance, MutableSet<Package>>()
+        val comparator: Comparator<Package> = compareBy<Package, String>(PATH_STRING_COMPARATOR) { it.id.name }.thenBy {
```
sschuberth, line 217: Drop the explicit `: Comparator<Package>`.

sschuberth (changes requested):

scanner/src/main/kotlin/ScanController.kt
```diff
@@ -212,7 +213,10 @@ class ScanController(
      * path. Packages without a resolved provenance are not included in the result.
      */
     fun getPackagesConsolidatedByProvenance(): Map<KnownProvenance, List<Package>> {
-        val packagesByProvenance = mutableMapOf<KnownProvenance, MutableList<Package>>()
+        val packagesByProvenance = mutableMapOf<KnownProvenance, MutableSet<Package>>()
+        val comparator = compareBy<Package, String>(PATH_STRING_COMPARATOR) { it.id.name }.thenBy {
+            it.id
+        }
```
sschuberth, line 219: This should fit into a single line now.

sschuberth (changes requested):

scanner/src/main/kotlin/ScanController.kt
```diff
@@ -33,6 +33,7 @@ import org.ossreviewtoolkit.model.UnknownProvenance
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.scanner.provenance.NestedProvenance
 import org.ossreviewtoolkit.scanner.provenance.NestedProvenanceScanResult
+import org.ossreviewtoolkit.utils.common.PATH_STRING_COMPARATOR
```
fviernau, line 36: Why is it value-able to have this choice be deterministic?

(To my understanding, the only reason to choose a reference package is because the API to scan packages still accepts a `Package` as input, while it actually wants to have only a provenance. So, it does not matter what package is passed, because only it's provenance is considered).

fviernau (changes requested):

fviernau:
In case you plan to do further refactorings in `Scanner.kt` or `ScanController.kt`, let's please sync - as I'm also currently doing so - to avoid rebase conflicts.

sschuberth (commented):
LGTM, but I'd like @mnonnenmacher to approve as he seems to know more about the context.

scanner/src/main/kotlin/ScanController.kt
nnobelis: In the snippet choice feature that I am developing, the user writes a snippet choice in his/her `.ort.yml`.

The syntax is roughly:
```
package_snippet_choices:
  - id: "<package id>"
    snippet_choices:
    - license: "<snippet licenses>"
      reasoning: ""
      match:
        path: "<source file path>"
        snippet: "<snippet purl>"
```
Because the snippets are coming from a specific package scan.

The scanner correlates the snippet choice with the current scanner results, to adapt the license findings in consequence.
I guess same issue would arise for snippet curation, if we would like to fix the snippet findings.

But I agree, the day the scanners accept a provenance instead of a package, then this "feature" won't be needed anymore (and I need would adapt the snippet choice in consequence).

Anyway I think we will discuss it in details when I will present the snippet choice to ORT developers.

nnobelis:
> @nnobelis - In case you plan to do further refactorings in `Scanner.kt` or `ScanController.kt`, let's please sync - as I'm also currently doing so - to avoid rebase conflicts.

I don't plan on touching the `ScanController` any further.
However I have, in a branch for the snippet choice, some small refactoring to the `Scanner.kt` and `ScanContext.kt` to expose the `RepositoryConfiguration` to the scanner.
But this is pretty trivial so I can rebase upon your changes.

utils/spdx/src/main/kotlin/Utils.kt
```diff
@@ -27,17 +27,15 @@ import java.net.URL
...
 import org.ossreviewtoolkit.utils.common.encodeHex
 import org.ossreviewtoolkit.utils.common.isSymbolicLink
 import org.ossreviewtoolkit.utils.common.realFile
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants.LICENSE_REF_PREFIX
 
-/**
- * A comparator that sorts parent paths before child paths.
- */
-internal val PATH_STRING_COMPARATOR = compareBy<String>({ path -> path.count { it == '/' } }, { it })
+
```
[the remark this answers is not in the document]
mnonnenmacher: Line is still there.

scanner/src/main/kotlin/ScanController.kt
mnonnenmacher, line 36: Commit message nit: "sorting the package**s**"

scanner/src/main/kotlin/ScanController.kt
```diff
@@ -212,7 +213,8 @@ class ScanController(
      * path. Packages without a resolved provenance are not included in the result.
```
mnonnenmacher, line 213: Please also mention the sorting in the function docs and ideally also explain the reason.

mnonnenmacher (changes requested):

scanner/src/main/kotlin/ScanController.kt
nnobelis: Please review the wording.

utils/spdx/src/main/kotlin/Utils.kt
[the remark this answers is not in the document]
nnobelis: This is an outdated version of the file. Line is ok now or ?!

scanner/src/main/kotlin/ScanController.kt
mnonnenmacher: Nit: "amount of path separator**s**", comma after "For instance"
I would remove the last TODO sentence, the `Scanner` itself will always have to take packages and then do the provenance resolution. This function is related to finding packages which share the same provenance. After the provenance resolution this data is passed to scanners which operate either on a reference package like FossId or that operate on provenances like the current ScanCode implementation.

mnonnenmacher (changes requested):

scanner/src/main/kotlin/ScanController.kt
nnobelis: If you want we can discuss if the migration to the provenance based interface for FossID makes sense.

mnonnenmacher (approved):

nnobelis:
Why the `funTest-docker` was skipped ?
Can it be merged or do the tests need to be rerun ?

sschuberth:
> @sschuberth here are the failed test logs:

That's a know issue currently, no need to rerun tests. But @fviernau needs to approve or dismiss his previous request for changes.

nnobelis:
Could we please merge this ?

PR #7093, opened 2023-06-05
refactor: Factor out `scannerRunOf()`

fviernau:
Simplify refactoring `ScannerRun` in upcoming changes.

Part of: #5950.


evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
```diff
@@ -36,10 +36,10 @@ import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
 import org.ossreviewtoolkit.model.ScanSummary
 import org.ossreviewtoolkit.model.ScannerDetails
-import org.ossreviewtoolkit.model.ScannerRun
 import org.ossreviewtoolkit.model.TextLocation
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.utils.test.scannerRunOf
```
sschuberth, line 42: Nit: To me, the commit message sounds as if `scannerRunOf()` was already existing before this change, whereas this change actually introduces `scannerRunOf()`.

How about improving this by referring to the already existing `advisorRunOf()` like

```
refactor: Extract code to create a `ScannerRun`

Similar like already done with `advisorRunOf()`, extract the code to create
a `ScannerRun` to a `scannerRunOf()` function.
```

sschuberth: And while at it, should another commit also move `advisorRunOf()` to `test-utils` for consistency, even if it's currently only being used in `FreeMarkerTemplateProcessorTest`?

sschuberth (changes requested):

sschuberth (approved):

PR #3573, opened 2021-02-02
NoticeTemplateReporter: Add an option for treating projects like packages

fviernau:
ORT has two different notice reporter templates. The notice summary
template just lists all licenses in one large list. The default
template has one joint list for the licenses of all projects and
separate lists for the licenses of each package.

In some cases it makes sense to also list the licenses separetly for (some
of the) projects. In particular, listing the licenses for
SpdxDocumentFile projects separately is interresting, because this
allows to create a notice by package report also in cases where no
package manager is involved and the source code of the project dependencies
is just included in the source tree of the analyzed project.

In order to support that use case the scan results have to be
deduplicated for the projects which are to be reported separately in the
notices, because the scan results of projects may overlap - which would
in turn result in licenses being reported twice, in the list for the
respective package and in the joint licenses list for the projects.

PR #7095, opened 2023-06-05
refactor(reporter)!: Remove the option `OPTION_PROJECT_TYPES_AS_PACKA…

fviernau:
…GES`

The option had been introduced [1] to enable producing a "by package" NOTICE file, in case the dependencies are included as source code into the project's source tree, without the use of any package manager.

A better solution would have been to implement treating certain sub directories of the projects source tree as dependencies consistently across stages of the analysis. For example, the evaluator could then execute the policy rules for dependencies, not (accidentally) the ones for projects. Doing so was not feasible back then due to the size of the task.

The implemented solution obviously is a shortcut, it is quite complex, not straight forward / obvious to use and also fragile, as debugging the behavior involves logic implemented inside the freemarker templates itself.

As nobody seems to be using this option anymore, remove it to get rid of the complexity. Mainly, because the implemented shortcut does not work well together with storing scan results by provenance in `OrtResult` which is going to be put in place in upcomining changes. The reason why it doesn't is because scan results won't be stored redundantly per package / project, but per provenance, which permits filtering the findings in the scan results with project / package context.

This revert 3aff6e7.

[1] https://github.com/oss-review-toolkit/ort/pull/3573

Part of: #5950.


sschuberth:
> This revert [3aff6e7](https://github.com/oss-review-toolkit/ort/commit/3aff6e7ee9a849061ba5121cfc7eb4510cc45bed).

s/revert/reverts/

sschuberth:
Typos also in:

- s/freemarker/FreeMarker/
- s/upcomining/upcoming/

sschuberth (approved):

sschuberth (approved):

PR #7096, opened 2023-06-05
Simplify detected license mapping

sschuberth:
@fviernau @nnobelis this may trivially conflict with your work. I'm fine with waiting and rebasing this last and resolve any conflicts.


sschuberth:
Current conflicts have been resolved and this is ready for review.

fviernau:
> @fviernau @nnobelis this may trivially conflict with your work. I'm fine with waiting and rebasing this last and resolve any conflicts.

Thanks @sschuberth! I'm fine if you merge before me, wrt. rebase conflicts.

sschuberth:
> Thanks @sschuberth! I'm fine if you merge before I merge mine, wrt. rebase conflicts.

Great! Mind also reviewing / approving then?

model/src/main/kotlin/LicenseFinding.kt
```diff
@@ -79,7 +61,7 @@ data class LicenseFinding(
  * Apply [detectedLicenseMapping] from the [org.ossreviewtoolkit.model.config.ScannerConfiguration] to any license
  * String.
  */
-private fun String.applyDetectedLicenseMapping(detectedLicenseMapping: Map<String, String>): String =
+fun String.mapLicense(detectedLicenseMapping: Map<String, String>): String =
```
fviernau, line 64: The new function name makes it less obvious that the function is only "detected" licenses. How about renaming the function to `mapDetectedLicense` instead? (Maybe the parameter name can then be shortened to just "mapping", which also make the tests a bit less verbose)

sschuberth: I was actually tempted to do the "opposite": Keep the function name and also generalize the parameter name. There's actually nothing "detected license"-specific in that function, and maybe not even something "(any) license"-specific. It's just about folding string replacements. What *makes* the function "detected license"-specific is the *call* to it with a detected license mapping. So at the call sites, it's IMO obvious what this is about. The only reason I kept "License" as part of "mapLicense" as "map" alone is already taken, and you could argue that the way the regex is design makes it specific to the license replacement use-case.

fviernau (approved):

PR #7097, opened 2023-06-06
fix(arm64): Use scancode-toolkit-mini for arm64 arch

heliocastro:
Arm64 are able to install scancode-toolkit-mini

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -153,11 +153,13 @@ RUN pip install --no-cache-dir -U \
     python-inspector=="${PYTHON_INSPECTOR_VERSION}"
 
 RUN arch=$(arch | sed s/aarch64/arm64/) \
-    && if [ "$arch" != "arm64" ]; then \
-        curl -Os https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt; \
-        pip install -U --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION; \
-        rm requirements.txt; \
-       fi
+    &&  if [ "$arch" == "arm64" ]; then \
+            pip install -U scancode-toolkit-mini==$SCANCODE_VERSION; \
+        else \
+            curl -Os https://raw.githubusercontent.com/nexB/scancode-toolkit/v$SCANCODE_VERSION/requirements.txt; \
+            pip install -U --constraint requirements.txt scancode-toolkit==$SCANCODE_VERSION; \
+            rm requirements.txt; \
+        fi
 
 FROM scratch AS python
 COPY --from=pythonbuild /opt/python /opt/python


sschuberth:
@pombredanne I was unable to find information about what exactly this "mini" version of ScanCode actually is. Is there a reference that explains how it differs from the regular version?

tsteenbe:
@sschuberth  See https://pypi.org/project/scancode-toolkit-mini/ - ScanCode is a tool to scan code for license, copyright, package and their documented dependencies and other interesting facts. scancode-toolkit-mini is a special build that does not come with pre-built binary dependencies by default. These are instead installed separately or with the extra_requires scancode-toolkit-mini

sschuberth:
> scancode-toolkit-mini is a special build that does not come with pre-built binary dependencies by default. These are instead installed separately or with the extra_requires scancode-toolkit-mini

So, since the PR doesn't install anything extra on top, would such a ScanCode installation even work?

tsteenbe:
Discussed in the ORT community meeting of June 15th - agreed with @heliocastro  @fviernau @mnonnenmacher  and @tsteenbe:

1. From @mnonnenmacher  Please invert the logic that the if (arm) { ScanCode mini} else { full ScanCode}
2. Update the commit message to:

```
fix(arm64): Use scancode-toolkit-mini for arm64 arch

ScanCode mini does not come with pre-compiled binaries which results in a slower execution.
However, these pre-compiled binaries are not needed on arm64 and are causing runtime issues.
Choosing to have a slower bootstrapped ScanCode rather than a broken one.
```

3. Look into rebuilding license database `--reindex-licenses` in a follow issue/PR.

sschuberth:
> ScanCode mini does not come with pre-compiled binaries which results in a slower execution.

I think that's still confusing. Not coming with pre-compiled binaries (without any substitute) leaves unclear how things could be working *at all*, and not just slower. Are binaries then compiled on-the-fly? Or are native binaries substituted with Python implementations, and that's why it's slower? Please clarify this in the commit message.

fviernau (approved):

sschuberth:
While the commit message is extensive now, it still uses the misleading term "pre-compiled" (Compiled to what? Python bytecode or native code?) and misses the point of what packages are used instead.

sschuberth:
> While the commit message is extensive now, it still uses the misleading term "pre-compiled" (Compiled to what? Python bytecode or native code?) and misses the point of what packages are used instead.

FYI, I've filed https://github.com/nexB/scancode-toolkit/issues/3432 to clarify independently of this PR.

sschuberth:
> FYI, I've filed [nexB/scancode-toolkit#3432](https://github.com/nexB/scancode-toolkit/issues/3432) to clarify independently of this PR.

Quoting @pombredanne from https://github.com/nexB/scancode-toolkit/issues/3432:

> [Missing dependencies] are not replaced. The feature that needs them will just not work or crash.

> In particular, I do not think that scancode-toolkit-mini is a solution to run on ARM.

PR #7099, opened 2023-06-06
OrtResult: Store scan results "by provenance"

### LINKED ISSUE #5950: OrtResult: Store scan results "by provenance"

fviernau:
This PR changes `OrtResult`s internal representation to store scan results by provenance, which eliminates redundancies in case multiple packages share the same provenance. It keeps `OrtResults` API _as-is_. That is, client code using `OrtResult.getScanResults()` or `OrtResult.getScanResults(id)` remains _as-is_, thereby hiding the complexities of the new internal representation completely. Also, information about the sub-repository structure is now available for all packages in `OrtResult`.

**Breaking change:** 

1. "old" ORT files containing a `ScannerRun` won't de-serialize anymore.
2. The scanner cannot output package verification codes anymore. This will be fixed again, by computing
    these codes on-the-fly from file lists which are not present yet but will be added soon.

Fixes #5950.

### DIFF
--- a/cli/src/funTest/assets/semver4j-ort-result.yml
+++ b/cli/src/funTest/assets/semver4j-ort-result.yml
@@ -184,125 +184,140 @@ scanner:
     - "postgresStorage"
     storage_writers:
     - "postgresStorage"
+  provenances:
+  - id: "Maven:com.vdurmont:semver4j:3.1.0"
+    package_provenance:
+      vcs_info:
+        type: "Git"
+        url: "https://github.com/vdurmont/semver4j.git"
+        revision: "7653e418d610ffcd2811bcb55fd72d00d420950b"
+        path: ""
+      resolved_revision: "7653e418d610ffcd2811bcb55fd72d00d420950b"
+  - id: "Maven:junit:junit:4.12"
+    package_provenance:
+      vcs_info:
+        type: "Git"
+        url: "https://github.com/junit-team/junit.git"
+        revision: "r4.12"
+        path: ""
+      resolved_revision: "64155f8a9babcfcf4263cf4d08253a1556e75481"
   scan_results:
-    Maven:com.vdurmont:semver4j:3.1.0:
-    - provenance:
-        vcs_info:
-          type: "Git"
-          url: "https://github.com/vdurmont/semver4j.git"
-          revision: "7653e418d610ffcd2811bcb55fd72d00d420950b"
-          path: ""
-        resolved_revision: "7653e418d610ffcd2811bcb55fd72d00d420950b"
-      scanner:
-        name: "ScanCode"
-        version: "3.2.1-rc2"
-        configuration: "--copyright --license --ignore *.ort.yml --info --strip-root\
-          \ --timeout 300 --ignore META-INF/DEPENDENCIES --json-pp"
-      summary:
-        start_time: "2020-09-30T09:27:12.023451Z"
-        end_time: "2020-09-30T09:28:20.525647Z"
-        package_verification_code: "48ba11487d53ce933b5d4db1d069b70a803ff19b"
-        licenses:
-        - license: "BSD-3-Clause"
-          location:
-            path: "pom.xml"
-            start_line: 28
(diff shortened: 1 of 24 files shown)


scanner/src/main/kotlin/Scanner.kt
```diff
@@ -148,14 +149,24 @@ class Scanner(
...
+            scanResults = (projectResults.scanResults + packageResults.scanResults).mapTo(mutableSetOf()) {
+                it.clearPackageVerificationCode()
+            }
+        )
 
         return ortResult.copy(scanner = scannerRun)
     }
 
-    suspend fun scan(packages: Set<Package>, context: ScanContext): Map<Identifier, List<ScanResult>> {
+    suspend fun scan(packages: Set<Package>, context: ScanContext): ScannerRun {
```
mnonnenmacher, line 167: As this is a public function I would have introduced a dedicated class for the return value which only contains the provenance resolution results and the scan results, because only those properties are required. Returning a `ScannerRun` here is misleading when using the function programmatically (e.g. the timestamps are wrong and the config is not filtered unlike with the other scan function).

mnonnenmacher (changes requested):

scanner/src/main/kotlin/Scanner.kt
fviernau: I spent quite a while trying to do that. It was not too easy.

Would you be fine if I addressed this in a follow up PR? (Currently, this function is only called from the test)

cli/src/funTest/assets/semver4j-ort-result.yml
```diff
@@ -184,125 +184,140 @@ scanner:
     - "postgresStorage"
     storage_writers:
     - "postgresStorage"
+  provenances:
+  - id: "Maven:com.vdurmont:semver4j:3.1.0"
```
mnonnenmacher, line 188: The commit message should describe the actual change of the internal representation better, it currently focuses only on compatibility. I would suggest to add a first paragraph that shortly explains that scan results are now stored by provenance and that the package provenance resolution results are stored as well to associate scan results with packages.

mnonnenmacher (changes requested):

scanner/src/funTest/assets/dummy-expected-scan-results-for-analyzer-result.yml
```diff
@@ -0,0 +1,132 @@
+---
```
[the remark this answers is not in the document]
mnonnenmacher: That's still open.

sschuberth, line 1: Commit message nits:
- `OrtResults` should probably be `OrtResult`s or `OrtResult`'s (as in the next commit)
- Not sure about the comma before "by"

scanner/src/main/kotlin/Scanner.kt
fviernau: @mnonnenmacher - I've extracted this to an issue and added it to the corresponding epic, see:
- https://github.com/oss-review-toolkit/ort/issues/7155
- https://github.com/oss-review-toolkit/ort/issues/6945

fviernau:
@mnonnenmacher I've made a final manual test, I compared scan results of main and this branch and verified that

1. static HTML reports are identical
2. output of the list licenses command is identical for some selected package IDs

I've done two scans:

1. Our `orthw` example mime types scan
2. This sub-repositoy case: https://github.com/fviernau/test-data.git

PR #7103, opened 2023-06-07
fix(fossid-webapp): Set severity to HINT when snippet license cannot be mapped

nnobelis:
With the introduction of snippet findings in the scan results, some
issues are also present in the result because licenses from FossID
snippets cannot always be mapped to SPDX, for instance "RETURNN license"
or "Apache 2-0".
While the usage of the declared license mapping file helped to
alleviate some of these issues, several remained, polluting the webapp
report.

Opening a support ticket at FossID to have them correct the license,
while being the correct solution, does not scale regarding the amount of
licenses to correct.
Therefore, this commit lowers for now the severity of these issues to
`HINT` and map them to `NO_ASSERTION`. This is a temporary solution
until a snippet curation mechanism is introduced to be able to correct
those licenses.

FYI @MarcelBochtler.

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
@@ -33,6 +33,7 @@ import org.ossreviewtoolkit.model.Issue
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.PackageProvider
 import org.ossreviewtoolkit.model.RemoteArtifact
+import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.model.Snippet as OrtSnippet
 import org.ossreviewtoolkit.model.SnippetFinding
 import org.ossreviewtoolkit.model.TextLocation
@@ -124,7 +125,8 @@ internal fun mapSnippetFindings(
                     if (expression == null) {
                         issues += FossId.createAndLogIssue(
                             source = "FossId",
-                            message = "Failed to map license '$it' as an SPDX expression."
+                            message = "Failed to map license '$it' as an SPDX expression.",
+                            severity = Severity.HINT
                         )
                     }
                 }
--- a/scanner/src/test/kotlin/scanners/fossid/FossIdLicenseMappingTest.kt
+++ b/scanner/src/test/kotlin/scanners/fossid/FossIdLicenseMappingTest.kt
@@ -35,6 +35,7 @@ import org.ossreviewtoolkit.clients.fossid.model.result.LicenseCategory
 import org.ossreviewtoolkit.clients.fossid.model.result.MatchType
 import org.ossreviewtoolkit.clients.fossid.model.result.Snippet
 import org.ossreviewtoolkit.model.Issue
+import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 import org.ossreviewtoolkit.utils.test.shouldNotBeNull
@@ -91,7 +92,9 @@ class FossIdLicenseMappingTest : WordSpec({
 
             issues should haveSize(1)
             issues.first() shouldNotBeNull {
-                message shouldStartWith "Failed to map license 'invalid license' as an SPDX expression."
+                message shouldStartWith
+                        "Failed to map license 'invalid license' as an SPDX expression."
+                severity shouldBe Severity.HINT
             }
             findings should haveSize(1)
             findings.first() shouldNotBeNull {


scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -33,6 +33,7 @@ import org.ossreviewtoolkit.model.Issue
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.PackageProvider
 import org.ossreviewtoolkit.model.RemoteArtifact
+import org.ossreviewtoolkit.model.Severity
```
sschuberth, line 36: Commit message:
- Is the double "N" in "RETURNN" really correct?
- "alleviate" without "s" at the end
- If the middle section is supposed to be its own paragraph, please make it stand out by adding empty lines around it.

sschuberth (dismissed):

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -124,10 +125,11 @@ internal fun mapSnippetFindings(
                     if (expression == null) {
                         issues += FossId.createAndLogIssue(
                             source = "FossId",
-                            message = "Failed to map license '$it' as an SPDX expression."
+                            message = "Failed to map license '$it' as a SPDX expression. Mapping it using licenseRef-.",
+                            severity = Severity.HINT
                         )
                     }
-                }
+                } ?: "LicenseRef-fossid-${it.replace("[^a-zA-Z0-9\\-.]".toRegex(), "-")}".toSpdx()
```
fviernau, line 132: In the analyzer we have `unmappedDeclaredLicenses` field in the model and only a warning (or hint?) if the license could not be mapped. This allows to write a (tweaked to users needs) policy rule to flag the mapping issue, under desired severity under desired preconditions. See, `rules.kts` in `ort-config` repository. 

How about doing the same here?

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -124,10 +125,11 @@ internal fun mapSnippetFindings(
                     if (expression == null) {
                         issues += FossId.createAndLogIssue(
                             source = "FossId",
-                            message = "Failed to map license '$it' as an SPDX expression."
+                            message = "Failed to map license '$it' as a SPDX expression. Mapping it using licenseRef-.",
```
fviernau, line 128: commit: Mapping to a `LicenseRef-*` in my view does not make sense at all if it is not clear what the semantics of the ID actually is. This seems to be the case here, which is why I propose to not do this.

I propose the workflow to be as follows:
1. unmapped license strings are flagged as hint. Optionally by a policy rule as must fix, e.g. error.
2. In order to fix the mapping, one has to manually define a "license mapping" similar to the declared license mapping curation.

What do you think?

fviernau (changes requested):

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
nnobelis: >     * Is the double "N" in "RETURNN" really correct?

image
Yes

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
nnobelis: Answered below.

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
sschuberth: > As FossID is a black box, we don't know if it is a declared license or a detected license.

Wasn't it so that as a user, you can add licenses in the FossID web UI, and instruct FossID to use licenses from there for identified snippets? If so, I think it would be rather "declared" licenses.

nnobelis: Yes in FossID you can add licenses and edit them a "License View".
Then, when a file is pending identification (i.e. has several matching snippets) you can:
- use the license of a snippet (without modification)
- manually assign a license of the "License View" of FossID

Therefore, I believe this license editor is NOT for snippets nor to curate snippets.

nnobelis: @fviernau An opinion on my comment ?

fviernau: > After discussion with @MarcelBochtler we also think you are right, that the license refs are not suitable here.

So, can we change the approach to output an issue saying that some (parts of the) license string could not be mapped, instead of creating `License-Ref-` IDs from strings with unknown semantics?

> However, dealing with an unmapped license with a declared license mapping is not correct.

Agree. However, I believe a mechanism which is quite a bit similar could make sense.


> Therefore, here are our suggestions:
> 
>     * Change the severity in case of unmappeable license to `HINT`, to remove the flooding of report with snippet related issues..
> 
>     * When there is a possibility to do snippet curation with ORT, bump up the severity to `WARNING`. Then the user will fix the issue with a (hopefully global) snippet curation.

I'm fine with these suggestions, as long as the `License-Refs` construction gets dropped. 
As future perspective, I believe it also makes sense to add an `unmapped_license` property to the snippet model.

fviernau: I'm ok if you don't do the unmapped license property at all in scope of your task. I just wanted to have some common vision for the future.

fviernau: Maybe setting the snippet license to `NO_ASSERTION` plus flagging an issue that the license could not be mapped, would be reasonable?

nnobelis: This is the current logic:
See https://github.com/bosch-io/oss-review-toolkit/blob/131fd5759440054ea77164713598c16b7abfefac/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt#L123-L133
and
https://github.com/bosch-io/oss-review-toolkit/blob/131fd5759440054ea77164713598c16b7abfefac/scanner/src/test/kotlin/scanners/fossid/FossIdLicenseMappingTest.kt#L94-L101

fviernau: The diff LGTM now. The commit message needs to be aligned with it.

nnobelis: I adapted a bit the commit message to mention the `NO_ASSERTION`.
Should anything more in it be changed ?

fviernau (approved):

PR #7104, opened 2023-06-07
Compare command

sschuberth:

### DIFF
--- a/plugins/commands/compare/src/main/kotlin/CompareCommand.kt
+++ b/plugins/commands/compare/src/main/kotlin/CompareCommand.kt
@@ -1,0 +1,130 @@
+/*
+ * Copyright (C) 2021 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


plugins/commands/compare/src/main/kotlin/CompareCommand.kt
```diff
@@ -0,0 +1,127 @@
...
+        .convert { it.absoluteFile.normalize() }
+
+    private val fileB by argument(help = "The second ORT result file to compare.")
+        .convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
+        .convert { it.absoluteFile.normalize() }
+
+    private val method by option(
+        "--method", "-m",
+        help = "The method to use when comparing ORT results."
```
mnonnenmacher, line 56: I don't remember, does Clikt automatically show the available options for enum parameters? Otherwise it would be nice to list them here.

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+    }
+}
+
+enum class CompareMethod {
+    TEXT_DIFF
+}
+
+fun unifiedDiff(a: File, b: File, replacements: Map<String, String> = emptyMap(), contextSize: Int = 7): List<String> {
```
mnonnenmacher, line 101: Could be private.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #7112, opened 2023-06-12
feat(docker): Build independent images

heliocastro:
All components no are built as independent docker images.


sschuberth:
I guess the successor is https://github.com/oss-review-toolkit/ort/pull/7447.

PR #7115, opened 2023-06-13
refactor: Turn some hard-coded directories into sets

sschuberth:

### DIFF
--- a/analyzer/src/main/kotlin/PackageManager.kt
+++ b/analyzer/src/main/kotlin/PackageManager.kt
@@ -80,7 +80,7 @@ abstract class PackageManager(
          */
         val ENABLED_BY_DEFAULT by lazy { ALL.values.filter { it.isEnabledByDefault } }
 
-        private val PACKAGE_MANAGER_DIRECTORIES = listOf(
+        private val PACKAGE_MANAGER_DIRECTORIES = setOf(
             // Ignore intermediate build system directories.
             ".gradle",
             ".yarn",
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/utils/FileListResolver.kt
```diff
@@ -74,18 +74,14 @@ private fun ProvenanceFileStorage.getFileList(provenance: KnownProvenance): File
...
-        !IGNORED_DIRECTORY_MATCHER.matches(it.relativeTo(dir).invariantSeparatorsPath)
+        it.name !in VCS_DIRECTORIES
     }.filter {
         it.isFile
     }.mapTo(mutableSetOf()) {
         FileEntry(path = it.relativeTo(dir).invariantSeparatorsPath, sha1 = HashAlgorithm.SHA1.calculate(it))
     }
 
-    return FileList(ignorePatterns = IGNORED_DIRECTORY_MATCHER.patterns.toSet(), files = files)
+    return FileList(ignorePatterns = VCS_DIRECTORIES.mapTo(mutableSetOf()) { "**/$it" }, files = files)
```
fviernau, line 86: Previously, the patterns passed to file list have been used to filter things. So, there was no risk that the patterns passed to `FileList` is inconsistent with the patterns used for the filtering. 

The new code requires the lambda in line 78 - 80 to be consistent with the patterns passed in line 86. Therefore, I slightly prefer the previous code.

What do you think?

fviernau (changes requested):

scanner/src/main/kotlin/utils/FileListResolver.kt
sschuberth: I've dropped that commit.

fviernau (approved):

PR #7116, opened 2023-06-13
fix(PackageCurationData): Preserve the operator when applying curations

MarcelBochtler:
When applying a curation to a package using a `SpdxCompoundExpression` the operator should not be overridden with the default `AND`. This fixes an issue that license choices were not taken into account when using the Downloader to download an archive for specified license categories.


MarcelBochtler:
I don't think that this change can be done on the basis of the `ProcessDeclaredLicense.spdxExpression` which states that it is `null` if no license was mapped.
Closing this PR for now.

PR #7117, opened 2023-06-13
Minor SPDX `calculatePackageVerificationCode()` improvements

sschuberth:


utils/spdx/src/main/kotlin/Utils.kt
```diff
@@ -91,6 +91,8 @@ fun calculatePackageVerificationCode(files: Sequence<File>, excludes: Sequence<S
 @JvmName("calculatePackageVerificationCodeForDirectory")
 fun calculatePackageVerificationCode(directory: File): String {
     val allFiles = directory.walk().onEnter { !it.isSymbolicLink() }.filter { !it.isSymbolicLink() && it.isFile }
+
+    // Filter twice as "partition" does not return sequences.
```
fviernau, line 97: My initial reaction about this comment was to try to find the line of code which uses partition.
So, I believe the comment can be confusing. How about saying something like "Filter twice instead of using partition to operate on sequances."?

fviernau (changes requested):

fviernau (approved):

Issue #7118, opened 2023-06-13
Consider refactoring `ProvenanceFileStorage`


sschuberth:
The current `ProvenanceFileStorage` interface and implementations have some potentially confusing quirks:

- The `fun putFile(provenance: KnownProvenance, file: File)` function takes a file, but actually only its *contents* are stored. In particular, the file *name* is lost and cannot be assumed to be maintained when calling `fun getFile(provenance: KnownProvenance): File?`. As a result, also the extension of the original file is lost, so one cannot deduce the file type (which implies how its contents are to be interpreted) anymore.
- As another consequence, only a single file can be associated by a given `KnownProvenance` per `ProvenanceFileStorage` instance.
- Also, as the interface operates on files, data that is in memory needs to be serialized to disk first before it can be stored, which may be inefficient.

As a straight-forward solution to address this, I'd propose to make the interface operator on Input- / OutputStreams rather than files. Alternatively (but with significant more work), the file name could also be (re-)stored, and multiple files (with different names) per storage could be allowed.

Issue #6765, opened 2023-03-29
Scanner: Storing the scan result via HTTP storage backend fails with WebDAV on a Caddy-based HTTP file server


AndreasHennig:
The documentation states that storing a scan result using an HTTP storage backend works with "any HTTP file server"
(https://github.com/oss-review-toolkit/ort#http-storage)

I set up a web server using Caddy (https://caddyserver.com/) with the WebDAV module using the following Caddyfile:

```
:8080 {
  root /var/www/ort-cache
  webdav / {
    scope /var/www/ort-cache/
    modify true
  }
  log stdout
}
```

The `scanner` section in my _config.yml_ contains the following definition:
```yml
scanner:
    # (...)
    storages:
      http:
        backend:
          httpFileStorage:
            url: 'https://url-to-my-webserver.de'
            headers:
              X-JFrog-Art-Api: "dummy" # See https://github.com/oss-review-toolkit/ort/issues/6758#issuecomment-1486477756

    storageReaders: ["http"]
    storageWriters: ["http"]
    # (...)
```

I am doing the scan with the following command:
```bash
./cli/build/install/ort/bin/ort --info scan --package-types PACKAGE -f JSON -i path/to/analyzer-result.json -o path/to/output/folder
```

However, storing the scan result fails with the following output:
```bash
[main] WARN  org.ossreviewtoolkit.scanner.storages.ProvenanceBasedFileStorage - Could not store scan result for 'RepositoryProvenance(vcsInfo=VcsInfo(type=Git, url=https://github.com/PrismLibrary/Prism.git, revision=15140a61976d0a224cd6ebb9ee1f7ca63db02b47, path=), resolvedRevision=15140a61976d0a224cd6ebb9ee1f7ca63db02b47)' at path 'repository/Git/https%3A%2F%2Fgithub.com%2FPrismLibrary%2FPrism.git/15140a61976d0a224cd6ebb9ee1f7ca63db02b47/scan-results.yml': IOException: Could not store file at 'https://url-to-my-webserver.de/repository/Git/https%3A%2F%2Fgithub.com%2FPrismLibrary%2FPrism.git/15140a61976d0a224cd6ebb9ee1f7ca63db02b47/scan-results.yml': 404 -
```

So the web server responses with a `404 Not Found` error, stating that the looked up resource cannot be found.

[...]

PR #7119, opened 2023-06-13
`ProvenanceFileStorage`-related improvements

sschuberth:

### DIFF
--- a/model/src/main/kotlin/utils/FileProvenanceFileStorage.kt
+++ b/model/src/main/kotlin/utils/FileProvenanceFileStorage.kt
@@ -33,7 +33,8 @@ import org.ossreviewtoolkit.utils.ort.createOrtTempFile
 import org.ossreviewtoolkit.utils.ort.storage.FileStorage
 
 /**
- * A [FileStorage] based implementation of [ProvenanceFileStorage].
+ * A [FileStorage]-based implementation of [ProvenanceFileStorage] that stores files associated by [KnownProvenance]
+ * in files.
  */
 class FileProvenanceFileStorage(
     /**
@@ -42,7 +43,7 @@ class FileProvenanceFileStorage(
     private val storage: FileStorage,
 
     /**
-     * The filename of the files.
+     * The name of the files to use for storing the contents of the associated files.
      */
     private val filename: String,
 ) : ProvenanceFileStorage {
--- a/model/src/main/kotlin/utils/ProvenanceFileStorage.kt
+++ b/model/src/main/kotlin/utils/ProvenanceFileStorage.kt
@@ -23,22 +23,22 @@ import java.io.File
 import org.ossreviewtoolkit.model.KnownProvenance
 
 /**
- * A storage which associates a file with a [KnownProvenance]
+ * A generic storage interface that associates a [KnownProvenance] with a file.
  */
 interface ProvenanceFileStorage {
     /**
-     * Return whether a file corresponding to [provenance] exists.
+     * Return whether a file is associated by [provenance].
      */
     fun hasFile(provenance: KnownProvenance): Boolean
 
     /**
-     * Associate the [file] to [provenance]. Overwrites any existing file corresponding to [provenance].
+     * Associate [provenance] with the given [file]. Overwrites any existing association by [provenance].
      */
     fun putFile(provenance: KnownProvenance, file: File)
 
     /**
-     * Return the file corresponding to [provenance] or null if no such file exists. The returned file is a
-     * temporary file.
+     * Return the file associated by [provenance], or null if there is no such file. Note that the returned file is a
+     * temporary file that the caller is responsible for.
      */
     fun getFile(provenance: KnownProvenance): File?
 }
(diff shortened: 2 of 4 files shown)


model/src/main/kotlin/utils/ProvenanceFileStorage.kt
```diff
@@ -23,21 +23,21 @@ import java.io.File
...
 
     /**
-     * Associate the [file] to [provenance]. Overwrites any existing file corresponding to [provenance].
+     * Associate [provenance] with the given [file]. Overwrites any existing association for [provenance].
      */
     fun putFile(provenance: KnownProvenance, file: File)
 
     /**
-     * Return the file corresponding to [provenance] or null if no such file exists. The returned file is a
+     * Return the file belonging to [provenance], or null if there is no such file. The returned file is a
```
fviernau, line 40: Should we use "associated with" instead of "belonging" to use same terminology as in `putFile` ?

(Same for docs of `hasFile()`)

model/src/funTest/kotlin/utils/DatabaseProvenanceFileStorageFunTest.kt
```diff
@@ -58,14 +58,14 @@ private fun TestConfiguration.writeTempFile(content: String): File =
         writeText(content)
     }
 
-class PostgresProvenanceFileStorageFunTest : WordSpec({
+class DatabaseProvenanceFileStorageFunTest : WordSpec({
```
fviernau, line 61: I believe is this renaming was done consistently, it'd be a large change.
From this diff I can see the following postgres specific things still
- PostgresListener
- PostgresStorageConfiguration
- val postgresStorage

Apart from that there are further storages, like ScanResultsStorage which also use the term postgres. Also AFAIK these only have been tested with Postgres, so generalizing might open the way for untested territory. Furthermore, do you think we may run into the situation / need to introduce postgres specific code? (It would become more difficult if we generalize). 

Given that, I'm not conviced yet that this generalization is the right way.
What do you think? Any other opinions @oss-review-toolkit/core-devs  ?

model/src/main/kotlin/utils/FileProvenanceFileStorage.kt
```diff
@@ -42,7 +43,7 @@ class FileProvenanceFileStorage(
     private val storage: FileStorage,
 
     /**
-     * The filename of the files.
+     * The name of the file to use for storing the contents of the associated files.
```
fviernau, line 46: maybe: file(s)

fviernau (changes requested):

model/src/main/kotlin/utils/ProvenanceFileStorage.kt
sschuberth: Having Kotlin's `associateWith` in mind, the thing that comes after "with" should be the value, not the key. That is, I'd prefer *not* to say "Return the file associated with ..." as that would again swap the key / value meaning. To be aligned with Kotlin wording, I'll consider rewording to "associated by".

model/src/funTest/kotlin/utils/DatabaseProvenanceFileStorageFunTest.kt
sschuberth: `PostgresListener` *is* PostgreSQL-specific as it uses `EmbeddedPostgres`. Similar for `PostgresStorageConfiguration` / `PostgresConnection` as they use various PostgreSQL-specific properties (like `sslmode`).

> Also AFAIK these only have been tested with Postgres, so generalizing might open the way for untested territory.

Which is fine, IMO. People also already use `HttpFileStorage` with [untested WebDAV servers](https://github.com/oss-review-toolkit/ort/issues/6765), and I'm grateful for these reports. Documenting tested infrastructure via the class name is a bad idea anyway, IMO.

> Furthermore, do you think we may run into the situation / need to introduce postgres specific code?

We already have that, see the use of `org.postgresql.util.PGobject` in [JsonbSupport](https://github.com/oss-review-toolkit/ort/blob/main/scanner/src/main/kotlin/storages/utils/JsonbSupport.kt).

> What do you think?

I think it's a step into the right direction.

mnonnenmacher: In the server development we use [large objects](https://jdbc.postgresql.org/documentation/binary-data/) to store binary data because it is more memory efficient. This allows to stream the data directly from the database to the API endpoint without ever loading it into memory. As #7118 suggests to change the API to use streams as well, maybe we should first decide if we want to implement this with a Postgres specific feature before making the name more generic.

sschuberth: Just to be clear: I do have nothing against using PostgreSQL-specific features and name classes accordingly if they do. But I find it confusing when classes claim by name to be PostgreSQL-specific if they aren't.

sschuberth: Anyway, this simple name change is already causing more churn that I'm willing to handle. Let me drop the name change from this PR.

fviernau (approved):

PR #7127, opened 2023-06-14
feat(reporter): Implement the FossID snippet reporter

nnobelis:

Example report (with how-to-fix hints):
[AsciiDoc_fossid_snippets.zip](https://github.com/oss-review-toolkit/ort/files/11742794/AsciiDoc_fossid_snippets.zip)

### DIFF
--- a/cli/src/main/resources/logback.xml
+++ b/cli/src/main/resources/logback.xml
@@ -34,6 +34,7 @@
     <logger name="org.ossreviewtoolkit.analyzer.managers.Yarn2" level="INFO" />
     <logger name="org.ossreviewtoolkit.clients.fossid.FossIdRestService" level="INFO" />
     <logger name="org.ossreviewtoolkit.plugins.reporters.fossid.FossIdReporter" level="INFO" />
+    <logger name="org.ossreviewtoolkit.plugins.reporters.fossid.FossIdSnippetReporter" level="INFO" />
     <logger name="org.ossreviewtoolkit.scanner.scanners.fossid.FossId" level="INFO" />
     <logger name="org.ossreviewtoolkit.scanner.scanners.fossid.FossIdConfig" level="INFO" />
     <logger name="org.ossreviewtoolkit.scanner.scanners.fossid.FossIdUrlProvider" level="INFO" />
--- a/cli/src/main/resources/logback.xml
+++ b/cli/src/main/resources/logback.xml
@@ -33,7 +33,7 @@
     <logger name="org.eclipse.jgit.internal.storage.file.FileSnapshot" level="ERROR" />
     <logger name="org.ossreviewtoolkit.analyzer.managers.Yarn2" level="INFO" />
     <logger name="org.ossreviewtoolkit.clients.fossid.FossIdRestService" level="INFO" />
-    <logger name="org.ossreviewtoolkit.reporter.reporters.fossid.FossIdReporter" level="INFO" />
+    <logger name="org.ossreviewtoolkit.plugins.reporters.fossid.FossIdReporter" level="INFO" />
     <logger name="org.ossreviewtoolkit.scanner.scanners.fossid.FossId" level="INFO" />
     <logger name="org.ossreviewtoolkit.scanner.scanners.fossid.FossIdConfig" level="INFO" />
     <logger name="org.ossreviewtoolkit.scanner.scanners.fossid.FossIdUrlProvider" level="INFO" />
--- a/plugins/reporters/fossid/src/main/kotlin/FossIdSnippetReporter.kt
+++ b/plugins/reporters/fossid/src/main/kotlin/FossIdSnippetReporter.kt
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.plugins.reporters.fossid
+
+import java.io.File
+
+import org.apache.logging.log4j.kotlin.Logging
+
+import org.ossreviewtoolkit.plugins.reporters.asciidoc.HtmlTemplateReporter
+import org.ossreviewtoolkit.plugins.reporters.freemarker.FreemarkerTemplateProcessor
+import org.ossreviewtoolkit.reporter.Reporter
+import org.ossreviewtoolkit.reporter.ReporterInput
+
+class FossIdSnippetReporter : Reporter by delegateReporter {
+    companion object : Logging {
+        private const val TEMPLATE_NAME = "fossid_snippet"
+
+        val delegateReporter = HtmlTemplateReporter()
+    }
+
+    override val type = "FossIdSnippet"
+
+    override fun generateReport(input: ReporterInput, outputDir: File, options: Map<String, String>): List<File> {
+        val extendedOptions = options + (FreemarkerTemplateProcessor.OPTION_TEMPLATE_ID to TEMPLATE_NAME)
+        return delegateReporter.generateReport(input, outputDir, extendedOptions)
+    }
+}
--- a/plugins/reporters/fossid/src/main/resources/META-INF/services/org.ossreviewtoolkit.reporter.Reporter
+++ b/plugins/reporters/fossid/src/main/resources/META-INF/services/org.ossreviewtoolkit.reporter.Reporter
@@ -1,1 +1,2 @@
 org.ossreviewtoolkit.plugins.reporters.fossid.FossIdReporter
+org.ossreviewtoolkit.plugins.reporters.fossid.FossIdSnippetReporter
--- a/plugins/reporters/freemarker/src/main/kotlin/FreemarkerTemplateProcessor.kt
+++ b/plugins/reporters/freemarker/src/main/kotlin/FreemarkerTemplateProcessor.kt
@@ -39,6 +39,7 @@ import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.RuleViolation
 import org.ossreviewtoolkit.model.Severity
+import org.ossreviewtoolkit.model.SnippetFinding
 import org.ossreviewtoolkit.model.Vulnerability
 import org.ossreviewtoolkit.model.VulnerabilityReference
 import org.ossreviewtoolkit.model.config.RuleViolationResolution
@@ -309,6 +310,20 @@ class FreemarkerTemplateProcessor(
         @Suppress("UNUSED") // This function is used in the templates.
         fun filterForUnresolvedVulnerabilities(vulnerabilities: List<Vulnerability>): List<Vulnerability> =
             vulnerabilities.filterNot { input.resolutionProvider.isResolved(it) }
+
+        /**
+         * Return a list of [SnippetFinding]s grouped by the source file being matched by those snippets.
+         */
+        @Suppress("UNUSED") // This function is used in the templates.
+        fun groupSnippetsByFile(snippetFindings: Collection<SnippetFinding>): Map<String, List<SnippetFinding>> =
+            snippetFindings.groupBy { it.sourceLocation.path }
+
+        /**
+         * Collect all the licenses present in a collection of [SnippetFinding]s.
+         */
+        @Suppress("UNUSED") // This function is used in the templates.
+        fun collectLicenses(snippetFindings: Collection<SnippetFinding>): Set<String> =
+            snippetFindings.mapTo(mutableSetOf()) { it.snippet.licenses.toString() }
 
         /**
          * Return a flag indicating that issues have been encountered during the run of an advisor with the given
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -168,6 +168,7 @@ val copyrightExcludedPaths = listOf(
     "integrations/completions/",
     "plugins/reporters/asciidoc/src/main/resources/pdf-theme/pdf-theme.yml",
     "plugins/reporters/asciidoc/src/main/resources/templates/freemarker_implicit.ftl",
+    "plugins/reporters/fossid/src/main/resources/templates/freemarker_implicit.ftl",
     "plugins/reporters/freemarker/src/main/resources/templates/freemarker_implicit.ftl",
     "plugins/reporters/static-html/src/main/resources/prismjs/",
     "plugins/reporters/web-app-template/yarn.lock",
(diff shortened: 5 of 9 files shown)


plugins/reporters/fossid-snippets/src/main/kotlin/FossIdSnippetReporter.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.ossreviewtoolkit.plugins.reporters.freemarker.FreemarkerTemplateProcessor
+import org.ossreviewtoolkit.reporter.Reporter
+import org.ossreviewtoolkit.reporter.ReporterInput
+
+class FossIdSnippetReporter : Reporter {
+    companion object : Logging {
+
+        private const val TEMPLATE_NAME = "fossid_snippets"
+
+        val delegateReporter = HtmlTemplateReporter()
```
sschuberth, line 35: Should we consider opening `HtmlTemplateReporter` and use inheritance instead? Or Kotlin built-in class delegation?

plugins/reporters/fossid-snippets/src/main/resources/templates/asciidoc/fossid_snippets.ftl
```diff
@@ -0,0 +1,87 @@
...
+*${filePath}* +
+License(s):
+[#list licenses as license]
+  ${license}[#sep],
+[/#list]
+
+[width=100%]
+[cols="1,1,3,1,3,3,1,1"]
+|===
+| ID | Match | PUrl | License | File | URL | Score | Release Date
```
sschuberth, line 70: Instead of "PUrl", say "pURL" or "PURL".

plugins/reporters/freemarker/src/main/kotlin/FreemarkerTemplateProcessor.kt
```diff
@@ -310,6 +311,20 @@ class FreemarkerTemplateProcessor(
...
+        @Suppress("UNUSED") // This function is used in the templates.
+        fun groupSnippetsByFile(snippetFindings: Collection<SnippetFinding>): Map<String, List<SnippetFinding>> =
+            snippetFindings.groupBy { it.sourceLocation.path }
+
+        /**
+         * Collect all the licenses present in a collection of [SnippetFinding]s.
+         */
+        @Suppress("UNUSED") // This function is used in the templates.
+        fun collectLicenses(snippetFindings: Collection<SnippetFinding>) : Set<String> =
+            snippetFindings.map { it.snippet.licenses.toString() }.toSet()
```
sschuberth, line 331: Use `mapTo(mutableSetOf())`.

plugins/reporters/fossid-snippets/src/main/resources/templates/asciidoc/fossid_snippets.ftl
```diff
@@ -27,10 +27,10 @@
 
 = FossID Snippets
 List of all the packages with their files and snippets.
-[#list ortResult.scanResults as package, scanResults]
+[#assign filteredResults = helper.filterResultsByVCS(ortResult.scanResults)]
```
sschuberth, line 30: Commit message:
- "*are* two packages"
- I don't get "This is because FossID does not support VCS path and repository is scanned once"
- Please separate the last paragraph by an empty line.

sschuberth (changes requested):

plugins/reporters/freemarker/src/main/kotlin/FreemarkerTemplateProcessor.kt
```diff
@@ -325,6 +330,22 @@ class FreemarkerTemplateProcessor(
         fun collectLicenses(snippetFindings: Collection<SnippetFinding>) : Set<String> =
             snippetFindings.map { it.snippet.licenses.toString() }.toSet()
 
+        /**
+         * Filter the scan results to remove the ones having a VCS provenance without an empty path: Since FossID does
+         * not support support VCS path, they are most likely duplicates of other results.
+         */
+        @Suppress("UNUSED") // This function is used in the templates.
+        fun filterResultsByVCS(scanResults: Map<Identifier, List<ScanResult>>): Map<Identifier, List<ScanResult>> {
```
sschuberth, line 338: Shouldn't this filtering already happen in the FossID scanner implementation, so such results do not show up in the first place?

sschuberth:
> Example report (with how-to-fix hints):

Ugh, this contains results for files like `.git/hooks/applypatch-msg.sample`. Should we filter out snippet matches for VCS paths before adding them to the results? (Really weird that FossID doe snot seem to ignore these by itself.)

plugins/reporters/fossid-snippets/src/main/kotlin/FossIdSnippetReporter.kt
nnobelis: `Reporter` has only one function `generateReport`. So if we use Kotlin built-in class delegation, we would nevertheless override the function to be able to set the option, nullifying the benefits of the class delegation.
Shall we open the `HtmlTemplateReporter ` instead ?

plugins/reporters/fossid-snippets/src/main/kotlin/FossIdSnippetReporter.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.File
+
+import org.apache.logging.log4j.kotlin.Logging
+import org.ossreviewtoolkit.plugins.reporters.asciidoc.HtmlTemplateReporter
+import org.ossreviewtoolkit.plugins.reporters.freemarker.FreemarkerTemplateProcessor
+import org.ossreviewtoolkit.reporter.Reporter
+import org.ossreviewtoolkit.reporter.ReporterInput
+
+class FossIdSnippetReporter : Reporter {
```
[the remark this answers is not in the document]
nnobelis: Before I change this, I will wait for the answer to the comment above.

nnobelis:
> > Example report (with how-to-fix hints):
> 
> Ugh, this contains results for files like `.git/hooks/applypatch-msg.sample`. Should we filter out snippet matches for VCS paths before adding them to the results? (Really weird that FossID doe snot seem to ignore these by itself.)

Please don't be shocked by these results: when one triggers a scan with FossID, the scan gets automatically an ignore rule for `.git`. Here I generated the report for a scan made with latest version from FossID. It seems there is a (FossId) bug that the ignore rule is missing.

plugins/reporters/freemarker/src/main/kotlin/FreemarkerTemplateProcessor.kt
nnobelis: Hmm I will try to better explain: the project I am scanning contains two packages with the same provenance, differing in the VCS path.
The current scanner logic in ORT sees that these packages have the same provenance, and triggers a scan for one of them in FossID ("Which one" could be influenced thanks to https://github.com/oss-review-toolkit/ort/pull/7092).

Then, after the scan, there is a scanner logic that de duplicates the scan results for packages with the same provenance. Therefore the scan results contains the FossID results two times, one for each package.

Therefore there is no filtering to be done in FossID : the scanner scans already only one package.
I think the whole duplication behavior will be fixed when FossID will scan by provenance and that scan results are stored by provenance.

nnobelis:
> Some more UI-things:
> 
>     * Would be nice if this was more ORT-branded, similar to like done [here](https://github.com/oss-review-toolkit/ort/pull/6290#issuecomment-1367564130).
> 

Yes this would be nice but I noticed the work you line has been done for the `PdfTemplateReporter.kt` and I am not sure it applies to the HTML reporter (cf. `OPTION_PDF_THEME_FILE`). Would it be ok to add it in a future commit ?

>     * Instead of "Click here" to expand, can we say "Show How-to-Fix-Hint" or similar?
> 

Yes this is already the case in my development branch. This PR does not contain the how-to-fix block will be added later. The attached file was just an example to justify the need for HTML.

>     * Should we display the score as a percentage so it's more clear what the maximum value is?

Well I have not idea what the score unit is for FossID as it doesn't show it in the UI and it is undocumented.
And when I see in their documentation that they have for components such properties:
`"score": 373.25446,`
I would not make the assumption it's a percentage.

plugins/reporters/freemarker/src/main/kotlin/FreemarkerTemplateProcessor.kt
fviernau: > there is a scanner logic that de duplicates the scan results for packages with the same provenance

Did you mean "duplicates"  instead of "de duplicates"? (I believe duplicating is what it actually does)

> Therefore there is no filtering to be done in FossID : the scanner scans already only one package.

I don't understand how the amount of packages being scanned is related to filtering findings by path. 
To me these topics seem unrelated, maybe there is a mis-understanding here. 

I believe it'd be worth looking at the functions `ScanSummary.filterByPath()` and `ScanSummary.filterByIgnorePatterns()`. These do not adhere to snippet findings yet. I believe they need to be extended, which IIUC would address @sschuberth 's ask.

nnobelis: In the Analyzer result, there are two packages with nearly the same provenance: it only differs by `vcs_info/path`. 

When the scanner is called, one package is chosen (=reference package) and scanned with FossID. The other package is NOT scanned. I believe this logic scans the reference package:
https://github.com/oss-review-toolkit/ort/blob/40eb0f0b0ac4709adf0c1ecfbea2a2c1136f04c0/scanner/src/main/kotlin/Scanner.kt#L291-L297

Then after the scan, the `ScanSummary` of the reference package is duplicated to the other package.
At the end, the scan-results contains the two packages having each identical summaries.

The reporter cannot be called on this as the two duplicated summaries are the same. Therefore, this function filters the packages in the scan-results to keep only one.

fviernau: > The reporter cannot be called on this as the two duplicated summaries are the same. Therefore, this function filters the packages in the scan-results to keep only one.

Shouldn't each package have the snippet findings associated with it, no matter if there is another package which includes the same snippets? If both packages are built from same VCS, then adjusting these filterBy functions I mentioned earlier would take care of filtering out the snippets which are not relevant.

nnobelis: But I don't want to filter out individual snippet findings: I want to have in the report only ONE set of snippet findings.
Therefore, I need to do filtering at package level to keep only one package.

How is the webapp reporter dealing with duplicated license findings with packages having the same VCS ?

fviernau: > How is the webapp reporter dealing with duplicated license findings with packages having the same VCS ?

AFAIK, the WebApp / static HTML report do report the licenses per package. If two packages are built from same VCS, they still report per package.

plugins/reporters/fossid-snippets/src/main/kotlin/FossIdSnippetReporter.kt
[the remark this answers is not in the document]
sschuberth: FYI, I'd have a slight preference for singular "Snippet".

plugins/reporters/build.gradle.kts
```diff
@@ -33,6 +33,7 @@ dependencies {
     api(project(":plugins:reporters:cyclonedx-reporter"))
     api(project(":plugins:reporters:evaluated-model-reporter"))
     api(project(":plugins:reporters:fossid-reporter"))
+    api(project(":plugins:reporters:fossid-snippets-reporter"))
```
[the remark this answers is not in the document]
mnonnenmacher: Yes, I would also prefer to put all FossID reporters in the same project.

plugins/reporters/freemarker/src/main/kotlin/FreemarkerTemplateProcessor.kt
nnobelis: Following this discussion, I realized this change was controversial so I removed the related commit to not hinder the current PR. I will rework it and add it to a future separate PR.

plugins/reporters/fossid-snippets/src/main/kotlin/FossIdSnippetReporter.kt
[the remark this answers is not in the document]
sschuberth: > would like to leave the reporter name as `FossIdSnippets`

What do you mean by "leave"? The code above actually shows the original name as "FossIdSnippetReporter", i.e. it is already singular as requested.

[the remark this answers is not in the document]
nnobelis: Sorry I meant type, not name:
https://github.com/oss-review-toolkit/ort/blob/396f7418e1aea08a37e794800767ad1266708f85/plugins/reporters/fossid/src/main/kotlin/FossIdSnippetReporter.kt#L38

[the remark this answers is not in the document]
nnobelis: No it's not really expensive, more like a extra task...
I will change the `type`.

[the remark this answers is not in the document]
nnobelis: Done

cli/src/main/resources/logback.xml
```diff
@@ -34,6 +34,7 @@
     <logger name="org.ossreviewtoolkit.analyzer.managers.Yarn2" level="INFO" />
     <logger name="org.ossreviewtoolkit.clients.fossid.FossIdRestService" level="INFO" />
     <logger name="org.ossreviewtoolkit.reporter.reporters.fossid.FossIdReporter" level="INFO" />
+    <logger name="org.ossreviewtoolkit.plugins.reporters.fossid.FossIdSnippetReporter" level="INFO" />
```
sschuberth, line 37: While touching this, mind also fixing the line above in all `logback.xml` files in a preceding commit?

plugins/reporters/fossid/src/main/kotlin/FossIdSnippetReporter.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.logging.log4j.kotlin.Logging
+
+import org.ossreviewtoolkit.plugins.reporters.asciidoc.HtmlTemplateReporter
+import org.ossreviewtoolkit.plugins.reporters.freemarker.FreemarkerTemplateProcessor
+import org.ossreviewtoolkit.reporter.Reporter
+import org.ossreviewtoolkit.reporter.ReporterInput
+
+class FossIdSnippetReporter : Reporter by delegateReporter {
+    companion object : Logging {
+        private const val TEMPLATE_NAME = "fossid_snippets"
```
sschuberth, line 33: Should this really be plural?

plugins/reporters/fossid/src/main/resources/META-INF/services/org.ossreviewtoolkit.reporter.Reporter
```diff
@@ -1 +1,2 @@
 org.ossreviewtoolkit.plugins.reporters.fossid.FossIdReporter
+org.ossreviewtoolkit.plugins.reporters.fossid.FossIdSnippetReporter
```
sschuberth, line 2: Missing final newline.

sschuberth (changes requested):

cli/src/main/resources/logback.xml
sschuberth: > in all `logback.xml` files in a preceding commit?

Emphasis was on *all*, you missed `helper-cli/src/main/resources/logback.xml`.

sschuberth (commented):
LGTM now, but I'd prefer @MarcelBochtler or @mnonnenmacher to approve as they have more insights about the use of the report.

mnonnenmacher (approved):

mnonnenmacher:
@nnobelis I have rebase the PR to get rid of the failing tests which were fixed in another PR.

PR #7129, opened 2023-06-14
Improvements to archive handling and storages

sschuberth:

### DIFF
--- a/model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
+++ b/model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
@@ -19,12 +19,10 @@
 
 package org.ossreviewtoolkit.model.utils
 
-import io.kotest.core.TestConfiguration
 import io.kotest.core.spec.style.WordSpec
-import io.kotest.engine.spec.tempfile
 import io.kotest.matchers.shouldBe
 
-import java.io.File
+import java.io.InputStream
 
 import org.ossreviewtoolkit.model.ArtifactProvenance
 import org.ossreviewtoolkit.model.Hash
@@ -46,18 +44,6 @@ private val VCS_PROVENANCE = RepositoryProvenance(
     resolvedRevision = "0000000000000000000000000000000000000000"
 )
 
-private fun File.readTextAndDelete(): String {
-    val text = readText()
-    delete()
-
-    return text
-}
-
-private fun TestConfiguration.writeTempFile(content: String): File =
-    tempfile().apply {
-        writeText(content)
-    }
-
 class PostgresProvenanceFileStorageFunTest : WordSpec({
     val postgresListener = PostgresListener()
     lateinit var storage: PostgresProvenanceFileStorage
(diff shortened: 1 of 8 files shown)


sschuberth:
After this PR, the `ProvenanceFileStorage` and `FileStorage` interface look awfully similar. They should probably be combined via a generic type, and aligning the semantics for `getData` / `read` if the key does not exist (return a nullable type vs, throwing an exception).

utils/common/src/test/kotlin/ArchiveUtilsTest.kt
```diff
@@ -59,175 +59,39 @@ class ArchiveUtilsTest : WordSpec() {
...
-
-                val fileA = outputDir.resolve("a")
-                val fileB = outputDir.resolve("dir/b")
-
-                fileA shouldBe aFile()
-                fileA.readText() shouldBe "a\n"
-                fileB shouldBe aFile()
-                fileB.readText() shouldBe "b\n"
-            }
+        (enumValues<ArchiveType>().asList() - ArchiveType.DEB - ArchiveType.NONE).forEach { type ->
```
mnonnenmacher, line 62: What's the reason for the special handling of DEB archives?

utils/common/src/main/kotlin/ArchiveUtils.kt
```diff
@@ -114,6 +114,8 @@ fun File.unpackTryAllTypes(targetDirectory: File, filter: (ArchiveEntry) -> Bool
     }.find { archiveType ->
         runCatching {
             unpack(targetDirectory, forceArchiveType = archiveType, filter)
+        }.onSuccess {
+            ArchiveUtils.logger.debug { "Unpacking stream as $archiveType to '$targetDirectory'." }
```
mnonnenmacher, line 118: "Unpacked"?

model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
```diff
@@ -19,12 +19,10 @@
 
 package org.ossreviewtoolkit.model.utils
 
-import io.kotest.core.TestConfiguration
 import io.kotest.core.spec.style.WordSpec
-import io.kotest.engine.spec.tempfile
 import io.kotest.matchers.shouldBe
 
-import java.io.File
+import java.io.InputStream
```
mnonnenmacher, line 25: Please explain the benefits in the commit message.

scanner/src/main/kotlin/utils/FileListResolver.kt
```diff
@@ -47,23 +44,16 @@ internal class FileListResolver(
...
 }
 
 private fun ProvenanceFileStorage.putFileList(provenance: KnownProvenance, fileList: FileList) {
-    val tempFile = createOrtTempFile(prefix = "file-list", suffix = ".yml")
-
-    tempFile.writeValue(fileList)
-    putFile(provenance, tempFile)
-
-    tempFile.delete()
+    putData(provenance, fileList.toYaml().byteInputStream())
```
mnonnenmacher, line 51: Would be nice if we could avoid serializing the list to memory first and instead directly serialize to the stream. But IIRC Jackson can only serialize to an output stream and converting that to an input stream requires to use the `PipedInputStream` where I had problems the last time I tried to use it.

scanner/src/main/kotlin/utils/FileListResolver.kt
```diff
@@ -47,23 +44,16 @@ internal class FileListResolver(
...
 }
 
 private fun ProvenanceFileStorage.getFileList(provenance: KnownProvenance): FileList? {
-    val file = getFile(provenance) ?: return null
-
-    // Cannot rely on the extension of the file to reflect its type, so use YAML explicitly.
-    return FileFormat.YAML.mapper.readValue<FileList>(file).also { file.delete() }
+    val data = getData(provenance) ?: return null
+    return String(data.use { it.readBytes() }).fromYaml<FileList>()
```
mnonnenmacher, line 56: This should work to avoid reading to a string first:
```kotlin
data.use { yamlMapper.readValue<FileList>(it) }
```

mnonnenmacher (changes requested):

utils/common/src/test/kotlin/ArchiveUtilsTest.kt
sschuberth: See its test implementation. It's different form the other, and thus cannot be covered by the parameterized loop.

model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
sschuberth: The benefits are implied by the linked issue #7118, but I can briefly repeat them in the commit message.

mnonnenmacher (approved):

Issue #5018, opened 2022-02-03
Possibility to exclude paths / files within a project from a scan


MarcelBochtler:
Scenario:
I do have a repository with some huge files as test resources. Sometimes even binary files.
These result in ScanCode timeouts, when it tries to scans them.

The `--skip-excluded` CLI parameter currently only provides the possibility to exclude `Packages` and whole `Projects`.
See: https://github.com/oss-review-toolkit/ort/blob/14a22cba16320feda72c52640eea4229781500cb/scanner/src/main/kotlin/Scanner.kt#L96-L103

Possible solution:
Use the `--ignore` option from ScanCode to ignore the files specified via a `pathExclude` in `.ort.yml`.

mnonnenmacher:
To implement this correctly it is required that the exclude configuration is reflected in the stored scan result, to make sure that the returned scan result matches the excludes if they are changed. This might be already solved by the scanner configuration compatibility check in the scan storage, if the excludes are passed as an option to ScanCode, but this needs to be verified.

mnonnenmacher:
Open questions:
* Should this apply only to projects or to packages as well?
* Should this use path excludes or a new separate configuration?
* Should this be implemented scanner specific (e.g. the `--ignore` option for ScanCode) or in a generic way (e.g. deleting files before scanning)?

Bosch will make a concrete suggestion for an implementation and propose it to the community.

PR #7131, opened 2023-06-14
feat(scanner): Expose the Excludes to the scanner 

nnobelis:
In a future commit, the FossID scanner will use the `Excludes` contained in the configuration to generate ignore rules for the scan.

### DIFF
--- a/scanner/src/main/kotlin/ScanContext.kt
+++ b/scanner/src/main/kotlin/ScanContext.kt
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.scanner
 
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.PackageType
+import org.ossreviewtoolkit.model.config.Excludes
 
 /**
  * Additional context information that can be used by a [ScannerWrapper] to alter its behavior.
@@ -34,5 +35,10 @@ data class ScanContext(
     /**
      * The [type][PackageType] of the packages to scan.
      */
-    val packageType: PackageType
+    val packageType: PackageType,
+
+    /**
+     * The [Excludes] of the project to scan.
+     */
+    val excludes: Excludes? = null
 )
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -109,7 +109,14 @@ class Scanner(
 
             logger.info { "Scanning ${packages.size} project(s) with ${projectScannerWrappers.size} scanner(s)." }
 
-            scan(packages, ScanContext(ortResult.labels + labels, PackageType.PROJECT))
+            scan(
+                packages,
+                ScanContext(
+                    ortResult.labels + labels,
+                    PackageType.PROJECT,
+                    ortResult.repository.config.excludes
+                )
+            )
         } else {
             logger.info { "Skipping project scan as no project scanner is configured." }
 
@@ -122,7 +129,7 @@ class Scanner(
 
             logger.info { "Scanning ${packages.size} package(s) with ${packageScannerWrappers.size} scanner(s)." }
 
-            scan(packages, ScanContext(ortResult.labels, PackageType.PACKAGE))
+            scan(packages, ScanContext(ortResult.labels, PackageType.PACKAGE, ortResult.repository.config.excludes))
         } else {
             logger.info { "Skipping package scan as no package scanner is configured." }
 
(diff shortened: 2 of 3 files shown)


scanner/src/funTest/kotlin/scanners/AbstractPathScannerWrapperFunTest.kt
```diff
@@ -33,6 +33,7 @@ import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.PackageType
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.FileArchiverConfiguration
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
```
fviernau, line 36: I'm against making the scanner consume the excludes.
One reason is consistency, the other is that adjusting the excludes would invalidate the scan results. In order to apply the changed excludes you would need to re-scan. This is IMO an unnecessary waste of human time. 

If such findings would be flagged by a policy rule, then consumption of excludes would be consistent with license findings and copyright findings handling.

fviernau (dismissed):

scanner/src/funTest/kotlin/scanners/AbstractPathScannerWrapperFunTest.kt
nnobelis: Point taken.
@sschuberth : Care to share your opinion ? This was originally your idea to generate FossID exclusions rules from the ORT exclusion rules.

For some context, I must say this is something we badly need, to make each FossID scan more independent (the current logic of copying the rules from last completed scan being too error prone).

nnobelis: @fviernau This would also help for https://github.com/oss-review-toolkit/ort/issues/5018.

nnobelis:
@fviernau : Another idea could be to expose the `Excludes` only for `PackageScannerWrapper`s.

Thus, we can make clear that `PathScannerWrapper`s  are independent from excludes.

scanner/src/main/kotlin/ScanContext.kt
```diff
@@ -34,5 +35,10 @@ data class ScanContext(
     /**
      * The [type][PackageType] of the packages to scan.
      */
-    val packageType: PackageType
+    val packageType: PackageType,
+
+    /**
+     * The [RepositoryConfiguration] of the project to scan.
+     */
+    val repositoryConfiguration: RepositoryConfiguration
```
mnonnenmacher, line 43: The `RepositoryConfiguration` contains many unrelated properties, like the analyzer configuration, resolutions, or curations. Therefore it would be better to only pass the `Excludes`.

scanner/src/funTest/kotlin/scanners/AbstractPathScannerWrapperFunTest.kt
mnonnenmacher: As long as excludes are only used by scanners for which the scan results are not stored, this would not be a problem. This is currently the case for all implementations of `ScannerWrapper` that set `override val criteria: ScannerCriteria? = null`, like `FossId`.
One way to ensure that `ScannerWrapper`s with `criteria != null` do not have access to the excludes would be to filter the `ScanContext` before passing it to such scanners.

mnonnenmacher (changes requested):

scanner/src/funTest/kotlin/scanners/AbstractPathScannerWrapperFunTest.kt
fviernau: I like @mnonnenmacher  idea. The reason why I was afraid of consuming excludes was complexity, along with the need to make the scanner scale better some day. The more complicated we make things here, the more afraid developers will become to touch it, the less likely we will have better scaling in the scanner. Martin's proposal addresses this, because for scanners which do not store anything, there is no performance speed up from caching anyway.

nnobelis: Thanks a lot @mnonnenmacher  for your idea !

scanner/src/main/kotlin/ScanContext.kt
```diff
@@ -34,5 +35,10 @@ data class ScanContext(
     /**
      * The [type][PackageType] of the packages to scan.
      */
-    val packageType: PackageType
+    val packageType: PackageType,
+
+    /**
+     * The [Excludes] of the project to scan.
+     */
+    val excludes: Excludes? = null,
```
sschuberth, line 43: Please drop the trailing comma.

mnonnenmacher (approved):

PR #7134, opened 2023-06-14
Some code inspection fixes

fviernau:
See individual commits.

### DIFF
--- a/clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
+++ b/clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
@@ -29,7 +29,6 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.runBlocking
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.json.Json
 
 import okhttp3.MediaType.Companion.toMediaType
--- a/clients/clearly-defined/src/main/kotlin/Extensions.kt
+++ b/clients/clearly-defined/src/main/kotlin/Extensions.kt
@@ -19,8 +19,6 @@
 
 package org.ossreviewtoolkit.clients.clearlydefined
 
-import java.util.Locale
-
 import okhttp3.RequestBody
 
 import okio.Buffer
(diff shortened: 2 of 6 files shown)


clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
```diff
@@ -29,7 +29,6 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.runBlocking
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.decodeFromString
```
sschuberth, line 32: Hmm, interesting that these were not found by neither detekt nor Qodana. Any idea why that is, @fviernau?

clients/clearly-defined/src/main/kotlin/Extensions.kt
```diff
@@ -19,8 +19,6 @@
 
 package org.ossreviewtoolkit.clients.clearlydefined
 
-import java.util.Locale
```
sschuberth, line 22: Hint: It's last use was removed in 809c89d4e865e8fbd5e39b24efdaa34ad1f41c06.

plugins/package-managers/gradle-plugin/src/main/kotlin/GradleModelExtensions.kt
```diff
@@ -51,9 +50,3 @@ internal fun Configuration.isRelevant(): Boolean {
 
     return canBeResolved && !isDeprecatedConfiguration && !isDependenciesMetadata
 }
-
-/**
- * Return whether the Android plugin for Gradle has been applied to the project.
- */
-internal fun Project.isAndroidProject(): Boolean =
```
sschuberth, line 58: I originally wanted to keep this around as I was planning to take it into use again, but as I don't know when that will be I'm fine with removing it. I know how to restore the implementation in any case.

sschuberth (approved):

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
fviernau: I don't know it for sure. But this file uses a function named `decodeFromString()` but from a different package.

fviernau:
Merge despite failing unrelated test.

PR #7136, opened 2023-06-15
test(clearly-defined): Try to make a test less flaky

fviernau:
The tests fail quite frequently randomly on CI. Try to reduce this annoyance by adding a retry mechanism. Start with 3 retries with a 3 seconds interval.

### DIFF
--- a/plugins/package-curation-providers/clearly-defined/src/funTest/kotlin/ClearlyDefinedPackageCurationProviderFunTest.kt
+++ b/plugins/package-curation-providers/clearly-defined/src/funTest/kotlin/ClearlyDefinedPackageCurationProviderFunTest.kt
@@ -19,11 +19,14 @@
 
 package org.ossreviewtoolkit.plugins.packagecurationproviders.clearlydefined
 
+import io.kotest.assertions.retry
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.collections.beEmpty
 import io.kotest.matchers.collections.shouldHaveSingleElement
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldNot
+
+import kotlin.time.Duration.Companion.seconds
 
 import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService.Server
 import org.ossreviewtoolkit.model.Identifier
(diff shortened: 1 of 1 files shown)


plugins/package-curation-providers/clearly-defined/src/funTest/kotlin/ClearlyDefinedPackageCurationProviderFunTest.kt
```diff
@@ -91,21 +110,36 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
...
     }
 })
 
 private fun createPackagesFromIds(vararg ids: String) =
     ids.map { Package.EMPTY.copy(id = Identifier(it)) }
+
+@OptIn(ExperimentalTypeInference::class)
+@ExperimentalKotest
+private suspend fun <T> withRetry(@BuilderInference test: suspend () -> T): T =
+    eventually(
```
sschuberth, line 139: Any reason you're not using Kotest's [built-in `retry`](https://kotest.io/docs/assertions/retry.html), which seems to be a better fit?

fviernau: I wasn't aware of it. Don't know why you believe retry is a better fit. I've updated it, the code is slightly longer now.

sschuberth: > Don't know why you believe retry is a better fit.

Because `retry`'s docs say "Retry is similar to eventually, but rather than attempt a block of code for a period of time, it attempts a block of code a maximum number of times", which IMO is exactly what we want.

sschuberth: > I've updated it, the code is slightly longer now.

Hmm, it actually looks shorter to me...

fviernau: Sure, you're right.

plugins/package-curation-providers/clearly-defined/src/funTest/kotlin/ClearlyDefinedPackageCurationProviderFunTest.kt
```diff
@@ -37,26 +40,33 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
         "return an existing curation for the javax.servlet-api Maven package" {
             val packages = createPackagesFromIds("Maven:javax.servlet:javax.servlet-api:3.1.0")
 
-            val curations = provider.getCurationsFor(packages)
+            withRetry {
```
sschuberth, line 43: I'd prefer to only wrap the actual `getCurationsFor` call in `withRetry`, like

    val curations = withRetry { provider.getCurationsFor(packages) }

plugins/package-curation-providers/clearly-defined/src/funTest/kotlin/ClearlyDefinedPackageCurationProviderFunTest.kt
```diff
@@ -91,21 +105,33 @@ class ClearlyDefinedPackageCurationProviderFunTest : WordSpec({
...
 })
 
 private fun createPackagesFromIds(vararg ids: String) =
     ids.map { Package.EMPTY.copy(id = Identifier(it)) }
+
+private suspend fun <T> withRetry(f: suspend () -> T): T =
+    retry(
+        maxRetry = 3,
+        timeout = 9.seconds,
+        delay = 3.seconds,
```
sschuberth, line 135: I wonder whether instead of a fixed delay we should rather make use of `multiplier = 5` (or so; and increase the `timeout` accordingly), because *if* there is a continuous problem, it probably does not disappear after a too short time.

sschuberth (changes requested):

plugins/package-curation-providers/clearly-defined/src/funTest/kotlin/ClearlyDefinedPackageCurationProviderFunTest.kt
fviernau: I would like to try this one out first. The first two runs worked. We can change it later as you propose if it turns out not to work. A believe as-is it's faster.

plugins/package-curation-providers/clearly-defined/src/funTest/kotlin/ClearlyDefinedPackageCurationProviderFunTest.kt
fviernau: IIUC it doesn't work, because `provider.getCurationsFor(packages)` sometimes returns empty set unexpectedly.

sschuberth: You're right.

sschuberth (approved):

plugins/package-curation-providers/clearly-defined/src/funTest/kotlin/ClearlyDefinedPackageCurationProviderFunTest.kt
sschuberth: We still get the timeouts: https://github.com/oss-review-toolkit/ort/actions/runs/5308587525/jobs/9608290182#step:5:825

So I'll prepare a PR that changes to using a multiplier.

sschuberth: See https://github.com/oss-review-toolkit/ort/pull/7149.

PR #7142, opened 2023-06-16
docs(model): Add ScanCode `commandLineNonConfig` to reference.yml

mnonnenmacher:
Add the missing `commandLineNonConfig` option for ScanCode to the reference.yml and explain the difference to the `commandLine` option.

### DIFF
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -193,7 +193,13 @@ ort:
     options:
       # A map of maps from scanner class names to scanner-specific key-value pairs.
       ScanCode:
+        # Command line options that affect the ScanCode output. If changed, stored scan results that were created with
+        # different options are not reused.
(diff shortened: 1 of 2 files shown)


model/src/main/resources/reference.yml
```diff
@@ -193,7 +193,11 @@ ort:
     options:
       # A map of maps from scanner class names to scanner-specific key-value pairs.
       ScanCode:
+        # Command line options that affect the ScanCode output. If changed, stored scan results that were created with
+        # different options are not reused.
         commandLine: '--copyright --license --info --strip-root --timeout 300'
+        # Command line options that do not affect the ScanCode output.
```
sschuberth, line 199: Nit: Should blank lines for grouping be added to make more clear that the comment does not refer to `parseLicenseExpressions` also?

sschuberth (approved):

PR #7145, opened 2023-06-16
build(Gradle): Extend the ANTLR work-around to all source sets

sschuberth:
This fixup for 6a73769 addresses the error [1]:

    Reason: Task ':utils:spdx-utils:compileTestKotlin' uses this output of
    task ':utils:spdx-utils:generateTestGrammarSource' without declaring
    an explicit or implicit dependency. This can lead to incorrect results
    being produced, depending on what order the tasks are executed.

[1]: https://github.com/oss-review-toolkit/ort/actions/runs/5288598268/jobs/9570557290#step:4:5223

### DIFF
--- a/utils/spdx/build.gradle.kts
+++ b/utils/spdx/build.gradle.kts
@@ -61,8 +61,10 @@ tasks.withType<AntlrTask>().configureEach {
     }
 }
 
-private val generateGrammarSource by tasks.existing
-sourceSets.main.get().java.srcDir(generateGrammarSource.map { files() })
+sourceSets.configureEach {
+    val generateGrammarSource by tasks.existing
+    sourceSets.main.get().java.srcDir(generateGrammarSource.map { files() })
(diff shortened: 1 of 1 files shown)


utils/spdx/build.gradle.kts
```diff
@@ -61,8 +61,10 @@ tasks.withType<AntlrTask>().configureEach {
     }
 }
 
-private val generateGrammarSource by tasks.existing
-sourceSets.main.get().java.srcDir(generateGrammarSource.map { files() })
+sourceSets.configureEach {
```
fviernau, line 64: commit message: How long are these job logs kept around? I wonder if it is better to inline the needed information into the commit message?!

fviernau (changes requested):

mnonnenmacher (approved):

fviernau (approved):

PR #7151, opened 2023-06-19
feat(fossid-webapp): Generate ignore rules from path excludes 

nnobelis:
Currently, the ignore rules are carried over from the last successful delta
scan. Since this process can lead to loss of rules when a new branch is
scanned, and to promote "configuration as code", this commit changes the
logic to create the ignore rules from the path excludes present in the
.ort.yml file.
ORT excludes path patterns is based on [1] and FossID supports only a
subset of it. Hence, this commit adds the mapping of this subset to the
FossID scanner. If a path exclude cannot be mapped to an ignore rule, an
issue is added to the scan summary.

Since it is not expected for the user to have already an .ort.yml
configured with path excludes, the rules of the last successful delta scan
are still carried over but, for each rule not existing as a path exclude,
an issue is also added to the scan summary.

This handling of "legacy ignore rules" will be removed in the future, after
a grace period allowing migration.

[1]: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/AntPathMatcher.html

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -303,15 +303,17 @@ class FossId internal constructor(
                         .checkResponse("list scans for project").data
                     checkNotNull(scans)
 
+                    val issues = mutableListOf<Issue>()
+
                     val (scanCode, scanId) = if (config.deltaScans) {
-                        checkAndCreateDeltaScan(scans, url, revision, projectCode, projectName, context)
+                        checkAndCreateDeltaScan(scans, url, revision, projectCode, projectName, context, issues)
                     } else {
                         checkAndCreateScan(scans, url, revision, projectCode, projectName)
                     }
 
                     if (config.waitForResult) {
                         val rawResults = getRawResults(scanCode)
-                        createResultSummary(startTime, provenance, rawResults, scanCode, scanId)
+                        createResultSummary(startTime, provenance, rawResults, scanCode, scanId, issues)
                     } else {
                         val issue = createAndLogIssue(
                             source = name,
@@ -468,7 +470,8 @@ class FossId internal constructor(
         revision: String,
         projectCode: String,
         projectName: String,
-        context: ScanContext
+        context: ScanContext,
+        issues: MutableList<Issue>
     ): Pair<String, String> {
         val projectRevision = context.labels[PROJECT_REVISION_LABEL]
 
@@ -535,19 +538,35 @@ class FossId internal constructor(
 
             logger.info { "Loading ignore rules from '$existingScanCode'." }
 
+            // TODO: This is the old way of carrying the rules to the new delta scan, by querying the previous scan.
+            //       With the introduction of support for the ORT excludes, this old behavior can be dropped.
             val ignoreRules = service.listIgnoreRules(config.user, config.apiKey, existingScanCode)
                 .checkResponse("list ignore rules")
             ignoreRules.data?.let { rules ->
                 logger.info { "${rules.size} ignore rule(s) have been found." }
 
+                val excludesRules = context.excludes?.let {
+                    convertRules(it, issues).also {
+                        logger.info { "${it.size} rules from ORT excludes have been found." }
+                    }
+                }.orEmpty()
+
+                // Create an issue for each legacy rule existing.
+                val legacyRules = excludesRules.filterLegacyRules(rules, issues)
+                if (legacyRules.isNotEmpty()) {
+                    logger.warn { "${legacyRules.size} legacy rules have been found." }
+                }
+
+                val allRules = excludesRules + legacyRules
+
                 // When a scan is created with the optional property 'git_repo_url', the server automatically creates
                 // an 'ignore rule' to exclude the '.git' directory.
                 // Therefore, this rule will be created automatically and does not need to be carried from the old scan.
-                rules.filterNot { it.type == RuleType.DIRECTORY && it.value == ".git" }.forEach {
+                allRules.filterNot { it.type == RuleType.DIRECTORY && it.value == ".git" }.forEach {
                     service.createIgnoreRule(config.user, config.apiKey, scanCode, it.type, it.value, RuleScope.SCAN)
                         .checkResponse("create ignore rules", false)
                     logger.info {
-                        "Ignore rule of type '${it.type}' and value '${it.value}' has been carried to the new scan."
+                        "Ignore rule of type '${it.type}' and value '${it.value}' has been created for the new scan."
                     }
                 }
             }
@@ -820,7 +839,8 @@ class FossId internal constructor(
         provenance: Provenance,
         rawResults: RawResults,
         scanCode: String,
-        scanId: String
+        scanId: String,
+        additionalIssues: MutableList<Issue>
     ): ScanResult {
         // TODO: Maybe get issues from FossID (see has_failed_scan_files, get_failed_files and maybe get_scan_log).
 
@@ -844,7 +864,7 @@ class FossId internal constructor(
             licenseFindings = licenseFindings,
             copyrightFindings = copyrightFindings,
             snippetFindings = snippetFindings,
-            issues = issues
+            issues = issues + additionalIssues
         )
 
         return ScanResult(
--- a/scanner/src/main/kotlin/scanners/fossid/FossId.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossId.kt
@@ -796,7 +796,8 @@ class FossId internal constructor(
                         }
                     }
 
-                    it to snippets.filterNot { it.matchType == MatchType.IGNORED }.toSet()
+                    val excludedMatchTypes = enumSetOf(MatchType.IGNORED, MatchType.NONE)
+                    it to snippets.filterNotTo(mutableSetOf()) { it.matchType in excludedMatchTypes }
                 }
             }.awaitAll().toMap()
         }
--- a/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
@@ -135,7 +135,7 @@ internal fun mapSnippetFindings(
             // FossID does not return the hash of the remote artifact. Instead, it returns the MD5 hash of the
             // matched file in the remote artifact as part of the "match_file_id" property.
             val url = checkNotNull(snippet.url) {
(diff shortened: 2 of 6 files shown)


clients/fossid-webapp/src/main/kotlin/model/rules/IgnoreRule.kt
```diff
@@ -50,4 +50,8 @@ data class IgnoreRule(
      * The date the rule was last updated (e.g. 2022-01-05 14:59:22).
      */
     val updated: String
-)
+
+) {
+    override fun equals(other: Any?): Boolean =
+        (other === this) || (other is IgnoreRule && other.type == type && other.value == value)
```
sschuberth, line 56: Overriding `equals` this way without also overriding `hashCode` might cause all kinds of nasty issues. Can this be replaced by a custom function with a telling name about which properties are being used / ignored?

sschuberth: What about using `distinctBy {}`?

nnobelis: How would you use it in the `filterLegacyRules` function ?
Phrasing it differently, what the intersection of the two lists would bring me ?

nnobelis: Maybe I should use `any` and do the comparison myself.

sschuberth: Right, `any` makes more sense, probably like

    rulesToTest.filterNot { ruleToTest ->
        any { it.type == ruleToTest.type && it.value == ruleToTest.value }
    }

scanner/src/main/kotlin/scanners/fossid/FossIdScanResults.kt
```diff
@@ -135,7 +135,7 @@ internal fun mapSnippetFindings(
             // FossID does not return the hash of the remote artifact. Instead, it returns the MD5 hash of the
             // matched file in the remote artifact as part of the "match_file_id" property.
             val url = checkNotNull(snippet.url) {
-                "The URL of snippet ${snippet.id} must not be null."
+                "The URL of snippet ${snippet.id} for file '$file' must not be null."
```
sschuberth, line 138: Commit message nit: This is not an error log, but an exception message.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -796,7 +796,7 @@ class FossId internal constructor(
                         }
                     }
 
-                    it to snippets.filterNot { it.matchType == MatchType.IGNORED }.toSet()
+                    it to snippets.filterNot { it.matchType in listOf(MatchType.IGNORED, MatchType.NONE) }.toSet()
```
sschuberth, line 818: Please use `enumSetOf()` and extract it to a variable outside of `filterNot` to not recreate it on each iteration.

sschuberth: While at it, you could also do `filterNotTo(mutableSetOf())` instead of `toSet()`.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -535,19 +538,35 @@ class FossId internal constructor(
 
             logger.info { "Loading ignore rules from '$existingScanCode'." }
 
+            // TODO: This is the old way of carrying the rules to the new delta scan, by querying the previous scan.
+            //  With the introduction of support for the ORT excludes, this old behavior can be dropped.
```
sschuberth, line 542: Wring indentation: Align with "This" from the previous line.

scanner/src/main/kotlin/scanners/fossid/FossId.kt
```diff
@@ -535,19 +538,35 @@ class FossId internal constructor(
...
+                val excludesRules = context.excludes?.let {
+                    convertRules(it, issues).also {
+                        logger.info { "${it.size} rules from ORT excludes have been found." }
+                    }
+                }.orEmpty()
+
+                // Create an issue for each legacy rule existing.
+                val legacyRules = excludesRules.filterLegacyRules(rules, issues)
+                if (legacyRules.isNotEmpty()) {
+                    logger.warn("${legacyRules.size} legacy rules have been found.")
```
sschuberth, line 557: Use `{}` syntax.

scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -0,0 +1,98 @@
...
+                FossId.logger.warn {
+                    "Path exclude  '${it.pattern}' cannot be converted to an ignore rule."
+                }
+            }
+        }
+    }
+}
+
+private fun PathExclude.mapToRule(): IgnoreRule? {
+    val match1 = DIRECTORY_PATTERN.matchEntire(pattern)
```
sschuberth, line 56: Probably a better name: `directoryMatch`

scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -0,0 +1,98 @@
...
+        val starStar = it.groupValues[2]
+
+        return if (starStar.isEmpty()) {
+            IgnoreRule(-1, RuleType.DIRECTORY, directory, -1, "")
+        } else {
+            IgnoreRule(-1, RuleType.DIRECTORY, "$directory/**", -1, "")
+        }
+    }
+
+    val match2 = EXTENSION_PATTERN.matchEntire(pattern)
```
sschuberth, line 69: Probably a better name: `extensionMatch`

scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -0,0 +1,98 @@
...
+import org.ossreviewtoolkit.clients.fossid.model.rules.IgnoreRule
+import org.ossreviewtoolkit.clients.fossid.model.rules.RuleType
+import org.ossreviewtoolkit.model.Issue
+import org.ossreviewtoolkit.model.Severity
+import org.ossreviewtoolkit.model.config.Excludes
+import org.ossreviewtoolkit.model.config.PathExclude
+
+private val DIRECTORY_PATTERN = "((?:\\w+/)*\\w+)/?(\\*\\*)?".toRegex()
+private val EXTENSION_PATTERN = "\\*\\.(\\w+)".toRegex()
+private val FILE_PATTERN = "([\\w.]+)".toRegex()
```
sschuberth, line 31: Consider using named capturing groups for readability.

scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    val match2 = EXTENSION_PATTERN.matchEntire(pattern)
+    match2?.let {
+        val extension = it.groupValues[1]
+        return IgnoreRule(-1, RuleType.EXTENSION, ".$extension", -1, "")
+    }
+
+    val match3 = FILE_PATTERN.matchEntire(pattern)
```
sschuberth, line 75: Probably a better name: `fileMatch`

scanner/src/main/kotlin/scanners/fossid/Utils.kt
sschuberth: Also: These are not patterns / strings, but compiled regexes, so the variable names should end in `_REGEX`.

sschuberth (changes requested):

scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+    }
+}
+
+private fun PathExclude.mapToRule(): IgnoreRule? {
+    val directoryMatch = DIRECTORY_REGEX.matchEntire(pattern)
+    directoryMatch?.let {
+        val directory = it.groups["directory"]!!.value
+        val starStar = it.groups["starstar"]?.value
```
sschuberth, line 59: Could now be written as

    DIRECTORY_REGEX.matchEntire(pattern)?.let { directoryMatch ->
        val directory = directoryMatch.groups["directory"]!!.value
        val starStar = directoryMatch.groups["starstar"]?.value

Similar below.

sschuberth (changes requested):

sschuberth (changes requested):
One last nit from my side.

scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -0,0 +1,94 @@
...
+import org.ossreviewtoolkit.model.Severity
+import org.ossreviewtoolkit.model.config.Excludes
+import org.ossreviewtoolkit.model.config.PathExclude
+
+private val DIRECTORY_REGEX = "(?<directory>(?:\\w+/)*\\w+)/?(?<starstar>\\*\\*)?".toRegex()
+private val EXTENSION_REGEX = "\\*\\.(?<extension>\\w+)".toRegex()
+private val FILE_REGEX = "(?<file>[\\w.]+)".toRegex()
+
+/**
+ * Convert an ORT path excludes in [excludes] to FossID [IgnoreRule]. If an error is encountered during the mapping, an
```
mnonnenmacher, line 34: "Convert the ORT [path excludes][Excludes.paths] in [excludes] to FossID [IgnoreRule]s."

scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    return null
+}
+
+/**
+ * Check if some elements of [rulesToTest] are legacy rules i.e. are not present in a reference list (current object).
+ * Create an issue on [issues] for each legacy rule and return a list of the latter.
+ */
+fun List<IgnoreRule>.filterLegacyRules(rulesToTest: List<IgnoreRule>, issues: MutableList<Issue>): List<IgnoreRule> =
```
mnonnenmacher, line 84: I think this function could be internal.

mnonnenmacher (changes requested):

scanner/src/main/kotlin/scanners/fossid/Utils.kt
nnobelis: I changed the visibility of `convertRules` too.

mnonnenmacher (approved):

PR #7157, opened 2023-06-20
test(downloader): Make `GitWorkingTreeFunTest` less flaky

fviernau:
Make the test resilient with regard to newly added remote branches by
only asserting that all expected branches are amongst the returned
branches.


sschuberth:
> Make the test resilient with regard to newly added remote branches by only asserting that all expected braanches are amongst the returned tags.

Typo in "braanches". Should say "returned branches" instead of "tags".

Issue #6944, opened 2023-04-28
OrtResult: Add a list of file paths and sha1 sum for all provenances


fviernau:
Pre-work:

1. As of #6943 a listing of files is stored by provenance. 
2. As of #5950 sub-module structure is available in `OrtResult` and scan results are stored "by provenance".

Task: Add file lists with a SHA1 sums to `OrtResult` for all provenances contained in the `OrtResult` as enabler for:
- #6947
- #6948

PR #7162, opened 2023-06-20
Minor fixes and preparations for adding file lists to `OrtResult`

fviernau:
See individual commits.

Part of #6944.

### DIFF
--- a/model/src/main/kotlin/utils/SortedSetConverters.kt
+++ b/model/src/main/kotlin/utils/SortedSetConverters.kt
@@ -31,6 +31,7 @@ import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageReference
 import org.ossreviewtoolkit.model.Project
+import org.ossreviewtoolkit.model.Provenance
 import org.ossreviewtoolkit.model.ProvenanceResolutionResult
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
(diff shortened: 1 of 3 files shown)


model/src/main/kotlin/utils/SortedSetConverters.kt
```diff
@@ -77,7 +77,7 @@ class ScanResultSortedSetConverter : StdConverter<Set<ScanResult>, Set<ScanResul
                     this += it.provenance.sourceArtifact.url
                 }
                 else -> {
-                    // Cannot happen as ScanResults in ScannerRun
+                    // Do not append anything, because unknown provenance does not have any properties.
```
sschuberth: Nit: "append" is the verb used for `buildString`, but for `buildList` it'd be "add".

model/src/main/kotlin/utils/SortedSetConverters.kt
```diff
@@ -93,4 +77,22 @@ class SnippetFindingSortedSetConverter : StdConverter<Set<SnippetFinding>, Sorte
         value.toSortedSet(compareBy<SnippetFinding> { it.sourceLocation.path }.thenByDescending { it.snippet.purl })
 }
 
+private fun Provenance.sortKey(): String =
```
sschuberth, line 80: The "sort" verb as part of the function name sounds as if the function itself would already sort. Maybe name this "toKey()", "getSortKey()", "getSelector()" or similar?

sschuberth (changes requested):
Generally, it's a bit hard to judge whether the refactorings make sense when the announced upcoming usages are not in this PR.

fviernau:
> Generally, it's a bit hard to judge whether the refactorings make sense when the announced upcoming usages are not in this PR.

Do you want me to close this PR?

fviernau:
> Not necessarily, I just wanted to point this out as a reviewer. Here, the refactorings are probably uncritical (though it seems a bit over the top that we have three variants of `prependPath` now), but for bigger code extractions for reuse it'd be nice to know what the reuse is. Because if the review turns out that the reuse does not make sense / is not needed (for whatever reason), the refactoring doesn't either, so it's easier to review both in the same context.

The planned use of the new `prependPath` is: https://github.com/oss-review-toolkit/ort/pull/7164/commits/077b5948a664de0e4f748dd7cfeb7154ec782e59#diff-c530c7a3069096e7ad7b11deffe116249ef3cc3938579dbdf08d9f22e7b030f3R290

sschuberth (approved):

PR #7163, opened 2023-06-20
Minor `ScanController` improvements

fviernau:
See individual commits.

This is follow-up to the implementation of #5950.

### DIFF
--- a/scanner/src/main/kotlin/ScanController.kt
+++ b/scanner/src/main/kotlin/ScanController.kt
@@ -53,14 +53,14 @@ class ScanController(
     val config: ScannerConfiguration
 ) {
     /**
-     * A map of package [Identifier]s to a list of [Issue]s that occurred during provenance resolution for the
-     * respective package.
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/ScanController.kt
```diff
@@ -62,7 +62,7 @@ class ScanController(
      * A map of package [KnownProvenance]s to an [Issue] that occurred during provenance resolution for the respective
      * package.
      */
-    private val nestedProvenanceResolutionIssues = mutableMapOf<KnownProvenance, Issue>()
+    private val nestedProvenanceResolutionIssues = mutableMapOf<Identifier, Issue>()
```
sschuberth, line 65: Commit message: I don't get the "it creates the resulting `Issue`s if any
specific to package(s) / IDs" part. Are maybe commas around "if any" missing?

fviernau: I've tried to further simplify the commit message a bit.

PR #7164, opened 2023-06-21
OrtResult: Include file lists for all known provenances in `ScannerRun`

fviernau:
The file lists are an enabler for

1. Computing SPDX package verification codes on-the-fly.
2. Adding file findings to SPDX reports
3. Several `helper-cli` automation opportunities like e.g. `PathExcludeGenerator`
 
Fixes: #6944.

**Breaking change:** `OrtResult` de-serialization is not backwards compatible.

### DIFF
--- a/cli/src/funTest/assets/semver4j-ort-result.yml
+++ b/cli/src/funTest/assets/semver4j-ort-result.yml
@@ -318,6 +318,7 @@ scanner:
           path: "LICENSE.md"
           start_line: 3
           end_line: 3
+  files: []
 advisor:
   start_time: "2021-04-29T14:54:16.562951Z"
   end_time: "2021-04-29T14:54:18.969210Z"
--- a/cli/src/funTest/assets/semver4j-ort-result.yml
+++ b/cli/src/funTest/assets/semver4j-ort-result.yml
@@ -318,7 +318,39 @@ scanner:
           path: "LICENSE.md"
           start_line: 3
           end_line: 3
-  files: []
+  files:
+  - provenance:
+      vcs_info:
+        type: "Git"
+        url: "https://github.com/junit-team/junit.git"
+        revision: "64155f8a9babcfcf4263cf4d08253a1556e75481"
+        path: ""
+      resolved_revision: "64155f8a9babcfcf4263cf4d08253a1556e75481"
+    files:
+    - path: "LICENSE-junit.txt"
+      sha1: "1dc94a80dbba8262f732e0b0d7235149861a1b57"
+    - path: "epl-v10.html"
+      sha1: "49f3f46520878abad864392fddf0da834c041bf8"
+    - path: "pom.xml"
+      sha1: "442292b8a7efeabbe4cc176709b833b1792140ec"
+    - path: "src/site/fml/faq.fml"
+      sha1: "220d7662a1fd7597e320e6e61589127349f9e5b0"
+    - path: "src/site/resources/css/hopscotch-0.1.2.min.css"
+      sha1: "ac230b8be930f5ee9b30a59562eda4cd27452b7e"
+    - path: "src/site/resources/scripts/hopscotch-0.1.2.min.js"
+      sha1: "8d68f8ccf1ae22b306e7cc0890df978e6c8a9abb"
+  - provenance:
+      vcs_info:
+        type: "Git"
+        url: "https://github.com/vdurmont/semver4j.git"
+        revision: "7653e418d610ffcd2811bcb55fd72d00d420950b"
+        path: ""
+      resolved_revision: "7653e418d610ffcd2811bcb55fd72d00d420950b"
+    files:
+    - path: "LICENSE.md"
+      sha1: "4a8486805915245bbbe2e3c8c1a9d680c6843a82"
+    - path: "pom.xml"
+      sha1: "442292b8a7efeabbe4cc176709b833b1792140ec"
 advisor:
   start_time: "2021-04-29T14:54:16.562951Z"
   end_time: "2021-04-29T14:54:18.969210Z"
--- a/model/src/main/kotlin/FileList.kt
+++ b/model/src/main/kotlin/FileList.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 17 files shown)


model/src/main/kotlin/FileList.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+private val HEX_SYMBOLS_REGEX = "^[0-9a-f]+$".toRegex()
+private const val SHA1_SUM_NUM_DIGITS = 40
+
+private fun String.checkSha1Sum() {
+    require(matches(HEX_SYMBOLS_REGEX)) {
+        "The sha1 value must only contain lower case hexadecimal symbols."
+    }
```
sschuberth, line 76: Just wondering: Do we need such a strict check at all? What circumstance should it guard against? Is maybe a simple not-blank / length check sufficient?

model/src/main/kotlin/ScannerRun.kt
```diff
@@ -207,11 +221,31 @@ data class ScannerRun(
...
+        val packageProvenance = resolutionResult.packageProvenance!!
+
+        val fileListsByPath = resolutionResult.getKnownProvenancesWithoutVcsPath().mapValues { (_, provenance) ->
+            fileListByProvenance.getValue(provenance)
+        }
+
+        return mergeFileLists(fileListsByPath)
+            .filterByVcsPath(packageProvenance.vcsPath)
+            .copy(provenance = packageProvenance)
+    }
```
sschuberth, line 238: Missing empty line after this one.

sschuberth (commented):
I only skimmed over the changes for now.

model/src/main/kotlin/FileList.kt
fviernau: Agreed, that was too strict.

model/src/main/kotlin/ScannerRun.kt
```diff
@@ -202,3 +265,36 @@ private fun scanResultForProvenanceResolutionIssue(packageProvenance: KnownProve
 
 private fun ScanSummary.addIssue(issue: Issue?): ScanSummary =
     if (issue == null) this else copy(issues = (issues + issue).distinct())
+
+private fun mergeFileLists(fileListByPath: Map<String, FileList>): FileList {
+    require(fileListByPath.containsKey("")) {
```
sschuberth, line 270: For maps, `containsKey` is the same as `contains`, and so the `in` operator can be used here: `"" in fileListByPath`.

sschuberth: Also, what's the semantics of `""` as a key? Is it the root path? Does it maybe make sense to define something like `ROOT_PATH_KEY = ""` for this for clarity?

model/src/main/kotlin/ScannerRun.kt
```diff
@@ -202,3 +265,36 @@ private fun scanResultForProvenanceResolutionIssue(packageProvenance: KnownProve
...
+        "There must be a file list associated with the empty path."
+    }
+
+    val provenance = fileListByPath.getValue("").provenance
```
sschuberth, line 274: So this could actually be combined with the above requirement:

    val provenance = requireNotNull(fileListByPath[""]?.provenance) { ... }

model/src/main/kotlin/FileList.kt
```diff
@@ -0,0 +1,68 @@
...
+        val path: String,
+
+        /**
+         * The sha1 checksum of the file, consisting of 40 lowercase hexadecimal digits.
+         */
+        val sha1: String
+    ) {
+        init {
+            require(path.isNotBlank()) { "The path must not be blank." }
+
```
sschuberth, line 56: Nit: Maybe drop the empty line between these similar one-liners.

model/src/main/kotlin/ScannerRun.kt
fviernau: > Also, what's the semantics of `""` as a key? Is it the root path? Does it maybe make sense to define something like `ROOT_PATH_KEY = ""` for this for clarity?

I believe that the error message for the require block clarifies that. Agree?

sschuberth: The require block still just says "There must be a file list associated with the empty path", which doe snot explain what it *means* to associate a file list with the empty path. Maybe instead: "There must be a file list associated with the root path" (if that's the correct semantics)?

cli/src/funTest/assets/semver4j-ort-result.yml
```diff
@@ -318,6 +318,7 @@ scanner:
           path: "LICENSE.md"
           start_line: 3
           end_line: 3
+  files: []
```
sschuberth, line 321: Maybe a stupid question, but why is this list empty? Given that we have findings above, shouldn't this list at least contain the files with findings? Similar in the other result files.

fviernau: This commits only adds the model classes, but the scanner outputs only empty file lists still. This is the next. The next next commit changes the scanner to add actual data. 

In that commit I added only the file lists needed for testing them, but I haven't added them to all assets, as these aren't used by the tests anyway. As adding could be quite a bit of effort, I thought it's better to do this only when need arises, and focus only on the test asset use for testing these file lists.

sschuberth: I get your point that it's effort, but I'm unsure whether it's ok to have test results with inconsistent / incomplete content... could your new helper-cli command to rewrite test assets become handy here maybe?

fviernau: Do you agree that adding the data would go into the topmost commit of this PR?

fviernau: Ok, I've updated all the assets to contain file list entries as you proposed. 
I've decided to make this a separate commit on top, so that the commit which implements returning the file lists in the scanner still nicely illustrates the impact: turning a previously empty list into a properly filled one.

sschuberth: Sorry, found a commit typo now: "Scann*er*Run"

scanner/src/funTest/assets/scanner-integration-all-pkgs-expected-ort-result.yml
```diff
@@ -406,7 +406,57 @@ scanner:
           path: "pkg4/pkg4.txt"
           start_line: -1
           end_line: -1
-  files: []
+  files:
```
sschuberth, line 409: Commit message: Something's wrong with the backticks around "sha1-sum", the opening one seems to be the wrong way. Why is is called `sha1-sum` instead of `sha1`, BTW?

cli/src/funTest/assets/semver4j-ort-result.yml
sschuberth: Thanks!

sschuberth (approved):

PR #7167, opened 2023-06-21
Start migrating scanner plugins to separate projects

sschuberth:

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -36,59 +36,6 @@ include(":evaluator")
 include(":helper-cli")
 include(":model")
 include(":notifier")
-include(":plugins:commands")
-include(":plugins:commands:advisor")
-include(":plugins:commands:analyzer")
-include(":plugins:commands:api")
-include(":plugins:commands:compare")
-include(":plugins:commands:config")
-include(":plugins:commands:downloader")
-include(":plugins:commands:evaluator")
-include(":plugins:commands:notifier")
-include(":plugins:commands:reporter")
-include(":plugins:commands:requirements")
-include(":plugins:commands:scanner")
-include(":plugins:commands:upload-curations")
-include(":plugins:commands:upload-result-to-postgres")
-include(":plugins:commands:upload-result-to-sw360")
-include(":plugins:package-curation-providers")
-include(":plugins:package-curation-providers:api")
-include(":plugins:package-curation-providers:clearly-defined")
-include(":plugins:package-curation-providers:file")
-include(":plugins:package-curation-providers:ort-config")
-include(":plugins:package-curation-providers:sw360")
-include(":plugins:package-managers")
-include(":plugins:package-managers:bower")
-include(":plugins:package-managers:bundler")
-include(":plugins:package-managers:cargo")
-include(":plugins:package-managers:carthage")
-include(":plugins:package-managers:cocoapods")
-include(":plugins:package-managers:composer")
-include(":plugins:package-managers:conan")
-include(":plugins:package-managers:gradle")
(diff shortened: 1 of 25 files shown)


settings.gradle.kts
```diff
@@ -106,61 +53,32 @@ project(":clients:osv").name = "osv-client"
...
 project(":utils:scripting").name = "scripting-utils"
 project(":utils:spdx").name = "spdx-utils"
 project(":utils:test").name = "test-utils"
 
+file("plugins").walk().maxDepth(3).filter {
+    it.isFile && it.name == "build.gradle.kts"
+}.mapTo(mutableListOf()) {
+    it.parentFile.toRelativeString(rootDir).replace(File.separatorChar, ':')
+}.forEach {
+    include(":$it")
```
fviernau, line 67: It wasn't super obvious to me what these two `it`s in line 65 and 67 refer to.
How about fixing this by using some variable, like e.g:

```
val moduleNames = file("plugins").walk().maxDepth(3).filter {
    it.isFile && it.name == "build.gradle.kts"
}.mapTo(mutableListOf()) {
    ":${it.parentFile.toRelativeString(rootDir).replace(File.separatorChar, ':')}" 
}

moduleNames.forEach { moduleName ->
     include(moduleName)
     ...
}
```

settings.gradle.kts
```diff
@@ -106,61 +53,32 @@ project(":clients:osv").name = "osv-client"
...
+    val parts = it.split(':')
+    if (parts.size == 3 && !parts[2].startsWith("gradle-") && !parts[2].endsWith("-template")) {
+        val singularName = parts[1].removeSuffix("s")
+
+        project(":$it").name = if (parts[2] == "api") {
+            "$singularName-api"
+        } else {
+            "${parts[2]}-$singularName"
+        }
+    }
```
fviernau, line 79: To me it's hard to grasp what

1. The condition in line 71 is about. Would it make sense to replace it with a hard coded ignore list for
     readability?
2. The reason why the "s" is stripped from the name in line 72

In general, would it make sense to add some code comment before line 62 (stating the purpose of the below code block) and to the logic in the function, to explain a bit the what and why / idea?

plugins/commands/scanner/build.gradle.kts
```diff
@@ -25,6 +25,8 @@ plugins {
 dependencies {
     api(project(":plugins:commands:command-api"))
 
+    implementation(platform(project(":plugins:scanners")))
```
fviernau, line 28: commit-msg: "This changes requires" -> "This change requires"?

fviernau:
LGTM now ok from my side. Maybe we can have one more reviewer? @oss-review-toolkit/core-devs ?

settings.gradle.kts
```diff
@@ -106,61 +53,36 @@ project(":clients:osv").name = "osv-client"
...
+    val parts = projectPath.split(':')
+    if (parts.size == 3 && parts[2] !in accompanyingProjects) {
+        // Convert the plural name for the type of plugin to singular.
+        val singularTypeName = parts[1].removeSuffix("s")
+
+        project(":$projectPath").name = if (parts[2] == "api") {
+            "$singularTypeName-api"
+        } else {
+            "${parts[2]}-$singularTypeName"
+        }
```
mnonnenmacher, line 82: I think a when expression would look nicer here:
```kotlin
        project(":$projectPath").name = when(parts[2]) {
            "api" -> "$singularTypeName-api"
            else -> "${parts[2]}-$singularTypeName"
        }
```

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #7168, opened 2023-06-21
fix(Scanner): Fix a typo

fviernau:


sschuberth (approved):

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -648,7 +648,7 @@ class Scanner(
             }
         }
 
-        logger.info { "Created file list for ${provenancesMissingFileLists.size} provenances in $duration." }
+        logger.info { "Created file lists for ${provenancesMissingFileLists.size} provenances in $duration." }
```
sschuberth, line 651: Commit message nit: The type could be "docs" as this is only a log statement, i.e. not a logic fix, and the word "fix" is already part of the actual title.

PR #7171, opened 2023-06-21
feat(web-app-reporter): Always show issue tab

### LINKED ISSUE #6232: Always show issues tab in Web App Reporter
When all issues are resolved, the issues tab disappears in the Web App Reporter. Unfortunately, this makes it hard to spot an issue, that was "falsely" resolved and is still making problems.

Also, the tab should appear IMO, even if there are no issues, since it is confusing, to have a tab only appear sometimes. If there are no issues, it just should state so, when selecting the tab.

hanna-modica:
This allows on one hand the view to be more consistent, by always showing the same tab, and on the other hand allows to check resolved issues, even if there are no open issues left.

resolves #6232

### DIFF
--- a/plugins/reporters/web-app-template/src/components/SummaryView.js
+++ b/plugins/reporters/web-app-template/src/components/SummaryView.js
@@ -373,87 +373,78 @@ class SummaryView extends React.Component {
                             <Col span={22} offset={1}>
                                 <Tabs tabPosition="top" className="ort-tabs-summary-overview">
                                     {
-                                        webAppOrtResult.hasRuleViolations()
-                                        && (
-                                            <TabPane
-                                                tab={(
-                                                    <span>
-                                                        <ExceptionOutlined />
-                                                        Rule Violations (
-                                                        {
-                                                            ruleViolations.length !== unresolvedRuleViolations
-                                                            && `${unresolvedRuleViolations}/`
-                                                        }
-                                                        {ruleViolations.length}
-                                                        )
-                                                    </span>
-                                                )}
-                                                key="ort-summary-rule-violations-table"
-                                            >
-                                                <RuleViolationsTable
-                                                    onChange={
-                                                        SummaryView.onChangeRuleViolationsTable
+                                        <TabPane
+                                            tab={(
+                                                <span>
+                                                    <ExceptionOutlined />
+                                                    Rule Violations (
+                                                    {
+                                                        ruleViolations.length !== unresolvedRuleViolations
+                                                        && `${unresolvedRuleViolations}/`
                                                     }
-                                                    ruleViolations={webAppOrtResult.ruleViolations}
-                                                    showExcludesColumn={webAppOrtResult.hasExcludes()}
-                                                    state={columns.ruleViolations}
-                                                />
-                                            </TabPane>
-                                        )
+                                                    {ruleViolations.length}
+                                                    )
+                                                </span>
+                                            )}
+                                            key="ort-summary-rule-violations-table"
+                                        >
+                                            <RuleViolationsTable
+                                                onChange={
+                                                    SummaryView.onChangeRuleViolationsTable
+                                                }
+                                                ruleViolations={webAppOrtResult.ruleViolations}
+                                                showExcludesColumn={webAppOrtResult.hasExcludes()}
+                                                state={columns.ruleViolations}
+                                            />
+                                        </TabPane>
                                     }
                                     {
-                                        webAppOrtResult.hasIssues()
-                                        && (
-                                            <TabPane
-                                                tab={(
-                                                    <span>
-                                                        <BugOutlined />
-                                                        Issues (
-                                                        {
-                                                            issues.length !== unresolvedIssues
-                                                            && `${unresolvedIssues}/`
-                                                        }
-                                                        {issues.length}
-                                                        )
-                                                    </span>
-                                                )}
-                                                key="ort-summary-issues-table"
-                                            >
-                                                <IssuesTable
-                                                    issues={webAppOrtResult.issues}
-                                                    onChange={
-                                                        SummaryView.onChangeIssuesTable
+                                        <TabPane
+                                            tab={(
+                                                <span>
+                                                    <BugOutlined />
+                                                    Issues (
+                                                    {
+                                                        issues.length !== unresolvedIssues
+                                                        && `${unresolvedIssues}/`
                                                     }
-                                                    showExcludesColumn={webAppOrtResult.hasExcludes()}
-                                                    state={columns.issues}
-                                                />
-                                            </TabPane>
-                                        )
+                                                    {issues.length}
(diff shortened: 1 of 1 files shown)


plugins/reporters/web-app-template/src/components/SummaryView.js
```diff
@@ -402,8 +402,7 @@ class SummaryView extends React.Component {
                                         )
                                     }
                                     {
-                                        webAppOrtResult.hasIssues()
-                                        && (
+                                        (
```
sschuberth, line 405: I guess we could drop the surrounding `(` / `)` now an de-indent the whole `<TabPane>` block (i.e. lines 378 - 401).

sschuberth: For consistency, IMO we should do the same for rule violations and vulnerabilities above and below.

sschuberth (commented):
Please hard-wrap commit message body lines to avoid *commit-lint* complaining. While at it, "resolves" should be capitalized.

plugins/reporters/web-app-template/src/components/SummaryView.js
hanna-modica: I also thought about this, but was not sure, if this was in scope of the initial ticket. Also, should those all be individual commits?

tsteenbe:
This should be configurable option as we had users that don't want to see violations, issues and vulnerabilities. Propose we move this change in behavior to a reporter option (e.g. boolean to show tabs always or not).

sschuberth:
> This should be configurable option as we had users that don't want to see violations, issues and vulnerabilities. Propose we move this change in behavior to a reporter option (e.g. boolean to show tabs always or not).

Personally, I don't think it's worth to put the burden of implementing a reporter-specific option / feature flag on the OP for this. We should have a consistent UI independent of the number of issues / violations / vulnerabilities, and say *in the tabs* that there are none, if so.

hanna-modica:
> Propose we move this change in behavior to a reporter option (e.g. boolean to show tabs always or not).

A user would need to select this behaviour every time the report is loaded. IMHO, this does not sound that user friendly and I agree with @sschuberth. How can we come to an agreement here?

tsteenbe:
>A user would need to select this behaviour every time the report is loaded. IMHO, this does not sound that user friendly and I agree with @sschuberth. How can we come to an agreement here?

@hanna-modica  That now what I proposed - it would be a configuration option at time of generation of the report, not upon opening the web report. No user action required only config change by an ORT admin.

plugins/reporters/web-app-template/src/components/SummaryView.js
mnonnenmacher: To me both a single commit and separate commits would be fine here, as you prefer.

mnonnenmacher:
> > This should be configurable option as we had users that don't want to see violations, issues and vulnerabilities. Propose we move this change in behavior to a reporter option (e.g. boolean to show tabs always or not).
> 
> Personally, I don't think it's worth to put the burden of implementing a reporter-specific option / feature flag on the OP for this. We should have a consistent UI independent of the number of issues / violations / vulnerabilities, and say _in the tabs_ that there are none, if so.

I agree with @sschuberth here. In the last community meeting there was no one present who actually preferred hiding the tab if there are no issues, but several participants preferred to always show it. As this also fixes the issue that it is currently impossible to review resolved issues if all issues are resolved, I would prefer to merge this as is and introduce an option only if there is an actual demand for it which justifies the work and increased complexity.

plugins/reporters/web-app-template/src/components/SummaryView.js
sschuberth: If the rationale is the same for all of there changes (which I believe it is), I'd put all into a single commit. But I'm also ok with either way.

tsteenbe:
@sschuberth  @mnonnenmacher  I am against merging this as is - as I said in the community meeting I don't believe this feature work as @hanna-modica  expects it to work - if a scan has only resolved issues to my recollection it won't show these in issues (cam from old HERE decision) - this needs to be tested first. 

We have ORT users that use SBOM and ORT web reports in the FOSS reporting to their customers as a replacement for Excel and HTML reports from their previous commercial vendor.  WebApp is provided as convenience for customer legal counsel to quickly browse the result (happy to get them off Excel). Just because ORT users are not at the community meeting doesn't mean their use cases should be taken into account.

hanna-modica:
> I don't believe this feature work as @hanna-modica expects it to work - if a scan has only resolved issues to my recollection it won't show these in issues (cam from old HERE decision) - this needs to be tested first.

In our reports, resolved issues are always shown in the issues tab. That aside, I believe it to be very confusing and not good UX nonetheless, if the shown tabs are not consistent for every report.

mnonnenmacher:
> @sschuberth @mnonnenmacher I am against merging this as is - as I said in the community meeting I don't believe this feature work as @hanna-modica expects it to work - if a scan has only resolved issues to my recollection it won't show these in issues (cam from old HERE decision) - this needs to be tested first.

I have verified that it works as expected, if all issues are resolved they are visible in the issues tab.

> We have ORT users that use SBOM and ORT web reports in the FOSS reporting to their customers as a replacement for Excel and HTML reports from their previous commercial vendor. WebApp is provided as convenience for customer legal counsel to quickly browse the result (happy to get them off Excel). Just because ORT users are not at the community meeting doesn't mean their use cases should be taken into account.

I still don't understand how an empty issues tab could break anyone's use of the web app report. To the contrary, if you are resolving issues, rule violations, or vulnerabilities one by one, you can always see the effects in the web app reporter, until you resolve the last one, because then those tabs just disappear which is a confusing UX.

plugins/reporters/web-app-template/src/components/SummaryView.js
```diff
@@ -373,87 +373,78 @@ class SummaryView extends React.Component {
...
-                                                        {ruleViolations.length}
-                                                        )
-                                                    </span>
-                                                )}
-                                                key="ort-summary-rule-violations-table"
-                                            >
-                                                <RuleViolationsTable
-                                                    onChange={
-                                                        SummaryView.onChangeRuleViolationsTable
+                                        <TabPane
```
mnonnenmacher, line 376: Commit message nits:
* "Always **show the** ... issue**s** ... tab**s**."
* "**the** one hand"
* "same tab**s**"

mnonnenmacher (changes requested):

plugins/reporters/web-app-template/src/components/SummaryView.js
hanna-modica: Thank you. I have updated the message.

hanna-modica: Hi Martin, I implemented your comments.

mnonnenmacher (approved):

PR #7179, opened 2023-06-22
fix(fossid-webapp): Accept path excludes for files with a '-'

nnobelis:

### DIFF
--- a/scanner/src/main/kotlin/scanners/fossid/Utils.kt
+++ b/scanner/src/main/kotlin/scanners/fossid/Utils.kt
@@ -28,7 +28,7 @@ import org.ossreviewtoolkit.model.config.PathExclude
 
 private val DIRECTORY_REGEX = "(?<directory>(?:\\w+/)*\\w+)/?(?<starstar>\\*\\*)?".toRegex()
 private val EXTENSION_REGEX = "\\*\\.(?<extension>\\w+)".toRegex()
-private val FILE_REGEX = "(?<file>[\\w.]+)".toRegex()
+private val FILE_REGEX = "(?<file>[^/]+)".toRegex()
 
 /**
  * Convert the ORT [path excludes][Excludes.paths] in [excludes] to FossID [IgnoreRule]s.. If an error is encountered
--- a/scanner/src/test/kotlin/scanners/fossid/MapIgnoreRulesTest.kt
+++ b/scanner/src/test/kotlin/scanners/fossid/MapIgnoreRulesTest.kt
@@ -140,6 +140,21 @@ class MapIgnoreRulesTest : WordSpec({
             issues should beEmpty()
         }
 
+        "map rule with files (with '-' in their names)" {
+            val exclude = Excludes(listOf(PathExclude("package-lock.json", PathExcludeReason.OTHER)))
+            val issues = mutableListOf<Issue>()
+
+            val ignoreRules = convertRules(exclude, issues)
+
+            ignoreRules shouldHaveSize 1
+            ignoreRules.first().shouldNotBeNull {
+                value shouldBe "package-lock.json"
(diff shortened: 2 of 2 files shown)


scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -28,7 +28,7 @@ import org.ossreviewtoolkit.model.config.PathExclude
 
 private val DIRECTORY_REGEX = "(?<directory>(?:\\w+/)*\\w+)/?(?<starstar>\\*\\*)?".toRegex()
 private val EXTENSION_REGEX = "\\*\\.(?<extension>\\w+)".toRegex()
-private val FILE_REGEX = "(?<file>[\\w.]+)".toRegex()
+private val FILE_REGEX = "(?<file>[\\w.\\-]+)".toRegex()
```
sschuberth, line 31: Could this be simplified to just not allow slashed, in which case the path exclude pattern should always point to a file? Like `"(?<file>[^/]+)".toRegex()`.

sschuberth: I'm specifically asking was there are more non-word characters that are valid in file names, so maintaining this as an "allow-list" becomes cumbersome.

nnobelis: Yeah I thought about it. For instance if somebody has Unicode characters...
Fixed.

scanner/src/test/kotlin/scanners/fossid/MapIgnoreRulesTest.kt
```diff
@@ -140,6 +140,21 @@ class MapIgnoreRulesTest : WordSpec({
             issues should beEmpty()
         }
 
+        "map rule with files (with '-' in their names" {
```
sschuberth: The test name lacks a closing `)`.

scanner/src/main/kotlin/scanners/fossid/Utils.kt
```diff
@@ -28,7 +28,7 @@ import org.ossreviewtoolkit.model.config.PathExclude
 
 private val DIRECTORY_REGEX = "(?<directory>(?:\\w+/)*\\w+)/?(?<starstar>\\*\\*)?".toRegex()
 private val EXTENSION_REGEX = "\\*\\.(?<extension>\\w+)".toRegex()
-private val FILE_REGEX = "(?<file>[\\w.]+)".toRegex()
+private val FILE_REGEX = "(?<file>[^/]+)".toRegex()
```
sschuberth, line 31: Please add to the commit message body: "Generally allow all characters except for a slash for file names now."

scanner/src/test/kotlin/scanners/fossid/MapIgnoreRulesTest.kt
```diff
@@ -125,7 +125,7 @@ class MapIgnoreRulesTest : WordSpec({
             issues should beEmpty()
         }
 
-        "map rule with files (with '.' in the names" {
+        "map rule with files (with '.' in their names" {
```
sschuberth, line 128: While at it, please also add the missing closing `)` here.

sschuberth (changes requested):

sschuberth (approved):

scanner/src/main/kotlin/scanners/fossid/Utils.kt
sschuberth: BTW, you should probably do something similar for `EXTENSION_REGEX`. Also, for a path exclude pattern like "*.tar.gz", what is `EXTENSION_REGEX` supposed to match / return? Really "tar.gz", or maybe just "gz"?

PR #7180, opened 2023-06-22
refactor: Move a private CLI helper function to the public analyzer API

mnonnenmacher:
Move the function to determine the enabled package managers to the public API of the analyzer to make it available when programmatically using ORT.

### DIFF
--- a/analyzer/src/main/kotlin/Extensions.kt
+++ b/analyzer/src/main/kotlin/Extensions.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2022 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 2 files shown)


analyzer/src/main/kotlin/Extensions.kt
```diff
@@ -0,0 +1,34 @@
...
+package org.ossreviewtoolkit.analyzer
+
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+
+/**
+ * Return the list of enabled [PackageManager]s based on the [AnalyzerConfiguration.enabledPackageManagers] and
+ * [AnalyzerConfiguration.disabledPackageManagers] configuration properties and the
+ * [default][PackageManagerFactory.isEnabledByDefault] of the [PackageManager]s.
+ */
+fun AnalyzerConfiguration.determineEnabledPackageManagers(): Set<PackageManagerFactory> {
```
sschuberth, line 29: @fviernau had the idea recently to not group functions by their type anymore, i.e. not put extensions functions into `Extensions.kt`, but by functionality. I basically think that's a good idea. So instead of introducing a *new* `Extensions.kt`, should we maybe move the function to e.g. `analyzer/src/main/kotlin/Analyzer.kt`?

mnonnenmacher: I'm not sure about that, maybe we should discuss in the next dev meeting if we want to define some rules here. To me `Extensions.kt` fits well in this case, because it's an API extension to the `AnalyzerConfiguration` class. Also, if moved it should probably go to `PackageManager.kt` instead of `Analyzer.kt` because that's the dependency of the function, but I'm also not a fan of putting too much functionality in a single file because it's against separation of concerns. Actually, I would also be for moving the `parseAuthorString` function out of `PackageManager.kt` for that reason.

sschuberth: Ok, then let's stick to our current conventions for now and bring this up in the next [Kotlin Developer Meeting](https://github.com/oss-review-toolkit/ort/wiki/Kotlin-Developer-Meeting).

sschuberth (dismissed):

sschuberth (approved):

PR #7181, opened 2023-06-22
ci(GitHub): Clean up the Gradle user home before caching

sschuberth:
See [1]. At a later point, this will become the default [2], and then this could be reverted.

[1]: https://github.com/gradle/gradle-build-action#removing-unused-files-from-gradle-user-home-before-saving-to-cache
[2]: https://github.com/gradle/gradle-build-action/issues/630

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -29,6 +29,7 @@ jobs:
     - name: Build all classes
       uses: gradle/gradle-build-action@v2
       with:
+        gradle-home-cache-cleanup: true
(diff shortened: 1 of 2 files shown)


fviernau (approved):

fviernau:
After merging this, will you delete existing cache entries?

sschuberth:
> After merging this, will you delete existing cache entries?

Yes, probably makes sense to do that, also given the pending build system refactoring like https://github.com/oss-review-toolkit/ort/pull/7167.

mnonnenmacher (approved):

PR #7187, opened 2023-06-22
fix(fossid-webapp): Remove issues for each pending file

nnobelis:
The FossId pending files were present in the scan result as issues with severity `HINT`. Now, the scan result contains the pending file's snippets.
However, counting the unique pending files having snippets is not an option since some pending files could have no snippet at all. Therefore, add a single issue with severity `HINT` with the count of pending files.

This is a breaking change.

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -846,10 +846,13 @@ class FossId internal constructor(
     ): ScanResult {
         // TODO: Maybe get issues from FossID (see has_failed_scan_files, get_failed_files and maybe get_scan_log).
 
-        // TODO: Deprecation: Remove the pending files in issues. This is a breaking change.
-        val issues = rawResults.listPendingFiles.mapTo(mutableListOf()) {
-            Issue(source = name, message = "Pending identification for '$it'.", severity = Severity.HINT)
-        }
+        val issues = mutableListOf(
+            Issue(
+                source = name,
+                message = "This scan has ${rawResults.listPendingFiles.size} file(s) pending identification in FossID.",
+                severity = Severity.HINT
+            )
+        )
 
         val snippetFindings = mapSnippetFindings(rawResults, issues)
 
--- a/plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
+++ b/plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
@@ -329,8 +329,7 @@ class FossIdTest : WordSpec({
             summary.licenseFindings shouldContainExactlyInAnyOrder expectedLicenseFindings
         }
 
-        // TODO: Deprecation: Remove the pending files in issues. This is a breaking change.
-        "report pending files as issues" {
+        "create an issue if there are files pending identification" {
             val projectCode = projectCode(PROJECT)
             val scanCode = scanCode(PROJECT, null)
             val config = createConfig(deltaScans = false)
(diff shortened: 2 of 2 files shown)


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -846,10 +846,12 @@ class FossId internal constructor(
     ): ScanResult {
         // TODO: Maybe get issues from FossID (see has_failed_scan_files, get_failed_files and maybe get_scan_log).
 
-        // TODO: Deprecation: Remove the pending files in issues. This is a breaking change.
-        val issues = rawResults.listPendingFiles.mapTo(mutableListOf()) {
-            Issue(source = name, message = "Pending identification for '$it'.", severity = Severity.HINT)
-        }
+        val issues = mutableListOf<Issue>()
+        issues += Issue(
```
mnonnenmacher, line 850: Nit: Could just be `val issues = listOf(Issue(...))`.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -846,10 +846,12 @@ class FossId internal constructor(
     ): ScanResult {
         // TODO: Maybe get issues from FossID (see has_failed_scan_files, get_failed_files and maybe get_scan_log).
 
-        // TODO: Deprecation: Remove the pending files in issues. This is a breaking change.
-        val issues = rawResults.listPendingFiles.mapTo(mutableListOf()) {
-            Issue(source = name, message = "Pending identification for '$it'.", severity = Severity.HINT)
-        }
+        val issues = mutableListOf<Issue>()
```
mnonnenmacher, line 849: Commit message question:
> However, counting the unique pending files having snippets is not an option since some pending files could have no snippet at all.

If you have the list of pending files and the snippet findings including their text locations a few lines below, can't you filter the pending files based on that? I'm not saying it should be done, I just wonder if I understand the commit message correctly.

mnonnenmacher (changes requested):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: Current state:
The list of file pending identifications are present as issues in the scan reults. Users use this data to find out if any file requires identification in FossID.

New states:
There is now the snippets in the scans results. One pending file can have one or more snippets.
However a file pending identification does not have necessary snippets. Or all its snippets have a match type ignored by ORT.

Therefore, for the user to find out if a file requires identification, collecting the pending files associated to the snippets is not enough. Hence, this commit add an issue containing the pending file count.
This could have been done before the introduction of the snippets model, but at the time we were unsure if the snippet model could help here.

sschuberth:
@nnobelis, please rebase to make some new / renamed CI checks work.

nnobelis:
> @nnobelis, please rebase to make some new / renamed CI checks work. Something went wrong anyway, as we have now 44 commits in here.

Yes this is really strange. I rebased on main to retrigger the tests but Github didn't detect the push.
I tried [this trick](https://stackoverflow.com/a/76590821/5321324) but it created this bug with the 44 commits :(

I tried then an hard reset on main + redo the commit, no luck :(

What should I do now ?

sschuberth:
> What should I do now ?

I've changed the "base" branch forth and back to "main", it now seems to be ok.

plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
```diff
@@ -350,9 +349,14 @@ class FossIdTest : WordSpec({
...
 
-            val expectedIssues = listOf(createPendingFile(4), createPendingFile(5)).map {
-                Issue(Instant.EPOCH, "FossId", "Pending identification for '$it'.", Severity.HINT)
-            }
+            val expectedIssues = listOf(
+                Issue(
+                    Instant.EPOCH,
+                    "FossId",
+                    "This scan has 2 file(s) pending identification in FossID.",
+                    Severity.HINT
```
sschuberth, line 357: When wrapping this anyway, please use named argument while at it.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -846,10 +846,13 @@ class FossId internal constructor(
     ): ScanResult {
         // TODO: Maybe get issues from FossID (see has_failed_scan_files, get_failed_files and maybe get_scan_log).
 
-        // TODO: Deprecation: Remove the pending files in issues. This is a breaking change.
-        val issues = rawResults.listPendingFiles.mapTo(mutableListOf()) {
-            Issue(source = name, message = "Pending identification for '$it'.", severity = Severity.HINT)
-        }
+        val issues = mutableListOf(
```
sschuberth, line 849: This can be a non-mutable `listOf()` now, or?

nnobelis: `mapSnippetFindings` and `mapSummary` expect a mutable list to add new issues if required.

PR #7196, opened 2023-06-24
Fix vulnerability count

mnonnenmacher:

### DIFF
--- a/plugins/commands/advisor/src/main/kotlin/AdvisorCommand.kt
+++ b/plugins/commands/advisor/src/main/kotlin/AdvisorCommand.kt
@@ -140,12 +140,16 @@ class AdvisorCommand : OrtCommand(
         println("The advice took $duration.")
 
         with(advisorRun.results.getVulnerabilities()) {
-            val totalPackageCount = ortResultOutput.getPackages(omitExcluded = true).size
+            val includedPackages = ortResultOutput.getPackages(omitExcluded = true).map { it.metadata.id }
+            val totalPackageCount = includedPackages.size
+            val vulnerablePackageCount = count { (id, vulnerabilities) ->
+                id in includedPackages && vulnerabilities.isNotEmpty()
+            }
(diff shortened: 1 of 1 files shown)


plugins/commands/advisor/src/main/kotlin/AdvisorCommand.kt
```diff
@@ -140,12 +140,16 @@ class AdvisorCommand : OrtCommand(
         println("The advice took $duration.")
 
         with(advisorRun.results.getVulnerabilities()) {
-            val totalPackageCount = ortResultOutput.getPackages(omitExcluded = true).size
+            val includedPackages = ortResultOutput.getPackages(omitExcluded = true).map { it.metadata.id }
```
sschuberth, line 143: > The map returned by `getVulnerabilities()` contains entries for all
packages, also those that do not have any vulnerabilities.

BTW, this is something that also occurred to me in the context of https://github.com/oss-review-toolkit/ort/pull/6613: The advisor's `retrievePackageFindings()` API definition does not make clear whether the returned map should contain entries for packages that have empty `defects` and `vulnerabilities` as part of the `AdvisorResult`. Should it? Just to get the `AdvisorDetails` and `AdvisorSummary` returned?

mnonnenmacher: I'm not sure about this as well, in some situations like package curations we decided that it has benefits to have explicit empty results to document that something was requested.

sschuberth (approved):

plugins/commands/advisor/src/main/kotlin/AdvisorCommand.kt
sschuberth: Discussed as part of the [Kotlin developer meeting](https://github.com/oss-review-toolkit/ort/wiki/Kotlin-Developer-Meeting#2023-06-26).

PR #7198, opened 2023-06-25
Introduce markdownlint

mnonnenmacher:
Introduce [markdownlint](https://github.com/DavidAnson/markdownlint) for linting Markdown files and apply the required fixes. See the commit messages for details.


.markdownlint-cli2.yaml
```diff
@@ -0,0 +1,11 @@
+#  Configuration for the markdownlint-cli2, see:
```
sschuberth, line 1: I'm curious: Could you explain in the commit message why you chose https://github.com/DavidAnson/markdownlint over https://github.com/markdownlint/markdownlint?

README.md
```diff
@@ -36,12 +36,12 @@ The OSS Review Toolkit (ORT) is a FOSS policy automation and orchestration toolk
 
 You can use it to:
 
-- Generate CycloneDX, SPDX SBOMs, or custom FOSS attribution documentation for your software project
-- Automate your FOSS policy using risk-based Policy as Code to do licensing, security vulnerability, InnerSource
+* Generate CycloneDX, SPDX SBOMs, or custom FOSS attribution documentation for your software project
```
sschuberth, line 39: Commit message: Typo in "asteri*s*ks".

ADOPTERS.md
```diff
@@ -1,9 +1,9 @@
 
 # Adopters
 
-_If you're using ORT in your organization, please add your organization's name to the alphabetically sorted table below.
+*If you're using ORT in your organization, please add your organization's name to the alphabetically sorted table below.
```
sschuberth, line 4: Commit message: Typo in "asteri*s*ks".

README.md
```diff
@@ -107,17 +107,23 @@ Depending on how ORT was installed, it can be run in the following ways:
 
 * If the Docker image was built, use
 
-      docker run ort --help
+  ```bash
```
sschuberth, line 110: We sometimes also used the more generic "shell", which is sufficient here. As I don't think we have any Bash-specific code blocks, would it make sense to align all code blocks on "shell"?

.markdownlint-cli2.yaml
mnonnenmacher: The main reason was that it can be run with `npx` which is already required for `commitlint`, while the other linter requires Ruby. I have added more information to the commit message.

mnonnenmacher: Btw, I did try the former first, but I did not do a proper comparison of the results.

README.md
mnonnenmacher: I have updated the commit and added another one to align on `shell`.

sschuberth (approved):

PR #7200, opened 2023-06-26
fix(fossid-webapp): Do not generate an issue for .git ignore rules

nnobelis:

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -545,6 +545,11 @@ class FossId internal constructor(
             ignoreRules.data?.let { rules ->
                 logger.info { "${rules.size} ignore rule(s) have been found." }
 
+                // When a scan is created with the optional property 'git_repo_url', the server automatically creates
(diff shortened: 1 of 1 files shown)


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -548,7 +548,9 @@ class FossId internal constructor(
                 // When a scan is created with the optional property 'git_repo_url', the server automatically creates
                 // an 'ignore rule' to exclude the '.git' directory.
                 // Therefore, this rule will be created automatically and does not need to be carried from the old scan.
-                val filteredRules = rules.filterNot { it.type == RuleType.DIRECTORY && it.value == ".git" }
+                val filteredRules = rules.filterNot {
```
sschuberth, line 551: Commit message:
- Typo in "somes".
- Maybe instead of "escape" say that some versions seem to use the literal pattern for a regular expression?
- Consider extracting the set to a variable outside of the loop implied by `filterNot`.

sschuberth (changes requested):

sschuberth (approved):

PR #7201, opened 2023-06-26
scanner: Stop including any irrelevant provenance in `ScannerRun`

### LINKED ISSUE #7192: scanner: Filter `ScannerRun.provenances` by VCS paths
A `ProvenanceResolutionResult` entry `ScannerRun.provenance` may contain sub-repositories which are outside of the VCS path of the corresponding package. In this case, `ScannerRun` may contain scan results for provenances which get filtered out anyway by `OrtResult.getScanResults()`. 

Drop such sub-repository entries to not add unneeded data to `OrtResult`. 
Consider also the special case where the VCS path of a package points into a sub-repository. One could argue that in this case all repositories on the path from VCS path to root need to be kept, since the root license matching algorithm needs these files as input. But one could also argue that that algorithm should stop at repository boundaries.

fviernau:
The findings returned by `OrtResult.getScanResults()` weren't wrong / superflous, but (unused) sub-repositories / findings contained inside of `OrtResult` have not yet been filtered by VCS path. This PR fixes that, see individual commits.

Fixes #7192.


model/src/main/kotlin/ScannerRun.kt
```diff
@@ -27,8 +27,7 @@ import java.time.Instant
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.utils.ProvenanceResolutionResultSortedSetConverter
 import org.ossreviewtoolkit.model.utils.ScanResultSortedSetConverter
-import org.ossreviewtoolkit.model.utils.alignRevisions
-import org.ossreviewtoolkit.model.utils.clearVcsPath
+import org.ossreviewtoolkit.model.utils.getKnownProvenancesWithoutVcsPath
```
sschuberth, line 30: Nit: This commit could nicely be squashed with the previous one, saying to move two extension functions from `ScannerRun` to `Extensions` (with the same rationale, that then only needs to be mentioned once in the single commit message).

scanner/src/funTest/assets/scanner-integration-expected-ort-result.yml
```diff
@@ -45,10 +45,38 @@ analyzer:
       scopes:
       - name: "deps"
         dependencies:
+        - id: "Dummy::pkg0:1.0.0"
```
sschuberth: Not strictly related to this commit alone, but would it make sense to give the packages names that reflect their properties, instead of using numbers? Like, "pkg0" could be renamed to "pkg-with-empty-vcs".

scanner/src/funTest/assets/scanner-integration-all-pkgs-expected-ort-result.yml
```diff
@@ -16,10 +16,10 @@ analyzer:
...
   environment:
     ort_version: "HEAD"
-    java_version: "<REPLACE_JAVA>"
-    os: "<REPLACE_OS>"
-    processors: "<REPLACE_PROCESSORS>"
-    max_memory: "<REPLACE_MAX_MEMORY>"
+    java_version: "17.0.7"
+    os: "Linux"
+    processors: 8
+    max_memory: 536870912
```
sschuberth, line 22: Placeholders should be kept.

scanner/src/funTest/kotlin/scanners/ScannerIntegrationFunTest.kt
```diff
@@ -79,6 +79,18 @@ class ScannerIntegrationFunTest : WordSpec({
                     matchExpectedResult(expectedResultFile)
         }
     }
+
+    "scan() a subset of the packages corresponding to a single VCS" should {
+        "return the expected ORT result" {
```
sschuberth, line 84: IIUC, this test does not pass yet and should be disabled here, and enabled in the next commit.

sschuberth (changes requested):

scanner/src/funTest/assets/scanner-integration-expected-ort-result.yml
fviernau: I deliberately went for this setup. This is easy to grasp as the package names align with the directory names and also the files in the repositories. Also, after working on `SpdxDocumentReporterFunTest` I regret the decision that  I've put intention of the test as properties of the package. Finally, also the `pkgs` can be used for testing different things, which are not forsee-able yet. I prefer to keep as-is, since benefit of changing for me is no clear and re-working it is also quite a bit of effort.

scanner/src/funTest/kotlin/scanners/ScannerIntegrationFunTest.kt
fviernau: The test does pass. To be sure, I just verified it again. Note that the expected result is updated in the next commit to illustrated the effect of the implemented filtering.

scanner/src/funTest/kotlin/scanners/ScannerIntegrationFunTest.kt
```diff
@@ -59,18 +59,18 @@ import org.ossreviewtoolkit.utils.test.matchExpectedResult
 import org.ossreviewtoolkit.utils.test.patchActualResult
 
 class ScannerIntegrationFunTest : WordSpec({
-    "scan()" should {
-        val analyzerResult = createAnalyzerResult()
+    "scan() all packages corresponding to a single VCS" should {
```
sschuberth, line 62: Nit: It should actually say "scanning", but that would not mention the function name anymore. While you probably could write something like "scan()'ing", maybe instead rephrase like

    "For packages with the same provenance, scan()" should

(I'm not sure whether that's the right wording, but you get my point.)

scanner/src/funTest/kotlin/scanners/ScannerIntegrationFunTest.kt
```diff
@@ -79,6 +79,18 @@ class ScannerIntegrationFunTest : WordSpec({
                     matchExpectedResult(expectedResultFile)
         }
     }
+
+    "scan() a subset of the packages corresponding to a single VCS" should {
```
sschuberth, line 83: See above for the wording.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -767,3 +779,10 @@ private fun ScanController.getSubRepositories(id: Identifier): Map<String, VcsIn
 
 private fun ScanResult.clearPackageVerificationCode(): ScanResult =
     copy(summary = summary.copy(packageVerificationCode = ""))
+
+private fun ProvenanceResolutionResult.filterByVcsPath(): ProvenanceResolutionResult =
+    copy(
+        subRepositories = subRepositories.filter { (path, _) ->
+            File(path).startsWith(File(packageProvenance?.vcsPath.orEmpty()))
```
sschuberth, line 786: `File.startsWith()` has an overload that takes a `String` that could be used here.

sschuberth (changes requested):

scanner/src/funTest/kotlin/scanners/ScannerIntegrationFunTest.kt
fviernau: I just went for "scanning". I believe it's good enough without mentioning the function name.

sschuberth: Ok, but as this is a regular word now that starts the sentence, it should be capitalized (in both tests).

sschuberth (approved):

PR #7202, opened 2023-06-26
fix(advisor): Log the correct number of distinct vulnerabilities

mnonnenmacher:
Previously, the code counted the number of distinct advisor results instead of the number of distinct vulnerabilities.

### DIFF
--- a/advisor/src/main/kotlin/Advisor.kt
+++ b/advisor/src/main/kotlin/Advisor.kt
@@ -90,8 +90,8 @@ class Advisor(
                         val providerResults = provider.retrievePackageFindings(packages)
 
                         logger.info {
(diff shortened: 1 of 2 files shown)


fviernau (dismissed):

sschuberth:
[Looks like](https://github.com/oss-review-toolkit/ort/actions/runs/5377002333/jobs/9766114494#step:4:209) a mock answer needs an update now.

mnonnenmacher:
> [Looks like](https://github.com/oss-review-toolkit/ort/actions/runs/5377002333/jobs/9766114494#step:4:209) a mock answer needs an update now.

A good example of why I'm not a fan of mock testing in such cases, implementation details too often leak into the test code.

sschuberth (approved):

PR #7204, opened 2023-06-26
CycloneDX reporter improvements

sschuberth:


plugins/reporters/cyclonedx/src/main/kotlin/CycloneDxReporter.kt
```diff
@@ -103,8 +103,8 @@ class CycloneDxReporter : Reporter {
     private fun mapHash(hash: org.ossreviewtoolkit.model.Hash): Hash? =
         enumValues<Hash.Algorithm>().find { it.spec == hash.algorithm.toString() }?.let { Hash(it, hash.value) }
 
-    private fun mapLicenseNamesToObjects(licenseNames: Collection<String>, origin: String, input: ReporterInput) =
-        licenseNames.map { licenseName ->
+    private fun Collection<String>.mapToLicenseObjects(origin: String, input: ReporterInput) =
```
fviernau, line 106: I believe the term "license names" had provided more value than "license objects" in terms of readability.
IMO we should add explicit return type `List<License>` so that "toLicenseObjects" becomes kind of redundant.
How about making it `licenseNamesToLicenses()`?

Just an idea: Maybe a separate commit could turn this into a (simpler?) singular form function? e.g.
`createLicense(licenseName, origin, input)` so that it can be called like `concludedLicenseNames.map { createLicense(it, origin, input)`.

fviernau (changes requested):

plugins/reporters/cyclonedx/src/main/kotlin/CycloneDxReporter.kt
fviernau: > I don't like the "license" doubling in there. Maybe (map)NamesToLicenses?

Agreed.

PR #7208, opened 2023-06-26
Generalize the ClearlyDefined scan storage implementation

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/storages/ClearlyDefinedStorage.kt
+++ b/scanner/src/main/kotlin/storages/ClearlyDefinedStorage.kt
@@ -52,24 +52,14 @@ import org.ossreviewtoolkit.scanner.ScanStorageException
 import org.ossreviewtoolkit.scanner.ScannerCriteria
 import org.ossreviewtoolkit.scanner.scanners.scancode.generateScannerDetails
 import org.ossreviewtoolkit.scanner.scanners.scancode.generateSummary
+import org.ossreviewtoolkit.utils.common.AlphaNumericComparator
 import org.ossreviewtoolkit.utils.common.collectMessages
+import org.ossreviewtoolkit.utils.common.withoutPrefix
 import org.ossreviewtoolkit.utils.ort.OkHttpClientHelper
 import org.ossreviewtoolkit.utils.ort.showStackTrace
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 
 import retrofit2.HttpException
-
-/** The name used by ClearlyDefined for the ScanCode tool. */
-private const val TOOL_SCAN_CODE = "scancode"
-
-/**
- * Given a list of [tools], return the version of ScanCode that was used to scan the package with the given
- * [coordinates], or return null if no such tool entry is found.
- */
-private fun findScanCodeVersion(tools: List<String>, coordinates: Coordinates): String? {
-    val toolUrl = "$coordinates/$TOOL_SCAN_CODE/"
-    return tools.find { it.startsWith(toolUrl) }?.substring(toolUrl.length)
-}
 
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/storages/ClearlyDefinedStorage.kt
```diff
@@ -122,13 +112,30 @@ class ClearlyDefinedStorage(
...
-            } else {
-                logger.debug { "$coordinates was not scanned with any version of ScanCode." }
+            val toolsVersionsByName = tools.mapNotNull { it.withoutPrefix("$coordinates/") }
+                .groupBy({ it.substringBefore('/') }, { it.substringAfter('/') })
+                .mapValues { (_, versions) -> versions.sortedWith(AlphaNumericComparator) }
+
+            toolsVersionsByName.mapNotNull { (name, versions) ->
+                // See https://github.com/clearlydefined/service#tool-name-registry.
+                when (name) {
+                    "scancode" -> {
```
fviernau, line 122: In the preceding commit one usage of the constant `TOOL_SCAN_CODE` has been removed. Should line 122 make use of a constant? Maybe of `TOOL_SCAN_CODE` ?

scanner/src/main/kotlin/storages/ClearlyDefinedStorage.kt
```diff
@@ -122,13 +112,30 @@ class ClearlyDefinedStorage(
                 coordinates.revision.orEmpty()
             )
 
-            val version = findScanCodeVersion(tools, coordinates)
-            if (version != null) {
-                listOfNotNull(loadScanCodeResult(coordinates, version))
-            } else {
-                logger.debug { "$coordinates was not scanned with any version of ScanCode." }
+            val toolsVersionsByName = tools.mapNotNull { it.withoutPrefix("$coordinates/") }
```
fviernau, line 115: The plural `tools` in `toolsVersionsByName` looked a bit unusual within our code base IIRC. 
Maybe instead `toolVersionsByName`, `toolVersionsByToolName`, `versionsForToolName` or `versionsForName` ?

scanner/src/main/kotlin/storages/ClearlyDefinedStorage.kt
```diff
@@ -122,13 +112,30 @@ class ClearlyDefinedStorage(
...
+                when (name) {
+                    "scancode" -> {
+                        loadToolData(coordinates, name, versions.last())?.let { result ->
+                            val summary = generateSummary(SpdxConstants.NONE, result)
+                            val details = generateScannerDetails(result)
+
+                            val provenance = getProvenance(coordinates)
+
+                            ScanResult(provenance, details, summary)
+                        }
```
fviernau, line 130: Is the case when `loadToolData()` returns null unexpected and if so should it be logged?

fviernau (changes requested):

scanner/src/main/kotlin/storages/ClearlyDefinedStorage.kt
sschuberth: I've deliberately removed the constant as the string now only occurs once, and I found it more convenient to see the actual string values directly as part of then `when` clause. Also, as I plan to introduce more `when` cases, I otherwise would have needed to introduce constants for these, too.

scanner/src/main/kotlin/storages/ClearlyDefinedStorage.kt
sschuberth: Yeah, the plural "tools" in there was a typo, so I went for "toolVersionsByName".

fviernau (approved):

Issue #6948, opened 2023-04-28
Compute package verification code on-the-fly


fviernau:
After including a full file listing (path / sha1) for all provenances in `OrtResult` (see #6946), the package
verification code can be computed from that. So:

1. Remove the package verification code from any `ScanResult` classes
2. Compute the package verification code instead from these file listings (probably outside of the scanner module).

PR #7212, opened 2023-06-28
SpdxDocumentReporter: Compute SPDX verification codes on-the-fly

fviernau:
Compute the SPDX package verification code based on the file lists which have been added to `OrtResult` recently.

See individual commits.

Fixes: #6948.

### DIFF
--- a/plugins/reporters/spdx/src/main/kotlin/Extensions.kt
+++ b/plugins/reporters/spdx/src/main/kotlin/Extensions.kt
@@ -23,6 +23,7 @@ package org.ossreviewtoolkit.plugins.reporters.spdx
 
 import org.ossreviewtoolkit.model.Hash
 import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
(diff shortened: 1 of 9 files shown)


plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -171,6 +174,20 @@ internal fun Package.toSpdxPackage(
...
 
+private fun OrtResult.getVcsScanResult(id: Identifier): ScanResult? =
+    getScanResultsForId(id).firstOrNull { it.provenance is RepositoryProvenance }
+
+private fun OrtResult.getResolvedRevision(id: Identifier): String? =
+    (getVcsScanResult(id)?.provenance as RepositoryProvenance?)?.resolvedRevision
+
+private fun OrtResult.getPackageVerificationCode(id: Identifier, type: SpdxPackageType): String? =
+    when (type) {
+        SpdxPackageType.VCS_PACKAGE -> getVcsScanResult(id)?.summary?.packageVerificationCode?.takeUnless {
```
sschuberth: What about `SpdxPackageType.PROJECT` here? Shouldn't we also be able to compute the package verification codes for projects?

fviernau: Yes, that's also possible. My intention was to, in the first place, only re-implement it for packages - to get back the functionality we had before. IIRC we haven't ever reported package verification codes for projects. If you're fine I'd add this later, in particular as I fear that for projects I've introduced a bug (mentioned in general slack by Nicola), which I need to investigate first. If you're ok, I'd leave this for now up to a following PR?

sschuberth: > IIRC we haven't ever reported package verification codes for projects.

That's indeed the question, but looks like you're right: The original code had

    val spdxProjectPackage = project.toPackage().toSpdxPackage(SpdxPackageType.PROJECT, licenseInfoResolver)

which does not pass a third argument of type `ScanResult` to `toSpdxPackage()`, so it defaults to `null`, which ripples down the package verification code being `null` for projects.

> If you're fine I'd add this later

Sure, esp. as I've convinced myself now that it's new functionality.

sschuberth (approved):

PR #7214, opened 2023-06-28
Scan summary drop package verification code

fviernau:
The SPDX package verification codes have started to turn obsolete once `ScanSummary` started to correspond (also) to a `Provenance`, as of the re-write of the scanner to scan "by provenance". However, scan summaries are still used to describe results of a package, e.g. as `OrtResult.getScanResult()` does. Recently reporters have been updated to compute these verification codes solely based on the file lists contained in `OrtResult` turning `ScanSummary.packageVerificationCode` fully obsolete. This PR removes that property.

Follow-up for #6948.

### DIFF
--- a/cli/src/funTest/assets/semver4j-ort-result.yml
+++ b/cli/src/funTest/assets/semver4j-ort-result.yml
@@ -376,7 +376,7 @@ advisor:
       - advisor:
           name: "NexusIQ"
           capabilities:
-            - "VULNERABILITIES"
+          - "VULNERABILITIES"
         summary:
           start_time: "2021-04-29T14:54:17.322191Z"
           end_time: "2021-04-29T14:54:18.966672Z"
--- a/cli/src/funTest/assets/semver4j-ort-result.yml
+++ b/cli/src/funTest/assets/semver4j-ort-result.yml
@@ -217,7 +217,6 @@ scanner:
     summary:
       start_time: "2020-09-30T09:28:32.817956Z"
       end_time: "2020-09-30T09:29:02.889213Z"
-      package_verification_code: ""
       licenses:
       - license: "Apache-2.0"
         location:
@@ -290,7 +289,6 @@ scanner:
     summary:
       start_time: "2020-09-30T09:27:12.023451Z"
       end_time: "2020-09-30T09:28:20.525647Z"
-      package_verification_code: ""
       licenses:
       - license: "BSD-3-Clause"
         location:
--- a/scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
+++ b/scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
@@ -102,7 +102,6 @@ abstract class AbstractStorageFunTest(vararg listeners: TestListener) : WordSpec
     private val scanSummaryWithFiles = ScanSummary.EMPTY.copy(
         startTime = Instant.EPOCH + Duration.ofMinutes(1),
         endTime = Instant.EPOCH + Duration.ofMinutes(2),
-        packageVerificationCode = "packageVerificationCode",
         licenseFindings = setOf(
             LicenseFinding("license-1.1", DUMMY_TEXT_LOCATION),
             LicenseFinding("license-1.2", DUMMY_TEXT_LOCATION)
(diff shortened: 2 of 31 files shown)


cli/src/funTest/assets/semver4j-ort-result.yml
```diff
@@ -376,7 +376,7 @@ advisor:
       - advisor:
           name: "NexusIQ"
           capabilities:
-            - "VULNERABILITIES"
+          - "VULNERABILITIES"
```
sschuberth, line 377: BTW, this is interesting in the context of https://github.com/oss-review-toolkit/ort/issues/7219. Actually both indentations are valid YAML (so it's not really a formatting issue that needs fixing). I'm curious to know what change in ORT causes the different serialization... probably a Jackson / SnakeYAML update.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -777,8 +771,7 @@ fun ScanResult.toNestedProvenanceScanResult(nestedProvenance: NestedProvenance):
 
     val provenances = nestedProvenance.getProvenances()
     val scanResultsByProvenance = provenances.associateWith { provenance ->
-        // TODO: Find a solution for the incorrect packageVerificationCode and for how to associate issues to the
-        //       correct scan result.
+        // TODO: Find a solution for how to associate issues to the correct scan result.
```
sschuberth, line 774: Isn't the change of this comment actually a fixup for calculating the PVC on the fly?

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
```diff
@@ -155,7 +154,9 @@ abstract class AbstractStorageFunTest(vararg listeners: TestListener) : WordSpec
 
             "not store a result for the same scanner and provenance twice" {
                 val summary1 = scanSummaryWithFiles
-                val summary2 = scanSummaryWithFiles.copy(packageVerificationCode = "anotherPackageVerificationCode")
+                val summary2 = scanSummaryWithFiles.copy(
+                    startTime = scanSummaryWithFiles.startTime.plusSeconds(10)
```
sschuberth, line 158: Unrelated change?

scanner/src/main/kotlin/Scanner.kt
sschuberth: BTW, while double-checking I also discovered the comment at https://github.com/oss-review-toolkit/ort/blob/70a13d2d70c7c54ddcc446b74580e459c01883ff/model/src/main/kotlin/utils/ScanResultUtils.kt#L37-L38 which could also be adjusted now.

sschuberth: And also https://github.com/oss-review-toolkit/ort/blob/70a13d2d70c7c54ddcc446b74580e459c01883ff/scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt#L67-L68

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
fviernau: This change was intended. IIUC the removed code intended to create a second summary which is different from the first one. In particular it was only targeting difference in general - not difference in terms of package verification code. This is why I choose to achive the (assumed) goal by modifying the timestamp instead.

scanner/src/main/kotlin/Scanner.kt
fviernau: > Isn't the change of this comment actually a fixup for calculating the PVC on the fly?

Maybe it depends on the point of view. If one say the scanner does not have any knowledge about any "reporters", which I believe is true, then probably not. Also my approach was: 1. Implement a replacement for the non working existing package verification code first (previous PR) 2. Then remove all the left-over in the scanner (this PR). That separation still makes sense to me, so I'd say it's not a fix-up.

cli/src/funTest/assets/semver4j-ort-result.yml
fviernau: > Actually both indentations are valid YAML

I prefer to use what ORT outputs programatically as the single source of truth what is the right intendation. Otherwise, we could not use our own output anymore without reformatting it causing manual effort.

sschuberth: > I prefer to use what ORT outputs programatically

My point is: What ORT outputs programmatically seems to have changed (at some point of time). Anyway, I prefer the "new" syntax with less indentation, so I like the change per se.

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
sschuberth: I see. This wasn't obvious to me, but it makes sense.

sschuberth (approved):

scanner/src/funTest/kotlin/storages/AbstractStorageFunTest.kt
fviernau: yeah, it would have been better I mentioned it in the commit.

PR #7217, opened 2023-06-28
refactor(web-app): Improve the name of a test asset

fviernau:


plugins/reporters/web-app/src/funTest/kotlin/WebAppReporterFunTest.kt
```diff
@@ -31,7 +31,7 @@ import org.ossreviewtoolkit.utils.test.readOrtResult
 class WebAppReporterFunTest : WordSpec({
     "WebAppReporter" should {
         "successfully export to a web application" {
-            val ortResult = readOrtResult(getAssetFile("dummy-expected-output-for-analyzer-result.yml"))
+            val ortResult = readOrtResult(getAssetFile("input-analyzer-result.yml"))
```
sschuberth, line 34: As this file also contains a scan result, I propose to instead use e.g. "scan-result-for-synthetic-gradle-lib.yml".

plugins/reporters/web-app/src/funTest/kotlin/WebAppReporterFunTest.kt
```diff
@@ -31,7 +31,7 @@ import org.ossreviewtoolkit.utils.test.readOrtResult
 class WebAppReporterFunTest : WordSpec({
     "WebAppReporter" should {
         "successfully export to a web application" {
-            val ortResult = readOrtResult(getAssetFile("dummy-expected-output-for-analyzer-result.yml"))
+            val ortResult = readOrtResult(getAssetFile("scan-result-for-synthetic-gradle-lib.yml.yml"))
```
sschuberth, line 34: There's a double `.yml` extension now.

sschuberth (changes requested):

sschuberth:
The pattern in `plugins/reporters/web-app/src/funTest/assets/.gitattributes` now does not match anymore, but actually I believe the whole file could be dropped, as the file is only used for deserialization, and not for comparison / as an expected result.

fviernau:
> but actually I believe the whole file could be dropped,

What would you use as a replacement?

sschuberth:
> > but actually I believe the whole file could be dropped,
> 
> What would you use as a replacement?

Nothing, the `.gitattributes` file should simply be removed. I don't think the file is needed at all, and I made a mistake of putting it there in the context of factoring out the reporters.

sschuberth:
> The pattern in `plugins/reporters/web-app/src/funTest/assets/.gitattributes` now does not match anymore, but actually I believe the whole file could be dropped, as the file is only used for deserialization, and not for comparison / as an expected result.

Mind adding the rationale for being able to remove the `.gitattributes` file to the commit message? At your choice, the removal could also happen in a preceding commit, as the file could have been removed independently already.

sschuberth (approved):

PR #7224, opened 2023-06-29
fix(spdx-utils): Make `copyrightText` default to `NOASSERTION`

sschuberth:
As clarified in the 2.3 specs (see [1] vs. [2]), an absent `copyrightText` should be interpreted as `NOASSERTION`.

[1]: https://spdx.github.io/spdx-spec/v2.2.2/package-information/#717-copyright-text-field
[2]: https://github.com/spdx/spdx-spec/blob/development/v2.2/schemas/spdx-schema.json#L311

### DIFF
--- a/utils/spdx/src/main/kotlin/model/SpdxFile.kt
+++ b/utils/spdx/src/main/kotlin/model/SpdxFile.kt
@@ -59,7 +59,7 @@ data class SpdxFile(
      * A text relating to a copyright notice, even if not complete. To represent a not present value
      * [SpdxConstants.NONE] or [SpdxConstants.NOASSERTION] must be used.
      */
-    val copyrightText: String,
+    val copyrightText: String = SpdxConstants.NOASSERTION,
 
     /**
      * The list of contributors which contributed to the file. Contributors could include names of copyright holders
--- a/utils/spdx/src/main/kotlin/model/SpdxPackage.kt
+++ b/utils/spdx/src/main/kotlin/model/SpdxPackage.kt
@@ -66,7 +66,7 @@ data class SpdxPackage(
      * A text relating to a copyright notice, even if not complete. To represent a not present value
      * [SpdxConstants.NONE] or [SpdxConstants.NOASSERTION] must be used.
      */
-    val copyrightText: String,
+    val copyrightText: String = SpdxConstants.NOASSERTION,
 
     /**
      * A more detailed description of the package as opposed to [summary], which may be an extract from the package.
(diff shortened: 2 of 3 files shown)


utils/spdx/src/main/kotlin/model/SpdxFile.kt
```diff
@@ -59,7 +59,7 @@ data class SpdxFile(
      * A text relating to a copyright notice, even if not complete. To represent a not present value
      * [SpdxConstants.NONE] or [SpdxConstants.NOASSERTION] must be used.
      */
-    val copyrightText: String,
+    val copyrightText: String = SpdxConstants.NOASSERTION,
```
fviernau, line 62: commit-msg: I wonder if the link for [2] is mixed up. Should it be as follows?

[2] https://spdx.github.io/spdx-spec/v2.3/package-information/#717-copyright-text-field

utils/spdx/src/main/kotlin/model/SpdxPackage.kt
```diff
@@ -66,7 +66,7 @@ data class SpdxPackage(
      * A text relating to a copyright notice, even if not complete. To represent a not present value
      * [SpdxConstants.NONE] or [SpdxConstants.NOASSERTION] must be used.
      */
-    val copyrightText: String,
+    val copyrightText: String = SpdxConstants.NOASSERTION,
```
fviernau, line 69: Have you considered to omit "default" values from serialization?

utils/spdx/src/main/kotlin/model/SpdxFile.kt
sschuberth: Actually, both links should link to the JSON specs, which is clearer in this context. Thanks, fixed.

utils/spdx/src/main/kotlin/model/SpdxPackage.kt
sschuberth: I haven't considered that TBH, but looking at other non-empty / non-null default values (like for `licenseInfoInFiles`) we seem to don't exclude these, either. And given the unclarity of the 2.2 specs, it's maybe better to be explicit. Do you agree?

fviernau: As-is when reading a file and writing it, omited values in the input will be filled with default values in the output. I considered this as a slightly negative thing. While at the same time the "require not blank" condition in the `init` block already guarantees that we are actually explicit in the sense that one has to specify to the constructor the value. It's not explicitly written out.

Well, I don't have a strong opinion here, actually I don't know which one is better. You can make the call.

fviernau (approved):

Issue #7225, opened 2023-06-29
scanner: The file lists contain paths following symbolic links


fviernau:
`FileListResolver` must not follow symbolic links. This became visible with the recent update of the `mime-types` example scan in the `orthw` repo.

PR #7226, opened 2023-06-29
fix(scanner): Ignore any symbolic links when creating file lists

fviernau:
Do not follow symbolic links to directories, because that leads to
undesired entries resulting from undesired duplicate traversal. Also,
ignore symbolic links to files, because the file it points to is
included already anyway if it is within the VCS or it is to be ignored
if it is not within.

Note: Programatically creating a symbolic link fails with [1] Operation
      not permitted, which is why this commit does not have unit test
      coverage.

[1] `Path.createSymbolicLinkPointingTo()`

Fixes: #7225.

**Important note:** Please delete existing file lists from your storage to make this change have effect also for
           previously created file list.

### DIFF
--- a/scanner/src/main/kotlin/utils/FileListResolver.kt
+++ b/scanner/src/main/kotlin/utils/FileListResolver.kt
@@ -33,6 +33,7 @@ import org.ossreviewtoolkit.scanner.FileList.FileEntry
 import org.ossreviewtoolkit.scanner.provenance.ProvenanceDownloader
 import org.ossreviewtoolkit.utils.common.FileMatcher
 import org.ossreviewtoolkit.utils.common.VCS_DIRECTORIES
+import org.ossreviewtoolkit.utils.common.isSymbolicLink
 
 internal class FileListResolver(
     private val storage: ProvenanceFileStorage,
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/utils/FileListResolver.kt
```diff
@@ -64,7 +65,7 @@ private val IGNORED_DIRECTORY_MATCHER by lazy {
 
 private fun createFileList(dir: File): FileList {
     val files = dir.walk().onEnter {
-        !IGNORED_DIRECTORY_MATCHER.matches(it.relativeTo(dir).invariantSeparatorsPath)
+        !IGNORED_DIRECTORY_MATCHER.matches(it.relativeTo(dir).invariantSeparatorsPath) && !it.isSymbolicLink()
     }.filter {
         it.isFile
```
sschuberth, line 70: What about symbolic links to files? I guess these should also not be followed: Either they point to a file inside the tree, and the the target file already gets hashed, or they point to outside the tree, in which case they should probably be ignored.

fviernau: Agree.

sschuberth (approved):

PR #7228, opened 2023-06-29
refactor(scanner): Apply the detected license mapping globally

sschuberth:
The `detectedLicenseMapping` is part of the global `ScannerConfiguration`, so also apply it globally on the `ScanSummary` returned by the scanners instead of leaving it to each scanner implementation to apply the mapping.

### DIFF
--- a/plugins/scanners/askalono/src/main/kotlin/Askalono.kt
+++ b/plugins/scanners/askalono/src/main/kotlin/Askalono.kt
@@ -32,7 +32,6 @@ import org.ossreviewtoolkit.model.TextLocation
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.jsonMapper
-import org.ossreviewtoolkit.model.mapLicense
 import org.ossreviewtoolkit.scanner.AbstractScannerWrapperFactory
 import org.ossreviewtoolkit.scanner.CommandLinePathScannerWrapper
 import org.ossreviewtoolkit.scanner.ScanContext
@@ -87,7 +86,7 @@ class Askalono internal constructor(
             val root = jsonMapper.readTree(line)
             root["result"]?.let { result ->
                 val licenseFinding = LicenseFinding(
-                    license = result["license"]["name"].textValue().mapLicense(scannerConfig.detectedLicenseMapping),
+                    license = result["license"]["name"].textValue(),
                     location = TextLocation(
                         // Turn absolute paths in the native result into relative paths to not expose any information.
                         relativizePath(scanPath, File(root["path"].textValue())),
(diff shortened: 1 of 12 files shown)


plugins/scanners/askalono/src/main/kotlin/Askalono.kt
```diff
@@ -87,7 +86,7 @@ class Askalono internal constructor(
             val root = jsonMapper.readTree(line)
             root["result"]?.let { result ->
                 val licenseFinding = LicenseFinding(
-                    license = result["license"]["name"].textValue().mapLicense(scannerConfig.detectedLicenseMapping),
+                    license = result["license"]["name"].textValue(),
```
fviernau, line 89: Do you maybe recall why the mapping had to be applied before writing the data into the storage?

sschuberth: The rationale is mentioned in 8a8626e1eaef45ee9f86f390b760592b09d6b0a1:

    The mapping is being done when creating the LicenseFinding after
    receiving the results from the scanners. This has the advantage that
    ORT has to handle only valid SPDX expressions internally. The
    disadvantage of this approach is, that mapped licenses are stored in the
    ScanResultStorage which might become invalid when the detected license
    mapping in the scanner configuration is changed.

fviernau: Thanks! I still wonder if we can get rid of the situation that the effect of applying `config.yml` gets written into scan result cache. Is this (still) our ultimate goal? 

If so, what if we distinguish between the following type of mapping entries:

1. `valid SPDX ID -> valid SPDX ID`
2. `string (not valid SPDX ID) -> valid SPDX ID`

Then,

1. Looks to me like it is user specific. Such entries could be applied only after reading scan results from storage.
2. Looks to me like some scanner specific fix-up of lack of scanner supporting proper license expression. In  
    particular it seems not user specific, so entries could be identical for all users and thus not configurable. In fact for
    two the need exists only in one scanner (not sure if this is correct). But independently from that, it seem the 
    entries can be hard-coded inside the scanner. 

I'm not sure if above is a reasonable approach. If it was, would it then make more sense to 1. hard code the 
FossID mapping into FossID integration and then 2. refactor this configurable mapping to be applied only after writing to / reading from any storage?

Anyhow, I do not want to annoy you with this bigger topic again, in this rather small change. We could also leave this discussion up to a separate ticket or so.

fviernau (approved):

plugins/scanners/askalono/src/main/kotlin/Askalono.kt
sschuberth: > Is this (still) our ultimate goal?

I have no idea actually 🤷🏻‍♂️ 😉 I was touching this code for pure refactoring purposes of the scanner API, not to imply any upcoming changes to the scan storage semantics.

> _edit: Whatever the approach would be, I just now realized that this PR seems a step in the right direction._

👍🏻

PR #7230, opened 2023-06-30
Generally relativize text locations for scan results

sschuberth:

### DIFF
--- a/model/src/main/kotlin/TextLocation.kt
+++ b/model/src/main/kotlin/TextLocation.kt
@@ -19,6 +19,8 @@
 
 package org.ossreviewtoolkit.model
 
+import java.io.File
+
 import kotlin.math.abs
 import kotlin.math.min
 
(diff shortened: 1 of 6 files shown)


sschuberth:
@fviernau and @mnonnenmacher, I've thought about what we've discussed in today's Kotlin developer meeting about this PR, and while I still agree that it makes sense to ensure relative paths via the individual scanner API, I'd prefer to first merge this as-is (modulo any other review comments) to "collect requirements" (as part of `Scanner.kt`) about that kind of checks / post-processing needs to be done. I have some more smaller refactoring in the pipeline that will probably make easy to move the respective check / post-processing code again.

model/src/main/kotlin/TextLocation.kt
```diff
@@ -90,4 +92,20 @@ data class TextLocation(
...
             else -> min(abs(other.startLine - endLine), abs(startLine - other.endLine))
         }
+
+    /**
+     * Return a [TextLocation] whose path is relative to [basePath], or throw an [IllegalArgumentException] if the paths
+     * have different roots.
+     */
+    fun withRelativePath(basePath: File): TextLocation {
+        val pathAsFile = File(path)
+        if (!pathAsFile.isAbsolute) return this
```
mnonnenmacher, line 102: Should this branch also call `invariantSeparatorsPath`?

sschuberth: Good catch, the corresponding code in `CommandLinePathScannerWrapper` did that as well. Fixed.

model/src/main/kotlin/TextLocation.kt
```diff
@@ -90,4 +92,20 @@ data class TextLocation(
...
         }
+
+    /**
+     * Return a [TextLocation] whose path is relative to [basePath], or throw an [IllegalArgumentException] if the paths
+     * have different roots.
+     */
+    fun withRelativePath(basePath: File): TextLocation {
+        val pathAsFile = File(path)
+
+        val relevantPath = when {
```
mnonnenmacher, line 103: Should this be called `relativePath` or `relativeFile` instead of `relevantPath`?

mnonnenmacher (approved):

PR #7240, opened 2023-07-04
Fix OSV mit working with PURLs

sschuberth:

### DIFF
--- a/advisor/src/funTest/kotlin/OsvFunTest.kt
+++ b/advisor/src/funTest/kotlin/OsvFunTest.kt
@@ -29,6 +29,7 @@ import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.config.OsvConfiguration
 import org.ossreviewtoolkit.model.readValue
+import org.ossreviewtoolkit.model.utils.toPurl
 import org.ossreviewtoolkit.utils.test.getAssetFile
 
 class OsvFunTest : StringSpec({
@@ -77,7 +78,7 @@ class OsvFunTest : StringSpec({
 })
 
 private fun identifierToPackage(id: String): Package =
-    Package.EMPTY.copy(id = Identifier(id))
+    Identifier(id).let { Package.EMPTY.copy(id = it, purl = it.toPurl()) }
 
 private fun createOsv(): Osv =
     Osv("OSV", OsvConfiguration(serverUrl = null))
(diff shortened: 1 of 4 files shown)


clients/osv/src/main/kotlin/OsvService.kt
```diff
@@ -67,7 +67,9 @@ class OsvService(serverUrl: String? = null, httpClient: OkHttpClient? = null) {
             val response = client.getVulnerabilityIdsForPackages(batchRequest).execute()
             val body = response.body()
 
-            if (!response.isSuccessful || body == null) return Result.failure(IOException(response.message()))
+            if (!response.isSuccessful || body == null) {
```
fviernau, line 70: Why is `|| body == null` necessary and / or correct?

advisor/src/funTest/kotlin/OsvFunTest.kt
```diff
@@ -77,7 +78,7 @@ class OsvFunTest : StringSpec({
 })
 
 private fun identifierToPackage(id: String): Package =
-    Package.EMPTY.copy(id = Identifier(id))
+    Identifier(id).let { Package.EMPTY.copy(id = it, purl = it.toPurl()) }
```
fviernau, line 81: Would swapping the order of this and the following commit lead to the situation that all tests pass of all of the three commits of this PR? If so, would it make sense to change the order of the top-most two commits?

fviernau (changes requested):

advisor/src/funTest/kotlin/OsvFunTest.kt
sschuberth: Yes, that would be the case, but I deliberately did not do so as I found it strange to commit a fixup for a commit that I reverted before. That could raise the question whether the fixup should have also been reverted.

clients/osv/src/main/kotlin/OsvService.kt
sschuberth: While theoretically (i.e. on HTTP specification level) a successful response could return a null body, in the context of OSV we always expect to get a (non-nullable) `VulnerabilitiesForPackageBatchResponse`, so handling a `null` body as an error seems to be correct.

In any case, the code was like that before and is out of scope for this change.

But I'll make another change to improve error reporting via `errorBody()`.

advisor/src/main/kotlin/advisors/Osv.kt
```diff
@@ -154,6 +154,7 @@ private fun createRequest(pkg: Package): VulnerabilitiesForPackageRequest? {
 
     if (name.isNotBlank() && pkg.id.version.isNotBlank() && !ecosystem.isNullOrBlank()) {
         return VulnerabilitiesForPackageRequest(
+            // Do not specify the purl here as it is mutually exclusive with the ecosystem.
```
fviernau, line 157: Consider making this comment also explain why it is preferable to use ecosystem.

fviernau (approved):

PR #7244, opened 2023-07-04
Upgrade to Clikt 4

sschuberth:


cli/src/main/kotlin/OrtMain.kt
```diff
@@ -68,7 +67,7 @@ fun main(args: Array<String>) {
     exitProcess(0)
 }
 
-class OrtMain : CliktCommand(name = ORT_NAME, invokeWithoutSubcommand = true) {
+class OrtMain : CliktCommand(name = ORT_NAME, epilog = "* denotes required options.", invokeWithoutSubcommand = true) {
```
fviernau, line 70: What do you think about extracting a constant for `REQUIRED_OPTION_MARKER`, so it gets harder to break? e.g. also used in line 105

cli/src/main/kotlin/OrtMain.kt
```diff
@@ -26,8 +26,7 @@ import com.github.ajalt.clikt.completion.completionOption
 import com.github.ajalt.clikt.core.CliktCommand
 import com.github.ajalt.clikt.core.context
 import com.github.ajalt.clikt.core.subcommands
-import com.github.ajalt.clikt.output.CliktHelpFormatter
-import com.github.ajalt.clikt.output.HelpFormatter
+import com.github.ajalt.clikt.output.MordantHelpFormatter
```
fviernau, line 29: commit: Is it necessary to do the  upgrade and the removal of the helper formatter in one commit?

sschuberth: Not strictly. I've made more fine-granular commits now.

fviernau (approved):

Issue #6947, opened 2023-04-28
SpdxDocumentReporter: Include file findings in the output


fviernau:
Task: Add a reporter specific option to toggle the inclusion of file findings (licenses & copyrights)

MVP: 

1. The generated `SpdxDocumentFile` should include `SpdxFile` instances for all
    files of all dependencies which contain license or copyright findings with the following attributes set
    - `spdxId`: generate
    - `checksum`: the SHA sum of the file
    - `filename`: the name of the file
    - `copyrightText`: All copyright statements which are matched to the file after the license info
                                   resolver has applied configuration: e.g. curations, copyright garbage. 
                                   TBC: should findings which get matched to the root license by `FindingsMatcher` 
                                            be included?
2. The `SpdxPackage.hasFile` property must contain reference to all related `SpdxFile` instances.

PR #7246, opened 2023-07-05
refactor(model): Change the type of unmatched copyrights

fviernau:
Use `ResolvedCopyrightFinding` instead of just `CopyrightFinding` as type, to prepare for an upcoming change which needs to know for each of the entries whether it is excluded by any path exclude.

Part of #6947.

### DIFF
--- a/model/src/main/kotlin/licenses/LicenseInfoResolver.kt
+++ b/model/src/main/kotlin/licenses/LicenseInfoResolver.kt
@@ -125,7 +125,7 @@ class LicenseInfoResolver(
         val filteredDetectedLicenseInfo =
             licenseInfo.detectedLicenseInfo.filterCopyrightGarbage(copyrightGarbageFindings)
 
-        val unmatchedCopyrights = mutableMapOf<Provenance, MutableSet<CopyrightFinding>>()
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/licenses/LicenseInfoResolver.kt
```diff
@@ -225,7 +225,17 @@ class LicenseInfoResolver(
                 }
             }
 
-            unmatchedCopyrights.getOrPut(findings.provenance) { mutableSetOf() } += matchResult.unmatchedCopyrights
+            unmatchedCopyrights.getOrPut(findings.provenance) {
+                mutableSetOf()
+            } += matchResult.unmatchedCopyrights.map { copyrightFinding ->
+                ResolvedCopyrightFinding(
```
sschuberth, line 231: Can we somehow reuse `LicenseInfoResolver.resolveCopyrights()` by making that function public and moving it out of the class context?

fviernau: Good catch. It draw me to `relativeFindingsPath` which I had overlooked before.

sschuberth (approved):

fviernau:
The failed checks (qodana crash + PubFunTest) are unrelated. Merge regardlessly.

PR #7258, opened 2023-07-07
refactor(scanner): Store scanner details in separate properties

sschuberth:
E.g. scan storages may need to access a CLI scanner's name without getting the version, which triggers running the local tool. This was not possible before as getting `ScannerDetails` contains the version. Solve that by storing the detail properties individually and constructing `ScannerDetails` only where needed.

### DIFF
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -149,14 +149,14 @@ class Scanner(
         val filteredScannerOptions = mutableMapOf<String, Options>()
 
         projectScannerWrappers.forEach { scannerWrapper ->
-            scannerConfig.options?.get(scannerWrapper.details.name)?.let { options ->
-                filteredScannerOptions[scannerWrapper.details.name] = scannerWrapper.filterSecretOptions(options)
+            scannerConfig.options?.get(scannerWrapper.name)?.let { options ->
+                filteredScannerOptions[scannerWrapper.name] = scannerWrapper.filterSecretOptions(options)
             }
         }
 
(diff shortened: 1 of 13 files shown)


scanner/src/main/kotlin/Scanner.kt
```diff
@@ -91,9 +92,10 @@ class Scanner(
 
         scannerWrappers.values.flatten().distinct().forEach { scannerWrapper ->
             scannerWrapper.criteria?.let { criteria ->
-                require(criteria.matches(scannerWrapper.details)) {
-                    "The scanner details of scanner '${scannerWrapper.details.name}' must satisfy the configured " +
-                        "criteria for looking up scan storage entries."
+                val details = with(scannerWrapper) { ScannerDetails(name, version, configuration) }
```
mnonnenmacher, line 95: It seems a bit tedious to have use this line everywhere the details are needed, is there a reason you didn't add the details as a lazy property to `ScannerWrapper`?

sschuberth: It felt like duplicating code to have the properties in there both individually and wrapped in `ScannerDetails`... but as you correctly noted we now have some duplication elsewhere. I'll play with the idea to see how the code would look like.

sschuberth: Kotlin says "delegated properties are not allowed on interfaces" 😞

sschuberth: I've introduced a convenience extension property now.

mnonnenmacher (approved):

PR #7260, opened 2023-07-10
fix(helper-cli): Stop failing with an `IllegalStateException`

fviernau:
The code introduced for the default value of `prefix` introduced by [1] is executed before the command line is parsed. So, the `provider` property is accessed before it is initialized. This cause an `InvalidStateException` each time `orthw` is executed. No matter which (sub-)command is run.

Revert [1] to fix this.

[1] ee63c878aa6cc6964d77eb862a723247f5002fe3.


sschuberth (approved):
Hmm, I'll research a substitute for this functionality.

PR #7268, opened 2023-07-10
Dockerfile cleanups

sschuberth:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -453,22 +453,22 @@ RUN mkdir -p /opt/php/bin \
     && curl -ksS https://getcomposer.org/installer | php -- --install-dir=/opt/php/bin --filename=composer --$COMPOSER_VERSION
 
 # nuget-inspector
-ENV NUGET_INSPECTOR_ROOT=/opt/nuget-inspector
-ENV NUGET_INSPECTOR_HOME=$NUGET_INSPECTOR_ROOT/bin
-ENV NUGET_DOTNET_HOME=$NUGET_INSPECTOR_ROOT/dotnet
+ENV NUGET_INSPECTOR_HOME=/opt/nuget-inspector
+ENV NUGET_INSPECTOR_BIN=$NUGET_INSPECTOR_HOME/bin
+ENV DOTNET_HOME=$NUGET_INSPECTOR_HOME/dotnet
 
-ENV PATH=$PATH:$NUGET_DOTNET_HOME:$NUGET_DOTNET_HOME/tools:$NUGET_INSPECTOR_HOME
+ENV PATH=$PATH:$DOTNET_HOME:$DOTNET_HOME/tools:$NUGET_INSPECTOR_BIN
 
 # Note: We are not installing a dotnet package directly because
 # debian packages from Ubuntu and Microsoft are incomplete
-RUN mkdir -p $NUGET_DOTNET_HOME \
+RUN mkdir -p $DOTNET_HOME \
     && curl -L https://aka.ms/dotnet/6.0/dotnet-sdk-linux-x64.tar.gz \
-    | tar -C $NUGET_DOTNET_HOME -xz
+    | tar -C $DOTNET_HOME -xz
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -130,10 +130,10 @@ ARG PYTHON_VERSION=3.10.8
 ARG PYENV_GIT_TAG=v2.3.7
 
 ENV PYENV_ROOT=/opt/python
-ENV PATH=$PATH:${PYENV_ROOT}/shims:${PYENV_ROOT}/bin
+ENV PATH=$PATH:$PYENV_ROOT/shims:$PYENV_ROOT/bin
```
mnonnenmacher, line 133: I haven't done much shell programming in recent years, but I remember that it was considered a best practice to always use braces for consistency and readability, and to reduce the possibility of mistakes. Sometimes they might not be required, but to the (unexperienced) user it might still be ambiguous how to read an expression.

mnonnenmacher: I looked up the related Shellcheck rule which has an example and rationale for why it might be good to keep the braces: https://www.shellcheck.net/wiki/SC2250

sschuberth: > I looked up the related Shellcheck rule which has an example and rationale for why it might be good to keep the braces: https://www.shellcheck.net/wiki/SC2250

Thanks for the reference, but even Shellcheck says

> This is an [optional](https://www.shellcheck.net/wiki/optional) suggestion. It must be explicitly enabled with a [directive](https://www.shellcheck.net/wiki/directive)

So it's *disabled* by default. And I actually disagree regarding the readability: To mee fewer special chars are less distracting than having more of them. Also, syntax-highlighting guards even the inexperienced user against escaping mistakes.

mnonnenmacher: Yes, it's optional, but it's still a suggestion. But I'm fine with removing them, I just wanted to make sure it's considered.

Dockerfile
```diff
@@ -196,7 +196,7 @@ RUN rbenv install $RUBY_VERSION -v \
     && gem install bundler cocoapods:$COCOAPODS_VERSION
 
 FROM scratch AS ruby
-COPY --from=rubybuild $RBENV_ROOT $RBENV_ROOT
+COPY --from=rubybuild /opt/rbenv /opt/rbenv
```
mnonnenmacher, line 199: Except for the issue with the `DART_SDK` variable was any of those not working as intended? I would have expected that those variables were always defined in the respective parent image? I just want to understand what's the benefit of the change as it seems to me it's making maintenance harder.

sschuberth: > Except for the issue with the `DART_SDK` variable was any of those not working as intended?

As I was trying to explain in the commit message, *all* of these lines were not working as intended:

- `FROM scratch AS ruby` starts a new stage with "scratch" (the empty image) as the parent. I.e. the parent image has no environment variables defined.
- `COPY --from=rubybuild $RBENV_ROOT $RBENV_ROOT` refers to "RBENV_ROOT" which is unset in this context. The "rubybuild" image used by `--from` *does* have "RBENV_ROOT" defined, but the source for a copy does not count as a parent / ancestor image.
- Thus the line expands to `COPY --from=rubybuild "" ""` which copies *everything*, not just the binaries from `/opt/rbenv`.

Dockerfile
sschuberth: As a side note, Kotlin inspection has exactly the opposite hint: It suggests to remove superfluous curly braces from string interpolations like `"${it}"`. I prefer the Kotlin style :wink:

Dockerfile
sschuberth: BTW, I've verified this by looking at the generated layers / stages with [dive](https://github.com/wagoodman/dive).

mnonnenmacher: > BTW, I've verified this by looking at the generated layers / stages with [dive](https://github.com/wagoodman/dive).

Nice tool!

mnonnenmacher: Thanks for the explanation, now it makes sense to me.

mnonnenmacher (approved):

Issue #6972, opened 2023-05-05
Allow multiple package-configurations with same ID and provenance


MarcelBochtler:
At the moment ORT does not allow package-configurations with the same ID and provenance, and throws an exception:

https://github.com/oss-review-toolkit/ort/blob/4e2f276fa5a459d05f30a69d34e6c38ae3f3135b/model/src/main/kotlin/utils/SimplePackageConfigurationProvider.kt#L74-L79

As package-configurations can be set up in multiple locations (`.ort.yml` and the directory given to `--package-configurations-dir`), it should be allowed to have multiple package-configurations with the same matchers.

For curations, this already works as expected, and the logic is to use the one from `.ort.yml` first, and the one configured for ORT afterward.

sschuberth:
Sounds like a duplicate of https://github.com/oss-review-toolkit/ort/issues/6873, or?

sschuberth:
As discussed with @MarcelBochtler, please verify whether your use-case still makes use of `SimplePackageConfigurationProvider` and its check:

https://github.com/oss-review-toolkit/ort/blob/811daef3d35abb440a1ad044bcbf14738fc95767/plugins/package-configuration-providers/api/src/main/kotlin/SimplePackageConfigurationProvider.kt#L47-L58

sschuberth:
I can confirm that the issue is still there. With the `OrtConfig` package configuration provider enabled in addition to our own one (implemented in a plugin) I get:

```
10:33:18.323 [main] INFO  org.ossreviewtoolkit.plugins.commands.api.utils.ExtensionsKt - Read ORT result from 'current-result.yml' (71.69 MiB) in 4.012595895s.
Exception in thread "main" java.lang.IllegalArgumentException: There must be at most one package configuration per Id and provenance, but found multiple for:
  Key(id=Identifier(type=Maven, namespace=org.apache.commons, name=commons-compress, version=1.26.2), sourceArtifactUrl=null, vcsMatcher=VcsMatcher(type=Git, url=https://gitbox.apache.org/repos/asf/commons-compress.git, revision=95727006cac0892c654951c4e7f1db142462f22a)).
	at org.ossreviewtoolkit.plugins.packageconfigurationproviders.api.SimplePackageConfigurationProviderKt.checkAtMostOneConfigurationPerIdAndProvenance(SimplePackageConfigurationProvider.kt:54)
[...]

PR #7270, opened 2023-07-11
package-configuration provider plugin

MarcelBochtler:
Move the package-configuration provider to a separate plugin.

This allows adding more providers through the plugin API, for instance adding the package-configurations from [ort-config](https://github.com/oss-review-toolkit/ort-config)

### DIFF
--- a/model/src/main/kotlin/config/OrtConfiguration.kt
+++ b/model/src/main/kotlin/config/OrtConfiguration.kt
@@ -85,9 +85,9 @@ data class OrtConfiguration(
      * providers: Providers that appear earlier in the list can overwrite curations for the same package from providers
      * that appear later in the list.
      */
-    val packageCurationProviders: List<PackageCurationProviderConfiguration> = listOf(
-        PackageCurationProviderConfiguration(type = "DefaultDir"),
-        PackageCurationProviderConfiguration(type = "DefaultFile")
+    val packageCurationProviders: List<ProviderPluginConfiguration> = listOf(
+        ProviderPluginConfiguration(type = "DefaultDir"),
+        ProviderPluginConfiguration(type = "DefaultFile")
     ),
 
     /**
--- a/model/src/main/kotlin/config/OrtConfiguration.kt
+++ b/model/src/main/kotlin/config/OrtConfiguration.kt
@@ -30,6 +30,7 @@ import org.apache.logging.log4j.kotlin.Logging
 
 import org.ossreviewtoolkit.model.Severity
 import org.ossreviewtoolkit.utils.common.EnvironmentVariableFilter
+import org.ossreviewtoolkit.utils.ort.ORT_PACKAGE_CONFIGURATIONS_DIRNAME
 import org.ossreviewtoolkit.utils.ort.ORT_PACKAGE_CURATIONS_DIRNAME
 import org.ossreviewtoolkit.utils.ort.ORT_PACKAGE_CURATIONS_FILENAME
 
@@ -78,6 +79,14 @@ data class OrtConfiguration(
      * The license file patterns.
      */
     val licenseFilePatterns: LicenseFilePatterns = LicenseFilePatterns.DEFAULT,
+
+    /**
+     * The package configuration providers to use. Defaults to the provider [ORT_PACKAGE_CONFIGURATIONS_DIRNAME]
+     * configuration location.
+     */
+    val packageConfigurationProviders: List<ProviderPluginConfiguration> = listOf(
+        ProviderPluginConfiguration(type = "DefaultDir")
+    ),
 
     /**
      * The package curation providers to use. Defaults to providers for the default [ORT_PACKAGE_CURATIONS_FILENAME] and
--- a/model/src/main/kotlin/config/ProviderPluginConfiguration.kt
+++ b/model/src/main/kotlin/config/ProviderPluginConfiguration.kt
@@ -23,25 +23,28 @@ import com.sksamuel.hoplite.ConfigAlias
 
 import org.ossreviewtoolkit.utils.common.Plugin
 
-data class PackageCurationProviderConfiguration(
+/**
+ * The configuration of provider plugins.
+ */
+data class ProviderPluginConfiguration(
     /**
-     * The [type][Plugin.type] of the package curation provider.
+     * The [type][Plugin.type] of the provider.
      */
     @ConfigAlias("name")
     val type: String,
 
     /**
-     * A unique identifier for the package curation provider.
+     * A unique identifier for the provider.
      */
     val id: String = type,
 
     /**
-     * Whether this curation provider is enabled.
+     * Whether this provider is enabled.
      */
     val enabled: Boolean = true,
 
     /**
-     * The configuration of the package curation provider. See the specific implementation for available configuration
+     * The configuration of the provider. See the specific implementation for available configuration
      * options.
      */
     val config: Map<String, String> = emptyMap()
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -36,6 +36,16 @@ ort:
     patentFilenames: ['patents']
     rootLicenseFilenames: ['readme*']
 
+  # Package configurations have to be unique by ID and provenance. Ensure that different providers do not provide
+  # configurations for the same package, see https://github.com/oss-review-toolkit/ort/issues/6972 for details.
+  packageConfigurationProviders:
+  - type: DefaultDir
+  - type: Dir
+    id: SomeConfigDir
+    config:
+      path: '/some-path/'
+      mustExist: true
+
   # Providers are listed from highest to lower priority. Technically, they are applied in reverse order: The provider
   # with the highest priority is applied last, so it can override any previously applied curations.
   packageCurationProviders:
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -37,7 +37,7 @@ ort:
     rootLicenseFilenames: ['readme*']
 
   # Providers are listed from highest to lower priority. Technically, they are applied in reverse order: The provider
-  # with the highest priority is applied last, so it can overwrite any previously applied curations.
+  # with the highest priority is applied last, so it can override any previously applied curations.
   packageCurationProviders:
   - type: DefaultFile
   - type: DefaultDir
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -39,33 +39,33 @@ ort:
   # Providers are listed from highest to lower priority. Technically, they are applied in reverse order: The provider
   # with the highest priority is applied last, so it can overwrite any previously applied curations.
   packageCurationProviders:
-    - type: DefaultFile
-    - type: DefaultDir
-    - type: File
-      id: SomeCurationsFile
-      config:
-        path: '/some-path/curations.yml'
-        mustExist: true
-    - type: File
-      id: SomeCurationsDir
-      config:
-        path: '/some-path/curations-dir'
-        mustExist: false
-    - type: OrtConfig
-      enabled: '${USE_ORT_CONFIG_CURATIONS:-true}'
-    - type: ClearlyDefined
-      config:
-        serverUrl: 'https://api.clearlydefined.io'
-        minTotalLicenseScore: 80
-    - type: SW360
-      config:
-        restUrl: 'https://your-sw360-rest-url'
-        authUrl: 'https://your-authentication-url'
-        username: username
-        password: password
-        clientId: clientId
-        clientPassword: clientPassword
-        token: token
+  - type: DefaultFile
+  - type: DefaultDir
+  - type: File
+    id: SomeCurationsFile
+    config:
+      path: '/some-path/curations.yml'
+      mustExist: true
+  - type: File
+    id: SomeCurationsDir
+    config:
+      path: '/some-path/curations-dir'
+      mustExist: false
+  - type: OrtConfig
+    enabled: '${USE_ORT_CONFIG_CURATIONS:-true}'
+  - type: ClearlyDefined
+    config:
+      serverUrl: 'https://api.clearlydefined.io'
+      minTotalLicenseScore: 80
+  - type: SW360
+    config:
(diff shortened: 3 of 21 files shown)


model/src/main/kotlin/config/PackageConfigurationProviderConfiguration.kt
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model.config
+
+import org.ossreviewtoolkit.utils.common.Plugin
+
+data class PackageConfigurationProviderConfiguration(
```
sschuberth, line 24: Please also document the class itself.

sschuberth: Can we reuse / generalize `PackageCurationProviderConfiguration` here?

plugins/package-configuration-providers/api/src/main/kotlin/PackageConfigurationProviderFactory.kt
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * The extension point for [PackageConfigurationProvider]s.
+ */
+interface PackageConfigurationProviderFactory<CONFIG> : ConfigurablePluginFactory<PackageConfigurationProvider> {
+    companion object {
+        val ALL = Plugin.getAll<PackageConfigurationProviderFactory<*>>()
+
+        /**
+         * Create a new (identifier, provider instance) tuple for each
+         * [enabled][PackageCurationProviderConfiguration.enabled] provider configuration in [configurations] ordered
```
sschuberth, line 38: Reference to wrong "curation" class (see also the import).

plugins/package-configuration-providers/file/src/main/kotlin/FilePackageConfigurationProvider.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.ossreviewtoolkit.model.utils.PackageConfigurationProvider
+import org.ossreviewtoolkit.plugins.packageconfigurationproviders.api.PackageConfigurationProviderFactory
+import org.ossreviewtoolkit.plugins.packageconfigurationproviders.api.SimplePackageConfigurationProvider
+import org.ossreviewtoolkit.utils.common.getDuplicates
+import org.ossreviewtoolkit.utils.ort.ORT_PACKAGE_CONFIGURATIONS_DIRNAME
+import org.ossreviewtoolkit.utils.ort.ortConfigDirectory
+
+class FilePackageConfigurationProviderConfig(
+    /**
+     * The path of the package curation file or directory.
```
sschuberth, line 39: s/curation/configuration/

plugins/package-configuration-providers/file/src/main/kotlin/FilePackageConfigurationProvider.kt
```diff
@@ -0,0 +1,128 @@
...
+                }
+
+                configurations.mapTo(allConfigurations) { it to configurationFile }
+            }
+
+            val duplicates = allConfigurations.getDuplicates { it.first }
+            if (duplicates.isNotEmpty()) {
+                val duplicatesInfo = buildString {
+                    duplicates.forEach { (curation, origins) ->
+                        appendLine("Curation for '${curation.id.toCoordinates()}' found in all of:")
```
sschuberth, line 113: s/curation/configuration/

model/src/main/kotlin/config/ProviderConfiguration.kt
```diff
@@ -23,25 +23,28 @@ import com.sksamuel.hoplite.ConfigAlias
 
 import org.ossreviewtoolkit.utils.common.Plugin
 
-data class PackageCurationProviderConfiguration(
+/**
+ * The configuration of a package curation and package configuration provider.
```
sschuberth, line 27: I wonder whether we should hard-code "curation" and "configuration" here, as they class might also be reused in the context of e.g. `AdviceProviders` (also see https://github.com/oss-review-toolkit/ort/pull/6613).

model/src/main/kotlin/config/ProviderConfiguration.kt
```diff
@@ -23,25 +23,28 @@ import com.sksamuel.hoplite.ConfigAlias
...
+ */
+data class ProviderConfiguration(
```
sschuberth, line 29: Also, is the class name too generic now? Should we maybe mention the plugin context by naming this e.g. `ProviderPluginConfiguration`, and maybe even moving it next to `PluginManager.kt`?

Note that I'm just thinking out aloud; I do not want to imply it's the right thing to do, but just start a discussion.

model/src/main/kotlin/config/OrtConfiguration.kt
```diff
@@ -79,6 +79,12 @@ data class OrtConfiguration(
      */
     val licenseFilePatterns: LicenseFilePatterns = LicenseFilePatterns.DEFAULT,
 
+    /**
+     * The package configuration providers to use. Does not define a default, because this is done by the
+     * EvaluatorCommand's `packageConfigurationsDir` and `packageConfigurationsFile` CLI options.
```
sschuberth, line 84: This seems like a weird rationale to me to not define a default, esp. when comparing to `packageCurationProviders` below. Do you plan to make a follow-up that removes the evaluator's CLI options in favor of these config options?

model/src/main/kotlin/config/PackageConfigurationProviderConfiguration.kt
fviernau: I believe that this falls under the category of too early refactoring, because I believe there will be provider specific configuration. E.g. we discussed already for curations, that one may be interested in consuming curations only of a certain type. Also, that is the benefit of using same class for different purposes? (If there should be code re-use potential, re-use cannot be achieved also with separate classes)

model/src/main/kotlin/config/OrtConfiguration.kt
fviernau: Note: `orthw` requires that it is possible to override `packageConfigurations` and `packageCurations` in the evaluator.
           I believe there should be a way which is simpler than generating a `config.yml` for ORT.

plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
```diff
@@ -273,19 +274,28 @@ class EvaluatorCommand : OrtCommand(
...
+            val providerFromOption = DirectoryPackageConfigurationProvider(packageConfigurationsDir)
+            add(providerFromOption)
+
+            val packageConfigurationProviders =
+                PackageConfigurationProviderFactory.create(ortConfig.packageConfigurationProviders)
+            addAll(packageConfigurationProviders.map { it.second })
         }
 
+        val packageConfigurationProvider =
+            CompositePackageConfigurationProvider(*enabledPackageConfigurationProviders.toTypedArray())
```
fviernau, line 297: For `PackageCurations` we had the requirement to associate the package configurations with the ID of the provider (in order to re-apply the curations for certain providers only).
Would that requirement also be here in an analog way?

I'm asking that because for `PackageCurations` the `Composite*Provider` had to be eliminated since it prevented exactly that.

model/src/main/kotlin/config/PackageConfigurationProviderConfiguration.kt
sschuberth: I guess that question boils down to whether provider-specific (not: plugin-specific) configuration should go to the existing `config` map, or whether a dedicated property should be introduced.

model/src/main/resources/reference.yml
```diff
@@ -36,6 +36,15 @@ ort:
     patentFilenames: ['patents']
     rootLicenseFilenames: ['readme*']
 
+  # Package configurations have to be unique by ID and provenance. Ensure that different providers do not provide
+  # configurations for the same package, see https://github.com/oss-review-toolkit/ort/issues/6972 for details.
```
fviernau, line 40: I believe that the requirement "Ensure that different providers do not provide configurations for the same package" is something we cannot ask, because this cannot be done easily in an automated way. 

How about the following approach: 
1. Add a map 'provider-name -> Set<PackageConfiguration>' to `OrtResult.resolvedConfiguration`.
2. Then consumption of resolved configuration is adjusted to `merge` the path excludes and `license finding curations` of all providers, thereby removing the requirement that there is only one package configuration across all providers applicable for a package.

fviernau (changes requested):

model/src/main/kotlin/config/PackageConfigurationProviderConfiguration.kt
fviernau: > I guess that question boils down to whether provider-specific (not: plugin-specific) configuration should go to the existing `config` map, or whether a dedicated property should be introduced.

Oh right, I missed that. So, then I'm fine with just staying with the current approach (no dedicated properties).

model/src/main/kotlin/config/ProviderConfiguration.kt
MarcelBochtler: > Also, is the class name too generic now?

Yes, I also thought that it might be too generic.

> Should we maybe mention the plugin context by naming this e.g. ProviderPluginConfiguration, and maybe even moving it next to PluginManager.kt?

I like `ProviderPluginConfiguration`. But I think we should keep it in the `model` module, as this is clearly a model, and moving it to `utils` would also break existing curation provider plugins.

model/src/main/kotlin/config/OrtConfiguration.kt
MarcelBochtler: > Do you plan to make a follow-up that removes the evaluator's CLI options in favor of these config options?

Yes I thought about that, but didn't want to break backwards compatibility for this initial introduction of the package-configuration plugin system.
The `orthw` requirement is another good reason to keep the CLI option for now.

model/src/main/kotlin/config/ProviderConfiguration.kt
MarcelBochtler: Renamed the class.

model/src/main/resources/reference.yml
MarcelBochtler: > I believe that the requirement "Ensure that different providers do not provide configurations for the same package" is something we cannot ask, because this cannot be done easily in an automated way.

I agree that we cannot ask that long-term.
However, this is how ORT currently works with package-configurations. There cannot be package-configurations with the same ID and source in the .ort.yml and the global package-configurations.

Are you saying that the implementation of #6972 is a precondition for this PR?

plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
MarcelBochtler: > in order to re-apply the curations for certain providers only

I'm not aware of that functionality. Why is it required for curations?
I'd need to understand that to be able to say if it would also be necessary for package-configurations.

mnonnenmacher (changes requested):

model/src/main/resources/reference.yml
mnonnenmacher: As this change is only about making package configuration providers plugins it should not have to address issues which already existed previously.

model/src/main/kotlin/config/ProviderPluginConfiguration.kt
```diff
@@ -23,25 +23,28 @@ import com.sksamuel.hoplite.ConfigAlias
 
 import org.ossreviewtoolkit.utils.common.Plugin
 
-data class PackageCurationProviderConfiguration(
+/**
+ * The configuration of plugin providers.
+ */
+data class ProviderPluginConfiguration(
```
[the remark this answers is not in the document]
sschuberth: I'd clearly vote for the former 😸

model/src/main/resources/reference.yml
fviernau: I'm ok to leave this out for now.

Anyhow, am I right that ORT previously supported only a single package configuration provider, while as of this change it allows multiple? (Asking because of above argumentation: "...should not have to address issues which already existed previously")

model/src/main/kotlin/config/OrtConfiguration.kt
MarcelBochtler: > if the evaluator command options are set, they should override the providers from the configuration file.

Yes, I think from a user's perspective that makes the most sense.

model/src/main/resources/reference.yml
MarcelBochtler: > am I right that ORT previously supported only a single package configuration provider, while as of this change it allows multiple?

Package-Configurations from two different sources were already applied before this PR (.ort.yml, and the one from the CLI option), if the feature was enabled:
https://github.com/oss-review-toolkit/ort/blob/dc870389cd61b4208f5252cf38c6a873d0de6e41/plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt#L277C6-L280

plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
```diff
@@ -273,19 +272,29 @@ class EvaluatorCommand : OrtCommand(
...
+            if (ortConfig.enableRepositoryPackageConfigurations) {
+                add(SimplePackageConfigurationProvider(repositoryPackageConfigurations))
+            } else {
+                if (repositoryPackageConfigurations.isNotEmpty()) {
+                    logger.info { "Local package configurations were not applied because the feature is not enabled." }
+                }
             }
 
-            DirectoryPackageConfigurationProvider(packageConfigurationsDir)
+            if (packageConfigurationsDir != null) {
```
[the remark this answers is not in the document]
MarcelBochtler: > What about `packageConfigurationsFile`?

That does no longer exist since #6831.
I guess you misread `packageCurationsFile`.

mnonnenmacher (approved):

mnonnenmacher:
@fviernau Are all your comments addressed?

fviernau:
> @fviernau Are all your comments addressed?

yes, LGTM.

PR #5092, opened 2022-02-23
analyzer: Add support for Swift Package Manager

UgniusV:
Swift Package Manager is still very young and no other open-source scanners support it. I thought it would benefit the community if ORT had SPM support. This commit adds a new analyzer that parses `Package.resolved` files and creates results that are enriched with as much data as possible.

PR #7271, opened 2023-07-11
Add initial support for the Swift Package Manager 

sschuberth:


MarcelBochtler (approved):
LGTM, I also tested the implementation on my Macbook with swift version `5.8.1`, which worked as expected.

sschuberth:
> LGTM, I also tested the implementation on my Macbook with swift version `5.8.1`, which worked as expected.

I also tested this after-the-fact on Linux with the [Vapor project](https://github.com/vapor/vapor) mentioned [here](https://github.com/oss-review-toolkit/ort/pull/5092#issuecomment-1052778698), and it successfully produced an analyzer result with a good-looking dependency tree.

plugins/package-managers/spm/src/funTest/assets/projects/synthetic/spm-app/Package.resolved
```diff
@@ -0,0 +1,48 @@
+{
```
fviernau, line 1: @UgniusV  I'm trying to enhance the implementation and I'm surprised by the entries in this lockfile.
Can you please explain how this file has been created? 

In particular, has the file been generated by the `swift` command? Has it been modified or been crafted by hand?
(I'm after the combination of `branch` `revision` `version`. E.g. which ones can be present. After playing around a while with `spm` I was unable to make it generate anything like this.

fviernau: If you could provide instructions how to reproduce the creation steps of this file...

PR #7276, opened 2023-07-12
fix(scanner): Filter scan results by scanner name

### LINKED ISSUE #7231: New detected licenses appears for SPDX project with submodule
Our customer has a repository containing a `project.spdx.yml` file. In this file, Git submodules, pointing to third party repositories, are referenced as packages.

In this repository, the customer defines a `license_finding_curations` in a `package_configurations` for one of these referenced packages. Thus, it should be possible to prevent rule violation for code in the third party repositories.

Unfortunately, since the 20.06, there `packages_configuration` are not applied anymore and the customer get some rule violations.

A minimal project reproducing the setup:
https://github.com/nnobelis/spdx-project

Last successful build: version 47bd691..

nnobelis:
When a project scanner is configured alongside a package scanner, the scan results could be duplicated if the provenances of those packages are overlapping (e.g. in the case of repositories with Git submodules). This commit addresses the bug by constructing a map of packages with their scanners and filtering the scan results using this map.

Fixes #7231.

**Breaking change: OrtResult deserialization is not backwards compatible.**

### DIFF
--- a/model/src/main/kotlin/ScannerRun.kt
+++ b/model/src/main/kotlin/ScannerRun.kt
@@ -29,6 +29,7 @@ import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.utils.FileListSortedSetConverter
 import org.ossreviewtoolkit.model.utils.ProvenanceResolutionResultSortedSetConverter
 import org.ossreviewtoolkit.model.utils.ScanResultSortedSetConverter
+import org.ossreviewtoolkit.model.utils.ScannersMapConverter
 import org.ossreviewtoolkit.model.utils.getKnownProvenancesWithoutVcsPath
 import org.ossreviewtoolkit.model.utils.mergeScanResultsByScanner
 import org.ossreviewtoolkit.model.utils.prependPath
@@ -73,6 +74,12 @@ data class ScannerRun(
     val scanResults: Set<ScanResult>,
 
     /**
+     * The names of the scanners which have been used to scan the package.
+     */
+    @JsonSerialize(converter = ScannersMapConverter::class)
+    val scanners: Map<Identifier, Set<String>>,
+
+    /**
      * The list of files for each resolved provenance.
      */
     @JsonSerialize(converter = FileListSortedSetConverter::class)
@@ -90,7 +97,8 @@ data class ScannerRun(
             config = ScannerConfiguration(),
             provenances = emptySet(),
             scanResults = emptySet(),
-            files = emptySet()
+            files = emptySet(),
+            scanners = emptyMap()
         )
     }
 
@@ -192,6 +200,8 @@ data class ScannerRun(
 
         val scanResultsByPath = resolutionResult.getKnownProvenancesWithoutVcsPath().mapValues { (_, provenance) ->
             scanResultsByProvenance[provenance].orEmpty()
+        }.mapValues { (_, scanResults) ->
+            scanResults.filter { it.scanner.name in scanners[id].orEmpty() }
         }
 
         // TODO: Handle the case of incomplete scan results (per scanner), e.g. propagate an issue.
--- a/model/src/main/kotlin/utils/SortedSetConverters.kt
+++ b/model/src/main/kotlin/utils/SortedSetConverters.kt
@@ -28,6 +28,7 @@ import java.util.SortedSet
 import org.ossreviewtoolkit.model.ArtifactProvenance
 import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.FileList
+import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageReference
@@ -72,6 +73,11 @@ class ProjectSortedSetConverter : StdConverter<Set<Project>, SortedSet<Project>>
 class ProvenanceResolutionResultSortedSetConverter :
     StdConverter<Set<ProvenanceResolutionResult>, SortedSet<ProvenanceResolutionResult>>() {
     override fun convert(value: Set<ProvenanceResolutionResult>) = value.toSortedSet(compareBy { it.id })
+}
+
+class ScannersMapConverter : StdConverter<Map<Identifier, Set<String>>, Map<Identifier, Set<String>>>() {
+    override fun convert(value: Map<Identifier, Set<String>>) =
+        value.mapValues { it.value.toSortedSet() }.toSortedMap()
 }
 
 /** Do not convert to SortedSet in order to not require a comparator consistent with equals */
--- a/scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
+++ b/scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.scanner.scanners
+
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.collections.containExactly
+import io.kotest.matchers.should
+
+import org.ossreviewtoolkit.model.AnalyzerResult
+import org.ossreviewtoolkit.model.AnalyzerRun
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.PackageType
+import org.ossreviewtoolkit.model.Project
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.utils.test.shouldNotBeNull
+
+private val PROJECT_ID = Identifier("Dummy", "", "project", "1.0.0")
+private val PACKAGE_ID = Identifier("Dummy", "", "pkg1", "1.0.0")
+
+class MultipleScannersTest : WordSpec({
+    "Scanning a project and a package with overlapping provenance and non overlapping scanners" should {
+        val analyzerResult = createAnalyzerResult()
+
+        val scannerWrappers = mapOf(
+            PackageType.PROJECT to listOf(DummyScanner("Dummy2")),
+            PackageType.PACKAGE to listOf(DummyScanner("Dummy1"))
+        )
+        val ortResult = createScanner(scannerWrappers).scan(analyzerResult, skipExcluded = false, emptyMap())
+
+        "return scan results with non-overlapping scanners" {
+            ortResult.scanner.shouldNotBeNull {
+                val results = getScanResults(PROJECT_ID)
+                results.map { it.scanner.name }.toSet() should containExactly("Dummy2")
+                val results2 = getScanResults(PACKAGE_ID)
+                results2.map { it.scanner.name }.toSet() should containExactly("Dummy1")
+            }
+        }
+    }
+})
+
+private fun createAnalyzerResult(): OrtResult {
+    val vcs = VcsInfo(
+        type = VcsType.GIT,
+        url = "https://github.com/oss-review-toolkit/ort-test-data-scanner.git",
+        revision = "97d57bb4795bc41f496e1a8e2c7751cefc7da7ec",
+        path = ""
+    )
+
+    val pkg = Package.EMPTY.copy(
+        id = PACKAGE_ID,
+        vcs = vcs,
+        vcsProcessed = vcs.normalize()
+    )
+
+    val project = Project.EMPTY.copy(
+        id = PROJECT_ID,
+        vcs = vcs,
+        vcsProcessed = vcs.normalize()
+    )
+
+    val analyzerRun = AnalyzerRun.EMPTY.copy(
+        result = AnalyzerResult.EMPTY.copy(
+            projects = setOf(project),
+            packages = setOf(pkg)
+        )
+    )
+
+    return OrtResult.EMPTY.copy(analyzer = analyzerRun)
+}
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -169,7 +169,8 @@ class Scanner(
             config = filteredScannerConfig,
             provenances = projectResults.provenances + packageResults.provenances,
             scanResults = projectResults.scanResults + packageResults.scanResults,
-            files = projectResults.files + packageResults.files
+            files = projectResults.files + packageResults.files,
+            scanners = projectResults.scanners + packageResults.scanners
         )
 
         return ortResult.copy(scanner = scannerRun)
@@ -238,11 +239,15 @@ class Scanner(
             }
         }
 
+        val scannerNames = scanners.mapTo(mutableSetOf()) { it.name }
+        val scannerNamesByPackageId = packages.associateBy({ it.id }) { scannerNames }
+
         return ScannerRun.EMPTY.copy(
             config = scannerConfig,
             provenances = provenances,
             scanResults = scanResults,
-            files = files
+            files = files,
+            scanners = scannerNamesByPackageId
         )
     }
 
--- a/utils/test/src/main/kotlin/Utils.kt
+++ b/utils/test/src/main/kotlin/Utils.kt
@@ -188,6 +188,9 @@ fun scannerRunOf(vararg pkgScanResults: Pair<Identifier, List<ScanResult>>): Sca
             }
         )
     }
+    val scanners = pkgScanResults.associate { (id, scanResultsForId) ->
+        id to scanResultsForId.mapTo(mutableSetOf()) { it.scanner.name }
+    }
 
     return ScannerRun.EMPTY.copy(
         provenances = pkgScanResultsWithKnownProvenance.mapTo(mutableSetOf()) { (id, scanResultsForId) ->
@@ -202,6 +205,7 @@ fun scannerRunOf(vararg pkgScanResults: Pair<Identifier, List<ScanResult>>): Sca
             )
         },
         scanResults = scanResults,
-        files = files
+        files = files,
+        scanners = scanners
     )
 }
--- a/cli/src/funTest/assets/semver4j-ort-result.yml
+++ b/cli/src/funTest/assets/semver4j-ort-result.yml
@@ -201,6 +201,15 @@ scanner:
         revision: "r4.12"
         path: ""
       resolved_revision: "64155f8a9babcfcf4263cf4d08253a1556e75481"
+  scanners:
+    "Maven:com.vdurmont:semver4j:3.1.0":
+    - "ScanCode"
+    "Maven:junit:junit:4.12":
+    - "ScanCode"
+    "Maven:org.hamcrest:hamcrest-core:1.3":
+    - "ScanCode"
+    "Maven:org.mockito:mockito-all:1.10.19":
+    - "ScanCode"
   scan_results:
   - provenance:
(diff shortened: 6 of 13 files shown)


nnobelis:
@fviernau The new test depends on my repository https://github.com/nnobelis/spdx-project which is not nice.

I saw you were using https://github.com/oss-review-toolkit/ort-test-data-scanner in your tests. Should we fork my repo in ORT or something ?

sschuberth:
> Fixes 7231.

Missing `#` before the issue number.

model/src/main/kotlin/ScannerRun.kt
```diff
@@ -72,6 +73,12 @@ data class ScannerRun(
     @JsonSerialize(converter = ScanResultSortedSetConverter::class)
     val scanResults: Set<ScanResult>,
 
+    /**
+     * The scanners for each scanned package.
+     */
+    @JsonSerialize(converter = ScannersMapConverter::class)
```
sschuberth, line 79: If the goal is to just sort the keys in the map by alphabetic string representation, using `@JsonPropertyOrder(alphabetic = true)` should be enough.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -111,12 +111,16 @@ class Scanner(
...
         val projectScannerWrappers = scannerWrappers[PackageType.PROJECT].orEmpty()
         val packageScannerWrappers = scannerWrappers[PackageType.PACKAGE].orEmpty()
+        val scannersByPackages = mutableMapOf<Identifier, Set<String>>()
 
         val projectResults = if (projectScannerWrappers.isNotEmpty()) {
             val packages = ortResult.getProjects(skipExcluded).mapTo(mutableSetOf()) { it.toPackage() }
 
             logger.info { "Scanning ${packages.size} project(s) with ${projectScannerWrappers.size} scanner(s)." }
 
+            val projectScanners = scannerWrappers[PackageType.PROJECT].orEmpty().map { it.name }.toSet()
```
sschuberth, line 121: Can we reuse `projectScannerWrappers` here?

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -137,6 +141,9 @@ class Scanner(
 
             logger.info { "Scanning ${packages.size} package(s) with ${packageScannerWrappers.size} scanner(s)." }
 
+            val packageScanners = scannerWrappers[PackageType.PACKAGE].orEmpty().map { it.name }.toSet()
```
sschuberth, line 144: Can we reuse `packageScannerWrappers` here?

model/src/main/kotlin/ScannerRun.kt
fviernau: Wouldn't ordering the values be needed as well to be deterministic, to allow for use with expected results?

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,54 @@
...
+
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.PackageType
+import org.ossreviewtoolkit.model.fromYaml
+import org.ossreviewtoolkit.utils.test.getAssetFile
+import org.ossreviewtoolkit.utils.test.shouldNotBeNull
+
+class MultipleScannersTest : WordSpec({
+    "Scanning a set of packages with overlapping provenance" should {
```
fviernau, line 33: How about sth. like the following?
```
"Scanning a project and a package with overlapping provenance and non overlapping scanners" should {
      "return scan results with non-overlapping scanners"
```

"

scanner/src/funTest/assets/analyzer-result_for_multiple_scanners.yml
```diff
@@ -0,0 +1,124 @@
+---
+repository:
```
fviernau, line 2: This just contains one package and one project. Would it make sense to create it programatically instead?

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -72,7 +72,7 @@ import org.ossreviewtoolkit.utils.spdx.toSpdx
 const val TOOL_NAME = "scanner"
 
 @Suppress("TooManyFunctions")
-class Scanner(
+data class Scanner(
```
fviernau, line 75: Why `data class`?

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -169,7 +176,8 @@ class Scanner(
             config = filteredScannerConfig,
             provenances = projectResults.provenances + packageResults.provenances,
             scanResults = projectResults.scanResults + packageResults.scanResults,
-            files = projectResults.files + packageResults.files
+            files = projectResults.files + packageResults.files,
+            scanners = scannersByPackages
```
fviernau, line 180: I believe line 180 should be `projectResult.scanners + packageResults.scanners`.

(Which implies the logic you added above should instead go into the other overload of `scan()`)

What do you think?

fviernau (changes requested):

scanner/src/main/kotlin/Scanner.kt
nnobelis: To be able to `copy()` the scanner with a different `scannerWrappers`: https://github.com/bosch-io/oss-review-toolkit/blob/b90e2a3b35ba8241be1d0d6d9c140d2a2cdebb71/scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt#L38

model/src/main/kotlin/ScannerRun.kt
nnobelis: @fviernau 
@sschuberth 
Is there a special annotation to do that of should I again create a converter ?

sschuberth: > Is there a special annotation to do that of should I again create a converter ?

If the values in a map should be sorted as well, I believe you need a converter again.

scanner/src/main/kotlin/Scanner.kt
nnobelis: Yes it is better.

scanner/src/funTest/assets/analyzer-result_for_multiple_scanners.yml
nnobelis: Done. Could you please review the changes ? I think I over complicated it and don't need to build the dependency graph by hand.

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        path = ""
+    )
+
+    val project = Project.EMPTY.copy(
+        id = createId("spdx-project"),
+        vcs = vcsProject,
+        vcsProcessed = vcsProject.normalize()
+    )
+
+    val scopes = mapOf(":spdx-project::default" to listOf(RootDependencyIndex(0, 0)))
```
fviernau, line 93: I believe that for this test it suffices to not link the package to the project at all.
So, you  could just omit any scopes and the dependency graph. Would that make sense?

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    val pkg = Package.EMPTY.copy(
+        id = createId("spdx-dependency"),
+        vcs = vcsPackage,
+        vcsProcessed = vcsPackage.normalize()
+    )
+
+    val vcsProject = VcsInfo(
+        type = VcsType.GIT,
+        url = "https://github.com/nnobelis/spdx-project.git",
```
fviernau, line 82: This is using a repo under you namespace. I believe that you could just use any `test-data` repo under `oss-review-toolkit`. IIUC you are after overlapping provenances, not necessarily covering submodules. So, maybe https://github.com/oss-review-toolkit/ort-test-data-scanner can be used?

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,114 @@
...
+    val dependencyGraph = mapOf(
+        DEFAULT_NS to DependencyGraph(
+            listOf(pkg.id),
+            scopes = scopes,
+            nodes = listOf(
+                DependencyGraphNode(0, 0, PackageLinkage.STATIC)
+            )
+        )
+    )
+    val analyzerRun = AnalyzerRun.EMPTY.copy(
```
fviernau, line 103: Empty lines are a bit inconsistent. in this function.

scanner/src/main/kotlin/Scanner.kt
fviernau: I'm not sure about the data class thing, have you considered adding a parameter to `createScanner()` as alternative?

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.ossreviewtoolkit.model.PackageLinkage
+import org.ossreviewtoolkit.model.PackageType
+import org.ossreviewtoolkit.model.Project
+import org.ossreviewtoolkit.model.RootDependencyIndex
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.utils.test.shouldNotBeNull
+
+private const val DEFAULT_NS = "SpdxDocumentFile"
+private const val DEFAULT_VERSION = ""
```
fviernau, line 41: Would it make sense to replace above two constants with two constants of type `Identifier` ? (One for the package, one for the project
) ?

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,114 @@
...
+            scannerWrappers = mapOf(
+                PackageType.PROJECT to listOf(dummyScanners.copy(name = "Dummy2")),
+                PackageType.PACKAGE to listOf(dummyScanners)
+            )
+        ).scan(analyzerResult, skipExcluded = false, emptyMap())
+
+        "return scan results with non-overlapping scanners" {
+            ortResult.scanner.shouldNotBeNull {
+                val results = getScanResults(Identifier(DEFAULT_NS, "", "spdx-project", DEFAULT_VERSION))
+                results.map { it.scanner.name }.toSet() shouldBe setOf("Dummy2")
```
fviernau, line 58: I'd prefer if `toSet()` was dropped and `should containExactly("Dummy")` be used instead.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -238,11 +239,16 @@ class Scanner(
             }
         }
 
+        val scannersByPackages = mutableMapOf<Identifier, Set<String>>()
+        val scannerNames = scanners.map { it.name }.toSet()
+        packages.associateByTo(scannersByPackages, { it.id }) { scannerNames }
```
fviernau, line 244: Can be simplified:
```
val scannerNames = scanners.mapTo(mutableSetOf()) { it.name }
val scannerNamesByPackageId = packages.associateBy({ it.id }, { scannerNames })
```

fviernau (changes requested):

model/src/main/kotlin/ScannerRun.kt
```diff
@@ -29,6 +29,7 @@ import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.model.utils.FileListSortedSetConverter
 import org.ossreviewtoolkit.model.utils.ProvenanceResolutionResultSortedSetConverter
 import org.ossreviewtoolkit.model.utils.ScanResultSortedSetConverter
+import org.ossreviewtoolkit.model.utils.ScannersMapConverter
```
fviernau, line 32: commit-message: This commit fixes a more general problem than what is described in the commit message. 

The scanner may used different set of scanners for scanning different packages. The same holds for projects. `ScannerRun` does not contain that information, and `ScannerRun.getScanResultsById()` just returns all scans matching the packages (nested) provenance, disregarding which scanners are to be used for the particular package. Extend the data model to contain that information and use it to adjust the internals of `getScanResultsById()` to adhere to the set of scanners to be used for the particular project or package.


----

Just as a hint: Which scanner to use for scanning a project / package depends on (haven't double checked):
- `--skip-excluded`
- `--skip-concluded`
- `--scanners`
- `--project-scanners

fviernau (changes requested):

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
nnobelis: I am really confused by this repository as there are no configuration file for any package managers ?!
How the project/package identifiers are supposed to look like ?

If I try to create a `Dummy::project:1.0.0` Project and  `Dummy::package:1.0.0` Package like you did in your `ScannerIntegrationFunTest`, the scanners are not workin.

I get a `ScannerDetails` with name `ProvenanceResolver` that contains a single issue:
`IOException: Could not resolve provenance for package 'Dummy::project:1.0.0' for source code origins [VCS, ARTIFACT].`.

model/src/main/kotlin/ScannerRun.kt
nnobelis: I don't understand the last sentence of your comment.

fviernau: I suppose you mean the sentence below the horizontal line, I meant to list the things which can influence which scanner are to be used for a package / project.

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
nnobelis: Yes, it is working !

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
nnobelis: Yes I had a VCS info but in the end, the bug was caused by the dependency graph. When I removed it, the `ProvenanceResolver` error is gone.

model/src/main/kotlin/ScannerRun.kt
nnobelis: Understand. Maybe also `--package-types` ?

model/src/main/kotlin/ScannerRun.kt
```diff
@@ -192,6 +200,8 @@ data class ScannerRun(
 
         val scanResultsByPath = resolutionResult.getKnownProvenancesWithoutVcsPath().mapValues { (_, provenance) ->
             scanResultsByProvenance[provenance].orEmpty()
+        }.mapValues { scanResults ->
```
fviernau, line 203: Could you use `(_ , scanResults) ->` instead of just `scanResults ->` ?

model/src/main/kotlin/utils/SortedSetConverters.kt
```diff
@@ -74,6 +75,11 @@ class ProvenanceResolutionResultSortedSetConverter :
     override fun convert(value: Set<ProvenanceResolutionResult>) = value.toSortedSet(compareBy { it.id })
 }
 
+class ScannersMapConverter : StdConverter<Map<Identifier, Set<String>>, Map<Identifier, Set<String>>>() {
+    override fun convert(value: Map<Identifier, Set<String>>) =
+        value.mapValues { it.value.toSortedSet() }.toSortedMap(compareBy { it })
```
fviernau, line 80: Quesion: Is it possible to call `toSortedMap` without parameter ?

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,97 @@
...
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.PackageType
+import org.ossreviewtoolkit.model.Project
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.utils.test.shouldNotBeNull
+
+private val PROJECT_IDENTIFIER = Identifier("Dummy", "", "project", "1.0.0")
```
fviernau, line 37: Consider renaming this to just `PROJECT_ID`, same right below.

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        val scannerWrappers = mapOf(
+            PackageType.PROJECT to listOf(dummyScanner.copy(name = "Dummy2")),
+            PackageType.PACKAGE to listOf(dummyScanner)
+        )
+        val ortResult = createScanner(scannerWrappers).scan(analyzerResult, skipExcluded = false, emptyMap())
+
+        "return scan results with non-overlapping scanners" {
+            ortResult.scanner.shouldNotBeNull {
+                val results = getScanResults(PROJECT_IDENTIFIER)
+                results.map { it.scanner.name }.toSet() should containExactly("Dummy2")
```
fviernau, line 54: The String `"Dummy2"` is ok, but `"Dummy"` is a bit magic. 
I propose to 1. inline the variable `dummyScanner` and also use the constructor parameter instead of `copy()`. This in turn allows to not make `DummyScanner` a data class.

```
PackageType.PROJECT to listOf(DummyScanner(name = "Dummy1")),
PackageType.PACKAGE to listOf(DummyScanner(name = "Dummy2"))
```

scanner/src/funTest/kotlin/scanners/MultipleScannersTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        vcs = vcsPackage,
+        vcsProcessed = vcsPackage.normalize()
+    )
+
+    val vcsProject = VcsInfo(
+        type = VcsType.GIT,
+        url = "https://github.com/oss-review-toolkit/ort-test-data-scanner.git",
+        revision = "97d57bb4795bc41f496e1a8e2c7751cefc7da7ec",
+        path = ""
+    )
```
fviernau, line 81: If I'm not missing anything `vcsPackage` and `vcsProject` are identical.
So, let's drop one of the two and rename the remaining to just `vcs`.

fviernau (changes requested):

model/src/main/kotlin/ScannerRun.kt
```diff
@@ -72,6 +73,12 @@ data class ScannerRun(
     @JsonSerialize(converter = ScanResultSortedSetConverter::class)
     val scanResults: Set<ScanResult>,
 
+    /**
+     * The scanners for each scanned package.
```
fviernau, line 77: The scanners for a package can also be empty. 
So, it should contain also IDs for not scanned packages.
How about saying instead:

The names of the scanners which have been used to scan the package.

model/src/main/kotlin/ScannerRun.kt
nnobelis: This is `mapValues`, there is only one parameter.

fviernau: Isn't the value a tuple `(path, scanResults)` so it can be decomposed ?

evaluator/src/test/kotlin/ProjectSourceRuleTest.kt
```diff
@@ -229,6 +229,6 @@ private fun createOrtResult(
                     )
                 )
             )
-        )
+        ).copy(scanners = mapOf(id to setOf("")))
```
fviernau, line 232: Please remove this `copy()` call and adjust `scannRunOf` to assign the scanner names internally, so that the return `ScannerRun` is consistent. (This way all callers of `scannerRunOf` get fixed.

model/src/main/kotlin/ScannerRun.kt
fviernau: This is not addressed yet.

utils/test/src/main/kotlin/Utils.kt
```diff
@@ -189,6 +189,12 @@ fun scannerRunOf(vararg pkgScanResults: Pair<Identifier, List<ScanResult>>): Sca
         )
     }
 
+    val scanners = mutableMapOf<Identifier, MutableSet<String>>()
```
fviernau, line 192: This can be simplified to:

```
val scanners = pkgScanResults.map { (id, scanResultsForId) ->
        id to scanResultsForId.mapTo(mutableSetOf()) { it.scanner.name}
    }.toMap()
``

**Review by fviernau — CHANGES_REQUESTED** [review]

File: utils/test/src/main/kotlin/Utils.kt (hunk 22 above)
Comment (nnobelis) [reply utils/test/src/main/kotlin/Utils.kt]: IntelliJ suggested even the further optimization:

```
val scanners = pkgScanResults.associate { (id, scanResultsForId) ->
        id to scanResultsForId.mapTo(mutableSetOf()) { it.scanner.name }
}
```

fviernau (dismissed):

fviernau (dismissed):

mnonnenmacher (dismissed):

fviernau (dismissed):

fviernau (approved):

PR #7282, opened 2023-07-13
ScanCode result parser simplifications

sschuberth:


plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
```diff
@@ -167,15 +167,13 @@ private fun getSpdxLicenseId(license: JsonNode): String {
     val idFromSpdxKey = license["spdx_license_key"].textValueOrEmpty().toSpdxId(allowPlusSuffix = true)
 
     // For regular SPDX IDs, return early here.
-    if (idFromSpdxKey.isNotEmpty() && !idFromSpdxKey.startsWith(LICENSE_REF_PREFIX)) return idFromSpdxKey
+    if (idFromSpdxKey.isNotEmpty()) return idFromSpdxKey
```
fviernau, line 165: Is this change non-functional? E.g. if ScanCode has `LicenseRef-` IDs which do not start with `LicenseRef-scancode-` it wouldn't be, right?

model/src/test/kotlin/LicenseFindingTest.kt
```diff
@@ -119,5 +119,29 @@ class LicenseFindingTest : WordSpec({
             "AGPL-1.0-or-later OR BSD (3-Clause) OR BSD (2-Clause)".mapLicense(mapping) shouldBe
                     "AGPL-1.0-or-later OR BSD-3-Clause OR BSD-2-Clause"
         }
+
```
fviernau, line 122: commit-message: Please indicate whether this commit is non-functional change, e.g. does it just remove code redundancy?

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
```diff
@@ -143,7 +142,7 @@ private fun getLicenseFindings(result: JsonNode, parseExpressions: Boolean): Set
             LicenseFinding(
                 license = spdxLicenseExpression,
                 location = TextLocation(
-                    path = file["path"].textValue().removePrefix(input),
+                    path = file["path"].textValue(),
```
fviernau, line 145: commit: Was this a bug before? If so, should the prefix of the title be `fix!` ?

If there was something broken, it would be nice to have the impact explained.

fviernau (changes requested):

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
sschuberth: Nothing was broken, so it's not a bug fix. It's just superfluous to relativize paths here already as it's now being done centrally. So this is about avoiding duplicate work. See how the mentioned 6e68575eefb693307170d9b7af978260fa1b80bb does the same thing for the other scanners by removing `relativizePath()` calls. The same should have been done for ScanCode back then, too, hence the mention of a "fixup".

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
sschuberth: > Is this change non-functional?

This is indeed a change that does not (intend to) change behavior.

> E.g. if ScanCode has `LicenseRef-` IDs which do not start with `LicenseRef-scancode-` it wouldn't be, right?

I still believe the change does not change behavior.

Just look at these lines in isolation:

    if (idFromSpdxKey.isNotEmpty() && !idFromSpdxKey.startsWith(LICENSE_REF_PREFIX)) return idFromSpdxKey
    return idFromSpdxKey.takeUnless { it.isEmpty() } // ...

The last line could be rewritten as

    if (idFromSpdxKey.isNotEmpty()) return idFromSpdxKey

So inserted into the first snippet we'd get

    if (idFromSpdxKey.isNotEmpty() && !idFromSpdxKey.startsWith(LICENSE_REF_PREFIX)) return idFromSpdxKey
    if (idFromSpdxKey.isNotEmpty()) return idFromSpdxKey

So the last line here basically is a just relaxed version of the first line, which contains the additional `!idFromSpdxKey.startsWith(LICENSE_REF_PREFIX)` condition. Which means we can drop the first line altogether, as the last line returns anyway if only `idFromSpdxKey` is not empty.

fviernau: I'm not sure if I get it. 

What if `idFromSpdxKey` gets assigned to `LicenseRef-HERE-license` in line `idFromSpdxKey` ?
I expect that previously, the return of the function would start with `LicenseRef-scancode-` while now it doesn't anymore?

sschuberth: If `idFromSpdxKey` was "LicenseRef-HERE-license", then in the original code line 170 would *not* have returned as the license starts with "LicenseRef-". But line 172 would have returned "LicenseRef-HERE-license" as it's non-empty.

The new code simplifies this by immediately returning "LicenseRef-HERE-license" as it's non-empty.

Before and after this change, the "scancode" infix is only added if no "LicenseRef-" prefix was present at all.

fviernau (approved):

PR #7285, opened 2023-07-14
Migrate the Scancode result parser to kotlinx-serialization

sschuberth:

### DIFF
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCodeErrorMappers.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCodeErrorMappers.kt
@@ -19,8 +19,6 @@
 
 package org.ossreviewtoolkit.plugins.scanners.scancode
 
-import com.fasterxml.jackson.databind.JsonNode
-
 import org.ossreviewtoolkit.model.Issue
 
 // Note: The "(File: ...)" part in the patterns below is actually added by our own getRawResult() function.
@@ -35,27 +33,20 @@ private val TIMEOUT_ERROR_REGEX = Regex(
             "ERROR: Processing interrupted: timeout after (?<timeout>\\d+) seconds. \\(File: (?<file>.+)\\)"
 )
 
-private fun getInputPath(result: JsonNode): String {
-    val header = result["headers"].single()
-    val input = header["options"]["input"]
-    val path = input.takeUnless { it.isArray } ?: input.single()
-    return path.textValue().let { "$it/" }
-}
-
 /**
  * Map scan errors for all files using messages that contain the relative file path.
  */
-internal fun mapScanErrors(result: JsonNode): List<Issue> {
-    val input = getInputPath(result)
-    return result["files"]?.flatMap { file ->
-        val path = file["path"].textValue().removePrefix(input)
-        file["scan_errors"].map {
+internal fun mapScanErrors(result: ScanCodeResult): List<Issue> {
+    val input = result.headers.single().options.input.single()
+    return result.files.flatMap { file ->
+        val path = file.path.removePrefix(input).removePrefix("/")
+        file.scanErrors.map { error ->
             Issue(
                 source = ScanCode.SCANNER_NAME,
-                message = "${it.textValue()} (File: $path)"
+                message = "$error (File: $path)"
             )
         }
-    }.orEmpty()
+    }
 }
 
 /**
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCodeErrorMappers.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCodeErrorMappers.kt
@@ -1,0 +1,119 @@
+/*
+ * Copyright (C) 2017 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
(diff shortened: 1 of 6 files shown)


plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
```diff
@@ -207,9 +207,9 @@ private fun getCopyrightFindings(result: JsonNode): Set<CopyrightFinding> {
 }
 
 /**
- * Get the list of [Issue]s for scanned files.
+ * Map scan errors for all files using messages that contains the relative file path.
```
fviernau: should be "contain" instead of "contains".

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModel.kt
```diff
@@ -0,0 +1,106 @@
...
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonTransformingSerializer
+
+@Serializable
+data class ScanCodeResult(
+    val headers: List<HeaderEntry>,
+    val files: List<FileEntry>
+)
+
+@Serializable
```
fviernau, line 35: consider moving all data classes below under `ScanCodeResult`.

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
```diff
@@ -52,137 +89,90 @@ private data class LicenseMatch(
...
                 )
             )
         }
     }
 
-    return copyrightFindings
+    return ScanSummary(
+        startTime = TIMESTAMP_FORMATTER.parse(header.startTimestamp).query(Instant::from),
+        endTime = TIMESTAMP_FORMATTER.parse(header.endTimestamp).query(Instant::from),
+        licenseFindings = associateLicensesWithExceptions(licenseFindings),
```
fviernau, line 174: Maybe `fun parseTimeStamp(String): Instant` could be extracted.

fviernau (changes requested):

fviernau:
Only skimmed through quickly, so far.

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModel.kt
sschuberth: You mean to nest them inside `ScanCodeResult`? I'd prefer not to do that to limit indentation. Esp. as that consequently would need to be done also for all other more deeply nested classes.

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
sschuberth: Right, but it was inlined also before, and introducing a function doesn't really pay off here, IMO.

plugins/scanners/scancode/src/main/kotlin/ScanCodeErrorMappers.kt
```diff
@@ -35,27 +33,20 @@ private val TIMEOUT_ERROR_REGEX = Regex(
...
  */
-internal fun mapScanErrors(result: JsonNode): List<Issue> {
-    val input = getInputPath(result)
-    return result["files"]?.flatMap { file ->
-        val path = file["path"].textValue().removePrefix(input)
-        file["scan_errors"].map {
+internal fun mapScanErrors(result: ScanCodeResult): List<Issue> {
+    val input = result.headers.single().options.input.single()
+    return result.files.flatMap { file ->
+        val path = file.path.removePrefix(input).removePrefix("/")
```
mnonnenmacher, line 42: Just wondering, why was `removePrefix("/")` not required before?

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
```diff
@@ -34,16 +42,45 @@ import org.ossreviewtoolkit.model.TextLocation
...
+        else -> SerializersModule {
+            polymorphicDefaultDeserializer(CopyrightEntry::class) { CopyrightEntry.Version2.serializer() }
+        }
+    }
+
+    val json = Json {
+        ignoreUnknownKeys = true
+        namingStrategy = JsonNamingStrategy.SnakeCase
+        serializersModule = module
+    }
```
mnonnenmacher, line 80: This could be a constant to reuse it when parsing multiple results.

plugins/scanners/scancode/src/main/kotlin/ScanCodeErrorMappers.kt
sschuberth: It was done differently before: `getInputPath()` was ensured to end with a `/`, so `removePrefix(input)` also removed that trailing slash.

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt
sschuberth: You mean the whole `Json` instance? I believe that's not possible as the modules are set dynamically depending on the output format version contained in each file.

mnonnenmacher (approved):

PR #7286, opened 2023-07-14
fix(spdx): Fix-up representing the relationship of files and packages

fviernau:
Use the dedicated field for representing that a file corresponds to a package, instead of the more verbose generic relationship (in the SPDX Document report).

This fix-up corresponds to #6947.

### DIFF
--- a/plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
+++ b/plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
@@ -61,6 +61,7 @@
       "referenceLocator" : "pkg:maven/first-package-group/first-package@0.0.1"
     } ],
     "filesAnalyzed" : true,
+    "hasFiles" : [ "SPDXRef-File-1" ],
     "homepage" : "first package's homepage URL",
     "licenseConcluded" : "NOASSERTION",
     "licenseDeclared" : "BSD-3-Clause AND (MIT OR GPL-2.0-only)",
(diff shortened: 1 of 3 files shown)


sschuberth:
Would you have a reference for the `hasFiles` field (and add it to the commit message)? I can't find it being documented at https://spdx.github.io/spdx-spec/v2.2.2/package-information/.

fviernau:
> Would you have a reference for the `hasFiles` field (and add it to the commit message)? I can't find it being documented at https://spdx.github.io/spdx-spec/v2.2.2/package-information/.

I only found it in the examples, see https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXJSONExample-v2.2.spdx.json.

sschuberth:
> > Would you have a reference for the `hasFiles` field (and add it to the commit message)? I can't find it being documented at https://spdx.github.io/spdx-spec/v2.2.2/package-information/.
> 
> I only found it in the examples, see https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXJSONExample-v2.2.spdx.json.

Ok, then probably the best reference is the schema: https://github.com/spdx/spdx-spec/blob/development/v2.2.1/schemas/spdx-schema.json#L367-L374

fviernau:
> Ok, then probably the best reference is the schema: https://github.com/spdx/spdx-spec/blob/development/v2.2.1/schemas/spdx-schema.json#L367-L374

I've added it to the commit message.

sschuberth (approved):

PR #7292, opened 2023-07-14
feat(model): Require size when writing to `ProvenanceFileStorage`

mnonnenmacher:
Add a size argument to the `putData` function of the `ProvenanceFileStorage` which represents the size of the input stream to be written.

This is a precondition for being able to use the PostgreSQL large object storage as a backend for the storage [1] which is a requirement in the ongoing ORT Server development.

The current implementations of `ProvenanceFileStorage` do not require the size of the input stream, but the size is easily available for all callers of the `putData` function.

[1]: https://www.postgresql.org/docs/current/lo-interfaces.html

### DIFF
--- a/model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
+++ b/model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
@@ -22,6 +22,7 @@ package org.ossreviewtoolkit.model.utils
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.shouldBe
 
+import java.io.ByteArrayInputStream
 import java.io.InputStream
 
 import org.ossreviewtoolkit.model.ArtifactProvenance
@@ -60,7 +61,7 @@ class PostgresProvenanceFileStorageFunTest : WordSpec({
         }
 
         "return true when data for the given provenance has been added" {
-            storage.putData(VCS_PROVENANCE, InputStream.nullInputStream())
+            storage.putData(VCS_PROVENANCE, InputStream.nullInputStream(), 0L)
 
             storage.hasData(VCS_PROVENANCE) shouldBe true
         }
@@ -68,8 +69,15 @@ class PostgresProvenanceFileStorageFunTest : WordSpec({
 
     "putData()" should {
         "return the data corresponding to the given provenance given such data has been added" {
-            storage.putData(VCS_PROVENANCE, "VCS".byteInputStream())
-            storage.putData(SOURCE_ARTIFACT_PROVENANCE, "source artifact".byteInputStream())
+            val vcsByteArray = "VCS".toByteArray()
+            val sourceArtifactByteArray = "source artifact".toByteArray()
+
+            storage.putData(VCS_PROVENANCE, ByteArrayInputStream(vcsByteArray), vcsByteArray.size.toLong())
+            storage.putData(
+                SOURCE_ARTIFACT_PROVENANCE,
+                ByteArrayInputStream(sourceArtifactByteArray),
+                sourceArtifactByteArray.size.toLong()
+            )
 
             storage.getData(VCS_PROVENANCE) shouldNotBeNull { String(use { readBytes() }) shouldBe "VCS" }
             storage.getData(SOURCE_ARTIFACT_PROVENANCE) shouldNotBeNull {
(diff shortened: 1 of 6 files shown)


model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
```diff
@@ -22,6 +22,7 @@ package org.ossreviewtoolkit.model.utils
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.shouldBe
 
+import java.io.ByteArrayInputStream
```
fviernau, line 25: How about adding as "syntactic sugar" a `fun putData(provenance, bytes: ByteArray()` for not making the caller that verbose?

model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
```diff
@@ -60,16 +61,23 @@ class PostgresProvenanceFileStorageFunTest : WordSpec({
         }
 
         "return true when data for the given provenance has been added" {
-            storage.putData(VCS_PROVENANCE, InputStream.nullInputStream())
+            storage.putData(VCS_PROVENANCE, InputStream.nullInputStream(), 0L)
```
fviernau, line 64: Do you have a reference which shows that / why the size is required for postgres?

fviernau (changes requested):

model/src/funTest/kotlin/utils/PostgresProvenanceFileStorageFunTest.kt
mnonnenmacher: You mean just for the test code? Because I think the change it `FileListResolver` is not that verbose.

mnonnenmacher: I don't enough about the implementation, maybe it's also a requirement of the JDBC driver or so. But after reading some docs I think it should be possible to remove that again later on. Therefore I have rephrased the commit message so that this is basically an ask of the people working on the ORT server to introduce this to unblock us for now pending later investigation if it can be removed again.

fviernau: hm, there is really no pointer to any API docs which shows that it's needed?

sschuberth: Maybe [this](https://www.postgresql.org/docs/7.1/jdbc-lo.html#JDBC-LO-EXAMPLE) is a better link as it refer to Java and not C code. So this

    ps.setBinaryStream(2, fis, file.length());

indeed requires the size of the `FileInputStream` to be passed.

mnonnenmacher: No, I was hoping that the new commit message made it clear that this is a requirement of our internal implementation. Unfortunately @oheger-bosch is not available anymore today, he could probably provide more details.

sschuberth: Looks like `length` is only needed if you don't want all of the stream to be read.

fviernau: Looking at another implementation of this API, the length is used for double checking that the expected amount of bytes has been read from the stream, e.g. fail if it doesn't match, see https://learn.microsoft.com/en-us/sql/connect/jdbc/reference/setbinarystream-method-int-java-io-inputstream-long?view=sql-server-ver16.

fviernau: I'll leave the decision whether to merge up to you guys, even tough I prefer to wait until there is a clear reason that it is really required. Because all API docs we've seen now seem to provide a way without the length.

sschuberth: TBH, I'm also reluctant towards this "dummy" change. @mnonnenmacher, could you maybe explain why it's easier to do this change here, than making the change in the server to not require `length`?

oheger-bosch: To give you a bit background information:
In ORT Server, we have introduced a generic storage API for storing all kinds of data. It can be backed by different implementations. Postgres is one of those, another one could be a file storage like Azure Storage Accounts. When designing the API we had the different possible implementations in mind, and some of them require the size of the data to be known beforehand. This is for instance needed by the [putBlob](https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=azure-ad) operation from Azure, where requests require a correct `Content-Length` header.

For Postgres itself, the data size is not required for writing or reading blobs. However, the implementation in ORT Server uses the size information to distinguish between data that can be read into memory and large data that needs to be streamed to the client. If the size is below a configurable threshold, it is read directly and send as response to the client. Otherwise, we need to work with temporary files to buffer the data; this is because Postgres requires that blobs can only be accessed in a transaction; however, when sending the HTTP response to the client, the transaction is already closed.

sschuberth: Thanks for the details @oheger-bosch. So if instead of Postgres the Azure backend would be mentioned in the commit message as an example that really *requires* the length (like you do above), and also the information from the last paragraph that ORT might use different "modes" depending on the size of the data would go into the commit message, I'd be ok with this change.

mnonnenmacher: I have updated the commit message.

sschuberth (approved):

mnonnenmacher:
Merged as the failing `GodModFunTest` and `GitWorkingTreeFunTest` are unrelated to the change.

PR #7293, opened 2023-07-14
Minor scanner improvements

fviernau:
See individual commits.

### DIFF
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -179,8 +179,8 @@ class Scanner(
     }
 
     suspend fun scan(packages: Set<Package>, context: ScanContext): ScannerRun {
-        val scanners = scannerWrappers[context.packageType].orEmpty()
-        if (scanners.isEmpty()) return ScannerRun.EMPTY
+        val scanners = scannerWrappers[context.packageType]
+        if (scanners.isNullOrEmpty()) return ScannerRun.EMPTY
 
         val controller = ScanController(packages, scanners, scannerConfig)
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/Scanner.kt
```diff
@@ -177,8 +177,7 @@ class Scanner(
     }
 
     suspend fun scan(packages: Set<Package>, context: ScanContext): ScannerRun {
-        val scanners = scannerWrappers[context.packageType].orEmpty()
-        if (scanners.isEmpty()) return ScannerRun.EMPTY
+        val scanners = scannerWrappers[context.packageType]?.takeIf { it.isNotEmpty() } ?: return ScannerRun.EMPTY
```
sschuberth: Personally, I actually find the new code harder to read. However, how about

        val scanners = scannerWrappers[context.packageType]
        if (scanners.isNullOrEmpty()) return ScannerRun.EMPTY

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -239,8 +239,9 @@ class Scanner(
             }
         }
 
-        val scannerNames = scannerWrappers.mapTo(mutableSetOf()) { it.name }
-        val scanners = packages.associateBy({ it.id }) { scannerNames }
+        val scanners = scannerWrappers.mapTo(mutableSetOf()) { it.name }.let { scannerNames ->
```
sschuberth, line 242: Here too, I find the new code harder to read than the old code.

sschuberth (approved):

PR #7294, opened 2023-07-14
refactor(scanner): Move storages to the `Scanner` constructor

mnonnenmacher:
Move the `archiver` and `fileListStorage` to the constructor of the `Scanner` class. This makes it possible to provide alternative implementations when using ORT programmatically.

### DIFF
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -55,6 +55,8 @@ import org.ossreviewtoolkit.model.config.createFileArchiver
 import org.ossreviewtoolkit.model.config.createStorage
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.mapLicense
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/Scanner.kt
```diff
@@ -80,7 +82,9 @@ class Scanner(
     val storageWriters: List<ScanStorageWriter>,
     val packageProvenanceResolver: PackageProvenanceResolver,
     val nestedProvenanceResolver: NestedProvenanceResolver,
-    val scannerWrappers: Map<PackageType, List<ScannerWrapper>>
+    val scannerWrappers: Map<PackageType, List<ScannerWrapper>>,
+    val archiver: FileArchiver? = scannerConfig.archive.createFileArchiver(),
+    val fileListStorage: ProvenanceFileStorage = scannerConfig.fileListStorage.createStorage()
```
sschuberth, line 87: Both could be `private`, and `fileListStorage` could also have `val` removed.

sschuberth (approved):

PR #7298, opened 2023-07-17
feat(Plugins): Add package-configuration provider for ort-config

MarcelBochtler:
Add a package-configuration provider for [ort-config](https://github.com/oss-review-toolkit/ort-config)

### DIFF
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -45,6 +45,7 @@ ort:
     config:
       path: '/some-path/'
       mustExist: true
+  - type: OrtConfig
 
   # Providers are listed from highest to lower priority. Technically, they are applied in reverse order: The provider
   # with the highest priority is applied last, so it can override any previously applied curations.
--- a/plugins/package-configuration-providers/ort-config/src/main/kotlin/OrtConfigPackageConfigurationProvider.kt
+++ b/plugins/package-configuration-providers/ort-config/src/main/kotlin/OrtConfigPackageConfigurationProvider.kt
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.plugins.packageconfigurationproviders.ortconfig
+
+import java.io.File
+
+import org.apache.logging.log4j.kotlin.Logging
+
+import org.ossreviewtoolkit.downloader.vcs.Git
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.Provenance
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.model.config.PackageConfiguration
+import org.ossreviewtoolkit.model.utils.PackageConfigurationProvider
+import org.ossreviewtoolkit.plugins.packageconfigurationproviders.api.DirectoryPackageConfigurationProvider
+import org.ossreviewtoolkit.plugins.packageconfigurationproviders.api.PackageConfigurationProviderFactory
+import org.ossreviewtoolkit.utils.common.safeMkdirs
+import org.ossreviewtoolkit.utils.ort.ortDataDirectory
+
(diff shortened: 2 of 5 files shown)


model/src/main/resources/reference.yml
```diff
@@ -40,6 +40,7 @@ ort:
   # configurations for the same package, see https://github.com/oss-review-toolkit/ort/issues/6972 for details.
   packageConfigurationProviders:
   - type: DefaultDir
+  - type: OrtConfig
```
sschuberth, line 43: Unrelated to this specific PR, but it just struck me that in line 44 below we have `type: Dir`, whereas for `packageCurationProviders` we have `type: File` even if it's in fact a directory. I believe we should align that.

And I also just realized that we have `DirectoryPackageConfigurationProvider` *and* `DirPackageConfigurationProvider`, both inheriting from `SimplePackageConfigurationProvider`... huh!? I guess some clean-up is needed here.

sschuberth: I just checked, and both `Dir(ectory)PackageConfigurationProvider` work when being passed just a single `File` instead of a directory.

MarcelBochtler: Allowing single files is an unintended functionality. It also should only work with single `PackageConfiguration`s in this file, as multiple package configurations in a single files are no longer supported since #6831.

So I think that this behavior should be fixed and only directories, instead of renaming `Dir` to `File`.

sschuberth: > So I think that this behavior should be fixed and only directories, instead of renaming `Dir` to `File`.

I'm not yet convinced, as it could make sense to align with `PackageCurationProvider` instead.

Anyway, this is all not strictly related to this PR and could be addressed independently.

plugins/package-configuration-providers/ort-config/src/main/kotlin/OrtConfigPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,78 @@
...
+    override fun getPackageConfigurations(packageId: Identifier, provenance: Provenance) =
+        DirectoryPackageConfigurationProvider(
+            configurationsDir.resolve(PACKAGE_CONFIGURATIONS_DIR)
+        ).getPackageConfigurations(packageId, provenance)
+}
+
+private fun updateOrtConfig(dir: File) {
+    dir.safeMkdirs()
+    Git().apply {
+        val workingTree = initWorkingTree(dir, VcsInfo.EMPTY.copy(type = VcsType.GIT, url = ORT_CONFIG_REPOSITORY_URL))
```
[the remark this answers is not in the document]
mnonnenmacher: I think it would be difficult to share an instance of `WorkingTreeCache` as both providers are located in separate plugin modules, we probably would have to make progress with Koin first to do this nicely. However, at least both providers use the same local directory for cloning the ort-config repository.

model/src/main/resources/reference.yml
mnonnenmacher: If we do a renaming, the curation provider should be renamed from `File` to `Dir`, treating the single file input as a special case, because for package configurations we won't support single files again.
And I agree that we should not have both a `DirectoryPackageConfigurationProvider` and a `DirPackageConfigurationProvider`.

sschuberth: > because for package configurations we won't support single files again.

Note that there are two distinct "single file things" that we should not mix up:

1. Storing a multiple package configuration in a single file. This is the case that is not supported anymore.
2. Pointing a `Dir(ectory)PackageConfigurationProvider` at a single file containing a single package configuration. This works coincidentally, and I do not see a reason to stop supporting that.

mnonnenmacher: Wouldn't it be weird to have a provider that can only provide a single package configuration for a single package?

sschuberth: That's not what I'm suggesting to do. I'm just saying that *if* someone configures a `Dir(ectory)PackageConfigurationProvider` to point to a file instead of a directory, I'm fine if that "just works" (as it is now) instead of e.g. throwing an exception.

mnonnenmacher: But as it makes little sense to configure a provider for a single package configuration it is likely a configuration mistake in which case it would be more helpful for the user to throw an exception.

sschuberth, line 43: Nit: Maybe make `OrtConfig` last in the list to keep `DefaultDir` and `Dir` next to each other?

plugins/package-configuration-providers/ort-config/src/main/kotlin/OrtConfigPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,78 @@
...
+    override fun create(config: Unit): PackageConfigurationProvider = OrtConfigPackageConfigurationProvider()
+
+    override fun parseConfig(config: Map<String, String>) = Unit
+}
+
+/**
+ * A [PackageConfigurationProvider] that provides [PackageConfiguration]s from the
+ * [ort-config repository](https://github.com/oss-review-toolkit/ort-config).
+ */
+open class OrtConfigPackageConfigurationProvider : PackageConfigurationProvider {
```
sschuberth, line 54: I believe this should be non-`open`. (I'll later provide an alignment PR that does the same for `OrtConfigPackageCurationProvider`.)

sschuberth (changes requested):

model/src/main/resources/reference.yml
MarcelBochtler: Yes, I also think that this makes more sense. Fixed.

sschuberth (approved):

Issue #5017, opened 2022-02-02
WebApp Reporter: License expressions from ScanCode should be displayed in report


MarcelBochtler:
The `scan-result.yml` contains license findings with a choice for the jhipster package:
```yml
- license: "Apache-2.0 OR GPL-1.0-or-later"
  location:
    path: "META-INF/maven/io.github.jhipster/jhipster-framework/pom.xml"
    start_line: 21
    end_line: 24
- license: "Apache-2.0 OR GPL-2.0-or-later WITH Classpath-exception-2.0"
  location:
    path: "META-INF/maven/io.github.jhipster/jhipster-framework/pom.xml"
    start_line: 18
  end_line: 21
```

The WebApp reporter does not display these license choices:
![SCR-20220202-nzp](https://user-images.githubusercontent.com/4022103/152193710-cac30589-cdd3-4737-a90b-f005449530ec.png)

tsteenbe:
@MarcelBochtler  The WebApp reporter may show the real issue is that the EvaluatedModel is still using [LicenseId](https://github.com/oss-review-toolkit/ort/blob/master/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedFinding.kt#L33) instead of SpdxExpression.

sschuberth:
More generally, we should make it easy for users of the web-app report to copy & paste the concluded license (that might still contain a choice) to a license choice configuration in order to tweak the effective license (/cc @mmurto). I'll adapt the issue title accordingly.

PR #7300, opened 2023-07-18
General license decomposition

sschuberth:


model/src/main/resources/reference.yml
```diff
@@ -210,8 +210,6 @@ ort:
         # Command line options that do not affect the ScanCode output.
         commandLineNonConfig: '--processes 4'
```
fviernau, line 213: commit-msg: Would it make sense to explicitly say that ORT doesn't want to support any version below 3?

model/src/main/resources/reference.yml
```diff
@@ -210,8 +210,6 @@ ort:
...
```
fviernau, line 212: > It anyway is better implemented as a general post-processing
of scan results from all scanners that decomposes the licenses if
required.

We recently had a discussion about the programmatic use of the scanners, and what kind of guarantees the interface should provide. To my understanding, the consensus was that the interface should remain strict and not make too much assumptions about further processing of the returned scan results.  Has there been any changes to that consensus which I missed? (I'm asking that because, IIUC this part of the commit message again assumes the "weakening" of the interface of the individual scanners to be the right way to go, which IMO contradicts to what we agreed upon. What do you think?

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -617,12 +617,16 @@ class Scanner(
 
     private fun postProcessScanSummary(summary: ScanSummary, downloadDir: File) =
         summary.copy(
-            licenseFindings = summary.licenseFindings.mapTo(mutableSetOf()) {
-                val licenseString = it.license.toString()
-                it.copy(
-                    license = licenseString.mapLicense(scannerConfig.detectedLicenseMapping).toSpdx(),
-                    location = it.location.withRelativePath(downloadDir)
-                )
+            licenseFindings = summary.licenseFindings.flatMapTo(mutableSetOf()) {
```
fviernau, line 620: Am I right this is done before writing the Summary to the storages? 
If so, I'm strongly vote not doing it here, but after the summary gets pulled out from storage, so that one can change this configuration parameter without invalidating the scan cache.

fviernau (changes requested):

model/src/main/resources/reference.yml
sschuberth: At least it was not my intention to say that. But as soon as version 3 is *also* supported, the code becomes a bit cumbersome to maintain if the `parseLicenseExpressions` feature stays.

model/src/main/resources/reference.yml
sschuberth: > To my understanding, the consensus was that the interface should remain strict and not make too much assumptions about further processing of the returned scan results.

That was not my understanding. *My* understanding was that we agreed that any post-processing should become part of the `(Path)ScannerWrapper`'s API eventually. That is, basically just the `postProcessScanSummary()` function would be moved from `Scanner` to `(Path)ScannerWrapper`, and the latter would call it as part of a default `scanPath()` implementation after calling (a yet to be introduced) `scanPathInternal()`.

Ultimately, I have two goals:
- avoid code duplication by not requiring scanner implementations to repeatedly implement the same post-processing steps (also not via helper functions that we provide, but must be called manually)
- safe-guard against "bad" scanner implementations by not allowing them to override / skip any of the mandatory post-processing

But I'm completely open to implementing these two points differently than outlined above. As mentioned in an earlier discussion, I'm not a fan of having both `scanPath()` and `scanPathInternal()`. An alternative could be to have a low-level scanner interface that scanners are supposed to implement, but for *using* scanners another high-level interface that implements the post-processing after calling the low-level interface must be used.

scanner/src/main/kotlin/Scanner.kt
sschuberth: My goal was to not change semantics compared to before. Both the original `parseLicenseExpressions` and now `decomposeLicenses` happen at the same time in this regard, i.e. before writing to the scan storage.

sschuberth: And BTW, I'd more more than happy to drop `parseLicenseExpressions` completely without any substituting feature. I just added `decomposeLicenses` as I feared there would be too much push-back for this PR if I do not provide a substituting feature.

sschuberth: @oss-review-toolkit/core-devs any feedback on whether `parseLicenseExpressions` for ScanCode can be dropped without introducing `decomposeLicenses` for all scanners?

sschuberth: BTW, I just realized there is https://github.com/oss-review-toolkit/ort/issues/5017, which IIUC means that even with `parseLicenseExpressions = true` at least the web-app report doe snot show license expressions. Which to me is more more reason that `parseLicenseExpressions` can safely be removed and make it default to `true` which changing behavior.

fviernau: > can safely be removed and make it default to `true` which changing behavior.

So, the target behavior is as follows, right? 
1. Each compound license expression returned by ScanCode will remain a compound license expression in   
    ScanSumamry?
2. The ScanCode e2e integration won't decompose any license expressions

sschuberth: That's my proposal, yes. (Note the "which" -> "without" typo I made in the text you quoted.) That is, depending on the discussion outcome, I'd simply drop the last commit form this PR.

mnonnenmacher: IIRC the main reason to introduce the `parseLicenseExpressions` flag was that at the time we were concerned that enabling it at HERE would require a large rework of the existing package configurations and might require an update of the rules, both of which we did not have capacity for back then (see  04584d2). However, I wonder if nowadays anyone is actually setting this flag to `false`.

sschuberth: > However, I wonder if nowadays anyone is actually setting this flag to `false`.

Well, it's `false` by default still (`isTrue()` returns `false` for `null`, i.e. if the option is unset).

sschuberth:
I propose to merge https://github.com/oss-review-toolkit/ort/pull/7331 instead for now.

PR #7308, opened 2023-07-19
Two FossID fixes

nnobelis:

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -500,7 +500,8 @@ class FossId internal constructor(
             }
         }
 
-        val mappedUrlWithoutCredentials = urlProvider.getUrl(urlWithoutCredentials)
+        val mappedUrl = urlProvider.getUrl(urlWithoutCredentials)
+        val mappedUrlWithoutCredentials = mappedUrl.replaceCredentialsInUri()
 
         // we ignore the revision because we want to do a delta scan
         val recentScans = scans.recentScansForRepository(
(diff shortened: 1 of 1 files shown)


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -500,7 +500,8 @@ class FossId internal constructor(
             }
         }
 
-        val mappedUrlWithoutCredentials = urlProvider.getUrl(urlWithoutCredentials)
+        val mappedUrl = urlProvider.getUrl(urlWithoutCredentials)
```
sschuberth, line 503: Commit message:
- Typo in "UR*L*"
- I believe we should drop "Previous commit introduced such a bug" as "Previous" is misleading (it's not the parent of this commit) and you anyway mention the commit to be fixed-up.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -534,6 +534,20 @@ class FossId internal constructor(
             .checkResponse("download data from Git", false)
 
         if (existingScan == null) {
+            val excludesRules = context.excludes?.let {
```
sschuberth, line 537: Commit message:
- "This commit change*s*"
- "in this case" should probably say "in the other case"

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -534,6 +534,20 @@ class FossId internal constructor(
...
+                convertRules(it, issues).also {
+                    logger.info { "${it.size} rules from ORT excludes have been found." }
+                }
+            }.orEmpty()
+
+            excludesRules.forEach {
+                service.createIgnoreRule(config.user, config.apiKey, scanCode, it.type, it.value, RuleScope.SCAN)
+                    .checkResponse("create ignore rules", false)
+                logger.info {
```
sschuberth, line 546: Nit: Let's add a blank line before this one.

sschuberth (changes requested):

sschuberth (approved):

sschuberth (approved):

PR #7309, opened 2023-07-19
Two Snippet Reporter fixes

nnobelis:

### DIFF
--- a/plugins/reporters/fossid/src/main/kotlin/FossIdSnippetReporter.kt
+++ b/plugins/reporters/fossid/src/main/kotlin/FossIdSnippetReporter.kt
@@ -38,6 +38,9 @@ class FossIdSnippetReporter : Reporter by delegateReporter {
     override val type = "FossIdSnippet"
 
     override fun generateReport(input: ReporterInput, outputDir: File, options: Map<String, String>): List<File> {
+        val hasFossIdResults = input.ortResult.scanner?.scanResults?.any { it.scanner.name == "FossId" } ?: false
+        require(hasFossIdResults) { "No FossID scan results have been found." }
+
         val extendedOptions = options + (FreemarkerTemplateProcessor.OPTION_TEMPLATE_ID to TEMPLATE_NAME)
         return delegateReporter.generateReport(input, outputDir, extendedOptions)
     }
--- a/plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
+++ b/plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
@@ -30,14 +30,20 @@ List of all the provenances with their files and snippets.
 [#list ortResult.scanner.scanResults as scanResult]
 
 [#if scanResult.scanner.name != "FossId"] [#continue] [/#if]
-== Provenance '${scanResult.provenance.vcsInfo.url}'
+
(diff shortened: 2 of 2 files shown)


plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
```diff
@@ -29,6 +29,7 @@
 List of all the provenances with their files and snippets.
 [#list ortResult.scanner.scanResults as scanResult]
 
+[#if scanResult.scanner.name == "FossId"]
```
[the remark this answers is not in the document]
nnobelis: Yes I agree. However please note that the line you commented on just filters the FossID scan results, for the case there is a mix of FossID and ScanCode scan results.

plugins/reporters/fossid/src/main/kotlin/FossIdSnippetReporter.kt
```diff
@@ -38,6 +38,9 @@ class FossIdSnippetReporter : Reporter by delegateReporter {
     override val type = "FossIdSnippet"
 
     override fun generateReport(input: ReporterInput, outputDir: File, options: Map<String, String>): List<File> {
+        val hasFossIdResults = input.ortResult.scanner?.scanResults?.any { it.scanner.name == "FossId" } ?: false
+        check(hasFossIdResults) { "No FossID scan results have been found." }
```
sschuberth, line 42: This now makes me think again about the general question how a reporter should be have in case of empty / invalid input. Failing like this IMO makes the reporter command exit with a code > 0, which will probably break your CI pipeline. And that might be a problem if you simply always activate FossID reports, even for pipeline runs that do not use FossID as a scanner.

plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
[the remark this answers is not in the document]
nnobelis: Yes I tried to add indentation but it messed up the AsciIDoc rendering, because it interpreted the tabs.

plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
```diff
@@ -29,6 +29,7 @@
 List of all the provenances with their files and snippets.
 [#list ortResult.scanner.scanResults as scanResult]
 
+[#if scanResult.scanner.name != "FossId"] [#continue][/#if]
```
sschuberth, line 32: Nit: For one-liners like this, I believe we should either also add a space after `[#continue]`, or omit the one before. In any case, be consistent WRT spacing. Or make it a three-liner with indentation 😉

plugins/reporters/fossid/src/main/kotlin/FossIdSnippetReporter.kt
sschuberth: > Should I change something ?

Not necessarily, I just wanted to put this up for consideration. Maybe you could cross-check how other reporters behave in this regard. I have a hunch that they do *not* fail, but print some custom text into the report instead.

But actually, if you keep this code, `check` should become `require` (as you're verifying input, not state).

sschuberth (approved):

PR #7310, opened 2023-07-19
fix(model): Use the correct PURL type for CocoaPods

ksg97031:
According to the purl-spec and ossindex, the name of CocoaPods is "CocoaPods" not "Pod".

[1] https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#cocoapods
[2] https://ossindex.sonatype.org/ecosystem/cocoapods

### DIFF
--- a/model/src/main/kotlin/utils/Extensions.kt
+++ b/model/src/main/kotlin/utils/Extensions.kt
@@ -207,6 +207,7 @@ fun Identifier.getPurlType() =
         "maven" -> PurlType.MAVEN
         "npm" -> PurlType.NPM
         "nuget" -> PurlType.NUGET
+        "pod" -> PurlType.COCOAPODS
         "pypi" -> PurlType.PYPI
         else -> lowerType
     }.toString()
--- a/plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/dep-tree-expected-output.yml
+++ b/plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/dep-tree-expected-output.yml
@@ -115,7 +115,7 @@ project:
     - id: "Pod::base64url:1.1.0"
 packages:
 - id: "Pod::Alamofire:5.4.3"
-  purl: "pkg:pod/Alamofire@5.4.3"
+  purl: "pkg:cocoapods/Alamofire@5.4.3"
   declared_licenses:
   - "MIT"
   declared_licenses_processed:
@@ -143,7 +143,7 @@ packages:
     revision: "5.4.3"
     path: ""
 - id: "Pod::CryptoSwift:1.4.0"
-  purl: "pkg:pod/CryptoSwift@1.4.0"
+  purl: "pkg:cocoapods/CryptoSwift@1.4.0"
   declared_licenses:
   - "Attribution"
   declared_licenses_processed:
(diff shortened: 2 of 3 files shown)


sschuberth (changes requested):
Thanks for your contribution! Please use `fix` as the Conventional Commit type, like `fix(model): Use the correct PURL type for CocoaPods`.

sschuberth:
> Thanks for your contribution! Please use `fix` as the Conventional Commit type, like `fix(model): Use the correct PURL type for CocoaPods`.

Also footers that refer to URLs should be referred to from the body text, and the foot notes should have a colon after the square brackets. See e.g. 4f9609e979ddfe1422e6241a1a1b201f510eb781.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -253,7 +253,7 @@ private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
 
     fun createPackageReference(name: String): PackageReference =
         PackageReference(
-            id = Identifier("Pod", "", name, versionForName.getValue(name)),
+            id = Identifier("CocoaPods", "", name, versionForName.getValue(name)),
```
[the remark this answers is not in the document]
sschuberth: This actually needs to be reverted and *only* the change in the other file be kept.

model/src/main/kotlin/utils/Extensions.kt
```diff
@@ -208,6 +208,7 @@ fun Identifier.getPurlType() =
         "npm" -> PurlType.NPM
         "nuget" -> PurlType.NUGET
         "pypi" -> PurlType.PYPI
+        "cocoapods" -> PurlType.COCOAPODS
```
sschuberth, line 211: This needs to map `"pod"` (lowercase name of ORT's internal type) to `PurlType.COCOAPODS`. And please sort this in alphabetically before the `"pypi"` line.

sschuberth (changes requested):

sschuberth:
> According to the [1] and [2], the name of CocoaPods is "CocoaPods" not "Pod".

Also, please amend your commit and hard-wrap this line after 75 chars.

model/src/main/kotlin/utils/Extensions.kt
ksg97031: I'm sorry, I just understood.

sschuberth (approved):

sschuberth:
Good point about the test fixes. Please also fix the occurrences in [plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/dep-tree-expected-output.yml](https://github.com/oss-review-toolkit/ort/blob/main/plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/dep-tree-expected-output.yml) and simply amend your first commit with all the changes to tests.

sschuberth:
Please don't put merge commits into PRs, but rebase instead.

ksg97031:
> Please don't put merge commits into PRs, but rebase instead.

I've rebased the commit to avoid merge commits. 
I apologize for any inconvenience this may cause.

sschuberth:
> I apologize for any inconvenience this may cause.

No need to apologize. Thanks for addressing all review comments so quickly!

sschuberth (approved):

sschuberth:
Merging despite the unrelated `OsvFunTest` failure.

PR #7316, opened 2023-07-20
fix(ScanSummary): Turn `licenses` into a `lazy` initialized property

fviernau:
When the SPDX expressions are mapped into a `MutableSet`, then `equals` gets called on each added SPDX expression instance. Equals in turn calls `validChoicesForDnf()` which is quite slow.

Turn the property into a lazy one, to remove the performance penalty from code paths which do not access that `licenses` property.


sschuberth (approved):

fviernau:
The failing test was unrelated, merge regardlessly.

PR #7317, opened 2023-07-20
chore(scanner): Log the beginning and end of merging scan results

fviernau:
For larger scan results this operation may take a couple of seconds. Add log output in order to indicate what is being processed.

### DIFF
--- a/model/src/main/kotlin/ScannerRun.kt
+++ b/model/src/main/kotlin/ScannerRun.kt
@@ -25,6 +25,10 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize
 import java.io.File
 import java.time.Instant
 
+import kotlin.time.measureTimedValue
+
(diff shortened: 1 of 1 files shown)


model/src/main/kotlin/ScannerRun.kt
```diff
@@ -25,6 +25,9 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize
 import java.io.File
 import java.time.Instant
 
+import kotlin.time.measureTimedValue
+
+import org.apache.logging.log4j.kotlin.Logging
```
sschuberth, line 30: Wrong import grouping.

sschuberth (changes requested):

sschuberth (approved):

fviernau:
Merge regardless of the failing unrelated OsvFunTest.

PR #7319, opened 2023-07-20
fix(ResolvedLicenseInfo): Eliminate unneeded DNF computations

fviernau:
Applying license choices is a slow operation, in particular `validChoices()` and `validChoicesForDnf()`. So, not executing the logic for applying the license choice in case the given flattened `licenseChoices` improves the performance quite a bit.

Furthermore, given some large SPDX expression the logic has been observed to execute for minutes at maximum CPU load eventually ending with a crash, for yet unknown reason. This change fixes that issue in case no license choice is given for the particular package.

### DIFF
--- a/model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
+++ b/model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
@@ -72,13 +72,17 @@ data class ResolvedLicenseInfo(
     fun effectiveLicense(licenseView: LicenseView, vararg licenseChoices: List<SpdxLicenseChoice>): SpdxExpression? {
         val resolvedLicenseInfo = filter(licenseView, filterSources = true)
 
-        return resolvedLicenseInfo.licenses.flatMap { resolvedLicense ->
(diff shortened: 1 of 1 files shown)


model/src/main/kotlin/licenses/ResolvedLicenseInfo.kt
```diff
@@ -72,13 +72,17 @@ data class ResolvedLicenseInfo(
     fun effectiveLicense(licenseView: LicenseView, vararg licenseChoices: List<SpdxLicenseChoice>): SpdxExpression? {
         val resolvedLicenseInfo = filter(licenseView, filterSources = true)
 
-        return resolvedLicenseInfo.licenses.flatMap { resolvedLicense ->
+        val detectedLicense = resolvedLicenseInfo.licenses.flatMap { resolvedLicense ->
```
sschuberth, line 75: Nit: These are not necessarily only "detected" licenses, or? So maybe better call this `resolvedLicenses`?

sschuberth: Thinking about it more, maybe `resolvedLicenseExpression` would be even more fitting. But I'm fine with keeping it as-is now.

PR #7324, opened 2023-07-21
Fix PURLs created by the `SpdxDocumentFile` package manager

sschuberth:


model/src/main/kotlin/utils/Extensions.kt
```diff
@@ -197,11 +197,11 @@ enum class PurlType(private val value: String) {
 }
 
 /**
- * Map a [Package]'s type to the string representation of the respective [PurlType], or fall back to the lower-case
```
MarcelBochtler, line 197: Commit message: Typo in `uyse`.

MarcelBochtler (changes requested):

MarcelBochtler (approved):

PR #7331, opened 2023-07-25
refactor(scanner)!: Always parse ScanCode license expressions

sschuberth:
Remove the feature to toggle parsing of ScanCode license expressions. Starting with ScanCode output format version 3, license expressions are first class citizens in ScanCode results, making this feature cumbersome to maintain.

Note that the web-app report still would not show license expressions anyway (see issue #5017).

> **Note**
> This is a breaking change for the configuration as the `parseLicenseExpressions` option for ScanCode is being removed. Users that rely on `parseLicenseExpressions = false` for historic reasons might need to update license curations to properly use expressions.

### DIFF
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -211,8 +211,6 @@ ort:
         # Command line options that do not affect the ScanCode output.
         commandLineNonConfig: '--processes 4'
 
-        parseLicenseExpressions: true
-
         # Criteria for matching stored scan results. These can be configured for any scanner that uses semantic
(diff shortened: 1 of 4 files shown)


sschuberth:
@oss-review-toolkit/kotlin-devs, as @bennati [confirmed](https://ort-talk.slack.com/archives/C9NNJ54B1/p1692606504242969?thread_ts=1692606220.078649&cid=C9NNJ54B1) that HERE is not using `parseLicenseExpressions = false`, please review.

mnonnenmacher (approved):

PR #7360, opened 2023-08-04
fix(FossId): Allow directories with dots to be added as ignore rules

MarcelBochtler:
Fix the FossID ignore rule creation for directories using a dot in the
name. While the regex is far from perfect [1], this fixes the very common
use case of ignoring `.git/` and `.idea/` directories.

[1]: For instance, `**/example/` is not added as directory ignore rule.


nnobelis (approved):
LGTM.

However using `.git` as an example is a bit unlucky as you should never have to manually create an ignore rule for this directory: FossID automatically creates it when scanning a Git repository.

mnonnenmacher (approved):

PR #7367, opened 2023-08-07
feat(spdx): Allow LicenseRef exceptions for licenseInfoInFiles 

MarcelBochtler:
While the [SPDX specification](https://spdx.github.io/spdx-spec/v2-draft/file-information/#861-description) doesn't explicitly allow exceptions for `licenseInfoInFiles` it also doesn't disallow them.
IMO ORT should allow SPDX expressions in it's created SpdxDocument.
Or am I misinterpreting the SPDX specifications @oss-review-toolkit/core-devs?

This fixes an issue in a project where scancode detected `GPL-2.0-or-later WITH LicenseRef-scancode-autoconf-simple-exception-2.0`, causing the `SpdxDocument` reporter break:
https://github.com/java-native-access/jna/blob/c32053e3137d7af5764ed4148e866d28eb8fd398/native/libffi/.ci/ar-lib#L10-L30

### DIFF
--- a/utils/spdx/src/main/kotlin/Extensions.kt
+++ b/utils/spdx/src/main/kotlin/Extensions.kt
@@ -74,11 +74,11 @@ fun String.isSpdxExpression(strictness: Strictness = Strictness.ALLOW_DEPRECATED
     runCatching { SpdxExpression.parse(this, strictness) }.isSuccess
 
 /**
- * Return true if and only if this String can be successfully parsed to an [SpdxExpression] or if it equals
- * [SpdxConstants.NONE] or [SpdxConstants.NOASSERTION].
+ * Return true if and only if this String can be successfully parsed to an [SpdxExpression] with the given [strictness]
+ * or if it equals [SpdxConstants.NONE] or [SpdxConstants.NOASSERTION].
  */
-fun String.isSpdxExpressionOrNotPresent(): Boolean =
-    SpdxConstants.isNotPresent(this) || isSpdxExpression()
+fun String.isSpdxExpressionOrNotPresent(strictness: Strictness = Strictness.ALLOW_DEPRECATED): Boolean =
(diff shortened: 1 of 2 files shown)


tsteenbe (changes requested):
While the [SPDX specification](https://spdx.github.io/spdx-spec/v2-draft/file-information/#861-description) doesn't explicitly allow exceptions for licenseInfoInFiles it also doesn't disallow them.

@MarcelBochtler  Custom exceptions (LicenseRef-scancode-autoconf-simple-exception-2.0) are not allowed per the SPDX specification - this will be fixed in 3.0 with the introduction of AdditionRef.  Exceptions are always part of license expression and per https://spdx.github.io/spdx-spec/v2-draft/SPDX-license-expressions/ license-exception-id are only those defined in section [A.2](https://spdx.github.io/spdx-spec/v2-draft/SPDX-license-list/#a2-exceptions-list)

A correct way to solve getting invalid SPDX expressions from ScanCode would be to convert cases such as 'GPL-2.0-or-later WITH LicenseRef-scancode-autoconf-simple-exception-2.0' into 'LicenseRef-ort-GPL-2.0-or-later-WITH-autoconf-simple-exception-2.0'.  Otherwise SPDX generated by ORT can not be consumed by other application capable of ingesting SPDX

mnonnenmacher:
To me it makes sense that the SPDX reporter strictly adheres to the specs by default. As a compromise, this feature could be made optional with a reporter option for users who want to allow license ref exceptions, as other reporters allow this as well.
This should also be reported as a bug to ScanCode, it should return a valid expression like "LicenseRef-scancode-GPL-2.0-or-later-WITH-autoconf-simple-exception-2.0" instead of an invalid exception.

MarcelBochtler:
Superseded by https://github.com/oss-review-toolkit/ort/pull/7399

PR #7374, opened 2023-08-11
test(spm): Update expected results

mnonnenmacher:


mnonnenmacher:
Self-approved as currently no reviewers are available.

PR #7377, opened 2023-08-15
test(python): Update an expected result

fviernau:


tsteenbe (approved):

fviernau:
self approved as no member from kotlin core devs is available for approval and the change is trivial.

PR #7385, opened 2023-08-17
build: Update Gradle to version 8.3

sschuberth:
See [1]. Address some deprecations that result in warnings along the way.

[1]: https://docs.gradle.org/8.3/release-notes.html

### DIFF
--- a/plugins/reporters/web-app/build.gradle.kts
+++ b/plugins/reporters/web-app/build.gradle.kts
@@ -22,7 +22,7 @@ plugins {
     id("ort-library-conventions")
 }
 
-val generatedResourcesDir = file("$buildDir/generated-resources/main")
+val generatedResourcesDir = layout.buildDirectory.dir("generated-resources/main")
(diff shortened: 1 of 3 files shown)


plugins/reporters/web-app/build.gradle.kts
```diff
@@ -22,7 +22,7 @@ plugins {
     id("ort-library-conventions")
 }
 
-val generatedResourcesDir = file("$buildDir/generated-resources/main")
+val generatedResourcesDir = layout.buildDirectory.dir("generated-resources/main")
```
fviernau, line 25: Just wondering `.get().asFile` isn't needed?

sschuberth: It's not needed here as `into()` below also accepts `Provider<...>`.

fviernau (approved):

sschuberth:
Merging despite the known / temporary SW360 build failure.

Issue #3296, opened 2020-11-06
scanner runs very slow / no status


nupis-FrankO:
Hi,

we are testing ORT at the moment to scan/analyze a maven project.

The scan runs very slow (> 1 day) and we don't see any state of the current execution. Is it possible to implement a "progress bar"?

Or is there a possibility to optimize the execution of the scan process?

Regards,

Frank

sschuberth:
I have also seen bad performance for scanning large Maven projects recently, and we're planning to investigate these, as even with debug output sometimes nothing seem to happen for minutes. Meanwhile, can you try increasing the max heap size for ORT to something like 16 GB (`-Xmx16g`) to see if that helps?

sachinshaji:
HI, We are also facing the same problem. Scanner takes lot of time.

sschuberth:
The scanner (i.e. ScanCode, by default) taking a lot of time is nothing ORT can change. However, we could probably try to do a better job at reporting progress (although you actually do get per package progress with `--info`).

In any case, you should set up a scan storage to benefit from existing scan results. Giving the ClearlyDefined scan storage a try is something you could do, for example.

sachinshaji:
Thanks for the replay. 
I have build a docker image form the Dockerfile given in the repo. I am triggering this command to scan the code
sudo docker run -v $PWD/:/project <ort-image> --info scan -i /project/analyse/analyzer-result.yml -o /project/scanner/scanner-result.json

We are not defining any storage backend, i guess this will default goes to filesystem.
Any suggestion to improve the scanning time? It took around 3 to 4 hours to scan the entire code.

sschuberth:
> We are not defining any storage backend, i guess this will default goes to filesystem.

Correct. And this only speeds up the scanning for consecutive scans that involve mostly the same packages, as you're populating our file-based scan storage yourself.

> Any suggestion to improve the scanning time?

The idea is that you either quickly build up some company-internal (Postgres-based) scan storage yourself to speed up future scans, or to use an existing public scan storage, like the one from ClearlyDefined as mentioned above. However, might be that [we currently have an issue there](https://github.com/oss-review-toolkit/ort/issues/4540).

sschuberth:
Closed as part of backlog grooming. Feel free to comment if you would like to contribute to this.

PR #7391, opened 2023-08-18
feat(cli): Add the flag `--monitor-stack-traces`

fviernau:
If `ort` executes slowly this can become handy for getting a first hint where the CPU work is happening.

Note: Print the stack traces in reverse order and the `main` thread's trace at the bottom to have the likely desired information at the most prominent place.

### DIFF
--- a/cli/src/main/kotlin/OrtMain.kt
+++ b/cli/src/main/kotlin/OrtMain.kt
@@ -42,6 +42,8 @@ import com.github.ajalt.mordant.rendering.Widget
 import com.github.ajalt.mordant.table.ColumnWidth
 import com.github.ajalt.mordant.table.grid
 
+import java.time.Duration
+
 import kotlin.system.exitProcess
 
 import org.apache.logging.log4j.kotlin.Logging
@@ -104,6 +106,8 @@ class OrtMain : CliktCommand(
         "--debug" to Level.DEBUG
     ).default(Level.WARN)
 
+    private val monitorStackTraces by option(help = "Continuously print out the stack traces of all threads.").flag()
+
     private val stacktrace by option(help = "Print out the stacktrace for all exceptions.").flag()
 
     private val configArguments by option(
@@ -145,6 +149,8 @@ class OrtMain : CliktCommand(
 
         // Make the parameter globally available.
         printStackTrace = stacktrace
+
+        if (monitorStackTraces) startStackTraceMonitoring()
 
         // Make options available to subcommands and apply static configuration.
         val ortConfig = OrtConfiguration.load(args = configArguments, file = configFile)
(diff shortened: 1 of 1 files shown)


sschuberth:
Hmm, I'm unsure whether this is a feature that we should add. It sounds more like something that should be done during debugging with the means of an IDE. Or are there really cases you'd like to inspect e.g. on CI that absolutely cannot be reproduced locally?

Also, this reminds me a bit of https://github.com/oss-review-toolkit/ort/issues/3296. IMO a much nicer (but also more involving) solution for monitoring would be to print better progress information on the CLI, similar to like Gradle does when building with multiple threads in parallel.

fviernau:
> Hmm, I'm unsure whether this is a feature that we should add. It sounds more like something that should be done during debugging with the means of an IDE. 

Can you explain which exact means you are referring to, and why these means should be preferable.
(If you're talking about a profiler, then I'd say the use cases are related but both approaches has pros and cons. If I used
the profiler for profiling my evaluator run, I would have waited ages for it to finish due to the large ORT file size. I'm not saying the profiler is not suitable, but looking at the thread dumps certainly gave me some insight more quickly.) 

> Or are there really cases you'd like to inspect e.g. on CI that absolutely cannot be reproduced locally?

It's not for CI, it's for local investigations. It could also be local on some other devs machine, so remote to myself.

> Also, this reminds me a bit of #3296. IMO a much nicer (but also more involving) solution for monitoring would be to print better progress information on the CLI, similar to like Gradle does when building with multiple threads in parallel.

In my use case the CLI output of ORT just didn't print anything, so I wanted to get a hunch on what it's doing.
If one could anticipate all performance problems and put logging making it visible everywhere, then the performance problem wouldn''t exist in the first place.

I've written this code already a couple of times and I wanted to save that effort for myself and others as well.
So, do you think seeing stack traces cannot be useful for any developer using ORT ?

sschuberth:
> Can you explain which exact means you are referring to

I wasn't thinking about using a profiler, but the threads view of the debugger tab in the IDE. You can pause execution of threads and look at the stacktrace at any time.

> and why these means should be preferable.

Since it's you who wants to introduce this change, could you please explain why your approach should be preferred over dedicated debugging tools?

> In my use case the CLI output of ORT just didn't print anything

Yes, that's exactly what #3296 is about: Missing information about progress. This should not necessarily be implemented by simply adding more log statement, but ideally by implementing progress interfaces. But that's, as mentioned, a bigger effort.

> If one could anticipate all performance problems and put logging making it visible everywhere, then the performance problem wouldn''t exist in the first place.

I don't get that sentence. How does more logging automatically solve performance problems?

> So, do you think seeing stack traces cannot be useful for any developer using ORT ?

Obviously, I cannot tell for the rest of @oss-review-toolkit/kotlin-devs, but I myself have never needed it / more than what the IDE debugging offers.

> Would it work for you if this feature was toggled via env var instead via a CI param, or is the feature itself not convincing?

To me, the feature itself is not convincing, as it seems to reimplement something that is already available in the IDE, or via some of the tools e.g. mentioned [here](https://www.baeldung.com/java-thread-dump).

fviernau:
`jstack` can be used for this purpose, e.g.

PR #7394, opened 2023-08-21
refactor(model)!: Make curations a list in `ResolvedPackageCurations`

mnonnenmacher:
The `curations` property is supposed to contain the curations in the order they were provided by the package curation provider, therefore change the type from set to list to preserve the order.

### DIFF
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -539,7 +539,7 @@ data class OrtResult(
                     if (it.provider.id != REPOSITORY_CONFIGURATION_PROVIDER_ID) {
                         it
                     } else {
-                        it.copy(curations = config.curations.packages.toSet())
+                        it.copy(curations = config.curations.packages)
                     }
                 }
             )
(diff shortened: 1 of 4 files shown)


sschuberth (approved):

model/src/main/kotlin/OrtResult.kt
```diff
@@ -539,7 +539,7 @@ data class OrtResult(
                     if (it.provider.id != REPOSITORY_CONFIGURATION_PROVIDER_ID) {
                         it
                     } else {
-                        it.copy(curations = config.curations.packages.toSet())
+                        it.copy(curations = config.curations.packages)
```
fviernau, line 542: commit-message: The set implementation used also preserves the order, but drops duplicates. Should this be mentioned instead in the commit message?

sschuberth: > The set implementation used also preserves the order, but drops duplicates.

I was thinking the same, but according to a [test failure](https://github.com/oss-review-toolkit/ort/actions/runs/5925765390/job/16065823206?pr=7394#step:6:1851) there's now a change in order!?

mnonnenmacher: Yes, the order in the result file was different than the order in the curations file, so it was actually wrong.

PR #7398, opened 2023-08-21
fix(SpdxExpression): Fix the DNF algorithm

MarcelBochtler:
Whenever a `SpdxExpression` contains an `AND` operator, the DNF needs to be calculated again.
For instance, the correct DNF for `a AND (b OR c OR d)` is `(a AND b) OR (a AND c) OR (a AND d)`. Before, the calculated DNF for this expression was `(a AND (b OR c)) OR (a AND d)`, which is not a valid DNF.


MarcelBochtler:
Pushed the wrong branch. Closing.

PR #7399, opened 2023-08-21
feat(spdx): Allow LicenseRef exceptions for licenseInfoInFiles

MarcelBochtler:


fviernau:
@tsteenbe: This PR now superseeds the linked one, for which you had some objections. As discussed, I've added the log output which lists the problematic LicenseRef- exceptions, to mitigate the concerns. Please take a look if you're fine with the PR now, not necessarily as a long term, but as a short term solution to eliminate the crash from the SPDX reporter.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
```diff
@@ -89,3 +104,25 @@ class SpdxDocumentReporter : Reporter {
         }
     }
 }
+
+private fun SpdxDocument.getLicenseRefExceptions(): Set<String> =
+    buildSet {
+        files.forEach { file ->
+            file.licenseInfoInFiles.forEach { license ->
```
sschuberth, line 111: Should we maybe implement this logic rather in a `List<String>`- / `String`-serializer that we could easily annotate more fields than just `licenseInfoInFiles` with?

fviernau: Not sure if I get your point. I guess, it's fine to do that once it is needed.

utils/spdx/src/main/kotlin/model/SpdxFile.kt
```diff
@@ -187,7 +188,9 @@ data class SpdxFile(
 
         // TODO: The check for [licenseInfoInFiles] can be made more strict, but the SPDX specification is not exact
         //       enough yet to do this safely.
-        licenseInfoInFiles.filterNot { it.isSpdxExpressionOrNotPresent() }.let {
+        licenseInfoInFiles.filterNot {
+            it.isSpdxExpressionOrNotPresent(SpdxExpression.Strictness.ALLOW_LICENSEREF_EXCEPTIONS)
```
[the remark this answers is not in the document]
sschuberth: > SPDX v2 does not allow using custom (`LicenseRef-`) IDs as exceptions
and an SPDX expression.

I believe this is still confusing for several reasons:
- There is no need to limit to SPDX v2. SPDX v1 is affacted as well, and for SPDX v3 it's not yet fully clear if and how the issue will be resolved.
- It's confusing to quote *License*Ref when talking about exceptions.
- I don't get the "and an SPDX expression" end of the sentence.

I propose:

   Currently, there is no syntax for custom exceptions to licenses defined in the SPDX specification.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
```diff
@@ -21,8 +21,14 @@ package org.ossreviewtoolkit.plugins.reporters.spdx
 
 import java.io.File
 
+import org.apache.logging.log4j.kotlin.Logging
+
 import org.ossreviewtoolkit.reporter.Reporter
 import org.ossreviewtoolkit.reporter.ReporterInput
+import org.ossreviewtoolkit.utils.spdx.SpdxCompoundExpression
+    import org.ossreviewtoolkit.utils.spdx.SpdxConstants.LICENSE_REF_PREFIX
```
sschuberth, line 29: Wrong indentation.

utils/spdx/src/main/kotlin/model/SpdxFile.kt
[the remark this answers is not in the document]
sschuberth: Also, I believe same change should be done for the very similar code in `SpdxPackage.kt` for consistency:

https://github.com/oss-review-toolkit/ort/blob/88138faf748d06e2c17eb2e624c6a96313449d49/utils/spdx/src/main/kotlin/model/SpdxPackage.kt#L219-L221

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
sschuberth: > Not sure if I get your point.

My point is to implement this in a way so that the check can easily be applied to other properties, like `SpdxPackage.licenseInfoFromFiles`, without too much hard-coding or code duplication.

sschuberth (changes requested):

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
fviernau: I've adjusted the code to also consider `licenseInfoFromFiles`, which shows that the approach almost as-is works without too much hard-coding or code duplication.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
```diff
@@ -89,3 +105,29 @@ class SpdxDocumentReporter : Reporter {
         }
     }
 }
+
+private fun SpdxDocument.getLicenseRefExceptions(): Set<String> {
+    val licenses = buildSet {
+        this@getLicenseRefExceptions.files.flatMapTo(this) { it.licenseInfoInFiles }
+        this@getLicenseRefExceptions.packages.flatMapTo(this) { it.licenseInfoFromFiles }
```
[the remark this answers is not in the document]
sschuberth: Anyway, I believe `buildSet` does not really pay of here, and personally I'd write this as

    val licenses = mutableSetOf<String>()
    files.flatMapTo(licenses) { it.licenseInfoInFiles }
    packages.flatMapTo(licenses) { it.licenseInfoFromFiles }

But I agree it's a matter of taste, and I'm ok to keep it as-is.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
```diff
@@ -89,3 +105,29 @@ class SpdxDocumentReporter : Reporter {
...
+    return buildSet {
+        licenses.forEach { license ->
+            SpdxExpression.parse(license, SpdxExpression.Strictness.ALLOW_ANY).getLicenseRefExceptions(this)
+        }
+    }
+}
+
+private fun SpdxExpression.getLicenseRefExceptions(result: MutableSet<String>) {
+    when (this) {
+        is SpdxCompoundExpression -> run {
```
sschuberth, line 124: `run` should simply be omitted here.

sschuberth (changes requested):

sschuberth (approved):

PR #7401, opened 2023-08-21
fix(fossid-webapp): Group snippets by source file matching lines 

nnobelis:
Following the [17.07 Kotlin Developer meeting](https://github.com/oss-review-toolkit/ort/wiki/Kotlin-Developer-Meeting#2023-07-17), it was agreed that, _before the Snippet Choice feature is implemented in ORT_, the snippet model has to be changed to group the snippets by the source file location including the line ranges.

### DIFF
--- a/model/src/main/kotlin/SnippetFinding.kt
+++ b/model/src/main/kotlin/SnippetFinding.kt
@@ -20,8 +20,8 @@
 package org.ossreviewtoolkit.model
 
 /**
- * A class representing a snippet finding for a source file. A snippet finding is a code snippet from another origin,
- * matching the code being scanned.
+ * A class representing the snippet findings for a source file location. A snippet finding is a code snippet from
+ * another origin, matching the code being scanned.
  * It is meant to be reviewed by an operator as it could be a false positive.
  */
 data class SnippetFinding(
@@ -31,7 +31,7 @@ data class SnippetFinding(
     val sourceLocation: TextLocation,
 
     /**
-     * The corresponding snippet.
+     * The corresponding snippets.
      */
-    val snippet: Snippet
+    val snippets: Set<Snippet>
 )
--- a/model/src/main/kotlin/utils/SortedSetConverters.kt
+++ b/model/src/main/kotlin/utils/SortedSetConverters.kt
@@ -90,7 +90,7 @@ class ScopeSortedSetConverter : StdConverter<Set<Scope>, SortedSet<Scope>>() {
 
 class SnippetFindingSortedSetConverter : StdConverter<Set<SnippetFinding>, SortedSet<SnippetFinding>>() {
     override fun convert(value: Set<SnippetFinding>) =
-        value.toSortedSet(compareBy<SnippetFinding> { it.sourceLocation.path }.thenByDescending { it.snippet.purl })
+        value.toSortedSet(compareBy<SnippetFinding> { it.sourceLocation }.thenByDescending { it.snippet.purl })
 }
 
 private fun Provenance.getSortKey(): String =
--- a/model/src/main/kotlin/utils/SortedSetConverters.kt
+++ b/model/src/main/kotlin/utils/SortedSetConverters.kt
@@ -89,8 +89,7 @@ class ScopeSortedSetConverter : StdConverter<Set<Scope>, SortedSet<Scope>>() {
 }
 
 class SnippetFindingSortedSetConverter : StdConverter<Set<SnippetFinding>, SortedSet<SnippetFinding>>() {
-    override fun convert(value: Set<SnippetFinding>) =
-        value.toSortedSet(compareBy<SnippetFinding> { it.sourceLocation }.thenByDescending { it.snippet.purl })
+    override fun convert(value: Set<SnippetFinding>) = value.toSortedSet(compareBy { it.sourceLocation })
 }
 
 private fun Provenance.getSortKey(): String =
--- a/plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
+++ b/plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
@@ -53,40 +53,49 @@ End time : ${summary.startTime} +
     [/#if]
 [/#if]
 
-[#assign snippets = helper.groupSnippetsByFile(summary.snippetFindings)]
+[#list helper.groupSnippetsByFile(summary.snippetFindings) as filePath, snippetFindings ]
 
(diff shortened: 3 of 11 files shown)


model/src/main/kotlin/SnippetFinding.kt
```diff
@@ -26,9 +26,9 @@ package org.ossreviewtoolkit.model
  */
 data class SnippetFinding(
     /**
-     * The text location in the scanned source file where the snippet has matched.
+     * The text locations in the scanned source file where the snippet has matched.
      */
-    val sourceLocation: TextLocation,
+    val sourceLocation: TextWithMultipleLocations,
```
sschuberth, line 31: Sorry for the (maybe naive) question, but why's this not simply a list / set of `TextLocation`?

mnonnenmacher: My assumption was that the reason for introducing the new class was to deduplicate the path.

sschuberth: Actually, looking at the change again, I'm surprised that `val sourceLocation: TextLocation` is touched at all; I though what we discussed was to allow multiple snippets to be referenced from a single code location, so it's `val snippet: Snippet` that should change and become a collection, no?

model/src/main/kotlin/TextWithMultipleLocations.kt
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+package org.ossreviewtoolkit.model
+
+import org.ossreviewtoolkit.model.utils.prependPath
+
+/**
+ * A [TextLocation] references text located in a file at multiple positions.
+ */
+data class TextWithMultipleLocations(
```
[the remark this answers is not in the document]
nnobelis: I agree but I will wait for the other developers opinion.

model/src/main/kotlin/SnippetFinding.kt
sschuberth: > This change of the first commit is something additional that needs to be done for compatibility with FossID.

Maybe I'm missing or mis-remembering something, but I thought we'd treat FossID snippet matches with multiple non-consecutive ranges as multiple individual `SnippetFinding`s in ORT's data model.

After all, we should create ORT's snippet data model after what we believe makes sense and what's able to cover all use-cases, and then map FossID's model to ORT's, instead of making ORT's model match FossID's exactly in all cases.

model/src/main/kotlin/utils/SortedSetConverters.kt
```diff
@@ -91,7 +91,7 @@ class ScopeSortedSetConverter : StdConverter<Set<Scope>, SortedSet<Scope>>() {
 
 class SnippetFindingSortedSetConverter : StdConverter<Set<SnippetFinding>, SortedSet<SnippetFinding>>() {
     override fun convert(value: Set<SnippetFinding>) =
-        value.toSortedSet(compareBy<SnippetFinding> { it.sourceLocation.path }.thenByDescending { it.snippet.purl })
+        value.toSortedSet(compareBy<SnippetFinding> { it.sourceLocation })
```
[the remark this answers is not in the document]
nnobelis: Because `snippet` is now a `Set<Snippet>` and there is now many purls for a single `SnippetFinding`,

model/src/main/kotlin/SnippetFinding.kt
nnobelis: I remember discussing that, if a snippet has multiple purls, it should be treated as multiple `SnippetFinding`. This is already the case in the implementation.
For non-consecutive line ranges, this is a good point and it could be that we agreed on the same and it felt under the radar.

However I worry this will again make the ORT result grow quite a bit. And I am not sure of the snippet choice could work with such split model.

I will put this PR back in draft and ponder this change. I also reached out to the ScanOSS team to see how they deal will non-consecutive line ranges as this could be interesting in designing the ORT model.

> map FossID's model to ORT's, instead of making ORT's model match FossID's exactly in all cases.

Yes of course and this is why I am always doing the splits when developing for Bosch and integrating in ORT 😄

nnobelis: @sschuberth I received a response from the ScanOSS team.

> I'd like to clarify that the second option you mentioned is indeed the correct one. In this case, the engine will generate a unified string containing all the line ranges, each separated by a comma. Here's an example for better understanding:
> `"lines": "0-15, 20-32, 33-50, 65-83, 124-131, 161-172, 183-192, 204-221, 236-245, 247-254, 256-266"`

Therefore FossID and ScanOSS have the same kind of results with non consecutive lines. However, after experimenting, I think splitting non consecutive lines across multiple `SnippetFinding`s is ok for ORT. It is not the most compact model but I don't see any drawback in using it.

As a side note, an interesting other comment from ScanOSS:
 
> Furthermore, I wanted to bring your attention to the availability of a feature called "High Precision Snippet Matching" (HPSM). This feature can be utilized to attain even more accurate line ranges. Additionally, by using the "-F2" flag, you can incorporate snippet IDs into the results. I want to emphasize that both of these features are accessible through our clients.

So snippets in ScanOSS can have IDs as FossID does. Maybe we should, in a future PR, add the ID to the snippet model in ORT. Hence, we wouldn't have to store the ids in the snippet's `additionalData` as we currently do for FossID.
What do you think ?

fviernau: If you copy one snipped into multiple files to different line locations, do these get detected with identical snipped ID ?

nnobelis: @fviernau For ScanOSS I cannot answer as I didn't test this `-F2` parameter.
For FossID, I would say 'no': If you scan a project and rescan it, the snippets are the same but with different ids.

Therefore, if we add this `id` property to the snippet model in ORT, I would put a documentation saying that its meaning is vendor specific.

fviernau: > For FossID, I would say 'no': If you scan a project and rescan it, the snippets are the same but with different ids.

Thanks, that's a pitty.

nnobelis:
@sschuberth I updated the PR to comply with your comment https://github.com/oss-review-toolkit/ort/pull/7401#discussion_r1309938664.

I attach
[AsciiDoc_fossid_snippet.zip](https://github.com/oss-review-toolkit/ort/files/12513831/AsciiDoc_fossid_snippet.zip)
 a snippet report using the new model.The interesting part is in `Requirement.java`.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #7402, opened 2023-08-21
fix(FossId): Ignore unknown Json properties for FossID scan description

MarcelBochtler:


nnobelis (approved):

clients/fossid-webapp/src/main/kotlin/model/status/ScanDescription.kt
```diff
@@ -19,8 +19,10 @@
 
 package org.ossreviewtoolkit.clients.fossid.model.status
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
```
sschuberth, line 22: Commit message: A short explanation *why* this is necessary *now* would have been nice, like FossID version XY introduced new properties that are not needed right now, and to guard against this happening in the future.

clients/fossid-webapp/src/main/kotlin/model/status/ScanDescription2021dot2.kt
```diff
@@ -19,11 +19,13 @@
 
 package org.ossreviewtoolkit.clients.fossid.model.status
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.annotation.JsonProperty
 
 /**
  * A description of scan status. This class is for FossID version 2021.2.
  */
+@JsonIgnoreProperties(ignoreUnknown = true)
```
sschuberth, line 28: Why do properties also need to be ignored for old version 2021.2, when the commit message says the new property was introduced in version (20)23.2? Or is this class being used *starting* with version 2021.2 and onward?

Maybe this is also something that could be clarified in the class description while at it: Add a commit that documents `ScanDescription` and `ScanDescription2021dot2` more clearly with regard to which versions they're used with.

sschuberth: Ok, found it, `ScanDescription2021dot2` is really being used with version 2021.2 *and onward*:

https://github.com/oss-review-toolkit/ort/blob/3b318d687c0c1aeb7fd790a961f56aeb829ad1f5/clients/fossid-webapp/src/main/kotlin/FossIdServiceWithVersion.kt#L36

Then I'd have the opposite question: Why ignore properties also in the (legacy) `ScanDescription`?

MarcelBochtler: You're right, I removed it from `ScanDescription`.

sschuberth (dismissed):

clients/fossid-webapp/src/main/kotlin/model/status/ScanDescription2021dot2.kt
```diff
@@ -23,7 +23,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.annotation.JsonProperty
 
 /**
- * A description of scan status. This class is for FossID version 2021.2.
+ * A description of scan status. This class is for FossID version 2021.2 and later.
```
sschuberth, line 26: Nit: I guess the commit message type should be `docs`. Otherwise it's a bit unclear what aspect of the `ScanDescription2021dot2` class the "improved accuracy" refers to.

sschuberth (approved):

sschuberth:
Merging despite the unrelated `PubFunTest`  failure.

Issue #730, opened 2018-08-01
Add AWS S3 as supported remote cache


tsteenbe:
For re-using already known scan results, ORT can use a remote cache. Currently only Artifactory is supported but we would like to add AWS S3 support as it more commonly used and cheaper to set up than Artifactory instance.

See also: https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/examples-s3.html

sschuberth:
This was implemented in https://github.com/oss-review-toolkit/ort/pull/7413.

PR #7409, opened 2023-08-22
Add support for using AWS S3 as online cache for scan results

carlos-ardila2:
Provides a S3FileStorage class which extends [FileStorage](https://github.com/oss-review-toolkit/ort/blob/main/utils/ort/src/main/kotlin/storage/FileStorage.kt) and uses official Amazon's  [AWS Kotlin SDK](https://aws.amazon.com/sdk-for-kotlin/) to store scan results.

Issue oss-review-toolkit/ort#730


sschuberth:
Your PR seem to contain ~13 commits that are not your own. Please rebase onto latest `main` to fix that and only keep your own commits in here.

PR #7416, opened 2023-08-25
fix(scanner): Merge `AdditionalData`s correctly

nnobelis:
Currently, when two scan results are merged, their `AdditionalData`s are merged into one and data is lost because key pairs are overridden. Such case can happen when a project with a Git submodule is scanned in FossID and leads to the creation of two scans in the latter. Two prevent this, this commit concatenates multiple additional data values as CSV. It also adapts the FossID reporter to accept such values.


model/src/main/kotlin/utils/ScanResultUtils.kt
```diff
@@ -69,11 +69,30 @@ fun mergeScanResultsByScanner(
...
+/**
+ * Merge together a list of [AdditionalData]. If there is only one of them, it is left untouched. If there are more than
+ * one, the collection of key pairs are each replaced by a single key pair: the key has the same name. The values are
+ * concatenated as CSV.
+ */
+private fun List<Map<String, String>>.mergeAdditionalData(): Map<String, String> {
+    return when (size) {
+        0 -> emptyMap()
+
+        1 -> first()
```
sschuberth, line 86: For simplicity, consider dropping these shortcut cases.

plugins/reporters/fossid/src/main/kotlin/FossIdReporter.kt
```diff
@@ -93,18 +93,23 @@ class FossIdReporter : Reporter {
...
-                                it.showStackTrace()
-                                logger.info {
-                                    "Error during report generation: ${it.collectMessages()}."
-                                }
-                            }.getOrNull()
-                    }
+            val scanResults = input.ortResult.getScanResults().values.asSequence().flatten()
+            val scanCodes = scanResults.mapNotNull { it.additionalData[SCAN_CODE_KEY] }
+                .distinct()
+                .map { it.split(",") }
```
sschuberth, line 99: Shouldn't you *frist* (flat) map to the split codes, and *then* deduplicate?

sschuberth (commented):
See above.

nnobelis:
> See above.

Let's see with the customer data I am working with:
```
input.ortResult.getScanResults().values.flatten() -> 370 elements
input.ortResult.getScanResults().values.flatten().mapNotNull { it.additionalData[SCAN_CODE_KEY] } -> 4 elements
input.ortResult.getScanResults().values.flatten().mapNotNull { it.additionalData[SCAN_CODE_KEY] }.distinct() -> 2 elements
input.ortResult.getScanResults().values.flatten().mapNotNull { it.additionalData[SCAN_CODE_KEY] }.distinct().map { it.split(",") }.flatten() -> 5 elements
input.ortResult.getScanResults().values.flatten().mapNotNull { it.additionalData[SCAN_CODE_KEY] }.distinct().map { it.split(",") .flatten().toSet() -> 2 elements
```
`split` is called twice.

Whereas in your solution, `split` is called 370 times. Yes, 366 times on the empty string, but it's still a function call with the stack and the clearing of registries. No ?

sschuberth:
> Yes, 366 times on the empty string, but it's still a function call with the stack and the clearing of registries.

TBH, I don't really care about the call stack, and I don't understand what "clearing of registries" refers to, but what I *would* be interested in is performance. As you seem to be willing to dig deeper, mind measuring some timings?

nnobelis:
> > Yes, 366 times on the empty string, but it's still a function call with the stack and the clearing of registries.
> 
> TBH, I don't really care about the call stack, and I don't understand what "clearing of registries" refers to, but what I _would_ be interested in is performance. As you seem to be willing to dig deeper, mind measuring some timings?

You know like me that the performance difference on such a small set are in the order of nanoseconds, and you would not be able to mesure it.
However, I always thought ORT development always aimed at writing the most performant code, ever, without wasting CPU cycles. Am I mistaken ?

sschuberth:
> However, I always thought ORT development always aimed at writing the most performant code, ever, without wasting CPU cycles. Am I mistaken ?

Partly. As always it's a balance between the elegance / simplicity of the code vs. the performance gains. And in areas where it does not matter, I prefer more readable code that's easier to understand and maintain over premature optimizations that, while algorithmically sensible, do not really have an impact.

plugins/reporters/fossid/src/main/kotlin/FossIdReporter.kt
```diff
@@ -93,18 +93,23 @@ class FossIdReporter : Reporter {
...
-                        logger.info { "Generating report for scan $scanCode." }
-                        service.generateReport(user, apiKey, scanCode, reportType, selectionType, outputDir)
-                            .onFailure {
-                                it.showStackTrace()
-                                logger.info {
-                                    "Error during report generation: ${it.collectMessages()}."
-                                }
-                            }.getOrNull()
-                    }
+            val scanResults = input.ortResult.getScanResults().values.asSequence().flatten()
```
sschuberth, line 96: BTW, what's the benefit of the `asSequence()` call? As always all values are processed, there's no savings in computations, or?

fviernau:
> > 
> 
> Partly. As always it's a balance between the elegance / simplicity of the code vs. the performance gains. And in areas where it does not matter, I prefer more readable code that's easier to understand and maintain over premature optimizations that, while algorithmically sensible, do not really have an impact. On the other hand, if readability is the same, I even aim for micro-optimizations like preferring `split(',')` over `split(",")`.

I believe it's time to put this into contributing guidelines, so that we have a clear agreement everybody can refer to.

nnobelis:
> > However, I always thought ORT development always aimed at writing the most performant code, ever, without wasting CPU cycles. Am I mistaken ?
> 
> Partly. As always it's a balance between the elegance / simplicity of the code vs. the performance gains. And in areas where it does not matter, I prefer more readable code that's easier to understand and maintain over premature optimizations that, while algorithmically sensible, do not really have an impact. On the other hand, if readability is the same, I even aim for micro-optimizations like preferring `split(',')` over `split(",")`.

Thanks for the explanation. I put your version in.

plugins/reporters/fossid/src/main/kotlin/FossIdReporter.kt
nnobelis: In my version with several calls in the chain, it was a recommendation from Intellij.
I removed in your version because as you said it is not needed.

sschuberth (approved):

sschuberth:
> Test failure is unrelated to the changes (`PubFunTest`).

Confirmed, see https://github.com/oss-review-toolkit/ort/pull/7417.

PR #7421, opened 2023-08-26
docs(ScanResultUtils): Improve docs for `mergeAdditionalData()`

sschuberth:
In particular, fix the reference and do not document the special case code that was removed during PR review.

### DIFF
--- a/model/src/main/kotlin/utils/ScanResultUtils.kt
+++ b/model/src/main/kotlin/utils/ScanResultUtils.kt
@@ -75,9 +75,8 @@ fun mergeScanResultsByScanner(
 }
 
 /**
- * Merge together a list of [AdditionalData]. If there is only one of them, it is left untouched. If there are more than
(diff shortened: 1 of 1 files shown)


nnobelis (approved):

model/src/main/kotlin/utils/ScanResultUtils.kt
```diff
@@ -75,9 +75,8 @@ fun mergeScanResultsByScanner(
 }
 
 /**
- * Merge together a list of [AdditionalData]. If there is only one of them, it is left untouched. If there are more than
- * one, the collection of key pairs are each replaced by a single key pair: the key has the same name. The values are
- * concatenated as CSV.
+ * Merge together a list of [ScanResult.additionalData] by keeping all keys and concatenating values to a string
```
fviernau, line 78: Maybe, while at it, "together" could be dropped?

fviernau (dismissed):

PR #7427, opened 2023-08-29
Improve resolution tests

mnonnenmacher:

### DIFF
--- a/model/src/test/kotlin/config/IssueResolutionTest.kt
+++ b/model/src/test/kotlin/config/IssueResolutionTest.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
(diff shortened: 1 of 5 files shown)


model/src/test/kotlin/config/IssueResolutionTest.kt
```diff
@@ -0,0 +1,68 @@
...
+})
+
+private fun resolution(message: String) = IssueResolution(
+    message = message,
+    reason = IssueResolutionReason.CANT_FIX_ISSUE,
+    comment = ""
+)
+
+private fun issue(message: String) =
+    Issue(
```
sschuberth, line 64: Nit: Formatting is inconsistent compared to line 57 above.

sschuberth (approved):

model/src/test/kotlin/config/IssueResolutionTest.kt
mnonnenmacher: As discussed I will merge this PR as is and make a separate PR which will enable a detekt rule to ensure consistent function signature formatting.

PR #7429, opened 2023-08-30
style(detekt): Enable the "FunctionSignature" rule

mnonnenmacher:
Enable the "FunctionSignature" rule [1] [2] to ensure consistent formatting of function signatures. Set the
`functionBodyExpressionWrapping` property to `multiline` to align with the Kotlin style guide which recommends to put a linebreak after the `=` sign if an expression body does not fit in a single line [3].

[1]: https://detekt.dev/docs/rules/formatting#functionsignature
[2]: https://pinterest.github.io/ktlint/0.50.0/rules/experimental/#function-signature
[3]: https://kotlinlang.org/docs/coding-conventions.html#expression-bodies

### DIFF
--- a/.detekt.yml
+++ b/.detekt.yml
@@ -38,8 +38,6 @@ formatting:
     functionBodyExpressionWrapping: multiline
   ImportOrdering:
     active: false
-  Indentation:
-    active: false
   MaximumLineLength:
     active: false
   NoUnusedImports:
--- a/advisor/src/main/kotlin/advisors/GitHubDefects.kt
+++ b/advisor/src/main/kotlin/advisors/GitHubDefects.kt
@@ -167,7 +167,7 @@ class GitHubDefects(name: String, config: GitHubDefectsConfiguration) : AdvicePr
                 ortIssues += createAndLogIssue(
                     providerName,
                     "Failed to load information about $itemType for package '${pkg.original.id.toCoordinates()}': " +
-                            exception.collectMessages(),
+                        exception.collectMessages(),
                     Severity.ERROR
                 )
             }.getOrNull().orEmpty()
@@ -231,9 +231,9 @@ class GitHubDefects(name: String, config: GitHubDefectsConfiguration) : AdvicePr
      */
     private fun List<GitHubIssue>.applyLabelFilters(): List<GitHubIssue> =
         filter { issue ->
-        val labels = issue.labels()
-        labelFilters.find { it.matches(labels) }?.including ?: false
-    }
+            val labels = issue.labels()
+            labelFilters.find { it.matches(labels) }?.including ?: false
+        }
 }
 
 /**
--- a/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
+++ b/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
@@ -48,7 +48,7 @@ class AnalyzerResultBuilder {
         val duplicates = (projects.map { it.toPackage() } + packages).getDuplicates { it.id }
         require(duplicates.isEmpty()) {
             "Unable to create the AnalyzerResult as it contains packages and projects with the same ids: " +
-                    duplicates.values
+                duplicates.values
         }
 
         return AnalyzerResult(projects, packages, issues, dependencyGraphs)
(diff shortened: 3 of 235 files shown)


analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
```diff
@@ -93,15 +94,17 @@ class AnalyzerResultBuilder {
      * Add the given [packageSet] to this builder. This function can be used for packages that have been obtained
      * independently of a [ProjectAnalyzerResult].
      */
-    fun addPackages(packageSet: Set<Package>) = apply {
+    fun addPackages(packageSet: Set<Package>) =
+        apply {
```
sschuberth, line 98: I'm also wondering why this did not become a one-liner. Looks like the auto-formatting is pretty much broken...

advisor/src/main/kotlin/advisors/GitHubDefects.kt
```diff
@@ -229,7 +229,8 @@ class GitHubDefects(name: String, config: GitHubDefectsConfiguration) : AdvicePr
     /**
      * Return a filtered list of [Issue]s according to the label filters defined in the configuration.
      */
-    private fun List<GitHubIssue>.applyLabelFilters(): List<GitHubIssue> = filter { issue ->
+    private fun List<GitHubIssue>.applyLabelFilters(): List<GitHubIssue> =
+        filter { issue ->
```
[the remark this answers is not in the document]
mnonnenmacher: I have added another commit that enables the "Indentation" rule because I found it annoying that we did not have a proper check for that.

analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
mnonnenmacher: It seems that code blocks are not collapsed automatically (I also could not find a rule for that), but apart from that the other auto-formatting looks good to me.

.detekt.yml
```diff
@@ -33,6 +33,9 @@ formatting:
     active: false
   FinalNewline:
     active: false
+  FunctionSignature:
```
sschuberth, line 36: > Note that this commit introduces some indentation issues which will be fixed in the following commit.

To a reviewer it might be unclear why theses issues were introduced in the first place. The missing link is that the reformatting was automatically done by `ktlint`, which introduced some temporary side effects. So, maybe mention the command that was run to apply the reformatting as part of the commit message?

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -338,7 +338,8 @@ enum class VcsHost(
             val unknownVcsInfo = VcsInfo.EMPTY.copy(url = projectUrl)
             val projectUri = projectUrl.toUri().getOrElse { return unknownVcsInfo }
 
-            fun URI.isTfsGitUrl() = path != null && host != null &&
+            fun URI.isTfsGitUrl() =
+                path != null && host != null &&
```
sschuberth, line 342: This line and the line below would now actually fit in one line. And shouldn't they according to the rules?

scanner/src/funTest/kotlin/scanners/ScannerIntegrationFunTest.kt
```diff
@@ -147,8 +147,7 @@ private fun createAnalyzerResult(vararg packages: Package): OrtResult {
     return OrtResult.EMPTY.copy(analyzer = analyzerRun)
 }
 
-private fun createId(name: String): Identifier =
-    Identifier("Dummy::$name:1.0.0")
+private fun createId(name: String): Identifier = Identifier("Dummy::$name:1.0.0")
```
sschuberth, line 150: Nit: While at it, should we introduce an empty line after this function?

downloader/src/main/kotlin/VcsHost.kt
mnonnenmacher: The rule does not touch existing line breaks within the function body, it only checks if the first line of the function body would fit on the same line as the signature. I have manually fixed it here.

mnonnenmacher:
Merged despite the failing `funTest-docker` which currently fails with a Node.js installation issue. Functional tests for previous iterations of the PR were successful.

PR #7435, opened 2023-08-31
fix(osv): Account for typed events being introduced in the model

sschuberth:
Some, but not all, OSV requests now return "typed" events where the
event type and value are serialized to separate fields. This is the
natural serialization format that kotlinx-serialization would use for
the `Event` class via the auto-generated deserializer.

However, some queries still return the single-line format for events.
This means the auto-generated deserializer should be used by default, and
a custom deserializer that understands the single-line format should be
used as a fallback. The current approach of a low-level `KSerializer`
is not very suitable to handle this case, as trying to decode one format
would advance in the input, leaving incomplete input to the fallback
deserializer if the first deserializer fails.

To solve this, a `JsonTransformingSerializer` is used instead that
transforms the single-line format to the typed format on a
`JsonElement` level before the auto-generated deserializer does its work.
As this serializer needs to be based on the auto-generated serializer,
another trick needs to be applied in order to not run into a recursion
problem: The `JsonTransformingSerializer` works on a list of events
instead of a single event. This way, it can refer to the auto-generated
serializer for a single event (instead of referring to itself) and
transform each events of a list instead of a single even.

### DIFF
--- a/clients/osv/src/main/kotlin/Model.kt
+++ b/clients/osv/src/main/kotlin/Model.kt
@@ -111,7 +111,7 @@ object Ecosystem {
     const val RUBY_GEMS = "RubyGems"
 }
 
-@Serializable(EventSerializer::class)
+@Serializable
 data class Event(
     val type: Type,
     val value: String
@@ -136,6 +136,7 @@ data class Package(
 data class Range(
     val type: Type,
     val repo: String? = null,
+    @Serializable(EventListSerializer::class)
     val events: List<Event>,
     val databaseSpecific: JsonObject? = null
 ) {
--- a/clients/osv/src/main/kotlin/Serializers.kt
+++ b/clients/osv/src/main/kotlin/Serializers.kt
@@ -23,46 +23,35 @@ import java.time.Instant
 import java.time.format.DateTimeFormatter
 
 import kotlinx.serialization.KSerializer
-import kotlinx.serialization.SerializationException
-import kotlinx.serialization.Serializer
+import kotlinx.serialization.builtins.ListSerializer
 import kotlinx.serialization.descriptors.PrimitiveKind
 import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
 import kotlinx.serialization.descriptors.SerialDescriptor
 import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
-import kotlinx.serialization.json.JsonDecoder
-import kotlinx.serialization.json.JsonEncoder
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
-import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.json.JsonTransformingSerializer
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonObject
 
 /**
- * Use a custom serializer in order to map the original data structure to a more strict and simple, enum based, model.
+ * Use a custom serializer to transform the legacy single-field events into newer typed events.
  */
-@Serializer(Event::class)
-internal object EventSerializer : KSerializer<Event> {
-    override fun deserialize(decoder: Decoder): Event {
-        val input = decoder as? JsonDecoder
-            ?: throw SerializationException("This serializer only works with the JSON format.")
-
-        val element = input.decodeJsonElement()
-        require(element is JsonObject)
-
-        require(element.entries.size == 1)
-        val (key, value) = element.entries.first()
-        val type = enumValueOf<Event.Type>(key.uppercase())
-
-        return Event(type, value.jsonPrimitive.content)
-    }
-
-    override fun serialize(encoder: Encoder, value: Event) {
-        val output = encoder as? JsonEncoder
-            ?: throw SerializationException("This serializer only works with the JSON format.")
-
-        val tree = JsonObject(mapOf(value.type.name.lowercase() to JsonPrimitive(value.value)))
-
-        output.encodeJsonElement(tree)
-    }
+internal object EventListSerializer : JsonTransformingSerializer<List<Event>>(ListSerializer(Event.serializer())) {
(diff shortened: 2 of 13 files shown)


clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
```diff
@@ -50,14 +48,12 @@ private val VULNERABILITY_FOR_PACKAGE_BY_INVALID_COMMIT_REQUEST = Vulnerabilitie
 private fun Vulnerability.patchIgnorableFields() =
     copy(
         modified = Instant.EPOCH,
-        databaseSpecific = emptyJsonObject.takeIf { databaseSpecific != null },
+        databaseSpecific = null,
```
fviernau, line 51: Why has line 51 been changed?

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
```diff
@@ -50,14 +48,12 @@ private val VULNERABILITY_FOR_PACKAGE_BY_INVALID_COMMIT_REQUEST = Vulnerabilitie
...
         affected = affected.mapTo(mutableSetOf()) { affected ->
-            affected.copy(ecosystemSpecific = emptyJsonObject.takeIf { affected.ecosystemSpecific != null })
+            affected.copy(ecosystemSpecific = null)
```
fviernau, line 53: Why has line 53 been changed?

clients/osv/src/main/kotlin/Serializers.kt
```diff
@@ -23,46 +23,35 @@ import java.time.Instant
...
-
-    override fun serialize(encoder: Encoder, value: Event) {
-        val output = encoder as? JsonEncoder
-            ?: throw SerializationException("This serializer only works with the JSON format.")
-
-        val tree = JsonObject(mapOf(value.type.name.lowercase() to JsonPrimitive(value.value)))
-
-        output.encodeJsonElement(tree)
-    }
+internal object EventListSerializer : JsonTransformingSerializer<List<Event>>(ListSerializer(Event.serializer())) {
```
fviernau, line 43: commit-msg: Please provide a rationale, why this refactor from `EventSerializer` to `EventListSerializer`. E.g.:
- Why is it better to do that on lists?
- Why is it ok to drop logic for serializing?

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
fviernau, line 51: commit: Why is the old format continued to be supported? Does OSV now return a mixture of both formats?

fviernau (changes requested):

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
sschuberth: See below.

clients/osv/src/main/kotlin/Serializers.kt
sschuberth: > Why is it ok to drop logic for serializing?

Because we never serialize the OSV model AFAIK.

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
sschuberth: Nevermind, I misunderstood the intention of the logic I believe, and have restored this in favor of updating the expected results.

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
fviernau: Looking at the expected result, it seems that only the new format is returned. So, can we adjust this PR to handle the new format only, thereby eliminating the unneeded JsonNames annotations again?

fviernau (dismissed):

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
sschuberth: > So, can we adjust this PR to handle the new format only, thereby eliminating the unneeded JsonNames annotations again?

I'm currently testing to do this. Initially, it seemed to me as if the old format is still returned under some circumstances, and also @mnonnenmacher reported something in that direction to me, but at least in my case the cause was likely outdated test data.

clients/osv/src/main/kotlin/Serializers.kt
sschuberth: > Why is it better to do that on lists?

This avoids running into a recursion problem when referring to the auto-generated serializer. I've explained this in the new commit message.

clients/osv/src/main/kotlin/Model.kt
```diff
@@ -136,6 +136,7 @@ data class Package(
 data class Range(
     val type: Type,
     val repo: String? = null,
+    @Serializable(EventListSerializer::class)
```
sschuberth, line 139: @fviernau, FYI, you can verify that deserialization of the old format is still required by removing this line and running `OsvServiceFunTest`. You'll see then that tests already fail deserializing the *actual* results (not the expected results).

mnonnenmacher (commented):
Is there a test case left which verifies that deserialization of the old format still works?

mnonnenmacher:
> > Is there a test case left which verifies that deserialization of the old format still works?
> 
> That's implicitly being tested when deserializing the _actual_ (not expected) data, as it partly contains the old format, see [this comment](https://github.com/oss-review-toolkit/ort/pull/7435#discussion_r1312789785) of mine.

Ok, I wasn't sure what actual data it is referring to. I think a small unit test for just the deserialization would be nice to have.

> _Edit:_ If we added the old format explicitly to the _expected_ data, we'd probably not notice if this is not required anymore when the actual data stops emitting the old format.

We won't notice either way :)

mnonnenmacher (dismissed):

mnonnenmacher (approved):

clients/osv/src/main/kotlin/Model.kt
sschuberth: This was a red herring. I was indeed able to revert most of the changes, see https://github.com/oss-review-toolkit/ort/pull/7444. Sorry @fviernau!

PR #7447, opened 2023-09-03
Use per language container strategy

heliocastro:
Ort language components are built now on independent containers. This enables the possibility to create custom tailored containers with only the required language components, reducing final size and improving build process.

Default runtime now contains Python and NodeJs components. Other language components have their own containers built, and can be included on custom images.


heliocastro:
Close due https://github.com/oss-review-toolkit/ort/pull/7112

sschuberth:
> Close due #7112

Now both PRs are closed!?

PR #7466, opened 2023-09-06
fix(fossid-webapp)!: Simplify the pattern for directories in ignore rules

nnobelis:
The commit b194cf37 added a change for supporting '.' in directory names. Currently, the dash symbol is also no supported. Therefore, the current commit simplifies the logic to allow all special characters in the directory name, by simplifying the regular expression.

This is a breaking change as the directory name "directory/sub1/sub2" is not accepted anymore and won't be mapped to a directory ignore rule. The directory name must now always end with '/', i.e. "directory/sub1/sub2/".


sschuberth:
Typo in "no*t* supported".

mnonnenmacher (approved):

PR #7470, opened 2023-09-07
 Change to grep/sed solution due Mac grep difference

heliocastro:
Mac BSD based grep tool has different options and behaviour than Linux GNU based, so a solution with basic sed and grep was needed.

### DIFF
--- a/scripts/docker_build.sh
+++ b/scripts/docker_build.sh
@@ -20,7 +20,7 @@
 DOCKER_ARGS=$@
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
-GIT_REVISION=$($GIT_ROOT/gradlew -q properties --property version | grep -oP "version: \K.+")
+GIT_REVISION=$($GIT_ROOT/gradlew -q properties --property version | sed -nr "s/version: (.+)/\1/p")
 
 echo "Setting ORT_VERSION to $GIT_REVISION."
 docker buildx build \


scripts/docker_build.sh
```diff
@@ -20,7 +20,7 @@
 DOCKER_ARGS=$@
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
-GIT_REVISION=$($GIT_ROOT/gradlew -q properties --property version | grep -oP "version: \K.+")
+GIT_REVISION=$("$GIT_ROOT/gradlew" -q properties --property version | grep "version" | sed -e "s/version: //")
```
sschuberth, line 23: Some things, please
- move the addition of quotes to a separate commit, which a separate rationale,
- explain what part of the `grep`command wasn't working on macOS,
- use either `grep` or `sed`, but not both.

sschuberth (changes requested):

scripts/docker_build.sh
heliocastro: Fixed

scripts/docker_build.sh
```diff
@@ -20,7 +20,7 @@
 DOCKER_ARGS=$@
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
-GIT_REVISION=$($GIT_ROOT/gradlew -q properties --property version | grep -oP "version: \K.+")
+GIT_REVISION=$($GIT_ROOT/gradlew -q properties --property version |  awk '/version: / {print $2}')
```
sschuberth, line 23: Please no heavy-weight `awk` either. How about `sed -nr "s/version: (.+)/\1/p"`, does that work on macOS?

sschuberth (changes requested):

scripts/docker_build.sh
sschuberth: There's now a merge commit in the PR.

scripts/docker_build.sh
heliocastro: Works, but i need just disagree about "heavy-weight" awk, when you need entire gradlew and parsing the entire project in the same line before :-)

sschuberth: Awk is a new dependency, Gradle is not. If it's in use already anyway, it doesn't matter that much how heavy-weight the tool is. And Gradle we bootstrap, whereas Awk might not be available on all platforms, like Windows (via WSL or Git Bash).

scripts/docker_build.sh
```diff
@@ -20,7 +20,7 @@
 DOCKER_ARGS=$@
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
-GIT_REVISION=$($GIT_ROOT/gradlew -q properties --property version | grep -oP "version: \K.+")
+GIT_REVISION=$($GIT_ROOT/gradlew -q properties --property version |  sed -nr "s/version: (.+)/\1/p")
```
sschuberth, line 23: Double-space before `sed`.

sschuberth (changes requested):

sschuberth (approved):

scripts/docker_build.sh
heliocastro: Done

PR #7474, opened 2023-09-07
Docker Next Generation - Per language container strategy

heliocastro:
Docker now builds independent language images, and the final image can be assembled with ORT main binary + languages.
All images are deployed under ghcr.io registry by default.

Default runtime image contains Java, Python, Rust, Golang, NodeJS and Ruby components, and other components can be be added using the new **custom-docker.sh** script.

Extra components available are:
Android, Swift, Dotnet, PHP, Haskell

This closes original external PR:
Closes #7447


fviernau (approved):

tsteenbe (approved):

mnonnenmacher:
Too late for a review, so just some general remarks from me:
* Indentation in workflow files is wrong, we always indent with two spaces.
* Some commits probably should have been squashed, why introduce so many new workflow files only to remove them again within the same PR?
* Last commit should (at least mostly) have been integrated into the previous commits.
* Some commit messages are fine, but for example the 2nd and 3rd commit would have needed an explanation to understand the motivation.
* I don't like the name "Mega image", it's not very telling.

sschuberth:
> Too late for a review, so just some general remarks from me:

Thanks for these remarks. In sum, very surprising that this was approved as-is by two (!) reviewers then.

PR #7477, opened 2023-09-07
refactor(PythonInspector): Remove default arguments

sschuberth:
Remove the default function arguments for the Python version and OS from `inspect()` as they are too easily being confused with the configured package manager options and their default values.


plugins/package-managers/python/src/funTest/kotlin/utils/PythonInspectorFunTest.kt
```diff
@@ -37,7 +37,8 @@ class PythonInspectorFunTest : StringSpec({
             PythonInspector.inspect(
                 workingDir = workingDir,
                 definitionFile = definitionFile,
-                pythonVersion = "27"
+                pythonVersion = "27",
```
mnonnenmacher, line 40: Commit message nit: "an**d** OS"

mnonnenmacher (changes requested):

mnonnenmacher (approved):

Issue #4099, opened 2021-06-01
Store dependencies as a tree of SPDX files (instead of a flat list)


silverhook:
Currently ORT’s SPDX output produces a flat list of all the dependencies and their (package-level) licenses.

This is a huge missed opportunity, as SPDX spec does actually include fields such as [`Relationship`](https://spdx.github.io/spdx-spec/7-relationships-between-SPDX-elements/#71-relationship) and [`ExternalDocumentRef`](https://spdx.github.io/spdx-spec/2-document-creation-information/#26-external-document-references), which could be used to:
- create a separate SPDX file for each package, which includes file-level info (ScanCode already does this, when used by ORT to scan the deps)
- create a SPDX file for the main package being analyzed, and in it
- link to SPDX files of other relevant packages using the appropriate relationships – e.g. for a dependency `DEPENDS_ON`

This would make the output not only easier to parse, since the hierarchy would not be lost; but would also make the results re-usable, as the SPDX file for a certain package could simply be reused if that package is a dependency in another scan later on.

See also https://github.com/oss-review-toolkit/ort/issues/1913#issuecomment-638809109 where relationships were mentioned as a todo.

cc: @maxhbr

sschuberth:
To clarify: IIUC the SPDX reporter already *does* maintain package hierarchy, see e.g. this expected test result file:

https://github.com/oss-review-toolkit/ort/blob/d9e78272c85273b8508b67f652bb511994ff5cba/reporter/src/funTest/assets/spdx-document-reporter-expected-output.spdx.yml#L155-L158

It's just that all packages are captured as part of a single SPDX file. So this request is not so much about "not loosing hierarchy", but simply to "externalize" the related packages to separate files.

maxhbr:
I generated the `document.spdx.yml` yesterday for `mime-types` and in the result each packages is always a dependeny_of the root package (and not of another package):

```
$  cat document.spdx.yml | grep DEPENDENCY_OF -B2
- spdxElementId: "SPDXRef-Package-1-acorn"
  relatedSpdxElement: "SPDXRef-Package-0-root-package"
  relationshipType: "DEPENDENCY_OF"
--
- spdxElementId: "SPDXRef-Package-10-ajv"
  relatedSpdxElement: "SPDXRef-Package-0-root-package"
  relationshipType: "DEPENDENCY_OF"
--
- spdxElementId: "SPDXRef-Package-100-character-entities-legacy"
[...]

Issue #3906, opened 2021-04-19
Capture the dependency graph in CycloneDX BOMs


sschuberth:
Since the CycloneDX 1.2 spec, [the dependency graph moved from an extension to a core feature](https://github.com/CycloneDX/cyclonedx-core-java/issues/99#issuecomment-820619180). So after https://github.com/oss-review-toolkit/ort/pull/3897 is merged, we could make use of it as the hierarchy of dependencies is valuable additional information.

mawl:
Hey,

we really would like to see see that this feature gets implemented. 
* Trivy can already generate CycloneDX SBOMs with a dependency graph included: https://trivy.dev/latest/docs/supply-chain/sbom/#generating * Since 2021 Dependency-Track can visualise dependency graphs: https://docs.dependencytrack.org/2021/03/17/v4.2.0/

mawl:
@sschuberth: Thanks for your work.

Unfortunately I get a stackoverflow error with ort v43.0.0:

```
 ______________________________                                                
/        \_______   \__    ___/ The OSS Review Toolkit, version 43.0.0,        
|    |   | |       _/ |    |    built with JDK 21.0.5+11-LTS, running under Jav
|    |   | |    |   \ |    |    Executing 'report' as 'ort' on Linux           
\________/ |____|___/ |____|    with 8 CPUs and a maximum of 8192 MiB of memory
                                                                               
Environment variables:                                                        
ORT_CONFIG_DIR = /home/ort/.ort/config                                        
ORT_DATA_DIR = /builds/compliance/license-scanning/.ort                
HOME = /home/ort                                                              
JAVA_HOME = /opt/java/openjdk                                                 
ANDROID_HOME = /opt/android-sdk                                               
                                                                              
Looking for ORT configuration in the following file:
        /home/ort/.ort/config/config.yml
Generating the 'WebApp' report in thread 'DefaultDispatcher-worker-1'...
Generating the 'CycloneDX' report in thread 'DefaultDispatcher-worker-3'...
Generating the 'PlainTextTemplate' report in thread 'DefaultDispatcher-worker-4'...
Exception in thread "main" java.lang.StackOverflowError
	at org.ossreviewtoolkit.model.DependencyRefCursor.visitDependencies(DependencyGraphNavigator.kt:165)
	at org.ossreviewtoolkit.model.DependencyGraphNavigator.packageDependencies$traverse(DependencyGraphNavigator.kt:104)
[...]

PR #7488, opened 2023-09-12
SPDX reporter: Stop creating dangling relationships to excluded packages

### LINKED ISSUE #7487: SpdxDocumentReporter accidentally includes broken references to excluded first level dependencies
The bug is illustrated in the functional tests and got introduced (IIUC) by [1].
In particular note that here [2], the relationship to the excluded package has been added.

[1] https://github.com/oss-review-toolkit/ort/pull/5126/commits/b47154482299b9b49c0e0c692e3f2dacdd53c395
[2] https://github.com/oss-review-toolkit/ort/pull/5126/commits/b47154482299b9b49c0e0c692e3f2dacdd53c395#diff-6de35dd2aff1f92b7f5ea558d3f77e02d0d596dd4ce2a8199056cfb31b47fcabR181-R184

fviernau:
Please see individual commits.

Fixes #7487.


plugins/reporters/spdx/src/main/kotlin/SpdxDocumentModelMapper.kt
```diff
@@ -64,21 +65,21 @@ internal object SpdxDocumentModelMapper : Logging {
...
         val projectPackages = projects.map { project ->
-            val spdxProjectPackage = project.toPackage().toSpdxPackage(
+            project.toPackage().toSpdxPackage(
                 SpdxPackageType.PROJECT,
                 licenseInfoResolver,
                 ortResult
             )
+        }
 
-            ortResult.getDependencies(project.id, 1).mapTo(relationships) { dependency ->
```
sschuberth, line 73: Just wondering, can't the same bug also appear around line 94, if a non-excluded package depends on an excluded package?

In general, I wonder whether a more general solution to the problem would be to also teach `getDependencies()` an `omitExcluded` parameter. If we had that, the fix would be a matter of just setting `omitExcluded = true` in current line 73 and 94, right?

fviernau:
```
root-project-1
|
+- sub-project-1
   |
   +- package (dependency of "sub-project" only)
root-project-2
|
+- sub-project-2
   |
   +- package (dependency of "sub-project" only)
```

If we have above, and consider the original (current) algorithm implemented by you, I guess the intention was to not mention subproject-1 and subproject-2 in the SPDX report, but link the dependencies directly to root-project-1 and root-project-2.

To my understanding the implementation did not match this mentioned intention, which I wanted to fix with this new implementation.

> are you saying there should be no relationship between "sub-project" and "package" being recorded in the SPDX document?

Given the above, I'd say no, because "package" should be a direct child of the "root" project, and nested submodules should not be modeled in the output.

sschuberth: > I guess the intention was to not mention subproject-1 and subproject-2 in the SPDX report, but link the dependencies directly to root-project-1 and root-project-2.

Actually no, that was no my intention. The intention was to create all of the following relationships in this case:

- `root-project-1` depends on `sub-project-1`
- `sub-project-1` depends on `package 1`

(And the same for the number 2 projects / packages.)

fviernau: This surprises me, because:

1. https://github.com/oss-review-toolkit/ort/commit/b47154482299b9b49c0e0c692e3f2dacdd53c395
    The commit message clearly says "...but to the real root projects"
2. That is reflected by the limitation of the depth to 1 here: https://github.com/oss-review-toolkit/ort/commit/b47154482299b9b49c0e0c692e3f2dacdd53c395#diff-a7d33dd6b251d36d6f79411935a6f2e35586240fa8e3337cb29b1c7a9b89403dR76

fviernau: But no matter what the original intention of that commit was, I believe we should do now what we think is the best.
In my opinion, only listing the top-level projects is better than exposing the whole sub-module structure. Currently, only top-level projects are exposed, so I would prefer to not start exposing sub-module structure of all nesting depths, but to stick with top level project and fix the underlying tree accordingly. 

Would that be ok with you?

sschuberth: > The commit message clearly says "...but to the real root projects"

The "but to the real root projects" part referred to the goal to link direct dependencies to the project they belong to, and not "to an arbitrary root project" like it was before.

> That is reflected by the limitation of the depth to 1 here

The idea behind that was the following: It is enough for each project to only link to its direct dependencies (with depth == 1), as in the other loop below, each package in turn gets linked to *its* direct dependencies. So even if each entity (project or package) only relates to its direct dependencies, in the end we get a full dependency tree in total.

fviernau: > Well, that differs from what @tsteenbe and others have expressed [here](https://github.com/oss-review-toolkit/ort/issues/4099#issuecomment-866253887) then. The intent of my [b471544](https://github.com/oss-review-toolkit/ort/commit/b47154482299b9b49c0e0c692e3f2dacdd53c395) was to partly address that linked issue.

I understand Thomas' comment differently, however. It's not mentioning sub-modules explicitly, so this is IMO speculation. I believe we should discuss sub-module case explicitly / separately.

> And we have the https://github.com/oss-review-toolkit/ort/issues/3906, BTW.

Ok.

> Not really, to be honest, as that takes us a step back again

Why is it a step back. My PR does not reduce information reported at all. But fixes this bug:

> However, I completely agree that I've introduced a bug in https://github.com/oss-review-toolkit/ort/commit/b47154482299b9b49c0e0c692e3f2dacdd53c395 by disregarding excludes.

fviernau: > So even if each entity (project or package) only relates to its direct dependencies, in the end we get a full dependency tree in total.

I'm not sure if this bug was clear to you already, but this sentence illustrates it pretty well. The idea does not work in case there are sub-projects, because then one does not iterate over the sub-projects dependencies at all in some cases.

mnonnenmacher: > ```
> root-project-1
> |
> +- sub-project-1
>    |
>    +- package 1
> root-project-2
> |
> +- sub-project-2
>    |
>    +- package 2  
> ```
> 
> [...]
>
> Given the above, I'd say no, because "package" should be a direct child of the "root" project, and nested submodules should not be modeled in the output.

To me it would be misleading to report "package-1" as a direct dependency of "root-project-1" here, especially if both "root-project-1" and "sub-project-1" are published. I do not yet understand the issue with the current approach to build the tree, especially:

> The idea does not work in case there are sub-projects, because then one does not iterate over the sub-projects dependencies at all in some cases.

Under which circumstances would that be the case?

fviernau:
I've split out the first commit as it will become rather unrelated to the bug fix, see #7567.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentModelMapper.kt
fviernau: @sschuberth  @mnonnenmacher  - 

I've adjusted the implementation to what I understood We agreed upon in todays call.
@sschuberth  I've incorporated your proposal from the PR comments to teach `getDependencies()` to omit excluded IDs.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentModelMapper.kt
```diff
@@ -70,7 +70,11 @@ internal object SpdxDocumentModelMapper : Logging {
                 ortResult
             )
 
-            ortResult.getDependencies(project.id, 1).mapTo(relationships) { dependency ->
+            ortResult.getDependencies(
+                id = project.id,
+                maxLevel = 1,
+                omitExcluded = false
```
sschuberth: At this commit, the `omitExcluded` parameter does not exist yet and should be removed here and below.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentModelMapper.kt
```diff
@@ -70,7 +70,11 @@ internal object SpdxDocumentModelMapper : Logging {
                 ortResult
             )
 
-            ortResult.getDependencies(project.id, 1).mapTo(relationships) { dependency ->
+            ortResult.getDependencies(
+                id = project.id,
```
sschuberth, line 74: Commit message nit: It's not "a type parameter" but "typed parameters" as also `id` is named now.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -188,17 +188,19 @@ data class OrtResult(
...
      * Return the dependencies of the given [id] (which can refer to a [Project] or a [Package]), up to and including a
      * depth of [maxLevel] where counting starts at 0 (for the [Project] or [Package] itself) and 1 are direct
-     * dependencies etc. A value below 0 means to not limit the depth.
+     * dependencies etc. A value below 0 means to not limit the depth. If [omitExcluded] is set to true, idenfitifers of
+     * excluded projects / packages are omitted from the result.
      */
-    fun getDependencies(id: Identifier, maxLevel: Int = -1): Set<Identifier> {
+    fun getDependencies(id: Identifier, maxLevel: Int = -1, omitExcluded: Boolean = false): Set<Identifier> {
         val dependencies = mutableSetOf<Identifier>()
+        val matcher = DependencyNavigator.MATCH_ALL.takeUnless { omitExcluded } ?: { !isExcluded(it.id) }
```
sschuberth, line 196: How about introducing a `DependencyNavigator.MATCH_NON_EXCLUDED` for `{ !isExcluded(it.id) }`?

sschuberth (changes requested):

model/src/main/kotlin/OrtResult.kt
fviernau: That's not possible, because the `DependencyNavigator` does not know about the excluded state.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -188,17 +188,19 @@ data class OrtResult(
     /**
      * Return the dependencies of the given [id] (which can refer to a [Project] or a [Package]), up to and including a
      * depth of [maxLevel] where counting starts at 0 (for the [Project] or [Package] itself) and 1 are direct
-     * dependencies etc. A value below 0 means to not limit the depth.
+     * dependencies etc. A value below 0 means to not limit the depth. If [omitExcluded] is set to true, idenfitifers of
```
sschuberth, line 191: Typo in "idenfitifers".

sschuberth (approved):

PR #7489, opened 2023-09-12
fix(reporter): Align setting `licenseInfoInFiles` with the spec v2.2

fviernau:
The description for the `filesAnalyzed` attribute says that "If false, the package shall not contain any files.", see [1]. This in turn implies that `licenseInfoInFiles` must be empty in that case.

Running [2] against an SPDX document which has a package with `filesAnalyzed=false` and a non-empty value for `licenseInfoInFiles` yields an error saying that the document is not valid for mentioned reason.

Fix that issue by reporting non-empty `licenseInfoInFiles` only for VCS and source artifact packages.

[1] https://github.com/spdx/spdx-spec/blob/development/v2.2.1/chapters/package-information.md#78-files-analyzed-field-
[2] https://github.com/spdx/tools-python

### DIFF
--- a/plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
+++ b/plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
@@ -47,7 +47,6 @@
     "homepage" : "first package's homepage URL",
     "licenseConcluded" : "BSD-2-Clause AND BSD-3-Clause AND MIT",
     "licenseDeclared" : "BSD-3-Clause AND (MIT OR GPL-2.0-only)",
-    "licenseInfoFromFiles" : [ "Apache-2.0", "BSD-2-Clause" ],
     "name" : "first-package",
     "summary" : "A package with all supported attributes set, with a VCS URL containing a user name, and with two scan results for the VCS containing copyright findings matched to a license finding.",
     "versionInfo" : "0.0.1"
@@ -89,7 +88,6 @@
     "homepage" : "first package's homepage URL",
     "licenseConcluded" : "NOASSERTION",
     "licenseDeclared" : "BSD-3-Clause AND (MIT OR GPL-2.0-only)",
-    "licenseInfoFromFiles" : [ "Apache-2.0", "BSD-2-Clause" ],
     "name" : "first-package",
     "summary" : "A package with all supported attributes set, with a VCS URL containing a user name, and with two scan results for the VCS containing copyright findings matched to a license finding.",
     "versionInfo" : "0.0.1"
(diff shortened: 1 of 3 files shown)


plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
mnonnenmacher, line 1: Commit message: "licenseInfoInFiles" -> "licenseInfoFromFiles" in several places.

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -170,15 +170,16 @@ internal fun Package.toSpdxPackage(
...
+        licenseInfoFromFiles = emptyList<String>().takeIf { packageVerificationCode == null } ?:
+            licenseInfoResolver.resolveLicenseInfo(id)
+                .filterExcluded()
+                .filter(LicenseView.ONLY_DETECTED)
+                .map { resolvedLicense ->
+                    resolvedLicense.license.takeIf { it.isValid(SpdxExpression.Strictness.ALLOW_DEPRECATED) }
+                        .nullOrBlankToSpdxNoassertionOrNone()
+                }
+                .distinct()
+                .sorted(),
```
mnonnenmacher, line 182: What I don't understand about the issue is that in the cases where the `licenseInfoFromFiles` is removed in the expected results files there were obviously detected licenses, but `filesAnalyzed` is `false`. As the `packageVerificationCode` is used to set the `filesAnalyzed` property, why is there no package verification code in cases where there are detected licenses?

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -170,15 +170,16 @@ internal fun Package.toSpdxPackage(
...
-        licenseInfoFromFiles = licenseInfoResolver.resolveLicenseInfo(id)
-            .filterExcluded()
-            .filter(LicenseView.ONLY_DETECTED)
-            .map { resolvedLicense ->
-                resolvedLicense.license.takeIf { it.isValid(SpdxExpression.Strictness.ALLOW_DEPRECATED) }
-                    .nullOrBlankToSpdxNoassertionOrNone()
-            }
-            .distinct()
-            .sorted(),
+        licenseInfoFromFiles = emptyList<String>().takeIf { packageVerificationCode == null } ?:
```
mnonnenmacher, line 173: I would prefer a simple `if/else` for readability.

mnonnenmacher (changes requested):

plugins/reporters/spdx/src/main/kotlin/Extensions.kt
fviernau: In the `SpdxDocument` the reporter produces, each dependency is represented by two packages:

1. The binary package.
2. The source artifact package OR the VCS package, depending on what was scanned.

Previously, `licenseInfoFromFiles` was included redundantly for both 1 and 2. Now, it is only reported for 2. Also, analog situation for the package verification code.

sschuberth (approved):

PR #7490, opened 2023-09-12
SpdxDocumentReporter: Adhere to detected licenses when computing extracted license info

fviernau:
This eliminates one type of violation of the SPDX spec V 2.2 from produced documents.

See individual commits.


plugins/reporters/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -228,7 +228,7 @@ internal fun SpdxDocument.addExtractedLicenseInfo(licenseTextProvider: LicenseTe
         packages.forEach {
             add(it.licenseConcluded)
             add(it.licenseDeclared)
-            // TODO: Also add detected non-SPDX licenses here.
+            addAll(it.licenseInfoFromFiles)
```
sschuberth, line 230: Commit message: The last sentence should end with `.` instead of `,`.

sschuberth (changes requested):

sschuberth (approved):

PR #7492, opened 2023-09-13
refactor(helper-cli): Replace the custom `runMain` with `CliktCommand.test`

sschuberth:
Note that a future version of clikt will also support a vararg for `test` for convenience [1]. According code can be migrated once that version is available.

[1]: https://github.com/ajalt/clikt/pull/451


fviernau (approved):
Great, thanks!

PR #7493, opened 2023-09-13
Miscellaneous small improvements to the requirements command

sschuberth:

### DIFF
--- a/plugins/commands/requirements/src/main/kotlin/RequirementsCommand.kt
+++ b/plugins/commands/requirements/src/main/kotlin/RequirementsCommand.kt
@@ -118,7 +118,6 @@ class RequirementsCommand : OrtCommand(
             echo("${category}s:")
 
             tools.forEach { tool ->
-                // TODO: State whether a tool can be bootstrapped, but that requires refactoring of CommandLineTool.
                 val message = buildString {
                     val (prefix, suffix) = if (tool.isInPath() || File(tool.command()).isFile) {
                         runCatching {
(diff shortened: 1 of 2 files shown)


sschuberth:
With this PR the `ort requirements` output looks like this:

image

plugins/commands/requirements/src/main/kotlin/RequirementsCommand.kt
```diff
@@ -177,20 +182,20 @@ class RequirementsCommand : OrtCommand(
         }
 
         echo("Prefix legend:")
-        echo("\t- The tool was not found in the PATH environment.")
-        echo("\t+ The tool was found in the PATH environment, but not in the required version.")
-        echo("\t* The tool was found in the PATH environment in the required version.")
+        echo("${DANGER_PREFIX}The tool was not found in the PATH environment.")
```
fviernau, line 185: Would it make sense to remove the trailing whitespace from the prefixes and get rid of the curly braces?

fviernau (approved):

plugins/commands/requirements/src/main/kotlin/RequirementsCommand.kt
sschuberth: I was a bit undecided about this myself, but in end went for the current solution to also reduce the duplication of the trailing whitespaces, and the `..._PREFIX` variable names IMO fit better if the whitespaces are included.

Issue #6465, opened 2023-02-09
AnalyzerResultBuilder: improve error message for duplicated packages


bennati:
The build fails if packages and projects have the same name, returning a list of the duplicated package IDs

https://github.com/oss-review-toolkit/ort/blob/13f5dd27592b20956cf359f4c83645a9dd3a2111/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt#L50

This error makes troubleshooting hard, as these IDs could be transitive dependencies and so they don't provide enough information.
Update the error message to also contain the section of the dependency tree that includes each of these packages, i.e. a sequence of packages and the files where these are defined. This would allow to understand which direct dependencies cause the issue and enable users to fix their build system.

bennati:
Happy to work on the implementation, if someone explains where to find the required information.

bennati:
I looked into this but I don't understand how to make sense of the dependency graph, are there any examples of how to work with it?

sschuberth:
@oheger-bosch can probably help here.

oheger-bosch:
Hi @bennati, I assume to better debug such problems something like the paths to dependencies as shown by the WebApp report would make sense. You might have a look how these paths are generated using the _dependency navigator_ API in [EvaluatedModelMapper](https://github.com/oss-review-toolkit/ort/blob/19c89ff9a0a7aa2a52d85c82fd477531da1ecf3d/reporter/src/main/kotlin/reporters/evaluatedmodel/EvaluatedModelMapper.kt#L692).

AIUI, the error happens for two different causes: If there is a project and a package with the same identifier, or if there are multiple packages with the same identifier. I am not sure whether these cases need to be distinguished when generating a more meaningful error message.

Recently, I encountered the latter case. Here a package was referenced by both the _GoMod_ and the _GoDep_ package managers. Unfortunately, the package managers produced slightly different metadata (which may also be caused by the complex repository layout using Git submodules), so two different `Package` objects were created.

HTH

devcooch:
Maybe @oheger-bosch and @sschuberth you could clarify on another question which pops up wrt this duplicate packages topic.

We are getting e.g. this duplicate:
`id=Identifier(type=Maven, namespace=junit, name=junit, version=4.12)`

The only difference between them is that one is referring company's internal artifactory, other one is in public Apache.
**INTERNAL:**
```
[...]

PR #7494, opened 2023-09-13
Enhanced functionality to improve developer experience by identifying…

dimitris-iliou:
… parent packages or projects from duplicate packages. This commit introduces new features that parse the dependency graph, starting from the duplicate package and ending at the parent package/project that encompasses it.

#6465

### DIFF
--- a/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
+++ b/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt
@@ -44,12 +44,48 @@ class AnalyzerResultBuilder {
     private val issues = mutableMapOf<Identifier, List<Issue>>()
     private val dependencyGraphs = mutableMapOf<String, DependencyGraph>()
 
+    private fun getPaths(
+        dependencyGraph: DependencyGraph?,
+        identifier: Identifier,
+    ): MutableList<Identifier> {
+
+        val packageName = dependencyGraph!!.packages.find { it == identifier   }  // gets the whole Identifier of the duplicate
+        val packageNameId = dependencyGraph.packages.indexOf(packageName) // gets the id from the duplicate
+        var packageList: MutableList<Identifier> = mutableListOf<Identifier>()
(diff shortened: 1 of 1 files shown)


sschuberth:
Thanks or the contribution @dimitris-iliou. But before you put too much effort in it, be informed that during one of the last ORT community meetings it was agreed that the best solution to address the issue is to not have duplicate IDs between project and packages to begin with, and deduplicate these by replacing references with package linkage in the dependency tree by references that link to the project instead.

scottschreckengaust:
> Thanks or the contribution @dimitris-iliou. But before you put too much effort in it, be informed that during one of the last ORT community meetings it was agreed that the best solution to address the issue is to not have duplicate IDs between project and packages to begin with, and deduplicate these by replacing references with package linkage in the dependency tree by references that link to the project instead.

@sschuberth : How might multiple subdirectories with the same python package dependencies within each `requirements.txt` file be constructed to not throw an exception?

```
20:18:16.195 [DefaultDispatcher-worker-1] ERROR java.lang.Object - Multiple projects with the same id 'PIP::src:1b11e6b573ecdb3a109776554dc04462bc00a82b' found. Not adding the project defined in 'https://github.com/organization/repository.git/src/module1/requirements.txt' to the analyzer results as it duplicates the project defined in 'https://github.com/organization/repository/src/module2/requirements.txt'.
Exception in thread "main" java.lang.IllegalArgumentException: Unable to create the AnalyzerResult as it contains packages and projects with the same ids: 
...
```

 from this TODO https://github.com/oss-review-toolkit/ort/blob/956d12e4ec820c39a5fc0931d30e0e3a3d692407/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt#L57-L58.

causing this exception https://github.com/oss-review-toolkit/ort/blob/956d12e4ec820c39a5fc0931d30e0e3a3d692407/analyzer/src/main/kotlin/AnalyzerResultBuilder.kt#L46-L47

For example:
```
`+ src
|_ src/module1/requirements.txt
|_ src/module2/requirements.txt
```

PR #7518, opened 2023-09-15
docs: Move the docs and toturial directories to the root

mnonnenmacher:
Move the docs and tutorial directories from the docusaurus directory to the root of the repository to make them easier to discover when looking at the repository.


mnonnenmacher:
@tsteenbe This moves the docs and tutorial directories to the root. Regarding the renaming of the docusaurus directory I'd like some other opinions, e.g. from @sschuberth or @fviernau. For me "website" would be fine but I'm open to suggestions.

sschuberth:
> For me "website" would be fine but I'm open to suggestions.

What I don't like so much about "website" is that IIUC the directory doesn't *really* contain the (generated) website, but the code of the framework to generate it. In that sense "docusaurus" was IMO more fitting. But I do not care all too strong either way.

sschuberth:
Commit message typo in "toturial".

PR #7522, opened 2023-09-18
Missing license info

heliocastro:
License page has only a link but no mention of what license ORT is using

### DIFF
--- a/website/docs/license.md
+++ b/website/docs/license.md
@@ -6,8 +6,7 @@ sidebar_position: 8
 
 Copyright (C) 2017-2023 [The ORT Project Authors](https://github.com/oss-review-toolkit/ort/blob/main/NOTICE).
 
-See the [LICENSE](https://github.com/oss-review-toolkit/ort/blob/main/LICENSE) file in the root of this project for
-license details.
+Licensed under the [Apache License, Version 2.0](https://github.com/oss-review-toolkit/ort/blob/main/LICENSE).
 
 OSS Review Toolkit (ORT) is a [Linux Foundation project](https://www.linuxfoundation.org) and part of
 [ACT](https://automatecompliance.org/).


docusaurus/docs/license.md
```diff
@@ -6,6 +6,7 @@ sidebar_position: 8
 
 Copyright (C) 2017-2023 [The ORT Project Authors](https://github.com/oss-review-toolkit/ort/blob/main/NOTICE).
 
+OSS Review Toolkit (ORT) is licensed under the Apache License, Version 2.0.
```
sschuberth, line 9: Why isn't the link directly below this line enough, as the docs are also hosted in the repo the link points to?

heliocastro: Unfortunately no, as the moment that i show to some persons, they asked why the license is not mention there.
And is really clear for persons to need to do a one extra jump to a not user friendly page of github as non-developers.

mnonnenmacher: I agree that the website should mention the license directly without requiring to click on a link.

mnonnenmacher (changes requested):
@heliocastro Please remove the merge commit and rebase your commit instead.

docusaurus/docs/license.md
heliocastro: Done

sschuberth: Ok, still it feels like duplication what we have here now. How about doing

    OSS Review Toolkit (ORT) is licensed under the [Apache License, Version 2.0](https://github.com/oss-review-toolkit/ort/blob/main/LICENSE).

and omitting the "See the LICENSE" sentence from below then?

heliocastro: Nope, because you need something indicating for the full license text.
This is a normal practice everywhere. A good example would be the projects page in github, where the main license ( or conjunction ) is displayed on the right, with a deep link for the full file.

sschuberth: I'm not following you. The full license text is still linked. It's just that the link is now called "Apache License, Version 2.0" instead of "LICENSE".

website/docs/license.md
```diff
@@ -6,8 +6,7 @@ sidebar_position: 8
 
 Copyright (C) 2017-2023 [The ORT Project Authors](https://github.com/oss-review-toolkit/ort/blob/main/NOTICE).
 
-See the [LICENSE](https://github.com/oss-review-toolkit/ort/blob/main/LICENSE) file in the root of this project for
-license details.
+OSS Review Toolkit (ORT) is licensed under the [Apache License, Version 2.0](https://github.com/oss-review-toolkit/ort/blob/main/LICENSE).
```
sschuberth, line 9: There's now an ugly duplication of "OSS Review Toolkit (ORT) is ..." here and below. I wonder whether we should combine this further to

```
OSS Review Toolkit (ORT) is a [Linux Foundation project]https://www.linuxfoundation.org) as part of [ACT](https://automatecompliance.org/) and licensed under the [Apache License, Version 2.0](https://github.com/oss-review-toolkit/ort/blob/main/LICENSE).
```

@mnonnenmacher what do you think?

heliocastro: I did exactly like that but then decided just commit first one, to see if was not my impression only.
Changing to that

heliocastro: How about this way:
image

tsteenbe (dismissed):

website/docs/license.md
```diff
@@ -6,8 +6,7 @@ sidebar_position: 8
 
 Copyright (C) 2017-2023 [The ORT Project Authors](https://github.com/oss-review-toolkit/ort/blob/main/NOTICE).
 
-See the [LICENSE](https://github.com/oss-review-toolkit/ort/blob/main/LICENSE) file in the root of this project for
-license details.
+Licensed under the [Apache License, Version 2.0](https://github.com/oss-review-toolkit/ort/blob/main/LICENSE)
```
sschuberth, line 9: Missing dot at end of sentence.

sschuberth (changes requested):

sschuberth (approved):

PR #7530, opened 2023-09-19
feat(migrate): Add path conventions for package configuration files

mnonnenmacher:
Apply the path conventions from the ort-config repository also to package configuration files.

### DIFF
--- a/plugins/commands/migrate/src/main/kotlin/MigrateCommand.kt
+++ b/plugins/commands/migrate/src/main/kotlin/MigrateCommand.kt
@@ -92,6 +92,7 @@ class MigrateCommand : OrtCommand(
             }
         }.toMap()
 
+        val configsDir = getCommonParentFile(pkgConfigFiles.keys)
         candidateFiles -= pkgConfigFiles.keys
         echo("Skipping ${candidateFiles.size} files of unknown format.")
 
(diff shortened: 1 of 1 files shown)


plugins/commands/migrate/src/main/kotlin/MigrateCommand.kt
```diff
@@ -169,3 +182,12 @@ private fun convertHoconToYaml(hocon: String): String {
...
 }
+
+/**
+ * Return the path for the directory to store [PackageConfiguration]s for this [Identifier].
+ *
+ * TODO: Replace with equivalent function from the package configuration provider for the ort-config repository, once
+ *       implemented.
+ */
+fun Identifier.toPackageConfigurationDir() =
+    "${type.encodeOr("_")}/${namespace.encodeOr("_")}/${name.encodeOr("_")}/${version.encodeOr("_")}"
```
sschuberth, line 193: Can be shortened to `toPath(emptyValue = "_")` like done in the [*helper-cli*'s `CreateCommand` for package configurations](https://github.com/oss-review-toolkit/ort/blob/5a92da88ce2176de3bb9e979a39a3e54cecb2282/helper-cli/src/main/kotlin/commands/packageconfig/CreateCommand.kt#L152C25-L152C50).

sschuberth (changes requested):

plugins/commands/migrate/src/main/kotlin/MigrateCommand.kt
```diff
@@ -169,3 +182,12 @@ private fun convertHoconToYaml(hocon: String): String {
...
     return yamlLines.joinToString("\n")
 }
+
+/**
+ * Return the path for the directory to store [PackageConfiguration]s for this [Identifier].
+ *
+ * TODO: Replace with equivalent function from the package configuration provider for the ort-config repository, once
+ *       implemented.
+ */
+fun Identifier.toPackageConfigurationDir() =
```
fviernau, line 192: Instead of this function, could you use `Identifier.toPath(emptyValue = "_")`?

fviernau (changes requested):

plugins/commands/migrate/src/main/kotlin/MigrateCommand.kt
sschuberth: Isn't this the same as my comment above?

fviernau: It is yes. I haven't seen it. Maybe I didn't hit F5 right before the review.

sschuberth (approved):

fviernau (approved):
LGTM, but unused imports are left....

sschuberth (approved):

fviernau (approved):

Issue #7476, opened 2023-09-07
Python 3.11 Support


mawl:
Hi there,

As a follow-up to https://github.com/oss-review-toolkit/ort/issues/7333:

How can I teach poetry / pip / python-inspector to use python version 3.11 instead of 3.10?

https://github.com/oss-review-toolkit/ort/issues/5740 says, that this should work:

.ort.yml
```
analyzer:
  package_managers:
    PIP:
      options:
        operatingSystem: 'linux'
        pythonVersion: '3.11'
```

I have build your Dockerfile with actual python components:
```
      --build-arg PYTHON_VERSION=3.11.4
      --build-arg PYENV_GIT_TAG=v2.3.23
      --build-arg PIPTOOL_VERSION=23.2.1      
      --build-arg PYTHON_POETRY_VERSION=1.5.1 
```

`pyenv versions` says: 
`* 3.11.4 (set by /opt/python/version)`

But this get's executed and the dependency tree stays empty:

`python-inspector --python-version 310 --operating-system linux ...`

Full log:
```
13:24:30.432 [main] INFO  org.ossreviewtoolkit.analyzer.Analyzer - Starting Poetry analysis.
13:24:30.438 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.analyzer.PackageManager - Using Poetry to resolve dependencies for path 'test/python/poetry/poetry.lock'...
13:24:30.448 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.plugins.packagemanagers.python.Poetry - Generating 'requirements-from-poetry.txt' file in '/builds/company/compliance/license-scanning/test/python/poetry' directory...
13:24:30.455 [DefaultDispatcher-worker-1] DEBUG org.ossreviewtoolkit.utils.common.EnvironmentVariableFilter - Filtering out these variables from the environment: [...
13:24:30.459 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.utils.common.ProcessCapture - Running 'poetry export --without-hashes --format=requirements.txt' in '/builds/company/compliance/license-scanning/test/python/poetry'...
13:24:30.997 [DefaultDispatcher-worker-1] DEBUG org.ossreviewtoolkit.utils.common.ProcessCapture - dnspython==2.4.1 ; python_version >= "3.11" and python_version < "3.12"
[...]

Issue #7333, opened 2023-07-25
Update Poetry to v1.5.1


mawl:
Hi there,

Would it be possible that you update Poetry to the latest released version [1.5.1](https://github.com/python-poetry/poetry/releases/tag/1.5.1)?

https://github.com/oss-review-toolkit/ort/blob/eef6f55fe215a0707d2b6833108f15be46b3fded/Dockerfile#L141

Or are there any issues preventing you from this?

We are using Poetry 1.5.1 in our projects and getting an error during `ort analyze`:

```
Poetry failed to resolve dependencies for path 'poetry.lock': IOException: Running 'poetry export --without-hashes --format=requirements.txt' in '...' failed with exit code 1: The Poetry configuration is invalid: - [source.0] Additional properties are not allowed ('priority' was unexpected) 
```

Thanks :)

sschuberth:
In any case @mawl, the `PYTHON_POETRY_VERSION` is a Docker build argument, so you should be able to easily customize it to your needs when building the Docker image, like `docker build --build-arg PYTHON_POETRY_VERSION=1.5.1 ...`.

mawl:
@sschuberth: Thanks - I updated it like this and the error as described above gets fixed - but I have additional issues with poetry and the python-exporter:

Non-working example using pendulum as dependency:
```
12:21:39.302 [DefaultDispatcher-worker-1] DEBUG org.ossreviewtoolkit.utils.common.ProcessCapture -   File "/opt/python/versions/3.11.4/lib/python3.11/site-packages/python_inspector/resolution.py", line 784, in get_setup_requirements
12:21:39.302 [DefaultDispatcher-worker-1] DEBUG org.ossreviewtoolkit.utils.common.ProcessCapture -     raise Exception(f"No setup.py or setup.cfg found in pypi sdist {sdist_location}")
12:21:39.302 [DefaultDispatcher-worker-1] DEBUG org.ossreviewtoolkit.utils.common.ProcessCapture - Exception: No setup.py or setup.cfg found in pypi sdist .cache/python_inspector/extracted_sdists/pendulum-2.1.2/pendulum-2.1.2
```

It works with pymongo, which contains a setup.py file - but when both dependencies (pendulum, pymongo) are in one project, no dependencies gets found either.

```
[tool.poetry]
name = "poetry-demo"
version = "0.1.0"
description = ""
authors = ["mawl@gmx.de"]
readme = "README.md"
packages = [{include = "poetry_demo"}]

[...]

PR #7533, opened 2023-09-19
feat(python): #7476 - Support Python 3.11

mawl:
Fix issue #7476 - Support Python 3.11

Python 3.11 is supported by python-inspector already: https://github.com/nexB/python-inspector/commit/8c6ee63491172bd0d9ddbc13c65c9a35092e35de 

As it is my first PR in your project, feel free to correct it and get me on the right track.


plugins/package-managers/python/src/main/kotlin/Pip.kt
```diff
@@ -39,8 +39,8 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
 private const val OPTION_OPERATING_SYSTEM_DEFAULT = "linux"
 private val OPERATING_SYSTEMS = listOf(OPTION_OPERATING_SYSTEM_DEFAULT, "macos", "windows")
 
-private const val OPTION_PYTHON_VERSION_DEFAULT = "3.10"
-private val PYTHON_VERSIONS = listOf("2.7", "3.6", "3.7", "3.8", "3.9", OPTION_PYTHON_VERSION_DEFAULT)
+private const val OPTION_PYTHON_VERSION_DEFAULT = "3.11"
```
sschuberth, line 42: We cannot really make this the default version already in this commit, if only the next commit is adding that version.

Dockerfile
```diff
@@ -126,8 +126,8 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     tk-dev \
     && sudo rm -rf /var/lib/apt/lists/*
 
-ARG PYTHON_VERSION=3.10.8
-ARG PYENV_GIT_TAG=v2.3.7
+ARG PYTHON_VERSION=3.11.4
```
sschuberth, line 129: Please explain in more detail in the commit message why bumping all of these tool versions is required. Is it really that all the previous versions did not support Python 3.11 yet?

sschuberth (changes requested):
Please rewrite commit messages to adhere to Conventional Commits, and capitalize proper nouns like "Python".

plugins/package-managers/python/src/main/kotlin/Pip.kt
mawl: To fix this, I changed the order of the commits.

Dockerfile
mawl: I cannot explain this for all tools, only for Pyenv and Poetry. Should I downgrade them and test it?

sschuberth: Well, basically yes. Or put the other way: You should have only upgraded those tools which are required to be updated in this context / to work with Python 3.11.

plugins/package-managers/python/src/main/kotlin/Pip.kt
mawl: To fix this finally I have squashed all commits to one.

Dockerfile
mawl: Update of not relevant tools reverted.

mawl:
@sschuberth: Thanks for your support. I have squashed the commits and upgraded the relevant tools (python, pyenv and poetry). Tests are green now. Is there anything else I can do or can we merge it now?

sschuberth:
@mawl by now the commit diff doesn't seem to match the commit message anymore, as only the Poetry version is bumped.

mawl:
> @mawl by now the commit diff doesn't seem to match the commit message anymore, as only the Poetry version is bumped.

I squashed too much and the changes are gone. I'll take a look.

mawl:
> @mawl by now the commit diff doesn't seem to match the commit message anymore, as only the Poetry version is bumped.

Changes are back now.

sschuberth:
I'm basically ok with the changes, but I'm not in a position to judge whether this would break things for people at Bosch or EPAM. So I'd prefer @mnonnenmacher / @MarcelBochtler and @fviernau to have a look.

Dockerfile
```diff
@@ -138,7 +138,7 @@ RUN curl -kSs https://pyenv.run | bash \
 ARG CONAN_VERSION=1.57.0
 ARG PYTHON_INSPECTOR_VERSION=0.9.8
 ARG PYTHON_PIPENV_VERSION=2022.9.24
-ARG PYTHON_POETRY_VERSION=1.2.2
+ARG PYTHON_POETRY_VERSION=1.6.1
```
mnonnenmacher, line 141: Please upgrade the Poetry version in Dockerfile-legacy as well. Dockerfile-legacy is used by CI, maybe that is related to the failing tests.

mnonnenmacher:
> I'm basically ok with the changes, but I'm not in a position to judge whether this would break things for people at Bosch or EPAM. So I'd prefer @mnonnenmacher / @MarcelBochtler and @fviernau to have a look.

We use the Dockerfile-legacy so the Python version upgrade would not affect us. Changing the default version used by python-inspector should be fine as it is easy to configure the version to be used.

Dockerfile
mawl: @mnonnenmacher: I hope you are fine with changing python-inspector to 0.9.8 as well, as this will resolve: https://github.com/oss-review-toolkit/ort/issues/7333#issuecomment-1667728868

mnonnenmacher: Yes please, versions should in general stay in sync. Installing Python 3.11 would be more difficult with the legacy file, so I think it's fine to stay with the default 3.10 there. python-inspector should still support 3.11 as long as no setup.py file which depends on 3.11 is involved.

heliocastro:
With the new docker merged, all version changes should be done in one single place, .versions
So both Docker, Docker extended and the local docker_build script will know how to build or not the modified image.

heliocastro:
Please modify the .versions file as well

heliocastro (approved):
Looks good to me

Dockerfile
sschuberth, line 138: > to fix https://github.com/oss-review-toolkit/ort/issues/7333

It's confusing to refer to that issue as it's already closed.

Dockerfile-legacy
```diff
@@ -33,7 +33,7 @@ ARG ORT_VERSION="DOCKER-SNAPSHOT"
 ARG NUGET_INSPECTOR_VERSION=0.9.12
 
 # Set this to the Python Inspector version to use.
-ARG PYTHON_INSPECTOR_VERSION="0.9.6"
+ARG PYTHON_INSPECTOR_VERSION="0.9.8"
```
sschuberth, line 36: > Both updates are necessary to fix https://github.com/oss-review-toolkit/ort/issues/7333 and the CI pipeline

Again, it's confusing to refer to that issue as it's already closed, and the previous already was said to fix it, too.

sschuberth (changes requested):
We do not allow merge commits in PRs.

Dockerfile
fviernau: Can you explain why upgrading python is needed in order to support 3.11?

In particular, would it suffice to stick to 3.10 (not saying we should) and only pass `-p 3.11` to python inspector by default ?

mawl:
I think we should stop the work here as more happens in https://github.com/oss-review-toolkit/ort/pull/7598 - I will close this PR for now.

PR #7537, opened 2023-09-19
fix(Scanner): Apply `detectedLicenseMapping` to package scanners

MarcelBochtler:
While the `detectedLicenseMapping` is already applied for path scanners, these mappings are not applied for package scanners. As FossId is not a path scanner, the mappings did not work. Also apply the mappings to package scanners.

This fixes an issue introduced in 6cc401a.


sschuberth:
> This fixes a regression introduced in [cdecb9b](https://github.com/oss-review-toolkit/ort/commit/cdecb9b7232c1e9c25a7f928d2a07e3851811987).

Sorry for having broken this. The intention of my commit was to "centralize" the mapping on the scanner-level, I just accidentally only applied it to *path* scanners. With FossId being a *package* scanner, can't we do the same centralization for package scanners (or even more ideally, for *any* scanner) instead of only applying it to FossId?

MarcelBochtler:
> With FossId being a _package_ scanner, can't we do the same centralization for package scanners (or even more ideally, for _any_ scanner) instead of only applying it to FossId?

I tried that. See: https://github.com/oss-review-toolkit/ort/compare/fb82e67c4972fc1e707a3c0735336206bd2e1b74..1385d27ac7ce6d1c22e4dec19e1bea26a704b88d (the removed code).

This was not working in my tests.
I don't see if or which central method FossID is using. That's why I implemented it in FossID directly.

mnonnenmacher:
> > With FossId being a _package_ scanner, can't we do the same centralization for package scanners (or even more ideally, for _any_ scanner) instead of only applying it to FossId?
> 
> I tried that. See: https://github.com/oss-review-toolkit/ort/compare/fb82e67c4972fc1e707a3c0735336206bd2e1b74..1385d27ac7ce6d1c22e4dec19e1bea26a704b88d (the removed code).
> 
> This was not working in my tests. I don't see if or which central method FossID is using. That's why I implemented it in FossID directly.

I looked at the diff and am really surprised that it didn't work, I don't understand why it shouldn't. I would be fine with merging this as a quick fix, but could you please add a TODO comment that the mapping of licenses should be moved to a central location?

mnonnenmacher (approved):

PR #7544, opened 2023-09-20
fix(CocoaPods): Added Sub Dependency to DependencyMap

### LINKED ISSUE #7523: Missing Key 'AFNetworking/UIKit' from CocoaPod in Dependency Map from ORT analyzer
While testing the  https://github.com/youusername/magnetX.git repository, I came across a crash issue. The problem is related to a missing key, specifically "AFNetworking/UIKit." It's worth noting that this issue is considered common in the majority of repositories that utilize AFNetworking.

```
9:33:11.814 [DefaultDispatcher-worker-1] INFO  org.ossreviewtoolkit.utils.common.ProcessCapture - Running 'pod repo add-cdn trunk https://cdn.cocoapods.org --allow-root' in '/Users/ksg/test/'...

java.util.NoSuchElementException: Key AFNetworking/UIKit is missing in the map.
        at kotlin.collections.MapsKt__MapsKt.getValue(Maps.kt:360)
        at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPodsKt.getPackageReferences$createPackageReference(CocoaPods.kt:256)
        at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPodsKt.getPackageReferences$createPackageReference(CocoaPods.kt:257)
        at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPodsKt.getPackageReferences(CocoaPods.kt:262)
  1 #!/bin/bash
        at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPodsKt.access$getPackageReferences(CocoaPods.kt:1)
        at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPods.resolveDependenciesInternal(CocoaPods.kt:130)
        at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPods.resolveDependencies(CocoaPods.kt:110)
        at org.ossreviewtoolkit.analyzer.PackageManager.resolveDependencies(PackageManager.kt:319)
        at org.ossreviewtoolkit.analyzer.PackageManagerRunner$run$3.invokeSuspend(Analyzer.kt:323)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
        at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:115)
        at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:103)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:584)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:793)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:697)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:684)
```

### LINKED ISSUE #7543: Incorrect PodSpec Name Leads to 'Key Missing' Error
While testing the [rn-tester](https://github.com/facebook/react-native/tree/27979dcd21ad43cfbf74ce249188594dbc8756e8/packages/rn-tester) project, I encountered the following errors:
```
java.util.NoSuchElementException: Key ScreenshotManager is missing in the map.
	at kotlin.collections.MapsKt__MapsKt.getValue(Maps.kt:360)
	at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPods.getPodspec(CocoaPods.kt:229)
	at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPods.getPackage(CocoaPods.kt:167)
	at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPods.resolveDependenciesInternal(CocoaPods.kt:133)
	at org.ossreviewtoolkit.plugins.packagemanagers.cocoapods.CocoaPods.resolveDependencies(CocoaPods.kt:110)
	at org.ossreviewtoolkit.analyzer.PackageManager.resolveDependencies(PackageManager.kt:319)
	at org.ossreviewtoolkit.analyzer.PackageManagerRunner$run$3.invokeSuspend(Analyzer.kt:323)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
	at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:115)
	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:103)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:584)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:793)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:697)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:684)
```

I believe this error is due to a different PodSpec filename than what we expected.
```
$ pod spec which "ScreenshotManager" --version=0.0.1 --allow-root --regex
/Users/REDACTED/.cocoapods/repos/trunk/Specs/3/3/1/KSScreenshotManager/0.0.1/KSScreenshotManager.podspec.json
->  It should be "ScreenshotManager," not "KSScreenshotManager"
```

The pod spec command searches for similar PodSpec files, not exact matches.

For example:
```
pod spec which "eenshotManageR" --version=0.0.1 --allow-root --regex
```
It may return a result like:
```
/Users/REDACTED/.cocoapods/repos/trunk/Specs/3/3/1/KSScreenshotManager/0.0.1/KSScreenshotManager.podspec.json
```

This means that it found a similar PodSpec file that matches the provided criteria. It's important to note that the search is not case-sensitive, and it tries to find the closest match based on the provided name and version. In this case, it found a PodSpec file named "KSScreenshotManager" that is similar to "eenshotManageR"

ksg97031:
Hi ORT Team,

Could you please review this?

Resolves #7523
ㄴ [fix(CocoaPods): Added Sub Dependency to DependencyMap](https://github.com/oss-review-toolkit/ort/pull/7544/commits/4daab7a72a74b4c2fe3f3aa05ef469ad2e2bb6f5)

~Resolves #7543
ㄴ [fix(cocoapods): Refined Podspec Name Matching for CocoaPods](https://github.com/oss-review-toolkit/ort/commit/9662fd312f0751323fb70541942698db47909808)~

### DIFF
--- a/plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
+++ b/plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
@@ -194,7 +194,7 @@ class CocoaPods(
 
         val podspecCommand = runCatching {
             run(
-                "spec", "which", podspecName,
+                "spec", "which", "^$podspecName$",
                 "--version=${id.version}",
                 "--allow-root",
                 "--regex",
@@ -249,6 +249,26 @@ private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
 
         val dependencies = node[entry]?.map { it.textValue().substringBefore(" ") }.orEmpty()
         dependenciesForName.getOrPut(name) { mutableSetOf() } += dependencies
+
+        // If this node has sub dependencies, append them to the dependency tracking.
+        // Sub dependencies are additional dependencies that are required.
+        // For more details, refer to issue #7523.
+        if (node is ObjectNode) {
+            val subEntries = node[entry].map { it.textValue() }
+            subEntries.filter { it.contains("(= ") }.forEach { subEntry ->
+                val (subDependencyName, subDependencyVersion) = NAME_AND_VERSION_REGEX.find(subEntry)!!.groups.let {
+                    it[1]!!.value to it[2]!!.value.removeSurrounding("(= ", ")")
+                }
+
+                // Add sub dependency if it's not already tracked
+                if (!versionForName.containsKey(subDependencyName)) {
+                    versionForName[subDependencyName] = subDependencyVersion
+
+                    val subDependencies = node[subEntry]?.map { it.textValue().substringBefore(" ") }.orEmpty()
+                    dependenciesForName.getOrPut(subDependencyName) { mutableSetOf() } += subDependencies
+                }
+            }
+        }
     }
 
     fun createPackageReference(name: String): PackageReference =


plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -194,7 +194,7 @@ class CocoaPods(
 
         val podspecCommand = runCatching {
             run(
-                "spec", "which", podspecName,
+                "spec", "which", "^$podspecName$",
                 "--version=${id.version}",
                 "--allow-root",
                 "--regex",
```
sschuberth, line 200: Just wondering, could we simply drop `--regex`, as I assume `podspecName` is always a literal string in our case?

ksg97031: > Just wondering, could we simply drop `--regex`, as I assume `podspecName` is always a literal string in our case?

I also experimented with removing the --regex option, but the result remained the same when I ran the following command:
```
$ pod spec which "coaSec" --version=1.2.4 --allow-root
/Users/REDACTED/.cocoapods/repos/cocoapods/Specs/d/e/9/CocoaSecurity/1.2.4/CocoaSecurity.podspec.json
```
I attempted to explore alternative options by using `pod spec which --help`, but I couldn't find any other options that seemed relevant to our situation.
```
$ pod spec which --help
Usage:

    $ pod spec which [QUERY]

      Prints the path of the .podspec file(s) whose name matches `QUERY`

Options:

    --regex        Interpret the `QUERY` as a regular expression
    --show-all     Print all versions of the given podspec
    --version      Print a specific version of the given podspec
    --allow-root   Allows CocoaPods to run as root
    --silent       Show nothing
    --verbose      Show more debugging information
    --no-ansi      Show output without ANSI codes
    --help         Show help banner of specified command
```

sschuberth: > but the result remained the same when I ran the following command:

But that's good, isn't it? If the result does not change without `--regex`, it means we can drop it, and avoid the need to make the regex more strict, right?

ksg97031: > > but the result remained the same when I ran the following command:
> 
> But that's good, isn't it? If the result does not change without `--regex`, it means we can drop it, and avoid the need to make the regex more strict, right?

What I meant by saying that the result remained the same is that it consistently produced an incorrect result. 
```
$ pod spec which "coaSec" --version=1.2.4 --allow-root
/Users/REDACTED/.cocoapods/repos/cocoapods/Specs/d/e/9/CocoaSecurity/1.2.4/CocoaSecurity.podspec.json
-> Resulted in a mismatch where you expected "coaSec," but it returned "CocoaSecurity."
```
To clarify, when I tested it, I found that in order to accurately match the exact string I was looking for, I needed to use `regex` checks with ^ and $.

sschuberth: But when run from ORT, `pod spec which` gets exclusively called with an `Identifier` name, which can never be a regex. Thus IMO the `--regex` is unneeded. And if it's not needed, also you stricter regex matching is not needed.

sschuberth: Here's my proposal for that: https://github.com/oss-review-toolkit/ort/pull/7552

sschuberth: Ah, now I believe to understand what you're saying. So the rationale for a proper commit message would be:

```
fix(CocoaPods): Restrict package name match to full matches

The both the default and the regex query string matching logic actually
match substrings case-insensitively. So if a package name is a substring of
another package name, invalid matches could occur. To circumvent that,
force the regex to be a full match.
```

If you agree with the rational, please copy & paste that to your commit as the message.

ksg97031: > Ah, now I believe to understand what you're saying. So the rationale for a proper commit message would be:
> 
> ```
> fix(CocoaPods): Restrict package name match to full matches
> 
> The both the default and the regex query string matching logic actually
> match substrings case-insensitively. So if a package name is a substring of
> another package name, invalid matches could occur. To circumvent that,
> force the regex to be a full match.
> ```
> 
> If you agree with the rational, please copy & paste that to your commit as the message.

I edited my commit message it as you suggested!

fviernau: I'm not sure about how controversial the other commit is, but this one now is IMO good to go.
We could split-out this commit to a separate PR and merge it already now? (makes sense @sschuberth ?)

sschuberth:
So I've debugged this and read up on the `Podfile.lock` structure to better understand what the problem is. I was able to solve this (I believe) in a simpler way, see https://github.com/oss-review-toolkit/ort/pull/7558.

@ksg97031 if you can confirm that my PR works for you, we can close this one.

sschuberth:
> @ksg97031 if you can confirm that my PR works for you, we can close this one.

Thanks for [confirming](https://github.com/oss-review-toolkit/ort/pull/7558#issuecomment-1731047363), closing this!

PR #7552, opened 2023-09-21
chore(CocoaPods): Do not use a regex for package name matching

sschuberth:
ORT exclusively runs `pod spec which` with an `Identifier`'s `name` which can never be a regex, so do not allow to interpret it as one.

### DIFF
--- a/plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
+++ b/plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
@@ -197,7 +197,6 @@ class CocoaPods(
                 "spec", "which", podspecName,
                 "--version=${id.version}",
                 "--allow-root",
-                "--regex",
                 workingDir = workingDir
             )
         }.getOrElse {


plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -197,7 +197,6 @@ class CocoaPods(
                 "spec", "which", podspecName,
                 "--version=${id.version}",
                 "--allow-root",
-                "--regex",
```
fviernau, line 200: If https://github.com/oss-review-toolkit/ort/pull/7544#discussion_r1333130483 is true, then we should keep the `--regex` and modify the regex such that it matches exactly as proposed in the linked PR, right ?

sschuberth: Yes, see https://github.com/oss-review-toolkit/ort/pull/7544#discussion_r1333153092.

PR #7558, opened 2023-09-22
More CocoaPods fixes

sschuberth:

Big thanks to @ksg97031 for reporting #7523 and proposing an initial solution in https://github.com/oss-review-toolkit/ort/pull/7544. This PR applies a simpler fix, plus two more.

### DIFF
--- a/plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/external-sources-expected-output.yml
+++ b/plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/external-sources-expected-output.yml
@@ -1,0 +1,286 @@
+---
+project:
+  id: "CocoaPods::src/funTest/assets/projects/synthetic/external-sources/Podfile:"
+  definition_file_path: "plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/external-sources/Podfile"
+  declared_licenses: []
+  declared_licenses_processed: {}
+  vcs:
+    type: ""
+    url: ""
+    revision: ""
+    path: ""
+  vcs_processed:
+    type: "Git"
+    url: "<REPLACE_URL_PROCESSED>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  homepage_url: ""
+  scopes:
+  - name: "dependencies"
+    dependencies:
+    - id: "Pod::AFNetworking:3.2.1"
+      dependencies:
+      - id: "Pod::AFNetworking/NSURLSession:3.2.1"
+        dependencies:
+        - id: "Pod::AFNetworking/Reachability:3.2.1"
+        - id: "Pod::AFNetworking/Security:3.2.1"
+        - id: "Pod::AFNetworking/Serialization:3.2.1"
+      - id: "Pod::AFNetworking/Reachability:3.2.1"
+      - id: "Pod::AFNetworking/Security:3.2.1"
+      - id: "Pod::AFNetworking/Serialization:3.2.1"
+      - id: "Pod::AFNetworking/UIKit:3.2.1"
+    - id: "Pod::CocoaSecurity:1.2.4"
+    - id: "Pod::MLHudAlert:0.0.4-3f23fa41632f8935fed028b04335576633ec53a2"
+    - id: "Pod::Ono:2.1.1"
+packages:
+- id: "Pod::AFNetworking:3.2.1"
+  purl: "pkg:cocoapods/AFNetworking@3.2.1"
+  declared_licenses:
+  - "MIT"
+  declared_licenses_processed:
+    spdx_expression: "MIT"
+  description: "A delightful iOS and OS X networking framework."
+  homepage_url: "https://github.com/AFNetworking/AFNetworking"
+  binary_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  source_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  vcs:
+    type: "Git"
+    url: "https://github.com/AFNetworking/AFNetworking.git"
+    revision: "3.2.1"
+    path: ""
+  vcs_processed:
+    type: "Git"
+    url: "https://github.com/AFNetworking/AFNetworking.git"
+    revision: "3.2.1"
+    path: ""
+- id: "Pod::AFNetworking/NSURLSession:3.2.1"
+  purl: "pkg:cocoapods/AFNetworking%2FNSURLSession@3.2.1"
+  declared_licenses:
+  - "MIT"
+  declared_licenses_processed:
+    spdx_expression: "MIT"
+  description: "A delightful iOS and OS X networking framework."
+  homepage_url: "https://github.com/AFNetworking/AFNetworking"
+  binary_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  source_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  vcs:
+    type: "Git"
+    url: "https://github.com/AFNetworking/AFNetworking.git"
+    revision: "3.2.1"
+    path: ""
+  vcs_processed:
+    type: "Git"
+    url: "https://github.com/AFNetworking/AFNetworking.git"
+    revision: "3.2.1"
+    path: ""
+- id: "Pod::AFNetworking/Reachability:3.2.1"
+  purl: "pkg:cocoapods/AFNetworking%2FReachability@3.2.1"
+  declared_licenses:
+  - "MIT"
+  declared_licenses_processed:
+    spdx_expression: "MIT"
+  description: "A delightful iOS and OS X networking framework."
+  homepage_url: "https://github.com/AFNetworking/AFNetworking"
+  binary_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  source_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  vcs:
+    type: "Git"
+    url: "https://github.com/AFNetworking/AFNetworking.git"
+    revision: "3.2.1"
+    path: ""
+  vcs_processed:
+    type: "Git"
+    url: "https://github.com/AFNetworking/AFNetworking.git"
+    revision: "3.2.1"
+    path: ""
+- id: "Pod::AFNetworking/Security:3.2.1"
+  purl: "pkg:cocoapods/AFNetworking%2FSecurity@3.2.1"
+  declared_licenses:
+  - "MIT"
+  declared_licenses_processed:
+    spdx_expression: "MIT"
+  description: "A delightful iOS and OS X networking framework."
+  homepage_url: "https://github.com/AFNetworking/AFNetworking"
+  binary_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  source_artifact:
+    url: ""
+    hash:
+      value: ""
+      algorithm: ""
+  vcs:
+    type: "Git"
+    url: "https://github.com/AFNetworking/AFNetworking.git"
+    revision: "3.2.1"
+    path: ""
+  vcs_processed:
+    type: "Git"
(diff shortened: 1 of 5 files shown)


plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -245,10 +248,8 @@ private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
             else -> node.textValue()
         }
 
-        val (name, version) = NAME_AND_VERSION_REGEX.find(entry)!!.groups.let {
-            it[1]!!.value to it[2]!!.value.removeSurrounding("(", ")")
-        }
-        versionForName[name] = version
+        val (name, version) = parseNameAndVersion(entry)
+        if (version != null) versionForName[name] = version
```
fviernau, line 266: Previously, for all names an entry existed in `versionForName`.
Now this does not hold anymore.

Can you explain why this is not problematic in line 261: `dependenciesForName.getValue(name)` ?

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -229,13 +229,16 @@ private const val LOCKFILE_FILENAME = "Podfile.lock"
...
 private const val SCOPE_NAME = "dependencies"
 
-private val NAME_AND_VERSION_REGEX = "(\\S+)\\s+(.*)".toRegex()
-
 private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
     val versionForName = mutableMapOf<String, String>()
     val dependenciesForName = mutableMapOf<String, MutableSet<String>>()
     val root = yamlMapper.readTree(podfileLock)
 
+    fun parseNameAndVersion(entry: String): Pair<String, String?> {
```
fviernau, line 245: This function does not use anything from this scope. Could you please move it out, so the code of `getPackageReferences()` is mixing less abstraction levels?

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -236,7 +236,13 @@ private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
...
     fun parseNameAndVersion(entry: String): Pair<String, String?> {
         val info = entry.split(' ', limit = 2)
-        return info[0] to info.getOrNull(1)?.removeSurrounding("(", ")")
+        val name = info[0]
+
+        // A version entry could look something like "(6.3.0)", "(= 2021.06.28.00-v2)", "(~> 8.15.0)", etc. Also see
+        // https://guides.cocoapods.org/syntax/podfile.html#pod.
+        val version = info.getOrNull(1)?.removeSurrounding("(", ")")?.substringAfterLast(' ')
+
+        return name to version
```
fviernau, line 253: Can it now happen that an entry in `versionForName` already exists and get overwritten in line 257?
If so, would the value always be overwritten by its identity, or could the value change?

plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/external-sources-expected-output.yml
```diff
@@ -0,0 +1,286 @@
+---
```
fviernau, line 1: commit message: as not metadata -> as  no...

ksg97031:
> Please have a look at the individual commit messages for the details.
> 
> Big thanks to @ksg97031 for reporting #7523 and proposing an initial solution in #7544. This PR applies a simpler fix, plus two more.

It works fine and looks good to me!!

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -229,7 +233,11 @@ private const val LOCKFILE_FILENAME = "Podfile.lock"
 
 private const val SCOPE_NAME = "dependencies"
 
-private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
+private fun parseLockFile(
```
fviernau, line 236: I prefer if the functioned kept on returning the model of the parse lockfile instead of making it modify two given mutable data structures, because IMO that remains simpler as there are no side effects.

```
data class Lockfile(
     val packageReferences: Set<PackageReference>
     val externalSources: List<Package>
)
```

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -271,10 +279,27 @@ private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
...
 
-    return root.get("DEPENDENCIES").mapTo(mutableSetOf()) { node ->
+    root.get("DEPENDENCIES").mapTo(references) { node ->
         val name = node.textValue().substringBefore(" ")
         createPackageReference(name)
     }
+
+    root.get("CHECKOUT OPTIONS")?.fields()?.asSequence()?.forEach {
+        val checkout = it.value as ObjectNode
+        val url = checkout[":git"].textValueOrEmpty()
```
fviernau, line 289: As currently only `git` is implemented here, how about skipping the iteration in case there is no key `:git` ?

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -271,10 +279,27 @@ private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
...
+        val revision = checkout[":commit"].textValueOrEmpty()
+        val id = Identifier("Pod", "", it.key, versionForName[it.key].orEmpty())
+
+        externalSources[id] = Package(
+            id = id,
```
fviernau, line 294: Would you know how the version string in the lockfile is determined?
I'm asking, because if commonly many revisions are used for same version string for such unpublished artifacts, maybe we should use the SHA1 in the version part of the identifier?

fviernau (changes requested):

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
sschuberth: No, I don't know. That why I'm using `versionForName[it.key].orEmpty()` when determining the `id` to eventually get a "real" version in addition to the SHA1.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
sschuberth: CocoaPods does not support having the same dependency in different versions, so *if* the version would be specified twice (which I've never seen in a `Podfile.lock`), it absolutely should be the same, and a value would get overwritten by its identity.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
sschuberth: > Now this does not hold anymore.

That's not actually true. The `version != null` condition will always be `true` here. Previously, this was hidden behind the `!!` assertions which I could have used here as well, but I consider their use as unclean. I'll change the code to use `checkNotNull()` instead to more clearly mark this as a precondition.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
sschuberth: > so that we do not risk that e.g. curation get filed for package versions which actually are used in combination with many different SHA1s.

Sorry, I do not understand that sentence / the risk. Could you maybe give a concrete example?

fviernau: > Sorry, I do not understand that sentence / the risk. Could you maybe give a concrete example?

I was wondering how the version string arrives at the lockfile. If the procedure starts with a specified the SHA1
and then determines the version based on

1. The package metadata found in the repo for given SHA1
2. Or the youngest Git tag preceeding the SHA1

then many SHA1's may correspond to a single ID.
That would be problematic with our package curation mechanism.
Does it become clearer with this?

sschuberth: > I was wondering how the version string arrives at the lockfile.

I've tested it at the example of `MLHudAlert` by specifying a custom commit in `Podfile` that does not point to a tag and regenerating the lockfile. The version that ends up in `Podfile.lock` is whatever the project's `.podspec` file contains as the version in the given commit.

So yes, the same `version` (as part of the `id`) could actually refer to / come from different SHA1s of the source code.

As I *do* like to have the version somewhere in there, what do you think about simply appending the SHA1 to the version in this case, like "0.0.4-3f23fa41632f8935fed028b04335576633ec53a2"?

fviernau: > > I was wondering how the version string arrives at the lockfile.
> 
> I've tested it at the example of `MLHudAlert` by specifying a custom commit in `Podfile` that does not point to a tag and regenerating the lockfile. The version that ends up in `Podfile.lock` is whatever the project's `.podspec` file contains as the version in the given commit.
> 
> So yes, the same `version` (as part of the `id`) could actually refer to / come from different SHA1s of the source code.
> 
> As I _do_ like to have the version somewhere in there, what do you think about simply appending the SHA1 to the version in this case, like "0.0.4-3f23fa41632f8935fed028b04335576633ec53a2"?

This solves indeed the issue I raised. I'm good with that solution.

sschuberth: > I'm good with that solution.

Implemented, please have another look.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -251,14 +257,18 @@ private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
         val (name, version) = parseNameAndVersion(entry)
         versionForName[name] = checkNotNull(version)
 
-        val dependencies = node[entry]?.map { it.textValue().substringBefore(" ") }.orEmpty()
+        val dependencies = node[entry]?.map { depNode ->
```
fviernau, line 267: Is this `depNode` of same format as `node`? e.g. would the case distinction from line 252 through 255 need to be applied as well?

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -251,14 +257,18 @@ private fun getPackageReferences(podfileLock: File): Set<PackageReference> {
...
+            val (depName, depVersion) = parseNameAndVersion(depNode.textValue())
+            depName.also { if (depVersion != null) versionForName[it] = depVersion }
```
fviernau, line 269: Can line 262 be simplified to just `if (depVersion != null) versionForName[it] = depVersion` ?

fviernau (changes requested):

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
fviernau: Or maybe `depVersion?.let { versionForName[depName] = it }`

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
fviernau: Thanks!

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
sschuberth: No, as `depName` needs to be returned for `map`.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
sschuberth: No, `depNode` is always a text node (as nesting can be max. 2 levels deep in the `PODS` section).

fviernau (approved):

PR #7560, opened 2023-09-22
fix(CocoaPods): Do not add version constraints as concrete versions

### LINKED ISSUE #5841: Support getting metadata for CocoaPods packages hosted in custom Git repos
Thanks for the report @unclejay80! As a stop-gap measure I've created https://github.com/oss-review-toolkit/ort/pull/5835. But you'll still be missing metadata for that package.

_Originally posted by @sschuberth in https://github.com/oss-review-toolkit/ort/issues/5834#issuecomment-1252838215_

sschuberth:

### DIFF
--- a/plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
+++ b/plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
@@ -237,7 +237,7 @@ private fun parseNameAndVersion(entry: String): Pair<String, String?> {
 
     // A version entry could look something like "(6.3.0)", "(= 2021.06.28.00-v2)", "(~> 8.15.0)", etc. Also see
     // https://guides.cocoapods.org/syntax/podfile.html#pod.
-    val version = info.getOrNull(1)?.removeSurrounding("(", ")")?.substringAfterLast(' ')
+    val version = info.getOrNull(1)?.removeSurrounding("(", ")")
 
     return name to version
 }
@@ -252,24 +252,24 @@ private fun parseLockfile(podfileLock: File): LockfileData {
     val dependencyConstraints = mutableMapOf<String, MutableSet<String>>()
     val root = yamlMapper.readTree(podfileLock)
 
-    // The "PODS" section lists the direct dependencies and, if applicable, their direct dependencies each. That is, the
-    // nesting never goes deeper than two levels, and in order to build up a full dependency tree, one needs to
-    // recursively step through all direct dependencies.
-    root.get("PODS").asIterable().forEach { node ->
-        val entry = when (node) {
-            is ObjectNode -> node.fieldNames().asSequence().first()
-            else -> node.textValue()
+    // The "PODS" section lists the resolved dependencies and, nested by one level, any version constraints of their
+    // direct dependencies. That is, the nesting never goes deeper than two levels.
+    root.get("PODS").forEach { node ->
+        when (node) {
+            is ObjectNode -> {
+                val (name, version) = parseNameAndVersion(node.fieldNames().asSequence().single())
+                resolvedVersions[name] = checkNotNull(version)
+                dependencyConstraints[name] = node.single().mapTo(mutableSetOf()) {
+                    // Discard the version (which is only a constraint in this case) and just take the name.
+                    parseNameAndVersion(it.textValue()).first
+                }
+            }
+
+            else -> {
+                val (name, version) = parseNameAndVersion(node.textValue())
+                resolvedVersions[name] = checkNotNull(version)
+            }
         }
-
-        val (name, version) = parseNameAndVersion(entry)
-        resolvedVersions[name] = checkNotNull(version)
-
-        val dependencies = node[entry]?.map { depNode ->
-            val (depName, depVersion) = parseNameAndVersion(depNode.textValue())
-            depName.also { if (depVersion != null) resolvedVersions[it] = depVersion }
-        }.orEmpty()
-
-        dependencyConstraints.getOrPut(name) { mutableSetOf() } += dependencies
     }
 
     val externalSources = root.get("CHECKOUT OPTIONS")?.fields()?.asSequence()?.mapNotNull {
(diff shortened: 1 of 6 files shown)


plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -266,7 +267,12 @@ private fun parseLockfile(podfileLock: File): LockfileData {
 
         val dependencies = node[entry]?.map { depNode ->
             val (depName, depVersion) = parseNameAndVersion(depNode.textValue())
-            depName.also { if (depVersion != null) versionForName[it] = depVersion }
+            depName.also {
+                // Versions on "second level" dependencies are actually constraints; only add it if is concrete.
```
fviernau, line 271: This makes me wonder if the previous PR which introduced parsing these constraints was actually the right thing to do. Could it be that each dependency which appears on the first level under  `DEPENDENCIES` has been resolved, but the second level are always really just constraints. Implying that if there is a dependency only on the second level, there was a constraint for it, which could not be resolved?   Might providing a pointer again to the rationale for the previous change?

fviernau (changes requested):

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
fviernau: I've just read this section, and this sound like with a lockfile the version can really be pinned down exactly, See https://guides.cocoapods.org/using/pod-install-vs-update.html#using-exact-versions-in-the-podfile-is-not-enough.

I wonder what that last sentence with "properly use pod install vs. pod update." refers to. Any idea?

fviernau: It's a guess but maybe we must distinguish the following two cases.
Let's consider package `D` which is a dependency of package `A`:

1. If `D` is a subspec of `A` then use the version of `A` as the version for `D`
2. Otherwise assume / assert that there is a top-level entry for `D` with a version

Could this algorithm make sense?

sschuberth: After doing some more research, I believe it's simpler than that. The (root) spec for AFNetworking over [here](https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking.podspec) lists no dependencies at all, only subspecs have [dependencies](https://guides.cocoapods.org/syntax/podspec.html#dependency). Still, the generated lockfile has [nested entries](https://github.com/oss-review-toolkit/ort/blob/4bbd26a8a4dde135054d4ceace5c90750882a3a0/plugins/package-managers/cocoapods/src/funTest/assets/projects/synthetic/external-sources/Podfile.lock#L3-L7) under AFNetworking. This tells me that these entries are indeed not dependencies, but pure constraint, say that *if* (and only if) one of the listed entries *also* appears as a dependency, then it must be in a version matching the constraint.

So for the AFNetworking case, AFNetworking/UIKit should not be part of the ORT dependency tree at all, but e.g. AFNetworking/NSURLSession should, as it also appears at the first level.

I'll make the necessary changes.

fviernau:
What do you think about adding an explicit revert commit for the commit which introduced (over-)reporting as dependencies? So, it becomes more obvious that this approach got discarded?

fviernau (changes requested):

sschuberth:
> What do you think about adding an explicit revert commit for the commit which introduced (over-)reporting as dependencies? So, it becomes more obvious that this approach got discarded?

I believe you refer to commit 18f9be2d73a7c834b5fcb2e410fcd2693272f8bb. However, about half of that commit should stay, and its over-reporting also did not become visible until the test added in 4bbd26a8a4dde135054d4ceace5c90750882a3a0. So I don't think there's a single commit that we could cleanly revert to make the diff simpler.

fviernau:
> I believe you refer to commit [18f9be2](https://github.com/oss-review-toolkit/ort/commit/18f9be2d73a7c834b5fcb2e410fcd2693272f8bb). However, about half of that commit should stay, and its over-reporting also did not become visible until the test added in [4bbd26a](https://github.com/oss-review-toolkit/ort/commit/4bbd26a8a4dde135054d4ceace5c90750882a3a0). So I don't think there's a single commit that we could cleanly revert to make the diff simpler.

I see. Would referencing 18f9be2d73a7c834b5fcb2e410fcd2693272f8bb from the top-most commit make sense to you?

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -304,8 +304,11 @@ private fun parseLockfile(podfileLock: File): LockfileData {
             dependencies = dependenciesForName[name].orEmpty().mapTo(mutableSetOf()) { createPackageReference(it) }
         )
 
+    // The "DEPENDENCIES" section lists direct dependencies, but only along with version constraints, not with their
+    // resolved versions, and eventually additional information about the source.
     val dependencies = root.get("DEPENDENCIES").mapTo(mutableSetOf()) { node ->
-        val name = node.textValue().substringBefore(" ")
+        // Discard the version (which is only a constraint in this case) and just take the name.
```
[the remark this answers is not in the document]
sschuberth: I believe the confusion arises from a misunderstanding about what "constraint" refers to exactly here: The whole line entry, the `name`, or the `version`. I use "constraint" synonymous to "version constraint", meaning that only the version is a constraint, but not the whole entry (including its name). Take this for example:

```
DEPENDENCIES:
  - Down (~> 0.11.0)
```

`Down` is a real concrete direct dependency, just its version is "inconcrete" as only a version constraint is given. Nevertheless, it should be added as a dependency, and as part of that its version needs to be resolved, which is done inside `createPackageReference()` which looks it up in `resolvedVersions`.

The situation is different for second-level `PODS` entries, however. Here, the full line entry is a constraint, and not even just the name should be taken (with a version being resolved later) as a dependency.

sschuberth:
> Would referencing [18f9be2](https://github.com/oss-review-toolkit/ort/commit/18f9be2d73a7c834b5fcb2e410fcd2693272f8bb) from the top-most commit make sense to you?

Yes, done.

PR #7562, opened 2023-09-22
FossID changes

nnobelis:


clients/fossid-webapp/src/main/kotlin/model/identification/markedAsIdentified/Comment.kt
```diff
@@ -0,0 +1,22 @@
+package org.ossreviewtoolkit.clients.fossid.model.identification.markedAsIdentified
+
```
sschuberth, line 2: Move this below the header.

sschuberth: Commit message:

- "in newer version of FossID" should say either "in *a* newer version of FossID" or "in newer version*s* of FossID".
- Please separate the last sentence ("It was not possible ...") by an empty line to make it a real paragraph.

sschuberth (changes requested):

clients/fossid-webapp/src/main/kotlin/model/identification/markedAsIdentified/Comment.kt
nnobelis: > Move this below the header.

There is unfortunately no detekt rule for this.
Loosely related: could we enable the detekt rule `MissingPackageDeclaration` ? I forgot some package declarations in the past.

sschuberth: > There is unfortunately no detekt rule for this.

Right, but the `checkLicenseHeaders` task that we run as part of CI checks should actually detects this.

> could we enable the detekt rule `MissingPackageDeclaration` ?

As long as it doesn't interfere without [our special conventions](https://github.com/oss-review-toolkit/ort/blob/main/detekt-rules/src/main/kotlin/OrtPackageNaming.kt), I have nothing against it. Feel free to propose a PR.

sschuberth (approved):

PR #7571, opened 2023-09-25
docs: Enable Algolia search for the website

mnonnenmacher:
Configure Algolia as search engine for the website as documented at [1].

[1]: https://docusaurus.io/docs/search#connecting-algolia


website/docusaurus.config.js
```diff
@@ -146,6 +146,39 @@ const config = {
...
+        // deployments using a different baseUrl. You can use regexp or string in the `from` param. For example:
+        // localhost:3000 vs myCompany.com/docs
+        // replaceSearchResultPathname: {
+        //   from: '/docs/', // or as RegExp: /\/docs\//
+        //   to: '/',
+        // },
+
+        // Optional: Algolia search parameters
+        //searchParameters: {},
+
```
sschuberth, line 176: The descriptions read as if we'd want to drop these for the time being...

website/docusaurus.config.js
```diff
@@ -146,6 +146,39 @@ const config = {
...
+        // Optional: path for search page that enabled by default (`false` to disable it)
+        searchPagePath: 'search',
+
+        //... other Algolia params
```
sschuberth, line 180: ... same here.

website/docusaurus.config.js
mnonnenmacher: I thought it's useful to just copy the whole config template to make it easier if we need those options in future. But I can remove it if you prefer?

sschuberth: I was unclear to me that this is the *whole* template. I though you manually kept *some* options that you felt could become useful in the future.

But in either case, I guess I'd prefer to drop them, also as this might go out of sync when new options are introduced upstream.

mnonnenmacher: I have also removed some redundant comments.

sschuberth (approved):

PR #7581, opened 2023-09-26
docker: set correct version for runtime

heliocastro:
Main version is caught in the same manner as the
docker_build script, with gradlew.
Tag event will create the version with defined tag.

### DIFF
--- a/.github/workflows/docker-ort-runtime-ext.yml
+++ b/.github/workflows/docker-ort-runtime-ext.yml
@@ -15,7 +15,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
-name: Mega Docker extended image
+name: Docker extended runtime image
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


tsteenbe:
@heliocastro  I was looking once more at the scripts in ./github/workflows and for consistency with the other scripts what do you think of renaming "[docker-ort-runtime.yml](https://github.com/oss-review-toolkit/ort/blob/main/.github/workflows/docker-ort-runtime.yml)" to something like build-and-deploy-ort-container-image-runtime? And change the name from "name: Mega Docker runtime image" to "name: Build and Deploy ORT Runtime Container Image?

heliocastro:
> @heliocastro I was looking once more at the scripts in ./github/workflows and for consistency with the other scripts what do you think of renaming "[docker-ort-runtime.yml](https://github.com/oss-review-toolkit/ort/blob/main/.github/workflows/docker-ort-runtime.yml)" to something like build-and-deploy-ort-container-image-runtime? And change the name from "name: Mega Docker runtime image" to "name: Build and Deploy ORT Runtime Container Image?

On this PR i already moved the Mega name away. Renaming the workflows was really too much i think, and the only sensible change as @fviernau mentioned is the name of the PR itself.
So my vote is keep the way it is only with the name change.

tsteenbe (approved):

PR #7588, opened 2023-09-27
chore(Jenkinsfile): Set the semver4j revision to the latest tag

sschuberth:
The latest tag is more likely to have results in public scan storages like ClearlyDefined.

### DIFF
--- a/integrations/jenkins/Jenkinsfile
+++ b/integrations/jenkins/Jenkinsfile
@@ -57,7 +57,7 @@ pipeline {
         string(
             name: 'PROJECT_VCS_REVISION',
             description: 'Optional VCS revision of the project (prefix Git tags with "refs/tags/").',
-            defaultValue: ''
+            defaultValue: 'refs/tags/v3.1.0'
         )
 
         credentials(


integrations/jenkins/Jenkinsfile
```diff
@@ -57,7 +57,7 @@ pipeline {
         string(
             name: 'PROJECT_VCS_REVISION',
             description: 'Optional VCS revision of the project (prefix Git tags with "refs/tags/").',
-            defaultValue: ''
+            defaultValue: 'refs/tags/v3.1.0'
```
fviernau, line 60: The value doesn't make sense if the user provided a value for `PROJECT_VCS_URL` different from its default value 'https://github.com/vdurmont/semver4j.git'. 

Have you considered that? ...does it break existing jobs which only provided `PROJECT_VCS_URL` ?

fviernau (changes requested):

integrations/jenkins/Jenkinsfile
sschuberth: > Have you considered that?

Yes, but I though it's not an issue as the coupling between the `PROJECT_VCS_*` parameters should be quite obvious.

> does it break existing jobs which only provided `PROJECT_VCS_URL`?

I'm not aware of any existing jobs that make direct use of this file, so I don't know. In any case, if the `PROJECT_VCS_URL` would be set programmatically, I'd expect the other `PROJECT_VCS_*` to be set, too.

fviernau: In my view the reason (to have more results from clearly defined) is not important enough to satisfy this default value for all scans. I'm not blocking this however, leaving the approval up to someone else.

sschuberth: Ok, let me just drop this then. It was mostly for convenience for anyone demoing ORT with the `Jenkinsfile`.

PR #7589, opened 2023-09-27
Improve scan storage logging

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/ScanResultsStorage.kt
+++ b/scanner/src/main/kotlin/ScanResultsStorage.kt
@@ -154,27 +154,30 @@ abstract class ScanResultsStorage : PackageBasedScanStorage {
             if (results.isEmpty()) {
                 results
             } else {
-                val scanResults = results.toMutableList()
+                val (matchingProvenance, nonMatchingProvenance) = results.partition { it.provenance.matches(pkg) }
 
-                // Only keep scan results whose provenance information matches the package information.
-                scanResults.retainAll { it.provenance.matches(pkg) }
-                if (scanResults.isEmpty()) {
+                if (matchingProvenance.isEmpty()) {
                     logger.debug {
                         "No stored scan results found for $pkg. The following entries with non-matching provenance " +
-                            "have been ignored: ${scanResults.map { it.provenance }}"
+                            "have been ignored: ${nonMatchingProvenance.map { it.provenance }}"
                     }
+
+                    matchingProvenance
                 } else {
-                    // Only keep scan results from compatible scanners.
-                    scanResults.retainAll { scannerCriteria.matches(it.scanner) }
-                    if (scanResults.isEmpty()) {
+                    val (matchingCriteria, nonMatchingCriteria) = matchingProvenance.partition {
+                        scannerCriteria.matches(it.scanner)
+                    }
+
+                    if (matchingCriteria.isEmpty()) {
                         logger.debug {
-                            "No stored scan results found for $scannerCriteria. The following entries with " +
-                                "incompatible scanners have been ignored: ${scanResults.map { it.scanner }}"
+                            "No stored scan results for '${pkg.id.toCoordinates()}' match $scannerCriteria. The " +
+                                "following entries with non-matching criteria have been ignored: " +
+                                nonMatchingCriteria.map { it.scanner }
                         }
                     }
-                }
 
-                scanResults
+                    matchingCriteria
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/ScanResultsStorage.kt
```diff
@@ -108,7 +108,8 @@ abstract class ScanResultsStorage : PackageBasedScanStorage {
 
         result.onSuccess { results ->
             logger.info {
-                "Read ${results.values.sumOf { it.size }} scan result(s) from ${javaClass.simpleName} in $duration."
+                val count = results.values.sumOf { it.size }
+                "Read $count matching scan result(s) from ${javaClass.simpleName} in $duration."
```
fviernau, line 112: While at it, would it make sense to say "Read all..."? Or maybe even "Successfully read all..."

scanner/src/main/kotlin/ScanResultsStorage.kt
```diff
@@ -154,27 +154,30 @@ abstract class ScanResultsStorage : PackageBasedScanStorage {
...
-                scanResults.retainAll { it.provenance.matches(pkg) }
-                if (scanResults.isEmpty()) {
+                if (matchingProvenance.isEmpty()) {
                     logger.debug {
                         "No stored scan results found for $pkg. The following entries with non-matching provenance " +
-                            "have been ignored: ${scanResults.map { it.provenance }}"
+                            "have been ignored: ${nonMatchingProvenance.map { it.provenance }}"
                     }
+
+                    matchingProvenance
```
fviernau, line 165: Can the code now return result entries which do match provenance, but do not match criteria in line 165?

scanner/src/main/kotlin/ScanResultsStorage.kt
```diff
@@ -154,27 +154,30 @@ abstract class ScanResultsStorage : PackageBasedScanStorage {
             if (results.isEmpty()) {
                 results
             } else {
-                val scanResults = results.toMutableList()
+                val (matchingProvenance, nonMatchingProvenance) = results.partition { it.provenance.matches(pkg) }
 
-                // Only keep scan results whose provenance information matches the package information.
-                scanResults.retainAll { it.provenance.matches(pkg) }
-                if (scanResults.isEmpty()) {
+                if (matchingProvenance.isEmpty()) {
```
fviernau, line 159: If the log output only makes sense if `nonMatchingProvenance` is not empty, this could be more directly expressed by just `nonMatchingProvenance.isNotEmpty()` in line 159.

fviernau (changes requested):

scanner/src/main/kotlin/ScanResultsStorage.kt
sschuberth: What are you trying to say with "all"?

scanner/src/main/kotlin/ScanResultsStorage.kt
sschuberth: No, as this is inside the `matchingProvenance.isEmpty()` block.

scanner/src/main/kotlin/ScanResultsStorage.kt
sschuberth: I prefer to keep is as-is, because this better expresses that we want to log if there is no matching provenance (and it avoids a double negation). The fact that the logging shows the non-matches is just a detail.

scanner/src/main/kotlin/ScanResultsStorage.kt
fviernau: That all matching results have been read, not just a subset of the matching results.

sschuberth: I think that adds more confusion than clarity, because we actually never deal with "a subset of the matching results". We have "all results" and "matching results" (which is a subset of "all results"), but not "all matching results" and thus also not "a subset of the / all matching results".

fviernau: That's something the reader of the logs doesn't know, because this is only in the code. Anyhow, not worth discussing any further I guess.

scanner/src/main/kotlin/ScanResultsStorage.kt
fviernau: I believe the current if condition makes it trickier to understand, if it's code block may get executed in case `nonMatchingProvenance` is actually empty. This is less obvious, which is why I think we should change the condition.

PR #7590, opened 2023-09-27
fix(ClearlyDefinedStorage): Properly parse returned VCS URLs

sschuberth:
ClearlyDefined returns VCS URLs like [1] which makes the provenance matcher fail as it checks for [2] (plus the revision independently). Properly parse the VCS URL returned by ClearlyDefined to fix this.

[1]: https://github.com/vdurmont/semver4j/tree/88912638db3f6112a2b345f1638ced33a0a606e1
[2]: https://github.com/vdurmont/semver4j.git

### DIFF
--- a/scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
+++ b/scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
@@ -54,7 +54,7 @@ class ClearlyDefinedStorageFunTest : StringSpec({
                     provenance = RepositoryProvenance(
                         vcsInfo = VcsInfo(
                             type = VcsType.GIT,
-                            url = "https://github.com/vdurmont/semver4j/tree/88912638db3f6112a2b345f1638ced33a0a606e1",
+                            url = "https://github.com/vdurmont/semver4j.git",
                             revision = "88912638db3f6112a2b345f1638ced33a0a606e1"
                         ),
                         resolvedRevision = "88912638db3f6112a2b345f1638ced33a0a606e1"
@@ -108,7 +108,7 @@ class ClearlyDefinedStorageFunTest : StringSpec({
                     provenance = RepositoryProvenance(
                         vcsInfo = VcsInfo(
                             type = VcsType.GIT,
-                            url = "https://github.com/sksamuel/hoplite/tree/b3bf5d7bd3814cb7576091acfecd097cb3a79e72",
+                            url = "https://github.com/sksamuel/hoplite.git",
                             revision = "b3bf5d7bd3814cb7576091acfecd097cb3a79e72"
                         ),
                         resolvedRevision = "b3bf5d7bd3814cb7576091acfecd097cb3a79e72"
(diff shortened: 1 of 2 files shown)


scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
```diff
@@ -54,7 +54,7 @@ class ClearlyDefinedStorageFunTest : StringSpec({
                     provenance = RepositoryProvenance(
                         vcsInfo = VcsInfo(
                             type = VcsType.GIT,
-                            url = "https://github.com/vdurmont/semver4j/tree/88912638db3f6112a2b345f1638ced33a0a606e1",
+                            url = "https://github.com/vdurmont/semver4j.git",
```
fviernau, line 57: Commit: Does clearly defined always return such URLs or only in some cases? ...in case of the latter, should we file an upstream issue?

fviernau: Would be nice to have some reference if available...

fviernau: Sorry, one more question: Previously, revision and resolved revision were guaranteed to be identical by this code. By the new code that guarantee is removed. Should we recover that by adjusting the new code slightly?

sschuberth: I don't believe restoring that guarantee is desirable; I rather regard it as a bug that this happened to be the case. And it does not matter in the context of scan storages, as provenance is always matched by resolved revision. So I prefer so allow revision and resolved revision to be different strings, as that can provide additional information about a symbolic name for the resolved revision which might be helpful.

fviernau: But isn't that additional information irrelevant? Because, for clearly defined it's undefined if this additional information has been created when the scan was executed or whether it's solely based on the request.
In the former case it does not matter, in the latter that information is present in the analyzer result.

fviernau (approved):

scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
sschuberth: > But isn't that additional information irrelevant?

It depends. It could be interesting during debugging to better understand if something strange happened, like a tag being recreated to point at a different commit.

Anyway, that's hypothetical, but generally I believe having more information at hand is better than having less information.

fviernau: Ok, makes sense. Only the resolved revision get's propagated by the scanner anyway. So the revision remains inside the scanner. See https://github.com/oss-review-toolkit/ort/blob/0d4b1f75932d3a6666d599dbf6359f7e8fcebdff/scanner/src/main/kotlin/Scanner.kt#L233.

PR #7592, opened 2023-09-27
chore(SW360): Use Maven Central and update version

heliocastro:
SW360 latest version artifacts now are published in Maven Central.


sschuberth (approved):

heliocastro:
> > SW360 latest version artifacts now are published in Maven Central.
> 
> Can we rely on this to also be the case going forward / for upcoming releases?

Yes, i'm responsible for this and i'm already prepare the needed changes for upstream as well. Next release will happens on end of the year probably.

PR #7595, opened 2023-09-28
fix(docker): Re-align the Poetry version

fviernau:
`Dockerfile` and `Dockerfile-legacy` should use the same version. So, align on the latest one.


sschuberth (approved):
Note that this conflicts with https://github.com/oss-review-toolkit/ort/pull/7533, but this change is better.

PR #7596, opened 2023-09-28
chore(docker): Adjust build frequency

heliocastro:
depends on:  https://github.com/oss-review-toolkit/ort/pull/7581

Intermediary Docker images will be built nightly at 04:00 and be tagged as <VERSION>-SNAPSHOT.
Same to workflow_dispatch events.

Push tag events will tag Docker images to the release version.

### DIFF
--- a/.github/workflows/docker-ort-runtime-ext.yml
+++ b/.github/workflows/docker-ort-runtime-ext.yml
@@ -19,14 +19,14 @@ name: Mega Docker extended image
 
 on:
   workflow_dispatch:
+  schedule:
+    - cron: '0 4 * * *'
   pull_request:
     paths:
       - '.versions'
       - 'Dockerfile'
       - 'Dockerfile-extended'
   push:
-    branches:
-      - main
     tags:
       - '*'
   workflow_run:
--- a/.github/workflows/docker-ort-runtime.yml
+++ b/.github/workflows/docker-ort-runtime.yml
@@ -19,13 +19,13 @@ name: Mega Docker runtime image
 
 on:
   workflow_dispatch:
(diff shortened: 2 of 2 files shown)


.github/workflows/docker-ort-runtime-ext.yml
```diff
@@ -15,23 +15,25 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
-name: Mega Docker extended image
+name: Docker extended image
```
sschuberth, line 18: This does more than just changing the schedule, which does not look right.

.github/workflows/docker-ort-runtime-ext.yml
```diff
@@ -19,14 +19,14 @@ name: Mega Docker extended image
...
+    - cron: '0 4 * * *'
   pull_request:
     paths:
       - '.versions'
       - 'Dockerfile'
       - 'Dockerfile-extended'
   push:
-    branches:
-      - main
     tags:
```
fviernau, line 30: Why can the main branch be removed? (Guess hotfix releases?)

heliocastro: Now push any tag. Tags can be in any branch you made it.
Before this push when tagged and when is pushed to main, not two at same time. They are individual entities.

.github/workflows/docker-ort-runtime-ext.yml
heliocastro: Solved

fviernau (approved):

tsteenbe (approved):

PR #7598, opened 2023-09-28
feat(package-managers/python)!: Support Python 3.11

fviernau:
Add "3.11" to the list of python versions which ORT allows to pass to
the `-p` option of `python-inspector`. This enables the successful
analysis of Python projects which require Python 3.11 via both, `Pip`
and `Poetry`. In particular, it does not seem to require upgrading
the Python installation to version 3.11.

The change is breaking, because also the default Python version is
changed to 3.11. For analyzing Python projects targeting 3.10 the
Python version as of now must be explicitly specified.

Note: The reason why the binary artifact URL of `MarkupSafe` became empty
in the expected result for the test is simply, because there is no
"Built Distribution", but only a "Source Distribution", see [1].
Furthermore, it's still correct to list `MarkupSafe` as dependency,
because it specifies `>= 3.6` as Python version constraint, mich
matches version 3.11.

[1] https://pypi.org/project/MarkupSafe/2.0.1/#files

Fixes #7476.

--------

Note: This PR conflicts with [1], I didn't intentionally repeat the work from [1], but have been debugging analyzing `pyproject.toml` without being aware [1] is related, and this is the result from it. So, I believe this PR minimizes the changes for supporting 3.11. 

[1] https://github.com/oss-review-toolkit/ort/pull/7533

PR #7599, opened 2023-09-28
chore(docker): Upgrade Python to the latest version

fviernau:
This prepares for supporting Python version 3.11.

Note: This is done in a separate (from supporting python 3.11 with Pip) PR in order to show that the existing `PoetryFunTest` does not break for the currently used Python version 3.10, when `poetry` is run under version 3.11.


sschuberth (approved):

fviernau:
I'm not sure if this is needed for supporting Python 3.11 with `Pip` / `Poetry` (https://github.com/oss-review-toolkit/ort/pull/7598). I cannot see any indication of need from the tests of that PR. Should we still upgrade Python now, or defer it a bit?

sschuberth:
> Should we still upgrade Python now, or defer it a bit?

I don't have an opinion here as I rarely analyze Python projects. What about @oss-review-toolkit/core-devs?

heliocastro:
Why not give proper instructions how to install multiple pythons with custom docker and done ?
This can be done ion documentation. We already have cases on TNG where i was asked how to have two pythons inside the env.
There's honestly nothing forcing to make 3.11.x default unless we want to have headaches, but in other side, 3.12.x is already banging the door.

sschuberth:
> Why not give proper instructions how to install multiple pythons with custom docker and done ?

To me, bumping the default Python version here simple severs as a proof that our functional tests still pass with that version. It's not exclusively / primarily for convenience of our users.

fviernau:
If one looks at EOL chart, we could establish ORT policy to always support the latest Python version in active support.
In the 3.11 case this will make the analysis work for 3.10 in most of the cases. In cases where it doesn't work (in particular for < 3.10 version, the user has to create a contain with also these Python versions installed to make it work).

So, I vote for a merge.

[1] https://endoflife.date/python

PR #7601, opened 2023-09-29
feat(PurlUtils): Add optional parameters to `toPurl`

nnobelis:
The commit https://github.com/bosch-io/oss-review-toolkit/commit/7da5ae93606390b4967755121a8932826931098a changed the visibility of `createPurl`, `toPurl` has
become the only public ORT function for plugins to generate Purls.
Since this function does not support qualifiers nor subpath, these two
parameters have to be added to the function signature with default values.

### DIFF
--- a/model/src/main/kotlin/utils/PurlUtils.kt
+++ b/model/src/main/kotlin/utils/PurlUtils.kt
@@ -86,12 +86,18 @@ fun Identifier.getPurlType() =
  * Create the canonical [package URL](https://github.com/package-url/purl-spec) ("purl") based on the properties of
  * the [Identifier]. Some issues remain with this specification
  * (see e.g. https://github.com/package-url/purl-spec/issues/33).
+ * Optional [qualifiers] may be given and will be appended to the purl as query parameters e.g.
+ * pkg:deb/debian/curl@7.50.3-1?arch=i386&distro=jessie
+ * Optional [subpath] may be given and will be appended to the purl e.g.
+ * pkg:golang/google.golang.org/genproto#googleapis/api/annotations
  *
  * This implementation uses the package type as 'type' purl element as it is used
  * [in the documentation](https://github.com/package-url/purl-spec/blob/master/README.rst#purl).
  * E.g. 'maven' for Gradle projects.
  */
-fun Identifier.toPurl() = if (this == Identifier.EMPTY) "" else createPurl(getPurlType(), namespace, name, version)
+@JvmOverloads
+fun Identifier.toPurl(qualifiers: Map<String, String> = emptyMap(), subpath: String = "") =
+    if (this == Identifier.EMPTY) "" else createPurl(getPurlType(), namespace, name, version, qualifiers, subpath)
 
 /**
  * Create the canonical [package URL](https://github.com/package-url/purl-spec) ("purl") based on given properties:


fviernau:
@sschuberth , mind sharing the rationale for 7da5ae93606390b4967755121a8932826931098a ?

sschuberth:
> @sschuberth , mind sharing the rationale for [7da5ae9](https://github.com/oss-review-toolkit/ort/commit/7da5ae93606390b4967755121a8932826931098a) ?

It's the same rational that you follow in similar commits like 38eacac94ab9d35411bbb7898965c80d5d5d9695. 0d75760adacbe336bfda7e38d71256c7a87931c5 and others, to reduce visibility of functions that are only supposed to be used internally, to not make too much promises regarding the public API.

fviernau:
> It's the same rational that you follow in similar commits like [38eacac](https://github.com/oss-review-toolkit/ort/commit/38eacac94ab9d35411bbb7898965c80d5d5d9695). [0d75760](https://github.com/oss-review-toolkit/ort/commit/0d75760adacbe336bfda7e38d71256c7a87931c5) and others, to reduce visibility of functions that are only supposed to be used internally, to not make too much promises regarding the public API.

I've been trying to include you into the discussion to figure out whether the rationale for 7da5ae93606 does still apply in context of @nnobelis change. The commits of mine are irrelevant in this context. In order to review this I would have like to have known what you thought.

model/src/main/kotlin/utils/PurlUtils.kt
```diff
@@ -91,7 +101,7 @@ fun Identifier.getPurlType() =
  * [in the documentation](https://github.com/package-url/purl-spec/blob/master/README.rst#purl).
  * E.g. 'maven' for Gradle projects.
  */
-fun Identifier.toPurl() = if (this == Identifier.EMPTY) "" else createPurl(getPurlType(), namespace, name, version)
+fun Identifier.toPurl() = toPurl(emptyMap(), "")
```
fviernau, line 104: Would it make sense to, instead of an overload, just add these two parameter to this function, with default values `emptyMap` and `""` ?

fviernau (changes requested):

model/src/main/kotlin/utils/PurlUtils.kt
sschuberth: That's how it was before [the force-push](https://github.com/oss-review-toolkit/ort/compare/ec7acb0377d80507a25bb5fbc199ad96e7336948..0f484d684dfc91c1c951c7ea27411680040cead1), and I also liked that better... @nnobelis why did you change it?

nnobelis: As explained in the commit message it broke the FreeMarker templates that call `toUrl` (the tests failed): I guess Kotlin's implementation of extension functions with default values is not compatible with FreeMarker Java's reflection.

fviernau: Have you considered annotation the function (from before) with `@JvmOverloads`? I might solve the problem as well.

nnobelis: I think we can go around with `@JvmOverloads`. Would it be ok to put this annotation on `toUrl` ?

fviernau: Where does `toUrl` now come from? Did you maybe mean this?

```
@JvmOverloads
fun Identifier.toPurl(qualifiers: Map<String, String> = emptyMap(), subpath: String = "") = ...
```

nnobelis: Yes `toPurl`. Sorry typo.

fviernau: If above signature works, IMO it'd bee good.

fviernau (approved):

sschuberth:
@nnobelis I guess this can be moved out of draft state?

model/src/main/kotlin/utils/PurlUtils.kt
sschuberth: @nnobelis, please also explain in the commit message what `@JvmOverloads` is needed for.

nnobelis:
@fviernau @sschuberth can we please merge this ?

fviernau (approved):

PR #7617, opened 2023-10-04
fix(docker): Enable build on schedule event

heliocastro:
After move to scheduled build, main images are building only on push and workflow_dispatch events.
Noticed after some mistake PR done on Ort Github CI repository.

### DIFF
--- a/.github/workflows/docker-ort-runtime-ext.yml
+++ b/.github/workflows/docker-ort-runtime-ext.yml
@@ -19,18 +19,6 @@ name: Docker extended runtime image
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: '0 4 * * *'
-  pull_request:
-    paths:
-      - '.versions'
-      - 'Dockerfile'
-      - 'Dockerfile-extended'
-      - '.github/workflows/docker-ort-runtime.yml'
-      - '.github/workflows/docker-ort-runtime-ext.yml'
-  push:
-    tags:
-      - '*'
   workflow_run:
     workflows:
       - 'Docker runtime image'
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-ort-runtime-ext.yml
```diff
@@ -20,7 +20,7 @@ name: Docker extended runtime image
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 4 * * *'
+    - cron: '0 5 * * *'
```
fviernau, line 23: This change IMO should be a separate commit. e.g. "align job dispatch time of ort-runtime-ext and ort-runtime"

fviernau (changes requested):

.github/workflows/docker-ort-runtime-ext.yml
fviernau: For the commit which re-enables it, IMO it would be nice to have a reference to the commit which accidentally disabled it.

heliocastro: @fviernau We never disabled. Actually, we moved from push events to scheduled events.
And after we moved, the schedule event were never handled.

heliocastro: @fviernau Splitted in two commits and commented about previous commit. 
The next window to build will be 4 AM and 5 AM.

tsteenbe (approved):

fviernau (approved):

heliocastro:
@fviernau @tsteenbe @sschuberth Anyone can push this one before tomorrow 4AM schedule ?

heliocastro:
@fviernau The pr is clean, can we go with approval now ?

tsteenbe (approved):

PR #7620, opened 2023-10-04
fix(PurlUtils): Add missing Debian Purl type to `getPurlType`

nnobelis:


model/src/main/kotlin/utils/PurlUtils.kt
```diff
@@ -65,6 +65,7 @@ enum class PurlType(private val value: String) {
 fun Identifier.getPurlType() =
     when (type.lowercase()) {
         "bower" -> PurlType.BOWER
+        "deb" -> PurlType.DEBIAN
```
fviernau, line 68: Can you please indicate _why_ you think this value is missing? (I believe so far this mapping was limited keys which are actually used by package managers within ORT.

fviernau (changes requested):

model/src/main/kotlin/utils/PurlUtils.kt
nnobelis: At Bosch, we have a package manager plugin for Debian.
It used previously `createPurl` and I plan to migrate to `toPurl` following the merge of https://github.com/oss-review-toolkit/ort/pull/7601.
`toPurl` calls this `getPurlType` function.

You wrote that `getPurlType` is meant to contains only package managers supported by ORT. In this case, we have a proprietary package manager  plugin. How do you see it coexist with the rule you described in your comment ?

fviernau: > You wrote that getPurlType is meant to

Actually, I didn't mean "is meant to" but I meant: it currently only contains values from the package managers in ORT.

> How do you see it coexist with the rule you described in your comment ?

ORT has a mechanism to plug-in package managers, but lacks a mechanism for plugging in logic related to the ecosystem that package manager uses, e.g. `Identifier.type`. In my view, that should be a plug-in as well for the exact same reason the package manager has been chosen to be a plugin. However, we do not have such mechanism yet.

There are further places in the code which would need to be considered.
For example, OSV has a mapping from `Identifier.type` to OSV ecosystem constants. So, currently OSV cannot be used for plugin package managers which would require extension of that mapping. 

In my view, it would make sense to define start a dedicated `PackageManagagerPlugin` interface, which would contain (very roughly spkeaking):

1. The existing factory method, currently called `PackageManagerFactory.create()`
2. A function which gives the purlType to the corresponding type (optional to implement)
3. A function which gives the OSV ecosystem to the corresponding type

What do other guys think? @oss-review-toolkit/core-devs

sschuberth: > However, we do not have such mechanism yet.

Exactly. We had discussed that in past, and I have several branches on roughly that topic lying around (`identifier-types`, `multiple-identifier-classes`) that also try to deal with the problem that a package manager's *project* type does not need to match with it *package* type(s).

So for package manager that support multiple package types, like `Pub`, we'd even need a *set* of package types / PURLs / OSV ecosystem coordinates.

fviernau: @sschuberth not sure if I get your point. Do you believe it's fine to add mappings for non-existing types now?

sschuberth: > not sure if I get your point.

I was just amending to your API proposal more cases that we'd need to cover with such a new API.

> Do you believe it's fine to add mappings for non-existing types now?

Basically no, but the right way to do it would require a rather large refactoring, I believe. Not sure whether that's feasible to do (now). I guess we should discuss this in a Kotlin developer meeting.

nnobelis: My 2 cents: for the time being I found a workaround so we can close this PR.
However this topic should definitively be discussed as it will surely arise again.

nnobelis:
Change not needed anymore.

PR #7622, opened 2023-10-04
chore(docker): Use the output of `printSemVersion` as `ORT_VERSION`

fviernau:
Previously, the value looked like 3.0.0-SNAPSHOT. Change it to the form like `3.0.0-389da0c-SNAPSHOT`, so that it is possible to figure out from which sources ORT has been built.

### DIFF
--- a/.github/workflows/docker-ort-runtime-ext.yml
+++ b/.github/workflows/docker-ort-runtime-ext.yml
@@ -169,7 +169,7 @@ jobs:
 
       - name: Get ORT current version
         run: |
-          ORT_VERSION=$(./gradlew -q properties --property version | sed -nr "s/version: (.+)/\1/p")
+          ORT_VERSION=$(./gradlew -q printSemVersion)
           echo "ORT_VERSION=${ORT_VERSION}" >> $GITHUB_ENV
 
       - name: Set up Docker build
--- a/.github/workflows/docker-ort-runtime.yml
+++ b/.github/workflows/docker-ort-runtime.yml
@@ -182,7 +182,7 @@ jobs:
 
       - name: Get ORT current version
         run: |
-          ORT_VERSION=$(./gradlew -q properties --property version | sed -nr "s/version: (.+)/\1/p")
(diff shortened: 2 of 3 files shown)


.github/workflows/docker-ort-runtime-ext.yml
```diff
@@ -169,7 +169,7 @@ jobs:
 
       - name: Get ORT current version
         run: |
-          ORT_VERSION=$(./gradlew -q properties --property version | sed -nr "s/version: (.+)/\1/p")
+          ORT_VERSION=$(./gradlew -q printSemVersion)
```
sschuberth, line 172: Note that when being exactly on tag `2.0.0` (with a clean working tree), this would print `2.0.0+sha.5893bd7` instead of `2.0.0`. Is this desired?

fviernau: I wasn't aware of that, but I think it's fine.

sschuberth: That would be a mismatch, however, for people running ORT also from the distribution archive, which shows `2.0.0` in this case. I'll look into [customizing the version format](https://github.com/jmongard/Git.SemVersioning.Gradle?tab=readme-ov-file#version-format) to address that.

fviernau: Shall we merge now this and then customize the version format in that gradle tasks later on?

sschuberth: My alternative proposal is already [here](https://github.com/oss-review-toolkit/ort/pull/7625=. That could be used instead of this PR (not in addition to it). The advantage is that also non-Docker user would get the SHA1 info for pre-release builds.

heliocastro: I can minimize it direct in the tagging, i can capture the sha without need to alter the current Gradle process.
But will solve only for the images in upstream, not the local build script/.

fviernau:
Close in favor of #7628 .

Issue #2896, opened 2020-07-31
Trouble generating report from scan-results


OctagonHex:
Hello,
I sucessfully created a scan result with the following command. (I used the scancode-toolkit examples.)
`cli\build\install\ort\bin\ort scan -p "C:\scancode-toolkit-3.1.1\samples" -o myOut`
Output:
```
Using scanner 'ScanCode' with storage 'FileBasedStorage with XZCompressedLocalFileStorage backend'.
Local file storage has 0 scan results files.
Writing scan result to 'myOut\scan-result.yml'.
```
If I look at the .yml file, it looks good and contains many licensed.

Now I try to generate any kind of report. My goal is to generate an attribution notice. So I run the command:
`cli\build\install\ort\bin\ort report -i myOut\scan-result.yml -o myOutReport -f Excel`
but the output shows the error
```
Creating the 'Excel' report...
15:15:23.583 [main] ERROR org.ossreviewtoolkit.commands.ReporterCommand - Could not create 'Excel' report: IllegalArgumentException: The provided ORT result does not contain an analyzer result.
Failed to create any report.
```
For -f NoticeSummary, or -f NoticeByPackage OSS-RT seems to work at first glance:
```
Creating the 'NoticeSummary' report...
Successfully created the 'NoticeSummary' report at [myOutReport\NOTICE_SUMMARY] in 0.012422699s.
Successfully created 1 of 1 report(s).
```

But despite the many licenses in the .yaml,  the resulting report is empty, i.e. it says:
`This project neither contains or depends on any third-party software components.`

What is the problem, or how can this be fixed?

sschuberth:
While running a reporter on an ORT result with only a scan result is not forbidden, this is a use-case that is not well tested. The usual (and well tested) workflow is to first create an ORT result with an analyzer result, and then use that as the input for the scanner, which creates another ORT result file that combines the analyzer and scan results. Such "rich" ORT result files should work fine to create reports.

OctagonHex:
Maybe you can give me a hint on how to accomplish my goal.
For example: I try to anaylze a unstructured directory of source code. I'll use the samples from ScanCode-toolkit.
I first analyze them, and the analyzer runs OK.
As expected, the result is very short and does not contain any dependencies.
[...]

PR #7624, opened 2023-10-04
refactor(poetry): Improve the definition file paths

fviernau:
In order to do it's job, `Poetry.kt` generates a requirements file and then relies on `Pip.kt` to do the analysis. The resulting `ProjectAnalyzerResult`s uses the paths to these generated requirements files as definiton file paths. This does not make sense, because:

1. The generation of the files is an implementation detail.
2. The generated files do not exist in the code repository which can lead to confusion, in particular also these get referenced from ORT configuration / path excludes.

Fix that by patching up the `ProjectAnalyzerResult`s to recover the original definition file name.

### DIFF
--- a/plugins/package-managers/python/src/main/kotlin/Poetry.kt
+++ b/plugins/package-managers/python/src/main/kotlin/Poetry.kt
@@ -25,6 +25,7 @@ import org.apache.logging.log4j.kotlin.Logging
 
 import org.ossreviewtoolkit.analyzer.AbstractPackageManagerFactory
 import org.ossreviewtoolkit.analyzer.PackageManager
+import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.ProjectAnalyzerResult
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
 import org.ossreviewtoolkit.model.config.RepositoryConfiguration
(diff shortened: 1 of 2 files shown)


plugins/package-managers/python/src/main/kotlin/Poetry.kt
```diff
@@ -72,6 +73,13 @@ class Poetry(
 
         return Pip(managerName, analysisRoot, analyzerConfig, repoConfig)
             .resolveDependencies(requirementsFile, labels)
+            .map { projectAnalyzerResult ->
+                projectAnalyzerResult.copy(
+                    project = projectAnalyzerResult.project.copy(
+                        definitionFilePath = VersionControlSystem.getPathInfo(definitionFile).path
```
sschuberth, line 79: I propose `definitionFile.relativeTo(analysisRoot)` instead to avoid adding more code that relies on the project to be version controlled.

fviernau: This does not introduce the requirement, because code which computes the value for `definitionFilePath` already relies on that prior to this change. Furthermore, `relativeTo(analysisRoot)` seems wrong because the definition file path is not meant to be relative to the analysis root, but the repository root.

sschuberth: > This does not introduce the requirement

And I didn't say that. I was talking about adding *more* code with that requirement.

> Furthermore, `relativeTo(analysisRoot)` seems wrong because the definition file path is not meant to be relative to the analysis root, but the repository root.

Indeed `Project.definitionFilePath` is documented like that 😞 IMO that's a mistake already, and it will make our lives harder in order to address https://github.com/oss-review-toolkit/ort/issues/2896.

Anyway, we already have a bunch of code like that, so we can bulk-address this later, I guess.

sschuberth (approved):

plugins/package-managers/python/src/main/kotlin/Poetry.kt
fviernau: > And I didn't say that. I was talking about adding _more_ code with that requirement.

I know, I didn't mean to say that you said that. I only intended to clarify that both, the old and the new code path depend on that same logic.

PR #7625, opened 2023-10-04
build: If on a pre-release, use the SemVer with SHA1 metadata

sschuberth:


build.gradle.kts
```diff
@@ -43,7 +43,7 @@ semver {
 
 // Only override a default version (which usually is "unspecified"), but not a custom version.
 if (version == Project.DEFAULT_VERSION) {
-    version = semver.version
+    version = semver.semVersion.takeIf { it.isPreRelease } ?: semver.version
```
fviernau, line 46: Does `it.isPreRelease` return true for any `semver` value which is not a "release version" ?

sschuberth: I'm not sure what you definition of a "release version" is, so I'll answer differently: `it.isPreRelease` returns `true` for any SemVer that contains the `-` separated part of the version (not the `+` separated part). You might want to have a look [here](https://github.com/jmongard/Git.SemVersioning.Gradle/blob/ac9b08ef51460d5b3cab126856cf0923676fe76e/src/test/kotlin/git/semver/plugin/semver/SemVersionTest.kt#L22-L53) to get the idea.

fviernau: Ok, so let me re-phrase the question: Does the new code guarantee that for any version which is not tagged with a (release) Git tag, the version string contains the SHA1?

sschuberth: Yes.

fviernau (approved):

fviernau (approved):

PR #7628, opened 2023-10-04
refactor(Poetry): Improve the IDs of projects

fviernau:
Do not use the filename of the generated requirements file as part of the identifier for a rationale analog to [1]. Furthermore, use include the definition file path relative to the analyzer root to ensure that the project ID is unique within the ORT file.

While at it, stop re-using any logic for deriving the identifier from `Pip`, because some parts of the logic do not make sense to use. For example, it does not make sense to derive a name and a version from a sibling setup.py, because that would be unrelated to the poetry lockfile being analyzed.

Note: This change could be regarded as a left-over from [1].

[1] d647cc7ffc1a83c607c46055c7fe8882ec88b16c

### DIFF
--- a/plugins/package-managers/python/src/funTest/assets/projects/synthetic/poetry-expected-output.yml
+++ b/plugins/package-managers/python/src/funTest/assets/projects/synthetic/poetry-expected-output.yml
@@ -1,6 +1,6 @@
 ---
 project:
-  id: "Poetry::poetry-from-poetry:<REPLACE_REVISION>"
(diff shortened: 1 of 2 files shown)


heliocastro (approved):
Looks good to me.
Small question, what will happens if project not have a VersionControlSystem ?

fviernau:
> Looks good to me.
> Small question, what will happens if project not have a VersionControlSystem ?

That's not supported currently. But, the code path used before this change also had VersionControlSystem related logic in it.

PR #7632, opened 2023-10-05
Poetry: Also analyze the dev dependencies

### LINKED ISSUE #7626: Poetry analysis results do not include any dev dependencies
The implementation relies on `poetry export`.
However, it does not pass `--dev` to make the export include development dependencies.

fviernau:
Extend the implementation to also report dev dependencies.

Fixes #7626.

Note: This is a best effort implementation which intentionally does not refactor any python inspector related code, 
           because the implementation in the long run should probably be anyway changed to support the lockfile format / 
           pyproject.toml directly.

### DIFF
--- a/plugins/package-managers/python/src/funTest/assets/projects/synthetic/poetry-expected-output.yml
+++ b/plugins/package-managers/python/src/funTest/assets/projects/synthetic/poetry-expected-output.yml
@@ -16,6 +16,39 @@ project:
     path: "<REPLACE_PATH>"
   homepage_url: ""
   scopes:
+  - name: "dev"
+    dependencies:
+    - id: "PyPI::black:19.10b0"
+      dependencies:
+      - id: "PyPI::appdirs:1.4.4"
+      - id: "PyPI::attrs:22.2.0"
+      - id: "PyPI::click:8.0.4"
+      - id: "PyPI::pathspec:0.9.0"
+      - id: "PyPI::regex:2023.8.8"
+      - id: "PyPI::toml:0.10.2"
+      - id: "PyPI::typed-ast:1.5.5"
+    - id: "PyPI::pylint:2.12.0"
+      dependencies:
+      - id: "PyPI::astroid:2.9.0"
+        dependencies:
+        - id: "PyPI::lazy-object-proxy:1.7.1"
+        - id: "PyPI::setuptools:59.6.0"
+        - id: "PyPI::wrapt:1.13.3"
+      - id: "PyPI::isort:4.3.21"
+      - id: "PyPI::mccabe:0.6.1"
+      - id: "PyPI::platformdirs:2.4.0"
+      - id: "PyPI::toml:0.10.2"
+    - id: "PyPI::pytest:6.2.5"
+      dependencies:
+      - id: "PyPI::attrs:22.2.0"
+      - id: "PyPI::iniconfig:1.1.1"
+      - id: "PyPI::packaging:21.3"
+        dependencies:
+        - id: "PyPI::pyparsing:3.0.7"
+      - id: "PyPI::pluggy:1.0.0"
+      - id: "PyPI::py:1.11.0"
+      - id: "PyPI::toml:0.10.2"
+    - id: "PyPI::rope:0.14.0"
   - name: "main"
     dependencies:
     - id: "PyPI::graphviz:0.19.1"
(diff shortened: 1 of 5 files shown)


plugins/package-managers/python/src/main/kotlin/Pip.kt
```diff
@@ -98,7 +98,7 @@ class Pip(
         return listOf(ProjectAnalyzerResult(project, packages))
     }
 
-    private fun runPythonInspector(definitionFile: File): PythonInspector.Result {
+    internal fun runPythonInspector(definitionFile: File): PythonInspector.Result {
```
sschuberth, line 101: Commit message:
- "based on / and for" -> "based on / for" (I guess)
- "meta-data" -> "metadata"
- "sibling" -> "siblings" (I guess; also remove the double whitespace afterwards)

plugins/package-managers/python/src/funTest/assets/projects/synthetic/poetry-expected-output.yml
```diff
@@ -22,7 +22,210 @@ project:
     - id: "PyPI::jinja2:3.0.3"
       dependencies:
       - id: "PyPI::markupsafe:2.0.1"
+  - name: "installDev"
```
sschuberth, line 25: Commit message:
- "not" -> "note"

plugins/package-managers/python/src/main/kotlin/Poetry.kt
```diff
@@ -64,9 +64,12 @@ class Poetry(
     override fun transformVersion(output: String) = output.substringAfter("version ").removeSuffix(")")
 
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
-        val result = inspectLockfile(definitionFile)
+        val dependenciesResult = inspectLockfile(definitionFile, includeDevDependencies = false)
+        val devDependenciesResult = inspectLockfile(definitionFile, includeDevDependencies = true)
```
sschuberth, line 68: `includeDevDependencies = true` sounds as if `devDependenciesResult` contains `dependenciesResult` *and* any dev dependencies. If that's so, shouldn't we substract `dependenciesResult` from `devDependenciesResult` and only use that for a `dev` (not `installDev`) scope?

sschuberth (changes requested):

plugins/package-managers/python/src/main/kotlin/Poetry.kt
fviernau: Yes, `devDependenciesResult` includes `dependenciesResult`, but I believe we should not substract, because:

1. The dependencies which end-up in the release artifacts are normally also needed for 
     e.g. test builds
2. It's unclear (to me) how substracting trees should work. We don't do this anywhere, 
    please correct if I'm wrong.
4. It would be inconsistent with e.g. `GoMod` integration and believe also what `.gradlew 
    dependencies` outputs for the respective scopes.

fviernau: @sschuberth - do you agree with above comment, so we can leave as-is?

sschuberth: I still don't understand why it should be correct to end up with two scopes, one named "install" and the other named "installDev", where "install" seems to be a subset of "installDev". If I'm not getting anything wrong, this overlap seems counter-intuitive to me, and it's something that we AFAIK do not do for any other package manager.

fviernau: > that we AFAIK do not do for any other package manager.

I doubt this is true. Have you actually read my comment above? I says:

It would be inconsistent with e.g. GoMod integration and believe also what .gradlew  dependencies outputs for the respective scopes.

sschuberth: > > that we AFAIK do not do for any other package manager.
> 
> I doubt this is true. Have you actually read my comment above

Yes, I have read it, but it even confused me more. For GoMod there seem to be two scopes created, "main" and "vendor". Simply judging from the names, there seems to be no overlap of dependencies here.

I also didn't get your comment about `./gradlew dependencies`. Gradle lists dependencies per configuration, and it's not combining dependencies of configurations. Note that I'm not talking about cases where a given dependency really happens to be part of multiple configurations / scopes.

Or phrased differently: The "installDev" scope sounds to me as if it should actually be an "install" scope (that we already have) and a "dev" scope. And the only reason why we don't have it like that seems to be because it's harder to implement. That's at least my understanding of the issue.

sschuberth: I was just reading up on the [`poetry export` command help](https://python-poetry.org/docs/cli/#export), and could we maybe use `--with dev --without main` to really only get `dev` dependencies?

sschuberth: Or yet better ``--only dev`?

fviernau: > Yes, I have read it, but it even confused me more. For GoMod there seem to be two scopes created, "main" and "vendor". Simply judging from the names, there seems to be no inherent overlap of dependencies here.

I believe the scope names do not provide sufficient information. Looking at https://github.com/oss-review-toolkit/ort/blob/1c4c0fc2ec0eca2aea1a6a4ad51dea8a19da98d0/analyzer/src/main/kotlin/managers/GoMod.kt#L126 it shows that `main` is a subset of `vender` in terms of nodes and edges. 

> I also didn't get your comment about `./gradlew dependencies`. Gradle lists dependencies per configuration, and it's not combining dependencies of configurations. Note that I'm not talking about cases where a given dependency really happens to be part of multiple configurations / scopes.

Hm, so are you saying that e.g. `compileClasspath` is not a subset of `testCompileClasspath`?  


> Or phrased differently: The "installDev" scope sounds to me as if it should actually be an "install" scope (that we already have) and a "dev" scope. 

I think it's inappropriate to derive some meaning from these names as they are rather arbitrarily chosen.  I believe it'd be more appropriate to pick better names if they don't fit the implementation we think makes most sens.

> And the only reason why we don't have it like that seems to be because it's harder to implement. That's at least my understanding of the issue.

My understanding may be wrong, but I thought that it would be consistent with what we already have in e.g. GoMod and Gradle. Furthermore, when resolving dev dependencies it could be that constraints coming from the non-dev dependencies are considered. Or phrased differently, could it be that resolving the dev dependcies without considering the main dependencies version constraints could lead to wrong resolved package version?

>I was just reading up on the poetry export command help, and could we maybe use --with dev --without main to really only get dev dependencies?

Interresting idea, that at least would not require one to subtract trees in some not yet clear / specified way and should be simple to implemented. However, I'm not sure whether the results are as good as with the current approach, because of the above mentioned constraints thing. What do you think about that?

sschuberth: > Hm, so are you saying that e.g. `compileClasspath` is not a subset of `testCompileClasspath`?

Not necessarily, yes. In practice, that will rarely be the case, but my point is that Gradle allows these to contain distinct dependencies without any overlap. They are not "hard-wired" to be a subset of each other. And if they aren't, our Gradle package manager implementation will still get that right.

Coming back to the Poetry docs, I believe what's called "dependency group" should map to a "scope" in ORT. And as "main" and "dev" (and probably more) are distinct dependency groups, they should also be distinct scopes.

fviernau: > Not necessarily, yes. In practice, that will rarely be the case, but my point is that Gradle allows these to contain distinct dependencies without any overlap. They are not "hard-wired" to be a subset of each other. And if they aren't, our Gradle package manager implementation will still get that right.

Thanks. But is this a theoretical consideration or is it like this in practice? (I'm asking because 
in my memories the scopes I've investigated seemed more like they were subsets)
Also my point about constraints would be interesting here: In practice, do the dev dependencies scopes
in gradle consider the version constraints of the release dependencies scopes? (Or phrasing it differently,
are the contraints of release dependencies scopes usually a subset of the dev dependencies scope) ..I'm asking this explicitly, because if they weren't, the resolution might lead to conflicting versions.  

> Coming back to the Poetry docs, I believe what's called "dependency group" should map to a "scope" in ORT. And as "main" and "dev" (and probably more) are distinct dependency groups, they should also be distinct scopes.

Letting aside the difficulty of the implementation for now, I believe we have to consider my point about constraints 
also for poetry. If we omit the release deps for analyzing the dev deps, then the release deps constraints are not considered. So, resolving the union of the constraints of both scopes can yield a result different from resolving 
dev dependencies constraints and release dependency constraints separately. So, one of the two approaches could yield a result inconsistent with the tooling (poetry). 

Would you agree that we should make correctness in this regard a priority for deciding whether to keep implementation as-is or to change it as suggested?

sschuberth: I fear that we again might be having problems understanding each other as we're using terms differently. To me "(dependency) constraints" are somethings like [this](https://docs.gradle.org/current/userguide/dependency_management_terminology.html#sub:terminology_dependency_constraint) in Gradle, i.e. an explicit mean to express additional requirements for transitive dependencies. I'm not aware of Poetry having such a mechanism, hence I believe you're using the term for any implicit "constraint" that results from the same dependency appearing multiple times in different versions / version ranges; i.e. the standard case for general version conflict resolution.

So when you say

> If we omit the release deps for analyzing the dev deps, then the release deps constraints are not considered.

I believe you mean that version conflict resolution might be different for "dev" alone than for a combination of "release" and "dev". But I'd assume that cannot happen as we're completely leaving the dependency resolution to Poetry in all cases; version conflict resolution is nothing that ORT should need to care about. Instead, I'd simply trust `poetry export --only <groupname>` to do the right thing.

fviernau: I've changed it as you suggested, e.g. use `poetry export --only dev`

plugins/package-managers/python/src/funTest/assets/projects/synthetic/poetry-expected-output.yml
sschuberth: Commit message: The last paragraph seem to just repeat the very first sentence on different word, and could IMO be dropped.

plugins/package-managers/python/src/main/kotlin/Poetry.kt
```diff
@@ -73,14 +76,17 @@ class Poetry(
...
+                Scope("installDev", devDependenciesResult.resolvedDependenciesGraph.toPackageReferences())
+            ),
             vcsProcessed = processProjectVcs(definitionFile.parentFile)
         )
 
         return listOf(ProjectAnalyzerResult(project, packages))
     }
 
-    private fun inspectLockfile(lockfile: File): PythonInspector.Result {
+    private fun inspectLockfile(lockfile: File,  devDependencies: Boolean): PythonInspector.Result {
```
sschuberth, line 89: I'd propose to generalize this further and take the dependency group name as a string instead, and pass that to `--only=`. The same names should be used for the scopes then. As a preparation, it might make sense to first rename "install" to "main" to reduce the diff in the expected output. So in total, we should have two scopes, "main" and "dev".

sschuberth (changes requested):

plugins/package-managers/python/src/main/kotlin/Poetry.kt
```diff
@@ -80,7 +80,7 @@ class Poetry(
         return listOf(ProjectAnalyzerResult(project, packages))
     }
 
-    private fun inspectLockfile(lockfile: File): PythonInspector.Result {
+    private fun inspectLockfile(lockfile: File, scopeName: String): PythonInspector.Result {
```
sschuberth, line 90: Nit: From Poetry-perspective this should probably be called `dependencyGroupName` as it's only the caller who decides to map this to ORT's scopes, but I'm fine with keeping the parameter name as-is.

sschuberth (approved):

plugins/package-managers/python/src/funTest/assets/projects/synthetic/poetry-expected-output.yml
```diff
@@ -16,13 +16,212 @@ project:
...
+      dependencies:
+      - id: "PyPI::attrs:22.2.0"
+      - id: "PyPI::iniconfig:1.1.1"
+      - id: "PyPI::packaging:21.3"
+        dependencies:
+        - id: "PyPI::pyparsing:3.0.7"
+      - id: "PyPI::pluggy:1.0.0"
+      - id: "PyPI::py:1.11.0"
+      - id: "PyPI::toml:0.10.2"
+    - id: "PyPI::rope:0.14.0"
```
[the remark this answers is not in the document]
fviernau: no idea, I haven't noticed. I'll take a look.

PR #7633, opened 2023-10-05
feat(pnpm): Add support for PNPM 8.x

MarcelBochtler:
According to [1] the only breaking change in PNPM 8.0.0 is that the support for Node.js 14 was dropped, which is not relevant for ORT's PNPM implementation.

[1]: https://github.com/pnpm/pnpm/blob/main/pnpm/CHANGELOG.md#800

**Breaking change**
The upgrade to PNPM 8.8.0 in the Docker image will cause analyzing PNPM projects that were created using PNPM 7.x resulting in an error.
To fix this, one has to update the PNPM lockfile using version PNPM version 8.0.0 or newer.


plugins/package-managers/node/src/main/kotlin/Pnpm.kt
```diff
@@ -70,7 +70,7 @@ class Pnpm(
 
     override fun command(workingDir: File?) = if (Os.isWindows) "pnpm.cmd" else "pnpm"
 
-    override fun getVersionRequirement(): RangesList = RangesListFactory.create("5.* - 7.*")
+    override fun getVersionRequirement(): RangesList = RangesListFactory.create("5.* - 8.*")
```
fviernau, line 73: Would it make sense to bump `PNPM_VERSION` in docker in this PR, so that we have it covered by the tests?

fviernau, line 73: The change of the default resolution mode in the change log IIUC may lead to change in dependency resolution results. However, I believe if the lockfile practice is followed, it shouldn't have any effect.

fviernau (approved):

.versions
```diff
@@ -14,7 +14,7 @@ NPM_VERSION=8.15.1
 NUGET_INSPECTOR_VERSION=0.9.12
 PHP_VERSION=8.1
 PIPTOOL_VERSION=22.2.2
-PNPM_VERSION=7.8.0
+PNPM_VERSION=8.8.0
```
fviernau, line 17: As this upgrade breaks the tests, I was wondering if updating lockfiles should be squashed. I'll leave it up to you. Without squashing I believe it makes sense to at least mention that tests break and lockfiles are updated in a following commit.

fviernau (changes requested):

fviernau (approved):

PR #7640, opened 2023-10-05
refactor(scanner)!: Use a property that holds all nested provenances

sschuberth:
All known provenance information is immutable, so there is no need to have a function recreate the set on each call. Use a property instead, name it `allProvenances` to avoid the need for the `@JsonIgnore` annotation, and inline some related code for simplicity while at it.

### DIFF
--- a/scanner/src/main/kotlin/provenance/NestedProvenance.kt
+++ b/scanner/src/main/kotlin/provenance/NestedProvenance.kt
@@ -42,6 +42,16 @@ data class NestedProvenance(
     val subRepositories: Map<String, RepositoryProvenance>
 ) {
     /**
+     * The set of all contained [KnownProvenance]s.
+     */
+    @JsonIgnore
+    val allProvenances: Set<KnownProvenance> =
+        buildSet {
+            add(root)
+            addAll(subRepositories.values)
+        }
+
+    /**
      * Return path of the provided [provenance] within this [NestedProvenance]. Throws an [IllegalArgumentException] if
      * the provided [provenance] cannot be found.
      */
@@ -52,14 +62,4 @@ data class NestedProvenance(
 
         throw IllegalArgumentException("Could not find entry for $provenance.")
     }
-
-    /**
-     * Return a set of all contained [KnownProvenance]s.
-     */
-    @JsonIgnore
-    fun getProvenances(): Set<KnownProvenance> =
-        buildSet {
-            add(root)
-            addAll(subRepositories.values)
-        }
 }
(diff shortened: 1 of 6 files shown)


scanner/src/main/kotlin/provenance/NestedProvenance.kt
```diff
@@ -41,6 +39,15 @@ data class NestedProvenance(
      */
     val subRepositories: Map<String, RepositoryProvenance>
 ) {
+    /**
+     * Return the set of all contained [KnownProvenance]s.
+     */
+    val allProvenances: Set<KnownProvenance> =
```
fviernau, line 45: Does this need a `@JsonIgnore` annotation?

scanner/src/main/kotlin/provenance/NestedProvenance.kt
```diff
@@ -41,6 +39,15 @@ data class NestedProvenance(
      */
     val subRepositories: Map<String, RepositoryProvenance>
 ) {
+    /**
```
fviernau, line 42: commit-message: Could it mention why this change is marked as breaking?

scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
```diff
@@ -40,16 +40,11 @@ data class NestedProvenanceScanResult(
...
-     * Return a set of all [KnownProvenance]s contained in [nestedProvenance].
-     */
-    fun getProvenances(): Set<KnownProvenance> = nestedProvenance.getProvenances()
-
     /**
      * Return true if [scanResults] contains at least one scan result for each of the [KnownProvenance]s contained in
      * [nestedProvenance].
      */
-    fun isComplete(): Boolean = getProvenances().all { scanResults[it]?.isNotEmpty() == true }
+    fun isComplete(): Boolean = nestedProvenance.allProvenances.all { scanResults[it]?.isNotEmpty() == true }
```
fviernau, line 47: The removal of the function in line 46 seems to me strictly speaking a bit unrelated / a thing which could become a separate commit. However, would it make sense to keep this function to not introduce these Law of Demeter violations?

fviernau (changes requested):

scanner/src/main/kotlin/provenance/NestedProvenance.kt
sschuberth: Ouch, looks like it's indeed needed. Thanks.

fviernau: Commit message needs an update then.

scanner/src/main/kotlin/provenance/NestedProvenanceScanResult.kt
sschuberth: I agree about maybe splitting this out to a separate commit, but how is the principle of least knowledge violated here?

fviernau: Just notice it's just once, the occurrence in NestedProvenanceScanResultTest.kt, so let's leave it as is.

fviernau (approved):

PR #7641, opened 2023-10-05
refactor(PackageBasedScanStorageReader)!: Align `read()` to take a `Package`

sschuberth:
A *package* based reader should indeed take a `Package` for all its `read()` functions. This aligns with the existing `read()` function that takes a `ScannerCriteria`, and it is a requirement for a proper implementation of `ClearlyDefinedStorage` which uses coordinates with a "provider" for lookup that can only be correctly constructed when provenance information is available. So far the `ClearlyDefinedStorage` hard-codes some assumptions in this regard, which will be corrected in upcoming changes.

### DIFF
--- a/scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
+++ b/scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
@@ -29,8 +29,12 @@ import java.time.Instant
 import kotlin.time.Duration.Companion.seconds
 
 import org.ossreviewtoolkit.clients.clearlydefined.ClearlyDefinedService.Server
+import org.ossreviewtoolkit.model.Hash
+import org.ossreviewtoolkit.model.HashAlgorithm
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.LicenseFinding
+import org.ossreviewtoolkit.model.Package
+import org.ossreviewtoolkit.model.RemoteArtifact
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
 import org.ossreviewtoolkit.model.ScanSummary
@@ -44,10 +48,19 @@ class ClearlyDefinedStorageFunTest : StringSpec({
     val storage = ClearlyDefinedStorage(ClearlyDefinedStorageConfiguration(Server.PRODUCTION.apiUrl))
 
     "Scan results for 'semver4j' from ScanCode should be read correctly" {
-        val id = Identifier("Maven:com.vdurmont:semver4j:3.1.0")
+        val pkg = Package.EMPTY.copy(
+            id = Identifier("Maven:com.vdurmont:semver4j:3.1.0"),
+            binaryArtifact = RemoteArtifact(
+                url = "https://repo1.maven.org/maven2/com/vdurmont/semver4j/3.1.0/semver4j-3.1.0.jar",
+                hash = Hash(
+                    value = "0de1248f09dfe8df3b021c84e0642ee222cceb13",
+                    algorithm = HashAlgorithm.SHA1
+                )
+            )
+        )
 
         withRetry {
-            val results = storage.read(id)
+            val results = storage.read(pkg)
 
             results.shouldBeSuccess {
                 it shouldContain ScanResult(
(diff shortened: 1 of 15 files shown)


scanner/src/main/kotlin/ScanResultsStorage.kt
```diff
@@ -55,15 +55,15 @@ abstract class ScanResultsStorage : PackageBasedScanStorage {
...
 
     /**
-     * Return all [ScanResult]s contained in this [ScanResultsStorage] corresponding to the package denoted by the given
-     * [id] wrapped in a [Result].
+     * Return all [ScanResult]s contained in this [ScanResultsStorage] corresponding to the [package][pkg] wrapped in a
+     * [Result].
      */
-    fun read(id: Identifier): Result<List<ScanResult>> {
-        val (result, duration) = measureTimedValue { readInternal(id) }
+    fun read(pkg: Package): Result<List<ScanResult>> {
```
fviernau, line 61: commit and this read() function: I have difficulties understanding which parts of the package will be used to determine which scan results are being returned and how. Can you help?

fviernau (changes requested):

scanner/src/main/kotlin/ScanResultsStorage.kt
sschuberth: It's indeed a bit hard to understand as the relevant parts are not implemented yet... because they couldn't, as the ClearlyDefined scan storage provider did not have access to the necessary information before this change.

Basically, the ClearlyDefined implementation would need to have access to the VCS location, source artifact location (and maybe also the binary source artifact location) in order to create a proper [Coordinate](https://github.com/oss-review-toolkit/ort/blob/59a4404f1f01434336635e6c6495f407bd6108a0/clients/clearly-defined/src/main/kotlin/Curation.kt#L103) for querying harvested (scan) results. A *Coordinate* includes a [ComponentType](https://github.com/oss-review-toolkit/ort/blob/681df5e2e5392d11901bfa6d40c0cbc821fe5005/clients/clearly-defined/src/main/kotlin/Enums.kt#L31) and [Provider](https://github.com/oss-review-toolkit/ort/blob/681df5e2e5392d11901bfa6d40c0cbc821fe5005/clients/clearly-defined/src/main/kotlin/Enums.kt#L60), and in order to determine these properly, more *Package* properties than just the *Identifier* are required.

mnonnenmacher: As `Package` contains many irrelevant properties I would prefer if the interface would only require the relevant values to have a cleaner API. But I would be fine with looking into it after fixing the ClearlyDefinedStorage.

scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
```diff
@@ -44,10 +48,24 @@ class ClearlyDefinedStorageFunTest : StringSpec({
...
+            binaryArtifact = RemoteArtifact(
+                url = "https://repo1.maven.org/maven2/com/vdurmont/semver4j/3.1.0/semver4j-3.1.0.jar",
+                hash = Hash(
+                    value = "0de1248f09dfe8df3b021c84e0642ee222cceb13",
+                    algorithm = HashAlgorithm.SHA1
+                )
+            ),
+            sourceArtifact = RemoteArtifact.EMPTY,
+            vcs = VcsInfo.EMPTY
+        )
```
mnonnenmacher, line 65: Use `Package.EMPTY.copy(...)` to set only the relevant properties, same two times below.

mnonnenmacher (changes requested):

scanner/src/funTest/kotlin/storages/ClearlyDefinedStorageFunTest.kt
sschuberth: I deliberately didn't do that to benefit from the default values for `purl` and `vcsProcessed`, but turns out that this was meaningful only in an earlier version of the PR.

sschuberth:
> But I would be fine with looking into it after fixing the ClearlyDefinedStorage.

Thanks. I was considering an API like for `PackageConfigurationProvider` which has `provenance: Provenance` in addition to `packageId: Identifier`:

https://github.com/oss-review-toolkit/ort/blob/89c626ed4cf659d50f2f0d9c9a4e37c49740b1c1/model/src/main/kotlin/utils/PackageConfigurationProvider.kt#L41

mnonnenmacher (approved):

PR #7643, opened 2023-10-05
deps: Upgrade the SW360 client to version 17.0.1-m2

sschuberth:
This gets rid of the transitive `jersey-json` dependency that is affected by vulnerabilities.


sschuberth:
Thanks @heliocastro for the fix in the SW360 client!

heliocastro (approved):
Loog good :-)

PR #7647, opened 2023-10-06
fix(analyzer): Add a test for dangling embed directives / GoMod

### LINKED ISSUE #5560: GoMod: go list -deps error: no matching files found

fviernau:
The `GoMod` integration does not execute `go generate/go build`, so the embed directive in `main.go` points to a non-existing file. When `GoMod` executes [1] it failed [2] with the previously used Go version 1.20.5. As of [3], with Go version 1.21.1, that issue is fixed.

Add a test to ensure it keeps on working.

Note: It is essential to pass `json=Module` instead of not just
      `-json` [4], because otherwise the issue would still persist on
      Go 1.21.1.

Fixes #5560.

[1] `go list -deps -json=Module -buildvcs=false ./...`
[2] `pattern file.txt: no matching`
[3] 47e45202a2cd04396fe929b8ac856c98f1b94649
[4] 43ad3a777c6b4f59c487faf255ca590e7d48822d

### DIFF
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed/go.mod
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed/go.mod
@@ -1,0 +1,3 @@
+module gomod_dangling_embed
+
+go 1.19
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -100,7 +100,7 @@ class GoMod(
 
     override fun transformVersion(output: String) = output.removePrefix("go version go").substringBefore(' ')
 
-    override fun getVersionRequirement(): RangesList = RangesListFactory.create(">=1.19.0")
+    override fun getVersionRequirement(): RangesList = RangesListFactory.create(">=1.21.1")
 
     override fun mapDefinitionFiles(definitionFiles: List<File>): List<File> =
         definitionFiles.filterNot { definitionFile ->
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed-expected-output.yml
@@ -1,0 +1,23 @@
+---
+project:
+  id: "GoMod::gomod_dangling_embed:<REPLACE_REVISION>"
+  definition_file_path: "analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed/go.mod"
+  declared_licenses: []
+  declared_licenses_processed: {}
+  vcs:
+    type: "Git"
+    url: "<REPLACE_URL_PROCESSED>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  vcs_processed:
+    type: "Git"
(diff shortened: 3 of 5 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -100,7 +100,7 @@ class GoMod(
 
     override fun transformVersion(output: String) = output.removePrefix("go version go").substringBefore(' ')
 
-    override fun getVersionRequirement(): RangesList = RangesListFactory.create(">=1.19.0")
+    override fun getVersionRequirement(): RangesList = RangesListFactory.create(">=1.21.1")
```
sschuberth, line 103: Commit message:
- "this projects" -> "the projects"

analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed/go.mod
```diff
@@ -0,0 +1,3 @@
+module gomod_dangling_embed
+
+go 1.19
```
sschuberth, line 3: Shouldn't this also be 1.21.1?

sschuberth (changes requested):

analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed/go.mod
fviernau: Why exactly should it be? (As-is it is consistent with the other test cases in `GoModFunTest`).

One aspect of the bug fix is in the tooling of version 1.21, but that tooling can be used with 1.19 projects, so I do not see an issue here.

sschuberth: > One aspect of the bug fix is in the tooling of version 1.21, but that tooling can be used with 1.19 projects

That was the missing piece of information for me here.

sschuberth:
Note that *commit-lint* still fails.

analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed/go.mod
fviernau: > That was the missing piece of information for me here.

Ok, so are we good to go?

fviernau:
> Note that _commit-lint_ still fails.

It seems it flag references to issues in the middle of a sentence without a preceding keyword like "Fixes".

sschuberth:
> My work around is to use a reference e.g. [5].

That's as well the only work-around I'm aware of.

PR #7651, opened 2023-10-06
Log4j simplifications

sschuberth:


utils/ort/src/main/kotlin/OkHttpClientHelper.kt
```diff
@@ -81,6 +76,11 @@ object OkHttpClientHelper {
         } ?: okHttpClient
 }
 
+/**
```
mnonnenmacher, line 79: Commit message: "frist" -> "first"

mnonnenmacher (changes requested):

mnonnenmacher (approved):

nnobelis:
IMO this is a breaking change: since `Logging` was public API before, external plugins could have used it.

sschuberth:
> IMO this is a breaking change: since `Logging` was public API before, external plugins could have used it.

Yes, right. For the PRs / version bumping it's too late, but I'll at least label this PR accordingly.

nnobelis:
@sschuberth: Before this PR, the loggers were throwing exception if the calling classes were not in `org.ossreviewtoolkit.`.
Therefore, the plugins used to piggyback on an ORT class logger to output their messages.

Now, they cannot use ORT classes' loggers but it should not be a problem since the aforementioned exception is not thrown anymore. Right ?

sschuberth:
> Before this PR, the loggers were throwing exception if the calling classes were not in `org.ossreviewtoolkit.`.

I don't believe that's true. That code was already removed way back in 606da3a305738e6bb68314f88b4eb7bd3ca84040.

However, I acknowledge that this PR reintroduces a behavior that 606da3a305738e6bb68314f88b4eb7bd3ca84040 circumvented, which is to "Stop polluting `Any` object with a `logger` property". This "pollution" is now back again (for files that import the extension property), but by now I actually regard this to be less of an issue, as also with the previous code logging on the wrong class could haven happened by accidentally using the base class's logger in derived classes.

> since the aforementioned exception is not thrown anymore. Right ?

The exception is not thrown in any case anymore.

nnobelis:
Understood thanks !

PR #7658, opened 2023-10-08
Plugin secret options

mnonnenmacher:
Introduce a separate map for secret options of configurable plugins. This improves the handling of secrets options, see the commit messages for details.


plugins/package-configuration-providers/dir/src/main/kotlin/DirPackageConfigurationProvider.kt
```diff
@@ -51,7 +51,7 @@ open class DirPackageConfigurationProviderFactory :
 
     override fun create(config: DirPackageConfigurationProviderConfig) = DirPackageConfigurationProvider(config)
 
-    override fun parseOptions(options: Options) =
+    override fun parseConfig(options: Options) =
```
sschuberth: I basically agree with the change, but I believe the rationale "to name the function after the output instead of the input" isn't a good one, because "parse" refers to the input, not the the output, and it makes more sense to say what is being parsed than what the output will be. A better rationale IMO would be to do this as a preparation for the `options` vs `secrets` split, to find a generic term that fits both.

cli/src/funTest/kotlin/OrtMainFunTest.kt
```diff
@@ -63,7 +63,7 @@ class OrtMainFunTest : StringSpec() {
                     packageCurationProviders = listOf(
                         ProviderPluginConfiguration(
                             type = "File",
-                            config = mapOf("path" to getAssetFile("gradle-curations.yml").path)
+                            options = mapOf("path" to getAssetFile("gradle-curations.yml").path)
```
sschuberth, line 66: Also here, I believe a better / stronger rationale for this change would be that we need to distinguish from `secrets`, which otherwise could also be seen as part of the generic "config".

model/src/main/kotlin/config/ProviderPluginConfiguration.kt
```diff
@@ -19,6 +19,8 @@
 
 package org.ossreviewtoolkit.model.config
 
+import com.fasterxml.jackson.annotation.JsonIgnore
```
sschuberth, line 22: Commit message: Typo in "configuraion".

sschuberth (changes requested):

sschuberth (approved):

PR #7664, opened 2023-10-10
Simplifications for ClearlyDefined coordinates

sschuberth:


fviernau (changes requested):

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
```diff
@@ -222,18 +222,12 @@ interface ClearlyDefinedService {
...
      */
-    @GET("curations/{type}/{provider}/{namespace}/{name}/{revision}")
-    suspend fun getCuration(
-        @Path("type") type: ComponentType,
-        @Path("provider") provider: Provider,
-        @Path("namespace") namespace: String,
-        @Path("name") name: String,
-        @Path("revision") revision: String
-    ): Curation
+    @GET("curations/{coordinates}")
```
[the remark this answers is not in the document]
sschuberth: I had to solve this differently as Retrofit escapes `/` contained in the string representation of `coordinates` as `%2F` instead of literally using it as a path.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
```diff
@@ -223,7 +223,7 @@ interface ClearlyDefinedService {
 
     /**
      * Get the curation for the component described by [type], [provider], [namespace] (use "-" if not applicable),
-     * [name] and [revision], see
+     * [name] and [revision] (when empty, the latest revision will be used if that makes sense for the provider), see
```
fviernau, line 226: commit message: Should it say "revision" instead of "provider" ?

clients/clearly-defined/src/main/kotlin/Coordinates.kt
```diff
@@ -62,10 +62,12 @@ data class Coordinates(
         provider = Provider.fromString(parts[1]),
         namespace = parts[2].takeUnless { it == "-" },
         name = parts[3],
-        revision = parts.getOrNull(4)
+        revision = parts[4].takeUnless { it.isEmpty() }
     )
 
-    override fun toString() = listOfNotNull(type, provider, namespace ?: "-", name, revision).joinToString("/")
+    val strings: List<String> = listOf(type.toString(), provider.toString(), namespace ?: "-", name, revision.orEmpty())
```
fviernau, line 68: Would `components` be a more speaking name maybe?

fviernau (changes requested):

clients/clearly-defined/src/main/kotlin/Coordinates.kt
sschuberth: "components" reminds a bit too much of the `ComponentType` class, and I'd like to emphasize that these are the string representations of all (also the nullable) properties. I had `entries` before, but that did also not reflect that bit good enough for my taste. So `strings` pragmatically was the best idea that I had.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
```diff
@@ -289,6 +289,28 @@ interface ClearlyDefinedService {
...
 }
 
+suspend fun ClearlyDefinedService.getCuration(coordinates: Coordinates): Curation =
+    @Suppress("DestructuringDeclarationWithTooManyEntries")
+    coordinates.strings.let { (_, _, namespace, name, revision) ->
+        getCuration(coordinates.type, coordinates.provider, namespace, name, revision)
+    }
+
+suspend fun ClearlyDefinedService.harvestTools(coordinates: Coordinates): List<String> =
+    @Suppress("DestructuringDeclarationWithTooManyEntries")
```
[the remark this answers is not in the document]
sschuberth: The called functions take `type` and `provider` as types enums, not as strings, so their string representations are "discarded" and their typed values are used instead. I liked that more than reconstructing the enums from the strings.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
fviernau, line 226: I believe the docs should also state what the behavior is, when revision is empty and it doesn't make sense for the provider to query the latest revision.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
```diff
@@ -257,7 +257,8 @@ interface ClearlyDefinedService {
 
     /**
      * Get information about the harvest tools that have produced data for the component described by [type],
-     * [provider], [namespace] (use "-" if not applicable), [name], and [revision], see
+     * [provider], [namespace] (use "-" if not applicable), [name], and [revision] (when empty, the latest revision will
```
fviernau, line 260: I believe the docs should also state what the behavior is, when revision is empty and it doesn't make sense for the provider to query the latest revision.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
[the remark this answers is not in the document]
fviernau: Hm, this means the first two items in the `strings` list are unused.

Do I get it correctly, that the only reason why the list is introduced is for reusing
`revision.orEmpty()` and ensuring that `-` represents the empty namespace.

I wonder if alternatively making `revision` not null in Coordinates, and disallowing empty strings as namespace, enforcing `-`, then all properties could just be as is - which would make the `string` property unnecessary. This code the could just look like

`with(coordinates) { harvestTools(type, provider, namespace, name, revision) }`

What do you think?

fviernau (changes requested):

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
sschuberth: This sentence just comes as-is from the [official docs](https://api.clearlydefined.io/api-docs/#/harvest/get_harvest__type___provider___namespace___name___revision_) that also don't state what happens if an empty revision is used with a provider that does not support it, so I can't tell.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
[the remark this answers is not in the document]
sschuberth: > Do I get it correctly, that the only reason why the list is introduced is for reusing
> `revision.orEmpty()` and ensuring that `-` represents the empty namespace.

Basically yes.

> What do you think?

The `Coordinates` class is also used in API responses, so we cannot make `revision` (or `namespace`) non-null as then deserialization of results would fail.

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
fviernau: I see, thanks.

fviernau:
> @fviernau do my answer make sense to you? Are we ok to merge?

Yes. I think the API simplication is good. I have a hunch that the complexity of the implementation can be reduced a bit, but I'd also need to look deeper into this to see find an alternative if exits.: e.g. the discussed unused first two values in the introduced `strings()` list. So, I'm ok to merge it.

fviernau (approved):

clients/clearly-defined/src/main/kotlin/ClearlyDefinedService.kt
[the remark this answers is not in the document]
sschuberth: I actually [*did* find a way to solve the escaping problem](https://github.com/oss-review-toolkit/ort/pull/7688). And yes, we need to rely on the `toString()` implementation, as that's what Retrofit uses to turn objects into paths.

PR #7682, opened 2023-10-11
Scanner: Take secret options into use

mnonnenmacher:
Take the secret options into use for scanner wrapper plugins. Please see the commit messages for details.

**Breaking change:**
The configuration for scanner wrappers was changed, see the diff of the `reference.yml` in this PR for examples.

### DIFF
--- a/model/src/main/kotlin/config/PluginConfiguration.kt
+++ b/model/src/main/kotlin/config/PluginConfiguration.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.model.config
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+
+import org.ossreviewtoolkit.utils.common.ConfigurablePluginFactory
+import org.ossreviewtoolkit.utils.common.Options
+
+/**
+ * The configuration of [configurable plugins][ConfigurablePluginFactory].
+ */
+data class PluginConfiguration(
+    /**
+     * The configuration options of the plugin. See the specific implementation for available configuration options.
+     */
+    val options: Options = emptyMap(),
+
+    /**
+     * The configuration secrets of the plugin. See the specific implementation for available secret options.
+     *
+     * This property is not serialized to ensure that secrets do not appear in serialized output.
+     */
+    @JsonIgnore
+    val secrets: Options = emptyMap()
+) {
+    /**
+     * Return a string representation of the object that does not contain the [secrets].
+     */
+    override fun toString() = "${this::class.simpleName}(options=$options)"
+}
--- a/model/src/main/kotlin/config/ProviderPluginConfiguration.kt
+++ b/model/src/main/kotlin/config/ProviderPluginConfiguration.kt
@@ -64,9 +64,8 @@ data class ProviderPluginConfiguration(
     @JsonIgnore
     val secrets: Options = emptyMap()
 ) {
-    override fun toString(): String {
-        // Do not use the generated toString function for the data class to ensure that the output does not contain the
-        // secret options.
-        return "${this::class.simpleName}(type=$type, id=$id, enabled=$enabled, options=$options)"
-    }
+    /**
+     * Return a string representation of the object that does not contain the [secrets].
+     */
+    override fun toString() = "${this::class.simpleName}(type=$type, id=$id, enabled=$enabled, options=$options)"
 }
--- a/model/src/main/kotlin/config/ProviderPluginConfiguration.kt
+++ b/model/src/main/kotlin/config/ProviderPluginConfiguration.kt
@@ -23,11 +23,15 @@ import com.fasterxml.jackson.annotation.JsonIgnore
 
 import com.sksamuel.hoplite.ConfigAlias
 
+import org.ossreviewtoolkit.utils.common.ConfigurablePluginFactory
 import org.ossreviewtoolkit.utils.common.Options
 import org.ossreviewtoolkit.utils.common.Plugin
 
 /**
- * The configuration of provider plugins.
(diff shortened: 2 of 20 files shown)


model/src/main/kotlin/config/ProviderPluginConfiguration.kt
```diff
@@ -64,9 +64,7 @@ data class ProviderPluginConfiguration(
     @JsonIgnore
     val secrets: Options = emptyMap()
 ) {
-    override fun toString(): String {
-        // Do not use the generated toString function for the data class to ensure that the output does not contain the
-        // secret options.
-        return "${this::class.simpleName}(type=$type, id=$id, enabled=$enabled, options=$options)"
-    }
+    // Do not use the generated toString function for the data class to ensure that the output does not contain the
+    // secret options.
```
sschuberth, line 68: Nit: How about also rephrasing this slightly into a proper function documentation now?

model/src/main/resources/reference.yml
```diff
@@ -202,44 +202,46 @@ ort:
...
+          maxVersion: '32.0.0'
 
       FossId:
-        serverUrl: 'https://fossid.example.com/instance/'
-        user: user
-        apiKey: XYZ
+        options:
+          serverUrl: 'https://fossid.example.com/instance/'
+          user: user
+          apiKey: XYZ
```
sschuberth: These should be secrets, I believe.

sschuberth: Ok, I see this happens in the next commit. I was expecting it to already happen in this commit as the commit message talk about secrets options.

model/src/main/resources/reference.yml
```diff
@@ -245,6 +245,12 @@ ort:
           user: user
           apiKey: XYZ
 
+      SCANOSS:
```
sschuberth, line 248: Commit message nit. In free text it should be written as "SCANOSS".

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -49,7 +49,6 @@ import org.ossreviewtoolkit.model.ScanSummary
 import org.ossreviewtoolkit.model.ScannerRun
 import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.config.DownloaderConfiguration
-import org.ossreviewtoolkit.model.config.PluginConfiguration
```
sschuberth: Commit message: "not stored" -> "now stored"

sschuberth (changes requested):

model/src/main/kotlin/config/PluginConfiguration.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+
+import org.ossreviewtoolkit.utils.common.ConfigurablePluginFactory
+import org.ossreviewtoolkit.utils.common.Options
+
+/**
+ * The configuration of [configurable plugins][ConfigurablePluginFactory].
+ */
+data class PluginConfiguration(
```
sschuberth, line 30: Is it really worth to introduce this class instead of just reusing `ProviderPluginConfiguration`? That's at least what I did in my [draft](https://github.com/oss-review-toolkit/ort/pull/6613/commits/5cfca0165879caeb7dc1cca111541010149f52a8) by renaming `PackageCurationProviderConfiguration` (meanwhile called `ProviderPluginConfiguration`) to `PluginConfiguration`.

model/src/main/resources/reference.yml
mnonnenmacher: Yes, this commit is only about changing the structure of the scanner configuration to support secret options.

model/src/main/kotlin/config/PluginConfiguration.kt
mnonnenmacher: I don't like reusing `ProviderPluginConfiguration` because it has the properties `type`, `id`, and `enabled` which are all meaningless here but would still be available via the API.

sschuberth: > it has the properties `type`, `id`, and `enabled` which are all meaningless here

I'm not yet convinced. `type` actually is the name in this context, and of course also scanners have names. `enabled` could become useful for scanners, too, e.g. when introducing a new experimental scanner for wider testing. Only `id` is not useful here currently, but that seems acceptable in exchange for more simplicity in the class hierarchy.

mnonnenmacher: `type` is used for package curation and configuration providers because their configuration is a list and there can be multiple instance of the same type, for scanners that is not the case. I see the issues illustrated below with the `type` property:

```yaml
# Current map
ScanCode:
  options:
    ...
  secrets:
    ...

# Using list approach like for provider plugins:
- type: ScanCode
  options:
    ...
  secrets:
    ..
- type: ScanCode # Second ScanCode entry is valid configuration syntax that needs additional checks at runtime.
  options:
    ...
  secrets:
    ..

# Using `ProviderPluginConfiguration` with the current map approach is no option because it would make this valid configuration:
ScanCode:
  type: ScanCode
  options:
    ...
FossId:
  type: ScanCode # Valid syntax but invalid configuration that needs additional checks at runtime.
```

Scanners are currently enabled via the `--scanners` and `--project-scanners` command line options. The `enabled` flag cannot be mapped to that concept, so what should it actually mean if the flag is set to `true` or `false`?

sschuberth: > `type` is used for package curation and configuration providers because their configuration is a list and there can be multiple instance of the same type

I guess you mixed up `type` and `id`?

> Scanners are currently enabled via the `--scanners` and `--project-scanners` command line options. The `enabled` flag cannot be mapped to that concept, so what should it actually mean if the flag is set to `true` or `false`?

An, looks like now I mixed up `Plugin.isEnabledByDefault` and `ProviderPluginConfiguration.enabled`.

model/src/main/kotlin/config/ProviderPluginConfiguration.kt
```diff
@@ -23,11 +23,15 @@ import com.fasterxml.jackson.annotation.JsonIgnore
...
 import com.sksamuel.hoplite.ConfigAlias
 
+import org.ossreviewtoolkit.utils.common.ConfigurablePluginFactory
 import org.ossreviewtoolkit.utils.common.Options
 import org.ossreviewtoolkit.utils.common.Plugin
 
 /**
- * The configuration of provider plugins.
+ * The configuration of provider [plugins][ConfigurablePluginFactory]. This class is used when multiple instances of the
+ * same plugins should be configurable, like it is for example possible for package curation providers. Therefore, each
```
sschuberth, line 32: Maybe say "same type of plugin".

model/src/main/kotlin/config/PluginConfiguration.kt
mnonnenmacher: > > `type` is used for package curation and configuration providers because their configuration is a list and there can be multiple instance of the same type
> 
> I guess you mixed up `type` and `id`?

What I meant is that in the map use for scanner configurations the type is the string key of the map (`Map<String, PluginConfiguration>`), so there can be only one instance per type by design. For the providers the map does not work and it was required to move the type into the configuration class because there multiple configurations for the same type should be possible.
That means, if we use a configuration class that contains the type we run into the issues illustrated above as for scanners (and advisors) there can be only one instance for each plugin type.

sschuberth (approved):

Issue #7649, opened 2023-10-06
GoMod: Support for local / sub module dependencies


fviernau:
In mono repo setups with multiple modules, some of the modules may depend on other modules of the same repository.
In that case listing the modules with `go list` as done by `GoMod` yields more than one module with `Main=true`, whereas all such modules do not have a version. This in turn makes ORT's precondition check fail with:
```
IllegalArgumentException: Expected exactly one unique package without version but got Identifier
```
This is solely, because this use case has deliberately not yet been implemented. 

The problem can happen in the use cases:
1. Replacing a dependency with a local one (#6710)
    ```
    replace gitlab.com/myProject/myRepo/GoProject => ../localProject
    ```
2. Analyzing mono repos with references between modules (#6532, #6519)
   ```
    ort/cli/build/install/ort/bin/ort -P ort.analyzer.allowDynamicVersions=true analyze -i $GOPATH/src/k8s.io/kubernetes -o ~/kubernetesORTAnalyzerOutput -f JSON
   ```

There were multiple bugs filed for this. I've decided to close them in favor of this single consolidated enhancement ticket, also because it's not a bug but a missing feature: 
- #6710
- #6532
- #6519

Interesting things:
- https://github.com/golang/go/issues/40513#issuecomment-667328207
- https://go.dev/ref/mod#go-mod-file-replace
- ```
  go list -json -m all
  go: errors parsing go.mod:
  go.mod:18: replacement module without version must be directory path (rooted or starting with ./ or ../) 
  ```

PR #7684, opened 2023-10-12
refactor(GoMod)!: Do not use the revision as project version

fviernau:
Go modules do not have a dedicated metadata field for specifying the version. The only way a version can be defined for a module is via a VCS tag. So, stop using the revision as version for consistency with other package manager integrations. This also simplifies an upcoming change.

Part of #7649.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -144,7 +144,7 @@ class GoMod(
                             type = managerName,
                             namespace = "",
                             name = projectId.name,
-                            version = projectVcs.revision
+                            version = ""
                         ),
                         definitionFilePath = VersionControlSystem.getPathInfo(definitionFile).path,
                         authors = emptySet(), // Go mod doesn't support author information.
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed-expected-output.yml
@@ -1,6 +1,6 @@
 ---
 project:
-  id: "GoMod::gomod_dangling_embed:<REPLACE_REVISION>"
+  id: "GoMod::gomod_dangling_embed:"
   definition_file_path: "analyzer/src/funTest/assets/projects/synthetic/gomod-dangling-embed/go.mod"
   declared_licenses: []
   declared_licenses_processed: {}
--- a/analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
+++ b/analyzer/src/funTest/assets/projects/synthetic/gomod-expected-output.yml
@@ -1,6 +1,6 @@
 ---
 project:
-  id: "GoMod::github.com/oss-review-toolkit/ort/gomod-synthetic-test-project:<REPLACE_REVISION>"
+  id: "GoMod::github.com/oss-review-toolkit/ort/gomod-synthetic-test-project:"
   definition_file_path: "<REPLACE_DEFINITION_FILE_PATH>"
   declared_licenses: []
   declared_licenses_processed: {}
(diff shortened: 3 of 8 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -144,7 +144,7 @@ class GoMod(
                             type = managerName,
                             namespace = "",
                             name = projectId.name,
-                            version = projectVcs.revision
+                            version = ""
```
sschuberth, line 147: I agree that using a plain revision / SHA1 seems wrong here, but while reading through https://go.dev/doc/modules/version-numbers I was wondering whether we should instead use a pseudo or pre-release version as calculated by some Go tooling (I don't know which one, but the docs indicate that a version number could be "generated").

sschuberth: Looks like `go list -f '{{.Version}}'` could do the trick.

fviernau: I just tried it:
```
analyzer/src/funTest/assets/projects/synthetic/gomod$ go list -f '{{.Version}}'
template: main:1:2: executing "main" at <.Version>: can't evaluate field Version in type *load.PackagePublic
```

fviernau: My main goal was to get rid of the version as it simplifies some upcoming work.

sschuberth: You might also need to pass `-m` for the module, [it seems](https://stackoverflow.com/a/69419917/1127485).

fviernau:
```
analyzer/src/funTest/assets/projects/synthetic/gomod$ go list -f '{{.Version}}' -m github.com/oss-review-toolkit/ort/gomod-synthetic-test-project
frank@EPDEFRAW0010:~/devel/ort/ort/analyzer/src/funTest/assets/projects/synthetic/gomod$ echo $?
0
```

sschuberth: This works:

```
$ go list -f '{{.Version}}' -m github.com/oss-review-toolkit/ort/analyzer/src/funTest/assets/projects/synthetic/gomod@main
v0.0.0-20231012190251-67ff91afd89e
```

fviernau: Anyhow, can we just make the version empty to simplify things? (As this was the original goal of this PR)

sschuberth: I get that the goal was to set the project's revision to empty, but I'm questioning the rationale. You wrote:

> The only way a version can be defined for a module is via a VCS tag.

This turns out not to be true for projects, which can have a calculated pseudo-version even without tags.

So even if it's convenient for your use-case to have an empty project revision, why is it *correct*?

fviernau: Yes, as the name says its a **pseudo-version**. Only the fact that it exists does not mean we have to use it.
For other package manager we also make the version empty. To me it's unneeded complexity. 
If it's not unneeded, then we should be able to define the need. Can you help?

sschuberth: > Yes, as the name says its a **pseudo**-version. Only the fact that it exists does not mean we have to use it.

And the fact that it is called "pseudo" does not mean that we should *not* use it. I believe the version is not called "pseudo" because it's bogus, but because it's calculated and in that sense "not real", but still well defined.

> For other package manager we also make the version empty. 

If I'm not mistaken, we only use an empty project version if there is absolutely no reasonable way to get a meaningful version. However, this is not the case here.

> To me it's unneeded complexity.

To me it's correctness, that unfortunately sometimes implies complexity.

> If it's not unneeded, then we should be able to tell why it is needed. Can you help?

Well, the ORT analyzer in general always had the goal to collect as many facts as possible about the analyzed projects, and we always preferred correctness over performance or complexity so far. Following that paradigm, if it possible to get a meaningful project version with reasonable effort, we should do so.

> If you really got that this was the goal to simplify, why are you proposing to make it more complex than?

Because I believe that the goal is wrong / not in line with ORT's principles.

fviernau: I'll implemented it as suggested, but after I merged my upcoming refactoring PR. Putting this on hold for now.

fviernau:
Closing this as it won't get priority from me anytime soon.

PR #7690, opened 2023-10-12
Migrate Advisor plugins to configurable plugin API

mnonnenmacher:
Migrate the advisor plugins to use the configurable plugin API. See the commit messages for details.

**Breaking change:**
The configuration of advice providers has changed, see the diff of `reference.yml` for examples.

**Breaking API change:**
The API of the `AdviceProviderFactory` has changed.


advisor/src/main/kotlin/advisors/GitHubDefectsConfiguration.kt
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (C) 2021 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
```
sschuberth, line 2: Should this simply be 2020 like in the source file, not going as deep as considering when the class was introduced? Same for the other extracted classes.

advisor/src/main/kotlin/advisors/GitHubDefects.kt
```diff
@@ -89,8 +107,18 @@ class GitHubDefects(name: String, config: GitHubDefectsConfiguration) : AdvicePr
...
-        override fun create(config: AdvisorConfiguration) = GitHubDefects(type, config.forProvider { gitHubDefects })
+    class Factory : AdviceProviderFactory<GitHubDefectsConfiguration>("GitHubDefects") {
+        override fun create(config: GitHubDefectsConfiguration) = GitHubDefects(type, config)
+
+        override fun parseConfig(options: Options, secrets: Options) =
+            GitHubDefectsConfiguration(
+                token = secrets["token"],
+                endpointUrl = options["endpointUrl"],
+                labelFilter = options["labelFilter"]?.split(",")?.map { it.trim() }
+                    ?: listOf("!duplicate", "!enhancement", "!invalid", "!question", "*"),
```
sschuberth, line 117: This now duplicates the default value of `GitHubDefectsConfiguration.labelFilter`, which IMO is not so nice. In [my old draft](https://github.com/oss-review-toolkit/ort/pull/6613) I tried to [solve this by extracting default values to constants](https://github.com/oss-review-toolkit/ort/pull/6613/commits/0db7e0f29d750b6e336cff02658f1673b1e8673d#diff-c683de376a76556ce1a5c1d790017956b5726c1b270ca08a1e3e2d61a76054cdR106-R112) (not fully implemented).

This raises the general question where default values should be applied. I see basically tree options:

1. in `parseConfig()`,
2. in the constructor of the output config class,
3. in the consumer of the config class.

My current preference would be 1., and I'd propose a clean-up commit that removes the default parameters from `GitHubDefectsConfiguration`'s constructor.

sschuberth (changes requested):

advisor/src/main/kotlin/advisors/GitHubDefectsConfiguration.kt
mnonnenmacher: Since I already took the time to figure out the correct years I'd rather keep them.

advisor/src/main/kotlin/advisors/GitHubDefects.kt
mnonnenmacher: Would you be fine with doing that clean up commit outside of this PR?

advisor/src/main/kotlin/advisors/GitHubDefectsConfiguration.kt
sschuberth: Ok, it wasn't clear to me that you already invested that effort as the years looked wrong at the first glance. For example, I thought `OssIndexConfiguration` was introduced earlier, but you're right, it seems to have been only in 2023.

advisor/src/main/kotlin/advisors/GitHubDefects.kt
sschuberth: Yes, if it's resolved within reasonable time, before we forget about it.

mnonnenmacher:
Merged as failing `GoModFunTest` is unrelated to the change.

nnobelis:
Once again this is a breaking change that breaks our plugins. Please add the corresponding tags to the PR.

sschuberth:
> Once again this is a breaking change that breaks our plugins. Please add the corresponding tags to the PR.

@nnobelis Conventional commits and release notes now supersede PR labels in this regard. Please take a look at the "Breaking Changes" section for a release instead (this time it was not forgotten to mark commits accordingly).

nnobelis:
Ah yes true my mistake.

Could we, as a convention, put also something in the PR title so we can have a quick oversight when browsing the Github notifications ?

sschuberth:
> put also something in the PR title

I'm not a fan of this as it's prone to typos etc., and we already have release notes, PR labels, and a note in the PR's top post; we can't put it just *everywhere*.

PR #7694, opened 2023-10-13
chore(stack)!: Temporarily remove Stack from Dockerfile-legacy and disable Stack and Pub tests

mnonnenmacher:
Temporarily remove Stack from the Dockerfile-legacy and disable the `StackFunTest` as GitHub runners are running out of disk space and the Stack installation requires a lot. Other candidates would have been Pub or SPM as they also take a lot of disk space, but those package managers are more actively used.


sschuberth (approved):

mnonnenmacher:
Only `YarnFunTest` fail in `funTest-docker` but it passes locally, it looks like some API requests failed during the test as the printed diff only removes some metadata.

sschuberth (approved):

sschuberth:
@mnonnenmacher did you also find out whether disabling `PubFunTest` alone would do the trick?

mnonnenmacher:
> @mnonnenmacher did you also find out whether disabling `PubFunTest` alone would do the trick?

@sschuberth No, I didn't. I could create a separate revert PR after merging this to check? Then we could at least already rebase other changes. If you agree, please just go ahead and merge this. Another option would be that I create another PR with just the Pub change and we wait for the results there.

PR #7697, opened 2023-10-13
Revert "chore(stack)!: Temporarily disable Stack in Dockerfile-legacy and tests"

mnonnenmacher:
Check if just disabling `PubFunTest` in #7694 would have been enough.


sschuberth:
BTW, should we use "revert:" as a conventional commit type for such commits?

sschuberth:
> BTW, should we use "revert:" as a conventional commit type for such commits?

Apparently we also accept Git's generated "Revert ..." wording:

https://github.com/oss-review-toolkit/ort/blob/a9bd271f53cbbd9fc1d20a290eff1d77c7231349/.commitlintrc.yml#L10

sschuberth (approved):

PR #7699, opened 2023-10-14
fix: Failing fun tests running inside Docker

fviernau:
Recently, PRs got merged where fun tests runs in Docker have been skipped. These two issues seem to have slipped through.

Note: I've verified locally that PoetryFunTest produces the same diff for `python-inspector` version 0.9.6 and 0.9.8.

### DIFF
--- a/Dockerfile-legacy
+++ b/Dockerfile-legacy
@@ -33,7 +33,7 @@ ARG ORT_VERSION="DOCKER-SNAPSHOT"
 ARG NUGET_INSPECTOR_VERSION=0.9.12
 
 # Set this to the Python Inspector version to use.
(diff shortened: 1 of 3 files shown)


sschuberth (approved):

sschuberth:
@fviernau I've pushed an additional SpmFunTest update to your branch to make tests pass.

sschuberth (approved):

sschuberth:
Let's merge this as it's certainly correct, even if more work is needed.

fviernau:
> Let's merge this as it's certainly correct, even if more work is needed.

The test failure of the most recent test run are not reproducible on my machine.

PR #7702, opened 2023-10-16
chore(plugins): Get all package configuration / curation plugins lazily

sschuberth:
Scanning the classpath for plugin implementations can be expensive, so only do this when needed for package configuration / curation providers, like already done for other plugin types.


sschuberth:
Merging despite the unrelated `GoModFunTest` failure.

PR #7707, opened 2023-10-18
build(scanner): Do not hard-code the dependency on scanner plugins

sschuberth:
Slightly change the logic for selecting the default scanner: If there is only a single scanner available, use that one, otherwise try to use ScanCode as usual.


plugins/commands/scanner/src/main/kotlin/ScannerCommand.kt
```diff
@@ -115,7 +114,8 @@ class ScannerCommand : OrtCommand(
     private val scanners by option(
         "--scanners", "-s",
         help = "A comma-separated list of scanners to use.\nPossible values are: ${ScannerWrapper.ALL.keys}"
-    ).convertToScannerWrapperFactories().default(listOf(ScanCode.Factory()))
+    ).convertToScannerWrapperFactories()
+        .default(listOfNotNull(ScannerWrapper.ALL.values.singleOrNull() ?: ScannerWrapper.ALL["ScanCode"]))
```
fviernau, line 118: Would it make sense to split-up this commit into two?

sschuberth: Done.

fviernau (approved):

sschuberth:
Merging despite the unrelated `NuGetFunTest` failure.

PR #7714, opened 2023-10-19
GoMod: Tremendously simplify applying replace directive and further things

fviernau:
The changes in this PR aim at applying the replace directive at the very end.
In particular, make it so that the construction of the dependency tree can disregard replace directives entirely.

The changes make it straight forward to implement #7649 on top.


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -119,35 +119,30 @@ class GoMod(
         stashDirectories(projectDir.resolve("vendor")).use { _ ->
             val moduleInfoForModuleName = getModuleInfos(projectDir)
             val graph = getModuleGraph(projectDir, moduleInfoForModuleName)
-            val projectId = moduleInfoForModuleName.getMainModuleId()
-            val packageIds = graph.nodes - projectId
-            val packages = packageIds.mapTo(mutableSetOf()) { moduleInfoForModuleName.getValue(it.name).toPackage() }
-            val projectVcs = processProjectVcs(projectDir)
+            val packages = graph.nodes.mapNotNullTo(mutableSetOf()) {
```
sschuberth, line 122: Commit message: Maybe https://go.dev/ref/mod#go-mod-file-replace is a better reference for `[2]`?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -221,13 +215,29 @@ class GoMod(
...
+        return if (moduleInfo.version.isBlank()) {
+            require(moduleInfo.main) {
+                "Found a local module dependency which is not supported."
+            }
+
+            Identifier(
+                type = managerName,
+                namespace = "",
+                name = moduleInfo.path,
+                version = processProjectVcs(File(moduleInfo.dir!!)).revision
```
sschuberth, line 230: Would it make sense to add a `requireNotNull(moduleInfo.dir) {}` with an appropriate message instead of using `!!`?

In general, I believe the `require...()` should be `check...()` as this is about program state, not user input.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -221,13 +215,29 @@ class GoMod(
...
+            )
+        } else {
+            Identifier(
+                type = "Go",
```
sschuberth, line 234: Why is this not `"GoMod"` / `managerName`? If it's intended, please add a code comment to explain.

sschuberth: And if this should be `managerName`, the code is very similar to the above clause, so it might make sense to only set a `version` conditionally, and construct the `Identifier` once in the end.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -236,46 +246,27 @@ class GoMod(
...
-                    // from the replaced module.
-                    val replace = moduleInfo.replace.copy(indirect = moduleInfo.indirect)
-                    put(moduleInfo.path, replace)
-                    put(moduleInfo.replace.path, replace)
-                } else {
-                    put(moduleInfo.path, moduleInfo)
-                }
-            }
-        }
+        return list.stdout.byteInputStream().use { JSON.decodeToSequence<ModuleInfo>(it) }.associateBy { it.path }
```
sschuberth, line 249: This is a bit long for my taste. Would it make sense to keep the original `moduleInfos` (which also slightly reduces the diff), and just `return moduleInfos.associateBy { it.path }` for clarity *what* is being associated?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -292,7 +283,10 @@ class GoMod(
         }
     }
 
-    private fun ModuleInfo.toPackage(): Package {
+    private fun ModuleInfo.toPackage(): Package? {
+        if (version.isBlank()) return null // A ModuleInfo with blank version should be represented by a Project.
+        if (replace != null) return replace.toPackage() // Apply the replace directive.
```
sschuberth, line 288: Nit: Consider moving the code comments to above the `if`s (and separating them with an empty line).

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -338,26 +332,30 @@ class GoMod(
...
             return PackageReference(
-                id = id,
+                id = moduleInfoForModuleName.getValue(module.name).toId(),
                 linkage = PackageLinkage.PROJECT_STATIC,
                 dependencies = dependencies
             )
         }
 
-        return getDependencies(root).mapTo(mutableSetOf()) { getPackageReference(it) }
+        val mainModule = moduleInfoForModuleName.values.single { it.main }.let {
```
sschuberth, line 354: Nit: Consider using `run` here for consistency with line 172.

sschuberth (changes requested):

analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
```diff
@@ -20,7 +20,9 @@
 package org.ossreviewtoolkit.analyzer.managers
 
 import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.collections.beEmpty
```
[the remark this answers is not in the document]
fviernau: > While you could change the message, I'd actually prefer to adjust the test accordingly.

What do you mean by this?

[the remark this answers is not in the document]
sschuberth: > Should the code assert the amount of issues?

Yes, that's my proposal.

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: I preferred [2], because of this comment: https://github.com/golang/go/issues/40513#issuecomment-668134054. I would have loved to link it directly, but this get rejected by commitlint.

analyzer/src/main/kotlin/managers/GoMod.kt
fviernau: Beyond that, I managed to split-out changed to this function into a separate preceding commit.

analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
```diff
@@ -95,10 +96,9 @@ class GoModFunTest : StringSpec({
     "Local module dependencies make the analysis fail" {
         // TODO: Implement support for local dependencies, see https://github.com/oss-review-toolkit/ort/issues/7649.
         val definitionFile = testDir.resolve("gomod-submodules/app/go.mod")
-        val expectedResultFile = testDir.resolve("gomod-submodules-embed-expected-output.yml")
 
         val result = create("GoMod").resolveSingleProject(definitionFile)
 
-        result.withInvariantIssues().toYaml() should matchExpectedResult(expectedResultFile, definitionFile)
+        result.issues.size shouldBe 1
```
sschuberth, line 102: Nit (sorry for not realizing earlier): Could also be `shouldHaveSize 1`.

sschuberth (approved):

analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
sschuberth: I realize we have more such code that could be changed, so I'll probably bulk-change after the merge of this.

sschuberth: See https://github.com/oss-review-toolkit/ort/pull/7716.

PR #1266, opened 2019-02-05
Do not override toString() for the Identifier anymore

sschuberth:
Please have a look at the individual commit messages for the details.

---
This change is [Reviewable](https://reviewable.io/reviews/heremaps/oss-review-toolkit/1266)

PR #7715, opened 2023-10-19
chore(model): Improve `Identifier.toString()`

fviernau:
The string returned by `toCoordinates()` seems nicer than the default implementation.

### DIFF
--- a/model/src/main/kotlin/Identifier.kt
+++ b/model/src/main/kotlin/Identifier.kt
@@ -124,4 +124,6 @@ data class Identifier(
      */
     fun toPath(separator: String = "/", emptyValue: String = "unknown"): String =
         sanitizedComponents.joinToString(separator) { it.encodeOr(emptyValue) }
+
+    override fun toString(): String = toCoordinates()
 }


model/src/main/kotlin/Identifier.kt
```diff
@@ -124,4 +124,6 @@ data class Identifier(
      */
     fun toPath(separator: String = "/", emptyValue: String = "unknown"): String =
         sanitizedComponents.joinToString(separator) { it.encodeOr(emptyValue) }
+
+    override fun toString(): String = toCoordinates()
```
sschuberth, line 128: I don't remember the exactly details, but I believe we had reasons in the past *not* to do this, as it breaks / might break things in a very subtle way. Maybe @mnonnenmacher remembers?

mnonnenmacher: I don't remember if this could cause any issues, but a quick search lead me to this commit:
https://github.com/oss-review-toolkit/ort/pull/1266/commits/a082f0c527ec333693b5b8ebb356660bff08a6e9

fviernau: So, @mnonnenmacher @sschuberth,

Can you shed a light on what you think the implications of above linked commit are on this PR? (I don't get it)

sschuberth: I believe we generally agreed at some point that we should not / only *very* sparsely override `toString()` for *any* class, as it's very hard to foresee the side effects when such classes are used e.g. in string interpolation (`"$id"`). Such places are usually not found during refactoring / when searching for uses of `toString()`.

And actually, IIRC, it was esp. you @fviernau who voted to not override `toString()` for these reasons. So I'm a bit surprised that you seem to have changed your mind here. Don't get me wrong, it's completely fine to change one's mind for a good reason, but I don't get what the good reason here is that outweighs above concerns.

fviernau: > And actually, IIRC, it was esp. you @fviernau who voted to not override `toString()` for these reasons.

It's already a while ago and I do not remember exactly how I phrased it anymore. However, my concern was 
as I recall it was primarily about code relying on `toString()` implementations, not the `toString()` implementations themselves. When filing this PR I though it would be fine if e.g. only log output made use of it.

However, thinking about it again I agree that a dedicated function is better as (like you already said) enables finding all places making use of it.

fviernau:
As per above discussion, using dedicated functions is preferrable.

PR #7718, opened 2023-10-19
build(application): Set the "Main-Class" attribute

sschuberth:
This is not actually done by Gradle's application plugin, so set it manually to allow the JAR to be run via `java -jar ...`.

### DIFF
--- a/buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
@@ -177,3 +177,7 @@ tasks.named<CreateStartScripts>("startScripts") {
         )
     }
 }
+
+tasks.withType<Jar>().configureEach {
+    manifest.attributes["Main-Class"] = application.mainClass
+}


sschuberth:
Not really needed as dependent JARs would not be on the classpath anyway.

buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
```diff
@@ -177,3 +177,7 @@ tasks.named<CreateStartScripts>("startScripts") {
         )
     }
 }
+
+tasks.withType<Jar>().configureEach {
+    manifest.attributes["Main-Class"] = application.mainClass
```
fviernau, line 182: Is it actually supported to run ORT in this way (apart from this entry)? Or would ORT run into further issues ?

sschuberth: See [above](https://github.com/oss-review-toolkit/ort/pull/7718#issuecomment-1771646498).

PR #7719, opened 2023-10-19
analyzer: Support references to local modules with `GoMod`

### LINKED ISSUE #7649: GoMod: Support for local / sub module dependencies

fviernau:
The only way of referencing a module on the local file system from another is by specifying a relative or absolute path via a replace directive. This is currently not supported by ORT's Go Modules integration.

Implement that support, so that a reference inbetween two modules on the local file system gets represent as a `PackageReference` between the corresponding `Project`s (in ORT speak). However, keep on failing in case a referenced module is not located within the analysis root, because that would violate ORT's package manager API.

Fixes #7649.

### DIFF
--- a/analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
+++ b/analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
@@ -20,9 +20,11 @@
 package org.ossreviewtoolkit.analyzer.managers
 
 import io.kotest.core.spec.style.StringSpec
-import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.should
 
+import java.io.File
+
+import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.toYaml
 import org.ossreviewtoolkit.utils.test.getAssetFile
 import org.ossreviewtoolkit.utils.test.matchExpectedResult
@@ -93,12 +95,28 @@ class GoModFunTest : StringSpec({
         result.toYaml() should matchExpectedResult(expectedResultFile, definitionFile)
     }
 
-    "Local module dependencies make the analysis fail" {
-        // TODO: Implement support for local dependencies, see https://github.com/oss-review-toolkit/ort/issues/7649.
-        val definitionFile = testDir.resolve("gomod-submodules/app/go.mod")
+    "Project dependencies with a (relative) local module dependency are detected correctly" {
+        val definitionFileApp = testDir.resolve("gomod-submodules/app/go.mod")
+        val definitionFileUtils = testDir.resolve("gomod-submodules/utils/go.mod")
+        val expectedResultFile = testDir.resolve("gomod-submodules-embed-expected-output.yml")
+        val expectedDefinitionFilePathUtils = getDefinitionFilePath(definitionFileUtils)
 
-        val result = create("GoMod").resolveSingleProject(definitionFile)
+        val result = create("GoMod").collateMultipleProjects(definitionFileApp, definitionFileUtils)
 
-        result.issues shouldHaveSize 1
+        result.withResolvedScopes().toYaml() should matchExpectedResult(
+            expectedResultFile,
+            definitionFileApp,
+            custom = mapOf(
+                "<REPLACE_DEFINITION_FILE_PATH_UTILS>" to expectedDefinitionFilePathUtils,
+                "<REPLACE_PATH_UTILS>" to expectedDefinitionFilePathUtils.substringBeforeLast('/')
+            )
+        )
     }
 })
+
+private fun getDefinitionFilePath(definitionFile: File): String {
+    val projectDir = definitionFile.parentFile
+    val vcsDir = checkNotNull(VersionControlSystem.forDirectory(projectDir))
+    val path = vcsDir.getPathToRoot(projectDir)
+    return "$path/${definitionFile.name}"
+}
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -220,15 +220,20 @@ class GoMod(
 
         return if (version.isBlank()) {
             // If the version is blank, it is a project in ORT speak.
-            check(main) { "Found a local module dependency which is not supported." }
-
             checkNotNull(dir) { "For projects, the directory is expected to not be null." }
+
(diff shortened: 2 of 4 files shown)


analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
```diff
@@ -22,10 +22,12 @@ package org.ossreviewtoolkit.analyzer.managers
 import io.kotest.core.spec.style.StringSpec
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.should
+import org.ossreviewtoolkit.downloader.VersionControlSystem
```
sschuberth, line 25: Wrong import ordering (also below).

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -220,15 +220,20 @@ class GoMod(
 
         return if (version.isBlank()) {
             // If the version is blank, it is a project in ORT speak.
-            check(main) { "Found a local module dependency which is not supported." }
-
             checkNotNull(dir) { "For projects, the directory is expected to not be null." }
 
+            val projectDir = File(dir)
+
+            require(projectDir.absoluteFile.startsWith(analysisRoot.absoluteFile)) {
```
sschuberth: As `analysisRoot` is absolute by convention, you can drop `.absoluteFile` here.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -220,15 +220,20 @@ class GoMod(
...
+                "A replace directive references a module in '${projectDir.absolutePath}' outside of analysis root " +
```
sschuberth: How about using `val projectDir = File(dir).absoluteFile` in line 225 so you can just use `$projectDir` here, and also simplify line 227?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -241,6 +246,19 @@ class GoMod(
         }
     }
 
+    private fun getProjectName(projectDir: File): String {
+        projectDir.resolve("go.mod").let { goModFile ->
```
sschuberth: Consider using `also` instead of `let` as the return value does not matter.

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -241,6 +246,19 @@ class GoMod(
...
+            require(goModFile.isFile) {
+                "Expected file '$goModFile.' which does not exist."
```
sschuberth: Is the dot after the variable intended?

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -241,6 +246,19 @@ class GoMod(
...
+            }
+        }
+
+        val list = runGo("list", "-m", "-json", "-buildvcs=false", workingDir = projectDir)
+
+        return list.stdout.byteInputStream().use { JSON.decodeToSequence<ModuleInfo>(it) }.single().path
+    }
+
+
```
sschuberth: One superfluous blank line.

analyzer/src/testFixtures/kotlin/Extensions.kt
```diff
@@ -63,19 +64,21 @@ fun PackageManager.resolveSingleProject(definitionFile: File, resolveScopes: Boo
...
 /**
- * Resolve the dependencies of a [definitionFile] which should create at least one project. All created projects will be
- * collated in an [AnalyzerResult] with their dependency graph.
+ * Resolve the dependencies of all [definitionFiles] which should create at least one project. All created projects will
+ * be collated in an [AnalyzerResult] with their dependency graph.
  */
-fun PackageManager.collateMultipleProjects(definitionFile: File): AnalyzerResult {
-    val managerResult = resolveDependencies(listOf(definitionFile), emptyMap())
+fun PackageManager.collateMultipleProjects(vararg definitionFiles: File): AnalyzerResult {
+    val managerResult = resolveDependencies(definitionFiles.toList(), emptyMap())
```
sschuberth, line 71: Prefer `.asList()` to avoid a conversion.

sschuberth (changes requested):

sschuberth (dismissed):

analyzer/src/funTest/kotlin/managers/GoModFunTest.kt
```diff
@@ -93,12 +95,28 @@
...
+                "<REPLACE_DEFINITION_FILE_PATH_UTILS>" to expectedDefinitionFilePathUtils,
+                "<REPLACE_PATH_UTILS>" to expectedDefinitionFilePathUtils.substringBeforeLast('/')
+            )
+        )
     }
 })
+
+private fun getDefinitionFilePath(definitionFile: File): String {
+    val projectDir = definitionFile.parentFile
+    val vcsDir = VersionControlSystem.forDirectory(projectDir)!!
```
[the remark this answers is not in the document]
fviernau: If pushed once more to fix this.

sschuberth (approved):

PR #7723, opened 2023-10-20
Resolve plugin config vs classpath exceptions

sschuberth:


oheger-bosch (approved):
The change looks good to me. Just wondering whether other plugin types could be affected as well. The commit message is rather generic about "provider plugins".

sschuberth:
> The change looks good to me. Just wondering whether other plugin types could be affected as well. The commit message is rather generic about "provider plugins".

AFAICT, other plugins do not get implicitly created by coming across their configuration, but by filtering enabled ones from those in the classpath.

PR #7724, opened 2023-10-20
Some small chores

sschuberth:


analyzer/src/funTest/kotlin/PackageManagerFunTest.kt
```diff
@@ -85,7 +85,7 @@ class PackageManagerFunTest : WordSpec({
 
             // The test project contains at least one file per package manager, so the result should also contain an
             // entry for each package manager.
-            val unmanagedPackageManagerFactory = PackageManager.ALL["Unmanaged"]
+            val unmanagedPackageManagerFactory = PackageManager.ALL.getValue("Unmanaged")
```
fviernau, line 88: out of scope: For this very same reason, we could finally get rid of all the `shouldNotBeNull` precondition assertions to drop some boiler plate in favor of just failing if it's not null.

fviernau (approved):

PR #7746, opened 2023-10-26
TrustSource reporter: update the data model and the reporter

gr-markin:


plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
```diff
@@ -22,34 +22,53 @@ package org.ossreviewtoolkit.plugins.reporters.trustsource
 import kotlinx.serialization.Serializable
 
 @Serializable
-data class TrustSourceModule(
+data class NewScan(
```
sschuberth, line 25: Commit message: Please use a title that adheres to [Conventional Commits](https://www.conventionalcommits.org/).

plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
```diff
@@ -22,34 +22,53 @@ package org.ossreviewtoolkit.plugins.reporters.trustsource
...
+    /** Name of the module */
     val module: String,
-    val moduleId: String,
-
-    val dependencies: List<TrustSourceDependency>
+    /** SCM tag of the module */
+    val tag: String = "",
+    /** SCM branch of the module */
+    val branch: String = "",
```
sschuberth, line 31: Would it be valid if both `tag` and `branch` are empty? If not, the docs should probably say that either or is required. Also, what if only a Git commit SHA1 is known, where would that go?

plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
```diff
@@ -22,34 +22,53 @@ package org.ossreviewtoolkit.plugins.reporters.trustsource
...
+    val purl: String,
+    val dependencies: List<Dependency>,
+
+    val description: String = "",
+    /** Package homepage, may differ or may be the same with the repositoryUrl **/
+    val homepageUrl: String = "",
+    /** Repository with the package's source code **/
+    val repositoryUrl: String = "",
+
+    /** Map where a key denotes a checksum algorithm and the value is the checksum itself **/
```
sschuberth, line 48: Can you be more specific about valid algorithm names? E.g. should it be "SHA1" or "SHA-1"? Ideally, the key would be an enum, not a string, so this becomes clear.

plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
```diff
@@ -22,34 +22,53 @@ package org.ossreviewtoolkit.plugins.reporters.trustsource
...
+    val checksum: Map<String, String> = emptyMap(),
+    val licenses: List<License> = emptyList(),
+
+    /** Dependencies source package **/
+    val `package`: Package? = null,
```
sschuberth, line 53: For ease of referencing the property, we could name it "pkg" but have a `@SerialName("package")` annotation.

plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
```diff
@@ -22,34 +22,53 @@ package org.ossreviewtoolkit.plugins.reporters.trustsource
...
 
 @Serializable
-class TrustSourceMeta
+data class Package(
+    /** Path to the license file within the package if available **/
+    val licenseFile: String = "",
+    /** Download URL to the source code package, e.g. Maven jars with sources **/
+    val sourcesUrl: String = "",
+    /** Checksum of the source code package **/
+    val sourcesChecksum: Map<String, String> = emptyMap()
```
sschuberth, line 73: Same here about the algorithm becoming an enum.

plugins/reporters/trustsource/src/main/kotlin/TrustSourceReporter.kt
```diff
@@ -69,27 +67,34 @@ private fun traverseDeps(input: ReporterInput, deps: Sequence<DependencyNode>):
             input.ortResult.getPackageLicenseChoices(dep.id),
             input.ortResult.getRepositoryLicenseChoices()
         )
+
         val licenses = effectiveLicense?.decompose()?.map {
             val name = it.toString()
             val url = it.getLicenseUrl().orEmpty()
 
-            TrustSourceLicense(name, url)
+            License(name, url = url)
```
sschuberth, line 75: Nit: There is no gain in using a named `url =` argument here, and it should be simply omitted.

sschuberth (changes requested):
Thanks @gr-markin for the update!

plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
gr-markin: Got it !

plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
gr-markin: Added comments and the commit field, thanks for the advise. These fields are normally filled by a developer submitting a scan of a module, hence nobody asked before to provide this info.

plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
gr-markin: It can be any name that is used by a package management system. Because scans are submitted over the RestAPI from  different sources, we have to check it on the server anyway, hence I just use the string representation.

plugins/reporters/trustsource/src/main/kotlin/TrustSourceModel.kt
gr-markin: Done

plugins/reporters/trustsource/src/main/kotlin/TrustSourceReporter.kt
gr-markin: Done

gr-markin:
Hi @sschuberth  thanks you for your review and advices. I pushed a new commit with changes with a title in the commit message, hope it's ok. Thanks in advance.

sschuberth (approved):

PR #7751, opened 2023-10-27
Support the CycloneDX vulnerability extension in Reporter

### LINKED ISSUE #4531: Support the CycloneDX vulnerability extension
The CycloneDX XML and JSON schemas have [extensions](https://github.com/CycloneDX/specification/tree/master/schema/ext) to also capture known vulnerabilities in CycloneDX itself. We should extend our reporter to support this when an advisor was run.

geoandri:
This PR adds support for CycloneDX vulnerabilities in Reporter.

Closes #4531

### DIFF
--- a/plugins/reporters/cyclonedx/src/funTest/assets/cyclonedx-reporter-expected-result.xml
+++ b/plugins/reporters/cyclonedx/src/funTest/assets/cyclonedx-reporter-expected-result.xml
@@ -15,7 +15,7 @@
     </licenses>
   </metadata>
   <components>
-    <component type="library">
+    <component type="library" bom-ref="NPM:@ort:concluded-license:1.0">
       <group>@ort</group>
       <name>concluded-license</name>
       <version>1.0</version>
@@ -49,7 +49,7 @@
       <modified>false</modified>
       <externalReferences><reference type="website"><url>https://github.com/oss-review-toolkit/ort</url></reference></externalReferences><ort:dependencyType xmlns:ort="http://www.w3.org/1999/xhtml">direct</ort:dependencyType>
     </component>
-    <component type="library">
+    <component type="library" bom-ref="NPM:@ort:declared-license:1.0">
       <group>@ort</group>
       <name>declared-license</name>
       <version>1.0</version>
@@ -72,7 +72,7 @@
       <modified>false</modified>
       <externalReferences><reference type="website"><url>https://github.com/oss-review-toolkit/ort</url></reference></externalReferences><ort:dependencyType xmlns:ort="http://www.w3.org/1999/xhtml">direct</ort:dependencyType>
     </component>
-    <component type="library">
+    <component type="library" bom-ref="NPM:@ort:license-file:1.0">
       <group>@ort</group>
       <name>license-file</name>
       <version>1.0</version>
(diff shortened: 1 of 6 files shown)


plugins/reporters/cyclonedx/src/funTest/assets/cyclonedx-reporter-expected-result.xml
```diff
@@ -131,4 +131,23 @@
       <externalReferences><reference type="website"><url>https://github.com/oss-review-toolkit/ort</url></reference></externalReferences><ort:dependencyType xmlns:ort="http://www.w3.org/1999/xhtml">direct</ort:dependencyType>
     </component>
   </components>
+  <vulnerabilities>
+    <vulnerability>
+      <id>CVE-2021-1234</id>
+      <ratings>
+        <rating>
```
sschuberth, line 138: Any idea why no severity value (should be "MEDIUM") shows up here?

sschuberth (changes requested):
Thanks a lot for the contribution!

plugins/reporters/cyclonedx/src/main/kotlin/CycloneDxReporter.kt
```diff
@@ -252,6 +257,30 @@ class CycloneDxReporter : Reporter {
         return outputFiles
     }
 
+    private fun addVulnerabilitiesToBom(input: ReporterInput, bom: Bom) {
```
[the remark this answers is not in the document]
geoandri: I think an argument like `advisorVulnerabilities:  Map<Identifier, List<Vulnerability>>` holds the data that's needed, since I also need the identifier to populate the `affects` field with [references](https://cyclonedx.org/docs/1.5/json/#vulnerabilities_items_affects_items_ref) pointing to the `bom-ref` components fields.

plugins/reporters/cyclonedx/src/funTest/assets/cyclonedx-reporter-expected-result.xml
geoandri: Yes, it's because of the `fromString(String text)` method in [CycloneDX java core](https://github.com/CycloneDX/cyclonedx-core-java/blob/master/src/main/java/org/cyclonedx/model/vulnerability/Vulnerability.java#L376) library, which is case-sensitive. Should I always pass the lowercase value e.g. 
```kotlin
Severity.fromString(reference.severity?.lowercase())
```

plugins/reporters/cyclonedx/src/main/kotlin/CycloneDxReporter.kt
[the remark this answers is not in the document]
sschuberth: Right! In essence, I was meaning to say to make the function take the return value of `input.ortResult.getVulnerabilities()` instead of the whole `input`, to make more clear that the implementation really only cares about vulnerability data.

plugins/reporters/cyclonedx/src/funTest/assets/cyclonedx-reporter-expected-result.xml
sschuberth: Let me quickly refactor some code on our side to provide better convenience functions for this use-case. I'll keep you posted.

sschuberth: Seems like you figured out yourself already how to make use of the refactored code, great! 🚀

sschuberth (approved):
Looks good to me now, thanks a lot for the contribution!

sschuberth:

PR #7752, opened 2023-10-27
fix(AdvisorRecord): Merge all properties of vulnerabilities

sschuberth:
Do not leave the `summary` and `description` at their `null` default values but take them from the first vulnerability. Update the documentation accordingly and also generally improve the wording while at it.

### DIFF
--- a/model/src/main/kotlin/AdvisorRecord.kt
+++ b/model/src/main/kotlin/AdvisorRecord.kt
@@ -121,8 +121,8 @@ data class AdvisorRecord(
 }
 
 /**
- * Merge this list of [Vulnerability] objects by combining vulnerabilities with the same ID and merging their
- * references.
+ * Merge this collection of [Vulnerability] objects by combining vulnerabilities with the same ID and merging their
+ * references. Other [Vulnerability] properties are taken from the first object which has any such property set.
  */
 private fun Collection<Vulnerability>.mergeVulnerabilities(): List<Vulnerability> {
     val vulnerabilitiesById = groupByTo(sortedMapOf()) { it.id }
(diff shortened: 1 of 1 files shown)


model/src/main/kotlin/AdvisorRecord.kt
```diff
@@ -121,19 +121,20 @@ data class AdvisorRecord(
...
- * Merge this (non-empty) list of [Vulnerability] objects (which are expected to have the same ID) by to a single
- * [Vulnerability] that contains all the references from the source vulnerabilities (with duplicates removed).
+ * Merge this (non-empty) collection of [Vulnerability] objects (which are expected to have the same ID) to a single
+ * [Vulnerability] that contains all the references from the original vulnerabilities (with duplicates removed). Other
+ * [Vulnerability] properties are arbitrarily taken from the first object.
  */
 private fun Collection<Vulnerability>.mergeReferences(): Vulnerability {
     val references = flatMapTo(mutableSetOf()) { it.references }
-    return Vulnerability(id = first().id, references = references.toList())
+    return first().copy(references = references.toList())
```
mnonnenmacher, line 139: How about searching for an instance that has a summary or description in case some have those values but the first has not? Something like `find { it.summary != null || it.description != null } ?: first()`.

sschuberth: Good idea, done.

model/src/main/kotlin/AdvisorRecord.kt
```diff
@@ -121,19 +121,21 @@ data class AdvisorRecord(
 }
 
 /**
- * Merge this list of [Vulnerability] objects by combining vulnerabilities with the same ID and merging their
- * references.
+ * Merge this collection of [Vulnerability] objects by combining vulnerabilities with the same ID and merging their
+ * references. Other [Vulnerability] properties are arbitrarily taken from the first object.
```
mnonnenmacher, line 125: This sentence needs to be updated as for the function below.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #7754, opened 2023-10-27
docs(configuration): Improve docs for curations in `.ort.yml`

sschuberth:


website/docs/configuration/ort-yml.md
```diff
@@ -85,7 +85,7 @@ The path exclude above has the following effects:
 
 * All projects found below the `test-data` directory are marked as excluded.
 * License findings in files below the `test-data` directory are marked as excluded. This can be used in
-  [evaluator rules](../../tutorial/intro#6-running-the-evaluator) to for instance change the severity from error to
+  [evaluator rules](../../tutorial/intro.md#6-running-the-evaluator) to for instance change the severity from error to
```
mnonnenmacher, line 88: The link on the website works just fine if you test it here: https://oss-review-toolkit.org/ort/docs/configuration/ort-yml#excluding-paths

Because the tutorial is a separate instance of the docs plugin currently URL links have to be used, see the warning at the bottom of [this page](https://docusaurus.io/docs/markdown-features/links) or [this comment](https://github.com/facebook/docusaurus/issues/9117#issuecomment-1617569344).

mnonnenmacher (changes requested):

website/docs/configuration/ort-yml.md
```diff
@@ -6,7 +6,9 @@ increased degree of automation and local configurations should only be done if t
 
 * [excludes](#excludes) - Mark [files, directories](#excluding-paths) or [package manager scopes](#excluding-scopes) as
   not included in released artifacts.
-* [curations](#curations) - Overwrite package metadata, set a concluded license or correct license findings.
+* [curations](#curations) - Override package metadata, set a concluded license or correct license findings. Needs to be
+  enabled in global ORT configuration ([config.yml](../getting-started/usage.md#ort-configuration-file)) via
+  `enableRepositoryPackageCurations: true`.
```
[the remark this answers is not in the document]
sschuberth: Those admonitions look nice, but I'm a bit concerned that they don't render correctly when indented in a list. Is there a way to preview this? Neither IntelliJ nor GitHub seem to be able to understand this syntax.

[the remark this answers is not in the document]
mnonnenmacher: My suggestion was not to add the admonition to the list, but to add it to the beginning of the curations section. Unfortunately I don't know a way to preview this except for running the site locally.

website/docs/configuration/ort-yml.md
mnonnenmacher: I just thought that we could merge the tutorial into the docs to avoid this issue by having only one instance of the docs plugin, what do you think?

website/docs/configuration/ort-yml.md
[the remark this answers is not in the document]
sschuberth: > My suggestion was not to add the admonition to the list, but to add it to the beginning of the curations section.

Hmm, I'm not sure I like that, if I'm understanding correctly what you mean. As enabling is only needed for this specific feature, I believe that note should stay here. And I don't want to repeat too much of that note at the section level either.

Does Docuaurus also support GitHub-style admonition syntax like

```
> **Note**
> Some text
```

which renders like

> **Note**
> Some text

?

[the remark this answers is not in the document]
sschuberth: > Does Docuaurus also support GitHub-style admonition syntax like

Not (yet), it seems: https://github.com/facebook/docusaurus/issues/7471

website/docs/configuration/ort-yml.md
sschuberth: > only one instance of the docs plugin

Sorry, which docs plugin?

Anyway, you're the expert here, so whatever you say 😆 I was only addressing this as [the markdown-links check failed](https://github.com/oss-review-toolkit/ort/actions/runs/6668910751/job/18125541974#step:4:41).

website/docs/configuration/ort-yml.md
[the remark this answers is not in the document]
sschuberth: In any case, I'd like to avoid syntax that's not valid in both (GitHub flavored) Markdown and Docusaurus. Just in case we ever switch so something else again, I'd like to avoid running into a "vendor lock-in".

website/docs/configuration/ort-yml.md
```diff
@@ -6,7 +6,9 @@ increased degree of automation and local configurations should only be done if t
 
 * [excludes](#excludes) - Mark [files, directories](#excluding-paths) or [package manager scopes](#excluding-scopes) as
   not included in released artifacts.
-* [curations](#curations) - Overwrite package metadata, set a concluded license or correct license findings.
+* [curations](#curations) - Override package metadata, set a concluded license or correct license findings. Needs to be
```
tsteenbe, line 9: Please change override back to overwrite as actions and decisions are overriden, information is overwritten.

tsteenbe (dismissed):

website/docs/configuration/ort-yml.md
sschuberth: But doesn't "overwritten" imply that the original information is not available anymore? That's not the case here, and I actually deliberately chose "overridden" to reflect that.

website/docs/configuration/ort-yml.md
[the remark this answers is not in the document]
mnonnenmacher: > Hmm, I'm not sure I like that, if I'm understanding correctly what you mean. As enabling is only needed for this specific feature, I believe that note should stay here. And I don't want to repeat too much of that note at the section level either.

I mean to put the admonition at the beginning of this section which describes the usage of curations in the `.ort.yml`. In the list at the beginning I think it's very easy to miss that important information, especially when navigating via the TOC on the right side: https://oss-review-toolkit.org/ort/docs/configuration/ort-yml#curations

> In any case, I'd like to avoid syntax that's not valid in both (GitHub flavored) Markdown and Docusaurus. Just in case we ever switch so something else again, I'd like to avoid running into a "vendor lock-in".

I wouldn't see this as a problematic vendor lock-in, it's easy to switch to another syntax if necessary and I don't expect that we will use too many admonitions. However, I tried [this remark plugin](https://github.com/myl7/remark-github-beta-blockquote-admonitions) to enable the GitHub syntax and I could get it to compile with [this hint](https://github.com/facebook/docusaurus/issues/5379#issuecomment-1525220908), but for some reason it didn't work. It's probably worth to wait a bit until Docusaurus 3 is released and then try again after the upgrade.

website/docs/configuration/ort-yml.md
mnonnenmacher: > Sorry, which docs plugin?

The docs plugin for docusaurus, we currently use two instances of the plugin, one for the docs and one for the tutorial. So links between those two must be URL links.

> Anyway, you're the expert here, so whatever you say 😆 I was only addressing this as the markdown-links check failed.

I will try to find the time to make that change in a separate PR, but until then the second commit should be removed as to my understanding it would actually break the link on the website. I wonder why the markdown lint check was failing, if this link is an issue it should also have failed before?

sschuberth: > we currently use two instances of the plugin, one for the docs and one for the tutorial.

That sounds weird... and I wasn't aware of that. Thanks for pointing that out.

> I wonder why the markdown lint check was failing, if this link is an issue it should also have failed before?

Correct. Apparently some caching takes place until the particular file is touched in a PR...

website/docs/configuration/ort-yml.md
mnonnenmacher: To me "overwrite" also sounds like the better fit here because we are replacing metadata (even if we document that it was overwritten by recording the applied curations), I checked the translations for "override" and they don't seem like a good fit to me.

sschuberth: Ok, I'll drop that hunk.

website/docs/configuration/ort-yml.md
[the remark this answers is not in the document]
sschuberth: > I mean to put the admonition at the beginning of this section which describes the usage of curations in the `.ort.yml`.

Ok, I hope to now have it like you suggest.

sschuberth:
I ended up rewriting / reordering larger parts than planned as I found the docs to be more and more lacking the more often I read it.

website/docs/configuration/ort-yml.md
```diff
@@ -6,8 +6,10 @@ increased degree of automation and local configurations should only be done if t
 
 * [excludes](#excludes) - Mark [files, directories](#excluding-paths) or [package manager scopes](#excluding-scopes) as
   not included in released artifacts.
-* [curations](#curations) - Overwrite package metadata, set a concluded license or correct license findings.
 * [resolutions](#resolutions) - Resolve any issues or policy rule violations.
+* [curations](#curations) - Overwrite package metadata, set a concluded license, or correct license findings in the
+  project.
+* [package configurations](#package-configurations) - Define path excludes or correct license finding in dependencies.
```
mnonnenmacher, line 12: Plural "finding**s**".

mnonnenmacher (changes requested):

website/docs/configuration/ort-yml.md
mnonnenmacher: I think the ignore pattern in the `mlc_config.json` needs to be changed to `"^../../tutorial/intro.*$"` to silence the markdown link check issue.

sschuberth: I gave it a try (with a slightly more specific pattern).

mnonnenmacher (approved):

PR #7761, opened 2023-10-30
Minor gomod improvements

fviernau:
See individual commits.

### DIFF
--- a/analyzer/src/main/kotlin/managers/GoMod.kt
+++ b/analyzer/src/main/kotlin/managers/GoMod.kt
@@ -209,6 +209,64 @@ class GoMod(
         return graph.breakCycles()
     }
 
+    /**
+     * Return the list of all modules available via the `go list` command.
+     */
+    private fun getModuleInfos(projectDir: File): List<ModuleInfo> {
+        val list = runGo("list", "-m", "-json", "-buildvcs=false", "all", workingDir = projectDir)
+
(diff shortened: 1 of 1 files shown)


analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -390,6 +388,8 @@ private const val PACKAGE_SEPARATOR = "# "
  */
 private const val WHY_CHUNK_SIZE = 32
 
+private val JSON: Json by lazy { Json { ignoreUnknownKeys = true } }
```
sschuberth, line 363: Nit: Creating this is cheap, and it's always required if the GoMod analyzer is used, so I'm unsure whether using `lazy` is beneficial here (also, it was not lazy before).

analyzer/src/main/kotlin/managers/GoMod.kt
```diff
@@ -390,6 +362,18 @@ private const val WHY_CHUNK_SIZE = 32
 
 private val JSON: Json by lazy { Json { ignoreUnknownKeys = true } }
 
+private const val DEFAULT_GO_PROXY = "https://proxy.golang.org"
```
sschuberth, line 365: Nit: This also moves this property (not only functions) and makes it private. For completeness, that should be mentioned as part of the commit message, IMO.

sschuberth (approved):

PR #7762, opened 2023-10-30
Vulnerability refactorings

sschuberth:


model/src/main/kotlin/vulnerabilities/VulnerabilityReference.kt
```diff
@@ -63,4 +63,9 @@ data class VulnerabilityReference(
                 else -> severity?.uppercase()?.takeIf { it in CVSS3_SEVERITIES }
             } ?: "UNKNOWN"
     }
+
+    /**
+     * Return a human-readable severity rating string.
+     */
+    val severityRating: String by lazy { getSeverityString(scoringSystem, severity) }
```
sschuberth, line 70: Note that this will serialize the rating (like "MEDIUM") as part of the advisor result, which slightly increases that section's size. If that's a problem I could use a member function instead, but then this class could not be used directly by the `EvaluatedModelMapper` anymore.

mnonnenmacher: How does Jackson handle lazy properties during deserialization? Are the serialized values ignored or are they assigend to the property somehow? The latter could be a problem if the function implementation changes.

mnonnenmacher: Is the motivation of this change only to get rid of the `EvaluatedVulnerabilityReference` or is there another reason to have the value in the ORT result? Because I'm not sure if I would like to align our default serialization to the requirements of the web app reporter.

sschuberth: > How does Jackson handle lazy properties during deserialization?

They are ignored.

> Is the motivation of this change only to get rid of the `EvaluatedVulnerabilityReference`

Actually, this wasn't my original motivation at all. I just wanted to have a convenience function (or now property) to get the severity rating string for a `VulnerabilityReference`, and then realized that this property is the only difference to `EvaluatedVulnerabilityReference`, so I also got rid of that class as a side effect.

mnonnenmacher (approved):

model/src/main/kotlin/vulnerabilities/VulnerabilityReference.kt
sschuberth: > > How does Jackson handle lazy properties during deserialization?
> 
> They are ignored.

My statement is false. Not sure what went wrong in my test, but lazy properties need to be explicitly ignored during deserialization, see https://github.com/oss-review-toolkit/ort/pull/7789.

PR #7765, opened 2023-10-30
Update VulnerableCode property names

sschuberth:


mnonnenmacher (approved):

nnobelis:
@sschuberth : This is a breaking change since the renamed properties are part of the public API.

sschuberth:
> @sschuberth : This is a breaking change since the renamed properties are part of the public API.

You're right, I'll label it as such. This is the downside of using data class models directly in the signature of the interface functions...

PR #7773, opened 2023-10-31
go: Minor `funTest` improvements

fviernau:


plugins/package-managers/go/src/funTest/kotlin/GoDepFunTest.kt
```diff
@@ -42,7 +42,7 @@ class GoDepFunTest : WordSpec({
     "GoDep" should {
         "resolve dependencies from a lockfile correctly" {
             val definitionFile = getAssetFile("projects/synthetic/godep/lockfile/Gopkg.toml")
-            val expectedResultFile = getAssetFile("projects/synthetic/godep-expected-output.yml")
+            val expectedResultFile = getAssetFile("projects/synthetic/lockfile-expected-output.yml")
```
sschuberth: I'm not sure about this, as this looses the information about which package manager this file belongs to. Maybe just use "lockfile" as an infix?

plugins/package-managers/go/src/funTest/kotlin/GoModFunTest.kt
```diff
@@ -37,7 +37,7 @@ class GoModFunTest : StringSpec({
 
     "Project dependencies are detected correctly" {
         val definitionFile = testDir.resolve("gomod/gomod/go.mod")
-        val expectedResultFile = testDir.resolve("gomod-expected-output.yml")
+        val expectedResultFile = testDir.resolve("gomod/gomod-expected-output.yml")
```
sschuberth: This file should also get the "gomod-" prefix stripped then.

plugins/package-managers/go/src/funTest/kotlin/GoDepFunTest.kt
sschuberth: > this looses the information about which package manager this file belongs to.

Ah, you're solving that in the next commit by introducing subdirectories, ok.

sschuberth (changes requested):

plugins/package-managers/go/src/funTest/kotlin/GoModFunTest.kt
fviernau: For this one the `gomod` prefix was kept intentionally, because the prefixes in the `*-expected-output.yml` files are aligned now with the names of the test projects (dirs).

sschuberth: Would it make sense to rename the inner "gomod" directory then? All other directories are named after the case they test, like "no-deps" etc. So maybe use just "project" as the directory name to somewhat match the "Project dependencies are detected correctly" test title?

fviernau: I've added one commit on top for a better name.

plugins/package-managers/go/src/funTest/kotlin/GoModFunTest.kt
```diff
@@ -35,9 +35,9 @@ import org.ossreviewtoolkit.utils.test.matchExpectedResult
 class GoModFunTest : StringSpec({
     val testDir = getAssetFile("projects/synthetic/gomod")
 
-    "Project dependencies are detected correctly" {
+    "Project dependencies are detected correctly given a project with test dependencies" {
         val definitionFile = testDir.resolve("gomod/go.mod")
-        val expectedResultFile = testDir.resolve("gomod-expected-output.yml")
+        val expectedResultFile = testDir.resolve("project-with-test-deps-expected-output.yml")
```
sschuberth, line 40: Hmm, the directory is now called "project-with-tests", so shouldn't this omit "deps-" and be named "project-with-tests-expected-output.yml"? Also, in line 39 above, shouldn't it be "project-with-tests/go.mod"?

sschuberth (approved):

sschuberth:
@fviernau I'd be fine with merging this and to resolve the unrelated CI issues later.

PR #7782, opened 2023-11-01
Upgrade Node.js to version 20

fviernau:
See individual commits.


plugins/package-managers/node/src/funTest/kotlin/NpmFunTest.kt
```diff
@@ -97,12 +96,9 @@ class NpmFunTest : WordSpec({
...
 
-            result.issues shouldHaveSize 1
-            with(result.issues.first()) {
-                source shouldBe "NPM"
-                severity shouldBe Severity.ERROR
-                message shouldContain "Unexpected token \"<\" (0x3C) in JSON at position 0 while parsing \"<>\""
-            }
+            result.issues.filter {
+                it.source == "NPM" && it.severity == Severity.ERROR && it.message.contains("Unexpected token")
+            } shouldHaveSize 1
```
sschuberth, line 101: How about using the `shouldHaveSingleElement { condition }` matcher? Also, prefer the `in` operator over `contains()`.

.versions
```diff
@@ -9,8 +9,8 @@ GO_DEP_VERSION=0.5.4
 GO_VERSION=1.21.1
 HASKELL_STACK_VERSION=2.7.5
 JAVA_VERSION=17
-NODEJS_VERSION=18.17.1
-NPM_VERSION=8.15.1
+NODEJS_VERSION=20.9.0
```
sschuberth, line 12: Commit message nit:

> Use the LTS version

Maybe say "*current* LTS version" as the LTS version changes over time.

plugins/package-managers/node/src/funTest/kotlin/NpmFunTest.kt
fviernau: I've done it slightly differently now, but following your underlying idea to use assertions instead of comparisons.

sschuberth (approved):

fviernau:
Merged regardless of unrelated detect failure.

PR #7783, opened 2023-11-01
chore(docker): Fix installing Node.js in the legacy image

fviernau:
Building the legacy docker on CI does not work anymore because the way
how Node.js has to be installed has changed, see [1]. So, migrate the
installation accordingly.

[1] https://github.com/nodesource/distributions/issues/1576#issuecomment-1698012034

### DIFF
--- a/Dockerfile-legacy
+++ b/Dockerfile-legacy
@@ -99,7 +99,8 @@ RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/
     apt-get install -y --no-install-recommends ca-certificates gnupg software-properties-common && \
     echo "deb https://repo.scala-sbt.org/scalasbt/debian all main" | tee -a /etc/apt/sources.list.d/sbt.list && \
     curl -ksS "https://keyserver.ubuntu.com/pks/lookup?op=get&options=mr&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | apt-key adv --import - && \
-    curl -sL https://deb.nodesource.com/setup_18.x | bash - && \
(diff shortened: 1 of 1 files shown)


sschuberth (approved):

plugins/package-managers/spm/src/funTest/assets/projects/synthetic/spm-expected-output-lib.yml
```diff
@@ -21,145 +21,117 @@ project:
     - id: "SPM:apple:swift-crypto:2.6.0"
     - id: "SPM:apple:swift-log:1.5.3"
     - id: "SPM:apple:swift-metrics:2.4.1"
-    - id: "SPM:apple:swift-nio:2.61.0"
+    - id: "SPM:apple:swift-nio:2.58.0"
```
sschuberth, line 24: This is a downgrade? Let's see if this passes...

sschuberth (dismissed):

plugins/package-managers/spm/src/funTest/assets/projects/synthetic/spm-expected-output-lib.yml
fviernau: Unbelievable how flaky it is today. Inbetween NuGetFunTest failed, which is now green again, but the SPM dep has been upgraded again.

fviernau: latest test run again has another result for SPM, which again different than my local result.

@sschuberth shall we just merge this as-is ? (and fix SPM on top later on)

sschuberth: Yes, let's fix one thing at a time today... I'll merge.

sschuberth (approved):

PR #7788, opened 2023-11-01
Scan function refactoring

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -109,41 +109,28 @@ class Scanner(
     suspend fun scan(ortResult: OrtResult, skipExcluded: Boolean, labels: Map<String, String>): OrtResult {
         val startTime = Instant.now()
 
-        val projectScannerWrappers = scannerWrappers[PackageType.PROJECT].orEmpty()
-        val packageScannerWrappers = scannerWrappers[PackageType.PACKAGE].orEmpty()
-
-        val projectResults = if (projectScannerWrappers.isNotEmpty()) {
-            val packages = ortResult.getProjects(skipExcluded).mapTo(mutableSetOf()) { it.toPackage() }
-
-            scan(
-                packages,
-                ScanContext(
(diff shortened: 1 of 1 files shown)


scanner/src/main/kotlin/Scanner.kt
```diff
@@ -109,41 +109,28 @@ class Scanner(
...
-            scan(
-                packages,
-                ScanContext(
-                    ortResult.labels,
-                    PackageType.PACKAGE,
-                    ortResult.repository.config.excludes,
-                    scannerConfig.detectedLicenseMapping
-                )
+        val packages = ortResult.getPackages(skipExcluded).map { it.metadata }.filterNotConcluded()
+            .filterNotMetadataOnly().toSet()
```
fviernau, line 125: Maybe add an empty line here.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -109,41 +109,28 @@ class Scanner(
...
+        val packageResults = scan(
+            packages,
+            ScanContext(
+                ortResult.labels,
```
fviernau, line 129: out of scope: Is it correct that line 128 is different from line 116?

fviernau (approved):

scanner/src/main/kotlin/Scanner.kt
sschuberth: I actually had it there first, but then it looked a bit odd compared to the equivalent block with lines 113 / 114, so I decided to not add an empty line here to keep more compact blocks of code. But I don't care that much, I could add an empty line here (and also after line 113) if you prefer.

fviernau: let's leave it as-is.

scanner/src/main/kotlin/Scanner.kt
sschuberth: Good question. I suspect this should also say `ortResult.labels + labels`, @mnonnenmacher?

PR #7789, opened 2023-11-01
fix(VulnerabilityReference): Do not deserialize a lazy property

sschuberth:
This is a fixup for 8b6fe4f which introduced the `severityRating` property.

### DIFF
--- a/model/src/main/kotlin/vulnerabilities/VulnerabilityReference.kt
+++ b/model/src/main/kotlin/vulnerabilities/VulnerabilityReference.kt
@@ -19,6 +19,8 @@
 
 package org.ossreviewtoolkit.model.vulnerabilities
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+
 import java.net.URI
 
 /**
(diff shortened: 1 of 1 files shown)


model/src/main/kotlin/vulnerabilities/VulnerabilityReference.kt
```diff
@@ -30,6 +32,7 @@ import java.net.URI
  * with a list of references; each reference points to the source of the information and has some detailed information
  * provided by this source.
  */
+@JsonIgnoreProperties(value = ["severity_rating"], allowGetters = true)
```
fviernau, line 35: Should we also not serialize the property, because it is redundant? If so, probably it would be good to instead annotate the property with @JsonIgnore.

sschuberth: > Should we also not serialize the property, because it is redundant?

That's more or less the discusion I had with @mnonnenmacher over [here](https://github.com/oss-review-toolkit/ort/pull/7762#discussion_r1376636535). Not serializing it would require to reintroduce the `EvaluatedVulnerabilityReference` class.

fviernau: My preference would (as well?) be to not have it serialized in OrtResult at all, because of the reasons mentioned there. However, that's out of scope of this PR.

fviernau (approved):

sschuberth:
Merging despite the unrelated `NuGetFunTest` failure.

PR #7790, opened 2023-11-01
feat(vulnerabilities): Support the CVSS4 qualitative severity rating scale

sschuberth:
As is it the same as CVSS 3's, just add alternative names to the `Cvss3Rating` enum.


tsteenbe (changes requested):
Not a fan of having a class and file called Cvss3Rating.kt and then have it support both Cvss3 and Cvss4. Propose the file and class are renamed to reflect it supports both versions e.g. `Cvss3or4Rating` or something similar.

sschuberth:
> Propose the file and class are renamed to reflect it supports both versions e.g. `Cvss3or4Rating` or something similar.

That would be a breaking API change again as the class is public. But I've added a separate class now for cleaner semantics, even if it implies some code duplication.

mnonnenmacher (approved):

PR #7800, opened 2023-11-02
helper-cli: Extend scope exclude generation for Poetry

fviernau:
See individual commits.


model/src/main/kotlin/config/ScopeExcludeReason.kt
```diff
@@ -46,6 +46,11 @@ enum class ScopeExcludeReason {
      */
     DEV_DEPENDENCY_OF,
 
+    /**
```
sschuberth, line 49: Commit message typo: "it's" -> "its"

sschuberth (approved):

Issue #7798, opened 2023-11-02
Docker:  Update of python-inspector version not reflected in 'ort-extended' image 6.0.0


fviernau:
As per c841f41d26c8ffe5fbe6127d935546ea94795043 python-inspector version should be 0.10.0 in the 6.0.0 extended image, but it actually  is 0.9.8.

_edit: Also `git-lsf` is absent, but shouldn't be as of:_ https://github.com/oss-review-toolkit/ort/commit/a4b12df0acc6329cc164712e189d2c865c4280b2

sschuberth:
Looks like the image simply hasn't been updated since Sep 8:

![image](https://github.com/oss-review-toolkit/ort/assets/349154/04d949b2-c89a-484b-abe9-c4ca5856822a)

sschuberth:
Hmm, on the other hand, [this page](https://github.com/oss-review-toolkit/ort/pkgs/container/ort-extended) shows version 6.0.0 from today, weird.

PR #7801, opened 2023-11-02
feat(docker): Rename images to new defaults

heliocastro:
Rename Docker image 'ort' to 'ort-minimal' and  'ort-extended' to 'ort' as agreed in community vote
to make it easier for new users to understand difference between ORT Docker images.
Additionally the language component images now have named version with a hash to enable checking if contents
of an image has changed without changing core language.

Closes: #7798


.github/actions/ortdocker/check_image.py
```diff
@@ -26,7 +27,11 @@
 token = os.getenv("INPUT_TOKEN")
 org = os.getenv("GITHUB_REPOSITORY_OWNER")
 name = os.getenv("INPUT_NAME")
-version = os.getenv("INPUT_VERSION")
+base_version = os.getenv("INPUT_VERSION")
+unique_id = hashlib.sha256(os.getenv("BUILD_ARGS").encode()).hexdigest()
```
fviernau, line 31: Is this hashing supposed to solve https://github.com/oss-review-toolkit/ort/issues/7798 ?

How does it work with changes which do not affect build args? 
e.g. I've added `git-lfs` a while ago, but in the published image it's not available for unknown reason. If I build image locally it works as expected.

sschuberth (changes requested):

.github/actions/ortdocker/check_image.py
heliocastro: Yes, this supposes to solve this specific bug.
When build args not changes, hash will be the same, so will check if the image tagged in ghcr.io with {version}-{hash} exists. If exists, not build, since is the same.

.github/actions/ortdocker/action.yml
```diff
@@ -20,81 +20,83 @@ description: "Check and create Docker image for ORT components"
...
-  name:
-    description: "Image name"
-    required: true
-  version:
-    description: "Image version"
-    required: true
-  build-args:
-    description: "List of build-time variables"
-    required: false
+    registry:
```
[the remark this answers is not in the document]
sschuberth: All these lines change indentation for no reason / other code changes:

image

.github/actions/ortdocker/action.yml
```diff
@@ -43,8 +43,8 @@ runs:
     - name: Install Python
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
-        cache: 'pip'
+        python-version: "3.10"
+        cache: "pip"
```
sschuberth, line 47: Please also don't mix-in other unrelated changes like this changing of quotes.

.github/actions/ortdocker/check_image.py
```diff
@@ -47,4 +52,4 @@
     if version in versions:
         print("found")
     else:
-        print("none")
+        print(version)
```
sschuberth, line 55: Is this change intended, and does it belong to the topic of this commit?

sschuberth (changes requested):
> Additionally the language component images now have named version with a hash to enable checking if contents of an image has changed without changing core language.

I propose to do this change in a preceding commit as it's not really on the topic of just renaming the Docker images.

.github/actions/ortdocker/check_image.py
sschuberth: Please explain this here instead of just marking my comment as "resolved".

heliocastro: Did i not explain one line above ?

.github/actions/ortdocker/action.yml
sschuberth: > or change all to double quote, or to single quote

I have no problem with that, but this should be done in a separate commit consistently for all files then.

.github/actions/ortdocker/check_image.py
fviernau: I don't understand yet why 'git-lfs' is not in the docker built by CI, see [1].
That commit [1] is not changing build args, but I wonder if your change here fixes that problem implicitly, e.g. so that docker layer caching starts detecting the modified `apt-get install` e.g.

Can you help me understanding this bit?

[1] https://github.com/oss-review-toolkit/ort/commit/a4b12df0acc6329cc164712e189d2c865c4280b2

heliocastro: Images suppose to be immutable, .
The check happens in this way:
current: base-17-jdk-jammy,  so the version is 17-jdk-jammy.

The test before was checking for version, so every time found a container in ghcr with the version that matches, simply pass and not build it. 

The problem is that we changed the internal content without change the main version. 
So when you added git-lfs, even changing inside, the new container was never generated.

Noe we have a hash based on the arguments as part of the version, so change the build arguments will declare as a new image and will build and push as new.

And the main registry will not receive pushes from content during pull_requests. Only in nighlty or in the tag events.
After this PR is approved, we trigger a workflow dispatch event and the issue will be gone,

.github/actions/ortdocker/action.yml
heliocastro: The name change and the fix are coupled, do it in two steps will not benefit at all. So the pr is done, just waiting for review.

sschuberth: > The name change and the fix are coupled

Changing the quoting style is not coupled to renaming images.

heliocastro: No, the hashing change not the quotes. quotes will be done other time as you suggested.

heliocastro: Is weird that this is threaded over the hashing changes. Sorry for the confusion.

.github/actions/ortdocker/check_image.py
fviernau: Thanks!

heliocastro:
> > Additionally the language component images now have named version with a hash to enable checking if contents of an image has changed without changing core language.
> 
> I propose to do this change in a preceding commit as it's not really on the topic of just renaming the Docker images.

Is mention before, the change and fix are coupled, no reason to do two separate entries.
Can we move forward now ?

tsteenbe (dismissed):

tsteenbe (dismissed):

fviernau (dismissed):

Dockerfile
```diff
@@ -432,8 +432,8 @@ FROM scratch AS ortbin
 COPY --from=ortbuild /opt/ort /opt/ort
 
 #------------------------------------------------------------------------
-# Main Minimal Runtime container
-FROM base as run
+# Minimal Runtime container
+FROM base as minimal
```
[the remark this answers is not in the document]
heliocastro: Because all builds relies on minimal target.  If you not do that, will break the build of all components. Is all about context usage.
Will not break the build, will just make it huge slow, as every sngle component will buils docker from scratch, like no cached exists

scripts/custom_docker.sh
```diff
@@ -27,10 +27,10 @@ DOCKER_IMAGE_ROOT="${DOCKER_IMAGE_ROOT:-ghcr.io/oss-review-toolkit}"
 valid_components=("android" "swift" "sbt" "dart" "dotnet" "php" "haskell")
 
 # Define the Dockerfile template
-dockerfile_template="FROM ghcr.io/oss-review-toolkit/ort\n"
+dockerfile_template="FROM ${DOCKER_IMAGE_ROOT}/ort-minimal\n"
 
 # Default output file
-output_file="Dockerfile.custom"
+output_file="Dockerfile-custom"
```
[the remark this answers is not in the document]
heliocastro: Do we really need to be so nitpicking ? At some point i had people that complained to me to have so much commits in the pull request.

scripts/custom_docker.sh
```diff
@@ -27,10 +27,10 @@ DOCKER_IMAGE_ROOT="${DOCKER_IMAGE_ROOT:-ghcr.io/oss-review-toolkit}"
 valid_components=("android" "swift" "sbt" "dart" "dotnet" "php" "haskell")
 
 # Define the Dockerfile template
-dockerfile_template="FROM ghcr.io/oss-review-toolkit/ort\n"
+dockerfile_template="FROM ${DOCKER_IMAGE_ROOT}/ort-minimal\n"
```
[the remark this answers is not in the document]
heliocastro: Maybe, but this should go for a different  PR, i won't touch this right now

.github/workflows/docker-ort.yml
```diff
@@ -162,9 +162,127 @@ jobs:
             GO_DEP_VERSION=${{ env.GO_DEP_VERSION }}
             GO_VERSION=${{ env.GO_VERSION }}
 
-  runtime_image:
-    needs: [ base_image, nodejs_image, python_image, rust_image, ruby_image, golang_image ]
-    name: Build ORT runtime image
+  android_image:
```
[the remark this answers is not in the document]
heliocastro: No, it can't, otherwise we end up with broken images again, because the chain of dependency is broken.

.github/actions/ortdocker/action.yml
```diff
@@ -54,6 +54,7 @@ runs:
         INPUT_TOKEN: ${{ inputs.token }}
         INPUT_NAME: ${{ inputs.name }}
         INPUT_VERSION: ${{ inputs.version }}
+        BUILD_ARGS: ${{ inputs.build-args }}
```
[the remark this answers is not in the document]
heliocastro: Changed the message.

scripts/custom_docker.sh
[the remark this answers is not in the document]
sschuberth: Well, the rule is simple: Everything that is not absolutely *required* to get the topic of the commit implemented should be moved to a separate commit with its own rationale. On the other hand, intermediate code states during development and fixups for previous commits in the same PR should be consolidated.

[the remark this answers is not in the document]
heliocastro: Moved to a new commit

scripts/docker_build.sh
```diff
@@ -45,127 +45,123 @@ image_build() {
...
 
     docker buildx build \
-        -f "$GIT_ROOT/Dockerfile" \
         --target "$target" \
         --tag "${DOCKER_IMAGE_ROOT}/$name:$version" \
         --tag "${DOCKER_IMAGE_ROOT}/$name:latest" \
         "$@" .
 }
 
+# Minimimal ORT image
```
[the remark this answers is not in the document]
heliocastro: So first there are complain that i made a big commit with al parts that change, and i should split for better understand.
Now i should squash all commits again because there are too many commits and mostly return to the initial situation ?

tsteenbe (approved):

tsteenbe (approved):

PR #7806, opened 2023-11-03
Delete temporary `FileListResolver` directories

sschuberth:

### DIFF
--- a/model/src/main/kotlin/utils/FileArchiver.kt
+++ b/model/src/main/kotlin/utils/FileArchiver.kt
@@ -30,6 +30,7 @@ import org.ossreviewtoolkit.model.KnownProvenance
 import org.ossreviewtoolkit.utils.common.FileMatcher
 import org.ossreviewtoolkit.utils.common.collectMessages
 import org.ossreviewtoolkit.utils.common.packZip
+import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 import org.ossreviewtoolkit.utils.common.unpackZip
 import org.ossreviewtoolkit.utils.ort.createOrtTempFile
 import org.ossreviewtoolkit.utils.ort.ortDataDirectory
@@ -94,7 +95,7 @@ class FileArchiver(
 
         logger.info { "Wrote archive of directory '$directory' to storage in $writeDuration." }
 
-        zipFile.delete()
+        zipFile.parentFile.safeDeleteRecursively(force = true)
     }
 
     /**
--- a/scanner/src/main/kotlin/utils/FileListResolver.kt
+++ b/scanner/src/main/kotlin/utils/FileListResolver.kt
@@ -35,6 +35,7 @@ import org.ossreviewtoolkit.scanner.provenance.ProvenanceDownloader
 import org.ossreviewtoolkit.utils.common.FileMatcher
 import org.ossreviewtoolkit.utils.common.VCS_DIRECTORIES
 import org.ossreviewtoolkit.utils.common.isSymbolicLink
+import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
 
 internal class FileListResolver(
     private val storage: ProvenanceFileStorage,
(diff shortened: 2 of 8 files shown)


sschuberth:
Thanks to @Etsija for pointing this out!

scanner/src/main/kotlin/utils/FileListResolver.kt
```diff
@@ -43,8 +44,11 @@ internal class FileListResolver(
     fun resolve(provenance: KnownProvenance): FileList {
         storage.getFileList(provenance)?.let { return it }
 
-        return provenanceDownloader.download(provenance).let { dir ->
-            createFileList(dir).also { storage.putFileList(provenance, it) }
+        val dir = provenanceDownloader.download(provenance)
+
+        return createFileList(dir).also {
+            storage.putFileList(provenance, it)
```
fviernau, line 50: I believe `putFileList()` could throw an exception.
So, the deletion should go into some `finally` block.

scanner/src/test/kotlin/utils/FileListResolverTest.kt
```diff
@@ -22,40 +22,39 @@ package org.ossreviewtoolkit.scanner.utils
...
-                    "src/cli/main.cpp"
-                )
-            }
+            provenanceDownloader = { dir }
         )
 
         val fileList = resolver.resolve(ArtifactProvenance(sourceArtifact = RemoteArtifact.EMPTY))
 
         fileList.toYaml() shouldBe File("src/test/assets/expected-file-list.yml").readText()
+        dir shouldNotBe aDirectory()
```
fviernau, line 52: Can we have a dedicated test which verifies that the directory is deleted, instead of adding the assertion to the existing unrelated test case ?

fviernau (changes requested):

model/src/main/kotlin/utils/FileArchiver.kt
```diff
@@ -94,7 +95,7 @@ class FileArchiver(
 
         logger.info { "Wrote archive of directory '$directory' to storage in $writeDuration." }
 
-        zipFile.delete()
+        zipFile.parentFile.safeDeleteRecursively(force = true)
```
fviernau, line 98: commit message: fiesl -> files

fviernau (changes requested):

model/src/main/kotlin/utils/FileArchiver.kt
```diff
@@ -30,6 +30,7 @@ import org.ossreviewtoolkit.model.KnownProvenance
 import org.ossreviewtoolkit.utils.common.FileMatcher
 import org.ossreviewtoolkit.utils.common.collectMessages
 import org.ossreviewtoolkit.utils.common.packZip
+import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
```
fviernau, line 33: commit: The operating system takes care of deleting temp files anyway.
So, the only reason I can see for deleting files manually is freeing up disk space.
In that regard deleting all the created temp file manually makes sense to me.

For empty directories, I doubt that we should add mechanims for their deletion as these will be deleted anyway? What do you think?

fviernau (changes requested):

model/src/main/kotlin/utils/FileArchiver.kt
sschuberth: > The operating system takes care of deleting temp files anyway.

Well, that depends on the OS and configuration. Usually temporary files are not deleted immediately, but after some hours / days, or only on the next reboot.

> So, the only reason I can see for deleting files manually is freeing up disk space.

I see people being confused by `ort-*` directory left-overs in the OS's temporary directory because that creates the impression that something went wrong and ORT crashed.

So it's also a way to double-check that nothing unexpected did happen.

fviernau: I see. So let's leave it as-is.

Anyhow, what do you think of doing something similar as `File.deleteOnExit()` inside `createOrtTempDir()`. E.g. register a hook triggered when JVM exits, which then deletes all still existing temp dirs recursively?

fviernau (approved):

model/src/main/kotlin/utils/FileArchiver.kt
Etsija: Regarding the OS deleting the temp files automagically: it really seems to depend on the system. I'm running ORT and developing plugins for it on WSL2, and it surely isn't very eager to delete anything from the /tmp/.  It's gone to a stage where I actually had to create a Bash script to controllably delete ORT-specific directories from it.

sschuberth: > doing something similar as `File.deleteOnExit()`

I'd rather not, as [`java.io.File.deleteOnExit()` can cause leaks](http://fxapps.blogspot.com/2017/10/how-javaiofiledeleteonexit-can-cause.html).

PR #7816, opened 2023-11-05
Log about configured but unavailable package managers 

sschuberth:

### DIFF
--- a/utils/common/src/main/kotlin/Extensions.kt
+++ b/utils/common/src/main/kotlin/Extensions.kt
@@ -34,6 +34,11 @@ import java.util.EnumSet
 import java.util.Locale
 
 /**
+ * Call [also] only if the receiver is null, e.g. for error handling, and return the receiver in any case.
+ */
+inline fun <T> T.alsoIfNull(block: (T) -> Unit): T = this ?: also(block)
+
+/**
  * Return a string of hexadecimal digits representing the bytes in the array.
  */
 fun ByteArray.encodeHex(): String = joinToString("") { String.format(Locale.ROOT, "%02x", it) }
--- a/analyzer/src/main/kotlin/Extensions.kt
+++ b/analyzer/src/main/kotlin/Extensions.kt
@@ -19,7 +19,10 @@
 
 package org.ossreviewtoolkit.analyzer
 
+import org.apache.logging.log4j.kotlin.logger
+
 import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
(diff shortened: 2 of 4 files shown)


utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -33,6 +33,11 @@ import java.nio.file.attribute.BasicFileAttributes
 import java.util.EnumSet
 import java.util.Locale
 
+/**
+ * Conveniently call [also] only if the receiver is null, e.g. for error handling, and return the receiver in any case.
```
fviernau, line 37: nit: "Conveniently" could be dropped.

utils/common/src/main/kotlin/Extensions.kt
```diff
@@ -33,6 +33,11 @@ import java.nio.file.attribute.BasicFileAttributes
...
+ */
+inline fun <T> T.alsoIfNull(block: (T) -> Unit): T = this ?: also(block)
```
fviernau, line 39: Should / can this have the `T?` ? `inline fun <T> T?.alsoIfNull`

fviernau: The code block is only executed if `T` is null, so the parameter could be
`() -> Unit` instead.

fviernau (changes requested):

utils/common/src/main/kotlin/Extensions.kt
sschuberth: > Should / can this have the `T?` ? `inline fun <T> T?.alsoIfNull`

No, in contrast to `Any.alsoIfNull` (which does not match `Any?`), using a generic `T` includes `T?`. FYI, I copied the function signature from Kotlin's `also` extension function.

sschuberth: > The code block is only executed if `T` is null, so the parameter could be `() -> Unit` instead.

No, it couldn't, as I'm passing `block` on to `also`, which requires this function signature.

fviernau: > No, it couldn't, as I'm passing block on to also

Hm, I basically meant something like the following:
```
inline fun <T> T.alsoIfNull(block: () -> Unit): T = this ?: run {
    block()
    null
}
```
e.g. not calling `also` at all. What do you think ?

sschuberth: I prefer calling `also` internally as it's slighly more compact and highlights the relation between these two functions.

fviernau (approved):

PR #7827, opened 2023-11-08
docs(evaluator): Fix-up the docs for two CLI options

fviernau:
Previously, the docs referred to package configurations which did not make sense, see also [1]. 

[1] 910454160c51bb783a6b40496ba68c782901e835

### DIFF
--- a/plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
+++ b/plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
@@ -149,7 +149,7 @@ class EvaluatorCommand : OrtCommand(
     private val packageCurationsFile by option(
         "--package-curations-file",
         help = "A file containing package curations. This replaces all package curations contained in the given ORT " +
-            "result file with the ones present in the given file and, if enabled, those from the package " +
+            "result file with the ones present in the given file and, if enabled, those from the repository " +
             "configuration."
     ).convert { it.expandTilde() }
         .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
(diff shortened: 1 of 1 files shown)


plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
```diff
@@ -148,19 +148,19 @@ class EvaluatorCommand : OrtCommand(
 
     private val packageCurationsFile by option(
         "--package-curations-file",
-        help = "A file containing package curations. This replaces all package curations contained in the given ORT " +
-            "result file with the ones present in the given file and, if enabled, those from the package " +
-            "configuration."
+        help = "A file containing package curations to replace the package curations in the given ORT file. The " +
```
sschuberth, line 151: Hmm, the docs *still* seem to refer to package curations, so I'm unsure about the intention of the change.

plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
```diff
@@ -148,19 +148,19 @@ class EvaluatorCommand : OrtCommand(
...
         .file(mustExist = true, canBeFile = true, canBeDir = false, mustBeWritable = false, mustBeReadable = true)
         .convert { it.absoluteFile.normalize() }
         .configurationGroup()
 
     private val packageCurationsDir by option(
         "--package-curations-dir",
-        help = "A directory containing package curation files. This replaces all package curations contained in the " +
-            "given ORT result file with the ones present in the given directory and, if enabled, those from the " +
-            "package configuration file."
+        help = "A file containing package curations to replace the package curations in the given ORT file. The " +
```
sschuberth, line 161: This should continue to say "dir" instead of "file".

plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
sschuberth: Ah, I guess you meant "package configuration" instead of "package curations" in the commit message.

All in all, I find the existing wording to be more clear, and I'd propose to just change "package configuration" to "repository configuration" in the last sentence.

sschuberth (changes requested):

PR #7831, opened 2023-11-09
fix: Add advisor plugins to the plugin classpath for distribution

sschuberth:
Without this, building the distribution would not necessarily build the advisor plugins.


sschuberth:
@nnobelis, this should fix your issue.

fviernau (approved):

PR #7838, opened 2023-11-09
docs(scanner): Improve logging for packages with incomplete scan results

sschuberth:
A user does not know that the reference package has its path removed, so the original message sounded as if the package is scanned as a substitute for others. Clarify that by rewording the log message.


sschuberth:
FYI @Etsija.

mnonnenmacher (approved):

PR #2773, opened 2020-06-24
spdx-utils: Implement the SPDX 2.2.1 document model

fviernau:
This change adds the model classes for SPDX documents for SPDX version
2.2.1 and supports serialization and deserialization of JSON and YAML
SPDX document files. All properties are supported except for 'ranges'
which are not properly specified yet [1], and thus cannot be implemented yet.

There is some design flexibility regarding how strong the typing is for
the respective properties, e.g. a propery could use a String or a
dedicated type. This change does not implement the strongest possible
typing for both, fault tollerance and most importantly to limit the
scope of this first iteration. Adding a dedicated type for SPDX IDs or
references are good candidates for future improvements.

The SPDX specification [2] is targeting mainly RDF and thus does not
provide enough information to derive the actual JSON or YAML format.
To fill the gap the examples [3] are taken as referrence additionally.

The tests proof that the JSON and YAML examples are (apart from the
mentioned ranges property) fully implemented, e.g.
`serialize(deserialize(example)) == example` holds.

[1] https://github.com/spdx/spdx-spec/issues/412
[2] https://github.com/spdx/spdx-spec/tree/development/v2.2.1/chapters
[3] https://github.com/spdx/spdx-spec/tree/development/v2.2.1/examples

Signed-off-by: Frank Viernau <frank.viernau@here.com>

PR #7839, opened 2023-11-09
Upgrade SPDX output to specification version 2.3

sschuberth:

### DIFF
--- a/README.md
+++ b/README.md
@@ -732,7 +732,7 @@ following formats are supported (reporter names are case-insensitive):
   * Customizable with [Apache Freemarker](https://freemarker.apache.org/) templates
 * Opossum input that can be visualized and edited in the [OpossumUI](https://github.com/opossum-tool/opossumUI)
   (`-f Opossum`)
-* [SPDX Document](https://spdx.dev/specifications/), version 2.2 (`-f SpdxDocument`)
+* [SPDX Document](https://spdx.dev/specifications/), version 2.3 (`-f SpdxDocument`)
 * Static HTML (`-f StaticHtml`)
 * [TrustSource](https://www.trustsource.io/) JSON file (`-f TrustSource`)
   * Use this as an alternative to [ts-scan](https://github.com/TrustSource/ts-scan) for support of more build systems.
--- a/plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
+++ b/plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.json
@@ -1,6 +1,6 @@
 {
   "SPDXID" : "SPDXRef-DOCUMENT",
-  "spdxVersion" : "SPDX-2.2",
+  "spdxVersion" : "SPDX-2.3",
   "creationInfo" : {
     "comment" : "some creation info comment",
     "created" : "<REPLACE_CREATION_DATE_AND_TIME>",
@@ -39,7 +39,7 @@
     "copyrightText" : "Copyright 2020 Some copyright holder in VCS\nCopyright 2020 Some copyright holder in source artifact\nCopyright 2020 Some other copyright holder in source artifact",
     "downloadLocation" : "https://some-host/first-package.jar",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/first-package-group/first-package@0.0.1"
     } ],
@@ -55,7 +55,7 @@
     "copyrightText" : "Copyright 2020 Some copyright holder in VCS\nCopyright 2020 Some copyright holder in source artifact\nCopyright 2020 Some other copyright holder in source artifact",
     "downloadLocation" : "git+ssh://github.com/path/first-package-repo.git@deadbeef#project-path",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/first-package-group/first-package@0.0.1"
     } ],
@@ -80,7 +80,7 @@
     "copyrightText" : "Copyright 2020 Some copyright holder in VCS\nCopyright 2020 Some copyright holder in source artifact\nCopyright 2020 Some other copyright holder in source artifact",
     "downloadLocation" : "https://some-host/first-package-sources.jar",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/first-package-group/first-package@0.0.1"
     } ],
@@ -96,7 +96,7 @@
     "copyrightText" : "NONE",
     "downloadLocation" : "NONE",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/fourth-package-group/fourth-package@0.0.1"
     } ],
@@ -112,7 +112,7 @@
     "copyrightText" : "NONE",
     "downloadLocation" : "NONE",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/second-package-group/second-package@0.0.1"
     } ],
@@ -128,7 +128,7 @@
     "copyrightText" : "Copyright 2020 Some copyright holder in source artifact",
     "downloadLocation" : "NONE",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/seventh-package-group/seventh-package@0.0.1"
     } ],
@@ -148,7 +148,7 @@
     "copyrightText" : "Copyright 2020 Some copyright holder in source artifact",
     "downloadLocation" : "https://some-host/seventh-package-sources.jar",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/seventh-package-group/seventh-package@0.0.1"
     } ],
@@ -169,7 +169,7 @@
     "copyrightText" : "NONE",
     "downloadLocation" : "NONE",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/sixth-package-group/sixth-package@0.0.1"
     } ],
@@ -185,7 +185,7 @@
     "copyrightText" : "NONE",
     "downloadLocation" : "NONE",
     "externalRefs" : [ {
-      "referenceCategory" : "PACKAGE_MANAGER",
+      "referenceCategory" : "PACKAGE-MANAGER",
       "referenceType" : "purl",
       "referenceLocator" : "pkg:maven/third-package-group/third-package@0.0.1"
     } ],
--- a/plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.yml
+++ b/plugins/reporters/spdx/src/funTest/assets/spdx-document-reporter-expected-output.spdx.yml
@@ -1,6 +1,6 @@
 ---
 SPDXID: "SPDXRef-DOCUMENT"
-spdxVersion: "SPDX-2.2"
+spdxVersion: "SPDX-2.3"
 creationInfo:
   comment: "some creation info comment"
   created: "<REPLACE_CREATION_DATE_AND_TIME>"
@@ -49,7 +49,7 @@ packages:
     \ in source artifact"
   downloadLocation: "https://some-host/first-package.jar"
   externalRefs:
-  - referenceCategory: "PACKAGE_MANAGER"
+  - referenceCategory: "PACKAGE-MANAGER"
     referenceType: "purl"
     referenceLocator: "pkg:maven/first-package-group/first-package@0.0.1"
   filesAnalyzed: false
@@ -67,7 +67,7 @@ packages:
     \ in source artifact"
   downloadLocation: "git+ssh://github.com/path/first-package-repo.git@deadbeef#project-path"
   externalRefs:
-  - referenceCategory: "PACKAGE_MANAGER"
+  - referenceCategory: "PACKAGE-MANAGER"
     referenceType: "purl"
     referenceLocator: "pkg:maven/first-package-group/first-package@0.0.1"
   filesAnalyzed: true
@@ -95,7 +95,7 @@ packages:
     \ in source artifact"
   downloadLocation: "https://some-host/first-package-sources.jar"
   externalRefs:
-  - referenceCategory: "PACKAGE_MANAGER"
+  - referenceCategory: "PACKAGE-MANAGER"
     referenceType: "purl"
     referenceLocator: "pkg:maven/first-package-group/first-package@0.0.1"
   filesAnalyzed: false
@@ -111,7 +111,7 @@ packages:
   copyrightText: "NONE"
   downloadLocation: "NONE"
   externalRefs:
-  - referenceCategory: "PACKAGE_MANAGER"
+  - referenceCategory: "PACKAGE-MANAGER"
     referenceType: "purl"
     referenceLocator: "pkg:maven/fourth-package-group/fourth-package@0.0.1"
   filesAnalyzed: false
@@ -125,7 +125,7 @@ packages:
   copyrightText: "NONE"
   downloadLocation: "NONE"
   externalRefs:
-  - referenceCategory: "PACKAGE_MANAGER"
+  - referenceCategory: "PACKAGE-MANAGER"
     referenceType: "purl"
     referenceLocator: "pkg:maven/second-package-group/second-package@0.0.1"
   filesAnalyzed: false
@@ -139,7 +139,7 @@ packages:
   copyrightText: "Copyright 2020 Some copyright holder in source artifact"
   downloadLocation: "NONE"
   externalRefs:
-  - referenceCategory: "PACKAGE_MANAGER"
+  - referenceCategory: "PACKAGE-MANAGER"
     referenceType: "purl"
     referenceLocator: "pkg:maven/seventh-package-group/seventh-package@0.0.1"
   filesAnalyzed: false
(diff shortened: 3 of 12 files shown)


tsteenbe (changes requested):
As SPDX 2.2 is the ISO version it has been adopted by many organizations as their standard (know several automotive companies that are 2.2-only ). By solely supporting SPDX 2.3 we likely reduce our potential user pool. Also we upcoming SPDX 3.0 and multiple CycloneDX versions

ORT TSC should imo decide on a supported versions scheme for open standards e.g.
A) We only support latest version
B) We support last 2 major versions
C) We support versions are are standard or popular with the community

Depending on TSC decision this pull request may need to be changed.

sschuberth:
> By solely supporting SPDX 2.3 we likely reduce our potential user pool.

Realistically, I believe that's not the case. As you can see from the diff of the spec, the differences between 2.2 and 2.3 are mostly about clarifying nature, and cleaning up with inconsistencies. We don't write out any 2.3-only fields, so unless SPDX consumers really look at the version in the header (which is very rarely the case), they can just read ORT's SPDX documents as if they were 2.2.

This PR actually fixes a source of confusion about `PACKAGE-MANAGER` vs. `PACKAGE-MANAGER` that the guys from Mercedes-Benz at the FOSS Convention told me about, so it's addressing direct feedback from our user base.

I don't think we need a TSC decision to move forward.

fviernau:
I can also share that I have seen the requirement from real world customers to produce SPDX documents in version 2.2.

sschuberth:
> produce SPDX documents in version 2.2.

But what does that actually mean? That `spdxVersion: "SPDX-2.2"` is literally written in the header, or that the actual document can syntactically be understood by a version 2.2 parser?

fviernau:
> But what does that actually mean? That `spdxVersion: "SPDX-2.2"` is literally written in the header, or that the actual document can syntactically be understood by a version 2.2 parser?

That it is valid WRT version 2.2. And that used SPDX tooling acccepts it as a valid 2.2.  input.

sschuberth:
> That it is valid WRT version 2.2. And that used SPDX tooling acccepts it as a valid 2.2. input.

Ok, so the version in written in the header does not actually matter as long as the document is property-wise SPDX 2.2-compliant. Which it is, even after this PR.

tsteenbe:
> Ok, so the version in written in the header does not actually matter as long as the document is property-wise SPDX 2.2-compliant

No, several consumption tools we have seen used actively check version header for both CycloneDX and SPDX to determine the right way to parse and validate the SBOM. Several enterprise standardized on SPDX 2.2 as it's ISO including in their procurement docs.

For me this matter of transparency & trust to our ORT users, we as TSC should set clear expectation - what SBOM version we support, when it support versions will change and why

sschuberth:
What about the first three commits in this PR, these also apply to SPDX 2.2. Can we merge these if I split them out?

fviernau:
> What about the first three commits in this PR, these also apply to SPDX 2.2. Can we merge these if I split them out?

I haven't reviewed these in detail yet, but if it's true that these are for version 2.2, then I'd have no objections moving these forward.

utils/spdx/src/main/kotlin/model/SpdxSnippet.kt
```diff
@@ -82,11 +82,38 @@ data class SpdxSnippet(
...
      */
     val snippetFromFile: String
 ) {
+    data class Range(
+        /** The start of the snippet range. */
+        val startPointer: Pointer,
+
+        /** The end of the snippet range. */
+        val endPointer: Pointer
+    )
```
fviernau, line 102: I couldn't find 100% proof in the spec, but I believe the intention of it is that the range is 
either 1. a line range or 2.  a byte range. This implies that `startPointer` and `endPointer`
must be of the same type (byte range vs line range).

Looking at the `Pointer` below, it should hold `lineNumber==null xor offset==null`. So, the model could be stronger typed also in terms of not null guarantee, if it modeled `ByteRangerPointer` and `LineRangePointer` separately.

Do you think it would make sense to add polymorphism here? (I'm not sure myself if this is a good approach, an alternative could be to add more documentation and also invariant checks into the cosntructor)

sschuberth: I was playing with having `sealed class Pointer` with derived `LineNumberPointer` and `ByteOffsetPointer`, but as either would then only have one `Int?` property, we'd run into a deserializer exception if someone specified both types of pointers, instead of having a probably nicer message triggered by `require` checks. I'm not sure what's better here.

This is BTW again an unfortunate example of SPDX model design choices, ping @goneall.

sschuberth: See https://github.com/oss-review-toolkit/ort/pull/7890 for the split-out SPDX 2.2 changes.

sschuberth:
Pinging @schr3gl3j, just curious, would you be more interested in SPDX 2.2 or SPDX 2.3 output?

goneall:
For reference - here's a comment with a list of changes from SPDX 2.2 to SPDX 2.3 that may impact tooling: https://github.com/spdx/spdx-spec/issues/691#issuecomment-1154198748

If you haven't already, you may want to use the above comment as a checklist to make sure the tooling is updated.

Also note that if you use the [SPDX Java Library](https://github.com/spdx/Spdx-Java-Library), it supports multiple versions of the spec and is in use in several other projects.

goneall:
W.R.T. the above comments on supporting version 2.2, I have also ran into organizations that will not use SPDX 2.3 since it has not gone through the ISO certification.

sschuberth:
> Also note that if you use the [SPDX Java Library](https://github.com/spdx/Spdx-Java-Library), it supports multiple versions of the spec

Yeah, that would be my main incentive to switching (and probably contributing) to it instead of maintaining our own model...

tsteenbe:
> Yeah, that would be my main incentive to switching (and probably contributing) to it instead of maintaining our own model...

Objections to using SPDX Java Library are still valid, nothing has changed - with all respect to Gary's continuous work maintaining it.

sschuberth:
> Objections to using SPDX Java Library are still valid, nothing has changed

I remember there were objections, but I cannot not find the discussion anymore. So could you please link to it / repeat the objections?

goneall:
> > Objections to using SPDX Java Library are still valid, nothing has changed
> 
> I remember there were objections, but I cannot not find the discussion anymore. So could you please link to it / repeat the objections?

If any of the objections are issues that can be reasonably fixed in the library, please let me know.  I'm not aware of the issues / objections other than the decision not to use the library.

sschuberth:
> If any of the objections are issues that can be reasonably fixed in the library, please let me know. I'm not aware of the issues / objections other than the decision not to use the library.

I've found [my original ask for a justification not to use the upstream SPDX Java library](https://github.com/oss-review-toolkit/ort/pull/2773#issuecomment-648890512) in the [PR from 2020 that introduces the SPDX classes](https://github.com/oss-review-toolkit/ort/pull/2773). In reply to my remark @fviernau added the following paragraph to the commit message:

> Note that there is already a SPDX java library [4] which is
intentionally not used mainly for keeping the development speed and
decisions concerning the SPDX model completely under our control.

Given that we didn't do a lot of development work in the SPDX area for quite some time, I tend to disagree with @tsteenbe's statement that "[Objections to using SPDX Java Library are still valid](https://github.com/oss-review-toolkit/ort/pull/7839#issuecomment-1828297120)", and we should probably revisit that decision.

heliocastro:
Not help much that the older Java library is still haunting us, because people are confused.

goneall:
> Not help much that the older Java library is still haunting us, because people are confused.

Good point - the old library definitely had issues (that's why we re-wrote it).

The old library README has been updated to point to the new library - let me know if you have any ideas on how to better communicate that there is a new and improved library available.

sschuberth:
Not pursuing this further.

sschuberth:
For reference, @fviernau mentioned [this comment](https://github.com/spdx/spdx-spec/issues/691#issuecomment-1154198748) which highlights the differences between SPDX 2.2 and 2.3.

utils/spdx-document/src/test/kotlin/model/SpdxDocumentTest.kt
```diff
@@ -58,7 +58,8 @@ class SpdxDocumentTest : WordSpec({
                 SpdxModelMapper.toYaml(it)
             }
 
-            reSerializedYaml lenientShouldEqualYaml yaml
+            // Account for the fact that ORT serializes SPDX 2.3 now.
+            reSerializedYaml.replace("PACKAGE-MANAGER", "PACKAGE_MANAGER") lenientShouldEqualYaml yaml
```
fviernau, line 62: `toYaml()` now outputs 2.3, but in this case the output property `spdxVersion` would be 2.2.

I'm not sure how to best deal with this / whether this is a problem.
The reading and writing use case IIUC is not used in production code, but just in the test. As it seemed a nice way to verify serializing / de-serializing actually works.

Maybe we can enhance `toYaml()` with a toggle to output 2.2.2 / 2.3. 
And the implementation then could also set the `spdxVersion` property accordingly + the dash / underscore patching above. This would then need to be enhanced to also patch another enum element. IMO 
two conversion functions `String.toSpdx22Yaml(): String` and `String.toSpdx22Json()` could be provided. These would then be re-usable also for the upcoming new reporter.

utils/spdx-document/src/test/kotlin/model/SpdxDocumentTest.kt
```diff
@@ -58,7 +58,8 @@ class SpdxDocumentTest : WordSpec({
                 SpdxModelMapper.toYaml(it)
             }
 
-            reSerializedYaml lenientShouldEqualYaml yaml
+            // Account for the fact that ORT serializes SPDX 2.3 now.
```
fviernau, line 61: I believe we should add an analog test for 2.3 in this commit. (I guess that patching the upstream example by removing ranges is not necessary in this case)

plugins/reporters/spdx/src/funTest/resources/synthetic-scan-result-expected-output.spdx.json
```diff
@@ -1,6 +1,6 @@
 {
   "SPDXID" : "SPDXRef-DOCUMENT",
-  "spdxVersion" : "SPDX-2.2",
+  "spdxVersion" : "SPDX-2.3",
```
fviernau: I propose to (in this commit), enhance the reporter so that it still outputs 2.2.
Under the hood it would serialize to 2.3 and then patch it up.
This would imo more adhere to doing one thing at a time, and capability to producec 2.2 would not be temporarily lost.

In a following commit / PR we can then add a reporter parameter, so that the version can be specified.

fviernau (changes requested):

utils/spdx-document/src/test/kotlin/model/SpdxDocumentTest.kt
sschuberth: > I believe we should add an analog test for 2.3 in this commit.

I agree that we should have test coverage for the plain 2.3 case here, but I'll leave this out of this commit / PR to focus on the desired patching logic.

plugins/reporters/spdx/src/funTest/resources/synthetic-scan-result-expected-output.spdx.json
sschuberth: > I propose to (in this commit), enhance the reporter so that it still outputs 2.2.
> Under the hood it would serialize to 2.3 and then patch it up.

I've done that now with a `patchSpdx23To22()` function, which should also cover your comment from [above](https://github.com/oss-review-toolkit/ort/pull/7839#discussion_r2169145769).

utils/spdx-document/src/main/kotlin/model/SpdxChecksum.kt
```diff
@@ -47,7 +52,10 @@ data class SpdxChecksum(
         SHA224(56),
         SHA256(64),
         SHA384(96),
-        SHA512(128)
+        SHA512(128),
+        SHA3_256(64),
+        SHA3_384(96),
+        SHA3_512(128)
```
fviernau: In order for the reporter to produce 2.2 compatible output, these new entries must not be used by the reporter. I believe `Hash.toSpdxChecksum()` should be modified accordingly (e.g. make the mapping more explicit, without relying on `find()`)?

utils/spdx-document/src/main/kotlin/model/SpdxPackage.kt
```diff
@@ -52,6 +52,12 @@ data class SpdxPackage(
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     val attributionTexts: List<String> = emptyList(),
 
+    /**
+     * The actual date the package was built in "YYYY-MM-DDThh:mm:ssZ" format.
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    val builtDate: String? = null,
```
fviernau: Should this use `Instant`, analog to `SpdxCreationInfo.created` ?

utils/spdx-document/src/main/kotlin/model/SpdxPackage.kt
```diff
@@ -184,12 +202,36 @@ data class SpdxPackage(
     @JsonInclude(JsonInclude.Include.NON_NULL)
     val supplier: String? = null,
 
+    /**
+     * The end of the support period for a package from the supplier in "YYYY-MM-DDThh:mm:ssZ" format.
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    val validUntilDate: String? = null,
```
fviernau: Should this use `Instant`, analog to `SpdxCreationInfo.created` ?

utils/spdx-document/src/main/kotlin/model/SpdxPackage.kt
```diff
@@ -162,6 +168,18 @@ data class SpdxPackage(
...
+     * This field provides information about the primary purpose of the identified package.
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    val primaryPackagePurpose: Purpose? = null,
+
+    /**
+     * The date the package was released in "YYYY-MM-DDThh:mm:ssZ" format.
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    val releaseDate: String? = null,
```
fviernau: Should this use `Instant`, analog to `SpdxCreationInfo.created` ?

utils/spdx-document/src/main/kotlin/model/SpdxPackage.kt
```diff
@@ -162,6 +168,18 @@ data class SpdxPackage(
     @JsonInclude(JsonInclude.Include.NON_NULL)
     val packageVerificationCode: SpdxPackageVerificationCode? = null,
 
+    /**
+     * This field provides information about the primary purpose of the identified package.
+     */
```
fviernau, line 176: This KDoc could reference https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field.

utils/spdx-document/src/test/kotlin/model/SpdxDocumentTest.kt
fviernau: Agree.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
```diff
@@ -167,3 +166,12 @@ private fun SpdxExpression.getLicenseRefExceptions(result: MutableSet<String>) {
         else -> {}
     }
 }
+
+private fun String.patchSpdx23To22() =
+    replace(spdxVersion22Regex, "$1SPDX-2.2$2")
+        .replace("PACKAGE-MANAGER", "PACKAGE_MANAGER")
+        .replace("PERSISTENT-ID", "PERSISTENT_ID")
```
fviernau: Would it make sense to use the same approach as for the `spdxVersion` version, e.g. include the property name in the regex?

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
```diff
@@ -167,3 +166,12 @@ private fun SpdxExpression.getLicenseRefExceptions(result: MutableSet<String>) {
         else -> {}
     }
 }
+
+private fun String.patchSpdx23To22() =
```
fviernau: Would it make sense to (in the future) enhance this to operate on a parsed tree, e.g. JsonNode?
If yes, would it make sense to leave a TODO comment?

utils/spdx-document/src/main/kotlin/model/SpdxChecksum.kt
sschuberth: There should be no risk in accidentally writing these out as ORT's `HashAlgorithm` does not support any of the newly added algorithms.

fviernau: If at some point ORT's supported algorithms get enhanced, in that case I believe it is very easy to miss this.
So, I'd prefer to have explicit entries for the mapping and a code comment which warns about adding 2.3 entries.

If you don't like the above still, how about addressing it with one of:
1. Adding a unit test for `Hash.toSpdxChecksum()` which verifies no 2.3 enum elements get returned?
2. or as a minimum, add a code comment (similar to what you wrote above)

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
sschuberth: See below.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
sschuberth: I've already now migrated this to operate on `ObjectNode` / `JsonNode`, and in the same run also generalized the enum value replacement .

utils/spdx-document/src/main/kotlin/model/SpdxChecksum.kt
sschuberth: I've added filtering now to `Hash.toSpdxChecksum()` in a way that should easily allow to make it configurable via a reporter option for the SPDX version later.

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
```diff
@@ -167,3 +168,23 @@ private fun SpdxExpression.getLicenseRefExceptions(result: MutableSet<String>) {
...
+            val refCat = ref.get("referenceCategory") ?: return@forEach
+
+            val refCatText = refCat.textValue()
+            if (refCatText != null && "-" in refCatText) {
+                (ref as ObjectNode).put("referenceCategory", refCatText.replace("-", "_"))
+            }
+        }
+    }
+
+    return this
```
fviernau, line 189: I hope my comment (on the outdated diff) was not mis-leading.

If we want to use this function (as-is) also for the new reporter (Spdx Java Library), then it should probably use KxS (and kaml, for the yaml case), and take a string as input.

Should we do this already now, or do you prefer to keep the function as-is and refactor it later again?

fviernau (changes requested):

plugins/reporters/spdx/src/main/kotlin/SpdxDocumentReporter.kt
sschuberth: > If we want to use this function (as-is) also for the new reporter (Spdx Java Library)

Hmm, actually I wasn't planning to use this function with the new implementation. My thinking was that the new implementation will only support SPDX 2.3 and 3.0, and the old implementation would support SPDX 2.2 and 2.3. So we could compare old and new implementations based on SPDX 2.3 only. We would keep old and new reporters for a while in parallel, and then at some (hopefully not too far away) time, drop the old implementation (and SPDX 2.2 support with it).

Or do you think we need to support SPDX 2.2 with the new implementation right from the start?

fviernau: > Or do you think we need to support SPDX 2.2 with the new implementation right from the start?

Yes, I believe this is needed. I think the new reporter should have the same features as the old one,
so that we can abandon the old one once the new one is battle tested. (As we see in this PR, supporting
2.2 also with the new one is not a big effort to do, and this was also a provided requirement)

Anyhow, I'm perfectly fine to keep this functions as-is for now, and get  into these details later. I simply wanted to raise awareness.

fviernau (approved):

PR #7845, opened 2023-11-10
chore(docker): Upgrade Conan to the latest 1.x version

fviernau:


sschuberth (dismissed):

sschuberth:
Looks like the major upgrade to Conan 2 needs some more changes: ['info' is not a Conan command. See 'conan --help'.](https://github.com/oss-review-toolkit/ort/actions/runs/6823601539/job/18557839607?pr=7845#step:6:2267)

fviernau:
> Looks like the major upgrade to Conan 2 needs some more changes: ['info' is not a Conan command. See 'conan --help'.](https://github.com/oss-review-toolkit/ort/actions/runs/6823601539/job/18557839607?pr=7845#step:6:2267)

Seems like in addition also the `conanfile` format has changed. Do we need to keep 1.x support then and have 2.x support in parallel?

sschuberth:
> Do we need to keep 1.x support then and have 2.x support in parallel?

I believe we need 1.x support, yes, but the people from Bosch might be better to tell. My question is rather: Do we need 2.x support (yet)?

mnonnenmacher:
> > Do we need to keep 1.x support then and have 2.x support in parallel?
> 
> I believe we need 1.x support, yes, but the people from Bosch might be better to tell. My question is rather: Do we need 2.x support (yet)?

Yes, we need 1.x support and will likely need to support both versions in parallel in future.

fviernau:
> Yes, we need 1.x support and will likely need to support both versions in parallel in future.

So, then let's just try latest 1.x version in this PR.

sschuberth (approved):

PR #7853, opened 2023-11-13
Remove the unmaintained Batect in favor of using a GitHub action container

sschuberth:


.github/workflows/build-and-test.yml
```diff
@@ -141,6 +141,8 @@ jobs:
         flags: funTest-non-docker
   funTest-docker:
     runs-on: ubuntu-22.04
+    container:
+      image: ghcr.io/oss-review-toolkit/ort
```
heliocastro, line 145: Show we specify the last nightly image or the latest released ? Somehow tests that depends on new features on base components that are not available on the curren published ones.

sschuberth: This should be as recent as possible, i.e. a daily build tag. Anyway, for testing I was trying to use `latest` (both implicitly by not specifying a tag, and by explicitly specifying the tag name), but for some reason it [does not work](https://github.com/oss-review-toolkit/ort/actions/runs/6855036442/job/18639246562?pr=7853#step:2:21).

sschuberth: Seems to be related to https://github.com/docker/build-push-action/issues/820.

sschuberth: > This should be as recent as possible, i.e. a daily build tag

Using the `snapshot` Docker tag now for that matter.

cli/src/funTest/kotlin/AnalyzerFunTest.kt
mnonnenmacher, line 1: Commit message: "ttest" -> "test"

.github/workflows/build-and-test.yml
mnonnenmacher, line 1: Commit message: "dependemcy" -> "dependency"

mnonnenmacher, line 1: Commit message: "Docherfile" -> "Dockerfile"

> The downside is that `Dockerfile` changes and e.g. package manager changes depending on it cannot be performed in the same commit / PR as the image first needs to get published.

Doesn't this introduce a serious issue if e.g. a tool version in Docker is upgraded and any test failures caused by this only appear in a random other PR latern on? This probably also creates the risk that we make broken releases.

I thought with the new setup building the Docker image is a lot faster, as most of the images can be reused from the registry, and that only Batect was blocking us from using the registry. E.g. if the Python version is upgraded only the Python image needs to be rebuilt, and if nothing is changed in Docker the build should be very fast?

mnonnenmacher (changes requested):

.github/workflows/build-and-test.yml
sschuberth: > Doesn't this introduce a serious issue if e.g. a tool version in Docker is upgraded and any test failures caused by this only appear in a random other PR later on?

Yes, that's a valid point. To solve this, we should probably do something like this in the PR:
1. Pull the current snapshot image.
2. (Re-)build the Docker image from source on the worker node, which should be fast due to stage image / layer caching, if not a lot has changed in the `Dockerfile` itself.
3. Run functional tests against the local Docker image.
4. Push the Docker image as the new snapshot if everything succeeded.

sschuberth: > Batect was blocking us from using the registry.

Not that it matters now anymore as Batect became unmaintained, but Batect was not blocking us from using the registry. Batect just made it hard / impossible to benefit from stage image / layer caching in conjunction with the `docker/build-push-action`.

> E.g. if the Python version is upgraded only the Python image needs to be rebuilt, and if nothing is changed in Docker the build should be very fast?

That's my understanding as well, but I haven't verified it.

sschuberth:
I've split out the first three commits to https://github.com/oss-review-toolkit/ort/pull/7894 in case we want to discuss the last two commits of this PR further.

sschuberth:
@heliocastro please don't force-push to branches owned by others without discussing this first! Not even if it's a simple rebase.

heliocastro (approved):
This solution is better since we are not anymore relying on two different codebases for Docker.
With the introduction of multiarch, debugging test issues will be a not be easy task, as now images from legacy and main are really disconnected.

heliocastro:
> @heliocastro please don't force-push to branches owned by others without discussing this first! Not even if it's a simple rebase.

Sorry for that. I didn't think that would be an issue.

sschuberth:
> I didn't think that would be an issue.

It breaks my checks which of my local branches are not in-sync with the remotes.

mnonnenmacher (approved):

mnonnenmacher:
As discussed on Slack I am approving that now despite my concerns as @heliocastro committed to making another PR that runs the functional tests with an image that matches the PR again.

PR #7860, opened 2023-11-14
fix(docker): No build or publish in pull_request events

heliocastro:


.github/workflows/docker-ort.yml
```diff
@@ -397,7 +399,7 @@ jobs:
         with:
           context: .
           target: run
-          push: true
+          push: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' || || github.event_name == 'schedule'}}
```
sschuberth, line 402: See above.

sschuberth (changes requested):

.github/workflows/docker-ort.yml
heliocastro: Fixed

sschuberth (approved):

PR #7863, opened 2023-11-15
docker: Align `Pip` versions and upgrade to the latest 

fviernau:
See individual commits.


Dockerfile-legacy
```diff
@@ -156,6 +157,8 @@ RUN /opt/ort/bin/export_proxy_certificates.sh /tmp/certificates/ && \
     # Install VCS tools (no specific versions required here).
     curl -ksS https://storage.googleapis.com/git-repo-downloads/repo > /usr/local/bin/repo && \
     chmod a+x /usr/local/bin/repo && \
+    # Upgrade pip.
+    pip install --no-cache-dir -U pip=="$PIPTOOL_VERSION" && \
```
sschuberth, line 161: Nits:
* Strictly speaking this could also be a "downgrade" depending on the default and configured versions. Maybe say "Install the configured version of pip.".
* How about moving this a few lines down to directly precede the other `pip install` lines?

sschuberth (changes requested):

sschuberth (approved):

PR #7864, opened 2023-11-15
test(scanner): Improve tests for `ScannerWrapperConfig`

mnonnenmacher:


scanner/src/test/kotlin/ScannerWrapperConfigTest.kt
```diff
@@ -59,5 +60,24 @@ class ScannerWrapperConfigTest : WordSpec({
...
+
+        "set unset storage options to null" {
+            with(ScannerWrapperConfig.create(emptyMap()).first) {
+                readFromStorage shouldBe null
+                writeToStorage shouldBe null
+            }
+        }
+
+        "fail if storage options are invalid" {
+            shouldThrow<IllegalArgumentException> {
```
fviernau, line 72: consider

```
listOf("1", "tru", "null").forAll {
    shouldThrow<IllegalArgumentException> {
          ScannerWrapperConfig.create(mapOf(ScannerWrapperConfig.PROP_READ_FROM_STORAGE to it))
    }
}

File: scanner/src/test/kotlin/ScannerWrapperConfigTest.kt [hunk 2]
```diff
@@ -59,5 +60,24 @@ class ScannerWrapperConfigTest : WordSpec({
 
             ScannerWrapperConfig.create(options).second shouldContainExactly mapOf("other" to "value")
         }
+
+        "set unset storage options to null" {
```
fviernau, line 64: I don't understand the test case name. Mind giving a hint?

fviernau (changes requested):

scanner/src/test/kotlin/ScannerWrapperConfigTest.kt
sschuberth: I guess you should read it like "storage option that are unset in the config should be explicitly set to null in the class properties".

mnonnenmacher: Is the new name better?

fviernau (approved):

PR #7865, opened 2023-11-15
 chore(docker): Extract .NET version to a variable

fviernau:
See individual commits.


fviernau:
I've tried upgrading to 8.0, but it seems that in order to analyze our test projects, .NET version 6.0 has to be installed.
E.g. with only 8.0, it fails with `No instances of MSBuild could be detected.`. So, we might need to install multiple .NET SDK versions in parallel.

sschuberth (approved):

PR #7876, opened 2023-11-16
test: Fixup the user home directory also when running tests (in Docker)

sschuberth:


heliocastro (approved):
Sounds's right

utils/test/src/main/kotlin/ProjectConfig.kt
```diff
@@ -23,12 +23,14 @@ import io.kotest.core.config.AbstractProjectConfig
...
 import io.kotest.extensions.junitxml.JunitXmlReporter
 
+import org.ossreviewtoolkit.utils.common.Os
 import org.ossreviewtoolkit.utils.ort.OrtProxySelector
 
 class ProjectConfig : AbstractProjectConfig() {
     override val specExecutionOrder = SpecExecutionOrder.Annotated
 
     init {
+        Os.fixupUserHomeProperty()
```
fviernau, line 33: Why is this necessary? (Does it fix any issue?)

fviernau (approved):

utils/test/src/main/kotlin/ProjectConfig.kt
sschuberth: It does not fix any issue in our current setup, but when using a different JDK (and maybe some other base image) then running tests in Docker could fail as the `user.home` Java system property bogusly is set to `?`. As this is rather hard to track down, I'm proactively addressing it here.

PR #7877, opened 2023-11-16
build: Align on tools from `.versions` also for the build / test workflow

sschuberth:


.github/workflows/build-and-test.yml
```diff
@@ -96,6 +92,8 @@ jobs:
     needs: build-web-app-reporter
     runs-on: ubuntu-22.04
     steps:
+    - name: Set tool version environment variables
```
fviernau, line 95: Does this commit actually align (change) any tool version?

sschuberth: Yes, the ScanCode version.

sschuberth: But weird that `.versions` is not found, as this step `needs: build-web-app-reporter`, so a checkout should be available...

sschuberth: Nevermind, different jobs do not share the workspace.

sschuberth:
Blocked by https://github.com/oss-review-toolkit/ort/pull/7883.

fviernau (approved):

PR #7883, opened 2023-11-16
fix: Remove any YAML front matter from ScanCode license files

sschuberth:
ScanCode 32.0.0 started to prepend its `*.LICENSE` files with YAML-encoded metadata, see [1]. This is a hot fix to remove this header, if present, from the license files. A better solution will be implemented later as part of a larger refactoring of license providers.

Different ScanCode versions also differ in whether license files come with a final newline or not. Align on not having a final newline to make tests pass either way.

[1]: https://github.com/nexB/scancode-toolkit/pull/3100

### DIFF
--- a/utils/spdx/src/main/kotlin/Utils.kt
+++ b/utils/spdx/src/main/kotlin/Utils.kt
@@ -130,8 +130,12 @@ fun getLicenseTextReader(
 ): (() -> String)? {
     return if (id.startsWith(LICENSE_REF_PREFIX)) {
         getLicenseTextResource(id)?.let { { it.readText() } }
-            ?: addScanCodeLicenseTextsDir(licenseTextDirectories).firstNotNullOfOrNull {
-                getLicenseTextFile(id, it)?.let { file -> { file.readText() } }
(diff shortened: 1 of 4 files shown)


utils/spdx/src/main/kotlin/Utils.kt
```diff
@@ -130,8 +130,18 @@ fun getLicenseTextReader(
...
+                    {
+                        val lines = file.readLines()
+
+                        // Remove any YAML front matter enclosed by "---" from ScanCode license files.
+                        val licenseLines = lines.takeUnless { it.first() == "---" }
+                            ?: lines.drop(1).dropWhile { it != "---" }.drop(1)
+
+                        licenseLines.joinToString("\n").trim()
+                    }
+                }
```
fviernau, line 144: Would it make sense to extract lines 134-144 into a functions e.g. `getScanCodeLicenseText(id, dir)`
and add a test which calls it for a file which has such YAML in the header, to show it works?

sschuberth: Done. I've actually improved the algorithm a bit as part of adding the test.

fviernau (approved):

PR #7888, opened 2023-11-17
fix(analyzer): Support uppercase-letters in Go module version

wkl3nk:
Uppercase-letters in version strings of dependent Go modules caused the analyzer to crash. 
With this fix, uppercase-letters are now properly escaped as any other paths of the Go modules in the file system.

### DIFF
--- a/plugins/package-managers/go/src/main/kotlin/GoMod.kt
+++ b/plugins/package-managers/go/src/main/kotlin/GoMod.kt
@@ -454,7 +454,8 @@ private fun ModuleInfo.toSourceArtifact(): RemoteArtifact {
 }
 
 private fun ModuleInfo.toVcsInfo(): VcsInfo? {
-    val infoFile = goMod?.let { File(it).resolveSibling("$version.info") } ?: return null
+    val escapedVersion = escapeVersion(version)
+    val infoFile = goMod?.let { File(it).resolveSibling("$escapedVersion.info") } ?: return null
     val info = infoFile.inputStream().use { JSON.decodeFromStream<ModuleInfoFile>(it) }
     val type = info.origin.vcs?.let { VcsType.forName(it) }.takeIf { it == VcsType.GIT } ?: return null
 
@@ -484,3 +485,19 @@ internal fun parseWhyOutput(output: String): Set<String> {
 
     return usedModules
 }
+
+/**
+ * Module paths appear as substrings of file system paths in the module cache on the file system.
+ * Go does not rely on the file system to be case-sensitive. For this reason, Go has decided to
+ * replace every uppercase letter in file system paths with an exclamation mark followed by the
+ * letter's lowercase equivalent.
+ *
+ * Details behind the reasoning and implementation in Go can be found in the Go source code at
+ * [module.go](https://github.com/golang/go/blob/5b6d3dea8744311825fd544a73edb8d26d9c7e98/src/cmd/vendor/golang.org/x/mod/module/module.go#L33-L42C64)
+ */
+internal fun escapeVersion(version: String): String {
+    require("!" !in version) { "Module versions must not contain exclamation marks: $version" }
+    return version.replace(upperCaseCharRegex) { "!${it.value.lowercase()}" }
+}
+
+private val upperCaseCharRegex = Regex("[A-Z]")
--- a/plugins/package-managers/go/src/test/kotlin/GoModTest.kt
+++ b/plugins/package-managers/go/src/test/kotlin/GoModTest.kt
@@ -19,10 +19,12 @@
 
 package org.ossreviewtoolkit.plugins.packagemanagers.go
 
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.core.spec.style.WordSpec
 import io.kotest.matchers.collections.beEmpty
 import io.kotest.matchers.collections.containExactlyInAnyOrder
 import io.kotest.matchers.should
+import io.kotest.matchers.shouldBe
 
 class GoModTest : WordSpec({
     "parseWhyOutput" should {
@@ -67,6 +69,23 @@ class GoModTest : WordSpec({
                 "cloud.google.com/go",
                 "github.com/fatih/color"
             )
+        }
+    }
+
+    "escapeVersion" should {
+        "escape uppercase letters" {
+            val version = "v0.1.0-MS4.0.20231102094829-08e0c3cd016c"
+            val escapedVersion = escapeVersion(version)
+
+            escapedVersion shouldBe "v0.1.0-!m!s4.0.20231102094829-08e0c3cd016c"
+        }
+
+        "throw exception if version string contains an exclamation mark" {
+            val version = "v1.0.0!"
+
+            shouldThrow<IllegalArgumentException> {
+                escapeVersion(version)
+            }
         }
     }
 })


plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
```diff
@@ -25,3 +25,24 @@ package org.ossreviewtoolkit.plugins.packagemanagers.go.utils
  * https://go.dev/ref/mod#incompatible-versions.
  */
 fun normalizeModuleVersion(moduleVersion: String): String = moduleVersion.removePrefix("v").substringBefore("+")
+
```
fviernau, line 28: This file only contains functions shared between `GoDep` and `GoMod`. For cohesion, can we move this into `GoMod` and reduce the visibility to `internal`?

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
```diff
@@ -25,3 +25,24 @@ package org.ossreviewtoolkit.plugins.packagemanagers.go.utils
...
+/**
+ * Module paths appear as substrings of file system paths in the module cache on the file system.
+ * Go does not rely on the file system to be case-sensitive. For this reason, Go has decided to
+ * replace every uppercase letter in file system paths with an exclamation mark followed by the
+ * letter's lowercase equivalent.
```
fviernau, line 33: It'd be nice to have a reference included in this comment. Do you maybe have one at hand?

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
```diff
@@ -25,3 +25,24 @@ package org.ossreviewtoolkit.plugins.packagemanagers.go.utils
...
+ */
+fun escapeVersion(version: String): String {
+    require(!version.contains("!")) { "Module versions must not contain exclamation marks: $version"}
+
+    val escaped = StringBuilder()
```
fviernau, line 38: please use instead of `StringBuilder` this:

```
buildString {
    append(..)
}
```

nit: maybe a to more compact could be to use the `map` function instead, up to you to decide.

plugins/package-managers/go/src/main/kotlin/GoMod.kt
```diff
@@ -454,7 +455,8 @@ private fun ModuleInfo.toSourceArtifact(): RemoteArtifact {
 }
 
 private fun ModuleInfo.toVcsInfo(): VcsInfo? {
-    val infoFile = goMod?.let { File(it).resolveSibling("$version.info") } ?: return null
+    val escapedVersion = escapeVersion(version)
```
fviernau, line 458: To increase readability I believe we should define an extension property:

```
var ModuleInfo.escapedVersion: String
    get() = escapeVersion(version)
```

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.ossreviewtoolkit.plugins.packagemanagers.go.utils
+
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldContainIgnoringCase
+
+class GoSupportFunTest : WordSpec({
```
fviernau, line 27: I believe this is more like a unit test, so it should not be under `funTest/kotlin/utils` but
instead under `test/kotlin/..`. The class name also needs to change to not have the term `Fun` included.

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
```diff
@@ -0,0 +1,45 @@
...
+    "escapeVersion()" should {
+        "escape uppercase letters from a version string" {
```
fviernau, line 29: This should be renamed to just "escape uppercase letters" because "from a version string" is redundant.

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
```diff
@@ -0,0 +1,45 @@
...
+            val version = "v0.1.0-M4.0.20231102094829-08e0c3cd016c"
+            val escapedVersion = escapeVersion(version)
+            escapedVersion shouldBe "v0.1.0-!m4.0.20231102094829-08e0c3cd016c"
+        }
+
+        "throw exception is exclamation mark is used in version string" {
```
fviernau, line 35: How about "throw given a version string containing an exclamation mark"?

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
```diff
@@ -0,0 +1,45 @@
...
+            val version = "v1.0.0-finally-it-works!"
```
fviernau, line 36: nit: I propose to make the version string just "v1.0.0!"

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
```diff
@@ -0,0 +1,45 @@
...
+
+            val exception = shouldThrow<IllegalArgumentException> {
+                escapeVersion(version)
+            }
+
+            exception.message shouldContainIgnoringCase "exclamation mark"
```
fviernau, line 42: I'd prefer to drop the assertion in line 42 for resilience of the test.

sschuberth:
Thanks for the contribution! Please add `Fixes #7880.` as a trailing paragraph to the commit message body in order to auto-close the issue upon merge.

fviernau (changes requested):

fviernau:
Nice finding, thank @wkl3nk !

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
```diff
@@ -0,0 +1,45 @@
+/*
```
sschuberth, line 1: Commit message:

> are now properly escaped

Would you have a reference to some Go documentation web page that defines the escaping rules?

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldContainIgnoringCase
+
+class GoSupportFunTest : WordSpec({
+    "escapeVersion()" should {
+        "escape uppercase letters from a version string" {
+            val version = "v0.1.0-M4.0.20231102094829-08e0c3cd016c"
+            val escapedVersion = escapeVersion(version)
```
sschuberth, line 31: Nit: Add empty lines before and after this line to resemble the AAA pattern.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
```diff
@@ -25,3 +25,24 @@ package org.ossreviewtoolkit.plugins.packagemanagers.go.utils
...
  */
 fun normalizeModuleVersion(moduleVersion: String): String = moduleVersion.removePrefix("v").substringBefore("+")
+
+/**
+ * Module paths appear as substrings of file system paths in the module cache on the file system.
+ * Go does not rely on the file system to be case-sensitive. For this reason, Go has decided to
+ * replace every uppercase letter in file system paths with an exclamation mark followed by the
+ * letter's lowercase equivalent.
+ */
+fun escapeVersion(version: String): String {
```
sschuberth, line 35: Please make this `internal` to avoid it becoming part of the public API beyond this module. Unrelated, but I'd appreciate if you could add another commit to this PR that also makes the above `normalizeModuleVersion` an `internal` function for consistency.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
```diff
@@ -25,3 +25,24 @@ package org.ossreviewtoolkit.plugins.packagemanagers.go.utils
...
+    require(!version.contains("!")) { "Module versions must not contain exclamation marks: $version"}
```
sschuberth, line 36: Syntax: `require("!" !in version)`

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
sschuberth, line 27: This should probably rather be a unit test than a functional test.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
```diff
@@ -25,3 +25,24 @@ package org.ossreviewtoolkit.plugins.packagemanagers.go.utils
...
+    val escaped = StringBuilder()
+    for (char in version) {
+        if (char.isUpperCase()) {
+            escaped.append("!${char.lowercaseChar()}")
+        } else {
+            escaped.append(char)
+        }
+    }
+
+    return escaped.toString()
```
sschuberth, line 47: Alternative implementation:

    return version.replace(Regex("[A-Z]")) {
        "!${it.value.lowercase()}"
    }

This could also extract `Regex("[A-Z]")` to a private property to the Regex is not recreated on each function call.

sschuberth (changes requested):

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
wkl3nk: Agreed. Moved "escapeVersion" to GoMod.kt with `internal` visiblity.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
wkl3nk: Yes. Added a link to the Go sources where it is described.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
wkl3nk: Done.

plugins/package-managers/go/src/main/kotlin/GoMod.kt
wkl3nk: Kept as it is. This Kotlin Woodoo confuses me. I think the readabliity is already good.

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
wkl3nk: Agreed. Moved to GoModTest.kt where there is already another `internal` helper function.

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
wkl3nk: Agreed.

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
wkl3nk: Agreed. Where was a typo. Should have been "if" instead of "is".

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
wkl3nk: Good argument.

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
wkl3nk: Done. Added link to go sources where the rules for escaping are described.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
wkl3nk: Done.

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
wkl3nk: Moved to GoModTest.kt where there is already another internal function to test.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
wkl3nk: Come on. Don't need a Regex for such an easy thing.

plugins/package-managers/go/src/main/kotlin/GoMod.kt
```diff
@@ -19,43 +19,28 @@
...
-import org.ossreviewtoolkit.model.Package
-import org.ossreviewtoolkit.model.PackageLinkage
-import org.ossreviewtoolkit.model.PackageReference
-import org.ossreviewtoolkit.model.Project
-import org.ossreviewtoolkit.model.ProjectAnalyzerResult
-import org.ossreviewtoolkit.model.RemoteArtifact
-import org.ossreviewtoolkit.model.Scope
-import org.ossreviewtoolkit.model.VcsInfo
-import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.model.*
```
fviernau, line 31: The imports will be flagged by static analysis. You may configure the IDE not to collapse
the inputs into some wildcard.

wkl3nk:
@sschuberth  @fviernau  Thanks for your comments, would you have a look at the PR again, please? And don't be too hard, I did not code Kotlin before ...

plugins/package-managers/go/src/main/kotlin/GoMod.kt
```diff
@@ -484,3 +470,28 @@ internal fun parseWhyOutput(output: String): Set<String> {
...
+
+/**
+ * Module paths appear as substrings of file system paths in the module cache on the file system.
+ * Go does not rely on the file system to be case-sensitive. For this reason, Go has decided to
+ * replace every uppercase letter in file system paths with an exclamation mark followed by the
+ * letter's lowercase equivalent.
+ *
+ * Details behind the reasoning and implementation in Go can be found in the Go source code at
+ * [module.go](https://github.com/golang/go/blob/master/src/cmd/vendor/golang.org/x/mod/module/module.go)
+ */
```
fviernau, line 482: We usually use perma-links, so that these do not  break. I propose to also include some line range as the file is large. Would the following fit?

https://github.com/golang/go/blob/5b6d3dea8744311825fd544a73edb8d26d9c7e98/src/cmd/vendor/golang.org/x/mod/module/module.go#L33-L42C64

plugins/package-managers/go/src/test/kotlin/GoModTest.kt
```diff
@@ -69,4 +71,19 @@ class GoModTest : WordSpec({
             )
         }
     }
+    "escapeVersion" should {
```
fviernau, line 74: Please add an empty line before this one.

plugins/package-managers/go/src/test/kotlin/GoModTest.kt
```diff
@@ -69,4 +71,19 @@ class GoModTest : WordSpec({
...
+        "escape uppercase letters" {
+            val version = "v0.1.0-M4.0.20231102094829-08e0c3cd016c"
+            val escapedVersion = escapeVersion(version)
+            escapedVersion shouldBe "v0.1.0-!m4.0.20231102094829-08e0c3cd016c"
```
fviernau, line 78: Please add an empty line below line 77.

fviernau (changes requested):

plugins/package-managers/go/src/main/kotlin/GoMod.kt
wkl3nk: Converted it back.

plugins/package-managers/go/src/main/kotlin/GoMod.kt
wkl3nk: Done

plugins/package-managers/go/src/test/kotlin/GoModTest.kt
wkl3nk: Done

plugins/package-managers/go/src/test/kotlin/GoModTest.kt
wkl3nk: Done.

fviernau (dismissed):

plugins/package-managers/go/src/funTest/kotlin/utils/GoSupportFunTest.kt
wkl3nk: Done.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
wkl3nk: Done. Extra commit added to this PR.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
```diff
@@ -24,4 +24,4 @@ package org.ossreviewtoolkit.plugins.packagemanagers.go.utils
  * also any suffix starting with '+', because build metadata is not involved in version comparison according to
  * https://go.dev/ref/mod#incompatible-versions.
  */
-fun normalizeModuleVersion(moduleVersion: String): String = moduleVersion.removePrefix("v").substringBefore("+")
+internal fun normalizeModuleVersion(moduleVersion: String): String = moduleVersion.removePrefix("v").substringBefore("+")
```
fviernau, line 27: IMO the consistency argument (as written) has vanished a bit, since `escapeVersion()` no more is in this file. 
In my view, reducing visibility does not necessarily be done in this PR. 

However, the commit message should adhere to conventional commits.

I propose to just say

```
refactor(go)!: Reduce the visibility of `normalizeModuleVersion()`

There is no need to expose this function.
```

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
sschuberth: Sure you don't. But I'd bet it's even faster (if the Regex is precompiled to a constant) and it's more readable (or at least shorter, if you're not used to Regexes) than your solution that manually iterates over each character in the string. But I'm also fine with keeping your approach.

plugins/package-managers/go/src/main/kotlin/GoMod.kt
```diff
@@ -484,3 +485,28 @@ internal fun parseWhyOutput(output: String): Set<String> {
...
+ * replace every uppercase letter in file system paths with an exclamation mark followed by the
+ * letter's lowercase equivalent.
+ *
+ * Details behind the reasoning and implementation in Go can be found in the Go source code at
+ * [module.go](https://github.com/golang/go/blob/5b6d3dea8744311825fd544a73edb8d26d9c7e98/src/cmd/vendor/golang.org/x/mod/module/module.go#L33-L42C64)
+ */
+internal fun escapeVersion(version: String): String {
+    require("!" !in version) { "Module versions must not contain exclamation marks: $version"}
+
+    val escaped = buildString {
```
sschuberth, line 501: Nit: You can omit the `escaped` variable and directly return from here: `return buildString {`.

plugins/package-managers/go/src/test/kotlin/GoModTest.kt
```diff
@@ -69,4 +71,21 @@ class GoModTest : WordSpec({
             )
         }
     }
+
+    "escapeVersion" should {
+        "escape uppercase letters" {
+            val version = "v0.1.0-M4.0.20231102094829-08e0c3cd016c"
+            val escapedVersion = escapeVersion(version)
```
sschuberth, line 78: Nit: The empty line before this one is still missing (the code that performs the action to test should stand by its own).

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
wkl3nk: Okay, using a Regex now, extended the test case to make sure **all** patterns matching are replaced.

plugins/package-managers/go/src/main/kotlin/utils/GoSupport.kt
wkl3nk: Yes. Removed the change, so this way sticking to just **one** commit in the PR.

plugins/package-managers/go/src/main/kotlin/GoMod.kt
wkl3nk: Done.

plugins/package-managers/go/src/test/kotlin/GoModTest.kt
wkl3nk: Done.

sschuberth (approved):
LGTM, I'll fixup some nits as a follow-up.

fviernau (approved):

sschuberth:
> LGTM, I'll fixup some nits as a follow-up.

I'll make my nit-fixes in-place in this PR as detekt fails.

fviernau (approved):

PR #7892, opened 2023-11-17
Make the evaluator command print the rules used as part of configuration 

sschuberth:


sschuberth:
Example output:

image

PR #7903, opened 2023-11-20
fix(opossum): Get license texts via the provider

sschuberth:
Do not directly call the low-level `getLicenseText()` function, but get the license texts via the provider passed as part of reporter input.


fviernau (approved):

sschuberth:
Merging despite the unrelated (temporary) test failures.

PR #7909, opened 2023-11-21
fix(helper-cli): Use the "pluginClasspath" approach to bundle plugins

sschuberth:
Use the same approach as for the main CLI to ensure that plugins required at runtime are bundled with the distribution.

### DIFF
--- a/helper-cli/build.gradle.kts
+++ b/helper-cli/build.gradle.kts
@@ -22,6 +22,9 @@ plugins {
     id("ort-application-conventions")
 }
 
+configurations.dependencyScope("pluginClasspath")
+configurations["runtimeClasspath"].extendsFrom(configurations["pluginClasspath"])
+
 application {
     applicationName = "orth"
     mainClass = "org.ossreviewtoolkit.helper.HelperMainKt"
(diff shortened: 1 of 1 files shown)


helper-cli/build.gradle.kts
```diff
@@ -43,4 +44,8 @@ dependencies {
     implementation(libs.jslt)
     implementation(libs.log4jApi)
     implementation(libs.slf4j)
+
+    "pluginClasspath"(platform(project(":plugins:package-configuration-providers")))
+    "pluginClasspath"(platform(project(":plugins:package-curation-providers")))
+    "pluginClasspath"(platform(project(":plugins:version-control-systems")))
```
sschuberth, line 50: @fviernau if there are more [plugin types](https://github.com/oss-review-toolkit/ort/tree/main/plugins) the helper CLI requires at runtime, they also need to be added here.

fviernau: I'll need to investigate this a bit. Shall we merge this one already before?

fviernau (dismissed):

fviernau (approved):

helper-cli/build.gradle.kts
sschuberth: I first need to fixup an compile issue. I'd say then test it, and merge yourself if your find it to resolve your issue.

fviernau: Thank you! I've just tested it, and verified it successfully downloads via `git` now.

fviernau (approved):

PR #7910, opened 2023-11-21
docs(jenkins): Clarify that the credentials type should be for HTTP

sschuberth:
Later steps use the `usernamePassword()` pipeline function [1] which assumes credentials to be composed of a username and password for HTTP authentication. Clarify this to avoid users specifying e.g. SSH credentials which would result in

    ERROR: Credentials 'jenkins-ssh' is of type 'SSH Username with private
    key' where 'com.cloudbees.plugins.credentials.common.StandardUsername
    PasswordCredentials' was expected

[1]: https://www.jenkins.io/doc/pipeline/steps/credentials-binding/

### DIFF
--- a/integrations/jenkins/Jenkinsfile
+++ b/integrations/jenkins/Jenkinsfile
@@ -62,7 +62,7 @@ pipeline {
 
         credentials(
             name: 'PROJECT_VCS_CREDENTIALS',
-            description: 'Optional Jenkins credentials to use for the VCS checkout.',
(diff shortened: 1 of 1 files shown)


integrations/jenkins/Jenkinsfile
mnonnenmacher, line 1: Commit message nit: Dot missing after "e.g"

mnonnenmacher (approved):

PR #7913, opened 2023-11-21
Allow to specify a VCS path for configuration in the Jenkins pipeline

sschuberth:

### DIFF
--- a/integrations/jenkins/Jenkinsfile
+++ b/integrations/jenkins/Jenkinsfile
@@ -413,7 +413,7 @@ pipeline {
 
                             currentBuild.displayName += ": $rootProjectId"
                         }
-                    } catch (IOException e) {
+                    } catch (IOException ignored) {
                         // Ignore and just skip setting a custom display name.
                     }
                 }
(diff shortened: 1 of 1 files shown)


sschuberth:
FYI @mmurto.

integrations/jenkins/Jenkinsfile
```diff
@@ -334,25 +340,31 @@ pipeline {
...
-                          enabled: ${USE_CLEARLY_DEFINED_CURATIONS}
-                    EOF
-                    fi
-
                     rm -f $HOME/.netrc
                     '''.stripIndent().trim()
                 }
+
+                script {
+                    env.ORT_CONFIG_DIR = "${env.ORT_DATA_DIR}/config/${env.ORT_CONFIG_VCS_PATH}"
```
fviernau, line 348: Should a trailing slash be stripped to avoid double slashes in the interpolated strings below, e.g. line 253?

sschuberth: I've verified that this works as expected also if `ORT_CONFIG_VCS_PATH` is empty, so I'd prefer to omit that logic for simplicity.

fviernau (approved):

PR #7915, opened 2023-11-22
feat(model): Sort the detected license mapping during serialization

mnonnenmacher:
Sort `ScannerConfiguration.detectedLicenseMapping` during serialization to have a stable order in ORT result files. This helps when comparing ORT result files from different sources.

### DIFF
--- a/model/src/main/kotlin/config/ScannerConfiguration.kt
+++ b/model/src/main/kotlin/config/ScannerConfiguration.kt
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.model.config
 
 import com.fasterxml.jackson.annotation.JsonAlias
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonPropertyOrder
 
 import org.ossreviewtoolkit.model.utils.FileArchiver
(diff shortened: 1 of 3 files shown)


model/src/main/kotlin/utils/SortedSetConverters.kt
```diff
@@ -92,6 +93,10 @@ class SnippetFindingSortedSetConverter : StdConverter<Set<SnippetFinding>, Sorte
     override fun convert(value: Set<SnippetFinding>) = value.toSortedSet(compareBy { it.sourceLocation })
 }
 
+class StringSortedMapConverter : StdConverter<Map<String, String>, SortedMap<String, String>>() {
```
fviernau, line 96: Would `@JsonPropertyOrder(alphabetic = true)` work instead?

fviernau (changes requested):

model/src/main/kotlin/utils/SortedSetConverters.kt
mnonnenmacher: Yes, it does.

sschuberth (approved):

fviernau (approved):

PR #7916, opened 2023-11-22
Minor StaticHTML reporter code improvements

fviernau:
See individual commits.


sschuberth:
> Once done, I plan to do some improvements to the StaticHtml report.

generally, I'd be interested for what features you're still using it, and whether it makes more sense to migrate any missing feature to the web app reporter instead of adding more feature to the static HTML reporter. Maybe a good topic for the community meeting to get a feeling for which reasons, if at all, people are still using the static report.

PR #7918, opened 2023-11-22
Teach package scanners about non-reference packages

sschuberth:

### DIFF
--- a/scanner/src/main/kotlin/ScanContext.kt
+++ b/scanner/src/main/kotlin/ScanContext.kt
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.scanner
 
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageType
 import org.ossreviewtoolkit.model.config.Excludes
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
@@ -53,5 +54,11 @@ data class ScanContext(
      * be used by scanners where scan results are stored, because then changes in the mapping would not be applied to
      * stored results.
      */
-    val detectedLicenseMapping: Map<String, String> = emptyMap()
+    val detectedLicenseMapping: Map<String, String> = emptyMap(),
+
+    /**
+     * The packages known to be covered in the context of this scan. For package scanners, this is the list of packages
+     * that have the same provenance as the reference package.
+     */
+    val coveredPackages: List<Package> = emptyList()
 )
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -356,9 +356,14 @@ class Scanner(
                     "Starting scan of '${referencePackage.id.toCoordinates()}' with package scanner '${scanner.name}."
                 }
 
-                // Filter the scan context to hide the excludes from scanner with scan matcher.
-                val filteredContext = if (scanner.matcher == null) context else context.copy(excludes = null)
-                val scanResult = scanner.scanPackage(referencePackage, filteredContext)
+                val adjustedContext = context.copy(
+                    // Hide excludes from scanners with a scanner matcher.
+                    excludes = context.excludes.takeUnless { scanner.matcher != null },
+                    // Tell scanners also about the non-reference packages.
+                    coveredPackages = packagesWithIncompleteScanResult
+                )
+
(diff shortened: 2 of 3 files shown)


sschuberth:
FYI @Etsija.

scanner/src/main/kotlin/ScanContext.kt
```diff
@@ -21,6 +21,7 @@ package org.ossreviewtoolkit.scanner
 
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.OrtResult
+import org.ossreviewtoolkit.model.Package
```
fviernau, line 24: commit-msg: Please provide a rationale _why_ some package scanners need to know that.

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -356,9 +356,15 @@ class Scanner(
                     "Starting scan of '${referencePackage.id.toCoordinates()}' with package scanner '${scanner.name}."
                 }
 
-                // Filter the scan context to hide the excludes from scanner with scan matcher.
-                val filteredContext = if (scanner.matcher == null) context else context.copy(excludes = null)
-                val scanResult = scanner.scanPackage(referencePackage, filteredContext)
+                val adjustedContext = context.run {
+                    // Hide excludes from scanners with a scanner matcher.
```
fviernau, line 360: nit: Consider re-write as follows?

```
val adjustedContext = context.copy(
       excludes = context.excludes.takeUnless { scanner.matcher != null },
       coveredPackages = packagesWithIncompleteScanResult
)
```

scanner/src/main/kotlin/ScanContext.kt
```diff
@@ -53,5 +54,11 @@ data class ScanContext(
...
      * stored results.
      */
-    val detectedLicenseMapping: Map<String, String> = emptyMap()
+    val detectedLicenseMapping: Map<String, String> = emptyMap(),
+
+    /**
+     * The packages known to be covered in the context of this scan. For package scanners, this is the list of packages
+     * that have the same provenance as the reference package.
+     */
+    val coveredPackages: List<Package> = emptyList()
```
fviernau, line 63: @mnonnenmacher Are package scanners actually more like a deprecated / phased out thing, or is there plans to keep them? (I wonder if it would make sense to introduce a dedicated `PackageScanContext` class, to not pollute the provenance scan user case with these attributes.)

fviernau (changes requested):

scanner/src/main/kotlin/Scanner.kt
```diff
@@ -356,9 +356,14 @@ class Scanner(
...
-                val filteredContext = if (scanner.matcher == null) context else context.copy(excludes = null)
-                val scanResult = scanner.scanPackage(referencePackage, filteredContext)
+                val adjustedContext = context.copy(
+                    // Hide excludes from scanners with a scanner matcher.
+                    excludes = context.excludes.takeUnless { scanner.matcher != null },
+                    // Tell scanners also about the non-reference packages.
+                    coveredPackages = packagesWithIncompleteScanResult
+                )
+
+                val scanResult = scanner.scanPackage(referencePackage, adjustedContext)
```
fviernau, line 366: Related to my comment in line 63: We could alternatively pass the `coveredPackages` as additional parameter to `scanPackage()` or pass a dedicated `PackageScanContext` class, so that we don't need to worry in other places about this new property.  What do you think?

sschuberth: > pass the `coveredPackages` as additional parameter to `scanPackage()`

As indicated in the commit message, I've not done it that way to avoid a breaking API change in the unction signature. Also, including `coveredPackages` into `ScanContext` paces the way for eventually also teach `ProvenanceScannerWrapper`s which packages are contained in the passed `provenance`.

scanner/src/main/kotlin/ScanContext.kt
mnonnenmacher: There are currently no plans to remove `PackageScannerWrapper`, at least as long as `FossId` is not migrated.

scanner/src/main/kotlin/Scanner.kt
mnonnenmacher: I think there's no general problem with having properties in the `ScanContext` that are not used by all scanners. For example, `ScanContext.labels` is only used by `FossId` and other scanners can simply ignore it.

mnonnenmacher (approved):

fviernau (approved):

sschuberth:
Merging independently of the unrelated Docker image build failure.

PR #7920, opened 2023-11-23
chore(reporter): Remove the unused `SummaryTable`

fviernau:


reporter/src/main/kotlin/ReportTableModel.kt
```diff
@@ -61,11 +59,6 @@ data class ReportTableModel(
      */
     val issueSummary: IssueTable,
 
-    /**
-     * A [SummaryTable] containing the dependencies of all [Project]s.
-     */
-    val summary: SummaryTable,
```
sschuberth, line 67: As `ReportTableModel` is public, would this need to be marked as breaking API change again?

sschuberth (commented):
@mnonnenmacher do you recall what the `SummaryTable` was used for? Are there internals reporters on your side that maybe make use of it?

sschuberth:
> @mnonnenmacher do you recall what the `SummaryTable` was used for? Are there internals reporters on your side that maybe make use of it?

@mnonnenmacher confirmed in the community meeting that this is not the case.

sschuberth (approved):

PR #7917, opened 2023-11-22
Fix `user.home` when running ORT via Jenkins in a Docker-in-Docker scenario

sschuberth:
Please have a look at the individual commit messages for the details.

PR #7924, opened 2023-11-23
feat(helper-cli): Extend path exclude generator by a couple of patterns

tsteenbe:
Add patterns for common CI configuration, build or test files and dirs.


fviernau (approved):

sschuberth:
Commitlint cannot cope with multiple signed-off footers, unfortunately (same with co-authored-by footers). The pragmatic solution is to simply remove the signoff for the non-author, and eventually use the other user as the committer.

fviernau:
> Commitlint cannot cope with multiple signed-off footers, unfortunately (same with co-authored-by footers). The pragmatic solution is to simply remove the signoff for the non-author, and eventually use the other user as the committer.

Ok if we just merge regardless of this failing check ?

sschuberth:
> Ok if we just merge regardless of this failing check ?

I'm not sure if this will result in problems with future PRs... maybe Commitlint looks at all the history instead of just the commits introduced in a PR. So I'd prefer to reword the commit messages, like I also did [here](https://github.com/oss-review-toolkit/ort/pull/7917#issuecomment-1822990085).

PR #7937, opened 2023-11-27
feat(CompareCommand): Add a new compare method `SEMANTIC_DIFF`

bs-ondem:
Check individual commit messages for details.

### DIFF
--- a/plugins/commands/compare/src/main/kotlin/CompareCommand.kt
+++ b/plugins/commands/compare/src/main/kotlin/CompareCommand.kt
@@ -34,6 +34,7 @@ import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.types.enum
 import com.github.ajalt.clikt.parameters.types.file
+import com.github.ajalt.mordant.rendering.Theme
 import com.github.difflib.DiffUtils
 import com.github.difflib.UnifiedDiffUtils
 
@@ -105,6 +106,22 @@ class CompareCommand : OrtCommand(
         val resultB = deserializer.readValue<OrtResult>(fileB)
 
         when (method) {
+            CompareMethod.SEMANTIC_DIFF -> {
+                echo(
+                    Theme.Default.warning(
+                        "The '${CompareMethod.SEMANTIC_DIFF}' compare method is not fully implemented. Some " +
+                            "properties may not be taken into account in the comparison."
+                    )
+                )
+
+                if (resultA == resultB) {
+                    echo("The ORT results are the same.")
+                    throw ProgramResult(0)
+                }
+
+                throw ProgramResult(1)
+            }
+
             CompareMethod.TEXT_DIFF -> {
                 val textA = deserializer.writeValueAsString(resultA)
                 val textB = deserializer.writeValueAsString(resultB)
@@ -149,6 +166,7 @@ class CompareCommand : OrtCommand(
 }
 
 private enum class CompareMethod {
+    SEMANTIC_DIFF,
     TEXT_DIFF
 }
 
--- a/plugins/commands/compare/src/main/kotlin/CompareCommand.kt
+++ b/plugins/commands/compare/src/main/kotlin/CompareCommand.kt
@@ -92,19 +92,19 @@ class CompareCommand : OrtCommand(
             throw ProgramResult(2)
         }
 
+        val deserializer = fileA.mapper()
+        val resultA = deserializer.readValue<OrtResult>(fileA)
+        val resultB = deserializer.readValue<OrtResult>(fileB)
+
         when (method) {
             CompareMethod.TEXT_DIFF -> {
                 // Reserialize file contents with some data types replaced by invariant strings.
-                val deserializer = fileA.mapper()
                 val serializer = deserializer.copy().registerModule(
                     SimpleModule().apply {
                         if (ignoreTime) addSerializer(InvariantInstantSerializer())
                         if (ignoreEnvironment) addSerializer(InvariantEnvironmentSerializer())
                     }
                 )
-
-                val resultA = deserializer.readValue<OrtResult>(fileA)
-                val resultB = deserializer.readValue<OrtResult>(fileB)
 
                 val textA = serializer.writeValueAsString(resultA)
                 val textB = serializer.writeValueAsString(resultB)
--- a/plugins/commands/compare/src/main/kotlin/CompareCommand.kt
+++ b/plugins/commands/compare/src/main/kotlin/CompareCommand.kt
@@ -19,10 +19,11 @@
 
 package org.ossreviewtoolkit.plugins.commands.compare
 
-import com.fasterxml.jackson.core.JsonGenerator
-import com.fasterxml.jackson.databind.SerializerProvider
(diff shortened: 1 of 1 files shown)


plugins/commands/compare/src/main/kotlin/CompareCommand.kt
```diff
@@ -92,7 +96,14 @@ class CompareCommand : OrtCommand(
             throw ProgramResult(2)
         }
 
-        val deserializer = fileA.mapper()
+        val deserializer = fileA.mapper().registerModule(
```
sschuberth, line 118: Note that making the results invariant already on deserialization should also work for the existing `TEXT_DIFF` method, which currently makes results only invariant on serialization.

That is, with this code, I believe the existing code

                val serializer = deserializer.copy().registerModule(
                    SimpleModule().apply {
                        if (ignoreTime) addSerializer(InvariantInstantSerializer())
                        if (ignoreEnvironment) addSerializer(InvariantEnvironmentSerializer())
                    }
                )

and the referenced serializers should not be required anymore.

Phrased differently, we should make results invariant either on deserialization or on serialization, but not both.

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
```diff
@@ -158,12 +169,22 @@ private class InvariantInstantSerializer : StdSerializer<Instant>(Instant::class
     }
 }
 
+private class EpochInstantDeserializer : StdDeserializer<Instant>(Instant::class.java) {
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): Instant =
+        Instant.EPOCH.also { parser.codec.readTree<JsonNode>(parser) }
```
sschuberth, line 338: Doesn't this need to say `Instant` instead of `JsonNode`, so the parser is advanced for the right amount of properties for the respective type?

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
```diff
@@ -158,12 +169,22 @@ private class InvariantInstantSerializer : StdSerializer<Instant>(Instant::class
...
+
 private class InvariantEnvironmentSerializer : StdSerializer<Environment>(Environment::class.java) {
     override fun serialize(value: Environment, gen: JsonGenerator, provider: SerializerProvider) {
         gen.writeString("invariant")
     }
 }
 
+private class DefaultEnvironmentDeserializer : StdDeserializer<Environment>(Environment::class.java) {
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): Environment =
+        Environment().also { parser.codec.readTree<JsonNode>(parser) }
```
sschuberth, line 349: See above.

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
```diff
@@ -155,12 +155,27 @@ class CompareCommand : OrtCommand(
 
                 throw ProgramResult(1)
             }
+
+            CompareMethod.SEMANTIC_DIFF -> {
+                echo(
+                    "The '${CompareMethod.SEMANTIC_DIFF}' compare method is not fully implemented. Some properties" +
```
sschuberth, line 180: Nit: Maybe print this formatted with `Theme.Default.warning`?

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
```diff
@@ -155,12 +155,27 @@ class CompareCommand : OrtCommand(
...
+                throw ProgramResult(1)
+            }
         }
     }
 }
 
 private enum class CompareMethod {
-    TEXT_DIFF
+    TEXT_DIFF,
+    SEMANTIC_DIFF
```
sschuberth, line 327: Nit: Should we maintain alphabetical sorting here (and also in the `when` above)?

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
```diff
@@ -167,12 +186,142 @@ class CompareCommand : OrtCommand(
...
         }
     }
 }
 
+private fun OrtResult.diff(other: OrtResult) =
+    OrtResultDiff(
+        repositoryDiff = repository.diff(other.repository)
+    )
+
+private fun Repository.diff(other: Repository): RepositoryDiff? =
```
sschuberth, line 204: Hmm, the following code looks the laborious. I wonder whether we could make some smart re-use of Kotest matcher code here?

sschuberth (changes requested):

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
bs-ondem: I would drop the serializers, because making it invariant on deserialization works for both methods.

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
bs-ondem: `readTree` works only for `JsonNode` ([docs](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.9.9/com/fasterxml/jackson/core/ObjectCodec.html#readTree-com.fasterxml.jackson.core.JsonParser-))

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
bs-ondem: `readTree` works only for `JsonNode` ([docs](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.9.9/com/fasterxml/jackson/core/ObjectCodec.html#readTree-com.fasterxml.jackson.core.JsonParser-))

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
sschuberth: Right, you'd obviously need to use `readValue` then.

bs-ondem: But I think I have to use `readTree` at this point, otherwise the parser would point to the same place after it was deserialized and this causes an exception.

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
bs-ondem: I also thought about using this, but I guess the output of those Kotest matcher would be not really readable for large files/big diffs...

plugins/commands/compare/src/main/kotlin/CompareCommand.kt
sschuberth: Ah, yes, you'd probably run into a recursive problem here.

But how does `readTree` know what the structure of an `Instant` is, and how many properties to read / skip? I guess this just happens to work because `Instant` is serialized as a string which happens to a single primitive `JsonNode` / `TextNode`.

Anyway, I believe you've verified this to work.

sschuberth: > read the "object" the parser is currently pointing at no matter if it is a simple value or a nested tree

You're right, it will read the whole next node, no matter whether it's a primitive text node or a complex object node.

PR #7948, opened 2023-11-28
Node package manager related cleanups

sschuberth:


plugins/package-managers/node/src/test/kotlin/utils/NpmSupportTest.kt
```diff
@@ -72,4 +78,142 @@ class NpmSupportTest : WordSpec({
...
+
+        "correct Artifactory API URLS" {
+            fixNpmDownloadUrl("http://my.repo/artifactory/api/npm/npm/all") shouldBe
+                "http://my.repo/artifactory/npm/all"
+        }
+    }
+
+    "parseNpmAuthors" should {
+        "get authors from a text node" {
+            val node = ObjectMapper().run {
```
fviernau, line 97: Have you considered using a multi-line JSON string 
and just parse it, instead of creating the node programmatically?
(To make it more obvious what the input of the test actually is)

sschuberth: > Have you considered using a multi-line JSON string
> and just parse it, instead of creating the node programmatically?

Yes, I have, and that was the approach I wanted to use at first, but then I thought it would be nice to not involve any parsing (and possible unrelated errors), but really test on the `JsonNode` type that is being expected as a parameter.

However, I do not really have a strong preference here. I can change it if you want.

fviernau: Let's just keep it then. Sounds reasonable to me...

fviernau (approved):

PR #7950, opened 2023-11-28
test(compare-command): Add a functional test for the text diff method

sschuberth:


plugins/commands/compare/src/funTest/kotlin/CompareCommandFunTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import com.github.ajalt.clikt.testing.test
+
+import io.kotest.assertions.withClue
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.shouldBe
+
+class CompareCommandFunTest : WordSpec({
+    "The text diff method" should {
+        "ignore time and environment differences in an analyzer result" {
```
fviernau, line 30: Looking at what this tests covers vs. the test assets, I believe the assets contain a lot of unneeded stuff.

I believe it'd be better if the test only had one result hard-coded, then it should make a copy with environment modified and then assert the diff is empty. 

Doing so could re-use an ORT file coded in Kotlin we already have, maybe.

fviernau (changes requested):

plugins/commands/compare/src/funTest/kotlin/CompareCommandFunTest.kt
sschuberth: Hmm, not sure about this. I deliberately created real-life analyzer results, once from a Jenkins run on Linux, once from a manual run on Windows, to ensure the command works for these, instead of coming up with some artificial results that I'm patching up, with having in mind how the implementations works.

sschuberth: So, would you be ok with sticking to the current approach @fviernau?

fviernau: Not really convinced, but the maybe unnecessary maintainance cost I tried to avoid are isolated and limited. So, ok let's make this compromise.

fviernau (approved):

PR #7952, opened 2023-11-28
Updating build toolchain and deps of Web App reporter

tsteenbe:
This pull request contains a significant changes to how ORT's WebApp (template) is generated and upgrade all dependencies to the latest available versions.

After @mennaElnemr9  created https://github.com/oss-review-toolkit/ort/pull/6598 and https://github.com/oss-review-toolkit/ort/pull/6552 I reviewed the impact of updating dependencies which let me to find issue after issues. Several build toolchain and app dependencies have introduced breaking changes or are not longer maintained. After a lots of research I came to the conclusion major rewrite of the WebApp was needed. Especially use create-react-app (CRA) and rescripts was problematic since rescripts has been archived and updated CRA ORT WebApp builds showed me several deprecated dependencies errors and warnings.

My current plan (subject to change):
Step 1: Change build toolchain and update deps
Step 2: Refactor React component code to use React v18 ways of working e.g. replace component classes with functional components. Possible look into replacing dependency on react-syntax-highlighter
Step 3: Implement new features / bug fixes


heliocastro (dismissed):
Looks good to me
Maybe on a second pass you can extend eslint verifications, but for this upgrade no need now.

sschuberth:
@tsteenbe can we please move forward with this? Or should some one take over?

plugins/reporters/web-app-template/build.gradle.kts
```diff
@@ -28,8 +28,8 @@ import org.jetbrains.kotlin.gradle.targets.js.yarn.YarnSetupTask
 
 // The Kotlin/JS plugins are only applied programmatically for Kotlin projects that target JavaScript. As we do not
 // directly target JavaScript from Kotlin, manually apply the plugins and configure the tool versions.
-NodeJsRootPlugin.apply(rootProject).nodeVersion = "16.13.0"
-YarnPlugin.apply(rootProject).version = "1.22.17"
+NodeJsRootPlugin.apply(rootProject).nodeVersion = "20.9.0"
+YarnPlugin.apply(rootProject).version = "1.22.19"
```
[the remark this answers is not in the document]
tsteenbe: @sschuberth  Yes, it would be nice if we could read them out programmatically - is this something you can implement in a follow up PR?

.ort.yml
```diff
@@ -148,7 +148,7 @@ curations:
     reason: "DOCUMENTATION_OF"
     comment: >-
       Documentation contains examples mentioning various licenses.
-  - path: "plugins/reporters/web-app-template/public/index.html"
+  - path: "plugins/reporters/web-app-template/index.html"
```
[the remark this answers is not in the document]
tsteenbe: > Why are all .js now .jsx files? And what are .jsx files after all?

*.js is plain javascript file whilst *.jsx are JSX files. JSX is a syntax extension for JavaScript coming from React project that lets HTML-like markup inside a JavaScript file. Not all *.js are *.jsx files now only those of the UI which have HTML within JavaScript are now JSX file. For Vite to work it requires correct extensions.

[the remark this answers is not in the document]
tsteenbe: > Why was logo.svg deleted?

Logo.svg actually contained the React logo (see https://github.com/oss-review-toolkit/ort/blob/c7d5c3a4/plugins/reporters/web-app-template/src/logo.svg), it was never used within the WebApp but was there to prevent create-react-app from throwing an error during build that logo.svg was missing. As we now migrate off create-react-app to vite logo.svg can savely be deleted.

plugins/reporters/web-app-template/build.gradle.kts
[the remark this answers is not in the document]
heliocastro: I can implement that part after the PR is merged.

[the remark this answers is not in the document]
sschuberth: Thinking about it again, I'm not sure whether it's meaningful to align these. We should probably not mix build- and runtime-requirements of ORT. For build-time, we probably always want the latest, whereas for runtime we might want to stick to older versions for backwards-compatibility. So it actually makes sense to decouple these, IMO.

sschuberth:
@etsija and @mmurto, as the front-end guys, would you have some time for sharing your thoughts on these changes as well?

plugins/reporters/web-app-template/vite.config.js
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
```
sschuberth, line 2: I guess this should be 2024 by now.

sschuberth (changes requested):
> fix(web-app-template): Minor index.html fixes
>
> Improve HTML head meta tags and fix up indentation.

I can't comment on the above commit as the diff is too large, but please split pure indentation / non-functional changes from the meta tags edits, so these can be reviewed.

sschuberth:
Looks like you need to adapt the expected size of the report again:

> 1588022 should be in range 1556000..1586000

tsteenbe:
> Looks like you need to adapt the expected size of the report again:
> 
> > 1588022 should be in range 1556000..1586000

I could not see this - when I look at the logs on GitHub it would hand somewhere in the tests on a pass without giving me any details of what went wrong. Thanks for the pointer.

tsteenbe:
@sschuberth  All feedback implemented and the tests now are green after https://github.com/oss-review-toolkit/ort/pull/8521 was merged.

plugins/reporters/web-app-template/src/components/TableView.jsx
```diff
@@ -501,7 +519,7 @@ class TableView extends React.Component {
                         {
                             defaultPageSize: 100,
                             hideOnSinglePage: true,
-                            pageSizeOptions: ['50', '100', '250', '500', '1000', '5000'],
+                            pageSizeOptions: ['50', '100', '250', '500'],
```
sschuberth: Please keep `, '1000', '5000'`; those were deliberately added in https://github.com/oss-review-toolkit/ort/pull/7144.

plugins/reporters/web-app-template/eslint.config.js
```diff
@@ -0,0 +1,470 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
```
sschuberth, line 2: This should now also say 2024.

plugins/reporters/web-app-template/src/sagas/index.js
```diff
@@ -30,15 +30,14 @@ export function * loadOrtResultData() {
     const ortResultDataNode = document.querySelector('script[id="ort-report-data"]');
 
     if (ortResultDataNode) {
-        const { type: dataType } = ortResultDataNode;
+        const {textContent: ortResultDataNodeContents, type: ortResultDataNodeType} = ortResultDataNode;
```
sschuberth, line 33: Is it style-wise really correct to no more have spaces after / before curly braces?

plugins/reporters/web-app/src/funTest/kotlin/WebAppReporterFunTest.kt
```diff
@@ -36,7 +36,7 @@ class WebAppReporterFunTest : WordSpec({
 
             val report = WebAppReporter().generateReport(ReporterInput(ortResult), outputDir).single()
 
-            report.length() should beInRange(1556000L..1586000L)
+            report.length() should beInRange(1559000L..1589000L)
```
sschuberth, line 39: This should be squashed to the commit that updated the dependencies.

sschuberth (changes requested):

heliocastro:
Tested this branch's changes using against ORT scans I have done - no issues found, looks good to me

heliocastro (approved):
Tested with full evaluator file from SW360 scan. Works properly.

plugins/reporters/web-app-template/src/components/TableView.jsx
tsteenbe: Seems did got lost in whilst I was resolving a merge conflict in a rebase - fix it now.

plugins/reporters/web-app-template/src/sagas/index.js
tsteenbe: No it isn't per ORT's eslint lint rules but wasn't caught as upgrade to ESLint causes some eslint plugins not to function properly and fail silently - downgraded to eslint 8.57 and now the above issue it get reported and fixed it by running `yarn lint --fix`.

tsteenbe:
Thank you @heliocastro for confirming this PR works as expect, @sschuberth  implemented all your comments hopefully it now good to go.

heliocastro (dismissed):
Good to Go, even with recent changes

fviernau:
I've tested it as well by using two evaluation-result.json files and comparing old vs. new report.
I've used Firefox version 123 under Ubuntu.
I've found a couple of nice formatting improvements.

Besides these, I was not sure about these formatting changes :

1. Margin between tab icon and text:
- old: Screenshot from 2024-04-16 09-19-13
- new: Screenshot from 2024-04-16 09-19-24
2. Same here:
- old: 
- new: 
3. Key of the key value pairs now gray:
- old: Screenshot from 2024-04-16 09-23-22
- new: Screenshot from 2024-04-16 09-23-27

Also I've found a bug with the filtering of findings in scan results. This can be reproduced in
multiple places where the scan result details are shown. E.g. in the tree tab, in the issues tab and so forth:

- old: Screenshot from 2024-04-16 09-24-55
- new: Screenshot from 2024-04-16 09-25-13

My suggestion would be to at minimum fix the bug, plus the changed margins. I'm not sure about the graying out of the key.

sschuberth:
> I've tested it as well by using two evaluation-result.json files and comparing old vs. new report.

Thanks for the effort you've put into testing this!

tsteenbe:
@fviernau  I debugged the filtering issue and it bug in upstream Ant Design (wrong input data is returned on setting of filters). Downgrade the version of Ant Design to a version without the bug that I had tested the WebApp report with previously and will fill a bug report upstream.

sschuberth:
> I debugged the filtering issue and it bug in upstream Ant Design

Could you please link the upstream issue (or create one if none exists yet) so we can upvote / subscribe to it?

tsteenbe:
> Could you please link the upstream issue (or create one if none exists yet) so we can upvote / subscribe to it?

Will do. Did already a quick check of the Ant Design issues (which is easy as many reports are in Mandarin) issue doesn't seem to be reported. Which does not surprise me as I am struggling to reproduce the bug outside of ORT -  bug with a small reproducable example usually get fixed fast by Ant Design maintainers.

fviernau:
I've tested again (similar as above) with latest updated (ba81082).
My previous findings are now all fixed.

Approving as it seems to work nicely, computes a bit faster and reports are smaller (without understanding every single JS detail).

fviernau (approved):

heliocastro (approved):
Results are same as previous iteraction, so good to go.

PR #7963, opened 2023-11-30
feat(model): Add the property `Issue.affectedPath`

fviernau:
This property can be set in case the issue is limited to a file or directory. This prepares for making use of it for ScanCode timeout issues, e.g. filter out irrelevant timeout issues if outside the VCS path or matched by a path exclude.

### DIFF
--- a/model/src/main/kotlin/Issue.kt
+++ b/model/src/main/kotlin/Issue.kt
@@ -19,6 +19,7 @@
 
 package org.ossreviewtoolkit.model
 
+import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.core.JsonGenerator
 import com.fasterxml.jackson.databind.SerializerProvider
 import com.fasterxml.jackson.databind.annotation.JsonSerialize
@@ -53,7 +54,13 @@ data class Issue(
     /**
      * The issue's severity.
      */
-    val severity: Severity = Severity.ERROR
+    val severity: Severity = Severity.ERROR,
+
+    /**
+     * The affected file or directory the issue is limited to, if any.
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    val affectedPath: String? = null
 ) {
     override fun toString(): String {
         val time = if (timestamp == Instant.EPOCH) "Unknown time" else timestamp.toString()


model/src/main/kotlin/Issue.kt
```diff
@@ -53,7 +54,13 @@ data class Issue(
     /**
      * The issue's severity.
      */
-    val severity: Severity = Severity.ERROR
+    val severity: Severity = Severity.ERROR,
+
+    /**
+     * The affected file or directory the issue is limited to, if any.
```
mnonnenmacher, line 60: Maybe "the issue is related to" or "that causes the issue"?

model/src/main/kotlin/Issue.kt
```diff
@@ -53,7 +54,13 @@ data class Issue(
...
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    val affectedPath: String? = null
```
mnonnenmacher, line 63: I'm not so happy with the name. `affectedPath` sounds like the file at this location is affected by the issue. For example, if a file system is corrupt, then files can be affected by this. But with the example of a ScanCode timeout, the timeout does not affect the file at all, the file is merely the cause of the issue because ScanCode cannot handle it for some reason.
Since we do not know what kinds of issues might make use of this property in future, shall we just call it `filePath` or `path`?

mnonnenmacher (changes requested):

model/src/main/kotlin/Issue.kt
sschuberth: > But with the example of a ScanCode timeout, the timeout does not affect the file at all

Well, I'm reading it differently, like "What thing was affected by the timeout? The file, because it's contents were not properly scanned."

> Since we do not know what kinds of issues might make use of this property in future, shall we just call it `filePath` or `path`?

As mentioned on Slack I was against `filePath` as also directories could be affected, and `path` sounds overly generic, thought I could live with it. Maybe `relatedPath`?

fviernau: I remove myself / my preference from this discussion and will implement what you guys agree @mnonnenmacher  @sschuberth.

mnonnenmacher: If both of you are happy with the current name please just go ahead.

sschuberth (approved):

Issue #7921, opened 2023-11-23
Scanner: Avoid returning irrelevant issues


fviernau:
The scanner might run into an issue which related only to a single file.
If an issue related to a file which not within the set of files resulting from filtering the scan result by VCS path, then
the issue should also be filtered out. This happens e.g. for TIMEOUT issues.

In this context, it makes sense to consider enabling the path excludes to silence timeouts as well.

fviernau:
Thinking about solving this by extending `ScanSummary` to have a dedicated property for `issues` which are specific to a single file. Maybe just a map `filePath -> issue` ? What do you think?

mnonnenmacher:
Do we get the affected file for such issues from the ScanCode result model or would we have to parse the error messages to try to find out if an issue belongs to a single file?

fviernau:
Finally, the following components adhere to the affected path of the issues WRT to path excludes and VCS:
- model such as `ScanSummary` / `OrtResult`
- `EvaluatedModelReporter`
- `StaticHtmlReporter`
- `WebAppReporter`

Therefore, closing the issue as completed. 

Note that the exist code / issues stats CLI output of the `ScannerCommand` does not yet adhere to excluded paths. 
Implementing that does require to also make the scanner command update the resolved config to import package
configurations and issue resolutions. This can be done as a (larger) follow-up task, which has its own ticket.

sschuberth:
> This can be done as a (larger) follow-up task, which has its own ticket.

Mind creating such an issue (if it does not exist yet) so that we do not forget about it?

fviernau:
> Mind creating such an issue (if it does not exist yet) so that we do not forget about it?

I've updated my above comment.

PR #7972, opened 2023-12-01
ScanCode: Improve mapping timeout errors

fviernau:
Relax the precondition for mapping timeout errors, so that the mapping is applied
regardless of the actual timeout value.

Also cover the mapping by the tests, which required to re-create the test assets for the output formats. 

Note: This prepares for setting `Issue.affectedPath` inside of `mapTimeoutIssues()`.

Part of #7921.


plugins/scanners/scancode/src/test/assets/scancode-output-format-2.0.0_mime-types-2.1.18.json
```diff
@@ -13,25 +13,27 @@
         "--license": true,
         "--processes": "3",
         "--strip-root": true,
-        "--timeout": "300.0"
+        "--timeout": "0.09"
```
sschuberth, line 16: Any reason why you using 0.09 here but 0.07 above? Similar for the next file.

plugins/scanners/scancode/src/test/kotlin/ScanCodeResultParserTest.kt
```diff
@@ -179,3 +178,5 @@ private fun containCopyrightsExactly(vararg copyrights: Pair<String, List<TextLo
         summary.copyrightFindings.groupBy { it.statement }.entries
             .map { (key, value) -> key to value.map { it.location } }
     }
+
+private fun getAssetFile(path: String): File = File("src/test/assets", path).absoluteFile
```
sschuberth, line 179: Please just import the existing function from `test-utils`. Also plugin projects have access to it.

sschuberth (changes requested):

plugins/scanners/scancode/src/test/assets/scancode-output-format-2.0.0_mime-types-2.1.18.json
fviernau: I played with the values until I got exactly one issue.

plugins/scanners/scancode/src/test/kotlin/ScanCodeResultParserTest.kt
fviernau: The existing functions only works for `funTest`.

sschuberth: It shouldn't, as the `test-utils` dependency is added for *all* test suite, not just funTests:

https://github.com/oss-review-toolkit/ort/blob/521640b4f55edfbfb357f65a12db480eb9521474/buildSrc/src/main/kotlin/ort-kotlin-conventions.gradle.kts#L62

I actually tried it quickly, and it worked for me. Just the import needs to be added.

fviernau: The implementation looks like this:

```
fun getAssetFile(path: String): File = File("src/funTest/assets", path).absoluteFile
```

but the asset is under `src/test/assets`, not `src/FunTest/assets`.

sschuberth: Ah, sorry, I misread your comment.

sschuberth (approved):

PR #7980, opened 2023-12-01
Make PackageManagers Closeable

### LINKED ISSUE #7978: DiskCache used for maven project gets corrupted sometimes
Running ort multiple times in a row for a project occasionally results in a corrupted cache:

```
DiskLruCache /home/tn/.ort/cache/analyzer/maven/remote-artifacts is corrupt: unexpected journal line: READ oREAD io-quarkus_quarkus-resteasy_jar_3-5-2__repo-eclipse-org__h-00000, removing
```

leading to the cache being discarded and all artifacts are fetched again.

This looks very much like this problem: https://github.com/JakeWharton/DiskLruCache/issues/52

Once I added proper closing of the cache, the problem was going away, will prepare a PR.

netomi:
This fixes #7978 .

This PR makes PackageManagers implement the Closeable interface to be able to close their resources once they are not used anymore (at the end of the analyzer run).

There is a default implementation that does nothing, the maven / gradle package manager override this method in order to close the DiskCache they are using via the MavenSupport class.


fviernau:
Thanks @netomi for providing this fix. 
After reading up on the comments of the GitHub issue, it seems that non-atomic reads to the journal
a suspected to be the root cause, see [1]. Assuming this is the case, I do not understand how / why this PR is supposed to fix the issue. 

@netomi do you have any explanation why this fix does work?
Can you share the reproducibility rate you achived before making this fix? e.g. difficult it was to reproduce, and also how you verified the fix works?

[1] https://github.com/JakeWharton/DiskLruCache/issues/52#issuecomment-1500755265

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -361,6 +362,11 @@
             }
         }
     }
+
+    /**
+     * [PackageManager] implementations should override this method if they need to clean resources.
+     */
+    override fun close() {}
```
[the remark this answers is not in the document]
netomi: ok, so is there a way to tell detekt that this is reasonable as its a default implementation that does nothing?

plugins/package-managers/gradle/src/main/kotlin/Gradle.kt
```diff
@@ -290,4 +290,8 @@ class Gradle(
             }
         }
     }
+
+    override fun close() {
+        maven.close()
```
netomi, line 295: is there a clever way in kotlin to do a silent close, i.e. ignore any exceptions that are thrown in the process?

sschuberth: See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html.

netomi: applied suggestion

analyzer/src/main/kotlin/PackageManager.kt
```diff
@@ -19,6 +19,7 @@
 
 package org.ossreviewtoolkit.analyzer
 
+import java.io.Closeable
```
sschuberth, line 22: Commit message: I believe saying "properly close PackageManager instances" is misleading, as before this change package managers were not closeable, so you also couldn't close them.

plugins/commands/analyzer/src/main/kotlin/AnalyzerCommand.kt
```diff
@@ -228,5 +228,9 @@ class AnalyzerCommand : OrtCommand(
         val issues = analyzerRun.result.getAllIssues().flatMap { it.value }
         SeverityStatsPrinter(terminal, resolutionProvider).stats(issues)
             .print().conclude(ortConfig.severeIssueThreshold, 2)
+
+        for (packageManager in info.managedFiles.keys) {
+            packageManager.close()
```
sschuberth, line 233: What I don't like so much about this approach is that every downstream user of `Analyzer.analyze()` has to call `close()` itself. I wonder whether a feasible / better approach would be for the respective package manager to override the existing `afterResolution()` and close their resources there.

plugins/package-managers/sbt/src/funTest/assets/projects/target-scala-2.12-projects_2.12-0.1.0-SNAPSHOT.pom
```diff
@@ -0,0 +1,20 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
sschuberth, line 1: The files committed here and below seem like temporary left-overs.

sschuberth (changes requested):

analyzer/src/main/kotlin/PackageManager.kt
netomi: ok, changed.

plugins/package-managers/sbt/src/funTest/assets/projects/target-scala-2.12-projects_2.12-0.1.0-SNAPSHOT.pom
netomi: indeed, using git add -A is not really a good idea.

plugins/commands/analyzer/src/main/kotlin/AnalyzerCommand.kt
netomi: I understand your point, the downside is that the Analyzer is fed with a a list of managedFiles with their associated PackageManager. Thus when the Analyzer would close the PackageManagers after it has finished analysis, the caller would not be able to use the package managers anymore in a safe way as they are already closed. This transfer of ownership would have to be documented imho, but I am indifferent where the closing should happen, unless it happens ofc :-)

sschuberth:
Closing this in favor of https://github.com/oss-review-toolkit/ort/pull/9781 which changes the `PackageManager` API to implement the changes in a better way. Thanks @netomi for the original implementation proposal!

PR #7981, opened 2023-12-01
feat: improve resolving of maven based dependencies

### LINKED ISSUE #7965: Improve analysis of maven based project
I tried to use the ORT GH action to analyse a maven project, but it took more than 30 min to run only the analysis (https://github.com/netomi/macos-notarization-service/actions/runs/7006736431).

The project has some additional repository setup and ORT seems to try find out from which repository a dependency is coming from by trying to download the dependency from each configured repository. Unfortunately mavencentral is coming last, so each dependency is first tried from the other. There seems to be additionally some network throttling in place when run as a GH action, when running the same analysis locally, it completed in a couple of minutes.

However, there should be a way to speed this up and I worked on improving the resolution of dependencies in maven projects. 

In my fork at https://github.com/netomi/ort/tree/disable-remote-verification I did some experiments to get the resolved repository from maven itself (it stores that information in the _remote.repositories file in the local cache).

With these changes the run of the analysis on GitHub could be completed in around 2min, see https://github.com/netomi/macos-notarization-service/actions/runs/7021936345).

My approach so far is quick and dirty, but I will be working on a PR to make this as clean as possible.

netomi:
This fixes #7965 .

Opening a draft PR to dicuss the speed improvements.

The idea of this change is as follow:

- let the pom file being resolve initially so that it is present in the local maven repository
- resolve the pom file of the artifact using resolveArtifact which will take advantage of the local repo and gives information from which repo it was resolved
- use the resolve repository to retrieve the checksum files for the binary / source artifacts

With this changes it works pretty fast. There are some assumptions that might not be ok for the general case:

- it is assumed that sources are located in the same repository as the binary artifact itself, imho this is a safe assumption to make
- when a checksum file is present it is assumed that the artifact is also present (otherwise the default resolution mechanism of maven would be broken imho)
- right now, if no checksum is found for an artifact, an empty hash is assumed, but the artifact is still returned as it exists, we could add a safe-guard to do a PeekTask in case no checksum is found to verify that the artifact exists

### DIFF
--- a/.github/workflows/docker-ort.yml
+++ b/.github/workflows/docker-ort.yml
@@ -19,8 +19,8 @@ name: ORT Docker Image
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: '0 4 * * *'
+#  schedule:
+#    - cron: '0 4 * * *'
   pull_request:
     paths:
       - '.versions'
--- a/plugins/package-managers/maven/src/main/kotlin/utils/MavenSupport.kt
+++ b/plugins/package-managers/maven/src/main/kotlin/utils/MavenSupport.kt
@@ -62,6 +62,7 @@ import org.eclipse.aether.repository.MirrorSelector
 import org.eclipse.aether.repository.RemoteRepository
 import org.eclipse.aether.repository.WorkspaceReader
 import org.eclipse.aether.resolution.ArtifactDescriptorRequest
+import org.eclipse.aether.resolution.ArtifactRequest
 import org.eclipse.aether.spi.connector.ArtifactDownload
 import org.eclipse.aether.spi.connector.layout.RepositoryLayout
 import org.eclipse.aether.spi.connector.layout.RepositoryLayoutProvider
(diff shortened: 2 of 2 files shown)


netomi:
gradle tests are failing, need to look into that, did mainly test with maven projects so far.

sschuberth:
> will update the PR accordingly.

Do you still plan to continue working on this @netomi?

Otherwise I could look into taking over your PR. But in that case I'd be grateful for any hints on how you were planning to resolve the Gradle issues.

netomi:
well I never got any feedback whether this approach would be interesting for the project.

Atm I do not work with ort, so we can close that for now I guess.

sschuberth:
> well I never got any feedback whether this approach would be interesting for the project.

I'd say that any performance improvement that does not break tests and that does not make the code extremely more complex is welcome 😸 

> Atm I do not work with ort, so we can close that for now I guess.

Then let me try to take a look at your code and eventually take this PR over.

sschuberth:
I've rebased the PR onto current `main` to continue working on it. However:

> Opening a draft PR to dicuss the speed improvements.
> ...
> With this changes it works pretty fast.

I'm not actually seeing any speed improvements with this PR. After removing my local `~/.m2` directory, running `./gradlew -p plugins/package-managers/maven/ funTest --rerun` is just as fast as on current `main` for me.

@netomi, in which scenario did you actually see speed improvements?

Anyway, as the code is nicer, I'll further look into the Gradle test failures to see if we can still merge this.

netomi:
so I was doing several experiments a long time ago. Maybe maven resolution did improve in the meantime. In my test scenarios, there was an additional repository configured, and ort was checking all configured repositories if any artifact is available from it, even though most are available from maven central ofc. This obviously resulted in a slow analysis process, also depending on the speed of the additionally configured repositories. In my tests / PR I wanted to take advantage of the builtin maven cache where it stores from which repository an artifact was resolved and use that information rather than checking all configured repositories again.

So obviously when doing multiple analysis runs for multiple projects sharing the maven cache this might lead to not-exact results, but in my case every analysis would be run on a clean or separate maven cache so that would not be a concern.

sschuberth:
> Maybe maven resolution did improve in the meantime.

I don't think so. The code hasn't been touched in a while, in way that would impact performance.

> In my test scenarios, there was an additional repository configured, and ort was checking all configured repositories if any artifact is available from it, even though most are available from maven central ofc.

Ok, so that could make the difference.

> So obviously when doing multiple analysis runs for multiple projects sharing the maven cache this might lead to not-exact results

Oh, wait... so you're saying running `ort analyze` on the same machine (and a single "shared" `~/.m2` directory) but with different projects could lead to wrong results for either project?

netomi:
this PR relies on maven to resolve an artifact to utilize its internal cache. You might poison your cache if you analyze projects that configure different repositories and resolve artifacts that are shared by multiple projects.

So thats why I wanted to discuss this PR as this will obviously not be a general solution for ORT, but for use-cases where you can ensure that you have a separate or clean cache for each project it will (hopefully) speed up the artifact resolution.

sschuberth:
> You might poison your cache if you analyze projects that configure different repositories and resolve artifacts that are shared by multiple projects.

Ah. I keep forgetting that Maven, in contrast to Gradle, does not store in its cache / local repository from which server a particular artifact was retrieved.

> So thats why I wanted to discuss this PR as this will obviously not be a general solution for ORT, but for use-cases where you can ensure that you have a separate or clean cache for each project it will (hopefully) speed up the artifact resolution.

Thanks for making that clear again. I somehow missed that exact point, or forgot about this.

So that means to me that we cannot merge this PR as-is, and also keeping the code as an optional special case when `~/.m2` does not seem feasible to me currently. Or what do @oss-review-toolkit/kotlin-devs think?

sschuberth:
> So that means to me that we cannot merge this PR as-is, and also keeping the code as an optional special case when `~/.m2` does not seem feasible to me currently. Or what do @oss-review-toolkit/kotlin-devs think?

Giving this more though, not being able to distinguish artifacts with the same GA-coordinates from different repositories actually is a problem that is inherent to Maven, and I believe it's thus also present with the current implementation.

PR #7987, opened 2023-12-04
fix(command): Drop an obsolete scanner command option

fviernau:
As of ef84184 the support for scanning a file or directory has been dropped. Remove the corresponding CLI option to fix-up ef84184.


plugins/commands/scanner/src/main/kotlin/ScannerCommand.kt
```diff
@@ -73,22 +69,15 @@ class ScannerCommand : OrtCommand(
     name = "scan",
     help = "Run external license / copyright scanners."
 ) {
-    private val input by mutuallyExclusiveOptions(
+    private val input by
```
sschuberth, line 72: Formatting / indentation should now also be changed, see e.g. line 82 and following.

sschuberth (changes requested):

sschuberth (approved):

PR #7989, opened 2023-12-04
scanner: Adhere to `Issue.affectedPath` when filtering scan issues

fviernau:
Eliminate some irrelevant issues from the report, to reduce the need for manually creating irrelevant resolutions. 

Note: The stored scan results do not have the `affectedPath` set. However, the deserialization of scan results has been          
           adjusted so that `affectedPath` is properly set during deserialization and there the filtering works even for old 
           stored scan results.

Part of #7921.

### DIFF
--- a/model/src/main/kotlin/ScanSummary.kt
+++ b/model/src/main/kotlin/ScanSummary.kt
@@ -23,7 +23,9 @@ import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
 import com.fasterxml.jackson.databind.annotation.JsonSerialize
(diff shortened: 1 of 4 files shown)


model/src/main/kotlin/ScanSummary.kt
```diff
@@ -144,3 +147,18 @@ data class ScanSummary(
...
+/**
+ * Set the `affectedPath` for scan timeout errors. So, scan results which have been created before `affectedPath` got
+ * introduced, still have the property value set.
+ */
+internal class IssueListConverter : StdConverter<List<Issue>, List<Issue>>() {
+    override fun convert(value: List<Issue>): List<Issue> =
+        value.map { issue ->
+            if (issue.affectedPath != null) return@map issue
+            val match = TIMEOUT_ERROR_REGEX.matchEntire(issue.message) ?: return@map issue
+            return@map issue.copy(affectedPath = match.groups["file"]!!.value)
```
sschuberth, line 160: Drop `return@map `.

sschuberth (changes requested):

sschuberth (approved):

PR #7996, opened 2023-12-05
refactor(config)!: Align on setting "skip excluded" in the config

sschuberth:
Previously, only the analyzer's "skip excluded" feature could be set via ORT's `config.yml` file. The other tools' "skip excluded" feature was only available via the `--skip-excluded` CLI option.

Align this by moving all "skip excluded" feature flags to the respective configuration, which still allows to set them via CLI options by using the generic syntax to set `config.yml` properties, e.g.

    ort -P ort.downloader.skipExcluded=true download ...


mnonnenmacher (approved):

sschuberth:
Merging despite the unrelated failure in the `markdown-links` check.

PR #8008, opened 2023-12-07
docs(cli): Distribute a `README.md` to show where to put plugins

sschuberth:


model/src/main/kotlin/ScanSummary.kt
```diff
@@ -149,10 +149,9 @@ data class ScanSummary(
 }
 
 /**
- * Set the `affectedPath` for scan timeout errors if otherwise it would have been unset. This way scan results which
- * have been created before `affectedPath` was introduced will still have that property set.
+ * Set the `affectedPath` for scan timeout errors if otherwise it had been unset. This way scan results which have been
```
mnonnenmacher, line 152: I find the sentence still clunky, what about: "Set the `affectedPath` for scan timeout errors if it is `null`."
And maybe also add "the" before "`affectedPath`".

sschuberth: > And maybe also add "the" before "`affectedPath`".

I guess you mean "remove" instead of "add"?

mnonnenmacher: No, I meant to add it in the second sentence.

mnonnenmacher (approved):

PR #8016, opened 2023-12-08
feat(scancode): Get the key to ID mapping without `--license-references`

sschuberth:
Usually, starting with ScanCode 32, explicitly passing the `--license-references` option is required in order to have information in the result file about which ScanCode key corresponds to which SPDX ID.

However, this information can also be reconstructed by correlating the strings from the `detected_license_expression(_spdx)` fields as they contain corresponding entries in order. Use that information as a fallback if `--license-references` was not used.


oheger-bosch (commented):
Changes look good to me, but since they are partly not trivial, shouldn't there be additional tests? I don't know the test coverage for the existing code, but the logic for `Version3` seems to be new and thus should be tested, shouldn't it?

sschuberth:
> shouldn't there be additional tests?

I've added a test for this new feature.

plugins/scanners/scancode/src/test/assets/scancode-32.0.8_spdx-expression-parse_no-license-references.json
```diff
@@ -0,0 +1,823 @@
...
+      "copyrights": [],
+      "holders": [],
+      "authors": [],
+      "files_count": 0,
+      "dirs_count": 0,
+      "size_count": 0,
+      "scan_errors": []
+    }
+  ]
+}
```
oheger-bosch, line 823: Nit: Missing newline at EOF.

oheger-bosch (dismissed):

plugins/scanners/scancode/src/test/assets/scancode-32.0.8_spdx-expression-parse_no-license-references.json
sschuberth: True. I took the file as-is from ScanCode, and it's inconsistent for our JSON assets anyway, so I'll align this another way 😸

sschuberth: As there was another fixup I had to do, I've actually decided to align this now, too.

oheger-bosch (approved):

PR #8018, opened 2023-12-08
feat(model): Associate licenses and exceptions from the same expression

sschuberth:

### DIFF
--- a/model/src/main/kotlin/utils/FindingsMatcher.kt
+++ b/model/src/main/kotlin/utils/FindingsMatcher.kt
@@ -27,9 +27,15 @@ import kotlin.math.min
 import org.ossreviewtoolkit.model.CopyrightFinding
 import org.ossreviewtoolkit.model.LicenseFinding
 import org.ossreviewtoolkit.model.TextLocation
+import org.ossreviewtoolkit.utils.spdx.SpdxCompoundExpression
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
+import org.ossreviewtoolkit.utils.spdx.SpdxExpression
 import org.ossreviewtoolkit.utils.spdx.SpdxLicenseException
+import org.ossreviewtoolkit.utils.spdx.SpdxLicenseIdExpression
 import org.ossreviewtoolkit.utils.spdx.SpdxLicenseWithExceptionExpression
+import org.ossreviewtoolkit.utils.spdx.SpdxOperator
+import org.ossreviewtoolkit.utils.spdx.SpdxSimpleExpression
+import org.ossreviewtoolkit.utils.spdx.SpdxSingleLicenseExpression
 import org.ossreviewtoolkit.utils.spdx.toSpdx
 
 /**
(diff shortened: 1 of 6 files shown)


model/src/main/kotlin/utils/FindingsMatcher.kt
```diff
@@ -27,9 +27,15 @@ import kotlin.math.min
 import org.ossreviewtoolkit.model.CopyrightFinding
```
oheger-bosch, line 27: Commit message: In the first sentence in the body something seems to be missing after "function", a "that"?

model/src/main/kotlin/utils/FindingsMatcher.kt
```diff
@@ -274,5 +280,67 @@ fun associateLicensesWithExceptions(
...
+        // Determine all licenses the exception is applicable to.
+        val applicableLicenses = SpdxLicenseException.mapping[exceptionString].orEmpty().mapTo(mutableSetOf()) {
+            SpdxLicenseIdExpression(it.id)
+        }
+
+        // Associate all remaining licenses that are applicable with the exception.
+        remainingLicenses.intersect(applicableLicenses).forEach {
+            associatedLicenses += SpdxLicenseWithExceptionExpression(it, exceptionString)
+            remainingLicenses -= it
+            i.remove()
```
oheger-bosch, line 333: Don't know much about SPDX processing, but is it correct that the `i.remove()` call is done inside the `forEach` loop? Is it guaranteed that it contains enough elements?

model/src/main/kotlin/utils/FindingsMatcher.kt
```diff
@@ -274,5 +280,67 @@ fun associateLicensesWithExceptions(
         exception.copy(license = "${SpdxConstants.NOASSERTION} ${SpdxExpression.WITH} ${exception.license}".toSpdx())
     }
 
-    return fixedLicenses
+    return fixedLicenses.mapTo(mutableSetOf()) { it.copy(license = associateLicensesWithExceptions(it.license)) }
+}
+
+/**
```
fviernau, line 286: This function to me seems quite complex.
I fear it can happen relatively easily to introduce some bug here.
As the function is underneath the call tree of `toSummary` of `ScanResultModelMapper`,
the returned summary may get stored in a scan  result storage. 

So if it turns out there is a bug in, then the stored scan results become kind of invalid.
I believe this is a bit risky, what do you think?

model/src/main/kotlin/utils/FindingsMatcher.kt
```diff
@@ -274,5 +280,67 @@ fun associateLicensesWithExceptions(
...
+ * Process [license] for stand-alone license exceptions as part of compound expressions and associate them with
+ * applicable licenses. Orphan license exceptions will get associated by [SpdxConstants.NOASSERTION]. Return a new
+ * expression that does not contain stand-alone license exceptions anymore.
+ */
+internal fun associateLicensesWithExceptions(license: SpdxExpression): SpdxExpression {
```
fviernau, line 291: out of scope: Can we move these two "associate*" functions to `ScanResultModelMapper` as these are unrelated to the other code in findings matcher but strongly related to `toSummary()` ?

model/src/main/kotlin/utils/FindingsMatcher.kt
sschuberth: Good finding! I'll add a test for this case.

model/src/main/kotlin/utils/FindingsMatcher.kt
sschuberth: > I fear it can happen relatively easily to introduce some bug here.

I agree that the code is rather complex, and that there could be lingering bugs, like with any code that we write.

> the returned summary may get stored in a scan result storage.

Yeah, we bump into this topic on a regular basis. Storing or not storing the results of post-processing scan results in the scan storage both have its pros and cons. However, as the result of associating exceptions across *different* findings already is stored as well, I don't think it makes a difference here.

> I believe this is a bit risky, what do you think?

Generally, I don't think it's too risky to have such code. To me, the scan storage is mostly a cache for speeding up scans, not a database of definitely correct license information. And as such it's fine to update the storage with corrections if needed. The bigger problem IMO would be with wrong stored scan results going unnoticed. But this is more unlikely to happen, I believe.

model/src/main/kotlin/utils/FindingsMatcher.kt
sschuberth: Actually, these functions are deliberately implemented in a scanner-agnostic way, and I'm using them already in code that is not using the `ScanCodeResultModelMapper`, so I'd prefer to leave them here.

oheger-bosch (approved):

model/src/main/kotlin/utils/FindingsMatcher.kt
fviernau: I just wanted to make sure it is considered by the devs approving this.

PR #8020, opened 2023-12-11
feat(Fossid-webapp): Increase the read timeout for `listMatchedLines`

nnobelis:
#

### DIFF
--- a/clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
+++ b/clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
@@ -254,6 +254,7 @@ interface FossIdRestService {
     suspend fun listSnippets(@Body body: PostRequestBody): PolymorphicResponseBody<Snippet>
 
     @POST("api.php")
+    @Headers("$READ_TIMEOUT_HEADER:${5 * 60 * 1000}")
     suspend fun listMatchedLines(@Body body: PostRequestBody): EntityResponseBody<MatchedLines>
 
     @POST("api.php")


clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -254,6 +254,7 @@ interface FossIdRestService {
     suspend fun listSnippets(@Body body: PostRequestBody): PolymorphicResponseBody<Snippet>
```
oheger-bosch, line 263: Commit message typo: "shis commit" -> "this commit"

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -41,6 +41,8 @@ import okhttp3.Interceptor
 import okhttp3.OkHttpClient
```
oheger-bosch, line 41: Commit message: "When the request interceptor *runs* in a timeout ..."

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
```diff
@@ -209,7 +211,14 @@ interface FossIdRestService {
 
             val readTimeout = newReadTimeout?.toIntOrNull() ?: chain.readTimeoutMillis()
 
-            return chain.withReadTimeout(readTimeout, TimeUnit.MILLISECONDS).proceed(request)
+            return runCatching {
+                chain.withReadTimeout(readTimeout, TimeUnit.MILLISECONDS).proceed(request)
+            }.onFailure {
+                logger.error { "Error when intercepting the call chain: ${it.message}" }
+                val requestBuffer = Buffer()
+                request.body?.writeTo(requestBuffer)
```
oheger-bosch, line 219: I assume, request bodies cannot contain sensitive information, right? So it should be safe to log them. Do all requests have a body?

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
nnobelis: Commit has been removed.

clients/fossid-webapp/src/main/kotlin/FossIdRestService.kt
nnobelis: Argh you are right: FossID request bodies contain the api key :(
I removed this commit as I don't see an elegant workaround.
Maybe the `HttpLoggingInterceptor` of OkHttp: https://stackoverflow.com/a/56892146.

sschuberth (changes requested):

sschuberth (approved):

PR #8025, opened 2023-12-12
FossID reporter improvements

oheger-bosch:
Refer to the individual commits for further information.

### DIFF
--- a/plugins/reporters/fossid/src/main/kotlin/FossIdReporter.kt
+++ b/plugins/reporters/fossid/src/main/kotlin/FossIdReporter.kt
@@ -96,7 +96,7 @@ class FossIdReporter : Reporter {
         return runBlocking(Dispatchers.IO) {
             val scanResults = input.ortResult.getScanResults().values.flatten()
             val scanCodes = scanResults.flatMapTo(mutableSetOf()) {
-                it.additionalData[SCAN_CODE_KEY].orEmpty().split(',')
+                it.additionalData[SCAN_CODE_KEY]?.split(',').orEmpty()
             }
 
             scanCodes.map { scanCode ->
--- a/plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
+++ b/plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
@@ -87,7 +87,7 @@ ${sourceLocation.startLine}-${sourceLocation.endLine}
 [#list snippetFinding.snippets as snippet ]
 [#assign matchType = snippet.additionalData["matchType"]]
 [#assign snippetFilePath = snippet.location.path!""]
-[#if matchType == "PARTIAL" && snippetFilePath?has_content]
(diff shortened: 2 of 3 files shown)


plugins/reporters/fossid/src/main/kotlin/FossIdReporter.kt
```diff
@@ -96,7 +96,7 @@ class FossIdReporter : Reporter {
         return runBlocking(Dispatchers.IO) {
             val scanResults = input.ortResult.getScanResults().values.flatten()
             val scanCodes = scanResults.flatMapTo(mutableSetOf()) {
-                it.additionalData[SCAN_CODE_KEY].orEmpty().split(',')
+                it.additionalData[SCAN_CODE_KEY]?.split(',').orEmpty()
```
sschuberth, line 99: Commit message nit: "*The* `FossIdReporter` ..."

Also in the commit message title the class name could be put in backticks.

sschuberth (approved):

plugins/reporters/fossid/src/main/kotlin/FossIdReporter.kt
sschuberth: > Also in the commit message title the class name could be put in backticks.

The "Also" was referring to use backticks in the text, too, though 😉

sschuberth (approved):

plugins/reporters/fossid/src/main/kotlin/FossIdReporter.kt
oheger-bosch: Backticks are cool. Here you go.

plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
```diff
@@ -87,7 +87,7 @@ ${sourceLocation.startLine}-${sourceLocation.endLine}
 [#list snippetFinding.snippets as snippet ]
 [#assign matchType = snippet.additionalData["matchType"]]
 [#assign snippetFilePath = snippet.location.path!""]
-[#if matchType == "PARTIAL" && snippetFilePath?has_content]
+[#if matchType == "PARTIAL" && snippetFilePath?has_content && snippet.additionalData['matchedLinesSnipped']?has_content]
```
nnobelis, line 90: This behavior that you now need to have matched lines for the snippet report was known for me, unfortunately it was on my TODO list.
My idea was to make the get matched lines option always enabled as is it required anyway to group snippets by source location.
If the performance cost of listing matched lines is problematic, maybe add an option "fetch snippets results": if enabled, the snippets with their matched lines you be listed. Hence, snippets and matched lines are always "enable together".

What do you think ?

oheger-bosch: No objections, although I cannot judge the possible performance penalty. Nevertheless, I think that the change made here is rather independent from your proposed enhancement; it is just an additional safe guard preventing the reporter to crash when there is unexpected input.

PR #8029, opened 2023-12-12
fix(evaluator): Apply repository license choices to the project

sschuberth:
This is a follow-up to e5e0f3f.

### DIFF
--- a/evaluator/src/main/resources/rules/osadl.rules.kts
+++ b/evaluator/src/main/resources/rules/osadl.rules.kts
@@ -37,7 +37,9 @@ val ruleSet = ruleSet(ortResult, licenseInfoResolver) {
             -isExcluded()
         }
 
-        val projectLicenseInfo = licenseInfoResolver.resolveLicenseInfo(project.id).filter(licenseView).filterExcluded()
+        val projectLicenseInfo = licenseInfoResolver.resolveLicenseInfo(project.id).filterExcluded()
+            .applyChoices(ortResult.getRepositoryLicenseChoices(), licenseView)
+
         val outboundLicenses = projectLicenseInfo.licenses.map { it.license }
 
         // Define a rule that is executed for each license of the dependency.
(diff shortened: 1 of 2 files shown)


evaluator/src/main/resources/rules/osadl.rules.kts
```diff
@@ -37,7 +37,9 @@ val ruleSet = ruleSet(ortResult, licenseInfoResolver) {
             -isExcluded()
         }
 
-        val projectLicenseInfo = licenseInfoResolver.resolveLicenseInfo(project.id).filter(licenseView).filterExcluded()
+        val projectLicenseInfo = licenseInfoResolver.resolveLicenseInfo(project.id).filterExcluded()
+            .applyChoices(ortResult.getRepositoryLicenseChoices(), licenseView)
```
sschuberth, line 41: Hmm, I find [our own docs about repository licenses choices](https://oss-review-toolkit.org/ort/docs/configuration/ort-yml#license-choice-for-the-project) actually a bit unclear. We say "License Choice for the Project" but also "he license choice is applied to each package". So are repository license choices also applied to project licenses?

fviernau: I'm also a bit uncertain about it. Looking at the evaluator, the package rule can also be executed for projects, and inside it's licenseRule the choices would get applied.

However, for the license compatibility matrix I guess that if a license choice is offered in the LICENSE file, then
compatibility be checked for all choices? However, if one integrates a code snipped which is under a license choice then it might make more sense to actually apply the choice as one of the options might leads to license issues?

sschuberth: Note that here, the project license is not retrieved via `licenseRule` (which would automatically apply choices) but directly, so I do believe we need to apply choices explicitly.

And as `getRepositoryLicenseChoices()` just return an unfiltered `List<SpdxLicenseChoice>` which itself does not contain any project / package information anymore, all choices indeed should get applied to the project' license. So I believe the proposed code to be correct.

fviernau (approved):

Issue #7997, opened 2023-12-05
GoMod failed to resolve dependencies for path 'go.mod': NoSuchElementException: Key go is missing in the map.


mawl:
Since Updating to https://github.com/oss-review-toolkit/ort/tree/ec6ff750b1a5b50bed9d24d5d0f479ecb0ef98bb

We get `GoMod failed to resolve dependencies for path 'go.mod': NoSuchElementException: Key go is missing in the map.` for some projects during analyze phase.

The  projects are using go 1.21

Issue can be reproduced with:

go.mod:
```
module my-module

go 1.21

require go.mongodb.org/mongo-driver v1.12.0 // indirect
```

go.sum:
```
github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
github.com/golang/snappy v0.0.1/go.mod h1:/XxbfmMg8lxefKM7IXC3fBNl/7bRcc72aCRzEWrmP2Q=
github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
github.com/klauspost/compress v1.13.6/go.mod h1:/3/Vjq9QcHkK5uEr5lBEmyoZ1iFhe47etQ6QUkpK6sk=
github.com/montanaflynn/stats v0.0.0-20171201202039-1bf9dbcd8cbe/go.mod h1:wL8QJuTMNUDYhXwkmfOly8iTdp5TEcJFWZD2D7SIkUc=
github.com/xdg-go/pbkdf2 v1.0.0/go.mod h1:jrpuAogTd400dnrH08LKmI/xc1MbPOebTwRqcT5RDeI=
github.com/xdg-go/scram v1.1.2/go.mod h1:RT/sEzTbU5y00aCK8UOx6R7YryM0iF1N2MOmC3kKLN4=
github.com/xdg-go/stringprep v1.0.4/go.mod h1:mPGuuIYwz7CmR2bT9j4GbQqutWS1zV24gijq1dTyGkM=
github.com/youmark/pkcs8 v0.0.0-20181117223130-1be2e3e5546d/go.mod h1:rHwXgn7JulP+udvsHwJoVG1YGAP6VLg4y9I5dyZdqmA=
[...]

PR #8032, opened 2023-12-13
fix(GoMod): Stop crashing with `NoSuchElementException`

fviernau:
The `GoMod` analysis crashes if both following conditions hold:

1. There is a constraint from 'main' to 'go' in the module graph.
2. There is a constraint from 'go' to on of it's dependencies in the
   module graph.

A similar bug has been spotted where only condition 1. held, but not
condition 2. This had been fixed by https://github.com/oss-review-toolkit/ort/commit/6b396606f23a2b0d9e593c9d0a62d3c6a01bc81a.

Generalize the fix to ignore all edges for which the module info is
missing. From the projects analyzed so far and from the functional
tests, there is strong evidence that the module info is only missing
the 'go' constraints as well as its transitive constraints. So, it
should be fine to skip edges without module info, because the
assumption is that these are always transitive constraints of the 'go'
module, or put otherwise: If a module is needed, then it's being
installed during the analysis which in turn implies that the module info
is available.

[1] `NoSuchElementException: Key go is missing in the map`


sschuberth (approved):
Should the commit message say to fix #7997?

PR #8034, opened 2023-12-13
refactor(model): Improve the name of a couple of setters

fviernau:


analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -48,7 +48,7 @@ import org.ossreviewtoolkit.model.config.RepositoryConfiguration
 import org.ossreviewtoolkit.model.orEmpty
 import org.ossreviewtoolkit.model.toYaml
 import org.ossreviewtoolkit.model.utils.PackageCurationProvider
-import org.ossreviewtoolkit.model.utils.addPackageCurations
+import org.ossreviewtoolkit.model.utils.setPackageCurations
```
sschuberth, line 51: Please also add to the commit message body something like this:

> The term "add" was misleading for these functions as data is actually overwritten instead of added. So use the term "set" instead to clarify.

sschuberth: Also strictly speaking, this is again changing public API, so a breaking change, though I'd personally be fine with regarding `OrtResultExtensions` not to be part of the "primary" public API as these are "extensions" only, so I'll leave it to you if you want to mark this as breaking or not.

sschuberth (changes requested):

sschuberth (approved):

PR #8036, opened 2023-12-14
chore(commands): Deprecate the `--skip-excluded` options

sschuberth:
This is a follow-up to 67297b2.


plugins/commands/advisor/src/main/kotlin/AdvisorCommand.kt
```diff
@@ -25,6 +25,7 @@ import com.github.ajalt.clikt.core.terminal
 import com.github.ajalt.clikt.parameters.options.associate
 import com.github.ajalt.clikt.parameters.options.convert
 import com.github.ajalt.clikt.parameters.options.default
+import com.github.ajalt.clikt.parameters.options.deprecated
```
fviernau, line 28: nit: Deprecate without `d` in commit message.

fviernau (changes requested):

fviernau (approved):

PR #8037, opened 2023-12-14
Minimal docker fixes

heliocastro:


Dockerfile
```diff
@@ -446,20 +446,20 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     subversion \
     && sudo rm -rf /var/lib/apt/lists/*
 
-RUN syft / --exclude '*/usr/share/doc' --exclude '*/etc' -o spdx-json --file /usr/share/doc/ort/ort-base.spdx.json
+RUN syft / --exclude '*/usr/share/doc' --exclude '*/etc' -o spdx-json=/usr/share/doc/ort/ort-base.spdx.json
```
sschuberth, line 449: Commit message: "Replace ... *with* ..."

But actually, I don't see any `--output` option being used, but a (weird) `=` syntax?!

sschuberth (changes requested):

Dockerfile
heliocastro: Changed the message as requested. Yep, new weird syntax

sschuberth: The title needs to be shortened now to pass the *commit-lint* check; any additional information need to go into the commit message body.

PR #8039, opened 2023-12-14
Remove `ReporterInput.packageConfigurationProvider`

### LINKED ISSUE #7921: Scanner: Avoid returning irrelevant issues

fviernau:
This PR first of all ensures that the `evaluator` and `reporter` and `list-copyrights` commands add all package configurations to the `OrtResult` before passing it around, and applies resulting simplifications.

This prepares for changing `OrtResult.getIssues()` to filter the issues by path excludes in a following PR.
As a I nice side effect, package configurations are now only resolved once per CLI command run instead of multiple times. 

Part of #7921.

### DIFF
--- a/helper-cli/src/main/kotlin/utils/Extensions.kt
+++ b/helper-cli/src/main/kotlin/utils/Extensions.kt
@@ -108,13 +108,12 @@ internal fun OrtResult.downloadSources(id: Identifier, sourceCodeOrigin: SourceC
 internal fun OrtResult.processAllCopyrightStatements(
     omitExcluded: Boolean = true,
     copyrightGarbage: Set<String> = emptySet(),
-    addAuthorsToCopyrights: Boolean = false,
-    packageConfigurationProvider: PackageConfigurationProvider = PackageConfigurationProvider.EMPTY
+    addAuthorsToCopyrights: Boolean = false
 ): List<ProcessedCopyrightStatement> {
     val result = mutableListOf<ProcessedCopyrightStatement>()
 
     val licenseInfoResolver = createLicenseInfoResolver(
-        packageConfigurationProvider = packageConfigurationProvider,
+        packageConfigurationProvider = this,
         copyrightGarbage = CopyrightGarbage(copyrightGarbage.toSortedSet()),
         addAuthorsToCopyrights = addAuthorsToCopyrights
     )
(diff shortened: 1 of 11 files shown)


sschuberth:
FYI, the SPM test update is part of my https://github.com/oss-review-toolkit/ort/pull/8119.

plugins/commands/reporter/src/main/kotlin/ReporterCommand.kt
```diff
@@ -54,6 +54,7 @@ import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.model.readValueOrDefault
 import org.ossreviewtoolkit.model.utils.CompositePackageConfigurationProvider
 import org.ossreviewtoolkit.model.utils.DefaultResolutionProvider
+import org.ossreviewtoolkit.model.utils.setPackageConfigurations
```
sschuberth, line 57: Commit message: "package curations" -> "package configurations"

plugins/commands/reporter/src/main/kotlin/ReporterCommand.kt
```diff
@@ -233,6 +234,7 @@ class ReporterCommand : OrtCommand(
                 DirPackageConfigurationProvider(packageConfigurationsDir)
             }
         }
+        ortResult = ortResult.setPackageConfigurations(packageConfigurationProvider)
```
sschuberth, line 237: Please add a blank line before this one.

helper-cli/src/main/kotlin/utils/Extensions.kt
```diff
@@ -113,7 +113,6 @@ internal fun OrtResult.processAllCopyrightStatements(
     val result = mutableListOf<ProcessedCopyrightStatement>()
 
     val licenseInfoResolver = createLicenseInfoResolver(
-        packageConfigurationProvider = this,
```
sschuberth: Commit message: Typo in `DefaultLicenseInforProvider` (remove the "r")

sschuberth (changes requested):

sschuberth (approved):

PR #8041, opened 2023-12-14
feat(ort-utils): Find names even if the version has an (ignorable) suffix

sschuberth:
So far ignorable suffixes were only ignored from the given (tag) names. Extend the logic to also ignore them from the given version.

### DIFF
--- a/utils/ort/src/main/kotlin/Utils.kt
+++ b/utils/ort/src/main/kotlin/Utils.kt
@@ -37,7 +37,8 @@ var printStackTrace = false
 private val versionSeparators = listOf('-', '_', '.')
 private val versionSeparatorsPattern = versionSeparators.joinToString("", "[", "]")
 
-private val ignorablePrefixSuffixPattern = listOf("rel", "release", "final").joinToString("|", "(", ")")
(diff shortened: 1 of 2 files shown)


utils/ort/src/main/kotlin/Utils.kt
```diff
@@ -37,7 +37,8 @@ var printStackTrace = false
 private val versionSeparators = listOf('-', '_', '.')
 private val versionSeparatorsPattern = versionSeparators.joinToString("", "[", "]")
 
-private val ignorablePrefixSuffixPattern = listOf("rel", "release", "final").joinToString("|", "(", ")")
+private val ignorablePrefixSuffix = listOf("rel", "release", "final")
```
fviernau, line 40: Could you help me understanding what the name means? e.g. why does it contain the word "prefix" ?

sschuberth: Have a look at line 43, which makes it more clear: These string may appear as prefixes as well as suffixes to tags (like "rel4.12" or "4.12.rel") and be safely ignored when trying to find a version match. This change now allows to ignore these strings also as a suffix to the given version.

fviernau (approved):

PR #8048, opened 2023-12-19
Preparations for removing `ReporterInput.resolutionProvider`

fviernau:
See individual commits.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/repoconfig/GenerateRuleViolationResolutionsCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/repoconfig/GenerateRuleViolationResolutionsCommand.kt
@@ -28,7 +28,6 @@ import com.github.ajalt.clikt.parameters.options.split
 import com.github.ajalt.clikt.parameters.types.enum
 import com.github.ajalt.clikt.parameters.types.file
 
-import org.ossreviewtoolkit.helper.utils.getUnresolvedRuleViolations
 import org.ossreviewtoolkit.helper.utils.readOrtResult
 import org.ossreviewtoolkit.helper.utils.replaceRuleViolationResolutions
 import org.ossreviewtoolkit.helper.utils.write
@@ -70,7 +69,7 @@ internal class GenerateRuleViolationResolutionsCommand : CliktCommand(
         val ortResult = readOrtResult(ortFile).replaceConfig(repositoryConfiguration)
 
         val generatedResolutions = ortResult
-            .getUnresolvedRuleViolations()
+            .getRuleViolations(omitResolved = true)
             .filter { it.severity in severity }
             .map {
                 RuleViolationResolution(
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -263,7 +263,7 @@ data class OrtResult(
         }
 
     /**
-     * Return all non-excluded issues which are not resolved by resolutions in the repository configuration of this
+     * Return all non-excluded issues which are not resolved by resolutions in the resolved configuration of this
      * [OrtResult] with severities equal to or over [minSeverity].
      */
     @JsonIgnore
@@ -272,7 +272,7 @@ data class OrtResult(
             .mapNotNull { (id, issues) -> issues.takeUnless { isExcluded(id) } }
             .flatten()
             .filter { issue ->
-                issue.severity >= minSeverity && getRepositoryConfigResolutions().issues.none { it.matches(issue) }
+                issue.severity >= minSeverity && getResolutions().issues.none { it.matches(issue) }
             }
 
     /**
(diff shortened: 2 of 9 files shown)


helper-cli/src/main/kotlin/commands/repoconfig/GenerateRuleViolationResolutionsCommand.kt
mnonnenmacher, line 1: Commit message: This is a breaking API change. (nit: and maybe "remove" instead of "eliminate")

model/src/main/kotlin/OrtResult.kt
mnonnenmacher, line 1: Commit message: Should now be changed to say "...to the resolved resolutions instead of those from the repository configuration".

plugins/commands/reporter/src/main/kotlin/ReporterCommand.kt
mnonnenmacher, line 1: Commit message: "ResolutionResolver" -> "ResolutionProvider"

plugins/commands/reporter/src/main/kotlin/ReporterCommand.kt
```diff
@@ -204,10 +205,10 @@ class ReporterCommand : OrtCommand(
             ortResult = ortResult.replaceConfig(config)
         }
 
-        val resolutionProvider =
-            ortResult.resolvedConfiguration.resolutions.takeUnless { refreshResolutions }?.let { resolutions ->
-                DefaultResolutionProvider(resolutions)
-            } ?: DefaultResolutionProvider.create(ortResult, resolutionsFile)
+        if (refreshResolutions) {
```
mnonnenmacher, line 208: Condition should be `refreshResolutions || ortResult.resolvedConfiguration.resolutions == null` to also resolve resolutions if they have not been resolved before (becuase the evaluator did not run).

plugins/reporters/evaluated-model/src/funTest/assets/reporter-test-input.yml
mnonnenmacher, line 1: Commit message:
* First line "resolution" -> "resolutions"
* "included in **the** `OrtResult` before passing it ~~around~~ to the reporters"

mnonnenmacher (changes requested):

helper-cli/src/main/kotlin/commands/repoconfig/GenerateRuleViolationResolutionsCommand.kt
fviernau: Eliminated eliminate as you prefer.

model/src/main/kotlin/OrtResult.kt
fviernau: The commit message was phrased how @sschuberth asked it to be. I don't understand the rationale why out should change now again, in particular because the commit diff hasn't change in the meantime. Please provide a rationale, and if you insist on the change it would be nice to have the exact sentence you want to see, because from the above I am unable to guess what your expectations are.

plugins/commands/reporter/src/main/kotlin/ReporterCommand.kt
fviernau: Nice catch.

model/src/main/kotlin/OrtResult.kt
mnonnenmacher: But didn't that discussion refer to the version of `getResolutions()` where it returned a merge of the resolutions in the repository config and the resolved config? As it now only returns those from the resolved config, I would also only refer to that. For example:

"Make the functions adhere to the resolutions from the resolved configuration instead of those from the repository configuration."

sschuberth: > But didn't that discussion refer to the version of `getResolutions()` where it returned a merge of the resolutions in the repository config and the resolved config?

Indeed it did.

fviernau: I've fixed it.

model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -146,29 +146,26 @@ class OrtResultTest : WordSpec({
...
                     result = AnalyzerResult(
                         projects = emptySet(),
                         packages = emptySet(),
                         issues = mapOf(
-                            Identifier("Maven:org.oss-review-toolkit:example:1.0") to
-                                listOf(
-                                    Issue(message = "Issue message to resolve", source = ""),
-                                    Issue(message = "Non-resolved issue", source = "")
-                                )
+                            Identifier("Maven:org.oss-review-toolkit:example:1.0") to listOf(
```
sschuberth, line 154: Just for my understanding: Is this change necessary because the original resolution from the repo config was not actually "baked" into the resolutions from the resolved configuration in the hard-coded test asset, although it usually *would* be if the repo config is read by ORT?

sschuberth: I.e. this is a bit similar to the adjustments of the test asset in your last commit 398a6215568581f12d567ed362aab1ed3834d68f, correct?

mnonnenmacher (approved):

PR #8069, opened 2023-12-28
feat(helper-cli): Add versioneer path exclude generator's patterns

fviernau:


helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -292,7 +292,8 @@ private val PATH_EXCLUDE_REASON_FOR_FILENAME = listOf(
     "setup.py" to BUILD_TOOL_OF,
     "test.js" to TEST_OF,
     "test.py" to TEST_OF,
-    "test_*.c" to BUILD_TOOL_OF
+    "test_*.c" to BUILD_TOOL_OF,
```
sschuberth, line 295: Unrelated, but it strikes me that `"test_*.c"` should probably rather be `TEST_OF`, or?

helper-cli/src/main/kotlin/utils/PathExcludeGenerator.kt
```diff
@@ -292,7 +292,8 @@ private val PATH_EXCLUDE_REASON_FOR_FILENAME = listOf(
...
+    "versioneer.py" to BUILD_TOOL_OF
```
sschuberth, line 296: Please add a reference to https://github.com/python-versioneer/python-versioneer (if that's the correct tool) to the commit message body so it becomes more apparent where this tools comes from and what it does.

sschuberth (changes requested):

sschuberth (approved):

PR #8072, opened 2023-12-29
build(Gradle): Use dashes to group dependencies

mnonnenmacher:
Use dashes in `libs.versions.toml` to group dependencies to improve overview. For example, `libs.jacksonDataformatXml` becomes `libs.jackson.dataformat.xml`.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -39,12 +39,12 @@ repositories {
 dependencies {
     implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
 
-    implementation(libs.detektPlugin)
-    implementation(libs.dokkatooPlugin)
-    implementation(libs.graalVmNativeImagePlugin)
     implementation(libs.jgit)
-    implementation(libs.kotlinPlugin)
-    implementation(libs.mavenPublishPlugin)
+    implementation(libs.plugin.detekt)
+    implementation(libs.plugin.dokkatoo)
+    implementation(libs.plugin.graalVmNativeImage)
+    implementation(libs.plugin.kotlin)
+    implementation(libs.plugin.mavenPublish)
 }
 
 val javaVersion = JavaVersion.current()
--- a/clients/oss-index/build.gradle.kts
+++ b/clients/oss-index/build.gradle.kts
@@ -30,5 +30,5 @@ dependencies {
     api(libs.retrofit)
 
     implementation(libs.bundles.kotlinxSerialization)
-    implementation(libs.retrofitConverterKotlinxSerialization)
+    implementation(libs.retrofit.converter.kotlinxSerialization)
 }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -73,77 +73,79 @@ kotlinSerialization = { id = "org.jetbrains.kotlin.plugin.serialization", versio
 versions = { id = "com.github.ben-manes.versions", version.ref = "versionsPlugin" }
 
 [libraries]
+# These are Maven coordinates for Gradle plugins, which is necessary to use them in precompiled plugin scripts.
+plugin-detekt = { module = "io.gitlab.arturbosch.detekt:detekt-gradle-plugin", version.ref = "detektPlugin" }
+plugin-dokkatoo = { module = "dev.adamko.dokkatoo:dokkatoo-plugin", version.ref = "dokkatooPlugin" }
+plugin-graalVmNativeImage = { module = "org.graalvm.buildtools:native-gradle-plugin", version.ref = "graalVmNativeImagePlugin" }
+plugin-kotlin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlinPlugin" }
+plugin-mavenPublish = { module = "com.vanniktech:gradle-maven-publish-plugin", version.ref = "mavenPublishPlugin" }
+
 antlr = { module = "org.antlr:antlr4", version.ref = "antlr" }
 asciidoctorj = { module = "org.asciidoctor:asciidoctorj", version.ref = "asciidoctorj" }
-asciidoctorjPdf = { module = "org.asciidoctor:asciidoctorj-pdf", version.ref = "asciidoctorjPdf" }
+asciidoctorj-pdf = { module = "org.asciidoctor:asciidoctorj-pdf", version.ref = "asciidoctorjPdf" }
 awsS3 = { module = "software.amazon.awssdk:s3", version.ref = "s3" }
 clikt = { module = "com.github.ajalt.clikt:clikt", version.ref = "clikt" }
 commonsCompress = { module = "org.apache.commons:commons-compress", version.ref = "commonsCompress" }
 cvssCalculator = { module = "us.springett:cvss-calculator", version.ref = "cvssCalculator" }
 cyclonedx = { module = "org.cyclonedx:cyclonedx-core-java", version.ref = "cyclonedx" }
-detektPlugin = { module = "io.gitlab.arturbosch.detekt:detekt-gradle-plugin", version.ref = "detektPlugin" }
-detektApi = { module = "io.gitlab.arturbosch.detekt:detekt-api", version.ref = "detektPlugin" }
-detektTest = { module = "io.gitlab.arturbosch.detekt:detekt-test", version.ref = "detektPlugin" }
+detekt-api = { module = "io.gitlab.arturbosch.detekt:detekt-api", version.ref = "detektPlugin" }
+detekt-test = { module = "io.gitlab.arturbosch.detekt:detekt-test", version.ref = "detektPlugin" }
 diffUtils = { module = "io.github.java-diff-utils:java-diff-utils", version.ref = "diffUtils" }
 diskLruCache = { module = "com.jakewharton:disklrucache", version.ref = "diskLruCache" }
-dokkatooPlugin = { module = "dev.adamko.dokkatoo:dokkatoo-plugin", version.ref = "dokkatooPlugin" }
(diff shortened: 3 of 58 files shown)


mnonnenmacher:
@oss-review-toolkit/core-devs This is a suggestions because I started using that syntax in my private projects. If you don't like it, I can close it.

clients/oss-index/build.gradle.kts
```diff
@@ -30,5 +30,5 @@ dependencies {
     api(libs.retrofit)
 
     implementation(libs.bundles.kotlinxSerialization)
-    implementation(libs.retrofitConverterKotlinxSerialization)
+    implementation(libs.retrofit.converter.kotlinxSerialization)
```
fviernau, line 33: Should the new name be entirely Lower case always? 
e.g. in this case `libs.retrofit.converter.kotlinx.serialization`?

gradle/libs.versions.toml
```diff
@@ -75,85 +75,85 @@ versions = { id = "com.github.ben-manes.versions", version.ref = "versionsPlugin
...
 diskLruCache = { module = "com.jakewharton:disklrucache", version.ref = "diskLruCache" }
 dokkatooPlugin = { module = "dev.adamko.dokkatoo:dokkatoo-plugin", version.ref = "dokkatooPlugin" }
-exposedCore = { module = "org.jetbrains.exposed:exposed-core", version.ref = "exposed" }
-exposedDao = { module = "org.jetbrains.exposed:exposed-dao", version.ref = "exposed" }
-exposedJavaTime = { module = "org.jetbrains.exposed:exposed-java-time", version.ref = "exposed" }
-exposedJdbc = { module = "org.jetbrains.exposed:exposed-jdbc", version.ref = "exposed" }
-exposedJson = { module = "org.jetbrains.exposed:exposed-json", version.ref = "exposed" }
+exposed-core = { module = "org.jetbrains.exposed:exposed-core", version.ref = "exposed" }
+exposed-dao = { module = "org.jetbrains.exposed:exposed-dao", version.ref = "exposed" }
+exposed-javaTime = { module = "org.jetbrains.exposed:exposed-java-time", version.ref = "exposed" }
```
fviernau, line 92: exposed-java-time?

gradle/libs.versions.toml
```diff
@@ -75,85 +75,85 @@ versions = { id = "com.github.ben-manes.versions", version.ref = "versionsPlugin
...
 jakartaMail = { module = "com.sun.mail:jakarta.mail", version.ref = "jakartaMail" }
 jgit = { module = "org.eclipse.jgit:org.eclipse.jgit", version.ref = "jgit" }
-jgitSshApache = { module = "org.eclipse.jgit:org.eclipse.jgit.ssh.apache", version.ref = "jgit" }
-jgitSshApacheAgent = { module = "org.eclipse.jgit:org.eclipse.jgit.ssh.apache.agent", version.ref = "jgit" }
-jiraRestClientApi = { module = "com.atlassian.jira:jira-rest-java-client-api", version.ref = "jiraRestClient" }
-jiraRestClientApp = { module = "com.atlassian.jira:jira-rest-java-client-app", version.ref = "jiraRestClient" }
+jgit-sshApache = { module = "org.eclipse.jgit:org.eclipse.jgit.ssh.apache", version.ref = "jgit" }
+jgit-sshApacheAgent = { module = "org.eclipse.jgit:org.eclipse.jgit.ssh.apache.agent", version.ref = "jgit" }
+jiraRestClient-api = { module = "com.atlassian.jira:jira-rest-java-client-api", version.ref = "jiraRestClient" }
+jiraRestClient-app = { module = "com.atlassian.jira:jira-rest-java-client-app", version.ref = "jiraRestClient" }
```
fviernau, line 115: jira-rest-client-* ?

gradle/libs.versions.toml
```diff
@@ -75,85 +75,85 @@ versions = { id = "com.github.ben-manes.versions", version.ref = "versionsPlugin
...
+kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serialization-json", version.ref = "kotlinxSerialization" }
 tomlkt = { module = "net.peanuuutz.tomlkt:tomlkt", version.ref = "tomlkt" }
-ktorClientCore = { module = "io.ktor:ktor-client-core", version.ref = "ktor" }
-ktorClientOkHttp = { module = "io.ktor:ktor-client-okhttp", version.ref = "ktor" }
+ktor-client-core = { module = "io.ktor:ktor-client-core", version.ref = "ktor" }
+ktor-client-okHttp = { module = "io.ktor:ktor-client-okhttp", version.ref = "ktor" }
 log4jApi = { module = "org.apache.logging.log4j:log4j-api", version.ref = "log4jApi" }
-log4jApiKotlin = { module = "org.apache.logging.log4j:log4j-api-kotlin", version.ref = "log4jApiKotlin" }
-log4jApiToSlf4j = { module = "org.apache.logging.log4j:log4j-to-slf4j", version.ref = "log4jApi" }
+log4jApi-kotlin = { module = "org.apache.logging.log4j:log4j-api-kotlin", version.ref = "log4jApiKotlin" }
```
fviernau, line 134: log4j-api-kotlin?

sschuberth:
> @oss-review-toolkit/core-devs This is a suggestions because I started using that syntax in my private projects. If you don't like it, I can close it.

I basically like it, but I agree with @fviernau's comments that we'd need to tweak the names a bit more to 1) never make them contain camel-case, and 2) always make them "hierarchic" by sometimes turning syllables around.

clients/oss-index/build.gradle.kts
mnonnenmacher: I tried to reflect the project hierarchy of the dependencies to have a meaningful grouping and not replace every dash in the module name with a dot.
For example, in this case the group is "converter" which contains all the retrofit converters for Jackson, kotlinx.serialization and so on, but there are not multiple converters for different "kotlinx" libraries, so to me it does not make sense to introduce a group for that.
Similar for the comments below:
* "exposed-java" is not a group of modules, so I used "exposed-javaTime".
* "jiraRestClient" is the project, so I kept it together.
* I thought about "logj4-api", but since we don't have any other log4j dependencies I found it reasonable to use "log4Api" as the group.

mnonnenmacher:
> > @oss-review-toolkit/core-devs This is a suggestions because I started using that syntax in my private projects. If you don't like it, I can close it.
> 
> I basically like it, but I agree with @fviernau's comments that we'd need to tweak the names a bit more to 1) never make them contain camel-case, and 2) always make them "hierarchic" by sometimes turning syllables around.

Please see my response to the first comment. I would prefer a logical hierarchy which does not prohibit camel-case, but I'm open to other approaches.

clients/oss-index/build.gradle.kts
sschuberth: > but there are not multiple converters for different "kotlinx" libraries

True, but on the other hand it seems awkward to eliminate *camelCase* in some cases, but not others.

mnonnenmacher: > True, but on the other hand it seems awkward to eliminate camelCase in some cases, but not others.

To me the goal is not to eliminate camel-case but to group dependencies. Therefore I didn't apply any changes to dependencies which do not belong to any group.

In the `libs.versions.toml` it does indeed look a bit inconsistent to have dashes and camel-case in the names, but my focus was more on the `build.gradle.kts` files. For example, `retrofit-converter-kotlinxSerialization` might look not so nice, but the resulting `retrofit.convert.kotlinxSerialization` in the Kotlin code is totally fine to me. The approach is taken from a project created using this wizard: https://kmp.jetbrains.com/

sschuberth: > In the `libs.versions.toml` it does indeed look a bit inconsistent to have dashes and camel-case in the names, but my focus was more on the `build.gradle.kts` files.

Ok, I could live with that, but I'd prefer if we could come up with some clear rules (that could also be implemented in code) on how to derive the TOML key name from the group / artifact name. But that seems to be difficult when looking at examples like

    exposed-javaTime = { module = "org.jetbrains.exposed:exposed-java-time", version.ref = "exposed" }

and

    jiraRestClient-api = { module = "com.atlassian.jira:jira-rest-java-client-api", version.ref = "jiraRestClient" }

where you need to understand the meaning of the dashed artifact name parts in order to apply some proper hierarchy.

sschuberth:
Interestingly, someone was just [asking](https://gradle-community.slack.com/archives/CAHSN3LDN/p1703970021215049) about conventions for aliases in version catalogs in Gradle Slack.

clients/oss-index/build.gradle.kts
mnonnenmacher: @fviernau: @sschuberth and me went through the change and agreed on the current state, if there is something you would like to change I think it would be better to do it in a separate smaller PR or to discuss it in the next core dev meeting.

fviernau: fine, sure just go ahead.

buildSrc/build.gradle.kts
```diff
@@ -39,12 +39,12 @@ repositories {
 dependencies {
     implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
 
-    implementation(libs.detektPlugin)
-    implementation(libs.dokkatooPlugin)
-    implementation(libs.graalVmNativeImagePlugin)
     implementation(libs.jgit)
-    implementation(libs.kotlinPlugin)
-    implementation(libs.mavenPublishPlugin)
+    implementation(libs.plugin.detekt)
```
sschuberth, line 43: @mnonnenmacher: I believe it was an oversight of us not to name this plural "plugin*s*". I'll amend your PR.

sschuberth: Ok, I just realized that's not possible due to a name clash with the `[plugins]` section, reverted.

sschuberth (approved):

PR #8076, opened 2023-12-29
Fossid simplifications

sschuberth:

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -248,6 +248,11 @@ class FossId internal constructor(
                 return ScanResult(provenance, details, summary)
             }
 
+            // FossId actually never uses the provenance determined by the scanner, but determines the source code to
+            // download itself based on the passed VCS information.
+            val provenance = pkg.vcsProcessed.revision.takeUnless { it.isBlank() }
+                ?.let { RepositoryProvenance(pkg.vcsProcessed, it) } ?: UnknownProvenance
+
             if (pkg.vcsProcessed.type != VcsType.GIT) {
                 val issue = createAndLogIssue(
                     source = name,
@@ -256,7 +261,7 @@ class FossId internal constructor(
                     severity = Severity.WARNING
                 )
 
-                return createSingleIssueResult(issue, UnknownProvenance)
+                return createSingleIssueResult(issue, provenance)
             }
 
             if (pkg.vcsProcessed.revision.isEmpty()) {
@@ -266,7 +271,7 @@ class FossId internal constructor(
                     severity = Severity.WARNING
                 )
 
-                return createSingleIssueResult(issue, UnknownProvenance)
+                return createSingleIssueResult(issue, provenance)
             }
 
             if (pkg.vcsProcessed.path.isNotEmpty()) {
@@ -276,7 +281,6 @@ class FossId internal constructor(
                         "path '${pkg.vcsProcessed.path}' set and scanning cannot be limited to paths.",
                     severity = Severity.WARNING
                 )
-                val provenance = RepositoryProvenance(pkg.vcsProcessed, pkg.vcsProcessed.revision)
 
                 return createSingleIssueResult(issue, provenance)
             }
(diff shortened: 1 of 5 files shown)


sschuberth:
FYI @nnobelis.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -275,9 +275,8 @@ class FossId internal constructor(
                         "path '${pkg.vcsProcessed.path}' set and scanning cannot be limited to paths.",
                     severity = Severity.WARNING
                 )
-                val provenance = RepositoryProvenance(pkg.vcsProcessed, pkg.vcsProcessed.revision)
 
-                return createSingleIssueResult(issue, provenance)
+                return createSingleIssueResult(issue, UnknownProvenance)
```
mnonnenmacher: The provenance parameter could now be removed from the `createSingleIssueResult` function as it is always the same.

But maybe it would be better to do the opposite change and provide the rejected provenance also in the two cases above to have more information in the ORT result?

sschuberth: > provide the rejected provenance also in the two cases above to have more information in the ORT result?

I was considering the same, but was uncertain about the intended semantics of `ScanResult.provenance` which is documented as "Provenance information about the *scanned* source code" (emphasis mine). As nothing is actually scanned in these cases, it felt wrong to return a known provenance. Do you recall the intended semantics here @mnonnenmacher?

sschuberth: > I agree with Martin, it feels we are losing information here by putting `UnknownProvenance`.

Well, we already "lost" information before this PR as some code paths already returned `UnknownProvenance`, and I just wanted to align that. But after having a chat with @mnonnenmacher I'm now doing some further refactoring to align this in the opposite way.

sschuberth:
@mnonnenmacher please have another look.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -276,9 +276,8 @@ class FossId internal constructor(
                         "path '${pkg.vcsProcessed.path}' set and scanning cannot be limited to paths.",
                     severity = Severity.WARNING
                 )
-                val provenance = RepositoryProvenance(pkg.vcsProcessed, pkg.vcsProcessed.revision)
 
-                return createSingleIssueResult(issue, provenance)
+                return createSingleIssueResult(issue, nestedProvenance?.root ?: UnknownProvenance)
```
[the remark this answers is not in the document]
sschuberth: Thanks for the hints, I've done all of it except for

> In this context it would also make sense to change the if conditions to work with `provenance` instead of `pkg.vcsProcessed`.

I'm running into a chicken-and-egg problem when using the provenance for error checking as I cannot construct a provenance with a blank revision. I'll think about this a bit more.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -236,110 +236,111 @@ class FossId internal constructor(
...
+                    }
                 }
+
+                ScanResult(provenance, details, summary)
             }
         }
 
-        logger.info { "Scan has been performed. Total time was $duration." }
+        logger.info {
+            val duration = with(result.summary) { java.time.Duration.between(startTime, endTime).toKotlinDuration() }
```
mnonnenmacher, line 341: Can't you just do `val duration = startTime - endTime`? (https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/-instant/#126222255%2FFunctions%2F-1003227268)

mnonnenmacher (changes requested):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
mnonnenmacher: Ok, I forgot that we use Java instants in the model...

mnonnenmacher: So you could do `endTime.toKotlinInstant() - startTime.toKotlinInstant()` but that's not better.

sschuberth: And `toKotlinInstant()` is only available with *kotlinx-datetime*, I believe.

mnonnenmacher (approved):

PR #8078, opened 2024-01-02
Minor python improvements

fviernau:
See individual commits.

### DIFF
--- a/plugins/package-managers/python/src/main/kotlin/Poetry.kt
+++ b/plugins/package-managers/python/src/main/kotlin/Poetry.kt
@@ -51,6 +51,13 @@ class Poetry(
     analyzerConfig: AnalyzerConfiguration,
     repoConfig: RepositoryConfiguration
 ) : PackageManager(name, analysisRoot, analyzerConfig, repoConfig), CommandLineTool {
+    companion object {
+        /**
+         * The name of the build system requirements and information file used by modern Python packages.
+         */
+        internal const val PYPROJECT_FILENAME = "pyproject.toml"
+    }
+
     class Factory : AbstractPackageManagerFactory<Poetry>("Poetry") {
         override val globsForDefinitionFiles = listOf("poetry.lock")
 
@@ -66,7 +73,7 @@ class Poetry(
     override fun transformVersion(output: String) = output.substringAfter("version ").removeSuffix(")")
 
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
-        val scopeName = parseScopeNamesFromPyproject(definitionFile.resolveSibling(Pip.PYPROJECT_FILENAME))
+        val scopeName = parseScopeNamesFromPyproject(definitionFile.resolveSibling(PYPROJECT_FILENAME))
         val resultsForScopeName = scopeName.associateWith { inspectLockfile(definitionFile, it) }
 
         val packages = resultsForScopeName
--- a/plugins/package-managers/python/src/test/kotlin/PoetryTest.kt
+++ b/plugins/package-managers/python/src/test/kotlin/PoetryTest.kt
@@ -25,17 +25,19 @@ import io.kotest.matchers.file.aFile
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNotBe
 
+import org.ossreviewtoolkit.plugins.packagemanagers.python.Pip.Companion.PYPROJECT_FILENAME
+
 class PoetryTest : WordSpec({
     "parseScopeNamesFromPyProject()" should {
         "return the 'main' scope even for a non-existing file" {
-            val pyprojectFile = tempdir().resolve(Pip.PYPROJECT_FILENAME)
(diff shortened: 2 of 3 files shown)


sschuberth:
Update for the test failure: https://github.com/oss-review-toolkit/ort/pull/8079

plugins/package-managers/python/src/test/kotlin/PoetryTest.kt
```diff
@@ -25,7 +25,7 @@ import io.kotest.matchers.file.aFile
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNotBe
 
-import org.ossreviewtoolkit.plugins.packagemanagers.python.Pip.Companion.PYPROJECT_FILENAME
+import org.ossreviewtoolkit.plugins.packagemanagers.python.Poetry.Companion.PYPROJECT_FILENAME
```
sschuberth, line 28: Semi-related: Should this apply the same `import`-simplification as in the previous commit?

plugins/package-managers/python/src/main/kotlin/Poetry.kt
```diff
@@ -51,6 +51,13 @@ class Poetry(
     analyzerConfig: AnalyzerConfiguration,
     repoConfig: RepositoryConfiguration
 ) : PackageManager(name, analysisRoot, analyzerConfig, repoConfig), CommandLineTool {
+    companion object {
+        /**
+         * The name of the build system requirements and information file used by modern Python packages.
+         */
+        internal const val PYPROJECT_FILENAME = "pyproject.toml"
```
sschuberth, line 58: FYI, Pip actually also uses `pyproject.toml` (implicitly), see https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/ and the commit message of ed0838125c562d171f48b1d2df0077e09e866cc6. So it's not wrong per se to keep the variable at `Pip`-scope, but from our current code base I see how it makes sense to move the constant. Overall, I'm undecided about the move.

plugins/package-managers/python/src/test/kotlin/PoetryTest.kt
fviernau: I'm not following. In which code location can the same simplification be applied?

plugins/package-managers/python/src/main/kotlin/Poetry.kt
fviernau: I preferred to move it because in our code base it is really only used by `Poetry` currently.

plugins/package-managers/python/src/test/kotlin/PoetryTest.kt
sschuberth: Nevermind, I somehow was mentally reading the diff of the first commit inverse, thinking that you've removed the import in favor of scoped access to the constant.

plugins/package-managers/python/src/test/kotlin/PoetryTest.kt
```diff
@@ -25,17 +25,19 @@ import io.kotest.matchers.file.aFile
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNotBe
 
+import org.ossreviewtoolkit.plugins.packagemanagers.python.Pip.Companion.PYPROJECT_FILENAME
+
 class PoetryTest : WordSpec({
     "parseScopeNamesFromPyProject()" should {
         "return the 'main' scope even for a non-existing file" {
-            val pyprojectFile = tempdir().resolve(Pip.PYPROJECT_FILENAME)
+            val pyprojectFile = tempdir().resolve(PYPROJECT_FILENAME)
```
sschuberth, line 33: I'm curious (really just trying to understand the rationale, no criticism): Does this make the (overall) code more compact in your eyes? Because comparing the number of characters, you're removing 12 chars in favor of adding ~90 chars. Granted, the chars are in a single import that's most of the time collapsed by the IDE, but I'm still interested in the "metric" you follow.

Or is this maybe rather about getting rid of the `Pip.` qualifier, so the refactoring in the next commit touches fewer places? I'd understand that, but it'd be a different rationale IMO.

plugins/package-managers/python/src/main/kotlin/Poetry.kt
sschuberth: Mind referring to ed0838125c562d171f48b1d2df0077e09e866cc6 then in order to document that you're aware of the original motivation to put this into `Pip`, but want to change that?

plugins/package-managers/python/src/test/kotlin/PoetryTest.kt
fviernau: It's about readability. As the reader is spending more time looking at the function implementation vs. looking at the imports, this IMO increases readability of the functions while sacrificing a bit on the readability of the imports. (I didn't aim at reducing the byte size of the file)

sschuberth:
A commit message body line is now [too long](https://github.com/oss-review-toolkit/ort/actions/runs/7388554136/job/20099625139?pr=8078#step:4:23).

sschuberth (approved):

PR #8080, opened 2024-01-02
Poetry detect python version

### LINKED ISSUE #7602: Poetry: Implement auto-detecting python version
When analysis is done via `Poetry.kt`, `OPTION_PYTHON_VERSION_DEFAULT` (see `Pip.kt`) is passed to the `-p` option of `python-inspector`.  The `poetry.lock` file always contains a version constraint for the Python version. If that constraint does not match with `OPTION_PYTHON_VERSION_DEFAULT`, then the analysis won't return any dependencies. Currently, the user can fix this by manually specifying the Python version via package manger `options`.

This is not necessary, because the value to pass to the `-p`option can be auto-detected, by just getting the Python version constraint from the `poetry.lock` and matching it against a `PYTHON_VERSIONS` (see again `Pip.kt`).

Note: In the long run it would make sense to have such logic present in `python-inspector` instead. However, this is a bigger effort, since support for e.g. Poetry would need to be implemented there. As several package managers (Poetry, PDM, ...) seem to be using the same file format for lockfiles, it could be more reasonable to add support for that lockfile format, instead of a package manager specific format. See also https://github.com/nexB/python-inspector/issues/12.

fviernau:
See individual commits.

Fixes #7602.

To be merged after https://github.com/oss-review-toolkit/ort/pull/8078.


plugins/package-managers/python/src/main/kotlin/Poetry.kt
```diff
@@ -120,10 +124,20 @@ class Poetry(
...
+        }.also {
             requirementsFile.parentFile.safeDeleteRecursively(force = true)
         }
     }
+
+    private fun detectPythonVersion(workingDir: File): String? {
+        val pyprojectFile = workingDir.resolve(PYPROJECT_FILENAME)
+        val constraint = getPythonVersionConstraint(pyprojectFile) ?: return null
+        return getPythonVersion(constraint)?.also {
+            logger().info { "Detected Python version '$it' from '$constraint'." }
```
sschuberth, line 138: This should be `logger.info`.

plugins/package-managers/python/src/main/kotlin/Poetry.kt
```diff
@@ -143,3 +157,37 @@ internal fun parseScopeNamesFromPyproject(pyprojectFile: File): Set<String> {
 
     return scopes
 }
+
+internal fun getPythonVersion(constraint: String): String? {
+    val rangeLists = constraint.split(",")
```
sschuberth, line 162: This could use the char-version of split: `split(',')`

plugins/package-managers/python/src/main/kotlin/Poetry.kt
```diff
@@ -143,3 +157,37 @@ internal fun parseScopeNamesFromPyproject(pyprojectFile: File): Set<String> {
...
+
+internal fun getPythonVersionConstraint(pyprojectTomlFile: File): String? {
+    val dependenciesSection = getTomlSectionContent(pyprojectTomlFile, "tool.poetry.dependencies")
+        ?: return null
+
+    return dependenciesSection
+        .split("\n")
+        .map { it.trim() }
+        .find { it.startsWith("python = ") }
+        ?.let { it.removePrefix("python = ") }?.removeSurrounding("\"")
```
sschuberth, line 182: I believe these 3 lines could be simplified to

        .firstNotNullOfOrNull { it.trim().withoutPrefix("python = ") }
        ?.removeSurrounding("\"")

sschuberth (changes requested):

sschuberth:
> Fixes #7602.

Please put this into the message of the commit that fixes the issue rather than in the PR.

sschuberth (approved):

PR #8083, opened 2024-01-05
Use `<pre>` block for issue messages

AJDurant:
Issue messages are mostly pre-formatted tool output, with linebreaks and alignment in the string. Wrapping the messages in `<pre>` blocks preserves this when displaying the messages, making them easier to read and understand.

### DIFF
--- a/plugins/reporters/web-app-template/src/components/IssuesTable.js
+++ b/plugins/reporters/web-app-template/src/components/IssuesTable.js
@@ -205,6 +205,7 @@ class IssuesTable extends React.Component {
                 key: 'message',
                 textWrap: 'word-break',
                 title: 'Message',
+                render: (text) => <pre>{text}</pre>,
                 ...getColumnSearchProps('message', filteredInfo, this)
             }
         );


mnonnenmacher:
@AJDurant: Thank you for the contribution! Can you please change the commit message to pass the commit-lint check?

```
✖   body's lines must not be longer than 75 characters [body-max-line-length]
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]
```

To fix the last two warnings I would suggest this title: "feat(reporter): Use `<pre>` block for issue messages"

@tsteenbe: Can you please review this change?

mnonnenmacher (approved):

mnonnenmacher:
> Thanks @mnonnenmacher, should be fixed now.

Thanks again @AJDurant.

@oss-review-toolkit/core-devs FYI, I have manually verified that issue messages are much more readable with this change.

mnonnenmacher:
Merged manually as the failing test is unrelated to the change.

PR #8089, opened 2024-01-08
refactor(scanner): Remove `findNestedProvenance()`

sschuberth:
As of 5559fdd this is the same as `getNestedProvenance()`.


scanner/src/main/kotlin/ScanController.kt
mnonnenmacher, line 1: `ScanController` is an internal class so this should not be a breaking change.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #8096, opened 2024-01-10
test(ort-utils): Add more Copyright symbol tests

sschuberth:


utils/ort/src/test/kotlin/CopyrightStatementsProcessorTest.kt
```diff
@@ -52,5 +52,32 @@ class CopyrightStatementsProcessorTest : WordSpec({
...
+                "Copyright (C) 2017 The ORT Project Authors",
+                "Copyright (c) 2022 The ORT Project Authors"
+            )
+
+            val actualResult = CopyrightStatementsProcessor.process(statements)
+
+            actualResult.processedStatements shouldHaveSize 2
+            actualResult.processedStatements.keys.first() shouldBe "Copyright (C) 2017 The ORT Project Authors"
+            actualResult.processedStatements.keys.last() shouldBe "Copyright (c) 2022 The ORT Project Authors"
+            actualResult.unprocessedStatements should beEmpty()
```
oheger-bosch, line 80: This is an interesting corner case I guess. From the logic, I would expect that the statements could be grouped to a single one, since the meaning of (C) and (c) should be the same. Is this not possible for legal reasons (IANAL)?

sschuberth: > Is this not possible for legal reasons (IANAL)?

That's exactly the question brought up by @vw-anton on [Stack](https://oss-review-toolkit.slack.com/archives/C9NNJ54B1/p1704886597287489), and [forwarded to ScanCode](https://github.com/nexB/scancode-toolkit/issues/3635). Obviously, opinions differ here. But that doesn't really matter in the context of this PR, as it's just supposed to highlight the current behavior.

oheger-bosch (approved):

PR #8097, opened 2024-01-10
fix(model): Correctly en- / decode a VCS subpath to / from PURLs

sschuberth:


model/src/main/kotlin/utils/PurlExtensions.kt
```diff
@@ -98,6 +97,8 @@ fun KnownProvenance.toPurlExtras(): PurlExtras =
                 subpath = vcsInfo.path
             )
         }
+
+        else -> PurlExtras()
```
mnonnenmacher, line 101: Better use `is UnknownProvenance` because then we get a compiler error in case a new provenance type is introduced (even though that's very unlikely).

mnonnenmacher (approved):

PR #8099, opened 2024-01-10
fix(common-utils): Do not extract TAR directory entries as files

### LINKED ISSUE #8098: Support .tar.gz archives that do NOT contain intermediate directory entries
When dealing with .tar.gz archives, ORT seems to have a problem in cases where the archive contains no intermediate diretory entries. For example, if the first entry in the archive is named _orjson-3.7.8/Cargo.toml_

You may argue that the archive in this case can be regarded as "corrupt", but other Unix tools like _gzip_ can handle that.

It seems to be a specific Java issue.

I currently have a problem in the scan step, when ORT downloads a _.tar.gz_ of source code of _orjson-3.7.8_.

**Quick steps to reproduce:**

* Download https://files.pythonhosted.org/packages/ac/4a/43daa65b7ed984612267983039a58703c48ccbc62561abb43ec7bc8b0663/orjson-3.7.8.tar.gz
* Replace the file _ort/utils/common/src/test/assets/test.tar.gz_ with _orjson-3.7.8.tar.gz_ (but keep the name, just exchange the test data)
* Run the tests in _ort/utils/common/src/test/kotlin/ArchiveUtilsTest.kt_
* This will result in test failures

sschuberth:
This works around [1]. Still keep the `isFile` check to be on the safe side and e.g. skip links.

Resolves #8098.

[1]: https://issues.apache.org/jira/browse/COMPRESS-657

### DIFF
--- a/utils/common/src/main/kotlin/ArchiveUtils.kt
+++ b/utils/common/src/main/kotlin/ArchiveUtils.kt
@@ -258,7 +258,9 @@ fun InputStream.unpackZip(targetDirectory: File) =
 fun InputStream.unpackTar(targetDirectory: File, filter: (ArchiveEntry) -> Boolean = { true }) =
     TarArchiveInputStream(this).unpack(
         targetDirectory,
-        { entry -> !(entry as TarArchiveEntry).isFile || File(entry.name).isAbsolute || !filter(entry) },
+        { entry ->
+            (entry as TarArchiveEntry).isDirectory || !entry.isFile || File(entry.name).isAbsolute || !filter(entry)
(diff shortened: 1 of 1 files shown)


oheger-bosch (approved):
Ideally, there would be a test to verify this fix.

sschuberth:
> Ideally, there would be a test to verify this fix.

As you'd need a specially crafted archive for this (which I'm not sure how to create), and the original archive from the issue report is rather large, I've abstained from adding a dedicated test case for now.

sschuberth:

PR #8104, opened 2024-01-11
refactor(spdx-utils): Improve the readability of `disjunctiveNormalFo…

sschuberth:
…rm()`

Shorten the case distinction via a helper function. The helper function can neither be local or an extension function due to its use of `contract`.

While at it, add some helpful code comment for the cases.


sschuberth:
See https://github.com/oss-review-toolkit/ort/pull/8106 instead.

PR #8107, opened 2024-01-12
fix(cargo): Only read checksum metadata entries as hashes

sschuberth:
Ignore metadata entries that do not start with "checksum ".

### DIFF
--- a/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
+++ b/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
@@ -51,6 +51,7 @@ import org.ossreviewtoolkit.model.orEmpty
 import org.ossreviewtoolkit.utils.common.CommandLineTool
 import org.ossreviewtoolkit.utils.common.splitOnWhitespace
 import org.ossreviewtoolkit.utils.common.unquote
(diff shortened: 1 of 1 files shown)


plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -121,7 +122,7 @@ class Cargo(
 
             else -> {
                 contents.metadata.mapNotNull { (k, v) ->
-                    (v as? String)?.let { k.unquote().removePrefix("checksum ") to v }
+                    k.unquote().withoutPrefix("checksum ")?.let { it to v }
```
mnonnenmacher, line 125: What was the purpose of checking if `v` is a string and why is it not required anymore?

sschuberth: This is a fixup for the migration done in 44523e412a504651cfad2f2617653818df7cc4a8 which did not really make sense to split out into a preceding commit (as I would then change `map {}` <-> `mapNotNull {}` and back again). I've clarified this in the commit message.

mnonnenmacher (approved):

PR #8111, opened 2024-01-12
fix(scanner): Keep the VCS path for a package scanner's reference package

sschuberth:
Previously, package scanner implementations did not get any information about which VCS path a package is located at in a VCS repository. However, that information might still be required, e.g. to point to the exact "file tree" for browsing the code of a package.

So keep the VCS path as part of the package. Now that package scanners do also get the (nested) provenance, that should be used to determine the root repository the package is located at.

The only package scanner implementation in ORT's code base is for FossId, which already handles a non-empty path by skipping the package as FossId cannot limit scans to sub-paths.

### DIFF
--- a/model/src/main/kotlin/Provenance.kt
+++ b/model/src/main/kotlin/Provenance.kt
@@ -60,13 +60,14 @@ data class ArtifactProvenance(
  */
 data class RepositoryProvenance(
     /**
-     * The VCS repository that was downloaded.
+     * The VCS information that was used for [resolving the revision][resolvedRevision]. It potentially still contains
+     * an unresolved "moving" VCS revision (e.g. a branch or tag name).
      */
     val vcsInfo: VcsInfo,
 
     /**
-     * The VCS revision of the source code that was downloaded, resolved from [vcsInfo] during download. Must not be
-     * blank, and must also be fixed revision, e.g. the SHA1 of a Git commit instead of a branch or tag name.
+     * The VCS revision resolved from [VCS information][vcsInfo]. It must neither be blank nor a "moving" VCS revision
+     * like a branch or tag name, but a fixed revision (e.g. the SHA1 of a Git commit).
      */
     val resolvedRevision: String
 ) : KnownProvenance {
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -252,10 +252,6 @@ class FossId internal constructor(
             pkg.vcsProcessed.revision.isEmpty() ->
                 "Package '${pkg.id.toCoordinates()}' has an empty VCS revision and cannot be scanned."
 
-            pkg.vcsProcessed.path.isNotEmpty() ->
-                "Ignoring package '${pkg.id.toCoordinates()}' from '${pkg.vcsProcessed.url}' as it has path " +
-                    "'${pkg.vcsProcessed.path}' set and scanning cannot be limited to paths."
-
             else -> null
         }
 
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -64,7 +64,6 @@ import org.ossreviewtoolkit.clients.fossid.model.status.ScanStatus
 import org.ossreviewtoolkit.clients.fossid.runScan
 import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.Issue
-import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.Provenance
 import org.ossreviewtoolkit.model.RepositoryProvenance
 import org.ossreviewtoolkit.model.ScanResult
@@ -236,11 +235,12 @@ class FossId internal constructor(
     private fun createSingleIssueSummary(startTime: Instant, issue: Issue) =
         ScanSummary.EMPTY.copy(startTime = startTime, endTime = Instant.now(), issues = listOf(issue))
 
-    override fun scanPackage(pkg: Package, nestedProvenance: NestedProvenance?, context: ScanContext): ScanResult {
+    override fun scanPackage(nestedProvenance: NestedProvenance?, context: ScanContext): ScanResult {
         val startTime = Instant.now()
 
         // FossId actually never uses the provenance determined by the scanner, but determines the source code to
-        // download itself based on the passed VCS information.
+        // download itself based on the passed VCS URL and revision, disregarding any VCS path.
+        val pkg = context.coveredPackages.first()
         val provenance = pkg.vcsProcessed.revision.takeUnless { it.isBlank() }
             ?.let { RepositoryProvenance(pkg.vcsProcessed, it) } ?: UnknownProvenance
 
--- a/scanner/src/test/kotlin/ScannerTest.kt
+++ b/scanner/src/test/kotlin/ScannerTest.kt
@@ -133,9 +133,9 @@ class ScannerTest : WordSpec({
             val pkgWithVcs = Package.new(name = "repository").withValidVcs()
 
             val scannerWrapper = spyk(FakePackageScannerWrapper()) {
-                every { scanPackage(pkgWithArtifact, any(), any()) } returns
+                every { scanPackage(any(), any()) } returns
                     createScanResult(pkgWithArtifact.artifactProvenance(), details)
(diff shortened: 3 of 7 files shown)


sschuberth:
> The only package scanner implementation in ORT's code base is for FossId, which already handles a non-empty path by skipping the package as FossId cannot limit scans to sub-paths.

@mnonnenmacher, @nnobelis, previously the VCS path could never have been non-empty it was cleared before FossId's `scanPackaage()` was called. Now it could be empty, causing FossId to skip packages that haven't been skipped before. So I ask myself, should we simply remove the check for a non-empty path from FossId? FossId itself does get passed the path anyway, and I wonder whether "scanning too much" does even hurt for a snippet scanner...

mnonnenmacher:
> The FossId server does get passed the path anyway, and I wonder whether "scanning too much" does even hurt for a snippet scanner...

Do you mean "does **not** get passed the path anyway"? In this case I would agree, better scan too much than scan nothing. Also, it sounds to me like removing the check would make this a non-breaking change from the perspective of FossId?

sschuberth:
> Do you mean "does **not** get passed the path anyway"?

Yes, right, sorry.

> removing the check would make this a non-breaking change from the perspective of FossId?

Correct. Removing the check would just allow to continue in the code path that doesn't make use of the path anyway. I'll amend the PR.

sschuberth:
> I'll amend the PR.

Done. Please have another look @mnonnenmacher.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
mnonnenmacher, line 1: Commit message nit: "scanner" -> "scanned" in the last line.

mnonnenmacher (changes requested):

sschuberth:
@mnonnenmacher, I've given this some more though and added three more commits, of which the last one actually removed the `pkg: Package` parameter from `scanPackage()` as its information is already contained in the other parameters.

nnobelis:
@sschuberth With your changes, should I expect some packages to be now skipped by FossID, where before they would be scanned ?

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -252,10 +252,6 @@ class FossId internal constructor(
             pkg.vcsProcessed.revision.isEmpty() ->
                 "Package '${pkg.id.toCoordinates()}' has an empty VCS revision and cannot be scanned."
 
-            pkg.vcsProcessed.path.isNotEmpty() ->
```
fviernau, line 255: commit-msg: "[..] for browsing the code of a package [..]" To me it's unclear to what "browsing" refers to in context of the scanner. Would it make sense to mention this explicitly?

fviernau (dismissed):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
sschuberth: I'm not certain how to phrase this in a better way (without going to broad either). It's about browsing the source code of a package in a third-party UI (similar to FOSSology) in order to create e.g. package configurations / license findings curation using a UI tool instead of writing YAML files.

sschuberth: I've mentioned "third-party UI" as part of the commit message now, hopefully that makes it clearer.

fviernau: Thanks! It's clear to me now.

fviernau:
LGTM.

nnobelis:
> > With your changes, should I expect some packages to be now skipped by FossID, where before they would be scanned ?
> 
> No, see [above](https://github.com/oss-review-toolkit/ort/pull/8111#issuecomment-1889670827) @nnobelis.

But if you remove the check, and since the path is not passed to FossID, couldn't it lead to some duplication e.g.

- scan X with path A
- scan X with path B
-> leads to X being scanned two times ?

Of course this cannot happen with the a simple repository where you give only one path. But I wonder what will happen for our customer with SPDX managed project and git submodules.

sschuberth:
> leads to X being scanned two times ?

That can't happen within a single scan as the scanner infrastructure [consolidates packages by provenance first](https://github.com/oss-review-toolkit/ort/blob/4d915d6e0aedf8548b6eb22116f06908b5fed11f/scanner/src/main/kotlin/Scanner.kt#L304).

In any case, this is nothing that this PR changes anyway.

nnobelis:
Ok, then it's ok with me.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -236,21 +236,24 @@ class FossId internal constructor(
...
         ScanSummary.EMPTY.copy(startTime = startTime, endTime = Instant.now(), issues = listOf(issue))
 
-    override fun scanPackage(pkg: Package, nestedProvenance: NestedProvenance?, context: ScanContext): ScanResult {
+    override fun scanPackage(nestedProvenance: NestedProvenance?, context: ScanContext): ScanResult {
         val startTime = Instant.now()
 
-        // FossId actually never uses the provenance determined by the scanner, but determines the source code to
-        // download itself based on the passed VCS information.
-        val provenance = pkg.vcsProcessed.revision.takeUnless { it.isBlank() }
-            ?.let { RepositoryProvenance(pkg.vcsProcessed, it) } ?: UnknownProvenance
```
mnonnenmacher, line 245: I don't like removing the `pkg` argument here for two reasons:
* Previously, FossId always used the VCS information, now it only does so if the resolved provenance is a `RepositoryProvenance`. That means that if someone is using FossId to scan packages this does not work properly anymore if they configure the source code origin priority to prioritize source artifacts over VCS.
* There could be package scanners that use other input than the provenance, like the package id or Purl. For example, doesn't clearly defined scan based on identifiers instead of provenance?

mnonnenmacher (changes requested):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
sschuberth: > Previously, FossId always used the VCS information

Good point, FossId never adhered to the configured `sourceCodeOrigin`.

> There could be package scanners that use other input than the provenance, like the package id or Purl.

That's actually exactly what the Double Open scanner does. But still the reference `pkg` is not needed, as `context.coveredPackages` is enought (and also better, in the sense that the packages still contains their VCS path).

So, I could solve both issues by operating on `context.coveredPackages`, and still remove the "artificial" `pkg` parameter. I really don't like it as it does reflect a real package when it comes from a VCS with a path.

model/src/main/kotlin/Provenance.kt
```diff
@@ -60,7 +60,8 @@ data class ArtifactProvenance(
  */
 data class RepositoryProvenance(
     /**
-     * The VCS repository that was downloaded.
+     * The VCS information that was passed to the downloader. This potentially still contains an unresolved "moving"
+     * VCS revision, like "master" for Git.
```
mnonnenmacher, line 64: Maybe "The VCS information that was used to resolve this provenance. ..." as it is a bit more precise. In the same sense the comment for `resolvedRevision` could be changed to say "The resolved VCS revision of the source code."

plugins/scanners/fossid/src/main/kotlin/FossId.kt
mnonnenmacher: That sounds good to me.

scanner/src/test/kotlin/ScannerTest.kt
```diff
@@ -158,9 +158,8 @@ class ScannerTest : WordSpec({
                 )
             )
 
-            verify(exactly = 1) {
-                scannerWrapper.scanPackage(pkgWithArtifact, any(), any())
-                scannerWrapper.scanPackage(pkgWithVcs, any(), any())
+            verify(exactly = 2) {
+                scannerWrapper.scanPackage(any(), any())
```
mnonnenmacher, line 162: This test now partly loses it's meaning, I think it would be good to replace the `any()` for `context` with an actual context to ensure that `context.coveredPackages` is correct (same below).

mnonnenmacher (changes requested):

model/src/main/kotlin/Provenance.kt
sschuberth: I've chosen a slightly different wording as I didn't want to say "resolve this provenance" because it's not the provenance itself that's being resolved, but provenance is the result of resolution.

mnonnenmacher (approved):

PR #8121, opened 2024-01-16
docs(model): Improve various `ProvenanceResolutionResult` texts

sschuberth:


model/src/main/kotlin/ProvenanceResolutionResult.kt
```diff
@@ -34,47 +34,47 @@ data class ProvenanceResolutionResult(
...
+                    "which is not allowed."
             }
         }
 
         if (packageProvenanceResolutionIssue != null) {
             require(nestedProvenanceResolutionIssue == null) {
-                "Nested provenance resolution issue is not null, even though nested provenance resolution was not " +
-                    "executed."
+                "Nested provenance resolution was not executed as already the (root) provenance resolution had an " +
+                    "issue, but still the nested provenance resolution also has an issue."
```
fviernau, line 77: How about replacing "also has an issue." with "issue is set.".

sschuberth: I believe "cannot be executed" is imprecise because it sounds like something that would happen in the future, if at all, but instead the execution, if any, has already happened, and this is an after-the-fact check of the results of resolution execution.

PR #8124, opened 2024-01-17
style: Conistently put @SerialName annotation on separate lines

fviernau:


plugins/package-managers/python/src/main/kotlin/utils/PythonInspector.kt
```diff
@@ -91,7 +91,8 @@ internal object PythonInspector : CommandLineTool {
 
     @Serializable
     internal data class Result(
-        @SerialName("files") val projects: List<Project>,
+        @SerialName("files")
+        val projects: List<Project>,
```
sschuberth, line 95: I can't find the PR discussion right now, but I recall that @mnonnenmacher preferred the one-liner syntax when introducing this, whereas I favored the two-liner syntax (with an empty line for separation). Personally, I still prefer the two-liner syntax (but then with the mentioned empty line added), but don't care enough anymore to spend time on the style discussion. Still, I'm curious what @mnonnenmacher says.

fviernau: Ah now I recall at least as well that there was some discussion, however no details anymore.

mnonnenmacher: Yes, I preferred the one-line syntax because you can see a lot more properties on a single screen without having to scroll.

fviernau: @mnonnenmacher so you still prefer to not merge a PR like this one ?

mnonnenmacher: Yes, I prefer the current formatting, but if you two prefer the other one you can outvote me.

fviernau: Ok, ditching it then

Issue #8123, opened 2024-01-17
SPM: Implement support for `PackageResolved` file format version 2


fviernau:
More recent XCode version use a different format for the `Package.resolved` file, see also [1].
Analyzing such file with ORT's SPM integration only leads to an analyzer issue due to failing to de-serialize [2].

[1] https://github.com/teufelaudio/SwiftPackageAcknowledgement/issues/7
[2] 
```
 ERROR java.lang.Object - SPM failed to resolve dependencies for path 'DiiaSimple.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved': MissingFieldException: Field 'object' is required for type with serial name 'org.ossreviewtoolkit.plugins.packagemanagers.spm.PackageResolved', but it was missing at path: $
Caused by: MissingFieldException: Field 'object' is required for type with serial name 'org.ossreviewtoolkit.plugins.packagemanagers.spm.PackageResolved', but it was missing
``

PR #8117, opened 2024-01-15
Fix Pub namespace

mnonnenmacher:
Do not set namespaces for Pub packages because Pub does not support namespaces. See the commit messages for details.

PR #8125, opened 2024-01-17
Several SPM related fixes

fviernau:
See individual commits.

Part of #8123.

### DIFF
--- a/plugins/package-managers/spm/src/main/kotlin/Spm.kt
+++ b/plugins/package-managers/spm/src/main/kotlin/Spm.kt
@@ -141,13 +141,13 @@ class Spm(
 
     private fun projectFromDefinitionFile(definitionFile: File): Project {
         val vcsInfo = VersionControlSystem.forDirectory(definitionFile.parentFile)?.getInfo().orEmpty()
-        val (author, project) = parseAuthorAndProjectFromRepo(repositoryURL = vcsInfo.url)
+        val (author, _) = parseAuthorAndProjectFromRepo(repositoryURL = vcsInfo.url)
 
         val projectIdentifier = Identifier(
             type = managerName,
             version = vcsInfo.revision,
             namespace = author.orEmpty(),
-            name = project ?: definitionFile.parentFile.relativeTo(analysisRoot).invariantSeparatorsPath
+            name = getFallbackProjectName(analysisRoot, definitionFile)
         )
 
         return Project(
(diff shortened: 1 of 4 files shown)


plugins/package-managers/spm/src/main/kotlin/Spm.kt
```diff
@@ -141,13 +141,13 @@ class Spm(
...
         val vcsInfo = VersionControlSystem.forDirectory(definitionFile.parentFile)?.getInfo().orEmpty()
-        val (author, project) = parseAuthorAndProjectFromRepo(repositoryURL = vcsInfo.url)
+        val (author, _) = parseAuthorAndProjectFromRepo(repositoryURL = vcsInfo.url)
 
         val projectIdentifier = Identifier(
             type = managerName,
             version = vcsInfo.revision,
             namespace = author.orEmpty(),
-            name = project ?: definitionFile.parentFile.relativeTo(analysisRoot).invariantSeparatorsPath
+            name = definitionFile.parentFile.relativeTo(analysisRoot).invariantSeparatorsPath
```
sschuberth, line 150: IMO we should strive to use the `getFallbackProjectName()` helper function like @mnonnenmacher recently did as part of https://github.com/oss-review-toolkit/ort/pull/8117.

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
```diff
@@ -85,11 +85,11 @@ data class LibraryDependency(
 
     override val id: Identifier
         get() {
-            val (author, project) = parseAuthorAndProjectFromRepo(repositoryUrl)
+            val (author, _) = parseAuthorAndProjectFromRepo(repositoryUrl)
```
sschuberth, line 86: IMO it's hard to understand why this is correct without looking up the implementation of `parseAuthorAndProjectFromRepo()`. What would have helped me during the review is a sentence like this in the commit message:

"The `parseAuthorAndProjectFromRepo()` function tries to parse the project name from the repository URL. However, this requires the URL to point to a known VCS host, and generally the dedicated `name` field from the SPM data model should be used instead."

sschuberth (changes requested):
I would have expected more clean-up commits regarding `parseAuthorAndProjectFromRepo()` in this PR, as all callers now only seem to use the `author`. On top of that, the `author` is also not used as an "author" anymore, but as an organization /  namespace.

So how about changing that function to just `parseNamespaceFromRepo()`?

fviernau:
> I would have expected more clean-up commits regarding `parseAuthorAndProjectFromRepo()` in this PR, as all callers now only seem to use the `author`. On top of that, the `author` is also not used as an "author" anymore, but as an organization / namespace.
> 
> So how about changing that function to just `parseNamespaceFromRepo()`?

I prefer to do so in a following PR, or rather remove the function entirely instead. (I believe it's likely this will become possible with upcoming changes)

sschuberth (approved):

PR #8126, opened 2024-01-17
fix(pub): Do not rely on the package `name` to be present

Syyllinen:
If `readPackageInfoFromCache` returns `EMPTY_JSON_NODE`, then `textValue()` throws java.lang.NullPointerException

### DIFF
--- a/plugins/package-managers/pub/src/main/kotlin/Pub.kt
+++ b/plugins/package-managers/pub/src/main/kotlin/Pub.kt
@@ -548,7 +548,7 @@ class Pub(
                         pkgInfoFromLockFile["source"].textValueOrEmpty() == "git" -> {
                             val pkgInfoFromYamlFile = readPackageInfoFromCache(pkgInfoFromLockFile, workingDir)
 
-                            rawName = pkgInfoFromYamlFile["name"].textValue() ?: packageName
+                            rawName = pkgInfoFromYamlFile["name"]?.textValue() ?: packageName
                             description = pkgInfoFromYamlFile["description"].textValueOrEmpty().trim()
                             homepageUrl = pkgInfoFromYamlFile["homepage"].textValueOrEmpty()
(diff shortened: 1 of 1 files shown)


sschuberth:
Thanks for the contribution!

Commit message:
- "bug" is not a valid Conventional Commit type, use "fix" instead.
- Limit the commit message title to ~75 chars, and put the rest into the commit message body (also hard-wrapped at 75 chars).

Syyllinen:
> Thanks for the contribution!
> 
> Commit message:
> 
> * "bug" is not a valid Conventional Commit type, use "fix" instead.
> * Limit the commit message title to ~75 chars, and put the rest into the commit message body (also hard-wrapped at 75 chars).

Thanks for the feedback! Commit message updated.

fviernau (dismissed):
@Syyllinen  please remove the  merge commit by just rebasing your branch on top of master. Merge commits are not allowed in this project.

sschuberth:
> fix(pub): Avoid java.lang.NullPointerException

I also still feel that this could be more telling from the title, like

   fix(pub): Do not rely on the package `name` to be present

That explains better in what context the NPE could have occurred.

Syyllinen:
> @Syyllinen please remove the merge commit by just rebasing your branch on top of master. Merge commits are not allowed in this project.

Reverted

Syyllinen:
> > fix(pub): Avoid java.lang.NullPointerException
> 
> I also still feel that this could be more telling from the title, like
> 
> ```
> fix(pub): Do not rely on the package `name` to be present
> ```
> 
> That explains better in what context the NPE could have occurred.

I trust your expertise on this one. Commit message updated.

sschuberth (approved):

sschuberth:
@Syyllinen there's not need to rebase frequently if just in order to get rid of unrelated test failures. We'll recognize that and either retrigger individual tests, rebase on our end, or override check and merge anyway.

PR #8132, opened 2024-01-18
Migrate Pub identifiers

mnonnenmacher:
Add an option to the migrate command to migrate Pub identifiers. See the commit messages for details.


plugins/commands/migrate/src/main/kotlin/MigrateCommand.kt
```diff
@@ -56,6 +56,12 @@ class MigrateCommand : OrtCommand(
     ).convert { it.expandTilde() }
         .file(mustExist = true, canBeFile = false, canBeDir = true, mustBeWritable = true, mustBeReadable = true)
 
+    private val pubIds by option(
+        "--pub-ids",
+        help = "Convert Pub package IDs in curations and configurations to the new format that has no namespace."
+    ).convert { it.expandTilde() }
+        .file(mustExist = true, canBeFile = false, canBeDir = true, mustBeWritable = true, mustBeReadable = true)
```
fviernau, line 63: Would it make sense to have a single option for all IDs where you specify a CSV e.g.: "NUGET,PUB"?

fviernau (approved):

plugins/commands/migrate/src/main/kotlin/MigrateCommand.kt
sschuberth: Instead, I was thinking of having shortcut-options like `--from-version` and `--to-version` that generally do all migrations required when upgrading from / to the given versions.

sschuberth (approved):

plugins/commands/migrate/src/main/kotlin/MigrateCommand.kt
mnonnenmacher: That probably doesn't make it easier to use.

sschuberth: It does not? From a user perspective, I believe that does *exactly* what they want: Users don't actually care so much *what* has changed, they just want to be able to upgrade from one version to the other without breakages.

sschuberth: Or were you referring to @fviernau's comment instead of mine, @mnonnenmacher?

PR #8136, opened 2024-01-19
feat(jenkins): Add a label to link back the the build URL

sschuberth:
This can be used e.g. from the web-app report's "About" screen to jump back to the build that created the report.

### DIFF
--- a/integrations/jenkins/Jenkinsfile
+++ b/integrations/jenkins/Jenkinsfile
@@ -487,7 +487,7 @@ pipeline {
                     /opt/ort/bin/set_gradle_proxy.sh
 
                     rm -fr out/results
-                    /opt/ort/bin/ort $ORT_OPTIONS analyze -i "$PROJECT_DIR/source" -o out/results/analyzer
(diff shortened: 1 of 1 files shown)


integrations/jenkins/Jenkinsfile
```diff
@@ -487,7 +487,7 @@ pipeline {
                     /opt/ort/bin/set_gradle_proxy.sh
 
                     rm -fr out/results
-                    /opt/ort/bin/ort $ORT_OPTIONS analyze -i "$PROJECT_DIR/source" -o out/results/analyzer
+                    /opt/ort/bin/ort $ORT_OPTIONS analyze -i "$PROJECT_DIR/source" -o out/results/analyzer -l JENKINS_BUILD_URL=$BUILD_URL
```
fviernau, line 490: I wonder if `CI_BUILD_URL` would make sense to use across CI integrations. What do you think?

fviernau (approved):

integrations/jenkins/Jenkinsfile
fviernau: Or `CI_JOB_URL`.

sschuberth: I deliberately went for the same name that Jenkins uses plus a `JENKINS_` prefix. I didn't see a need to use a generic name here because the label is not going to be used programmatically or so.

PR #8139, opened 2024-01-19
Marika 20240112

marikanieminen:
Reasonings for moving any mappings from unambiguous to ambiguous comprise of the following:
-	Mapping is incorrect / might be incorrect
o	The mapped text refers to a modified version of the license
o	The mapped text refers to a completely different license
-	Mapping is ambiguous on the part of missing version numbering 
o	if license only has one version, I found it not to be ambiguous
-	The mapped text contains a link that does not function hence the accuracy cannot be evaluated
-	The mapped text contains mention of multiple licenses and does not specify whether these both or either apply (AND/OR)
-	The mapped text, e.g. an acronym, could be referring to multiple different licenses and it is not clear which is meant

### DIFF
--- a/utils/spdx/src/main/resources/declared-license-mapping.yml
+++ b/utils/spdx/src/main/resources/declared-license-mapping.yml
@@ -83,14 +83,12 @@
 "BSD - See ndg/httpsclient/LICENCE file for details": BSD-3-Clause
 "BSD License for HSQL": BSD-3-Clause
 "BSD-Style + Attribution": BSD-3-Clause-Attribution
-"BSD-derived (http://www.repoze.org/LICENSE.txt)": LicenseRef-scancode-repoze
 "Berkeley Software Distribution (BSD) License": BSD-2-Clause
 "Bouncy Castle Licence": MIT
 "Bouncy Castle License": MIT
 "CDDL v1.1 / GPL v2 dual license": CDDL-1.0 OR GPL-2.0-only
 "CUP Parser Generator Copyright Notice, License, and Disclaimer": HPND
 "CeCILL Licence française de logiciel libre": CECILL-2.0
-"Commons Clause": LicenseRef-scancode-commons-clause
 "Eiffel Forum License (EFL)": EFL-2.0
 "Eiffel Forum License": EFL-2.0
 "Expat (MIT/X11)": MIT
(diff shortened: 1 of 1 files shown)


sschuberth:
FYI @oss-review-toolkit/core-devs, Marika works at HH Partners and has legal background.

tsteenbe (changes requested):
I applaud Marika for her first contribution to ORT. The contents of this file requires us as community to define some clear governance rules and we need to define as a community to avoid endless discussions. I propose to halt this PR until rules are defined and I would appreciate it if @marikanieminen could help us define such rules with the community.

sschuberth:
> The contents of this file requires us as community to define some clear governance rules and we need to define as a community to avoid endless discussions.

Note that "define as a community" and "avoid endless discussions" are usually somewhat contradicting, depending on the size of the community. The more people you involve, the harder it gets to reach a consensus. Therefore I propose to decide by ORT TSC majority vote.

My proposal for the concrete mapping is as follows: In general, we should aim for only listing 100% clear license strings as unambiguous, no fuzzyness or room for interpretation here. And that should hold for past, current, and future times.

That is (out of my head, no guarantee for completeness yet),
- I'd disagree with "if license only has one version, I found it not to be ambiguous" as that could change in the future if a new version of a license gets published, which would require us to update the mapping (if we even realize that a new version of a license is published),
- an entry without a version although the license is versioned cannot be unambiguous,
- an entry with only a major version, esp. if there are multiple versions of the license, cannot be unambiguous no matter how "common" a specific minor version of the license may be,
- URLs, unless permalinks, cannot be unambiguous as their contents might change over time,
- an entry that refers to a project name only cannot be unambiguous as the project might change its license,
- an entry that refers to other files, like `LICENSE`, cannot be unambiguous as the contents of the file might change.

On the other hand,
- clear acronyms should be accepted as unambiguous,
- typos that don't create any ambiguity should be tolerated,
- changes in whitespace, punctuation (unless part of the version) and special chars should be tolerated,
- additional words (like "license") that do not change the meaning should be tolerated.

tsteenbe:
Discussed in Kotlin dev meeting of Feb 17, 2026  agreed with @mnonnenmacher  @MarcelBochtler , @fviernau  @tsteenbe we need configurability of declared license mappings (https://github.com/oss-review-toolkit/ort/issues/10498) before removing mappings so any changes the ORT maintainers make to ORT-built-in declared license mappings can be easily changed by the end user by specifying their desired declared license mappings in config.yml a la [detected license mapping](https://github.com/oss-review-toolkit/ort/blob/main/model/src/main/resources/reference.yml#L205-L207).

fviernau:
> Discussed in Kotlin dev meeting of Feb 17, 2026 agreed with @mnonnenmacher @MarcelBochtler , @fviernau @tsteenbe we need configurability of declared license mappings (#10498) before removing mappings so any changes

I've created a dedicated ticket, to have the discussion more focused on just the declared license mapping:
https://github.com/oss-review-toolkit/ort/issues/11459.

PR #8140, opened 2024-01-19
Further `Spm` improvements

fviernau:
Primarily separate the models for `Package.resolved` (lockfile) and `Package.swift` (definitionFile) in order to prepare for
handling multiple file format versions of `Package.resolved`. Furthermore, clarify miscellaneous related things a bit.

Part of #8123.

### DIFF
--- a/plugins/package-managers/spm/src/main/kotlin/Spm.kt
+++ b/plugins/package-managers/spm/src/main/kotlin/Spm.kt
@@ -89,8 +89,8 @@ class Spm(
     }
 
     /**
-     * Resolves dependencies when the final build target is an app and no Package.swift is available.
-     * This method parses dependencies from the Package.resolved file.
+     * Resolves dependencies when only a lockfile aka `Package.Resolved` is available. This commonly applies to e.g.
+     * Xcode projects which only have a lockfile, but no `Package.swift` file.
      */
     private fun resolveAppDependencies(definitionFile: File): List<ProjectAnalyzerResult> {
         val resolved = definitionFile.inputStream().use { json.decodeFromStream<PackageResolved>(it) }
--- a/plugins/package-managers/spm/src/main/kotlin/Spm.kt
+++ b/plugins/package-managers/spm/src/main/kotlin/Spm.kt
@@ -83,7 +83,7 @@ class Spm(
         requireLockfile(definitionFile.parentFile) { definitionFile.name != PACKAGE_SWIFT_NAME }
 
         return when (definitionFile.name) {
-            PACKAGE_SWIFT_NAME -> resolveLibraryDependencies(definitionFile)
+            PACKAGE_SWIFT_NAME -> resolveDefinitionFileDependencies(definitionFile)
             else -> resolveLockfileDependencies(definitionFile)
         }
     }
@@ -108,11 +108,11 @@ class Spm(
      * This method parses dependencies from `swift package show-dependencies --format json` output.
      * Also, this method provides parent-child associations for parsed dependencies.
      */
-    private fun resolveLibraryDependencies(definitionFile: File): List<ProjectAnalyzerResult> {
-        val project = projectFromDefinitionFile(definitionFile)
+    private fun resolveDefinitionFileDependencies(packageSwiftFile: File): List<ProjectAnalyzerResult> {
+        val project = projectFromDefinitionFile(packageSwiftFile)
 
         val result = run(
-            definitionFile.parentFile,
+            packageSwiftFile.parentFile,
             "package",
             "show-dependencies",
             "--format",
--- a/plugins/package-managers/spm/src/main/kotlin/Spm.kt
+++ b/plugins/package-managers/spm/src/main/kotlin/Spm.kt
@@ -104,11 +104,9 @@ class Spm(
     }
 
     /**
-     * Resolves dependencies when the final build target is a library and Package.swift is available.
+     * Resolves dependencies of a `Package.swift` file.
      * This method parses dependencies from `swift package show-dependencies --format json` output.
      * Also, this method provides parent-child associations for parsed dependencies.
-     *
-     * Only used when analyzerConfig.allowDynamicVersions is set to true.
      */
     private fun resolveLibraryDependencies(definitionFile: File): List<ProjectAnalyzerResult> {
         val project = projectFromDefinitionFile(definitionFile)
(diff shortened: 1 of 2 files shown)


plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
```diff
@@ -33,14 +33,6 @@ import org.ossreviewtoolkit.utils.ort.normalizeVcsUrl
 
 val json = Json { ignoreUnknownKeys = true }
 
-abstract class SpmDependency {
```
sschuberth, line 36: Commit message nits:
- "aka." without "." at the end.
- Foot notes should have an additional ":", like "[1]:".

sschuberth: > The inheritance was used for code re-use

While that's true, I'm not sure whether that was the *only* purpose. I believe it was also used to express that "both libraries and applications are dependencies". This semantic is now lost. It might make sense to keep `SpmDependency` as an `interface` to still express that semantic. I still need to review the following commits to make up my mind about this myself.

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
```diff
@@ -99,31 +99,27 @@ data class AppDependency(
...
-    val vcs: VcsInfo
-        get() {
-            val vcsInfoFromUrl = VcsHost.parseUrl(repositoryUrl)
-
-            if (vcsInfoFromUrl.revision.isBlank() && state != null) {
-                when {
-                    !state.revision.isNullOrBlank() -> return vcsInfoFromUrl.copy(revision = state.revision)
-                    !state.version.isNullOrBlank() -> return vcsInfoFromUrl.copy(revision = state.version)
-                }
+    fun toPackage(): Package {
```
sschuberth: Actually, this could be a regular / lazy property now, to avoid re-evaluation on each function call as the input properties (like `repositoryUrl`) never change.

sschuberth: Why isn't the same done for `LibraryDependency`?

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
```diff
@@ -90,35 +90,35 @@ data class AppDependency(
...
-                    !revision.isNullOrBlank() -> "revision-$revision"
-                    !branch.isNullOrBlank() -> "branch-$branch"
-                    else -> ""
-                }
-            }.orEmpty()
-        )
-
-        val vcsInfoFromUrl = VcsHost.parseUrl(repositoryUrl)
-        val vcsInfo = if (vcsInfoFromUrl.revision.isBlank() && state != null) {
+internal fun AppDependency.toPackage(): Package {
```
sschuberth: Ok, now this cannot be a property anymore, also no an extension property, as access to `state` is required, which comes from the parent class.

plugins/package-managers/spm/src/main/kotlin/Spm.kt
```diff
@@ -89,8 +89,8 @@ class Spm(
     }
 
     /**
-     * Resolves dependencies when the final build target is an app and no Package.swift is available.
-     * This method parses dependencies from the Package.resolved file.
+     * Resolves dependencies when only a lockfile aka. `Package.Resolved` is available. This commonly applies to e.g.
```
sschuberth, line 92: "aka." -> "aka"

plugins/package-managers/spm/src/main/kotlin/Spm.kt
```diff
@@ -84,20 +84,20 @@ class Spm(
...
         }
     }
 
     /**
      * Resolves dependencies when only a lockfile aka. `Package.Resolved` is available. This commonly applies to e.g.
      * Xcode projects which only have a lockfile, but no `Package.swift` file.
      */
-    private fun resolveAppDependencies(definitionFile: File): List<ProjectAnalyzerResult> {
-        val resolved = definitionFile.inputStream().use { json.decodeFromStream<PackageResolved>(it) }
+    private fun resolveLockfileDependencies(packageResolvedFile: File): List<ProjectAnalyzerResult> {
```
sschuberth, line 95: Commit message: The function signature as such is not change, it's still the same amount of parameters of the same type. It's just that the function and parameter are renamed.

plugins/package-managers/spm/src/main/kotlin/Spm.kt
```diff
@@ -83,7 +83,7 @@ class Spm(
         requireLockfile(definitionFile.parentFile) { definitionFile.name != PACKAGE_SWIFT_NAME }
 
         return when (definitionFile.name) {
-            PACKAGE_SWIFT_NAME -> resolveLibraryDependencies(definitionFile)
+            PACKAGE_SWIFT_NAME -> resolveDefinitionFileDependencies(definitionFile)
```
sschuberth, line 86: Some comment as before: This just giving the function and parameter different names.

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
fviernau: For `LibraryDependency` the `id` property is used from elsewhere.
So, doing the exact same thing is not possible. 

The library dependency code of course could be changed as well. 
I didn't want to increase the commit count any further as I think the PR is already large enough.
And my primary concern is the lockfile bits.

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
fviernau: > It might make sense to keep `SpmDependency` as an `interface` to still express that semantic.

Having things independent does make it easier for me to refactor code. Once done, and it still seems the best thing to do to have a common interface, this can also be done on top in no time.

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
sschuberth: > For `LibraryDependency` the `id` property is used from elsewhere.

Yes, but that other place could have simply accessed the `id` via the package. 

> I didn't want to increase the commit count any further as I think the PR is already large enough.

As it follows the same idea, it would have been fine with me to do it in the same commit.

Anyway, I'll probably make a proposal once you're done with your refactoring.

sschuberth (approved):

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
fviernau: Thanks!

sschuberth: > Anyway, I'll probably make a proposal once you're done with your refactoring.

So, @fviernau please let me know once you're done.

fviernau: Will do.

PR #8149, opened 2024-01-22
Rename the `Spm` package manager to `SwiftPM`

### LINKED ISSUE #8134: Rename `SPM` to `SwiftPM`
The term `SPM` isn't used by the [SwiftPM](https://www.swift.org/package-manager) folks.

So, renames should be done like:

1. SPM.kt -> SwiftPM.Kt
2. project.id.type -> SwiftPM
3. package.id.type -> Swift
4. dir: ./plugins/package-managers/spm -> ./plugins/package-managers/swiftpm 
5. doc updates

...look for further analog alignments in the code.

fviernau:
Align the name with upstream terminology.

See individual commits.

Fixes #8134.


plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
```diff
@@ -34,10 +34,10 @@ import org.ossreviewtoolkit.utils.ort.normalizeVcsUrl
 val json = Json { ignoreUnknownKeys = true }
 
 /**
- * The output of the `spm dependencies` command.
+ * The data model for the output of the command `swift package show-dependencies --format json`.
```
sschuberth, line 38: Out of curiosity: Was `spm dependencies` the correct command at some point in time, or was it always wrongly documented?

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
```diff
@@ -92,6 +74,21 @@ data class Pin(
...
+    get() = Identifier(
+        type = PACKAGE_TYPE,
+        namespace = "",
+        name = getCanonicalName(repositoryUrl),
+        version = version
+    )
+
+internal fun LibraryDependency.toPackage(): Package {
+    val vcsInfoFromUrl = VcsHost.parseUrl(repositoryUrl)
+    val vcsInfo = vcsInfoFromUrl.takeUnless { it.revision.isBlank() } ?: vcsInfoFromUrl.copy(revision = version)
```
sschuberth, line 88: Nit: The "While at it" part of the commit message could also mention that the `vcs` property was inlined.

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
```diff
@@ -43,16 +44,16 @@ data class SwiftPackageDependencies(
     val url: String,
     val version: String,
     val path: String,
-    val dependencies: List<LibraryDependency>
-)
-
-@Serializable
-data class LibraryDependency(
```
sschuberth, line 50: Commit message: This is not turning `AppDependency` into a nested class, but `LibraryDependency`.

sschuberth (changes requested):

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
fviernau: When the initial implementation was merged, the command was named `swift`, not `spm`. Do, I believe it's been wrongly documented from the beginning.

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
sschuberth: In the commit message body "Get rid of the misleading term `App`" still mentions "App".

sschuberth: Semi-related: I'm not so happy with the long / nested `org.ossreviewtoolkit.plugins.packagemanagers.spm.SwiftPackageDependencies.Dependency` imports. How about also renaming `SwiftPackageDependencies` to `SwiftPackage` (as it contains more than just the `dependencies` anyway) and then using `SwiftPackage.Dependency` instead of `Dependency`?

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
```diff
@@ -28,6 +28,7 @@ import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.RemoteArtifact
 import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.plugins.packagemanagers.spm.SwiftPackageDependencies.Dependency
```
sschuberth: To me, it seems weird to need this import for the nested class that comes from the same file. See my other comment for a proposed solution.

sschuberth (changes requested):

plugins/package-managers/spm/src/main/kotlin/SpmModel.kt
fviernau: I've amended the commit at the bottom to rename to `SwiftPackage` instead of to `SwiftPackageDependencies`.

sschuberth (dismissed):

sschuberth (approved):

fviernau:
The failing test is unrelated.

PR #8152, opened 2024-01-22
Add an option to prefer ScanCode's file- over line-level findings 

sschuberth:

### DIFF
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
@@ -56,6 +56,13 @@ import org.semver4j.Semver
  * * **"commandLineNonConfig":** Command line options that do not modify the result and should therefore not be
  *   considered in [configuration], like "--processes". Defaults to
  *   [ScanCodeConfig.DEFAULT_COMMAND_LINE_NON_CONFIG_OPTIONS].
+ * * **preferFileLicense**: A flag to indicate whether the "high-level" per-file license reported by ScanCode starting
+ *   with version 32 should be used instead of the individual "low-level" per-line license findings. The per-file
+ *   license may be different from the conjunction of per-line licenses and is supposed to contain fewer
+ *   false-positives. However, no exact line numbers can be associated to the per-file license anymore. If enabled, the
+ *   start line of the per-file license finding is set to the minimum of all start lines for per-line findings in that
+ *   file, the end line is set to the maximum of all end lines for per-line findings in that file, and the score is set
+ *   to the arithmetic average of the scores of all per-line findings in that file.
  */
 class ScanCode internal constructor(
     name: String,
@@ -91,7 +98,14 @@ class ScanCode internal constructor(
             }
         }
 
-    override val configuration by lazy { config.commandLine.joinToString(" ") }
+    override val configuration by lazy {
+        buildList {
+            addAll(config.commandLine)
+
+            // Add this in the style of a fake command line option for consistency with the above.
+            if (config.preferFileLicense) add("--prefer-file-license")
+        }.joinToString(" ")
+    }
 
     override val matcher by lazy { ScannerMatcher.create(details, wrapperConfig.matcherConfig) }
 
@@ -141,7 +155,7 @@ class ScanCode internal constructor(
     }
 
     override fun createSummary(result: String, startTime: Instant, endTime: Instant): ScanSummary =
-        parseResult(result).toScanSummary()
+        parseResult(result).toScanSummary(config.preferFileLicense)
 
     /**
      * Execute ScanCode with the configured arguments to scan the given [path] and produce [resultFile].
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
@@ -69,13 +69,6 @@ class ScanCode internal constructor(
         const val SCANNER_NAME = "ScanCode"
 
         private const val LICENSE_REFERENCES_OPTION_VERSION = "32.0.0"
-        private const val OUTPUT_FORMAT = "json-pp"
-
-        private val OUTPUT_FORMAT_OPTION = if (OUTPUT_FORMAT.startsWith("json")) {
-            "--$OUTPUT_FORMAT"
-        } else {
-            "--output-$OUTPUT_FORMAT"
-        }
     }
 
     class Factory : ScannerWrapperFactory<ScanCodeConfig>(SCANNER_NAME) {
@@ -158,7 +151,7 @@ class ScanCode internal constructor(
             command(),
             *commandLineOptions.toTypedArray(),
             // The output format option needs to directly precede the result file path.
-            OUTPUT_FORMAT_OPTION, resultFile.absolutePath,
+            "--json-pp", resultFile.absolutePath,
             path.absolutePath
         )
 }
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
@@ -97,7 +97,7 @@ class ScanCode internal constructor(
             }
         }
 
-    val commandLineOptions by lazy { getCommandLineOptions(version) }
+    private val commandLineOptions by lazy { getCommandLineOptions(version) }
 
     override fun command(workingDir: File?) =
         listOfNotNull(workingDir, if (Os.isWindows) "scancode.bat" else "scancode").joinToString(File.separator)
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
@@ -23,6 +23,7 @@ import java.io.File
 import java.time.Instant
 
 import kotlin.math.max
+import kotlin.time.Duration.Companion.minutes
 
 import org.apache.logging.log4j.kotlin.logger
 
@@ -72,7 +73,8 @@ class ScanCode internal constructor(
 
         private const val LICENSE_REFERENCES_OPTION_VERSION = "32.0.0"
         private const val OUTPUT_FORMAT = "json-pp"
-        private const val TIMEOUT = 300
+
+        private val TIMEOUT = 5.minutes
 
         /**
          * Configuration options that are relevant for [configuration] because they change the result file.
@@ -82,7 +84,7 @@ class ScanCode internal constructor(
             "--license",
             "--info",
             "--strip-root",
-            "--timeout", TIMEOUT.toString()
+            "--timeout", "${TIMEOUT.inWholeSeconds}"
         )
 
         /**
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
@@ -22,9 +22,6 @@ package org.ossreviewtoolkit.plugins.scanners.scancode
 import java.io.File
 import java.time.Instant
 
-import kotlin.math.max
-import kotlin.time.Duration.Companion.minutes
-
 import org.apache.logging.log4j.kotlin.logger
 
 import org.ossreviewtoolkit.model.ScanSummary
@@ -41,7 +38,6 @@ import org.ossreviewtoolkit.utils.common.Options
 import org.ossreviewtoolkit.utils.common.Os
 import org.ossreviewtoolkit.utils.common.ProcessCapture
 import org.ossreviewtoolkit.utils.common.safeDeleteRecursively
-import org.ossreviewtoolkit.utils.common.splitOnWhitespace
 import org.ossreviewtoolkit.utils.common.withoutPrefix
 import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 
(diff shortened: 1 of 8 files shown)


plugins/scanners/scancode/src/main/kotlin/ScanCodeConfig.kt
```diff
@@ -19,19 +19,47 @@
...
+        private val DEFAULT_TIMEOUT = 300.seconds
+
+        /**
+         * The default list of command line options that might have an impact on the scan results.
+         */
+        private val DEFAULT_COMMAND_LINE_OPTIONS = listOf(
+            "--copyright",
+            "--license",
+            "--info",
+            "--strip-root",
```
fviernau, line 46: out of scope: Shall we move `--strip-root` to `DEFAULT_COMMAND_LINE_NON_CONFIG_OPTIONS` as it does not have impact on the `ScanSummary` ?

plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
```diff
@@ -90,18 +90,15 @@ class ScanCode internal constructor(
...
-            addAll(config.commandLine)
-            add(OUTPUT_FORMAT_OPTION)
-        }.joinToString(" ")
-    }
+    override val configuration by lazy { config.commandLine.joinToString(" ") }
 
     internal fun getCommandLineOptions(version: String) =
         buildList {
             addAll(config.commandLine)
             addAll(config.commandLineNonConfig)
```
fviernau, line 92: I do not understand why output format does not become part of `commandLineNonConfig`. Can you help me there?

plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
```diff
@@ -77,13 +77,7 @@ class ScanCode internal constructor(
         override fun parseConfig(options: Options, secrets: Options) = ScanCodeConfig.create(options)
     }
 
-    override val matcher by lazy { ScannerMatcher.create(details, wrapperConfig.matcherConfig) }
```
fviernau, line 110: commit-msg: "Make the propert**ies** **g**o first" ?

plugins/scanners/scancode/src/test/kotlin/ScanCodeResultParserTest.kt
```diff
@@ -80,9 +80,11 @@ class ScanCodeResultParserTest : FreeSpec({
 
                 val summary = parseResult(resultFile).toScanSummary()
 
-                summary.licenseFindings.find {
-                    it.location == TextLocation("README.md", 100) && it.score == 100.0f
-                }?.license.toString() shouldBe "GPL-2.0-only WITH GCC-exception-2.0"
+                with(summary.licenseFindings) {
+                    shouldHaveSize(18)
```
fviernau, line 85: commit-msg: Please provide a rationale. E.g. during scancode upgrade this test would fail because the findings changed to 19, I would not know what the 18 is about.

fviernau (changes requested):

plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
sschuberth: Good point, it should be, but it was more difficult before as `OUTPUT_FORMAT_OPTION` was private to `ScanCode`. Will add another commit after the inlining of that variable to do that, which avoid several rebasing conflicts for me.

plugins/scanners/scancode/src/main/kotlin/ScanCodeConfig.kt
sschuberth: I agree, done in a new commit.

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModelMapper.kt
```diff
@@ -91,19 +91,31 @@ fun ScanCodeResult.toScanSummary(): ScanSummary {
...
                 location = TextLocation(
                     path = file.path,
-                    startLine = license.startLine,
-                    endLine = license.endLine
+                    startLine = licenses.minOf { it.startLine },
+                    endLine = licenses.maxOf { it.endLine }
                 ),
-                score = license.score
+                score = licenses.map { it.score }.average().toFloat()
             )
```
sschuberth, line 103: FYI @Etsija, @lamppu, @mmurto: This should resemble DOS's logic.

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModelMapper.kt
```diff
@@ -91,19 +91,31 @@ fun ScanCodeResult.toScanSummary(): ScanSummary {
...
+                license = file.detectedLicenseExpressionSpdx,
                 location = TextLocation(
                     path = file.path,
-                    startLine = license.startLine,
-                    endLine = license.endLine
+                    startLine = licenses.minOf { it.startLine },
+                    endLine = licenses.maxOf { it.endLine }
                 ),
-                score = license.score
+                score = licenses.map { it.score }.average().toFloat()
```
fviernau, line 102: Would you know the rational DOS use when coming up with an `avg` score? 
Also, I wonder in what way the CLI option `--license-score` influences the per file license? (and whether this should become a command line config option).

plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
```diff
@@ -90,7 +97,14 @@ class ScanCode internal constructor(
             }
         }
 
-    override val configuration by lazy { config.commandLine.joinToString(" ") }
+    override val configuration by lazy {
```
fviernau, line 100: Adding the per file findings as entries into the text location based license findings which correspond to a line range an a specific file, also implies that the existing license finding curation mechanism becomes applicable to these per file entries. 

So, from the perspective of creating the license finding curations, I wonder about the implication such as would we accept (e.g. in ort-config) also license finding curations which match such per file license findings?

fviernau:
Can you help me understand whether this PR changes the "cache key" of ORT scan results? So that stored scan results would not match anymore with the new ORT version? (I have in mind changes like moving the `strip-root` option)

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModelMapper.kt
sschuberth: > Would you know the rational DOS use when coming up with an `avg` score?

@Etsija and @lamppu probably know better, but the average is pretty much the only metric that somewhat makes sense and is reasonably easy to calculate. Min. does not make sense, max. does not make sense.

Of course, you could do something "crazy" like only taking the scores of those line findings into account whose licenses are also contained in the file finding, and maybe even weight the score according to the number of lines matched, but that's much more work for questionable benefit.

> I wonder in what way the CLI option --license-score influences the per file license?

Let's ask the author's: @pombredanne, does ScanCode's [`--license-score` option](https://scancode-toolkit.readthedocs.io/en/stable/cli-reference/basic-options.html#license-score-option) have an (implicit) impact on which licenses will be part of the per-file `detected_license_expression(_spdx)` fields?

plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
sschuberth: > would we accept (e.g. in ort-config) also license finding curations which match such per file license findings?

I think we should, mostly for two reasons:
- there's hardly any negative impact compared to not doing so,
- we should also accept curations for other supported scanners than ScanCode, and these other scanners *could* theoretically find exactly the file-level finding that ScanCode finds. If we'd accept such a curation for another scanner, we should accept it for ScanCode, too. (This somewhat reminds me of @mnonnenmacher's remark that our license finding curations should also include the scanner name and version.)

sschuberth:
> Can you help me understand whether this PR changes the "cache key" of ORT scan results?

First of all, this PR does not change any behavior of ORT when merged, as the option is disabled by default (`String?.toBoolean()` only returns `true` if the string is "true", and returns `false` in all other cases, including `null`).

Secondly, as the "fake" `--prefer-file-license` option becomes part of the scanner configuration if `ort.scanner.config.ScanCode.options.preferFileLicense` is `true`, then yes, the "cache key" changes, and results will not be "mixed up" compared to not using that options. Phrased differently, enabling that option for the first time causes rescans.

plugins/scanners/scancode/src/test/kotlin/ScanCodeResultParserTest.kt
sschuberth: Note that we're parsing a committed test asset here, so the number of findings would not change unless we also change the committed result file.

fviernau:
> Ah, right, that would trigger rescans then, too. So I guess we should drop it, even if basically it's the right thing to do?

I'm not sure which  option we'd choose. An alternative could be to migrate the scan result storage entries somehow.

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModelMapper.kt
fviernau: > Of course, you could do something "crazy" like only taking the scores of those line findings into account whose licenses are also contained in the file finding, and maybe even weight the score according to the number of lines matched, but that's much more work for questionable benefit.

Another alternative would be to omit `lines` and `score` entirely, with the rational that they stop being meaningful.
The reason I'm picking this up is, because (1) I suspect these values are a bit more likely to exhibit some bit of flakyness at least across scancode version (2) the result of the heuristic gets backed into the stored scan results and is thus very expensive to change.

sschuberth:
> > Ah, right, that would trigger rescans then, too. So I guess we should drop it, even if basically it's the right thing to do?
> 
> I'm not sure which option we'd choose. An alternative could be to migrate the scan result storage entries somehow.

I've anyway dropped the last `fix(scancode): Move --strip-root to non-config options` commit now to not make it delay the merge of this PR further.

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModelMapper.kt
sschuberth: > Another alternative would be to omit `lines` and `score` entirely, with the rational that they stop being meaningful.

Making lines optional would be a bigger change in the data model, and it would prevent us from matching copyright to license findings, and probably have more implications.

However, I forgot that the `score`actually is optional already. Would you prefer to omit it instead of calculating the average?

> I suspect these values are a bit more likely to exhibit some bit of flakyness at least across scancode version

More likely to be flaky than what, than the individual per-line findings? As the lines and score are calculated from the per-line findings, they can only be flaky if the per-line findings themselves are already flaky.

> the result of the heuristic gets backed into the stored scan results and is thus very expensive to change.

Correct, but that's the way forward how ScanCode will operate, I guess. As [this discussion](https://github.com/nexB/scancode-toolkit/issues/3458) shows, there seem to be no endeavors to enhance the quality of the per-line findings themselves, but to achieve better quality by post-processing these findings into the per-file finding.

fviernau: > Correct, but that's the way forward how ScanCode will operate, I guess. As [this discussion](https://github.com/nexB/scancode-toolkit/issues/3458) shows, there seem to be no endeavors to enhance the quality of the per-line findings themselves, but to achieve better quality by post-processing these findings into the per-file finding.

I don't much about the processing inside of scancode, so I don't want to move this discussion off-topic,
but it's already interesting that this is actually a post-processing step of per-line findings.
If the only input of this post processing step is the a per-line findings, then it could
also / in the long run make more sense to have that postprocessing step applied by ORT (using ScanCodes logic), not by ScanCode (at least not in a single step) in order to make use of the per-line finding curations. The process could be:

1. Compute per line findings with ScanCode
2. Apply per line finding curations aka license finding curations
3. Take result from 2. as input for the post processing to compute the per-file findings using ScanCode's heuristic 
     feeding in the curated license findings.

I believe this would exhibit several advantages over the current solution. 

@pombredanne - what do you think about this idea?

sschuberth: > make more sense to have that postprocessing step applied by ORT (using ScanCodes logic)

I agree, *if* that's possible, and ScanCode does not require internal data for this that's not written to the result file.

sschuberth: > However, I forgot that the `score`actually is optional already. Would you prefer to omit it instead of calculating the average?

@fviernau I still need you input here. I believe that's the last thing to clarify before we could merge?

fviernau: I'd have slight tendency to omit it, but I also believe my knowledge is too limited in that area to make the call.
The reason I believe why its better to omit it, is because I believe the way the score would be defined would even harder to grasp and hardly be actionable. E.g. if an SPDX expression would consist of several 100% findings and a  few 80% findings, the final score would have a relatively low score even though some parts of the expression apply a 100%. So,e.g. filtering out on score would not be feasible anymore, because one would have two different ways the score could have been computed.

So, all in all I prefer to omit it, but would be fine if you kept it. Up to you.

fviernau: @mnonnenmacher what do you think about https://github.com/oss-review-toolkit/ort/pull/8152#discussion_r1464745604 ?

sschuberth:
@oss-review-toolkit/kotlin-devs as there was no feedback on the topic of settings the `score` to `null` or keeping it at the average, I prefer to merge this as-is with the average to resemble what the Double Open Server (DOS) currently does.

fviernau (approved):

PR #8154, opened 2024-01-22
SwiftPM: Fix test flakyness and further issues

fviernau:


plugins/package-managers/swiftpm/src/funTest/assets/projects/synthetic/spm-lib-no-lockfile/Package.resolved
```diff
@@ -0,0 +1,187 @@
+{
```
sschuberth, line 1: Hmm, the rationale is "to avoid the need for constantly updating the expected results", but won't that still be the case for the copied `spm-lib` project without the lockfile?

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -76,7 +76,10 @@ class SwiftPm(
         PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages())
 
     override fun mapDefinitionFiles(definitionFiles: List<File>): List<File> {
-        return definitionFiles.filterNot { file -> file.path.contains(".build/checkouts") }
+        return definitionFiles.filterNot { file ->
+            file.path.contains(".build/checkouts") ||
```
sschuberth, line 80: Mind changing this to use the `in` operator while at it?

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -80,7 +80,9 @@ class SwiftPm(
     }
 
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
-        requireLockfile(definitionFile.parentFile) { definitionFile.name != PACKAGE_SWIFT_NAME }
+        if (definitionFile.name == PACKAGE_SWIFT_NAME) {
```
sschuberth, line 86: I don't understand why this is necessary. If the `definitionFile.name` is `PACKAGE_RESOLVED_NAME`, then `definitionFile.resolveSibling(PACKAGE_RESOLVED_NAME)` points to itself and the `isFile` check succeeds, and no `require()` check is triggered.

As a side node, IMO the name `requireLockfile()` is misleading / wrong, as the function does not require a lockfile to be present, but only `allowDynamicVersions || condition()` to be true. The function should probably be renamed, but I don't have a good proposal right now to what.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -76,7 +76,10 @@ class SwiftPm(
         PackageManagerResult(projectResults, graphBuilder.build(), graphBuilder.packages())
 
     override fun mapDefinitionFiles(definitionFiles: List<File>): List<File> {
-        return definitionFiles.filterNot { file -> file.path.contains(".build/checkouts") }
+        return definitionFiles.filterNot { file ->
+            file.path.contains(".build/checkouts") ||
+                (file.name == PACKAGE_RESOLVED_NAME && file.resolveSibling(PACKAGE_SWIFT_NAME).isFile)
```
sschuberth, line 81: Shouldn't it be the opposite? Like if both `PACKAGE_SWIFT_NAME` and `PACKAGE_RESOLVED_NAME` (i.e. the lockfile for `PACKAGE_SWIFT_NAME`) and present, we should prefer / only use `PACKAGE_RESOLVED_NAME` as it provides the stable versions?

I mean, if you'd run `swiftpm` in such a directory, if would also install the dependencies from `PACKAGE_RESOLVED_NAME` even if there are later version for the dependencies in `PACKAGE_SWIFT_NAME`. At least that's my understanding.

sschuberth (changes requested):

plugins/package-managers/swiftpm/src/funTest/assets/projects/synthetic/spm-lib-no-lockfile/Package.resolved
fviernau: For the copied project, the test only verifies that the analyzer rejects analysis because `allowDynamicVersions = false`.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
fviernau: > I don't understand why this is necessary. If the `definitionFile.name` is `PACKAGE_RESOLVED_NAME`, then `definitionFile.resolveSibling(PACKAGE_RESOLVED_NAME)` points to itself and the `isFile` check succeeds, and no `require()` check is triggered.

That's correct behavior then, because `Package.resolved` is the lockfile.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
fviernau: > Shouldn't it be the opposite? Like if both `PACKAGE_SWIFT_NAME` and `PACKAGE_RESOLVED_NAME` (i.e. the lockfile for `PACKAGE_SWIFT_NAME`) and present, we should prefer / only use `PACKAGE_RESOLVED_NAME` as it provides the stable versions?

I think it's better in this case to run `switch packages show-dependencies..` because that command does adhere to the lockfile, but it gives more information: it has the edges of the dependency tree.

> I mean, if you'd run swiftpm in such a directory, if would also install the dependencies from PACKAGE_RESOLVED_NAME even if there are later version for the dependencies in PACKAGE_SWIFT_NAME. At least that's my understanding.

I think the assumption (similar to other package managers) should be that the lockfile `Package.resolved` corresponds to the sibling definition file `Package.swift` and just pins / locks down the dependency versions. 

E.g. analog in NPM world, if there is `package-lock.json` commited, then resolution of deps for `package.son` does adhere to the versions present in the lockfile.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
sschuberth: > That's correct behavior then

I'm not trying to say the new code introduces incorrect behavior, but I was trying to explain why I believe the introduced check to be superfluous. So, do you agree the added check is not strictly necessary? That's still not clear to me.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
fviernau: So, no change needed for addressing this comment, right?

sschuberth: Right.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
fviernau: > I'm not trying to say that the new code introduces incorrect behavior, but I was trying to explain why I believe the introduced check to be superfluous.

I need to check this by running ORT. From reading the code, I assumed that when both files are present, the SwiftPm.kt would run the analysis twice: once for the code path for the lockfile and once based on `swift package show-dependencies`. (If you already know why this isn't the case, then let me know)

fviernau: You were right. I've dropped the commit. Good catch!

sschuberth: Hmm, it now seems to me that you've dropped the wrong commit? This discussion belongs to the commit that adds the `if (definitionFile.name == PACKAGE_SWIFT_NAME)` check. You should be able to drop that as well.

fviernau: Hm. If we have the situation that `Package.swift` exists, and `allowDynamicVersions=false`, then we must do a check whether the `lockfile` exists, and fail in case it doesn't. This is the reason for the existance of the check. So, why can it be removed?

sschuberth: > If the definition file is `Package.resolved` then of course no check is necessary, because we know that a lockfile exists as the definition file is a lockfile.

Right, the check is not *necessary* in that case, but it also does not hurt to have it.

> So, why is the check entirely unnecessary in your view?

Like I tried to explain [above](https://github.com/oss-review-toolkit/ort/pull/8154#discussion_r1462040805) it's simply not strictly required. It's maybe an optimization, but at the cost of added complexity and reduced readability. The original code would have worked just as well.

Maybe we should have a Huddle to do through this PR?

PR #8159, opened 2024-01-23
test(spm): Update expected results

sschuberth:


sschuberth:
I'm aware that https://github.com/oss-review-toolkit/ort/pull/8154 is in progress, but in case the review takes more time, here's another intermediate SPM test fix.

fviernau (approved):

PR #8161, opened 2024-01-23
SwiftPM: Support lockfile format version 2

### LINKED ISSUE #8123: SPM: Implement support for `PackageResolved` file format version 2

fviernau:
See individual commits.

Fixes #8123.


plugins/package-managers/swiftpm/src/funTest/kotlin/SwiftPmFunTest.kt
```diff
@@ -31,30 +31,34 @@ import org.ossreviewtoolkit.utils.test.getAssetFile
 import org.ossreviewtoolkit.utils.test.matchExpectedResult
 
 class SwiftPmFunTest : WordSpec({
-    "Parsing 'Package.resolved' dependencies" should {
+    "Analyzing a lockfile with file format version 1" should {
         "return the correct result" {
-            val definitionFile = getAssetFile("projects/synthetic/spm-app/Package.resolved")
-            val expectedResultFile = getAssetFile("projects/synthetic/spm-expected-output-app.yml")
+            val definitionFile = getAssetFile("projects/synthetic/lockfile-v1/Package.resolved")
```
sschuberth, line 36: Nit: To isolate this even more from "project-with-lockfile", maybe call this "lockfile-v1-*only*" (or "*only*-lockfile-v1")?

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -159,7 +172,21 @@ class SwiftPm(
     }
 }
 
-private fun parseLockfile(packageResolvedFile: File): Set<Package> {
-    val resolved = packageResolvedFile.inputStream().use { json.decodeFromStream<PackageResolved>(it) }
-    return resolved.objects["pins"].orEmpty().mapTo(mutableSetOf()) { it.toPackage() }
+private fun parseLockfile(packageResolvedFile: File): Result<Set<Package>> {
+    val root: JsonObject = packageResolvedFile.inputStream().use { json.decodeFromStream(it) }
```
sschuberth, line 176: Better use

    val root = json.parseToJsonElement(packageResolvedFile.readText()).jsonObject

to read into an "untyped" `JsonObject`.

sschuberth: To also guard against JSON parsing errors, I propose to wrap the whole function body in `runCatching {}`, like

```
private fun parseLockfile(packageResolvedFile: File): Result<Set<Package>> =
    runCatching {
        val root = json.parseToJsonElement(packageResolvedFile.readText()).jsonObject

        when (val version = root.getValue("version").jsonPrimitive.content) {
            "1" -> {
                val pinsJson = root.getValue("object").jsonObject.getValue("pins")
                val pins = json.decodeFromJsonElement<List<PinV1>>(pinsJson)
                pins.mapTo(mutableSetOf()) { it.toPackage() }
            }

            else -> {
                throw IOException(
                    "Could not parse lockfile '${packageResolvedFile.invariantSeparatorsPath}'. Unknown file format " +
                        "version '$version'."
                )
            }
        }
    }
```

My code also already moves `version` to `when` to address an inspection hint that otherwise would be triggered by the next commit. And I've inlined the `messages` variable.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -95,13 +100,21 @@ class SwiftPm(
      * Resolves dependencies when only a lockfile aka `Package.Resolved` is available. This commonly applies to e.g.
      * Xcode projects which only have a lockfile, but no `Package.swift` file.
      */
-    private fun resolveLockfileDependencies(packageResolvedFile: File): List<ProjectAnalyzerResult> =
-        listOf(
+    private fun resolveLockfileDependencies(packageResolvedFile: File): List<ProjectAnalyzerResult> {
+        val issues = mutableListOf<Issue>()
+
+        val packages = parseLockfile(packageResolvedFile).onFailure {
+            issues += Issue(source = managerName, message = it.message!!)
```
sschuberth, line 107: With my proposal below, this might catch more exceptions that just our own, so we cannot guarantee that `message`is always non-`null`, and should probably use `.orEmpty()` instead.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
```diff
@@ -70,6 +71,23 @@ data class PinV1(
     )
 }
 
+/**
+ * See https://github.com/apple/swift-package-manager/blob/3ef830dddff459e569d6e49c186c3ded33c39bcc/Sources/PackageGraph/PinsStore.swift#L387-L462.
+ */
+@Serializable
+data class PinV2(
+    val identity: String,
+    val state: PinV1.State?,
```
sschuberth, line 80: Shouldn't this be `(PinV2.)State`? Otherwise that new state class would be unused.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
```diff
@@ -70,6 +71,23 @@ data class PinV1(
...
+    val location: String
+) {
+    @Serializable
+    data class State(
+        val version: String? = null,
+        val revision: String? = null,
+        val branch: String? = null
+    )
```
sschuberth, line 88: Remove this if indeed `PinV1.State` should be used. Also, consider unnested the state class from `PinV1` then to better explain that it's shared.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -159,7 +172,21 @@ class SwiftPm(
...
-    val resolved = packageResolvedFile.inputStream().use { json.decodeFromStream<PackageResolved>(it) }
-    return resolved.objects["pins"].orEmpty().mapTo(mutableSetOf()) { it.toPackage() }
+private fun parseLockfile(packageResolvedFile: File): Result<Set<Package>> {
+    val root: JsonObject = packageResolvedFile.inputStream().use { json.decodeFromStream(it) }
+    val version = root.getValue("version").jsonPrimitive.content
+
+    return when {
+        version == "1" -> {
+            val pinsJson = root["object"]?.jsonObject?.get("pins")?.toString() ?: "{}"
+            val pins = json.decodeFromString<List<PinV1>>(pinsJson)
```
sschuberth, line 182: There's no need to convert back to a string and forth to a type again, just use `decodeFromJsonElement()`. I'll add this to my code proposal above.

plugins/package-managers/swiftpm/src/funTest/assets/projects/synthetic/lockfile-v2/Package.resolved
```diff
@@ -0,0 +1,14 @@
...
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/Alamofire/Alamofire.git",
+      "state" : {
+        "revision" : "d120af1e8638c7da36c8481fd61a66c0c08dc4fc",
+        "version" : "5.4.4"
+      }
+    }
+  ],
+  "version" : 2
+}
```
sschuberth, line 14: Nit: Missing final newline.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -183,6 +183,12 @@ private fun parseLockfile(packageResolvedFile: File): Result<Set<Package>> {
             val packages = pins.mapTo(mutableSetOf()) { it.toPackage() }
             Result.success(packages)
         }
+        version == "2" -> {
+            val pinsJson = root["pins"]?.toString() ?: "{}"
+            val pins = json.decodeFromString<List<PinV2>>(pinsJson)
```
sschuberth, line 188: Same hint about using `decodeFromJsonElement()` applies here.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
```diff
@@ -57,6 +57,7 @@ data class SwiftPackage(
 /**
  * See https://github.com/apple/swift-package-manager/blob/3ef830dddff459e569d6e49c186c3ded33c39bcc/Sources/PackageGraph/PinsStore.swift#L285-L384.
  */
+@Serializable
```
sschuberth, line 60: This needs to go to the previous commit.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
```diff
@@ -114,6 +132,35 @@ internal fun PinV1.toPackage(): Package {
...
+    val vcsInfoFromUrl = VcsHost.parseUrl(location)
+    val vcsInfo = if (vcsInfoFromUrl.revision.isBlank() && state != null) {
+        when {
+            !state.revision.isNullOrBlank() -> vcsInfoFromUrl.copy(revision = state.revision)
+            !state.version.isNullOrBlank() -> vcsInfoFromUrl.copy(revision = state.version)
+            else -> vcsInfoFromUrl
+        }
+    } else {
+        vcsInfoFromUrl
+    }
```
sschuberth, line 159: Any plans to reduce the code duplication here with `PinV1.toPackage()`?

sschuberth (changes requested):

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
fviernau: I've copied your approach and only removed the assumptions that the entries for the key "object" and "pins" are always present, as I haven't found proof for this to be true.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
fviernau: I've changed it according to above comment to use PinV2.State to keep things independent.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
fviernau: Nothing concrete yet. I thought it could make sense to first look into version 3 format, and to see how the contract is for the properties of `state` and also considering `kind` property before using a common path.

plugins/package-managers/swiftpm/src/funTest/kotlin/SwiftPmFunTest.kt
fviernau: I'd do this with a new commit (PR) on top, to avoid some hazzle with updating expected file, if this is ok for you.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -95,13 +99,21 @@ class SwiftPm(
...
      * Xcode projects which only have a lockfile, but no `Package.swift` file.
      */
-    private fun resolveLockfileDependencies(packageResolvedFile: File): List<ProjectAnalyzerResult> =
-        listOf(
+    private fun resolveLockfileDependencies(packageResolvedFile: File): List<ProjectAnalyzerResult> {
+        val issues = mutableListOf<Issue>()
+
+        val packages = parseLockfile(packageResolvedFile).onFailure {
+            issues += Issue(source = managerName, message = it.message.orEmpty())
+        }.getOrNull() ?: emptySet()
```
sschuberth, line 107: Nicer: `.getOrDefault(emptySet())`

sschuberth (approved):
Just an optional code beautification.

sschuberth:
As a general remark, the handling for `PinV1` vs `PinV2` could have also been implemented via polymorphic deserializers, similar to like in the [ScanCodeResultParser](https://github.com/oss-review-toolkit/ort/blob/a0ea68220b985d0e2d1c2f69429ca0966af2ed7d/plugins/scanners/scancode/src/main/kotlin/ScanCodeResultParser.kt#L45). But I guess this was deliberately not done for this small case to reduce complexity.

fviernau:
Merge despite the failing NpmVersionUrlTest which is unrelated.

plugins/package-managers/swiftpm/src/funTest/kotlin/SwiftPmFunTest.kt
fviernau: Addressed by https://github.com/oss-review-toolkit/ort/pull/8165.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
fviernau: Addressed by https://github.com/oss-review-toolkit/ort/pull/8165.

PR #7556, opened 2023-09-21
Fix updating the Git working tree for outdated branches

mnonnenmacher:
Please see the commit messages for details.

PR #8170, opened 2024-01-24
Fix resetting a Git working tree to the latest revision of a branch

sschuberth:

### DIFF
--- a/plugins/version-control-systems/git/src/main/kotlin/Git.kt
+++ b/plugins/version-control-systems/git/src/main/kotlin/Git.kt
@@ -195,17 +195,16 @@ class Git : VersionControlSystem(), CommandLineTool {
 
             // See https://git-scm.com/docs/gitrevisions#_specifying_revisions for how Git resolves ambiguous
             // names. In particular, tag names have higher precedence than branch names.
-            runCatching { fetch.setRefSpecs(revision).call() }
-                .recoverCatching {
-                    // Note that in contrast to branches / heads, Git does not namespace tags per remote.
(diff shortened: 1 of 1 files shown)


plugins/version-control-systems/git/src/main/kotlin/Git.kt
```diff
@@ -230,19 +233,41 @@ class Git : VersionControlSystem(), CommandLineTool {
             } else {
                 workingTree.runGit("fetch", "--tags", "origin")
             }
```
fviernau, line 235: commit-msg: Would it make sense to link [1] (I assume this commit introduced the bug) ?

[1]: https://github.com/oss-review-toolkit/ort/pull/7556/commits/53e755d882106c99beacd35ad4ad7d170a8c45c3

sschuberth: > I assume this commit introduced the bug

I'm actually not sure how guilty that commit really is. So far, running `git reset --hard FETCH_HEAD` in this scenario has always worked for me, and also @nnobelis was only able to reproduce this with a *single* of the hundreds repos they're scanning. So I assume this rather to be a Git server implementation bug, but it could be a JGit bug as well... I simply could find any specs on the ordering of advertised refs.

So that's why I didn't explicitly refer to that commit, as I'm still unsure where the bug lies.

fviernau: Makes sense, thanks for explaining!

fviernau (approved):

PR #8178, opened 2024-01-25
Pub: Parse source artifacts for hosted packages

mnonnenmacher:
Add support for source artifacts for hosted packages. See the commit messages for details.

### DIFF
--- a/plugins/package-managers/pub/src/funTest/assets/projects/synthetic/pub-expected-output-multi-module.yml
+++ b/plugins/package-managers/pub/src/funTest/assets/projects/synthetic/pub-expected-output-multi-module.yml
@@ -38,9 +38,9 @@ projects:
     path: ""
   vcs_processed:
     type: "Git"
-    url: "ssh://git@github.com/oss-review-toolkit/ort.git"
-    revision: "10242bd48f0ebcf37ac3c1d50b69236b144a97e4"
-    path: "plugins/package-managers/pub/src/funTest/assets/projects/synthetic/multi-module/subproject"
+    url: "<REPLACE_URL_PROCESSED>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>/subproject"
   homepage_url: "https://ossreviewtoolkit.org/"
   scopes:
   - name: "dependencies"
--- a/plugins/package-managers/pub/src/funTest/assets/projects/synthetic/pub-expected-output-multi-module.yml
+++ b/plugins/package-managers/pub/src/funTest/assets/projects/synthetic/pub-expected-output-multi-module.yml
@@ -61,10 +61,10 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: ""
+    url: "https://pub.dev/packages/collection/versions/1.16.0.tar.gz"
     hash:
-      value: ""
-      algorithm: ""
+      value: "ef7e3a5529178ce8f37a9d0b11cbbc8b1e025940f9cf9f76c42da6796301219d"
+      algorithm: "SHA-256"
   vcs:
     type: "Git"
     url: "https://github.com/dart-lang/collection.git"
@@ -87,10 +87,10 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: ""
+    url: "https://pub.dev/packages/typed_data/versions/1.3.1.tar.gz"
     hash:
-      value: ""
-      algorithm: ""
+      value: "26f87ade979c47a150c9eaab93ccd2bebe70a27dc0b4b29517f2904f04eb11a5"
+      algorithm: "SHA-256"
   vcs:
     type: "Git"
     url: "https://github.com/dart-lang/typed_data.git"
--- a/plugins/package-managers/pub/src/main/kotlin/Pub.kt
+++ b/plugins/package-managers/pub/src/main/kotlin/Pub.kt
@@ -532,8 +532,10 @@ class Pub(
                     var vcs = VcsInfo.EMPTY
                     var authors = emptySet<String>()
 
+                    val source = pkgInfoFromLockFile["source"].textValueOrEmpty()
+
                     when {
-                        pkgInfoFromLockFile["source"].textValueOrEmpty() == "path" -> {
+                        source == "path" -> {
                             rawName = packageName
                             val path = pkgInfoFromLockFile["description"]["path"].textValueOrEmpty()
                             vcs = VersionControlSystem.forDirectory(workingDir.resolve(path))?.getInfo() ?: run {
@@ -545,7 +547,7 @@ class Pub(
                             }
                         }
 
-                        pkgInfoFromLockFile["source"].textValueOrEmpty() == "git" -> {
+                        source == "git" -> {
                             val pkgInfoFromYamlFile = readPackageInfoFromCache(pkgInfoFromLockFile, workingDir)
 
                             rawName = pkgInfoFromYamlFile["name"]?.textValue() ?: packageName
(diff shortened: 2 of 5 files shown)


plugins/package-managers/pub/src/main/kotlin/Pub.kt
```diff
@@ -532,8 +532,10 @@ class Pub(
                     var vcs = VcsInfo.EMPTY
                     var authors = emptySet<String>()
 
+                    val source = pkgInfoFromLockFile["source"].textValueOrEmpty()
```
sschuberth, line 537: How about doing the same for `pkgInfoFromLockFile["description"].textValueOrEmpty()`, too, while at it?

plugins/package-managers/pub/src/funTest/assets/projects/synthetic/pub-expected-output-multi-module.yml
```diff
@@ -61,10 +61,10 @@ packages:
       value: ""
       algorithm: ""
   source_artifact:
-    url: ""
+    url: "https://pub.dev/packages/collection/versions/1.16.0.tar.gz"
```
sschuberth, line 64: Could you explain in the commit message what a "hosted packages" is? I.e., how does it compare to "non-hosted (regular?) packages"? If there is something like the latter, are these not supported? And why?

plugins/package-managers/pub/src/main/kotlin/Pub.kt
```diff
@@ -597,6 +599,17 @@ class Pub(
                         logger.warn { "No version information found for package $rawName." }
                     }
 
+                    val hostUrl = pkgInfoFromLockFile["description"]["url"].textValueOrEmpty()
```
fviernau, line 602: Line 588 is written as if `pkgInfoFromLockFile["description"]` could be null. So, I believe line 602 needs to handle this as well. I'm not 100% certain, but would `pkgInfoFromLockFile["description"]["url"]` throw an NPE if there is no value for key `"description"`?

plugins/package-managers/pub/src/main/kotlin/Pub.kt
sschuberth: On a second thought, maybe just extract `val description = pkgInfoFromLockFile["description"]` so you can also make use of it in your last commit.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
```diff
@@ -597,6 +599,17 @@ class Pub(
...
+
+                    val sourceArtifact = if (source == "hosted" && hostUrl.isNotEmpty() && version.isNotEmpty()) {
+                        val sha256 = pkgInfoFromLockFile["description"]["sha256"].textValueOrEmpty()
+                        val hash = if (sha256.isNotEmpty()) Hash(sha256, HashAlgorithm.SHA256) else Hash.NONE
```
sschuberth, line 606: YOu should be able to just use `Hash.create()`. I just realized that its docs are slightly wrong, as for an empty string `Hash.NONE` is returned instead of `Hash.UNKNOWN`.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
```diff
@@ -613,8 +626,7 @@ class Pub(
                         homepageUrl = homepageUrl,
                         // Pub does not create binary artifacts, therefore use any empty artifact.
                         binaryArtifact = RemoteArtifact.EMPTY,
-                        // Pub does not create source artifacts, therefore use any empty artifact.
-                        sourceArtifact = RemoteArtifact.EMPTY,
+                        sourceArtifact = sourceArtifact,
```
sschuberth, line 629: What about the original comment here, was it wrong? Is it worth mentioning that in the commit message?

sschuberth (changes requested):

plugins/package-managers/pub/src/main/kotlin/Pub.kt
sschuberth: I'll make a docs fix for that.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
fviernau: The reason is probably because there is already a `var description`, see line 529.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
sschuberth: See https://github.com/oss-review-toolkit/ort/pull/8179.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
sschuberth: Right... but that could easily be solved by using a different variable name, like `descriptionNode`. Anyway, it was just a proposal.

mnonnenmacher: I think the whole function needs a refactoring, especially the large when condition. But I don't have time to look deeper into it, esp. since the test data is incomplete, e.g. there are no tests for the `host == "git"` case, so I only did the small change that was useful for my change.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
mnonnenmacher: Which line do you mena, line 588 reads just `}`?

plugins/package-managers/pub/src/main/kotlin/Pub.kt
mnonnenmacher: I'm adding support for source artifacts, isn't it obvious that the sentence was wrong?

plugins/package-managers/pub/src/main/kotlin/Pub.kt
fviernau: 588 in old file, 590 in new file.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
sschuberth: The comment said "Pub does not create source artifacts" which to me sounded like Pub does not have the concept of source artifacts as all, so I was surprised that you can add support for them.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
mnonnenmacher: `pkgInfoFromLockfile["description"]` is never null (at least in our test data), but it can be either a node or a text. If it's a text `pkgInfoFromLockfile["description"]["url"]` will return `null` so there is no need for special handling.

plugins/package-managers/pub/src/main/kotlin/Pub.kt
mnonnenmacher: TBH, I have no idea why that sentence was added in the first place, maybe it was wrong all the time, maybe it was correct back in 2019 when it was added.

sschuberth (approved):

plugins/package-managers/pub/src/main/kotlin/Pub.kt
fviernau: But then the `textValueOrEmpty()` should be changed to just `textValue()` shouldn't it?

sschuberth: > If there is really guarantee that it's never null, then the `textValueOrEmpty()` should be changed to just `textValue()` shouldn't it?

To me, ideally any such assumptions are not subtly expressed via `textValueOrEmpty()` vs `textValue()`, but via explicit `checkNotNull()` statements.

mnonnenmacher: `pkgInfoFromLockfile["description"]` cannot be null but `pkgInfoFromLockfile["description"]["url"]` can be, so `textValueOrEmpty()` is required.

PR #8193, opened 2024-01-30
Further NPM CLI issue adjustments

sschuberth:

The second commits is just a proposal, I'm fine to also drop it again.

### DIFF
--- a/plugins/package-managers/node/src/main/kotlin/Npm.kt
+++ b/plugins/package-managers/node/src/main/kotlin/Npm.kt
@@ -572,10 +572,13 @@ open class Npm(
         val lines = process.stderr.lines()
         val issues = mutableListOf<Issue>()
 
+        // Generally forward issues from the NPM CLI to the ORT NPM package manager. Lower the severity of warnings to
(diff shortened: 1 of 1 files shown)


fviernau (approved):
LGTM, I prefer to have another approval as I'm not certain about using "NPM CLI" as `Issue.source`.

fviernau (approved):

mnonnenmacher:
> LGTM, I prefer to have another approval as I'm not certain about using "NPM CLI" as `Issue.source`. Maybe @mnonnenmacher ?

I thought about this and am not sure if I like the approach. The reason is that for a user who does not know the background it's not clear why some issues use "NPM" and others "NPM CLI" as source. Also, if you want to filter NPM issues (e.g. in the workbench) you now have potentially two different values for the source to filter by.
Another approach would be to prefix the messages with an explanation, that would not require any background knowledge, for example "The NPM CLI reported the following warning during dependency resolution: ...".

sschuberth:
> I thought about this and am not sure if I like the approach.

Ok, let me just drop the second commit then to simplify the decision making here.

fviernau (approved):

PR #8200, opened 2024-01-31
build: Only sign when making official releases

sschuberth:
Signing is not set up for JitPack, so simply disable it as these are just snapshot builds.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,6 +40,7 @@ jobs:
         env:
           ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.OSSRH_USERNAME }}
           ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.OSSRH_PASSWORD }}
+          ORG_GRADLE_PROJECT_RELEASE_SIGNING_ENABLED: true
           ORG_GRADLE_PROJECT_signingInMemoryKey: ${{ secrets.SIGNING_KEY }}
           ORG_GRADLE_PROJECT_signingInMemoryKeyPassword: ${{ secrets.SIGNING_PASSWORD }}
         run: ./gradlew publishAndReleaseToMavenCentral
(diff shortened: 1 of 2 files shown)


sschuberth:
The build log should soon be available at https://jitpack.io/com/github/oss-review-toolkit/ort/jitpack-no-sign-SNAPSHOT/build.log.

fviernau (dismissed):

sschuberth:
Unfortunately, this does not work for unknown reasons.

nnobelis:
@sschuberth :

How about we put a condition in `ort-publication-conventions.gradle.kts`
 ```
if (project.property("SIGN_PUBLISHED_ARTIFACTS") == true) {
        signAllPublications()
}
```
Set it to `true` in `gradle.properties` and override it to `false` in `jitpack.yml` ?

sschuberth:
> How about we put a condition in `ort-publication-conventions.gradle.kts`

I don't believe that's going to help as [signing is built into the plugin we use](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/789644bcdac4442a799bb49323105bbbbb39897d/plugin/src/main/kotlin/com/vanniktech/maven/publish/MavenPublishPlugin.kt#L16-L19).

nnobelis:
Yes I saw this code but I am not sure it's relevant, as setting `RELEASE_SIGNING_ENABLED` to `false` didn't disable the signing as expected.

Anyway putting the code I mentionned earlier and 
`./gradlew --no-daemon --stacktrace -x dokkatooGeneratePublicationJavadoc -PSIGN_PUBLISHED_ARTIFACTS=false publishToMavenLocal` in `jitpack.yml` doesn't  trigger the signing error anymore.

sschuberth:
> I am not sure it's relevant, as setting `RELEASE_SIGNING_ENABLED` to `false` didn't disable the signing as expected.

Right, but this is the only place where the plugin calls `signAllPublications()`. So I assume it's always called, and for some reason setting the surrounding condition to `false` does not work.

> doesn't cause the signing error anymore.

That doesn't make any sense to me. Setting `-PSIGN_PUBLISHED_ARTIFACTS=false` should have the same effect as not adding the code you propose at all, which is the same state that we have now, so why should the signing error go away then suddenly?

sschuberth:
> That doesn't make any sense to me.

Ah! We *already* have `signAllPublications()` in *our* code! That's the issue. That should go away completely, as we should rely on the plugin functionality here.

oheger-bosch (approved):

PR #8204, opened 2024-01-31
Support namespace-level package curations from `ort-config`

sschuberth:

### DIFF
--- a/plugins/package-curation-providers/ort-config/src/funTest/kotlin/OrtConfigPackageCurationProviderFunTest.kt
+++ b/plugins/package-curation-providers/ort-config/src/funTest/kotlin/OrtConfigPackageCurationProviderFunTest.kt
@@ -28,7 +28,7 @@ import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
 
 class OrtConfigPackageCurationProviderFunTest : StringSpec({
-    "provider can load curations from the ort-config repository" {
+    "The provider succeeds to return known curations for packages" {
         val azureCore = Identifier("NuGet:Azure:Core:1.22.0")
         val azureCoreAmqp = Identifier("NuGet:Azure.Core:Amqp:1.2.0")
         val packages = createPackagesFromIds(azureCore, azureCoreAmqp)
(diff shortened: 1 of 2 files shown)


plugins/package-curation-providers/file/src/main/kotlin/FilePackageCurationProvider.kt
```diff
@@ -103,7 +103,7 @@ class FilePackageCurationProvider(
                     "The path '$it' does not exist."
                 }
 
-                if (it.isDirectory) FileFormat.findFilesWithKnownExtensions(it) else listOf(it)
+                if (it.isDirectory) FileFormat.findFilesWithKnownExtensions(it).sorted() else listOf(it)
```
fviernau, line 106: We had this topic of ordering vs. added complexity already before, but I do not recall the details.

My preference still is to not assume any ordering but instead require each curation to be correct on its own.
Because for correct curations, the order does not matter. Anyhow, it'd be interested to hear further opinions on this. Also how much of an edge case this is. IMO it'd would be that problematic to not use namespace curations 
if the curation does not apply to all packages in the namespace.

fviernau (changes requested):

plugins/package-configuration-providers/dir/src/main/kotlin/DirPackageConfigurationProvider.kt
```diff
@@ -88,7 +88,7 @@ class DirPackageConfigurationProvider(
                     "The path '$it' does not exist."
                 }
 
-                if (it.isDirectory) FileFormat.findFilesWithKnownExtensions(it) else listOf(it)
+                if (it.isDirectory) FileFormat.findFilesWithKnownExtensions(it).sorted() else listOf(it)
```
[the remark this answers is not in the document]
sschuberth: No. While `_.yml` are conventional with the `ort-config` provider, they could be used with a `file`  provider as well. Phrased differently, when pointing a `file` provider to the local checkout of the `ort-config` provider, you should get the same curations in the same order.

plugins/package-curation-providers/file/src/main/kotlin/FilePackageCurationProvider.kt
sschuberth: > I'm happy to drop the last commit.

Done.

fviernau (approved):

PR #8210, opened 2024-02-01
chore(vulnerable-code): Add more details to an issue message

sschuberth:


plugins/advisors/vulnerable-code/src/main/kotlin/VulnerableCode.kt
```diff
@@ -106,7 +107,11 @@ class VulnerableCode(name: String, config: VulnerableCodeConfiguration) : Advice
                 // issues that are not associated to any package.
                 allVulnerabilities += chunk.associateWith { emptyList() }
 
-                issues += Issue(source = providerName, message = it.collectMessages())
+                issues += Issue(
+                    source = providerName,
+                    message = "Failed to request chunk ${index + 1} of ${chunks.size} (chunk size ${chunk.size}): " +
```
fviernau, line 112: What's the reason for adding these details? e.g. for what purpose could the index and the size of the chunk be used?

sschuberth: It helped me debug https://github.com/nexB/vulnerablecode/issues/1411.

sschuberth: > or at least do not add the chunk size

What's the problem with adding the chunk size?

> 2\. which contains all `purl` values contained in the batch / chunk which failed.

That would have been 91 PURLs in this case, blowing up the log message quite a bit.

In all, I still prefer the current solution.

fviernau: In general: I believe that for debugging it's ok to look into the log output and that not all information needed for debugging needs to go into issue message. It suffices if it points one to the relevant bits of the logs IMO.

> What's the problem with adding the chunk size?

I believe that the chunk size is not sufficient information anyway, so it would not bring much value in debugging,
as one needs to dig deeper anyway. 

> That would have been 91 PURLs in this case, blowing up the log message quite a bit.

Apart from that, my assumption was (maybe I was wrong here) that one of the purls must have cause the
issue. So, I assumed that figuring out the root cause would involve making individual requests for each purl in the chunk, to figure out the problematic one. This is why I proposed to log the PURL''s which to me translates to logging the request which failed.

sschuberth: > Apart from that, my assumption was (maybe I was wrong here) that one of the purls must have cause the
> issue.

While that turned out to be correct, *my* initial assumption was that the timeout is related to a too big chunk size, which is why I was interested in it.

Also, [`OkHttpClientHelper` already logs these kind of errors in general](https://github.com/oss-review-toolkit/ort/blob/505f2a2ca05c72affceb24f9993a859429d2898a/utils/ort/src/main/kotlin/OkHttpClientHelper.kt#L117-L119), so I wanted to add some additional information here to the issue.

Anyway, to move forward, I've changed to code to what you proposed.

fviernau (approved):

PR #8214, opened 2024-02-02
Hash value fixes

fviernau:
Today's failing `ConanFunTest` showed that it can happen that the hash value is uppercase.
Try to make it lowercase in more cases.

### DIFF
--- a/plugins/package-managers/conan/src/main/kotlin/Conan.kt
+++ b/plugins/package-managers/conan/src/main/kotlin/Conan.kt
@@ -32,6 +32,7 @@ import org.ossreviewtoolkit.analyzer.parseAuthorString
 import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.Hash
+import org.ossreviewtoolkit.model.HashAlgorithm
 import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageReference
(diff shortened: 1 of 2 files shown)


plugins/package-managers/conan/src/main/kotlin/Conan.kt
```diff
@@ -415,7 +416,7 @@ class Conan(
             val url = artifactEntry["url"].let { urlNode ->
                 (urlNode.takeIf { it.isTextual } ?: urlNode.first()).textValueOrEmpty()
             }
-            val hash = Hash.create(artifactEntry["sha256"].textValueOrEmpty())
+            val hash = Hash.create(artifactEntry["sha256"].textValueOrEmpty(), HashAlgorithm.SHA256.name)
```
sschuberth, line 419: Yet better: `Hash(artifactEntry["sha256"].textValueOrEmpty(), HashAlgorithm.SHA256)`

sschuberth (changes requested):

sschuberth:
I've created a small "companion" PR: https://github.com/oss-review-toolkit/ort/pull/8217

plugins/package-managers/conan/src/main/kotlin/Conan.kt
sschuberth: No, I didn't. Yet another drawback of us using data classes that get serialized also as interfaces 😞

But I wonder whether it matters: Serialization now always happens lower-case, and comparison already is case-insensitve. So maybe we could even consider dropping your second commit instead? Or at least document in its message that is only as consistency reason, but is not strictly required.

fviernau: If you compare the `Hash` instance with the `equals` function, then my second commit is required. 
So, it does have an effect IMO. What do you think?

sschuberth: Makes sense. Then please remind us about that and add to your second commit's message that lowercasing is not required for verification, but for equality checks of instances.

My "Yet better" can be disregarded then, or changed to `Hash(artifactEntry["sha256"].textValueOrEmpty().lowercase(), HashAlgorithm.SHA256)`.

sschuberth (approved):

PR #8215, opened 2024-02-02
Revert "refactor(scancode): Disregard the output format in scanner configuration

mnonnenmacher:
This reverts commit a8d6171433ac0549ecf880cccb4f4c9be4036416.

While the change itself made sense, it breaks the matching with existing scan results that still contain the `--json-pp` option. This means that all stored `ScanCode` results cannot be used anymore.

As there is no way to fix this in the user configuration, revert the change until a better solution is found.

### DIFF
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
@@ -76,6 +76,7 @@ class ScanCode internal constructor(
         const val SCANNER_NAME = "ScanCode"
 
         private const val LICENSE_REFERENCES_OPTION_VERSION = "32.0.0"
+        private const val OUTPUT_FORMAT_OPTION = "--json-pp"
     }
 
     class Factory : ScannerWrapperFactory<ScanCodeConfig>(SCANNER_NAME) {
@@ -101,6 +102,7 @@ class ScanCode internal constructor(
     override val configuration by lazy {
         buildList {
             addAll(config.commandLine)
+            add(OUTPUT_FORMAT_OPTION)
 
             // Add this in the style of a fake command line option for consistency with the above.
             if (config.preferFileLicense) add("--prefer-file-license")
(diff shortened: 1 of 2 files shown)


oheger-bosch (approved):

fviernau:
Have you considered making a script which migrates the scan results in the storage, so that they match again?

mnonnenmacher:
> Have you considered making a script which migrates the scan results in the storage, so that they match again?

No, but that should be part of the "better solution" mentioned in the commit message.
Making such a script is not super trivial, because:
* The `ScanResultStorage` API does not have a function to delete or update existing results.
* Everyone who already upgraded ORT after change (like us) will have duplicate scan results in the database that only differ in the configuration, so the script needs to avoid unique constraint violations when updating such entries.

fviernau:
> > Have you considered making a script which migrates the scan results in the storage, so that they match again?
> 
> No, but that should be part of the "better solution" mentioned in the commit message. Making such a script is not super trivial, because:
> 
>     * The `ScanResultStorage` API does not have a function to delete or update existing results.
> 
>     * Everyone who already upgraded ORT after the change (like us) will have duplicate scan results in the database that only differ in the configuration, so the script needs to avoid unique constraint violations when updating such entries.

Would it be sufficient to have such migration code directly operate on the data interface instead of through the Storage facade? ...and would it suffice to only have migration for PostGres storage, not for all types of storage ?

mnonnenmacher:
> Would it be sufficient to have such migration code directly operate on the data interface instead of through the Storage facade? 

I would say yes, whatever works.

> ...and would it suffice to only have migration for PostGres storage, not for all types of storage ?

That would be unfair for users that do not use PostgreSQL.

> (I wonder if we could come up with the migration code before the next release tag creation, which could avoid a revert)

To me this is a critical issue that affects potentially all ORT users, so the revert should be merged ASAP. The change itself was not a big deal, so I think there is no time pressure to re-apply it. I would actually make a new release directly after merging this.

mnonnenmacher:
Merged as failing tests were unrelated and all relevant tests passed.

PR #8217, opened 2024-02-02
Small hash-related improvements

sschuberth:

### DIFF
--- a/plugins/package-managers/pub/src/main/kotlin/Pub.kt
+++ b/plugins/package-managers/pub/src/main/kotlin/Pub.kt
@@ -38,6 +38,7 @@ import org.ossreviewtoolkit.downloader.VcsHost
 import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.EMPTY_JSON_NODE
 import org.ossreviewtoolkit.model.Hash
+import org.ossreviewtoolkit.model.HashAlgorithm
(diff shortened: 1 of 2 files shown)


plugins/package-managers/pub/src/main/kotlin/Pub.kt
```diff
@@ -609,7 +610,7 @@ class Pub(
 
                         RemoteArtifact(
                             url = "$hostUrl/packages/$rawName/versions/$version.tar.gz",
-                            hash = Hash.create(sha256)
+                            hash = Hash(sha256, HashAlgorithm.SHA256)
```
fviernau, line 613: After my related PR is merged, using constructor instead of `create()` would by-pass turning the `value` lowercase. At least it was my intention to make `create()` ensure the `value` passed to the constructor is `lowercase`. 

Maybe we can discuss it here in one place: https://github.com/oss-review-toolkit/ort/pull/8214#discussion_r1475838326

fviernau (changes requested):

plugins/package-managers/pub/src/main/kotlin/Pub.kt
fviernau: So, from the linked discussion, I believe this could now just be changed to call the `create()` function, passing again, but pass the algorithm, right?

fviernau (approved):

PR #8218, opened 2024-02-02
feat(vulnerable-code): Make the read timeout configurable

sschuberth:
See [1] for background information. Also [2] is good related read.

[1]: https://github.com/nexB/vulnerablecode/issues/1411
[2]: https://www.baeldung.com/okhttp-timeouts

### DIFF
--- a/model/src/main/resources/reference.yml
+++ b/model/src/main/resources/reference.yml
@@ -152,6 +152,7 @@ ort:
       VulnerableCode:
         options:
           serverUrl: 'http://localhost:8000'
+          readTimeout: 40
         secrets:
           apiKey: 0123456789012345678901234567890123456789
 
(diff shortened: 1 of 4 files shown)


model/src/main/resources/reference.yml
```diff
@@ -152,6 +152,7 @@ ort:
       VulnerableCode:
         options:
           serverUrl: 'http://localhost:8000'
+          readTimeout: 40
```
fviernau, line 155: Maybe `readTimeoutSeconds`?

fviernau (approved):

model/src/main/resources/reference.yml
sschuberth: I was considering `readTimeout*In*Seconds`, but then went for the shorter name as other timeouts that are configurable (FossID, ScanCode) also do not mention the unit.

Issue #3537, opened 2021-01-26
Allow excluding a specific dependency


tsteenbe:
Not all package manager have scopes such as Python where its `requirements.txt` is basically a flat list of dependencies. What if in https://github.com/heremaps/xyz-spaces-python/blob/master/requirements.txt the developer was to exclude [geopandas](https://geopandas.org/) as [PROVIDED_BY](https://github.com/oss-review-toolkit/ort/blob/master/model/src/main/kotlin/config/PathExcludeReason.kt#L63) e.g. be provided by end user or to be installed on the system.

Propose we introduce way for ORT users to exclude a specific package and was thinking of below:

```
---
excludes:
  paths:
  - pattern: "requirements_dev.txt"
    reason: "BUILD_TOOL_OF"
    comment: "Optional dependencies only used for development / testing."
  - pattern: "tests/**"
    reason: "TEST_OF"
    comment: "Directory only used for testing."
  dependencies:
  // Pattern matches the id of the package, ommitting the version number here so it will always match the package 
  - pattern: "PyPI::geopandas"
    reason: "PROVIDED_BY"
    comment: "Packages to be provided at runtime by user."
```

nnobelis:
I think is a good idea, especially if the scanner afterwards does not scan the dependency at all if `--skip-excluded` is set.
See also https://github.com/oss-review-toolkit/ort/pull/3508 even is this is not the same use case.

brueggenthies-ams:
@nnobelis we are currently evaluating ORT, and ran into the same "issue". Would the current architecture support the extension of the excludes mechanic to also work with packages?
If yes, i would look into this.

nnobelis:
@sschuberth @mnonnenmacher  @fviernau  Could you please answer @brueggenthies-ams ?

fviernau:
I believe ORT only reflects what is supported by the respective package manager. E.g. if the package managers have scopes for "test dependencies" then these dependencies really do not go into the released artifact. So, the exclusion mechanism exists on the package manager side, not in ORT, ORT only reflects that. In my view we should really try to stick to that approach, for clarity and simplicity.

For `requirements.txt` there is the option to have a `requirements-*.txt` to define a separate set of dependencies (which could be excluded. But also, it seems that using a `setup.py` with a `install_requires` would solve the outlined issue, see [1].

[1] https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/

sschuberth:
[...]

PR #8219, opened 2024-02-02
feat(jenkins): Add a parameter to skip excluded scopes and paths

sschuberth:
Enable this by default for performance reasons of showcasing ORT. For simplicity, only use a single parameter for all tools as there is barely a need to configure this flag differently per tool.

### DIFF
--- a/integrations/jenkins/Jenkinsfile
+++ b/integrations/jenkins/Jenkinsfile
@@ -97,6 +97,12 @@ pipeline {
             defaultValue: ''
         )
 
+        booleanParam(
+            name: 'SKIP_EXCLUDED',
+            description: 'Enable to skip scopes and paths that have been configured to be excluded, where supported.',
+            defaultValue: true
+        )
+
         string(
             name: 'ENVIRONMENT_VARIABLES',
             description: 'Optional list of comma-separated key=value pairs of environment variables to set.',
@@ -311,7 +317,7 @@ pipeline {
             steps {
                 withCredentials(projectVcsCredentials) {
                     sh '''
-                    ORT_OPTIONS="$LOG_LEVEL"
+                    ORT_OPTIONS="$LOG_LEVEL -P ort.downloader.skipExcluded=$SKIP_EXCLUDED"
 
                     if [ "$STACKTRACE" = "true" ]; then
                         ORT_OPTIONS="$ORT_OPTIONS --stacktrace"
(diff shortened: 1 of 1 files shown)


integrations/jenkins/Jenkinsfile
```diff
@@ -97,6 +97,12 @@ pipeline {
             defaultValue: ''
         )
 
+        booleanParam(
+            name: 'SKIP_EXCLUDED',
+            description: 'Enable to skip scopes and paths that have been excluded via configuration.',
```
fviernau, line 102: Question: One what granularity does the "excluding paths" work? (file vs. project)

If the latter, would the comment be clearer if it said: "Enable to skip excluded scopes and projects" ?

fviernau (changes requested):

integrations/jenkins/Jenkinsfile
sschuberth: > One what granularity does the "excluding paths" work? (file vs. project)

While of the tools are you referring to specifically?

fviernau: The scanner.

sschuberth: > The scanner won't skip excluded paths

For reference, that's https://github.com/oss-review-toolkit/ort/issues/5018 BTW.

> which is why I proposed the alternative comment.

> "Enable to skip excluded scopes, projects and packages." 

As [packages cannot be excluded on their own](https://github.com/oss-review-toolkit/ort/issues/3537) either, this wording has problems as well. I'll think of something else and update the PR.

fviernau: Packages can be excluded by either a scope exclude or by a path exclude matching the project's definition file.

sschuberth: > Packages can be excluded by either a scope exclude or by a path exclude

Exactly. So packages are never excluded explicitly, but only implicitly via scopes or via projects. My goal was to not document this "implictness", but focus on the first-party entities that we can explicitly exclude. I'm open for other wordings that consider this.

fviernau: I believe it's clearer to speak about "what" is excluded (package, project) rather than about "how" its being excluded (scope exclude, path exclude). But I'll leave it up to you.

fviernau (approved):

integrations/jenkins/Jenkinsfile
sschuberth: > I believe it's clearer to speak about "what" is excluded (package, project)

In terms of ORT configuration, we have "*scope* excludes" and "*path* excludes", so to me "scope" and "path" are the "what" here, as those are the entities the user has to deal with.

While you were approving, I came up with

> Enable to skip scopes and paths (and affected packages / projects) that have been configured to be excluded.

Would that be better than the current wording in your eyes?

fviernau: > Would that be better than the current wording in your eyes?

I'd still have similar comments on the new version, but this is really opinionated. I'm fine with your version as well.

fviernau: I don't have an opinion on that one. A bit similar. But let's just do as you prefer, it's too opinionated I guess

sschuberth: I had pushed this [here](https://github.com/oss-review-toolkit/ort/pull/8223) for reference, but decided to close it.

PR #8221, opened 2024-02-02
SwiftPM: Make the model file dedicated for just the model and its parsing

fviernau:
This separates concerns, and ensures we have `SwiftPmModel` only be about the upstream data model and its parsing.
Mapping this to ORT's model is moved out to the package manager implementation, which makes it easier to change that mapping in following changes.

See individual commits.

Part of: #8167.


plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
mnonnenmacher, line 1: Commit message: "reductions"

mnonnenmacher, line 1: Commit message: "refactoring"

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -164,3 +169,72 @@ class SwiftPm(
...
+    get() = Identifier(
+        type = PACKAGE_TYPE,
+        namespace = "",
+        name = getCanonicalName(repositoryUrl),
+        version = version
+    )
+
+internal fun SwiftPackage.Dependency.toPackage(): Package {
+    val vcsInfoFromUrl = VcsHost.parseUrl(repositoryUrl)
+    val vcsInfo = vcsInfoFromUrl.takeUnless { it.revision.isBlank() } ?: vcsInfoFromUrl.copy(revision = version)
```
mnonnenmacher, line 185: Not introduced in this commit, but is there something sepcial in SwiftPM that allows us to set the revision to the version? We don't do this for other package managers and instead rely on the downloader to find a matching tag.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
```diff
@@ -52,6 +52,26 @@ data class SwiftPackage(
     )
 }
 
+/**
+ * See https://github.com/apple/swift-package-manager/blob/3ef830dddff459e569d6e49c186c3ded33c39bcc/Sources/PackageGraph/PinsStore.swift#L387-L462.
+ */
+@Serializable
+data class PinV2(
```
mnonnenmacher, line 59: Not introduced in this commit, but shouldn't these classes be internal instead of public?

mnonnenmacher (changes requested):

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
fviernau: Yes, there is something special. The implementation using `swift package show-dependencies` involves resolving the revisions. These resolved revisions are present in the lockfile. `SwifPM` operates directly on `Git` repository, similar as `GoMod` does. So, for these two package managers, revisions get resolved during building of the dependency tree.

Note: My next PR on top of this, will change the implementation exactly on this topic. I want to always set the resolved revision as revision. I'll do that obtaining these from the lockfile, which either already exists or is created when obtaining the dependencies.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
fviernau: I don't mind if it's public or internal, what do you prefer?
(But I believe the visibility should be consistent with the two `parse*()` functions below.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
mnonnenmacher: Ok, but this doesn't explain why the `version` is used as fallback revision.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPmModel.kt
fviernau: Done in commit on top.

plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
fviernau: True, using the version as fallback is wrong IMO. 
I'll fix this. Next PR will improve setting the version string, and the revision.

mnonnenmacher (approved):

PR #8224, opened 2024-02-02
fix(vulnerable-code): Correct an URL escape fixup case

sschuberth:
This is a fixup for 6ec2a31.


plugins/advisors/vulnerable-code/src/main/kotlin/VulnerableCode.kt
mnonnenmacher, line 1: Commit message nit: "**used** as"

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #8227, opened 2024-02-04
New SPDX expression parser

mnonnenmacher:
Replace the ANTLR parser with a handwritten parser. See the commit messages for details.


utils/spdx/src/main/kotlin/parser/Exceptions.kt
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.utils.spdx.parser
+
+import kotlin.reflect.KClass
```
sschuberth, line 22: Just double-checking as the use of reflection always is suspicious to me in the context to multiplatform: Would thin work with transpiling to JS?

utils/spdx/src/main/kotlin/parser/Exceptions.kt
```diff
@@ -0,0 +1,50 @@
...
+
+import org.ossreviewtoolkit.utils.spdx.SpdxException
+
+/**
+ * An exception to indicate that an [SpdxExpressionLexer] error occurred.
+ */
+class SpdxExpressionLexerException(val char: Char?, val position: Int) :
```
sschuberth, line 29: Why is `char` nullable? What are the semantics if it is `null`? Would "Unexpected character 'null'" then mean that the string ended earlier than expected?

utils/spdx/src/main/kotlin/parser/Exceptions.kt
```diff
@@ -0,0 +1,50 @@
...
+    SpdxException("Unexpected character '$char' at position $position.")
+
+/**
+ * An exception to indicate that an [SpdxExpressionParser] error occurred.
+ */
+class SpdxExpressionParserException(
+    val token: Token?,
```
sschuberth, line 36: Similar question like above. Clarifying this in the class docs could help.

utils/spdx/src/main/kotlin/parser/Exceptions.kt
```diff
@@ -0,0 +1,50 @@
...
+    vararg val expectedTokenTypes: KClass<out Token> = emptyArray()
+) : SpdxException(
+    buildString {
+        append("Unexpected token '$token'")
+
+        if (expectedTokenTypes.size == 1) {
+            append(", expected ${expectedTokenTypes.first().simpleName}")
+        } else if (expectedTokenTypes.size > 1) {
```
sschuberth, line 44: Maybe use `when`?

utils/spdx/src/main/kotlin/parser/Exceptions.kt
```diff
@@ -0,0 +1,50 @@
...
+            append(", expected one of ${expectedTokenTypes.joinToString { it.simpleName.orEmpty() }}")
```
sschuberth, line 45: What if `expectedTokenTypes` is empty? Then this message does not make sense.

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
```diff
@@ -0,0 +1,102 @@
...
+    constructor(input: String) : this(input.asSequence())
+
+    companion object {
+        private val UPPERCASE = 'A'..'Z'
+        private val LOWERCASE = 'a'..'z'
+        private val DIGITS = '0'..'9'
+
+        private fun Char.isUpper() = this in UPPERCASE
+        private fun Char.isLower() = this in LOWERCASE
+        private fun Char.isDigit() = this in DIGITS
```
sschuberth, line 38: Why not use the Kotlin built-in functions?

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private val iterator = input.iterator()
+    private var next = iterator.nextOrNull()
+    private var position = 0
+
+    fun tokens(): Sequence<Token> {
+        var complete = false
+        return generateSequence {
+            if (complete) null else nextToken().also { if (it == null) complete = true }
```
sschuberth, line 49: Once `nextToken()` returned `null` for the first time, won't further calls also (and cheaply) return `null`? So does the "shortcut" via the `complete` variable really pay off, or could we get rid of it?

sschuberth (commented):
Pausing my review at the second commit and file utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt.

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
sschuberth: Also the [docs](https://kotlinlang.org/docs/sequences.html#from-a-function) say "The sequence generation stops when the provided function returns null".

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
```diff
@@ -0,0 +1,102 @@
...
+                    while (next?.isIdentifier() == true) {
+                        cur = consumeChar()
+                        append(cur)
+                    }
+                }
+
+                return when (value) {
+                    "AND" -> Token.AND(start)
+                    "OR" -> Token.OR(start)
+                    "WITH" -> Token.WITH(start)
```
sschuberth, line 81: FYI, lowercase will soon be accepted as well, it seems: https://github.com/spdx/spdx-spec/pull/876

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        return null
+    }
+
+    private fun consumeChar(): Char? {
+        return next?.also {
+            position++
+            next = iterator.nextOrNull()
+        }
+    }
```
sschuberth, line 101: Turn into an expression function?

utils/spdx/src/main/kotlin/parser/Exceptions.kt
mnonnenmacher: Yes, it should: https://kotlinlang.org/docs/js-reflection.html#class-references
Only the `simpleName` property is used.

utils/spdx/src/main/kotlin/parser/Exceptions.kt
mnonnenmacher: Yes it does, it will say for example `Unexpected token 'PLUS(position = 5).'`.

utils/spdx/src/main/kotlin/parser/Exceptions.kt
mnonnenmacher: When actually makes the expression longer in this case as the cases do not fit on a single line, so it requires two levels of nesting.

utils/spdx/src/main/kotlin/parser/Exceptions.kt
mnonnenmacher: I have added an explanation to the class docs.

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
mnonnenmacher: The built-in functions would allow many UTF characters that are not allowed by the SPDX specs, I have added docs.

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
mnonnenmacher: I think this was a left-over from an earlier iteration where I still had an `EOF` token which I have removed to simplify code.

utils/spdx/src/main/kotlin/parser/Exceptions.kt
sschuberth: I was thinking of a "trick" using a dummy case for 0:

    when (expectedTokenTypes.size) {
        0 -> {}
        1 -> append(", expected ${expectedTokenTypes.first().simpleName}")
        else -> append(", expected one of ${expectedTokenTypes.joinToString { it.simpleName.orEmpty() }}")
    }

But maybe that's not better after all.

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
```diff
@@ -0,0 +1,103 @@
...
+                    "WITH" -> Token.WITH(start)
+                    else -> when {
+                        value.startsWith("DocumentRef-") -> Token.DOCUMENTREF(start, position, value)
+                        value.startsWith("LicenseRef-") -> Token.LICENSEREF(start, position, value)
+                        else -> Token.IDENTIFIER(start, position, value)
+                    }
+                }
+            }
+
+            cur?.let { throw SpdxExpressionLexerException(it, position) }
```
sschuberth, line 92: Nit: Maybe `also` is more fitting as the return value of the block does not matter?

sschuberth: Or maybe just use pain `if` to align with line 90 of `SpdxExpressionParser.kt`.

gradle/libs.versions.toml
```diff
@@ -11,7 +11,6 @@ kotlinPlugin = "1.9.22"
 mavenPublishPlugin = "0.27.0"
 versionsPlugin = "0.51.0"
 
-antlr = "4.13.1"
```
sschuberth, line 14: Commit message: Typo in "Relaates".

sschuberth (changes requested):

gradle/libs.versions.toml
sschuberth: And probably no comma before "as".

utils/spdx/src/main/kotlin/parser/SpdxExpressionParser.kt
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * Also, the rules for `license-id` and `license-exception-id` are changed to allow any valid `idstring` as the
+ * [strictness] decides if only the SPDX identifiers are allowed for license and exception ids and therefore these rules
+ * cannot be part of the grammar.
+ *
+ * For backward compatibility with the previously used SPDX expression parser, operators are case-insensitive. This is
+ * also planned for future SPDX versions, see https://github.com/spdx/spdx-spec/pull/876.
+ */
+class SpdxExpressionParser(
+    tokens: Sequence<Token>,
```
fviernau, line 77: Why does the parser accept a `Sequence<Token>`, instead of just a `String`?

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
```diff
@@ -0,0 +1,103 @@
...
+                        append(cur)
+                    }
+                }
+
+                return when (value.uppercase()) {
+                    "AND" -> Token.AND(start)
+                    "OR" -> Token.OR(start)
+                    "WITH" -> Token.WITH(start)
+                    else -> when {
+                        value.startsWith("DocumentRef-") -> Token.DOCUMENTREF(start, position, value)
```
fviernau, line 85: Can SPDX license expressions really contain `DocumentRef-`? Did the previous parser support that, and should we actually support that?

utils/spdx/src/main/kotlin/parser/SpdxExpressionParser.kt
```diff
@@ -0,0 +1,196 @@
...
+ * This parser implements the grammar defined in the
+ * [SPDX specification](https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/):
+ *
+ * ```
+ * license-expression   -> simple-expression | compound-expression
+ * compound-expression  -> simple-expression |
+ *                         simple-expression "WITH" license-exception-id |
+ *                         compound-expression "AND" compound-expression |
+ *                         compound-expression "OR" compound-expression |
+ *                         "(" compound-expression ")" )
```
fviernau, line 42: Line 42 has a closing brace without a corresponding opening brace.

Can the braces be dropped?

utils/spdx/src/test/kotlin/parser/SpdxExpressionLexerTest.kt
```diff
@@ -0,0 +1,255 @@
...
+import io.kotest.matchers.sequences.shouldContainExactly
+import io.kotest.matchers.shouldBe
+
+class SpdxExpressionLexerTest : FunSpec({
+    context("identifiers") {
+        verifyTokens(
+            "a" to sequenceOf(
+                Token.IDENTIFIER(1, 1, "a")
+            ),
+            "a+" to sequenceOf(
```
fviernau, line 35: Would an `or-later` also fit in here? (Not sure if it falls under the category of a token)

utils/spdx/src/main/kotlin/parser/SpdxExpressionParser.kt
```diff
@@ -0,0 +1,196 @@
...
+ * simple-expression    -> license-id | license-id"+" | license-ref
+ * license-ref          -> ["DocumentRef-" idstring ":"] "LicenseRef-" idstring
+ * license-exception-id -> <short form license exception identifier in Annex A.2>
+ * license-id           -> <short form license identifier in Annex A.1>
+ * idstring             -> 1*(ALPHA / DIGIT / "-" / "." )
+ * ```
+ *
+ * For more efficient parsing, this is transformed into the following form which implements the operator precedence as
```
fviernau, line 50: Would you know why this approach is more efficient? (Would it be worth mentioning here briefly)

fviernau (dismissed):

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
sschuberth: Yes, yes, and yes 😃

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
mnonnenmacher: Somehow the type resolved is confused here, for the line 

```kotlin
if (cur != null) throw SpdxExpressionLexerException(cur, position)
```

it says `Condition 'cur != null' is always true` but also `Smart cast to 'Char' is impossible, because 'cur' is a local variable that is captured by a changing closure`. So I went with the current line. Maybe that's a detail we can figure out after the merge.

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
mnonnenmacher: Yes, the specs indeed just call it `ALPHA` in the grammar. I assumed that meant the standard English alphanumeric characters, but it could also many any alphanumeric characters. Anyway, the previous implementation supported the same character set, so I think we could discuss this also after merge.

utils/spdx/src/main/kotlin/parser/Exceptions.kt
mnonnenmacher: Yes, I think the code is well readable the way it is.

utils/spdx/src/main/kotlin/parser/SpdxExpressionParser.kt
```diff
@@ -0,0 +1,196 @@
...
+ * part of the grammar:
+ *
+ * ```
+ * license-expression -> or-expression
+ * or-expression      -> and-expression ( "OR" and-expression ) *
+ * and-expression     -> primary ( "AND" primary ) *
```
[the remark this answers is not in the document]
mnonnenmacher: The `(` is part of the grammar, the line means that `"AND" primary` can repeat 0-∞ times, like in a regular expression.

utils/spdx/src/main/kotlin/parser/SpdxExpressionParser.kt
mnonnenmacher: I have reworded the sentence. It's not about efficient parsing, but about having a grammar that is easy to implement. Having the operator precedence as part of the grammar makes it easy that each line in the grammar matches a function in the parser and no additional logic is required to handle the operator precedence. The original grammar does not say anything about operator precedence, it's only explained in the documentation. Most examples I found for expression parsers use a grammar like this.

utils/spdx/src/main/kotlin/parser/SpdxExpressionLexer.kt
sschuberth: BTW, I believe this is a kind of bug that will go away with the K2 Kotlin compiler.

utils/spdx/src/test/kotlin/SpdxExpressionParserTest.kt
```diff
@@ -176,32 +176,31 @@ class SpdxExpressionParserTest : WordSpec({
                 SpdxExpression.parse("license WITH exception+")
             }
 
-            exception.message shouldBe "extraneous input '+' expecting <EOF>"
+            exception.message shouldBe "Unexpected token 'PLUS(position=23)'."
```
sschuberth, line 179: IMO this reads a bit odd with the position being part of the quoted token. Instead, I'd expect something like "Unexpected PLUS token at position 23." (I'd also not quote 'PLUS' as that's not the literal character, or use '+', similar to line 195.)

But I'd be fine with improving this after the merge. Similar below in line 187.

sschuberth (approved):

utils/spdx/src/test/kotlin/SpdxExpressionParserTest.kt
mnonnenmacher: As discussed in the core-dev meeting I will make a follow-up PR that will improve the error messages.

PR #8233, opened 2024-02-07
SwiftPM: A couple of minor fixes

fviernau:
In context of #8167.


plugins/package-managers/swiftpm/src/main/kotlin/SwiftPm.kt
```diff
@@ -141,7 +141,7 @@ class SwiftPm(
     }
 
     private fun getSwiftPackage(packageSwiftFile: File): SwiftPackage {
-        // TODO: Issues might be thrown into stderr. Parse them and add it them to the result as well.
+        // TODO: Handle errors from standard output.
```
mnonnenmacher, line 144: The original message referred to stderr instead of stdout.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #8242, opened 2024-02-09
Bumping all java versions to 21

ToBeHH:
Bumping all Java versions to 21, especially because Gradle does not work, if the project you want to test is already on v21.

It may be a bit of an overkill to bump *all* Java to v21, but this increases consistency.


mnonnenmacher:
@ToBeHH Changing the default Java version can also break the analysis of older projects, for example when they use a Gradle version that is not compatible with Java 21. Maybe it would be a better solution to publish Docker images for both Java versions. For the artifacts released to Maven Central this is even more problematic because if we build them with Java 21 they cannot be used with any older Java versions anymore. I will create an issue to discuss that topic with the community.

Please note that we are currently having issues with publishing the Docker images, so if you are in urgent need for an image built with Java 21 you would have to build it on your own.

> Will retry with proper signed commit messages

No need to make a new PR for such fixes, just use Git rebase interactive to fixup your commits and force push your branch.

sschuberth:
> Will retry with proper signed commit messages

Also note that commits do not need to be cryptographically signed (`git commit -S`) but just require a Signed-off-by footer (`git commit -s`).

mnonnenmacher:
I have summarized the issue in #8249.

PR #8243, opened 2024-02-09
Updating all Java versions to v21

ToBeHH:
Updating all JAVA versions to Java v21.

The main reason for this is, that Gradle does not work anymore, if you have set Gradle to v21 locally.

Pushing all Java version to 21 might be an overkill, on the other side it increases consistency across the project.


sschuberth:
Sorry @ToBeHH, we can't accept this due to the reasons mentioned by @mnonnenmacher over [here](https://github.com/oss-review-toolkit/ort/pull/8242#issuecomment-1935820113). Also see related PRs https://github.com/oss-review-toolkit/ort/pull/4948 and https://github.com/oss-review-toolkit/ort/pull/6311.

PR #8245, opened 2024-02-09
Adding IconScout simple license

ToBeHH:
Adding the IconScout Simple License, which is granted for using the Unicons icon library.

See https://github.com/Iconscout/unicons for details and the original license.

### DIFF
--- a/plugins/reporters/web-app-template/public/index.html
+++ b/plugins/reporters/web-app-template/public/index.html
@@ -1782,6 +1782,9 @@
             }, {
               "_id" : 260,
               "id" : "ZPL-2.1"
+            }, {
+              "_id" : 261,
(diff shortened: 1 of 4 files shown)


sschuberth:
I'm sorry @ToBeHH, but I have to close this PR as well. The file `utils/spdx/src/main/kotlin/SpdxLicense.kt` is auto-generated via the `generateSpdxLicenseEnum` Gradle task, also see the docs of the `SpdxLicense` enum. As such the file only contains the licenses from https://spdx.org/licenses/ and must not be edited manually.

If you want to add custom / non-SPDX licenses to ORT, please follow the docs at https://oss-review-toolkit.org/ort/docs/configuration/license-texts.

PR #8248, opened 2024-02-09
revert(docker): Disable Arm64 docker build

heliocastro:


.github/actions/ortdocker/action.yml
```diff
@@ -98,7 +98,7 @@ runs:
       uses: docker/build-push-action@v5
       with:
         context: .
-        platforms: linux/amd64,linux/arm64
+        platforms: linux/amd64
```
mnonnenmacher, line 101: This commit should only contain the platform changes, all the formatting changes [should be in separate commits](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md):

> [Make separate commits for logically separate changes.](https://github.com/git/git/blob/e6932248fcb41fb94a0be484050881e03c7eb298/Documentation/SubmittingPatches#L43) For example, pure formatting changes that do not affect software behavior usually do not belong in the same commit as changes to program logic.

mnonnenmacher (dismissed):

.github/actions/ortdocker/action.yml
heliocastro: It was a drat before. I updated and added as a PR now.

heliocastro:
@sschuberth This PR will fix python build as well.

.github/actions/ortdocker/action.yml
```diff
@@ -15,29 +15,29 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
-name: "ORT Docker image"
-description: "Check and create Docker image for ORT components"
-author: "The ORT Project Authors"
+name: 'ORT Docker image'
```
sschuberth, line 18: Please add to the commit message *which* linter was used for this normalization.

.github/actions/ortdocker/action.yml
```diff
@@ -65,9 +65,6 @@ runs:
         result=$(python ./.github/actions/ortdocker/check_image.py)
         echo "result=$result" >> $GITHUB_OUTPUT
 
-    - name: Set up QEMU
```
sschuberth, line 68: Please explain in the commit message *why* this is reverted, i.e. what problem it caused.

sschuberth (changes requested):

.github/actions/ortdocker/action.yml
heliocastro: Done

.github/actions/ortdocker/action.yml
heliocastro: Done

sschuberth: > Used RedHat VSCode YAML Language server.

I'm not sure we want this. Each linter lints differently, and VSCode is not an editor that we officially support for ORT (only IntelliJ IDEA is). Also, it seems a bit arbitrary to only "normalize" the Docker-related YAML files in the `.github` directory, and not all of them.

Personally, I'm for dropping this commit. What do @oss-review-toolkit/core-devs say?

heliocastro: I dropped the linter commit.
Maybe we should agreed with some default with prettier, or similar that can be added in the code base later.

sschuberth (approved):

PR #8258, opened 2024-02-12
chore(mailmap): Update Nicolas' email address

nnobelis:


.mailmap
```diff
@@ -29,7 +29,7 @@ Martin Nonnenmacher <martin.nonnenmacher@bosch.io> <mnonnenmacher@users.noreply.
 Mikko Murto <mikko.murto@doubleopen.org> <mikko.murto@hhpartners.fi>
 Mo Khan <mo.khan@gmail.com>
 Maximilian Huber <maximilian.huber@tngtech.com> <gh@maxhbr.de>
-Nicolas Nobelis <nicolas.nobelis@bosch.com> <nicolas.nobelis@bosch.io>
+Nicolas Nobelis <nicolas.nobelis@bosch.com> <nicolas.nobelis@bosch.com>
```
sschuberth, line 32: There should be no need to list your `.com` email address twice, or? Both Bosch addresses already get consolidated to `Nicolas Nobelis <nicolas.nobelis@bosch.com>`, so it should be good as-is.

sschuberth (changes requested):

.mailmap
nnobelis: Oh sorry Monday morning...

The point is the .io address should not be used anymore. Should I remove the mail or leave as is as you said ?

nnobelis: Is is important regarding past commits ?

sschuberth: > Is is important regarding past commits ?

Yes, so we should keep it.

nnobelis:
Not needed (see comment).

PR #8263, opened 2024-02-12
feat(jenkins): Allow to specify an input path within the repository

sschuberth:
This can be useful when enabling only a single package manager and pointing the analyzer to a single file for testing analysis of a single project.

### DIFF
--- a/integrations/jenkins/Jenkinsfile
+++ b/integrations/jenkins/Jenkinsfile
@@ -130,6 +130,12 @@ pipeline {
             description: 'An optional existing analyzer result file to use instead of running the analyzer.'
         )
 
+        string(
+            name: 'ANALYZER_INPUT_PATH',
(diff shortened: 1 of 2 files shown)


integrations/jenkins/Jenkinsfile
```diff
@@ -130,6 +130,12 @@ pipeline {
             description: 'An optional existing analyzer result file to use instead of running the analyzer.'
         )
 
+        string(
+            name: 'ANALYZER_INPUT_PATH',
+            description: 'The path to analyze, relative to the repository root.',
```
mnonnenmacher, line 135: You could add here that it is possible to set the path to a definition file to analyze only that single project.

mnonnenmacher (changes requested):

fviernau (approved):

mnonnenmacher (approved):

PR #8273, opened 2024-02-14
helper-cli: Add some more scope exclude patterns

fviernau:
See individual commits.

### DIFF
--- a/helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
+++ b/helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
@@ -146,6 +146,16 @@ private fun getScopeExcludesForPackageManager(packageManagerName: String): List<
                 comment = "Packages for static code analysis only."
             ),
             ScopeExclude(
+                pattern = ".*debug.*",
+                reason = ScopeExcludeReason.DEV_DEPENDENCY_OF,
+                comment = "Packages for debug builds."
+            ),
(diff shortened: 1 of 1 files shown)


helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
```diff
@@ -145,6 +145,16 @@ private fun getScopeExcludesForPackageManager(packageManagerName: String): List<
                 reason = ScopeExcludeReason.DEV_DEPENDENCY_OF,
                 comment = "Packages for static code analysis only."
             ),
+            ScopeExclude(
+                pattern = ".*debug.*",
+                reason = ScopeExcludeReason.DEV_DEPENDENCY_OF,
+                comment = "Packages for debug builds."
+            ),
+            ScopeExclude(
+                pattern = ".*Debug.*",
```
sschuberth, line 154: As these are regexes, do you think it makes sense to combine this (and the previous similarly) to `".*[Dd]ebug.*"`?

sschuberth (approved):

helper-cli/src/main/kotlin/commands/repoconfig/GenerateScopeExcludesCommand.kt
fviernau: I believe it's better to not combine them, because it seems that it most of the cases leads to simpler entries in the `ort.yml`.

PR #8277, opened 2024-02-14
Fix some SPDX license processing issues

sschuberth:


utils/spdx/src/main/kotlin/SpdxExpression.kt
mnonnenmacher, line 1: Commit message: "~~any~~ anyway", "is ~~be~~ too strict"

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #8285, opened 2024-02-15
Some code inspection fixes

fviernau:


sschuberth:
Ah, you dropped the Gradle-Inspector commit while I was reviewing. For reference, here's what I was about to write:

Note that until Gradle 8.1 this was nullable, see

https://github.com/gradle/gradle/blob/40ba32cde9d6daf2b92c39376d2758909dd6b813/subprojects/core-api/src/main/java/org/gradle/internal/deprecation/DeprecatableConfiguration.java#L61-L62

I guess I wasn't sure what counts, the Gradle (Tooling API) version *we* use, or the Gradle version that the project to analyze uses, so I kept the `null` check to be on the safe side.

sschuberth (approved):

fviernau:
> Ah, you dropped the Gradle-Inspector commit while I was reviewing. For reference, here's what I was about to write:
> 
> Note that until Gradle 8.1 this was nullable, see
> 
> https://github.com/gradle/gradle/blob/40ba32cde9d6daf2b92c39376d2758909dd6b813/subprojects/core-api/src/main/java/org/gradle/internal/deprecation/DeprecatableConfiguration.java#L61-L62
> 
> I guess I wasn't sure what counts, the Gradle (Tooling API) version _we_ use, or the Gradle version that the project to analyze uses, so I kept the `null` check to be on the safe side.

I dropped it because indeed the value was actually null during (failed) docker funTest execution.

PR #8289, opened 2024-02-15
docs(spdx): Consistently refer to patch-level version 2 of SPDX 2.2

sschuberth:
While at it, simplify the `SpdxDocument` docs to just point at the Markdown specification for version 2.2.2.

### DIFF
--- a/utils/spdx/src/main/kotlin/model/SpdxDocument.kt
+++ b/utils/spdx/src/main/kotlin/model/SpdxDocument.kt
@@ -33,8 +33,7 @@ private const val SPDX_VERSION_MAJOR_MINOR = "SPDX-2.2"
 private val DATA_LICENSE = SpdxLicense.CC0_1_0.id
 
 /**
- * An SPDX document as specified by https://github.com/spdx/spdx-spec/tree/development/v2.2.1/chapters and
- * https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/ in revision 947271b.
+ * An SPDX document as specified by https://spdx.github.io/spdx-spec/v2.2.2/.
  */
 data class SpdxDocument(
     /**
--- a/utils/spdx/src/test/kotlin/model/SpdxDocumentTest.kt
+++ b/utils/spdx/src/test/kotlin/model/SpdxDocumentTest.kt
@@ -33,10 +33,10 @@ import org.ossreviewtoolkit.utils.spdx.SpdxModelMapper
 import org.ossreviewtoolkit.utils.spdx.yamlMapper
 
 /**
- * This test uses the following test assets copied from the SPDX 2.2.1 specification examples.
+ * This test uses the following test assets copied from the SPDX 2.2.2 specification examples.
  *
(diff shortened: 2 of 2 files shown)


utils/spdx/src/main/kotlin/model/SpdxDocument.kt
```diff
@@ -33,8 +33,7 @@ private const val SPDX_VERSION_MAJOR_MINOR = "SPDX-2.2"
 private val DATA_LICENSE = SpdxLicense.CC0_1_0.id
 
 /**
- * An SPDX document as specified by https://github.com/spdx/spdx-spec/tree/development/v2.2.1/chapters and
- * https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/ in revision 947271b.
+ * An SPDX document as specified by https://spdx.github.io/spdx-spec/v2.2.2/.
```
fviernau, line 36: Does the implementation implement 2.2.1 or 2.2.2 ?

sschuberth: AFAIK there is no implementation-wise differences between these versions. Also the `spdxVersion` field we write out is major / minor only, omitting the patch level, see

https://github.com/oss-review-toolkit/ort/blob/8b346d177342b5abd854b0f33a575485d27686c1/utils/spdx/src/main/kotlin/model/SpdxDocument.kt#L48-L51

utils/spdx/src/test/kotlin/model/SpdxDocumentTest.kt
```diff
@@ -33,10 +33,10 @@ import org.ossreviewtoolkit.utils.spdx.SpdxModelMapper
 import org.ossreviewtoolkit.utils.spdx.yamlMapper
 
 /**
- * This test uses the following test assets copied from the SPDX 2.2.1 specification examples.
+ * This test uses the following test assets copied from the SPDX 2.2.2 specification examples.
  *
- * 1. https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXYAMLExample-2.2.spdx.yaml
- * 2. https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXJSONExample-v2.2.spdx.json
+ * 1. https://github.com/spdx/spdx-spec/blob/development/v2.2.2/examples/SPDXYAMLExample-2.2.spdx.yaml
+ * 2. https://github.com/spdx/spdx-spec/blob/development/v2.2.2/examples/SPDXJSONExample-v2.2.spdx.json
```
fviernau, line 39: Have you checked whether these files are identical?

sschuberth: I believe I did check when I was originally doing these (old) changes, but I just did again: On https://github.com/spdx/spdx-spec/compare/v2.2.1...v2.2.2 the "Files changes" tab does not list these files, so there are no changes.

fviernau (approved):

PR #8300, opened 2024-02-18
fix(docker): Check if owner is an organization or user

heliocastro:
GitHub API endpoints differ for organisations and users, and before check_image script assuming organisations only. The side effect is individual users using Ort Docker workflow having all layers rebuilt on every commit that triggers Docker builds.


heliocastro:
Superseded by #8458

PR #8303, opened 2024-02-19
fix(scanner): Catch a `DownloadException` instead of `IOException`

sschuberth:
The `downloadRecursively()` function rethrows `IOException` as `DownloadException`, so the latter needs to be catched here.

### DIFF
--- a/scanner/src/main/kotlin/Scanner.kt
+++ b/scanner/src/main/kotlin/Scanner.kt
@@ -20,7 +20,6 @@
 package org.ossreviewtoolkit.scanner
 
 import java.io.File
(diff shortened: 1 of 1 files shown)


fviernau (approved):

fviernau (changes requested):
Can we auto close 8116 with this commit?

sschuberth:
> Can we auto close 8116 with this commit?

I'm not sure whether this is the only cause, so I left it open for now for the OP to confirm the bug to be resolved, and close manually.

fviernau (approved):

PR #8304, opened 2024-02-19
Dry-run fixes for the dowloader

sschuberth:


downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -425,6 +425,7 @@ enum class VcsHost(
          */
         fun toArchiveDownloadUrl(vcsInfo: VcsInfo): String? {
             val normalizedVcsInfo = vcsInfo.normalize()
+            if (normalizedVcsInfo.revision.isEmpty()) return null
```
fviernau, line 428: This logic is only used in the analyzer, while the code is located in the `downloader`.

The `fix(downloader)` prefix, IMO can easily be misleading one into thinking that this fixes something in the downloader. However, it fixes an issue in the analyzer result. Would you be fine using `fix(analyzer)` as the prefix ?

downloader/src/main/kotlin/VcsHost.kt
```diff
@@ -518,7 +519,7 @@ enum class VcsHost(
      */
     fun toArchiveDownloadUrl(vcsInfo: VcsInfo): String? {
         val normalizedVcsInfo = vcsInfo.normalize()
-        if (!isApplicable(normalizedVcsInfo)) return null
+        if (normalizedVcsInfo.revision.isEmpty() || !isApplicable(normalizedVcsInfo)) return null
```
fviernau, line 522: commit: The callers of both functions have a fallback to `downloadUrl`, from which they created the `VcsInfo` passed as parameter. Have you actually tested this fix, and did this fallback (not) exhibit the same issue?

fviernau (changes requested):

downloader/src/main/kotlin/VcsHost.kt
sschuberth: Well, the `downloader` is a library, and as such it has no knowledge about where its code is being called from. So I believe it's not meaningful to change the scope to the name of the (currently) only caller. But I'll think about adding more detail to the commit message.

downloader/src/main/kotlin/VcsHost.kt
sschuberth: In combination with the next commit, strictly speaking this commit would probably not be required anymore functionality-wise, but it's still much faster to statically check for a valid URL than trying to download from it and running into a 404 error.

So I'll consider swapping commits and making this a `perf` commit.

downloader/src/main/kotlin/VcsHost.kt
sschuberth: > This logic is only used in the analyzer

Actually, that's not true, it's also used with `ort download --dry-run`.

downloader/src/main/kotlin/VcsHost.kt
fviernau: Note: my comments above actually belong to the function in line 462. I mixed them up.

sschuberth: > I believe the question is still open: does the same issue happen for the fallback `downloadUrl` in the callers?

My main point was to change the non-static version of `toArchiveDownloadUrl()`, which is only called at

https://github.com/oss-review-toolkit/ort/blob/25c97903c4ab89ad2adbe1886da2c50754fe480b/downloader/src/main/kotlin/Downloader.kt#L259-L260

where there is no `downloadUrl` fallback. I only change also the static version of `toArchiveDownloadUrl()` for consistency.

fviernau: Ok understood. It can be considered anyhow out of scope of this commit IMHO. However, would you agree that after this change, the mentioned fallback value seems questionable and should be removed - or does it have to be kept for the case where the respective host is not present in the enum?

fviernau (approved):

downloader/src/main/kotlin/VcsHost.kt
sschuberth: > or does it have to be kept for the case where the respective host is not present in the enum?

Yes, that's the actual purpose of the fallback.

PR #8305, opened 2024-02-19
GradleInspector: Ignore artifacts of zero byte size

fviernau:
See individual commits.

### DIFF
--- a/plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
+++ b/plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
@@ -350,15 +350,15 @@ private fun Collection<OrtDependency>.toPackageRefs(
     }
 
 /**
- * Create a [RemoteArtifact] based on the given [pomUrl], [classifier], [extension] and hash [algorithm]. The hash value
- * is retrieved remotely.
+ * Create a [RemoteArtifact] based on the given [pomUrl], [classifier] and [extension]. The hash value is retrieved
+ * remotely.
  */
 private fun createRemoteArtifact(
     pomUrl: String?,
     classifier: String? = null,
-    extension: String? = null,
-    algorithm: String = "sha1"
+    extension: String? = null
 ): RemoteArtifact {
+    val algorithm = "sha1"
     val artifactBaseUrl = pomUrl?.removeSuffix(".pom") ?: return RemoteArtifact.EMPTY
 
     val artifactUrl = buildString {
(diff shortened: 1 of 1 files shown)


plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -87,6 +87,11 @@ private val GRADLE_USER_HOME = Os.env["GRADLE_USER_HOME"]?.let { File(it) } ?: O
  */
 const val OPTION_GRADLE_VERSION = "gradleVersion"
 
+/**
```
sschuberth, line 90: Commit message nits:
- Should we say "Maven" instead of "Gradle" artifacts?
- Typo in "rot" -> "root.

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -87,6 +87,11 @@ private val GRADLE_USER_HOME = Os.env["GRADLE_USER_HOME"]?.let { File(it) } ?: O
...
+ * The sha1 sum for a zero by size file.
+ */
+private const val ZERO_BYTES_FILE_SHA1 = "da39a3ee5e6b4b0d3255bfef95601890afd80709"
```
sschuberth, line 93: We already have this as `EMPTY_PACKAGE_VERIFICATION_CODE`. How about refactoring that first to a constant in `HashAlgorithm`?

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
```diff
@@ -371,7 +376,11 @@ private fun createRemoteArtifact(
     val checksum = okHttpClient.downloadText("$artifactUrl.$algorithm")
         .getOrElse { return RemoteArtifact.EMPTY }
 
-    return RemoteArtifact(artifactUrl, parseChecksum(checksum, algorithm))
+    // Ignore file with zero byte size, because it cannot be a valid archive.
+    val hash = parseChecksum(checksum, algorithm).takeIf { it.value != ZERO_BYTES_FILE_SHA1 }
```
sschuberth, line 380: Nit: Do better document the expected default, consider `takeUnless { it == ZERO_BYTES_FILE_SHA1 }`.

sschuberth (changes requested):

plugins/package-managers/gradle-inspector/src/main/kotlin/GradleInspector.kt
fviernau: `EMPTY_PACKAGE_VERIFICATION_CODE` is in `spdx-utils` and does not know about the `HashAlgorithm` class. What do you think?

sschuberth: It's only used in `UtilsTest` which could add the `model` as a dependency. I'll propose a follow-up refactoring.

sschuberth (approved):

PR #8312, opened 2024-02-20
Simplify storage code

sschuberth:

### DIFF
--- a/utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
+++ b/utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
@@ -80,10 +80,13 @@ class HttpFileStorage(
         }
     }
 
-    override fun exists(path: String): Boolean {
-        val request = Request.Builder()
+    private fun requestBuilder(): Request.Builder =
+        Request.Builder()
             .headers(headers.toHeaders())
             .cacheControl(CacheControl.Builder().maxAge(cacheMaxAgeInSeconds, TimeUnit.SECONDS).build())
+
+    override fun exists(path: String): Boolean {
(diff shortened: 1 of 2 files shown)


utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
```diff
@@ -67,6 +67,12 @@ class HttpFileStorage(
      */
     private val cacheMaxAgeInSeconds: Int = 0
 ) : FileStorage {
+    private val requestBuilder by lazy {
+        Request.Builder()
+            .headers(headers.toHeaders())
+            .cacheControl(CacheControl.Builder().maxAge(cacheMaxAgeInSeconds, TimeUnit.SECONDS).build())
+    }
```
fviernau, line 74: Why is there no risk of re-using a single instance of the builder for building multiple requests?

fviernau (changes requested):

utils/ort/src/main/kotlin/storage/HttpFileStorage.kt
sschuberth: Why should there be any risk? The actual request is only created upon the `build()` call, which still is performed in the individual functions.

fviernau: Because if on user of the builder alters the state, the altered state may apply to the other user?

sschuberth: Ah, right. I was falsely assuming that each method call on the builder would return a new instance instead of modifying the existing one. Thanks, fixed.

PR #8321, opened 2024-02-20
Refactor compound expressions

mnonnenmacher:
Reactor compound expressions to have multiple children, see the commit messages for details.

### DIFF
--- a/utils/spdx/src/main/kotlin/SpdxExpression.kt
+++ b/utils/spdx/src/main/kotlin/SpdxExpression.kt
@@ -201,29 +201,41 @@ sealed class SpdxExpression {
     /**
      * Concatenate [this][SpdxExpression] and [other] using [SpdxOperator.AND].
      */
-    infix fun and(other: SpdxExpression) = SpdxCompoundExpression(this, SpdxOperator.AND, other)
+    infix fun and(other: SpdxExpression) = SpdxCompoundExpression(SpdxOperator.AND, listOf(this, other))
 
     /**
      * Concatenate [this][SpdxExpression] and [other] using [SpdxOperator.OR].
      */
-    infix fun or(other: SpdxExpression) = SpdxCompoundExpression(this, SpdxOperator.OR, other)
+    infix fun or(other: SpdxExpression) = SpdxCompoundExpression(SpdxOperator.OR, listOf(this, other))
 }
 
 /**
- * An SPDX expression compound of a [left] and a [right] expression with an [operator] as defined by version 2.1 of the
- * [SPDX specification, appendix IV][1].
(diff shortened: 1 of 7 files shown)


utils/spdx/src/main/kotlin/SpdxExpression.kt
```diff
@@ -200,29 +200,33 @@ sealed class SpdxExpression {
...
-    val operator: SpdxOperator,
-    val right: SpdxExpression
+    val children: List<SpdxExpression>,
+    val operator: SpdxOperator
 ) : SpdxExpression() {
-    override fun decompose() = left.decompose() + right.decompose()
+    constructor(left: SpdxExpression, operator: SpdxOperator, right: SpdxExpression) :
+        this(listOf(left, right), operator)
+
+    constructor(operator: SpdxOperator, vararg children: SpdxExpression) : this(children.toList(), operator)
```
sschuberth, line 224: How about only having a constructor like this:

```
class SpdxCompoundExpression(
    val operator: SpdxOperator,
    left: SpdxExpression,
    right: SpdxExpression,
    vararg children: SpdxExpression
) : SpdxExpression() {
    val children = listOf(left, right, *children)

    // ...
```

That way you could drop the third commit that introduces `require()` completely.

mnonnenmacher: That constructor would be inconvenient to use if you actually have a list of children, like in the parser. Also the names `left`, `right` and `children` are confusing, and this introduces a mandatory use of the spread operator.
The current constructors each serve a specific use case:
* `list, operator` -> when you have a list of children, like in the parser
* `operator, vararg children` -> most convenient when manually building the expression in code like in the tests
* `left, operator, right` -> for backward compatibility

utils/spdx/src/main/kotlin/SpdxExpression.kt
```diff
@@ -201,29 +201,41 @@ sealed class SpdxExpression {
...
  *
  * [1]: https://spdx.dev/spdx-specification-21-web-version#h.jxpfx0ykyb60
  */
-class SpdxCompoundExpression(
-    val left: SpdxExpression,
-    val operator: SpdxOperator,
-    val right: SpdxExpression
+class SpdxCompoundExpression internal constructor(
+    val children: List<SpdxExpression>,
+    val operator: SpdxOperator
```
sschuberth, line 220: Sorry, one more question: Is the order of parameters intended like this? To me it seems like we should align to line 232 now and make the `operator` go first.

utils/spdx/src/main/kotlin/SpdxExpression.kt
mnonnenmacher: I have made the list constructor internal as discussed offline.

sschuberth (approved):

PR #8322, opened 2024-02-21
refactor(scanner)!: Align provenance storages on `write` instead of `put`

sschuberth:
Say `write` instead of `put` to match `read` functions and to align with the `ScanStorage` interfaces.

### DIFF
--- a/scanner/src/main/kotlin/provenance/NestedProvenanceStorage.kt
+++ b/scanner/src/main/kotlin/provenance/NestedProvenanceStorage.kt
@@ -58,10 +58,10 @@ interface NestedProvenanceStorage {
     fun readNestedProvenance(root: RepositoryProvenance): NestedProvenanceResolutionResult?
 
     /**
-     * Put the resolution [result] for the [root] provenance into the storage. If the storage already contains an entry
-     * for [root] it is overwritten.
+     * Write the resolution [result] for the [root] provenance into the storage. If the storage already contains an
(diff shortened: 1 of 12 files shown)


sschuberth:
As discussed, @mnonnenmacher.

scanner/src/main/kotlin/provenance/NestedProvenanceStorage.kt
```diff
@@ -58,10 +58,10 @@ interface NestedProvenanceStorage {
     fun readNestedProvenance(root: RepositoryProvenance): NestedProvenanceResolutionResult?
 
     /**
-     * Put the resolution [result] for the [root] provenance into the storage. If the storage already contains an entry
-     * for [root] it is overwritten.
+     * Write the resolution [result] for the [root] provenance into the storage. If the storage already contains an
+     * entry for [root] it is overwritten.
```
fviernau, line 62: out of scope: I've recently mentioned (slack) the constraint violations in the logs which occur when the entry already exists. Without looking into it, I guess that it's actually not overwritten but ignored in the Postgres based implementation.

mnonnenmacher: @fviernau I'm only aware of possible constraint violations related to storing scan results, but not for provenance resolution results.

mnonnenmacher (approved):

PR #8326, opened 2024-02-21
build(Gradle): Force color for the `run` task if the terminal supports it

sschuberth:
When running through Gradle, the app's output is redirected to terminal detection does not work. Force color in this case if the host's terminal actually supports color. This way e.g.

    ./gradlew :cli:run --args="--help"

shows colorful help output.

### DIFF
--- a/buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
@@ -184,3 +184,10 @@ tasks.named<CreateStartScripts>("startScripts") {
         )
     }
 }
+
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
```diff
@@ -184,3 +184,10 @@ tasks.named<CreateStartScripts>("startScripts") {
         )
     }
 }
+
```
fviernau, line 187: commit message: "to terminal detection" "**s**o terminal detection"?

fviernau (approved):

PR #8329, opened 2024-02-22
Improve the Vcs Info fallback mechanism for node package managers

fviernau:
See individual commits.

### DIFF
--- a/plugins/package-managers/node/src/funTest/assets/projects/synthetic/npm-babel-expected-output.yml
+++ b/plugins/package-managers/node/src/funTest/assets/projects/synthetic/npm-babel-expected-output.yml
@@ -2184,7 +2184,7 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "git+https://github.com/chalk/ansi-regex.git"
+    url: "https://github.com/chalk/ansi-regex.git"
     revision: "7c908e7b4eb6cd82bfe1295e33fdf6d166c7ed85"
     path: ""
   vcs_processed:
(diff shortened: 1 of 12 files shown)


plugins/package-managers/node/src/funTest/assets/projects/synthetic/npm-babel-expected-output.yml
```diff
@@ -2245,7 +2245,7 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "git+https://github.com/es128/anymatch.git"
+    url: "https://github.com/es128/anymatch"
```
sschuberth, line 2248: Just an observation: Weird that the `.git` suffix sometimes gets stripped, and sometimes not.

plugins/package-managers/node/src/funTest/kotlin/NpmFunTest.kt
```diff
@@ -126,9 +126,9 @@ class NpmFunTest : WordSpec({
             val expectedResult = patchExpectedResult(expectedResultFile, definitionFile)
                 .fromYaml<ProjectAnalyzerResult>()
 
-            val actualResult = create("NPM").resolveSingleProject(definitionFile, resolveScopes = true)
+            val result = create("NPM").resolveSingleProject(definitionFile, resolveScopes = true)
```
sschuberth, line 129: I'm not sure *what* this aligns to (the commit message does not say), but I believe I've shared earlier that personally I'd prefer to align on `actualResult` as it better matches `expectedResult`.

sschuberth (approved):
Approving despite my personal preference to not align on `result` but on `actualResult`.

PR #8330, opened 2024-02-22
test(osv): Update an expected result

fviernau:


clients/osv/src/funTest/assets/vulnerabilities-by-name-and-version-expected-result.json
```diff
@@ -258,7 +258,7 @@
             },
             {
                 "type": "WEB",
-                "url": "http://jinja.pocoo.org/docs/changelog/"
+                "url": "http://jinja.pocoo.org/docs/changelog"
```
sschuberth, line 261: As the trailing slash goes back and forth, should we normalize this, e.g. by always trimming it?

fviernau: can we merge this one now and I'll look into normalization into a PR on top ?

sschuberth: Yes, I'll approve. Though the final newline was also again stripped. And the `GoDep` test still fails. But let's try to solve one problem at a time here.

sschuberth (approved):

clients/osv/src/funTest/assets/vulnerabilities-by-name-and-version-expected-result.json
fviernau: here is the follow up: https://github.com/oss-review-toolkit/ort/pull/8331

PR #8331, opened 2024-02-22
test(osv): Make the test resilient WRT trailing slashes in URLs

fviernau:
Recently, the tests failed because the results kept on alternating between URLs with and without trailing slashes. So, make the test tolerate both in order to reduce the flakyness.

### DIFF
--- a/clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
+++ b/clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
@@ -53,9 +53,14 @@ private fun Vulnerability.patchIgnorableFields() =
         }
     )
 
+private fun Vulnerability.normalizeUrls(): Vulnerability {
+    val references = references.mapTo(mutableSetOf()) { it.copy(url = it.url.removeSuffix("/")) }
+    return copy(references = references)
+}
+
 private val emptyJsonObject = JsonObject(emptyMap())
 
-private fun List<Vulnerability>.patchIgnorableFields() = map { it.patchIgnorableFields() }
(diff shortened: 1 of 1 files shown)


sschuberth (changes requested):
See the detekt issue about the unused import.

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
```diff
@@ -53,9 +53,14 @@ private fun Vulnerability.patchIgnorableFields() =
         }
     )
 
+private fun Vulnerability.normalizeUrls(): Vulnerability {
+    val references = references.mapTo(mutableSetOf()) { it.copy(url = it.url.removeSuffix("/")) }
```
sschuberth, line 57: Just a thought: We could also address this more intrinsically by changing `val url: String` to `val url: URI` in the `Reference` class. As according to the [spec](https://ossf.github.io/osv-schema/#references-field) this field always contains a URL, that should work, and avoid the need for normalization. How about that?

The only downside I see is that kotlinx-serialization requires a custom (de-)serializer for the URI Java type. We have that already (e.g. `URISerializer` as part of ClearlyDefined), it just need to be moved to a common place. Or we take https://github.com/Kantis/ks3 into use, which is something I was planning to do anyway.

fviernau: If at all, I'd again propose to make this a future iteration and keep this PR as-is, as it is an improvement with low effort.

sschuberth (approved):

PR #8332, opened 2024-02-22
Migrate from custom kotlinx-serializers for Java types to ks3

sschuberth:

### DIFF
--- a/clients/clearly-defined/build.gradle.kts
+++ b/clients/clearly-defined/build.gradle.kts
@@ -30,6 +30,7 @@ dependencies {
     api(libs.retrofit)
 
     implementation(libs.bundles.kotlinxSerialization)
+    implementation(libs.bundles.ks3)
     implementation(libs.kotlinx.coroutines)
     implementation(libs.retrofit.converter.kotlinxSerialization)
     implementation(libs.retrofit.converter.scalars)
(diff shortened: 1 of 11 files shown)


nnobelis:
Could you please give a bit of context on this?

Is there currently 3  JSON/Yaml serializers in ORT ?

- kotlinx
- ks3
- jackson

Is ks3 the new serializer we should always use now ?

clients/clearly-defined/src/main/kotlin/Curation.kt
```diff
@@ -17,15 +17,12 @@
  * License-Filename: LICENSE
  */
 
-@file:UseSerializers(FileSerializer::class, URISerializer::class)
-
 package org.ossreviewtoolkit.clients.clearlydefined
 
-import java.io.File
-import java.net.URI
+import io.ks3.java.`typealias`.FileAsString
```
mnonnenmacher, line 22: That's a very inconvenient choice for a package name.

clients/clearly-defined/build.gradle.kts
mnonnenmacher, line 1: Please add a link to the ks3 project / documentation to the commit message.

mnonnenmacher (changes requested):

clients/clearly-defined/src/main/kotlin/Curation.kt
sschuberth: I've [reported](https://github.com/Kantis/ks3/issues/113) it.

PR #8333, opened 2024-02-22
NodeSupport: Further improve determining VCS info

fviernau:
Adjust the fallback algorithm in various ways mainly to:

1. Derive revisions from the URL only if they are a `sha1`, to avoid branch names as `vcs.revision` such as "master" / "main".
2. Derive the `vcs.path` from the URLs
3. Derive the `vcs.type` from the URLs 

For our test data (see diff) this seems to be an improvement in all cases.

### DIFF
--- a/plugins/package-managers/node/src/funTest/assets/projects/synthetic/npm-babel-expected-output.yml
+++ b/plugins/package-managers/node/src/funTest/assets/projects/synthetic/npm-babel-expected-output.yml
@@ -2552,13 +2552,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-cli"
+    url: "https://github.com/babel/babel.git"
     revision: ""
-    path: ""
+    path: "packages/babel-cli"
   vcs_processed:
     type: "Git"
     url: "https://github.com/babel/babel.git"
-    revision: "master"
+    revision: ""
     path: "packages/babel-cli"
 - id: "NPM::babel-code-frame:6.26.0"
   purl: "pkg:npm/babel-code-frame@6.26.0"
@@ -2582,13 +2582,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-code-frame"
+    url: "https://github.com/babel/babel.git"
     revision: ""
-    path: ""
+    path: "packages/babel-code-frame"
   vcs_processed:
     type: "Git"
     url: "https://github.com/babel/babel.git"
-    revision: "master"
+    revision: ""
     path: "packages/babel-code-frame"
 - id: "NPM::babel-core:6.26.3"
   purl: "pkg:npm/babel-core@6.26.3"
@@ -2612,13 +2612,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-core"
+    url: "https://github.com/babel/babel.git"
     revision: ""
-    path: ""
+    path: "packages/babel-core"
   vcs_processed:
     type: "Git"
     url: "https://github.com/babel/babel.git"
-    revision: "master"
+    revision: ""
     path: "packages/babel-core"
 - id: "NPM::babel-generator:6.26.1"
   purl: "pkg:npm/babel-generator@6.26.1"
@@ -2642,13 +2642,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-generator"
+    url: "https://github.com/babel/babel.git"
     revision: ""
-    path: ""
+    path: "packages/babel-generator"
   vcs_processed:
     type: "Git"
     url: "https://github.com/babel/babel.git"
-    revision: "master"
+    revision: ""
     path: "packages/babel-generator"
 - id: "NPM::babel-helpers:6.24.1"
   purl: "pkg:npm/babel-helpers@6.24.1"
@@ -2672,13 +2672,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-helpers"
+    url: "https://github.com/babel/babel.git"
     revision: ""
-    path: ""
+    path: "packages/babel-helpers"
   vcs_processed:
     type: "Git"
     url: "https://github.com/babel/babel.git"
-    revision: "master"
+    revision: ""
     path: "packages/babel-helpers"
 - id: "NPM::babel-messages:6.23.0"
   purl: "pkg:npm/babel-messages@6.23.0"
@@ -2702,13 +2702,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-messages"
+    url: "https://github.com/babel/babel.git"
     revision: ""
-    path: ""
+    path: "packages/babel-messages"
   vcs_processed:
     type: "Git"
     url: "https://github.com/babel/babel.git"
-    revision: "master"
+    revision: ""
     path: "packages/babel-messages"
 - id: "NPM::babel-polyfill:6.26.0"
   purl: "pkg:npm/babel-polyfill@6.26.0"
@@ -2732,13 +2732,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-polyfill"
+    url: "https://github.com/babel/babel.git"
     revision: ""
-    path: ""
+    path: "packages/babel-polyfill"
   vcs_processed:
     type: "Git"
     url: "https://github.com/babel/babel.git"
-    revision: "master"
+    revision: ""
     path: "packages/babel-polyfill"
 - id: "NPM::babel-register:6.26.0"
   purl: "pkg:npm/babel-register@6.26.0"
@@ -2762,13 +2762,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-register"
+    url: "https://github.com/babel/babel.git"
     revision: ""
-    path: ""
+    path: "packages/babel-register"
   vcs_processed:
     type: "Git"
     url: "https://github.com/babel/babel.git"
-    revision: "master"
+    revision: ""
     path: "packages/babel-register"
 - id: "NPM::babel-runtime:6.26.0"
   purl: "pkg:npm/babel-runtime@6.26.0"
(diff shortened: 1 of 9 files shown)


plugins/package-managers/node/src/funTest/assets/projects/synthetic/npm-babel-expected-output.yml
```diff
@@ -2552,13 +2552,13 @@ packages:
       algorithm: "SHA-1"
   vcs:
     type: "Git"
-    url: "https://github.com/babel/babel/tree/master/packages/babel-cli"
+    url: "https://github.com/babel/babel.git"
```
sschuberth, line 2555: Commit message:
- "Improve *the* algorithm"
- "by deriving the path either from the given URL in addition to from" -> "by deriving the path from the URL in addition to"

sschuberth: Looking purely at the data without having looked at the code changes yet, this does not look like an improvement to me, because:
- The previous (unprocessed) `vcs` is now processed, i.e. not taken as-is from metadata anymore. Even if the URL was wrong in the original metadata, it should be taken here, as it's the sole responsibility of `vcs_processed` to "correct" the metadata.
- We loose information about what branch was referred in the original metadata.

plugins/package-managers/node/src/main/kotlin/utils/NpmSupport.kt
```diff
@@ -153,15 +154,22 @@ internal fun parseNpmVcsInfo(node: JsonNode): VcsInfo {
         revision = head
     )
 
-    val type = repository["type"].textValueOrEmpty()
-    val url = repository.textValue() ?: repository["url"].textValueOrEmpty()
+    if (repository.isTextual) {
+        val url = expandNpmShortcutUrl(repository.textValue())
+        val vcsInfo = VcsHost.parseUrl(url)
+        return vcsInfo.copy(revision = head.takeIf { it.isSha1() } ?: vcsInfo.revision.takeIf { it.isSha1() }.orEmpty())
```
sschuberth, line 160: While I did read point 2. of your commit message, I do not understand what problem it solves. Even if there was "master" in the metadata, still a tag that matches the version would get precedence if I'm reading the code correctly:

https://github.com/oss-review-toolkit/ort/blob/6d04874140e4bf760ade6f1657173b67d0f4db41/downloader/src/main/kotlin/VersionControlSystem.kt#L346-L356

We first add our guessed revisions as candidates, and then any revisions from metadata. So how does it help to only take revisions if they are commit hashes, and map "master" to ""?

plugins/package-managers/node/src/main/kotlin/utils/NpmSupport.kt
```diff
@@ -153,15 +154,22 @@ internal fun parseNpmVcsInfo(node: JsonNode): VcsInfo {
...
+    val vcsInfo = VcsHost.parseUrl(url)
 
     return VcsInfo(
-        type = VcsType.forName(type),
+        type = type ?: vcsInfo.type,
         url = expandNpmShortcutUrl(url),
-        revision = head,
-        path = path
+        revision = head.takeIf { it.isSha1() } ?: vcsInfo.revision.takeIf { it.isSha1() }.orEmpty(),
+        path = path.takeIf { it.isNotBlank() } ?: vcsInfo.path
```
sschuberth, line 172: I like this line, though. We can simplify to `isEmpty()` probably.

sschuberth (changes requested):

plugins/package-managers/node/src/main/kotlin/utils/NpmSupport.kt
fviernau: I wasn't aware of this bit of the code, so my comment about the tag matching was wrong.

However, when you have "master" as revision and no release tags, then either
1. Re-scanning (scanner) has no cache hit which can lead to long execution times. 
2. If there is cache hit because provenance resolution stored, then you scan some arbitrary 
    old revisions of the branch without noticing it. The revision is determined by when it was first scanned.

mnonnenmacher: If a package provenance was resolved from a branch the resolution is always re-attempted:
https://github.com/oss-review-toolkit/ort/blob/e62e68b008ea13967cd16bef1404c5d1eab42622/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt#L173-L190

I would agree that if resolving the provenance from a branch is an issue we should solve this globally and not on the package manager level. For example, we could introduce a configuration option to disable this and enforce curations in this case. Currently `allowMovingRevisions` is hardcoded to `true` when getting the revision candidates:
https://github.com/oss-review-toolkit/ort/blob/e62e68b008ea13967cd16bef1404c5d1eab42622/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt#L208-L209

plugins/package-managers/node/src/funTest/assets/projects/synthetic/npm-babel-expected-output.yml
fviernau: > this does not look like an improvement to me, because:

@mnonnenmacher  do you also see all aspects of this PR as "not an improvement" / "undersirable", or do you see certain aspects of it which make sense? e.g. we've talked mostly about the revision so far, but also `vcs.path` and `vcs.type` do change, as well as the URLs.

sschuberth: > do you also see all aspects of this PR as "not an improvement" / "undersirable"

For the record, I'm not seeing *all* aspects of this PR as a whole negatively. The first commit is a nice refactoring, and as [noted](https://github.com/oss-review-toolkit/ort/pull/8333#discussion_r1500430367) used the URL as a fallback for determining the path also makes sense to me.

fviernau: I've split out that first commit to https://github.com/oss-review-toolkit/ort/pull/8336.

fviernau: > Looking purely at the data without having looked at the code changes yet, this does not look like an improvement to me, because:
> 
>     * The previous (unprocessed) `vcs` is now processed, i.e. not taken as-is from metadata anymore. Even if the URL was wrong in the original metadata, it should be taken here, as it's the sole responsibility of `vcs_processed` to "correct" the metadata.
> 
>     * We loose information about what branch was referred in the original metadata.

I tried to start working on this PR again, but I'm having difficulties solving the issue you mention here. 
Let me try to explain. I understand your ask as follows:

The `Package.vcs` field should only hold the data exactly as specified in package metadata.

That ask implies to me that the function extracting the VCS info `NpmSupport: parseNpmVcsInfo()` must not process it, but processing must happen separately. The `parseNpmVcsInfo()` is a couple of callers, but looking just at the caller inside `Npm` already shows that your ask is already not the case, because the algorithm:

1. Gathers package metadata in `var vcsFromPackage`
2. Gathers VCS from download in `val vcsFromDownloadUrl`
3. Gathers VCS from file system directory in `val vcsFromDirectory`
4. merges 2 or 3 into 1.
5. Uses the result als the (unprocessed) `Package.vcs`

So, what you are asking from my code is already before heavily violation. 
I'm not objecting your ask, but wanted to double check whether you are aware that
this is basically asking for a bit larger refactoring. I can do that, but I would like to agree
upfront, what exactly has to go into `Package.vcs`. (My guess is that only things from package metadata should go there. ) 

Are we that strict as well in other package managers? (I would also be open for the approach to not refactor it and leave the change basically as-is) What do you think?

fviernau:
I've slit-out handling textal repository nodes: https://github.com/oss-review-toolkit/ort/pull/8512

plugins/package-managers/node/src/funTest/assets/projects/synthetic/npm-babel-expected-output.yml
sschuberth: > The `Package.vcs` field should only hold the data exactly as specified in package metadata.

Yes, at least that's how the property is currently documented:

https://github.com/oss-review-toolkit/ort/blob/d39d163d136ddd9d8661f5c1350f2a3644ccde16/model/src/main/kotlin/Package.kt#L109-L112

Note, though, that at least to me "package metadata" does not exclude that more sources (like a package registry) than just `package.json` could be consulted for metadata.

But it's a valid question where to draw the line to `Package.vcsProcessed` then.

> I would like to agree upfront, what exactly has to go into Package.vcs.

Makes sense. Could you please add it to the agenda of the next Kotlin dev meeting?

plugins/package-managers/node/src/main/kotlin/utils/NpmSupport.kt
sschuberth, line 160: Maybe something like this instead (similar in line 171):

    listOf(head, vcsInfo.revision).firstOrNull { it.isSha1() }.orEmpty()

plugins/package-managers/node/src/main/kotlin/utils/NpmSupport.kt
```diff
@@ -153,15 +154,22 @@ internal fun parseNpmVcsInfo(node: JsonNode): VcsInfo {
...
 
-    val type = repository["type"].textValueOrEmpty()
-    val url = repository.textValue() ?: repository["url"].textValueOrEmpty()
+    if (repository.isTextual) {
+        val url = expandNpmShortcutUrl(repository.textValue())
+        val vcsInfo = VcsHost.parseUrl(url)
+        return vcsInfo.copy(revision = head.takeIf { it.isSha1() } ?: vcsInfo.revision.takeIf { it.isSha1() }.orEmpty())
+    }
+
+    val type = VcsType.forName(repository["type"].textValueOrEmpty()).takeUnless { it == VcsType.UNKNOWN }
```
sschuberth, line 163: At this point, we know that `repository` is an object node with mandatory `type` and `url` nodes, so we could use just `textValue()` instead of `textValueOrEmpty()` here any below.

sschuberth: Maybe move the `.takeUnless { it == VcsType.UNKNOWN }` part to line 169 to highlight why the `?:` fallback is needed there?

plugins/package-managers/node/src/main/kotlin/utils/NpmSupport.kt
```diff
@@ -173,3 +181,7 @@ internal fun splitNpmNamespaceAndName(rawName: String): Pair<String, String> {
     val namespace = rawName.removeSuffix(name).removeSuffix("/")
     return Pair(namespace, name)
 }
+
+private val SHA1_REGEX = "[0-9a-f]{7,40}".toRegex()
+
+private fun String.isSha1(): Boolean = SHA1_REGEX.matches(this.lowercase())
```
sschuberth, line 187: Th explicit `this.` should be dropped here.

fviernau:
After looking at it again, the changes take effect in our funTest projects primarily on unprocessed VCS while processed VCS remains the same. So, I've double checked whether the changes are needed for the scans for which I've originally developed the PR for, but it turned out the scans figure out metadata successfully without this PR.

So, closing this as not needed / change unprocessed VCS only is not desired.

PR #8334, opened 2024-02-22
Detekt fixes

sschuberth:


scanner/src/testFixtures/kotlin/AbstractPathScannerWrapperFunTest.kt
nnobelis, line 1: Typo in commit message "unsed"

analyzer/src/testFixtures/kotlin/TestUtils.kt
```diff
@@ -122,7 +121,7 @@ fun ProjectAnalyzerResult.withInvariantIssues() =
         issues = issues.sortedBy { it.message }.map { it.copy(timestamp = Instant.EPOCH) }
     )
 
-fun Spec.analyze(
+fun analyze(
```
fviernau, line 124: I also stumbled across these recently. I assumed that the reason for limiting the function to `Spec`
was to reduce the scope and not pollute the whole namespace with this function. What do you think about this?

evaluator/src/main/kotlin/PackageRule.kt
```diff
@@ -54,7 +54,10 @@ open class PackageRule(
     private val licenseRules = mutableListOf<LicenseRule>()
 
     @Suppress("UNUSED") // This is intended to be used by rule implementations.
-    val uncuratedPkg: Package by lazy { ruleSet.ortResult.getUncuratedPackageOrProject(pkg.metadata.id)!! }
+    val uncuratedPkg: Package by lazy {
```
fviernau, line 57: Have you considered to disable the reporting of this type of issue entirely temporarily? 
...Or alternatively disabling that violation for each file it appears on file level? To me the suppress things seem to reduce readability.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
```diff
@@ -177,7 +177,7 @@ class CocoaPods(
         return Package(
             id = id,
             authors = emptySet(),
-            declaredLicenses = podspec.license.takeUnless { it.isEmpty() }?.let { setOf(it) } ?: emptySet(),
+            declaredLicenses = podspec.license.takeUnless { it.isEmpty() }?.let { setOf(it) }.orEmpty(),
```
fviernau, line 180: How about `.orEmpty().toSet()` to get rid of this `let` ?

plugins/package-managers/conan/src/main/kotlin/Conan.kt
```diff
@@ -488,5 +488,5 @@ class Conan(
      * author name; otherwise, return an empty set.
      */
     private fun parseAuthors(node: JsonNode): Set<String> =
-        parseAuthorString(node["author"]?.textValue(), '<', '(')?.let { setOf(it) } ?: emptySet()
+        parseAuthorString(node["author"]?.textValue(), '<', '(')?.let { setOf(it) }.orEmpty()
```
fviernau, line 491: How about `.orEmpty().toSet()` to get rid of this `let` ?

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetInspectorExtensions.kt
```diff
@@ -113,6 +113,7 @@ internal fun Collection<NuGetInspector.PackageData>.toOrtPackages(): Set<Package
 
         fun NuGetInspector.PackageData.getHash(): Hash =
             Hash.create(
+                @Suppress("UseOrEmpty")
```
fviernau, line 116: Have you considered supressing this for the whole function, so that readability of function body is not affected?

plugins/package-managers/stack/src/main/kotlin/Stack.kt
```diff
@@ -347,7 +347,7 @@ class Stack(
                 .map(String::trim)
                 .filter(String::isNotEmpty)
                 .mapNotNullTo(mutableSetOf(), ::parseAuthorString),
-            declaredLicenses = map["license"]?.let { setOf(it) } ?: emptySet(),
+            declaredLicenses = map["license"]?.let { setOf(it) }.orEmpty(),
```
fviernau, line 350: How about `.orEmpty().toSet()` to get rid of this `let` ?

fviernau (changes requested):

analyzer/src/testFixtures/kotlin/TestUtils.kt
sschuberth: It was both: I was looking for a reasonable justification to put these into `Extensions.kt` back then, without the need to introduce a new file, and though it would be a nice "trick" limit the functions to the `Spec` context, although the context is never used. But it always annoying me that the IDE is actually dimming `Spec` to hint it being unused.

Anyway, by now these are in the `testFixture` source set, and the required file name change now includes "Test" as well, so I think the scope should be clear, and name clashes are very unlikely.

evaluator/src/main/kotlin/PackageRule.kt
sschuberth: > Have you considered to disable the reporting of this type of issue entirely temporarily?

Yes, and I decided against it. I want to highlight the exact places where these unsafe calls are used in order to more easily find them and address them one-by one going forward.

> To me the suppress things seem to reduce readability.

I partly agree, the code looks "uglier" now. But that's deliberate, to have an incentive to clean it up via a better implementation.

analyzer/src/testFixtures/kotlin/TestUtils.kt
fviernau: > Anyway, by now these are in the `testFixture` source set, and the required file name change now includes "Test" as well, so I think the scope should be clear, and name clashes are very unlikely.

Agree.

evaluator/src/main/kotlin/PackageRule.kt
fviernau: So, would you be ok to move these at the minimum to the function level? (In my view the effect on readability is not problematic then anymore. The scope should still be small then, if not then like the cause is too large function.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
sschuberth: I don't see how that should work (`podspec.license` is a string), but I've done a commit on top that simplifies this via `setOfNotNull()`.

plugins/package-managers/conan/src/main/kotlin/Conan.kt
sschuberth: See above.

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetInspectorExtensions.kt
sschuberth: Yes, I have considered this. In general I prefer to have `@Suppress` as narrow-scoped as possible, to not inadvertently suppress other occurrences (that might be added later to the code). I only lift `@Suppress` to a higher level if it make sense to suppress multiple clearly belonging statements at once.

plugins/package-managers/stack/src/main/kotlin/Stack.kt
sschuberth: See above.

plugins/package-managers/cocoapods/src/main/kotlin/CocoaPods.kt
fviernau: Sorry, my fault. Thought podspec.license was some collection time.

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetInspectorExtensions.kt
fviernau: Ok, then let's leave it as is. Maybe we can collect other opinions in the dev meeting how they would see moving these to function level, but leave that out of scope here.

fviernau (approved):

plugins/package-managers/nuget/src/main/kotlin/utils/NuGetInspectorExtensions.kt
sschuberth: Sure, feel free to put that topic onto the agenda 👍🏻

PR #8338, opened 2024-02-25
Reformat README

mnonnenmacher:

### DIFF
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -4,6 +4,7 @@
 globs:
   - "**/*.md"
 ignores:
+  - "**/build/**"
   - "**/node_modules/**"
   - "**/projects/external/**"
   - "**/multi-kotlin-project/**"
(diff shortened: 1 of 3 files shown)


sschuberth (approved):
I'm ok to take it into use, but we should ASAP implement checks to enforce this (for already migrated files).

mnonnenmacher:
> I'm ok to take it into use, but we should ASAP implement checks to enforce this (for already migrated files).

Like I mentioned in the core dev meeting, the problem is that there is no rule for that. There are two markdownlint plugins which we might use, but both of them don't seem to be actively maintained and I don't know how well they work:
https://www.npmjs.com/package/markdownlint-rule-max-one-sentence-per-line
https://www.npmjs.com/package/sentences-per-line

I could give it a try anyway.

sschuberth:
What's blocking to merge this @mnonnenmacher?

sschuberth:
> I wasn't sure if you are stil ok with the merge

I'm still fine. Let's merge, and if the feel comfortable with the style, come up with automated checks and roll it out further. And if are not fine, this change is small enough to revert (even manually).

PR #8342, opened 2024-02-26
Snippet Choice continuation

nnobelis:


clients/fossid-webapp/src/main/kotlin/model/result/MatchedLines.kt
```diff
@@ -24,4 +24,9 @@ import org.ossreviewtoolkit.clients.fossid.PolymorphicList
 data class MatchedLines(
     val localFile: PolymorphicList<Int>,
     val mirrorFile: PolymorphicList<Int>
-)
+) {
```
sschuberth, line 27: Commit message: "allow*s* to control"

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -190,8 +195,30 @@ internal fun mapSnippetFindings(rawResults: RawResults, issues: MutableList<Issu
...
+                        }
+                        true
+                    }
+
+                    else -> false
+                }
+
+                if (!isSnippetChoice) {
+                    findings.getOrPut(sourceLocation) { mutableSetOf(ortSnippet) } += ortSnippet
+                }
```
sschuberth: I tried to come up with something without the explicit `true`s and that does not iterate the whole list of choices twice, and ended up with this:

```kotlin
val choicesForLocation = snippetChoices.filter { it.given.sourceLocation == sourceLocation }
val choicesForLocationAndPurl = choicesForLocation.filter { it.choice.purl == purl }

val isChosenSnippet = (choicesForLocation.isNotEmpty() && choicesForLocationAndPurl.isNotEmpty()).also {
    logger.info {
        "Ignoring snippet $purl for file ${sourceLocation.prettyPrint()}, " +
            "as this is a chosen snippet."
    }
}

val isChosenSnippetLocation = choicesForLocation.isNotEmpty().also {
    logger.info {
        "Ignoring snippet $purl for file ${sourceLocation.prettyPrint()}, " +
            "as there is a snippet choice for this source location."
    }
}

if (!isChosenSnippet && !isChosenSnippetLocation) {
    findings.getOrPut(sourceLocation) { mutableSetOf(ortSnippet) } += ortSnippet
}
```

Not sure if it's much nicer, though.

plugins/scanners/fossid/src/test/kotlin/FossIdSnippetChoiceTest.kt
```diff
@@ -0,0 +1,259 @@
+/*
+ * Copyright (C) 2023 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
```
sschuberth, line 2: This now should be 2024, I guess.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -216,7 +221,17 @@ internal fun mapSnippetFindings(
                     else -> false
                 }
 
-                if (!isSnippetChoice) {
+                val isLocationsWithFalsePositives = locationsWithFalsePositives.any {
+                    it.given.sourceLocation == sourceLocation
+                }
+                if (isLocationsWithFalsePositives) {
```
sschuberth, line 227: Nit: Empty line before this one.

plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
```diff
@@ -76,12 +76,12 @@ License(s):
...
 | Source Location | pURL | License | File | URL | Score | Release Date
 
-.${snippetCount}+|
+.${snippetCount*2}+|
 [#if helper.isFullFileLocation(sourceLocation)]
 Full match
 [#else]
 Partial match +
-${sourceLocation.startLine}-${sourceLocation.endLine}
+${sourceLocation.startLine?c}-${sourceLocation.endLine?c}
```
sschuberth, line 84: Out of curiosity, what does `?c` do?

plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
```diff
@@ -94,6 +94,46 @@ ${sourceLocation.startLine}-${sourceLocation.endLine}
 | ${snippet.purl!""}
 | ${snippet.licenses!""} | ${snippetFilePath} | ${snippet.provenance.sourceArtifact.url!""}[URL]
 | ${snippet.score!""} | ${snippet.additionalData["releaseDate"]}
+6+a|
```
sschuberth, line 97: It this line intended? The syntax looks a bit odd.

plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
nnobelis: https://freemarker.apache.org/docs/ref_builtins_number.html#ref_builtin_c

It converts the number to a number "for the machine".
Without it, the number are rendered in the locale e.g. 1000->1.000

plugins/reporters/fossid/src/main/resources/templates/asciidoc/fossid_snippet.ftl
nnobelis: Yes

`6+` is to span 6 columns:
https://docs.asciidoctor.org/asciidoc/latest/tables/span-cells/

For the `a` I don't remember exactly. I think it was the `AsciiDoc` format that was required:
https://docs.asciidoctor.org/asciidoc/latest/tables/format-cell-content/

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -809,10 +810,12 @@ class FossId internal constructor(
                     }
                     logger.info { "${snippets.size} snippets." }
 
+                    val filteredSnippets = snippets.filterNotTo(mutableSetOf()) { it.matchType in excludedMatchTypes }
```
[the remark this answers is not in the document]
nnobelis: I moved the logic to a function in `MatchType`.
I don't understand your second comment though: this is not moved code but new code.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: It creates double logging in case `choicesForLocation.isNotEmpty()` is `true`, while the previous code didn't.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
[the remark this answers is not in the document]
sschuberth: To me, this basically just moves former line 826 to line 801, and also moves related filtering to happen earlier, but it's not entirely new code. Or am I missing something?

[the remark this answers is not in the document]
nnobelis: I looked again at the commit diffs and I see what you mean. Yes the filtering is in fact not new, I mixed things up.

I adapted the commit message to reflect that.

clients/fossid-webapp/src/main/kotlin/model/result/MatchType.kt
```diff
@@ -26,5 +26,12 @@ enum class MatchType {
...
     INTAKE,
     NONE,
-    PARTIAL
+    PARTIAL;
+
+    companion object {
+        /**
+         * Test if the match type is a valid one i.e. the scanner can work with it.
+         */
+        fun MatchType.isValidType() = this == PARTIAL || this == FULL
```
sschuberth, line 35: Probably my fault, as I believe I've mentioned the term "companion object" before, but this should just be a regular function on the enum itself:

    fun isValidType() = this == FULL || this == PARTIAL

Nit: I've also swapped the order of values to check to make them match the definition order.

sschuberth: Also strictly speaking, the introduction of this helper function is a refactoring that could have happened independently before. While I'm fine with keeping it in this commit, the commit message could at least say something like

> While at it, introduce a helper function for valid match types.

sschuberth (changes requested):

sschuberth (changes requested):

sschuberth (approved):

PR #8346, opened 2024-02-26
Fix exception associations

sschuberth:


model/src/test/kotlin/utils/FindingsMatcherTest.kt
```diff
@@ -313,6 +313,14 @@ class FindingsMatcherTest : WordSpec() {
                             "GPL-2.0-only WITH Classpath-exception-2.0 AND " +
                             "GPL-3.0-only WITH Classpath-exception-2.0"
                         ).toSpdx()
+
+                    associateLicensesWithExceptions(
+                        @Suppress("MaxLineLength")
+                        "(CDDL-1.1 OR GPL-2.0-only WITH Classpath-exception-2.0) AND GPL-2.0-only AND Classpath-exception-2.0".toSpdx()
+                    ) shouldBe (
+                        "(CDDL-1.1 OR GPL-2.0-only WITH Classpath-exception-2.0) AND " +
+                            "GPL-2.0-only WITH Classpath-exception-2.0"
```
mnonnenmacher, line 322: For consistency, either suppress MaxLineLength here as well or also break line 319.

model/src/main/kotlin/utils/FindingsMatcher.kt
```diff
@@ -292,62 +291,52 @@ internal fun associateLicensesWithExceptions(license: SpdxExpression): SpdxExpre
...
-        val children = license.children.map { associateLicensesWithExceptions(it) }
-        return SpdxCompoundExpression(
-            license.operator,
-            children[0],
-            children[1],
-            *children.drop(2).toTypedArray()
-        )
+    // Only search for stand-alone exceptions as part of AND-expressions.
+    if (license.operator == SpdxOperator.OR) {
+        return license.children.map { associateLicensesWithExceptions(it) }.reduce(SpdxExpression::or)
```
mnonnenmacher, line 296: Note that using reduce this will create a nested tree of compound expressions. I was thinking about adding a normalization to the `SpdxCompoundExpression` class either as a function or within the constructor to automatically merge with child compound expressions that use the same constructor. But for now you should better call:
```kotlin
SpdxCompoundExpression(SpdxExpression.OR, license.children.map { associateLicensesWithExceptions(it) })
```

model/src/main/kotlin/utils/FindingsMatcher.kt
```diff
@@ -292,62 +291,48 @@ internal fun associateLicensesWithExceptions(license: SpdxExpression): SpdxExpre
...
-        SpdxLicenseWithExceptionExpression(SpdxLicenseIdExpression(SpdxConstants.NOASSERTION), it.toString())
+    val orphanExceptions = standAloneExceptions.map { it.toString() } - handledExceptions
+
+    orphanExceptions.mapTo(associatedLicenses) { exceptionId ->
+        SpdxLicenseWithExceptionExpression(SpdxLicenseIdExpression(SpdxConstants.NOASSERTION), exceptionId)
     }
 
     // Recreate the compound AND-expression from the associated licenses.
-    return handledLicenses.reduce(SpdxExpression::and)
+    return associatedLicenses.reduce(SpdxExpression::and)
```
mnonnenmacher, line 337: See comment above regarding the use of `reduce`.

model/src/main/kotlin/utils/FindingsMatcher.kt
```diff
@@ -292,62 +291,48 @@ internal fun associateLicensesWithExceptions(license: SpdxExpression): SpdxExpre
...
-        }
+            is SpdxSimpleExpression -> {
+                val licenseId = childLicense.toString()
 
-        // Associate all remaining licenses that are applicable with the exception and remove the exception.
-        val licenses = simpleLicenses.intersect(applicableLicenses)
-        if (licenses.isEmpty()) continue
+                val validLicenseExceptionCombinations = standAloneExceptions.mapNotNull { exception ->
+                    val exceptionId = exception.toString()
+                    val applicableLicenseId = SpdxLicenseException.mapping[exceptionId].orEmpty().map { it.id }
```
mnonnenmacher, line 314: Use plural `applicableLicenseIds`.

model/src/main/kotlin/utils/FindingsMatcher.kt
```diff
@@ -292,62 +291,48 @@ internal fun associateLicensesWithExceptions(license: SpdxExpression): SpdxExpre
...
 
-        licenses.forEach {
-            handledLicenses += SpdxLicenseWithExceptionExpression(it, exceptionString)
-        }
+                    SpdxLicenseWithExceptionExpression(childLicense, exceptionId).takeIf {
+                        licenseId in applicableLicenseId
+                    }
```
mnonnenmacher, line 318:
```kotlin
if (licenseId in applicableLicenseIds) SpdxLicenseWithExceptionExpression(childLicense, exceptionId) else null
```

would be a tiny bit more efficient.

mnonnenmacher (changes requested):

model/src/main/kotlin/utils/FindingsMatcher.kt
mnonnenmacher: I was hesitant about adding this normalization to the constructor because I think it would be good if `"a AND (b AND c)".toSpdx().toString == "a AND (b AND c)"`, even though "a AND b AND c" would be semantically correct as well.

model/src/main/kotlin/utils/FindingsMatcher.kt
sschuberth: Using reduce, though, has the advantage of transparently dealing with `associatedLicenses.size < 2` cases...

mnonnenmacher (approved):

PR #8347, opened 2024-02-27
Sbt clean ups

sschuberth:

### DIFF
--- a/plugins/package-managers/sbt/src/main/kotlin/Sbt.kt
+++ b/plugins/package-managers/sbt/src/main/kotlin/Sbt.kt
@@ -54,37 +54,6 @@ class Sbt(
     analyzerConfig: AnalyzerConfiguration,
     repoConfig: RepositoryConfiguration
 ) : PackageManager(name, analysisRoot, analyzerConfig, repoConfig), CommandLineTool {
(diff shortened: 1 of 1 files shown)


plugins/package-managers/sbt/src/main/kotlin/Sbt.kt
```diff
@@ -182,7 +180,11 @@ class Sbt(
             targetDir.walk().maxDepth(1).filterTo(pomFiles) { it.extension == "pom" }
         }
 
-        return pomFiles.distinct().map { moveGeneratedPom(it) }
+        return pomFiles.distinct().map { pomFile ->
+            moveGeneratedPom(pomFile).onFailure {
+                logger.error { "Moving '$pomFile' failed: ${it.message}" }
```
mnonnenmacher, line 154: Does this new message print the absolute paths of both the source and target files?

sschuberth: Yes, I've just verified this. Actually, the paths are even absolute without the explicit conversion, so I've removed it in a new commit.

mnonnenmacher (approved):

PR #8352, opened 2024-02-28
fix(fossid-webapp): Add a version check in `waitDownloadComplete`

nnobelis:
Previous versions of FossID could stay blocked in the state NOT_FINISHED while downloading the repository.  Therefore, during the call to `waitDownloadComplete`, a workaround that checks Git output was put in place.
However, with the newest version of FossID (23.3.0), trying to call "trigger scan" when the download is not finished could lead to an error "Downloading scan files from the git repository is NOT FINISHED. Please wait a little and click rescan button.".
Therefore, this commit adds a version check to the workaround.

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -739,7 +739,9 @@ class FossId internal constructor(
                     // stays in state "NOT FINISHED". Therefore, we check the output of the Git fetch to find out
                     // whether the download is actually done.
                     val message = response.message
-                    if (message == null || !GIT_FETCH_DONE_REGEX.containsMatchIn(message)) return@wait false
+                    if (version >= "20.2" || message == null || !GIT_FETCH_DONE_REGEX.containsMatchIn(message)) {
+                        return@wait false
+                    }
 
                     logger.warn { "The download is not finished but Git Fetch has completed. Carrying on..." }
 


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -739,7 +739,9 @@ class FossId internal constructor(
                     // stays in state "NOT FINISHED". Therefore, we check the output of the Git fetch to find out
                     // whether the download is actually done.
                     val message = response.message
-                    if (message == null || !GIT_FETCH_DONE_REGEX.containsMatchIn(message)) return@wait false
+                    if (version >= "20.2" || message == null || !GIT_FETCH_DONE_REGEX.containsMatchIn(message)) {
```
sschuberth, line 742: I don't think the `>=` comparison on strings does what you want it to do. You probably need to use the `Semver` class for a proper version number comparison.

sschuberth (changes requested):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: Well we are already using this kind of comparison in other places: https://github.com/oss-review-toolkit/ort/blob/main/plugins/scanners/fossid/src/main/kotlin/FossId.kt#L703

Do you have a concrete example where the comparison would not be the correct one ?

nnobelis: Ah I see this is the case "1" > "100" or something ?
Currently FossID versions have been pretty simple such as 20.2, 21.3.2, 23.3.0 so it should be ok.

As we urgently need this fix to repair our production environment, could we merge this and do the refactoring to`Semver" in a future PR ?

sschuberth: > Do you have a concrete example where the comparison would not be the correct one ?

E.g. `"20.10" >= "20.2"` would return `false`, though it should be `true`.

sschuberth: > As we urgently need this fix to repair our production environment

Please consider decoupling our production environment from upstream development to avoid such situations in the future.

> could we merge this and do the refactoring to`Semver" in a future PR ?

As you need to fix other places as well anyway, I'm ok with that. Please try to file that fixup PR before tomorrow's release.

sschuberth (approved):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
sschuberth: > You probably need to use the `Semver` class for a proper version number comparison.

Note to myself, another option could be to use Java's built-in [ModuleDescriptor.Version](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/module/ModuleDescriptor.Version.html).

PR #8353, opened 2024-02-28
fix(fossid-webapp): Replace version comparison with Semver

nnobelis:
`coerce` needs to be used because FossID versions are not valid semantic versions e.g. "23.3".

This follows discussion on https://github.com/oss-review-toolkit/ort/pull/8352.

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -87,6 +87,8 @@ import org.ossreviewtoolkit.utils.common.enumSetOf
 import org.ossreviewtoolkit.utils.common.replaceCredentialsInUri
 import org.ossreviewtoolkit.utils.ort.showStackTrace
 
+import org.semver4j.Semver
+
 /**
  * A wrapper for [FossID](https://fossid.com/).
  *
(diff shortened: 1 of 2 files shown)


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -700,7 +702,9 @@ class FossId internal constructor(
             // Scans that were added to the queue are interpreted as an error by FossID before version 2021.2.
             // For older versions, `waitScanComplete()` is able to deal with queued scans. Therefore, not checking the
             // response of queued scans.
-            if (version >= "2021.2" || scanResult.error != "Scan was added to queue.") {
+            if (Semver.coerce(version)?.isGreaterThan("2021.2.0") == true ||
```
sschuberth, line 705: Both comparisons should use `isGreaterThanOrEqualTo()` to match the original code. As that's quite long, consider doing something like

            val currentVersion = checkNotNull(Semver.coerce(version))
            val minVersion = checkNotNull(Semver.coerce("2021.2"))
            if (currentVersion >= minVersion || scanResult.error != "Scan was added to queue.") {

Similar below, and probably also in the test.

sschuberth (changes requested):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: Done but I am not too sure about the `checkNotNull` though: If FossID does sometimes exotic one day in the fiture, e.g. `25.6-patch6`, the integration will be broken ?

sschuberth: Actually, `coerce` turns "25.6-patch6" into "25.6.0", i.e. it's not `null`. So I thing we should be good.

PR #8362, opened 2024-02-29
build(Docker): Remove explicit Cargo version

nnobelis:
Since Cargo package manager has no version requirement, always use the version of Cargo available in Ubuntu.

### DIFF
--- a/Dockerfile-legacy
+++ b/Dockerfile-legacy
@@ -72,7 +72,6 @@ ARG SCANCODE_VERSION
 ENV \
     # Package manager versions.
     BOWER_VERSION=1.8.12 \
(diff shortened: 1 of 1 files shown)


Dockerfile-legacy
```diff
@@ -128,7 +127,7 @@ RUN --mount=type=cache,target=/var/cache/apt --mount=type=cache,target=/var/lib/
         mercurial \
         subversion \
         # Install package managers (in versions known to work).
-        cargo=$CARGO_VERSION \
+        cargo \
```
sschuberth, line 130: Please remove the `Nicolas Nobelis <nicolas.nobelis@bosch.com>` line (above the signoff) from the commit message.

sschuberth (changes requested):

sschuberth (approved):

PR #8364, opened 2024-02-29
deps(docker): Upgrade Python to the latest 3.11.x version

sschuberth:


mnonnenmacher:
@sschuberth I think you also have to upgrade pyenv to the latest version (2.3.36) so that it knows about Python 3.11.8.

sschuberth:
> @sschuberth I think you also have to upgrade pyenv to the latest version (2.3.36) so that it knows about Python 3.11.8.

Weird that `pyenv` would not be able to fetch this information dynamically... anyway, thanks for the hint, I've added an according commit. Let's see if it helps.

fviernau (approved):

Issue #8366, opened 2024-02-29
Fails if project has .python-version with a version that is not supported


elnygren:
eg. `3.11.1`

causes analyzer to not be able to analyze Python requirements.txt because 3.11.1 is not installed

solution was to remove .python-version (not desirable, but worked)

sschuberth:
I guess we could solve this by implementing something like in https://github.com/oss-review-toolkit/ort/pull/8080, but for Pip, eventually also temporarily stashing away the `.python-version` file.

sschuberth:
@elnygren, could you please check whether https://github.com/oss-review-toolkit/ort/pull/8367 fixes the issue for you?

heliocastro:
It goes a little deeper. .python-version is not strict to python version, but to python environments as well, even if we would be able develop a mechanism to install the required python, we would never be able to guess for example, if someone decided that the local env is something like "mypython". 
It could be even called "venv" and we don't know which version originally intended.

Stash could be a solution, but more an patch instead over solve the real problem.
On the direction of simplify things, i would say to check stash + get rid of entire mechanisms added like #8080 and focus on two simplifications, minimal python required over pyproject.toml or for older packages, setup.cfg. Both provides the minimum python required version as a minimal. And if not, the version of python used to analyze is the current one.

Beyond that, instruction on website explain how the user should prepare the code ahead.

sschuberth:
> It goes a little deeper. .python-version is not strict to python version

I couldn't find a formal specification for the syntax in that file, and all examples from `pyenv` I saw just contain the version number. Do you have a link to a formal specification?

heliocastro:
Python version is not a formal PEP and AFAIK is only used by pyenv.
You can find here how is described:
https://github.com/pyenv/pyenv?tab=readme-ov-file#understanding-python-version-selection

PR #8367, opened 2024-02-29
Support the `.python-version` file

sschuberth:

### DIFF
--- a/plugins/package-managers/python/src/main/kotlin/Pip.kt
+++ b/plugins/package-managers/python/src/main/kotlin/Pip.kt
@@ -88,7 +88,7 @@ class Pip(
     }
 
     override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
-        val result = runPythonInspector(definitionFile)
+        val result = runPythonInspector(definitionFile) { detectPythonVersion(definitionFile.parentFile) }
 
         val project = result.toOrtProject(managerName, analysisRoot, definitionFile)
         val packages = result.packages.toOrtPackages()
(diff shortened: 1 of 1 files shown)


plugins/package-managers/python/src/main/kotlin/Pip.kt
```diff
@@ -130,4 +130,10 @@ class Pip(
             }
         }.getOrThrow()
     }
+
+    private fun detectPythonVersion(workingDir: File): String? {
+        val pythonVersionFile = workingDir.resolve(".python-version")
+        if (!pythonVersionFile.isFile) return null
+        return pythonVersionFile.readText().trim().takeUnless { it.isEmpty() }
```
fviernau, line 137: Do you have any reference for the format of the file? Would be nice to link it from a code comment.

Furthermore, would it make sense to make the check a tad more strict? e.g. return null if it's multiline string.

sschuberth: > Do you have any reference for the format of the file?

No, see https://github.com/oss-review-toolkit/ort/issues/8366#issuecomment-1972676318.

> Furthermore, would it make sense to make the check a tad more strict? e.g. return null if it's multiline string.

Will do.

plugins/package-managers/python/src/main/kotlin/Pip.kt
```diff
@@ -130,4 +130,12 @@ class Pip(
             }
         }.getOrThrow()
     }
+
+    private fun detectPythonVersion(workingDir: File): String? {
+        // While there seems to be no formal specification, a `.python-version` file seems to be supposed to simply
+        // conatin the plan version, see e.g. https://github.com/pyenv/pyenv/blob/21c2a3d/test/version.bats#L28.
```
fviernau, line 136: typo: 
- splan -> plain
- conatin -> contain

maybe add a "just" before contain?

fviernau (changes requested):

plugins/package-managers/python/src/main/kotlin/Pip.kt
sschuberth: > maybe add a "just" before contain?

As it should be either "just" or "simply", IMO, I've swapped out "simply" for "just".

fviernau (approved):

Issue #8361, opened 2024-02-29
Implement alternative analyzers that favor speed over accuracy


sschuberth:
Historically, the ORT analyzer has been pedantic about getting things right (i.e. resolving *exactly* the same dependencies as the build system does), and gathering *all* metadata known about a package (even if irrelevant in pretty much every compliance check).

To do so, the ORT analyzer requires the build system configuration of the project under analysis to be self-contained, in good shape, and to follow best practices WRT the build system's own recommendations (e.g. for Gradle, to not evaluate environment variables at task configuration time).

While it's somewhat nice that the ORT analyzer implicitly checks for adhering to those best practices, which can be seen as a build health analysis feature, there are cases where it hinders getting the required information, e.g. when analyzing legacy code bases and projects that are in maintenance mode.

Also, for many compliance checks, e.g. the hierarchy in a dependency graph does not really matter, but a flat list of packages would be enough, as [only the combination of distributed packages and their licenses might matter](https://github.com/oss-review-toolkit/ort/pull/4977#discussion_r997947353).

So the proposal is to implement more "limited" analyzers that are faster than the current analyzers (which are eager to get things fully correct at all costs). Ideas for making things faster include:

- lockfile-only analyzers
- only get a flat list of packages with metadata, but no dependency graph
- statically parse files instead of using the original build system
- leverage [information from deps.dev](https://github.com/oss-review-toolkit/ort/issues/4112)
- fully offline analyzers that require no network connection
- ...

References: https://github.com/oss-review-toolkit/ort/issues/4112#issuecomment-1131931210, https://github.com/oss-review-toolkit/ort/issues/5175#issuecomment-1076783293, https://github.com/oss-review-toolkit/ort/issues/8278, https://github.com/oss-review-toolkit/ort/issues/10129

heliocastro:
First of all, i think this is a dangerous approach.

One of the main reasons Ort being adopted more and more is because the accuracy and level of detailed information we can provide.
I know the cost and the burden to keep this, but indeed is why make the whole software simply better.

Going to the "easy path" of good enough, we would simply downplaying the entire project to the same level of other tools, including commercial ones, and then give enough ammunition to say anything over an OSS project, like the usual arguments.

Said that, there's actually a compromise that could be done in the terms of caching the analyser process similar way is done on the scan front.
Technically, unless we have some vector attach that replace a package in a registry with same version, the package not changes, therefore, not even the transient dependencies. We act over direct dependencies.
 
[...]

Issue #5626, opened 2022-08-01
Create a command line flag to specify the maximum depth of analysis


dgutson:
We sometimes need to get the 1st-level (direct) dependencies of a project.
I think that setting a --max-depth N could be helpful, N being 1 in the case of direct dependencies. It also could help to run ORT faster when resolving issues "iteratively" (eg by depth level): first level 1, then level 2, etc.
If maintainers don't think having the ability to set an arbitrary depth is useful, then just a --direct-deps-only flag could also be useful.

dgutson:
(FWIW we're currently working on this, so feel free to assign this to @qequ)

fviernau:
@dgutson 

Would you mind sharing the use case, e.g. for what purpose is the analysis result needed?

I'm asking this, since in case it is used in order to evaluate license compliance, then the definition of the `depth` may be a bit inaccurate. E.g. if you have a software project and you factor out a component X of that project into a separate library, then the direct of X, which have been first level before, become 2nd level dependencies of the project. So, the decision whether a package is reported or not would become dependent on the internal sub module structure of the project somewhat.

dgutson:
Hi @fviernau, sure.
I see what you mean. 
We have the "third party libraries" compliance audits, where we need to fill a number of data about the libraries _we_ use. This is what is under our direct control (choosing what libs to use), rather than modifying those external libs, so this is our action space. We could do this by post-processing the report, but this requires a script and wasting processing power and time by ORT running on libraries we don't need to be analyzed anyways (just the "first level"). Another CLI option could tell ORT where/which are the "internal" dependencies, so depth count could start outside of that set.

fviernau:
Hi @dgutson 

here's a list of approaches to consider, I tried to make it rather complete even if I don't really like some of these:

1. Add logic to each package manager integration to handle the `maxDepth`. I have doubts about this due to added complexity /   maintenance costs, in particular the cost / benefit ratio as this would IIUC not lead to a significant speed-up for several (most?)  the package managers.
2. Keep the package managers integrations as is, but make the analyzer command filter out the packages, after each respective 
    package manager integration determined the full dependency tree. This is cheap to maintain and reduces the size of the analyzer result, so the succeeding stages (scanner, evaluator, reporter) will speed-up and the report will contain just what you need.
3. Same as 2. but implement a separate command to filter a given analyzer result by depth, or maybe even highly customizable: pass a `.kts` script which decides whether a package needs to be filtered out.
[...]

Issue #2293, opened 2020-02-28
A Question about Restraining the Analyzer Depth 


CleanHit:
Is it possible to restrain the depth of the `analyser` (and `scanner` if needed) to the first level of seen package? For example with some additional `--max-depth` command line option, instead of going recursively into the sub dependencies of each package and again into their sub dependencies and so on......

CleanHit:
I understand that the replies in this post #2295 give an indirect answer to my question. However in some cases it is reasonable to limit the result to the first level of detected package dependencies. For example in case of creating a static contribution HTML for an application with the OSS Licenses that don't require the deeper knowledge of their recursive dependencies. 

If it is not possible to define an `--max-depth` option for the `Analyzer`, then I hope that the `analyzer-result.yml` file structure e.g.:

1. lib-dependency 1

    1. sub-dependency 1.1
        1. sub-dependency 1.1.1
        1. sub-dependency 1.1.2
    1. sub-dependency 1.2

2. lib-dependency 2

    1. sub-dependency 2.1
    1. sub-dependency 2.2
        1. sub-dependency 2.2.1
        1. sub-dependency 2.2.2

allows to minimize the information only to the lib-dependencies and create a smaller `analyzer-result.yml` file, that can be used by the `Scanner`, `Evaluator` and `Reporter`.

sschuberth:
> I understand that the replies in this post #2295 give an indirect answer to my question.

Correct, the Analyzer is always supposed to get the full picture.

So the correct way to implement the feature you're looking for would to introduce some new type of [Exclude](https://github.com/heremaps/oss-review-toolkit/blob/05fa56bd1989b549161bfee7763c3552dc28ea05/model/src/main/kotlin/config/Excludes.kt#L31) that allows to exclude packages based on certain properties, like their depth in the dependency tree.

However, personally I have mixed feelings about implementing a feature like this, because from a legal point of view it does not matter how deep a dependency occurs in your tree. You have to comply to its license obligations independently of the depth.

CleanHit:
I've went over my `analyzer-result.yml` file and the produced **cycloneDX** `bom.xml` again. And I've noticed that the packages where listed only once (per version) in the `bom.xml` file even if they appeared many times recursively (as a sub-dependency) in the `analyzer-result.yml` file. Therefore, the produced `bom.xml` file is sufficient to create those **static contribution HTMLs**.

sschuberth:
[...]

PR #8372, opened 2024-03-04
feat(scanner): add '--packages-depth' parameter.

fb33:
Sometimes we don't want to audit the whole dependencies of a project but only those declared in the project's package manager files.
To do this, I add a new parameter to the scanner to manage the depth in the dependency tree to scan.


fviernau:
We may consider to add a helper command to strip out dependencies for a given `depth` from an analyzer result as an alternative for supporting this use case. (That would not add complexity to the scanner)

sschuberth:
Thank for the contribution @fb33!

A basic issue I see with the implementation is related to ORT's transparency about what has been scanned: When only looking at the ORT scan result, how do you know whether only a subset of the packages have been scanned vs. really only the listed findings were present?

What I'm trying to say is: We need a way to record as part of the ORT result / in the scan result that only a subset (and which subset) of packages has been scanned. This could probably be solved as easily as adding the used depth (if `!= -1`) to the scan result.

But thinking further, if only scanning of e.g. direct dependencies is desired, should we also limit analysis to the configured depth already? That would automatically limit the scanning to those packages. Related issues: https://github.com/oss-review-toolkit/ort/issues/2293, https://github.com/oss-review-toolkit/ort/issues/5626, https://github.com/oss-review-toolkit/ort/issues/8361

tsteenbe:
This pull request was discussed in the ORT community meeting of March 21st, 2024.

The people attending can see the usefulness of this feature as:
- Limiting the result saves work and enable user to gradually improve their compliance over time e.g. first up to level 1 then 3, 6 etc.
- Don't want to do a full depth scan when open-sourcing code - you only care about first-level
- Limiting the scan will give user faster results which may be useful within running ORT in pull requests

However there are various concerns:
- Legally the depth of what is scanned is irrelevant. If OSS is included you have to respect its licenses
- End-to-end traceability, how can we make it visible in reports that not a full scan was done. We should implement a way to make ORT configurations visible. 
- Maintainability of the pull request
- Should we instead of CLI parameter implement this within config.yml

Question is if we should protect users against themselves? Given other options in ORT our believe we should not.

Issue #8359, opened 2024-02-29
`DEPENDS_ON` does not for to refer to external SPDX documents


sschuberth:
> Yes this works for inline documents when package relations are defined in the root/project document with DEPENDS_ON but it seems this is broken for external documents. I have changed the relations in analyzer/src/funTest/assets/projects/synthetic/spdx/libs/my-lib/package.spdx.yml to

```
relationships:
- spdxElementId: "SPDXRef-Package-my-lib"
  relationshipType: "DEPENDS_ON"
  relatedSpdxElement: "DocumentRef-curl-7.70.0:SPDXRef-Package-curl"
- spdxElementId: "SPDXRef-Package-my-lib"
  relationshipType: "DEPENDS_ON"
  relatedSpdxElement: "DocumentRef-openssl-1.1.1g:SPDXRef-Package-openssl"
- spdxElementId: "SPDXRef-Package-my-lib"
  relationshipType: "DYNAMIC_LINK"
  relatedSpdxElement: "DocumentRef-curl-7.70.0:SPDXRef-Package-curl"

```
and now the test for "retrieve transitive dependencies" fails

_Originally posted by @JSurf in https://github.com/oss-review-toolkit/ort/pull/4844#discussion_r1507168403_

JSurf:
Thanks for creating the issue.
To clarify:
DEPENDS_ON works fine when used in the "main" document but fails when packages are referenced with DEPENDS_ON in an external library document

Background:
We want to generate an aggregated report from multiple modules. For that we want to use ORT to create SPDX documents for every module first, then use one "main project" SPDX document to reference those submodules to generate an aggregated report. 

ORT spdx reporter generates DEPENDS_ON relationships in SPDXs. So in the aggregated report, no packages are found due to this bug. When manually changing the relations in the module SPDX to DEPENDENCY_OF the results contain all packages.

It does not matter if the packages itself are inline in the referenced module SPDX or also referenced from an external SPDX.

JSurf:
It seems the bug is even more simple and has nothing to do with referencing external documents and can be reproduced with a single document.

It seems DEPENDS_ON is only supported when the spdxElementId is the "root" ID (same as documentDescribes). 
For anything nested, it seems only DEPENDENCY_OF is supported.

[...]

PR #8376, opened 2024-03-05
fix(SpdxDocumentFile): Support nested DEPENDS_ON relations

JSurf:
This fixes nested DEPENDS_ON relations as discussed in:
https://github.com/oss-review-toolkit/ort/issues/8359

The main issue was that the dependsOnCase function is not always correctly defined.

There might be a better way to restructure the code and also support other nested relationships (if this is wanted) but my kotlin skills are very limited to do further improvements


sschuberth:
> The main issue was that the dependsOnCase function is not always correctly defined.

Thanks a lot for the contribution, @JSurf! I believe we should explain the above in more detail in the commit message. I propose to add:

```
While the `getDependencies()` function overload that takes the
`dependencyOfRelation` argument is generic with respect to the
relationship, whenever `DEPENDENCY_OF` is passed, also `dependsOnCase`
should be passed to handle dependency relationships bidirectionally. This
is exactly what the `getDependencies()` convenience function that only
takes three arguments is there for, so make use of that.

Also add a new test case to cover the `DEPENDS_ON` case.

Fixes #8359.
```

The last line will cause the issue to get auto-closed upon merge of the PR.

plugins/package-managers/spdx/src/funTest/kotlin/SpdxDocumentFileFunTest.kt
```diff
@@ -175,6 +175,25 @@ class SpdxDocumentFileFunTest : WordSpec({
             }
         }
 
+        "retrieve nested dependson dependencies" {
```
sschuberth, line 178: Nit: Let's use "DEPENDS_ON" instead of "dependson" here to resemble the enum name.

plugins/package-managers/spdx/src/funTest/assets/projects/synthetic/dependson-packages/project-xyz.spdx.yml
```diff
@@ -0,0 +1,92 @@
+SPDXID: "SPDXRef-DOCUMENT"
```
sschuberth, line 1: Nit: Let's use "DEPENDS_ON" instead of "dependson" in the path to resemble the enum name.

sschuberth: Commit message title nit: Consider putting "DEPENDS_ON" into Markdown-style backticks to format it as a keyword.

sschuberth (changes requested):

JSurf:
@sschuberth Made requested changes, please review again

PR #8379, opened 2024-03-06
feat(reporter): Sort license finding paths with localeCompare

AJDurant:
Use localeCompare to sort paths textually.

Currently paths are sorted by their length, which produces an output that is hard to use. It seems much more natural to sort paths textually.

Old behaviour:
image

New behaviour:
image


sschuberth (approved):

sschuberth:
Thanks @AJDurant!

PR #8387, opened 2024-03-08
fix(advisors): Use potentially customized PURLs in advisor queries

### LINKED ISSUE #8385: bug(VulnerableCode): In SPDX-based projects, no vulnerabilities are found
Given you have a SPDX-based project, with dependencies defined like as follows in _analyzer-result.yml:_

* id: "SpdxDocumentFile:wolfSSL Inc. (support@wolfssl.com):wolfSSL:" / purl: "pkg:conan/wolfssl@4.6.0"
* id: "SpdxDocumentFile:NanoPB, Inc.:nanopb:0.3.9.5" / purl: "pkg:pypi/nanopb@0.3.9.5"

If you use VulnerableCode as advisor, this will return known vulnerabilities, as for any other package type too.

But when processing the returned results from VulnerableCode, the code messes up.
See [this place in the code](https://github.com/oss-review-toolkit/ort/blob/d201f9ebed4bf9aa40b957140b8d32eaa511615a/plugins/advisors/vulnerable-code/src/main/kotlin/VulnerableCode.kt#L133)

A few lines above, a map is used to store the _affected_by_vulnerabilities_ items. The **purl** is used as key in the map. Example purls would be _pkg:conan/wolfssl@4.6.0_ or _pkg:pypi/nanopb@0.3.9.5._

But then, when the results are processed, a key of _pkg.id.toPurl()_ is used to access these items in the map.

Unfortunately, this _pkg.id.toPurl()_ do not match the **purls** that where used when the items where put into the map.

_pkg.id.toPurl()_ returns strange things like

* _pkg:generic/wolfSSL%20Inc.%20%28support%40wolfssl.com%29/wolfSSL@_
* _pkg:generic/NanoPB%2C%20Inc./nanopb@0.3.9.5_

For me, it seems more appropriate to use _pkg.purl_ instead of _pkg.id.toPurl()_ to access the items in the map.

sschuberth:
Do not regenerate the PURL from the package ID, but use the PURL that is already stored as part of the package as that might be a custom PURL, e.g. coming from a curation.

For VulnerableCode, this fixes a regression introduced the the refactoring in 70916bf.

Fixes #8385.

### DIFF
--- a/plugins/advisors/nexus-iq/src/main/kotlin/NexusIq.kt
+++ b/plugins/advisors/nexus-iq/src/main/kotlin/NexusIq.kt
@@ -42,7 +42,6 @@ import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.config.PluginConfiguration
 import org.ossreviewtoolkit.model.utils.PurlType
 import org.ossreviewtoolkit.model.utils.getPurlType
-import org.ossreviewtoolkit.model.utils.toPurl
 import org.ossreviewtoolkit.model.vulnerabilities.Vulnerability
 import org.ossreviewtoolkit.model.vulnerabilities.VulnerabilityReference
 import org.ossreviewtoolkit.utils.common.Options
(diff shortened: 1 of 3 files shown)


plugins/advisors/nexus-iq/src/main/kotlin/NexusIq.kt
mnonnenmacher, line 1: Commit message: "the the"  -> "by the"

mnonnenmacher (changes requested):

mnonnenmacher (approved):

PR #8390, opened 2024-03-08
Markdown one sentence per line

mnonnenmacher:
During reformatting I discovered a lot of minor issues with grammer or spelling, I will make a follow-up PR to fix those.


.github/workflows/static-analysis.yml
```diff
@@ -81,8 +81,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: Setup Node
```
sschuberth, line 84: > the `markdownlint-cli2-action` has no option to install custom rules.

I've reported https://github.com/DavidAnson/markdownlint-cli2-action/issues/134 about this.

.github/workflows/static-analysis.yml
```diff
@@ -81,8 +81,12 @@ jobs:
...
+        uses: actions/setup-node@v4
       - name: Check for Markdown issues
-        uses: DavidAnson/markdownlint-cli2-action@v15
+        run: |
+          npm install -g markdownlint-rule-max-one-sentence-per-line@0.0.2
```
sschuberth, line 88: How about omitting the version here, as you also don't specify one for `markdownlint-cli2`? Or vice-versa, specify it for `markdownlint-cli2`.

sschuberth (changes requested):

.github/workflows/static-analysis.yml
mnonnenmacher: As the rule is a not very active one man project I wanted to avoid any issues here by using fixed version, for markdownlint itself I'm not so concerned about upgrades.

sschuberth: Mind adding this rationale to the commit message then?

sschuberth (approved):

PR #8396, opened 2024-03-10
Small refactor docker build to solve race condition with functional tests

heliocastro:


heliocastro:
Different pull_request will be issued soon

PR #8400, opened 2024-03-11
feat(fossid-webapp): Mark files with all qualified snippets as identified

nnobelis:
If a file has ALL its snippet findings source locations ranges qualified in the .ort.yml file, either with ORIGINAL_FINDING, NO_RELEVANT_FINDING or a combination of both, it is marked as identified.

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -27,6 +27,7 @@ import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 import kotlin.time.toKotlinDuration
 
+import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
@@ -53,6 +54,7 @@ import org.ossreviewtoolkit.clients.fossid.listMatchedLines
 import org.ossreviewtoolkit.clients.fossid.listPendingFiles
 import org.ossreviewtoolkit.clients.fossid.listScansForProject
 import org.ossreviewtoolkit.clients.fossid.listSnippets
+import org.ossreviewtoolkit.clients.fossid.markAsIdentified
 import org.ossreviewtoolkit.clients.fossid.model.Project
 import org.ossreviewtoolkit.clients.fossid.model.Scan
 import org.ossreviewtoolkit.clients.fossid.model.result.MatchType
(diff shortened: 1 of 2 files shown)


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -922,4 +927,42 @@ class FossId internal constructor(
...
+    ) {
+        runBlocking(Dispatchers.IO) {
+            val candidatePathsToMark = snippetChoices.groupBy({ it.given.sourceLocation.path }) {
+                it.choice.reason
+            }
+
+            val requests = mutableListOf<Deferred<Any>>()
+
+            candidatePathsToMark.forEach { (path, reasons) ->
+                val hasNonChosenLocations = snippetFinding.any { path == it.sourceLocation.path }
```
sschuberth, line 949: Maybe use `none` instead of `any` (and rename the variable accordingly) to avoid the double-negation in the condition below?

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -922,4 +927,42 @@ class FossId internal constructor(
...
+
+                if (!hasNonChosenLocations) {
+                    if (listPendingFiles.none { path == it }) {
+                        logger.info { "Not marking $path as it it not pending." }
+                    } else {
+                        logger.info {
+                            "Marking $path as identified as it has the following snippet choices $reasons"
+                        }
+                        requests += async {
```
sschuberth, line 958: Nit: Empty line before this new block.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -922,4 +927,43 @@ class FossId internal constructor(
...
+            }
+
+            val requests = mutableListOf<Deferred<Any>>()
+
+            candidatePathsToMark.forEach { (path, reasons) ->
+                val allLocationsChosen = snippetFinding.none { path == it.sourceLocation.path }
+
+                if (allLocationsChosen) {
+                    if (listPendingFiles.none { path == it }) {
+                        logger.info { "Not marking $path as it it not pending." }
```
mnonnenmacher, line 953: `"Not marking $path as identified as it is not pending."`

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -922,4 +927,43 @@ class FossId internal constructor(
...
+                    } else {
+                        logger.info {
+                            "Marking $path as identified as it has the following snippet choices $reasons"
```
mnonnenmacher, line 956: I'm not sure if I understand that message, how about that?
`"Marking $path as identified as for all snippets a choice has been made. The used reasons are: $[reasons.joinToString()}"`

mnonnenmacher (changes requested):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: I adapted the sentence a bit further.

mnonnenmacher (approved):

Issue #8249, opened 2024-02-09
Improve support for different Java versions


mnonnenmacher:
Currently the Docker image and the artifacts published to Maven Central are both built with Java 17. This can cause two kinds of issues:

* Docker: As the Maven and Gradle analyzers are built into ORT and compiled with the same Java version, sometimes projects cannot be analyzed that require a different Java version. For example, the project could use a Gradle version that is not compatible with Java 17, or the project could require Java 21.
* Maven Central: Projects that use ORT programmatically as a library, like the ORT workbench, must be compiled with at least the same Java version as the ORT artifacts. However, for the same reason as above it could be required to build those tools with a different Java version.

One approach to solve this would be to release the Docker images and the Maven Central artifacts with different variants for each of the Java LTS releases, currently relevant are Java 11, 17, and 21.

Another more complex approach would be to decouple the Maven and Gradle analyzers from ORT itself and release them as standalone tools, comparable to the python-inspector. These tools could then be released with a bundled JVM for the different Java versions.

Also see some related PRs and issues: #4948, #6311, #8027, #8242

sschuberth:
I'm also thinking into the direction of simply installing multiple JDK versions in parallel in the Docker image, and finding a way to choose the JDK at runtime for the package manager implementations. At least for SBT that would work, as it's called as an external CLI tool. Also for Gradle projects that leverage the [toolchains](https://docs.gradle.org/current/userguide/toolchains.html) mechanism that should work. Not sure about Maven, though.

mnonnenmacher:
> Not sure about Maven, though.

I'm not sure if have seen the issue with Maven projects so far, maybe because it's more declarative than Gradle.

heliocastro:
Maven you can have in the properties:
```xml
    <java_source.version>17</java_source.version>
    <java_target.version>17</java_target.version>
```

But then, this is the model applied, so, if you are using java 21 to compile, it will apply only 17 characteristics. Again. sometimes not work well to the plugins.

sschuberth:
> Maven you can have in the properties

You have the same in Gradle, but these have no impact on the chosen JDK (if there are multiple ones on your system), but only on the command line options passed to the compiler.

sschuberth:
My current idea, at least for SBT for now, is to bootstrap the required JDK via https://github.com/foojayio/discoclient at ORT analyzer runtime.

fb33:
Hi, 
Just a feedback with jdk21.
[...]

PR #8403, opened 2024-03-12
build: Set the target JVM to 11

mnonnenmacher:
Configure Gradle to build class files compatible with Java 11. This allows to use the published ORT artifacts also in Projects that use Java 11. Previously, Java 17 was required.

The main use case for this is to be able to build the Analyzer Docker image for the ORT Server [1] with Java 11 which is required to analyze Gradle projects that are not compatible with newer Java versions.

Relates to #8249.

[1]: https://github.com/eclipse-apoapsis/ort-server

### DIFF
--- a/buildSrc/src/main/kotlin/ort-kotlin-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-kotlin-conventions.gradle.kts
@@ -111,8 +111,9 @@ detekt {
 }
 
 val javaVersion = JavaVersion.current()
-val maxKotlinJvmTarget = runCatching { JvmTarget.fromTarget(javaVersion.majorVersion) }
+val jvmSourceCompatibility =  runCatching { JvmTarget.fromTarget(javaVersion.majorVersion) }
     .getOrDefault(enumValues<JvmTarget>().max())
+val jvmTargetCompatibility = JvmTarget.JVM_11
 
 val mergeDetektReportsTaskName = "mergeDetektReports"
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/ort-kotlin-conventions.gradle.kts
```diff
@@ -111,8 +111,9 @@ detekt {
 }
 
 val javaVersion = JavaVersion.current()
-val maxKotlinJvmTarget = runCatching { JvmTarget.fromTarget(javaVersion.majorVersion) }
+val jvmSourceCompatibility =  runCatching { JvmTarget.fromTarget(javaVersion.majorVersion) }
     .getOrDefault(enumValues<JvmTarget>().max())
+val jvmTargetCompatibility = JvmTarget.JVM_11
```
mnonnenmacher, line 116: @sschuberth Does it make sense to keep the logic to determine the max version and use different versions for source and target or should we just hard code both to Java 11?

mnonnenmacher: @sschuberth Any chance we can get this change into tomorrow's release?

sschuberth: That's my goal, yes, and I'm on it. Just learned a few unexpected things about Gradle toolchains that require me to think about it a bit more.

sschuberth: See https://github.com/oss-review-toolkit/ort/pull/8411, which I propose to use instead of this PR.

PR #8409, opened 2024-03-13
feat(package-manager): Add initial support for Bazel

haikoschol:
This PR introduces a very limited Analyzer plugin for the Bazel build system. It only covers projects that use [bzlmod for managing dependencies](https://bazel.build/external/overview#bzlmod). Therefore it expects Bazel 7.0 or later, where bzlmod is the default.

So far the implementation has only been tested with [a small toy example project](https://github.com/haikoschol/bazel-android-py-cc-bzlmod/tree/main/cc).

A few things that are not covered, which very likely need to be added in subsequent iterations are:
* Patches that can be applied to the source artifact of dependencies during the build process (e.g. https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/glog/0.5.0/source.json#L4)
* Linking: Since this is C/C++, dynamic and static linking is possible. Right now all dependency relationships have static linkage because Bazel is generally biased towards static linking.
* Build targets are not taken into account. I'm not sure how this would look like yet, maybe as different scopes.
* There are no Bazel specific PURLs yet.


clients/bazel-module-registry/src/funTest/kotlin/BazelModuleRegistryServiceFunTest.kt
```diff
@@ -0,0 +1,52 @@
...
+package org.ossreviewtoolkit.clients.bazelmoduleregistry
+
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.collections.shouldContain
+import io.kotest.matchers.comparables.shouldBeGreaterThan
+import io.kotest.matchers.shouldBe
+
+import org.ossreviewtoolkit.model.HashAlgorithm
+
+class BazelModuleRegistryServiceTest : WordSpec({
```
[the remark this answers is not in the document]
sschuberth: Solve this by renaming the class to `BazelModuleRegistryServiceFunTest`.

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
```diff
@@ -0,0 +1,169 @@
...
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+
+@Serializer(URI::class)
+object URISerializer : KSerializer<URI> {
+    override fun serialize(encoder: Encoder, value: URI) = encoder.encodeString(value.toString())
+    override fun deserialize(decoder: Decoder) = URI(decoder.decodeString())
+}
+
+/**
```
fviernau, line 54: I propose to do the following with the "model" classes.

1. Move all model classes to the top / file level. Optionally into a separate dedicated file.
2. Rename ModuleMaintainer to just Maintainer and move it inside ModuleMetadata
3. Rename ModuleSource to ModuleSourceInfo to align with the naming of the API

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
```diff
@@ -0,0 +1,169 @@
...
+         * The service uses the Bazel Central Registry by default.
+         */
+        const val DEFAULT_URL = "https://bcr.bazel.build"
+
+        /**
+         * Create a Bazel Module Registry service instance for communicating with a server running at the given [url],
+         * defaulting to the Bazel Central Registry, optionally with a pre-built OkHttp [client].
+         */
+        fun create(
+            url: String = DEFAULT_URL,
```
fviernau, line 78: Should we use `url = String? = null` and internally fallback to the default value if null is provided? (IIRC this was the approach the code base has been moving towards)

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
```diff
@@ -0,0 +1,169 @@
...
+         */
+        private val JSON = Json {
+            ignoreUnknownKeys = true
+            namingStrategy = JsonNamingStrategy.SnakeCase
+        }
+
+        /**
+         * The service uses the Bazel Central Registry by default.
+         */
+        const val DEFAULT_URL = "https://bcr.bazel.build"
```
fviernau, line 71: Both above constants can be moved to top / file level and made private.

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
```diff
@@ -0,0 +1,169 @@
...
+    @Serializable
+    data class ModuleMetadata(
+        @Serializable(URISerializer::class) val homepage: URI?,
+        val maintainers: List<ModuleMaintainer>? = null,
+        val repository: List<String>? = null,
+        val versions: List<String>,
+        // The key in the map is the version, the value the reason for yanking it.
+        val yankedVersions: Map<String, String>,
+    ) {
+        fun vcsInfo(): VcsInfo {
```
fviernau, line 103: I have doubts this is a good location for this function. It maps the upstream model to ORT's model,
which I believe this should be place near by the code of the bazel analyzer.

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
```diff
@@ -0,0 +1,169 @@
...
+                type = VcsType.GIT,
+                url = url,
+                revision = "",
+                path = "",
+            )
+        }
+    }
+
+    @Serializable
+    data class ModuleMaintainer(
```
fviernau, line 128: Do you maybe have a link / reference to the upstream data model to add e.g. into KDoc for the data classes?

clients/bazel-module-registry/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+plugins {
+    // Apply precompiled plugins.
+    id("ort-library-conventions")
+
+    // Apply third-party plugins.
+    alias(libs.plugins.kotlinSerialization)
+}
+
+dependencies {
+    api(projects.model)
```
fviernau, line 31: Historically, IIRC the projects under `clients` had the goal to not have any ORT dependencies. I believe the goal has never been canceled. I've commented in the Kotlin code as well, that I believe mapping between upstream model and ORT's model better fits into the analyzer, then into the client module. If you do so, I believe we can get rid of this `model` dependency?

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
```diff
@@ -0,0 +1,169 @@
...
+object URISerializer : KSerializer<URI> {
+    override fun serialize(encoder: Encoder, value: URI) = encoder.encodeString(value.toString())
+    override fun deserialize(decoder: Decoder) = URI(decoder.decodeString())
+}
+
+/**
+ * Interface for a Bazel Module Registry, based on the directory structure of https://bcr.bazel.build
+ * and the git repository it is based on (https://github.com/bazelbuild/bazel-central-registry/).
+ */
+interface BazelModuleRegistryService {
```
fviernau, line 58: Do you think `BazelModuleRegistryClient` would be a naming improvement?

fviernau (changes requested):

mnonnenmacher (commented):
I haven't looked at the code, but please also add an entry to the package manager list in `website/docs/tools/advisor.md` and mention the limitations / mark it as experimental.

plugins/package-managers/bazel/src/funTest/assets/projects/synthetic/bzlmod-cc-has-lockfile-expected-output.yml
```diff
@@ -0,0 +1,276 @@
...
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  vcs_processed:
+    type: "Git"
+    url: "<REPLACE_URL_PROCESSED>"
+    revision: "<REPLACE_REVISION>"
+    path: "<REPLACE_PATH>"
+  homepage_url: ""
+  scopes:
+  - name: ""
```
fviernau, line 19: The scope should have a name.

sschuberth:
> I haven't looked at the code, but please also add an entry to the package manager list in `website/docs/tools/advisor.md` and mention the limitations / mark it as experimental.

@haikoschol that should say `website/docs/tools/analyzer.md`, obviously.

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
haikoschol: `DEFAULT_URL` is accessed in the Bazel plugin.

fviernau: Haven't reviewed second commit. Will do once comments in first one are addressed.

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
haikoschol: I moved them to the Analyzer plugin.

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
haikoschol: I added links to example files on the Bazel Central Registry.

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
haikoschol: Doesn't make a difference to me. I've renamed it to `BazelModuleRegistryClient`.

plugins/package-managers/bazel/src/main/kotlin/BazelModel.kt
```diff
@@ -0,0 +1,152 @@
...
+        return parts.getOrElse(1) { "" }
+    }
+}
+
+fun toVcsInfo(metadata: ModuleMetadata): VcsInfo {
+    if (metadata.repository.isNullOrEmpty()) {
+        return VcsInfo.EMPTY
+    }
+
+    val repo = metadata.repository!![0]
```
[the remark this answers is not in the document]
haikoschol: I don't know how to avoid this.

clients/bazel-module-registry/src/funTest/kotlin/BazelModuleRegistryClientFunTest.kt
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 2021 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
```
fviernau, line 2: 2024

plugins/package-managers/bazel/src/funTest/assets/projects/synthetic/bzlmod-cc-no-lockfile-expected-output.yml
```diff
@@ -0,0 +1,276 @@
...
+  vcs:
+    type: "Git"
+    url: "ssh://git@github.com/haikoschol/oss-review-toolkit.git"
+    revision: "0dbad696baf76543a45f7beb2cb64eb8a26a1ea5"
+    path: "plugins/package-managers/bazel/src/funTest/assets/projects/synthetic/bzlmod-cc-no-lockfile"
+  vcs_processed:
+    type: "Git"
+    url: "ssh://git@github.com/haikoschol/oss-review-toolkit.git"
+    revision: "0dbad696baf76543a45f7beb2cb64eb8a26a1ea5"
+    path: "plugins/package-managers/bazel/src/funTest/assets/projects/synthetic/bzlmod-cc-no-lockfile"
```
fviernau, line 16: `url`, `revision`, `path` need to use `REPLACE_*` patterns.

plugins/package-managers/bazel/src/funTest/kotlin/BazelFunTest.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import org.ossreviewtoolkit.analyzer.create
+import org.ossreviewtoolkit.analyzer.resolveSingleProject
+import org.ossreviewtoolkit.model.toYaml
+import org.ossreviewtoolkit.utils.test.getAssetFile
+import org.ossreviewtoolkit.utils.test.matchExpectedResult
+
+class BazelFunTest : StringSpec({
+    "Dependencies for a project with a lock file are detected correctly" {
+        val definitionFile = getAssetFile("projects/synthetic/bzlmod-cc-has-lockfile/MODULE.bazel")
```
fviernau, line 33: Would it make sense to rename the directory of that synthetic project e.g. `module-with-lockfile` or `project-with-lockfile`?

plugins/package-managers/bazel/src/funTest/kotlin/BazelFunTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        val expectedResultFile = getAssetFile("projects/synthetic/bzlmod-cc-has-lockfile-expected-output.yml")
+
+        val result = create("Bazel").resolveSingleProject(definitionFile)
+
+        result.toYaml() should matchExpectedResult(expectedResultFile, definitionFile)
+    }
+
+    "Dependencies for a project without a lock file are detected correctly" {
+        val definitionFile = getAssetFile("projects/synthetic/bzlmod-cc-no-lockfile/MODULE.bazel")
```
fviernau, line 42: Analog here, the project could be named `project-without-lockfile`. IMO it would align better with the test name and be a tad more human readable.

..The names of expected result files should be aligned then.

plugins/package-managers/bazel/src/funTest/kotlin/BazelFunTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        val expectedResultFile = getAssetFile("projects/synthetic/bzlmod-cc-no-lockfile-expected-output.yml")
+
+        val result = create("Bazel", allowDynamicVersions = true).resolveSingleProject(definitionFile)
+
+        result.toYaml() should matchExpectedResult(expectedResultFile, definitionFile)
+
+        // HACK Remove the lock file that was created by invoking Bazel above.
+        getAssetFile("projects/synthetic/bzlmod-cc-no-lockfile/MODULE.bazel.lock").delete()
```
fviernau, line 50: Why is this needed?

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,309 @@
...
+    override fun getVersionRequirement(): RangesList = RangesListFactory.create(">=7.0")
+
+    override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> {
+        val projectDir = definitionFile.parentFile
+        val lockFile = projectDir.resolve(BZLMOD_LOCK_FILE)
+        requireLockfile(projectDir) { lockFile.isFile }
+
+        val projectVcs = processProjectVcs(projectDir)
+        val moduleMetadata = Parser(definitionFile.readText()).parse()
+        val depDirectives = moduleMetadata.dependencies.associateBy { "${it.name}@${it.version}" }
```
fviernau, line 92: Is it possible that a single module (name) can occur in the dependency try with multiple times with different versions ?

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,309 @@
...
+    private fun getDependencyGraph(
+        projectDir: File,
+        depDirectives: Map<String, BazelDepDirective>,
+        lockFile: LockFile?
+    ): Set<Scope> {
+        if (lockFile != null) {
+            return parseLockFileGraph(lockFile.moduleDepGraph, depDirectives, managerName)
+        } else {
+            val modGraphProcess = run("mod", "graph", "--output", "json", workingDir = projectDir)
+            return parseBzlmodGraph(modGraphProcess.stdout, depDirectives, managerName)
```
fviernau, line 191: Can you help me understanding why you decided to use two code paths for construction the graph?

In particular, why isn't `mod graph` always used?

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,309 @@
...
+        ),
+        // Static linking is the default for cc_binary targets. According to the documentation, it is off for cc_test,
+        // but according to experiments, it is on.
+        // See https://bazel.build/reference/be/c-cpp#cc_binary.linkstatic
+        linkage = PackageLinkage.STATIC,
+        dependencies = dependencies
+    )
+}
+
+internal fun parseBzlmodGraph(
```
fviernau, line 266: How about naming this `parseModuleGraph` to avoid short forms and not repeat the term Bazel as that is clear from the context of the Kotlin module.

plugins/package-managers/bazel/src/main/kotlin/BazelModel.kt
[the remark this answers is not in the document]
sschuberth: Since `repository` is a nullable property (and also the list could be empty), I'd use

    val repo = metadata.repository?.firstOrNull()

and handle the `null`-case of `repo` accordingly in the code.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,309 @@
...
+
+    private fun getDependencyGraph(
+        projectDir: File,
+        depDirectives: Map<String, BazelDepDirective>,
+        lockFile: LockFile?
+    ): Set<Scope> {
+        if (lockFile != null) {
+            return parseLockFileGraph(lockFile.moduleDepGraph, depDirectives, managerName)
+        } else {
+            val modGraphProcess = run("mod", "graph", "--output", "json", workingDir = projectDir)
```
fviernau, line 190: Do edges in this Graph represent version constraints or actual (resolved) dependencies? (Asking as I'm a bit reminded of `GoMod` looking at the `cli` command)

fviernau (changes requested):

clients/bazel-module-registry/src/main/kotlin/BazelModuleRegistryService.kt
fviernau: Thanks!

haikoschol:
I realized that I added a quick & dirty way to install Bazel in the Dockerfile for testing purposes. I removed that from the latest commit on this branch and am currently working on adding it "properly" in a subsequent commit.

plugins/package-managers/bazel/src/funTest/kotlin/BazelFunTest.kt
haikoschol: Running `bazel mod graph --output json` creates the lock file as a side effect. If the file is not deleted, subsequent executions of the test won't actually test the code for analyzing projects without a lock file.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
haikoschol: It is not needed when a lock file is present, since that contains a dependency graph under `moduleDepGraph`. The usual approach in ORT as far as I can tell is to use the lock file when it is available.

fviernau: Hm, I'm not convinced from that yet. Do both approaches (lockfile parsing vs. module graph) yield the exact same information?

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
fviernau: The code could be simplified to associate the entries just by name. 
This would remove the (for me) open question, what happens in case of an empty version string.
E.g. could the current code swallow some entries by associating different meteadata with the same key in that case?

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
haikoschol: I don't know. AFAIK bzlmod uses the same dependency resolution approach as Go mod, i.e. finding the lowest version that satisfies all constraints. But I don't know in what scenario it actually picks versions. [The documentation on `bazel_dep`](https://bazel.build/rules/lib/globals/module#bazel_dep) just says:

> The version of the module to be added as a direct dependency.

It also says the default of this attribute is `''`, which to me suggests that the version is optional. However, when I try that I get an error:

```
$ cat MODULE.bazel
bazel_dep(name = "glog", repo_name = "com_github_google_glog")
bazel_dep(name = "googletest", version = "1.14.0", repo_name = "com_google_googletest", dev_dependency = True)

$ bazel mod graph --output json
ERROR: in module dependency chain <root> -> glog@_: bad bazel_dep on module 'glog' with no version. Did you forget to specify a version, or a non-registry override?. Type 'bazel help mod' for syntax and help.
```

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
haikoschol: The reason I'm reshuffling `moduleMetadata.dependencies` into this map is because Bazel uses the format `name@version` in the list of dependencies in the output of `bazel mod graph` and in the lock file. The map is used to split the direct dependencies into `main` and `dev` scope.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
haikoschol: They yield the same result for the dependencies that are declared with `bazel_dep` in `MODULE.bazel`, but not for all dependencies. The lock file contains more "Bazel internal" extra stuff that possibly nobody who uses ORT for Bazel will care about.

I'm happy to get rid of the code that extracts the dependency tree from the lock file.

fviernau: I'd appreciate if the lockfile code was removed then.

haikoschol:
I have no idea why this markdown-links check thinks https://bazel.build/ is a dead link.

haikoschol:
I made the egregious mistake of assuming the "update branch" button would do the right thing. Instead it introduced a merge commit, which AFAIK is extremely verboten in ORT. How do I fix this?

sschuberth:
> Instead it introduced a merge commit, which AFAIK is extremely verboten in ORT.

Exactly! 👮🏻‍♂️ 😆 

> How do I fix this?

Well, your local branch should still have your latest changes, so just force-push that to the remote branch for this PR. Afterwards, you can either click on "Update *with rebase*" from the drop-down, or just rebase locally and force-push once more.

clients/bazel-module-registry/src/funTest/kotlin/BazelModuleRegistryClientFunTest.kt
fviernau: This has been fixed in the wrong commit.

plugins/package-managers/bazel/src/main/kotlin/StarlarkParser.kt
```diff
@@ -0,0 +1,239 @@
...
+    val devDependency: Boolean
+)
+
+enum class TokenType {
+    IDENTIFIER, STRING, NUMBER, BOOLEAN, EQUALS, COMMA, LPAREN, RPAREN, LBRACKET, RBRACKET, EOF
+}
+
+data class Token(val type: TokenType, val value: String)
+
+class Lexer(private val input: String) {
```
fviernau, line 45: Is this code generated or manually written?

plugins/package-managers/bazel/src/main/kotlin/StarlarkParser.kt
```diff
@@ -0,0 +1,239 @@
...
+                        advance() // Skip anything else to avoid an endless loop.
+                    }
+                }
+            }
+        }
+        return Token(TokenType.EOF, "")
+    }
+}
+
+class Parser(input: String) {
```
fviernau, line 159: Please reduce visibility of all classes in this file to `internal` or `private` if possible.

plugins/package-managers/bazel/src/test/kotlin/BazelTest.kt
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+package org.ossreviewtoolkit.plugins.packagemanagers.bazel
+
+import io.kotest.core.spec.style.WordSpec
+import io.kotest.matchers.shouldBe
+
+import org.ossreviewtoolkit.utils.test.getAssetFile
+
+class BazelTest : WordSpec({
```
fviernau, line 27: Does this unit test cover things which are not already covered by the functional test?

fviernau (changes requested):

plugins/package-managers/bazel/src/main/kotlin/StarlarkParser.kt
haikoschol: It's generated.

plugins/package-managers/bazel/src/test/kotlin/BazelTest.kt
haikoschol: Maybe not any more. It was useful during development and would probably be useful again if the functions it tests need to be updated. Should I remove it? If yes, I assume the functions it tests should be made `private` as well.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.utils.common.CommandLineTool
+
+import org.semver4j.RangesList
+import org.semver4j.RangesListFactory
+
+import retrofit2.HttpException
+
+const val BZLMOD_LOCK_FILE = "MODULE.bazel.lock"
```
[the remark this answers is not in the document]
haikoschol: Renamed and made `private`.

clients/bazel-module-registry/src/funTest/kotlin/BazelModuleRegistryClientFunTest.kt
```diff
@@ -24,13 +24,13 @@ import io.kotest.matchers.collections.shouldContain
 import io.kotest.matchers.comparables.shouldBeGreaterThan
 import io.kotest.matchers.shouldBe
 
-class BazelModuleRegistryClientTest : WordSpec({
-    val service = BazelModuleRegistryClient.create()
+class BazelModuleRegistryClientFunTest : WordSpec({
```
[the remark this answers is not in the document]
haikoschol: Should be fixed now.

plugins/package-managers/bazel/src/main/kotlin/StarlarkParser.kt
fviernau: Shall we add a KDoc comment which states how it was generated? 

Is there any particular license which applies to the generated code?

plugins/package-managers/bazel/src/test/kotlin/BazelTest.kt
fviernau: In my view, with the current state of the tests to me they seem to be redundant and covered sufficiently by the fun test. However, if you or others really prefer having these, I'd be ok as well.

plugins/package-managers/bazel/src/main/kotlin/StarlarkParser.kt
haikoschol: The code was [generated by GPT-4](https://chat.openai.com/share/6cc4f4e1-9ac8-4eb7-8dd0-fd8476a38537). Should I add the link to that chat log to the code? Does ORT have a policy regarding code generated by an LLM?

sschuberth: > Does ORT have a policy regarding code generated by an LLM?

Argh, I wish you wouldn't have asked...

plugins/package-managers/bazel/src/test/kotlin/BazelTest.kt
haikoschol: I don't have a strong preference about keeping it or not, but I did think of one thing now that the functional tests don't cover: the function `transformBazelVersion`. However I'm not sure how useful that is. Especially since I just noticed that those tests are not even run as part of the CI. Did I fail to add them in some Gradle config file or is that expected behaviour?

fviernau (changes requested):

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+
+        val bmrClient = BazelModuleRegistryClient.create(lockFileData.registryUrl())
+        val packages = scopes.flatMap { getPackages(it.dependencies, bmrClient) }.filterNotNull()
+
+        return listOf(
+            ProjectAnalyzerResult(
+                project = Project(
+                    id = Identifier(
+                        type = managerName,
+                        name = moduleMetadata.module?.name ?: "root",
```
fviernau, line 106: As per `OrtResult` invariant the identifier is unique. In particular the fallback to `root` is problematic in this regard. In other places IIRC we utilize the definition file path to come to something unique.

problematic case: multiple definition files

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+                        version = moduleMetadata.module?.version.orEmpty(),
+                        namespace = ""
```
fviernau, line 108: Please re-order to: `type`, `namespace`, `name`, `version` for consistency.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+                    ),
+                    definitionFilePath = VersionControlSystem.getPathInfo(definitionFile).path,
+                    authors = emptySet(),
```
fviernau, line 111: Please avoid passing redundant default values, such as `authors`.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+                packages = packages.toSet()
+            )
+        )
+    }
+
+    private fun getPackages(
+        graph: Set<PackageReference>,
+        bmrClient: BazelModuleRegistryClient,
+        seen: MutableSet<Identifier> = mutableSetOf()
+    ): Set<Package?> {
```
fviernau, line 127: Is there a particular reason for using `Set<Package?>` instead of `Set<Package>`?

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+        val result = mutableSetOf<Package?>()
```
fviernau, line 128: You may utilize `buildSet` and turn this function into an extension function, e.g. 
`getPackages(..): Set<Package> = buildSet { ... }` 

This would separate concerns: 
1. Obtaining a set of identifiers from given package references
2. Creating a package for a given id

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+                    homepageUrl = "",
+                    scopeDependencies = scopes
+                ),
+                packages = packages.toSet()
+            )
+        )
+    }
+
+    private fun getPackages(
+        graph: Set<PackageReference>,
```
fviernau, line 124: IIUC the function only needs `Identifiers`. So, could `Set<PackageReference>` be changed to either `Set<Identifier>` or `Collection<Identifier>` ?

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
fviernau: Actually I believe it'd be nice to additionally separate also the retrieval of the `ModuleMetadata`, to arrive at something like:
```
val ids = packageRefs.....// some mapping code from the refs to the ids
val moduleMetadataForId = ids.map { getModuleMetedata(it) }
val packages = ids.map { id -> getPackage(id, moduleMetadataForId[id]) }
```

This would make it easier to parallelize the querying later on, because no shared mutable data structure is involved anymore. After writing noticed that moduleSourceInfo retrieval is missing and would need to be incorporated above in analog way.

plugins/package-managers/bazel/src/test/kotlin/BazelTest.kt
fviernau: Agree, for `transformBazelVersion` I believe the test actually adds value, it clarifies what the function does.
So, only removing `parseModuleGraph()` would be the option IMO.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+        return result
+    }
+
+    private fun packageFromPackageReference(ref: PackageReference, bmrClient: BazelModuleRegistryClient): Package? {
+        val metadata: ModuleMetadata
+        try {
+            metadata = runBlocking { bmrClient.getModuleMetadata(ref.id.name) }
+        } catch (e: HttpException) {
+            logger.warn { "Failed to fetch metadata for Bazel module ${ref.id.name}: $e" }
+            return null
```
fviernau, line 150: Should we return a `Package` with just the `id` instead of a `null` one, to not have dangling references in the result? (Same for the `null` return below).

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,221 @@
...
+            dependencies = mainDeps.map { parseModuleGraphNode(it, idRefType) }.toSet()
+        ),
+        Scope(
+            name = "dev",
+            dependencies = devDeps.map { parseModuleGraphNode(it, idRefType) }.toSet()
+        )
+    )
+}
+
+private fun parseModuleGraphNode(node: ModuleGraphNode, refType: String): PackageReference {
```
fviernau, line 199: Please turn this into an expression function.

fviernau (changes requested):

plugins/package-managers/bazel/src/test/kotlin/BazelTest.kt
haikoschol: Should I move the functions `parseModuleGraph` and `parseModuleGraphNode` into the class then? That would get rid of the parameter for setting `Identifier.type`.

fviernau: Makes sense to me.

haikoschol: I removed `parseModuleGraph` and moved the code into `getDependencyGraph`.

plugins/package-managers/bazel/src/main/kotlin/BazelModel.kt
```diff
@@ -0,0 +1,119 @@
...
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+
+private val json = Json { ignoreUnknownKeys = true }
+
+/**
+ * The data model for MODULE.bazel.lock.
+ */
+@Serializable
+internal data class LockFile(
```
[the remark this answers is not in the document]
haikoschol: I believe this is done now.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
haikoschol: The reason was that `packageFromPackageReference` may have returned `null`. I've changed it as you suggested to return a `Package` with only an `id`.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
haikoschol: Not directly related, but I just realized that there are now two data classes called `ModuleMetadata`. One in `StarklarkParser` and one in `BazelModuleRegistryClient`. Seems quite confusing. Do you have any thoughts on this?

haikoschol: > Actually I believe it'd be nice to additionally separate also the retrieval of the `ModuleMetadata`, to arrive at something like:
> 
> ```
> val ids = packageRefs.....// some mapping code from the refs to the ids
> val moduleMetadataForId = ids.map { getModuleMetedata(it) }
> val packages = ids.map { id -> getPackage(id, moduleMetadataForId[id]) }
> ```

I've made some changes that I believe implement what you suggested. I didn't use `buildSet` though.

I guess failure to retrieve data from the module registry should result in `Issue`s being created. However, I don't know how to implement that without using a shared mutable set that is either passed around or an instance variable.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,266 @@
...
+
+    private fun getPackage(id: Identifier, metadata: ModuleMetadata?, sourceInfo: ModuleSourceInfo?): Package {
+        return Package(
+            id = id,
+            declaredLicenses = emptySet(),
+            description = "",
+            homepageUrl = metadata?.homepage?.toString().orEmpty(),
+            binaryArtifact = RemoteArtifact.EMPTY,
+            sourceArtifact = sourceInfo?.remoteArtifact() ?: RemoteArtifact.EMPTY,
+            vcs = metadata?.vcsInfo() ?: VcsInfo.EMPTY
```
fviernau, line 142: Please use `VcsInfo.orEmpty()`.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    private fun getPackage(id: Identifier, metadata: ModuleMetadata?, sourceInfo: ModuleSourceInfo?): Package {
+        return Package(
+            id = id,
+            declaredLicenses = emptySet(),
+            description = "",
+            homepageUrl = metadata?.homepage?.toString().orEmpty(),
+            binaryArtifact = RemoteArtifact.EMPTY,
+            sourceArtifact = sourceInfo?.remoteArtifact() ?: RemoteArtifact.EMPTY,
```
fviernau, line 141: Please use `RemoteArtifact.orEmtpy()`.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,266 @@
...
+    private fun getPackages(scopes: Set<Scope>, registryUrl: String?): Set<Package> {
+        val registry = BazelModuleRegistryClient.create(registryUrl)
+        val identifiers = scopes.flatMap { getIdentifiers(it.dependencies) }
+        val moduleMetadataForId = identifiers.mapNotNull { getModuleMetadata(it, registry) }.toMap()
+        val moduleSourceInfoForId = identifiers.mapNotNull { getModuleSourceInfo(it, registry) }.toMap()
+        return identifiers.map { getPackage(it, moduleMetadataForId[it], moduleSourceInfoForId[it]) }.toSet()
+    }
+
+    private fun getPackage(id: Identifier, metadata: ModuleMetadata?, sourceInfo: ModuleSourceInfo?): Package {
+        return Package(
```
fviernau, line 135: Please turn this into an expression function.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,266 @@
...
+                packages = packages
+            )
+        )
+    }
+
+    private fun getPackages(scopes: Set<Scope>, registryUrl: String?): Set<Package> {
+        val registry = BazelModuleRegistryClient.create(registryUrl)
+        val identifiers = scopes.flatMap { getIdentifiers(it.dependencies) }
+        val moduleMetadataForId = identifiers.mapNotNull { getModuleMetadata(it, registry) }.toMap()
+        val moduleSourceInfoForId = identifiers.mapNotNull { getModuleSourceInfo(it, registry) }.toMap()
```
fviernau, line 130: I believe that `getModuleMetadata()` and `getModuleSourceInfo()` now should be changed to just return the `nullable` result, but not the `id` as a `Pair`.  The you can just do:
```
val moduleMetadataForId = identifiers.associate { getModuleMetadata(it, registry) }
```

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,266 @@
...
+                    scopeDependencies = scopes
+                ),
+                packages = packages
+            )
+        )
+    }
+
+    private fun getPackages(scopes: Set<Scope>, registryUrl: String?): Set<Package> {
+        val registry = BazelModuleRegistryClient.create(registryUrl)
+        val identifiers = scopes.flatMap { getIdentifiers(it.dependencies) }
```
fviernau, line 128: Consider `id` instead of `identifiers` to align with the two variable names directly below.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,266 @@
...
+        val moduleMetadataForId = identifiers.mapNotNull { getModuleMetadata(it, registry) }.toMap()
+        val moduleSourceInfoForId = identifiers.mapNotNull { getModuleSourceInfo(it, registry) }.toMap()
+        return identifiers.map { getPackage(it, moduleMetadataForId[it], moduleSourceInfoForId[it]) }.toSet()
```
fviernau, line 131: We commonly use `mapTo(mutbaleSetOf())` in such case, instead of using `toSet()` at the end.

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
```diff
@@ -0,0 +1,266 @@
...
+internal fun transformBazelVersion(output: String) =
+    output.lineSequence()
+        .find { it.startsWith("Build label") }
+        .orEmpty()
+        .split(":")
+        .getOrElse(1) { "" }
+        .trim()
+
+private fun ModuleMetadata.vcsInfo(): VcsInfo {
+    val repo = repository?.firstOrNull() ?: return VcsInfo.EMPTY
```
fviernau, line 235: Please use `VcsInfo.orEmpty()`.

fviernau (changes requested):

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
fviernau: I haven't considered the issue before. As it's not implemented, shall we leave this consideration aside for this iteration?

plugins/package-managers/bazel/src/main/kotlin/StarlarkParser.kt
fviernau: Shall we just add a comment to either the class or the file, which indicates that / how the file was generated?

plugins/package-managers/bazel/src/main/kotlin/Bazel.kt
haikoschol: > I haven't considered the issue before. As it's not implemented, shall we leave this consideration aside for this iteration?

Works for me.

plugins/package-managers/bazel/src/main/kotlin/StarlarkParser.kt
haikoschol: I've added a comment to the top of the file.

website/docs/tools/analyzer.md
```diff
@@ -13,6 +13,9 @@ It can be further processed or manually edited before passing it to one of the o
 Currently, the following package managers (grouped by the programming language they are most commonly used with) are supported:
 
 * C / C++
+  * [Bazel](https://bazel.build/) (**experimental**) (limitations:
+    bzlmod only, *very* basic parsing of `MODULE.bazel` files, patches to source artifacts are not included,
```
fviernau, line 17: Could we re-word the `bzlmod only` item a bit. I do not (as a person not into bazel) understand which limitation it refers to, e.g. what is missing? 

Also for the second item, could we make the text indicate what is missing? (basic parsing is not a limitation in case all needed data is available).

fviernau (changes requested):

fviernau:
I could probably find a couple of more nits, but I propose to draw a line here and merge it, once my last remaining open comments are addressed. (Some nits can still be fixed in future iterations) Is this ok with you as well @sschuberth @mnonnenmacher  ?

fviernau (dismissed):

Dockerfile
```diff
@@ -403,6 +403,25 @@ RUN mkdir -p $DOTNET_HOME/bin \
 FROM scratch AS dotnet
 COPY --from=dotnetbuild /opt/dotnet /opt/dotnet
 
+#------------------------------------------------------------------------
+# BAZEL
+FROM base as bazelbuild
+
+ARG BAZEL_VERSION=7.0.1
```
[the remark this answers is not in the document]
fviernau: @heliocastro would you review the docker related bits as well, I guess you know this `.versions` mechanism ;-)

website/docs/tools/analyzer.md
haikoschol: I've reworded the text and mentioned the limited testing.

Dockerfile
```diff
@@ -403,6 +403,25 @@ RUN mkdir -p $DOTNET_HOME/bin \
 FROM scratch AS dotnet
 COPY --from=dotnetbuild /opt/dotnet /opt/dotnet
 
+#------------------------------------------------------------------------
```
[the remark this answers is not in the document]
sschuberth: Yes, it's e.g. used as part of the `Jenkinsfile`.

sschuberth:
> I could probably find a couple of more nits, but I propose to draw a line here and merge it, once my last remaining open comments are addressed. (Some nits can still be fixed in future iterations) Is this ok with you as well @sschuberth @mnonnenmacher ?

I first wanted the discussion between you and @haikoschol to settle down, but now that it comes to an end, I'd like to do a brief review myself, too. But please approve nonetheless if you're happy. As auto-merge is not on, we can have more reviews after that.

Dockerfile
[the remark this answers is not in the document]
fviernau: So, `Dockerfile-legacy` should be updated in this PR / commit as well ?!

website/docs/tools/analyzer.md
fviernau: Thanks!

fviernau:
> I first wanted the discussion between you and @haikoschol to settle down, but now that it comes to an end, I'd like to do a brief review myself, too. But please approve nonetheless if you're happy. As auto-merge is not on, we can have more reviews after that.

Kotlin wise, it LGTM I agree with merge. Not approving now as there are open comments regarding the Docker bits. You may however start yourself already reviewing the Kotlin bits (I won't request further change on Kotlin side here).

Dockerfile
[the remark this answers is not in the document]
sschuberth: Ideally, yes. But going forward I'd like to avoid the need to maintain `Dockerfile-legacy` in addition to `Dockerfile`, and I hope @mnonnenmacher's Docker experiments to be the basis for such a change.

That said, I'm not sure whether it's worth to put effort into maintaining `Dockerfile-legacy` just now, as part of this initial implementation. I'd also be fine either adding Bazel support to `Dockerfile-legacy` later, or replacing the whole Docker stuff with a more streamlined solution, whatever comes first.

sschuberth:
@haikoschol, `import com.jakewharton.retrofit2.converter.kotlinx.serialization.asConverterFactory` now needs to be changed to `import retrofit2.converter.kotlinx.serialization.asConverterFactory`, see https://github.com/oss-review-toolkit/ort/pull/8440/files.

sschuberth:
Regarding the failing `markdown-links` check, please add an entry for https://bazel.build/ to https://github.com/oss-review-toolkit/ort/blob/main/mlc_config.json.

sschuberth:
Apparently `funTest-docker` still fail as [`bazel` cannot be found](https://github.com/oss-review-toolkit/ort/actions/runs/8369458047/job/22915230016?pr=8409#step:4:1351). I suspect this is due to the known problems with our current Docker / PR setup, which doe snot reliably use a Docker image build from changes within the same PR. Not sure if it's worth fixing that setup as part of this PR. But I'll at least ensure that the Bazel funTests will pass for me locally.

haikoschol:
> Apparently `funTest-docker` still fail as [`bazel` cannot be found](https://github.com/oss-review-toolkit/ort/actions/runs/8369458047/job/22915230016?pr=8409#step:4:1351). I suspect this is due to the known problems with our current Docker / PR setup, which doe snot reliably use a Docker image build from changes within the same PR. Not sure if it's worth fixing that setup as part of this PR. But I'll at least ensure that the Bazel funTests will pass for me locally.

@fviernau suggested to move the Docker changes to a separate PR that gets merged before. Should I give that a go?

sschuberth:
> @fviernau suggested to move the Docker changes to a separate PR that gets merged before. Should I give that a go?

If I can verify that funTest pass locally for me, I'd be fine with keeping things as-is. But I did not yet find the time to test, nor for a code review. I'll try today.

heliocastro:
The PR for fixing the race condition on funtest Docker is almost done, but the tests need be done in my private repo to avoid polute ort registry, since it involves upload docker images and version matching
Sorry for the delay, but i'm working on this in private time, with is diminished

sschuberth (approved):

PR #8411, opened 2024-03-13
Use Gradle toolchains to build ORT with a consistent Java version

sschuberth:


gradle.properties
```diff
@@ -23,6 +23,9 @@ org.gradle.parallel = true
 
 kotlin.code.style = official
 
+# The version of the JDK to use for building ORT.
+javaLanguageVersion = 11
```
sschuberth, line 27: BTW, I'd be fine with giving this another / shorter name. Maybe just `buildJdk`?

sschuberth:
Based on this, we could also do some simplifications regarding the GitHub workflows and Docker images, but I'll do these separately as a follow-up.

gradle.properties
mnonnenmacher: I'm fine with either name, but add `javaToolchainVersion` to the candidates.

mnonnenmacher (approved):

sschuberth:
> I'll do these separately as a follow-up.

See https://github.com/oss-review-toolkit/ort/pull/8414 for a start.

PR #8418, opened 2024-03-14
feat(cli): Use the resolved resolutions in `NotifierCommand`

mnonnenmacher:
If available, use only the resolved resolutions in the `NotifierCommand` instead of resolving them again. This increases consistency with previously executed ORT commands.

### DIFF
--- a/plugins/commands/notifier/src/main/kotlin/NotifierCommand.kt
+++ b/plugins/commands/notifier/src/main/kotlin/NotifierCommand.kt
@@ -77,11 +77,14 @@ class NotifierCommand : OrtCommand(
 
     override fun run() {
         val ortResult = readOrtResult(ortFile).mergeLabels(labels)
-        val notifier = Notifier(
-            ortResult,
-            ortConfig.notifier,
-            DefaultResolutionProvider.create(ortResult, resolutionsFile)
-        )
+
+        // If available, use only the resolved resolutions.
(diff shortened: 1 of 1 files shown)


plugins/commands/notifier/src/main/kotlin/NotifierCommand.kt
```diff
@@ -77,11 +78,14 @@ class NotifierCommand : OrtCommand(
...
     override fun run() {
         val ortResult = readOrtResult(ortFile).mergeLabels(labels)
-        val notifier = Notifier(
-            ortResult,
-            ortConfig.notifier,
-            DefaultResolutionProvider.create(ortResult, resolutionsFile)
-        )
+
+        // If available, use only the resolved resolutions.
+        val resolutionProvider: ResolutionProvider = when (ortResult.resolvedConfiguration.resolutions) {
```
sschuberth, line 83: Nit: I believe the type should not be needed.

sschuberth (changes requested):

plugins/commands/notifier/src/main/kotlin/NotifierCommand.kt
sschuberth: And now the import can be removed as well 😉

sschuberth (approved):

PR #8422, opened 2024-03-14
Simplify storage readers

mnonnenmacher:
Simplify the storage reader interfaces by merging the read functions, and apply several refactorings as clean up after switching to the new scanner implementation a while ago. This is a preparation for turning scan storages into plugins. See the commit messages for details.

### DIFF
--- a/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
+++ b/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
@@ -30,10 +30,10 @@ import org.ossreviewtoolkit.model.config.PluginConfiguration
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.scanner.CommandLinePathScannerWrapper
 import org.ossreviewtoolkit.scanner.ScanContext
+import org.ossreviewtoolkit.scanner.ScanStorage
 import org.ossreviewtoolkit.scanner.ScannerMatcher
 import org.ossreviewtoolkit.scanner.ScannerWrapperConfig
 import org.ossreviewtoolkit.scanner.ScannerWrapperFactory
-import org.ossreviewtoolkit.scanner.storages.ScanResultsStorage
 import org.ossreviewtoolkit.utils.common.Options
 import org.ossreviewtoolkit.utils.common.Os
 import org.ossreviewtoolkit.utils.common.ProcessCapture
@@ -52,7 +52,7 @@ import org.semver4j.Semver
  * configuration [options][PluginConfiguration.options]:
  *
  * * **"commandLine":** Command line options that modify the result. These are added to the [ScannerDetails] when
- *   looking up results from the [ScanResultsStorage]. Defaults to [ScanCodeConfig.DEFAULT_COMMAND_LINE_OPTIONS].
+ *   looking up results from a [ScanStorage]. Defaults to [ScanCodeConfig.DEFAULT_COMMAND_LINE_OPTIONS].
  * * **"commandLineNonConfig":** Command line options that do not modify the result and should therefore not be
  *   considered in [configuration], like "--processes". Defaults to
  *   [ScanCodeConfig.DEFAULT_COMMAND_LINE_NON_CONFIG_OPTIONS].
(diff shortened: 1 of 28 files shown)


scanner/src/main/kotlin/ScanResultsStorage.kt
```diff
@@ -74,27 +69,6 @@ abstract class ScanResultsStorage : PackageBasedScanStorage {
...
 
-    /**
-     * Return all [ScanResult]s contained in this [ScanResultsStorage] corresponding to the given [packages] that
-     * are [compatible][ScannerMatcher.matches] with the provided [scannerMatcher] wrapped in a [Result]. Also,
-     * [Package.sourceArtifact], [Package.vcs], and [Package.vcsProcessed] are used to check if the scan result matches
-     * the expected source code location. That check is important to find the correct results when different revisions
-     * of a package using the same version name are used (e.g. multiple scans of a "1.0-SNAPSHOT" version during
-     * development).
-     */
-    fun read(packages: Collection<Package>, scannerMatcher: ScannerMatcher): Result<Map<Identifier, List<ScanResult>>> {
```
sschuberth, line 103: Any comment on *why* this function is actually unused? I mean, as we're pretty much always interested in results of multiple packages, it would seem more efficient to *only* use this function in order to not apply the scanner matcher for every package we request.

plugins/scanners/scancode/src/main/kotlin/ScanCode.kt
```diff
@@ -30,10 +30,10 @@ import org.ossreviewtoolkit.model.config.PluginConfiguration
 import org.ossreviewtoolkit.model.config.ScannerConfiguration
 import org.ossreviewtoolkit.scanner.CommandLinePathScannerWrapper
 import org.ossreviewtoolkit.scanner.ScanContext
-import org.ossreviewtoolkit.scanner.ScanResultsStorage
 import org.ossreviewtoolkit.scanner.ScannerMatcher
 import org.ossreviewtoolkit.scanner.ScannerWrapperConfig
 import org.ossreviewtoolkit.scanner.ScannerWrapperFactory
+import org.ossreviewtoolkit.scanner.storages.ScanResultsStorage
```
sschuberth: Commit message: Typo in "packaage".

sschuberth (changes requested):

scanner/src/main/kotlin/ScanResultsStorage.kt
mnonnenmacher: I initially did not use it in the new scanner implementation for simplicity. Also, the new scanner converts package scan results into provenance scan results and reuses them if another package has the same provenance. And since the switch to the new scanner two years ago the downloading of stored scan results never seemed to be a performance problem so no one looked into optimizing the performance, e.g. by taking this function into use again.

sschuberth: Thanks for the details. I'll leave it up to you whether you think it's worth putting some of that into the commit message.

sschuberth (approved):

scanner/src/main/kotlin/ScanResultsStorage.kt
mnonnenmacher: I have added some of it to the commit message.

sschuberth: Just FYI @mnonnenmacher, we might have a use-case to add back the API for reading multiple packages at once: Is there is a general problem with some scanner backend, there is no point in continuing reading from it if the first read already fails, and we'd like to be able to fail all of the reads at once.

mnonnenmacher: Taking it back into use might require a bigger refactoring of the `Scanner`, losing some of the performance optimizations along the way, e.g. not downloading scan results multiple times for packages that share the same provenance.
If the goal is to disable a storage if a certain error occurred, that could be implemented also without introducing such a function.

sschuberth: > that could be implemented also without introducing such a function.

What do you have in mind @mnonnenmacher? Like a special `Result` value / exception being throw that signals the caller that no further packages should be tried?

mnonnenmacher: Yes, something like that. I think the tricky part would be to decide which kind of error should trigger this behavior. In most cases it might be better to fix the storage than to rescan everything, but that depends on the specific setup.

sschuberth: > I think the tricky part would be to decide which kind of error should trigger this behavior.

I believe this should simply be signaled via a new special exception that implementations may throw...

PR #8426, opened 2024-03-15
refactor(scanner): Rename `sourceCodeOriginsPriority`

fviernau:
Drop the `Priority`-suffix to align with the `sourceCodeOriginsProperty` in other places, e.g. in `Package`. Adjust the KDoc to make clear that the given list of source code origins is in order of priority.

While at it, re-word the comment a bit further, and re-align the analog part of the comment in the `PackageProvenanceResolver` interface with the one in the `DefaultPackageProvenanceResolver` class.

### DIFF
--- a/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
+++ b/scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
@@ -47,11 +47,12 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
  */
 interface PackageProvenanceResolver {
     /**
-     * Resolve the [KnownProvenance] of [pkg] based on the provided [defaultSourceCodeOriginsPriority].
+     * Resolve the [Provenance] of [pkg] based on [Package.sourceCodeOrigins] if specified, or else on
+     * [defaultSourceCodeOrigins]. Each source code origins are listed in order of priority.
      *
      * Throws an [IOException] if the provenance cannot be resolved.
      */
-    fun resolveProvenance(pkg: Package, defaultSourceCodeOriginsPriority: List<SourceCodeOrigin>): KnownProvenance
+    fun resolveProvenance(pkg: Package, defaultSourceCodeOrigins: List<SourceCodeOrigin>): KnownProvenance
 }
 
 /**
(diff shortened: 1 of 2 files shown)


scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
```diff
@@ -47,11 +47,12 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
  */
 interface PackageProvenanceResolver {
     /**
-     * Resolve the [KnownProvenance] of [pkg] based on the provided [defaultSourceCodeOriginsPriority].
+     * Resolve the [Provenance] of [pkg] based on [pkg.sourceCodeOrigins] if specified, or else on
```
sschuberth, line 50: `[pkg.sourceCodeOrigins]` -> `[Package.sourceCodeOrigins]`

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
```diff
@@ -47,11 +47,12 @@ import org.ossreviewtoolkit.utils.ort.showStackTrace
...
+     * [defaultSourceCodeOrigins], in order of priority.
```
sschuberth, line 51: IMO the way it's currently written, it's unclear what "in order of priority" (it could also refer to `Package.sourceCodeOrigins` taking precedence over `defaultSourceCodeOrigins`). I propose to start a new sentence saying e.g. "Each source code origins are listed in order of priority."

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
sschuberth: Similar below n the implementation.

scanner/src/main/kotlin/provenance/PackageProvenanceResolver.kt
sschuberth: Similar below n the implementation.

sschuberth (changes requested):

sschuberth (approved):

PR #8437, opened 2024-03-18
gradle: Misc. minor fixes

fviernau:


build.gradle.kts
```diff
@@ -187,7 +187,7 @@ val checkGitAttributes by tasks.registering {
                 .map { it.trim() }
                 .filter { it.isNotEmpty() && it.first() !in commentChars }
                 // The patterns are located before the first whitespace.
-                .mapTo(mutableSetOf()) { line -> line.takeWhile { !it.isWhitespace() } }
+                .mapTo(mutableSetOf()) { it.substringBefore(" ") }
```
mnonnenmacher, line 190: This is a bit different as it does not handle other whitespace characters like `\t`. But I don't know if this is relevant. (@sschuberth)

build.gradle.kts
```diff
@@ -186,7 +186,7 @@ val checkGitAttributes by tasks.registering {
                 // Skip empty and comment lines.
                 .map { it.trim() }
                 .filter { it.isNotEmpty() && it.first() !in commentChars }
-                // The patterns is the part before the first whitespace.
+                // The patterns are located before the first whitespace.
```
sschuberth, line 189: Actually, I believe this should get fixed in the opposite direction, and use singular "pattern" (because there can be only one), reading

> The pattern is the part before the first whitespace.

build.gradle.kts
sschuberth: It *is* relevant and breaks the tests, because tabs are used as spacers in that file.

sschuberth (changes requested):

fviernau:
Failing test was unrelated.

PR #8441, opened 2024-03-19
feat(fossid-webapp): Identify snippet choice entries that have been removed

nnobelis:
When one or several source locations of a 'marked as identified' file have been deleted from the snippet choice section of the .ort.yml file, put again the file to the 'pending' state. Its snippets will be then listed as for all pending files.
To find out if some locations have been removed, the comment in the marked as identified file is used to compare the location counts with the entries in the .ort.yml file.
Please note that FossID UI should not be used by the user to manually change the file state: The snippet choice feature is non-interactive and doing so would disturb the logic.

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -66,6 +66,7 @@ import org.ossreviewtoolkit.clients.fossid.model.rules.RuleType
 import org.ossreviewtoolkit.clients.fossid.model.status.DownloadStatus
 import org.ossreviewtoolkit.clients.fossid.model.status.ScanStatus
 import org.ossreviewtoolkit.clients.fossid.runScan
+import org.ossreviewtoolkit.clients.fossid.unmarkAsIdentified
 import org.ossreviewtoolkit.downloader.VersionControlSystem
 import org.ossreviewtoolkit.model.Issue
 import org.ossreviewtoolkit.model.LicenseFinding
@@ -328,7 +329,7 @@ class FossId internal constructor(
                             "false positives."
                     }
 
-                    val rawResults = getRawResults(scanCode)
+                    val rawResults = getRawResults(scanCode, snippetChoices?.choices.orEmpty())
                     createResultSummary(
                         startTime,
                         provenance,
@@ -820,7 +821,7 @@ class FossId internal constructor(
      * Get the different kind of results from the scan with [scanCode]
      */
     @Suppress("UnsafeCallOnNullableType")
-    private suspend fun getRawResults(scanCode: String): RawResults {
+    private suspend fun getRawResults(scanCode: String, snippetChoices: List<SnippetChoice>): RawResults {
         val identifiedFiles = service.listIdentifiedFiles(config.user, config.apiKey, scanCode)
             .checkResponse("list identified files")
             .data!!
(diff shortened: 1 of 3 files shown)


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -920,14 +925,25 @@ class FossId internal constructor(
...
+        val pendingFilesCount = (rawResults.listPendingFiles - newlyMarkedFiles.toSet()).size
+
+        issues.add(
+            0,
+            Issue(
+                source = name,
+                message = "This scan has $pendingFilesCount file(s) pending identification in FossID.",
+                severity = Severity.HINT
+            )
+        )
```
mnonnenmacher, line 945: Why not `issues += Issue(...)`?

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -326,6 +326,46 @@ private fun getLicenseFindingFromSnippetChoice(
...
+        }
+        val comment = markedAsIdentifiedFile.comments.values.firstOrNull {
+            it.comment.contains(ORT_NAME)
+        }?.runCatching {
+            jsonMapper.readValue(this.comment, OrtComment::class.java)
+        }?.onFailure {
+            logger.warn {
+                "File $markedFileName is marked as identified but it does have a valid comment ('${it.message}'). " +
+                    "It will be ignored."
+            }
```
mnonnenmacher, line 351: This seems wrong, `it` is the exception if `readValue` failed, not the `OrtComment`. Or should it say "does **not** have a valid comment"?

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -326,6 +326,46 @@ private fun getLicenseFindingFromSnippetChoice(
...
+        }?.getOrDefault(null)
+
+        when {
+            comment != null -> {
+                val snippetChoiceCount = snippetChoicesByName.count {
```
mnonnenmacher, line 356: `snippetChoicesCount` for consistency.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -326,6 +326,46 @@ private fun getLicenseFindingFromSnippetChoice(
...
+                    it.choice.reason == SnippetChoiceReason.ORIGINAL_FINDING
+                }
+                val locationWithFalsePositives = snippetChoicesByName.count {
```
mnonnenmacher, line 359: `notRelevantSnippetsCount` for consistency.

mnonnenmacher (changes requested):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: Because I would like to keep the same behavior as before the commit: this "special" issue should be the first of the issues.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: Yes I missed the _not_. I reworded the log message too.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: I also fixed the second variable.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: It would be cool if one could write `issues =+ Issue(...)` to achieve that.

mnonnenmacher: I think that does not work as intended, issues are sorted during serialization:

https://github.com/oss-review-toolkit/ort/blob/d7af7364be933aef0b9af20c66c58c7c03867b95/model/src/main/kotlin/ScanSummary.kt#L79-L85
https://github.com/oss-review-toolkit/ort/blob/d7af7364be933aef0b9af20c66c58c7c03867b95/model/src/main/kotlin/ScanSummary.kt#L151-L164

And they can also be sorted again by any reporter. Therefore I would not care about the order at this place.

mnonnenmacher (approved):

PR #8447, opened 2024-03-21
Code scanning cleanups

sschuberth:

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
@@ -148,7 +148,7 @@ private fun mapLicense(
         //       which cannot be parsed to an SpdxExpression. A better solution could be to automatically
         //       convert the strings into a form that can be parsed, then the mapping could be applied globally.
         LicenseFinding(license.mapLicense(detectedLicenseMapping), location)
-    }.onSuccess { licenseFinding ->
+    }.map { licenseFinding ->
         licenseFinding.copy(license = licenseFinding.license.normalize())
     }.onFailure { spdxException ->
         issues += FossId.createAndLogIssue(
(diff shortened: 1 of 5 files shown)


plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -148,7 +148,7 @@ private fun mapLicense(
         //       which cannot be parsed to an SpdxExpression. A better solution could be to automatically
         //       convert the strings into a form that can be parsed, then the mapping could be applied globally.
         LicenseFinding(license.mapLicense(detectedLicenseMapping), location)
-    }.onSuccess { licenseFinding ->
+    }.map { licenseFinding ->
```
sschuberth, line 151: @nnobelis please double-check.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
mnonnenmacher, line 1: Commit message: Typo in "onSuccess". But the message also confused me at first, it should say something like "`onSuccess` returns the original result, so the copied license finding is never used".

mnonnenmacher (changes requested):

mnonnenmacher (approved):

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: We discussed with @mnonnenmacher : it looks wrong and should have been picked up by the `FossIdLicenseMappingTest`. I will look at it in the future.

nnobelis: I found why the test was previously passing. Please see https://github.com/oss-review-toolkit/ort/pull/8495.

PR #8456, opened 2024-03-23
Attempt to fix funtest docker race condition

heliocastro:
To fix Ort docker functional tests, it was necessary to adapt the build to check strict versions of docker images built and add a special check when running the docker functional tests.

## Main changes
* Images are built using an external generic action, so maintenance is not anymore inse Ort
* Built images are always published with their correct version + sha
* Bazel image is included


heliocastro:
@mnonnenmacher This is the race condition solution

heliocastro:
Close this pull request as not possible upload images coming from external organization. A new PR is added from Ort organization.

PR #8458, opened 2024-03-24
Attempt to fix funtest docker race condition

heliocastro:
It was necessary to adapt the build to check strict versions of docker images built and add a special check when running the docker functional tests to fix Ort docker functional tests.

## Main changes
* Images are built using an external generic action, so maintenance is not anymore inside Ort code
* Built images are always published with their correct version + sha
* Bazel image is included


heliocastro:
@mnonnenmacher I needed to push the PR from inside Ort org, due permissions.
I will need help from you and @sschuberth I will need some help to understand why one of the tests is failing now.

mnonnenmacher:
@heliocastro To me this PR is going in the wrong direction, as it makes the already very complex Docker build even more complex. I would rather go in the opposite direction and just use a registry cache to improve the build time of the Docker image. I made some tests with that and in combination with the multi-stage build it looks very promising. Before I create a PR for that I would like to know if you address any other use cases than build performance with the current setup that I need to consider.

heliocastro:
> @heliocastro To me this PR is going in the wrong direction, as it makes the already very complex Docker build even more complex. I would rather go in the opposite direction and just use a registry cache to improve the build time of the Docker image. I made some tests with that and in combination with the multi-stage build it looks very promising. Before I create a PR for that I would like to know if you address any other use cases than build performance with the current setup that I need to consider.

Back from vacation.
We tried the registry cache before, but it also failed and led to size problems.
Remember that we need to rever the arm build? I discovered during the tests in this PR that the cause was loading the last stage to cache to try to avoid exactly what this PR does.

This PR is actually less complex than we have today. I offloaded the maintenance of the action of the image to an external one, and the images are really using the correct version, even on the test side.

We can bring back Arm buildings after this PR is merged since the out of disk space will not happen anymore ( hint: no caching on disk ).

I spent some time over this to clean up the exact problem of docker image versioning, so it is up to you if you see value in trying caching, but I can tell you ahead that you will probably find some new complexity there as well.

mnonnenmacher:
> Back from vacation. We tried the registry cache before, but it also failed and led to size problems.

What kind of size problems?

> Remember that we need to rever the arm build? I discovered during the tests in this PR that the cause was loading the last stage to cache to try to avoid exactly what this PR does.

I thought the problem was related to the webpack build of the web-app reporter?
Which cache do you refer to in "loading the last stage to cache", the one of the local Docker daemon?

> This PR is actually less complex than we have today. I offloaded the maintenance of the action of the image to an external one, and the images are really using the correct version, even on the test side.

Moving the code to another repository does not reduce the overall complexity and looking at the workflow diffs they definitely become more complex, for example as they now need to figure out which workflow is running the functional tests.

> I spent some time over this to clean up the exact problem of docker image versioning, so it is up to you if you see value in trying caching, but I can tell you ahead that you will probably find some new complexity there as well.

Ok, I'll do some more tests then and share the outcome.

mnonnenmacher:
> Ok, I'll do some more tests then and share the outcome.

My tests have been successful, for the implementation and description see #8527.

heliocastro:
Too bad I didnt have time to review, right. Seen like my expertise worth
nothing

sschuberth:
@heliocastro, your opinion on the matter is still appreciated, and as I've mentioned in the past to other persons as well, also after-the-fact reviews and reviews from domain experts, even if they were not committers, are appreciated. So feel free to add comments to #8527 that we could address as follow-ups (like I myself did as well).

That said, the elegance in @mnonnenmacher's solution IMO lies in the fact that it vastly simplifies the code and does not require additional Python scripting, neither internal nor external to this repository. He's been testing this for weeks in his own fork, and we finally needed a maintainable solution for the problem. So I believe the quick merge was justified.

heliocastro:
Sorry, this  sound a little strange, as usually most of us non core developers are stopped for days, sometime weeks to "wait for an expert review" and subject to a hard scrutiny.
Which  would be ok if not the case of double standards happens, like these PR.
In one day, with no questions asked except with minor fixes, was approved, and no one else could have an opinion over that.  Not even the person that knows the pre work better than the approver.
I'm ok with that, this opens better the way you see how contributors need perceive the project governance acceptance.

sschuberth:
Closing this as @mnonnenmacher's solution is in place now.

PR #8463, opened 2024-03-27
fix(gradle-plugin): Do not fail with NPE when dependency POMs are missing

brigaccess:
This commit fixes an issue I reported earlier in Slack conversation. Sometimes, POM cannot be resolved by ORT for one reason or another. When this happens, the analysis finishes early, omitting dependencies that have POMs but were unlucky to be listed after the POMless one.

The fix itself is simple: instead of requiring a non-null value in POMs map (through `getValue`), add a warning and leave the fields empty when the returned POM is null.

In my particuar case, ORT fails scanning [1] and [2] without the patch due to Gradle IntelliJ Plugin [3] injecting dependencies without POMs.

[1] https://github.com/JetBrains/Grammar-Kit
[2] https://github.com/JetBrains/ideavim
[3] https://github.com/JetBrains/intellij-platform-gradle-plugin

### DIFF
--- a/plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
+++ b/plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
@@ -198,7 +198,13 @@ internal class OrtModelBuilder : ToolingModelBuilder {
                                 null
                             }
 
-                            val modelBuildingResult = poms.getValue(id.toString())
+                            val modelBuildingResult = poms[id.toString()]
+                            if (modelBuildingResult == null && id !in visitedDependencies) {
+                                val message = "No POM found for $id"
+                                logger.warn(message)
+                                warnings += message
+                            }
+
                             val dependencies = if (id in visitedDependencies) {
                                 // Cut the graph on cyclic dependencies.
                                 emptyList()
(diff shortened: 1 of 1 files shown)


brigaccess:
Fixed detekt issue

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -207,23 +207,31 @@ internal class OrtModelBuilder : ToolingModelBuilder {
                                 selectedComponent.dependencies.toOrtDependencies(poms)
                             }
 
+                            if (modelBuildingResult == null) {
+                                val message = "POM is missing for $id"
+                                logger.warn(message)
+                                warnings += message
+                            }
+
```
sschuberth, line 215: Let's move this directly after line 201 to handle the `null` case where may occur.

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -207,23 +207,31 @@ internal class OrtModelBuilder : ToolingModelBuilder {
...
                                 warning = null,
                                 pomFile = pomFile,
-                                mavenModel = OrtMavenModelImpl(
-                                    licenses = modelBuildingResult.effectiveModel.collectLicenses(),
-                                    authors = modelBuildingResult.effectiveModel.collectAuthors(),
-                                    description = modelBuildingResult.effectiveModel.description.orEmpty(),
-                                    homepageUrl = modelBuildingResult.effectiveModel.url.orEmpty(),
-                                    vcs = modelBuildingResult.getVcsModel()
-                                ),
+                                mavenModel = modelBuildingResult?.let {
```
sschuberth, line 226: Maybe use `run` instead of `let` to avoid the need for `it.` inside the block.

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
```diff
@@ -207,23 +207,31 @@ internal class OrtModelBuilder : ToolingModelBuilder {
                                 selectedComponent.dependencies.toOrtDependencies(poms)
                             }
 
+                            if (modelBuildingResult == null) {
+                                val message = "POM is missing for $id"
```
sschuberth, line 211: Let's phrase this like "No POM found for $id."

sschuberth (changes requested):
Some additional commit message nits: We prefer passive voice in commit messages, so no mention of I / my / we etc.

> This commit fixes an issue I reported earlier in Slack conversation:

I'd simply drop this sentence to avoid the use of non-passive voice, and the mention of the Slack conversation is not super useful unless also a link would be added (but that could expire with our free plan of Slack).

> In my particuar case

Just say "In this specific case...".

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
brigaccess: Come to think of it, I suppose it should also check if the dependency ID is not in `visitedDependencies` to avoid having multiple warnings?

sschuberth: Yeah, extending the check to `if (modelBuildingResult == null && id !in visitedDependencies)` sounds like a good idea.

sschuberth (approved):

plugins/package-managers/gradle-plugin/src/main/kotlin/OrtModelBuilder.kt
sschuberth: Thanks for your contribution!

PR #8470, opened 2024-03-28
Upgrade ScanCode to version 32.1.0

sschuberth:


plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModel.kt
mnonnenmacher, line 1: Commit message: Could you explain the change a bit more detailed? IIUC license findings that are references to license findings in other files are now ignored, because they already appear as findings for those other files?

mnonnenmacher (changes requested):

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModel.kt
sschuberth: Exactly. I've used your wording for the commit message.

fviernau: I've added another comment to the ScanCode issue, it could be considered here (and in the commit message) as well: https://github.com/nexB/scancode-toolkit/issues/3648#issuecomment-2032239964

mnonnenmacher (approved):

plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModel.kt
fviernau: > Note that keeping findings that are in fact references is not currently supported by the ORT data model. That's because line ranges always have to refer to the file from the finding, not a references file. 

This indicates to me that you likely didn't get my point. Consider this:

1. In older ScanCode versions we had `License-scancode-unkown-license-reference` findings
    with start / endLine pointing to the scanned file.
2. In newer ScanCode version findings such as 1. do not appear, but they are replaced with the licenses 
    detected in the referenced file.

This commit now filters out the findings. So, compared to older ScanCode version scan results the `License-scancode-unkown-license-reference` findings are missing. I think this is undesired and should be avoided.

sschuberth: I've asked for clarification [over here](https://github.com/nexB/scancode-toolkit/issues/3648#issuecomment-2033662900).

PR #8471, opened 2024-03-28
fix(maven): return coordinates-only `Package` on `parsePackage` exception

brigaccess:
When analyzing a Maven project with a dependency on `opensymphony:quartz:1.6.0`, ORT does not include it in the packages list of the analyzer, but it is still present in the `dependency_graphs.Maven.packages`. The reason for it turned out to be a lack of POM for the dependency in the Central (see [1]) causing an exception in `MavenSupport.parsePackage` call.

This PR fixes the issue by returning a coordinates-only `Package` from `getOrElse` block instead of `null`. Another way to fix this would be to return that `Package` from the `MavenSupport.parsePackage` itself. However, `GradleDependencyHandler` seems to rely on the exception thrown by it, and other (potential) consumers might want to do that as well.

[1] https://repo.maven.apache.org/maven2/opensymphony/quartz/1.6.0/

### DIFF
--- a/plugins/package-managers/maven/src/main/kotlin/utils/MavenDependencyHandler.kt
+++ b/plugins/package-managers/maven/src/main/kotlin/utils/MavenDependencyHandler.kt
@@ -28,6 +28,8 @@ import org.ossreviewtoolkit.model.Identifier
 import org.ossreviewtoolkit.model.Issue
 import org.ossreviewtoolkit.model.Package
 import org.ossreviewtoolkit.model.PackageLinkage
+import org.ossreviewtoolkit.model.RemoteArtifact
+import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.createAndLogIssue
 import org.ossreviewtoolkit.model.utils.DependencyHandler
 import org.ossreviewtoolkit.plugins.packagemanagers.maven.Maven
@@ -112,7 +114,24 @@ class MavenDependencyHandler(
                 message = "Could not get package information for dependency '" +
                     "${dependency.artifact.identifier()}': ${e.collectMessages()}"
             )
-        }.getOrNull()
+        }.getOrElse {
+            Package(
+                id = dependency.artifact.run {
+                    Identifier(
+                        type = "Maven",
+                        namespace = groupId,
+                        name = artifactId,
+                        version = version
+                    )
+                },
+                binaryArtifact = RemoteArtifact.EMPTY,
+                declaredLicenses = emptySet(),
+                description = "",
+                homepageUrl = "",
+                sourceArtifact = RemoteArtifact.EMPTY,
+                vcs = VcsInfo.EMPTY
+            )
+        }
     }
 
     /**
--- a/plugins/package-managers/maven/src/test/kotlin/utils/MavenDependencyHandlerTest.kt
+++ b/plugins/package-managers/maven/src/test/kotlin/utils/MavenDependencyHandlerTest.kt
@@ -24,6 +24,7 @@ import io.kotest.matchers.collections.beEmpty
 import io.kotest.matchers.collections.containExactly
 import io.kotest.matchers.collections.haveSize
 import io.kotest.matchers.nulls.beNull
+import io.kotest.matchers.nulls.shouldNotBeNull
 import io.kotest.matchers.should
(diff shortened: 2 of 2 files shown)


plugins/package-managers/maven/src/main/kotlin/utils/MavenDependencyHandler.kt
```diff
@@ -112,7 +114,24 @@ class MavenDependencyHandler(
...
+            Package(
+                id = dependency.artifact.run {
+                    Identifier(
+                        type = "Maven",
+                        namespace = groupId,
+                        name = artifactId,
+                        version = version
+                    )
+                },
+                binaryArtifact = RemoteArtifact.EMPTY,
```
sschuberth, line 127: I wonder why we shouldn't be able to at least get the binary artifact here... I mean, apparently software with a dependency on `opensymphony:quartz:1.6.0` *does* build, so the binary can be retrieved, and we should be able to get its coordinates here.

Maybe instead of this fallback here, can we narrow down which part of `parsePackage()` fails, and try to get other data from it on a more fine-granular basis?

plugins/package-managers/maven/src/test/kotlin/utils/MavenDependencyHandlerTest.kt
```diff
@@ -219,7 +222,24 @@ class MavenDependencyHandlerTest : WordSpec({
...
+                namespace shouldBe artifact.groupId
+                name shouldBe artifact.artifactId
+                version shouldBe artifact.version
+            }
+
+            with(pkg) {
+                binaryArtifact shouldBe RemoteArtifact.EMPTY
+                declaredLicenses should beEmpty()
+                description shouldBe ""
+                homepageUrl shouldBe ""
```
sschuberth, line 239: This should be `should beEmpty()` similar to like one line above, but it requires to import either version via an alias to avoid a name clash. Probably import the line above like `import io.kotest.matchers.collections.beEmpty as beEmptyCollection`.

plugins/package-managers/maven/src/test/kotlin/utils/MavenDependencyHandlerTest.kt
```diff
@@ -219,7 +222,24 @@ class MavenDependencyHandlerTest : WordSpec({
...
-            handler.createPackage(dependency, issues) should beNull()
+            val pkg = handler.createPackage(dependency, issues)
+            pkg.shouldNotBeNull()
+
+            with(pkg.id) {
+                type shouldBe "Maven"
+                namespace shouldBe artifact.groupId
+                name shouldBe artifact.artifactId
+                version shouldBe artifact.version
+            }
```
sschuberth, line 233: Maybe nest this in the below?

sschuberth (changes requested):

plugins/package-managers/maven/src/main/kotlin/utils/MavenDependencyHandler.kt
brigaccess: >can we narrow down which part of parsePackage() fails

The Maven call ([MavenSupport.kt#698](https://github.com/oss-review-toolkit/ort/blob/4f32b500b2d29404d2260c191f25f4aea09d5fa3/plugins/package-managers/maven/src/main/kotlin/utils/MavenSupport.kt#L698) => [MavenSupport.kt#715](https://github.com/oss-review-toolkit/ort/blob/4f32b500b2d29404d2260c191f25f4aea09d5fa3/plugins/package-managers/maven/src/main/kotlin/utils/MavenSupport.kt#L715)) raises `ProjectBuildingException`. Not sure if it's reasonable to break the contract of `parsePackage` rethrowing these. Same goes for `requestRemoteArtifact` in `MavenDependencyHandler` (given that it is a private method).

brigaccess: >try to get other data from it on a more fine-granular basis

I've checked the exception and it doesn't seem like there's much to extract from it, except for the fact that the cause is `ArtifactResolutionException`. This could be used to make a conditional that tries to fetch the binary, I guess?

Another way would be to try and fetch the remote artifact before the POM, but it's probably a bad idea given how rare the case of POMless dependency is.

PR #8476, opened 2024-04-02
refactor: Use Kotest's own `tempdir()` in tests

sschuberth:
These are cleaned up automatically.

### DIFF
--- a/helper-cli/src/funTest/kotlin/commands/CreateAnalyzerResultFromPackageListCommandFunTest.kt
+++ b/helper-cli/src/funTest/kotlin/commands/CreateAnalyzerResultFromPackageListCommandFunTest.kt
@@ -22,6 +22,7 @@ package org.ossreviewtoolkit.helper.commands
 import com.github.ajalt.clikt.testing.test
 
 import io.kotest.core.spec.style.WordSpec
+import io.kotest.engine.spec.tempdir
 import io.kotest.matchers.shouldBe
 
 import org.ossreviewtoolkit.helper.HelperMain
@@ -29,14 +30,13 @@ import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.ResolvedConfiguration
 import org.ossreviewtoolkit.model.readValue
 import org.ossreviewtoolkit.utils.ort.Environment
-import org.ossreviewtoolkit.utils.ort.createOrtTempDir
 import org.ossreviewtoolkit.utils.test.getAssetFile
 
 class CreateAnalyzerResultFromPackageListCommandFunTest : WordSpec({
     "The command" should {
         "generate the expected analyzer result file" {
             val inputFile = getAssetFile("package-list.yml")
-            val outputFile = createOrtTempDir().resolve("analyzer-result.yml")
+            val outputFile = tempdir().resolve("analyzer-result.yml")
             val expectedOutputFile = getAssetFile("create-analyzer-result-from-pkg-list-expected-output.yml")
 
             HelperMain().test(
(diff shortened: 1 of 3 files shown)


helper-cli/src/funTest/kotlin/commands/CreateAnalyzerResultFromPackageListCommandFunTest.kt
```diff
@@ -22,21 +22,21 @@ package org.ossreviewtoolkit.helper.commands
 import com.github.ajalt.clikt.testing.test
 
 import io.kotest.core.spec.style.WordSpec
+import io.kotest.engine.spec.tempdir
```
fviernau, line 25: Could the remaining two places also be migrated?
* `scanner/src/test/kotlin/ScannerTest.kt`
* `scanner/src/test/kotlin/utils/FileListResolverTest.kt`

fviernau (changes requested):

helper-cli/src/funTest/kotlin/commands/CreateAnalyzerResultFromPackageListCommandFunTest.kt
sschuberth: I had investigated these already, and they should stay as-is:

In `ScannerTest.kt`, while the `FakeProvenanceDownloader` deliberately dose not delete the temporary directory, the wrapping scanner does.

And in `FileListResolverTest.kt`, there is an explicit test that the directory already gets deleted, see

https://github.com/oss-review-toolkit/ort/blob/48ae81634c5a3bd4d55e89d2474dc05825a5ae03/scanner/src/test/kotlin/utils/FileListResolverTest.kt#L64-L80

fviernau: I'm not sure if I follow the rationale. Which of the following do you mean?

1. migration is not needed because production code takes care of deleting
2. migration is not possible because it would break something or invalidate the test case.

In case of 1. should we migrate anyway, just for consistency?

sschuberth: It's 1., plus deletion would defeat the purpose of the test (though not cause an error), so IMO these should stay as-is.

fviernau: Why would deletion defeat the purpose of the test? To my understanding the kotest function would delete the dir only after the spec has been entirely executed and assertion about deletion has been made ?

sschuberth: Ok, "defeat the purpose of the test" probably was the wrong wording.

My point for `FakeProvenanceDownloader` is that it should be implemented like any other "real" `ProvenanceDownloader` and not make use of test-only functions as part of its implementation. `tempdir()` cannot be called from `FakeProvenanceDownloader` anyway as the function requires `TestConfiguration` context. And `ProvenanceDownloader.download()` documents "The caller is responsible for deleting the directory", so it really should not delete the directory itself.

For `createTempDirWithFiles()` in `FileListResolverTest` there would be no harm in using `tempdir()`, but it's completely unnecessary as there already is `dir shouldNotBe aDirectory()` as part of the test.

sschuberth: > For `createTempDirWithFiles()` in `FileListResolverTest` there would be no harm in using `tempdir()`

Ok, I did that now just for consistency, like you requested.

fviernau (approved):

PR #8493, opened 2024-04-05
Cargo refactorings

sschuberth:


plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -198,8 +201,8 @@ class Cargo(
 
         val scopes = setOfNotNull(
             getTransitiveDependencies(groupedDependencies[""], "dependencies"),
```
fviernau, line 192: Should `""` also be extracted to a constant for conistency?

sschuberth: That will get removed in some more upcoming refactorings, so I left it as-is.

fviernau (approved):

PR #8497, opened 2024-04-08
More Cargo improvements

sschuberth:

### DIFF
--- a/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
+++ b/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
@@ -49,7 +49,6 @@ import org.ossreviewtoolkit.utils.common.splitOnWhitespace
 import org.ossreviewtoolkit.utils.common.unquote
 import org.ossreviewtoolkit.utils.common.withoutPrefix
 import org.ossreviewtoolkit.utils.ort.DeclaredLicenseProcessor
-import org.ossreviewtoolkit.utils.ort.ProcessedDeclaredLicense
 import org.ossreviewtoolkit.utils.spdx.SpdxConstants
 import org.ossreviewtoolkit.utils.spdx.SpdxOperator
 
@@ -234,16 +233,14 @@ private fun parseDeclaredLicenses(pkg: CargoMetadata.Package): Set<String> {
     return declaredLicenses
 }
 
-private fun processDeclaredLicenses(licenses: Set<String>): ProcessedDeclaredLicense =
+private fun parsePackage(pkg: CargoMetadata.Package, hashes: Map<String, String>): Package {
+    val declaredLicenses = parseDeclaredLicenses(pkg)
+
     // While the previously used "/" was not explicit about the intended license operator, the community consensus
     // seems to be that an existing "/" should be interpreted as "OR", see e.g. the discussions at
     // https://github.com/rust-lang/cargo/issues/2039
     // https://github.com/rust-lang/cargo/pull/4920
-    DeclaredLicenseProcessor.process(licenses, operator = SpdxOperator.OR)
-
-private fun parsePackage(pkg: CargoMetadata.Package, hashes: Map<String, String>): Package {
-    val declaredLicenses = parseDeclaredLicenses(pkg)
-    val declaredLicensesProcessed = processDeclaredLicenses(declaredLicenses)
+    val declaredLicensesProcessed = DeclaredLicenseProcessor.process(declaredLicenses, operator = SpdxOperator.OR)
 
     return Package(
         id = Identifier(
--- a/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
+++ b/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
@@ -201,7 +201,7 @@ class Cargo(
             definitionFilePath = VersionControlSystem.getPathInfo(definitionFile).path,
             authors = projectPkg.authors,
             declaredLicenses = projectPkg.declaredLicenses,
-            declaredLicensesProcessed = processDeclaredLicenses(projectPkg.declaredLicenses),
+            declaredLicensesProcessed = projectPkg.declaredLicensesProcessed,
             vcs = projectPkg.vcs,
             vcsProcessed = processProjectVcs(workingDir, projectPkg.vcs, projectPkg.homepageUrl),
             homepageUrl = projectPkg.homepageUrl,
--- a/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
+++ b/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
@@ -175,7 +175,7 @@ class Cargo(
 
         val hashes = readHashes(resolveLockfile(metadata))
         val projectPkg = packageById.getValue(projectId).let { cargoPkg ->
-            parsePackage(cargoPkg, hashes).let { it.copy(id = it.id.copy(type = managerName)) }
+            cargoPkg.toPackage(hashes).let { it.copy(id = it.id.copy(type = managerName)) }
         }
 
         val project = Project(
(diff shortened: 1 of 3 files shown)


plugins/package-managers/cargo/src/main/kotlin/CargoMetadata.kt
```diff
@@ -93,6 +93,31 @@ internal data class CargoMetadata(
...
+         * The dependencies of this package. This is an alternative to "dependencies" which contains additional
+         * information. In particular, this handles renamed dependencies.
+         */
+        val deps: List<Dep>
+    )
+
+    @Serializable
+    data class Dep(
+        /**
+         * The name of the dependency's library target. If this is a renamed dependency, this is the new name. */
```
mnonnenmacher, line 108: Wrong linebreak after `/**`.

plugins/package-managers/cargo/src/main/kotlin/CargoMetadata.kt
```diff
@@ -93,6 +93,31 @@ internal data class CargoMetadata(
         /** The Package ID of this node. */
         val id: String,
         /** The dependencies of this package, an array of Package IDs. */
-        val dependencies: List<String>
+        val dependencies: List<String>,
+
+        /**
+         * The dependencies of this package. This is an alternative to "dependencies" which contains additional
+         * information. In particular, this handles renamed dependencies.
```
mnonnenmacher, line 100: Sounds to me like `dependencies` would contain additional information, but looking at the datatypes it seems that `deps` contains additional dependencies. Consider replacing "which" with "that".

plugins/package-managers/cargo/src/main/kotlin/CargoMetadata.kt
```diff
@@ -93,6 +93,31 @@ internal data class CargoMetadata(
...
+        val name: String,
+        /** The Package ID of the dependency. */
+        val pkg: String,
+        /** Array of dependency kinds. Added in Cargo 1.40. */
+        val depKinds: List<DepKind>
+    )
+
+    @Serializable
+    data class DepKind(
+        /** The dependency kind. "dev", "build", or null for a normal dependency. */
```
mnonnenmacher, line 118: Don't we use \`null\` (same below)?

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -163,34 +138,43 @@ class Cargo(
             "Virtual workspaces are not supported."
         }
 
-        val projectNode = metadata.packages.single { it.id == projectId }
-        val groupedDependencies = projectNode.dependencies.groupBy { it.kind.orEmpty() }
+        val projectNode = metadata.resolve.nodes.single { it.id == projectId }
+        val depNodesByKind = mutableMapOf<String, MutableList<CargoMetadata.Node>>()
+        projectNode.deps.forEach { dep ->
+            dep.depKinds.forEach { depKind ->
+                val depNode = metadata.resolve.nodes.single { it.id == dep.pkg }
```
mnonnenmacher, line 143: Could this line be outside of this `forEach` loop?

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -266,7 +266,8 @@ private fun parseSourceArtifact(pkg: CargoMetadata.Package, hashes: Map<String,
 
     if (source == "registry+https://github.com/rust-lang/crates.io-index") {
         val url = "https://crates.io/api/v1/crates/${pkg.name}/${pkg.version}/download"
-        val hash = Hash.create(hashes[pkg.id].orEmpty())
+        val key = "${pkg.name} ${pkg.version} (${pkg.source})"
+        val hash = Hash.create(hashes[key].orEmpty())
```
mnonnenmacher, line 260: Commit message: Why not?

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
mnonnenmacher, line 1: Commit message: Typo in "property"

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -199,15 +189,15 @@ class Cargo(
             scopeDependencies = scopes
         )
 
-        val nonProjectPackages = packages
-            .filterNot { isProjectDependency(it.key) }
-            .mapTo(mutableSetOf()) { it.value }
+        val nonProjectPackages = packageById.values.mapNotNullTo(mutableSetOf()) { cargoPkg ->
+            cargoPkg.takeUnless { it.isProject() }?.let { parsePackage(cargoPkg, hashes) }
```
mnonnenmacher: Nit: Could be `parsePackage(it, hashes)`, fits better to the usage of `let`.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -199,15 +189,15 @@ class Cargo(
...
+        }
 
         return listOf(ProjectAnalyzerResult(project, nonProjectPackages))
     }
 }
 
-private val PATH_DEPENDENCY_REGEX = Regex("""^.*\(path\+file://(.*)\)$""")
+private fun CargoMetadata.Package.isProject(): Boolean = source == null
```
mnonnenmacher, line 200: Unnecessary return type declaration.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -190,7 +190,7 @@ class Cargo(
         )
 
         val nonProjectPackages = packageById.values.mapNotNullTo(mutableSetOf()) { cargoPkg ->
-            cargoPkg.takeUnless { it.isProject() }?.let { parsePackage(cargoPkg, hashes) }
+            cargoPkg.takeUnless { it.isProject() }?.let { cargoPkg.toPackage(hashes) }
```
mnonnenmacher, line 193: `let` is now unnecessary.

mnonnenmacher (changes requested):

plugins/package-managers/cargo/src/main/kotlin/CargoMetadata.kt
sschuberth: I've taken all of these comments as-is from [here](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html#json-format), so I'd prefer not to change them, or keep changes to a minimum.

plugins/package-managers/cargo/src/main/kotlin/CargoMetadata.kt
sschuberth: See above; I'm undecided whether this is a formatting change that I still should do or not.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
sschuberth: I've explained it in the commit message now.

fviernau (changes requested):

plugins/package-managers/cargo/src/main/kotlin/CargoMetadata.kt
```diff
@@ -93,6 +93,30 @@ internal data class CargoMetadata(
...
+        val kind: String? = null,
+        /** The target platform for the dependency. null if not a target dependency. */
```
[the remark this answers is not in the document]
sschuberth: Please see [above](https://github.com/oss-review-toolkit/ort/pull/8497#discussion_r1556944075), if possible I prefer to stick to the exact wording of the official docs.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
mnonnenmacher: It's obsolete with the following change anyway.

PR #8507, opened 2024-04-10
model: Improve and align `OrtResult.getIssues()`

fviernau:
Align with `getRuleViolations()` and `getVulnerabilities()` and make the API more flexible.
This prepares for filtering the issues by their affected path in a following change.

Context: #7921.

### DIFF
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -293,7 +293,7 @@ data class OrtResult(
      */
     @JsonIgnore
     fun getOpenIssues(minSeverity: Severity = Severity.WARNING) =
-        getIssues(omitExcluded = true, omitResolved = true, minSeverity = minSeverity).values.flatten()
+        getIssues(omitExcluded = true, omitResolved = true, minSeverity = minSeverity).values.flatten().distinct()
 
     /**
      * Return a list of [PackageConfiguration]s for the given [packageId] and [provenance].
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -246,21 +246,29 @@ data class OrtResult(
     /**
      * Return a map of all de-duplicated [Issue]s associated by [Identifier]. If [omitExcluded] is set to true, excluded
      * issues are omitted from the result. If [omitResolved] is set to true, resolved issues are omitted from the
-     * result.
+     * result. Issues with [severity][Issue.severity] below [minSeverity] are omitted from the result.
      */
     @JsonIgnore
-    fun getIssues(omitExcluded: Boolean = false, omitResolved: Boolean = false): Map<Identifier, Set<Issue>> {
+    fun getIssues(
+        omitExcluded: Boolean = false,
+        omitResolved: Boolean = false,
+        minSeverity: Severity = Severity.entries.min()
+    ): Map<Identifier, Set<Issue>> {
         val analyzerIssues = analyzer?.result?.getAllIssues().orEmpty()
         val scannerIssues = scanner?.getAllIssues().orEmpty()
         val advisorIssues = advisor?.results?.getIssues().orEmpty()
 
         val allIssues = analyzerIssues.zipWithCollections(scannerIssues).zipWithCollections(advisorIssues)
 
-        return allIssues.mapValues { (_, issues) ->
-            issues.filterTo(mutableSetOf()) { !omitResolved || !isResolved(it) }
-        }.filter { (id, issues) ->
-            issues.isNotEmpty() && (!omitExcluded || !isExcluded(id))
-        }
+        return allIssues.mapNotNull { (id, issues) ->
+            if (omitExcluded && isExcluded(id)) return@mapNotNull null
+
+            val filteredIssues = issues.filterTo(mutableSetOf()) {
+                (!omitResolved || !isResolved(it)) && it.severity >= minSeverity
+            }
+
+            filteredIssues.takeUnless { it.isEmpty() }?.let { id to it }
+        }.toMap()
     }
 
     /**
@@ -285,10 +293,7 @@ data class OrtResult(
      */
     @JsonIgnore
     fun getOpenIssues(minSeverity: Severity = Severity.WARNING) =
-        getIssues(omitExcluded = true, omitResolved = true)
-            .values
-            .flatten()
-            .filter { it.severity >= minSeverity }
+        getIssues(omitExcluded = true, omitResolved = true, minSeverity = minSeverity).values.flatten()
 
     /**
      * Return a list of [PackageConfiguration]s for the given [packageId] and [provenance].
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -252,8 +252,7 @@ data class OrtResult(
         val scannerIssues = scanner?.getAllIssues().orEmpty()
         val advisorIssues = advisor?.results?.getIssues().orEmpty()
 
-        val analyzerAndScannerIssues = analyzerIssues.zipWithCollections(scannerIssues)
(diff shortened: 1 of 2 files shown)


model/src/main/kotlin/OrtResult.kt
```diff
@@ -246,18 +246,25 @@ data class OrtResult(
...
      * issues are omitted from the result. If [omitResolved] is set to true, resolved issues are omitted from the
+     * result. if [minSeverity] is not null, issues with [Issue.severity] below [minSeverity] are omitted from the
      * result.
      */
     @JsonIgnore
-    fun getIssues(omitExcluded: Boolean = false, omitResolved: Boolean = false): Map<Identifier, Set<Issue>> {
+    fun getIssues(
+        omitExcluded: Boolean = false,
+        omitResolved: Boolean = false,
+        minSeverity: Severity? = null
```
sschuberth, line 256: In order to not have to deal with nullability, maybe use `Severity.entries.min()` as the default argument?

Note that this only makes sense if there is no special code that skips filtering by severity like there is at

https://github.com/oss-review-toolkit/ort/blob/968f956962301538b5e99c82a58f2da2a8cd72d9/model/src/main/kotlin/OrtResult.kt#L439-L445

model/src/main/kotlin/OrtResult.kt
```diff
@@ -292,7 +292,7 @@ data class OrtResult(
      */
     @JsonIgnore
     fun getOpenIssues(minSeverity: Severity = Severity.WARNING) =
-        getIssues(omitExcluded = true, omitResolved = true, minSeverity = minSeverity).values.flatten()
+        getIssues(omitExcluded = true, omitResolved = true, minSeverity = minSeverity).values.flatten().distinct()
```
sschuberth, line 295: Commit message: Maybe say "The same issue can occur for different identifiers, ..."

model/src/main/kotlin/OrtResult.kt
```diff
@@ -245,17 +245,22 @@ data class OrtResult(
...
         val scannerIssues = scanner?.getAllIssues().orEmpty()
         val advisorIssues = advisor?.results?.getIssues().orEmpty()
 
         val allIssues = analyzerIssues.zipWithCollections(scannerIssues).zipWithCollections(advisorIssues)
 
-        return allIssues.filterKeys { id -> !omitExcluded || !isExcluded(id) }
+        return allIssues.mapValues { (_, issues) ->
+            issues.filterTo(mutableSetOf()) { !omitResolved || !isResolved(it) }
+        }.filter { (id, issues) ->
+            issues.isNotEmpty() && (!omitExcluded || !isExcluded(id))
```
sschuberth, line 268: IMO it's not so nice that `issues.isNotEmpty()` is required here to "clean up" after the first filtering, and I'm wondering whether there's a more elegant way to only filter / map once.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -246,18 +246,24 @@ data class OrtResult(
...
         val scannerIssues = scanner?.getAllIssues().orEmpty()
         val advisorIssues = advisor?.results?.getIssues().orEmpty()
 
         val allIssues = analyzerIssues.zipWithCollections(scannerIssues).zipWithCollections(advisorIssues)
 
         return allIssues.mapValues { (_, issues) ->
-            issues.filterTo(mutableSetOf()) { !omitResolved || !isResolved(it) }
+            issues.filterTo(mutableSetOf()) {
+                (!omitResolved || !isResolved(it)) && it.severity >= minSeverity
+            }
```
sschuberth, line 266: All in all, this could now look something like this:

        return allIssues.mapNotNull { (id, issues) ->
            if (omitExcluded && isExcluded(id)) return@mapNotNull null

            val filteredIssues = issues.filterTo(mutableSetOf()) {
                (!omitResolved || !isResolved(it)) && it.severity >= minSeverity
            }

            filteredIssues.takeUnless { it.isEmpty() }?.let { id to it }
        }.toMap()

model/src/main/kotlin/OrtResult.kt
```diff
@@ -446,11 +446,14 @@ data class OrtResult(
      * [omitResolved] and remove violations below the [minSeverity].
      */
     @JsonIgnore
-    fun getRuleViolations(omitResolved: Boolean = false, minSeverity: Severity? = null): List<RuleViolation> {
+    fun getRuleViolations(
```
sschuberth, line 449: The commit message does not refer to `getRuleViolations()`, which makes it sound as if `minSeverity` is a general model property. I propose:

    refactor(model)!: Align severity filtering in `getRuleViolations()`

    Avoid dealing with a nullable `minSeverity` in the same was as in
    `getIssues()`.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -446,11 +446,14 @@ data class OrtResult(
...
+        omitResolved: Boolean = false,
+        minSeverity: Severity = Severity.entries.min()
+    ): List<RuleViolation> {
         val allViolations = evaluator?.violations.orEmpty()
 
         val severeViolations = when (minSeverity) {
-            null -> allViolations
+            Severity.entries.min() -> allViolations
```
sschuberth, line 456: With nullable `minSeverity` it was natural to only filter if `minSeverity != null`, but now this optimization seems a bit premature and referring to `Severity.entries.min()` twice is not so nice, IMO. So maybe just drop this optimization and always filter? When doing so, this should of course also be mentioned in the commit message as a conscious decision.

model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -270,7 +270,7 @@ class OrtResultTest : WordSpec({
     }
 
     "getRuleViolations()" should {
-        "return unfiltered rule violations if omitResolved is false" {
+        "return the issues, even tough it is resolved and a hint" {
```
sschuberth, line 273: These are rule violations, not issues. To me, the new name is less readable as it does not refer to `omitResolved`, so it sounds as if resolved violations should be returned unconditionally.

model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -304,7 +304,7 @@ class OrtResultTest : WordSpec({
                 .shouldContainExactly("rule id")
         }
 
-        "drop resolved rule violations if omitResolved is true" {
+        "filter out issues which are resolved or below warning severity" {
```
sschuberth, line 307: Here, basically the same remark as above applies.

model/src/main/kotlin/OrtResult.kt
fviernau: I believe the same applies to the performance optimization below for `omitResolved`. So, I did that in a following separate commit.

model/src/test/kotlin/OrtResultTest.kt
fviernau: Do you have a proposal ?

model/src/test/kotlin/OrtResultTest.kt
fviernau: Do you have a proposal?

model/src/main/kotlin/OrtResult.kt
fviernau: This is now addressed in the next commit, where I took your sniplet with `mapNotNull`

model/src/main/kotlin/OrtResult.kt
```diff
@@ -260,13 +260,15 @@ data class OrtResult(
 
         val allIssues = analyzerIssues.zipWithCollections(scannerIssues).zipWithCollections(advisorIssues)
 
-        return allIssues.mapValues { (_, issues) ->
-            issues.filterTo(mutableSetOf()) {
+        return allIssues.mapNotNull { (id, issues) ->
```
sschuberth, line 263: This hunk seems to have gone to the wrong commit.

PR #8512, opened 2024-04-12
test(node): Cover textual repository nodes

fviernau:
See individual commits.

### DIFF
--- a/model/src/main/kotlin/Mappers.kt
+++ b/model/src/main/kotlin/Mappers.kt
@@ -73,3 +73,5 @@ val EMPTY_JSON_NODE: JsonNode = MissingNode.getInstance()
 inline fun <reified T> String.fromYaml(): T = yamlMapper.readValue(this)
 
 fun Any?.toYaml(): String = yamlMapper.writeValueAsString(this)
+
+fun String.readJsonTree(): JsonNode = jsonMapper.readTree(this)
--- a/plugins/package-managers/node/src/test/kotlin/utils/NpmSupportTest.kt
+++ b/plugins/package-managers/node/src/test/kotlin/utils/NpmSupportTest.kt
@@ -31,7 +31,7 @@ import org.ossreviewtoolkit.model.VcsInfo
 import org.ossreviewtoolkit.model.VcsType
 
 class NpmSupportTest : WordSpec({
-    "expandNpmShortcutUrl" should {
+    "expandNpmShortcutUrl()" should {
         "do nothing for empty URLs" {
             expandNpmShortcutUrl("") shouldBe ""
         }
@@ -80,7 +80,7 @@ class NpmSupportTest : WordSpec({
         }
     }
 
-    "fixNpmDownloadUrl" should {
+    "fixNpmDownloadUrl()" should {
         "replace HTTP with HTTPS for the NPM registry only" {
             fixNpmDownloadUrl("http://registry.npmjs.org/babel-cli/-/babel-cli-6.26.0.tgz") shouldBe
                 "https://registry.npmjs.org/babel-cli/-/babel-cli-6.26.0.tgz"
@@ -93,7 +93,7 @@ class NpmSupportTest : WordSpec({
         }
     }
 
-    "parseNpmAuthors" should {
+    "parseNpmAuthors()" should {
         "get authors from a text node" {
             val node = ObjectMapper().run {
                 createObjectNode().apply {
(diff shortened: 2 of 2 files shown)


plugins/package-managers/node/src/test/kotlin/utils/NpmSupportTest.kt
```diff
@@ -196,7 +195,7 @@ class NpmSupportTest : WordSpec({
             @Suppress("Wrapping")
             val node = ObjectMapper().run {
                 createObjectNode().apply {
-                    replace("gitHead", TextNode(HashAlgorithm.SHA1GIT.emptyValue))
+                    replace("gitHead", TextNode("bar"))
```
sschuberth: Commit message: This is not about the "VCS path", but the "VCS revision".

model/src/main/kotlin/Mappers.kt
```diff
@@ -73,3 +73,5 @@ val EMPTY_JSON_NODE: JsonNode = MissingNode.getInstance()
 inline fun <reified T> String.fromYaml(): T = yamlMapper.readValue(this)
 
 fun Any?.toYaml(): String = yamlMapper.writeValueAsString(this)
+
+fun String.readJsonTree(): JsonNode = jsonMapper.readTree(this)
```
sschuberth, line 77: Commit message: s/programatical/programatic/

plugins/package-managers/node/src/test/kotlin/utils/NpmSupportTest.kt
```diff
@@ -192,17 +193,16 @@ class NpmSupportTest : WordSpec({
...
-            @Suppress("Wrapping")
-            val node = ObjectMapper().run {
-                createObjectNode().apply {
-                    replace("gitHead", TextNode("bar"))
-                    replace("repository", createObjectNode().apply {
-                        replace("type", TextNode("Git"))
-                        replace("url", TextNode("https://example.com/"))
-                        replace("directory", TextNode("foo"))
-                    })
+            val node = """
```
sschuberth, line 196: It'd be some work, but I guess ideally this should be done for all tests in this file now, or? That would nice decouple the test from Jackson as well.

plugins/package-managers/node/src/main/kotlin/utils/NpmSupport.kt
```diff
@@ -153,16 +153,25 @@ internal fun parseNpmVcsInfo(node: JsonNode): VcsInfo {
...
+    return if (repository.isTextual) {
+        VcsInfo(
+            type = VcsType.UNKNOWN,
+            url = expandNpmShortcutUrl(repository.textValue()),
+            path = "",
+            revision = head
+        )
+    } else {
+        val type = repository["type"].textValueOrEmpty()
+        val url = repository.textValue() ?: repository["url"].textValueOrEmpty()
```
sschuberth, line 165: I believe the original `repository.textValue()` here already handles the `repository.isTextual` case, or?

At least I'm pretty sure that original code already intended to fully handle that case:
- `type` would be empty, so `VcsType.forName(type)` returns `VcsType.UNKNOWN`
- `path` would be empty as well

So what additional case does your new code cover?

sschuberth (changes requested):

plugins/package-managers/node/src/main/kotlin/utils/NpmSupport.kt
fviernau: HM, are you sure that the previous code already worked in this case?

When I wrote this code (in my other PR) I do re-call that the analyzer previously did not detect VCS info but afterwards did. So, thought it was needed but probably I made a mistake during testing and had made further changes which made it work.

fviernau: Test proofs that you were right.

plugins/package-managers/node/src/test/kotlin/utils/NpmSupportTest.kt
fviernau: I would do a follow up PR for this

sschuberth (approved):

plugins/package-managers/node/src/test/kotlin/utils/NpmSupportTest.kt
fviernau: here you go: https://github.com/oss-review-toolkit/ort/pull/8518

PR #8526, opened 2024-04-16
refactor(DependencyHandler): Consume abstract but generate concrete types

sschuberth:
Change the interface to consume abstract collections but generate concrete collection types.


oheger-bosch:
This change makes me a bit nervous, since there could be undesired side effects when different collections with different equals semantics are used. Please test carefully that now not accidentally full dependency graphs are traversed by equals() implementations when inserting new nodes in the graph - or similar scenarios.

sschuberth:
@oheger-bosch, I'm not sure how what you're describing can be caused by this change as I'm not changing any of the `DependencyHandler` *implementations* really, but just the interface. Meaning, any concrete implementations of `equals()` would still be the same as before, or?

oheger-bosch:
Well, just take this as a kind request to test carefully.

sschuberth:
> Well, just take this as a kind request to test carefully.

I'll trust on the coverage of your tests here 😸

fviernau (approved):

PR #8527, opened 2024-04-16
Refactor Docker build

mnonnenmacher:
Refactor the Docker build to use the registry cache instead of language specific images for speeding up the build. The out of disk space issue is solved by removing unneeded tools from the GitHub Actions runner.

The new approach has been tested on my private fork at https://github.com/mnonnenmacher/ort. The images built during testing can be found at https://github.com/mnonnenmacher/ort/pkgs/container/ort and https://github.com/mnonnenmacher/ort/pkgs/container/ort-minimal.

Note that in this PR the fun-test job will always have to do a full build of the Docker image, as the registry cache is not yet populated. Tests on my fork have shown that if no changes are done to the Dockerfile the build takes about 2 minutes, a full build takes about 10 minutes.

Please see the commit messages for further details.

I will update the documentation in a follow-up PR, also with information about how the registry cache can be used for local builds.


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,90 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+    tags:
+      - '*'
```
sschuberth, line 11: Should we align this to `[0-9]+.[0-9]+.[0-9]+` from `release.yml`?

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,90 @@
...
+  workflow_dispatch:
+
+env:
+  GRADLE_OPTS: -Dorg.gradle.daemon=false
+  REGISTRY: ghcr.io
+  IMAGE_CACHE_TAG: ghcr.io/mnonnenmacher/ort:cache
+  IMAGE_NAME: ghcr.io/mnonnenmacher/ort
```
sschuberth, line 18: I guess these names should not include "mnonnenmacher"?

.github/workflows/build-and-test.yml
```diff
@@ -10,6 +10,8 @@ on:
 
 env:
   GRADLE_OPTS: -Dorg.gradle.daemon=false -Dkotest.assertions.multi-line-diff=unified
+  REGISTRY: ghcr.io
```
sschuberth, line 13: Commit message:
- "to speed *up*"
- "buil*d*s only"

.github/workflows/build-and-test.yml
```diff
@@ -116,19 +118,45 @@ jobs:
...
+        # Change the ownership of the Gradle user home and the workspace to the user in the Docker container.
+        sudo chown -R 1000:1000 /home/runner/.gradle
+        sudo chown -R 1000:1000 ${{ github.workspace }}
+
+        # Run the functional tests in the Docker container.
+        docker run \
+          -v ${{ github.workspace }}:/workspace \
+          -v /home/runner/.gradle:/home/ort/.gradle \
+          -w /workspace \
+          --entrypoint=sh \
```
sschuberth, line 153: Interesting that just `sh` works... maybe still use an absolute `/bin/sh` to not rely on the `PATH` environment? (I guess the latter is why it works.)

.github/workflows/build-and-test.yml
```diff
@@ -116,19 +118,45 @@ jobs:
...
+    - name: Setup Gradle
       uses: gradle/actions/setup-gradle@v3
       with:
         gradle-home-cache-cleanup: true
-        arguments: --scan -Ptests.include=org.ossreviewtoolkit.plugins.packagemanagers.* funTest jacocoFunTestReport
+    - name: Run functional tests that do require external tools
+      run: |
+        # Change the ownership of the Gradle user home and the workspace to the user in the Docker container.
+        sudo chown -R 1000:1000 /home/runner/.gradle
+        sudo chown -R 1000:1000 ${{ github.workspace }}
```
sschuberth, line 146: Instead of this (and the restoring code below), would it be an alternative to use `docker run --user 1001:121 ...`?

sschuberth: And probably use `$(id -u)` / `$(id -g)` instead of hard-coding the values.

mnonnenmacher: I wrote this part of the code several weeks ago, I don't remember exactly which options I tried, but I think I could not get it to work with the `--user` option. I would rather do any such changes in a follow-up PR, as currently I'm just happy that it works, and doing the changes in this PR might require a few additional rounds of testing.

.github/workflows/build-and-test.yml
```diff
@@ -10,6 +10,8 @@ on:
 
 env:
   GRADLE_OPTS: -Dorg.gradle.daemon=false -Dkotest.assertions.multi-line-diff=unified
+  REGISTRY: ghcr.io
+  TEST_IMAGE_TAG: ort:test
```
sschuberth, line 14: Just a though: Should we name this `FUNTEST_...` and `ort:funtest` to make more clear it's only used for functional tests?

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,88 @@
...
+          echo "ORT_VERSION=${ORT_VERSION}" >> $GITHUB_ENV
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract Docker Metadata for ort Image
```
sschuberth, line 43: Nit: Maybe spell it like "ORT"?

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          images: |
+            ${{ env.REGISTRY }}/${{ github.repository_owner }}/ort
+          tags: |
+            type=pep440,pattern={{version}}
+            type=raw,value=${{ env.ORT_VERSION }}
+            type=ref,event=branch
+            type=ref,event=tag
+            type=sha
+      - name: Build & Push ort Docker Image
```
sschuberth, line 55: Nit: Maybe spell it like "ORT"?

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta-ort-minimal.outputs.tags }}
+          labels: ${{ steps.meta-ort-minimal.outputs.labels }}
+          target: minimal
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/ort:cache
+          build-args: ORT_VERSION=${{ env.ORT_VERSION }}
+      - name: Print Disk Space
+        run: df -h
```
sschuberth, line 88: Just double-checking: I guess this is intended here, and not a debug left-over?

.github/workflows/build-and-test.yml
mnonnenmacher: I'm fine with both, if you want I can change it.

.github/workflows/docker-build.yml
mnonnenmacher: Yes, I thought it's useful in case we see out of disk space issues again, then we can easily check how much disk space was left at the end of the last successful job. But I have the impression that Docker deletes some temporary data after the build as the last time I checked, the value was higher than expected, so I could also remove it.

.github/workflows/build-and-test.yml
sschuberth: Let's keep it as-is then to reduce the iterations for this PR. Maybe we'll use the image also for integration or performance tests, and I can't make up my mind whether it should rather be `FUN_TEST_...` and `ort:funTest` instead 😅

.github/workflows/docker-build.yml
mnonnenmacher: That was supposed to refer to the image name, so I went with quoted `'ort'` instead and `'ort-minimal'` for the other image.

.github/workflows/docker-build.yml
mnonnenmacher: See above.

sschuberth (approved):

PR #8528, opened 2024-04-16
Evaluator chores

sschuberth:

### DIFF
--- a/plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
+++ b/plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
@@ -317,12 +317,18 @@ class EvaluatorCommand : OrtCommand(
         val scripts = scriptUris.map { it.toURL().readText() }
         val evaluatorRun = evaluator.run(*scripts.toTypedArray())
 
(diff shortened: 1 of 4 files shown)


plugins/commands/evaluator/src/main/kotlin/EvaluatorCommand.kt
```diff
@@ -317,13 +317,17 @@ class EvaluatorCommand : OrtCommand(
         val scripts = scriptUris.map { it.toURL().readText() }
         val evaluatorRun = evaluator.run(*scripts.toTypedArray())
 
-        val duration = with(evaluatorRun) { Duration.between(startTime, endTime).toKotlinDuration() }
-        echo("The evaluation of ${scriptUris.size} script(s) took $duration.")
+        if (evaluatorRun.violations.isNotEmpty()) {
+            echo("The following ${evaluatorRun.violations.size} rule violations were found:")
```
fviernau, line 321: What do you think about inserting: `else echo "No rule violations have been found.`

Should it be `have been` instead of `were` ?

fviernau (changes requested):

fviernau (approved):

PR #8544, opened 2024-04-19
feat(model): Allow configuring further PostgreSQL connection parameters

### LINKED ISSUE #8543: Make further database connection (pool) parameters configurable for Postgres based storages
On our CI with 20-30 scans execution in parallel on different runners we've started observing the Postgres database to reject new connections due to reached connection limit. From looking into PgAdmin one could see that plenty of the open connection were in idle state. One way to relax the situation a bit is to tweak the connection (pool) parameters on the client side.
In particular, according to @mnonnenmacher 's experiments it helped to adjust:
- minimum idle connections
- pool size

In scope of this ticket, these two parameters should be made configurable and while at it it would be nice to also make further
parameters configurable, analog to ORT server, see [1].

[1]: https://github.com/eclipse-apoapsis/ort-server/commit/f0f601e1c20d0e6f94d9e9c0240ea33098410ab8

fviernau:
The implementation has been done based on [1].

Fixes #8543.

[1]: https://github.com/eclipse-apoapsis/ort-server/commit/f0f601e1c20d0e6f94d9e9c0240ea33098410ab8

### DIFF
--- a/integrations/schemas/ort-configuration-schema.json
+++ b/integrations/schemas/ort-configuration-schema.json
@@ -384,6 +384,24 @@
                 },
                 "parallelTransactions": {
                     "type": "integer"
+                },
+                "connectionTimeout": {
+                    "type": "integer"
+                },
+                "idleTimeout": {
+                    "type": "integer"
+                },
+                "keepaliveTime": {
+                    "type": "integer"
+                },
+                "maxLifetime": {
+                    "type": "integer"
+                },
+                "maximumPoolSize": {
+                    "type": "integer"
+                },
+                "minimumIdle": {
+                    "type": "integer"
                 }
             },
             "required": [
--- a/integrations/schemas/ort-configuration-schema.json
+++ b/integrations/schemas/ort-configuration-schema.json
@@ -382,9 +382,6 @@
                 "sslrootcert": {
                     "type": "string"
                 },
-                "parallelTransactions": {
-                    "type": "integer"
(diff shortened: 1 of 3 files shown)


model/src/main/kotlin/config/PostgresConnection.kt
```diff
@@ -75,7 +75,45 @@ data class PostgresConnection(
...
      * The number of parallel transactions to use for the storage dispatcher.
      */
-    val parallelTransactions: Int = 5
+    val parallelTransactions: Int = 5,
+
+    /**
+     * The maximum number of milliseconds to wait for connections from the pool. For details see the
+     * [Hikari documentation](https://github.com/brettwooldridge/HikariCP#frequently-used).
+     */
+    val connectionTimeout: Long?,
```
sschuberth, line 84: Should these all have `@JsonInclude(Include.NON_NULL)` and `null` assigned as the default value, similar to the nullable values above?

integrations/schemas/ort-configuration-schema.json
```diff
@@ -382,9 +382,6 @@
                 "sslrootcert": {
                     "type": "string"
                 },
-                "parallelTransactions": {
```
sschuberth, line 385: Please add to the commit message that this is a fixup for f7dd71961496b92f293a97ef39bc9b9cfba68da9 which removed the use of this parameter.

PR #8549, opened 2024-04-22
chore(model): Add new PostgreSQL connection params to `reference.yml`

fviernau:
Add the parameter which have been recently added by 5b64b6c to the reference config file. Use the default values from [1] as default values.

[1]: https://github.com/brettwooldridge/HikariCP#frequently-used


sschuberth:
Just as a cross-reference, this reminds me that it would be nice to have https://github.com/oss-review-toolkit/ort/pull/5516 finally.

sschuberth (approved):

fviernau:
> Just as a cross-reference, this reminds me that it would be nice to have #5516 finally.

Yes, this would also help to relax the original issue (too many open /idle connections) as it would reduce the amount of connection pools.

Issue #8373, opened 2024-03-04
reporter: Allow hiding issues for excluded packages in WebApp report


fviernau:
It should be possible to filter issues which related to excluded packages.
Consider to have the hidden per default, because not relevant.

fviernau:
Hiding issues for excluded packages was already possible when this issue was created. The feature was just not discovered. Anyhow, it is now possible to also file issues with excluded affected path, as of https://github.com/oss-review-toolkit/ort/pull/8562.

PR #8552, opened 2024-04-22
test(reporters): Add three scan issues to report test assets

fviernau:
First preparations for implementing #8373.

### DIFF
--- a/plugins/reporters/evaluated-model/src/funTest/assets/evaluated-model-reporter-test-deduplicate-expected-output.yml
+++ b/plugins/reporters/evaluated-model/src/funTest/assets/evaluated-model-reporter-test-deduplicate-expected-output.yml
@@ -65,6 +65,15 @@ issue_resolutions:
   comment: "A comment explaining why the issue can be ignored."
 issues:
 - _id: 0
+  timestamp: "2024-04-22T10:36:10.661544294Z"
+  type: "SCANNER"
+  source: "FakeScanner"
+  message: "ERROR: Timeout after 300 seconds while scanning file 'project/file-within-excluded-project.dat'."
+  severity: "ERROR"
+  pkg: 0
+  scan_result: 0
+  how_to_fix: "Some how to fix text."
+- _id: 1
   timestamp: "1970-01-01T00:00:00Z"
   type: "SCANNER"
   source: "Dummy"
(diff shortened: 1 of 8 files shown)


plugins/reporters/evaluated-model/src/funTest/assets/evaluated-model-reporter-test-deduplicate-expected-output.yml
```diff
@@ -65,59 +65,80 @@ issue_resolutions:
   comment: "A comment explaining why the issue can be ignored."
 issues:
 - _id: 0
+  timestamp: "2024-04-22T10:36:10.661544294Z"
```
sschuberth, line 68: > Add one issue to an included- and another to an excluded project

"Add one issue to an included and another one to an excluded project"

> This prepares adding the ability

"This prepares for adding the ability" or "This prepares to add the ability"

plugins/reporters/evaluated-model/src/funTest/assets/evaluated-model-reporter-test-expected-output.yml
```diff
@@ -190,6 +221,97 @@ scan_results:
   package_verification_code: "da39a3ee5e6b4b0d3255bfef95601890afd80709"
 packages:
 - _id: 0
+  id: "Gradle:org.ossreviewtoolkit:nested-fake-project:1.0.0"
```
sschuberth, line 224: Any idea why the numeric IDs of the packages seem to be swapped now? That makes the diff a bit difficult to read...

fviernau: No idea. This is always imho difficult to figure out as the IDs are assigned by Jackson.

fviernau: I've debugged this for quite a bit, and it seems to just depend on the order things get serialized. E.g. the first identifier which gets serialized gets the `0` assigned. The next identifier the `1` and so forth. So, with this change `Gradle:org.ossreviewtoolkit:nested-fake-project:1.0.0` gets serialized first. 

I believe such diffs we have to live with, as long as we use the above mentioned approach to assign identifiers.

sschuberth: > E.g. the first identifier which gets serialized gets the `0` assigned.

Right, but why does the order of identifiers change with your PR, shouldn't it all be deterministic? In any case, thanks for looking into it.

fviernau: > Right, but why does the order of identifiers change with your PR, shouldn't it all be deterministic?

Hm, each time one make a small change to the test asset one needs to debug why jackson assigns the IDs in the way it does is imo raising the bar for contributions quite a bit. Can we shortcut this and live without the explanation ?

fviernau: Investigating further. Maybe @mnonnenmacher does the issue here ring any bell? (can you help?)

fviernau: I believe it is as follows:

1. The `EvaluatedModel.issues` property comes before `EvaluatedModel.packages`.
2. The issues do reference packages
3. So, the assignment of the package identifiers start when `issues` are serialized, and the ordering of
    the issues determines the IDs of the referenced packages.
4. As this commit adds a new _first_ issue, the package referenced gets the first id: `0`.

So, it seemed natural to just change property order so that `packages` are serialized before the issues, but 
that would re-create a similar issues, because there is a cyclic reference between these instances. 

@sschuberth what do you think?

sschuberth: That sounds reasonable as an explanation, good enough for me, thanks!

sschuberth (approved):

PR #8554, opened 2024-04-22
fix(git): Properly configure `insteadOf` for submodules

sschuberth:
This partly reverts c4a628d which introduced an URL replacement when getting nested provenance for Git working trees. That was only necessary because Git's `insteadOf` was not properly configured for submodules: `git config` needs to be called for each submodule separately instead of using Git's `-c` option.

### DIFF
--- a/plugins/version-control-systems/git/src/main/kotlin/Git.kt
+++ b/plugins/version-control-systems/git/src/main/kotlin/Git.kt
@@ -63,7 +63,7 @@ import org.semver4j.RangesListFactory
 const val GIT_HISTORY_DEPTH = 50
 
 // Replace prefixes of Git submodule repository URLs.
-private val REPOSITORY_URL_PREFIX_REPLACEMENTS = mapOf(
+private val REPOSITORY_URL_PREFIX_REPLACEMENTS = listOf(
     "git://" to "https://"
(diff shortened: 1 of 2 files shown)


plugins/version-control-systems/git/src/main/kotlin/Git.kt
```diff
@@ -63,7 +63,7 @@ import org.semver4j.RangesListFactory
 const val GIT_HISTORY_DEPTH = 50
 
 // Replace prefixes of Git submodule repository URLs.
-private val REPOSITORY_URL_PREFIX_REPLACEMENTS = mapOf(
+private val REPOSITORY_URL_PREFIX_REPLACEMENTS = listOf(
```
fviernau, line 66: From the commit message I do not understand whether there was an issue with the previous approach. 
Did it not work in some cases?

sschuberth: The previous approach only did the URL replacement on the call of `getNested()`, whereas the underlying repository configuration (for Git, inside `.git`) still contained the original URLs. Which means that any incremental update to that working tree would have failed.

fviernau (approved):

PR #8555, opened 2024-04-23
`OrtResultTest` improvements

fviernau:
See individual commits.

Context: #7921.

### DIFF
--- a/model/src/test/kotlin/OrtResultTest.kt
+++ b/model/src/test/kotlin/OrtResultTest.kt
@@ -25,6 +25,7 @@ import io.kotest.matchers.collections.containExactly
 import io.kotest.matchers.collections.containExactlyInAnyOrder
 import io.kotest.matchers.collections.haveSize
 import io.kotest.matchers.collections.shouldContainExactly
+import io.kotest.matchers.collections.shouldHaveSingleElement
 import io.kotest.matchers.should
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.string.shouldMatch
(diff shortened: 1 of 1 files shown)


model/src/test/kotlin/OrtResultTest.kt
```diff
@@ -173,10 +173,7 @@ class OrtResultTest : WordSpec({
 
             val openIssues = ortResult.getOpenIssues(Severity.WARNING)
 
-            openIssues should haveSize(1)
-            with(openIssues.first()) {
-                message shouldBe "Non-resolved issue"
-            }
+            openIssues.map { it.message }.shouldContainExactly("Non-resolved issue")
```
sschuberth, line 183: Nit: Closer to the original semantics would be `shouldHaveSingleElement` (which also does not take a `vararg` and thus has an `infix` version).

fviernau: Why does it have to be "close" to the previous code?
I don't see why it is closer. The semantics is the same, isn't it.

fviernau: `shouldHaveSingleElement "Non-resolved issue"` has the exact same semantics as `.shouldContainExactly("Non-resolved issue")`, doesn't it?

Changing it anyway, as `shouldHaveSingleElement` is a slight abbrevation.

sschuberth: > `shouldHaveSingleElement "Non-resolved issue"` has the exact same semantics as `.shouldContainExactly("Non-resolved issue")`, doesn't it?

`shouldContainExactly` *alone* does not have the same semantics as `shouldHaveSingleElement`, you also need to take the amount of parameters into account to match semantics. In other words, `shouldHaveSingleElement` better expresses that you're *never* expecting there to be more than a single element.

sschuberth (approved):

fviernau:
Failing tests were unrelated.

PR #8562, opened 2024-04-24
WebApp: Reflect excluded affected paths of issues in the UI

fviernau:
The effect on the UI is visible only under the issues summary view and is as follows:

1. The excluded state icon of issues now reflects excluded affected paths
2. The filtering of excluded issues now filters out issues with excluded affected paths

Note: This PR can be easily verified by creating a WebApp report from `./plugins/reporters/evaluated-model/src/funTest/assets/reporter-test-input.yml`.

Part of: #7921.

### UI Changes

1. Excluded scan timeout issue for non-excluded package `Gradle:org.ossreviewtoolkit.gradle.example:lib:1.0.0` now grayed out: 
Screenshot from 2024-04-25 13-11-58
2. After adjusting the filter for excluded issues, the issue is properly hidden:
Screenshot from 2024-04-25 13-12-20

### DIFF
--- a/model/src/main/kotlin/OrtResult.kt
+++ b/model/src/main/kotlin/OrtResult.kt
@@ -153,7 +153,7 @@ data class OrtResult(
         resolvedConfiguration.packageConfigurations.orEmpty().groupBy { it.id }
     }
 
-    private val excludedAffectedPathIssuesForId: Map<Identifier, Set<Issue>> by lazy {
+    private val issuesWithExcludedAffectedPathById: Map<Identifier, Set<Issue>> by lazy {
         buildMap<Identifier, MutableSet<Issue>> {
             scanner?.getAllScanResults().orEmpty().forEach { (id, scanResults) ->
                 scanResults.forEach { scanResult ->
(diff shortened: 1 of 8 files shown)


heliocastro (dismissed):
Looks good.

model/src/main/kotlin/OrtResult.kt
```diff
@@ -567,6 +567,12 @@ data class OrtResult(
             isPackageExcluded(id)
         }
 
+    /**
+     * Return `true` if and only if the given [issue] is excluded in scope of the given [id].
```
mnonnenmacher, line 571: Can you better describe what "in scope of the given id" means? IIUC it's either that the id is excluded completely or that the `affectedPath` of the issue matches a path exclude configured for id? The `excludedAffectedPathIssuesForId` property could use some documentation. And maybe also a renaming to make the purpose clearer, something like `issuesWithExcludedAffectedPathById`.

plugins/reporters/web-app-template/src/components/IssuesTable.jsx
mnonnenmacher, line 1: Commit message nit: "issue**s**" at the end.

mnonnenmacher (changes requested):

mnonnenmacher (approved):

tsteenbe (approved):

PR #8574, opened 2024-04-25
fix(cargo): Treat projects outside the analyer root as packages

### LINKED ISSUE #8571: [BUG] Cargo: submodules and local packages are being skipped
After the last releases, local packages and submodules that were defined with a 'path' in the Cargo.toml file are being skipped from the report file.

To reproduce it use a package dependency with a local path, for example:
```
# Cargo.toml
...
[dependencies]
bytes = { version = "^1.3", optional = true }
futures = "^0.3"
redis = { path = "../submodules/redis-rs/redis", features = ["aio", "tokio-comp", "tokio-rustls-comp", "connection-manager","cluster", "cluster-async"] }
logger_core = {path = "../logger_core"}

```
The generated NOTICE_DEFAULT file wouldn't contain attributions for both the `redis` and `logger_core` crates, as opposed to older ORT versions were these packages were included.

sschuberth:
E.g. when analyzing a single Cargo project in a sub-directory with a `cargo.toml` file that points to "path dependencies" in a parent directory, these dependencies should not be seen as projects but as packages. This restores the behavior from before 7522a0c.

Resolves #8571.

### DIFF
--- a/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
+++ b/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
@@ -162,7 +162,7 @@ class Cargo(
                 val pkg = packageById.getValue(node.id)
                 PackageReference(
                     id = Identifier("Crate", "", pkg.name, pkg.version),
-                    linkage = if (pkg.isProject()) PackageLinkage.PROJECT_STATIC else PackageLinkage.STATIC,
+                    linkage = if (pkg.isProject(analysisRoot)) PackageLinkage.PROJECT_STATIC else PackageLinkage.STATIC,
                     dependencies = dependencyNodes.toPackageReferences()
                 )
             }
@@ -191,14 +191,30 @@ class Cargo(
         )
 
         val nonProjectPackages = packageById.values.mapNotNullTo(mutableSetOf()) { cargoPkg ->
-            cargoPkg.takeUnless { it.isProject() }?.toPackage(hashes)
+            cargoPkg.takeUnless { it.isProject(analysisRoot) }?.toPackage(hashes)
         }
 
         return listOf(ProjectAnalyzerResult(project, nonProjectPackages))
     }
 }
 
-private fun CargoMetadata.Package.isProject() = source == null
+/**
+ * Return the local path for this Cargo package if applicable, or null if the Cargo package is not local.
+ */
+private fun CargoMetadata.Package.getLocalPath(): File? =
+    id.substringAfter("path+file://", "").ifEmpty { null }
+        ?.removeSuffix(")")?.substringBefore("#")?.let { File(it) }
+
+/**
+ * Return whether this Cargo package is supposed to be regarded as an ORT project. The [analysisRoot] is used to check
+ * whether this Cargo package lives within the analyzer root.
+ */
+private fun CargoMetadata.Package.isProject(analysisRoot: File): Boolean {
+    val isWithinAnalyzerRoot = getLocalPath()?.startsWith(analysisRoot.absoluteFile) == true
+
+    // If a package cannot be retrieved from anywhere but lies within the analyzer root, treat it as a project.
+    return source == null && isWithinAnalyzerRoot
+}
 
 private fun CargoMetadata.Package.toPackage(hashes: Map<String, String>): Package {
     val declaredLicenses = parseDeclaredLicenses()
@@ -208,6 +224,9 @@ private fun CargoMetadata.Package.toPackage(hashes: Map<String, String>): Packag
     // https://github.com/rust-lang/cargo/issues/2039
     // https://github.com/rust-lang/cargo/pull/4920
     val declaredLicensesProcessed = DeclaredLicenseProcessor.process(declaredLicenses, operator = SpdxOperator.OR)
+
+    val vcs = repository?.let { VcsHost.parseUrl(it) }.orEmpty()
+    val vcsProcessed = getLocalPath()?.let { PackageManager.processProjectVcs(it) } ?: vcs.normalize()
 
     return Package(
         id = Identifier(
@@ -225,7 +244,8 @@ private fun CargoMetadata.Package.toPackage(hashes: Map<String, String>): Packag
         binaryArtifact = RemoteArtifact.EMPTY,
         sourceArtifact = parseSourceArtifact(hashes).orEmpty(),
         homepageUrl = homepage.orEmpty(),
-        vcs = repository?.let { VcsHost.parseUrl(it) }.orEmpty()
+        vcs = vcs,
+        vcsProcessed = vcsProcessed
     )
 }
 
--- a/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
+++ b/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
@@ -225,7 +225,7 @@ private fun CargoMetadata.Package.toPackage(hashes: Map<String, String>): Packag
         binaryArtifact = RemoteArtifact.EMPTY,
         sourceArtifact = parseSourceArtifact(hashes).orEmpty(),
         homepageUrl = homepage.orEmpty(),
-        vcs = VcsHost.parseUrl(repository.orEmpty())
+        vcs = repository?.let { VcsHost.parseUrl(it) }.orEmpty()
     )
 }
 
--- a/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
+++ b/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
@@ -250,7 +250,7 @@ private fun CargoMetadata.Package.parseDeclaredLicenses(): Set<String> {
 // to.
 // See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories
 // for the specification for this kind of dependency.
-private val GIT_DEPENDENCY_REGEX = Regex("git\\+(https://.*)\\?(?:rev|tag|branch)=.+#([0-9a-zA-Z]+)")
+private val GIT_DEPENDENCY_REGEX = Regex("git\\+(https://.*)\\?(?:rev|tag|branch)=.+#([0-9a-fA-F]{7,40})")
 
 private fun CargoMetadata.Package.parseSourceArtifact(hashes: Map<String, String>): RemoteArtifact? {
     val source = source ?: return null


plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -191,14 +191,18 @@ class Cargo(
...
         return listOf(ProjectAnalyzerResult(project, nonProjectPackages))
     }
 }
 
-private fun CargoMetadata.Package.isProject() = source == null
+private fun CargoMetadata.Package.isProject(root: File? = null): Boolean {
+    val path = id.substringAfter("path+file://", "").removeSuffix(")").ifEmpty { null }
+    val isWithinRoot = root != null && path != null && File(path).startsWith(root)
+    return source == null && isWithinRoot
+}
```
fviernau, line 205: Tests still fail due to linkage type.

fviernau (changes requested):

plugins/package-managers/cargo/src/funTest/assets/projects/synthetic/cargo-expected-output.yml
```diff
@@ -12,10 +12,10 @@ project:
   declared_licenses_processed:
     spdx_expression: "Apache-2.0 OR MIT OR NOASSERTION"
   vcs:
-    type: ""
-    url: ""
-    revision: ""
-    path: ""
+    type: "Git"
```
fviernau, line 15: commit-msg: analyer -> analyzer

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -209,6 +215,10 @@ private fun CargoMetadata.Package.toPackage(hashes: Map<String, String>): Packag
     // https://github.com/rust-lang/cargo/pull/4920
     val declaredLicensesProcessed = DeclaredLicenseProcessor.process(declaredLicenses, operator = SpdxOperator.OR)
 
+    val vcs = repository?.let { VcsHost.parseUrl(it) }
+        ?: getLocalPath()?.let { PackageManager.processProjectVcs(it) }
```
fviernau, line 219: This is using the result of `processProjectVcs()` as (unprocessed) `vcs`. 
Just in recent discussion, IIUC, this has been seen undesired. What do you think?
(Looking at other package managers, many places used `VcsInfo.EMPY`)

I'm not against this change, but believe this needs to be clarified first.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -191,14 +192,19 @@ class Cargo(
...
         return listOf(ProjectAnalyzerResult(project, nonProjectPackages))
     }
 }
 
-private fun CargoMetadata.Package.isProject() = source == null
+private fun CargoMetadata.Package.getLocalPath(): File? =
+    id.substringAfter("path+file://", "").ifEmpty { null }
+        ?.removeSuffix(")")?.substringBefore("#")?.let { File(it) }
+
+private fun CargoMetadata.Package.isProject(root: File): Boolean =
```
fviernau, line 206: Looking at the signature, I have difficulties understanding what is expected to be passed as `root`. 
Can you help me understanding it? ..and can the name (root) be made more speaking ?

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -191,14 +192,19 @@ class Cargo(
...
+    source == null && getLocalPath()?.startsWith(root) == true
```
fviernau, line 207: Can we have a code comment, explaining the idea besides the expression ?

fviernau (changes requested):

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
sschuberth: I've added docs to clarify.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
sschuberth: I've extracted a speaking variable instead.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
sschuberth: I also recalled that discussion when writing this code and thought about it. The problem here is that without this code, a package that actually is a project outside of the working tree would not have *any* VCS information at all, neither processed nor unprocessed. Adding this code was the easiest way to fix that, even if strictly speaking the "bonus" of getting the unprocessed VCS non-empty for "real" packages would not be required.

fviernau: > The problem here is that without this code, a package that actually is a project outside of the working tree would not have any VCS information at all, 

I do not understand this, can you help? (I wonder why not `vcs ? VcsInfo.EMPTY; vcsProcesses=...` is possible)

fviernau (changes requested):

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
sschuberth: > I wonder why not `vcs ? VcsInfo.EMPTY; vcsProcessesed=...` is possible

Ah, I think I now got what you mean: Basically, use `processProjectVcs()` only for `vcsProcessed`. That I could do.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -191,14 +192,30 @@ class Cargo(
...
+            cargoPkg.takeUnless { it.isProject(workingDir) }?.toPackage(hashes)
         }
 
         return listOf(ProjectAnalyzerResult(project, nonProjectPackages))
     }
 }
 
-private fun CargoMetadata.Package.isProject() = source == null
+/**
+ * Return the local path for this Cargo package if applicable, or null if the Cargo package is not local.
```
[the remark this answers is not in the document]
sschuberth: > Is this always an absolute path

Yes, because it's derived from `definitionFile` which is guaranteed to be absolute.

> should it be mentioned in the docs or even checked with an assertion?

I'm simply calling `absoluteFile` now to make it work in any case.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
fviernau: Should have been `=` instead of `?`, but you got it. 

You don't have to change it. I just wanted to have it clear, since in my NPMSupport improvements PR we discarded the changes for `vcs` unprocessed recently. And I believe it was because of rationale that it shouldn't be processed. I'm also fine with changing the decision and just keep your code as-is in this regard. up to you.

sschuberth: I've already changed it locally, and it reduces the diff in test output, so that's probably good thing for review readability anyway.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
```diff
@@ -191,14 +192,30 @@ class Cargo(
...
+        ?.removeSuffix(")")?.substringBefore("#")?.let { File(it) }
+
+/**
+ * Return whether this Cargo package is supposed to be regarded as an ORT project. The [workingDir] is used to check
+ * whether a local path Cargo package lies outside the project directory.
+ */
+private fun CargoMetadata.Package.isProject(workingDir: File): Boolean {
+    val isWithinWorkingDir = getLocalPath()?.startsWith(workingDir) == true
+
+    // If a package cannot be retrieved from anywhere but lies inside the working directory, treat it as a project.
```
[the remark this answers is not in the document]
sschuberth: `analysisRoot` is constant per analyzer run, whereas `workingDir` changes with the `definitionFile`. We want the latter to understand whether a path dependency is out-of-tree of the *current* project the `definitionFile` belongs to.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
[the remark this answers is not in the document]
fviernau: > Yes, because it's derived from `definitionFile` which is guaranteed to be absolute.

My question was about the `File` returned by `getLocalPath()` not about the parent of the `definitionFile`.

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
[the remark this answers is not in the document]
fviernau: Hm, so what if we have a repository with these two files:
- `Cargo.toml`
- `dir/Cargo.toml`

Then `isProject( `Cargo.toml`)`  would either return `true` or `false` depending on which definition file is being currently processed. In my intuition the return value of whether a file is a project or not should remain constant 
in an analyzer result. (Maybe its just a matter of naming)  What do you think ?

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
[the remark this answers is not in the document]
sschuberth: Yes, that's also guaranteed to be absolute (as it's derived from a URI).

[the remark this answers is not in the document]
fviernau: Would a `require` check make sense / mention in the function docs of `getLocalPath()` ?

plugins/package-managers/cargo/src/main/kotlin/Cargo.kt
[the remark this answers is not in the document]
sschuberth: Thanks for insisting, I know know what confused me: The tests compare an isolated `ProjectAnalyerResult`, not a whole `AnalyzerResult`. This explains why the expected tests results actually should not change with this PR. I've manually verified that #8571 is still fixed.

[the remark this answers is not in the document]
fviernau: So, are you saying that the multiple definition files scenario I've mentioned works with this PR? (So, no issue there)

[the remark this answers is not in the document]
sschuberth: Your example, IIUC, only pledged for a given Cargo project to be consistently treated as an ORT project *or* as an ORT package *independently* of the current working directory. And yes, that's the case, and the current code now also resembles the original code more closely which was also using `analyzerRoot`:

https://github.com/oss-review-toolkit/ort/blob/adf14d48ec85fab9ab0ef6ec8250f644e1e29814/plugins/package-managers/cargo/src/main/kotlin/Cargo.kt#L130-L138

fviernau (approved):

Issue #8038, opened 2023-12-14
Show effective licenses in WebApp report's summary tab


tsteenbe:
The WebApp report currently show the declared and detected licenses in Summary tab it would be more useful if it showed the effective licenses in the summary e.g. license that actually applies to release artifacts including made license choices. In this way a report report can faster understand which licenses are applicable to the project and thus determine their license obligations.

sschuberth:
Sound like a duplicate of https://github.com/oss-review-toolkit/ort/issues/7436, or?

tsteenbe:
@sschuberth  #7436 is likely a bug in the EvaluatedModel not including the effective license in the output in some case, this ticket is about visualizing the effective license. Do agree that #7436 likely has an effect on #8038 e.g. not all effective licenses may be show in Summary Tab.

@fviernau  To continue our discussions the fastest way for me to show effective licenses in the WebApp is for you to expand the `statistics` in the EvaluatedModel to add `effective` within `licenses` so you get something like what is shown below. Maybe it makes sense to also implemented a `concluded` at the same time so report readers can instantly see for how many packages the license was concluded. What do you think?

```
"statistics": {
    "open_issues": {...},
    "open_rule_violations": {...},
    "dependency_tree": {...},
    "licenses": {
        "declared": {...},
        "detected": {...},
        "effective": {...}
```

hendrikebbers:
In the "Declared Licenses" not only the "effective license" but also the "filtered out" license is shown. Would be great if that can be changed / configurable.
<img width="714" alt="Image" src="https://github.com/user-attachments/assets/af08df7a-6260-4ed8-a248-7ea9ae402685" />

PR #8579, opened 2024-04-26
Add statistics for the effective licenses to the evaluated model's license stats

fviernau:
See individual commits.

Part of: #8038.


tsteenbe (approved):
@fviernau  After you fix-up the last commit message I did a retest and this looks good to me now

PR #8606, opened 2024-05-02
Static HTML report model clean-ups

fviernau:
A couple of trivial refactors focusing on cleaning-up the model classes of the static HTML report.

Please see the individual commit messages.

### DIFF
--- a/plugins/reporters/static-html/src/main/kotlin/ReportTable.kt
+++ b/plugins/reporters/static-html/src/main/kotlin/ReportTable.kt
@@ -35,9 +35,9 @@ import org.ossreviewtoolkit.model.config.ScopeExclude
 import org.ossreviewtoolkit.model.licenses.ResolvedLicense
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
 
-internal fun Collection<ReportTableModel.ResolvableIssue>.containsUnresolved() = any { !it.isResolved }
+internal fun Collection<ReportTable.ResolvableIssue>.containsUnresolved() = any { !it.isResolved }
 
-internal data class ReportTableModel(
+internal data class ReportTable(
     /**
      * The [VcsInfo] for the scanned project.
      */
--- a/plugins/reporters/static-html/src/main/kotlin/ReportTable.kt
+++ b/plugins/reporters/static-html/src/main/kotlin/ReportTable.kt
@@ -153,7 +153,7 @@ internal data class ReportTable(
 
     data class IssueTable(
         val type: Type,
-        val rows: List<IssueRow>
+        val rows: List<Row>
     ) {
         val errorCount = rows.count { it.issue.severity == Severity.ERROR }
         val warningCount = rows.count { it.issue.severity == Severity.WARNING }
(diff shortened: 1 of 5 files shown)


plugins/reporters/static-html/src/main/kotlin/ReportTable.kt
```diff
@@ -35,7 +35,7 @@ import org.ossreviewtoolkit.model.config.ScopeExclude
 import org.ossreviewtoolkit.model.licenses.ResolvedLicense
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
 
-internal fun Collection<ReportTable.ResolvableIssue>.containsUnresolved() = any { !it.isResolved }
+internal fun Collection<ResolvableIssue>.containsUnresolved() = any { !it.isResolved }
```
sschuberth: Commit message: The class is actually not called "ReportTableModel" anymore, but "ReportTable".

plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
```diff
@@ -72,7 +72,7 @@ class StaticHtmlReporter : Reporter {
     private val licensesSha1 = mutableMapOf<String, String>()
 
     override fun generateReport(input: ReporterInput, outputDir: File, config: PluginConfiguration): List<File> {
-        val reportTableModel = ReportTableModelMapper.map(input)
+        val reportTableModel = TableReportModelMapper.map(input)
```
sschuberth, line 75: Commit message:

> Rename the model to `TableReport`

"... model and mapper ..."

> table based and consist

"table-based and consist*s*"

sschuberth: If you want to emphasize the multiple tables, just swapping "ReportTable" to "TableReport" does not add much clarity IMO. I'd rather go for plural "tables" and use "ReportTables" or "TablesReport" (personally, I'd have a preference for the former as it more resembles "IssueTable" etc.).

sschuberth: The variable `reportTableModel` here isn't renamed accordingly, although the function parameter below *are* renamed.

plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
```diff
@@ -174,7 +174,7 @@ class StaticHtmlReporter : Reporter {
         return document.serialize().normalizeLineBreaks()
     }
 
-    private fun getRuleViolationSummaryString(ruleViolations: List<ResolvableViolation>): String {
+    private fun getRuleViolationSummaryString(ruleViolations: List<TableReportViolation>): String {
```
sschuberth, line 177: Commit message: Double "of of"

sschuberth (changes requested):

plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
sschuberth, line 177: Should this also use plural "Table*s*Report..." to not have too many variants of the spelling, and to document that it belongs to the scope of `TablesReport`?

sschuberth: Or maybe just move these classes inside of `TablesReport` and omit the prefix, similar to like you did for the `Row` classes?

sschuberth (approved):

PR #8610, opened 2024-05-03
Advisor results optimizations

sschuberth:

### DIFF
--- a/plugins/advisors/osv/src/main/kotlin/Osv.kt
+++ b/plugins/advisors/osv/src/main/kotlin/Osv.kt
@@ -111,9 +111,12 @@ class Osv(name: String, config: OsvConfiguration) : AdviceProvider(name) {
         val result = service.getVulnerabilityIdsForPackages(requests.map { it.second })
         val results = mutableListOf<Pair<Identifier, List<String>>>()
 
-        result.map {
-            it.mapIndexedTo(results) { i, vulnerabilities ->
-                requests[i].first.id to vulnerabilities
+        result.map { allVulnerabilities ->
+            // OSV returns vulnerability results in the same order as packages were requested, so use the list index to
+            // identify to which package a result belongs. This means that also empty results are returned as otherwise
+            // list indices would not match, so filter these out.
+            allVulnerabilities.mapIndexedNotNullTo(results) { i, pkgVulnerabilities ->
+                pkgVulnerabilities.takeUnless { it.isEmpty() }?.let { requests[i].first.id to it }
             }
         }.onFailure {
             logger.error {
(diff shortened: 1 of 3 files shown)


plugins/advisors/osv/src/main/kotlin/Osv.kt
```diff
@@ -111,9 +111,12 @@ class Osv(name: String, config: OsvConfiguration) : AdviceProvider(name) {
...
 
-        result.map {
-            it.mapIndexedTo(results) { i, vulnerabilities ->
-                requests[i].first.id to vulnerabilities
+        result.map { allVulnerabilities ->
+            // OSV returns vulnerability results in the same order as packages were requested, so use the list index to
+            // identify to which package a result belongs. This means that also empty results are returned as otherwise
+            // list indices would not match, so filter these out.
+            allVulnerabilities.mapIndexedNotNullTo(results) { i, pkgVulnerabilities ->
+                pkgVulnerabilities.takeUnless { it.isEmpty() }?.let { requests[i].first.id to it }
```
fviernau, line 119: I believe the following use case to be important to support in the future:
Make it obvious for which packages vulnerabilities have actually been queried.

"In order to do that, it is required to be able to distinguish between a) successful query with empty result b) no query has been made. "

As this change goes into the opposite direction, should we align the other way round?

fviernau (changes requested):

plugins/advisors/osv/src/main/kotlin/Osv.kt
sschuberth: > Make it obvious for which packages vulnerabilities have actually been queried.

As for all advisors the same packages are queried (and the list of packages mostly matches the `analyzer.packages` field, modulo excluded packages maybe), I'd rather store this information once on the `AdvisorRun` / `AdvisorRecord` level, and not per advisor.

> As this change goes into the opposite direction, should we align the other way round?

I don't believe that makes sense at this point. Allowing all advisors to return and store empty results per package blows up the result file massively with dummy entries like

      Go::golang.org/x/mod:0.10.0:
      - advisor:
          name: "OSV"
          capabilities:
          - "VULNERABILITIES"
        summary:
          start_time: "2024-05-03T07:38:15.948083720Z"
          end_time: "2024-05-03T07:38:23.883181733Z"
        defects: []
        vulnerabilities: []

So while I acknowledge the use-case, I believe the storage format needs to be totally different.

fviernau: > As for all advisors the same packages are queried

Which packages are queried successfully may depend on whether the ecosystem of the package is supported by the advisor.

> I don't believe that makes sense at this point. 

Agree, let's consider that use case out of scope here / bigger future topic.

fviernau (approved):

PR #8612, opened 2024-05-03
SPDX vs NPM project-package name improvements

sschuberth:

### DIFF
--- a/plugins/package-managers/node/src/main/kotlin/Npm.kt
+++ b/plugins/package-managers/node/src/main/kotlin/Npm.kt
@@ -534,8 +534,11 @@ open class Npm(
 
         val rawName = json["name"].textValueOrEmpty()
         val (namespace, name) = splitNpmNamespaceAndName(rawName)
-        if (name.isBlank()) {
-            logger.warn { "'$packageJson' does not define a name." }
+
+        val projectName = name.ifBlank {
+            getFallbackProjectName(analysisRoot, packageJson).also {
+                logger.warn { "'$packageJson' does not define a name, falling back to '$it'." }
+            }
         }
 
         val version = json["version"].textValueOrEmpty()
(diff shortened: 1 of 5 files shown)


plugins/package-managers/node/src/main/kotlin/Npm.kt
```diff
@@ -534,8 +534,13 @@ open class Npm(
 
         val rawName = json["name"].textValueOrEmpty()
         val (namespace, name) = splitNpmNamespaceAndName(rawName)
-        if (name.isBlank()) {
-            logger.warn { "'$packageJson' does not define a name." }
+
```
fviernau, line 537: commit-message: like to one -> like the one

plugins/package-managers/node/src/main/kotlin/Npm.kt
```diff
@@ -534,8 +534,13 @@ open class Npm(
...
+        val projectName = if (name.isBlank()) {
+            getFallbackProjectName(analysisRoot, packageJson).also {
+                logger.warn { "'$packageJson' does not define a name, falling back to '$it'." }
```
fviernau, line 540: We have a couple of callers of `getFallbackProjectName()`, but none logs any message about the use of the fallback. What doy ou think about simply dropping this here as well?

fviernau: Then, code can be shortened, e.g. to:
`= name.takeUnless { it.isBlank() } ?: getFallbackProjectName(analysisRoot, packageJson)`

fviernau (changes requested):

plugins/package-managers/node/src/main/kotlin/Npm.kt
sschuberth: > What doy ou think about simply dropping this here as well?

This is basically just the previously existing log statement, extended with the fallback name. It did not feel right to me to remove that log statement altogether in this context, esp. as there is a similar log statement about the version, too, which should then also be removed for consistency.

sschuberth: However, I've simplified the code via `ifBlank {}` now.

fviernau (approved):

PR #8613, opened 2024-05-03
fix(helper-cli): Fix broken reading of scan results by identifier

fviernau:
See commit message.


sschuberth:
Commit message: Typo in "provnenance".

tsteenbe (commented):
Tested this PR on several of the projects I currently doing license clearance for .. it working again as expected.

mnonnenmacher (approved):
> So, [1] obviously broke the reading of scan results in `helper-cli`'s `CreateCommand`

Having tests for the helper-cli would certainly help preventing such issues in future.

PR #8616, opened 2024-05-03
FossID: Limit fossid snippets

nnobelis:

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -32,6 +32,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withTimeoutOrNull
 
@@ -857,36 +858,37 @@ class FossId internal constructor(
         }
 
         val matchedLines = mutableMapOf<Int, MatchedLines>()
-        val snippets = runBlocking(Dispatchers.IO) {
-            pendingFiles.map {
-                async {
-                    logger.info { "Listing snippet for $it..." }
-                    val snippetResponse = service.listSnippets(config.user, config.apiKey, scanCode, it)
-                        .checkResponse("list snippets")
-                    val snippets = checkNotNull(snippetResponse.data) {
-                        "Snippet could not be listed. Response was ${snippetResponse.message}."
-                    }
-                    logger.info { "${snippets.size} snippets." }
+        val pendingFilesIterator = pendingFiles.iterator()
+        val snippets = flow {
+            while (pendingFilesIterator.hasNext()) {
+                val file = pendingFilesIterator.next()
+                logger.info { "Listing snippet for $file..." }
 
-                    val filteredSnippets = snippets.filterTo(mutableSetOf()) { it.matchType.isValidType() }
-
-                    if (config.fetchSnippetMatchedLines) {
-                        logger.info { "Listing snippet matched lines for $it..." }
-
-                        filteredSnippets.filter { it.matchType == MatchType.PARTIAL }.map { snippet ->
-                            val matchedLinesResponse =
-                                service.listMatchedLines(config.user, config.apiKey, scanCode, it, snippet.id)
-                                    .checkResponse("list snippets matched lines")
-                            val lines = checkNotNull(matchedLinesResponse.data) {
-                                "Matched lines could not be listed. Response was ${matchedLinesResponse.message}."
-                            }
-                            matchedLines[snippet.id] = lines
-                        }
-                    }
-
-                    it to filteredSnippets
+                val snippetResponse = service.listSnippets(config.user, config.apiKey, scanCode, file)
+                    .checkResponse("list snippets")
+                val snippets = checkNotNull(snippetResponse.data) {
+                    "Snippet could not be listed. Response was ${snippetResponse.message}."
                 }
-            }.awaitAll().toMap()
+                logger.info { "${snippets.size} snippets." }
+
+                val filteredSnippets = snippets.filterTo(mutableSetOf()) { it.matchType.isValidType() }
+
+                if (config.fetchSnippetMatchedLines) {
+                    logger.info { "Listing snippet matched lines for $file..." }
+
+                    filteredSnippets.filter { it.matchType == MatchType.PARTIAL }.map { snippet ->
+                        val matchedLinesResponse =
+                            service.listMatchedLines(config.user, config.apiKey, scanCode, file, snippet.id)
+                                .checkResponse("list snippets matched lines")
+                        val lines = checkNotNull(matchedLinesResponse.data) {
+                            "Matched lines could not be listed. Response was ${matchedLinesResponse.message}."
+                        }
+                        matchedLines[snippet.id] = lines
+                    }
+                }
+
+                emit(file to filteredSnippets.toSet())
+            }
         }
 
         return RawResults(
@@ -903,7 +905,7 @@ class FossId internal constructor(
      * Construct the [ScanSummary] for this FossID scan.
      */
     @Suppress("LongParameterList")
-    private fun createResultSummary(
+    private suspend fun createResultSummary(
         startTime: Instant,
         provenance: Provenance,
         rawResults: RawResults,
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -922,6 +922,7 @@ class FossId internal constructor(
         val snippetLicenseFindings = mutableSetOf<LicenseFinding>()
         val snippetFindings = mapSnippetFindings(
             rawResults,
+            config.snippetsLimit,
             issues,
             detectedLicenseMapping,
             snippetChoices,
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -31,6 +31,7 @@ import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.runBlocking
@@ -876,14 +877,19 @@ class FossId internal constructor(
                 if (config.fetchSnippetMatchedLines) {
                     logger.info { "Listing snippet matched lines for $file..." }
 
-                    filteredSnippets.filter { it.matchType == MatchType.PARTIAL }.map { snippet ->
-                        val matchedLinesResponse =
-                            service.listMatchedLines(config.user, config.apiKey, scanCode, file, snippet.id)
-                                .checkResponse("list snippets matched lines")
-                        val lines = checkNotNull(matchedLinesResponse.data) {
-                            "Matched lines could not be listed. Response was ${matchedLinesResponse.message}."
-                        }
-                        matchedLines[snippet.id] = lines
+                    coroutineScope {
+                        filteredSnippets.filter { it.matchType == MatchType.PARTIAL }.map { snippet ->
+                            async {
+                                val matchedLinesResponse =
+                                    service.listMatchedLines(config.user, config.apiKey, scanCode, file, snippet.id)
+                                        .checkResponse("list snippets matched lines")
+                                val lines = checkNotNull(matchedLinesResponse.data) {
+                                    "Matched lines could not be listed. Response was " +
+                                        "${matchedLinesResponse.message}."
+                                }
+                                matchedLines[snippet.id] = lines
+                            }
+                        }.awaitAll()
                     }
                 }
 
--- a/plugins/scanners/fossid/src/main/kotlin/FossIdConfig.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossIdConfig.kt
@@ -115,41 +115,41 @@ data class FossIdConfig(
 ) {
     companion object {
         /** Name of the configuration property for the server URL. */
-        private const val SERVER_URL_PROPERTY = "serverUrl"
+        private const val PROP_SERVER_URL = "serverUrl"
 
         /** Name of the configuration property for the username. */
-        private const val USER_PROPERTY = "user"
+        private const val PROP_USER = "user"
 
         /** Name of the configuration property for the API key. */
-        private const val API_KEY_PROPERTY = "apiKey"
+        private const val PROP_API_KEY = "apiKey"
 
         /** Name of the configuration property controlling whether ORT should wait for FossID results. */
-        private const val WAIT_FOR_RESULT_PROPERTY = "waitForResult"
+        private const val PROP_WAIT_FOR_RESULT = "waitForResult"
 
         /** Name of the configuration property defining the naming convention for projects. */
-        private const val NAMING_PROJECT_PATTERN_PROPERTY = "namingProjectPattern"
+        private const val PROP_NAMING_PROJECT_PATTERN = "namingProjectPattern"
 
         /** Name of the configuration property defining the naming convention for scans. */
-        private const val NAMING_SCAN_PATTERN_PROPERTY = "namingScanPattern"
+        private const val PROP_NAMING_SCAN_PATTERN = "namingScanPattern"
 
         /** Name of the configuration property defining whether to keep failed scans. */
-        private const val KEEP_FAILED_SCANS_PROPERTY = "keepFailedScans"
+        private const val PROP_KEEP_FAILED_SCANS = "keepFailedScans"
 
         /** Name of the configuration property controlling whether delta scans are to be created. */
-        private const val DELTA_SCAN_PROPERTY = "deltaScans"
+        private const val PROP_DELTA_SCAN = "deltaScans"
 
         /** Name of the configuration property that limits the number of delta scans. */
-        private const val DELTA_SCAN_LIMIT_PROPERTY = "deltaScanLimit"
+        private const val PROP_DELTA_SCAN_LIMIT = "deltaScanLimit"
 
-        private const val DETECT_LICENSE_DECLARATIONS_PROPERTY = "detectLicenseDeclarations"
+        private const val PROP_DETECT_LICENSE_DECLARATIONS = "detectLicenseDeclarations"
 
-        private const val DETECT_COPYRIGHT_STATEMENTS_PROPERTY = "detectCopyrightStatements"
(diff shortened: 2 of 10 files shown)


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -857,9 +857,9 @@ class FossId internal constructor(
         }
 
         val matchedLines = mutableMapOf<Int, MatchedLines>()
-        val snippets = runBlocking(Dispatchers.IO) {
-            pendingFiles.map {
-                async {
+        val snippets = pendingFiles.associateWith {
+            val filteredSnippets = lazy {
```
sschuberth, line 861: Instead of having lazy sets of snippets, could this be solved with sequences of snipptes instead of maybe simplify the code a bit?

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -191,6 +191,16 @@ internal fun mapSnippetFindings(
         results += mappedSnippets
     }
 
+    if (runningSnippetCount >= snippetsLimit) {
```
sschuberth, line 194: Shouldn't this only be `>`? (And the message below say "exceeded" instead of "reached".)

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -64,7 +64,7 @@ internal data class RawResults(
     val markedAsIdentifiedFiles: List<MarkedAsIdentifiedFile>,
     val listIgnoredFiles: List<IgnoredFile>,
     val listPendingFiles: List<String>,
-    val listSnippets: Map<String, Set<Snippet>>,
+    val listSnippets: Map<String, Lazy<Set<Snippet>>>,
```
fviernau, line 67: IIUC the `LazzySet` is introduced to change the order of evaluating something, introducing the assumption that
the set is not accesses too early. I believe such magic must not be put into the scan result data structure, but rather made explicit. Would it be possible to make the order of processing explicit in the code, e.g. have one function which does things in the desired order. And finally once done, constructs the result without any lazy evaluated properties?

fviernau (changes requested):

plugins/scanners/fossid/src/main/kotlin/FossIdConfig.kt
```diff
@@ -110,6 +110,9 @@ data class FossIdConfig(
     /** Whether matched lines of snippets are to be fetched. */
     val fetchSnippetMatchedLines: Boolean,
 
+    /** A limit on the amount of snippets to fetch. **/
```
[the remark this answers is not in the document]
nnobelis: To keep the logic simple, I would prefer not to give the possibility to the user to disable the limit.
A workaround, for now, would be to put an arbitrary big number smaller than `Int.MAX_VALUE` as limit, e.g. 2000000000.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: Could you please suggest with a code snippet how to achieve that ?
Would it be as simple as:
```
sequence{
  yieldAll(querySnippets())
}
``` 
? Would the same objective to lazy load the snippets be achieved?

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: I don't really understand your comment.
The lazy set is not introduced to manipulate the order of snippets in any way, it is here to load lazily the snippets at the last moment possible, to be able to load only the snippets as long as the limit has not been reached.

fviernau: I see, I guess I lack the full understanding. Can you help me understanding _why_ the "last moment possible" matter here? E.g. wouldn't a couple of moments earlier the limit also be known already?

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
sschuberth: Not sure if it's intended, but the message still says "reached" instead of "exceeded".

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
oheger-bosch: This would probably require a major rework, but would it make sense to do the mapping directly after the listing on a per-file basis? This would have the additional benefit that it could be done in parallel as well.

fviernau: I'm surprised that snippet choice is applied in the `scanner` stage. This implies that if you want a choice to 
take effect you need to re-scan. I suppose the turn-around time of making changes and seeing the effect is quite long. Given that, have there been any plans to move the application of the choices out of the scanner stage, or should there be? (If so, it'd mean that a solution for mapping snippets is needed, which works without knowing about the choices)

plugins/scanners/fossid/src/main/kotlin/FossId.kt
sschuberth: Probably anyway the right way is to stick with `async`, but make it [start lazily](https://kotlinlang.org/docs/composing-suspending-functions.html#lazily-started-async).

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: @fviernau This is the current snippet choice workflow:

- The ORT pipeline is run
- The snippets are collected in the snippet report
- Based on this results, the snippet choices are added to `.ort.yml`
- An ORT pipeline is re-triggered: chosen snippets are hidden from the results and the snippet report 

If you want to apply the snippet choice without rescanning (e.g. in the Evaluator step ?), it means that every scanner run should _always_ list ALL snippets to allow the choice loop to occurs latter.

As I wrote in the commit message, we have customer monorepos with more than 400 000 snippets, and IMO embedding all these in the ORT Results is not scalable.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: So it means the `snippets` in the `RawResults` are now `Map<String, Deferred<MutableSet<Snippet>>>` ?

sschuberth: I guess something like `MutableSet<Deferred<Snippet>>` would make more sense as values, because that allows to call `awaitAll()` on the values. But I don't see through the required changes right now; you'd have to find our yourself 😉

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
fviernau: Thanks for explaining this again. I think it's clearer to me now. Anyhow, I'd need to invest some time to deeper understand things for making a proper review. As I do not have a lot of time today, I step back from this review for now.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
oheger-bosch: If the strategy is to process files one by one until the limit is reached, the `async` is probably a bit pointless, since actually the requests are executed sequentially. It would probably be possible to still run things in parallel (maybe with a lower number of concurrent jobs) and cancelling jobs early when the limit has already been reached; in the worst case, more results than configured would be fetched and ignored afterwards. But I fear that this approach does not work well with the current separation between listing snippets and mapping them/applying choices.

sschuberth: Ok, it was just an untested idea anyway. Use whatever solution you see fit.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: @sschuberth I come back to you on this review because the tests are currently failing, for instance:
https://github.com/bosch-io/oss-review-toolkit/blob/nnobelis/limit_fossid_snippets/plugins/scanners/fossid/src/test/kotlin/FossIdSnippetChoiceTest.kt#L724-L767

This is because the test does the following:
- List the two pending files
- List the snippet of each file with a limit of two snippets.
- Since each pending file has two snippets, the scanner should stop after the first pending file, and output an issue.

So here `runningSnippetCount` is 2 and `snippetsLimit` is also 2, so the comparison was correct.
You can argue that this issue should only be present if the limit is strictly exceeded, but this the above example this is not the case and you need this issue to know there are more snippets results.

sschuberth: Indeed, my expectation would be that the issue only gets created if there are more snippets than the limit allows. In other words, if there are exactly as many snippets as the limit allows, there should be no issue filed. Is that the case currently?

nnobelis: It is the case currently but I see some problems with that:

- Right now, if we stop exactly at the limit, no issue is produced: what does it means for the user:
  - Is it because there were exactly `<limit>` snippets ?
  - Is it because the limit has been reached (and there are more snippets) ?
- There _must_ be an issue if there a more snippets than what the ORT results currently contains. This issue is present in the snippet report and this is the only way for the user to know that some snippet choices must be done to see the rest of the snippets (See my comment to Frank: https://github.com/oss-review-toolkit/ort/pull/8616#discussion_r1593635819)

We can't peek to see if there are more snippets in the remaining files because of the lazy loading. I mean we could but it would defeat the optimization that the limit allowed.

nnobelis: That's why I went with  `>=`: if we stop exactly at the limit, an issue is produced and the user hence knows an another run must be done.

After the second run, if there are more snippets, they are shown in the report. If there are none, the report is empty. Which is a minor annoyance but at least no information is hidden.

sschuberth: > I mean we could but it would defeat the optimization that the limit allowed.

Ok, that was the relevant bit of information for me.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -857,9 +857,9 @@ class FossId internal constructor(
         }
 
         val matchedLines = mutableMapOf<Int, MatchedLines>()
-        val snippets = runBlocking(Dispatchers.IO) {
-            pendingFiles.map {
-                async {
+        val snippets = pendingFiles.associateWith {
```
oheger-bosch, line 860: The new implementation is now fully sequential; every concurrent processing of files has been turned off. Could this be a performance issue if there are many pending files?

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -857,9 +857,9 @@ class FossId internal constructor(
...
+            val filteredSnippets = lazy {
+                runBlocking(Dispatchers.IO) {
```
oheger-bosch, line 862: Since `getRawResults()` is a *suspend* function, it should not call `runBlocking()`. Use `withContext(Dispatchers.IO)` instead.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: This is correct. Let's take an example with _n_ pending files and _s_ snippets in total.

Before the changes, one request was made to list the snippets per pending file, so n requests in total. Then, one request to list the matched lines _per_ snippet. Therefore _n + s_ in total, done concurrently.

With the changes, there is now _n + min(s, limit)_ requests in total, done sequentially,

I would assume this is less efficient but this is nevertheless a huge improvement for diminishing the amount of requests, as soon as the amount of snippet is growing.

Unfortunately, with the introduction of lazy loading, the initial concurrency cannot be get back. However, all the matched lines for all snippets of _one_ pending file can be made concurrently ?

Any suggestion ?

nnobelis: My guts feeling is _n << s_, by a factor 10 as least.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: Thanks for the hint. Unfortunately, `withContext` is a suspending function and it cannot be used with `lazy`, because the latter does not support suspension.

nnobelis: To avoid `runBlocking` altogether, I refactored the code to generate a `Sequence` as @sschuberth suggested.
Unfortunately, the function building a `Sequence` cannot be suspendable.
Therefore, I went with a `Flow`: values are emitted asynchronously and if the limit has been reached, the collecting job is cancelled.

plugins/scanners/fossid/src/main/kotlin/FossId.kt
nnobelis: I put again the code to list snippets of a pending file concurrently.

As we discussed, the proper solution to avoid all these problem would be to put the listing and mapping code together.
However, this is a big refactoring and I worry that the snippet choice feature relies on having all snippets information at once for a given pending file.

Additionally, it seems FossID is moving toward OpenAPI to replace the legacy API, hence I would like to wait for their new release first, before doing such a refactoring.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -167,29 +169,40 @@ private fun mapLicense(
...
-            snippetChoices,
-            remainingSnippetChoices,
-            snippetLicenseFindings
-        )
-    }.collect {
-        allFindings += it
+    val results = mutableSetOf<SnippetFinding>()
+    var runningSnippetCount = 0
+
+    coroutineScope {
```
oheger-bosch, line 182: Is this construct with `coroutineScope` and `launch` required? I would assume that the flow can be collected directly in this `suspend` function. Or is this done because of the `cancel()` to enforce the limit?

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -167,29 +169,40 @@ private fun mapLicense(
...
+                    detectedLicenseMapping,
+                    snippetChoices,
+                    remainingSnippetChoices,
+                    snippetLicenseFindings
+                )
+
+                runningSnippetCount += mappedSnippets.size
+                results += mappedSnippets
+
+                if (runningSnippetCount >= snippetsLimit) {
```
oheger-bosch, line 198: This `cancel()` looks rather drastic. Would it instead be possible to use `takeWhile` on the flow to enforce the limit?

plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -31,6 +31,7 @@ import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
```
oheger-bosch, line 31: Commit message nit: "This commit *enforces*...". Or maybe something like: "This commit adds concurrent processing again."

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: Yes it is required: `cancel()` is a function of the `coroutineScope`, not from the `Flow`.

oheger-bosch: Sorry, I am confused. Could you elaborate what is the exact purpose of this construct? And would the `takeWhile` approach I suggested in the other comment be an alternative?

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: I thought about `takeWhile` too. I tried:
```
 rawResults.listSnippets.takeWhile { runningSnippetCount < snippetsLimit }.collect { (file, rawSnippets) ->
                val mappedSnippets = rawSnippets.mapSnippetFindingsForFile(
                    [...]
                )

                runningSnippetCount += mappedSnippets.size
                results += mappedSnippets
            }
        }
```

Unfortunately, when condition becomes `false`, the first element that must be ignored is nevertheless emitted to the flow.

fun <[T](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/take-while.html)> [Flow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/index.html)<[T](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/take-while.html)>.[takeWhile](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/take-while.html)(predicate: suspend ([T](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/take-while.html)) -> [Boolean](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-boolean/index.html)): [Flow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-flow/index.html)<[T](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/take-while.html)>

This is because `T` must be emitted before the condition can be evaluated

oheger-bosch: There is a more generic version called `transformWhile`. It allows more control over the elements that are emitted. I think, this could be used here.

nnobelis: Ok, it works because it allows to evaluate the condition _after_ the emission.

plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
nnobelis: `cancel()` was needed to stop the coroutine collecting the emitted values.
I removed it for now as `transformWhile` achieves this requirement.

oheger-bosch (approved):
Looks good to me. Any objections from anybody against merging this?

sschuberth:
> Any objections from anybody against merging this?

There's a [unit test failure](https://github.com/oss-review-toolkit/ort/actions/runs/9078200843/job/24944835693?pr=8616#step:4:7559) now. Please first have a look at this.

oheger-bosch (approved):

oheger-bosch:
@sschuberth, are you fine with merging?

sschuberth:
> @sschuberth, are you fine with merging?

I haven't looked at the latest iterations code-wise, and I'm not deep enough into the topic to do so in reasonable time. So I don't feel comfortable to formally approve, but if you've approved and all tests pass (the Docker Build failure seems unrelated, GitHub apparently just had some problems today), I don't see a reason not to merge. Please go ahead as you see fit.

PR #8624, opened 2024-05-06
fix(spdx): Accept "additional-terms" as part of LicenseRef exceptions

sschuberth:
This is a hotfix for not failing SPDX reports to get generated for ScanCode license expressions like

    AGPL-3.0-only WITH LicenseRef-scancode-agpl-generic-additional-terms

A proper future fix could be to introduce a license classification provider interface that can query whether a given string is an exception or not.

### DIFF
--- a/utils/spdx/src/main/kotlin/SpdxExpression.kt
+++ b/utils/spdx/src/main/kotlin/SpdxExpression.kt
@@ -411,7 +411,7 @@ class SpdxLicenseWithExceptionExpression(
     val exception: String
 ) : SpdxSingleLicenseExpression() {
     companion object {
-        private val EXCEPTION_STRING_PATTERN = Regex("\\bexception\\b", RegexOption.IGNORE_CASE)
+        private val EXCEPTION_STRING_PATTERN = Regex("\\b(exception|additional-terms)\\b", RegexOption.IGNORE_CASE)
 
         /**
          * Parse a string into an [SpdxLicenseWithExceptionExpression]. Throws an [SpdxException] if the string cannot


utils/spdx/src/main/kotlin/SpdxExpression.kt
```diff
@@ -411,7 +411,7 @@ class SpdxLicenseWithExceptionExpression(
     val exception: String
 ) : SpdxSingleLicenseExpression() {
     companion object {
-        private val EXCEPTION_STRING_PATTERN = Regex("\\bexception\\b", RegexOption.IGNORE_CASE)
+        private val EXCEPTION_STRING_PATTERN = Regex("\\b(exception|additional-terms)\\b", RegexOption.IGNORE_CASE)
```
fviernau, line 414: I'd prefer to stay on the safe-side and list `LicenseRef-scancode-agpl-generic-additional-terms` explicitly, to not risk matching anything else. What do you think?

fviernau (changes requested):

utils/spdx/src/main/kotlin/SpdxExpression.kt
sschuberth: That would miss at least `LicenseRef-scancode-gpl-generic-additional-terms` (`gpl` vs `agpl`) which exists, too. In any case, I believe it's safe enough to assume something to be an exception that if it

1. is used with the `WITH` operator, and
2. starts with "LicenseRef-", and
3. contains "additional-terms"

fviernau: Out of scope:

Looking into `ScanCode`'s data it turns out that this identifier has `is_generic = true` which implies that there is not license text defined for that license identifier. (So, it's similar to this LicenseRef-scancode-other-permissive and so forth). Just wanted to share this, I don't have any idea on how this could be best addressed. 

My guess would be that we need a mechanism which omits generic licenses from SPDX reports, for both license IDs and exception IDs, as it does not make sense to list license IDs which are undefined in terms of the corresponding text.

fviernau (approved):

PR #8629, opened 2024-05-07
reporter: Further clean-up the code of the static HTML plugin

fviernau:
See individual commits.

### DIFF
--- a/plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
+++ b/plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
@@ -161,7 +161,7 @@ class StaticHtmlReporter : Reporter {
                         issueTable(tablesReport.advisorIssueSummary)
                     }
 
-                    tablesReport.projectDependencies.forEach { (_, table) ->
+                    tablesReport.projectDependencies.forEach { table ->
                         projectTable(table)
                     }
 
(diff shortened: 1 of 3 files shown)


plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
```diff
@@ -445,10 +445,8 @@ class StaticHtmlReporter : Reporter {
             }
 
             tbody {
-                val projectRow = table.rows.single { it.id == project.id }
-                projectRow(project.id.toCoordinates(), 1, projectRow)
-                (table.rows - projectRow).forEachIndexed { rowIndex, pkg ->
-                    projectRow(project.id.toCoordinates(), rowIndex + 2, pkg)
+                table.rows.forEachIndexed { rowIndex, pkg ->
```
sschuberth: Commit message: Could you go into a bit more detail what the "desired ordering" is? Am I reading the code right that the desire is to make the root project go first, and it's that logic that you're simplifying?

plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
```diff
@@ -161,7 +161,7 @@ class StaticHtmlReporter : Reporter {
                         issueTable(tablesReport.advisorIssueSummary)
                     }
 
-                    tablesReport.projectDependencies.forEach { (_, table) ->
+                    tablesReport.projectDependencies.forEach { table ->
```
sschuberth: Commit message: Maybe "from the previous commit" -> "due to the previous commit" (or "thanks to", "because of")

plugins/reporters/static-html/src/main/kotlin/StaticHtmlReporter.kt
sschuberth: Much better, thank you!

sschuberth (approved):

fviernau:
Merged despite the failing unrelated test.

PR #8633, opened 2024-05-08
static-html: Clean-up the CSS file a bit

fviernau:
This PR does not intend to arrive at something perfect, but targets a better overview as a preparation
for a re-design of the project table which will involve some further refactoring of CSS structure.

Part of: #7921.


plugins/reporters/static-html/src/funTest/assets/static-html-reporter-test-expected-output.html
```diff
@@ -63,21 +63,25 @@
     width: 110px;
 }
 
-.ort-report-labels {
+/*
```
sschuberth, line 70: Commit message: "So, Use" -> "So, use"

plugins/reporters/static-html/src/funTest/assets/static-html-reporter-test-expected-output.html
```diff
@@ -183,17 +183,17 @@
     border-right: 1px solid rgba(34, 36, 38, .15);
 }
 
-.ort-report-table.ort-report-rule-violation-table tr.ort-resolved {
+.ort-report-rule-violation-table tr.ort-resolved {
```
sschuberth: Commit message: In the first sentence, the "specializations" / "spezialize" duplication reads a bit odd, IMO. Can you think of a way to reword? Maybe just say "refers to" instead of "spezialize"? In any case, it should say "specialize".

sschuberth (changes requested):

plugins/reporters/static-html/src/funTest/assets/static-html-reporter-test-expected-output.html
fviernau: I ended up re-wording it to avoid the term "specialize" entirely.

sschuberth (approved):

PR #8634, opened 2024-05-08
test: Fix two failing fun tests

fviernau:

PR #8635, opened 2024-05-08
test(osv): Update expected resuts

fviernau:


fviernau:
Commit added to https://github.com/oss-review-toolkit/ort/pull/8634.

PR #8642, opened 2024-05-13
Add an initial Bitbake analyzer implementation

sschuberth:


Dockerfile
```diff
@@ -420,6 +420,17 @@ RUN mkdir -p $BAZEL_HOME/bin \
 FROM scratch as bazel
 COPY --from=bazelbuild /opt/bazel /opt/bazel
 
+#------------------------------------------------------------------------
+# BITBAKE
+FROM base as bitbakebuild
+
+ARG BITBAKE_VERSION=2.8.0
+
+RUN git clone --branch $BITBAKE_VERSION --depth 1 https://git.openembedded.org/bitbake /opt/bitbake
```
fviernau, line 429: Why did you choose to clone the _branch_, not the _tag_? 

Note: There is also an archive URL available https://git.openembedded.org/bitbake/snapshot/bitbake-2.8.0.tar.gz.

plugins/package-managers/bitbake/README.md
```diff
@@ -0,0 +1,13 @@
...
+It supersedes the combination of the [meta-doubleopen] and [do-convert] projects by relying on upstream [SBOM] generation in [SPDX] format, and converting the generated files to an ORT analyzer result file via ORT's [SPDX document file analyzer].
+
+[ORT]: https://github.com/oss-review-toolkit/ort
+[BitBake]: https://docs.yoctoproject.org/bitbake.html
+[Yocto]: https://www.yoctoproject.org/
+[meta-doubleopen]: https://github.com/doubleopen-project/meta-doubleopen
+[do-convert]: https://github.com/doubleopen-project/do-convert
+[SBOM]: https://docs.yoctoproject.org/dev/dev-manual/sbom.html
+[SPDX]: https://spdx.dev/
+[SPDX document file analyzer]: https://oss-review-toolkit.org/ort/docs/tools/analyzer
```
fviernau, line 13: commit: Why did you choose to use the analyzer for mapping the SPDX file? 

(I don't know that analyzer inside out, but I recall that it was implemented with some specific use case in mind which is beyond the SDPX spec)

plugins/package-managers/bitbake/src/funTest/kotlin/BitBakeToolFunTest.kt
```diff
@@ -0,0 +1,102 @@
...
+                val singleFileInfo = fileInfo.copy(
+                    managedFiles = fileInfo.managedFiles.map { (packageManager, definitionsFiles) ->
+                        packageManager to definitionsFiles.filter { it.name == recipeFileName }
+                    }.toMap()
+                )
+                analyze(singleFileInfo)
+            }
+
+            result.analyzer?.result shouldNotBeNull {
+                projects shouldHaveSize 90
```
fviernau, line 98: I think it would be valuable to make the fun test consistent with the ones for other package managers,
to use an expected output file which contains the expected ORT model. This would make it more obvious what the package manager outputs and protect from unintended breaking changes when changing the SPDX analyzer.

plugins/package-managers/bitbake/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+ * License-Filename: LICENSE
+ */
+
+plugins {
+    // Apply precompiled plugins.
+    id("ort-library-conventions")
+}
+
+dependencies {
+    api(project(":analyzer"))
```
fviernau, line 26: Should this be written as `api(projects.analyzer)` ? (Similar below)

plugins/package-managers/bitbake/src/funTest/kotlin/BitBakeToolFunTest.kt
```diff
@@ -0,0 +1,102 @@
...
+import org.ossreviewtoolkit.analyzer.create
+import org.ossreviewtoolkit.model.Identifier
+import org.ossreviewtoolkit.model.VcsInfo
+import org.ossreviewtoolkit.model.VcsType
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.plugins.versioncontrolsystems.git.Git
+import org.ossreviewtoolkit.utils.test.ExpensiveTag
+import org.ossreviewtoolkit.utils.test.shouldNotBeNull
+
+class BitBakeToolFunTest : WordSpec({
```
fviernau, line 41: Class name should be `BitBakeFunTest`.

plugins/package-managers/bitbake/src/funTest/kotlin/BitBakeToolFunTest.kt
```diff
@@ -0,0 +1,102 @@
...
+        val projectDir = tempdir()
+        val pokyVcsInfo = VcsInfo(VcsType.GIT, "https://git.yoctoproject.org/poky", "kirkstone-4.0.17")
+
+        Git().run {
+            val workingTree = initWorkingTree(projectDir, pokyVcsInfo)
+            updateWorkingTree(workingTree, pokyVcsInfo.revision)
+        } shouldBeSuccess pokyVcsInfo.revision
+
+        "create an SPDX file for the 'quilt-native' package" {
+            val recipeFileName = "quilt-native_0.67.bb"
```
fviernau, line 62: Can you help me understand _why_ this test tests via `Analyzer` instead of directory via `BitBake` class?

plugins/package-managers/bitbake/src/main/kotlin/BitBake.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    private val scriptFile by lazy { extractResourceToTempFile(BITBAKE_SCRIPT_NAME).apply { setExecutable(true) } }
+    private val spdxConfFile by lazy { extractResourceToTempFile(SPDX_CONF_NAME) }
+
+    private val spdxManager by lazy { SpdxDocumentFile(name, analysisRoot, analyzerConfig, repoConfig) }
+
+    override fun resolveDependencies(definitionFiles: List<File>, labels: Map<String, String>): PackageManagerResult {
+        val commonDefinitionDir = getCommonParentFile(definitionFiles)
+        val workingDir = requireNotNull(commonDefinitionDir.searchUpwardsForFile(INIT_SCRIPT_NAME)) {
```
fviernau, line 72: Should the search be limited to remain within the analyzer root?

plugins/package-managers/bitbake/src/main/kotlin/BitBake.kt
```diff
@@ -0,0 +1,166 @@
...
+        ) = BitBake(type, analysisRoot, analyzerConfig, repoConfig)
+    }
+
+    private val scriptFile by lazy { extractResourceToTempFile(BITBAKE_SCRIPT_NAME).apply { setExecutable(true) } }
+    private val spdxConfFile by lazy { extractResourceToTempFile(SPDX_CONF_NAME) }
+
+    private val spdxManager by lazy { SpdxDocumentFile(name, analysisRoot, analyzerConfig, repoConfig) }
+
+    override fun resolveDependencies(definitionFiles: List<File>, labels: Map<String, String>): PackageManagerResult {
+        val commonDefinitionDir = getCommonParentFile(definitionFiles)
```
fviernau, line 71: Can you help me understand where the assumption comes from, that all definition files _always share a single init script_? Or rather, could it be that there are multiple init scripts for independent definition files?

plugins/package-managers/bitbake/src/main/kotlin/BitBake.kt
```diff
@@ -0,0 +1,166 @@
...
+
+        logger.info { "Determined the working directory to be '$workingDir'." }
+
+        val localVersion = getBitBakeVersion(workingDir)
+        val globalVersion = createOrtTempDir().let { dir ->
+            getBitBakeVersion(dir).also { dir.safeDeleteRecursively(force = true) }
+        }
+
+        if (localVersion != globalVersion) {
+            logger.warn { "Local $managerName version $localVersion differs from global version $globalVersion." }
```
fviernau, line 84: Why is this a warning? e.g. what problem could this cause?

plugins/package-managers/bitbake/src/main/kotlin/BitBake.kt
```diff
@@ -0,0 +1,166 @@
...
+        val deployDirs = mutableSetOf<File>()
+
+        definitionFiles.forEach { definitionFile ->
+            val target = definitionFile.nameWithoutExtension.substringBeforeLast('_')
+
+            val deployDir = getDeployDir(workingDir, target)
+            deployDirs += deployDir
+
+            val spdxFile = deployDir.findSpdxFiles().find { it.name == "recipe-$target.spdx.json" }
+            if (spdxFile != null) {
```
fviernau, line 96: In order to ensure the SPDX file contains the right content, I believe the `if` case should be removed. What do you think ?

plugins/package-managers/bitbake/src/main/kotlin/BitBake.kt
```diff
@@ -0,0 +1,166 @@
...
+        logger.info { "Found ${spdxFiles.size} SPDX file(s) in '$commonDeployDir'." }
+
+        return spdxManager.resolveDependencies(spdxFiles, labels)
+    }
+
+    override fun resolveDependencies(definitionFile: File, labels: Map<String, String>): List<ProjectAnalyzerResult> =
+        throw NotImplementedError("This function is not supported for $managerName.")
+
+    private fun getDeployDir(workingDir: File, target: String): File {
+        val bitbakeEnv = runBitBake(workingDir, "-e", target)
```
fviernau, line 123: Should this be `bitBakeEnv` ?

Alternative:
```
val lines = runBitBake(workingDir, "-e", target).stdout.lineSequence()
return lines.mapNotNull { it.withoutPrefix("DEPLOY_DIR=") }.first().let { File(it.removeSurrounding("\"")) }
```

plugins/package-managers/bitbake/src/funTest/kotlin/BitBakeToolFunTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+            val version = bitBake.getBitBakeVersion(tempdir())
+
+            version shouldMatch "\\d+\\.\\d+\\.\\d+"
+        }
+    }
+
+    "Analyzing recipes from Poky" should {
+        val projectDir = tempdir()
+        val pokyVcsInfo = VcsInfo(VcsType.GIT, "https://git.yoctoproject.org/poky", "kirkstone-4.0.17")
```
fviernau, line 54: Not sure how long the clone takes, but it seems the repository is rather large.

Not just for speed, but also for minimizing the test, what do you think about crafting a more minimal project under the `assets` directory? This would enable adding further (corner) cases easily, on future increments of this implementation.

fviernau (changes requested):

plugins/package-managers/bitbake/src/main/kotlin/BitBake.kt
```diff
@@ -0,0 +1,166 @@
...
+
+        if (!scriptFile.delete()) logger.warn { "Unable to delete the temporary '$scriptFile' file." }
+        if (!spdxConfFile.delete()) logger.warn { "Unable to delete the temporary '$spdxConfFile' file." }
+
+        val commonDeployDir = deployDirs.singleOrNull() ?: getCommonParentFile(deployDirs)
+        val spdxFiles = commonDeployDir.findSpdxFiles().toList()
+
+        logger.info { "Found ${spdxFiles.size} SPDX file(s) in '$commonDeployDir'." }
+
+        return spdxManager.resolveDependencies(spdxFiles, labels)
```
fviernau, line 116: Without mapping the result from the manager, I suspect that `Identifiers` (e.g. type), purl, cpe might not match what is desired. Maybe also the concluded licenses should explicitly be nullified.

fviernau (changes requested):

Dockerfile
sschuberth: There is no `--tag` for `git clone`, but [`--branch` "can also take tags"](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-code--branchcodeemltnamegtem).

Dockerfile
```diff
@@ -420,6 +420,17 @@ RUN mkdir -p $BAZEL_HOME/bin \
 FROM scratch as bazel
 COPY --from=bazelbuild /opt/bazel /opt/bazel
 
+#------------------------------------------------------------------------
+# BITBAKE
+FROM base as bitbakebuild
+
+ARG BITBAKE_VERSION=2.8.0
```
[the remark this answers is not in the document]
sschuberth: It has bothered me for a while now that we have versions defined both in `.versions` and as part of Dockerfile `ARG` default values, so I've created https://github.com/oss-review-toolkit/ort/pull/8648 to consolidate that.

plugins/package-managers/bitbake/README.md
sschuberth: > Why did you choose to use the analyzer for mapping the SPDX file?

That was the easiest way to make use of `SpdxDocumentCache` / `SpdxResolvedDocument` and related resolution logic right now.

> I recall that it was implemented with some specific use case in mind which is beyond the SDPX spec

Yes, indeed we might run into issues here when fully implementing document resolution. Ideally, any such issues could be fixed in `SpdxDocumentFile` itself (e.g. by generalizing code or making some behavior configurable). But if that does not work out, we should consider making direct use of the "low-level" `spdx-utils` in `BitBake` instead.

sschuberth:
Closing in favor of https://github.com/oss-review-toolkit/ort/pull/11430.

Issue #7478, opened 2023-09-07
Bundler fails to parse dependency "NoSuchElementException: Key spoon is missing in the map."


bennati:
One of our projects gets the following scan issue:
`Failed to parse dependency 'spoon' of project 'Bundler::Gemfile:' in '/builds/oss/oss-review-toolkit/ort-gitlab-ci/project': NoSuchElementException: Key spoon is missing in the map.`

I was able to reproduce the issue with this minimal Gemfile:
```
# frozen_string_literal: true

source 'https://rubygems.org'

ruby '3.0.2'

source '<ARTIFACTORY>' do
  gem 'bootstrap-sass-2', '2.3.2.2.pre.5'
end

group :test, :development, :ui_test do
  gem 'pry-rails'
end
```

Interestingly, the error disappears when changing the Gemfile to the following:
```
# frozen_string_literal: true

source 'https://rubygems.org'

ruby '3.0.2'

gem 'bootstrap-sass', '2.3.2.2'

group :test, :development, :ui_test do
  gem 'pry-rails'
end
```

[Here is the diff](https://github.com/oss-review-toolkit/ort/files/12551019/patch.diff.zip) between `bootstrap-sass-2:2.3.2.2.pre.5` and `bootstrap-sass:2.3.2.2`

sschuberth:
Can you please share a stacktrace, or a real value for the `<ARTIFACTORY>` placeholder?

bennati:
There is no stacktrace: the error is not fatal and the analyzer continues

```
 ______________________________                                                
/        \_______   \__    ___/ The OSS Review Toolkit, version fdd83a3.       
|    |   | |       _/ |    |                                                   
|    |   | |    |   \ |    |    Running 'analyze' as 'ort' under Java 17.0.8 on
\________/ |____|___/ |____|    with 4 CPUs and a maximum of 27648 MiB of memor
                                                                               
Environment variables:                                                        
ORT_CONFIG_DIR = ort-configuration                                            
ORT_DATA_DIR = /builds/oss/oss-review-toolkit/ort-gitlab-ci/.ort              
JAVA_HOME = /opt/java/openjdk                                                 
ANDROID_HOME = /opt/android-sdk                                               
                                                                              
Looking for ORT configuration in the following file:
        /builds/oss/oss-review-toolkit/ort-gitlab-ci/config.yml

Looking for analyzer-specific configuration in the following files and directories:
[...]

PR #8643, opened 2024-05-14
Testing adding a comment to a Gemfile

lancedikson:
TBD if the test works out

### DIFF
--- a/plugins/package-managers/bundler/src/funTest/assets/projects/synthetic/lockfile/Gemfile
+++ b/plugins/package-managers/bundler/src/funTest/assets/projects/synthetic/lockfile/Gemfile
@@ -7,4 +7,5 @@ gem 'rack', '~>2.1'
 
 group :test do
     gem 'rspec', :require => 'spec'
+    gem 'elasticsearch-rails', github: 'elastic/elasticsearch-rails' # some random comment
 end


sschuberth:
I guess this PR related to [this comment](https://github.com/oss-review-toolkit/ort/issues/7478#issuecomment-2108588425) of yours. I ran the same test locally (as that's much faster) and it passes for me.

sschuberth:
> I guess this PR related to [this comment](https://github.com/oss-review-toolkit/ort/issues/7478#issuecomment-2108588425) of yours. I ran the same test locally (as that's much faster) and it passes for me.

Interesting, in CI it seems to fail. I'll investigate.

lancedikson:
> I guess this PR related to https://github.com/oss-review-toolkit/ort/issues/7478#issuecomment-2108588425 of yours.

Yes, sorry for not linking the PR with the issue proactively

> Interesting, in CI it seems to fail. I'll investigate.

Thanks! 🙌 Though the problem is hardly related to the open issue, but most likely to [the comment only](https://github.com/oss-review-toolkit/ort/issues/7478#issuecomment-1969987962).

sschuberth:
> Thanks! 🙌 Though the problem is hardly related to the open issue, but most likely to [the comment only](https://github.com/oss-review-toolkit/ort/issues/7478#issuecomment-1969987962).

Actually, I can reproduce the problem also without the comment, so it's not related to that (and explains why I could not reproduce the problem by just adding a comment to an existing file). Instead, the problem seems to be related to the `github` dependency type.

lancedikson:
Okay, interesting :) It somehow works without a comment for me in my actual project and doesn't work with a comment. That's why I thought it was a comment-related issue.

sschuberth:
Digging deeper, looks like this is related to Ruby calling native `git` to fetch from GitHub underneath, and then running into https://github.com/jruby/jruby/issues/6721. Actually, we already apply some of the work-arounds mentioned there. I'll investigate why they seem to have no effect.

sschuberth:
> and then running into [jruby/jruby#6721](https://github.com/jruby/jruby/issues/6721).

That was a red herring. The real cause was slightly different leading console output when fetching from GitHub, and that output was not stripped properly before parsing. https://github.com/oss-review-toolkit/ort/pull/8644 fixes this.

lancedikson:
Thanks! I'll test it with my project today.

PR #8648, opened 2024-05-15
refactor(docker): Define tool versions only once

sschuberth:
Previously, tool versions have been defined both inside the `.versions` file and as part of default `ARG` values. Consolidate these to `ARG` values only that get used where needed.


fviernau (commented):
LGTM.

sschuberth:
@oheger-bosch, @MarcelBochtler, @heliocastro can I get an approval from your side?

heliocastro (approved):
Is a good approach use versions.dockerfile with include now the new syntax is available.
I think the cost of uglify with sed the github action is ok'ish.

sschuberth:
> I think the cost of uglify with sed the github action is ok'ish.

I agree it's not super-nice, and I once filed [this issue](https://codeberg.org/devthefuture/dockerfile-x/issues/3) about it, but did not find the time so far to implement it.

PR #8655, opened 2024-05-16
Implement signing of ORT releases

sschuberth:

### DIFF
--- a/buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
@@ -33,6 +33,7 @@ private val Project.libs: LibrariesForLibs
 plugins {
     // Apply core plugins.
     application
+    signing
 
     // Apply precompiled plugins.
     id("ort-kotlin-conventions")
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/ort-application-conventions.gradle.kts
```diff
@@ -185,10 +186,23 @@ tasks.named<CreateStartScripts>("startScripts") {
...
+val distZip = tasks.named<Zip>("distZip")
+
+signing {
+    val signingInMemoryKey: String? by project
+    val signingInMemoryKeyPassword: String? by project
+
+    if (signingInMemoryKey != null && signingInMemoryKeyPassword != null) {
+        useInMemoryPgpKeys(signingInMemoryKey, signingInMemoryKeyPassword)
+        sign(distTar.get())
+        sign(distZip.get())
```
fviernau, line 202: Unfortunately, I'm lacking the Gradle knowledge to understand _when_ the signing (above two lines) is executed.
In particular, I wonder if always both, `.zip` and `.tar`, get signed. If building dist archive, does it always create both archive types? Can you give a hint?

sschuberth: The changes to `.github/workflows/release.yml` in the fore-last commit give you a hint: You have to call the respective `distTar` / `distZip` or `signDistTar` / `signDistZip` tasks explicitly. The above code does not execute anything, but it just *configures* the signing (and makes the `signDistTar` / `signDistZip` tasks available).

fviernau: Thank you!

fviernau (approved):

Issue #8211, opened 2024-02-01
Show the `source` of an issue on the web-app report's `Issues` tab


sschuberth:
A column before `Message` should be added about the `Source`. The columns should be sortable / filterable just like the `Package` column.

Issue #4041, opened 2021-05-19
WebAppReporter: Add more times (scanner start time, reporter creation time) to metadata and About dialog


neubs-bsi:
This is a follow up ticket to #3727 which was closed after merging #3978.

In the PR there was a discussion of adding more times, like the run of the scanner and the actual date of creation of the web app reporter. But the PR was merged and the ticket closed since an initial timestamp indicating a rough time frame was added with the time the anaylzer started. 

@sschuberth you expressed a wish for the time of creation of the report to be added.

Additionally those timestamps should include the exact time (hh:mm:ss), as currently they just display a date (e.g. Friday, 15th of June)

tsteenbe:
@neubs-bsi  I propose not only add the start time for all ORT components but also include the duration and move this info into ORT result statistics. What we then can do is add a new tab stats where we can not only show times but also numbers for issues/violations in the webapp report. Believe would be a better approach then just cramming it in the About dialog.

sschuberth:
Closed as part of backlog grooming. Feel free to comment if you would like to contribute to this.

PR #8668, opened 2024-05-21
feat(static-html): Make the descriptions of issues more compact

fviernau:
Only show the `Issue.message`, but stop showing `Issue.timestamp` and `Issue.severity`, to make the representation more compact and to align with the WebApp report. Note that the timestamp is rarely relevant and the severity is represented by a dedicated styling anyway.


sschuberth:
I was a bit hesitant about the rationale at first, as we seem to have tickets that go a little bit into the other direction for the web-app report (like https://github.com/oss-review-toolkit/ort/issues/4041, https://github.com/oss-review-toolkit/ort/issues/8211), but these tickets do not refer to the issue time and severity specifically, so I believe this change is ok.

sschuberth (approved):

sschuberth (approved):

PR #8674, opened 2024-05-22
chore(docker): Update NodeJS to the current version

fviernau:
Node version 21 is the "current" version until end of Q2 2024, see also [1]. According to [1], the corresponding NPM version is 10.5.0. So update the NPM version too.

[1]: https://nodejs.org/en/about/previous-releases


sschuberth (approved):

sschuberth:
@fviernau would you have a problem with reverting this, and / or going to Node.js v20.14.0 and belonging NPM v10.7.0, as Node 20.x is the current LTS release and I'm having trouble analyzing some projects with Node 21? If you're ok with this, I'll create that PR.

fviernau:
Don't have a problem with this in general, but would like to understand which update policy we would like to follow for node. Are you proposing now to only use LTS releases, and do the upgrade once the next LTS release becomes current (soon) ?

sschuberth:
> Are you proposing now to only use LTS releases, and do the upgrade once the next LTS release becomes current (soon) ?

I believe that's a good default policy, yes.

However, as I continue to run into Node projects that require special versions of Node, I believe we need to re-think our decision to leave it to the environment (in this case our Docker image) to provide a (fixed) version of Node (and also other tools). Instead, we could leverage utilities like e.g. https://github.com/beeequeue/nvm-rust to install the version of Node that the project requires at runtime.

sschuberth:
> upgrade to 22 already now.

My currently problem is with the version of Node, not of NPM. I cannot use Node > 20 currently.

fviernau:
> I believe that's a good default policy, yes.

If we start following this policy, then your problem will pop up soon again, because the next LTS release is about to become current.

sschuberth:
> because the next LTS release is about to become current.

That's why I wrote "default policy". We should probably allow for so time for projects to adapt to the next LTS version, and not update to it immediately after its release.

PR #8675, opened 2024-05-22
fix(fossid-webapp): Count snippets when enforcing the snippet limit

nnobelis:
The current logic is wrong: the snippet limit is applied to ORT snippet findings. However, a snippet finding refers to a single source code location and contains generally multiple snippets. Hence, those snippets must be counted when applying the limit. Please note that the new logic still garantees that all the snippets of a snippet finding are returned, for unity.

This is a fixup for 9223e90e.

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
@@ -191,8 +191,12 @@ internal suspend fun mapSnippetFindings(
             snippetLicenseFindings
         )
 
-        runningSnippetCount += mappedSnippets.size
-        results += mappedSnippets
+        val snippetFindingIterator = mappedSnippets.iterator()
+        while (runningSnippetCount < snippetsLimit && snippetFindingIterator.hasNext()) {
+            val snippetFinding = snippetFindingIterator.next()
+            runningSnippetCount += snippetFinding.snippets.size
+            results += snippetFinding
+        }
     }
 
     if (runningSnippetCount >= snippetsLimit) {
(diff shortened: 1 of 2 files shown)


plugins/scanners/fossid/src/main/kotlin/FossIdScanResults.kt
```diff
@@ -191,8 +191,12 @@ internal suspend fun mapSnippetFindings(
...
         )
 
-        runningSnippetCount += mappedSnippets.size
-        results += mappedSnippets
+        val snippetFindingIterator = mappedSnippets.iterator()
+        while (runningSnippetCount < snippetsLimit && snippetFindingIterator.hasNext()) {
+            val snippetFinding = snippetFindingIterator.next()
+            runningSnippetCount += snippetFinding.snippets.size
+            results += snippetFinding
+        }
```
sschuberth, line 199: Maybe instead something like

        results += mappedSnippets.takeWhile {
            runningSnippetCount += it.snippets.size
            runningSnippetCount < snippetsLimit
        }

sschuberth: And BTW, is the condition in line 182 then still needed / correct?

nnobelis: Your suggestion with `takeWhile` does not work because the condition is evaluated _before_ adding the element to the list. Therefore, when you are going over the limit (condition == true) you are not adding the current element, while in the _iterator_ logic it is added.

And yes the condition line 182 is still needed as it allows to stop loading snippets from FossID when the limit has been reached.

sschuberth:
There are two commits now that should simply be squashed.

oheger-bosch (approved):

PR #8676, opened 2024-05-22
chore(docker): Update Pnpm to the latest version

fviernau:
See individual commits.


sschuberth (approved):

MarcelBochtler (approved):

sschuberth:
Interestingly, this [adds some new packages](https://github.com/oss-review-toolkit/ort/actions/runs/9193909850/job/25288128247?pr=8676#step:7:1411).

fviernau:
@MarcelBochtler  do you recall, whether it was intended that `package-b` does not appear in the expected result of the workspaces project, but `package-a` does?

I'd like to understand why not both should be part of the result, see also 
https://github.com/oss-review-toolkit/ort/pull/5611/commits/205349609f4c71d72c2ecffff1a0c75856844153.

sschuberth:
> @MarcelBochtler do you recall, whether it was intended that `package-b` does not appear in the expected result of the workspaces project, but `package-a` does?
> 
> I'd like to understand why not both should be part of the result, see also [2053496](https://github.com/oss-review-toolkit/ort/commit/205349609f4c71d72c2ecffff1a0c75856844153).

As meanwhile [PNPM version 9.2.0 was released](https://github.com/pnpm/pnpm/releases/tag/v9.2.0) which I'm interested in, I'd have the same question to @MarcelBochtler to move this forward.

MarcelBochtler:
> @MarcelBochtler do you recall, whether it was intended that `package-b` does not appear in the expected result of the workspaces project, but `package-a` does?
> 
> I'd like to understand why not both should be part of the result, see also [2053496](https://github.com/oss-review-toolkit/ort/commit/205349609f4c71d72c2ecffff1a0c75856844153).

I don't recall it from the top of my head. I would need to investigate this. I had a quick look at the setup and for me it also looks like both packages should be part of the result.

fviernau:
I also believe it is correct to add the two new packages. I've updated the commit to cover it with a rationale.

docker/versions.dockerfile
```diff
@@ -17,7 +17,7 @@ ARG NPM_VERSION=10.7.0
 ARG NUGET_INSPECTOR_VERSION=0.9.12
 ARG PHP_VERSION=8.1
 ARG PIPTOOL_VERSION=24.0
-ARG PNPM_VERSION=8.10.3
+ARG PNPM_VERSION=9.2.0
```
sschuberth, line 20: Commit message nit: "for unknown reason" -> "for unknown reason*s*" or "for *an* unknown reason"

plugins/package-managers/node/src/funTest/assets/projects/synthetic/pnpm-workspaces/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
```
sschuberth, line 1: Just FYI, running `install` [with `--lockfile-only`](https://pnpm.io/cli/install#--lockfile-only) would probably have been sufficient.

sschuberth (changes requested):

sschuberth (approved):

PR #8677, opened 2024-05-22
chore(docker): Update Poetry to the latest version

fviernau:
See also [1].

[1]: https://github.com/python-poetry/poetry/releases/tag/1.8.3


sschuberth (approved):

sschuberth (dismissed):

sschuberth:
The upgrade seems to [introduce some issue](https://github.com/oss-review-toolkit/ort/actions/runs/9193028331/job/25288108834?pr=8677#step:7:2569).

Dockerfile
```diff
@@ -173,6 +173,12 @@ RUN ARCH=$(arch | sed s/aarch64/arm64/) \
     rm requirements.txt; \
     fi
 
+# Poetry 1.8.3 requires packaging version >= 22.0, see
```
sschuberth, line 175: Should we say "Poetry >= 1.8.0" here as that's the version the below issue is reported against?

Dockerfile
```diff
@@ -173,6 +173,12 @@ RUN ARCH=$(arch | sed s/aarch64/arm64/) \
...
+# https://github.com/python-poetry/poetry/issues/9029#issue-2153105325.
+# However, the above installation of ScanCode's requirements installs packaging version 21.3, see
```
sschuberth, line 177: Could this also be fixed by simply swapping the `RUN` statements from lines 156 and 167 to install ScanCode before Poetry, in the hope that installing the latter will then get the `packaging` right?

fviernau: I didn't try this because it's less minimal invasive. 
This way we keep dependencies as close as possible to ScanCode's requirements.txt.

But, if you prefer the swapping solution still, I can try if it work. What do you think?

sschuberth: I'm simply curious if it'd work. But actually, swapping might be a better solution as it does not hard-code `packaging==24.0.0`, which might be more future-proof with later Poetry updates.

fviernau: Ok, then let's just try here on CI.

fviernau: I've ended up building it locally...it works and ends up packaging version 24.0.

sschuberth (approved):

PR #8685, opened 2024-05-23
Docker build script fixes

sschuberth:


scripts/docker_build.sh
mnonnenmacher, line 1: > ... but this is the least intrusive fix

Maybe, but wouldn't it be cleaner to just remove the `--build-arg` lines instead?

heliocastro (approved):
Pretty much correct. Docker doesn't have a unified way to get a general build arts by file environment, so this solution fix the local issue.

Issue #7828, opened 2023-11-08
evaluator: Declared license mapping silently fails if mapped license is not a valid SPDX expression


fviernau:
I've filed a declared license mapping curation which did not have any effect. 
The reason for it seems to be that I forgot to add the `LicenseRef-` prefix to the value in the mapping, so that it is not a valid SPDX license id.

PR #8691, opened 2024-05-24
feat(model)!: Stop silently ignoring invalid declared license mappings

fviernau:
Previously, `PackageCuration.apply()` silently ignored declared license mapping entries with invalid SPDX expressions. For example, if one accidentally omits the `LicenseRef-` prefix, the mapping is just silently ignored.

Add a check that all values in the `Map` are valid SPDX expression into the constructor, to fail as early as possible. When used via a `FilePackageCurationProvider`, ORT now fails with the error message pointing to the problematic curation file path.

Fixes: #7828.

### DIFF
--- a/model/src/main/kotlin/PackageCurationData.kt
+++ b/model/src/main/kotlin/PackageCurationData.kt
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.annotation.JsonInclude
 import org.ossreviewtoolkit.utils.common.zip
 import org.ossreviewtoolkit.utils.ort.DeclaredLicenseProcessor
 import org.ossreviewtoolkit.utils.spdx.SpdxExpression
+import org.ossreviewtoolkit.utils.spdx.SpdxExpression.Strictness.ALLOW_LICENSEREF_EXCEPTIONS
 
 /**
  * This class contains curation data for a package. It is used to amend the automatically detected metadata for a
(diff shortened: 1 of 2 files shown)


fviernau:
@oss-review-toolkit/core-devs gentle ask for review

fviernau:
> I'd rather see a check being implemented in that function, as it then would also apply to built-in declared license mappings, to guards against typos etc. in [declared-license-mapping.yml](https://github.com/oss-review-toolkit/ort/blob/main/utils/spdx/src/main/resources/declared-license-mapping.yml).

I deliberately put it there for the reason mentioned in the commit message [1], as I believe knowing the file path of the incorrect entry must be in the error. 

[1] ... ORT now fails with the error message pointing to the problematic curation file path.

model/src/main/kotlin/PackageCurationData.kt
```diff
@@ -108,6 +109,14 @@ data class PackageCurationData(
     @JsonInclude(JsonInclude.Include.NON_NULL)
     val sourceCodeOrigins: List<SourceCodeOrigin>? = null
 ) {
+    init {
+        declaredLicenseMapping.values.forEach { spdxExpression ->
+            require(spdxExpression.isValid(ALLOW_LICENSEREF_EXCEPTIONS)) {
+                "The value '$spdxExpression' within the declared license mapping is not a valid SPDX expression."
```
sschuberth, line 115: IMO "within" sounds a bit strange here. How about:

"The declared license mapping for '$key' maps to '$value' which is not a valid SPDX expression."

(So I would also include the key into the exception message for clarity.)

sschuberth (approved):

PR #8693, opened 2024-05-24
Percent encoding / decoding test enhancements

sschuberth:


utils/common/src/test/kotlin/ExtensionsTest.kt
```diff
@@ -375,13 +376,21 @@ class ExtensionsTest : WordSpec({
...
 
                 unreserved.forEach {
-                    val singleCharString = it.toString()
-                    singleCharString.percentEncode() shouldBe singleCharString
+                    val decoded = it.toString()
+
+                    val encoded = decoded.percentEncode()
+
+                    encoded shouldBe decoded
+                    URLDecoder.decode(encoded, Charsets.UTF_8) shouldBe decoded
```
fviernau, line 393: To me it seems that the first assertion (line 392 / 383) has become redundant. 
If we extract a function `fun String.decode() = URLDecoder.decode(this, Charsets.UTF_8)`, then the test can be rewritten as a one liner, such as: `original.encode().decode() shouldBe original`

What do you think?

utils/common/src/test/kotlin/ExtensionsTest.kt
```diff
@@ -375,13 +376,21 @@ class ExtensionsTest : WordSpec({
 
             assertSoftly {
```
fviernau, line 377: out of scope: It would imo be nice to replace `assertSoftly / forEach` with `reserved.forAll`, maybe in a separate commit.

fviernau (changes requested):

utils/common/src/test/kotlin/ExtensionsTest.kt
sschuberth: Doing so would omit the check that the encoded string looks like expected, i.e. having hex chars, or not being different to the decoded string. However, I want to explicit check for properties of the encoded string.

fviernau: I see, then let's leave it as-is.

fviernau (approved):

fviernau (approved):

utils/common/src/test/kotlin/ExtensionsTest.kt
fviernau: Thanks!

fviernau (approved):

PR #8694, opened 2024-05-24
model: Eliminate the use of a `SortedSet`

fviernau:
Please see individual commits.

Part of #6235.


sschuberth:
> The file contains both, `SortedSet` and `SortedMap` converters.

Correct, but maps are also not collections 😅 But I cannot think of a better name either 😸

sschuberth (dismissed):

fviernau:
> Correct, but maps are also not collections 😅 But I cannot think of a better name either 😸

hehe, right. I missed that 😅

sschuberth (approved):

PR #8696, opened 2024-05-24
model: Serialize dependency graph edges in some explicit order

fviernau:
Make serialized order independent of insertion order in order to remove the need to update the tests when changing insertion order in upcoming changes.

Part of https://github.com/oss-review-toolkit/ort/issues/6235.

### DIFF
--- a/model/src/main/kotlin/DependencyGraph.kt
+++ b/model/src/main/kotlin/DependencyGraph.kt
@@ -99,10 +99,10 @@ data class DependencyGraph(
     val nodes: List<DependencyGraphNode>? = null,
 
     /**
-     * A list with the edges of this dependency graph. By traversing the edges, the dependencies of packages can be
+     * A set with the edges of this dependency graph. By traversing the edges, the dependencies of packages can be
      * determined.
      */
-    val edges: List<DependencyGraphEdge>? = null
+    val edges: Set<DependencyGraphEdge>? = null
 ) {
     companion object {
         /**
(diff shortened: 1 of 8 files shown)


model/src/main/kotlin/DependencyGraph.kt
```diff
@@ -102,7 +102,7 @@ data class DependencyGraph(
      * A list with the edges of this dependency graph. By traversing the edges, the dependencies of packages can be
      * determined.
      */
-    val edges: List<DependencyGraphEdge>? = null
+    val edges: Set<DependencyGraphEdge>? = null
```
sschuberth, line 108: The comment in line 102 should now also say "set" instead of "list". Also, the commit message should probably also mention that order is not important either (even if a set usually maintains insertion order).

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -179,7 +179,7 @@ class DependencyGraphBuilder<D>(
         )
     }
 
-    private fun Collection<DependencyGraphEdge>.removeCycles(): List<DependencyGraphEdge> {
+    private fun Set<DependencyGraphEdge>.removeCycles(): Set<DependencyGraphEdge> {
```
sschuberth, line 182: In the spirit of "be lenient when consuming, be strict when producing", should the receiver stay a `Collection`?

sschuberth (changes requested):

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
fviernau: I believe in general it makes sense, in this case I decided against it so that one does not have to even think about what happens when input contains duplicates, because the API is private (small scope), and because there is no use besides the Sets one.

sschuberth (approved):

PR #8697, opened 2024-05-24
refactor(model): Stop implementing Comparable with DependencyReference 

### LINKED ISSUE #6235: model: Eliminate sorted collections

fviernau:
This is a single commit, based on #8705. This PR has to be merged afterwards.

### DIFF
--- a/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
+++ b/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
@@ -105,13 +105,13 @@ class AnalyzerResultBuilderTest : WordSpec() {
 
     private val graph1 = DependencyGraph(
         packages = dependencies1,
-        scopeRoots = sortedSetOf(DependencyGraph.DEPENDENCY_REFERENCE_COMPARATOR, depRef1, depRef2),
+        scopeRoots = setOf(depRef1, depRef2),
         scopes = scopeMapping1
     )
 
     private val graph2 = DependencyGraph(
         packages = dependencies2,
-        scopeRoots = sortedSetOf(DependencyGraph.DEPENDENCY_REFERENCE_COMPARATOR, depRef3),
+        scopeRoots = setOf(depRef3),
         scopes = scopeMapping2
     )
 
--- a/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
+++ b/analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
@@ -103,18 +103,17 @@ class AnalyzerResultBuilderTest : WordSpec() {
         DependencyGraph.qualifyScope(project2, "scope-3") to listOf(RootDependencyIndex(0))
     )
 
-    private val graph1 =
-        DependencyGraph(
-            dependencies1,
-            sortedSetOf(DependencyGraph.DEPENDENCY_REFERENCE_COMPARATOR, depRef1, depRef2),
-            scopeMapping1
-        )
-    private val graph2 =
-        DependencyGraph(
-            dependencies2,
-            sortedSetOf(DependencyGraph.DEPENDENCY_REFERENCE_COMPARATOR, depRef3),
-            scopeMapping2
-        )
+    private val graph1 = DependencyGraph(
+        packages = dependencies1,
+        scopeRoots = sortedSetOf(DependencyGraph.DEPENDENCY_REFERENCE_COMPARATOR, depRef1, depRef2),
+        scopes = scopeMapping1
+    )
+
+    private val graph2 = DependencyGraph(
+        packages = dependencies2,
+        scopeRoots = sortedSetOf(DependencyGraph.DEPENDENCY_REFERENCE_COMPARATOR, depRef3),
+        scopes = scopeMapping2
+    )
 
     private val analyzerResult1 = ProjectAnalyzerResult(
         project1, setOf(package1), listOf(issue3, issue4)
--- a/model/src/main/kotlin/DependencyGraph.kt
+++ b/model/src/main/kotlin/DependencyGraph.kt
@@ -23,8 +23,6 @@ import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.databind.annotation.JsonSerialize
 
-import java.util.SortedSet
-
 import org.ossreviewtoolkit.model.utils.DependencyGraphEdgeSortedSetConverter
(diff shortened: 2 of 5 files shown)


analyzer/src/test/kotlin/AnalyzerResultBuilderTest.kt
```diff
@@ -105,13 +105,13 @@ class AnalyzerResultBuilderTest : WordSpec() {
 
     private val graph1 = DependencyGraph(
         packages = dependencies1,
-        scopeRoots = sortedSetOf(DependencyGraph.DEPENDENCY_REFERENCE_COMPARATOR, depRef1, depRef2),
+        scopeRoots = setOf(depRef1, depRef2),
```
sschuberth, line 108: Commit message: s/wich/which/

model/src/main/kotlin/DependencyGraph.kt
```diff
@@ -295,7 +295,7 @@ data class RootDependencyIndex(
  * Note: This is by intention no data class. Equality is tested via references and not via the values contained.
  */
 @JsonInclude(JsonInclude.Include.NON_DEFAULT)
-class DependencyReference(
+data class DependencyReference(
```
sschuberth, line 298: Commit message: Could you please add a permalink to the exact test case? I'm having trouble finding it.

model/src/main/kotlin/DependencyGraph.kt
```diff
@@ -326,17 +326,7 @@ class DependencyReference(
...
-     * Define an order on [DependencyReference] instances. Instances are ordered by their indices and fragment indices.
-     */
-    override fun compareTo(other: DependencyReference): Int =
-        if (pkg != other.pkg) {
-            pkg - other.pkg
-        } else {
-            fragment - other.fragment
-        }
-}
+)
```
sschuberth, line 329: I'm not sure what the original motivation for this `compareTo()` implementation was... like, was the implementation more or less arbitrary, just to be able to use the class in sorted sets, or was there more to it, like and intention to sort by indices in some representation?

Maybe @oheger-bosch can chime in here to confirm that this change (by now, with the preceding changes) is uncritical?

sschuberth (changes requested):

model/src/main/kotlin/DependencyGraph.kt
fviernau: I've also re-written the commit message as the thing with the `copy()` was not exactly right.

model/src/main/kotlin/DependencyGraph.kt
oheger-bosch: It's been a while... But as far as I remember, the purpose of the `compareTo()` implementation was to have a more deterministic order in dependency graphs. During tests, it popped up that the order in which items were added to a graph builder had an impact on the resulting graphs. They were isomorphic, but not identical. By defining an ordering, the effect could be reduced, but not fully eliminated.

model/src/main/kotlin/DependencyGraph.kt
```diff
@@ -302,7 +295,7 @@ data class RootDependencyIndex(
  * Note: This is by intention no data class. Equality is tested via references and not via the values contained.
  */
 @JsonInclude(JsonInclude.Include.NON_DEFAULT)
-class DependencyReference(
+data class DependencyReference(
```
oheger-bosch, line 298: This change makes me worry. The class comment says "This is by intention no data class. Equality is tested via references and not via the values contained." If you now have a generated `equals()` implementation that checks all properties, you can get very expensive traversals over whole graphs.

model/src/main/kotlin/DependencyGraph.kt
fviernau: > "They were isomorphic, but not identical" 

Do you remember in what way this could be problematic? Was it non-deterministic without implementing `compareTo()` ?

model/src/main/kotlin/DependencyGraph.kt
fviernau: Hm, the tests which started failing when stopping to implement `Comparable`, could it be that they asserted a bit too much? (e.g. is there maybe a fix for this possible in test code?)

model/src/main/kotlin/DependencyGraph.kt
oheger-bosch: I am not 100% sure, but I think for a given order in which the graph builder was called, the results were the same.

model/src/main/kotlin/DependencyGraph.kt
fviernau: Another alternative could be to hold these dependency references always in lists, not sets, to avoid the `equals()` calls in the first place.

fviernau:
@oheger-bosch @sschuberth given above comments, do you have any preference how the top commit should be changed to move this forward?

fviernau:
I've split-out the non-controversial commits to https://github.com/oss-review-toolkit/ort/pull/8705.

fviernau:
@oheger-bosch another question which just came up to me: Do you recall why it is necessary to have tests for de-serialization and serialization?  It seems that no custom serializers are at play, so normally we do no test serialization in that scenario IIUC. So, could it be an option to just drop these serialization tests from AnalyzerResultBuilderTest ?

model/src/main/kotlin/DependencyGraph.kt
oheger-bosch: `equals()` is called when constructing the dependency graph to match subtrees. This was the reason why this class was not a data class. I am pretty sure that this change will blow up for bigger graphs. Did you test this with a larger project?

oheger-bosch:
> @oheger-bosch another question which just came up to me: Do you recall why it is necessary to have tests for de-serialization and serialization? It seems that no custom serializers are at play, so normally we do no test serialization in that scenario IIUC. So, could it be an option to just drop these serialization tests from AnalyzerResultBuilderTest ?

I don't recally exactly any more. But I think these tests have been added in the progress of migrating from the old format for dependency trees to the dependency graph format. Since this was a major rewrite, it was useful to have tests in place.

fviernau:
No plans to continue with this.

PR #8700, opened 2024-05-27
refactor: Remove some unnecessary uses of sorted sets

fviernau:
See individual commits.

### DIFF
--- a/model/src/main/kotlin/licenses/LicenseClassifications.kt
+++ b/model/src/main/kotlin/licenses/LicenseClassifications.kt
@@ -70,7 +70,7 @@ data class LicenseClassifications(
     /** A property allowing convenient access to the names of all categories defined in alphabetical order. */
     @get:JsonIgnore
     val categoryNames: Set<String> by lazy {
-        categories.mapTo(sortedSetOf()) { it.name }
+        categories.mapTo(mutableSetOf()) { it.name }
     }
 
     init {
--- a/model/src/test/kotlin/DependencyTreeNavigatorTest.kt
+++ b/model/src/test/kotlin/DependencyTreeNavigatorTest.kt
@@ -82,7 +82,7 @@ class DependencyTreeNavigatorTest : AbstractDependencyNavigatorTest() {
 
         "dependencyTreeDepth" should {
             "return 0 if the scope does not contain any package" {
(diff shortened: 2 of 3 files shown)


model/src/main/kotlin/utils/DependencyGraphBuilder.kt
```diff
@@ -553,8 +553,8 @@ private fun constructSortedScopeMappings(
 ): Map<String, List<RootDependencyIndex>> {
     val orderedMappings = mutableMapOf<String, List<RootDependencyIndex>>()
 
-    scopeMappings.keys.toSortedSet().forEach { scope ->
-        orderedMappings[scope] = scopeMappings.getValue(scope)
+    scopeMappings.entries.sortedBy { it.key }.forEach { (scope, rootDependencyIndices) ->
+        orderedMappings[scope] = rootDependencyIndices
```
sschuberth, line 557: Hmm, I believe we need a different rationale here, as the code does not really look simpler to me, but longer and more complex. What's the real goal here, I guess to remove the use of `toSortedSet()`? Then we should say so.

sschuberth (changes requested):

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
fviernau: My goal was indeed to simplify. I've just dropped this commit.

sschuberth (approved):

model/src/main/kotlin/utils/DependencyGraphBuilder.kt
sschuberth: While reviewing the code base as part of https://github.com/oss-review-toolkit/ort/pull/8705 I've come across this place again, and I've changed my mind a bit:

While the new code still is not really simpler IMO, it's better as it avoids an unneeded `getValue()` on `scopeMappings` (and the mention of `toSortedSet()` which we prefer to get rid of).

So, mind re-introducing the commit with a slightly adjust commit message?

fviernau: Sure, here you go: https://github.com/oss-review-toolkit/ort/pull/8706.

PR #8703, opened 2024-05-27
feat(Yarn): Add basic support for Corepack [1].

oheger-bosch:
According to the Yarn 2+ documentation, Corepack is the preferred way to install this package manager. If this method is used, the name of the executable has to be determined differently.

[1]: https://yarnpkg.com/corepack

### DIFF
--- a/plugins/package-managers/node/src/main/kotlin/Yarn2.kt
+++ b/plugins/package-managers/node/src/main/kotlin/Yarn2.kt
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.MappingIterator
 import com.fasterxml.jackson.databind.node.NullNode
 import com.fasterxml.jackson.databind.node.ObjectNode
+import com.fasterxml.jackson.module.kotlin.contains
 import com.fasterxml.jackson.module.kotlin.readValues
 
 import java.io.File
@@ -87,6 +88,10 @@ private enum class YarnDependencyType(val type: String) {
  * - *disableRegistryCertificateVerification*: If true, the `yarn npm info` commands called by this package manager will
  *   not verify the server certificate of the HTTPS connection to the NPM registry. This allows to replace the latter by
  *   a local one, e.g. for intercepting the requests or replaying them.
+ * - *corepackOverride*: Per default, this class determines via auto-detection whether Yarn has been installed via
+ *   [Corepack](https://yarnpkg.com/corepack), which impacts the name of the executable to use. With this option,
+ *   auto-detection can be disabled, and the enabled status of Corepack can be explicitly specified. This is useful to
+ *   force a specific behavior in some environments.
  */
 class Yarn2(
     name: String,
@@ -99,6 +104,11 @@ class Yarn2(
          * The name of the option to disable HTTPS server certificate verification.
          */
         const val OPTION_DISABLE_REGISTRY_CERTIFICATE_VERIFICATION = "disableRegistryCertificateVerification"
+
+        /**
+         * The name of the option that allows overriding the automatic detection of Corepack.
+         */
+        const val OPTION_COREPACK_OVERRIDE = "corepackOverride"
 
         /**
          * The name of Yarn 2+ resource file.
@@ -119,10 +129,35 @@ class Yarn2(
          * The amount of package details to query at once with `yarn npm info`.
          */
         private const val BULK_DETAILS_SIZE = 1000
+
+        /**
+         * The name of the manifest file used by Yarn 2+.
+         */
+        private const val MANIFEST_FILE = "package.json"
+
+        /**
+         * The name of the property that defines the package manager and its version if Corepack is enabled.
+         */
+        private const val PACKAGE_MANAGER_PROPERTY = "packageManager"
+
+        /**
+         * The name of the default executable. This is used when the [OPTION_COREPACK_OVERRIDE] option is set.
+         */
+        private const val DEFAULT_EXECUTABLE_NAME = "yarn"
+
+        /**
+         * Check whether Corepack is enabled based on the `package.json` file in [workingDir]. If no such file is found
+         * or if it cannot be read, assume that this is not the case.
+         */
+        private fun isCorepackEnabledInManifest(workingDir: File): Boolean {
+            return runCatching {
+                jsonMapper.readTree(workingDir.resolve(MANIFEST_FILE)).contains(PACKAGE_MANAGER_PROPERTY)
+            }.getOrDefault(false)
+        }
     }
 
     class Factory : AbstractPackageManagerFactory<Yarn2>("Yarn2") {
-        override val globsForDefinitionFiles = listOf("package.json")
+        override val globsForDefinitionFiles = listOf(MANIFEST_FILE)
 
         override fun create(
             analysisRoot: File,
@@ -135,7 +170,8 @@ class Yarn2(
      * The Yarn 2+ executable is not installed globally: The program shipped by the project in `.yarn/releases` is used
      * instead. The value of the 'yarnPath' property in the resource file `.yarnrc.yml` defines the path to the
      * executable for the current project e.g. `yarnPath: .yarn/releases/yarn-3.2.1.cjs`.
-     * This map holds the mapping between the directory and their Yarn 2+ executables.
+     * This map holds the mapping between the directory and their Yarn 2+ executables. It is only used if Yarn has not
+     * been installed via Corepack; then it is accessed under a default name.
      */
     private val yarn2ExecutablesByPath: MutableMap<File, String> = mutableMapOf()
 
@@ -156,6 +192,13 @@ class Yarn2(
 
     override fun command(workingDir: File?): String {
         if (workingDir == null) return ""
+
+        val corepackEnabled = if (OPTION_COREPACK_OVERRIDE in options) {
+            options[OPTION_COREPACK_OVERRIDE].toBoolean()
+        } else {
+            isCorepackEnabledInManifest(workingDir)
+        }
+        if (corepackEnabled) return DEFAULT_EXECUTABLE_NAME
 
         return yarn2ExecutablesByPath.getOrPut(workingDir) {
             val yarnConfig = yamlMapper.readTree(workingDir.resolve(YARN2_RESOURCE_FILE))
--- a/plugins/package-managers/node/src/test/kotlin/Yarn2Test.kt
+++ b/plugins/package-managers/node/src/test/kotlin/Yarn2Test.kt
@@ -1,0 +1,151 @@
+/*
+ * Copyright (C) 2024 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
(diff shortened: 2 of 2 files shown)


analyzer/src/funTest/kotlin/Yarn2Test.kt
```diff
@@ -0,0 +1,89 @@
+/*
```
fviernau, line 1: commit-msg: Would it make sense to move the reference `[1]` behind the first mention of "Corepack" in the message _body_?

analyzer/src/funTest/kotlin/Yarn2Test.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.plugins.packagemanagers.node.Yarn2
+
+class Yarn2Test : WordSpec({
+    "command" should {
+        "return the executable defined in yarnrc.yml" {
+            val workingDir = tempdir()
+            val executable = "yarn-wrapper.js"
+            workingDir.resolve(".yarnrc.yml").writeText("yarnPath: $executable")
+            val executableFile = workingDir.resolve(executable).also {
```
sschuberth, line 39: Nit: Use `apply` instead of `also` to avoid the need for `it` below.

analyzer/src/funTest/kotlin/Yarn2Test.kt
sschuberth, line 1: Commit message: It's a bit unusual to have a foot note like `[1]` in the title. Better move it to the body text, after "documentation" or "Corepack".

analyzer/src/funTest/kotlin/Yarn2Test.kt
```diff
@@ -0,0 +1,89 @@
...
+import io.kotest.matchers.string.shouldContain
+
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.PackageManagerConfiguration
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.plugins.packagemanagers.node.Yarn2
+
+class Yarn2Test : WordSpec({
+    "command" should {
+        "return the executable defined in yarnrc.yml" {
```
sschuberth, line 35: Use ".yarnrc.yml" also in the test name to match the file name exactly. Similar below.

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
```diff
@@ -100,6 +103,11 @@ class Yarn2(
          */
         const val OPTION_DISABLE_REGISTRY_CERTIFICATE_VERIFICATION = "disableRegistryCertificateVerification"
 
+        /**
+         * The name of the option that determines whether Yarn 2+ was installed via Corepack.
+         */
+        const val OPTION_COREPACK_ENABLED = "corepackEnabled"
```
fviernau, line 109: commit-message: Can you explain _why_ you choose to implement this as an option as opposed to adding detection whether Corepack is enabled? Looking at [1] it says:
```
 You can quickly check whether Corepack is enabled by
running `yarn exec env`: if you get a path as output, Corepack
 is properly installed. If not, something may be messing with
how the shims are installed.
```

^[1]: https://yarnpkg.com/corepack

analyzer/src/funTest/kotlin/Yarn2Test.kt
```diff
@@ -0,0 +1,89 @@
...
+            val yarn = Yarn2("yarn", workingDir, AnalyzerConfiguration(), RepositoryConfiguration())
+
+            val exception = shouldThrow<IllegalArgumentException> {
+                yarn.command(workingDir)
+            }
+
+            exception.localizedMessage shouldContain executable
+        }
+
+        "return the default executable name if corepack is enabled" {
```
sschuberth, line 76: Probably use the capitalized product name "Corepack" here.

analyzer/src/funTest/kotlin/Yarn2Test.kt
```diff
@@ -0,0 +1,89 @@
...
+import io.kotest.engine.spec.tempdir
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldContain
+
+import org.ossreviewtoolkit.model.config.AnalyzerConfiguration
+import org.ossreviewtoolkit.model.config.PackageManagerConfiguration
+import org.ossreviewtoolkit.model.config.RepositoryConfiguration
+import org.ossreviewtoolkit.plugins.packagemanagers.node.Yarn2
+
+class Yarn2Test : WordSpec({
```
fviernau, line 33: The class should be named `Yarn2FunTest`.

fviernau (dismissed):

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
```diff
@@ -157,6 +171,8 @@ class Yarn2(
     override fun command(workingDir: File?): String {
         if (workingDir == null) return ""
 
+        if (options[OPTION_COREPACK_ENABLED].toBoolean()) return DEFAULT_EXECUTABLE_NAME
```
sschuberth, line 174: IMO it would be nicer if this could be (reliably) auto-detected. Does the `yarn exec env` approach mentioned over [here](https://yarnpkg.com/corepack#installation) work for that?

sschuberth (changes requested):

analyzer/src/funTest/kotlin/Yarn2Test.kt
oheger-bosch: Oops, IntelliJ has created the class at a completely wrong path. I moved it to the test folder of the Node package manager plugin. It is acutally a unit test.

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
oheger-bosch: This is an interesting point, I was also thinking about this, but am not sure how to handle this in the best way. It would also be possible to even go a step further and design the `Yarn2` class in a way that it - optionally - executes the `corepack enable` command, which is required to use Corepack.

My line of thinking was that ORT is rather low-level. It should focus on obtaining the dependencies from package managers and assume that the environment is properly set up. The setup of the environment falls in the responsibility of integrators, e.g. ORT Server. (This is also related to the discussion we had in the last ORT Community meeting.) Now, the integration environment typically knows whether Corepack should be used or not and can directly provide the information to `Yarn2`; so there is no need to do an additional check.

But I am open for other opinions.

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
oheger-bosch: See my [other comment](https://github.com/oss-review-toolkit/ort/pull/8703#discussion_r1616777462).

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
oheger-bosch: Any other opinions or comments on that?

sschuberth: Coincidentally, I've read about Corepack today also in the context of https://github.com/oss-review-toolkit/ort/issues/8715. Could we simply auto-enable the use of Corepack if `package.json` has the [`packageManager` node](https://nodejs.org/api/packages.html#packagemanager)?

oheger-bosch: Well, this does not really play well with the API: Whether Corepack is enabled or not affects the `command` function of `Yarn2`. However, this function has no obvious way to check the status of Corepack. It is called early in the analysis process to determine the tool version. IIUC, at that time, `Yarn2` does not know about the definition files it will have to process.

oheger-bosch: This relies on `resolveDependencies` always being called before `command()`, which may change in a refactoring.

But aside from that, this brings me back to my original question: Even if feasible, should the responsibility to determine the Corepack status not be shifted to the execution environment of ORT? Here, the information is typically available without major hickhacks.

sschuberth: Ah, yes, I forgot to reply to that earlier:

> My line of thinking was that ORT is rather low-level. It should focus on obtaining the dependencies from package managers and assume that the environment is properly set up.

I'd agree to the second sentence, but not so much to the former. Whenever feasible to detect and apply, ORT should use the correct tools and versions automatically. Just out of pure pragmatism we deviated from this rule of thumb for ecosystems where there's no single defined way to declare the tool version, or bootstrapping the tool (on all supported platform) would be infeasible / too much effort.

Or phrased differently: Development teams that take ORT into use should not be required to configure their tools / version for ORT again if they already did so in their build system, as we'd want to avoid teams to maintain basically the same information in multiple different places. That thinking is also the motivation behind the already linked https://github.com/oss-review-toolkit/ort/issues/8715.

However, whether the auto-detection implementation is feasible needs to be decided on a case-by-case basis. And for this particular Yarn / Corepack case, I'm not yet convinced that is is not feasible.

oheger-bosch: Then I propose the following: The default behavior is to auto-detect whether Corepack is enabled based on the content of the `package.json` file. But this behavior can be overridden with the configuration property; then this mechanism can be disabled if it is not needed.

Would this be okay?

oheger-bosch: I have changed the implementation accordingly.

sschuberth: Thanks. But I wonder whether we really needs this:

> But this behavior can be overridden with the configuration property

IMO too many things in ORT are configurable already, not always with a proven use-case (anymore). So, you do already see a concrete case where one would opt-out of auto-detection, or was the configuration added "just in case"?

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
```diff
@@ -119,6 +128,31 @@ class Yarn2(
          * The amount of package details to query at once with `yarn npm info`.
          */
         private const val BULK_DETAILS_SIZE = 1000
+
+        /**
+         * The name of the manifest file used by Yarn 2+.
+         */
+        private const val MANIFEST = "package.json"
```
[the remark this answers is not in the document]
sschuberth: And one more thing: While the trend has not been started with this PR, we should not continue adding `private` constants to a `companion object`. Using a `companion object` makes sense to scope `public` constants only.

IMO these should become `private` top-level constants (some applies to the function), but this can be addressed separately.

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
```diff
@@ -87,6 +88,9 @@ private enum class YarnDependencyType(val type: String) {
  * - *disableRegistryCertificateVerification*: If true, the `yarn npm info` commands called by this package manager will
  *   not verify the server certificate of the HTTPS connection to the NPM registry. This allows to replace the latter by
  *   a local one, e.g. for intercepting the requests or replaying them.
+ * - *corepackEnabled*: If true, this implementation assumes that Yarn has been installed via
```
sschuberth, line 91: IMO the name of the property (and also partly the description) is a bit misleading now, as it sounds as if Corepack would only be supported at all if `corepackEnabled` is set to `true`.

I'd rather propose to rename this e.g. to `corepackOverride` or `forceCorepack` and explain that Corepack support is usually auto-detected, but can be overridden / enforced using this property.

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
oheger-bosch: Yes, for ORT Server (at least our installation), we will probably enable Corepack by default. So, the result of the auto-detection must be overridden. The reason for this is that Corepack does not support other installation options; for instance, it refuses to work with a version of yarn that was installed via the package manager of the OS. This means in our case that if we want to support Corepack, we cannot provide default installations of certain package managers, but must rely on Corepack to do the installation.

plugins/package-managers/node/src/main/kotlin/Yarn2.kt
oheger-bosch: Renamed the property and reworded the description.

sschuberth (approved):

sschuberth:
Test failures are related to outdated Pub tests only, see https://github.com/oss-review-toolkit/ort/pull/8722.

PR #8705, opened 2024-05-27
model: Remove the remaining uses of SortedSet / SortedMap from classes which get serialized

### LINKED ISSUE #6235: model: Eliminate sorted collections

fviernau:
See individual commits.

Fixes #6235.

### DIFF
--- a/model/src/test/kotlin/DependencyGraphTest.kt
+++ b/model/src/test/kotlin/DependencyGraphTest.kt
@@ -35,12 +35,11 @@ class DependencyGraphTest : WordSpec({
                 id("org.apache.commons", "commons-collections4", "4.4"),
                 id(group = "org.junit", artifact = "junit", version = "5")
             )
-            val fragments =
-                sortedSetOf(
-                    DependencyReference(0),
-                    DependencyReference(1),
-                    DependencyReference(2)
-                )
+            val fragments = sortedSetOf(
+                DependencyReference(0),
+                DependencyReference(1),
+                DependencyReference(2)
+            )
             val scopeMap = mapOf(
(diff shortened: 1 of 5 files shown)


model/src/test/kotlin/DependencyGraphTest.kt
```diff
@@ -35,12 +35,11 @@ class DependencyGraphTest : WordSpec({
                 id("org.apache.commons", "commons-collections4", "4.4"),
                 id(group = "org.junit", artifact = "junit", version = "5")
             )
-            val fragments =
-                sortedSetOf(
-                    DependencyReference(0),
-                    DependencyReference(1),
-                    DependencyReference(2)
-                )
+            val fragments = sortedSetOf(
```
sschuberth: Commit message: Probably prefer `style` over `test` as the commit type.

fviernau:
Note to myself: Test in AnalyzerResultBuilder now fails because the assertion now checks:
`AnalyzerResult.equals() -> DependencyGraph.equals() -> scopeRoots.equals()`
which return `false`, while previously, `scopeRoots()` was a `SortedSet` which used `compareTo()`.

fviernau:
No plans to continue with this.

PR #8708, opened 2024-05-27
refactor(clearly-defined): Get enums for serial names directly

sschuberth:
Look up enums for serial names directly instead of searching for them by string representation.

### DIFF
--- a/clients/clearly-defined/src/main/kotlin/Enums.kt
+++ b/clients/clearly-defined/src/main/kotlin/Enums.kt
@@ -43,7 +43,8 @@ enum class ComponentType(val defaultProvider: Provider? = null) {
 
     companion object {
         @JvmStatic
-        fun fromString(value: String) = ComponentType.entries.single { it.toString() == value }
+        fun fromString(value: String) =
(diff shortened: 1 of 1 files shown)


clients/clearly-defined/src/main/kotlin/Enums.kt
```diff
@@ -43,7 +43,8 @@ enum class ComponentType(val defaultProvider: Provider? = null) {
 
     companion object {
         @JvmStatic
-        fun fromString(value: String) = ComponentType.entries.single { it.toString() == value }
+        fun fromString(value: String) =
+            ComponentType.entries[ComponentType.serializer().descriptor.getElementIndex(value)]
```
mnonnenmacher, line 47: Why not use `ComponentType.valueOf(value)` (and maybe remove this function altogether as it seems to do the same as `valueOf`)?

sschuberth: `valueOf()` expects `ComponentType.name` as the `value`, but here we want to construct the enum from the name used for serialization (see `@SerialName`).

mnonnenmacher (approved):

PR #8712, opened 2024-05-28
feat(evaluated-model): Allow for shorter turn-around times seeing narrowed down VCS paths in findings

fviernau:
See individual commits.

### DIFF
--- a/plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
+++ b/plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
@@ -426,21 +426,19 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
                 calculatePackageVerificationCode(fileList.files.map { it.sha1 }.asSequence())
             }.orEmpty(),
             issues = issues
-        )
-
-        val actualScanResult = scanResults.addIfRequired(evaluatedScanResult)
+        ).run { scanResults.addIfRequired(this) }
 
         issues += addIssues(
             result.summary.issues,
             EvaluatedIssueType.SCANNER,
             pkg,
-            actualScanResult,
+            evaluatedScanResult,
             null
         )
 
-        addLicensesAndCopyrights(pkg.id, result, actualScanResult, findings)
+        addLicensesAndCopyrights(pkg.id, result, evaluatedScanResult, findings)
 
-        return actualScanResult
+        return evaluatedScanResult
     }
 
     private fun addDependencyTree(project: Project, pkg: EvaluatedPackage, deduplicateDependencyTree: Boolean) {
--- a/plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
+++ b/plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
@@ -277,9 +277,7 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
 
         issues += addAnalyzerIssues(project.id, evaluatedPackage)
 
-        input.ortResult.getScanResultsForId(project.id).mapTo(scanResults) { result ->
-            convertScanResult(result, findings, evaluatedPackage)
-        }
+        scanResults += convertScanResultsForPackage(evaluatedPackage, findings)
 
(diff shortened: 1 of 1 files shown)


plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
```diff
@@ -426,21 +426,19 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
                 calculatePackageVerificationCode(fileList.files.map { it.sha1 }.asSequence())
             }.orEmpty(),
             issues = issues
-        )
-
-        val actualScanResult = scanResults.addIfRequired(evaluatedScanResult)
+        ).run { scanResults.addIfRequired(this) }
```
sschuberth, line 440: Hmm... but shouldn't `evaluatedScanResult` in line 420 be renamed to `actualScanResult` then? Because `addIfRequired()` might not add the `EvaluatedScanResult` at all, so it's a bit odd to still call it `evaluatedScanResult` in any case, or?

Doing the rename would also reduce the diff below.

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
```diff
@@ -410,6 +406,14 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
         )
     }
 
+    private fun convertScanResultsForPackage(
```
sschuberth, line 410: Maybe make this an extension function on `EvaluatedPackage` and just call it "convertScanResult()"?

Although not having an extension function fits better the existing code... but maybe still use the simpler "convertScanResult()" name, as an overload to the function at line 417?

sschuberth (changes requested):

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
fviernau: I do not really see the problem with calling it `evaluatedScanResult`.

sschuberth: Ah, now I see what confused me: Your first commit renames the `result` function parameter (of type `ScanResult`) to `scanResult`, but there is also the `scanResults` member variable (plural) which is a list of `EvaluatedScanResult`. As the only difference between those variables is the singular vs. plural naming, I thought they would be of the same type (`ScanResult`), which is why I found the name misleading.

From that perspective this particular renaming is ok for me now. Though I still feel that `scanResult` and `scanResults` not being of the same base type is a bit misleading.

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
sschuberth: > but maybe still use the simpler "convertScanResult()" name

So, what do you think about this?

fviernau: I'd tend to keep this as-is, if this is ok for you.

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
```diff
@@ -411,17 +411,17 @@ internal class EvaluatedModelMapper(private val input: ReporterInput) {
     }
 
     private fun convertScanResult(
-        result: ScanResult,
+        scanResult: ScanResult,
```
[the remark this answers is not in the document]
fviernau: > but is it really required as the function is already called "convert_Scan_Result()", and the other parameters also omit the "Evaluated" prefix from the type, for example.

Often times variables named `result` are used for the value which is later on returned by the function.

> hat I'm after is that dropping this commit could help to understand that the "result" function parameter and the "scanResults" member variable are not of the same base type.

.. anyhow, I've dropped the commit.

plugins/reporters/evaluated-model/src/main/kotlin/EvaluatedModelMapper.kt
fviernau: I've dropped the commit which renamed `result`.

sschuberth (approved):

fviernau:
The failing `PubFunTest` was unrelated.

PR #8724, opened 2024-06-04
refactor(scanner,model): Move NestedProvenance to model

pepper-jk:
Moving the `NestedProvenance` data structure to model, allows it to be used within the model without creating a circular dependency between scanner and model.

This would enable a future `AnalayzerInput` class [1] to utilize it and still be placed into the model,
where its predecessor `Repostory` is already located.

[1] https://github.com/oss-review-toolkit/ort/issues/2896#issuecomment-2069742269

Signed-off-by: Jens Keim <jens.keim@forvia.com>

### DIFF
--- a/model/src/main/kotlin/NestedProvenance.kt
+++ b/model/src/main/kotlin/NestedProvenance.kt
@@ -17,13 +17,9 @@
  * License-Filename: LICENSE
  */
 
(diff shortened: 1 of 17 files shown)


pepper-jk:
Closing this PR in favor of a different approach, see https://github.com/oss-review-toolkit/ort/issues/2896#issuecomment-2151813209.

PR #8730, opened 2024-06-06
Misc. SPDX / license mapping related improvements

sschuberth:


sschuberth:
@oss-review-toolkit/kotlin-devs can we agree that it makes sense to merge this now independently of the outcome of the discussion in https://github.com/oss-review-toolkit/ort/pull/8691?

sschuberth:
> @oss-review-toolkit/kotlin-devs can we agree that it makes sense to merge this now independently of the outcome of the discussion in #8691?

I've dropped the actual warning log from the PR, but the other small refactoring commits can still be merged IMO. Please have a look @mnonnenmacher and / or @fviernau.

utils/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -85,6 +94,17 @@ fun String.isSpdxExpressionOrNotPresent(strictness: Strictness = Strictness.ALLO
 fun String.toSpdx(strictness: Strictness = Strictness.ALLOW_ANY): SpdxExpression =
     SpdxExpression.parse(this, strictness)
 
+/**
+ * Parses the string as an [SpdxExpression] of the given [strictness] and returns the result on success, or null if the
+ * string cannot be parsed.
+ */
+fun String.toSpdxOrNull(strictness: Strictness = Strictness.ALLOW_ANY): SpdxExpression? =
```
fviernau, line 101: Parses -> Parse
returns -> return
maybe: the string -> this string

utils/spdx/src/main/kotlin/Extensions.kt
```diff
@@ -85,6 +94,17 @@ fun String.isSpdxExpressionOrNotPresent(strictness: Strictness = Strictness.ALLO
...
+    runCatching {
+        toSpdx(strictness)
+    }.onFailure {
+        logger.debug { "Could not parse license '$this' to SPDX: ${it.collectMessages()}" }
```
fviernau, line 105: As $this is not necessary a license, if it could not be parse, should we just drop the mention of "license" from the log message?

utils/spdx/src/main/kotlin/SpdxExpression.kt
```diff
@@ -107,10 +107,9 @@ sealed class SpdxExpression {
         }
 
     /**
-     * Normalize all license IDs using a mapping containing common misspellings of license IDs. If [mapDeprecated] is
-     * `true`, also deprecated IDs are mapped to their current counterparts. The result of this function is not
-     * guaranteed to contain only valid IDs. Use [validate] to check the returned [SpdxExpression] for validity
-     * afterwards.
+     * Normalize all license IDs using [SpdxSimpleLicenseMapping]. If [mapDeprecated] is `true`, also deprecated IDs are
+     * mapped to their current counterparts. The result of this function is not guaranteed to contain only valid IDs.
+     * Use [validate] to check the returned [SpdxExpression] for validity afterwards.
```
fviernau, line 112: out of scope: Should this be `isValid()` instead of `validate()`, because the latter does not return any value?

utils/spdx/src/main/resources/declared-license-mapping.yml
```diff
@@ -15,6 +15,10 @@
 # SPDX-License-Identifier: Apache-2.0
 # License-Filename: LICENSE
 
+# A mapping from license strings collected from the declared licenses of Open Source packages to SPDX expressions. This
+# mapping only contains license strings which can *not* be parsed by [SpdxExpression.parse], for example because the
+# license names contain white spaces. See [SpdxSimpleLicenseMapping] for a mapping of varied license names.
```
fviernau, line 20: Should it be various instead of varied?

fviernau (changes requested):

utils/spdx/src/main/kotlin/Extensions.kt
sschuberth: I've reworded the message.

utils/spdx/src/main/kotlin/SpdxExpression.kt
sschuberth: I believe the mention of `validate()` better matches the `normalize()` context, but I've now simply mentioned `isValid()` as well.

utils/spdx/src/main/resources/declared-license-mapping.yml
sschuberth: No, "varied" is correct (and also used in the original text that I copied this from) as it's supposed to mean "variants of license names" and not "several license names".

fviernau (approved):

PR #8734, opened 2024-06-07
deps(docker): Downgrade Node to the latest LTS version

sschuberth:
Be able to analyze a wider range of projects by not updating beyond the latest LTS version of Node and its belonging NPM version. Also see the discussion at [1].

[1]: https://github.com/oss-review-toolkit/ort/pull/8674#issuecomment-2154385295


fviernau:
As the next LTS version is right around the corner to become the current LTS version, I'm not certain whether this is ok for the community as opposed to building a custom docker as a workaround, so I prefer to have another approval.

mnonnenmacher:
> As the next LTS version is right around the corner to become the current LTS version, I'm not certain whether this is ok for the community as opposed to building a custom docker as a workaround, so I prefer to have "another" approval or opinion @oss-review-toolkit/core-devs .

I do agree with the policy you discussed in #8674 to always use the current LTS release as default, based on that I will approve the PR. If a new LTS is release within the next few days we can make just another update. But actually I think you misunderstood the the [docs](https://nodejs.org/en/about/previous-releases): Node 22 will switch from "current" to "active" in Q4 and only then it will become the LTS release.

mnonnenmacher (dismissed):

fviernau (approved):

fviernau (dismissed):

mnonnenmacher (approved):

PR #8750, opened 2024-06-11
chore(model): Remove the unused `createMissingArchives` scanner option

sschuberth:
The condition that used that option was removed in c71628d, and there is no plan to reintroduce it as the current provenance-based scanner logic works differently.

### DIFF
--- a/integrations/schemas/ort-configuration-schema.json
+++ b/integrations/schemas/ort-configuration-schema.json
@@ -198,9 +198,6 @@
                 "archive": {
                     "$ref": "#/definitions/Archive"
                 },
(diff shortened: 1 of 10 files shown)


integrations/schemas/ort-configuration-schema.json
mnonnenmacher, line 1: Commit message nit: "options" -> "option"

fviernau (dismissed):

mnonnenmacher (changes requested):

mnonnenmacher (approved):

fviernau (approved):

PR #8751, opened 2024-06-11
refactor(downloader)!: Make `getDefaultBranchName()` non-nullable

sschuberth:
All current (non-test) implementations return a non-null name already as they fall back to a hard-coded name eventually.

### DIFF
--- a/downloader/src/main/kotlin/VersionControlSystem.kt
+++ b/downloader/src/main/kotlin/VersionControlSystem.kt
@@ -179,9 +179,10 @@ abstract class VersionControlSystem(
     abstract fun getVersion(): String
 
     /**
-     * Return the name of the default branch for the repository at [url], or null if there is no default remote branch.
+     * Return the name of the default branch for the repository at [url]. It is expected that there always is a default
+     * branch name that implementations can fall back to, and that the returned name is non-empty.
      */
-    abstract fun getDefaultBranchName(url: String): String?
+    abstract fun getDefaultBranchName(url: String): String
 
     /**
      * Return a working tree instance for this VCS.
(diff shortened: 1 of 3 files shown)


downloader/src/main/kotlin/VersionControlSystem.kt
```diff
@@ -179,9 +179,9 @@ abstract class VersionControlSystem(
     abstract fun getVersion(): String
 
     /**
-     * Return the name of the default branch for the repository at [url], or null if there is no default remote branch.
+     * Return the name of the default branch for the repository at [url].
```
fviernau, line 182: Should this now say: "[..], or empty if there is no default remote branch."?

fviernau (changes requested):

downloader/src/main/kotlin/VersionControlSystem.kt
sschuberth: No, because all current implementations always return a non-empty default branch as a fallback.

fviernau: > No, because all current implementations always return a non-empty default branch as a fallback.

But isn't this an interface of a plugin, so that potential unknown implementations should be considered as opposed to just the ones existing currently within ORT?

sschuberth: While that's technically possible, I'd propose that it's simply not allowed by convention, as I believe there should always by a default branch name that any implementation would be able to fall back to.

sschuberth: I've added more docs / text to the commit message.

PR #8752, opened 2024-06-11
Fix Docker builds for external PRs

sschuberth:

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -55,14 +55,24 @@ jobs:
             type=ref,event=tag
             type=sha
       - name: Build & Push 'ort' Docker Image
+        if: ${{ github.event_name != 'pull_request' }}
         uses: docker/build-push-action@v5
         with:
           context: .
-          push: ${{ github.event_name != 'pull_request' }}
+          push: true
           tags: ${{ steps.meta-ort.outputs.tags }}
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build.yml
```diff
@@ -96,6 +96,7 @@ jobs:
           labels: ${{ steps.meta-ort-minimal.outputs.labels }}
           target: minimal
           cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/ort:cache
+          cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/ort:cache,mode=max
```
mnonnenmacher, line 99: That was left out on purpose, all layers from the ort-minimal image are contained in the full ort image built before, so using the cache of the full image here in cache-from makes the build really fast. Writing the cache would overwrite the cache image with just the layers from the minimal cache. I thought I had documented that, but I can't find it anywhere, so maybe add an explanatory comment instead.

mnonnenmacher (changes requested):

.github/workflows/docker-build.yml
sschuberth: I've added a comment instead.

sschuberth: Ah, right. Thanks, fixed.

mnonnenmacher (approved):

Issue #8695, opened 2024-05-24
Gemfile parsing for Bundler (Ruby) doesn't correctly take into account platforms (ruby, java etc.)


clemens:
### Describe the bug

Disclaimer: Strictly speaking, this is probably a mixture of a bug and a feature request. But since it does cause unexpected error messages in the output, I'll qualify it as a bug in this ticket.

The gist is that in a Ruby project, ORT doesn't seem to take into account Bundler's concept of "platforms" (ruby, java etc.) and subsequently might look up wrong transitive dependencies and/or complain about missing dependencies.

### To Reproduce

- Use GitLab CI setup as listed below
- Use regular Ruby (not JRuby; any version should be fine)
- Use a very minimal Gemfile like below with a gem that has different dependencies depending on platform, e.g. `psych`, which [depends on `stringio` on the `ruby` platform](https://rubygems.org/gems/psych/versions/5.1.2), but [`jar-dependencies` on the `java` platform](https://rubygems.org/gems/psych/versions/5.1.2-java)
- Run `bundle install`
- Push to GitLab to ensure job runs

Note that running `bundle lock --add-platform java` also doesn't make a difference – while it includes the dependencies correctly in `Gemfile.lock`, the error persists

``` yaml
# GitLab CI configuration
include:
  - https://raw.githubusercontent.com/oss-review-toolkit/ort-ci-gitlab/main/templates/ort-scan.yml

ort-scan:
  stage: test
  extends: .ort-scan
  variables:
    ORT_LOG_LEVEL: debug
```

``` ruby
# Minimal Gemfile example

source 'https://rubygems.org'

gem 'psych', '5.1.2'
```

### Expected behavior

- I'd like to not see errors related to platforms that I don't care about (= haven't declared in my `Gemfile.lock`). In the example, I'd expect to not see an error that tells me that `jar-dependencies` couldn't be found.
- I'd like dependencies to be handled correctly for all platforms that I specify – in other words, if the `Gemfile.lock` includes only `psych (5.1.2)`, ORT should correctly handle it with its `ruby` platform dependencies; if it includes `psych (5.1.2-java)`, ORT should correctly handle it with its `java` platform dependencies; and if it includes multiple platforms, ORT should handle both versions

[...]

PR #8754, opened 2024-06-12
Enforce "ruby" platform when fetching version data

clemens:
When fetching version data for individual Ruby gems, we don't want to rely on creation order of the gems on RubyGems.org. Instead, we use the standard `ruby` platform as a sensible default (as we do elsewhere).

Partially addresses https://github.com/oss-review-toolkit/ort/issues/8695.

### DIFF
--- a/plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
+++ b/plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
@@ -358,8 +358,11 @@ class Bundler(
         workingDir.walk().maxDepth(1).filter { it.isFile && it.extension == "gemspec" }.firstOrNull()
 
     private fun queryRubygems(name: String, version: String, retryCount: Int = 3): GemSpec? {
-        // See http://guides.rubygems.org/rubygems-org-api-v2/.
-        val url = "https://rubygems.org/api/v2/rubygems/$name/versions/$version.yaml"
+        // NOTE: Explicitly use platform=ruby here to enforce the same behavior here that is also used in the Bundler
+        //       resolving logic.
+        // See plugins/package-managers/bundler/src/main/resources/resolve_dependencies.rb
+        // See <http://guides.rubygems.org/rubygems-org-api-v2/>.
+        val url = "https://rubygems.org/api/v2/rubygems/$name/versions/$version.yaml?platform=ruby"
 
         return okHttpClient.downloadText(url).mapCatching {
             GemSpec.createFromGem(yamlMapper.readTree(it))


plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
```diff
@@ -359,7 +359,8 @@ class Bundler(
 
     private fun queryRubygems(name: String, version: String, retryCount: Int = 3): GemSpec? {
         // See http://guides.rubygems.org/rubygems-org-api-v2/.
-        val url = "https://rubygems.org/api/v2/rubygems/$name/versions/$version.yaml"
+        // See discussion in https://github.com/oss-review-toolkit/ort/issues/8695 on why we use the "platform=ruby".
```
fviernau, line 362: What do you think about adjusting the comment to summarize the rationale, and then link the discussion with a "see also $link"? ...Because, the whole discussion is quite lengthy to read through.

plugins/package-managers/bundler/src/funTest/assets/projects/synthetic/lockfile/Gemfile
```diff
@@ -4,6 +4,7 @@ source "https://rubygems.org"
 
 gem 'nokogiri'
 gem 'rack', '~>2.1'
+gem 'psych'
```
fviernau, line 7: Can you help me understanding whether it is necessary to and why you choose to update the lockfiles in this same commit, as opposed to a separate commit?

fviernau (changes requested):

plugins/package-managers/bundler/src/funTest/assets/projects/synthetic/bundler-expected-output-lockfile.yml
```diff
@@ -381,3 +463,35 @@ packages:
...
+  vcs:
+    type: "Git"
+    url: "https://github.com/ruby/stringio.git"
+    revision: ""
+    path: ""
+  vcs_processed:
+    type: "Git"
+    url: "https://github.com/ruby/stringio.git"
+    revision: ""
+    path: ""
```
sschuberth, line 497: Missing final newline.

plugins/package-managers/bundler/src/funTest/assets/projects/synthetic/lockfile/Gemfile
sschuberth: I guess the question is not about the lockfile actually, but why this dependency was added to the test `Gemfile` (implying lockfile updates) in the first place.

If you want to showcase the effect of your change at the example of this new dependency, IMO the dependency should be added in a preceding commit (incl. any necessary lockfile changes), and then a new commit should introduce your actual change and the necessary test assets changes.

So in short, separate the addition of new test dependencies from the effect of code changes.

plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
sschuberth: I'd actually prefer to not link to our own issue discussions at all, but keep the issue self-contained. To me, the gist is to make the platform when querying metadarta consistent with the platform we resolve dependencies for at

https://github.com/oss-review-toolkit/ort/blob/503edeedc8de2bee866e2d308ac5473b69f1d42e/plugins/package-managers/bundler/src/main/resources/resolve_dependencies.rb#L31-L32

plugins/package-managers/bundler/src/funTest/assets/projects/synthetic/lockfile/Gemfile
clemens: The intention here is that the psych gem is one example where the logic currently "breaks" in the sense that due to the order in which the different versions of the gem have been uploaded to RubyGems.org, fetching the data without an explicit `platform` parameter returns the Java version of it instead of the standard Ruby version.

Check these 2 requests:

- `curl 'https://rubygems.org/api/v2/rubygems/psych/versions/5.1.2.yaml'`
- `curl 'https://rubygems.org/api/v2/rubygems/psych/versions/5.1.2.yaml?platform=ruby'`

In contrast, other gems like nokogiri, which happen to have a different upload order, return the standard Ruby version first:

- `curl 'https://rubygems.org/api/v2/rubygems/nokogiri/versions/1.16.5.yaml'`
- `curl 'https://rubygems.org/api/v2/rubygems/nokogiri/versions/1.16.5.yaml?platform=java'`

According to RubyGems' own source code, [this (arguably non-deterministic) behavior is intentional](https://github.com/rubygems/rubygems.org/blob/master/app/models/rubygem.rb#L146-L159).

---

That being said: I had assumed that when adding the gem to the test `Gemfile`, some test in addition to the one that did fail (due to an additional gem being installed) would fail because of the change. But it didn't, so I could probably revert these changes.

clemens:
@sschuberth @fviernau Thanks for the review comments. I had intentionally marked it as a Draft MR in order not to bother you to review yet, while I was sorting out the build etc. :) I guess my intention failed there.

Either way: It seems that there currently isn't a test that fails with the `Gemfile` changes aside from the obvious one which checks the resulting list.

My question here would be: If I revert all the changes to the `Gemfile` and `Gemfile.lock` and keep the change to the URL plus add a comment that references to the corresponding `resolve_dependencies.rb` change that it mimicks, would the change be acceptable for you even without an explicit test for it?

I just ran it in CI and here's what I'm getting:

Before the change:

Screenshot 2024-06-12 at 13 49 41

After the change:

Screenshot 2024-06-12 at 13 49 54

plugins/package-managers/bundler/src/funTest/assets/projects/synthetic/lockfile/Gemfile
sschuberth: > The intention here is that the psych gem is one example where the logic currently "breaks"

Yeah, then this new dependency plus needed test assest chnages should be added in a separate commit before the commit that makes the code changes.

However...

> But it didn't, so I could probably revert these changes.

...if that sentence means there would be no `psych`-related changes to the test assets before and after your change, then I'd indeed drop the introduction of that dependency for simplicity.

sschuberth:
> I guess my intention failed there.

No worries. Sometimes we're just curious and willing to spend time on reviewing draft PRs, well knowing that our comments might become null and void due to ongoing changes.

> would the change be acceptable for you even without an explicit test for it?

For me, it would be acceptable, as it's obviously doing the right thing, even if you cannot prove that it makes a difference according to test assets.

> After the change:

Interesting that this writes out a smaller file although it has the same number of projects / packages. Mind investigating where the size difference comes from?

clemens:
The main difference that I can spot is that the `issues` are captured in the `analyzer-result.json`:

Bundler Issues

There are 32 such entries. If I extract a single one and put it into a file, it's roughly 350-400 bytes (depending on the length of the respective project name) => 400 x 32 = ~12 KB. This comes fairly close to the (presumably rounded) 20 KB difference that the output lists.

plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
```diff
@@ -358,8 +358,11 @@ class Bundler(
...
     private fun queryRubygems(name: String, version: String, retryCount: Int = 3): GemSpec? {
-        // See http://guides.rubygems.org/rubygems-org-api-v2/.
-        val url = "https://rubygems.org/api/v2/rubygems/$name/versions/$version.yaml"
+        // NOTE: We explicitly use platform=ruby here to enforce the same behavior here that we also have in our
+        // Bundler resolving logic.
+        // See <https://github.com/oss-review-toolkit/ort/blob/main/plugins/package-managers/bundler/src/main/resources/resolve_dependencies.rb#L31-L32>
+        // See <http://guides.rubygems.org/rubygems-org-api-v2/>.
+        val url = "https://rubygems.org/api/v2/rubygems/$name/versions/$version.yaml?platform=ruby"
 
         return okHttpClient.downloadText(url).mapCatching {
```
fviernau, line 367: Merge commits are not allowed in this repository. Please do rebase this commit onto head of main instead.

fviernau (dismissed):

plugins/package-managers/bundler/src/main/kotlin/Bundler.kt
```diff
@@ -358,8 +358,11 @@
         workingDir.walk().maxDepth(1).filter { it.isFile && it.extension == "gemspec" }.firstOrNull()
 
     private fun queryRubygems(name: String, version: String, retryCount: Int = 3): GemSpec? {
-        // See http://guides.rubygems.org/rubygems-org-api-v2/.
-        val url = "https://rubygems.org/api/v2/rubygems/$name/versions/$version.yaml"
+        // NOTE: We explicitly use platform=ruby here to enforce the same behavior here that we also have in our
+        // Bundler resolving logic.
+        // See <https://github.com/oss-review-toolkit/ort/blob/main/plugins/package-managers/bundler/src/main/resources/resolve_dependencies.rb#L31-L32>
```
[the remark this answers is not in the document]
sschuberth: Please use passive voice in code comments and commit messages, i.e. do not use "we" or "I" etc. Also see our [contribution guideline WRT the use of "imperative mood"](https://github.com/oss-review-toolkit/.github/blob/main/CONTRIBUTING.md#git-history).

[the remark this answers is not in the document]
clemens: Understood!

Should I `@Suppress("MaxLineLength")` on the line with the long URL or would you want to treat this differently?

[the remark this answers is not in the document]
sschuberth: I think URLs are allowed to exceed the line length. But anyway maybe just use a relative path to the file instead of a URL.

clemens:
So I adapted everything, but now we seem to have what looks like random/unrelated test failures…? I'm a bit stuck here. :/

sschuberth:
> I'm a bit stuck here. :/

No worries. We'll realize this and eventually retrigger flaky tests etc.

sschuberth:
> So I adapted everything

The commit message still talks about "we". Please reword as described above.

sschuberth (approved):

PR #8758, opened 2024-06-13
chore(conan): Update test fixture

rettichschnidi:
This change is a (not strictly needed) preparation for Conan 2.x support in ORT.

Updating the Poco recipe to the current version offered by the Conan Center Index allows it to work with both, version 1.x and 2.x of Conan.

The lock file has been created using Conan version 1.64.1:

```shell
rm -rf ~/.conan/data
conan lock create --lockfile-out lockfile.lock conanfile.py
```

Conan Center Index: https://github.com/conan-io/conan-center-index
Commit: f78dc3729f97c1554eae5f3843cabb2e2198c09a

### DIFF
--- a/plugins/package-managers/conan/src/funTest/assets/projects/synthetic/conan-expected-output-py.yml
+++ b/plugins/package-managers/conan/src/funTest/assets/projects/synthetic/conan-expected-output-py.yml
@@ -1,17 +1,14 @@
 ---
 project:
-  id: "Conan::Poco:1.9.3"
+  id: "Conan::poco:null"
   definition_file_path: "plugins/package-managers/conan/src/funTest/assets/projects/synthetic/conan-py/conanfile.py"
-  authors:
-  - "The Author"
   declared_licenses:
-  - "The Boost Software License 1.0"
+  - "BSL-1.0"
   declared_licenses_processed:
-    unmapped:
-    - "The Boost Software License 1.0"
+    spdx_expression: "BSL-1.0"
   vcs:
     type: "Git"
-    url: "http://github.com/pocoproject/conan-poco.git"
+    url: "https://github.com/conan-io/conan-center-index.git"
     revision: ""
     path: ""
   vcs_processed:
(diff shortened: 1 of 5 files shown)


plugins/package-managers/conan/src/funTest/assets/projects/synthetic/conan-expected-output-py.yml
```diff
@@ -1,36 +1,197 @@
 ---
 project:
-  id: "Conan::Poco:1.9.3"
+  id: "Conan::poco:"
```
sschuberth, line 3: Looks like the version is gone here as it's now also missing from `conanfile.py`; however, we should add it back to have comparable results.

rettichschnidi: I put it back. In a separate commit to make it clear what exactly got modified and why.

sschuberth: So, are you saying that Poco 1.9.3 upstream *did* specific a version, but Poco 1.13.3 upstream does not anymore? Or was our `conanfile.py` for Poco 1.9.3 not matching upstream in the first place?

plugins/package-managers/conan/src/funTest/assets/projects/synthetic/conan-expected-output-py.yml
```diff
@@ -1,6 +1,6 @@
 ---
 project:
-  id: "Conan::poco:"
+  id: "Conan::poco:1.13.3"
```
sschuberth, line 3: I believe this should regenerate the lockfile as well then.

plugins/package-managers/conan/src/funTest/assets/projects/synthetic/conan-expected-output-py.yml
rettichschnidi: > So, are you saying that Poco 1.9.3 upstream _did_ specific a version, but Poco 1.13.3 upstream does not anymore?

Yes

The new upstream recipe works for multiple versions of Poco. The versions are now defined in [conandata.yml](https://github.com/conan-io/conan-center-index/blob/master/recipes/poco/all/conandata.yml).

> Or was our `conanfile.py` for Poco 1.9.3 not matching upstream in the first place?

With respect to the version field, it was matching this (latest) upstream [version](https://github.com/pocoproject/conan-poco/blob/b2ead4e2a5b3f8ef85b386a6d2dff101f4c7c0a7/conanfile.py).

There where however two local modifications (c43b09d79e5d1fa2e269a3d71f9d53406f400892, 8f2869aea45b386cdfa42d1c3950f7fcde141e72) on top of the upstream version.

My preferred solution would be:

1. Keep this file exactly as upstream
2. Relay on the [synthetic test](https://github.com/oss-review-toolkit/ort/blob/f545e5ea3d917129b0e52ed4329d4a6e75e51d1b/plugins/package-managers/spdx/src/funTest/assets/projects/synthetic/subproject-conan/subproject/conanfile.py) to verify the version field functionality.

sschuberth: As [`conandata.yml` seem to be an actual Conan feature](https://docs.conan.io/1/reference/config_files/conandata.yml.html), not a Poco invention, I believe my favorite would be a third option to actually add support for `conandata.yml` in ORT, so that we'd properly get the version here.

Let's see if I can find some time for that...

rettichschnidi: Is there really a need for this?

My understanding is that in the real world, the top level `conanfile.txt`/`.py` files would specify the (one and only) wanted version of Poco and everything works as expected already as of today.

sschuberth: After taking a look again, I believe you're correct. Real-world Conan projects should (but apparently don't have to) declare their version as part of `conanfile.py`, and that version is used for a lookup in `conandata.yml` eventually, but `conandata.yml` itself does not declare the version of the project.

rettichschnidi: I will then remove the version, keep the Poco `conanfile.py` exactly the same as upstream.

rettichschnidi: Done

plugins/package-managers/conan/src/funTest/assets/projects/synthetic/conan-expected-output-py.yml
```diff
@@ -1,36 +1,197 @@
 ---
 project:
-  id: "Conan::Poco:1.9.3"
+  id: "Conan::poco:null"
```
sschuberth, line 3: We should actually never get `null` for the version of a project. Let me take over your PR and fix this on top of it. Nothing to be done from your side @rettichschnidi.

sschuberth (approved):

Issue #8383, opened 2024-03-07
Conan 2 support


rettichschnidi:
It seems like ORT currently supports only Conan 1.x. For my needs, I need support for version 2.

Questions:
- Are there any fundamental and/or hidden challenges to expect or can I expect adding support for it to be straightforward?
- For me, creating a new conan2 package-manager plugin (based on the existing one) would be the easiest way to achieve my goal. Do you consider this an acceptable approach? Or would the existing Conan plugin need to be modified to support both versions?

sschuberth:
> For me, creating a new conan2 package-manager plugin (based on the existing one) would be the easiest way to achieve my goal.

The way the ORT analyzer works is that it needs to be able to decide "from the outside" of the project under analysis (like by files being present or not), which package manager implementation it should use. So if there is a reliable way to tell whether a project is a Conan 1 or Conan 2, then implementing Conan 2 as a separate plugin should work straightly.

As I'm not a Conan expert, for me there are a bunch of more questions to clarify, like: Is Conan 2 backwards compatible to Conan 1? So could we use *only* a Conan 2 binary to power both Conan 1 and Conan 2 analyzers? Or only have a single analyzer plugin for both on the ORT side after all?

Maybe @jsteinhofff can chime in here?

sschuberth:
I guess this is good read:
- https://docs.conan.io/2/whatsnew.html
- https://docs.conan.io/1/conan_v2.html

rettichschnidi:
> The way the ORT analyzer works is that it needs to be able to decide "from the outside" of the project under analysis (like by files being present or not), which package manager implementation it should use. So if there is a reliable way to tell whether a project is a Conan 1 or Conan 2, then implementing Conan 2 as a separate plugin should work straightly.

With some support of the user, this should be possible by adding [required_conan_version](https://docs.conan.io/1/reference/conanfile/other.html#requiring-a-conan-version-for-the-recipe) to a `conanfile.py`. This does not seem to be possible for `conanfile.txt` files.

Some additional thoughts/questions
- ORT runs just one project at a time, correct? If so, can we just deduct the used Conan version from the binary available on the system. There would be no question whether version 1 or 2 should be used.
  - If not ...:
    - Can we expect ORT users to to set the `required_conan_version` variable in their `conanfile.py`?
      - This would imply that `conanfile.txt` users need  migrate to `conanfile.py` (easy to do) if they want to stick with Conan 1.x.
[...]

PR #8759, opened 2024-06-13
chore(conan): Avoid deprecated section name

rettichschnidi:
This change is a preparation for Conan 2.x support in ORT.

With Conan 2, the section `build_requires` in conanfile.txt has been renamed to `tool_requires`. The old name is no longer supported and causes Conan (2.x) to fail with an error:

```shell
$ conan install conanfile.txt
...
ConfigParser: Unrecognized field 'build_requires'
```

Starting with Conan 1.44, `tool_requires` has been added to Conan 1.x as an alias for `build_requires`. Ever since, the official documentation does not mention a section named `build_requires` anymore.

Current 1.x docs: https://docs.conan.io/en/1.64/reference/conanfile_txt.html
Old 1.x docs: https://docs.conan.io/en/1.43/reference/conanfile_txt.html
Relevant issue: https://github.com/conan-io/conan/pull/10168


sschuberth:
I have no problems with raising the min. supported Conan version to 1.44.0, but we should double-check with the community, so I've added it to the [agenda of today's call](https://github.com/oss-review-toolkit/ort/wiki/ORT-Community-Meeting#2024-06-13) (which I cannot attend, however).

However, dosing this as a preparation for Conan 2 support probably does not count, as over [here](https://github.com/oss-review-toolkit/ort/issues/8383) we assumed that Conan 2 support would be implemented via a new independent package manager with its own test assets.

sschuberth:
> I've added it to the [agenda of today's call](https://github.com/oss-review-toolkit/ort/wiki/ORT-Community-Meeting#2024-06-13) (which I cannot attend, however).

I see that the community had no objections during the call, so let's move forward with this, even if not strictly required for Conan 2 support, which still could be implemented as a separate analyzer plugin.

sschuberth (approved):

PR #8764, opened 2024-06-17
Draft: feat(Repository): Use KnownProvenance instead of VcsInfo

pepper-jk:
This PR is still an early draft.

In order to allow source artifacts as well as local source code to
be scanned, we require a more generallized Repository, which allows
any type of `KnownProvenance`. While we still set the signature of
`Repository` to allow `KnownProvenance`s, this commit focuses on
accomidateing `RepositoryProvenance` as then main input for now.

Therefore we wrap `VcsInfo` with `RepositoryProvenance`, whenever
it is used as input, and unwrap it, when it is produced as output.
This gives us a quick update of the now depreacted code, without
the necessity to support all `KnownProvenance` types from the get
go.

This commit compiles without errors, but still fails tests, due
to the now changed `OrtResult` structure and output. Fixed tests
must be squshed into this commit before merging.

Signed-off-by: Jens Keim <jens.keim@forvia.com>


model/src/main/kotlin/VcsInfo.kt
```diff
@@ -83,6 +83,11 @@ data class VcsInfo(
      * Return a [VcsInfoCurationData] with the properties from this [VcsInfo].
      */
     fun toCuration() = VcsInfoCurationData(type, url, revision, path)
+
```
sschuberth, line 86: Commit message nits:
- Duplicate "method" in first sentence.
- Please use passive voice, i.e. no "we", "I", "us" etc.

model/src/main/kotlin/VcsInfo.kt
```diff
@@ -83,6 +83,11 @@ data class VcsInfo(
...
+    /**
+     * Return true if this vcs information matches the other vcs information.
```
sschuberth, line 88: s/vcs/VCS/

sschuberth: The docs for this function require some special attention: The matching disregards the `path` property (otherwise just `vcsInfoA == vcsInfoB` could have been used for matching). So in a way this is a bit similar to a `PackageCuration`'s `isApplicableDisregardingVersion()` function, and I'd name it as `equalsDisregardingPath()`.

model/src/main/kotlin/Provenance.kt
```diff
@@ -79,6 +87,13 @@ data class RepositoryProvenance(
      * Return true if this provenance matches the processed VCS information of the [package][pkg].
      */
     override fun matches(pkg: Package): Boolean = vcsInfo == pkg.vcsProcessed
+
+    /**
+     * Return true if the [Provenance][other] is a [RepositoryProvenance] and its normalized vcsInfo matches this
+     * provenance's normalized [VcsInfo][vcsInfo].
+     */
+    override fun matches(other: Provenance): Boolean =
+        other is RepositoryProvenance && other.vcsInfo.normalize().matches(vcsInfo.normalize())
```
sschuberth, line 98: Why is this disregarding the `resolvedRevision`? And if that's a mistake, why not simply use `provenanveA == provenanceB` for matching `Provenance`s?

model/src/main/kotlin/Provenance.kt
```diff
@@ -80,7 +80,9 @@ data class RepositoryProvenance(
     val resolvedRevision: String
 ) : KnownProvenance {
     init {
-        require(resolvedRevision.isNotBlank()) { "The resolved revision must not be blank." }
+        require(vcsInfo == VcsInfo.EMPTY || resolvedRevision.isNotBlank()) {
```
sschuberth, line 83: Commit message nits:
- s/vcs/VCS/
- s/git/Git/
- Odd indentation before "Both"
- Again use passive voice

sschuberth: Logic-wise, I'd prefer to have the clause order swapped. Usually we require `resolvedRevision.isNotBlank()` but if it is blank, the whole `VcsInfo` must be empty: `resolvedRevision.isNotBlank() || vcsInfo == VcsInfo.EMPTY`.

analyzer/src/main/kotlin/Analyzer.kt
```diff
@@ -42,6 +42,7 @@ import org.ossreviewtoolkit.model.AnalyzerResult
 import org.ossreviewtoolkit.model.AnalyzerRun
 import org.ossreviewtoolkit.model.OrtResult
 import org.ossreviewtoolkit.model.Repository
+import org.ossreviewtoolkit.model.RepositoryProvenance
```
sschuberth, line 45: Commit message:
- Again use passive voice
- Typo in "generallized" -> "generalized"
- "Repository" -> "`Repository` class" (use backticks so it's clear you refer to the class name)
- "accomidateing" -> "accommodating"
- "then main" -> "the main"
- "depreacted" -> "deprecated"
- "realted" -> "related"

model/src/main/kotlin/Repository.kt
```diff
@@ -31,48 +31,51 @@ data class Repository(
...
-    val nestedRepositories: Map<String, VcsInfo> = emptyMap(),
+    val nestedRepositories: Map<String, KnownProvenance> = emptyMap(),
 
     /**
      * The configuration of the repository, parsed from [ORT_REPO_CONFIG_FILENAME].
      */
     val config: RepositoryConfiguration = RepositoryConfiguration()
 ) {
     companion object {
+        private const val DEFAULT_REVISION = ""
```
sschuberth, line 49: This is more "empty" than "default". And I'd simply inline it.

model/src/main/kotlin/Repository.kt
```diff
@@ -31,48 +31,51 @@ data class Repository(
...
         val EMPTY = Repository(
-            vcs = VcsInfo.EMPTY,
-            vcsProcessed = VcsInfo.EMPTY,
+            provenance = RepositoryProvenance(VcsInfo.EMPTY, DEFAULT_REVISION),
             nestedRepositories = emptyMap(),
             config = RepositoryConfiguration()
         )
     }
 
+    fun vcs(): VcsInfo {
```
sschuberth, line 62: Alternatively, this could be a getter-property annotated with `JsonIgnore`. Same below.

model/src/main/kotlin/Provenance.kt
```diff
@@ -80,7 +80,9 @@ data class RepositoryProvenance(
     val resolvedRevision: String
 ) : KnownProvenance {
     init {
```
fviernau, line 82: Why haven't you decided for keeping the require as-is, and make the tests construct instances with non-empty resolved revision?

I'm asking to me it seems that if repository provenance is sued, the VCS info should not be empty, because otherwise it would be an Unkown provenance, wouldn't it?

fviernau (changes requested):

model/src/main/kotlin/Repository.kt
```diff
@@ -31,48 +31,51 @@ data class Repository(
...
-    val vcsProcessed: VcsInfo = vcs.normalize(),
+    val provenance: KnownProvenance,
 
     /**
      * A map of nested repositories, for example Git submodules or Git-Repo modules. The key is the path to the
      * nested repository relative to the root of the main repository.
      */
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
-    val nestedRepositories: Map<String, VcsInfo> = emptyMap(),
+    val nestedRepositories: Map<String, KnownProvenance> = emptyMap(),
```
fviernau, line 41: Nested repositories can remain VcsInfo or alternatively repository provenance, because by definition this can only be a repository.

fviernau (changes requested):

fviernau:
Can we have a GitHub issue which describes exactly the goal of this effort? In particular, mention explicitly in which cases we want to support local file provenance. (e.g. analyzer, scanner, scan storages, ...). I'm still worried about the added complexity of this refactoring, compared to the value it will provide, but having the goal clearly statet would at least ensure we are reviewing against the same outlined goal.

sschuberth:
> Can we have a GitHub issue which describes exactly the goal of this effort?

I have started such an issue at https://github.com/oss-review-toolkit/ort/issues/8803 for people to comment on.

model/src/main/kotlin/VcsInfo.kt
pepper-jk: overhauled the commit message after the changes to the commit.

model/src/main/kotlin/VcsInfo.kt
pepper-jk: When I change the name and the doc, should the same doc changes be applied to the `matches` method in `RespositoryProvenance` in the next commit:
https://github.com/oss-review-toolkit/ort/pull/8764/commits/f43b5d9c97c427d8b8ae41561514660e44828116#diff-db17c034876de657175ab28d452f2812913baba034a764c609990260ce101751

sschuberth: Not necessarily. The current docs already make clear the only the normalized VCS info is compared, so I think we're good.

model/src/main/kotlin/Provenance.kt
pepper-jk: It was not done on purpose.

> why not simply use `provenanveA == provenanceB` for matching `Provenance`s?

Because `Repository` used the `equalsDisregardingPath` style `matches` before, and I wanted to ensure the same behavior as before. IIRC some tests would also fail with just `==` between `VcsInfo`, which `provenanveA == provenanceB` would imply, correct?

sschuberth: > which `provenanveA == provenanceB` would imply, correct?

Yes, `==` on `data class`es is a "recursive equals()".

model/src/main/kotlin/Provenance.kt
pepper-jk: You are probably correct. Changing the tests¹ to use `"main"` as revision instead of blank, like I did for [model/src/test/kotlin/OrtResultTest.kt](https://github.com/oss-review-toolkit/ort/pull/8764/files?file-filters%5B%5D=.kt&show-viewed-files=true#diff-c52bc83bc22aea9f28992016b2b705f488d2f9f231508bd1705cd1f9b2754c9e), might be the better way to solve the issue with the tests, than allowing blank `resolvedRevision`s on for `VcsInfo.Empty`.

However, the [Analyzer](https://github.com/oss-review-toolkit/ort/pull/8764/commits/2f69343732b3629d16b47feb9edfa3c5cf8f6c80?file-filters%5B%5D=.kt&show-viewed-files=false#diff-35aa45acd6a9c6882501f95739dbb3cb4ec27cf6ec36089d5473b967b8bd2e8f) also allows `VcsInfo.Empty` to be passed from `workingTree` to the `Repository`. And with `VcsInfo.Empty` `workingTree` will very likely also return an empty revision. So in order to not change that behavior of the Analyzer and Repository in this case, I added this exception to `RepositoryProvenance`.

I have not yet noticed any issues regarding non-`Repository` uses of the `RepositoryProvenance`.

Still you are bringing up some good points, maybe the `Analyzer` behavior could change in this case? Then we could revert the exception to `RepositoryProvenance`. We could dial this in once we handle other `Provenance`s in the `Analyzer`.

¹ [evaluator/src/test/kotlin/TestData.kt](https://github.com/oss-review-toolkit/ort/pull/8764/files?file-filters%5B%5D=.kt&show-viewed-files=true#diff-45f6c3132de3d7712b3032b29f22f011ea18979ab3f9f0d2c744354dc4b15a19)
[model/src/test/kotlin/licenses/TestData.kt](https://github.com/oss-review-toolkit/ort/pull/8764/files?file-filters%5B%5D=.kt&show-viewed-files=true#diff-0cd329b641afb2f5622dad4a87a827922b7580fda2e8dad92def8c12c26b3203)
[reporter/src/testFixtures/kotlin/TestData.kt](https://github.com/oss-review-toolkit/ort/pull/8764/files?file-filters%5B%5D=.kt&show-viewed-files=true#diff-00f03b3f88f750595dffb7115e0e27eea71f21c58426e7d1d36b3d54974d9efe)

model/src/main/kotlin/Provenance.kt
pepper-jk: >  * s/vcs/VCS/

`vcs` is part of a test case name string. so changing the capitalization would erode the reference.
Maybe I should change the format to `"fail if no vcs matches"` to make this more clear and also explicitly call them test cases?

model/src/main/kotlin/Repository.kt
pepper-jk: So Provenances have no hierarchy? The information required to create `nestedRepositories` is only available when using Git as a VCS?

model/src/main/kotlin/Repository.kt
pepper-jk: Done. Let's see if the tests like me.

model/src/main/kotlin/Repository.kt
pepper-jk: Done.

model/src/main/kotlin/Provenance.kt
pepper-jk: Swapped the logic around.

model/src/main/kotlin/Provenance.kt
pepper-jk: Added a comparison of `resolvedRevision`.

model/src/main/kotlin/VcsInfo.kt
pepper-jk: Done.

model/src/main/kotlin/VcsInfo.kt
pepper-jk: Pushed new commits with code changes only.
I will squash them once they are reviewed and accepted.
Afterwards I will update the commit messages and force push.

model/src/main/kotlin/Repository.kt
sschuberth: Correct, `Git` and `GitRepo` are the only implementors of `WorkingTree.getNested()`.

model/src/main/kotlin/VcsInfo.kt
pepper-jk: Just pushed the squashed commits with the overhauled messages.

model/src/main/kotlin/Repository.kt
pepper-jk: I have now reverted the `nestedRepositories` changes (https://github.com/oss-review-toolkit/ort/commit/ac06be18706675938b99da8a6b3a2e1f8173551e) and will squash them into the prior commit (https://github.com/oss-review-toolkit/ort/commit/665ab205ade0560c97e1a4929b07398c1d8fc484) once all checks pass.

Rebasing on current main was also no issue, so I guess I will be done soon.

pepper-jk: Just pushed the rebased and squashed version.

pepper-jk:
@sschuberth @fviernau I know, you are still considering if this refactoring is the right way, but if you could find the time to review the current changes that would be great.

Since most changes regard code quality and commit messages, it would be great to know if the changes know meet your standards. Not only for this PR, but also for future work.

pepper-jk:
Closed in favor of #9476.

PR #8768, opened 2024-06-18
Rename some classes according to Retrofit conventions

sschuberth:


clients/osv/src/main/kotlin/OsvClient.kt
```diff
@@ -0,0 +1,129 @@
...
+import java.util.concurrent.atomic.AtomicReference
+
+import okhttp3.OkHttpClient
+
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+
+/**
+ * This class wraps the OSV Rest API client in order to make its use simpler and less error-prone.
```
fviernau, line 34: This should now say: "..wraps the OSV service"

clients/osv/src/main/kotlin/OsvClient.kt
```diff
@@ -0,0 +1,129 @@
...
+ */
+class OsvClient(serverUrl: String? = null, httpClient: OkHttpClient? = null) {
+    private val client = OsvService.create(serverUrl, httpClient)
```
fviernau, line 37: this should now be named `service`.

clients/osv/src/main/kotlin/OsvService.kt
```diff
@@ -19,111 +19,136 @@
...
-        } else {
-            Result.failure(IOException(response.message()))
+interface OsvService {
+    companion object {
+        const val BATCH_REQUEST_MAX_SIZE = 1000
+
+        val JSON = Json { namingStrategy = JsonNamingStrategy.SnakeCase }
+
+        /**
+         * Create an OsvApiClient instance for communicating with the given [server], optionally using a pre-built
```
fviernau, line 53: Also `OsvApiClient` is obsolete now.

plugins/advisors/osv/src/main/kotlin/Osv.kt
```diff
@@ -69,7 +69,7 @@ class Osv(name: String, config: OsvConfiguration) : AdviceProvider(name) {
 
     override val details: AdvisorDetails = AdvisorDetails(providerName, enumSetOf(AdvisorCapability.VULNERABILITIES))
 
-    private val service = OsvService(
+    private val service = OsvClient(
```
fviernau, line 72: here as well

fviernau (changes requested):

clients/osv/src/main/kotlin/OsvClient.kt
```diff
@@ -0,0 +1,129 @@
...
+import okhttp3.OkHttpClient
+
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+
+/**
+ * This class wraps the OSV Rest API client in order to make its use simpler and less error-prone.
+ */
+class OsvClient(serverUrl: String? = null, httpClient: OkHttpClient? = null) {
```
fviernau, line 36: As you're renaming this class, I believe a better name would be `OsvServiceWrapper`. It's inline with the intention and the already existing class docs.

fviernau (changes requested):

clients/osv/src/main/kotlin/OsvClient.kt
sschuberth: IIUC, per definition the code that wraps the service *is* the client, so I'd prefer to keep the shorter (and more commonly used) "client".

clients/bazel-module-registry/src/funTest/kotlin/BazelModuleRegistryClientFunTest.kt
```diff
@@ -25,7 +25,7 @@ import io.kotest.matchers.comparables.shouldBeGreaterThan
 import io.kotest.matchers.shouldBe
 
 class BazelModuleRegistryClientFunTest : WordSpec({
-    val client = BazelModuleRegistryClient.create()
```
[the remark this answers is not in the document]
sschuberth: > couldn't find any indication that the `Service` suffix is a convention.

It's used in their official example code as part of the introduction, which is good enough for me to confirm a convention.

Other than that, "Service" is also what's mostly used in [Retrofit's own code base](https://github.com/search?q=repo%3Asquare%2Fretrofit+%22interface+%22+language%3AKotlin&type=code). The only other suffix I see is "Api", which would be fine with me as well, but as most of our existing code already uses "Service", I've aligned on that.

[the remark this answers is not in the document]
fviernau: > It's used in their official example code as part of the introduction, which is good enough for me to confirm a convention.

For me a single use of a suffix in example code is not sufficient proof of a convention. However, would you be ok to change the rationale to align on "Service" for consistency within ORT's code base?

clients/osv/src/main/kotlin/OsvClient.kt
fviernau: > IIUC, per definition the code that wraps the service is the client,

I see, so what is the code that _uses_, not _wraps_ the service then?
To me it seems that client would better fit to the code using that wrapper. 

> so I'd prefer to keep the shorter (and more commonly used) "client".

ok, normally I don't care too much, in this case I would appreciate if the compromise was made towards my preference.

clients/osv/src/funTest/kotlin/OsvServiceFunTest.kt
```diff
@@ -66,21 +66,21 @@ class OsvServiceFunTest : StringSpec({
     "getVulnerabilitiesForPackage() returns the expected vulnerability when queried by commit" {
```
fviernau, line 66: OsvServiceFunTest -> OsvClientFunTest

fviernau (changes requested):

clients/bazel-module-registry/src/funTest/kotlin/BazelModuleRegistryClientFunTest.kt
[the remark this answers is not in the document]
sschuberth: > For me a single use of a suffix in example code is not sufficient proof of a convention.

I've also provided a link to multiple other examples in Retrofit's code base.

> However, would you be ok to change the rationale to align on "Service" for consistency within ORT's code base?

Sure, if that rationale works better for you.

clients/osv/src/main/kotlin/OsvClient.kt
sschuberth: I've checked the code base again, and OSV seems to currently be the only client library project that providers a higher-level "wrapper" around the low-level `Service` interface. In my thinking that "wrapper" is something between the service and the (advisor) client, so going for "ServiceWrapper" here is fine with me.

sschuberth: Or actually, thinking about it again, we could also keep `OsvServiceWrapper` at `OsvClient` to keep thing simple, as you prefer.

clients/bazel-module-registry/src/funTest/kotlin/BazelModuleRegistryClientFunTest.kt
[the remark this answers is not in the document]
fviernau: Thanks!

clients/osv/src/main/kotlin/OsvClient.kt
fviernau: I'd prefer to keep `OsvServiceWrapper` as it is right now.

fviernau (approved):

PR #8770, opened 2024-06-18
Add the Double Open Server (DOS) client and package configuration provider

sschuberth:
The DOS scanner wrapper will be contributed as a follow-up.

### DIFF
--- a/clients/dos/src/main/kotlin/DosClient.kt
+++ b/clients/dos/src/main/kotlin/DosClient.kt
@@ -1,0 +1,179 @@
+/*
+ * Copyright (C) 2024 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.clients.dos
+
+import java.io.File
+
+import okhttp3.MediaType.Companion.toMediaType
+import okhttp3.RequestBody.Companion.asRequestBody
+
+import org.apache.logging.log4j.kotlin.logger
+
+import org.ossreviewtoolkit.utils.common.formatSizeInMib
+
+/**
+ * A client implementation on top of the DOS service.
+ */
+class DosClient(private val service: DosService) {
+    /**
+     * Get the package configuration for the given [purl]. Return a [PackageConfigurationResponseBody] or null on error.
+     */
+    suspend fun getPackageConfiguration(purl: String): PackageConfigurationResponseBody? {
+        if (purl.isEmpty()) {
+            logger.error { "The Package URL is required to get the package configuration." }
+            return null
+        }
+
+        val requestBody = PackageConfigurationRequestBody(purl)
+        val response = service.getPackageConfiguration(requestBody)
+        val responseBody = response.body()
+
+        return if (response.isSuccessful && responseBody != null) {
+            responseBody
+        } else {
+            logger.error { "Error getting the package configuration for PURL $purl: ${response.errorBody()}" }
+
+            null
+        }
+    }
+
+    /**
+     * Return a pre-signed URL as provided by the DOS API to upload a package to S3 object storage for scanning, or null
+     * if the request did not succeed.
+     */
+    suspend fun getUploadUrl(key: String): String? {
+        if (key.isEmpty()) {
+            logger.error { "Need the name of the zipped packet to upload" }
+            return null
+        }
+
+        val requestBody = UploadUrlRequestBody(key)
+        val response = service.getUploadUrl(requestBody)
+        val responseBody = response.body()
+
+        return if (response.isSuccessful && responseBody != null && responseBody.success) {
+            response.body()?.presignedUrl
+        } else {
+            logger.error { "Unable to get a pre-signed URL for $key: ${response.errorBody()?.string()}" }
+            null
+        }
+    }
+
+    /**
+     * Upload the file at [file] to S3 using the [pre-signed URL][presignedUrl] and return true on success or false
+     * other on failure.
+     */
+    suspend fun uploadFile(file: File, presignedUrl: String): Boolean {
+        logger.info { "Uploading file $file of size ${file.formatSizeInMib} to S3..." }
+
+        val contentType = "application/zip".toMediaType()
+        val response = service.uploadFile(presignedUrl, file.asRequestBody(contentType))
+
+        return if (response.isSuccessful) {
+            logger.info { "Successfully uploaded $file to S3." }
+            true
+        } else {
+            logger.error { "Failed to upload $file to S3: ${response.errorBody()}" }
+            false
+        }
+    }
+
+    /**
+     * Add a new scan job [zipFileKey] and [purls]. Return a [JobResponseBody] or null on error.
+     */
+    suspend fun addScanJob(zipFileKey: String, purls: List<String>): JobResponseBody? {
+        if (zipFileKey.isEmpty() || purls.isEmpty()) {
+            logger.error { "The ZIP file key and Package URLs are required to add a scan job." }
+            return null
+        }
+
+        val requestBody = JobRequestBody(zipFileKey, purls)
+        val response = service.addScanJob(requestBody)
+        val responseBody = response.body()
+
+        return if (response.isSuccessful && responseBody != null) {
+            responseBody
+        } else {
+            logger.error { "Error adding a new scan job for $zipFileKey and $purls: ${response.errorBody()}" }
+
+            null
+        }
+    }
+
+    /**
+     * Get scan results for a list of [purls]. In case multiple purls are provided, it is assumed that they all refer to
+     * the same provenance (like a monorepo). If [fetchConcluded] is true, return concluded licenses instead of detected
+     * licenses. Return either existing results, a "pending" message if the package is currently being scanned, a
+     * "no-results" message if a scan yielded no results, or null on error.
+     */
+    suspend fun getScanResults(purls: List<String>, fetchConcluded: Boolean): ScanResultsResponseBody? {
+        if (purls.isEmpty()) {
+            logger.error { "The list of PURLs to get scan results for must not be empty." }
+            return null
+        }
+
+        val options = ScanResultsRequestBody.ReqOptions(fetchConcluded)
+        val requestBody = ScanResultsRequestBody(purls, options)
+        val response = service.getScanResults(requestBody)
+        val responseBody = response.body()
+
+        return if (response.isSuccessful && responseBody != null) {
+            when (responseBody.state.status) {
+                "no-results" -> logger.info { "No scan results found for $purls." }
+                "pending" -> logger.info { "Scan pending for $purls." }
+                "ready" -> {
+                    logger.info { "Scan results ready for $purls." }
+
+                    if (fetchConcluded) {
+                        logger.info { "Returning concluded licenses instead of detected licenses." }
+                    }
+                }
+            }
+
+            responseBody
+        } else {
+            logger.error { "Error getting scan results: ${response.errorBody()}" }
+
+            null
+        }
+    }
+
+    /**
+     * Get the state of the scan job with the given [id]. Return a [JobStateResponseBody] with one of waiting / active /
+     * completed / failed state, or null on error.
+     */
+    suspend fun getScanJobState(id: String): JobStateResponseBody? {
+        if (id.isEmpty()) {
+            logger.error { "Need the job ID to check for job state" }
+            return null
+        }
+
+        val response = service.getScanJobState(id)
+        val responseBody = response.body()
+
+        return if (response.isSuccessful && responseBody != null) {
+            responseBody
+        } else {
+            logger.error { "Error getting the scan state for job $id: ${response.errorBody()}" }
+
+            null
+        }
+    }
+}
--- a/clients/dos/src/main/kotlin/DosModel.kt
+++ b/clients/dos/src/main/kotlin/DosModel.kt
@@ -1,0 +1,164 @@
+/*
+ * Copyright (C) 2024 The ORT Project Authors (see <https://github.com/oss-review-toolkit/ort/blob/main/NOTICE>)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ * License-Filename: LICENSE
+ */
+
+package org.ossreviewtoolkit.clients.dos
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonObject
+
+import org.ossreviewtoolkit.model.config.PathExcludeReason
+
+/** The JSON (de-)serializer to use for the DOS model. */
+val JSON = Json { ignoreUnknownKeys = true }
+
+@Serializable
+data class UploadUrlRequestBody(
+    /** The desired key name for the file to upload to S3. */
+    val key: String
+)
+
+@Serializable
+data class UploadUrlResponseBody(
+    /** A flag to indicate whether requesting the upload URL was successful. */
+    val success: Boolean,
+
+    /** The pre-signed upload URL, or null on failure. */
+    val presignedUrl: String? = null,
+
+    /** The success or error message from the server, if any. */
+    val message: String? = null
+)
+
+@Serializable
+data class ScanResultsRequestBody(
+    /**
+     * The list of purls to get scan results for. In case multiple purls are provided, it is assumed that they all
+     * refer to the same provenance (like a monorepo).
+     */
+    val purls: List<String>? = null,
+
+    /** Options fort requesting scan results. */
+    val options: ReqOptions? = null
+) {
+    @Serializable
+    data class ReqOptions(
+        /** An option to fetch the concluded licenses along with the scan results. Defaults to "false". */
+        val fetchConcluded: Boolean? = false
+    )
+}
+
+@Serializable
+data class ScanResultsResponseBody(
+    /** The state of the scan job. */
+    val state: State,
+
+    /** The list of purls that was originally requested. */
+    val purls: List<String>? = null,
+
+    /** The raw result from the scanner. Currently only ScanCode is supported. */
+    val results: JsonObject? = null
+) {
+    @Serializable
+    data class State(
+        /** Named status of the scan job. One of: "no-results", "pending", "ready". */
+        val status: String,
+
+        /** The ID of the scan job. Only non-null of the job is "pending". */
+        val jobId: String? = null
+    )
+}
+
+@Serializable
+data class JobRequestBody(
+    /** The key of the previously uploaded ZIP file to scan. */
+    val zipFileKey: String,
+
+    /** The list of purls whose packages' source code is contained in the ZIP file. */
+    val purls: List<String>
+)
+
+@Serializable
+data class JobResponseBody(
+    /** The ID of the scan job that was created, or null on error. */
+    val scannerJobId: String? = null,
+
+    /** The success or error message from the server, if any. */
+    val message: String? = null
+)
+
+@Serializable
+data class JobStateResponseBody(
+    /** The state of the scan job. */
+    val state: State
+) {
+    @Serializable
+    data class State(
+        /**
+         * Named status of the scan job. One of: "created", "processing", "queued", "waiting", "active", "delayed",
+         * "paused", "stuck", "resumed", "stalled", "savingResults", "completed", "failed".
+         */
+        val status: String? = null,
+
+        /** The success or error message from the server, if any. */
+        val message: String? = null
+    )
+}
+
+@Serializable
+data class PackageConfigurationRequestBody(
+    /** The purl of the package to request configuraion for. */
+    val purl: String
+)
+
+@Serializable
+data class PackageConfigurationResponseBody(
+    /** The license conclusions for the file(s) in the package(s). */
+    val licenseConclusions: List<LicenseConclusion>,
+
+    /** The path exclusions for the file(s) in the package(s). */
+    val pathExclusions: List<PathExclusion>
+) {
+    @Serializable
+    data class LicenseConclusion(
+        /** The path to the file the conclusion belongs to. */
+        val path: String,
+
+        /** The SPDX expression for the detected license of the whole file. */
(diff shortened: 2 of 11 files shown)


utils/ort/src/test/kotlin/UtilsTest.kt
```diff
@@ -74,14 +74,15 @@ class UtilsTest : WordSpec({
...
+                "PROD_1_1_0",
                 "PROD_1_0_1",
                 "PROD_1_0_2",
                 "PROD_1_0_3",
                 "before_SourceForge",
                 "before_debug_changes"
             )
 
-            filterVersionNames("1.0.3", names) shouldHaveSingleElement "PROD_1_0_3"
+            filterVersionNames("1.1.0", names) shouldHaveSingleElement "PROD_1_1_0"
```
fviernau, line 85: unrelated / left-over?

fviernau (changes requested):

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+    val url: String,
+
+    /** The secret token to use with the DOS backend. */
+    val token: String,
+
+    /** The timeout for communicating with the DOS backend, in seconds. */
+    val timeout: Long?
+)
+
+open class DosPackageConfigurationProviderFactory :
```
fviernau, line 60: Can you help understanding why the `open` modifier is present?

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+            timeout = options["timeout"]?.toLongOrNull(),
+            token = secrets.getValue("token")
+        )
+}
+
+/**
+ * A [PackageConfigurationProvider] that loads [PackageConfiguration]s from a Double Open Server instance.
+ */
+class DosPackageConfigurationProvider(config: DosPackageConfigurationProviderConfig) : PackageConfigurationProvider {
+    private val service = DosService.create(
```
fviernau, line 78: Would it make sense to inline `service`, as all use is through `client` ?

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+    private val client = DosClient(service)
+
+    override fun getPackageConfigurations(packageId: Identifier, provenance: Provenance): List<PackageConfiguration> {
+        val purl = packageId.toPurl(provenance.toPurlExtras())
+
+        val packageResults = runBlocking { client.getPackageConfiguration(purl) }
+        if (packageResults == null) {
+            logger.info { "Unable to find package configuration for '$purl'." }
```
fviernau, line 91: Is this case a successful execution with an empty result, or an unsuccessful execution ?

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+
+            logger.info { "Found license conclusions for $purl: ${packageResults.licenseConclusions}" }
+            logger.info { "Found path exclusions for $purl: ${packageResults.pathExclusions}" }
+
+            listOf(packageConfiguration)
+        }
+    }
+}
+
+internal fun generatePackageConfiguration(
```
fviernau, line 112: I propose to turn this into `PackageConfigurationResponseBody.toPackageConfiguration(id, provenance)`.
This eliminates the word generate which IMO is slightly misleading, and also avoids the `packageResults` variable name.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+        )
+    } else {
+        null
+    }
+
+    val licenseFindingCurations = packageResults.licenseConclusions.map { licenseConclusion ->
+        val detected = licenseConclusion.detectedLicenseExpressionSPDX?.takeUnless { it.isEmpty() }
+            ?.let { SpdxExpression.parse(it) }
+
+        check(licenseConclusion.concludedLicenseExpressionSPDX.isNotEmpty())
```
fviernau, line 137: Would it make sense to filter empty ones, and log some info instead of failing hard?

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+        val concluded = SpdxExpression.parse(licenseConclusion.concludedLicenseExpressionSPDX)
+
+        LicenseFindingCuration(
+            path = licenseConclusion.path,
+            detectedLicense = detected?.let { associateLicensesWithExceptions(it) },
```
fviernau, line 142: I don't understand what / why associate with exception is done here.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+            )
+
+            logger.info { "Found license conclusions for $purl: ${packageResults.licenseConclusions}" }
+            logger.info { "Found path exclusions for $purl: ${packageResults.pathExclusions}" }
+
+            listOf(packageConfiguration)
+        }
+    }
+}
+
```
fviernau, line 111: Can below function be private?

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,164 @@
...
+            emptyList()
+        } else {
+            val packageConfiguration = generatePackageConfiguration(
+                id = packageId,
+                provenance = provenance,
+                packageResults = packageResults
+            )
+
+            logger.info { "Found license conclusions for $purl: ${packageResults.licenseConclusions}" }
+            logger.info { "Found path exclusions for $purl: ${packageResults.pathExclusions}" }
```
fviernau, line 105: Consider merging this into a single log statement.

fviernau (changes requested):

clients/dos/src/main/kotlin/DosService.kt
```diff
@@ -0,0 +1,122 @@
...
+
+            val authorizationInterceptor = AuthorizationInterceptor(token)
+
+            val okHttpClient = client ?: OkHttpClient.Builder()
+                .apply {
+                    if (timeout != null) {
+                        callTimeout(timeout)
+                        connectTimeout(timeout)
+                        readTimeout(timeout)
+                        writeTimeout(timeout)
```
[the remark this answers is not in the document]
sschuberth: I guess it's simply a convenient way to increase all timeouts in cause you ran into timeouts, and just want to get rid of the problem, not caring too much *which* timeout is the root cause. As @Etsija is on vacation, maybe @mmurto can share more historic details?

clients/dos/src/main/kotlin/DosService.kt
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Put the [file] into the S3 object storage at [url].
+     */
+    @SkipAuthorization
+    @PUT
+    suspend fun uploadFile(@Url url: String, @Body file: RequestBody): Response<Unit>
+
+    /**
+     * Add new scanner job for specified [purls][JobRequestBody.purls]. In case multiple PURLs are provided, it is
+     * assumed that they all refer to the same provenance (like a monorepo).
```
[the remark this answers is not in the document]
sschuberth: While personally I like all-caps "PURL" better, indeed https://github.com/package-url/purl-spec/ seems to use "purl" in free text, so I've aligned with that. I'll maybe do another PR to also align on that in existing code.

clients/dos/src/main/kotlin/DosService.kt
```diff
@@ -0,0 +1,122 @@
...
+import retrofit2.converter.kotlinx.serialization.asConverterFactory
+import retrofit2.http.Body
+import retrofit2.http.GET
+import retrofit2.http.POST
+import retrofit2.http.PUT
+import retrofit2.http.Path
+import retrofit2.http.Url
+
+/**
+ * The service for the REST API client of the Double Open Server (DOS), see https://github.com/doubleopen-project/dos/.
```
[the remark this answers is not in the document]
sschuberth: The API link is on that linked project page, i.e. https://api.doubleopen.io/docs/. As I did not want to mention both links, I went for only the more generic one.

clients/dos/src/main/kotlin/DosModel.kt
```diff
@@ -0,0 +1,115 @@
...
+    val licenseConclusions: List<LicenseConclusion>,
+    val pathExclusions: List<PathExclusion>
+) {
+    @Serializable
+    data class LicenseConclusion(
+        val path: String,
+        val detectedLicenseExpressionSPDX: String? = null,
+        val concludedLicenseExpressionSPDX: String,
+        val comment: String? = null
+    )
```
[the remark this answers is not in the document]
sschuberth: @lamppu or @mmurto can you assist here?

clients/dos/src/main/kotlin/DosModel.kt
```diff
@@ -0,0 +1,115 @@
...
+package org.ossreviewtoolkit.clients.dos
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonObject
+
+/**
+ * The JSON (de-)serialization object used by this service.
+ */
+val JSON = Json { ignoreUnknownKeys = true }
```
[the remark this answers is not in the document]
sschuberth: As the way the JSON serializer is configured is specific to the model (think e.g. about a configured naming strategy), and the model is / should be public, as this is a generic API client, I prefer the JSON serializer to remain public, too.

clients/dos/src/main/kotlin/DosModel.kt
```diff
@@ -0,0 +1,115 @@
...
+ */
+
+package org.ossreviewtoolkit.clients.dos
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonObject
+
+/**
+ * The JSON (de-)serialization object used by this service.
```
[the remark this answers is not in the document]
sschuberth: I've used a similar wording now.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
sschuberth: Good catch, it's not needed and I've removed it.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
sschuberth: As that would result in more nesting, I'd prefer to keep it as-is in this case.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
sschuberth: It's the latter. As the current log message is misleading, and the `client.getPackageConfiguration(purl)` call already logs, I've simply removed the log completely here.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
sschuberth: This `check` (compared to `require`) is supposed to document in code that this case is never supposed to happen. But if it indeed does, we should indeed fail hard.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
sschuberth: The DOS scanner uses ScanCode on the server-side, so it suffers from the same shortcomings in ScanCode's results as when running locally. This means we also need to fixup stand-alone exception findings here.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
sschuberth: The reason why I didn't was that I wanted to get separate log prefixes per line.

clients/dos/src/main/kotlin/DosModel.kt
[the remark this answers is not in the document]
mmurto: I'm not certain what this is.

clients/dos/src/main/kotlin/DosService.kt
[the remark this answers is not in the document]
mmurto: Unfortunately no historic details here. My guess is the same as Sebastian's.

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
fviernau: Hm, but these are license finding curations, not license findings. 
Shouldn't the curations target the actual detected licenses (which has associated exceptions)? (Instead of matching against the unpatched detected licenses which the curator normally does not know about)

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
```diff
@@ -0,0 +1,156 @@
...
+        config.timeout?.let { Duration.ofSeconds(it) }
+    )
+
+    private val client = DosClient(service)
+
+    override fun getPackageConfigurations(packageId: Identifier, provenance: Provenance): List<PackageConfiguration> {
+        val purl = packageId.toPurl(provenance.toPurlExtras())
+
+        val packageResults = runBlocking { client.getPackageConfiguration(purl) }
+        if (packageResults == null) return emptyList()
```
fviernau, line 90: line 90 can be added to line 89 with `?: return emptyList()`

fviernau: Maybe even like this, to eliminate the `if / else` below.
```
val packageResults = runBlocking { client.getPackageConfiguration(purl) }
    ?.takeUnless { it.licenseConclusions.isEmpty() && it.pathExclusions.isEmpty() } ?: return emptyList()
```

plugins/package-configuration-providers/dos/src/main/kotlin/DosPackageConfigurationProvider.kt
sschuberth: > Shouldn't the curations target the actual detected licenses (which has associated exceptions)?

Exactly that's the reason for calling `associateLicensesWithExceptions()` also here: Remember that this is the *matcher* for the detected license that ORT already "knows", and that known detected license was post-processed with `associateLicensesWithExceptions()` previously. So in order to match that post-processed detected license, we need to apply the same post-processing here as well.

fviernau: Just curious, the license finding curations in ORT we write against the "post processed" findings. In particular, we do not run `associateLicensesWithExceptions()`. Why is this different for DOS and why don't these curations get fixed on the backend / in the database ?

sschuberth: > Just curious, the license finding curations in ORT we write against the "post processed" findings.

I agree with this, but...

> In particular, we do not run `associateLicensesWithExceptions()`.

... this is IMO a contradiction to that. Post-processing includes running `associateLicensesWithExceptions()`, and we do that at

https://github.com/oss-review-toolkit/ort/blob/f995050ff24808ffc95d77772963ca19d15f8141/plugins/scanners/scancode/src/main/kotlin/ScanCodeResultModelMapper.kt#L154

So any detected license returned by ScanCode got post-processed by `associateLicensesWithExceptions()`. Hence also everything that we match against this license needs to be be post-processed by `associateLicensesWithExceptions()`.

fviernau: > So any detected license returned by ScanCode got post-processed by `associateLicensesWithExceptions()`. Hence also everything that we match against this license needs to be be post-processed by `associateLicensesWithExceptions()`.

This bit I still do not understand. What I do understand is that in ORT we call the postprocessing in the scanner, but not for the license finding curations. Because the license finding curation are written against the post processed licenses, not against the raw ones. In DOS this aspect (for the license finding curations, not scanner) seems to be different. Why?

sschuberth: I'm still not 100% certain about your question, but I guess it can be rephrased as "Why does DOS here need to call `associateLicensesWithExceptions()`, but the other `PackageConfigurationProvider`s don't?".

The answer to that question is bit difficult to give without knowing the DOS scanner implementation (which I probably should have contributed first, but I wanted to do the "easier" `PackageConfigurationProvider` contribution first, as it's less code):

The DOS scanner backend is written in TypeScript, not in Koltin, and thus cannot make use of any ORT functionality (that was a design decision that probably needs to be revisited in the context of ORT server, but that's another topic). This means that `associateLicensesWithExceptions()` is not available to the DOS scanner, so it can't post-process the ScanCode results with it. So we "postpone" the post-processing that would usually be done by the `ScanCodeResultModelMapper`, which the DOS scanner cannot use, to here.

fviernau: This comment clarifies what I've been after. If this information is added as a code comment, I'd be ok with keeping the `associate*` call.

fviernau: Thanks!

clients/dos/src/main/kotlin/DosModel.kt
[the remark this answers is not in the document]
sschuberth: I did look at the DOS code, and as now also mentioned [in my added code comment](https://github.com/oss-review-toolkit/ort/compare/7e641ee5b689eecee1d939e53292fdfdc80c8226..a686d5314f9ac24883809c2589bed5085c4dd599#diff-654c73eb0d695d1a0f435c8e015b9fa1bb44d8d9b820c607b9581f5b88c42ef8R122), this property holds the raw data as reported by ScanCode.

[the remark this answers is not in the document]
fviernau: In ORT, the semantics of the property is: It selects matching license LicenseFindings with equal detected license, whereas a LicenseFinding is defined as (path, license, start_line, end_line).  

How is it in DOS?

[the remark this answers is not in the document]
sschuberth: I'm again unsure if I get the question correctly. What confuses me is: The *matching* of a license finding curation is always done by ORT core, no matter who the provider of the package configuration was. So how could it be different for DOS if it just provides the package curation, but the matching of a contained license finding curation is done by ORT core?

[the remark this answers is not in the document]
fviernau: If the semantics of a license finding curation is identical with the one of ORT it could not be different. 
However, I'm trying to understand the semantics of a license finding curation in DOS. I've asked about this several times now, but it remains unanswered. 

The data model is not identical, as it lacks start / end lines, which is why I'm asking this question. It's not obvious at all to me, and documentation pointers I don't have any either.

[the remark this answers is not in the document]
sschuberth: > I've asked about this several times now, but it remains unanswered.

That's not out of ill intent, but simply because I don't understand your questions, and what exactly you're after. Asking for the "semantic of DOS license finding curations" is such a broad question that I have no idea how to answer it pointedly.

So again I'm *guessing* that the concrete question is: Why does DOS not return any `startLines` / `lineCount` as part of a `LicenseFindingCuration`. This is something I don't know and need to check with the team, ping @lamppu.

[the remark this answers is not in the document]
fviernau: > That's not out of ill intent,

I didn't mean to say it was out of ill intent. 

> Asking for the "semantic of DOS license finding curations" is such a broad question

Hm, I didn't anticipate that this could have been very brought. However, let me try to further narrow down:

Is that detected license property supposed to be matched against

1. Single License findings in ORT `(path, detected,license, start_line, end_line)` or
2. Detected license of the entire file

I do recall some feature flag, which has been introduced not too long ago, which enables some combining to license per file. But I cannot find that commit. However, IIRC it was supposed to be used in DOS context.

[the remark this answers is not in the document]
lamppu: Yes, so the workflow that we have decided on is that we make curations for the whole file, instead of for individual line matches, so my understanding was that the line numbers could be omitted here. The `detectedLicenseExpressionSPDX` is actually the file-level scanner finding detected by Scancode, as that is what we actually return in our scan results, using the first and last lines of the individual matches as the line numbers. We do save the individual matches in our database, so that curators can inspect these in our UI, but we only return the file-level one to ORT.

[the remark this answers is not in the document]
fviernau: Thanks @lamppu ! 

So, this means that as-is, ORT would match the license finding curations against single license findings, while it would be required to match them against the license of the entire file. This seems to be a mismatch. Any idea how this could be best addressed, maybe @sschuberth? (Does this curation provider actually require to turn on the feature to make ORT operate on file level licenses instead of single licenses, @sschuberth do you recall your commit for this?)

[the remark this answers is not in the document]
sschuberth: After having another clarifying discussion with @lamppu, the answer is that using the DOS package configuration provider only makes sense when enabling the [`preferFileLicense` configuration](https://github.com/oss-review-toolkit/ort/blob/59681807ea5c4ab9f468e783f2bdb7fa1de5dd30/plugins/scanners/scancode/src/main/kotlin/ScanCode.kt#L59-L65) for ScanCode, or in conjunction with the DOS scanner wrapper. I'll add a `README.md` that documents this.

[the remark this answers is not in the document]
sschuberth: Done. Please have another look @fviernau.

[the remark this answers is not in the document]
fviernau: Thanks!

clients/dos/src/main/kotlin/DosClient.kt
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    /**
+     * Return a pre-signed URL as provided by the DOS API to upload a package to S3 object storage for scanning, or null
+     * if the request did not succeed.
+     */
+    suspend fun getUploadUrl(key: String): String? {
+        if (key.isEmpty()) {
+            logger.error { "Need the name of the zipped packet to upload" }
```
fviernau, line 63: Is `packet` correct? 

Missing trailing dot.

clients/dos/src/main/kotlin/DosClient.kt
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    /**
+     * Get the state of the scan job with the given [id]. Return a [JobStateResponseBody] with one of waiting / active /
+     * completed / failed state, or null on error.
+     */
+    suspend fun getScanJobState(id: String): JobStateResponseBody? {
+        if (id.isEmpty()) {
+            logger.error { "Need the job ID to check for job state" }
```
fviernau, line 164: Missing trailing dot.

fviernau (approved):

clients/dos/src/main/kotlin/DosClient.kt
sschuberth: I'll fix these in a follow-up, thanks.

sschuberth: See https://github.com/oss-review-toolkit/ort/pull/8789.

PR #8773, opened 2024-06-19
fix(fossid-webapp): Generate ignore rules also for non-delta scans

wkl3nk:
Generate ignore rules from path excludes in _.ort.yml_ also if FossID delta scans are disabled.

### DIFF
--- a/plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
+++ b/plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
@@ -935,6 +935,46 @@ class FossIdTest : WordSpec({
             }
         }
 
+        "apply exclusion rules to a non-delta scan" {
+            val projectCode = projectCode(PROJECT)
+            val scanCode = scanCode(PROJECT, null)
+            val config = createConfig(deltaScans = false)
+            val vcsInfo = createVcsInfo()
+            val scan = createScan(vcsInfo.url, "${vcsInfo.revision}_other", scanCode)
+
+            val service = FossIdRestService.create(config.serverUrl)
+                .expectProjectRequest(projectCode)
+                .expectListScans(projectCode, listOf(scan))
+                .expectCheckScanStatus(scanCode, ScanStatus.NEW, ScanStatus.FINISHED)
+                .expectCreateScan(projectCode, scanCode, vcsInfo, "")
+                .expectDownload(scanCode)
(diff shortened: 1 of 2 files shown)


plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
```diff
@@ -935,6 +935,45 @@ class FossIdTest : WordSpec({
             }
         }
 
+        "apply exclusion rules to a origin scan" {
+            val projectCode = projectCode(PROJECT)
+            val scanCode = scanCode(PROJECT, FossId.DeltaTag.ORIGIN)
+            val config = createConfig()
```
nnobelis, line 941: Your PR is adding the logic to convert rules in the case when this is NOT a delta scope.
Therefore you need a `createConfig(deltaScans= false)`.
You can test with the debugger that you are in the right branch of the code.

plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
```diff
@@ -935,6 +935,45 @@ class FossIdTest : WordSpec({
             }
         }
 
+        "apply exclusion rules to a origin scan" {
+            val projectCode = projectCode(PROJECT)
+            val scanCode = scanCode(PROJECT, FossId.DeltaTag.ORIGIN)
```
nnobelis, line 940: You need a scancode without the delta part (no really critical but consistency).

plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
```diff
@@ -935,6 +935,45 @@ class FossIdTest : WordSpec({
             }
         }
 
+        "apply exclusion rules to a origin scan" {
```
nnobelis, line 938: Origin scans are only for delta scans.

How about "apply exclusion rules to a non-delta scan"

plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
wkl3nk: Thanks for the hint. Using the debugger I am now sure that the desired code is covered.

plugins/scanners/fossid/src/test/kotlin/FossIdTest.kt
wkl3nk: Had a look at other non-delta scans and modified the code in the same way.

nnobelis (approved):

nnobelis (approved):

mnonnenmacher (approved):

PR #8778, opened 2024-06-20
FossId refactorings

mnonnenmacher:

### DIFF
--- a/plugins/scanners/fossid/src/main/kotlin/FossId.kt
+++ b/plugins/scanners/fossid/src/main/kotlin/FossId.kt
@@ -63,6 +63,7 @@ import org.ossreviewtoolkit.clients.fossid.model.Project
 import org.ossreviewtoolkit.clients.fossid.model.Scan
 import org.ossreviewtoolkit.clients.fossid.model.result.MatchType
 import org.ossreviewtoolkit.clients.fossid.model.result.MatchedLines
+import org.ossreviewtoolkit.clients.fossid.model.rules.IgnoreRule
 import org.ossreviewtoolkit.clients.fossid.model.rules.RuleScope
 import org.ossreviewtoolkit.clients.fossid.model.rules.RuleType
 import org.ossreviewtoolkit.clients.fossid.model.status.DownloadStatus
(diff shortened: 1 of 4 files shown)


plugins/scanners/fossid/src/main/kotlin/FossId.kt
```diff
@@ -671,6 +628,34 @@ class FossId internal constructor(
...
+            convertRules(it, issues).also {
+                logger.info { "${it.size} rules from ORT excludes have been found." }
+            }
+        }.orEmpty()
+
+        // Create an issue for each legacy rule.
+        val legacyRules = excludesRules.filterLegacyRules(existingRules, issues)
+        if (legacyRules.isNotEmpty()) {
+            logger.warn { "${legacyRules.size} legacy rules have been found." }
+        }
```
sschuberth: Could you briefly describe in the commit message how you are ensuring that this code is only executed for the originally third code block? It seems that the "trick" is that `existingRules` is left empty in the other cases, which results in empty `legacyRules`.

I guess what's confusing me here is the way `filterLegacyRules()` works. from its name, I would have guessed that it returns a result based on its receiver `excludesRules`, i.e. that it filters `excludesRules`, whereas in fact it filters the `existingRules` argument.

plugins/scanners/fossid/src/main/kotlin/Utils.kt
```diff
@@ -84,15 +84,16 @@ private fun PathExclude.mapToRule(): IgnoreRule? {
...
+ * add it to [issues].
  */
 internal fun List<IgnoreRule>.filterLegacyRules(
-    rulesToTest: List<IgnoreRule>,
+    referenceRules: List<IgnoreRule>,
     issues: MutableList<Issue>
 ): List<IgnoreRule> =
-    rulesToTest.filterNot { ruleToTest ->
-        any { it.value == ruleToTest.value && it.type == ruleToTest.type }
+    filterNot { rule ->
```
sschuberth: Thank you! Much better now!

sschuberth (approved):

plugins/scanners/fossid/src/main/kotlin/FossId.kt
sschuberth: Ok, this more or less became obsolete with [this later change of yours](https://github.com/oss-review-toolkit/ort/pull/8778#discussion_r1648461079)!

fviernau (approved):

nnobelis (approved):

PR #8781, opened 2024-06-21
chore(package-managers): Do not log all Gradle `stderr` output as warnings

sschuberth:
Not everything that Gradle outputs to `stderr` is a warning in the sense that results are wrong or something needs to be done. Avoid some potential confusion by only logging `stderr` output at `debug` level, which aligns with logging of `stdout` output.


sschuberth:
FYI @wkl3nk.

fviernau (approved):

